PR #23, opened 2018-12-25
[WIP] Bug 1515140 - set up tc-builder

djmitche:
A few of these correspond to PRs in the source repos.  The first commit ("REDO") will be better accomplished with a re-import of the repo, I think (that's still planned, right?).

The support is pretty hacked-in.  It's slow and makes huge images.  The idea is to make minimal changes, and we can follow up once the monorepo is the only repo.


djmitche:
@imbstack what happened to close this?

djmitche:
Looks like this got closed when upstream changed with no common ancestor.

PR #28, opened 2019-01-04
Update lint

imbstack:
This should get us to a nice place. I've wanted 90% of these things for a long time. We have a lot of rules that I want to enable but will just take too much work for now to do. Let's do them later (will file bug) if we decide to enforce them. They're in one of the sections under `rules`.

This also gets rid of `eslint-config-taskcluster`. @owlishDeveloper was right, it wasn't getting us very much now. Many of these rules are copied from there to here.

Please feel free to ask me to remove or add any rules below. I've taken a stricter view of things like spacing than we have in the past.

Each of these commits should be relatively self contained and make it possible to actually review this. Obviously some can be skipped since they touch 100s of files.


libraries/pulse/src/consumer.js
```diff
@@ -164,7 +164,7 @@ class PulseConsumer {
           // channel is probably sick; but if this is an IllegalOperationError,
           // there's no need to report it (that is basically saying the channel
           // has closed, so we'll re-connect)
-          if (!err instanceof amqplib.IllegalOperationError) {
+          if (!(err instanceof amqplib.IllegalOperationError)) {
```
djmitche: that might explain some things in Sentry :)

libraries/pulse/test/publisher_test.js
```diff
@@ -206,7 +206,7 @@ suite('publisher_test.js', function() {
         folder: path.join(__dirname, 'schemas'),
       });
 
-      publisher = await exchanges.publisher({
```
djmitche: Is this a false-positive for the rule?  My guess is, it's seeing `publisher` twice here and assuming it's the same variable?  Also, it's an ineffective assignment (`publisher` is never used), so removing the assignment might be a better fix.

libraries/testing/src/pulse.js
```diff
@@ -47,7 +47,7 @@ var PulseTestReceiver = function(credentials, mocha) {
     // the PulseConnection closes eventually... But it's nice to keep things
     // clean, errors are more likely to surface at the right test this way.
     return Promise.all(that._listeners.map(function(listener) {
-      listener.close();
+      return listener.close();
```
djmitche: Wow, lots of bugs I introduced in this library..

.eslintrc.yml
```diff
@@ -43,10 +43,13 @@ rules:
   complexity:
     - error
     - max: 60 # We may want to eventually bring this number down
+  no-multi-spaces: error
+  key-spacing:
+    - error
+    - mode: strict
```
djmitche: RIP @jonasfj's manual alignment in vim..

owlishDeveloper:
The taskgroup looks very green to me in Tools...

imbstack:
I reran a task there to make it look more green :)

@jhford I'm going to land this now to avoid it bitrotting since it is a pain to keep up to date. Please let me know if there's anything about this you don't like! Should be quite easy to change back and I'm not super locked down on anything in here.

PR #29, opened 2019-01-04
Bug 1517732 - Do not specify route "statuses" twice

JohanLorenzo:
Fixes https://github.com/mozilla-mobile/android-components/pull/1650#commitcomment-31840040

### DIFF
--- a/services/github/src/tc-yaml.js
+++ b/services/github/src/tc-yaml.js
@@ -101,8 +101,10 @@ class VersionZero extends TcYaml {
   }
   compileTasks(config, cfg, payload, now) {
     config.tasks = config.tasks.map((task) => {
-      task.routes = task.routes || [];
-      task.routes.push(cfg.app.statusTaskRoute);
+      task.routes = _.union(
+        [cfg.app.statusTaskRoute],
+        task.routes,
+      );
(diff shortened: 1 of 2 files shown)


services/github/src/tc-yaml.js
```diff
@@ -101,8 +101,10 @@ class VersionZero extends TcYaml {
   }
   compileTasks(config, cfg, payload, now) {
     config.tasks = config.tasks.map((task) => {
-      task.routes = task.routes || [];
-      task.routes.push(cfg.app.statusTaskRoute);
+      task.routes = _.union(
```
owlishDeveloper: Thank you for the fix, @JohanLorenzo! This looks very good; I'm running tests locally - everything seems to be passing! The only nit pick is - using lodash is not entirely necessary (I'm trying to get rid of this dependency because it mostly just duplicates what JS can do on its own). `Array.from(new Set([...a, ...b]))` would do the trick here.

JohanLorenzo:
Thank you very much for the super quick review and the suggestion, @owlishDeveloper ! TIL a new JS idiom :smiley: 

I just addressed them. Please let me know what you think.

PR #31, opened 2019-01-04
Adds listRoles2 api with pagination

arshadkazmi42:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1517494


arshadkazmi42:
@djmitche We don't have travis integrated to this yet? I am not able to test it in my local, so thought travis here will verify it.

djmitche:
Sorry, that's https://bugzilla.mozilla.org/show_bug.cgi?id=1517754

I'll run the tests locally.

djmitche:
Please remove the `libraries/client` change.  We'll be fixing things so that client generation is automatic, but for the moment that change is unnecessary.

Also, were you unable to run the tests yourself?  What issues did you see?  You can run tests just for auth with `yarn --workspace services/auth test`.

arshadkazmi42:
> Please remove the libraries/client change. We'll be fixing things so that client generation is automatic, but for the moment that change is unnecessary.

I will do these changes now

> Also, were you unable to run the tests yourself? What issues did you see? You can run tests just for auth with yarn --workspace services/auth test.

Call to `listRoles2` was returning 500.

djmitche:
ah -- if you set DEBUG=* you should see what's going on there.

arshadkazmi42:
Running this command `yarn --workspace services/auth/ test` is running all services tests, and logs for specific auth tests are getting trimmed in terminal

services/auth/test/role_test.js
```diff
@@ -192,6 +192,62 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
...
+        description: 'test role',
+        scopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'],
+      });
+    }
+
+    let result = await helper.apiClient.listRoles2();
+
+    assert(result.roles.some(role => role.roleId === `thing-id:${clientId}`));
+    assert(result.roles.some(role => role.description === 'test role'));
+    assert(result.roles.some(role => role.scopes === ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2']));
```
djmitche: you'll need a deep comparison here, like `_.isEqual`

arshadkazmi42:
@djmitche Just review reminder, if you might have missed this.
Also, I am still not able to test it with this command `yarn --workspace services/auth test` its running all the tests

libraries/client/src/apis.js
```diff
@@ -180,6 +180,22 @@ module.exports = {
...
+          "output": "v1/list-roles2-response.json#",
+          "query": [
+            "continuationToken",
+            "limit"
+          ],
+          "route": "/roles2/",
+          "stability": "stable",
+          "title": "List Roles",
+          "type": "function"
+        },
```
djmitche: Let's remove the change to this file.  It will get regenerated eventually, but at the moment it differs from the API definition (in `title` at least)

arshadkazmi42:
@djmitche Which command to use for testing? `yarn workspace taskcluster-auth test` this is also running all the tests.

djmitche:
With a new enough yarn (just install the latest) that should just run auth tests.  But auth has a *lot* of tests!  If it starts with `azure_test.js (real)` then it's doing the right thing.

arshadkazmi42:
@djmitche I got the tests running.
Getting this error
` data.roles[0] should have required property 'expandedScopes'`
And when I am passing `expandedScopes` in createRole api, its giving error `expandScopes` is not allowed.
Its for this test
https://github.com/taskcluster/taskcluster/pull/31/files#diff-e3b008c569902b4f662445817d135a8fR202

djmitche:
Yep, if you look at [the API definition](https://docs.taskcluster.net/docs/reference/platform/taskcluster-auth/references/api#createRole), createRole does not take expandedScopes -- that value is calculated after the role is created.

arshadkazmi42:
In reponse yml I have added this
https://github.com/taskcluster/taskcluster/pull/31/files#diff-e5aa7653ec93596045d8dec431ee5391R14

In `get-role-response` expandedScopes is a mandatory field.
How can I get it to return exapndedScopes on createRole api call? As for its response, validator is expecting expandedScopes. Or the response yml needs to be changed?

djmitche:
Maybe I'm misunderstanding why this is an issue.. but when you call createRole, only pass `scopes`.  When you list roles or get a single role, or create a role, you'll get back a result with both `scopes` and `expandedScopes`.  This little change fixes the problem with createRole:

```diff
diff --git a/services/auth/test/role_test.js b/services/auth/test/role_test.js
index 2aa26100d..73d6e5fa3 100644
--- a/services/auth/test/role_test.js
+++ b/services/auth/test/role_test.js
@@ -194,23 +194,23 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
 
   test('listRoles2', async () => {
     // Clear existing roles
     await helper.Roles.modify((roles) => roles.splice(0));
 
     // Create 4 dummy roles
     await helper.apiClient.createRole(`thing-id:${clientId}`, {
       description: 'test role',
-      expandedScopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'],
+      scopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'],
     });
     for (let i=0;i<3;i++) {
       let tempRoleId = `${clientId}${i}`;
       await helper.apiClient.createRole(tempRoleId, {
         description: 'test role',
-        expandedScopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'],
+        scopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'],
       });
     }
 
     let result = await helper.apiClient.listRoles2();
 
     assert(result.roles.some(role => role.roleId === `thing-id:${clientId}`));
     assert(result.roles.some(role => role.description === 'test role'));
     assert(result.roles.some(role => _.isEqual(role.scopes, ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'])));
```

then you'll see an error about a *missing* expandedScopes in the response from listRoles2.  But the response from listRoles doesn't have this problem, because it includes expandedScopes.  The difference is in the function `roleToJson`.  Hopefully that's enough get you going in the right direction :)

arshadkazmi42:
`roleToJson` was the missing part. got it working. tests passed

djmitche:
Deployed, and seems to be working fine.  Nice work!!

arshadkazmi42:
> This looks good to me! As a quick followup, once this lands, you could refactor the old listRoles method to use the helper, too.

What I should for this change, which you suggested earlier for the old listRoles?

djmitche:
Actually now that I look again, that's probably not worthwhile -- the "old" listRoles method is only a few lines long, so there's no pressing need to refactor it.  Forget I said anything :)

PR #33, opened 2019-01-04
Bug 1516463 - use minItems, not minLength, for array length

djmitche:
(ported from https://github.com/taskcluster/taskcluster-queue/pull/309)


imbstack:
I reran the tests and they pass.

PR #49, opened 2019-01-04
Bug 1260285 - add a metaschema that enforces some basic sanity

djmitche:
This factors out a new '/schemas/common/metaschema.json#' that enforces
some basic things on our schemas.  This is strictly *more* restrictive
than vanilla draft-06 json-schema, so a compliant schema remains usable
by any other schema-handling tools.

(ported from https://github.com/taskcluster/taskcluster-lib-references/pull/6)


djmitche:
See #91.

PR #50, opened 2019-01-04
Use hasOwnProperty as it is faster and more sane

djmitche:
This was r+'d in https://github.com/taskcluster/taskcluster-lib-scopes/pull/26 -- should we just land it?


owlishDeveloper:
What's with the failing tests?

imbstack:
It passed on a retry so who knows! We're definitely still getting flakes.

djmitche:
Please add flaky failures to https://bugzilla.mozilla.org/show_bug.cgi?id=1517662

PR #52, opened 2019-01-05
Symlink local dependencies rather than copying

imbstack:
I'm not exactly sure why the tweaks were needed, but it seems reasonable to do anyway.


services/login/src/main.js
```diff
@@ -1,4 +1,4 @@
-const config = require('taskcluster-lib-config');
```
djmitche: wow..

imbstack:
The only thing worrying me about this links stuff is that `yarn add`ing new stuff occasionally seems to break things. In addition the tweaks I made here to make it work seem like they shouldn't be needed. I'm going to try to dig a bit deeper before I land this.

imbstack:
Ok, I've tried for 6 hours and all I can come up with is that under certain circumstances (which I cannot figure out), yarn makes two entries in `yarn.lock` for a single package unnecessarily and then later trips over itself [here](https://github.com/yarnpkg/yarn/blob/master/src/package-linker.js#L490). You can get around the issue by specifying `--no-bin-links` to `yarn install` or by deleting `yarn.lock` and remaking it. 

You can also fix it by making sure each version of the package required in every `package.json` is identical.

I'm landing this for now and we can see how often this bites us.

PR #54, opened 2019-01-07
Bug 1518137 - better libraries/api error handling, fix InsufficientScopes

djmitche:
The important thing here is not losing track of the message template, so
users still see the explanation of their InsufficientScopes error.  But
this also factors out the translation of AuthorizationError ->
InsufficientScopes and AuthenticationError -> AuthenticationFailed.

### DIFF
--- a/libraries/api/src/middleware/auth.js
+++ b/libraries/api/src/middleware/auth.js
@@ -97,7 +97,7 @@ const authLog = process.env.NODE_ENV === 'production' ?
  * client satisfies the scope expression in `options.scopes` after the
  * parameters denoted by `<...>` and `{for: ..., each: ..., in: ...}` are
  * substituted in. If the client does not satisfy the scope expression, it
- * throws an Error with code = 'AuthorizationError'.
+ * throws an Error.
  *
  * The `req.scopes()` method returns a Promise for the set of scopes the caller
  * has. Please, note that `req.scopes()` returns `[]` if there was an
@@ -244,7 +244,7 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
        * Create method to check if request satisfies the scope expression. Given
        * extra parameters.
        * Return true, if successful and if unsuccessful it throws an Error with
-       * code = 'AuthorizationError'.
+       * code = 'AuthenticationFailed'.
        */
       req.authorize = async (params) => {
         // This lint can be disabled because authenticate() will always return the same value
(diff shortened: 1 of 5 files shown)


libraries/api/src/middleware/auth.js
```diff
@@ -253,12 +253,11 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
...
-          const err = new Error('Authentication failed'); // This way instead of subclassing due to babel/babel#3083
-          err.name = 'AuthenticationError';
-          err.code = 'AuthenticationError';
-          err.message = result.message;
-          err.details = result;
-          throw err;
+          throw new ErrorReply({
+            code: 'AuthenticationFailed',
+            message: result.message,
+            details: result,
```
owlishDeveloper: So that `result` object - does it have the information about which scopes are missing anywhere in it? Should we be showing that information at all in the authorization case?

libraries/api/src/middleware/auth.js
```diff
@@ -319,11 +317,6 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
         next();
       }
     } catch (err) {
-      if (err.code === 'AuthorizationError') {
-        return next(new ErrorReply({code: 'InsufficientScopes', message: err.messageTemplate, details: err.details}));
```
owlishDeveloper: I'm quite puzzled. The error that Simon reported lacked `details` field completely (or at least it wasn't showing up in the logs). Why would that be?

libraries/api/src/middleware/auth.js
djmitche: This error is for bad Hawk credentials, not bad scopes - so no, not in this case.

libraries/api/src/middleware/auth.js
djmitche: The original bug was, indeed, right here -- we passed err.messageTemplate (which is undefined) instead of err.message (which contains the message simon expected).  The minimal fix to that bug was just messageTemplate -> message, but I took the chance to clean up some messiness in handling errors while I was at it.

djmitche: Yeah, it was a bit of leftovers from the refactor that @arshadkazmi42 worked on.

libraries/api/src/middleware/auth.js
```diff
@@ -253,12 +253,11 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
         // If authentication failed
         if (result.status === 'auth-failed') {
           res.set('www-authenticate', 'hawk');
-          const err = new Error('Authentication failed'); // This way instead of subclassing due to babel/babel#3083
```
imbstack: Bye bye babel.

djmitche:
deploying now

PR #59, opened 2019-01-07
Switch to Node 10.14

owlishDeveloper:
Later version of Node (`10.14.2` and up) could break tests in tc-lib-references (there was a change in Node 10.14.2 for which `fs-mock` library we use to mock the file system needs a patch).

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -84,7 +84,7 @@ tasks:
                 then: true
               TEST_BUCKET: 'test-bucket-for-any-garbage' # TODO: Allow each project to specify its own env
             maxRunTime: 3600
-            image: "node:8"
+            image: "node:10.14.0"
             command:
               - "/bin/bash"
               - "-lc"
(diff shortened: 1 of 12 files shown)


owlishDeveloper:
Oh, I forgot the test. Let me just add it in here

owlishDeveloper:
A later thought - I'll file a bug

owlishDeveloper:
As I was creating a bug - don't we have a metatest already? The way our `.taskcluster.yml` is structured, we have one Node for every package. If you simply change Node version in `package.json`, the build will fail. If you also change `taskcluster.yml` - that will be obvious in the PR, and it would be easy to pull the plug on the change?

owlishDeveloper:
We have to either wait for `fs-mock` to be patched up, or to patch it ourselves. Currently that library is what keeping us from using the latest version of Node 10. Other than that - keeping up with the latest releases of Node 10 should be easy!

> it's based on engineStrict being set to true and a different version of Node being specified. Should we add a very explicit test that ensures all versions are specified with equal values?

I would argue there's no need for that, because in order for the tests to pass you'd need to make changes to `package.json` and to `.taskcluster.yml`... Those are hard to miss in a PR. Unless there's a concern we'll forget we wanted all services to run the same Node version and just approve the PR?

PR #60, opened 2019-01-08
Switch to Auth0 library for correct runtime env

owlishDeveloper:
We used to use an Auth0 library for headless browser 🕸🎃🕷👻, latest version of which is completely broken in Node. This PR is to switch to the Node library


owlishDeveloper:
Cool, I'll deploy this first thing tomorrow morning

PR #70, opened 2019-01-09
Start using checks api for our .tc.yml tests

imbstack:
As it says on the tin.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -1,4 +1,5 @@
 version: 1
+reporting: checks-v1
 tasks:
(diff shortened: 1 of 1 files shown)


owlishDeveloper:
yee-haw!

owlishDeveloper:
(what's with the failure?)

imbstack:
That failure has been showing up a lot more after the node 10 update. I think once we can publish a newer version of taskcluster/remotely-signed-s3 we will be on a more modern libxmljs and be fine.

PR #74, opened 2019-01-10
Add tc-web-server to monorepo

owlishDeveloper:


yarn-error.log
```diff
@@ -0,0 +1,6421 @@
+Arguments:
```
helfi92: I don't think we  want to include `yarn-error.log` since this is only useful to debug your own copy of the code. Let's remove it and include `yarn-error.log` in `.gitignore`.

services/web-server/package.json
```diff
@@ -43,13 +21,11 @@
     "dataloader": "^1.4.0",
     "debug": "^3.1.0",
     "deepmerge": "^2.0.1",
-    "ejs": "^2.6.1",
```
helfi92: Is there a reason for removing this package? https://github.com/taskcluster/taskcluster-web-server/blob/c5442223d2e6994a8bfc928afa13ae7a48ff4bc3/src/servers/createApp.js#L12 uses it.

.github/CODEOWNERS
```diff
@@ -11,6 +11,7 @@ services/purge-cache @imbstack
 services/queue @jhford
```
helfi92: Out of curiosity, how were you able to preserve the git history? It's awesome :)

yarn-error.log
owlishDeveloper: Oh, yes, it sneaked in by accident. I'll add it to `.gitignore`. Thank you for noticing!

services/web-server/package.json
owlishDeveloper: Weird - our metatests ruled this dependency to be unused. Good catch here, thank you!

.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 lerna-debug.log
 .eslintcache
 user-build-config.yml
+**/yarn-error.log
```
helfi92: Assuming we never want to push logs, we can replace `lerna-debug.log` and `yarn-error.log` with `*.log`.

services/web-server/.neutrinorc.js
```diff
@@ -0,0 +1,36 @@
...
+      neutrino.config.merge({
+        externals: [
+          nodeExternals({
+            whitelist: [/^webpack/]
+          }),
+          nodeExternals({
+            modulesFromFile: true,
+          }),
+          nodeExternals({
+            modulesDir: '../../node_modules',
```
helfi92: Using forward slashes might cause trouble for users running on windows. Another way to write this could be with:

```jsx
const { join } = require('path');

// ...

modulesDir: join('..', '..', 'node_modules');
```

services/web-server/.neutrinorc.js
```diff
@@ -0,0 +1,36 @@
+const nodeExternals = require('webpack-node-externals');
```
helfi92: There seems to be 2 `.neutrinorc.js` files. Let's delete `.neutrinorc.js` and keep `services/web-server/.neutrinorc.js`.

services/web-server/.neutrinorc.js
owlishDeveloper: oh cool!

services/web-server/.neutrinorc.js
owlishDeveloper: yeah, a left-over from Friday banging on the wall :)

.gitignore
```diff
@@ -1,5 +1,4 @@
 node_modules
-lerna-debug.log
 .eslintcache
 user-build-config.yml
-**/yarn-error.log
+**/*.log
```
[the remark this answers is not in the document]
owlishDeveloper: I want to target all logs in all the dirs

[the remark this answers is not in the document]
helfi92: `*.log` will target all logs in all the dirs.

owlishDeveloper:
linting test fails, but we decided I sort that out in a separate PR

owlishDeveloper:
I would probably have to start the whole shebanga again :( Forgot to warn you not to merge anything while this is happening - I didn't know it would take this long tho. The change in that commit looks pretty small - we can make a PR with that change here

PR #77, opened 2019-01-11
Have a better task info in Checks

owlishDeveloper:
I retain task id in the check names to ensure different tasks with the same name will get different check runs without throwing errors. But I have the task name from `taskDefinition.metadata` preceding that, so that when you look at the list of 27 tasks, you could instantly tell which is which. I am also missing link to the task group - so I added it to the task description.

Downside of this particular solution is that I make additional network call. Alternative solution would be to add metadata to the pulse messages. (I would actually prefer the alternative solution, unless there are objections. I think queue service won't die if it will send some additional bit of data, wheres additional network calls here make tc-gh more brittle. On the other hand, those messages are public API, so changing them...)

### DIFF
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -85,6 +85,8 @@ class Handlers {
     this.deprecatedResultStatusPq = null;
     this.initialTaskStatusPq = null;
     this.deprecatedInitialStatusPq = null;
+
+    this.queueClient = null;
   }
 
   /**
@@ -97,6 +99,11 @@ class Handlers {
     assert(!this.initialTaskStatusPq, 'Cannot setup twice!');
     assert(!this.deprecatedResultStatusPq, 'Cannot setup twice!');
     assert(!this.deprecatedInitialStatusPq, 'Cannot setup twice!');
+
+    this.queueClient = new taskcluster.Queue({
+      rootUrl: this.context.cfg.taskcluster.rootUrl,
+      credentials: this.context.cfg.taskcluster.credentials,
+    });
 
     // Listen for new jobs created via the api webhook endpoint
     const GithubEvents = taskcluster.createClient(this.reference);
@@ -218,12 +225,8 @@ class Handlers {
 
   // Create a collection of tasks, centralized here to enable testing without creating tasks.
   async createTasks({scopes, tasks}) {
-    let queue = new taskcluster.Queue({
-      rootUrl: this.context.cfg.taskcluster.rootUrl,
-      credentials: this.context.cfg.taskcluster.credentials,
-      authorizedScopes: scopes,
-    });
-    await Promise.all(tasks.map(t => queue.createTask(t.taskId, t.task)));
+    const scopedQueueClient = this.queueClient.use({authorizedScopes: scopes});
+    await Promise.all(tasks.map(t => scopedQueueClient.createTask(t.taskId, t.task)));
   }
 
   // Send an exception to Github in the form of a comment.
@@ -406,14 +409,18 @@ async function statusHandler(message) {
         check_run_id: checkRun.checkRunId,
       });
     } else {
+      const taskDefinition = await this.queueClient.task(taskId).catch(debug);
+      debug(`Result status. Got task build from DB and task definition for ${taskId} from Queue service`);
+
       const checkRun = await instGithub.checks.create({
         owner: organization,
         repo: repository,
-        name: `Task ${taskId}`,
+        name: `${taskDefinition.metadata.name}: task ${taskId}`,
         head_sha: sha,
-        output: { // TODO: a helpful output
-          title: `TaskGroup: queued for ${eventType})`,
-          summary: `Check for ${eventType}`,
+        output: {
+          title: `${this.context.cfg.app.statusContext} (${eventType.split('.')[0]})`,
+          summary: `${taskDefinition.metadata.description}`,
+          text: `[Task group](${libUrls.ui(this.context.cfg.taskcluster.rootUrl, `/groups/${taskGroupId}}`)})`,
         },
         details_url: libUrls.ui(
           this.context.cfg.taskcluster.rootUrl,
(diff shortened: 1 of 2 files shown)


owlishDeveloper:
Ah, there we go - you can see the result in this PR already

owlishDeveloper:
[](url)

owlishDeveloper:
🐞 There are 81 checks here. So, if you force-push, it creates new ones, and the old ones stay. Have you noticed this before, bstack?

services/github/src/handlers.js
```diff
@@ -406,14 +409,18 @@ async function statusHandler(message) {
...
       });
     } else {
+      const taskDefinition = await this.queueClient.task(taskId).catch(debug);
+      debug(`Result status. Got task build from DB and task definition for ${taskId} from Queue service`);
+
       const checkRun = await instGithub.checks.create({
         owner: organization,
         repo: repository,
-        name: `Task ${taskId}`,
+        name: `${taskDefinition.metadata.name}: task ${taskId}`,
```
imbstack: I think I'd prefer to leave the `taskId` out of the name. It makes it quite long and hard to see the how-long-it-took portion of the reports. It is really an internal implentation detail of how taskcluster works. The user shouldn't really care or know that it exists.

I think to ensure uniqueness of `metadata.name`, we can assert that they are unique at task submission time.

services/github/src/handlers.js
```diff
@@ -406,14 +409,18 @@ async function statusHandler(message) {
...
         head_sha: sha,
-        output: { // TODO: a helpful output
-          title: `TaskGroup: queued for ${eventType})`,
-          summary: `Check for ${eventType}`,
+        output: {
```
imbstack: I find myself wondering if we should not report group status. Is the purpose of it to allow somebody to block on us being entirely done?

services/github/src/handlers.js
```diff
@@ -710,6 +717,9 @@ async function taskDefinedHandler(message) {
     installationId,
   } = await this.context.Builds.load({taskGroupId});
 
+  const taskDefinition = await this.queueClient.task(taskId).catch(debug);
```
imbstack: Same thing with `debug` here.

services/github/src/handlers.js
```diff
@@ -98,6 +100,11 @@ class Handlers {
     assert(!this.deprecatedResultStatusPq, 'Cannot setup twice!');
     assert(!this.deprecatedInitialStatusPq, 'Cannot setup twice!');
 
+    this.queueClient = new taskcluster.Queue({
```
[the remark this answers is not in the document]
owlishDeveloper: New things aren't, but creation of tasks is. And the scoped client (from `createTasks` function) cannot be saved to the `Handlers` object. Unscoped one can.

services/github/src/handlers.js
owlishDeveloper: What do you mean? (I'm not reporting group status, just add the link to the task group in the task. I find myself wanting to see the whole taskgroup often, but now there's no link to it)

services/github/src/handlers.js
owlishDeveloper: Would this work with the tasks created by a decision task as well? `jobHandler` doesn't touch those

Or are you proposing having a state (like an Azure table with all the names for particular sha or whatever)?

imbstack: Ah, you're very right about that. I definitely don't want more state. I'm reading the docs and their developer forums but don't see anything about what happens if you try to create two runs with the same title. Does it just reject the second one?

owlishDeveloper: I know what happens: the first run will get replaced by the second

owlishDeveloper: Yeah, I can't find any other solution that wouldn't include asking some DB for existing checks. Besides, it seems we are kinda pushed into that path by github... If github does it, why can't we... So be it

I'll include this into our documentation tho

services/github/src/handlers.js
```diff
@@ -406,14 +412,18 @@ async function statusHandler(message) {
...
         repo: repository,
-        name: `Task ${taskId}`,
+        name: `${taskDefinition.metadata.name}`,
         head_sha: sha,
-        output: { // TODO: a helpful output
-          title: `TaskGroup: queued for ${eventType})`,
-          summary: `Check for ${eventType}`,
+        output: {
+          title: `${this.context.cfg.app.statusContext} (${eventType.split('.')[0]})`,
+          summary: `${taskDefinition.metadata.description}`,
```
imbstack: Nice

services/github/src/handlers.js
```diff
@@ -98,6 +100,14 @@ class Handlers {
     assert(!this.deprecatedResultStatusPq, 'Cannot setup twice!');
     assert(!this.deprecatedInitialStatusPq, 'Cannot setup twice!');
 
+    // This is a simple Queue client without scopes to use throughout the handlers for simple things
+    // Where scopes are needed, use this.queueClient.use({authorizedScopes: scopes}).blahblah
+    // (see this.createTasks for example)
+    this.queueClient = new taskcluster.Queue({
```
imbstack: I sorta don't like having this around in here since it is powerful and easy to make a mistake with. The comment says that it is for simple things where scopes aren't needed, but this client actually has _all_ of tc-github's scopes. I think we can just use the `authorizedScopes` client like before since fetching a task's status doesn't require any scopes.

owlishDeveloper: client with authorized scopes is created in `jobHandler` that has access to the scopes in `.taskcluster.yml`. Do you mean I should just create unscoped client every time it's needed as opposed to having it in the object?

owlishDeveloper: Also, I thought that when we create the scoped client, it has all the github scopes _and_ the scopes from `.taskcluster.yml`. Is that incorrect?

imbstack: Almost. When we use `authorizedScopes` it limits the set of scopes that the client can use to whatever the task is allowed based on the roles we assign it. Here's the scenario I'm scared of with the way this is built now:

Imagine in 6 months I come along and want to make a change to this file. I will look for the first queue client I have sitting around in here which very well could be `this.queueClient` and then I will use it to do some action that a task has requested. This is now a scope escalation because I didn't use the client with limited scopes. I would feel much safer if we just didn't ever generate a non-limited client in here.

Making a new client every time is a fine way to do it. I don't think it costs very much to do.

PR #79, opened 2019-01-13
[WIP] Bug 1518251 - sketch of postgres interface

djmitche:
This is a sketch of a new taskcluster-lib-postgres, plus use of that
library in the secrets service.  This is *not* meant to land as-is; instead
please see it as a working sketch of a general idea.

# Overview

The central idea here is to make all DB access go via stored procedures,
and to keep the behavior of those stored procedures fixed forever, such
that old and new versions of a TC service can run simultaneously with no
downtime, even after a schema upgrade.  That schema upgrade will include
modifications to all affected stored procedures such that they continue
to have the same behavior despite changes to the underlying tables.

# Discipline

As we have for Azure tables, we will need to commit to only ever
appending to a Schema object (.addVersion, .addMethod) and never
modifying the existing definitions, except for bugfixes to existing
stored procedures. This has been easy to verify in review for Azure.

# Deployment

Users can deploy all services on a single postgres server (by setting
all *_DB_URL to the same value), in which case the services will not
collide with one another since they use distinct schema names.  Users
can also choose to deploy some or all services to different databases or
even servers, or to split read and write traffic to different servers.

# Updates

We will work a one-off run of `updateDb` into every terraform run, such
that the running application won't be updated to the new version until
`updateDb` has run successfully.  That will perform schema migrations
where necessary (as specified by `addVersion`) and will re-write the
stored functions on every run so that updates that only affect functions
do not require a database version change.

# Testing DB Definitions

An advantage of keeping *all* history of a schema is that we can test
inter-version compatibility.  For example, we can upgrade to version 1,
insert some test data, verify that function X returns expected data,
then upgrade to the latest version and verify that it continues to
return precisely the same data.

With a little more thought, we can regularize and automate this approach
so that it's easy to test that method behaviors don't change from
version to version.

# Testing Non-DB Code

Because all access to the DB is via regularized DB functions that appear
as normal JS functions, replacing the DB object with a fake is trivial.
So, most services can be tested both with and without a DB, just as we
do now for Azure, meaning that contributors can just run `yarn test` and
get useful results, without setting up a postgres server.

# TODO

 - secrets tests for DB definitions
 - secrets tests updated to use fakes
 - terraform support
 - deploy secrets to a dev environment using postgres

# Future

 - pg-cursor supports streaming responses, so we could have some methods
   return streaming data, rather than the full set of rows
 - performance metrics (slow queries, etc.)
 - wrapper for continuation / limit handling
 - custom pg datatype for on-the-fly encryption / decryption (which is
   what azure-entities does)

### DIFF
--- a/libraries/postgres/src/index.js
+++ b/libraries/postgres/src/index.js
@@ -1,0 +1,256 @@
+const {Pool, Client} = require('pg');
+const {dollarQuote} = require('./util');
+const assert = require('assert');
+const debug = require('debug')('taskcluster-lib-postgres');
+
+// Overall plan is: access is only through plpgsql functions, and those must
+// have a consistent API (args + result).  Upgrades can redefine these
+// functions and add new functions, but not change API as existing software may
+// be using those functions concurrently.
+
+const [READ, WRITE] = [Symbol('read'), Symbol('write')];
+
+class Schema{
+  /**
+   * Create a new Schema
+   *
+   * serviceName must be provided, and is used to namespace this service's
+   * tables apart from other services (note that the deployment may put
+   * other services' tables on other database servers entirely, so it is
+   * not allowed to rely on other services' tables)
+   *
+   * script is a script to create the schema, suitable as an argument to
+   * the Postgres DO statment; that is usually 'BEGIN stmt; stmt; .. END'.
+   */
+  constructor({serviceName, script}, {version, versions, methods}={}) {
+    assert(serviceName, 'serviceName is required');
+    this.serviceName = serviceName;
+    assert(script, 'script is required');
+    this.script = script;
+
+    this.version = version || 1;
+    this.versions = versions || [null];
+
+    assert.equal(this.versions.length, this.version);
+    this.versions.push(this);
+
+    this.methods = new Map(methods);
+  }
+
+  /**
+   * Add a new version to this schema; version must be one more than the
+   * previous version.  A script must be provided to upgrade from the previous
+   * version. All defined methods are reset.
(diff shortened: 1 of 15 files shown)


imbstack:
This is making sense to me so far! I need to research a lot of concepts more deeply before I'm completely on board, but thanks so much for starting this.

libraries/postgres/src/index.js
```diff
@@ -0,0 +1,256 @@
...
+   *
+   * serviceName must be provided, and is used to namespace this service's
+   * tables apart from other services (note that the deployment may put
+   * other services' tables on other database servers entirely, so it is
+   * not allowed to rely on other services' tables)
+   *
+   * script is a script to create the schema, suitable as an argument to
+   * the Postgres DO statment; that is usually 'BEGIN stmt; stmt; .. END'.
+   */
+  constructor({serviceName, script}, {version, versions, methods}={}) {
```
owlishDeveloper: I experimented with `{version, versions, methods}={}` - if an empty object is passed in, or an object holding `undefined` in any of the props, or no object was passed in, - in all the cases all these variables are `undefined` in the function. Was that the intended effect?

owlishDeveloper: oic. You're avoiding the destructuring type error

djmitche: Yeah, that second positional argument is sort of for "internal use only".  I don't really like this approach (two options-style positional arguments) and would change it before deploying this "for real".

djmitche:
(see bug)

PR #80, opened 2019-01-13
Bug1518564 stop bluebird

Biboswan:


services/notify/src/irc.js
```diff
@@ -1,12 +1,26 @@
...
-const irc = Promise.promisifyAll(require('irc-upd'));
+const util = require('util');
+const irc = require('irc-upd');
 const assert = require('assert');
 const aws = require('aws-sdk');
 const _ = require('lodash');
 
 const MAX_RETRIES = 5;
 
+Object.keys(irc).forEach( key => {
```
imbstack: I think I'd rather see us directly wrap things in promises below or avoid using promises when they are unsupported below. This is a big scary blob to have at the top of this file.

Biboswan: Yeah I felt the same. Just promify where it is required directly.

services/github/src/main.js
```diff
@@ -186,12 +186,12 @@ const load = loader({
       return monitor.oneShot('syncInstallations', async () => {
         const gh = await github.getIntegrationGithub();
         const installations = (await gh.apps.getInstallations({})).data;
-        await Promise.map(installations, inst => {
+        await Promise.all(_.map(installations, inst => {
```
owlishDeveloper: There's no need for adding a dependency: `await Promise.all(installations.map(inst => { etc.` Please prune lodash from this PR

Biboswan: Lodash was already being required in the files i changed. But yes i felt that too from chat on irc

owlishDeveloper: Yeah, wit's already used in a lot of places. I'm trying to get rid of it in tc-github, so I'd prefer we don't create new usages of lodash, especially in places like this. Here you don't even have to use any obscure ES7 trickery - just plain ES5 `map`

services/github/src/main.js
```diff
@@ -3,7 +3,7 @@ const exchanges = require('./exchanges');
 const Handlers = require('./handlers');
 const Intree = require('./intree');
 const data = require('./data');
-const Promise = require('bluebird');
+const _ = require('lodash');
```
imbstack: No longer needed!

services/notify/package.json
```diff
@@ -11,8 +11,7 @@
   },
   "dependencies": {
     "aws-sdk": "^2.379.0",
-    "bluebird": "^3.5.1",
-    "debug": "^3.1.0",
+    "debug": "^4.0.0",
```
imbstack: Why this change?

PR #86, opened 2019-01-15
Bug 1520061 - build monorepo from a git clone, not working copy

djmitche:
This will get changed pretty radically when we build into a single docker image (bug 1516347). But the fundamental problem is still there: balancing ease of development against security.

Building from your working copy makes it easy to, for example, build an image of your current work in progress and test it in a dev deployment.  As I've implemented this, doing otherwise would require pushing to a github repo, then editing the spec to point to that branch on that repo, then building.

Other ideas:
 * `git clone` from the working copy by default, and warn if `git status` shows un-checked-in files.  This would safely ignore everything that git ignores
 * `git clone` from the working copy when given a command-line option (e.g., `--local-repo`)

thoughts?

### DIFF
--- a/infrastructure/builder/src/build/repo.js
+++ b/infrastructure/builder/src/build/repo.js
@@ -25,39 +25,33 @@ const generateRepoTasks = ({tasks, baseDir, spec, cfg, name, cmdOptions}) => {
     ],
     locks: ['git'],
     run: async (requirements, utils) => {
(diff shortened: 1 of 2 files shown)


djmitche:
Yeah, that makes sense.  I can give it more thought in the bug I mentioned.  Ideally we will eventually not have a spec..

PR #87, opened 2019-01-15
Use req.clientId() to generate InsufficientScopes message

djmitche:


djmitche:
Deployed to queue and confirmed it still gives the right error.  I'm not sure under what conditions result.clientId was undefined, so I didn't test those conditions -- but at least this didn't make the situation worse.

PR #88, opened 2019-01-15
Make services-reviewers responsible for some services, libraries

djmitche:
Note that the last match for a file wins, so this does not override the
more-specific assignments to specific people.


owlishDeveloper:
Can be merged or rebased now. I investigated tc-gh - it really is a bug, consistent repro, and I know the reason. Thank you!

PR #91, opened 2019-01-16
Bug 1260285 - add a metaschema and validate against it

djmitche:
..a newer version of #49.


libraries/references/schemas/metadata-metaschema.yml
```diff
@@ -27,9 +37,14 @@ allOf:
           version:
             title: "Version of the document format"
             type: integer
-          additionalProperties: false
```
imbstack: :man_facepalming:

libraries/validate/src/schemaset.js
```diff
@@ -121,7 +107,14 @@ class SchemaSet {
   async validator(rootUrl) {
     await publish({cfg: this.cfg, schemaset: this, rootUrl});
 
-    const ajv = Ajv({useDefaults: true, format: 'full', verbose: true, allErrors: true});
+    const ajv = Ajv({
+      useDefaults: true,
+      format: 'full',
+      verbose: true,
+      // schema validation occurs in the tests and need not be done here
```
imbstack: Thank you for this comment. I would ask this exact question a year from now otherwise.

djmitche:
It's all in one repo now, so we can pretty easily upgrade to draft-7 at some point.

djmitche:
(rebased to update the schemas added/modified since the first version -- nothing exciting)

djmitche:
The risk on deployment here was that the service would not start up due to some failure.  Auth started up just fine.

libraries/references/schemas/metadata-metaschema.yml
```diff
@@ -27,9 +37,14 @@ allOf:
...
+        additionalProperties: false
         required:
           - version
           - name
+    additionalProperties: true
     required:
       - metadata
+
+allOf:
+  - {$ref: "./metaschema.json#"}
```
petemoore: Is this definitely legal?
I think this might need to be fully qualified, like `$schema`. Then `$schema` tells us that this document conforms to our `metaschema.json`, and this `$ref` says that anything that conforms to this schema also conforms to `metaschema.json`. Would `$ref: "/schemas/common/metaschema.json#"` get expanded like `$schema` gets expanded on line 1, if we used that instead? Or does expansion only occur on the `$schema` property?

djmitche: (discussed in irc) yep, `$ref` can be relative, and that avoids the need to find and rewrite `$ref` entries when making absolute schemas.

PR #95, opened 2019-01-16
Bug1496890- show clientId of hook triggerer

Biboswan:


services/hooks/src/v1.js
```diff
@@ -494,9 +495,9 @@ builder.declare({
 /**
  * Common implementation of triggerHook and triggerHookWithToken
  */
-const triggerHookCommon = async function({req, res, hook, payload, firedBy}) {
+const triggerHookCommon = async function({req, res, hook, payload, clientId = null, firedBy}) {
```
djmitche: this will automatically default to `undefined` (which is what we want -- the context item shouldn't be set at all, rather than being set to null), so no need for the `= null` here.

services/hooks/test/api_test.js
```diff
@@ -619,7 +619,7 @@ helper.secrets.mockSuite('api_test.js', ['taskcluster'], function(mock, skipping
       assume(helper.creator.fireCalls).deep.equals([{
         hookGroupId: 'foo',
         hookId: 'bar',
-        context: {firedBy: 'triggerHookWithToken', payload},
+        context: {firedBy: 'triggerHookWithToken', clientId: null, payload},
```
djmitche: can be omitted here and in the previous test, too.

Biboswan: it should be then replaced with clientId: undefined in the test for the tests to pass

Biboswan:
> Please also update the docs (in `services/hooks/docs`)
yeah waited for the patch to be reviewed before going to the docs

PR #97, opened 2019-01-17
Add web-server to monorepo: Fix linting

owlishDeveloper:
Review after this one is merged https://github.com/taskcluster/taskcluster/pull/74

### DIFF
--- a/services/web-server/src/PulseEngine/index.js
+++ b/services/web-server/src/PulseEngine/index.js
@@ -175,34 +175,35 @@ export default class PulseEngine {
     });
   }
 
-  @serialize
   async innerReconcileSubscriptions() {
-    debug('Reconciling subscriptions');
+    return serialize(async () => {
+      debug('Reconciling subscriptions');
 
-    const { connection, client } = this;
+      const { connection, client } = this;
 
-    // if there's no connection, there's nothing to do; reconciliation
-    // will occur again on the next connection
-    if (!connection) {
-      return;
-    }
+      // if there's no connection, there's nothing to do; reconciliation
+      // will occur again on the next connection
+      if (!connection) {
+        return;
+      }
 
-    if (!this.channel) {
-      this.channel = await connection.amqp.createChannel();
-    }
+      if (!this.channel) {
+        this.channel = await connection.amqp.createChannel();
+      }
 
-    const { channel } = this;
+      const { channel } = this;
 
-    await Promise.all(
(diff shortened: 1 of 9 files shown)


services/web-server/src/servers/createApp.js
```diff
@@ -31,7 +31,7 @@ export default async ({ cfg }) => {
 
   await Promise.all(
     cfg.app.loginStrategies.map(async strategy => {
-      const { default: loginStrategy } = await import(`../login/${strategy}`);
+      const { default: loginStrategy } = require(`../login/${strategy}`);
```
helfi92: `require` is a synchronous function. We would need to remove the `async` keyword from the function definition since we're not using `await` anymore. We would also need to remove `await Promise.all`.

I think you can just rewrite as:

```js
cfg.app.loginStrategies.forEach(strategy => {
  const { default: loginStrategy } = require(`../login/${strategy}`);

  loginStrategy(app, cfg);
});
```

owlishDeveloper:
Can you make a PR with the change in that last commit here? I forgot to warn you not to diverge histories..... I didn't know it would take this long tho

services/web-server/src/PulseEngine/index.js
```diff
@@ -175,34 +175,35 @@ export default class PulseEngine {
     });
   }
 
-  @serialize
   async innerReconcileSubscriptions() {
-    debug('Reconciling subscriptions');
+    return serialize(async () => {
```
djmitche: When I was working with @ydidwania to use decorators, they had a more complex function signature than this (taking three arguments).  But I didn't totally understand things, so I'll just ask: does this work? :)

owlishDeveloper: I was, indeed, able to start the server... I may not have used the functionality that calls this. Which annotations library did you use with yd?

djmitche: I tihnk we were using serialize.  Anyway, we can find out how it works after landing.

helfi92: We are using the legacy way of decorators. This is coming from https://github.com/taskcluster/taskcluster/blob/38c7219dc1b6f1040e64518964d2257f6500d6e8/services/web-server/.neutrinorc.js#L10. Now that we are no longer using decorators, we should perhaps remove `@babel/plugin-proposal-decorators` from the repo.

helfi92:
I'll pull the code to test subscriptions.

PR #99, opened 2019-01-17
Pass errors in body parsing to Express

djmitche:


djmitche:
Deployed to hooks and confirmed there.

PR #102, opened 2019-01-17
move more CODEOWNERS to services-reviewers

djmitche:


djmitche:
good point -- I just assigned * to the services-reviewers team

PR #110, opened 2019-01-18
Increase workerType, provisionerID & schedulerId length

OjaswinM:
I think I covered everything in `services/*/schemas/v1/*` and `services/*/src/{v1,api}.js` but I'm not sure if I should look outside those directories. 

Also, how should I go about implementing the tests?


OjaswinM:
Dumb question but what are the other remaining parts? :)

djmitche:
(commented in the bug)

OjaswinM:
I changed all the relevant tests to have longer Ids and checked that they were for Ids more than 40 chars long. I also fixed some length assertions in some places I had missed.

For the lib-pulse checks, should I make a separate PR?

OjaswinM:
I think we already do it [here](https://github.com/taskcluster/taskcluster/blob/30372ec731730592a7bf22d4b349a61f75663883/libraries/pulse/src/publisher.js#L165)

djmitche:
Great!  I think this is done, then (well, once I merge and deploy the PR, which I will do on Tuesday when I'm back at work)

petemoore:
@djmitche This is a major change, and can have a huge impact. Please let's discuss changes like this first. For example, AMQP routing keys are of limited length, so increasing length of identifiers can break task routes, which can impact many of our services. We should also consider how this impacts workers/clients, and work out (together) if we need to roll out changes there at the same time.

djmitche:
We did account for the AMQP route length (the last few comments disucss that, as well as some content in the bug).  Let's discuss in the bug.

djmitche:
(the bug = https://bugzilla.mozilla.org/show_bug.cgi?id=1520579)

services/hooks/schemas/constants.yml
```diff
@@ -1,6 +1,7 @@
 identifier-pattern:     "^([a-zA-Z0-9-_]*)$"
 identifier-min-length:  1
 identifier-max-length:  22
+identifier-extended-max-length:  40
```
jhford: If we're switching to a longer format, do we need to keep the old length around?  All old IDs are valid new IDs and my understanding is that this patch is designed to update *everything* to the new format in one go.

Could we have a single field for max identifier length?  If not, maybe we should hold off on landing until we can have a single field?  I'm concerned that the old identifier (indentifier-max-length) will end up living in perpetuity and will be confusing to future implementers.

djmitche:
We'll take this through the RFC process -- I'll link to that from the bug.  @OjaswinM ping me on IRC and I'll fill you in on how that works :)

djmitche:
https://github.com/taskcluster/taskcluster-rfcs/pull/139 is the RFC for this

djmitche:
OK, it's [official](https://github.com/taskcluster/taskcluster-rfcs/blob/master/rfcs/0139-longer-identifiers.md)!

djmitche:
@OjaswinM please have a look at the RFC and adjust this implementation to match. I think that will mean naming the formats "identifier" and "slugid", and double-checking that all of the named identifiers have been addressed.

We'll also need to make parallel changes in [ec2-manager](https://github.com/taskcluster/ec2-manager) and [aws-provisioner](https://github.com/taskcluster/aws-provisioner) which are not included in this repository.

OjaswinM:
Ohh great :) 

I'll start working on it asap

OjaswinM:
alright so the RFC says we need to change `organization` and `repository` in tc-github to "identifier" of length 40 but they are at length 100 right now([1] & [2]). Should I reduce them to 40? 

[1] https://github.com/taskcluster/taskcluster/blob/79ee65e91d30f813285c8425a906a56368b3ba0e/services/github/schemas/v1/build-list.yml#L18-L29

[2] https://github.com/taskcluster/taskcluster/blob/79ee65e91d30f813285c8425a906a56368b3ba0e/services/github/schemas/constants.yml#L9

djmitche:
Haha, good find.. I wish we had brought that up in the RFC!  No, let's leave those at 100.

djmitche:
OK, RFC is resolved -- 38 characters it is!

OjaswinM:
Awesome, i made the changes and will be proceeding with the tests. @djmitche should I change all the sample ids to be of greater length in the test like I did the last time around?

Also, is this taskId related to other taskIds, and if so, should its value be changed to 22? 
Asking because it's set to 50 right now rather than the usual 22

https://github.com/taskcluster/taskcluster/blob/887b4d9ce952992edb734a34208965d8558b29b1/services/treeherder/schemas/v1/pulse-job.yml#L17-L25

djmitche:
Yes, testing with longer identifiers is a good idea.

That "Task ID" field is actually a different thing (it's for treeherder, and it's shaped like a [uuid](https://en.wikipedia.org/wiki/Universally_unique_identifier)) so leave it as-is.

djmitche:
Also please run `yarn generate` and check in the results of that run (it will make some changes under generated/)

OjaswinM:
Ahh okay, there's also an error in taskcluster-queue due to expired deadline i think

djmitche:
https://bugzilla.mozilla.org/show_bug.cgi?id=1526381 for the queue error -- I don't think that's related to this bug.

OjaswinM:
Alright i'll start with the tests. I do have some questions regarding changes in ec2-manager but I guess [the bug would be a better place to discuss that](https://bugzilla.mozilla.org/show_bug.cgi?id=1520579).

OjaswinM:
@djmitche @jhford  I have modified the tests and made sure they were failing for lengths above 38. I changed some parts of `worker-manager` which mainly involved asserting that `workerType` is lesser than 38 in `schema/`, `tests/` and `lib/`. Kindly review the changes.

Also, doubts :) 

1) https://github.com/taskcluster/taskcluster/blob/732ff6c8cf8020335c2e7d170b3a6dbdb0387b2a/services/worker-manager/schemas/v1/worker-configuration.yml#L7-L13

is this same as workerId? I tried to restrict it to `identifier-pattern` but a lot of tests have used `.` in the naming which is rejected by the schema.


2) I'm getting this error when running `yarn generate` but it seems to pass here:- 

```
✖ Generate References (Error: /home/ojaswin/TaskCluster/taskcluster-services/generated/docs/worker-manager/schemas/v1/ruleset.json: schema.items.
properties['conditions'] should have properties type, additionalProperties, required when property properties is present; /home/ojaswin/TaskClust
er/taskcluster-services/generated/docs/worker-manager/schemas/v1/ruleset.json: schema.items should be array; /home/ojaswin/TaskCluster/taskcluste
r-services/generated/docs/worker-manager/schemas/v1/ruleset.json: schema.items should match some schema in anyOf)
94% finished
{ Error: /home/ojaswin/TaskCluster/taskcluster-services/generated/docs/worker-manager/schemas/v1/ruleset.json: schema.items.properties['conditions'] should have properties type, additionalProperties, required when property properties is present; /home/ojaswin/TaskCluster/taskcluster-services/generated/docs/worker-manager/schemas/v1/ruleset.json: schema.items should be array; /home/ojaswin/TaskCluster/taskcluster-services/generated/docs/worker-manager/schemas/v1/ruleset.json: schema.items should match some schema in anyOf
    at exports.validate (/home/ojaswin/TaskCluster/taskcluster-services/libraries/references/src/validate.js:175:11)
    at References.validate (/home/ojaswin/TaskCluster/taskcluster-services/libraries/references/src/references.js:122:5)
    at References.validate (/home/ojaswin/TaskCluster/taskcluster-services/libraries/references/src/references.js:117:16)
    at References.makeSerializable (/home/ojaswin/TaskCluster/taskcluster-services/libraries/references/src/references.js:141:10)
    at Object.run (/home/ojaswin/TaskCluster/taskcluster-services/infrastructure/builder/src/generate/generators/service-docs.js:82:31)
  problems:
   [ '/home/ojaswin/TaskCluster/taskcluster-services/generated/docs/worker-manager/schemas/v1/ruleset.json: schema.items.properties[\'conditions\'] should have properties type, additionalProperties, required when property properties is present',
     '/home/ojaswin/TaskCluster/taskcluster-services/generated/docs/worker-manager/schemas/v1/ruleset.json: schema.items should be array',
     '/home/ojaswin/TaskCluster/taskcluster-services/generated/docs/worker-manager/schemas/v1/ruleset.json: schema.items should match some schema in anyOf' ] }
```

djmitche:
Errors from JSON-schema validation can be confusing.  In this case, all it's saying is that in `service/worker-manager/schemas/v1/ruleset.yml`, `items.properties.conditions` needs a `required` property (it already has `type` and `additionalProperties`).  I'm not sure why you're seeing that from `yarn generate` but it's not happening in CI.  The only thing I can think is that maybe you are using a different version of Ajv?  At any rate, it should be an easy fix -- just add `required: []`.

generated/docs/worker-manager/schemas/v1/ruleset.json
```diff
@@ -10,10 +10,15 @@
...
-        ],
+        "type": "object",
+        "properties": {
+          "workerType": {
+            "type": "string",
+            "minLength": 1,
+            "maxLength": 38,
+            "pattern": "^([a-zA-Z0-9-_]*)$"
+          }
+        },
```
djmitche: Oh, this is why it's giving you that error -- you've added a required property of conditions, and when you specify `properties` you must also specify `type`, `required` and `additionalProperties`.  However, I'm not sure this is the appropriate place to add a workerType -- @jhford would know for sure.

djmitche:
Mystery solved: https://bugzilla.mozilla.org/show_bug.cgi?id=1527699

generated/docs/worker-manager/schemas/v1/ruleset.json
OjaswinM: Ahh alright, I'll try that.

Actually, I noticed that conditions sometimes has the workerType so I added a definition for it which is optional.

I also removed "null" from the type and instead made conditions optional by removing it from required. I believe that would have the same effect?

djmitche: I believe conditions can *optionally* contain a workerType, but can also be null.  I don't think there's a need to enforce length on that particular occurrence of workerType -- if someone has a condition with `{"workerType": "someReallyReallyReallyReallyReallyReallyReallyLongWorkerTypeName"}`, it just won't match any workerTypes since none of them have that name.

OjaswinM: Ahh okay I get it.  I've also added some checks for workerType occurences in conditions in some other places. Should I remove all the checks?

djmitche: I think so.  The worker-manager is still new and not yet in production, so we don't need to worry too much about it.

OjaswinM:
@djmitche I've fixed the schema. Hopefully all the CI tests should pass now.

I've also opened the PR in aws-provisioner which can be found here: [https://github.com/taskcluster/aws-provisioner/pull/196](https://github.com/taskcluster/aws-provisioner/pull/196)

djmitche:
Thanks for your diligence on this project @OjaswinM!

djmitche:
Deployed everywhere without incident!!

PR #116, opened 2019-01-18
Bug 1519870 - drop Hoek dependency

djmitche:
Compare with the original implementation at
  https://github.com/hapijs/hoek/blob/v5.0.4/lib/index.js#L793


djmitche:
I'm rather concerned that all of the tests in #16 passed, when this function didn't even exist.  Let me see if I can add a test case.

djmitche:
Oh -- I missed @imbstack's changes in that PR.  This is fine.  The [B64 implementation](https://github.com/hapijs/b64/blob/master/lib/index.js#L32) matches the one in the old Hoek version.

djmitche:
good job meta test!

PR #118, opened 2019-01-18
Bug in checks: creates a checkrun for the whole group

owlishDeveloper:
I decided to try this: since we get all the tasks anyway, I added a check to examine their `routes` field. If at least a single task belongs to the checks API - stop the whole handler.

I also modified the loop so that we stop iterating as soon as we found a failing task. For all the above I have to use good old `for` loop (I know, I'm so fancy ✨)

Finally, I scratched my lodash itch

### DIFF
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -295,11 +295,17 @@ async function deprecatedStatusHandler(message) {
     do {
       let group = await this.queueClient.listTaskGroup(message.payload.taskGroupId, params);
       params.continuationToken = group.continuationToken;
-      group.tasks.forEach(task => {
-        if (['failed', 'exception'].includes(task.status.state)) {
+
+      for (let i = 0; i < group.tasks.length; i++) {
+        // don't post group status for checks API
+        if (group.tasks[i].task.routes.includes(this.context.cfg.app.checkTaskRoute)) {return;}
+
+        if (['failed', 'exception'].includes(group.tasks[i].status.state)) {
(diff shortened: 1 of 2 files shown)


services/github/src/handlers.js
```diff
@@ -3,7 +3,6 @@ const taskcluster = require('taskcluster-client');
 const libUrls = require('taskcluster-lib-urls');
 const yaml = require('js-yaml');
 const assert = require('assert');
-const _ = require('lodash');
```
djmitche: :100:

services/github/src/handlers.js
```diff
@@ -295,11 +295,17 @@ async function deprecatedStatusHandler(message) {
     do {
       let group = await this.queueClient.listTaskGroup(message.payload.taskGroupId, params);
       params.continuationToken = group.continuationToken;
-      group.tasks.forEach(task => {
-        if (['failed', 'exception'].includes(task.status.state)) {
+
+      for (let i = 0; i < group.tasks.length; i++) {
+        // don't post group status for checks API
+        if (group.tasks[i].task.routes.includes(this.context.cfg.app.checkTaskRoute)) {return;}
```
djmitche: I kinda thought eslint would complain, but IMHO the return statement should be on its own line

owlishDeveloper: Eslint complained only about my very first variant:
`if (blablabla) return;` <---- it wanted the return statement to be in curlies

PR #121, opened 2019-01-21
Include hooks listlastFire api support

Biboswan:


Biboswan:
I have checked query's response in the graghql playground

services/web-server/src/graphql/Hooks.graphql
```diff
@@ -33,6 +33,12 @@ enum HookFireResult {
   NO_FIRE
 }
 
+enum HookFiredBy {
+  SCHEDULE
+  TRIGGERHOOK
```
helfi92: Rename to `TRIGGER_HOOK`.

services/web-server/src/graphql/Hooks.graphql
```diff
@@ -33,6 +33,12 @@ enum HookFireResult {
...
+  TRIGGERHOOKWITHTOKEN
```
helfi92: Rename to `TRIGGER_HOOK_WITH_TOKEN`.

services/web-server/src/graphql/Hooks.graphql
```diff
@@ -89,11 +95,22 @@ input HookInput {
   triggerSchema: JSON
 }
 
+type HookLastFire {
+  hookGroupId: ID!,
+  hookId: ID!,
+  firedBy: HookFiredBy!,
```
helfi92: Should `firedBy` be required? I don't see it set as required in https://github.com/taskcluster/taskcluster/blob/30372ec731730592a7bf22d4b349a61f75663883/services/hooks/schemas/v1/list-lastFires-response.yml#L41-L47.

If it's not required, then it should be `firedBy: HookFiredBy,`. If it's required, then `list-lastFires-response.yml
` needs to be updated.

services/web-server/src/graphql/Hooks.graphql
```diff
@@ -89,11 +95,22 @@ input HookInput {
...
+  result:HookFireResult!,
+  error: String!,
+}
+
 extend type Query {
   hookGroups(filter: JSON): [HookGroup]
   hooks(hookGroupId: ID!, filter: JSON): [Hook]
   hook(hookGroupId: ID!, hookId: ID!): Hook
   hookStatus(hookGroupId: ID!, hookId: ID!): HookStatus
+  hookLastFires(hookGroupId: ID!, hookId: ID!): [HookLastFire]
```
helfi92: When the response is an array, we should allow a user to filter on the result. Let's add an additional optional parameter `filter: JSON`. See the `diff` for how to fully integrate the filter feature.

Biboswan: So I think of adding a searchbar for finding apt lastfires by success, time etc.

helfi92: The graphQL endpoint should still have a filter parameter. It doesn't mean we will use it but as a convention, for all response returning an array, we should expose a filter parameter.

services/web-server/src/graphql/Hooks.graphql
```diff
@@ -89,11 +95,22 @@ input HookInput {
   triggerSchema: JSON
 }
 
+type HookLastFire {
+  hookGroupId: ID!,
+  hookId: ID!,
+  firedBy: HookFiredBy!,
+  taskId: String!,
+  taskCreateTime: DateTime!,
+  result:HookFireResult!,
```
helfi92: Add a space after the column:

`result: HookFireResult!,`

services/web-server/src/resolvers/Hooks.js
```diff
@@ -40,6 +45,9 @@ export default {
     hookStatus(parent, { hookGroupId, hookId }, { loaders }) {
       return loaders.hookStatus.load({ hookGroupId, hookId });
     },
+    hookLastFires(parent, {hookGroupId, hookId, filter}, {loaders}) {
```
helfi92: Add a space after `{` and before `}` when destructuring:

```js
hookLastFires(parent, { hookGroupId, hookId, filter }, { loaders }) {
```

services/web-server/src/resolvers/Hooks.js
```diff
@@ -40,6 +45,9 @@ export default {
...
+      return loaders.hookLastFires.load({hookGroupId, hookId, filter});
```
helfi92: Add a space after `{` and before `}` when destructuring:

`return loaders.hookLastFires.load({ hookGroupId, hookId, filter });`

PR #123, opened 2019-01-22
Add worker manager to Mono Repo

jhford:
This PR is to add the worker manager codebase to the TC monorepo.  I  have followed the repoifier steps as best as I could, but I'm not 100% sure if I did it correctly.  I also need to see if the unit tests are working correctly with the monorepo, so please expect that this branch has failing unit tests until I can fix them.


jhford:
I am removing the automatic review request because I'm not sure if the tests will pass or fail.

imbstack:
It looks like it just needs added to the root `package.json`. It can either be added manually or with the `-W` flag that yarn requires to do things at that level. `yarn add --dev -W memory-streams` should work!

imbstack:
This is an exciting PR. Thanks for this

services/worker-manager/schemas/v1/ruleset.yml
```diff
@@ -0,0 +1,24 @@
+$schema: http://json-schema.org/draft-06/schema#
```
djmitche: This will need to point to the metaschema (see other `schemas/v1/*.yml` for an example)

djmitche:
#124 should fix the permissions thing.

djmitche:
I rebased and addressed the review comments.  I did a `taskcluster-builder build` and (as designed) the worker-manager service was automatically included in the resulting monoimage and started up just fine, happily serving me an empty array for `/api/worker-manager/v1/worker-configurations`.

djmitche:
(We can build k8s deployments for this in a subsequent PR)

djmitche:
@jhford I can't flag you for review but -- please have a look.  I'll get this landed tomorrow.

djmitche:
Woo!

PR #125, opened 2019-01-22
Bug 1426512 - update all hawks to ^7.0.10

djmitche:
I would still like to remove use of superagent-hawk, which requires hawk@1.0.0 (it's only used for testing, so no biggie..)

@imbstack I have a monorepo question here: if we're using the same version of Hawk everywhere, should that just be in the toplevel `package.json`?  But I see that all external packages in the top level are `devDependencies`.  Anyway, is there more I could do here, or is it best to just list `"hawk": "^7.0.10"` in all of the relevant package.jsons?


imbstack:
I've so far tended to put the production dependencies in their own repos other than the taskcluster libs but there's no real reason for that. Either way is fine I think

djmitche:
I will continue to swim with that tide :)

djmitche:
now:
```
(v10.14.0) dustin@lamport ~/p/taskcluster2 [bug1426512*] $ yarn why hawk
yarn why v1.13.0
[1/4] Why do we have the module "hawk"...?
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "hawk@7.0.10"
info Reasons this module exists
   - "_project_#taskcluster-lib-api" depends on it
   - Hoisted from "_project_#taskcluster-lib-api#hawk"
   - Hoisted from "_project_#taskcluster-auth#hawk"
   - Hoisted from "_project_#taskcluster-client#hawk"
   - Hoisted from "_project_#taskcluster-lib-testing#hawk"
info Disk size without dependencies: "148KB"
info Disk size with unique dependencies: "348KB"
info Disk size with transitive dependencies: "424KB"
info Number of shared dependencies: 5
Done in 1.41s.
```

libraries/client/src/client.js
```diff
@@ -628,8 +628,8 @@ exports.createClient = function(reference, name) {
       throw new Error('accessToken must be given');
     }
 
-    // Create bewit (this is messed up, function differs in browser)
```
imbstack: :fireworks:

services/auth/src/signaturevalidator.js
```diff
@@ -205,104 +205,59 @@ const createSignatureValidator = function(options) {
...
-  let loadCredentials = function(clientId, ext, callback) {
+
+  const loadCredentials = async (clientId, ext) => {
     // We may have two clientIds here: the credentialName (the one the caller
     // sent in the Hawk Authorization header) and the issuingClientId (the one
     // that signed the temporary credentials).
     let credentialName = clientId,
       issuingClientId = clientId;
 
-    (async () => {
```
imbstack: I wonder why this was this way before.

djmitche: It was an awkward mix of async/await and Hawk's old callback-based approach.

PR #126, opened 2019-01-23
Add tc-web FE to monorepo

owlishDeveloper:
I don't quite understand how to go about `.taskcluster.yml` - add this dir to the root one, just like other services?

Also, linting. Does it get to have its own linting rules?


.github/CODEOWNERS
```diff
@@ -2,6 +2,7 @@
 
 services/queue @jhford
 services/web-server @taskcluster/frontend-reviewers
+taskcluster-web @taskcluster/frontend-reviewers
```
djmitche: You're going to hate me, sorry, but I kind of want to just call it `web/` or something else short and sweet (`frontend`? `ui`?).  We've eliminated `taskcluster-` from all of the other pathnames in this repo.  Hopefully once a name is selected this isn't a hard fix..

owlishDeveloper: Oh, let's call it `ui`, PLEASE!

owlishDeveloper:
@helfi92 I wonder if we should try deploying it on Netlify? Before merging. Are you able to start this thing locally?

netlify.toml
```diff
@@ -0,0 +1,26 @@
+[build]
```
djmitche: It's probably worth a header comment here explaining how we use Netlify (and that it's not used in production)

test/meta_test.js
```diff
@@ -34,6 +36,10 @@ suite('Repo Meta Tests', function() {
     assert.equal(taskclusterYml.tasks.$let.node, packageJson.engines.node);
   });
 
+  test('Node version for UI matches the rest of the repo', function() {
+    assert.equal(taskclusterYml.tasks.$let.node, uiPackageJson.engines.node);
+  });
+
```
djmitche: :tada:

helfi92:
Please refrain from merging today. I would like to review it tomorrow morning.

djmitche:
> sheep it!

that's my opinion only - sorry if it sounded too eager :)

helfi92:
No worries :)

netlify.toml
owlishDeveloper: I think once it goes to production there'll be no Netlify anyways?

djmitche: I think we'll still use netlify to test PRs?

PR #127, opened 2019-01-23
Implement notifications blacklist as a user editable table

OjaswinM:
Related to [https://bugzilla.mozilla.org/show_bug.cgi?id=1520918](https://bugzilla.mozilla.org/show_bug.cgi?id=1520918)


djmitche:
that commit seems to keep popping up and closing things..

OjaswinM:
I'll start with the tests so I can make sure everything's working

OjaswinM:
Alright so I think some bits are still missing to get the tests running. Here's the error I'm getting
 
image

djmitche:
It's usually easier to comment on something texutal as text, rather than a screenshot :)

```
  1) test of test:
     TypeError: Cannot read property 'list' of undefined
      at Context.<anonymous> (test/api_test.js:69:37)
      at process.topLevelDomainCallback (domain.js:121:23)
  
  2) api_test.js (mock)
       "before all" hook for "test of test":
     Error: Target BlacklistedNotification is not defined
      at /home/dustin/p/taskcluster/libraries/loader/src/loader.js:184:13
      at Object.sticky [as load] (/home/dustin/p/taskcluster/libraries/testing/src/stickyloader.js:23:25)
      at Context.<anonymous> (test/helper.js:413:53)

```

So the first error is occurring because you're running a test outside of the suite:
```js
// Just to check if the table is loading (TODO: Investigate client )
test('test of test', async function() {
  let item = await helper.apiClient.list();                                                                                                                                                                                                                                                                                    
  assert.deepEqual(item.addresses, []);
});
```
and it's the suite's responsibility to make sure that `apiClient` is set up and working.

The second error is because there's no loader component named `BlacklistedNotification` defined.  If you look in `services/secrets/src/main.js` you'll see that there's a loader component named `Secret` that defines the Secret table.  You'll want something similar for BlacklistedNotification.

This is looking good -- keep it up!

OjaswinM:
> It's usually easier to comment on something texutal as text, rather than a screenshot :)
> 
> ```
>   1) test of test:
>      TypeError: Cannot read property 'list' of undefined
>       at Context.<anonymous> (test/api_test.js:69:37)
>       at process.topLevelDomainCallback (domain.js:121:23)
>   
>   2) api_test.js (mock)
>        "before all" hook for "test of test":
>      Error: Target BlacklistedNotification is not defined
>       at /home/dustin/p/taskcluster/libraries/loader/src/loader.js:184:13
>       at Object.sticky [as load] (/home/dustin/p/taskcluster/libraries/testing/src/stickyloader.js:23:25)
>       at Context.<anonymous> (test/helper.js:413:53)
> ```

Oh I'll keep that in mind next time!

> So the first error is occurring because you're running a test outside of the suite:
> 
> ```js
> // Just to check if the table is loading (TODO: Investigate client )
> test('test of test', async function() {
>   let item = await helper.apiClient.list();                                                                                                                                                                                                                                                                                    
>   assert.deepEqual(item.addresses, []);
> });
> ```
> 
> and it's the suite's responsibility to make sure that `apiClient` is set up and working.
> 
> The second error is because there's no loader component named `BlacklistedNotification` defined. If you look in `services/secrets/src/main.js` you'll see that there's a loader component named `Secret` that defines the Secret table. You'll want something similar for BlacklistedNotification.
> 
> This is looking good -- keep it up!

Ahh I get it now, I'll try to fix this

OjaswinM:
So I go the the table working but now I'm getting a different error. 

```
1) api_test.js (mock)
       Blacklist: addBlacklistAddress():
     Error: Internal error, unknown error code: EntityAlreadyExists
Notification address already exists in the blacklist

---

* method:     addBlacklistAddress
* errorCode:  InternalServerError
* statusCode: 500
* time:       2019-01-29T12:01:40.158Z
      at /home/ojaswin/TaskCluster/taskcluster-services/libraries/client/src/client.js:374:21
      at process._tickCallback (internal/process/next_tick.js:68:7)

```
What I noticed is that for some reason `await helper.apiClient.addBlacklistAddress(dummyAddress);` in the last test is being called multiple times which is causing the error. It's actually called 6 times which is equal to the number of tests before the failing test which makes me suspect that it's being called in the beforeEach hook (although I might be completely off the mark here haha). I'll keep look around till you can review it :)

djmitche:
It sounds from your last comment like you're still on track?  @mention me when you want me to have a look again!

OjaswinM:
@djmitche I do have one doubt now.

Right now we are using the `notificationType` as our partition key and `notificationAddress` as the row key and for methods like `BlacklistedNotification.remove()` or `BlacklistedNotification.load()` we need to provide both of the props for it to identify the entity, leading to the following behavior:-
```
let dummyAddress = {
    notificationType: "email",
    notificationAddress: "name1@name.com"
};

item = await this.BlacklistedNotification.load(dummyAddress);

assert.deepEqual(item,dummyAddress);
```

thus making the call redundant. 

One possible fix could be to maybe change the partition key to a constant key as done in [service/secret](https://github.com/taskcluster/taskcluster/blob/2736c6d52c501bfeb6e46918c0827a1d4e39b76e/services/secrets/src/data.js#L8). That way we can fetch and remove by just providing the `notificationAddress` as the param. What do you think?

djmitche:
I don't know if I'd say that was redundant.  If that address isn't in the table, the `load` call will fail (or, if called with `(dummyAddress, true)` will return `false`, which is probably more useful).

This is an example of a database table where presence or absence of a row is the important piece of data, rather than any additional information in the row.

OjaswinM:
>If that address isn't in the table, the `load` call will fail (or, if called with `(dummyAddress, true)` will return `false`, which is probably more useful).

Using it this way is causing a schema error when it returns null. How should I handle that?

services/notify/src/api.js
```diff
@@ -89,3 +91,194 @@ builder.declare({
...
+builder.declare({
+  method: 'get',
+  route: '/blacklist/get',
+  name: 'getBlacklistAddress',
+  input: 'notification-address.yml',
+  //output: 'notification-address.yml',
+  //scopes: 'secrets:get:<name>',
+  title: 'Read Blacklisted Notification',
+  description: [
+    'Read the notification, associated with some address, from the blacklist.',
```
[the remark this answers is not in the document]
OjaswinM: Seeing how we will have to check the blacklist for existence of an address time and again won't it be more efficient to use this method rather than `list()`?

djmitche:
You'll need to add a call to [req.authorize](https://github.com/taskcluster/taskcluster/tree/master/libraries/api#scopes) for those methods.

OjaswinM:
> You'll need to add a call to [req.authorize](https://github.com/taskcluster/taskcluster/tree/master/libraries/api#scopes) for those methods.

Right, that fixed it :)

So I have added blacklist checks in the `notify` methods and written tests for them as well. Everything is working fine except I think the taskcluster-notify test which is failing due to the missing table names etc. Please review and suggest changes. @djmitche 

I'll comment in the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1520918) to discuss further

djmitche:
I pushed a commit to fix up the tests to use the newly added table.  It:
 * configures the table name with `app.blacklistedNotificationTableName` instead of `azure.tableName` (since someday we may have more than one table).
 * calls `withBlacklist` first -- order of the `helper.withXxx` things is important
 * adds TASKCLUSTER_* secrets which are configured in CI.

This should make tests run in CI.  Note, though, that I'm seeing the following when running with `DEBUG=*`:
```
  taskcluster-lib-monitor reportError: Error: Internal error, unknown error code: BlacklistedAddress
undefined is blacklisted level=error tags={} Error: Internal error, unknown error code: BlacklistedAddress
undefined is blacklisted
    at /home/dustin/p/taskcluster/libraries/api/src/middleware/express-error.js:69:21
    at Layer.handle_error (/home/dustin/p/taskcluster/node_modules/express/lib/router/layer.js:71:5)
    at next (/home/dustin/p/taskcluster/node_modules/express/lib/router/route.js:135:13)
    at Promise.resolve.then.then.catch (/home/dustin/p/taskcluster/libraries/api/src/middleware/handle.js:30:14)
    at process._tickCallback (internal/process/next_tick.js:68:7) +1s
```

I am not sure why the tests are succeeding despite the 500 error?  Also, that error code should be declared in the APIBuilder (see the `libraries/api` README).

OjaswinM:
> I pushed a commit to fix up the tests to use the newly added table. It:
> 
>     * configures the table name with `app.blacklistedNotificationTableName` instead of `azure.tableName` (since someday we may have more than one table).
> 
>     * calls `withBlacklist` first -- order of the `helper.withXxx` things is important
> 
>     * adds TASKCLUSTER_* secrets which are configured in CI.
> 
> 
> This should make tests run in CI. Note, though, that I'm seeing the following when running with `DEBUG=*`:
> 
> ```
>   taskcluster-lib-monitor reportError: Error: Internal error, unknown error code: BlacklistedAddress
> undefined is blacklisted level=error tags={} Error: Internal error, unknown error code: BlacklistedAddress
> undefined is blacklisted
>     at /home/dustin/p/taskcluster/libraries/api/src/middleware/express-error.js:69:21
>     at Layer.handle_error (/home/dustin/p/taskcluster/node_modules/express/lib/router/layer.js:71:5)
>     at next (/home/dustin/p/taskcluster/node_modules/express/lib/router/route.js:135:13)
>     at Promise.resolve.then.then.catch (/home/dustin/p/taskcluster/libraries/api/src/middleware/handle.js:30:14)
>     at process._tickCallback (internal/process/next_tick.js:68:7) +1s
> ```
> 
> I am not sure why the tests are succeeding despite the 500 error? Also, that error code should be declared in the APIBuilder (see the `libraries/api` README).

I found the mistake. I was throwing `BlacklistedAddress` but asserting `InternalServerError` in the tests hence they were passing. It's all fixed now

services/notify/schemas/v1/modify-notification-address.yml
```diff
@@ -0,0 +1,12 @@
+$schema:  "/schemas/common/metaschema.json#"
+title:  "Request For Modifying Address"
```
djmitche: This whole file can be removed.

services/notify/schemas/v1/notification-address-list.yml
```diff
@@ -0,0 +1,20 @@
+$schema:  "/schemas/common/metaschema.json#"
+title:  "List of notification adresses"
+description:  |
+  List of notification addresses.
```
djmitche: List of *blacklisted* notification addresses

services/notify/src/api.js
```diff
@@ -50,7 +66,19 @@ builder.declare({
...
+    notificationAddress: req.body.routingKey,
+  };
+  // Ensure that the address is not in the blacklist
+  let response = await this.BlacklistedNotification.load(notificationAddress, true);
+  if(!response) {
+    await this.notifier.pulse(req.body);
+    res.sendStatus(200);
+  } else {
+    return res.reportError('BlacklistedAddress', `${req.body.address} is blacklisted`, {});
+  }
```
djmitche: Minor stylistic thing: it might be better to reverse the arms of this conditional.  Then the code to handle blacklisting (the `load` and the `reportError`) will be close together, and separated from the "normal path" of an address that is not blacklisted.

services/notify/src/api.js
```diff
@@ -86,6 +114,110 @@ builder.declare({
...
+  title: 'Blacklist Given Address',
+  description: [
+    'Add the given address to the notification blacklist. The address',
+    'can be of either of the three supported address type namely pulse, email',
+    'or IRC(user or channel). Addresses in the blacklist will be ignored',
+    'by the notification service.',
+  ].join('\n'),
+}, async function(req, res) {
+  // The address to blacklist
+  let address = Object.assign({}, {
```
djmitche: I think the `Object.assign` isn't required here?  Just `= { notificationType: .., notificationAddress: ..}` should create a new object for you.

services/notify/src/api.js
```diff
@@ -86,6 +114,110 @@ builder.declare({
...
+    'If you are not interested in listing all the members at once, you may',
+    'use the query-string option `limit` to return fewer.',
+  ].join('\n'),
+}, async function(req, res) {
+  const continuation = req.query.continuationToken || null;
+  const limit = Math.min(parseInt(req.query.limit || 1000, 10), 1000);
+  const query = await this.BlacklistedNotification.scan({}, {continuation, limit});
+
+  return res.reply({
+    addresses: query.entries.map(address => address._properties),
```
djmitche: It's probably better to construct this explicitly, rather than relying on `_properties` (which is internal, after all).  So `address => {notificationType: address.notificationType, ..}`

services/notify/src/main.js
```diff
@@ -52,6 +54,22 @@ const load = loader({
...
+    setup: ({cfg, monitor, process}) => data.BlacklistedNotification.setup({
+      tableName: cfg.app.blacklistedNotificationTableName,
+      credentials: sasCredentials({
+        accountId: cfg.azure.accountId,
+        tableName: cfg.app.blacklistedNotificationTableName,
+        rootUrl: cfg.taskcluster.rootUrl,
+        credentials: cfg.taskcluster.credentials,
+      }),
+      cryptoKey: cfg.azure.cryptoKey,
+      signingKey: cfg.azure.signingKey,
```
djmitche: These two can be omitted, since we are neither encrypting nor signing any columns in this table.

services/notify/test/api_test.js
```diff
@@ -17,6 +28,17 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['aws'], function(mock, s
       assert.deepEqual(m.payload.message, {test: 123});
       assert.deepEqual(m.CCs, ['route.notify-test']);
     });
+    // Add an address to the blacklist
```
djmitche: Let's put this in a separate test case.

services/notify/test/api_test.js
```diff
@@ -29,6 +51,22 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['aws'], function(mock, s
     helper.checkEmails(email => {
       assert.deepEqual(email.delivery.recipients, ['success@simulator.amazonses.com']);
     });
+    // Add an address to the blacklist
```
djmitche: separate test case, please

services/notify/test/api_test.js
```diff
@@ -60,5 +98,75 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['aws'], function(mock, s
       assert.equal(body.channel, '#taskcluster-test');
       assert.equal(body.message, 'Does this work?');
     });
+    // Add an irc-channel address to the blacklist
```
djmitche: separate test case, please

services/notify/test/api_test.js
```diff
@@ -60,5 +98,75 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['aws'], function(mock, s
...
+      notificationType: 'irc-channel',
+      notificationAddress: '#taskcluster-test',
+    });
+    // Ensure sending notification to that address fails with appropriate error
+    try {
+      await helper.apiClient.irc({message: 'Does this work?', channel: '#taskcluster-test'});
+    } catch(e) {
+      assert(e.code, 'BlacklistedAddress');
+    }
+    // Repeat for a blacklisted user
```
djmitche: separate test case, please

services/notify/test/helper.js
```diff
@@ -387,3 +398,39 @@ exports.withServer = (mock, skipping) => {
...
+      return;
+    }
+
+    if (mock) {
+      const cfg = await exports.load('cfg');
+      exports.load.inject('BlacklistedNotification', data.BlacklistedNotification.setup({
+        tableName: 'BlacklistedNotification',
+        credentials: 'inMemory',
+        cryptoKey: cfg.azure.cryptoKey,
+        signingKey: cfg.azure.signingKey,
```
djmitche: These can be ommitted (sorry, I know I added them!)

services/notify/config.yml
```diff
@@ -12,17 +12,21 @@ defaults:
     maxMessageTime:               600
     # Email address to blacklist, useful to mitigate temporary misuse
     emailBlacklist:               !env:json EMAIL_BLACKLIST
```
djmitche: Let's plan to come back in a later PR and remove this support -- after we've gotten the UI working and transferred the current blacklisted addresses to it!

OjaswinM: Sounds good!

services/notify/schemas/v1/notification-address-list.yml
OjaswinM: I thought to keep it general so it could be used in the future or will that be not needed?

services/notify/src/api.js
OjaswinM: haha you're right, I don't even remember why I used it there

services/notify/schemas/v1/notification-address-list.yml
djmitche: Ah, that's a fair point!

PR #128, opened 2019-01-23
Bug 1441625 - add built-in-workers service

djmitche:
This is an import from https://github.com/taskcluster/taskcluster-built-in-workers and has already been reviewed in https://github.com/taskcluster/taskcluster-built-in-workers/pull/1 so I'm just looking for sign-off that I've imported it correctly.


services/built-in-workers/config.yml
```diff
@@ -0,0 +1,24 @@
+defaults:
+  app:
+    publishMetaData: !env:bool PUBLISH_METADATA
+
+  # TaskCluster credentials
```
imbstack: Big 'C' Taskcluster!

Not actually an issue that needs to be fixed.

services/built-in-workers/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "1.0.0",
+  "private": true,
+  "scripts": {
+    "heroku-prebuild": "echo $SOURCE_VERSION > .git-version",
+    "test": "mocha test/*_test.js",
+    "lint": "eslint src/*.js test/*.js"
+  },
+  "dependencies": {
+    "debug": "^4.1.0",
+    "taskcluster-lib-urls": "^12.0.0"
```
imbstack: This should be removed.

PR #129, opened 2019-01-24
Bug 1516347 - monoimage

djmitche:
This took a while!  This builds all of the services that are currently in the repo into a single docker image.  Microcommits with good messages for your reviewing pleasure.  Some notes:
 * The new image is based on the `node` image, not on an (outdated) Heroku stack
 * This now builds tc-web-server at build time cc @helfi92 (f49761b)
 * We now clone the local working copy of the monorepo.  This is a compromise between just copying the working copy (which can accidentally pull in .gitignore'd files, as I found out) and pulling from github (which for some kinds of changes would require committing and pushing each change).  There's a warning if you have files checked out, but that can be ignored (and it's generally safe to do so when modifying files in `infrastructure/builder`, making development of the tool efficient).
 * I tried to pull docs in here, and failed badly:
   * Merging docs into the monorepo didn't work, as gulp doesn't work on node 10
   * I worked for a bit on a thing that would clone the docs repo and include its static output in the monoimage, but the complexity of that approach was so high that it wasn't worth it, since the tc-docs repo will go away soon.

Tc-builder is still pretty complicated.  It's dealing with a bunch of special cases, many of which will go away or get simpler once we're off Heroku and have more stuff merged to the monorepo.  I expect a lot of this to get refactored once or twice more in the next few months, so as long as it can support our near-term needs (cc @owlishDeveloper and @imbstack) I'm not terribly concerned with the tool's architecture.  For example, `infrastructure/builder/taskcluster-spec/build.yml` is basically just a list of some of the special cases at this point.  Once they're gone, there will be no more "spec" to think about.


infrastructure/builder/src/build/monoimage.js
```diff
@@ -123,6 +123,33 @@ const generateMonoimageTasks = ({tasks, baseDir, spec, cfg, cmdOptions}) => {
...
+        utils,
+        baseDir,
+      });
+    },
+    entrypoints: async (requirements, utils, procs) => {
+      // since we ran `yarn build` already, there's no need to run it again
+      // on startup, so remove it from the entrypoint commands
+      Object.keys(procs).forEach(process => {
+        if (process.startsWith('web-server/')) {
+          procs[process] = procs[process].replace('yarn build && ', '');
```
imbstack: yeessssssss

services/web-server/src/index.js
```diff
@@ -46,6 +36,24 @@ const load = loader(
         }),
     },
 
+    docs: {
```
owlishDeveloper: What does this section do?

services/web-server/src/index.js
```diff
@@ -20,18 +21,7 @@ const load = loader(
   {
     cfg: {
       requires: ['profile'],
-      setup: ({ profile }) => {
-        const cfg = config({ profile });
-
-        // apply some sanity-checks
-        assert(cfg.server.port, 'config server.port is required');
```
owlishDeveloper: these are not needed anymore?

services/web-server/src/index.js
owlishDeveloper: ah, ok - it adds the web-server section to the documentation on the website

infrastructure/builder/src/build/monoimage.js
```diff
@@ -0,0 +1,492 @@
...
+ *    - generate an "entrypoint" script to allow things like "docker run <monoimage> service/web"
+ *  - Generate the docs for each service, bundling them into a single docsDir
+ *  - Build the docker image containing both the app (/app) and built docs directory (/docs)
+ *
+ *  All of this is done using a "hooks" approach to allow segmenting the various oddball bits of
+ *  this process by theme.
+ */
+const generateMonoimageTasks = ({tasks, baseDir, spec, cfg, cmdOptions}) => {
+  const packageJson = JSON.parse(fs.readFileSync(path.join(appRootDir.get(), 'package.json')));
+  const nodeVersion = packageJson.engines.node;
```
owlishDeveloper: 👍

services/web-server/src/index.js
djmitche: I moved them down to the server section, since those configs aren't set for writeDocs (and aren't needed in that case)

djmitche:
Huh, does webpack pass its arguments to the started process?  Is there any way to prevent that?  I can't see a quick answer to that from the docs.

The docker-image startup process takes a single command-line argument telling what sort of process to run (either `server` or `writeDocs` in this case).  That's passed from the Procfile.  So I think we could do either of two things:
 * define the index entrypoint for webpack to load `loader` from a second "main" entrypoint and call `loader({process: 'web', profile: process.env.NODE_ENV})`, then have the Procfile run `node build/main.js web`; or
 * update the command-line handling in `src/main.js` to look for a `--process` argument, defaulting to `web`, and ignore any other arguments, then update the Procfile to run `node build/index.js --process web` and `node build/index.js --process writeDocs`.

helfi92:
> Huh, does webpack pass its arguments to the started process? Is there any way to prevent that?

I'm not sure but one option we could probably do is define all the arguments in `.neutrinorc.js`. `.neutrinorc.js` gets called from `webpack.config.js`.

djmitche:
Can you show what that would look like?

helfi92:
> Huh, does webpack pass its arguments to the started process? Is there any way to prevent that?

In order to remove webpack options, I was thinking something like:

```diff
diff --git a/services/web-server/.neutrinorc.js b/services/web-server/.neutrinorc.js
index 40babcc97..13e1234f7 100644
--- a/services/web-server/.neutrinorc.js
+++ b/services/web-server/.neutrinorc.js
@@ -1,6 +1,8 @@
 const nodeExternals = require('webpack-node-externals');
 const { join } = require('path');
 
+const isEnvDevelopment = process.env.NODE_ENV === 'development';
+
 module.exports = {
   use: [
     ['@neutrinojs/node', {
@@ -13,6 +15,8 @@ module.exports = {
     }],
     (neutrino) => {
       neutrino.config.merge({
+        mode: isEnvDevelopment ? 'development' : 'production',
+        watch: isEnvDevelopment,
         externals: [
           nodeExternals({
             whitelist: [/^webpack/]
diff --git a/services/web-server/package.json b/services/web-server/package.json
index 2233beb57..3bd80e2f0 100644
--- a/services/web-server/package.json
+++ b/services/web-server/package.json
@@ -6,8 +6,8 @@
   "author": "Eli Perelman <eli@eliperelman.com>",
   "license": "MPL-2.0",
   "scripts": {
-    "build": "webpack --mode production",
-    "start": "webpack --watch --mode development",
+    "build": "NODE_ENV=production webpack",
+    "start": "NODE_ENV=development webpack",
     "lint": "eslint --cache --format codeframe --ext mjs,js src test",
     "test": "mocha --require mocha.config.js --recursive"
   },
```

Would this work?

djmitche:
Ah, thanks!  I think that won't quite work, but it gets me on the right track.  Please hold, an agent will have an updated PR in a few moments ;)

djmitche:
rebased and one commit added to address the above issue

djmitche:
@helfi92 thanks for the help :)

PR #148, opened 2019-01-25
Fix: Worker types view modal not opening

helfi92:
A [commit](https://github.com/taskcluster/taskcluster-web/commit/508ff89e65304d8e67723f039b62502c5ed91dbe) later added after the migration of tc-web to the monorepo has started.


helfi92:
I don't think I asked for review :)

djmitche:
Hah, GitHub did it did automatically, I forgot :)

PR #154, opened 2019-01-25
Add configuration file for web-ui

owlishDeveloper:


infrastructure/terraform/web-ui.tf
```diff
@@ -0,0 +1,26 @@
+module "web_ui_secrets" {
+  source = "modules/service-secrets"
+  project_name = "taskcluster-ui"
```
imbstack: isn't it called taskcluster-web?

owlishDeveloper: I don't want to call it taskcluster-web in Kubernetes, because it leads to names like "taskcluster-web-web" in there, so I decided to go with the dir name

imbstack: We already have a secrets_secrets in there somewhere iirc. May as well continue the trend. It's just nice to map it to the name of the package if we can.

owlishDeveloper: The package in the monorepo resides in the `ui` dir? I was hoping to have more distinguishable names than `web` and `web-server`

djmitche: I think a thorough renaming to `taskcluster-ui` could make sense, if @helfi92 is ok with it.

helfi92: I’m okay with both -web and -ui.

infrastructure/builder/src/build/service/web-ui-nginx-site.conf
```diff
@@ -26,7 +26,7 @@ gzip_types
 server {
   listen 80;
   server_name localhost;
-  root /app/build;
+  root /app/ui/build;
 
   location / {
     try_files $uri @rewrites;
```
helfi92: The docs entry file is docs.html. We should do something similar to https://github.com/taskcluster/taskcluster/blob/32673bfefcb072fecce052d814a2d1738f5be5df/ui/.neutrinorc.js#L38-L40 so that when a user navigates to a url that starts with `/docs`, we use `docs.html` instead of the default `index.html`. I think this can be done under `location @rewrites` directive.

_Note: The redirect in `neutrinorc.js` is only for the dev environment. That's why we still need nginx to do it for us in production._

infrastructure/builder/src/build/monoimage.js
```diff
@@ -150,6 +150,9 @@ const generateMonoimageTasks = ({tasks, baseDir, spec, cfg, cmdOptions}) => {
     },
   });
 
+  // TODO: add hooks push for web-ui
+  // TODO: add entry point for docker run for web-ui
+
```
djmitche: It might be better to record these in a bugzilla bug or github issue than in the source..

infrastructure/builder/taskcluster-spec/build.yml
```diff
@@ -32,16 +32,16 @@ repositories:
...
 
-# tools
-- name: tools
+# tc-web ui
+- name: ui
   kind: service
   docs:
     tier: integrations
-    projectName: taskcluster-tools
+    projectName: taskcluster-web
```
djmitche: or taskcluster-ui??

infrastructure/builder/src/build/service/web-ui-nginx-site.conf
owlishDeveloper: So we already have

```
location @rewrites {
    rewrite ^(.+)$ /index.html last;
  }
```

in the nginx configuration. do I just add the docs part in there? like

```
location @rewrites {
    rewrite ^(.+)$ /index.html last;
    rewrite ^\/docs/$ /docs.html last;
  }
```

?

owlishDeveloper: and is it something that needs to be done in this PR? Like, do docs exist in tc-web?

infrastructure/builder/src/build/monoimage.js
owlishDeveloper: I wanted to mark the specific place in the code where I need to look while I still remember it 😔

infrastructure/builder/src/build/service/web-ui-nginx-site.conf
helfi92: Yes, docs partially exist in tc-web https://taskcluster-web.netlify.com/docs. tc-web is somewhat a two single page applications (index.html and docs.html).

> do I just add the docs part in there?

Looks good to me but I haven't dabbled much with nginx in the past. You'd have to see if https://owlish.imbstack.com/docs works for you. It should look like that:

screen shot 2019-02-05 at 5 03 57 pm

helfi92: > and is it something that needs to be done in this PR?

I guess it can wait.

infrastructure/builder/taskcluster-spec/build.yml
owlishDeveloper: yeah, most likely so

infrastructure/builder/src/build/service/web-ui-nginx-site.conf
```diff
@@ -34,5 +34,6 @@ server {
 
   location @rewrites {
     rewrite ^(.+)$ /index.html last;
+    rewrite ^\/docs/$ /docs.html last;
```
helfi92: > the docs page just keeps reloading in a loop https://owlish.imbstack.com/docs/

Here's what I think is happening:

nginx is not serving `docs.html` when the user types `https://owlish.imbstack.com/docs` in the URL bar. It's serving `index.html`. `index.html` has a [react-router route](https://github.com/taskcluster/taskcluster-web/blob/ade83943bdf93957f8e55c30c9dfd62e0a4cc536/src/App/routes.jsx#L6-L10) which tries to switch entry points whenever the user navigates to a route from `index.html` that starts with `/docs` (for example clicking a link in tc-web that should navigate the user to the docs site). This calls the [SwitchEntryPoint](https://github.com/taskcluster/taskcluster-web/blob/ade83943bdf93957f8e55c30c9dfd62e0a4cc536/src/views/SwitchEntryPoint/index.jsx) component which reloads the page but since nginx fails to route to `docs.html` after reloading, it continuously refreshes the page.

helfi92: In production, I am using the following rewrite rules on the netlify site: https://github.com/taskcluster/taskcluster-web/blob/ade83943bdf93957f8e55c30c9dfd62e0a4cc536/netlify.toml#L12-L21. Hope that  helps to get the rewrites work with nginx.

owlishDeveloper: Netlify has such a wonderful user interface there, comparing to nginx's regexes :( Thank you Hassan! I'll correct my regexes

PR #155, opened 2019-01-25
Bug 1522914, 1522917 - improvements to reconciliation

djmitche:
This didn't work out-of-the-box when I tried it, so I made some improvements as shown here..

### DIFF
--- a/services/hooks/src/listeners.js
+++ b/services/hooks/src/listeners.js
@@ -125,7 +125,15 @@ class HookListeners {
         await this.client.withChannel(async channel => channel.unbindQueue(fullQueueName, exchange, routingKeyPattern));
       }
       for (let {exchange, routingKeyPattern} of addBindings) {
-        await this.client.withChannel(async channel => channel.bindQueue(fullQueueName, exchange, routingKeyPattern));
+        try {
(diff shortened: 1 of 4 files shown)


services/hooks/src/listeners.js
```diff
@@ -125,7 +125,15 @@ class HookListeners {
         await this.client.withChannel(async channel => channel.unbindQueue(fullQueueName, exchange, routingKeyPattern));
       }
       for (let {exchange, routingKeyPattern} of addBindings) {
-        await this.client.withChannel(async channel => channel.bindQueue(fullQueueName, exchange, routingKeyPattern));
+        try {
+          await this.client.withChannel(async channel => channel.bindQueue(fullQueueName, exchange, routingKeyPattern));
+        } catch (err) {
+          if (err.isOperational) {
```
imbstack: Mind leaving a comment as to what this means w.r.t. taskcluster working?

djmitche: I'm a little unclear on it myself.  amqplib doesn't seem to provide a great way to identify a particular error (e.g., exchange does not exist).  It might be best to refactor this so that it continues to retry this periodically, regardless of the error.

PR #157, opened 2019-01-27
[Bug 1491551] continue - expire limit lastfires to 100 per hookId

Biboswan:
Another alternative solution is to get the hookIds from the Hook table. Then load from `Lastfires`  for  each hookId and delete instead of one long scan in the current patch. Because i'm think there can be memory leak with the present solution.

### DIFF
--- a/services/hooks/package.json
+++ b/services/hooks/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "heroku-prebuild": "echo $SOURCE_VERSION > .git-version",
     "lint": "eslint src/*.js test/*.js",
-    "test": "mocha test/expires_test.js"
+    "test": "mocha test/*_test.js"
   },
   "author": "",
   "license": "MPL-2.0",
--- a/services/hooks/package.json
+++ b/services/hooks/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "heroku-prebuild": "echo $SOURCE_VERSION > .git-version",
     "lint": "eslint src/*.js test/*.js",
-    "test": "mocha test/*_test.js"
+    "test": "mocha test/expires_test.js"
   },
   "author": "",
   "license": "MPL-2.0",
--- a/services/hooks/src/data.js
+++ b/services/hooks/src/data.js
@@ -222,37 +222,39 @@ LastFire.prototype.definition = function() {
 };
 
 LastFire.expires = async function(Hook, now, n = 100) {
-  const hookIds = new Set();
+  const hookKeys = [];
   let count = 0;
   await Hook.scan({},
     {
       handler: async hook => {
-        const { hookId } = await hook.definition();
-        hookIds.add(hookId);
+        const { hookGroupId, hookId } = await hook.definition();
+        hookKeys.push({ hookGroupId, hookId });
       },
     },
   );
 
-  for (hookId of hookIds) {
+  for ({ hookGroupId, hookId } of hookKeys) {
     const hookData = [];
     await this.scan({
       taskCreateTime: Entity.op.lessThan(now),
+      hookGroupId: Entity.op.equal(hookGroupId),
       hookId: Entity.op.equal(hookId)}, {
       limit: 500,
       handler: async item => {
-        const { hookGroupId, taskId, taskCreateTime } = await item.definition();
-        hookData.push({hookGroupId, taskId, taskCreateTime});
+        const { taskId, taskCreateTime } = await item.definition();
+        hookData.push({taskId, taskCreateTime});
       },
     },
     );
     hookData.sort((a, b) =>
       new Date(b.taskCreateTime) - new Date(a.taskCreateTime));
 
-    for(let i=n;i<hookData.length;i++) {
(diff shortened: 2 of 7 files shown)


djmitche:
Good thinking.  I'll take a closer look shortly.

services/hooks/src/data.js
```diff
@@ -221,5 +221,42 @@ LastFire.prototype.definition = function() {
   });
 };
 
+LastFire.expires = async function(Hook, now, n = 100) {
+  const hookIds = new Set();
+  let count = 0;
+  await Hook.scan({},
+    {
+      handler: async hook => {
+        const { hookId } = await hook.definition();
```
djmitche: you'll need to get the hookGroupId as well, and use that in the `this.scan` below, to ensure that hooks with different hookGroupIds but the same hookId don't get confused.

services/hooks/src/data.js
```diff
@@ -221,5 +221,42 @@ LastFire.prototype.definition = function() {
...
+      handler: async item => {
+        const { hookGroupId, taskId, taskCreateTime } = await item.definition();
+        hookData.push({hookGroupId, taskId, taskCreateTime});
+      },
+    },
+    );
+    hookData.sort((a, b) =>
+      new Date(b.taskCreateTime) - new Date(a.taskCreateTime));
+
+    for(let i=n;i<hookData.length;i++) {
```
djmitche: This is more better written as `for (let hook of hookData) {..}`, as it avoids calling `hookData.length` on every iteration.

services/hooks/src/data.js
Biboswan: hookId is unique how there would be confusion? Later in line 243 I extracted the hookGroupId as well before removing the entity. Yeah but maybe using hookGroupId will lead to faster scan in the LastFires table. Would using `query` instead of `scan` in LastFires make it faster as in somewhere it says avoids full table scan.

Biboswan: ah i yeah got it.

services/hooks/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "heroku-prebuild": "echo $SOURCE_VERSION > .git-version",
     "lint": "eslint src/*.js test/*.js",
-    "test": "mocha test/*_test.js"
```
djmitche: Was this temporary?  You can achieve the same thing by running `node_modules/.bin/mocha test/expires_test.js`

services/hooks/src/v1.js
```diff
@@ -354,6 +354,22 @@ builder.declare({
...
+    await this.LastFire.query({
+      hookGroupId: req.params.hookGroupId,
+      hookId: req.params.hookId,
+    }, {
+      handler: async (lastFire) => {
+        await lastFire.remove(false, true);
+      },
+    });
+
+    this.publisher.hookLastFiresDeleted({hookGroupId, hookId});
```
djmitche: I don't think we need to publish a message when this occurs -- nobody really needs to know..

services/hooks/src/v1.js
```diff
@@ -354,6 +354,22 @@ builder.declare({
...
+    // eslint-disable-next-line no-empty
+  } catch (err) {
+    // The hook may have not been ever triggered
```
djmitche: Generally ignoring *all* errors is a mistake -- in this case, it's hiding the fact that you're calling a nonexistent `hookLastFiresDeleted` method.  If there's a specific kind of error you're looking for, you can do something like
```js
} catch (err) {
  if (err.code == 'whatever') {
    // handle it
  } else {
    throw err;
  }
}
```
that way anything unexpected still gets thrown.

services/hooks/test/api_test.js
```diff
@@ -259,10 +259,23 @@ helper.secrets.mockSuite('api_test.js', ['taskcluster'], function(mock, skipping
...
       await helper.hooks.removeHook('foo', 'bar');
       await helper.hooks.hook('foo', 'bar').then(
-        () => { throw new Error('The resource should not exist'); },
+        () => { throw new Error('The resource in Hook Table should not exist'); },
         (err) => { assume(err.statusCode).equals(404); });
       helper.checkNextMessage('hook-deleted', ({payload}) =>
         assume({hookGroupId: 'foo', hookId: 'bar'}).deep.equals(payload));
+      await helper.hooks.listLastFires('foo', 'bar').then(
+        () => { throw new Error('The resource in LastFires table should not exist'); },
+        (err) => { assume(err.statusCode).equals(404); });
```
djmitche: good thinking!

services/hooks/test/expires_test.js
```diff
@@ -0,0 +1,82 @@
+const helper = require('./helper');
+const taskcluster = require('taskcluster-client');
+const assume = require('assume');
+
+suite('expires_test', function() {
```
djmitche: nice tests!

djmitche:
Looks good!  I'll merge this and set up a scheduled task to run it periodically.

PR #163, opened 2019-01-28
Bug 1521798 - run auth tests more quickly

djmitche:
This brings the auth tests down to 20 seconds total.

### DIFF
--- a/services/auth/test/trie_test.js
+++ b/services/auth/test/trie_test.js
@@ -6,6 +6,14 @@ const ScopeSetBuilder = require('../src/scopesetbuilder');
 const trie = require('../src/trie');
 const trietestcases = require('./trietestcases');
 
+// This test suite was designed to test every conceivable combination of
+// inputs to the trie implementation, in an effort to suss out any hidden bugs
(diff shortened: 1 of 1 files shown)


services/auth/test/trie_test.js
```diff
@@ -6,6 +6,14 @@ const ScopeSetBuilder = require('../src/scopesetbuilder');
 const trie = require('../src/trie');
 const trietestcases = require('./trietestcases');
 
+// This test suite was designed to test every conceivable combination of
+// inputs to the trie implementation, in an effort to suss out any hidden bugs
+// by a kind of fuzzing approach.  That's great, but the result is slow (about
+// 4 minutes) and unnecessary unless the trie implementation has beem modified.
+// So by default we run tests only only a smaller set of inputs.  To run the
+// total test suite, just set FULL_TESTS=1.
+const FULL_TESTS = !!process.env.FULL_TESTS;
```
imbstack: Should we call this `AUTH_FULL_TESTS` or make this a services-wide standard? I'm happy either way.

djmitche: It's easy to rename later.  There's a good chance we'll never look at this again.

PR #165, opened 2019-01-30
Bug 1522477 - introduce yarn generate

djmitche:
This sets up `yarn build` then introduces `yarn generate`.  Generating stuff and checking it in results in a pretty nice simplification of the build process!

This creates a lot of files, but most of those are `.md` files that are identical to those elsewhere in the repo, so they're not actually adding anything to the git repository (git deduplicates).  Those will probably eventually migrate into tc-web.


djmitche:
well, that fared badly in CI..

README.md
```diff
@@ -39,7 +39,7 @@ To run tests for a specific package, you can either cd into the directory and `y
 
 ### Build
 
-To build the Taskcluster services, run `./taskcluster-builder build`.
+To build the Taskcluster services, run `yarn build`.
```
imbstack: Nice

infrastructure/builder/src/generate/generators/node-version.js
```diff
@@ -0,0 +1,44 @@
+const fs = require('fs');
+const path = require('path');
+const {readFile, modifyFile, modifyJSON} = require('../util');
+
+/**
+ * Update the node version to match everywhere, treating that in `package.json`
```
imbstack: YESSSSSSSSS

infrastructure/builder/src/generate/generators/node-version.js
```diff
@@ -0,0 +1,44 @@
...
+    if (!nodeVersion || !nodeVersion.match(/[0-9.]+/)) {
+      throw new Error(`invalid node version ${nodeVersion} in package.json`);
+    }
+    utils.step({title: `Setting node version ${nodeVersion}`});
+
+    utils.status({message: '.taskcluster.yml'});
+    await modifyFile('.taskcluster.yml',
+      contents => contents.replace(
+        /^( *node: ')[0-9.]+(')$/m,
+        `$1${nodeVersion}$2`));
```
jhford: Is the plan to keep this as a regular expression search and replace?  Parsing a yaml document as a string can be a little dangerous.

djmitche: I'm not totally happy with it either, but I don't know of tools that can reliably re-write a file including all its comments, etc.  If you know of something, let me know!  At any rate, this will produce diffs that people will see, so if something goes wrong then we'll notice.

djmitche:
(rebased)

djmitche:
I rebased and force-pushed again, having added three things:
 * some fixes to the worker-manager schemas to make them valid against the metaschema (using `required` instead of the unrecognized `requiredProperties`, including `additionalProperties`)
 * special-case that worker-manager's source is in `lib` instead of `src`
 * actually check that `yarn generate` doesn't generate any changes in CI

djmitche:
I'm a bit concerned that my `yarn generate` in `.taskcluster.yml` magically started working.  But, ok?

PR #170, opened 2019-01-31
remove unused file services/notify/src/hack.js

djmitche:


djmitche:
No, it was due to some funkiness with yarn.

PR #172, opened 2019-02-01
Bug 1524187 - Internal Server Errors when calling queue.completeArtifact

jhford:
This changes 500 ISE into 400 BadInput errors

Bugzilla Bug: [1524187](https://bugzilla.mozilla.org/show_bug.cgi?id=1524187)

### DIFF
--- a/services/queue/src/artifacts.js
+++ b/services/queue/src/artifacts.js
@@ -752,7 +752,7 @@ builder.declare({
       });
 
       if (urllib.parse(url).protocol !== 'https:') {
-        throw new Error('Only HTTPS is allowed for artifacts');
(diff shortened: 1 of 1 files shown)


jhford:
These two latest commits should also be deployed.  The first is status code check.  I suspect that the reason the metadata header isn't being set is that the request is failing.  The second commit is only a debugging commit which we should deploy in staging to see what's going on.

djmitche:
I don't think we want to merge and deploy this as-is (removing label).

djmitche:
rebased and the DEBUGGING commit removed..

imbstack:
Conservation of Bugmentum

PR #174, opened 2019-02-01
Bug 1522891 - update triggerHook output to allow no task creation

djmitche:
Bugzilla Bug: [1522891](https://bugzilla.mozilla.org/show_bug.cgi?id=1522891)


djmitche:
Works fine now :)

PR #179, opened 2019-02-04
Remove prettier, eslint-{config, plugin}-prettier

helfi92:
Now that we make use of @mozilla-frontend-infra/react-lint we won't neeed the prettier dependencies here.

Closes https://github.com/taskcluster/taskcluster/pull/160.


helfi92:
Tests are failing. I think I know why. Will work on it now.

PR #183, opened 2019-02-05
[Bug 1522320] Removing all capatialised C from Taskcluster

iFlameing:
Bugzilla Bug: [1522320](https://bugzilla.mozilla.org/show_bug.cgi?id=1522320)


test/meta_test.js
```diff
@@ -40,6 +40,11 @@ suite('Repo Meta Tests', function() {
     assert.equal(taskclusterYml.tasks.$let.node, uiPackageJson.engines.node);
   });
 
+  test('git grep test', async function() {
+    const Taskcluster = await exec('git grep "TaskCluster" ').stdout;
```
jhford: should there also be a check for "Task Cluster" and "Task cluster"?

iFlameing: @jhford should I write a test for "Task Cluster" and "Task cluster"? since the repo name is pretty clear I don't think that it is needed.what you think?

djmitche: I only see one occurrence of that string, but .. yeah, let's test for that too.

iFlameing: ok

djmitche:
It looks like the [meta test failed](https://tools.taskcluster.net/groups/MaZOhUlMRAy5WZiuxRbSZg/tasks/A98Tml8JSsGSLDLEHQSArA/runs/0/logs/public%2Flogs%2Flive_backing.log).  Ignore the `yarn generate` test (I'll fix that).

iFlameing:
@djmitche I am working on it.

iFlameing:
@djmitche How I can open the [webhook_pullrequestfile](https://github.com/taskcluster/taskcluster/blob/master/services/github/test/data/webhooks/webhook.pull_request.close.json)  because when I open it, in vscode or gedit editor it is not saving the modify code.

djmitche:
I suspect that's because it's a large JSON object and those editors must be unable to handle it.

It is also signed by GitHub, so changing it will change the signature.   I will try to find a new version of that file that does not include "TaskCluster" but is properly signed by GitHub.  That explains why the taskcluster-github checks failed, then :/

djmitche:
Also, this will need a rebase.  The conflicts shouldn't be major -- I deleted some uses of "TaskCluster" in READMEs and API docs.

iFlameing:
@djmitche I will do the rebase in next commit and squash all my commits in one. what should I do now? Because Github test is falling because of the webhooks_pullrequest file. I tried in vim also but it remains the same.

djmitche:
I think the best is to just revert that file (so don't include any changes to that file in the PR) and ignore for the moment that the test is failing because of that one instance of "TaskCluster".  I'll take a closer look and find an updated copy of the file and push it here.  I think it's something I need to do because I would like to download it from the GitHub administrative interface, which only I have access to.

iFlameing:
ok

djmitche:
OK, I've fixed the github webhook data in 9228f38.  If you run `git fetch upstream` to fetch changes to this repository, you should be able to `git cherry-pick 9228f38` into your branch.

iFlameing:
ok thanks!

iFlameing:
@djmitche I push a new commit with the added test for "Task cluster" and "Task Cluster". The GitHub meta:test is falling because I use "TaskCluster" for checking in git grep. can you take look at it?

djmitche:
A trick I learned a long time ago to make regular expressions not match themselves is to use [] around one letter, so `git grep Task[C]luster`.  You can include a comment too to tell anyone reading the code why that's done.

iFlameing:
@djmitche when I doing it giving me  Error: Command failed: git grep "Task[C]luster"

djmitche:
I think that's because `git grep` returns a failure status when it doesn't find a match -- which is what we want :)

[The docs](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback) say that a nonzero value is converted to an Error, so you'll need to catch that error and consider it success if `error.code === 1`, and otherwise consider it a failure and show the stdout.

Since there are three patterns now, it might make sense to do this in a loop or with a utility function.

imbstack:
Just wanted to say this is looking great so far. Hope your exams went well!

iFlameing:
@djmitche I push a new commit which solves the test meta. can you finally review it? I didn't have any idea why this netlify test are failing.
@imbstack thanks

iFlameing:
@djmitche when I am doing `assert.equal(error.code === 1)` and running command `yarn run test:meta` in console it giving me some weired error 

UnhandledPromiseRejectionWarning: AssertionError [ERR_ASSERTION]: true == undefined
    at exec.then.catch (/home/iflameing/Documents/github/taskcluster/test/meta_test.js:53:16)

but when I am doing console.log(error.code) it giving me 1

iFlameing:
@djmitche I encounter some problem while testing some test so I asked some question above but now it all resolved. can you merge this now?

test/meta_test.js
```diff
@@ -42,6 +42,17 @@ suite('Repo Meta Tests', function() {
     assert.equal(taskclusterYml.tasks.$let.node, uiPackageJson.engines.node);
   });
 
+  test('git grep test', async function() {
+    const Taskcluster = ["Task[C]luster", "Task [c]luster", "Task [C]luster"]
+    Taskcluster.forEach( value => {
+      exec(`git grep ${value}`).then((res)=>{
+        console.log(res)
```
djmitche: This will echo something to stdout, but does not cause the test to fail:
```
  Repo Meta Tests
    ✓ All packages in CI (381ms)
    ✓ Node version in .taskcluster.yml matches that in package.json
    ✓ Node version for UI matches the rest of the repo
    ✓ git grep test
{ stdout: 'README.md:TaskCluster\n', stderr: '' }
    ✓ Dependencies are not missing/unused (6697ms)


  5 passing (7s)
```

Also, I think I see why you were getting unhandled errors -- nothing is waiting for the `exec` Promise, so mocha carries on with the next test case immediately.  If you were to add an `assert` here, and it failed you would get an unhandled exception:
```
  Repo Meta Tests
    ✓ All packages in CI (376ms)
    ✓ Node version in .taskcluster.yml matches that in package.json
    ✓ Node version for UI matches the rest of the repo
    ✓ git grep test
{ stdout: 'README.md:TaskCluster\n', stderr: '' }
(node:14947) UnhandledPromiseRejectionWarning: AssertionError [ERR_ASSERTION]: Input A expected to strictly equal input B:
+ expected - actual

- undefined
+ 1
    at exec.then.catch (/home/dustin/p/taskcluster/test/meta_test.js:52:16)
(node:14947) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:14947) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

The easiest way to fix this is to use a `for .. of` loop, which allows you to use `await` in the body of the loop:
```js
for (let pattern of Taskcluster) {
  try {
    res = await exec(`git grep '${pattern}'`); // <-- note quoting the pattern for the shell
    // if the grep succeeded, then something matched
    throw new Error(`misspellings found: ${res.stdout}`);
  } catch (err) {
    if (err.code === 1) {
      // git grep found nothing
      return;
    }
    throw err;
  }
}
```

As a general rule, it's best to avoid mixing async/await (and you've used `async` to define this test function) with `.then/.catch`.  And async/await work better with `for` loops than with array methods like `foreach` and `map`.

test/mocha.opts
```diff
@@ -1,4 +1,5 @@
 --ui tdd
+-R spec
```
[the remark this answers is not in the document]
iFlameing: ok, I will revert this but during my debug, without this console.log not outputting anything in terminal so I added it.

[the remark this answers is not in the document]
djmitche: That's strange, but ultimately doesn't matter -- we don't want to be calling `console.log` in a test case, as that doesn't communicate success or failure of the test to mocha.  Better to include the error information in a thrown Error.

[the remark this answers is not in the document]
iFlameing: ok

iFlameing:
@djmitche I amend a new commit which contains all changes as you suggested in the review.

djmitche:
Awesome!  I added some missing semicolons before merging..

iFlameing:
@djmitche thanks I learned a lot in solving just one problem which not so much big.

PR #185, opened 2019-02-05
Add link to readme anchor

imbstack:
This just makes the links a bit nicer imho. Feel free to reject if you like the original behavior.


README.md
```diff
@@ -30,45 +30,45 @@ It is not possible to run a full Taskcluster deployment directly from this repos
...
+    * [Index Service](services/index#readme)
+    * [Hooks Service](services/hooks#readme)
+    * [Auth Service](services/auth#readme)
+* [Taskcluster Web](ui#readme)
+    * [ui/src/components/DateDistance](ui/src/components/DateDistance#readme)
+    * [ui/src/components/StatusLabel](ui/src/components/StatusLabel#readme)
+    * [ui/src/components/SpeedDial](ui/src/components/SpeedDial#readme)
+    * [ui/src/components/Snackbar](ui/src/components/Snackbar#readme)
+    * [ui/src/components/Markdown](ui/src/components/Markdown#readme)
+    * [ui/src/components/Search](ui/src/components/Search#readme)
```
helfi92: These README's are used by the tc-web's [styleguide](https://taskcluster-web-styleguide.netlify.com/). We should probably remove these links from here 😅 We could take care of that in a separate PR though.

djmitche: That should be fine.  There's a few `.filter(..)`'s in the generation code -- this would just mean adding one more.  Feel free!

PR #186, opened 2019-02-05
only show a diff when generate produces one, and don't use a pager

djmitche:
In one of the failures of `yarn generate`, the task timed out with `--more--` at the bottom, suggesting it was running a pager and waiting for input..

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -67,11 +67,9 @@ tasks:
     - name: yarn generate
       command: |
           yarn generate | cat &&
-          git status &&
-          git diff &&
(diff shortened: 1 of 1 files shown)


owlishDeveloper:
The only that concerns me - the failing tests?

djmitche:
Yeah, it looks like the queue tests are flaky.. but the generate errors need further investigation :(

djmitche:
https://unix.stackexchange.com/questions/155046/determine-if-git-working-directory-is-clean-from-a-script seems a bit better..

djmitche:
Yay, green!

PR #188, opened 2019-02-06
Sort README tables of contents

djmitche:


djmitche:
It also avoids spurious diffs from yarn-generate :)

PR #189, opened 2019-02-07
Bug 1523807 - bring tc-client-web into the monorepo

djmitche:
Bugzilla Bug: [1523807](https://bugzilla.mozilla.org/show_bug.cgi?id=1523807)

This moves both JS clients to `clients/`.  Both are released from here, so they keep all of their dependencies in their own `package.json`.  Taskcluster-client is both required by other libraries and services, and requires other libraries for its tests, so it is included as a workspace.  Taskcluster-client-web is neither used by any other service or library, nor uses any of them, and also has its own build process (via neutrino) and lint (with trailing commas forbidden, argh!), so it is not included as a workspace.

Other features:
* check that a packaged tc-client can be required
* consolidate the image used to run tests in a browser
* disable the pre-commit hook (since it would apply to all commits, not just those affecting tc-client-web tests)
* Fix up CODEOWNERS


clients/client-web/package.json
```diff
@@ -4,7 +4,7 @@
   "main": "build/index.js",
   "author": "Eli Perelman <eli@eliperelman.com>",
   "license": "MPL-2.0",
-  "repository": "taskcluster/taskcluster-client-web",
+  "repository": "taskcluster/taskcluster",
```
helfi92: Thoughts on https://github.com/taskcluster/taskcluster/tree/master/clients/client-web instead?

djmitche: Huh, judging from https://yarnpkg.com/en/docs/package-json#toc-repository even this syntax is invalid, but yarn figured it out.  I'll make the suggested change (and to tc-client)

PR #196, opened 2019-02-09
Bug 1523215 - improve index tests

djmitche:
Bugzilla Bug: [1523215](https://bugzilla.mozilla.org/show_bug.cgi?id=1523215)

I managed to reproduce a race condition locally, where the polling detected one index route being added and started the rest of the assertions before the other index routes had been added.

For the timeout in `api_test.js`, I took a wild guess that it was an actual timeout -- from creating a bunch of entities sequentially -- rather than a hang.  I couldn't reproduce that issue locally with or without the fix.

### DIFF
--- a/services/index/src/api.js
+++ b/services/index/src/api.js
@@ -133,8 +133,7 @@ builder.declare({
     'Find a task by index path, returning the highest-rank task with that path. If no',
     'task exists for the given path, this API end-point will respond with a 404 status.',
   ].join('\n'),
-}, function(req, res) {
-  var that = this;
+}, async function(req, res) {
   var indexPath = req.params.indexPath || '';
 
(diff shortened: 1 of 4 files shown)


services/index/src/api.js
```diff
@@ -145,23 +144,25 @@ builder.declare({
...
+    });
+  } catch (err) {
     // Re-throw the error, if it's not a 404
     if (err.code !== 'ResourceNotFound') {
       throw err;
     }
     return res.reportError('ResourceNotFound', 'Indexed task not found', {});
-  });
+  }
+  if (_.isEmpty(tasks.entries)) {
```
owlishDeveloper: `if (!tasks.entries.length)` 😶

PR #197, opened 2019-02-09
Refactor 🤪

owlishDeveloper:
Bugzilla Bug: [1518680](https://bugzilla.mozilla.org/show_bug.cgi?id=1518680)

This is not super-needed because the service is working just fine :) I did this in my free time because when I was working with `login` I found it difficult to read (please keep in mind that I am a very primitive creature, only emojis are easy to read for me and I am confused very easily). The main objective here was reorginizing the code to reduce the amount of functions in the auth0 module; the logic itself is preserved exactly as it was (I hope). The tests suffered from me extensively as well.

`checkIfActive` boolean is worth a honorary mention, because I noticed in the code when we get `user` from `clientId` we don't check if the user is active, and when we get it from request - we do. I wonder why is that.

Look and tell me if you think this really makes the code easier to read, doesn't break anything, has all the important test cases.

### DIFF
--- a/services/login/src/handlers/mozilla-auth0.js
+++ b/services/login/src/handlers/mozilla-auth0.js
@@ -38,11 +38,7 @@ class Handler {
 
     this._managementApiExp = null;
     this._managementApi = null;
-    this._identityProviderId = 'mozilla-auth0';
-  }
-
-  get identityProviderId() {
-    return this._identityProviderId;
+    this.identityProviderId = 'mozilla-auth0';
   }
 
   // Get a management API instance, by requesting an API token as needed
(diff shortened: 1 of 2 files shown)


services/login/src/handlers/mozilla-auth0.js
```diff
@@ -80,12 +76,30 @@ class Handler {
...
+    }
 
-    return a0.getUser({id: userId});
+    // take a user and attach roles to it
+    this.addRoles(userProfile, user);
+
+    return user;
   }
 
+  // exposed method
```
djmitche: It would probably be good to document the Handler API in a README or maybe in a parent class.  But again, maybe not useful if we'll be rewriting all of this code into services/web-server soon, maybe even with a totally different "handler" API.

services/login/src/handlers/mozilla-auth0.js
```diff
@@ -162,7 +145,9 @@ class Handler {
       encodedUserId = encodedUserId.replace(/\|[^|]*$/, '');
     }
 
-    return decode(encodedUserId);
+    const userId = decode(encodedUserId);
+
+    return this.getUser({userId, checkIfActive: false});
```
djmitche: Nice job spotting and replicating the existing behavior, but I think this should be `checkIfActive: true`. The clientId -> user translation is used in `scanner.js` when checking whether existing clients have more scopes than their corresponding user does. It seems that process does not consider whether the user is inactive!

(and then it makes sense to drop the `checkIfActive` argument)

services/login/src/handlers/mozilla-auth0.js
```diff
@@ -172,10 +157,10 @@ class Handler {
     // array as we do not use Auth0 user linking.
     profile.identities.forEach(({provider, connection}) => {
       if (
-        provider === 'ad' && connection === 'Mozilla-LDAP' ||
+        (provider === 'ad' && connection === 'Mozilla-LDAP') ||
```
djmitche: I think I remember the lint rules at the time complaining about this.  They definitely help humans read the expression!

services/login/src/handlers/mozilla-auth0.js
owlishDeveloper: If this is going to be rewritten soon - I'd just leave these in the code as comments

owlishDeveloper:
@djmitche do you have any suggestions left? I'd merge then?

djmitche:
:ship: it

PR #199, opened 2019-02-09
Bug 1515673 - YML changed for v1 in QuickStart

arshadkazmi42:
Bugzilla Bug: [1515673](https://bugzilla.mozilla.org/show_bug.cgi?id=1515673)

### DIFF
--- a/ui/src/views/Quickstart/index.jsx
+++ b/ui/src/views/Quickstart/index.jsx
@@ -27,31 +27,33 @@ import urls from '../../utils/urls';
 import ErrorPanel from '../../components/ErrorPanel';
 import githubQuery from './github.graphql';
 
+
 const initialYaml = {
   version: 1,
-  tasks: [
-    {
+  policy: {
+    pullRequests: ''
+  },
+  tasks: {
+    $match: {
+      taskId: {'$eval': 'as_slugid("pr_task")'},
       provisionerId: '{{ taskcluster.docker.provisionerId }}',
       workerType: '{{ taskcluster.docker.workerType }}',
-      extra: {
-        github: {
-          env: true,
-          events: [],
-        },
-      },
+      scopes: [
+        'secrets:get:project/taskcluster/testing/taskcluster-github'
+      ],
       payload: {
         maxRunTime: 3600,
         image: 'node',
-        command: [],
+        command: []
       },
       metadata: {
         name: '',
         description: '',
-        owner: '{{ event.head.user.email }}',
-        source: '{{ event.head.repo.url }}',
-      },
-    },
-  ],
+        owner:'{{ event.sender.login }}@users.noreply.github.com',
+        source:'{{ event.repository.url }}'
+      }
+    }
+  }
 };
 const initialState = {
   events: new Set([
@@ -210,29 +212,28 @@ export default class QuickStart extends Component {
   renderEditor() {
     const newYaml = safeDump({
       ...initialYaml,
-      access: this.state.access,
-      tasks: [
-        {
-          ...initialYaml.tasks[0],
-          ...{
-            metadata: {
-              ...initialYaml.tasks[0].metadata,
-              name: this.state.taskName,
-              description: this.state.taskDescription,
-            },
-            extra: {
-              github: {
-                events: [...this.state.events].sort(),
+      policy: {
+        pullRequests: this.state.access
+      },
+      tasks: {
+        $match: {
+          [`tasks_for == "github-pull-request" && event["action"] in ${[...this.state.events].sort()}`]: {
+            ...initialYaml.$match,
+            ...{
+              metadata: {
+                ...initialYaml.tasks.$match.metadata,
+                name: this.state.taskName,
+                description: this.state.taskDescription,
               },
-            },
-            payload: {
-              ...initialYaml.tasks[0].payload,
-              command: this.state.commands,
-              image: this.state.image,
+              payload: {
+                ...initialYaml.tasks.$match.payload,
+                command: this.state.commands,
+                image: this.state.image,
+              },
             },
           },
         },
-      ],
+      },
     });
 
     return (
--- a/ui/src/views/Quickstart/index.jsx
+++ b/ui/src/views/Quickstart/index.jsx
@@ -34,7 +34,7 @@ const initialYaml = {
   },
   tasks: {
     $match: {
-      taskId: { '$eval': 'as_slugid("pr_task")' },
+      taskId: { $eval: 'as_slugid("pr_task")' },
       provisionerId: '{{ taskcluster.docker.provisionerId }}',
       workerType: '{{ taskcluster.docker.workerType }}',
       scopes: ['secrets:get:project/taskcluster/testing/taskcluster-github'],
@@ -215,8 +215,8 @@ export default class QuickStart extends Component {
       tasks: {
         $match: {
           [`tasks_for == "github-pull-request" && event["action"] in ${[
-              ...this.state.events
-            ].sort()}`]: {
+            ...this.state.events,
+          ].sort()}`]: {
             ...initialYaml.$match,
             ...{
               metadata: {
--- a/ui/src/views/Quickstart/index.jsx
+++ b/ui/src/views/Quickstart/index.jsx
@@ -27,7 +27,7 @@ import urls from '../../utils/urls';
 import ErrorPanel from '../../components/ErrorPanel';
 import githubQuery from './github.graphql';
 
-const initialYaml = {
+const taskDefinition = {
   version: 1,
   policy: {
     pullRequests: 'collaborators',
@@ -86,7 +86,7 @@ const getMatchCondition = events => {
   return condition;
 };
 
-const getNewYaml = state => {
+const getTaskDefinition = state => {
   const {
     access,
     commands,
@@ -97,22 +97,22 @@ const getNewYaml = state => {
   } = state;
 
   return safeDump({
-    ...initialYaml,
+    ...taskDefinition,
     policy: {
       pullRequests: access,
     },
     tasks: {
       $match: {
         [condition]: {
-          ...initialYaml.tasks.$match,
+          ...taskDefinition.tasks.$match,
           ...{
             metadata: {
-              ...initialYaml.tasks.$match.metadata,
+              ...taskDefinition.tasks.$match.metadata,
               name: taskName,
               description: taskDescription,
             },
             payload: {
-              ...initialYaml.tasks.$match.payload,
+              ...taskDefinition.tasks.$match.payload,
               image,
               command: commands,
             },
@@ -194,12 +194,15 @@ const cmdDirectory = (type, org = '<YOUR_ORG>', repo = '<YOUR_REPO>') =>
   },
 }))
 export default class QuickStart extends Component {
+  initialEvents = new Set([
+    'pull_request.opened',
+    'pull_request.reopened',
+    'pull_request.synchronize',
+  ]);
+
   initialState = {
-    events: new Set([
-      'pull_request.opened',
-      'pull_request.reopened',
-      'pull_request.synchronize',
-    ]),
+    events: this.initialEvents,
+    condition: getMatchCondition(this.initialEvents),
     owner: '',
     repo: '',
     access: 'collaborators',
@@ -211,10 +214,7 @@ export default class QuickStart extends Component {
     taskDescription: '',
   };
 
-  state = {
-    ...this.initialState,
-    condition: getMatchCondition(this.initialState.events),
-  };
+  state = this.initialState;
 
   getInstalledState = debounce(async (owner, repo) => {
     const { data } = await this.props.client.query({
@@ -286,12 +286,12 @@ export default class QuickStart extends Component {
 
     this.setState({
       ...resetState,
-      editorValue: getNewYaml(resetState),
+      editorValue: getTaskDefinition(resetState),
     });
   };
 
   renderEditor() {
-    const newYaml = getNewYaml(this.state);
+    const newYaml = getTaskDefinition(this.state);
 
     return (
       <CodeEditor
--- a/ui/src/views/Quickstart/index.jsx
+++ b/ui/src/views/Quickstart/index.jsx
@@ -27,20 +27,17 @@ import urls from '../../utils/urls';
 import ErrorPanel from '../../components/ErrorPanel';
 import githubQuery from './github.graphql';
 
-
 const initialYaml = {
   version: 1,
   policy: {
-    pullRequests: ''
+    pullRequests: '',
   },
   tasks: {
     $match: {
-      taskId: {'$eval': 'as_slugid("pr_task")'},
+      taskId: { '$eval': 'as_slugid("pr_task")' },
       provisionerId: '{{ taskcluster.docker.provisionerId }}',
       workerType: '{{ taskcluster.docker.workerType }}',
-      scopes: [
-        'secrets:get:project/taskcluster/testing/taskcluster-github'
-      ],
+      scopes: ['secrets:get:project/taskcluster/testing/taskcluster-github'],
       payload: {
         maxRunTime: 3600,
         image: 'node',
--- a/ui/src/views/Quickstart/index.jsx
+++ b/ui/src/views/Quickstart/index.jsx
@@ -158,22 +158,6 @@ const cmdDirectory = (type, org = '<YOUR_ORG>', repo = '<YOUR_REPO>') =>
       ].join(' && '),
     ],
   }[type]);
-const initialState = {
-  events: new Set([
-    'pull_request.opened',
-    'pull_request.reopened',
-    'pull_request.synchronize',
-  ]),
-  owner: '',
-  repo: '',
-  access: 'collaborators',
-  image: 'node',
-  commands: cmdDirectory('node'),
-  commandSelection: 'standard',
-  installedState: null,
-  taskName: '',
-  taskDescription: '',
-};
 
 @hot(module)
 @withApollo
@@ -210,9 +194,26 @@ const initialState = {
   },
 }))
 export default class QuickStart extends Component {
+  initialState = {
+    events: new Set([
+      'pull_request.opened',
+      'pull_request.reopened',
+      'pull_request.synchronize',
+    ]),
+    owner: '',
+    repo: '',
+    access: 'collaborators',
+    image: 'node',
+    commands: cmdDirectory('node'),
+    commandSelection: 'standard',
+    installedState: null,
+    taskName: '',
+    taskDescription: '',
+  };
+
   state = {
-    ...initialState,
-    condition: getMatchCondition(initialState.events),
+    ...this.initialState,
+    condition: getMatchCondition(this.initialState.events),
   };
 
   getInstalledState = debounce(async (owner, repo) => {
@@ -279,8 +280,8 @@ export default class QuickStart extends Component {
 
   handleReset = () => {
     const resetState = {
-      ...initialState,
-      condition: getMatchCondition(initialState.events),
+      ...this.initialState,
+      condition: getMatchCondition(this.initialState.events),
     };
 
     this.setState({
--- a/ui/src/views/Quickstart/index.jsx
+++ b/ui/src/views/Quickstart/index.jsx
@@ -51,15 +51,6 @@ const initialYaml = {
     },
   },
 };
-const initialState = {
-  events: new Set([
-    'pull_request.opened',
-    'pull_request.reopened',
-    'pull_request.synchronize',
-  ]),
-  taskName: '',
-  taskDescription: '',
-};
 const baseCmd = [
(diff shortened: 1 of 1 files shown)


ui/src/views/Quickstart/index.jsx
```diff
@@ -210,29 +209,30 @@ export default class QuickStart extends Component {
...
-            },
-            extra: {
-              github: {
-                events: [...this.state.events].sort(),
+      policy: {
+        pullRequests: this.state.access,
+      },
+      tasks: {
+        $match: {
+          [`tasks_for == "github-pull-request" && event["action"] in ${[
```
owlishDeveloper: here pull request is hard-coded. We want the group of checkmarks `This task should run on` to work: 
* if you tick `Push` (for example) checkmark, this line should have `tasks_for == "push"` in here.
* if you untick all those checkmarks, it should be an empty string
* `event["action"]` is relevant for PRs only. If no pull-request -related checkmark is ticked, this condition should not be there

Please refer to this section of documentation: https://github.com/taskcluster/taskcluster/blob/master/services/github/docs/taskcluster-yml-v1.md#github-events

ui/src/views/Quickstart/index.jsx
```diff
@@ -210,29 +209,30 @@ export default class QuickStart extends Component {
...
-            metadata: {
-              ...initialYaml.tasks[0].metadata,
-              name: this.state.taskName,
-              description: this.state.taskDescription,
-            },
-            extra: {
-              github: {
-                events: [...this.state.events].sort(),
+      policy: {
+        pullRequests: this.state.access,
```
owlishDeveloper: 👍

ui/src/views/Quickstart/index.jsx
```diff
@@ -47,11 +46,11 @@ const initialYaml = {
       metadata: {
         name: '',
         description: '',
-        owner: '{{ event.head.user.email }}',
-        source: '{{ event.head.repo.url }}',
+        owner: '{{ event.sender.login }}@users.noreply.github.com',
```
owlishDeveloper: In json-e, we don't use `{{ }}` syntax. It would be `${ }`

It's possible that `${ }` characters need to be escaped

ui/src/views/Quickstart/index.jsx
```diff
@@ -29,16 +29,15 @@ import githubQuery from './github.graphql';
...
 const initialYaml = {
   version: 1,
-  tasks: [
-    {
+  policy: {
+    pullRequests: '',
+  },
+  tasks: {
+    $match: {
+      taskId: { $eval: 'as_slugid("pr_task")' },
```
owlishDeveloper: For some reason this line and several lines below it do not show up for me on the deployed website. If you run a local build while developing, it will be much easier to see these issues and to debug them.

ui/src/views/Quickstart/index.jsx
arshadkazmi42: What if multiple checkboxes are checked, then how it needs to be handled?
`tasks_for == "github-pull-request" || tasks_for == "github-push"` 
Something like this?

ui/src/views/Quickstart/index.jsx
arshadkazmi42: If I change it to `${}`
it needs event variable. I don't think its available, am I missing something here?

ui/src/views/Quickstart/index.jsx
owlishDeveloper: Yes! The only thing, there are several checkboxes that are all about PRs, so `tasks_for` is `github-pull-request`, but each checkbox has it's own action. I would put PR-related expression in parenthesis, like `(tasks_for ... && event["action"] ...) || tasks_for ...` - something like that.
 
as far json-e goes, you can refer to its boolean expressions docs here https://taskcluster.github.io/json-e/#Language/boolean-operations (and that website contains much more docs)

ui/src/views/Quickstart/index.jsx
owlishDeveloper: `${}` is part of json-e syntax, not javascript's template literal. That's what I mean when I say these need to be escaped. Its meaning is the same as meaning of `{{ }}` syntax. Actually, I think I was wrong about escaping `${}` - should be fine in a string, just like `{{ }}`

ui/src/views/Quickstart/index.jsx
```diff
@@ -210,29 +209,30 @@ export default class QuickStart extends Component {
...
-          ...{
-            metadata: {
-              ...initialYaml.tasks[0].metadata,
-              name: this.state.taskName,
-              description: this.state.taskDescription,
-            },
-            extra: {
-              github: {
-                events: [...this.state.events].sort(),
+      policy: {
```
helfi92: `this.state` is used many times in this method. Let's do:

```jsx
const { access, taskName, // ... } = this.state;


pullRequests: access,

// ...

```

ui/src/views/Quickstart/index.jsx
```diff
@@ -47,11 +46,11 @@ const initialYaml = {
       metadata: {
         name: '',
         description: '',
-        owner: '{{ event.head.user.email }}',
-        source: '{{ event.head.repo.url }}',
+        owner: `{{ event.sender.login }}@users.noreply.github.com`,
```
owlishDeveloper: Please replace `{{ }}`  template with json-e

ui/src/views/Quickstart/index.jsx
```diff
@@ -29,16 +29,15 @@ import githubQuery from './github.graphql';
...
-      },
+  policy: {
+    pullRequests: '',
+  },
+  tasks: {
+    $match: {
+      taskId: { $eval: 'as_slugid("pr_task")' },
+      provisionerId: `{{ taskcluster.docker.provisionerId }}`,
+      workerType: `{{ taskcluster.docker.workerType }}`,
+      scopes: ['secrets:get:project/taskcluster/testing/taskcluster-github'],
```
owlishDeveloper: please remove scopes

ui/src/views/Quickstart/index.jsx
```diff
@@ -47,11 +46,11 @@ const initialYaml = {
       metadata: {
         name: '',
         description: '',
-        owner: '{{ event.head.user.email }}',
-        source: '{{ event.head.repo.url }}',
+        owner: `{{ event.sender.login }}@users.noreply.github.com`,
+        source: `{{ event.repository.url }}`,
```
owlishDeveloper: Please replace `{{ }}`  template with json-e

ui/src/views/Quickstart/index.jsx
```diff
@@ -29,16 +29,15 @@ import githubQuery from './github.graphql';
...
-          events: [],
-        },
-      },
+  policy: {
+    pullRequests: '',
+  },
+  tasks: {
+    $match: {
+      taskId: { $eval: 'as_slugid("pr_task")' },
+      provisionerId: `{{ taskcluster.docker.provisionerId }}`,
```
owlishDeveloper: Please replace `{{ }}`  template with json-e

ui/src/views/Quickstart/index.jsx
```diff
@@ -29,16 +29,15 @@ import githubQuery from './github.graphql';
...
+      workerType: `{{ taskcluster.docker.workerType }}`,
```
owlishDeveloper: Please replace `{{ }}`  template with json-e

ui/src/views/Quickstart/index.jsx
arshadkazmi42: I did all the changes.
This part is giving some error. I changed it to 
`owner: { $eval: 'event.sender.login' }}`@users.noreply.github.com`,`
What will be the json-e format for eval with string contact?

owlishDeveloper: I am noticing you switched from string to template literal (I am talking about JavaScript types)? Can you please switch back? :)

What exactly is the error?

I don't see json-e syntax here. I still see the `{{ }}` templates. Please let me know if you have any questions

arshadkazmi42: Do you mean change it like this
` owner: `${event.sender.login}@users.noreply.github.com`,`?

owlishDeveloper: Yes! Except please don't use template literal - there is no need for it. Use a string literal:

```
owner: '${event.sender.login}@users.noreply.github.com',
```

The part `${event.sender.login}` is not part of JavaScript's template literal, it's json-e language https://github.com/taskcluster/json-e . In the YML it needs to be a string.

arshadkazmi42: Doing like this
`owner: '${event.sender.login}@users.noreply.github.com'`
yarn lint is throwing this error
` Unexpected template string expression (no-template-curly-in-string)`

owlishDeveloper: Disable that linter rule for the affected code

ui/src/views/Quickstart/index.jsx
```diff
@@ -29,16 +29,14 @@ import githubQuery from './github.graphql';
...
-          events: [],
-        },
-      },
+  policy: {
+    pullRequests: '',
+  },
+  tasks: {
+    $match: {
+      taskId: { $eval: 'as_slugid("pr_task")' },
+      provisionerId: '${taskcluster.docker.provisionerId}', // eslint-disable-line no-template-curly-in-string
```
owlishDeveloper: Let's make `provisionerId` and `workerType` just hardcoded like so:

```
provisionerId: 'aws-provisioner-v1'
workerType: 'github-worker'
```

helfi92:
I plan on reviewing this tomorrow :)

ui/src/views/Quickstart/index.jsx
```diff
@@ -207,32 +205,68 @@ export default class QuickStart extends Component {
...
-              command: this.state.commands,
-              image: this.state.image,
             },
           },
         },
-      ],
+      },
     });
 
     return (
```
helfi92: Clicking the "Reset Form & File" button doesn't seem to reset the task definition editor. Would you be able to make it work, please? Happy to help if you have any questions.

screen shot 2019-02-19 at 2 21 04 pm

_Note: It was probably never working :( so I take the full blame._

ui/src/views/Quickstart/index.jsx
```diff
@@ -29,16 +29,14 @@ import githubQuery from './github.graphql';
...
-      provisionerId: '{{ taskcluster.docker.provisionerId }}',
-      workerType: '{{ taskcluster.docker.workerType }}',
-      extra: {
-        github: {
-          env: true,
-          events: [],
-        },
-      },
+  policy: {
+    pullRequests: '',
```
helfi92: The initial editor seem to have pullRequests set to `collaborators`, not `''`.

ui/src/views/Quickstart/index.jsx
```diff
@@ -207,32 +205,68 @@ export default class QuickStart extends Component {
...
-            },
-            extra: {
-              github: {
-                events: [...this.state.events].sort(),
+      policy: {
+        pullRequests: access,
+      },
+      tasks: {
+        $match: {
+          [`${this.getMatchCondition(events)}`]: {
```
helfi92: Th function is being run every time a user types in the editor which is not ideal (editor becomes slow when typing). Given that we only require `events` to determine the condition, we should probably be able to run this method whenever `handleEventsSelection` is triggered. Perhaps something like:

```jsx

  handleEventsSelection = ({ target: { value } }) => {
    const events = new Set(this.state.events);
    const condition = getMatchesCondition(events);

    events.has(value) ? events.delete(value) : events.add(value);

    this.setState({
      events,
      condition,
      editorValue: null,
    });
  };
```

arshadkazmi42:
@helfi92 @owlishDeveloper I had one general question regarding this UI repo.
Whenever there is some error in the code, and on hot reload if fails due to the wrong code.
The hot reload keeps on reloading the error, its like it crashes and tries to reload again. And if left for sometime it freezes the browser and my laptop.
Is this know issue? Or it's some issue happening for me? It makes it very difficult, if there is any error in my code, it just crashes the whole system and I had to restart my laptop

ui/src/views/Quickstart/index.jsx
arshadkazmi42: Doing this, its making the initial value of condition to null. As render function is using `condition` from state and initially that value is null.

arshadkazmi42: Fixed it

ui/src/views/Quickstart/index.jsx
arshadkazmi42: Done

ui/src/views/Quickstart/index.jsx
arshadkazmi42: Done

arshadkazmi42:
Which test it failing? On checking logs, it shows some timeout error for that tests. I can't seem to figure out which exact test is failing

ui/src/views/Quickstart/index.jsx
```diff
@@ -179,8 +207,12 @@ export default class QuickStart extends Component {
 
     events.has(value) ? events.delete(value) : events.add(value);
 
+    // This should be called after events has been modified in above line
```
helfi92: Change the comment to ``` // Note: this should be called after `events` has been modified in the above line ```

arshadkazmi42:
> Love the code improvement! Before we merge, is it possible you forgot to commit the change for https://github.com/taskcluster/taskcluster/pull/199/files#r258191614? Clicking the reset button after modifying the task definition editor is still not working for me.

I did fixed the reset button in the last commit, is it not working for you?

ui/src/views/Quickstart/index.jsx
arshadkazmi42: Update this

helfi92:
> I did fixed the reset button in the last commit, is it not working for you?

Doesn't seem like it. Using the deploy preview https://deploy-preview-199--taskcluster-web.netlify.com/quickstart

kapture 2019-02-25 at 11 34 29

arshadkazmi42:
> Doesn't seem like it. Using the deploy preview https://deploy-preview-199--taskcluster-web.netlify.com/quickstart

@helfi92  Actually issue was at my end, I didn't validated the editorValues on reset, I was validating it only for the checkoxes.
I made the changes now, its working.

ui/src/views/Quickstart/index.jsx
```diff
@@ -103,6 +158,22 @@ const cmdDirectory = (type, org = '<YOUR_ORG>', repo = '<YOUR_REPO>') =>
       ].join(' && '),
     ],
   }[type]);
+const initialState = {
```
owlishDeveloper: rather than having a global variable with component's state, it's a much better idea to save state in the component's class

arshadkazmi42: I tried moving the initialState inside class as static variable, but on accessing it in constructor using `QuickStart.initialState` it returns `undefined`.
I there something I need to do while defining static variables in class?

> When I press it, the `${match}` condition becomes empty, although pull-request-related flags are set.

I have fixed this part, the method I am using for this fix, might needs a review.
Also, this https://tools.taskcluster.net/quickstart does not supports resetting editor values which are changed manually.
In this I am also resetting value of editor to default if its getting change. Due to which the implementation is a bit different from the old one

owlishDeveloper: Usually class props are accessed using `this` keyword: `this.initialState`

It doesn't have to be static variable (in fact, I'm not sure if the transpiler we use here even allows for keyword `static`), but I would like to see it defined inside the class

> Also, this https://tools.taskcluster.net/quickstart does not supports resetting editor values which are changed manually.

You can add that functionality on top of the existing code, I believe

arshadkazmi42: > Usually class props are accessed using `this` keyword: `this.initialState`

I will work on this change.
Resetting I implemented.

arshadkazmi42: > It doesn't have to be static variable (in fact, I'm not sure if the transpiler we use here even allows for keyword `static`), but I would like to see it defined inside the class

I was using it as static, as it was similarly getting used in https://github.com/taskcluster/taskcluster-tools/blob/master/src/views/QuickStart/index.jsx#L94

ui/src/views/Quickstart/index.jsx
```diff
@@ -135,15 +194,26 @@ const cmdDirectory = (type, org = '<YOUR_ORG>', repo = '<YOUR_REPO>') =>
...
     commandSelection: 'standard',
     installedState: null,
+    taskName: '',
+    taskDescription: '',
+  };
+
+  state = {
+    ...this.initialState,
+    condition: getMatchCondition(this.initialState.events),
   };
```
helfi92: This would make more sense if it was:

```jsx
state = this.initialState;
```

I think we can achieve that by doing something like:

```jsx
initialEvents = new Set([
  'pull_request.opened',
  'pull_request.reopened',
  'pull_request.synchronize',
]);

initialState = {
  events: this.initialEvents,
  condition: getMatchCondition(this.initialEvents),
  owner: '',
  repo: '',
  access: 'collaborators',
  image: 'node',
  commands: cmdDirectory('node'),
  commandSelection: 'standard',
  installedState: null,
  taskName: '',
  taskDescription: '',
};

state = this.initialState;
```

ui/src/views/Quickstart/index.jsx
```diff
@@ -204,36 +279,19 @@ export default class QuickStart extends Component {
...
   handleReset = () => {
-    this.setState(initialState);
+    const resetState = {
+      ...this.initialState,
+      condition: getMatchCondition(this.initialState.events),
+    };
+
+    this.setState({
+      ...resetState,
+      editorValue: getNewYaml(resetState),
```
helfi92: I have mixed feelings about the name `getNewYaml`. Let's find a more descriptive name.

arshadkazmi42: How about `formatYml` or `generateYml` ?

helfi92: What is that we are generating/formatting with `formatYml` and `generateYml`? For example, if a function is responsible in returning the yml of a task definition, then I would probably name it something like `getTaskDefinition`. Let me know what you think.

arshadkazmi42: I was thinking about it in the other way, basically we have initialYml value defined as json
[Line 30](https://github.com/taskcluster/taskcluster/pull/199/files/66a5d75339efcf46929e1410f3f7cf8137abb546#diff-ab453666b30fa170ad0870e42fc3b8deR30)

And the json is getting formatted into YML, in `getNewYml` function, based on this thinking, I was suggesting, `formatYml` or `generateYml`.

helfi92: I think we can also rename `initialYml` to something else. The only reason we have `initialState` is because state means something to a React component. `initialYml` doesn't say much about what it is.

arshadkazmi42: Ok. So, function name will be `getTaskDefinition`
How about changing `initialYml` to just `yml` ?

helfi92: I don't think `yml` says a lot. What is inside that yml?

arshadkazmi42: It has json structure of the yml which is formatted into a real yml and its rendered in the editor.
https://github.com/taskcluster/taskcluster/pull/199/files#diff-ab453666b30fa170ad0870e42fc3b8deR30

helfi92: What does that json structure represent?

helfi92:
I'll be out for next week. @owlishDeveloper whenever you get the chance, do you mind reviewing this, please. Keep up the great work @arshadkazmi42 :)

ui/src/views/Quickstart/index.jsx
arshadkazmi42: It represents the data which will be converted to yml script and it will be used while running auto scripts to CI/CD

arshadkazmi42:
> Keep up the great work @arshadkazmi42 :)

Thanks @helfi92

ui/src/views/Quickstart/index.jsx
helfi92: > It represents the data which will be converted to yml script

What's inside that data? Is it a task definition? I think once you are able to answer this question, you'll be in a better measure of picking the right variable name :)

arshadkazmi42: Yes. One word for whatever I was trying to explain is `task defintion`.
I think we can name the variable `ymlTaskDefinition` or simply `taskDefinition`

arshadkazmi42: @helfi92 Does these variable names makes sense? I can update the PR with changed variable and function name once confirmation

owlishDeveloper:
I'm in the process of reviewing this. May take a while

PR #200, opened 2019-02-10
Add badges API endpoint for Checks

owlishDeveloper:
Bugzilla Bug: [1526713](https://bugzilla.mozilla.org/show_bug.cgi?id=1526713)


owlishDeveloper:
What do I do about `yarn generate`? 😳

owlishDeveloper:
> The only nits I have is that `completed.svg` and `cancelled.svg` are the no-info-yet badges when really we have the info

I thought if something was cancelled, we don't have the info about the build. What would you suggest?

As for `completed` - it is actually a status, not a conclusion - and for that status, the conclusion _must_ be present, so the svg will be chosen by the conclusion. I added that file for some crazy case where Github adds a new conclusion for which we don't have the badge yet - instead of error it would.... oh, maybe the badge for this should be, in fact the error svg? What do you think?

> Also is there no way to make the badge endpoint smartly decide which sort of status to return without adding a new endpoint?

That was my first implementation :) I like this variant better because 
* it's a bit more readable (if-else logic became a bit awkward in the original endpoint), 
* it will be easier to edit when Statuses are deprecated (simply remove or deprecate the existing endpoint),
* if we decide to redesign the badges or add new ones (not that it's likely to happen, but you never know) - we will have to simply replace the svg file without the need of editing the code.

The downside, of course, is that people who use badges and want to switch to Checks will have to edit their READMEs. I thought it should be fine

services/github/src/api.js
```diff
@@ -389,6 +389,60 @@ builder.declare({
...
+  let instGithub = await installationAuthenticate(owner, this.OwnersDirectory, this.github);
+
+  if (instGithub) {
+    try {
+      const checksuites = await instGithub.checks.listSuitesForRef({
+        owner,
+        repo,
+        ref,
+        app_id: this.cfg.github.credentials.integrationId,
+      });
```
djmitche: What if this is a private repo - are we leaking data about a private repo's status?  I'm not sure if this is worth worrying about or not.. Certainly, the tasks that led to the status are public..

owlishDeveloper: What do you mean by "leaking data"?

djmitche: Are we provided a public way to get data that GitHub otherwise keeps secret?

owlishDeveloper:
> Could we merge this with the existing badge API method, so that it first looks for a checkSuite and, if none is found, shows the information from the Builds table?

Hm, that's an interesting idea - didn't occur to me about the Builds table. I am thinking: will the data always be in the Builds table? For example, people might run tasks only on PRs, and the Builds table wouldn't have anything for `master` branch (or whatever branch)? Also, the information from Builds can be retrieved using `taskGroupId` only. So we would have to procure the `taskGroupId` first

djmitche:
I'm suggesting only using Builds when the checkSuite can't be found -- so only for status reporting.

owlishDeveloper:
We would have to map from `ref` or `branch` to `taskGroupId` - it seems to me it's easier to just call Github's API?

services/github/src/api.js
owlishDeveloper: Don't we already do that with the existing badges API?

djmitche:
Oh, I thought for some reason that the old badges API method used the Builds table.  It looks like it just calls listStatusesForRef instead of listSuitesForRef.  So let me revise what I'm considering:

Define one API badge method which first calls listSuitesForRef and, if it finds nothing, then calls listStatusesForRef.

owlishDeveloper:
> Define one API badge method which first calls listSuitesForRef and, if it finds nothing, then calls listStatusesForRef.

That was what I did at first :) I already replied to that suggestion here:
https://github.com/taskcluster/taskcluster/pull/200#issuecomment-462402938

djmitche:
Oh, I misunderstood what Brian was suggesting!  Sorry, it seems obvious now :)

imbstack:
>  it's a bit more readable (if-else logic became a bit awkward in the original endpoint),

If there are readability issues we can break it out into sub-functions and make it easier to read that way.

> it will be easier to edit when Statuses are deprecated (simply remove or deprecate the existing endpoint),

Not that different to change, right? It is just a bit bigger of a change.

> if we decide to redesign the badges or add new ones (not that it's likely to happen, but you never know) - we will have to simply replace the svg file without the need of editing the code.

That seems possible if we do it with one endpoint too, right? The code is nearly identical, just inside and if/else.

> The downside, of course, is that people who use badges and want to switch to Checks will have to edit their READMEs. I thought it should be fine?

I'd like to keep it the same especially since the `badge` endpoint is a nice name and anything more specific is kinda a leaky abstraction. We may even have to end up having multiple endpoints of this in the future if we keep adding new ways to do checks or something like that.

owlishDeveloper:
> If there are readability issues we can break it out into sub-functions and make it easier to read that way.

In my experience sub-functions make the code harder to read, especially when there's a lot of really small ones

owlishDeveloper:
It's a fair point about the leaky abstraction and nice name. I also do view the need to edit your README as a downside. I'll see what I can do about the readability

owlishDeveloper:
@helfi92 netlify is not happy - do you have any idea what is wrong?

owlishDeveloper:
oh - I did try "retrigger all", but nothing seemed to be happening. Thank you! Did you use the "retrigger all" button in Checks tab, @helfi92 ?

helfi92:
I clicked on the details link of 
screen shot 2019-02-13 at 2 39 43 pm

I sent you an invitation to join the netlify app. Another way I do to fix these kinds of intermittent issues is to amend the last commit message and force push.

owlishDeveloper:
> Also is there no way to make the badge endpoint smartly decide which sort of status to return without adding a new endpoint?

So I thought a bit about this, and to me it seems the there are two ways of "smartly decide":
* add a[n optional] query parameter to the existing endpoint: `/v1/repository/taskcluster/taskcluster/master/badge.svg?reporting=checks`
* make the api first read `.taskcluster.yml` and then decide based on the `reporting` setting in there.

In all honesty, I prefer the first method.

Stuff like "if no statuses are found make a check call" or "if no checks are found make statuses call" will not work because i) if people switch to checks - there _will_ be statuses for their repo; ii) even if people supposedly never used taskcluster with checks (like `servo/servo` repo) - there might be some checksuites (github creates them automatically for installed app). So this way is inaccurate, and just as many network calls in the worst case scenario as reading `.taskcluster.yml`

djmitche:
Does looking forward to a future where there are no statuses change that calculation?

owlishDeveloper:
> Does looking forward to a future where there are no statuses change that calculation?

(Which calculation? I understood your question as "what happens with these options you're suggesting when the statuses are deprecated by Github")

So when statuses are deprecated by Github, we will have to replace all `installation.status. ...` with `installation.checks. ...` equivalent.

For the query parameter - I would make the endpoint work for both having the query parameter `?reporting=checks` and not - the endpoint would just use Checks API in both cases.

For reading the yml - I would remove reading the yml part from the endpoint and just call the Checks APIs.

services/github/assets/in_progress.svg
```diff
@@ -0,0 +1,127 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
helfi92: Not that it matters, but I usually like to feed SVGs through https://jakearchibald.github.io/svgomg/ to reduce their size and get rid of unneeded tags. For example, I was able to have the following reduction on this file :)

screen shot 2019-02-19 at 8 02 48 am

djmitche:
Yes, that's what I was asking.  I have a difficult time understanding why in some circumstances we want to make the change from status to checks really easy for users, to the point of choosing a design with lots of other downsides; while in other cases (here) the design seems to require users to configure status vs. checks in a context where tc-github could just figure it out.  And all of this in a short time window before statuses are deprecated by GitHub.

services/github/assets/in_progress.svg
owlishDeveloper: oh thanks for the tip!

owlishDeveloper:
>  I have a difficult time understanding why in some circumstances we want to make the change from status to checks really easy for users, to the point of

Don't you think I can turn that around 180 degrees? ;) Let's stick to more constructive reasoning?

> short time window

Are we sure about that? It can be years for all I know

PR #202, opened 2019-02-10
Hooks: Disable save button unless some values have changed #390

akshithashetty:
I have compared the two hook objects using the equals method and disabled the save button accordingly. It would be of great help if you could review the code and provide some direction further. Thanks.

Resolves https://github.com/taskcluster/taskcluster/issues/227.

### DIFF
--- a/ui/src/components/HookForm/index.jsx
+++ b/ui/src/components/HookForm/index.jsx
@@ -358,6 +358,13 @@ export default class HookForm extends Component {
       triggerSchemaValidJson
     );
   };
+  isHookValuesSaved = () => {
+    const { hook, previousHook } = this.state;
+
+    if (equals(hook, previousHook)) return true;
+
+    return false;
+  };
 
   handleHookGroupIdChange = e =>
     this.setState({
@@ -673,7 +680,7 @@ export default class HookForm extends Component {
             requiresAuth
             classes={{ root: classes.successIcon }}
             variant="round"
-            disabled={!this.validHook() || actionLoading}
+            disabled={!this.validHook() || this.isHookValuesSaved()}
             onClick={this.handleCreateHook}>
             <ContentSaveIcon />
           </Button>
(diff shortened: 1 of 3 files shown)


ui/src/components/HookForm/index.jsx
```diff
@@ -358,6 +358,13 @@ export default class HookForm extends Component {
       triggerSchemaValidJson
     );
   };
+  isHookValuesSaved = () => {
```
helfi92: Instead of `isHookValuesSaved`, we could simply used `isHookDirty` directly in the `render` method.

ui/src/components/HookForm/index.jsx
```diff
@@ -673,7 +680,7 @@ export default class HookForm extends Component {
             requiresAuth
             classes={{ root: classes.successIcon }}
             variant="round"
-            disabled={!this.validHook() || actionLoading}
+            disabled={!this.validHook() || this.isHookValuesSaved()}
```
helfi92: We still need to check for `actionLoading`.

ui/src/components/HookForm/index.jsx
akshithashetty: Could you please explain this ?

helfi92: From https://github.com/helfi92/taskcluster/commit/e16ae414dccbb983cc89a01cdd196c37038e3030, it is explained in [line 417](https://github.com/helfi92/taskcluster/commit/e16ae414dccbb983cc89a01cdd196c37038e3030#diff-6be593b857389d964b6f77d9ff45a3a8R417). Let me know if it's still confusing to understand.

akshithashetty: Yes that is clear, Thanks

ui/src/components/HookForm/index.jsx
```diff
@@ -690,10 +692,11 @@ export default class HookForm extends Component {
               requiresAuth
               classes={{ root: classes.successIcon }}
               variant="round"
-              disabled={!this.validHook() || actionLoading}
+              disabled={!this.validHook() || actionLoading || !isHookDirty}
               onClick={this.handleUpdateHook}>
               <ContentSaveIcon />
             </Button>
+            <Snackbar open={!isHookDirty} message="Your data has been saved" />
```
helfi92: We only want to show the snackbar when the hook is saved/triggered.

akshithashetty:
Thanks a lot, I have added the changes into the commit.

PR #207, opened 2019-02-11
Omit DEBUG when running a node subprocess

djmitche:
Without this, the DEBUG output obscures the expected stdout..

(found while looking for flaky tests)

### DIFF
--- a/libraries/monitor/test/uncaught_test.js
+++ b/libraries/monitor/test/uncaught_test.js
@@ -50,7 +50,9 @@ suite('Uncaught Errors', () => {
       path.resolve(__dirname, './should_exit_with_error.js'),
       ['--correct'],
       {
-        env: process.env,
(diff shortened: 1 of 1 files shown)


libraries/monitor/test/uncaught_test.js
```diff
@@ -50,7 +50,7 @@ suite('Uncaught Errors', () => {
       path.resolve(__dirname, './should_exit_with_error.js'),
       ['--correct'],
       {
-        env: process.env,
+        env: _.omit(process.env, ['DEBUG']),
```
owlishDeveloper: `env: Object.assign(process.env, {DEBUG: undefined})` ? 🤔

imbstack:
Although all of this code is going away soon hopefully.

djmitche:
I figured, and that's why I asked you for review.

PR #210, opened 2019-02-11
Bug 1524406 - Fix keys for LastFire table

djmitche:
Since this changes the keys, we cannot use the same table.  So this also
updates the table name in the Terraform configuration to point to a
fresh new table.

Bugzilla Bug: [1524406](https://bugzilla.mozilla.org/show_bug.cgi?id=1524406)


djmitche:
Note that https://github.com/taskcluster/azure-entities/pull/45 would have made this bug cause tests to fail, but that PR is not required for the fix.

PR #211, opened 2019-02-11
Bug 1523376 - use crypto-js instead of stealing Hawk's copy

djmitche:
This also adds a check that the library can be require`d in a Node
environment, which helps at least one user (Treeherder) to run some
of its tests in a Node environment.  We do *not* guarantee that any of
the module's functionality actually works outside of a browser.

Bugzilla Bug: [1523376](https://bugzilla.mozilla.org/show_bug.cgi?id=1523376)

### DIFF
--- a/clients/client-web/README.md
+++ b/clients/client-web/README.md
@@ -3,7 +3,7 @@
 **A Taskcluster client library for the browser.**
 
 This client library is generated from the auto-generated API reference. taskcluster-client-web differs from
-[taskcluster-client](https://github.com/taskcluster/taskcluster-client) by providing a version that is compatible
+[taskcluster-client](https://yarnpkg.com/en/package/taskcluster-client) by providing a version that is compatible
 with the browser out of the box and does not require a build step to use.
 
 ## Installation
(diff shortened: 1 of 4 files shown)


clients/client-web/README.md
```diff
@@ -95,6 +95,9 @@ require(['taskcluster-client-web'], ({ Queue }) => {
 </script>
 ```
 
+Note that this module can be imported in a Node.js environment without error, but is not guaranteed to work.
+If you need Taskcluster support in a Node.js environment, see the `taskcluster-client` package.
```
helfi92: A link to taskcluster-client would be neat.

djmitche:
```
error An unexpected error occurred: "https://registry.yarnpkg.com/@apollographql/apollo-tools/-/apollo-tools-0.2.9.tgz: Request failed \"502 Bad Gateway\"".
```
yarn has not been having a great time the last few days

djmitche:
well, it was a README update and tc-client-web passed, so I'm going to merge :)

https://bugzilla.mozilla.org/show_bug.cgi?id=1527375 might help with the yarn issues.

PR #216, opened 2019-02-13
Documentation

helfi92:
In the currently documentation structure, docs are in a separate repo on a separate domain. This pull-request brings it in this repo.

A couple of points:

* [infrastructure/builder/src/generate/generators/docs-tocs.js](https://github.com/taskcluster/taskcluster/pull/216/files#diff-58baaa0db0003fe7ebac8c7b92d2c8c4) creates the documentation table of contents. The JSON output is located in  `ui/src/autogenerated/docs-table-of-contents.json`.

* A good place to start reviewing this PR would be in the [main documentation view](https://github.com/taskcluster/taskcluster/pull/216/files#diff-a52e0ed0a07958b7394b2309676f90d0).

* The `SchemaTable`component  uses https://github.com/taskcluster/react-schema-viewer but without downloading Bootstrap. The relevant bootstrap styles were copied as seen [here](https://github.com/taskcluster/taskcluster/pull/216/files#diff-31f0cbd772a1eeea5191de4d9383fbeaR31). I consider this component temporary until we have [a better component to use](https://bugzilla.mozilla.org/show_bug.cgi?id=1473155).

* The netlify build is presently failing because of https://github.com/taskcluster/taskcluster/pull/216#discussion_r256458452. I'm considering changing the location of the generated docs.

Resolves https://github.com/taskcluster/taskcluster-web/issues/318.
Resolves https://github.com/taskcluster/taskcluster-web/issues/54.

Apologies for the heavy PR. If you are having any issues or not understanding something, please let me know. I would be happy to assist.


package.json
```diff
@@ -14,6 +14,7 @@
     "clients/client"
   ],
   "dependencies": {
+    "mdx-loader": "^2.2.1",
```
helfi92: This dependency is required in the root repo if the generated docs are not located in the `ui/`. I currently have `mdx-loader` defined both in `package.json` and `ui/package.json`...

ui/src/views/Documentation/index.jsx
```diff
@@ -106,63 +182,220 @@ export default class Documentation extends Component {
...
+      }
+
+      return `https://schemas.taskcluster.net/${schemaId.replace(
+        /^\/schemas\//,
+        ''
+      )}`;
+    }
+
+    return schemaId;
+  }
```
helfi92: This was copied from https://github.com/taskcluster/taskcluster-docs/blob/af1b485785a6ff00a7ef507d02ef7806d5fcff0d/lib/utils.js. hard-coding `taskcluster` seems odd a bit in the realm of redeployability. Thoughts on what this should be?

ui/src/views/Documentation/Reference/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+import Typography from '@material-ui/core/Typography';
+import List from '@material-ui/core/List';
+import ListItem from '@material-ui/core/ListItem';
+import ListItemText from '@material-ui/core/ListItemText';
+import Entry from './Entry';
+import Markdown from '../../../components/Markdown';
+import HeaderWithAnchor from '../../../components/HeaderWithAnchor';
+
+@withRouter
+export default class Reference extends Component {
```
helfi92: Idea of a better name? :)

ui/src/views/Documentation/Reference/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+  static propTypes = {
+    /** The JSON object representation of api.json or events.json.  */
+    json: object.isRequired,
```
helfi92: Idea of a better prop name? :)

package.json
helfi92: This is also why netlify is failing since it's not doing `yarn install` in the root directory. Thoughts on removing `generated/docs` and `ui/src/docs` and instead have everything under `ui/docs` (i.e.,  `ui/docs/static` and `ui/docs/generated`)?

helfi92:
A couple of points:

* [infrastructure/builder/src/generate/generators/docs-tocs.js](https://github.com/taskcluster/taskcluster/pull/216/files#diff-58baaa0db0003fe7ebac8c7b92d2c8c4) creates the documentation table of contents. The JSON output is located in  `ui/src/autogenerated/docs-table-of-contents.json`.

* A good place to start reviewing this PR would be in the [main documentation view](https://github.com/taskcluster/taskcluster/pull/216/files#diff-a52e0ed0a07958b7394b2309676f90d0). The latter uses 

* The `SchemaTable`component  uses https://github.com/taskcluster/react-schema-viewer but without downloading Bootstrap. The relevant bootstrap styles were copied as seen [here](https://github.com/taskcluster/taskcluster/pull/216/files#diff-31f0cbd772a1eeea5191de4d9383fbeaR31). I consider this component temporary until we have [a better component to use](https://bugzilla.mozilla.org/show_bug.cgi?id=1473155).

package.json
djmitche: That sounds 110% fine to me.

ui/src/views/Documentation/index.jsx
```diff
@@ -1,35 +1,95 @@
...
   state = {
     error: null,
-    Document: null,
+    Page: null,
     pageInfo: null,
+    referenceJson: null,
   };
 
-  componentDidMount() {
+  async componentDidMount() {
```
djmitche: why async?

ui/src/views/Documentation/index.jsx
```diff
@@ -106,63 +182,220 @@ export default class Documentation extends Component {
     return this.findChildFromRootNode(rootNode) || rootNode;
   }
```
djmitche: I know this has already landed, but I noticed that this `findChildFromRootNode` is doing a depth-first search for the named node.  I think that's done because the "filename" structure doesn't match the TOC structure, so a tree traversal doesn't work.  Let's make those correspond so URL pathname and TOC position match one another!

ui/src/views/Documentation/index.jsx
```diff
@@ -106,63 +182,220 @@ export default class Documentation extends Component {
...
+        (pathname.endsWith('api') || pathname.endsWith('events'));
+      const { loader, path } = readDocFile(
+        `${pathname}.${isJsonFile ? 'json' : 'md'}`
       );
+      const { default: Page } = await loader;
+
+      absolutePath = path;
+
+      // use `isMDXComponent` once https://github.com/mdx-js/mdx/pull/369 is merged
+      if (!Page.prototype) {
```
djmitche: I'm guessing this is meant to distinguish the two different possible return values of readDocFile?  Where a markdown file will return a Page instance, which has a prototype, but a `.json` file returns .. I guess, the JSON data?  So it doesn't have a prototype?  That seems pretty fragile, and certainly not reader-friendly!  Even using `.isMDXComponent` seems to violate encapsulation.  The OO solution would seem to be to wrap both data types in objects (or as components) with a common set of methods.

ui/src/views/Documentation/index.jsx
```diff
@@ -106,63 +182,220 @@ export default class Documentation extends Component {
...
+
   async load() {
     try {
-      const { default: Document } = await this.handleImport(
-        this.props.match.params.path
+      const { params } = this.props.match;
+      const pathname = params.path || 'index';
+      const isJsonFile =
+        pathname.startsWith('reference') &&
+        (pathname.endsWith('api') || pathname.endsWith('events'));
```
djmitche: !?! (see comments above)

ui/src/views/Documentation/index.jsx
```diff
@@ -106,63 +182,220 @@ export default class Documentation extends Component {
...
+    const page = renderToString(
+      <BrowserRouter>
+        <Page components={this.components()} />
+      </BrowserRouter>
     );
+
+    return page
+      .split(DOCS_SCHEMA_REGEX)
+      .map(html =>
+        !isUrl(html) && html.startsWith('<') ? (
```
djmitche: Maybe I'm misunderstanding this, but I think this is splitting on a regexp that matches a group and thus returns an array of `[ unmatched, matchGroup, unmatched, matchGroup, .. ]`, then "guessing" whether each element is unmatched data (because it doesn't look like a URL but does look like HTML in that it begins with `<`) or a URL.  In fact, it adds a dependency to help with that guessing.  But the positions of the unmatched data and the matchGroup are fixed within the array, so you could use `.map((val, i) => (i & 1 == 0) ? val : <SchemaTable ..>)`.

Couldn't this be handled with a `components` entry, anyway?  We are not committed to using `data-render-schema` -- we could easily change all of those to e.g., `<jsonSchema path="/schemas/notify/v1/irc-request.json" />`.  I suspect that would save a lot of difficulty breaking HTML into pieces and re-assembling it with `Fragment` and `dangerouslySetInnerHtml`!

ui/src/views/Documentation/index.jsx
```diff
@@ -106,63 +182,220 @@ export default class Documentation extends Component {
...
+          html
+        ) : (
+          <SchemaTable key={html} schema={html} />
+        )
+      );
   }
 
+  getReferenceEntries = entries => {
+    const projectName = this.props.match.params.path.split('/')[2];
```
djmitche: I'd prefer to not try to parse paths like this -- best to get this from some metadata attached to the object when it was created.  Otherwise, if we move things around then suddenly assumptions like this start generating unexpected behavior.

ui/src/views/Documentation/index.jsx
```diff
@@ -106,63 +182,220 @@ export default class Documentation extends Component {
...
+
+    return Promise.all(
+      entries.map(entry =>
+        ['input', 'output', 'schema'].reduce(async (acc, prop) => {
+          const accumulator = await acc;
```
djmitche: I spent a good ten minutes figuring out how this `reduce` works.  Points for cleverness, points off for readability :)

ui/src/views/Documentation/index.jsx
```diff
@@ -106,63 +182,220 @@ export default class Documentation extends Component {
...
+
+          const schemaName = entry[prop].replace(/#$/, '');
+          let {
+            default: schema,
+          } = await import(/* webpackChunkName: 'Documentation.Schema' */ `../../../../generated/docs/${projectName}/schemas/${schemaName}`);
+
+          schema = this.sanitizeSchema(schema);
+
+          // `dereference` maintains object reference equality
+          const deref = await RefParser.dereference(
```
djmitche: Rather than flatten references, I'd like to eventually have react-schema-viewer able to follow references (shown by expanding a collapsed panel).  That will mean feeding the entire universe of schemas to react-schema-viewer and then telling it the ID of the schema it should render.  That universe is currently included in `generated/references.json` -- it's not huge (69K gzipped).  So, we might want to look forward to a day when all schemas are in a single data structure in memory. For now, I think this is OK..

ui/src/views/Documentation/index.jsx
```diff
@@ -106,63 +182,220 @@ export default class Documentation extends Component {
...
+              resolve: {
+                http: false,
+                file: false,
+                any: {
+                  order: 1,
+                  canRead: /^http*|^\/schemas|^taskcluster:\/schemas/,
+                  read: async (file, callback) => {
+                    const url = new URL(file.url);
+                    const schemaName =
+                      url.hostname === 'schemas.taskcluster.net'
```
djmitche: We can ignore this case, as we won't be using this on the legacy cluster.

ui/src/views/Documentation/index.jsx
```diff
@@ -106,63 +182,220 @@ export default class Documentation extends Component {
...
+          const deref = await RefParser.dereference(
+            schema.$id || schema.id,
+            schema,
+            {
+              resolve: {
+                http: false,
+                file: false,
+                any: {
+                  order: 1,
+                  canRead: /^http*|^\/schemas|^taskcluster:\/schemas/,
```
djmitche: I suspect this will "just work" with abstract schemas if it only supports `/schemas`.

ui/src/views/Documentation/index.jsx
djmitche: Crazy idea: could we handle *everything* as a markdown document, and generate some of them from other data structures during generation?  So for example `api.md` would just be `<reference path="/references/auth/v1/api.json" />`, and there a a `components` entry for `reference` that knows how to render that reference based on data in `generated/references.json`.

Then the entire `/docs` chunk is just displaying a tree of Page objects, each one like the next.

infrastructure/builder/src/generate/generators/docs-tocs.js
```diff
@@ -0,0 +1,204 @@
...
+const { join } = require('path');
+const { readJSONSync, writeJSON } = require('../util');
+const removeExtension = require('../../../../../ui/src/utils/removeExtension');
+
+const DOCS_LOCATIONS = {
+  GENERATED: join('generated', 'docs'),
+  STATIC: join('ui', 'src', 'docs'),
+};
+
+const projectMetadata = {};
```
owlishDeveloper: we can avoid a global variable if we define it right in `readGeneratedDocs` and get rid of `readProjectMetadata` function

infrastructure/builder/src/generate/generators/docs-tocs.js
```diff
@@ -0,0 +1,204 @@
...
+          }),
+        }
+      );
+      const files = { ...mdFiles, ...jsonFiles };
+
+      // Rename key
+      Object.keys(files).forEach(oldKey => {
+        const path = `reference/${metadata.tier}/${projectName}/${oldKey}`;
+
+        delete Object.assign(files, {
```
owlishDeveloper: 🤔 how does this work? what is the intended effect?

package.json
owlishDeveloper: 👍

ui/src/utils/removeExtension.js
```diff
@@ -0,0 +1,2 @@
+// Remove extension from filename
+module.exports = filename => filename.replace(/\.[^/.]+$/, '');
```
owlishDeveloper: I wonder if something like `filename.split('.').pop()` would work instead of this module?

ui/src/views/Documentation/index.jsx
owlishDeveloper: should we `await` `this.load()`?

ui/.neutrinorc.js
```diff
@@ -25,6 +29,7 @@ module.exports = {
       rules: {
         'react/no-access-state-in-setstate': 'off',
         'babel/no-unused-expressions': 'off',
+        'linebreak-style': 'off',
```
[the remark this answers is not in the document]
helfi92: I believe It affected windows users: https://github.com/taskcluster/taskcluster-web/issues/306#issuecomment-446643627.

ui/src/views/Documentation/index.jsx
helfi92: Good catch. It should be removed.

ui/.neutrinorc.js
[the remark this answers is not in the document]
djmitche: Ah, cool.  Git ends up fixing those on commit, but for local editing it makes sense to ignore that lint.

infrastructure/builder/src/generate/util.js
```diff
@@ -75,3 +75,8 @@ exports.gitLsFiles = async () => {
     .stdout.split(/\0/);
   return files;
 };
+
+/**
+ * Remove extension from `filename`
+ */
+exports.removeExtension = filename => filename.replace(/\.[^/.]+$/, '');
```
owlishDeveloper: is there a reason why regex is more preferable than methods from Node's `path` module?

services/queue/docs/task-schema.md
```diff
@@ -9,9 +9,10 @@ docref:       true
 title:        "Task Schema"
 order: 100
 ---
+import SchemaTable from '../../../../src/components/SchemaTable'
```
owlishDeveloper: Is the amount of `../` correct? Also, it feels a bit weird importing something like that, although I can't formulate why. I wonder if we could have a separate small library of components and use it like a regular `taskcluster-lib-etc.`?

infrastructure/builder/src/generate/util.js
helfi92: Good point. We could probably just use the `path` module. This was copy pasted after going through https://github.com/taskcluster/taskcluster/pull/216#discussion_r256600172.

ui/src/views/Documentation/index.jsx
helfi92: @djmitche 

> Crazy idea: could we handle _everything_ as a markdown document

Yes, we definitely can. What should I change so that in addition to api.json and events.json, services also get `api.md` and `events.md`? I'm thinking of having the markdown be:

```
---
title: ...
order: 1
inline: true
---
import Reference from 'path/to/the/Reference/component'
import api from './api.json'

<Reference json={api} />
```

djmitche: I like that!  I think you'd just need to add some support to `yarn generate` to generate such files. 

In their definition, api and exchange (and soon log) references are "sort of" all the same thing.  They are all listed in an array in `manifest.json` and differentiated by getting the schema identified by their $schema property (for example, https://dustin.taskcluster-dev.net/references/queue/v1/exchanges.json points to https://dustin.taskcluster-dev.net/schemas/common/exchanges-reference-v0.json#) and looking at its metadata.name and metadata.version fields.  Perhaps `Reference` can do that.  Or, perhaps `yarn generate` could do that lookup and use distinct components for different `metadata.name`'s (and maybe even helpfully include the version):

```
import ApiReference from '..';

<ApiReference json={api} version={0} />
```

Given all this, it sounds like we could safely generate *everything* under `ui/`, leaving only `references.json` in `generated/` (and I can make that go away if necessary, too).

infrastructure/builder/src/generate/generators/docs-tocs.js
helfi92: This rename the keys in order to include the section, tier, and service name that they belong to. This helps to provide a neat URL in the browser. I will add some comments :)

ui/src/views/Documentation/index.jsx
helfi92: @djmitche How do you propose we do this?

services/queue/docs/task-schema.md
helfi92: > Is the amount of `../` correct?

Yes, the path is actually relative to the generated doc.

> Also, it feels a bit weird importing something like that, although I can't formulate why.

It's weird because 1 ) you probably haven't seen imports in markdown files and 2) `'../../../../` makes your eyes bleed. I admit it's not the prettiest but it really simplifies things in the UI. For example, we won't need to traverse the markdown file to insert react components like the `SchemaTable` component.

ui/src/views/Documentation/index.jsx
djmitche: I think abandoning filenames in favor of a TOC-path would help (that TOC path is probably partly based on the filename, though).  Then the node data structure can either be a tree which is traversed following that path, or a flat Map data structure keyed by the path.

ui/src/components/SchemaTable/index.jsx
```diff
@@ -0,0 +1,175 @@
...
+    };
+  },
+  { withTheme: true }
+)
+/**
+ * Display a SchemaTable
+ */
+export default class SchemaTable extends Component {
+  static defaultProps = {
+    schema: oneOf([string, object]).isRequired,
```
djmitche: I think this only allows a string now?

ui/src/views/Documentation/index.jsx
djmitche: Now that filenames are a little easier to see in the source tree, this makes a lot more sense.  I don't think the DFS itself is a problem (we don't have a lot of pages of docs).  Aside from some of the references paths, the TOC path and filename seem to be well-aligned.  And those references docs will likely change as we host those docs permanently in `ui/docs` the way @owlishDeveloper suggested today.  So I'll mark this resolved.

ui/src/components/SchemaTable/index.jsx
```diff
@@ -82,58 +82,61 @@ import references from '../../../docs/references.json';
...
 export default class SchemaTable extends Component {
+  static propTypes = {
+    schema: string.isRequired,
+    /**
+     * The service name in which the entry belongs to.
+     * Required for a reference document (api or exchanges) because
+     * the `input` and `output` property values of the schema are relative URIs.
+     * The service name helps us find the referenced schema given
+     * the relative path.
+     * */
```
djmitche: :100:

djmitche:
Hm, https://deploy-preview-216--taskcluster-web.netlify.com/docs/reference/platform/queue/docs/worker-interaction doesn't show the image - it seems to have src="/docs/generated/queue/docs/queue-worker-interaction.svg" and of course that location loads docs.html :)

There may be a few more bugs like this lingering.  Since these docs won't immediately go into production, if you'd prefer to land this as-is and fix any remaining such bugs afterward, I think that's fine.

helfi92:
> Oh, and just to check -- were the docs copied from tc-docs into ui/docs recently, or is it possible there are changes to tc-docs that haven't been reflected here?

The docs are old. I will put in a fresh copy from taskcluster-docs before merging.

> Hm, https://deploy-preview-216--taskcluster-web.netlify.com/docs/reference/platform/queue/docs/worker-interaction doesn't show the image

Interesting. It works locally but doesn't seem to work in production. I'm not sure why at the top of my head but I'll look into it.

> This mysterious "API" section now floats to the bottom, but it's still titled strangely:

I'll look into it.

---

I'm planning to most probably merge either friday/monday. @djmitche Should we archive https://github.com/taskcluster/taskcluster-docs once this gets merged?

djmitche:
Sounds like a good plan!  I'll mark my review as approving in anticipation, and maybe @owlishDeveloper can look again before the merge as a final check.

Yes, let's archive tc-docs so we don't accidentally modify the manual there.  It will still be used to generate https://docs.taskcluster.net based on reference data uploaded from heroku.  I think we can leave it in that state for a while -- even until June.

helfi92:
@owlishDeveloper @imbstack I'm planning to merge this today if there are no major complaints: https://deploy-preview-216--taskcluster-web.netlify.com/docs.

PR #228, opened 2019-02-13
[Bug 1527723] Allow setting defaults for loader virtual components

imbstack:
This simplifies the setup at the bottom of the `main.js` in our services and allows for easier requiring of components in modules that otherwise are not loaded by the loader. An example of what I want to do is:

```js
const monitor = require('./main')('monitor').prefix('script');
monitor.info('whatever', {foo: 5});
```
This way we can configure monitor in one place, using configuration loaded in the same way as all of our other components but also allow people to have modules that exist outside the `main.js` setup.

Bugzilla Bug: [1527723](https://bugzilla.mozilla.org/show_bug.cgi?id=1527723)

### DIFF
--- a/libraries/loader/README.md
+++ b/libraries/loader/README.md
@@ -96,7 +96,9 @@ let server = await load('server', {port: 8080});
 ```
 
 Finally, you can specify virtual components, for example you may wish to force
-the caller of `load` to always specify a port.
+the caller of `load` to always specify a port or provide a default. If you
+set the default of the virtual component to a falsy value, you will force
+the user to provide a value for you.
 
 ```js
 let loader = require('taskcluster-lib-loader');
@@ -116,7 +118,7 @@ let load = loader({
       return server;
     }
   },
-  
+
   // Definition of 'express' component, notice that even through the 'server'
   // components setup function returns a promise, `ctx.server` is a server
   // object as resolved.
@@ -130,10 +132,10 @@ let load = loader({
     }
   }
 
-  // Virtual components that must always be specified for `load` to work
-}, ['port']);
+  // Virtual components that can be specified for `load` to work
+}, {'port': 900});
 
-// Create express (here we're forced to specify port)
+// Create express (here we specify a port)
 let server = await load('express', {port: 80});
 ```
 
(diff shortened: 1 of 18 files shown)


imbstack:
Ah yeah, we can do:

```js
const loader = require('./main');
const monitor = await loader('monitor').prefix('foo');
```

Ah yeah, we need to catch the error because of unhandled rejection stuff. I can fix that. Does this general plan make sense though?

libraries/loader/package.json
```diff
@@ -28,7 +28,6 @@
   "dependencies": {
     "assume": "^2.1.0",
     "debug": "^4.1.0",
-    "lodash": "^4.17.11",
```
owlishDeveloper: 😊

libraries/loader/README.md
```diff
@@ -150,18 +152,11 @@ particularly useful for getting a fresh component between tests.
...
 // If this file is executed launch component from first argument
 if (!module.parent) {
-  load(process.argv[2], {
-    process: process.argv[2],
-    profile: process.env.NODE_ENV,
-  }).catch(err => {
-    console.log(err.stack);
-    process.exit(1);
-  });
+  load(process.argv[2]);
```
djmitche: This needs a catch, too.

I kind of like the idea of some syntax to say "this is a top-level invocation of the loader" that will automatically handle errors, as it will remove another 2-3 lines of repetitive code from the services.  But this works too!

libraries/loader/src/loader.js
```diff
@@ -61,142 +54,76 @@ function renderGraph(componentDirectory, sortedComponents) {
...
- * }, ['profile']);
- * ```
- *
- * Then you'll be able to load config as:
- * ```js
- * let config = await load('config', {profile: 'test'});
- * ```
+ * Usage is detailed in README.
  */
-function loader(componentDirectory, virtualComponents = []) {
```
jhford: Rather than breaking backwards compatibility, what about taking either a list or an object and converting lists into null value objects?

```javascript
> let x = {}
undefined
> ['port'].forEach(y => x[y] = null)
undefined
> x
{ port: null }
>
```

imbstack: Nice. Will update that to catch any cases I missed while updating.

libraries/loader/package.json
imbstack: I thought you might like that :)

libraries/loader/README.md
imbstack: Yeah, I would like that eventually too but this was easier to change for now. Created [bug 1527969](https://bugzilla.mozilla.org/show_bug.cgi?id=1527969) for this.

PR #229, opened 2019-02-14
Bug 1527583 - Queue doesn't take jobs from "unscheduled" to "pending"

jhford:
This is missing whatever is needed to start and stop the pollers in the
queue unit tests because I cannot make sense of how polling is managed
in the tests.

Bugzilla Bug: [1527583](https://bugzilla.mozilla.org/show_bug.cgi?id=1527583)

### DIFF
--- a/services/queue/src/dependencyresolver.js
+++ b/services/queue/src/dependencyresolver.js
@@ -45,12 +45,14 @@ class DependencyResolver {
     this._parallelism = options.parallelism;
 
     // do iteration
+    const pollingDelaySecs = this._pollingDelay / 1000;
+    const maxIterationTimeSecs = 600;
     this.iterator = new Iterate({
       maxFailures: 10,
-      waitTime: this._pollingDelay,
-      watchDog: this._pollingDelay,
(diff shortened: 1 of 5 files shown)


services/queue/src/dependencyresolver.js
```diff
@@ -43,67 +44,65 @@ class DependencyResolver {
...
 
-    // Promise that polling is done
-    this._done = null;
-    // Boolean that polling should stop
-    this._stopping = false;
+    // do iteration
+    this.iterator = new Iterate({
+      maxFailures: 10,
+      waitTime: this._pollingDelay,
+      watchDog: this._pollingDelay,
```
djmitche: I think that this wants to be larger than `maxIterationTime`, since we don't call `watchDog.touch()`.

djmitche:
merging as a break-fix

djmitche:
The queue tests were failing to run, which is why they were green despite needing 6e5c4fda9eb46c9182c9d3921497b17f47e45c5e

djmitche:
I think @imbstack has fixed that issue in master already.

PR #233, opened 2019-02-14
Bug 1527428 - fix removal of listeners, and add tests

djmitche:
Bugzilla Bug: [1527428](https://bugzilla.mozilla.org/show_bug.cgi?id=1527428)

### DIFF
--- a/services/hooks/src/listeners.js
+++ b/services/hooks/src/listeners.js
@@ -57,13 +57,17 @@ class HookListeners {
     );
     debug('Listening to hook exchanges');
     this.pulseHookChangedListener = consumer;
-    this.listeners = [];
+    this.listeners = {};
     // Reconcile on start up
     await this.reconcileConsumers();
   }
 
   /** Create a new pulse consumer for a hook */
-  async createListener(hook, queueName) {
+  async createListener(hookGroupId, hookId, queueName) {
+    if (this.listeners[queueName]) {
+      return;
+    }
+
     debug(`${queueName}: creating listener (and queue if necessary)`);
 
     const client = this.client;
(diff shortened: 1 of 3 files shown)


services/hooks/src/listeners.js
```diff
@@ -76,31 +80,29 @@ class HookListeners {
...
+    this.listeners[queueName] = listener;
   }
 
   /** Delete a listener for the given queueName  */
   async removeListener(queueName) {
     debug(`${queueName}: stop listening`);
-    let removeIndex = this.listeners.findIndex(({_queueName}) => _queueName === queueName);
-    if (removeIndex > -1) {
-      const listener = this.listeners[removeIndex];
+    const listener = this.listeners[queueName];
```
djmitche: This is the key bit -- the `_queueName` search was not working (and after all is depending on an internal thing).  The rest is just tests and some minor refactorings.

services/hooks/src/listeners.js
```diff
@@ -57,15 +57,19 @@ class HookListeners {
...
-    this.listeners = [];
+    this.listeners = {};
     // Reconcile on start up
     await this.reconcileConsumers();
   }
 
   /** Create a new pulse consumer for a hook */
-  async createListener(hook, queueName) {
+  async createListener(hookGroupId, hookId, queueName) {
     debug(`${queueName}: creating listener (and queue if necessary)`);
```
imbstack: Should this come after the new check below?

services/hooks/src/listeners.js
```diff
@@ -76,31 +80,29 @@ class HookListeners {
...
+    if (listener) {
       await listener.stop();
-      this.listeners.splice(removeIndex, 1);
     }
   }
 
   haveListener(queueName) {
-    const index = this.listeners.findIndex(({_queueName}) => _queueName === queueName);
-    return index > -1;
+    return !!this.listeners[queueName];
```
owlishDeveloper: Isn't `if (!this.haveListener(queue.queueName))` the same as `if (!this.listeners[queueName])`?

services/hooks/src/listeners.js
djmitche: Probably!

services/hooks/src/listeners.js
djmitche: Yeah, that might be simpler.

PR #236, opened 2019-02-14
Bug 1527427 - fix event emission in lib-pulse

djmitche:
* Be more careful to remove events on client, since it is long-lived
* Allow more than the default 10 events on client and connection, since
  in normal use the number might be much larger.

Bugzilla Bug: [1527427](https://bugzilla.mozilla.org/show_bug.cgi?id=1527427)


djmitche:
I think those are both flakes..

djmitche:
(bugs filed)

PR #238, opened 2019-02-14
Revert "Catch one more case of loader update"

helfi92:
Reverts taskcluster/taskcluster#232

```
/Users/haali/Documents/Mozilla/projects/taskcluster/libraries/loader/src/loader.js:110
      throw new Error(`Target is type ${typeof target}, not string`);
      ^

Error: Target is type undefined, not string
```


imbstack:
The fix is in #240 I believe

PR #241, opened 2019-02-14
Build new docker images, drop some docs-service-related stuff

djmitche:
This includes the addition of taskcluster-ui and the dropping of taskcluster-docs.  I added a few hunks to clean that up -- no longer deploying docs, and setting the node version for the taskcluster-ui build.

But, this also gets some nice new functionality in hooks (notably, expiration) as well as the built-in-workers service.


infrastructure/terraform/ingress.yaml
```diff
@@ -18,10 +18,6 @@ spec:
           backend:
             serviceName: taskcluster-ui
             servicePort: 80
-        - path: '/docs'
-          backend:
-            serviceName: taskcluster-ui
-            servicePort: 80
```
djmitche: This will get handled by the `/` path above

owlishDeveloper: I guess that will happen after Hassan's work is merged?

djmitche: It should be the case already -- everything that's not matched by a more specific route goes to taskcluster-ui, so there's no need to specifically list `/docs` as going to taskcluster-ui.

owlishDeveloper: hm but do we have that path in netlify config then?

owlishDeveloper: anyways, I approved - we can always go back, it's not a big deal

owlishDeveloper: Actually, I just tried deploying with this change - docs don't seem to work for me. Have you tried deploying this on your cluster?

owlishDeveloper:
Also - should we remove

```
  case 'docs':
    assert(!isMonorepo, 'monorepo not supported for this buildtype');
    docsTasks({tasks, baseDir, spec, cfg, name, cmdOptions, repository, workDir});
    break;
```

From `service.js`? In builder.

djmitche:
OK, I'll leave the /docs rule in place.  I think it's redundant, but it doesn't hurt :)

And good catch in service.js.

djmitche:
I'll generate images again to get #253

owlishDeveloper:
>  I think it's redundant, but it doesn't hurt :)

Can you explain more why it's redundant? Have you tried deploying with this configuration on your cluster? Does `/docs` route work for you?

What is the purpose of [this configuration](https://github.com/taskcluster/taskcluster/blob/master/netlify.toml#L17-L20) if the application does the routing itself?

djmitche:
I think it's redundant because the ingress is just mapping path prefixes, and selects the longest one.  So just like `/auth/clients` goes to the UI because nothing else matches it, I expect `/docs` to go to the UI if nothing else matches it.

I had deployed, but not checked /docs.  I was going to test it before asking for your review again.  Reality often disagrees with my understanding :)

owlishDeveloper:
I added configuration for `/docs` to both ingress controller and nginx. I honestly don't remember if I did that because it actually didn't work or because I was just copying our Netlify configuration. Having only nginx doesn't work for sure. I can try removing both and deploying that and see what happens.

djmitche:
I'm pretty sure it *is* required for nginx and netlify, just not the ingress.

djmitche:
I'll experiment and find out.

owlishDeveloper:
> I'll experiment and find out.

Yeah please check that out, because only nginx doesn't seem to work for me. It's possible I got nginx wrong?

owlishDeveloper:
> I think it's redundant because the ingress is just mapping path prefixes

It's mapping them to the services tho, doesn't it? And once the path is mapped to the service - the nginx config kicks in. That's my current understanding.

djmitche:
Yep, it's mapping path prefixes to services, so without the `/docs` clause, it maps anything that starts with `/` to tc-ui (unless it matches something else, but nothing else matches `/docs`)

owlishDeveloper:
Oh, ok. In that case, we probably should take a closer look at the current nginx configuration - it looks like it might be wrong. Please let me know the results of your experiments!

djmitche:
I think the nginx configuration for the ui is correct -- it needs to do sometihng different with `/docs` and with everything else (reply with `docs.html` and `index.html`, respectively).  That's different from the ingress config, though.

Regarding the ingress config: I have data!
 * With `/docs` in the ingress config: https://dustin.taskcluster-dev.net/docs/ works fine
 * Without `/docs` in the ingress config: https://dustin.taskcluster-dev.net/docs/ works fine

Still, I shouldn't have mixed this into a PR that purports to just update docker images.  So I'll land this as-is if you're OK.  And if the data is convincing, I'll make a new PR to drop that ingress config.

owlishDeveloper:
I will try deploying without the ingress config again. I will make sure I use the correct image (the one from this PR should be good?) - the way it didn't work for me, I was experiencing the reloading bug. Now that I think of it, I might have deployed an image with a bad regex in the nginx config.

Let me try that again.

djmitche:
Sounds good, thanks.  Yeah, it's confusing for me too to keep track of which images have which support.

owlishDeveloper:
Yes, with the image from this PR it works! Ingress config is indeed redundant! 🎉 Please feel free to make the PR (or even return that change here if you like)

PR #259, opened 2019-02-15
Rename Blacklist to Denylist

djmitche:
@imbstack suggested using a term with less cultural baggage (at least in the US) than "blacklist".  Seems reasonable! /cc @OjaswinM


djmitche:
I've already updated the testing and production scopes as well as the config in Heroku.

djmitche:
https://notify.taskcluster.net/v1/denylist/list !

PR #263, opened 2019-02-15
Reorganize and correct the flow in jobHandlers

owlishDeveloper:
Bugzilla Bug: [1528329](https://bugzilla.mozilla.org/show_bug.cgi?id=1528329)


owlishDeveloper:
basically the problematic line was this https://github.com/taskcluster/taskcluster/blob/master/services/github/src/pr-allowed.js#L39 but somehow I ended up with that bunch of changes 😳

as per my usual, I removed all the functions I could and got the whole thing in jobHandler. I plan to abstract the getting and parsing `.taskcluster.yml` logic in a separate function because we do that twice

Changes in tests will follow if you think this PR makes sense.

services/github/src/pr-allowed.js
```diff
@@ -1,81 +0,0 @@
...
-    let content = await instGithub.repos.getContents({
-      owner: organization,
-      repo: repository,
-      path: '.taskcluster.yml',
-      ref: branch,
-    });
-    taskclusterYml = yaml.safeLoad(Buffer.from(content.data.content, 'base64').toString());
-  } catch (e) {
-    if (e.code === 404) {
-      return DEFAULT_POLICY;
```
owlishDeveloper: explanation: in this case it should return something like `null` because error 404 means there's no `.taskcluster.yml` in that branch.

services/github/src/pr-allowed.js
```diff
@@ -1,81 +0,0 @@
-const yaml = require('js-yaml');
-
-const DEFAULT_POLICY = 'collaborators';
-
-async function prAllowed(options) {
```
owlishDeveloper: Explanation: This function return a boolean, but there are three cases: `true`, `false` and `neither 😳` (if true, we should check if the author of the PR is collaborator, if false we should comment* to that effect and exit, if neither - we should just exit)

* in case it's PR.opened - in other cases we can just exit

owlishDeveloper:
> Makes sense to me so far! It was kinda easier for me to read when split over multiple files but if that's not true for you then it is fine this way as well.

It was not a question of readability in this case. I added some explanation comments to the deleted module. Basically, My line of reasoning was this:

1. `getRepoPolicy` should return something like `null` if error 404
2. If `getRepoPolicy` returns `null`, what should `prAllowed` return then? This case is not quite binary - we have three branches, and we need to check for them in jobHandler... Maybe there isn't a need for `prAllowed`... deleted it
3. Now we are calling `isCollaborator` and `getRepoPolicy` from jobHandler. 
4. Looks like `getRepoPolicy` just wraps `installation.repo.checkCollaborator` - why don't I just call that.... there goes another function.
5. The code in `getRepoPolicy` looks familiar - we already do large part of that in jobHandler. What if instead of calling this function, I abstract all the getting-and-parsing-the-yml into a new separate function and call that twice from `jobHandler`? So I extract that logic and for now, as a first step, place it into `jobHandler`.
6. At this point, all I have in that module is tiny little function that gets policy from `.taskcluster.yml`. It must feel lonely, so I make it a Handler's method. There goes the whole module 😳

Part 7 is to abstract reading `.taskcluster.uml` - it's not done yet

services/github/src/handlers.js
```diff
@@ -270,6 +269,61 @@ class Handlers {
...
+    return DEFAULT_POLICY;
+  }
+
+  // Try to get `.taskcluster.yml` from a certain ref. Returns either YML, or error
+  async getYml({instGithub, owner, repo, ref}) {
+    let response = await instGithub.repos.getContents({owner, repo, path: '.taskcluster.yml', ref})
+      .catch(e => {
+        if (e.code === 404) {
+          debug(`${organization}/${repository}@${ref} has no '.taskcluster.yml' in default branch. Skipping.`);
+          return null;
```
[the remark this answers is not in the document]
owlishDeveloper: the function is not finished yet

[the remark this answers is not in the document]
owlishDeveloper: corrected the function and the comment

services/github/test/pr-allowed_test.js
```diff
@@ -1,151 +0,0 @@
-const debug = require('debug')('test');
-const helper = require('./helper');
-const assert = require('assert');
-const prAllowed = require('../src/pr-allowed');
-
-suite('allowPullRequests', function() {
```
djmitche: Did these tests get lost in the shuffle?

owlishDeveloper: We test the PR-allowed checking logic in handlers tests already. I can add more test cases if you think there's anything missing

Issue #225, opened 2018-10-26
[UI] always allow collapsing the drawer


davehouse:
Could you add an action to collapse the drawer or leave the accordion button when the screen is wide (then I can choose to collapse it)?

As a user, I would like to hide the navigation when I am looking at content. This would provide more space to view content and prevent horizontal scrolling.

Here is a screenshot showing the limited content space (half of the table is displayed, and I must scroll horizontally to see the other half):
![image](https://user-images.githubusercontent.com/5540936/47586104-6e70d480-d91c-11e8-95a4-45ba674568f5.png)

iFlameing:
@helfi92 I would like to work on this issue :)

helfi92:
Great to hear! Go for it.

helfi92:
Here's how I would tackle the issue:

1. Add a "<" icon button on the right hand side of the "Taskcluster" text

![screen shot 2019-02-16 at 7 53 18 am](https://user-images.githubusercontent.com/3766511/52900108-0e01ed80-31c0-11e9-9bc4-08278cf34052.png)

You can import it in the code via:

```
import ChevronLeftIcon from 'mdi-react/ChevronLeftIcon;
```

2. Use an `IconButton` to wrap the icon so that it's clickable

3. Add a tooltip to the button that says "Collapse side panel"

4. When the user closes the menu, there should be a hamburger menu that allows a user to open the panel.

![screen shot 2019-02-16 at 7 57 19 am](https://user-images.githubusercontent.com/3766511/52900151-8a94cc00-31c0-11e9-8015-a64dc4d0fe4e.png)

_Note: We already have that hamburger menu show up by default when the window size is small._

5. On small screens, when the panel is open, there is a hamburger menu that we can remove and replace with the "<" icon however it should be located to the right of "Taskcluster".

<img width="126" alt="screen shot 2019-02-16 at 8 01 15 am" src="https://user-images.githubusercontent.com/3766511/52900205-1149a900-31c1-11e9-9fb6-6b8eb9a3ddbc.png">

helfi92:
Actually, I think there might be an easier approach which is by simply adding a hamburger menu at the left hand side of "Taskcluster" :) No need for the "<" icon.

Aditya-Kolla:
Can I give this a go?

iFlameing:
@Aditya-Kolla I already working on this. You can work on another issue.

iFlameing:
[...]

PR #271, opened 2019-02-18
Always allow collapsing the drawer

iFlameing:
Regarding issue #225 

@helfi92 can you make some comment on this. The only problem is that when I am making `drawerOpen true`  I am able to show the drawer in desktop mode but I am not able to remove it when I am clicking on any sidebar list. can you take a look or suggest me anything after looking this?

### DIFF
--- a/ui/src/components/Dashboard/index.jsx
+++ b/ui/src/components/Dashboard/index.jsx
@@ -7,7 +7,7 @@ import Drawer from '@material-ui/core/Drawer';
 import AppBar from '@material-ui/core/AppBar';
 import Toolbar from '@material-ui/core/Toolbar';
 import IconButton from '@material-ui/core/IconButton';
-import Hidden from '@material-ui/core/Hidden';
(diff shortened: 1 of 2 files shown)


iFlameing:
@helfi92 thanks, I think it is ready for review. can you review this? :)

iFlameing:
@helfi92 I modify my code as you said and saw that previously for Documentation also we have a temporary variant drawer. But with these changes, we have a permanent drawer as it is important for a newcomer. But there is a problem in mobile view which I didn't like i.e( when we are clicking on the DocsSidebarList, drawer is still open and we have to click on the backdrop to close it.) What do you think and also what should I do now after these changes?

helfi92:
You can try to figure it out. I'm also ok merging after you add the requested changes then we could the issue you pointed out in a different PR.

iFlameing:
@helfi92 I pushed the commit which is required. you can merge this now :) thanks for the help.

PR #272, opened 2019-02-18
Upgrade octokit, tc-github auth stuff

djmitche:
In experimenting with Github logins, it looks like the newer version
does authentication a bit differently, and in particular the
`authenticate` method is deprecated.

@owlishDeveloper I'll test this in my dev environment before putting it up for real review, but FYI..


services/github/src/github-auth.js
```diff
@@ -22,50 +22,49 @@ const retryPlugin = (octokit, options) => {
...
+  let setupToken = () => {
     let inteToken = jwt.sign(
       {iss: cfg.github.credentials.integrationId},
       cfg.github.credentials.privatePEM,
       {algorithm: 'RS256', expiresIn: '1m'},
     );
     try {
-      github.authenticate({type: 'app', token: inteToken});
+      return new Octokit({
+        promise: Promise,
```
owlishDeveloper: I thought we don't need to do this once we don't use BlueBird anymore

owlishDeveloper: oh nevermind

djmitche:
Hm, I'm getting something wrong in setting up my dev environment with a github app.  Even without this patch applied, I get an error when a pull-request-related webhook arrives:

screenshot_2019-02-21 logs viewer - dustin-dev-2 - google cloud platform

I've configured my test hook with what I believe to be identical permissions (before i got that right, the webhooks weren't delivered..).  The webhook secret seems to work OK - payloads are being delivered.  I've configured my deployment with
 * `export TF_VAR_github_integration_id=25812` (the "App ID" from the app settings page)
 * `github_private_pem` is from the "Private Keys" section at the bottom of that page
 * it seems like `github_oauth_token` is unused, so I removed it

I'll keep looking for a bit, but if there's anything obviously wrong in the above please let me know :)

djmitche:
Somehow I managed to use an App ID that was off by 2 (25812 instead of 25814).  Trying again.

djmitche:
OK, that fixed it so things work before this patch.  They still don't work after, but I can debug that.

djmitche:
OK, this version works fine for me (although it has structured logging, so not much output -- but no errors)

services/github/src/github-auth.js
```diff
@@ -22,50 +22,31 @@ const retryPlugin = (octokit, options) => {
...
-      debug('Authentication as app failed!');
-      throw e;
-    }
-    return github;
+
+    return new Octokit({auth: `bearer ${inteToken}`});
+  };
+
+  const getInstallationGithub = async (inst_id) => {
+    const inteGithub = await getAppGithub();
```
owlishDeveloper: Let's not mutate global variables, please 👍

Lovely refactor of this module in general

services/github/app.json
```diff
@@ -1,44 +0,0 @@
-{
```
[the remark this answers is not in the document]
djmitche: It's used by Heroku for the apps it runs for GitHub PRs.  We don't do that anymore, so it's not required.  I only noticed the file because I grepped for `GITHUB_OAUTH_TOKEN`.

services/github/package.json
```diff
@@ -9,7 +9,7 @@
     "checkStaging": "node test/checkStaging.js"
   },
   "dependencies": {
-    "@octokit/rest": "^16.0.5",
+    "@octokit/rest": "^16.15.0",
```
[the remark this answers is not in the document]
djmitche: I think `yarn.lock` still has 16.15.0.  I'd like to keep that, since it's what I tested with.  It was only released 25 days ago, so it's not ancient.

PR #273, opened 2019-02-19
Add Search feature to Secrets

### LINKED ISSUE #175: [taskcluster-web] Add ability to search for secrets
If I want to jump to existing secrets, the new UI doesn't allow me to search for them.

**Potential solution:**
Like the other pages, show a search in the top-right that will filter the results

Aditya-Kolla:
Added Search bar to Secret where you can filter required secrets. 
This closes #175

### DIFF
--- a/ui/src/views/Secrets/ViewSecrets/index.jsx
+++ b/ui/src/views/Secrets/ViewSecrets/index.jsx
@@ -6,6 +6,7 @@ import Spinner from '@mozilla-frontend-infra/components/Spinner';
 import { withStyles } from '@material-ui/core/styles';
 import PlusIcon from 'mdi-react/PlusIcon';
 import Dashboard from '../../../components/Dashboard';
+import Search from '../../../components/Search';
 import SecretsTable from '../../../components/SecretsTable';
 import HelpView from '../../../components/HelpView';
 import Button from '../../../components/Button';
@@ -20,6 +21,9 @@ import secretsQuery from './secrets.graphql';
       secretsConnection: {
         limit: VIEW_SECRETS_PAGE_SIZE,
       },
+      filter: {
+        name: { $regex: '' },
+      },
     },
   }),
 })
@@ -29,6 +33,28 @@ import secretsQuery from './secrets.graphql';
   },
 }))
 export default class ViewSecrets extends Component {
+  state = {
+    secretSearch: '',
+    value: null,
+  };
+
+  handleSecretSearchSubmit = secretSearch => {
+    const {
+      data: { refetch },
+    } = this.props;
+
+    this.setState({ secretSearch });
+
+    refetch({
+      secretsConnection: {
+        limit: VIEW_SECRETS_PAGE_SIZE,
+      },
+      filter: {
+        ...(secretSearch ? { $regex: secretSearch } : null),
+      },
+    });
+  };
+
   handleCreate = () => {
     this.props.history.push('/secrets/create');
   };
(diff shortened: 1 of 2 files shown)


ui/src/views/Secrets/ViewSecrets/index.jsx
```diff
@@ -20,6 +21,9 @@ import secretsQuery from './secrets.graphql';
       secretsConnection: {
         limit: VIEW_SECRETS_PAGE_SIZE,
       },
+      filter: {
+        name: { $regex: '' },
+      },
```
helfi92: No need to define `filter` here.

ui/src/views/Secrets/ViewSecrets/index.jsx
```diff
@@ -65,19 +98,33 @@ export default class ViewSecrets extends Component {
...
       <Dashboard
         title="Secrets"
-        helpView={<HelpView description={description} />}>
+        helpView={<HelpView description={description} />}
+        search={
+          <Search
+            disabled={loading}
+            onSubmit={this.handleSecretSearchSubmit}
+            onChange={this.handleSecretSearchChange}
+            value={value}
```
helfi92: No need to have `value` and `onChange`. You can grab the value directly from `onSubmit`.

ui/src/views/Secrets/ViewSecrets/index.jsx
```diff
@@ -29,6 +33,28 @@ import secretsQuery from './secrets.graphql';
...
+
+    this.setState({ secretSearch });
+
+    refetch({
+      secretsConnection: {
+        limit: VIEW_SECRETS_PAGE_SIZE,
+      },
+      filter: {
+        ...(secretSearch ? { $regex: secretSearch } : null),
+      },
```
helfi92: The reason this is not working is because filter needs to be:

```jsx
filter: secretSearch ? { name: { $regex: secretSearch } } : null,
```

ui/src/views/Secrets/ViewSecrets/index.jsx
```diff
@@ -46,6 +72,13 @@ export default class ViewSecrets extends Component {
           cursor,
           previousCursor,
         },
+        ...(this.state.secretSearch
```
helfi92: This will need to be fixed. See above.

ui/src/views/Secrets/ViewSecrets/index.jsx
Aditya-Kolla: Alright

Aditya-Kolla:
@helfi92 Please check the changes.

Aditya-Kolla:
@helfi92 I fixed the pagination and also the error. 
One thing I noticed was that the filter operation is being done on the secrets in that particular page rather than on all secrets as a whole. Is that alright for now?

Aditya-Kolla:
> Could you give me an example maybe to understand? :)

Let us assume there are 10 secrets per page and a total of 30 secrets. 
When we're using search the filter is only considering the secrets on that page(first 10 secrets).
**Page 1 filtered results**
screenshot from 2019-02-21 22-38-56

When I'm moving to the next page the filter is applied on the next page(next 10 secrets)
**Page 2 filtered results**
screenshot from 2019-02-21 22-39-00

Wouldn't it be better if we could filter all secrets and then paginate the resulting secrets :thinking: 
Could be a good enhancement if that's possible :D

helfi92:
> Wouldn't it be better if we could filter all secrets and then paginate the resulting secrets 🤔
Could be a good enhancement if that's possible :D

We use azure for storing our data and it's super good for handling pagination. We are trying to move away from Azure to Postgres. It should become better in the near future :)

Aditya-Kolla:
>  Thank you

You're welcome :D

Issue #274, opened 2019-02-19
New UI: Getting to task from id in log more complicated


pocmo:
Quite often I need to go through the following flow:

* Open a task that was triggered by a hook, e.g. this one: https://tools.taskcluster.net/groups/eOMZxg6lS-mst_yYNZBTKg
* Go to the log and get the group ID of the scheduled task
* Paste the group id into the box at the top and then look at the new graph.

In the new UI that is way more complicated:
https://taskcluster-web.netlify.com/tasks/groups/eOMZxg6lS-mst_yYNZBTKg

* I click on the task and need to find the link to the log (that's more hidden now in the list of data that shows up)

* In the list of logs I can't jump to the end any more, I think ("follow log"). So I have to grab the toolbar and scroll down manually.

* I copy the task id but I can't jump to the task immediately. I have to go to "view task" first and then search for it there.

* When I finally end up on the task page I do not see the whole group. I have to do one more click to get to the group.

* Getting back to the group from the log of a task is also complicated; it requires me to mostly rely on my browser's back stack - and once back I need to refresh the group list.

pocmo:
Sorry, a bit vague. :)

helfi92:
>  I click on the task and need to find the link to the log (that's more hidden now in the list of data that shows up)

There's a "View Live Log" list item above the Artifacts dropdown which should get you to the Log faster. Perhaps we should make that a bit clearer by adding one of these "LOG" labels:

![screen shot 2019-02-19 at 12 45 51 pm](https://user-images.githubusercontent.com/3766511/53035837-5c272300-3444-11e9-9226-d18148e04891.png)

>  In the list of logs I can't jump to the end any more, I think ("follow log"). So I have to grab the toolbar and scroll down manually.

I'll create a separate bug to bring that back :) We currently only show the "follow log" button for logs that are currently "live".

> * I copy the task id but I can't jump to the task immediately. I have to go to "view task" first and then search for it there.
> * When I finally end up on the task page I do not see the whole group. I have to do one more click to get to the group.

A faster flow would be to click "Task Groups" from the navigation drawer and paste the text in the search bar. The whole process would become:

1.  Click on "View Live Log"
2. Copy the group ID of the scheduled task
3. Click on "Task Groups"
4. Paste the group ID in the search box

I agree with you that we need some way to jump to a different task group/task ID in less clicks. I'll try to think about ways we could improve the workflow.

djmitche:
[...]

PR #276, opened 2019-02-19
Add 'follow log' to TaskLog

### LINKED ISSUE #275: [UI] Bring back the "Follow Log" button for logs that are not live
STR:

1. Navigate to https://taskcluster-web.netlify.com/tasks/eOMZxg6lS-mst_yYNZBTKg/runs/0/logs/https%3A%2F%2Fqueue.taskcluster.net%2Fv1%2Ftask%2FeOMZxg6lS-mst_yYNZBTKg%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log

Actual: There is no follow log button
screen shot 2019-02-19 at 1 06 04 pm

Expected: There is a follow log button
screen shot 2019-02-19 at 1 05 37 pm

We currently don't show the "follow log" button when a log has finished running. Let's make it show at all times. 

Relates to https://github.com/taskcluster/taskcluster/issues/274.

Aditya-Kolla:
Changed the TaskLog component such that it displays "follow log" options all the time.

This closes #275


ui/src/views/Tasks/TaskLog/index.jsx
```diff
@@ -36,8 +42,9 @@ export default class TaskLog extends Component {
       <Dashboard disablePadding>
         <Log
           url={url}
-          stream={false}
+          stream
```
helfi92: A finished log is no longer streaming. We should keep it set to `false`.

Aditya-Kolla:
What's gonna follow do in a non live log? Get to the end of the log?

helfi92:
It's going to resolve one of the points in https://github.com/taskcluster/taskcluster/issues/274#issue-412021119:

> In the list of logs I can't jump to the end any more, I think ("follow log"). So I have to grab the toolbar and scroll down manually.

Aditya-Kolla:
The 'shouldStartFollowing' method was dependent on stream value so I tried setting 'startFollowing' prop of the non live log to be dependent on the follow in state which is controlled by the user.
Please advise if this is the proper way to make this work.

Aditya-Kolla:
@helfi92 I did as you suggested and it works fine. 
One thing I noticed was the button css was getting triggered by the mouse click which shouldn't happen for non live logs so I fixed that. Also the tooltip and the font color.
Please review :)

Aditya-Kolla:
@helfi92 Thank you! :) 
The only difference between them would be the stream value. 
Should we consider sending a 'stream' boolean to `TaskLog` as a prop?

helfi92:
> Should we consider sending a 'stream' boolean to `TaskLog` as a prop?

Sounds like a good plan.

Aditya-Kolla:
@helfi92 All done, I hope?

ui/src/views/Tasks/index.jsx
```diff
@@ -84,11 +81,13 @@ export default class Task extends Component {
...
         <RouteWithProps
           path={`${path}/:taskId/runs/:runId/logs/live/:logUrl`}
+          stream
           {...props}
-          component={LiveTaskLog}
+          component={TaskLog}
         />
         <RouteWithProps
           path={`${path}/:taskId/runs/:runId/logs/:logUrl`}
+          stream={false}
```
helfi92: `stream` is `false` by default. We can remove this line.

ui/src/components/Log/index.jsx
```diff
@@ -276,12 +245,14 @@ export default class Log extends Component {
...
               variant="round"
               className={classNames({
-                [classes.followButtonFollowing]: follow,
+                [classes.followButtonFollowing]: follow && stream,
               })}
-              color={follow ? 'inherit' : 'secondary'}
+              color={follow && stream ? 'inherit' : 'secondary'}
               onClick={this.handleFollowClick}
               {...FollowLogButtonRest}>
               <ArrowDownBoldCircleOutlineIcon />
```
helfi92: For a log that is not streaming, we should be okay removing the `onScroll` props. I think you can do:

```jsx
onScroll={stream ? this.handleScroll : undefined}
```

ui/src/views/Tasks/index.jsx
Aditya-Kolla: Yeah. I was wondering whether it would be obligatory to show the same.

ui/src/components/Log/index.jsx
Aditya-Kolla: Alright.

Aditya-Kolla:
> Could you look into it, please? I suspect it has to do with a state (maybe `follow`) not passing a certain condition.

I checked the clickHandler for the follow button and it sets the boolean value of follow. 
So basically, when we're clicking it once it gets set to true. 
Second click, `follow: false` so nothing happens. 

I'm thinking to fix this by setting back the state to initial follow value in the callback when stream is false i.e when it's a non live log.
Suggestions?

PR #279, opened 2019-02-20
Deploy tc-web-ui from monoimage

owlishDeveloper:
Bugzilla Bug: [1525180](https://bugzilla.mozilla.org/show_bug.cgi?id=1525180)

I am not being able to test this :(  I am getting errors during `yarn install` at various stages. Errors also differ. For example:

```
yarn install v1.13.0
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
info fsevents@1.2.6: The platform "linux" is incompatible with this module.
info "fsevents@1.2.6" is an optional dependency and failed compatibility check. Excluding it from installation.
[4/5] Linking dependencies...
warning "workspace-aggregator-63abe0b2-6123-4fbf-b9b5-154b7691f5cd > taskcluster-web-server > apollo-server-express > apollo-server-core > apollo-tracing@0.4.0" has incorrect peer dependency "graphql@0.10.x - 14.0.x".
error An unexpected error occurred: "EIO: i/o error, scandir '/cache/v4/npm-walk-2.3.14-60ec8631cfd23276ae1e7363ce11d626452e1ef3/node_modules/walk'".
info If you think this is a bug, please open a bug report with the information provided in "/app/yarn-error.log".
```
or:
```
info There appears to be trouble with your network connection. Retrying...
```

All these errors happen during `Build Hook: API Services` stage, for `gce-provider` and `events`, although one time the builds for the latter two were able to complete

I'll try testing in the library on faster internets after the meeting. Decided to show the code for now at least


owlishDeveloper:
I tried several more times, and it reached the `web-ui` hooks, where it failed with very weird logs:

```
yarn run v1.13.0
warning Skipping preferred cache folder "/app/.cache/yarn" because it is not writable.
warning Selected the next writable cache folder in the list, will be "/tmp/.yarn-cache-501".
$ node infrastructure/builder/src/main.js build
warning Cannot find a suitable global folder. Tried these: "/usr/local, /app/.yarn"
[17:43:47] Repo gce-provider - Clone: running
[17:43:47] Pull Docker Image heroku/heroku:16: running
[17:43:47] Pull Docker Image heroku/heroku:16-build: running
[17:43:47] Clone heroku-buildpack-nodejs: running
[17:43:47] Repo events - Clone: running
[17:43:47] Repo docker-worker - Clone: running
[17:43:47] Repo generic-worker - Clone: running
[17:43:47] Pull Docker Image node:10.15.1: running
[17:43:47] Pull Docker Image node:10.15.1-alpine: running
[17:43:47] Clone Monorepo from Working Copy: running
[17:43:47] Pull Docker Image heroku/heroku:16: failed
[17:43:47] Pull Docker Image heroku/heroku:16: fail: Error: connect ENOENT /var/run/docker.sock
[17:43:47] Pull Docker Image heroku/heroku:16-build: failed
[17:43:47] Pull Docker Image heroku/heroku:16-build: fail: Error: connect ENOENT /var/run/docker.sock
[17:43:47] Pull Docker Image node:10.15.1: failed
[17:43:47] Pull Docker Image node:10.15.1: fail: Error: connect ENOENT /var/run/docker.sock
[17:43:47] Pull Docker Image node:10.15.1-alpine: failed
[17:43:47] Pull Docker Image node:10.15.1-alpine: fail: Error: connect ENOENT /var/run/docker.sock
[17:43:47] Clone Monorepo from Working Copy: failed
[17:43:47] Clone Monorepo from Working Copy: fail: Error: Command failed: git status --porcelain
fatal: Not a git repository (or any parent up to mount point /app/ui)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

[17:43:49] Repo events - Clone: finished
[17:43:49] Repo gce-provider - Clone: finished
[17:43:52] Repo generic-worker - Clone: finished
[17:43:53] Clone heroku-buildpack-nodejs: finished
[17:44:24] Repo docker-worker - Clone: finished
{ Error: connect ENOENT /var/run/docker.sock
    at PipeConnectWrap.afterConnect [as oncomplete] (net.js:1104:14)
  errno: 'ENOENT',
  code: 'ENOENT',
  syscall: 'connect',
  address: '/var/run/docker.sock' }
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
I would expect these logs to contain only what's happening in the monoimage.... Or are all those things happening in the monoimage?

infrastructure/builder/src/build/monoimage.js
```diff
@@ -147,8 +156,54 @@ const generateMonoimageTasks = ({tasks, baseDir, spec, cfg, cmdOptions}) => {
...
+        fs.mkdirSync(cacheDir);
+      }
+
+      utils.step({title: 'Run Yarn Install'});
+
+      await dockerRun({
+        image: nodeImage,
+        workingDir: '/app/ui',
+        env: ['YARN_CACHE_FOLDER=/ui/cache'],
+        command: ['yarn'],
```
djmitche: Will this work with `--production` to avoid devDependencies?

infrastructure/builder/src/build/service.js
```diff
@@ -24,7 +17,6 @@ const generateServiceTasks = ({tasks, baseDir, spec, cfg, name, cmdOptions}) =>
     break;
 
   case 'web-ui':
-    assert(!isMonorepo, 'monorepo not supported for this buildtype'); // TODO build from monorepo
     webUiTasks({tasks, baseDir, spec, cfg, name, cmdOptions, repository, workDir});
     break;
```
djmitche: This whole stanza can go now, I think?  Along with `service/web-ui.js`.  `web-ui` never appears in the spec.

infrastructure/builder/src/build/monoimage.js
```diff
@@ -147,8 +156,54 @@ const generateMonoimageTasks = ({tasks, baseDir, spec, cfg, cmdOptions}) => {
     },
   });
 
-  // TODO: add hooks push for web-ui
-  // TODO: add entry point for docker run for web-ui
+  hooks.push({
+    name: 'web-ui',
+    build: async (requirements, utils, procs) => {
+      const cacheDir = path.join(workDir, 'ui/cache');
```
djmitche: This can probably benefit from using the same cacheDir as the 'API Services' section, rather than its own cache.  That should save downloading any packages common between the two.

djmitche:
I'm working on a review -- but to the errors above: those look like an OS-level error (I/O Error) and then a crashed Docker daemon (`connect ENOENT /var/run/docker.sock`).  By any chance is the disk full?  Otherwise all I can think of is the old "turn it off and on again"..

djmitche:
Oh, I think the docker-crashed issue is actually from running `yarn build` inside docker.  More notes in the review comments.

infrastructure/builder/src/build/monoimage.js
```diff
@@ -147,8 +156,54 @@ const generateMonoimageTasks = ({tasks, baseDir, spec, cfg, cmdOptions}) => {
...
+
+      await dockerRun({
+        image: nodeImage,
+        workingDir: '/app/ui',
+        env: ['YARN_CACHE_FOLDER=/ui/cache'],
+        command: ['yarn'],
+        logfile: `${workDir}/yarn.log`,
+        utils,
+        binds: [
+          `${appDir}:/app/ui`,
```
djmitche: This is mounting `appDir`, which is the repository root, at `/app/ui`, so the `yarn` command is running in the repository root -- and the next `dockerRun` thus runs `yarn build` in the root, which tries to use Docker, which doesn't work in a Docker container.  I think you'll want to mount `appDir` at `/app`, and rely on the `workingDir` definition above to make the `yarn` command run inside the `ui` subdirectory.  Same for the `dockerRun` below.

infrastructure/builder/src/build/monoimage.js
```diff
@@ -147,8 +156,54 @@ const generateMonoimageTasks = ({tasks, baseDir, spec, cfg, cmdOptions}) => {
...
+      });
+    },
+    entrypoints: async (requirements, utils, procs) => {
+      // since we ran `yarn build` already, there's no need to run it again
+      // on startup, so remove it from the entrypoint commands
+      Object.keys(procs).forEach(process => {
+        if (process.startsWith('web-ui/')) {
+          procs[process] = procs[process].replace('yarn build && ', '');
+        }
+      });
```
djmitche: This worked for web-server because it has a Procfile - it just had a bit in there I didn't want.

For `web-ui`, we need to set up an entrypoint that will run nginx.  This will work more like it does for the "References" hook.

infrastructure/builder/src/build/monoimage.js
owlishDeveloper: > Running yarn with no command will run yarn install, passing through any provided flags.

From the docs

djmitche: Yes, and adding `--production` will cause it to skip installing any `devDependencies`, which should save some space in the docker iamge.

infrastructure/builder/src/build/monoimage.js
owlishDeveloper: What does "mounting" mean here? I was actually trying to find out what `binds` means - my understanding is this comes from `dockerode`, so I tried looking up docks in there and found no explanation, and then I tried to search in Docker docs for `docker run` comand - also didn't find anything

djmitche: It's equivalent to [-v](https://docs.docker.com/engine/reference/run/#volume-shared-filesystems).  I'm not sure why it's named "binds" instead of "volumes".  The format is "<dir-on-host>:<dir-in-container>" so `${appDir}:/app/ui` mounts whatever `appDir` points at, on the host filesystem, at `/app/ui` in the docker container.

owlishDeveloper: There are `volumes`: https://github.com/apocas/dockerode/blob/master/examples/external_volume.js#L10 `binds` appears to be separate

infrastructure/builder/src/build/monoimage.js
owlishDeveloper: Why don't we add `--production` anywhere? I was simply copying your code

infrastructure/builder/src/build/monoimage.js
djmitche: Docker's pretty inconsistent in what it calls things.  Since dockerode is a wrapper around the Docker API, I start with the API docs.  So corresponding to the link in your comment, the createContainer docs are at https://docs.docker.com/engine/api/v1.37/#operation/ContainerCreate.  That has both Volumes and HostConfig.Binds.  I think Volumes just mounts empty temporary directories (that will be destroyed when the container exits) while Binds connects host directories to container directories:

> A list of volume bindings for this container. Each volume binding is a string in one of these forms:
> 
>     host-src:container-dest to bind-mount a host path into the container. Both host-src, and container-dest must be an absolute path.
>     host-src:container-dest:ro to make the bind mount read-only inside the container. Both host-src, and container-dest must be an absolute path.
>     volume-name:container-dest to bind-mount a volume managed by a volume driver into the container. container-dest must be an absolute path.
>     volume-name:container-dest:ro to mount the volume read-only inside the container. container-dest must be an absolute path.

Now that I look through that documentation, it looks like `HostConfig.Mounts` could do the same job.  Maybe that's a newer option or an older one?  Dunno, Docker is weird.

I don't specifically remember figuring this out, but my general sources for Docker learnings are:
 * The Docker API, because dockerode docs say "Input options are directly passed to Docker. Check Docker API documentation for more details."
 * Other things that use docker.  I vaguely recall looking at [docker-worker](https://github.com/taskcluster/docker-worker/blob/a930de7b6101f95caa1f91e986f40e27f3039abb/test/dockerworker.js#L96-L100)
 * Trial and error.  Docker API docs are only for the most recent version, which isn't necessarily what we're running, and sometimes things change incompatibly.

infrastructure/builder/src/build/monoimage.js
djmitche: I think there's a bug to add that.  I tried it once and it didn't work, and I didn't try too hard to figure out why.  If it doesn't work for tc-ui, we can file another bug and leave out `--production` for now.

djmitche: https://bugzilla.mozilla.org/show_bug.cgi?id=1521648

If it's too much trouble, we can drop this and just add tc-ui to that bug.

owlishDeveloper: I see. It sounded as if there is `--production` everywhere, and I just missed it for some reason :)

infrastructure/builder/src/build/monoimage.js
owlishDeveloper: oh, API docs... I see my mistake. I was looking at the CLI docs. I found `binds` in API docs. ty!

infrastructure/builder/src/build/monoimage.js
owlishDeveloper: I tried both `'yarn', '--production'` and `'yarn', 'install', '--production'` - I still see things like `@neutrinojs/karma@9.0.0-beta.1` in the logs (that's a dev dependency)

djmitche: OK, let's worry about it later..

infrastructure/builder/src/build/utils.js
```diff
@@ -262,8 +261,9 @@ exports.dockerPull = async ({baseDir, image, utils}) => {
           // calculate overall progress by assuming that every image must be
           // downloaded and extracted, and that those both take the same amount
           // of time per byte.
-          total = _.sum(Object.values(totals)) * 2;
-          current = _.sum(Object.values(downloading)) + _.sum(Object.values(extracting));
+          let total = Object.values(totals).reduce((a, b) => a + b, 0) * 2;
```
helfi92: I think this can be a `const` since it's not being reassigned.

infrastructure/builder/src/build/utils.js
```diff
@@ -262,8 +261,9 @@ exports.dockerPull = async ({baseDir, image, utils}) => {
...
+          let current = Object.values(downloading).reduce((a, b) => a + b, 0) +
```
helfi92: `const current`.

infrastructure/builder/src/build/utils.js
```diff
@@ -326,7 +326,7 @@ exports.dockerRegistryCheck = async ({tag}) => {
     // Acces the registry API directly to see if this tag already exists, and do not push if so.
     // TODO: this won't work with custom registries!
     const res = await got(`https://index.docker.io/v1/repositories/${repo}/tags`, {json: true});
-    if (res.body && _.includes(res.body.map(l => l.name), imagetag)) {
+    if (res.body && res.body.map(l => l.name).includes(imagetag)) {
```
helfi92: What's `l`? Let's rename to something more descriptive.

infrastructure/builder/src/build/index.js
```diff
@@ -1,21 +1,18 @@
-const _ = require('lodash');
```
helfi92: 👌

infrastructure/builder/src/build/utils.js
owlishDeveloper: it's a common pattern to use non-descriptive one-letter names like this in lamda-functions. As to why `l` was chosen in this case - I am not sure :)

infrastructure/builder/src/build/index.js
```diff
@@ -1,21 +1,18 @@
...
 const mkdirp = util.promisify(require('mkdirp'));
 const {ClusterSpec} = require('../formats/cluster-spec');
 const {TerraformJson} = require('../formats/tf-json');
 const {TaskGraph, Lock, ConsoleRenderer, LogRenderer} = require('console-taskgraph');
-const {gitClone} = require('./utils');
 const generateRepoTasks = require('./repo');
 const generateMonoimageTasks = require('./monoimage');
 
-const _kindTaskGenerators = {
+const kindTaskGenerators = {
```
imbstack: Should we just get rid of this if it doesn't return anything anymore?

ui/web-ui-nginx-site.conf
```diff
@@ -0,0 +1,68 @@
...
+                      '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /var/log/nginx/access.log  main;
+
+    sendfile        on;
+    #tcp_nopush     on;
+
+    keepalive_timeout  65;
+
+    #gzip  on;
```
imbstack: Why disabled?

imbstack: Ah, I see it is also down below. Let's just delete the line in that case.

infrastructure/builder/src/build/index.js
owlishDeveloper: It does - the functions collected in this object are used below (see line 60)

ui/web-ui-nginx-site.conf
owlishDeveloper: ah, I just copied the default nginx file. I can delete these

PR #283, opened 2019-02-22
Bug 1527969  Add function to lib-loader that is for using when file is executed

sudipt1999:
Bugzilla Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1527969

`crashOnError()` method added to the `loader.js` file 

ran the test locally !

test result : https://gist.github.com/sudipt1999/a3180149f259b1139e88f18172b957c4

kindly cross check for update 

Thank You

### DIFF
--- a/libraries/loader/src/loader.js
+++ b/libraries/loader/src/loader.js
@@ -103,7 +103,7 @@ function loader(componentDirectory, virtualComponents = {}) {
     setup: () => console.log(renderGraph(componentDirectory, topoSorted)),
   };
 
-  return function(target, options = {}) {
+  let load = function(target, options = {}) {
     options = Object.assign({}, options);
 
     if (typeof target !== 'string') {
@@ -131,15 +131,14 @@ function loader(componentDirectory, virtualComponents = {}) {
     Object.entries(options).forEach(([key, comp]) => {
       loaded[key] = Promise.resolve(comp);
     });
-
     // Load a component
-    function load(target) {
+    function recursiveLoad(target) {
       if (!loaded[target]) {
         var def = componentDirectory[target];
         // Initialize component, this won't cause an infinite loop because
         // we've already check that the componentDirectory is a DAG
         let requires = def.requires || [];
-        return loaded[target] = Promise.all(requires.map(load)).then(deps => {
+        return loaded[target] = Promise.all(requires.map(recursiveLoad)).then(deps => {
           let ctx = {};
           for (let i = 0; i < deps.length; i++) {
             ctx[def.requires[i]] = deps[i];
@@ -158,9 +157,16 @@ function loader(componentDirectory, virtualComponents = {}) {
       }
       return loaded[target];
     }
-
-    return load(target);
+    return recursiveLoad(target);
   };
+
+  load.crashOnError = function(target) {
+    load(target).catch(err => {
+      console.log(err.stack);
+      process.exit(1);
+    });
+  };
+  return load;
 }
 
 module.exports = loader;
--- a/libraries/loader/test/loader_tests.js
+++ b/libraries/loader/test/loader_tests.js
@@ -518,7 +518,7 @@ suite('component loader', () => {
     assume(orderCalled).eql(['dep4', 'dep3', 'dep2', 'dep1', 'base']);
   });
 
-  it('should fail when specified component didn\'t load', async () => {
+  test('should fail when specified component didn\'t load', async () => {
     let load = subject({
       fail: {
         requires: [],
--- a/libraries/loader/test/loader_tests.js
+++ b/libraries/loader/test/loader_tests.js
@@ -517,4 +517,16 @@ suite('component loader', () => {
     assume(await load('base', {})).equals(rv);
     assume(orderCalled).eql(['dep4', 'dep3', 'dep2', 'dep1', 'base']);
   });
+
+  it('should fail when specified component didn\'t load', async () => {
+    let load = subject({
+      fail: {
+        requires: [],
+        setup: () => Promise.reject(new Error('uhoh!')),
+      },
+    });
+
+    assert.throws( function() { load.crashOnError(true); }, false);
+  });
+
 });
--- a/services/purge-cache/src/main.js
+++ b/services/purge-cache/src/main.js
@@ -115,10 +115,7 @@ const load = loader({
 
 // If this file is executed launch component from first argument
 if (!module.parent) {
-  load(process.argv[2]).catch(err => {
-    console.log(err.stack);
-    process.exit(1);
-  });
+  load.crashOnError(process.argv[2]);
 }
 
 module.exports = load;
--- a/libraries/loader/README.md
+++ b/libraries/loader/README.md
@@ -157,9 +157,7 @@ process running the specified component.
 ```js
 // If this file is executed launch component from first argument
 if (!module.parent) {
-  load(process.argv[2]).catch(err => {
-    console.log(err.stack);
-    process.exit(1);
-  });
+  // method to crash the program if different components get loaded
+  load.crashOnError(process.argv[2]);
 }
 ```
--- a/services/auth/src/main.js
+++ b/services/auth/src/main.js
@@ -241,10 +241,7 @@ const load = Loader({
 
 // If this file is executed launch component from first argument
 if (!module.parent) {
-  load(process.argv[2]).catch(err => {
-    console.log(err.stack);
-    process.exit(1);
-  });
+  load.crashOnError(process.argv[2]);
 }
 
 module.exports = load;
--- a/services/built-in-workers/src/main.js
+++ b/services/built-in-workers/src/main.js
@@ -1,5 +1,6 @@
 const loader = require('taskcluster-lib-loader');
 const monitorManager = require('./monitor');
+const monitor = require('taskcluster-lib-monitor');
 const docs = require('taskcluster-lib-docs');
 const taskcluster = require('taskcluster-client');
 const config = require('typed-env-config');
@@ -73,10 +74,7 @@ const load = loader({
 
 // If this file is executed launch component from first argument
 if (!module.parent) {
-  load(process.argv[2]).catch(err => {
-    console.log(err.stack);
-    process.exit(1);
-  });
+  load.crashOnError(process.argv[2]);
 }
 
 module.exports = load;
--- a/services/github/src/main.js
+++ b/services/github/src/main.js
@@ -262,10 +262,7 @@ const load = loader({
 
 // If this file is executed launch component from first argument
 if (!module.parent) {
-  load(process.argv[2]).catch(err => {
-    console.log(err.stack);
-    process.exit(1);
-  });
+  load.crashOnError(process.argv[2]);
 }
 
 module.exports = load;
--- a/services/hooks/src/main.js
+++ b/services/hooks/src/main.js
@@ -245,10 +245,7 @@ const load = loader({
 
 // If this file is executed launch component from first argument
 if (!module.parent) {
-  load(process.argv[2]).catch(err => {
-    console.log(err.stack);
-    process.exit(1);
-  });
+  load.crashOnError(process.argv[2]);
 }
 
 module.exports = load;
--- a/services/index/src/main.js
+++ b/services/index/src/main.js
@@ -6,6 +6,7 @@ const Handlers = require('./handlers');
 const builder = require('./api');
 const Config = require('typed-env-config');
 const loader = require('taskcluster-lib-loader');
+const monitor = require('taskcluster-lib-monitor');
 const monitorManager = require('./monitor');
 const SchemaSet = require('taskcluster-lib-validate');
 const App = require('taskcluster-lib-app');
@@ -189,10 +190,7 @@ var load = loader({
 
 // If this file is executed launch component from first argument
 if (!module.parent) {
-  load(process.argv[2]).catch(err => {
-    console.log(err.stack);
-    process.exit(1);
-  });
+  load.crashOnError(process.argv[2]);
 }
 
 module.exports = load;
--- a/services/login/src/main.js
+++ b/services/login/src/main.js
@@ -107,10 +107,7 @@ let load = loader({
 
 // If this file is executed launch component from first argument
 if (!module.parent) {
-  load(process.argv[2]).catch(err => {
-    console.log(err.stack);
-    process.exit(1);
-  });
+  load.crashOnError(process.argv[2]);
 }
 
 module.exports = load;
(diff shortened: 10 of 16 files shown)


libraries/loader/src/loader.js
```diff
@@ -2,7 +2,7 @@ const util = require('util');
 const assume = require('assume');
 const TopoSort = require('topo-sort');
 const debug = require('debug')('taskcluster-lib-loader');
-
+var myLoader;//variable to store method load() so that er can use it globally
```
jhford: We usually prefer to use `let` or `const` for declaring variables.  Is there a reason to pick var?

I'm not sure the reason for storing a module level reference to a load function?  The load function is meant to be returned as a closure, so storing a global reference to the last closure created by the module could have implications that I'm not totally sure of.

libraries/loader/src/loader.js
```diff
@@ -158,9 +158,20 @@ function loader(componentDirectory, virtualComponents = {}) {
...
   };
 }
 
-module.exports = loader;
+
+
+/*
+  Method to crash the process if error occours
+*/
+function crashOnError(target)
```
jhford: Ahh, the global assignment to the last created loader is so that this method can access it here.

What might be a prefered solution is to set the `crashOnError` function as a property of the closure which is returned, rather than as a module function.

libraries/loader/src/loader.js
sudipt1999: Will change to `let` . My bad Sorry!

libraries/loader/src/loader.js
sudipt1999: I thought of making a `class` first but I failed when I tested it out so , as this idea was given by bstack . So I thought of doing that way only.

sudipt1999:
@jhford  some pointers would be really helpful

sudipt1999:
@jhford kindly change `subject` in `loader_test.js` file from
`let subject = require('../src/loader');` 
to 
`let subject = require('../src/loader').loader;`
since previously I didn't commit it , but test ran successfully locally so I guess this need to be done!

sudipt1999:
All services files changed according to updated `loader.js` file ,
task group report [task group report](https://tools.taskcluster.net/groups/eK5WbVpfQOOVbsLbZxT0ww) 
Kindly review it !

libraries/loader/src/loader.js
jhford: There's not really a need to make this a class, and that would likely result in a fairly large change to the the projects which use the loader.

Since a closure is a function which returns a function, and a function is an object which can have properties, the `crashOnError` function could be attached as a property of the function that's returned.

For example:
```javascript
function makeClosure(x) {
  let closure =  function(y) {
    console.log('From closure scope:', x, 'From argument:', y);
  }

  closure.doubler = function(y) {
    console.log('From closure scope:', x * 2, 'From argument:', y * 2);
  }

  return closure;
}

let closure = makeClosure(10);

closure(20);
closure.doubler(20);
```
will print
```bash
$ node closure.js 
From closure scope: 10 From argument: 20
From closure scope: 20 From argument: 40
```

sudipt1999:
Thanks a lot !
Will update it accordingly now .
Thank You!

sudipt1999:
Thank you sir for your guidance .
Will do changes quickly and update the pr asap.
Thank You

libraries/loader/src/loader.js
```diff
@@ -3,6 +3,7 @@ const assume = require('assume');
 const TopoSort = require('topo-sort');
 const debug = require('debug')('taskcluster-lib-loader');
 
+let myLoader;//variable to store method load() so that er can use it globally
```
jhford: I don't think we need to declare this at module level (globally), since we're only referencing it inside the closure and not exporting it anywhere.

libraries/loader/src/loader.js
```diff
@@ -103,7 +104,7 @@ function loader(componentDirectory, virtualComponents = {}) {
     setup: () => console.log(renderGraph(componentDirectory, topoSorted)),
   };
 
-  return function(target, options = {}) {
+  myLoader = function(target, options = {}) {
```
jhford: Could this be `let myLoader = ...` instead of having the module-level variable?

jhford:
@sudipt1999 it might also be a good idea to squash the commits.  Are you familiar with `git rebase -i origin/master`?

sudipt1999:
I am not familiar but will look to it !

sudipt1999:
Have updated the changes requested @jhford ! Kindly review it.

jhford:
The queue unit test failures are from an `ECONNECT`, so I've requested it to be run again.

sudipt1999:
I guess it was due to my slow internet connection ! Hope it will be
successful this time .


On 26-Feb-2019 8:14 PM, "John Ford" <notifications@github.com> wrote:

The queue unit test failures are from an ECONNECT, so I've requested it to
be run again.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/taskcluster/taskcluster/pull/283#issuecomment-467464921>,
or mute the thread

sudipt1999:
As I thought it was due to my slow internet connection ! Hola ! I guess now its done kindly update me if any changes required. Thanks for teaching me `closure function` @jhford

jhford:
@sudipt1999 actually, those tests aren't being run on your machine :)  It looks like it was a timeout in our testing infrastructure running on EC2 :)

Did you have a chance to read up on `git rebase -i origin/master`?  Need any pointers on that?

sudipt1999:
I did saw a tutorial of rebase on youtube ! I also ran the git rebase
command successfully . Guess i made some mistakes somewhere  . Yeah please
provide some pointers

jhford:
Sure!  So what I do when I want to rebase, I can run the command `git rebase -i origin/master` which takes the current branch and shows me which commits have been added to my branch on top of the `origin/master` reference (what is master on github).

What'll happen next is that your editor will open with a file.

```
 1 pick 55752e2 Add pagination and streaming
 2 pick d8beeb4 Lots of inserters
 3 pick 4222f94 Use batches of 2000
 4
 5 # Rebase d401c1c..4222f94 onto d401c1c (3 commands)
 6 #
 7 # Commands:
 8 # p, pick <commit> = use commit
 9 # r, reword <commit> = use commit, but edit the commit message
10 # e, edit <commit> = use commit, but stop for amending
11 # s, squash <commit> = use commit, but meld into previous commit
12 # f, fixup <commit> = like "squash", but discard this commit's log message
13 # x, exec <command> = run command (the rest of the line) using shell
14 # b, break = stop here (continue rebase later with 'git rebase --continue')
15 # d, drop <commit> = remove commit
16 # l, label <label> = label current HEAD with a name
17 # t, reset <label> = reset HEAD to a label
18 # m, merge [-C <commit> | -c <commit>] <label> [# <oneline>]
19 # .       create a merge commit using the original merge commit's
20 # .       message (or the oneline, if no original merge commit was
21 # .       specified). Use -c <commit> to reword the commit message.
22 #
23 # These lines can be re-ordered; they are executed from top to bottom.
24 #
25 # If you remove a line here THAT COMMIT WILL BE LOST.
26 #
27 # However, if you remove everything, the rebase will be aborted.
28 #
29 # Note that empty commits are commented out
```

This file can be edited to do reflect what you're wanting to do.  It has explanations for what each change needs to be.  Say that I want to squash `55752e2` and `d8beeb4` into a single commit, with a single change message, I could edit the top lines to read:

```
 1 pick 55752e2 Add pagination and streaming
 2 fixup d8beeb4 Lots of inserters
 3 pick 4222f94 Use batches of 2000
```

And if I wanted to squash them but have a different commit message, I could edit the file to do:

```
 1 reword 55752e2 Add pagination and streaming
 2 fixup d8beeb4 Lots of inserters
 3 pick 4222f94 Use batches of 2000
```

Once you've decided what changes you'd like to make, you can save an exit and git will do the rebase.  If there are conflicts applying each commit, you can use `git status` to find the conflicts, then follow the advice in that command's output to figure out what to do next :)  Oh, and you'll have to force push to your crashOnError branch (`git push origin crashOnError -f`) after you've done a rebase.

On your first attempt, I'd recommend copying the whole repository so you don't loose state... `cp -r taskcluster taskcluster-backup`.  After a couple times, this will become second nature, but at the beginning it's not a bad idea in case you want to revert back.

Let me know how it goes!

sudipt1999:
@jhford  thanks for guiding me in rebase ! I am really happy to learn this git command and closure from you! I am seeing lot of changed files in pr I hope everything went good with rebase

sudipt1999:
I went confused between squash and fixup ! So i ended up doing squash finally worked ! Thank you

libraries/loader/package-lock.json
```diff
@@ -0,0 +1,96 @@
+{
```
djmitche: This file was created by using `npm` instead of `yarn` here.  You should be able to just delete the file.  You may also need to run `yarn` in the root directory of the repository in order to fix the `yarn.lock` file contents.

services/purge-cache/src/main.js
```diff
@@ -4,6 +4,7 @@ const path = require('path');
 const _ = require('lodash');
 const config = require('typed-env-config');
 const loader = require('taskcluster-lib-loader');
+const monitor = require('taskcluster-lib-monitor');
```
djmitche: Why are these added in this file and in `services/index` and `services/built-in-workers`?  Perhaps they were added accidentally during the rebase?

libraries/loader/src/loader.js
```diff
@@ -102,8 +102,8 @@ function loader(componentDirectory, virtualComponents = {}) {
   componentDirectory['dump-dot'] = {
     setup: () => console.log(renderGraph(componentDirectory, topoSorted)),
   };
-
-  return function(target, options = {}) {
+  // Implementing the closure property to stablish load.crashOnError() functionality
+  let myLoader = function(target, options = {}) {
```
djmitche: We typically don't name variables in "production" code starting with `my` -- it's common in examples and demos as a way of saying "you can name this whatever you want".  Let's call this `load`, since that's what it will be called by services that are using the library.

To avoid confusion, you can rename the `load` sub-function below (around line 134) to something like `recursiveLoad`.

libraries/loader/src/loader.js
```diff
@@ -158,9 +157,17 @@ function loader(componentDirectory, virtualComponents = {}) {
       }
       return loaded[target];
     }
-
+    //method to crash the process if error occured
+    myLoader.crashOnError = function(target)
```
djmitche: This is adding the `crashError` property within the body of the `myLoader` function -- so it won't be available until after that function is called.  I think you want to define it *outside* of that function's body -- just before `return myLoader;`.

sudipt1999: As soon as the 
```
import loader from 'taskcluster-lib-loader'
let load = loader();
load.crashOnError(arg);//this can be called like this 
```
since we are passing as an object there wont be any problem !
if we write it as you suggested @djmitche  the scope of `load()` would not be accessible to `crashOnError()` .

libraries/loader/src/loader.js
sudipt1999: Yeah ! I will do the needful ! Thank You

libraries/loader/src/loader.js
djmitche: Add some tests and you'll see what I mean :)

sudipt1999: I got what you were trying to convey @djmitche  until we don't invoke the `load()`  method  `crashOnError()` method won't come to existence . It would be undefined until then . I should have thought about that , will update the code accordingly.

sudipt1999: Thank you

sudipt1999: Thank you

sudipt1999:
@djmitche `||` @jhford  was https://github.com/taskcluster/taskcluster/blob/master/libraries/loader/test/loader_tests.js#L230
the test for `crashOnError()` , because when the component won't be found it will throw an `error` and not reject a promise ;

djmitche:
There's no test for crashOnError because you haven't written one yet!

That test is checking that plain old `load` correctly throws an exception (asynchronously) when a loader function fails.

Note that when using `await`/`async`, an `await` that gets a rejected promise looks exactly like a thrown exception.  So, the two really aren't any different.

sudipt1999:
@jhford  error is occurring from the test I wrote 
which is 

```
it('should fail when specified component didn\'t load', async () => {
    let load = subject({
      fail: {
        requires: [],
        setup: () => Promise.reject(new Error('uhoh!')),
      },
    });

     assert.throws( function() { load.crashOnError(true); }, false);
  });
```
Kindly correct me out please !

services/purge-cache/src/main.js
djmitche: ^^ still needs an answer, I think?

libraries/loader/test/loader_tests.js
```diff
@@ -517,4 +517,16 @@ suite('component loader', () => {
     assume(await load('base', {})).equals(rv);
     assume(orderCalled).eql(['dep4', 'dep3', 'dep2', 'dep1', 'base']);
   });
+
+  test('should fail when specified component didn\'t load', async () => {
```
djmitche: :100:

libraries/loader/src/loader.js
```diff
@@ -102,8 +102,8 @@ function loader(componentDirectory, virtualComponents = {}) {
   componentDirectory['dump-dot'] = {
     setup: () => console.log(renderGraph(componentDirectory, topoSorted)),
   };
-
-  return function(target, options = {}) {
+  // Implementing the closure property to stablish load.crashOnError() functionality
```
djmitche: I'm not sure what this means.. perhaps remove it?

libraries/loader/src/loader.js
```diff
@@ -158,9 +157,16 @@ function loader(componentDirectory, virtualComponents = {}) {
       }
       return loaded[target];
     }
-
-    return load(target);
+    return recursiveLoad(target);
+  };
+  //method to crash the process if error occured
```
djmitche: rather than document here, it'd be best to include this documentation in `libraries/loader/README.md` where the rest of this library is documented.  Note, too, the commenting style in the rest of this file: whitespace both before and after `//`.

libraries/loader/src/loader.js
```diff
@@ -158,9 +157,16 @@ function loader(componentDirectory, virtualComponents = {}) {
...
+  load.crashOnError = function(target) {
+    load(target).catch(err => {
+      console.log(err.stack);//will print the error
```
djmitche: This comment is redundant -- it says the same thing as the code on this line.  Please remove :)

PR #284, opened 2019-02-22
[bug 1529998] Try out pnp

imbstack:
Obviously not ready to land yet, but this could make things simpler/faster/more robust!

Bugzilla Bug: [1529998](https://bugzilla.mozilla.org/show_bug.cgi?id=1529998)


imbstack:
Well that is harder than I expected. Finding binaries is hard here still. Closing for now

PR #299, opened 2019-02-25
Upgrade azure-entities to drop buffertools dep

djmitche:
```
(v10.15.1) dustin@lamport ~/p/taskcluster [upgrade-azure-entities*] $ yarn why buffertools
yarn why v1.13.0
[1/4] Why do we have the module "buffertools"...?
[3/4] Finding dependency...
error We couldn't find a match!
Done in 1.43s.
```

Thanks to @sudipt1999!


djmitche:
(dropping this dependency makes it easier to develop taskcluster on windows)

PR #308, opened 2019-02-25
Change terraform to deploy ui from the monoimage

owlishDeveloper:
Bugzilla Bug: [1525180](https://bugzilla.mozilla.org/show_bug.cgi?id=1525180)


owlishDeveloper:
because I just copied over what I already had in mozilla-terraform's modules :)

PR #310, opened 2019-02-26
Bump client-web version

helfi92:
CHANGELOG: https://github.com/taskcluster/taskcluster/commit/6f881ee1fc0ac547189cd2982cea1649f910f159

### DIFF
--- a/clients/client-web/package.json
+++ b/clients/client-web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "taskcluster-client-web",
-  "version": "9.0.0",
+  "version": "9.1.0",
   "main": "build/index.js",
   "author": "Eli Perelman <eli@eliperelman.com>",
   "license": "MPL-2.0",


helfi92:
> What do you think about using a single version for everything in the monorepo?

What packages are acutally published from the monorepo other than taskcluster-client and taskcluster-client-web?

djmitche:
It's those two, and my thinking is that we'd co-version those with a release of the services.  So basically you'd run `yarn release --minor`, and that would:

* bump the version in `package.json` and run `yarn generate` to disperse that version everywhere;
* commit and tag (like `yarn version --minor` does);
* build and upload a new docker image `taskcluster/taskcluster:21.2.3`;
* create a github release (maybe eventually containing a change log?); and
* publish new clients `taskcluster-client@21.2.3` and `taskcluster-client-web@21.2.3`.

This is probably an odd place to talk about.  I'm just thinking out loud :)

helfi92:
This sounds good to me. I like it.

Issue #300, opened 2019-02-25
[UI] Copy-and-paste task definition into "create task" fails validation


mitchhentges:
**Describe the bug**
I frequently need to run tasks very similar to existing tasks, such as to diagnose a failing build.
However, when I copy-and-paste the definition of the existing task, it fails validation, such as for `.requires`: when I copy it ([e.g.: from here](https://queue.taskcluster.net/v1/task/KtK5-mBtQ3uMuHsCgDpnGQ)), it has a value in lowercase with hyphens (like `all-completed`), but when I run the task, validation fails saying it expected something in uppercase with underscores (like `ALL_COMPLETED`)

**To Reproduce**
1. Find an [existing task](https://taskcluster-web.netlify.com/tasks/KtK5-mBtQ3uMuHsCgDpnGQ)
2. Click to "View Task Definition"
3. Copy  the raw JSON
4. Go to [create a new task](https://taskcluster-web.netlify.com/tasks/create)
5. Paste the raw JSON
6. Create the task

**Expected behavior**
A task should be created that's identical to the task that was copied

**Platform (please complete the following information):**
 - OS: Ubuntu 18.10
 - Browser: Firefox 65.0

helfi92:
Sorry for that, It's most probably a bug from our end. I will take care of that one. Thanks for the issue!

PR #312, opened 2019-02-26
[UI] Fix task definition mutation validation

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/300.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -67,6 +67,7 @@
     "prop-types": "^15.6.2",
     "qs": "^6.6.0",
     "react": "^16.8.2",
+    "ramda": "^0.26.1",
     "react-apollo": "^2.3.3",
     "react-codemirror2": "^5.1.0",
     "react-copy-to-clipboard": "^5.0.1",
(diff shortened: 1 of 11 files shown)


ui/package.json
```diff
@@ -67,6 +67,7 @@
     "prop-types": "^15.6.2",
     "qs": "^6.6.0",
     "react": "^16.8.2",
+    "ramda": "^0.26.1",
```
helfi92: We used ramda in many places in the UI. Not sure why it wasn't specified in package.json.

ui/src/utils/formatTaskMutation.js
```diff
@@ -0,0 +1,18 @@
+import upperSnakeCase from './upperSnakeCase';
+
+// This should be called just before sending a request to the
+// server when mutating a task payload.
+// The GraphQL server requires that enums are capitalized.
```
djmitche: This makes me uneasy -- we're transforming a task definition in the UI, hoping that we precisely reverse that transform in web-server before passing it on to the queue.  If our hopes aren't satisfied, then users will get strange errors about inputs they did not provide.  Let's keep an eye on this, and if things get more complicated perhaps consider alternative approaches that ferry task definitions intact between the queue and the UI?

ui/package.json
djmitche: Looks like material-ui-treeview depends on it.

PR #313, opened 2019-02-26
Bug 1361375 - implement query string options for the testAuthenticate endpoint

sudipt1999:
Bugzilla Bug: [1361375](https://bugzilla.mozilla.org/show_bug.cgi?id=1361375)

implement query string options for the testAuthenticate endpoint . Changes done
1. created a variable `credential` to store the object `{clientId, scopes}`
2. for query string ,used `history.pushState()` to add query string to current url
3. for setting header of credential used `res.setHeader()` which will be set as `X-test_auth`

### DIFF
--- a/services/auth/schemas/v1/test-authenticate-response.yml
+++ b/services/auth/schemas/v1/test-authenticate-response.yml
@@ -22,7 +22,7 @@ properties:
     uniqueItems:            true
   queryString:
     description: |
-      Query string for the testAuthenticate
-    type:                       string
-    additionalProperties:       false
-required: [clientId, scopes]
+      Query string for the testAuthenticate. Will be sent by test-authenticate route for testing endpoint.
+    type:                       object
+    default:                    {}
+required: [clientId, scopes, queryString]
--- a/services/auth/schemas/v1/test-authenticate-response.yml
+++ b/services/auth/schemas/v1/test-authenticate-response.yml
@@ -20,4 +20,9 @@ properties:
       type:                 string
       pattern: "^[\x20-\x7e]*$"
     uniqueItems:            true
+  queryString:
+    description: |
+      Query string for the testAuthenticate
+    type:                       string
+    additionalProperties:       false
 required: [clientId, scopes]
--- a/services/auth/src/api.js
+++ b/services/auth/src/api.js
@@ -1031,7 +1031,9 @@ builder.declare({
     'The request is validated, with any certificate, authorizedScopes, etc.',
     'applied, and the resulting scopes are checked against `requiredScopes`',
     'from the request body. On success, the response contains the clientId',
-    'and scopes as seen by the API method.',
+    'scopes and queryString as seen by the API method.',
+    'For testing the endpoints queryString will be sent in response, will',
+    'contain clientId and scopes for testing',
   ].join('\n'),
 }, async function(req, res) {
   await new Promise(next => APIBuilder.middleware.remoteAuthentication({
@@ -1060,10 +1062,9 @@ builder.declare({
     req.clientId(),
     req.scopes(),
   ]);
-  //setting headers as purposed by bug 161375
-  req.params.test_auth = {clientId, scopes};
-  res.setHeader('X-test_auth', {clientId, scopes});
-  res.reply({clientId, scopes});
+  
+  const queryString = {clientId, scopes};
+  res.reply({clientId, scopes, queryString});
 });
 
 builder.declare({
@@ -1120,24 +1121,3 @@ builder.declare({
   res.reply({clientId, scopes});
 });
 
-/** Query string to authenticate **/
-builder.declare({
-  method: 'get',
-  route: '/test-authenticate?q={auth}',
-  query: {
-    prefix: /^[A-Za-z0-9!@/:.+|_-]+$/, // should match clientId above
-    continuationToken: /./,
-    limit: /^[0-9]+$/,
-  },
-  name: 'queryString',
-  output: 'test-authenticate-response.yml',
-  stability: 'stable',
-  title: 'Query String',
-  description: [
-    'Get the query string passed in the url for authentication',
-    'A object will be query String present else null String would be received',
-  ].join('\n'),
-}, async function(req, res) {
-  let response = req.params.test_auth || '';
-  res.reply(response);
-});
--- a/services/auth/src/api.js
+++ b/services/auth/src/api.js
@@ -1135,7 +1135,7 @@ builder.declare({
   title: 'Query String',
   description: [
     'Get the query string passed in the url for authentication',
-    'A object will be query String present else null String would be received'
+    'A object will be query String present else null String would be received',
   ].join('\n'),
 }, async function(req, res) {
   let response = req.params.test_auth || '';
--- a/services/auth/src/api.js
+++ b/services/auth/src/api.js
@@ -1119,3 +1119,25 @@ builder.declare({
   ]);
   res.reply({clientId, scopes});
 });
+
+/** Query string to authenticate **/
+builder.declare({
+  method: 'get',
+  route: '/test-authenticate?q={auth}',
+  query: {
+    prefix: /^[A-Za-z0-9!@/:.+|_-]+$/, // should match clientId above
+    continuationToken: /./,
+    limit: /^[0-9]+$/,
+  },
+  name: 'queryString',
+  output: 'test-authenticate-response.yml',
+  stability: 'stable',
+  title: 'Query String',
+  description: [
+    'Get the query string passed in the url for authentication',
+    'A object will be query String present else null String would be received'
+  ].join('\n'),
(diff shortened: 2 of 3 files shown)


jhford:
Hi @sudipt1999 I think you pushed the wrong branch to this pull request?

jhford:
Ahh, i realised what's happening.  @sudipt1999 you're using the same branch for two different Pull Requests.  That's generally not ideal because it can cause unrelated changes to get lumped together.

One workflow that I use (using my username) is:

```shell
# Setting up my git remotes
git remote add jhford github.com:taskcluster/taskcluster
git remote add upstream https://github.com:taskcluster/taskcluster
git fetch --all

# starting project-1
git checkout -t upstream/master -b project-1
vim index.js # or edit the project as you wish
git push origin project-1

# starting project-2
git checkout -t upstream/master -b project-2
vim index.js # or edit the project as you wish
git push origin project-2
```

Then in the web UI for creating the user interface can be done by loading

`https://github.com/taskcluster/taskcluster/compare/master...sudipt1999:project-1`

and 

`https://github.com/taskcluster/taskcluster/compare/master...sudipt1999:project-2`

Then you can always use `git rebase -i` to squash commits that aren't needed.

sudipt1999:
Really sorry ! I wasnt knowing that's going to happen!

On 27-Feb-2019 1:46 PM, "John Ford" <notifications@github.com> wrote:

Ahh, i realised what's happening. @sudipt1999
<https://github.com/sudipt1999> you're using the same branch for two
different Pull Requests. That's generally not ideal because it can cause
unrelated changes to get lumped together.

One workflow that I use (using my username) is:

# Setting up my git remotes
git remote add jhford github.com:taskcluster/taskcluster
git remote add upstream https://github.com:taskcluster/taskcluster
git fetch --all
# starting project-1
git checkout -t upstream/master -b project-1
vim index.js # or edit the project as you wish
git push origin project-1
# starting project-2
git checkout -t upstream/master -b project-2
vim index.js # or edit the project as you wish
git push origin project-2

Then in the web UI for creating the user interface can be done by loading

https://github.com/taskcluster/taskcluster/compare/master...sudipt1999:
project-1

and

https://github.com/taskcluster/taskcluster/compare/master...sudipt1999:
project-2

Then you can always use git rebase -i to squash commits that aren't needed.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/taskcluster/taskcluster/pull/313#issuecomment-467766250>,
or mute the thread

jhford:
not a problem!  It's something I done many times :)

sudipt1999:
Thanks for guiding me @jhford  really appreciate your help !

services/auth/src/v1.js
```diff
@@ -1064,6 +1064,12 @@ builder.declare({
     req.clientId(),
     req.scopes(),
   ]);
+  //setting headers as purposed by bug 161375
+  let credentials = {clientId, scopes};
+  let currentUrl = window.location.href;
+  let urlWithQueryString = currentUrl + '?' + 'test_auth' + '=' + credentials;
+  history.pushState(null, null, urlWithQueryString);//will add the query params to current url
```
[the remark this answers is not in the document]
sudipt1999: @djmitche  I am guessing you want me to embed  query params in req object similar to https://github.com/taskcluster/taskcluster/blob/c7670d403092259fe67ac1010e95275a09088e92/services/index/src/api.js#L191 ,

something like ` let continuation = req.query.continuationToken || null;` 

Kindly correct me if I am going in wrong direction!

Thank You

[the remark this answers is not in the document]
jhford: To figure out something similar window.location.href, you might find the [request object](https://nodejs.org/api/http.html#http_message_url) docs useful

sudipt1999:
@jhford  kindly review the changes then I can move on to writing test and updating description part for bug ! Thank You

djmitche:
I'll have a look

djmitche:
This is looking closer!  I'm not sure how clientId and accessToken fit in, though.

What we want is to be able to call a URL like `https://auth.taskcluster.net/v1/test-authenticate?q=abc` and get a response back indicating that `query-param` was received correctly.  John suggested returning that in a header, but it is probably easier to include the value in the response.  That will mean adding a new value to the [output schema](https://github.com/taskcluster/taskcluster/blob/master/services/auth/schemas/v1/test-authenticate-response.yml) for the method, and defining a new `query` parameter (similar to [this](https://github.com/taskcluster/taskcluster/blob/master/services/auth/src/v1.js#L169)) named `q`.  And then adding a line or two in the API method definition to set the response value when `q` is present, and omit it otherwise.

Once this is done, a good next step will be to add some tests to https://github.com/taskcluster/taskcluster-client.py that *use* this new functionality, to check that the client is passing queries correctly.

sudipt1999:
> What we want is to be able to call a URL like `https://auth.taskcluster.net/v1/test-authenticate?q=abc` and get a response back indicating that `query-param` was received correctly. John suggested returning that in a header, but it is probably easier to include the value in the response. That will mean adding a new value to the [output schema](https://github.com/taskcluster/taskcluster/blob/master/services/auth/schemas/v1/test-authenticate-response.yml) for the method, and defining a new `query` parameter (similar to [this](https://github.com/taskcluster/taskcluster/blob/master/services/auth/src/v1.js#L169)) named `q`. And then adding a line or two in the API method definition to set the response value when `q` is present, and omit it otherwise.
How can I test this end point ! Kindly point me I was unable to do that

djmitche:
There are tests in [`testauth_test.js`](https://github.com/taskcluster/taskcluster/blob/master/services/auth/test/testauth_test.js)

services/auth/schemas/v1/test-authenticate-response.yml
```diff
@@ -20,4 +20,9 @@ properties:
       type:                 string
       pattern: "^[\x20-\x7e]*$"
     uniqueItems:            true
+  queryString:
+    description: |
+      Query string for the testAuthenticate
+    type:                       string
+    additionalProperties:       false
```
djmitche: I like the plan to return the query string as a property of the response body like this.  Please also add `queryString` to the `required` section just below.

services/auth/src/api.js
```diff
@@ -1060,6 +1060,9 @@ builder.declare({
     req.clientId(),
     req.scopes(),
   ]);
+  //setting headers as purposed by bug 161375
+  req.params.test_auth = {clientId, scopes};
+  res.setHeader('X-test_auth', {clientId, scopes});
```
djmitche: I apologize for contracting jhford, but I don't think a header is a good choice here -- all of the other data about the request comes back in the response body.  Also, the Taskcluster client libraries don't make response headers available, so there would be no way to look for this header.  So please change this bit to include `queryString` in the response (`res.reply`) instead.

services/auth/src/api.js
```diff
@@ -1060,6 +1060,9 @@ builder.declare({
     req.clientId(),
     req.scopes(),
   ]);
+  //setting headers as purposed by bug 161375
+  req.params.test_auth = {clientId, scopes};
```
djmitche: I'm not sure why this is *setting* something in `req.params`.  The request should be read-only..

services/auth/src/api.js
```diff
@@ -1060,6 +1060,9 @@ builder.declare({
     req.clientId(),
     req.scopes(),
   ]);
+  //setting headers as purposed by bug 161375
```
djmitche: Including a bug number in a comment is useful when a reader might otherwise wonder "why was this written??".  But in this case, at least once the patch is finished, the reasoning should be pretty clear from reading the code, so there won't be a need to reference the bug, and even a comment may feel somewhat redundant.

services/auth/src/api.js
```diff
@@ -1116,3 +1119,25 @@ builder.declare({
...
+  stability: 'stable',
+  title: 'Query String',
+  description: [
+    'Get the query string passed in the url for authentication',
+    'A object will be query String present else null String would be received',
+  ].join('\n'),
+}, async function(req, res) {
+  let response = req.params.test_auth || '';
+  res.reply(response);
+});
```
djmitche: I don't think we need to add a new API method -- the task is to add functionality to the existing testAuthentication endpoint.

services/auth/schemas/v1/test-authenticate-response.yml
```diff
@@ -20,4 +20,9 @@ properties:
       type:                 string
       pattern: "^[\x20-\x7e]*$"
     uniqueItems:            true
+  queryString:
+    description: |
+      Query string for the testAuthenticate
```
djmitche: Let's expand this a little: where did this value come from?  Help the reader of the API documentation figure out how to use this value.

sudipt1999:
Yes ! I will finish this up

sudipt1999:
As you wish sir !

PR #319, opened 2019-02-27
Readme.md file updated for bug 1527969

sudipt1999:
Bug fix -> Add function to lib-loader that is for using when file is executed
Updated the documented for its use!


djmitche:
Please add this change to the same branch as you used in #283 (`crashOnError`) and push it there, rather than in a new PR.

PR #323, opened 2019-02-28
Bug 1527424 - clean up error handling in TaskCreator

djmitche:
This refactors TaskCreator.fire to do a better job of error handling,
consistently updating LastFire and returning an API response or an
error.

This also drops an unused schema file and deletes some additional (now
moot) modifications of the lastFire field in the Hooks table.

Bugzilla Bug: [1527424](https://bugzilla.mozilla.org/show_bug.cgi?id=1527424)

cc: @ydidwania

### DIFF
--- a/services/hooks/src/listeners.js
+++ b/services/hooks/src/listeners.js
@@ -84,7 +84,12 @@ class HookListeners {
         hookId: hookId,
       }, true);
       if (latestHook) {
-        await this.taskcreator.fire(latestHook, {firedBy: 'pulseMessage', payload});
+        try {
+          await this.taskcreator.fire(latestHook, {firedBy: 'pulseMessage', payload});
+        } catch (err) {
+          // any errors were already reported via the LastFire table, so they
+          // can be safely ignored here
+        }
       }
     });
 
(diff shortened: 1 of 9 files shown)


services/hooks/src/listeners.js
```diff
@@ -84,7 +84,12 @@ class HookListeners {
         hookId: hookId,
       }, true);
       if (latestHook) {
-        await this.taskcreator.fire(latestHook, {firedBy: 'pulseMessage', payload});
+        try {
+          await this.taskcreator.fire(latestHook, {firedBy: 'pulseMessage', payload});
```
imbstack: Nit: Would it make more sense to have the try/catch inside taskcreator if it should never throw an error anyway?

services/hooks/src/taskcreator.js
```diff
@@ -79,81 +83,88 @@ class TaskCreator {
...
-      return {status: {taskId: options.taskId}};
-    }
-
-    let lastFire, taskCreateRes, fireError;
-    try {
-      taskCreateRes = await queue.createTask(options.taskId, task);
-      lastFire = {
-        result: 'success',
+    // Inner implementation, returning
+    // `lastFie` (entry to insert into lastFire row, if any)
```
imbstack: `lastFie`

services/hooks/src/listeners.js
djmitche: For both `Hooks.triggerHook` and the time-based scheduling, we *do* want to handle this error (the former to pass it along to the user; the latter to decide whether to retry the fire on the next iteration).  I thought of adding a `throwError: ..` option to `taskcreator.fire`, but this seemed a bit more natural.

djmitche:
landed and deployed in heroku.

PR #329, opened 2019-03-01
[bug 1531576] Add clientId to express logs and remove unstructured version

imbstack:
According to the metrics in the bug, the `console.log` at the end of authorization is the source of a lot of the logging volume currently. If we merge some extra information into the express metrics, we can avoid logging two lines.

Bugzilla Bug: [1531576](https://bugzilla.mozilla.org/show_bug.cgi?id=1531576)

### DIFF
--- a/libraries/api/src/middleware/auth.js
+++ b/libraries/api/src/middleware/auth.js
@@ -130,22 +130,22 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
     // so this is probably a fair policy for now. We can always allow more.
     if (req.headers && req.headers.authorization &&
         req.query && req.query.bewit) {
-      return Promise.resolve({
+      return {
         status: 'auth-failed',
         message: 'Cannot use two authentication schemes at once ' +
                   'this request has both bewit in querystring and ' +
                   'and \'authorization\' header',
-      });
+      };
     }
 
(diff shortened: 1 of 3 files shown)


libraries/api/src/middleware/auth.js
```diff
@@ -130,22 +130,22 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
     // so this is probably a fair policy for now. We can always allow more.
     if (req.headers && req.headers.authorization &&
         req.query && req.query.bewit) {
-      return Promise.resolve({
```
djmitche: eww, good fix

libraries/monitor/src/monitor.js
```diff
@@ -131,6 +131,8 @@ class Monitor {
 
           this.log.expressTimer({
             name,
+            hasAuthed: req.hasAuthed,
+            clientId: req.hasAuthed ? await req.clientId() : '',
```
djmitche: neat, so this will avoid logging `auth-failed` for things that didn't need auth..

imbstack: I think it is also required because an endpoint with no scopes will need to talk to auth to do `req.clientId()` whereas it is cheap to do for an already authed endpoint.

imbstack:
Superseded by #417

PR #341, opened 2019-03-02
Bug 1519969 - Sending an email to an invalid address fails

sudipt1999:
Bug 1519969 
https://bugzilla.mozilla.org/show_bug.cgi?id=1519969

currently stuck at the implementation to find email of the person creating the Hook !

Kindly help me out here !

### DIFF
--- a/generated/references.json
+++ b/generated/references.json
@@ -4963,6 +4963,7 @@
         },
         "owner": {
           "description": "Email of the person or group responsible for this hook.",
+          "format": "email",
           "maxLength": 255,
           "title": "Owner",
           "type": "string"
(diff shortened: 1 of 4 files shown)


djmitche:
The schemas control the input to the API method.  You can find out about JSON schemas [here](https://json-schema.org/understanding-json-schema/).

Since we already copy the hook's "owner" field into the notify service's "email" field, we need to make sure that the schema for the "owner" field is the same as that for the "email" field.  So, find the schemas describing those two fields, and make the "owner" field's requirements match those of the "owner" field.

sudipt1999:
I currently added the `format: email` to owner field and remove `maxLength:255` since it was not mentioned in email field of taskcluster-notify. I hope I am moving in right direction here , since you mentioned  

> we already copy the hook's "owner" field into the notify service's "email" field

I guess that will fix the bug

PR #343, opened 2019-03-02
No more check staging

djmitche:
This model likely reappear in [Bug 1492271](https://bugzilla.mozilla.org/show_bug.cgi?id=1492271), and can be found in the Git history if necessary.  For the moment staging checks are not working, so let's remove them.


djmitche:
By "this model" I mean doing quick checks to make sure a deployed service is more-or-less working.

PR #344, opened 2019-03-02
Renaming .md to .mdx

iFlameing:
Fixes: Bugzilla Bug: [1530790](https://bugzilla.mozilla.org/show_bug.cgi?id=1530790)

@djmitche can you review this?


iFlameing:
@djmitche I think we should wait for @helfi92. Because right now no option solve this bug completely. what do you think?

djmitche:
It's true, but it will correctly display README.md files :)

djmitche:
Maybe we should do this work *after* we move the files from `services/*/docs` into `ui/docs` -- it will probably make a lot more sense at that point.  And the READMEs will not move (although some of their content might be moved out to documentation files).

helfi92:
> Maybe we should do this work after we move the files from services/*/docs into ui/docs -- it will probably make a lot more sense at that point. And the READMEs will not move (although some of their content might be moved out to documentation files).

Agreed. I think that's a good plan!

djmitche:
Closing to continue after the move..

PR #355, opened 2019-03-04
Fix oneshot exit in test

imbstack:


imbstack:
Interestingly this seems to have been broken for a long time. Long enough that there was an old `new Monitor` invocation hidden by this last fix. It seems like zurvan and iterate just don't play nicely together.

imbstack:
Tests pass on retry.

PR #360, opened 2019-03-05
Adding versions in docs path for api and event reference

iFlameing:
Bugzilla Bug: [1461815](https://bugzilla.mozilla.org/show_bug.cgi?id=1461815)

@djmitche I read the code present in the (infrastructure/builder/src/generate) and found that service-docs.js and docs-tocs.js are used for copying files from service folder. I made changes to only service folder because when we run yarn generate, ui/docs/generated file will automatically update with the added version.

can you comment If I am in the right direction or not? And what will be my next step to solve this issue.

### DIFF
--- a/infrastructure/builder/src/generate/generators/docs-tocs.js
+++ b/infrastructure/builder/src/generate/generators/docs-tocs.js
@@ -70,12 +70,11 @@ async function readGeneratedDocs() {
     // Rename keys to include the section, tier and service name
     // e.g., docs/intro -> reference/integrations/github/docs/intro
     Object.keys(mdFiles).forEach(oldKey => {
-      const path = `reference/${metadata.tier}/${projectName}/${oldKey}`;
+      let path = `reference/${metadata.tier}/${projectName}/${oldKey}`;
 
       if(oldKey === "references/events" || oldKey === "references/api"){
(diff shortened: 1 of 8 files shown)


infrastructure/builder/src/generate/generators/docs-tocs.js
```diff
@@ -70,8 +70,11 @@ async function readGeneratedDocs() {
     // Rename keys to include the section, tier and service name
     // e.g., docs/intro -> reference/integrations/github/docs/intro
     Object.keys(mdFiles).forEach(oldKey => {
-      const path = `reference/${metadata.tier}/${projectName}/${oldKey}`;
+      let path = `reference/${metadata.tier}/${projectName}/${oldKey}`;
 
+      if(oldKey === "references/events" || oldKey === "references/api"){
+        path = `reference/${metadata.tier}/${projectName}/v1/${oldKey}`;
+      }
```
djmitche: I think there's an earlier phase where these `mdFiles` are created - that would be the place to do this naming.  There, you'll have access to the reference document itself, which has an `apiVersion` property that should be used to give the version -- in theory, someday that may be something other than `v1` :)

djmitche:
There was a big docs reorganization that made this conflict quite a bit.  The task is still open, though!

PR #362, opened 2019-03-06
[Bug 1532360] queue: lastClaimed api call

aerickson:
Bugzilla Bug: [1532360](https://bugzilla.mozilla.org/show_bug.cgi?id=1532360)

TODO: 
- move test to better place?


aerickson:
I added a breaking test. This should be red.

aerickson:
This won't work (well). I'm sure we run multiple instances of this service... so a value will probably come back, but it likely won't be the latest time a task was consumed.

services/queue/src/api.js
```diff
@@ -2127,6 +2128,34 @@ builder.declare({
...
+  route: '/last_claimed/:provisionerId/:workerType',
+  name: 'lastClaimed',
+  stability: APIBuilder.stability.experimental,
+  output: 'last-claimed-response.yml',
+  title: 'Get Timestamp When Task Last Claimed',
+  description: [
+    'Get a unix epoch time indicating when a task was claimed for the',
+    'given `provisionerId` and `workerType`.',
+    '',
+    'The call returns -1 if it has no record of a task being claimed.',
```
djmitche: We typically use "JSON" timestamps (ISO8601).

services/queue/src/queueservice.js
```diff
@@ -144,6 +144,10 @@ class QueueService {
     // Keep a cache of pending counts as mapping:
     //    <provisionerId>/<workerType> -> {lastUpdated, count: promise}
     this.countPendingCache = {};
+
+    // Keep a record of last claimed for each queue as mapping:
+    //    <provisionerId>/<workerType> -> lastClaimed
+    this.lastClaimedMap = {};
```
djmitche: We currently run 25 instances of the REST API server, so yeah, this single-process map won't work :)

PR #364, opened 2019-03-06
[Bug 1523801] Blacklist listening to some specific exchanges

ydidwania:
This PR involves checking the exchanges in the hook definition against a `denyList`at creation/updation of a hook. I have also written 2 tests one for each.
The `config.yml` looks like
```
pulse:
      denylist: 
           exchanges: [taskcluster-queue, taskcluster-hooks]
```
@djmitche I am guessing you'd want some changes to the above format. Do let me know while reviewing :)

Bugzilla Bug: [1523801](https://bugzilla.mozilla.org/show_bug.cgi?id=1523801)

### DIFF
--- a/services/hooks/src/main.js
+++ b/services/hooks/src/main.js
@@ -129,7 +129,7 @@ const load = loader({
     requires: ['cfg', 'schemaset', 'Hook', 'LastFire', 'taskcreator', 'monitor', 'publisher', 'pulseClient'],
     setup: ({cfg, schemaset, Hook, LastFire, taskcreator, monitor, publisher, pulseClient}) => builder.build({
       rootUrl: cfg.taskcluster.rootUrl,
-      context: {Hook, LastFire, taskcreator, publisher, denyList: cfg.pulse.denylist.exchanges},
+      context: {Hook, LastFire, taskcreator, publisher, denylist: cfg.pulse.denylist},
       schemaset,
       publish: cfg.app.publishMetaData,
       aws: cfg.aws.validator,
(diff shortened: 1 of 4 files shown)


services/hooks/src/main.js
```diff
@@ -129,7 +129,7 @@ const load = loader({
     requires: ['cfg', 'schemaset', 'Hook', 'LastFire', 'taskcreator', 'monitor', 'publisher', 'pulseClient'],
     setup: ({cfg, schemaset, Hook, LastFire, taskcreator, monitor, publisher, pulseClient}) => builder.build({
       rootUrl: cfg.taskcluster.rootUrl,
-      context: {Hook, LastFire, taskcreator, publisher},
+      context: {Hook, LastFire, taskcreator, publisher, denyList: cfg.pulse.denylist.exchanges},
```
djmitche: I think we'll introduce bugs by spelling it both `denylist` and `denyList`.  Let's just use the first (all-lowercase).

services/hooks/src/v1.js
```diff
@@ -573,6 +594,16 @@ const triggerHookCommon = async function({req, res, hook, payload, clientId, fir
   }
 };
 
+const isDeniedBinding = async ({bindings, denyList}) => {
+  let denyPattern = new RegExp(`^exchange/(${denyList.join('|')})/`);
+  let denied=false;
+  bindings.forEach((binding) => {
```
djmitche: this could also be done with return [`!bindings.every(..)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) (but no need to change it).

djmitche:
I'll merge when the tests pass

ydidwania:
Okay!

PR #365, opened 2019-03-07
Point bug-filers to Bugzilla

djmitche:
Based on discussion in this week's team meeting.  I tried to be brief, rather than enumerating a long decision tree for figuring out where to post something.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,6 +7,11 @@ assignees: ''
 
 ---
 
+<!-- NOTE: The Taskcluster team uses Bugzilla as its primary bug-tracking tool.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -7,6 +7,11 @@ assignees: ''
 
 ---
 
+<!-- NOTE: The Taskcluster team uses Bugzilla as its primary bug-tracking tool.
```
imbstack: Probably "if this issue is not a UI bug or is more than a quick fix". At least to my eye, the current way accidentally implies that UI bugs are simple.

djmitche: Yeah, def not what I meant :)

djmitche:
[pinned issue](https://github.com/taskcluster/taskcluster/issues/366)

PR #369, opened 2019-03-07
Speed back up queue tests

imbstack:
This makes some weirdness from changing time mocks before teardown a non-issue I think.

The tests are now fast again but also actually run.

### DIFF
--- a/services/queue/test/helper.js
+++ b/services/queue/test/helper.js
@@ -461,16 +461,21 @@ exports.withPollingServices = (mock, skipping) => {
       helper.load.remove(service);
       return svc;
     };
+    helper.stopPollingService = async () => {
+      if (svc) {
+        await svc.terminate();
+        svc = null;
+      }
+    };
   });
 
   teardown(async function() {
     if (svc) {
-      await svc.terminate();
-      svc = null;
+      throw new Error('Must call stopPollingService if you have started a service');
     }
   });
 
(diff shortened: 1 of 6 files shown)


services/queue/test/helper.js
```diff
@@ -461,16 +461,18 @@ exports.withPollingServices = (mock, skipping) => {
...
+    };
   });
 
-  teardown(async function() {
+  suiteTeardown(async function() {
     if (svc) {
-      await svc.terminate();
-      svc = null;
+      throw new Error('Must call stopPollingService if you have started a service');
     }
```
djmitche: Why does this happen after the suite (`suiteTeardown`) and not after each test (`teardown`)?  Also, in a case where an assertion fails in a test, it won't get to the `stopPollingService`, so this error will appear as well -- but the next test case will run anyway (perhaps with zurvan still active?)

imbstack: This should not have been in `suiteTeardown`. My brain just stuck it there when it saw "teardown". We can't do the stopping in a `teardown` because that happens _after_ zurvan has stopped and somehow that breaks things. I haven't been able to figure out exactly what it breaks.

djmitche: Ahhhh, got it, thanks and sorry.  Maybe add that to the error message or in a comment so next time I forget I will be reminded?

PR #371, opened 2019-03-08
Bug 1532469 - use a yarn cache

djmitche:
Bugzilla Bug: [1532469](https://bugzilla.mozilla.org/show_bug.cgi?id=1532469)


djmitche:
I pushed a test change to this which counted the number of files under /yarn-cache, and it was nonzero.  So I think this is working.

PR #372, opened 2019-03-09
[Bug 1518568] Part 2: Enable remaining useful lints (other than 1)

imbstack:
After this lands, I consider the work on linting Good Enough For Now. 

These specific rules seem to have caught a lot of actual bugs!

Bugzilla Bug: [1518568](https://bugzilla.mozilla.org/show_bug.cgi?id=1518568)


services/queue/test/blobstore_test.js
```diff
@@ -1,198 +0,0 @@
-suite.skip('queue/tasks_store', function() {
```
djmitche: Is this removed because we never use aritfacts with type = `azure`?

services/queue/test/querytasks_test.js
```diff
@@ -70,17 +69,7 @@ helper.secrets.mockSuite(__filename, ['taskcluster', 'aws', 'azure'], function(m
...
-    await testing.poll(async () => {
-      // At some point in the future we have to got fetch a new result saying
-      // more tasks are now in the queue...
-      const r3 = await helper.queue.pendingTasks(
-        'no-provisioner-extended-extended',
-        'query-test-worker-extended-extended',
-      );
-      assume(r3.pendingTasks).is.greaterThan(r1.pendingTasks);
-    }, 30, 1000);
+    return;
```
djmitche: Remove the return statement and the comment, too..

libraries/validate/test/validate_test.js
```diff
@@ -80,19 +80,15 @@ suite('Valid Schema Tests', () => {
...
-        folder: 'test/schemas',
-        constants: 'test/schemas/constants.yml',
-        serviceName: 'whatever',
-      });
-      let error = v(
-        {value: 43},
-        libUrls.schema(rootUrl, 'whatever', '/v1/yml-test-schema#'));
-      return assert.equal(error, null);
-    } catch (err) {
-      return err;
```
djmitche: wat

.eslintrc.yml
```diff
@@ -68,9 +68,6 @@ rules:
   no-unreachable: error
   no-undef: error
 
-  # These are disabled for now and we will slowly turn them on
-  require-await: off
```
djmitche: The ["When not to use it"](https://eslint.org/docs/rules/require-await#when-not-to-use-it) section seems to apply here!

libraries/validate/test/validate_test.js
imbstack: This blames to me but I had the same reaction when I saw it.

services/queue/test/blobstore_test.js
imbstack: The first line is just a `return` so I figured if we were never running it, may as well delete it.

djmitche: Oh I totally missed that.  I added that in fe4b31917e, probably intending to fix it later.  But, I don't envision a world where we ever use azure blob artifacts via this API, so I think deleting it is reasonable.

PR #376, opened 2019-03-09
Bug 1533937 - centralize azure table setup in testing

djmitche:
Bugzilla Bug: [1533937](https://bugzilla.mozilla.org/show_bug.cgi?id=1533937)


djmitche:
Now ready for review :)

I'm not sure how or even if I'll continue on this bug (to try to delete unused tables) but this part is self-sufficient.

PR #385, opened 2019-03-11
Fix table name in github tests

djmitche:
This changes the table name to match elsewhere in the code.  #376 is centralizing withEntities and will require that these match everywhere.


services/github/test/helper.js
```diff
@@ -98,7 +98,7 @@ exports.withEntities = (mock, skipping) => {
     exports.buildsTableName = `TaskclusterGithubBuildsV${tableVersion}`;
     exports.ownersTableName = `TaskclusterIntegrationOwnersV${tableVersion}`;
     exports.checkRunsTableName = `TaskclusterCheckRunsV${tableVersion}`;
-    exports.checksToTasksTableName = `TaskclustersToTasksV${tableVersion}`;
```
owlishDeveloper: oh deer

djmitche: :deer:

PR #387, opened 2019-03-11
Allow listening to the queue's task-completed exchange

djmitche:
This suits bastien's needs, in particluar.

/cc @ydidwania

### DIFF
--- a/services/hooks/config.yml
+++ b/services/hooks/config.yml
@@ -51,7 +51,13 @@ defaults:
     password:       !env PULSE_PASSWORD
     hostname:       !env PULSE_HOSTNAME
     vhost:          !env PULSE_VHOST
-    denylist:       [taskcluster-queue, taskcluster-hooks]
+    # a list of regular expressions for exchanges that are prohibited
+    denylist:
+     # triggering a hook on task-created could create a "task bomb" that just spews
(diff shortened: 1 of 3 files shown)


services/hooks/config.yml
```diff
@@ -51,7 +51,13 @@ defaults:
...
     hostname:       !env PULSE_HOSTNAME
     vhost:          !env PULSE_VHOST
-    denylist:       [taskcluster-queue, taskcluster-hooks]
+    # a list of regular expressions for exchanges that are prohibited
+    denylist:       
+     # triggering a hook on task-created could create a "task bomb" that just spews
+     # out tasks with no end, and isn't useful otherwise
+     - exchange/taskcluster-queue/v1/task-created
+     # changes to hooks themselves are internal and shouldn't trigger hooks
+     - exchange/taskcluster-hooks/.*
```
ydidwania: The `.*` in the end does not help anywhere I believe but its not bad either :)

djmitche: yeah, makes it clear that it's a regexp

djmitche:
I refactored the isDeniedBinding function.  Something about its construction of a regexp wasn't working correctly.

PR #389, opened 2019-03-11
Bug 1523801 : Blacklist listening to some specific exchanges

ydidwania:
Display the bindings of the hook definition in the error message in case of denied exchanges being used.

Bugzilla Bug: [1523801](https://bugzilla.mozilla.org/show_bug.cgi?id=1523801)


djmitche:
> One or more of the exchanges below have been denied access to hooks [{"exchange":"exchange/taskcluster-hooks/v1/abc","routingKeyPattern":"#"}]

looks good!

PR #400, opened 2019-03-12
Send notification on failures/exceptions in master

imbstack:
This will currently send an email for each task that fails. Perhaps later we can make a summary task that sends a single email instead.


djmitche:
conceptual :100: from me :)

PR #401, opened 2019-03-12
Bug 1534711 - pass rootUrl when initializing tc-client

djmitche:
There don't appear to be any tests to run the scanner.  Login never got a lot of tests -- it used to be mostly UI.  And it will soon be merged into web-server, so I'm not sure it's worth adding any.  Even if we did have tests, they would have faked out the auth service, so this *particular* bug would have snuck through.

Bugzilla Bug: [1534711](https://bugzilla.mozilla.org/show_bug.cgi?id=1534711)

### DIFF
--- a/services/login/src/scanner.js
+++ b/services/login/src/scanner.js
@@ -16,7 +16,10 @@ async function scanner(cfg, handlers) {
 
   // NOTE: this function performs once auth operation at a time.  It is better
   // for scans to take longer than for the auth service to be overloaded.
-  let auth = new taskcluster.Auth({credentials: cfg.app.credentials});
+  let auth = new taskcluster.Auth({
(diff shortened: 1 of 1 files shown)


djmitche:
Maybe our deadline is a little too strict?

imbstack:
20 minutes should be plenty for small tests like this. I keep checking the queues and number of workers the provisioner has and it seems to not be at the top.  Maybe it is just taking too long a time to spin up new workers.

I'd almost rather have us figure out how to get this to work correctly than just increase our timeout.

djmitche:
I think deadline can be longer to include provisioning issues -- better late than failed.  The maxRunTime is already set pretty short to avoid waiting for long-running test suites.

PR #403, opened 2019-03-12
Fix auth with entities

djmitche:
That's at least three fewer of my dumb mistakes in the codebase.  A drop in the bucket..


imbstack:
Conservation of Bugmentum

PR #417, opened 2019-03-13
[Bug 1535125] Part 1: Update critical part of lib-api and auth/queue

imbstack:
Still lots of services/libraries to go after this. Also haven't begun to tackle `debug()` usage yet.

Bugzilla Bug: [1535125](https://bugzilla.mozilla.org/show_bug.cgi?id=1535125)

### DIFF
--- a/generated/references.json
+++ b/generated/references.json
@@ -11535,15 +11535,15 @@
           "fields": {
             "clientId": "The clientId that was authenticated.",
             "expires": "After this time the authentication is no longer valid.",
-            "hash": "The hash of the request payload, if there was one.",
+            "hash": "The hash of the request payload of the original request, if there was one.",
             "host": "The hostname of the authorizer of the request.",
             "message": "A reason for failure if failed. Otherwise empty.",
             "method": "HTTP method of the original request.",
-            "port": "The port of the request from the authorizer.",
+            "port": "The port of the original request.",
             "resource": "The path of the original request",
             "scheme": "Currently always `hawk`.",
             "scopes": "All scopes that this client has available.",
-            "sourceIp": "The ip address the original request came from.",
+            "sourceIp": "The ip address of the original request.",
             "status": "Whether or not the authentication was successful."
           },
           "type": "signature-validation",
--- a/generated/references.json
+++ b/generated/references.json
@@ -7524,10 +7524,15 @@
           "version": 1
         },
         {
-          "description": "A timer for express endpoints.",
+          "description": "A timer and audit for express endpoints.\nYou can combine this with auth audit logs to get\na complete picture of what was authorized when.\n\nHere, anything that is not public should have authenticated\nand if it authenticated, we will tell the clientId here. Given\nthat, it is not necessarily true that the endpoint was\n_authorized_. You can tell that by the statusCode.",
           "fields": {
+            "clientId": "The clientId that made the request if there was one.",
             "duration": "The duration in ms of the endpoint.",
+            "hasAuthed": "Will be true if the request had req.authorize() called on it.",
+            "method": "The http method of the request. Should match the one in auth audit log.",
             "name": "The name of the endpoint.",
+            "public": "Will be true if the endpoint requires no scopes.",
+            "resource": "The path of the http request. Should match the one in auth audit log.",
             "statusCode": "The http status code that the endpoint resolved with."
           },
           "type": "monitor.express",
@@ -7763,10 +7768,15 @@
           "version": 1
         },
         {
-          "description": "A timer for express endpoints.",
+          "description": "A timer and audit for express endpoints.\nYou can combine this with auth audit logs to get\na complete picture of what was authorized when.\n\nHere, anything that is not public should have authenticated\nand if it authenticated, we will tell the clientId here. Given\nthat, it is not necessarily true that the endpoint was\n_authorized_. You can tell that by the statusCode.",
           "fields": {
+            "clientId": "The clientId that made the request if there was one.",
             "duration": "The duration in ms of the endpoint.",
+            "hasAuthed": "Will be true if the request had req.authorize() called on it.",
+            "method": "The http method of the request. Should match the one in auth audit log.",
             "name": "The name of the endpoint.",
+            "public": "Will be true if the endpoint requires no scopes.",
+            "resource": "The path of the http request. Should match the one in auth audit log.",
             "statusCode": "The http status code that the endpoint resolved with."
           },
           "type": "monitor.express",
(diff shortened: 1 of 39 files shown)


generated/references.json
```diff
@@ -11436,6 +11436,33 @@
...
+          "description": "Records results of authenticating a client.\nThis contains the verified hawk headers that were sent\nto auth in addition to info about the client.\n\nIn these fields, there are two requests that might be\nmentioned. The first is the original request from a\nclient to a service and the second is the request from\nthe service to auth itself.",
+          "fields": {
+            "clientId": "The clientId that was authenticated.",
+            "expires": "After this time the authentication is no longer valid.",
+            "hash": "The hash of the request payload, if there was one.",
+            "host": "The hostname of the authorizer of the request.",
+            "message": "A reason for failure if failed. Otherwise empty.",
+            "method": "HTTP method of the original request.",
+            "port": "The port of the request from the authorizer.",
+            "resource": "The path that the original request",
```
djmitche: *of

services/auth/src/scoperesolver.js
```diff
@@ -276,11 +276,14 @@ class ScopeResolver extends events.EventEmitter {
       // such expansion we'll continue and compute one.
       const cacheKey = queue.join('\n');
       const cacheResult = lru.get(cacheKey);
+
+      this._monitor.log.scopeResolver({
+        cacheHit: cacheResult ? true : false,
+      });
+
```
djmitche: Thoughts on dropping this entirely?

services/queue/src/monitor.js
```diff
@@ -4,4 +4,46 @@ const manager = new MonitorManager({
   serviceName: 'queue',
 });
 
+manager.register({
+  name: 'resolvedQueuePoll',
+  type: 'resolved-queue-poll',
```
djmitche: This name's a little confusing, since the log format is used for the resolved queue but also the deadline queue and the claim queue.  Maybe `azureQueuePoll`?

services/queue/src/monitor.js
```diff
@@ -4,4 +4,46 @@ const manager = new MonitorManager({
...
+  version: 1,
+  level: 'info',
+  description: 'Report result of polling messages from an azure queue.',
+  fields: {
+    messages: 'Number of messages fetched.',
+    failed: 'Number of these messages that failed to be handled.',
+    resolver: 'The name of the resolver that polled the queue.',
```
djmitche: Could this be the name of the queue?

libraries/monitor/src/builtins.js
```diff
@@ -25,13 +25,25 @@ module.exports = [
   {
     name: 'expressTimer',
     type: 'monitor.express',
```
djmitche: maybe this should be `apiMethod`/`monitor.api-method`?

ui/src/views/Documentation/Reference/index.jsx
```diff
@@ -24,7 +24,12 @@ export default class Reference extends Component {
     const topicExchangeEntries =
       json.entries &&
       json.entries.filter(({ type }) => type === 'topic-exchange');
-    const logTypes = json.types;
+    const builtinLogTypes = json.types.filter(l =>
```
djmitche: maybe "commonLogTypes" and adjust the heading in the DOM to match?

services/auth/src/scoperesolver.js
imbstack: Happy to do so.

services/queue/src/monitor.js
imbstack: The resolver itself doesn't actually know the name of the queue it is pulling from. I could add it in if wanted?

djmitche: I think I was just suggesting a change to the description: "The name of the queue being polled".

djmitche: (it doesn't have to be the actual name of the Azure object)

libraries/monitor/src/monitor.js
```diff
@@ -120,7 +124,7 @@ class Monitor {
     return (req, res, next) => {
       let sent = false;
       const start = process.hrtime();
-      const send = () => {
+      const send = async () => {
```
owlishDeveloper: I see this is now `async` - are we `await`ing something?

libraries/api/src/middleware/auth.js
```diff
@@ -130,22 +122,22 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
     // so this is probably a fair policy for now. We can always allow more.
     if (req.headers && req.headers.authorization &&
         req.query && req.query.bewit) {
-      return Promise.resolve({
+      return {
```
owlishDeveloper: We used to return a `Promise`, and now just an object - is that ok?

libraries/api/src/middleware/auth.js
```diff
@@ -159,14 +151,14 @@ const remoteAuthentication = ({signatureValidator, entry}) => {
     }
 
     // Send input to signatureValidator (auth server or local validator)
-    let result = await Promise.resolve(signatureValidator({
+    let result = await signatureValidator({
```
owlishDeveloper: Same here: used to be a `Promise`, and now a value. (in fact, everywhere in this file). Is that what's expected by callers?

services/auth/test/scoperesolver_test.js
```diff
@@ -2,15 +2,14 @@ const helper = require('./helper');
 const ScopeResolver = require('../src/scoperesolver');
 const exchanges = require('../src/exchanges');
 const {scopeCompare} = require('taskcluster-lib-scopes');
-const MonitorManager = require('taskcluster-lib-monitor');
+const monitorManager = require('../src/monitor');
```
owlishDeveloper: it's the same as was before, isn't it? 🤔

services/queue/src/bucket.js
```diff
@@ -168,7 +168,7 @@ Bucket.prototype.setupCORS = async function() {
     this.monitor.reportError(err, 'warning');
   }
 
-  // Set CURS
```
owlishDeveloper: 🧙‍♂️

libraries/monitor/src/monitor.js
imbstack: Yep, the `req.clientId` below need to be awaited now.

services/auth/test/scoperesolver_test.js
imbstack: Not quite. The `src/monitor` does some setup to lib-monitor first.

services/queue/src/bucket.js
imbstack: :star2:

libraries/api/src/middleware/auth.js
imbstack: I'll go through and double check to make sure. I definitely don't want to break this part.

imbstack: Ah, there was one `signatureValidator` that wasn't an `async`! I've updated it to be one and also brought back in the `Promise.resolve` for that specific part just in case.

Nice catch.

imbstack:
Ok, I have pushed this to my dev instance, ran through a few simple actions to see if things still work:

1. Logged in as root and created a new client with no scopes
2. Logged in as the new client and attempted to create a new user (properly rejected and no client was created upon checking)
3. Created and viewed secrets as the root client

This makes me think that auth still works properly and is not accidentally permissive or broken. The logs seem to make sense in all of the cases as well.

PR #418, opened 2019-03-14
A foolish consistency

djmitche:


djmitche:
OK, I think I finally got this right.  This was probably a good use for a "draft" PR.

PR #419, opened 2019-03-14
Worker manager in dev

imbstack:
This seems to mostly work for me? I can't see the provisioner doing anything but the api runs and responds to pings.


owlishDeveloper:
It probably won't because `InMemoryDataStore` is used only for `test`. I am hoping we should be able to implement an Azure data store class

infrastructure/terraform/auth-service.tf
```diff
@@ -77,6 +77,13 @@ locals {
         "queue:get-artifact:*",
       ]
     },
+    {
+      clientId    = "static/taskcluster/worker-manager"
+      accessToken = "${random_string.worker_manager_access_token.result}"
+      description = "..."
+
+      scopes = []
```
owlishDeveloper: Why don't we add `'worker-manager:manage-worker-configuration:*` and `worker-manager:manage-worker-configuration:*`?

infrastructure/terraform/ingress.yaml
```diff
@@ -36,10 +36,10 @@ spec:
             - {service_name: queue, project_name: taskcluster-queue}
             - {service_name: hooks, project_name: taskcluster-hooks}
             - {service_name: index, project_name: taskcluster-index}
-            - {service_name: gce-provider, project_name: gce-provider}
```
owlishDeveloper: ❗️

infrastructure/terraform/worker-manager-service.tf
```diff
@@ -0,0 +1,48 @@
...
+  override_special = "_-"
+}
+
+module "worker_manager_secrets" {
+  source            = "modules/service-secrets"
+  project_name      = "taskcluster-worker-manager"
+  disabled_services = "${var.disabled_services}"
+
+  secrets = {
+    TASKCLUSTER_CLIENT_ID    = "static/taskcluster/worker-manager"
```
owlishDeveloper: aws creds as well?

infrastructure/terraform/auth-service.tf
imbstack: A service has its own scopes, if that makes any sense. I think it will definitely need some more scopes to work correctly, but this at least gets it running.

infrastructure/terraform/ingress.yaml
imbstack: Haha, as discussed in irc, this was a crappy prototype that I made.

infrastructure/terraform/worker-manager-service.tf
imbstack: Yep, we can make that work once it needs them.

PR #421, opened 2019-03-14
tc-lib-iterate updates

djmitche:
This isn't everything I want to do with this library, but it's a start: simplification, better docs, remove some unused features, refactorings.

The big thing I want to finish up is the interface by which this is used in a process.  It has some weird behavior around throwing an unhandled error in the nextTick right now.  I think a better approach might be to provide an async `run` method that returns a promise when iteration is complete and rejects if there are too many failures.  But I'll do that in another PR.

### DIFF
--- a/libraries/iterate/src/iterate.js
+++ b/libraries/iterate/src/iterate.js
@@ -59,14 +59,8 @@ class Iterate extends events.EventEmitter {
     }
     this.monitor = opts.monitor;
 
-    // Count the iteration that we're on.
-    this.currentIteration = 0;
-
     // Decide whether iteration should continue
     this.keepGoing = false;
-
-    // Store the list of exceptions of the last few iterations
-    this.failures = [];
 
     // We want to be able to share state between iterations
     this.sharedState = {};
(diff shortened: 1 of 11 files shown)


libraries/iterate/src/iterate.js
```diff
@@ -139,7 +139,7 @@ class Iterate extends events.EventEmitter {
             }, this.maxIterationTime);
           }),
           watchdogRejector.promise(),
-          this.handler(watchdog, this.sharedState).then(() => {
+          Promise.resolve(this.handler(watchdog, this.sharedState)).then(() => {
```
imbstack: Huh.

libraries/iterate/src/iterate.js
```diff
@@ -76,103 +70,105 @@ class Iterate extends events.EventEmitter {
...
       }
+      const d = process.hrtime(start);
+      const duration = d[0] * 1000 + d[1] / 1000000;
 
-      // Wait until we've done all the checks to emit success
-      this.emit('iteration-success');
-      debug(`ran handler in ${diff} seconds`);
+      this.emit(iterError ? 'iteration-failure' : 'iteration-success');
       if (this.monitor) {
         this.monitor.info('iteration', {
```
imbstack: Should we make this a registered log type? Not blocking landing this.

libraries/iterate/src/watchdog.js
```diff
@@ -22,7 +22,7 @@ class WatchDog extends events.EventEmitter {
       } else {
         // This is being safe rather than sorry
         debug('exiting becase there is no expired event listener');
-        process.exit(1); // eslint-disable-line no-process-exit
+        //process.exit(1); // eslint-disable-line no-process-exit
```
imbstack: Should we just remove this line and the debug before it?

libraries/iterate/src/iterate.js
djmitche: Yeah, I think so.  I figured that would find itself in your dragnet..

libraries/iterate/src/watchdog.js
djmitche: That happened in a later commit.  I didn't mean to commit the outcommenting, but it happened anyway!  I intend to squash this on landing -- the split commits were just a way for me to double back when I accidentally broke tests.

PR #430, opened 2019-03-14
Retry 404 and higher errors

owlishDeveloper:
Bugzilla Bug: [1534294](https://bugzilla.mozilla.org/show_bug.cgi?id=1534294)

### DIFF
--- a/services/github/src/github-auth.js
+++ b/services/github/src/github-auth.js
@@ -12,10 +12,10 @@ const retryPlugin = (octokit, options) => {
       try {
         return await request(options);
       } catch (err) {
-        if (attempt === retries || err.name !== 'HttpError' || err.status !== 404) {
+        if (attempt === retries || err.name !== 'HttpError' || err.status < 500) {
           throw err;
         }
-        debug(`404 getting retried for eventual consistency. attempt: ${attempt}`);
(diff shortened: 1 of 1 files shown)


services/github/src/github-auth.js
```diff
@@ -12,10 +12,10 @@ const retryPlugin = (octokit, options) => {
       try {
         return await request(options);
       } catch (err) {
-        if (attempt === retries || err.name !== 'HttpError' || err.status !== 404) {
+        if (attempt === retries || err.name !== 'HttpError' || err.status < 404) {
```
djmitche: I think we should retry 404 and >=500.  Most 4xx's shouldn't be retried.  For example, 409 conflict is a permanent error.

owlishDeveloper: oh ok

services/github/src/github-auth.js
```diff
@@ -12,10 +12,10 @@ const retryPlugin = (octokit, options) => {
       try {
         return await request(options);
       } catch (err) {
-        if (attempt === retries || err.name !== 'HttpError' || err.status !== 404) {
+        if (attempt === retries || err.name !== 'HttpError' || err.status !== 404 || err.status < 500) {
```
imbstack: Need to remove the `err.status !== 404` now I think? Otherwise r+

Issue #285, opened 2019-02-22
[client-web] Bundle size increase in v9.0.0


edmorley:
**Describe the bug**
The minified bundle size of the `taskcluster-client-web` package was 92KB (18KB gzipped) in v8.1.1, but increased to 145KB (35KB) in v9.0.0:
https://bundlephobia.com/result?p=taskcluster-client-web@8.1.1
https://bundlephobia.com/result?p=taskcluster-client-web@9.0.0

This resulted in our webpack build exceeding the size threshold for the Renovate PR updating us to v9.0.0:
https://github.com/mozilla/treeherder/pull/4683
https://travis-ci.org/mozilla/treeherder/jobs/497201761#L539

**Expected behavior**
The package size is kept as small as is possible.

**Additional context**
Looking at the size breakdown chart at the bottom of the bundlephobia page, it seems that perhaps `crypto-js` is not listed as an external? Maybe doing so would help allow webpack to tree-shake the unused imports more easily?

Another thing that would help is if an ES6 module version was included in the package (listed under the `module` key in `package.json`), since webpack could then tree shake it?

helfi92:
> if ES6 module version was included

The ES6 version is the default export of the package. I don't know if adding a `module` property would help in this case.

> it seems that perhaps `crypto-js` is not listed as an external? Maybe doing so would help allow webpack to tree-shake the unused imports more easily?

I think that might be it :) We could certainly try.

edmorley:
It seems be to packaged as CJS? See the `module.exports` here:
https://unpkg.com/taskcluster-client-web@9.0.0/build/index.js

helfi92:
> It seems be to packaged as CJS

Apologies for my confusion. You're absolutely right. Is there a way to generate an es6 version? I don't seem to see  a `module` value for webpack's `libraryTarget` property. https://webpack.js.org/configuration/output/#outputlibrarytarget. Would having the `module` property point to [`src/index.js`](https://github.com/taskcluster/taskcluster/blob/b34aacc27c4e96565b598abf7c638e8fa3ce5530/clients/client-web/src/index.js) be good enough?

helfi92:
Added crypto-js to webpack's externals in `v9.1.0`.

sarah-clements:
Unfortunately, the bundle size is still high: 144.7kB (Minified 34.5 kB) https://bundlephobia.com/result?p=taskcluster-client-web@9.1.0

djmitche:
[...]

PR #435, opened 2019-03-15
Add esm output to client-web

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/285.

### DIFF
--- a/clients/client-web/.neutrinorc.js
+++ b/clients/client-web/.neutrinorc.js
@@ -1,45 +1,69 @@
-const { merge } = require('@neutrinojs/compile-loader');
+const name = 'taskcluster';
+const envs = ['@neutrinojs/env', ['NODE_ENV', 'TASKCLUSTER_ROOT_URL']];
+const lint = ['neutrino-preset-mozilla-frontend-infra/node-lint', {
+  rules: {
+    'no-underscore-dangle': 'off'
+  }
+}];
+
+const middleware = (neutrino) => {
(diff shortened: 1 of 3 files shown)


helfi92:
`yarn build` successfully generates both bundles however the process doesn't seem to exit...

helfi92:
@edmorley Any idea/thoughts about why the process hangs even though the bundles are successfully generated? https://github.com/taskcluster/taskcluster/blob/ae45db83fbcf5947d271b7674718dc8fd5932469/clients/client-web/.neutrinorc.js#L44-L69

edmorley:
`@neutrinojs/fork` is removed in Neutrino 9 -- I would update to Neutrino 9 first and then use the new approach mentioned here:
https://master.neutrinojs.org/migration-guide/#neutrino-v8-to-v9

edmorley:
Also Neutrino 8 uses webpack 3, which doesn't support tree shaking.

helfi92:
@edmorley What's stopping us from moving to neutrino v9 was https://github.com/neutrinojs/neutrino/issues/1251 which was recently resolved :) Any idea about when this can be published?

edmorley:
Ah I forgot that issue was from this package. A new release will be out soon (likely once merged), however the issue you link can be worked around on the caller's side too, via a manual `neutrino.config.externals.clear()` call (or similar) during testing only.

djmitche:
Does #445 supersede this?

helfi92:
Indeed.

PR #436, opened 2019-03-15
Simplify lib-config

imbstack:
* Remove `!env:flag` due to it being confusing and `!env:bool` is good enough
* Allow specifying which config files are allowed to be missing
* Stop catching and logging errors that should just be thrown
* Refactor to be easier to read

This is work in preparation for modifying lib-config to allow it to tell us what all of the `!env` vars are in order to automatically generate kubernetes configs.

Apologies for the one-giant-commit

### DIFF
--- a/infrastructure/builder/src/build/index.js
+++ b/infrastructure/builder/src/build/index.js
@@ -18,8 +18,8 @@ class Build {
   async run() {
     this.cfg = config({
       files: [
-        'build-config.yml',
-        'user-build-config.yml',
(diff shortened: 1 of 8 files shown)


djmitche:
> in order to automatically generate kubernetes configs.

:1st_place_medal:

djmitche:
This caused issues starting the tc-web queue service:
```
AssertionError [ERR_ASSERTION]: forceSSL must be defined
```

PR #445, opened 2019-03-15
client-web neutrino v9, esm output, share linting

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/285.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -89,9 +89,9 @@ tasks:
       - name: taskcluster-client-web
         image: 'taskcluster/browser-test:${node}'
         command: |
+            yarn lint clients/client-web &&
             cd clients/client-web &&
             yarn &&
-            yarn lint &&
             { Xvfb :99 -screen 0 640x480x8 -nolisten tcp & } &&
             sleep 2 &&
             CHROME_BIN=firefox DISPLAY=:99 yarn test
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -89,7 +89,6 @@ tasks:
       - name: taskcluster-client-web
         image: 'taskcluster/browser-test:${node}'
         command: |
-            yarn lint clients/client-web &&
             cd clients/client-web &&
             yarn &&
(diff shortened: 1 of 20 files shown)


clients/client-web/README.md
```diff
@@ -96,7 +96,6 @@ require(['taskcluster-client-web'], ({ Queue }) => {
 </script>
 ```
 
-Note that this module can be imported in a Node.js environment without error, but is not guaranteed to work.
```
djmitche: This was something that the treeherder team had requested in order to support server-side testing.  Is it no longer needed?

.taskcluster.yml
```diff
@@ -89,9 +89,9 @@ tasks:
       - name: taskcluster-client-web
         image: 'taskcluster/browser-test:${node}'
         command: |
+            yarn lint clients/client-web &&
```
djmitche: It looks like this is already included in the `lint` task by virtue of the changes you made in the top-level `package.json`.  In fact, the comand here just adds `clients/client-web` to the end of that command-line, which I think will just result in the linter scanning that directory twice.  So, this can probably be omitted..

clients/client-web/src/Client.js
```diff
@@ -4,31 +4,23 @@ import hawk from 'hawk';
...
-    credentialAgent: null,
-  };
-
-  static create(reference) {
-    return class extends Client {
-      constructor(options) {
-        super({ ...options, reference });
-      }
+  constructor(options = {}) {
+    this.defaults = {
```
djmitche: Could this just be a local variable?

clients/client-web/README.md
helfi92: Requiring this file in the node environment seem to throw an error about `window` not being defined.The `pretest` package.json script was failing.

djmitche: Oh, that's no fun.  @camd is that going to be a problem?  (ref https://bugzilla.mozilla.org/show_bug.cgi?id=1523376)

PR #451, opened 2019-03-15
Fix my wrong review

imbstack:
Sorry I don't know what I was thinking when I wrote the last review. If this makes sense to you it will be basically what you had before.

### DIFF
--- a/services/github/src/github-auth.js
+++ b/services/github/src/github-auth.js
@@ -12,11 +12,11 @@ const retryPlugin = (octokit, options) => {
       try {
         return await request(options);
       } catch (err) {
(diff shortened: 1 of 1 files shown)


services/github/src/github-auth.js
```diff
@@ -12,11 +12,11 @@ const retryPlugin = (octokit, options) => {
       try {
         return await request(options);
       } catch (err) {
-        if (attempt === retries || err.name !== 'HttpError' || err.status < 500) {
-          throw err;
+        if (attempt < retries && err.name === 'HttpError' && (err.status > 500 || err.status === 404)) {
```
owlishDeveloper: err.status >= 500 I think

Issue #440, opened 2019-03-15
[UI] make a nicer docs getting-started page


djmitche:
Instead of "plain" markdown with `hr` lines, this could be much more attractive and inviting.

The tutorial and the manual's first page could also probably stand some revision.

PR #463, opened 2019-03-18
[UI] Add custom get started view

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/440.

### DIFF
--- a/ui/docs/README.md
+++ b/ui/docs/README.md
@@ -1,39 +1,6 @@
 ---
 noAnchors: true
 ---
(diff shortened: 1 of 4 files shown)


helfi92:
> Nice! Any chance we could make the titles clickable, too?

I'm not sure it would make sense to have a button with a clickable title. I could perhaps add more emphasis on the "See More" button or make the whole bottom row clickable.

djmitche:
OK, I only asked that because that's the first place I clicked and nothing happened..

helfi92:
No worries. I’ll work something out.

Issue #450, opened 2019-03-15
[UI] support searching docs


djmitche:
The docs are spread over quite a few pages, and it's not always obvious where to find some key piece of information.

A search feature would help!

helfi92:
We could perhaps use [Docsearch](https://community.algolia.com/docsearch/) to provide the search feature. The normal way requires sending the documentation URL to the algolia team. In order to not have to reach out to the Algolia team for every spawned cluster, we could [run our own crawler](https://community.algolia.com/docsearch/run-your-own.html) instead. We could potentially use the [documentation-scrapper docker image](https://hub.docker.com/r/algolia/documentation-scrapper) to do the crawling and indexing.

djmitche:
We could potentially run that during `yarn build`, which runs lots of other stuff in docker images.  But from a glance at the "run your own crawler" link, it seems like it scans static pages via GET requests -- would it have any success trying to scan a react app?

It seems like something that indexes the `.md`/`.mdx` files (and maybe the references?) directly would be a better fit for our needs.

helfi92:
Setting [`start_urls`](https://community.algolia.com/docsearch/config-file.html#start_urls) to https://taskcluster-web.netlify.com/docs and [`js_render`](https://community.algolia.com/docsearch/config-file.html#js_render-optional) to `true` in the config file will perform the crawling in a Selenium browser. I believe we should be okay.

djmitche:
True, but that's not a great URL to crawl, as it's not necessarily running the version of the docs that we're building.  To use selenium to do the crawling, we'd need to run a full copy of tc-ui in another docker container at the same time and link the two.  That seems like a super-complicated way to index markdown files :)

helfi92:
> That seems like a super-complicated way to index markdown files :)

Many of the files are not markdown. They contain Javascript.

djmitche:
Give it a shot, then :)

helfi92:
Sure. Just wanted to see if there was consensus before doing this :)

PR #468, opened 2019-03-19
[WIP] Add doc search

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/450.

### DIFF
--- a/ui/src/components/DocSearch/index.jsx
+++ b/ui/src/components/DocSearch/index.jsx
@@ -1,0 +1,42 @@
+import React, { Component } from 'react';
+import { func } from 'prop-types';
+import docsearch from 'docsearch.js/dist/cdn/docsearch';
+import 'docsearch.js/dist/cdn/docsearch.css';
+import Search from '../Search';
+
+export default class DocSearch extends Component {
+  static propTypes = {
+    onSubmit: func.isRequired,
+  };
+
+  componentDidMount() {
+    docsearch({
+      apiKey: 'c7d22dbe731b27772278220687a87208',
+      // Used for identification when using Algolia's API
+      appId: 'FUFMXT38CF',
+      indexName: 'taskcluster',
+      inputSelector: '#algolia-doc-search',
+      handleSelected: this.handleSelected,
+    });
+  }
+
+  handleSelected = (input, event, suggestion) => {
+    const { onSubmit } = this.props;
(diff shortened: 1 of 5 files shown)


helfi92:
In order to run the crawler, I:

1. Created a file `doc-env` with content:

```bash
APPLICATION_ID=FUFMXT38CF
API_KEY=8ddc2f77936ff02ce46e8db68c6b14ee
```

2. Ran the docker image

```bash
brew install jq
docker run -it --env-file=doc-env -e "CONFIG=$(cat docsearch.json | jq -r tostring)" algolia/docsearch-scraper
```

ui/src/components/DocSearch/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+export default class DocSearch extends Component {
+  static propTypes = {
+    onSubmit: func.isRequired,
+  };
+
+  componentDidMount() {
+    docsearch({
+      apiKey: 'c7d22dbe731b27772278220687a87208',
+      // Used for identification when using Algolia's API
```
djmitche: Is it OK to make this information public?  If this is all generated locally, how is the API involved?  Could someone else use this key and ID to break our search?  Or possibly to inject malicious code into our users' browser sessions?

helfi92:
> Does that end up spidering the entire ui (for example if a doc links to /hooks, does it then spider every hook?)

[stop_urls](https://community.algolia.com/docsearch/config-file.html#stop_urls-optional) can be set to only scan and index links that fall under `/docs`.

> When we add a schema viewer that recursively expands $ref, will it expand those infinitely?

Any links not pertaining to the same domain will automatically not be scanned. `stop_urls` can also be used which accepts a regex.

>  Where does the index end up?

In Algolia servers

>  If it's on the algolia servers, then we'll need to have a disctinct index per Taskcluster build, and somehow reference that from the build (maybe a different API key per build?).

I belive the [`index_name`](https://community.algolia.com/docsearch/config-file.html#index_name) should be able to handle that.

> 10 minutes seems really long

Agreed. Note that the indexing can happen asynchronously in the background and doesn't necessarily need to block the `yarn build` command. Perhaps it can be run as a task?

djmitche:
From talking amongst the team just now, this takes too long to index and we'll look to take a simpler, less-fancy but faster approach.

PR #476, opened 2019-03-21
Bug 1535755 - simplify docs generation

djmitche:
Bugzilla Bug: [1535755](https://bugzilla.mozilla.org/show_bug.cgi?id=1535755)

This puts all docs under ui/docs, and simplifies a bunch of stuff:
 * all references are rendered based on data in `generated/references.json`
 * only use one copy of `references.json`
 * put the table of contents in `generated/` too

It also makes some user-visible changes:
 * `exchanges` instead of `events` (we used to use these interchangeably, about 50%/50%, and I think this is finally the last use of `events`)
 * no /references/ or /docs/ in the paths anymore, so paths largely match the table of contents

The super-long relative import paths were getting on my last nerve, so I added a resolver alias such that `import 'taskcluster-ui/foo/bar'` will resolve to `ui/src/foo/bar`.  It might make sense to try to clean up the `../../../` elsewhere in `ui/`, but not in this PR.

We're still using `writeDocs` from tc-lib-docs to export a tarball, but it doesn't contain any docs that we end up using.  All we really need is a JSON data structure -- that could speed up `yarn generate` a bit and avoid use of a temporary directory.  I'll work on that someday, unless someone beats me to it.


ui/docs/reference/core/hooks/exchanges.md
```diff
@@ -0,0 +1,9 @@
+---
+order: 2
+inline: true
+title: events
```
helfi92: The title should probably be `exchanges`.

djmitche: Good point!  I'm impressed you spotted that in a 27,000 line PR :)

Issue #469, opened 2019-03-20
New UI: Task status updates aren't seen as accurately as the old


jonalmeida:
**Describe the bug**

When comparing a new PR in https://github.com/mozilla-mobile/android-components which leads to may individual tasks in a group being created, I see the old using is updated far quicker than the new UI (which seems to poll for updates).

**To Reproduce**
1. Create a PR in android-components.
2. Observe the old and new UI side-by-side.
3. Observe when task statuses are updated in both UIs.

**Expected behavior**
The new UI should subscribe for task status updates instead of polling for them.

**Platform (please complete the following information):**
 - OS: Mac OS
 - Browser if applicable: Firefox
 - Version if applicable: 67.0a1 (2019-03-15) (64-bit)

helfi92:
Thanks for the issue. I do agree that it would make more sense to subscribe to a task instead of polling in this situation. Polling becomes a problem when you have to support task groups that takes a while to load (10k+ tasks) for something like a release job. I will experiment with switching to a subscription model and see if that improves the experience.

helfi92:
Resolved in https://github.com/taskcluster/taskcluster/commit/fd837dfde1cce4e813c054c41091f9c1d57bc519. Thank you for bringing this up.

Issue #248, opened 2019-02-14
New UI feels slower


glandium:
Loading a large task group feels slower than in the old UI, despite being actually faster, because it only displays the list once it has everything while the old UI would start displaying things before everything is downloaded.

Displaying one task, though, _is_ slower.

helfi92:
The new UI should now feel a lot faster. We no longer wait for the whole list to load before displaying it in the UI. Resolved in https://github.com/taskcluster/taskcluster/commit/fd837dfde1cce4e813c054c41091f9c1d57bc519.

PR #477, opened 2019-03-21
[UI] Task Group - use a subscription model instead of polling

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/469 and https://github.com/taskcluster/taskcluster/issues/248.


services/web-server/src/createContext.js
```diff
@@ -20,11 +20,18 @@ export default ({ pulseEngine, rootUrl }) => ({ req, connection }) => {
...
+    const currentClients = clients({ rootUrl });
+    const currentLoaders = loaders(
+      currentClients,
+      false,
+      rootUrl
+    );
     // if connection is set, this is for a subscription
     return {
       pulseEngine,
       // subscriptions do not need credentials (all public data)
```
djmitche: this comment should move up with the clients(..) call

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -254,6 +260,67 @@ export default class TaskGroup extends Component {
...
+          'tasksCompleted',
+          'tasksFailed',
+          'tasksException',
+        ],
+      },
+      updateQuery(previousResult, { subscriptionData }) {
+        const { tasksSubscriptions } = subscriptionData.data;
+        const newTask = tasksSubscriptions.status.task;
+        const isFromSameTaskGroupId =
+          newTask.status.taskGroupId === taskGroupId;
```
djmitche: This should always be true, right?  I think that if it's not, that's a signal that something's pretty broken (the browser is going to be consuming 5-10 messages/second!).  My GraphQL is still pretty weak, but I think the way this is designed ends up [back here](https://github.com/taskcluster/taskcluster/blob/6704c2694a011b35de304abab1f1161e3ba64d2c/services/web-server/src/resolvers/Tasks.js#L137-L138) where a proper routingKeyPattern is defined to select only tasks in this task group.

So, perhaps omit this or turn it into an assertion?

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -174,6 +175,7 @@ export default class TaskGroup extends Component {
 
     if (prevProps.match.params.taskGroupId !== taskGroupId) {
       updateTaskGroupIdHistory(taskGroupId);
+      this.subscribeToMore();
```
djmitche: There's a bit of a race condition here, since the "current" state is loaded, then the pulse listeners are established.  If the state of a task changes in that interval, the UI will never show it.  I can't really tell if this is a big deal or not, though -- it depends on the propagation delay of pulse messages and how much time elapses between fetching the base task state and beginning the subscription.

We have the advantage that applying pulse messages redundantly still yields a correct result.  Is it possible to switch the order, so that the subscription begins *before* the current state is fetched?

ui/.neutrinorc.js
```diff
@@ -31,6 +31,7 @@ module.exports = {
         'react/no-access-state-in-setstate': 'off',
         'babel/no-unused-expressions': 'off',
         'linebreak-style': 'off',
+        'no-underscore-dangle': 'off',
```
djmitche: Is this to allow the use of `__typename`?  Is it normal that a dataIdFromObject function has to use that property?  If so, perhaps it should just be disabled for that function?

ui/src/views/Tasks/TaskGroup/index.jsx
helfi92: This is to make sure data is not from another task group which can happen when a message is in flight and a user searches for a different task group. I'll add a comment.

djmitche: Ah, neat.  Thanks.

ui/src/views/Tasks/TaskGroup/index.jsx
helfi92: Good observation. I'll look into it.

helfi92: According to [react docs](https://reactjs.org/docs/react-component.html#mounting), the order in which the methods are called when mounting is:

1. constructor()
2. static getDerivedStateFromProps()
3. render()
4. componentDidMount()

I added a commit that moves the listener creation from `componentDidMount` to `getDerivedStateFromProps`. This should help in making it less prone to a race condition.

helfi92:
> What is the purpose of the dataIdFromObject function?

dataIdFromObject is used to properly cache GraphQL responses. GraphQL usually caches using the `__typename` property which causes subscriptions to not cache properly since their `__typename` is not `TaskStatus` but rather `TaskDefined`, `TaskFailed`, etc...In order to properly update the cache, we need to make sure an object of type`TaskStatus` uses the same cache key as `TaskCompleted`, `TaskFailed`, etc..

> Perhaps the dataId could be ${taskId}/${runId}?

Agreed.

djmitche:
> In order to properly update the cache, we need to make sure an object of typeTaskStatus uses the same cache key as TaskCompleted, TaskFailed, etc..

I'm curious why there are five different types for a task status, instead of just using one everywhere?

ui/src/App/index.jsx
```diff
@@ -40,7 +41,57 @@ export default class App extends Component {
...
+          return `${taskId}/${runId}` || defaultDataIdFromObject(object);
+        }
+
+        default: {
+          // fall back to default handling
+          return defaultDataIdFromObject(object);
+        }
+      }
+    },
+    /* eslint-enable no-underscore-dangle */
```
djmitche: I have questions :)

I think all of the `|| defaultDataIdFromObject(object)` will never get used, because the left side of `||` will never be false.  Also, this function begins by carefully "guessing" at where taskId and runId might be, in such a way that it will not throw an error for *any* object, even if it might get a bogus taskId or runId or more likely just undefined.  *Then* it consults `__typename` which could have indicated what sort of object it had to begin with.  And all of the branches of the switch do the same thing, so they could probably be coalesced.  So perhaps

```js
  switch (object.__typename) {
    case 'TaskDefined':
    case 'TaskPending':
    case 'TaskRunning':
    case 'TaskCompleted':
    case 'TaskFailed':
    case 'TaskException': {
      const taskId = ..; // wherever a taskId is in these messages
      const runId = ..; // similar
      return `${taskId}/${runId}`;
    }
    case 'TaskStatus': {
      const taskId = ..; // wherever a taskId is in this API response
      const runId = ..; // similar
      return `${taskId}/${runId}`;
    }
  }
```

I can't tell for sure from [the apollo-cache-inmemory docs](https://github.com/apollographql/apollo-client/tree/master/packages/apollo-cache-inmemory) but it looks like the dataId is scoped by the typename, so even if a TaskPending and a TaskStatus both resolve to the same `<taskId>/<runId>`, they won't be treated as the same object in the cache.  And this is probably for the best -- for run 0 there will be a TaskDefined, then TaskPending, then TaskRunning, then TaskResolved message.  All of which would be represented as a TaskStatus with `<taskId>/0` depending on when you fetched it.  So perhaps TaskStatus should not be cached at all (since it's not immutable)?  Or perhaps all of the events should be transmuted into TaskStatus objects?

helfi92: >  Or perhaps all of the events should be transmuted into TaskStatus objects?

I think this approach make things simpler. I'll see what I can do in web-server to have this fixed. Thanks for bringing that up.

helfi92:
The netlify preview won't work properly because there are changes to `web-server`. The netlify preview currently reads from the production instance of web-server.

helfi92:
@djmitche This one is ready for review :)

djmitche:
Haha, yeah, it just came up in my task list :)

djmitche:
I started testing this by running `yarn start` for both web-server and ui, and when I load https://dustin-ui.ngrok.io/tasks/groups/Zki7rsASTBSw-p3jCve2zA I get
```
Warning: [JSS] Could not find the referenced rule input in TaskGroup. browser.js:49
Unhandled GraphQL subscription error DOMException: "The operation is insecure." bundle.esm.js:279
Warning: Failed prop type: The prop `taskGroupConnection` is marked as required in `TaskGroupTable`, but its value is `undefined`.
```

am I doing something wrong?

ui/src/App/index.jsx
djmitche: Questions answered :)

helfi92:
@djmitche I see a lot of CORS failures when clicking on that link.

djmitche:
whoops, I didn't mean to make that a review :)

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -164,16 +169,98 @@ export default class TaskGroup extends Component {
...
+              );
+            })
+          : tasks.set(tasksSubscriptions.taskId) &&
+            previousResult.taskGroup.edges.concat({
+              // eslint-disable-next-line no-underscore-dangle
+              __typename: 'TasksEdge',
+              node: {
+                ...cloneDeep(tasksSubscriptions.task),
+              },
+            });
```
djmitche: I'm not going to lie - I cant' figure out what this statement (starting with `const edges = `) does.

I'm *guessing* it's either updating an entry that is in previousResult and tasks, or creating a new entry in previousResult.  But aside from the confusing syntax, I'm not sure why the state is stored in parallel in `tasks` and `previousResult`?  Could those get out of sync?

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -147,6 +151,7 @@ export default class TaskGroup extends Component {
     super(props);
 
     previousCursor = INITIAL_CURSOR;
+    tasks = new Map();
```
djmitche: Why are these module-level global variables, and not instance properties?

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -164,16 +169,98 @@ export default class TaskGroup extends Component {
...
+                return edge;
+              }
+
+              return dotProp.set(edge, 'node', node =>
+                dotProp.set(node, 'status', status =>
+                  dotProp.set(status, 'state', tasksSubscriptions.state)
+                )
+              );
+            })
+          : tasks.set(tasksSubscriptions.taskId) &&
```
djmitche: I think most readers don't know whether `:` or `&&` has higher precedence.  I don't.  This is probably a good time to use multiple statements instead of stringing together expressions with && in order to achieve sequential execution.  I won't tell the functional-programming police if you don't :)

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -164,16 +169,98 @@ export default class TaskGroup extends Component {
...
+          'tasksDefined',
+          'tasksPending',
+          'tasksRunning',
+          'tasksCompleted',
+          'tasksFailed',
+          'tasksException',
+        ],
+      },
+      updateQuery(previousResult, { subscriptionData }) {
+        const { tasksSubscriptions } = subscriptionData.data;
```
djmitche: I think `tasksSubscriptions` has a single task status in it?  The doubly-plural variable name suggests otherwise, so I'm not sure.

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -164,16 +169,98 @@ export default class TaskGroup extends Component {
     searchTerm: null,
   };
 
+  static unsubscribe() {
+    if (!TaskGroup.listener) {
+      return;
+    }
+
+    TaskGroup.listener.unsubscribe();
+    TaskGroup.listener = null;
```
djmitche: Maybe the same question as about the module-level globals, but why are these class-level properties, and static methods?

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -164,16 +169,98 @@ export default class TaskGroup extends Component {
...
+          ? previousResult.taskGroup.edges.map(edge => {
+              if (tasksSubscriptions.taskId !== edge.node.status.taskId) {
+                return edge;
+              }
+
+              return dotProp.set(edge, 'node', node =>
+                dotProp.set(node, 'status', status =>
+                  dotProp.set(status, 'state', tasksSubscriptions.state)
+                )
+              );
```
djmitche: That's six lines to accomplish `edge.node.status.state = tasksSubscriptions.state`.  Assignment operators are OK :)

djmitche: Also, is it just the state you want to update here, and not the other properties of the task status?

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -164,16 +169,98 @@ export default class TaskGroup extends Component {
...
+            })
+          : tasks.set(tasksSubscriptions.taskId) &&
+            previousResult.taskGroup.edges.concat({
+              // eslint-disable-next-line no-underscore-dangle
+              __typename: 'TasksEdge',
+              node: {
+                ...cloneDeep(tasksSubscriptions.task),
```
djmitche: Does this also need to create a `status` property?

ui/src/views/Tasks/TaskGroup/taskGroupSubscription.graphql
```diff
@@ -0,0 +1,26 @@
...
+    taskId
+    taskGroupId
+    task {
+      metadata {
+        name
+      }
+      taskGroupId
+      taskId
+      status {
+        taskId
```
djmitche: Are all three `taskIds` always the same?  Both `taskGroupIds`?

helfi92:
Does it work for you when you run it outside the cluster?

djmitche:
That's where I'm running it -- just on my development machine with two `yarn start` invocations.  Hmm.

helfi92:
> I started testing this by running yarn start for both web-server and ui, and when I load https://dustin-ui.ngrok.io/tasks/groups/Zki7rsASTBSw-p3jCve2zA I get

Shouldn't it be `localhost` instead of `dustin-ui.ngrok.io`?

helfi92:
What about another view like `/provisioners`. Does it work for you?

djmitche:
Since I don't develop on my laptop / desktop, but on an EC2 instance, I use ngrok to proxy to the running server.  This has worked fine for me in the past with tc-web-server / tc-ui.  But, I will experiment :)

helfi92:
I just cloned a fresh copy of the monorepo and pulled a copy of this branch. I was not able to replicate the issue. I'm using the following config:

```bash
# ui/.env
APPLICATION_NAME="Taskcluster"
PORT=5080
```

```bash
# services/web-server/user-config.yml
defaults:
  taskcluster:
    rootUrl: "https://taskcluster.net"
  pulse:
    namespace: taskcluster-user-haali
    username: taskcluster-user-haali
    password: <password>
    hostname: pulse.mozilla.org
    vhost: /

```

djmitche:
Ah,

```
Blocked loading mixed active content “http://localhost:5080/graphql”
```

I hadn't set the GRAPHQL_* env vars :(

djmitche:
OK, so I've got this running and I set some breakpoints on subscribe / unsubscribe.  They don't seem to be called on page load (and I don't see /subscribe accessed, nor anything updating in the UI).  If I click on one of the colored boxes at the top, then it calls `TaskGroup.subscribe` but even then, it returns `TaskGroup.listener` right away and no subscription takes place.

ui/src/views/Tasks/TaskGroup/index.jsx
djmitche: ah, I see now that this is `dot-prop-immutable` and is thus creating an entirely new data structure.

ui/src/views/Tasks/TaskGroup/taskGroupSubscription.graphql
djmitche: Yup :)

ui/src/views/Tasks/TaskGroup/index.jsx
helfi92: It's used inside static methods which don't have access to the `this` instance. What I can do is make `tasks` be static so it can be referred to as `TaskGroup.tasks`.

ui/src/views/Tasks/TaskGroup/index.jsx
helfi92: This is a similar reasoning as https://github.com/taskcluster/taskcluster-web-server/pull/49#issuecomment-429380389. We are subscribing to tasks, not the return value.

ui/src/views/Tasks/TaskGroup/index.jsx
helfi92: I'll see if there is a way to remove the subscription from happening in `static getDerivedStateFromProps`.

ui/src/views/Tasks/TaskGroup/index.jsx
helfi92: It already has one.

Screen Shot 2019-04-12 at 2 43 40 PM

ui/src/views/Tasks/TaskGroup/index.jsx
helfi92: Done.

PR #478, opened 2019-03-22
[Bug 1537907] Elide secret values

imbstack:
This should be one layer in helping protect us from accidental leaks.

Bugzilla Bug: [1537907](https://bugzilla.mozilla.org/show_bug.cgi?id=1537907)

### DIFF
--- a/libraries/monitor/src/logger.js
+++ b/libraries/monitor/src/logger.js
@@ -36,6 +36,37 @@ const LEVELS_REVERSE_COLOR = [
   chalk.magenta('DEBUG'),
 ];
 
+const ELIDED = new Set([
+  'credentials',
+  'accessToken',
+  'password',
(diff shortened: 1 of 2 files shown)


libraries/monitor/src/logger.js
```diff
@@ -36,6 +36,35 @@ const LEVELS_REVERSE_COLOR = [
   chalk.magenta('DEBUG'),
 ];
 
+const ELIDED = new Set([
+  'credentials',
+  'accessToken',
+  'password',
+  'secretAccessKey',
```
djmitche: how about `secret` (since that's the name of the content of a secret in the secret service)

imbstack: Anybody else who reviews this, if you have other ideas, please let me know! I think we can safely be relatively strict here.

PR #482, opened 2019-03-22
Use allcontributors in docs

djmitche:
With some other fixes:
 * drop the nonfunctional badge
 * add a "staff" contribution type and gather staff at the beginning


.all-contributorsrc
```diff
@@ -4,18 +4,25 @@
...
   "files": [
-    "README.md"
+    "README.md",
+    "ui/docs/people.md"
   ],
   "imageSize": 100,
   "commit": false,
+  "types": {
+    "staff": {
+      "symbol": "👷",
```
imbstack: How about :wrench:

.all-contributorsrc
```diff
@@ -4,18 +4,25 @@
...
+      "description": "Current or former Taskcluster staff member"
```
imbstack: I'm sorta less interested in having former staff members on here. Part of the point to me is a list of people who can be asked questions, etc. Maybe we need an alumni type as well? This is a tiny nit though, either way is fine.

ui/src/views/Documentation/index.jsx
```diff
@@ -120,6 +120,7 @@ export default class Documentation extends Component {
 
       this.setState({ Page, pageInfo, error: null });
     } catch (error) {
+      console.log(Object.assign({}, error));
```
imbstack: Left over debug?

.all-contributorsrc
djmitche: Good point.. what's the emoji for that?

ui/src/views/Documentation/index.jsx
djmitche: ack, yes.  I'm working on another PR to make debugging docs easier (since right now any error is treated as "not found")

djmitche:
Other opinions on the right emoji?  @owlishDeveloper ? :)

djmitche:
Oh, good point!  I assume both are due to underlying links.  I'll see if I can figure out how to apply styles to specific markdown files :)

djmitche:
Hm, it turns out that the mdx `components={a}` is not catching the literal `a` tags that allcontributors creates, so they open in the same tab/window.  That seems like a bit of a deep bug to pursue in this PR, but is it something we should worry about at all?

djmitche:
Is https://github.com/jxnblk/mdx-style the right way to accomplish this?  I can't figure out any other approach..

ui/docs/people.md
```diff
@@ -0,0 +1,13 @@
...
...g src="https://avatars2.githubusercontent.com/u/12398942?v=4" width="100px;" alt="Hashini Galappaththi"/><br /><sub><b>Hashini Galappaththi</b></sub></a><br /><a href="https://github.com/taskcluster/taskcluster/commits?author=hashi93" title="Code">💻</a></td><td align="center"><a href="https://github.com/fiennyangeln"><img src="https://avatars1.githubusercontent.com/u/24544912?v=4" width="100px;" alt="Fienny Angelina"/><br /><sub><b>Fienny Angelina</b></sub></a><br /><a href="https://github.com/taskcluster/taskcluster/commits?author=fiennyangeln" title="Code">💻</a></td><td align="center"><a href="https://github.com/kanikasaini"><img src="https://avatars2.githubusercontent.com/u/20171105?v=4" width="100px;" alt="Kanika Saini"/><br /><sub><b>Kanika Saini</b></sub></a><br /><a href="https://github.com/taskcluster/taskcluster/commits?author=kanikasaini" title="Code">💻</a></td><td align="center"><a href="https://github.com/Biboswan"><img src="https://avatars2.githubusercontent.com/u/22202556?v=4" width="100px;" alt="Biboswan Roy"/><br /><sub><b>Biboswan Roy</b></sub></a><br /><a href="https://github.com/taskcluster/taskcluster/commits?author=Biboswan" title="Code">💻</a></td></tr></table>
```
helfi92: html tags are not picked up by mdx. Are we able to switch `<a />` to `[]()`. I think this would fix our issue.

djmitche: Not if we want to use allcontributors, no.  We could easily do so with `yarn generate` but then we lose the benefit of the bot.  (We could still use the bot, but would have to run `yarn generate` on every PR from the bot, which pretty much defeats the purpose).

Will mdx-style not find the `<a>` elements, either?  If so, the we may be on our own and not able to use allcontributors..

djmitche: (I can try it and find out, just wondering if you were suggesting it wouldn't work)

helfi92: > Is https://github.com/jxnblk/mdx-style the right way to accomplish this? I can't figure out any other approach..

It would probably work. I would probably just add a `<style />` tag https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style instead.

djmitche: It doesn't appear that's valid JSX (or maybe MDX or MDHASH or HASH.. not sure):
```
SyntaxError: /home/dustin/p/taskcluster/ui/docs/people.md: Unexpected token, expected "}" (25:17)

  23 | <style>
  24 | a {
> 25 |   text-decoration: none;
     |                  ^
  26 | }
  27 | </style>
```

djmitche: And using dangerouslySetInnerHTML causs it to get stuck on the comment:

```
  25 |   text-decoration: none;
  26 | }`}} />
> 27 | <!-- ALL-CONTRIBUTORS-LIST:START - Do not remove or modify this section -->
     |  ^
  28 | <!-- prettier-ignore -->
```

djmitche:
The emoji are less critical now since they are only in the README and not on the docs page, but I'm still open to suggestions :)

ui/src/views/Documentation/People/index.jsx
```diff
@@ -0,0 +1,68 @@
+import React, { Component } from 'react';
+import { withStyles } from '@material-ui/core/styles';
+import { func } from 'prop-types';
+import { withRouter } from 'react-router-dom';
+import { contributors } from '../../../../../.all-contributorsrc';
```
helfi92: I think we can remove `json-loader` and instead do:

```jsx
import { contributors } from '../../../../../.all-contributorsrc.json';
```

Could you give it a try to see if it works, please?

ui/src/utils/constants.js
```diff
@@ -194,6 +195,12 @@ export const DOCS_MENU_ITEMS = [
     hasChildren: true,
     icon: BookOpenOutlineIcon,
   },
+  {
+    label: 'People',
+    path: join(DOCS_PATH_PREFIX, 'people'),
+    hasChildren: false,
+    icon: AccountHeartOutlineIcon,
+  },
```
helfi92: https://taskcluster-web.netlify.com/docs has a link to the people's page. Is it possible to not have it part of the sidebar?

ui/src/views/Documentation/People/index.jsx
djmitche: No love:
```
Module not found: Error: Can't resolve '../../../../../.all-contributorsrc.json' in '/home/dustin/p/taskcluster/ui/src/views/Documentation/People'
```

ui/src/utils/constants.js
djmitche: I kind of want to highlight it a little more -- I think this is more important than "resources" for example.  Could we remove it from the sidebar but make it one of the four buttons?

ui/src/views/Documentation/People/index.jsx
helfi92: I was able to make it work via:

```diff
diff --git a/ui/src/views/Documentation/People/index.jsx b/ui/src/views/Documentation/People/index.jsx
index ddbfd9751..335a8e831 100644
--- a/ui/src/views/Documentation/People/index.jsx
+++ b/ui/src/views/Documentation/People/index.jsx
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import { withStyles } from '@material-ui/core/styles';
 import { func } from 'prop-types';
 import { withRouter } from 'react-router-dom';
-import { contributors } from '../../../../../.all-contributorsrc';
+import contributorsRc from '../../../../../.all-contributorsrc.json';
 
 @withRouter
 @withStyles(() => ({
@@ -43,6 +43,7 @@ export default class People extends Component {
 
   render() {
     const { classes, filter } = this.props;
+    const { contributors } = contributorsRc;
 
     return (
       <div className={classes.peopleList}>
```

djmitche: Did you also remove the config in `.neutrinorc`?  I did almost the same thing except that I kept the destructuring assignment.  Does that not work in this specific case for some reason?

helfi92: > Did you also remove the config in .neutrinorc? 

Yes

>  Does that not work in this specific case for some reason?

Seems like it doesn't. There's probably an explanation for that. I'm not sure tbh.

helfi92:
I added a commit to make it look like

Screen Shot 2019-03-25 at 5 52 14 PM

Let me know if you disagree with any of the changes.

djmitche:
If we rename the file, then all-contributors won't work anymore, as it won't find the file..

The UX changes look good, though.

ui/src/views/Documentation/People/index.jsx
djmitche: I bet the explanation is #495.

helfi92:
@djmitche If you're happy with the state of this PR, feel free to merge :)

PR #487, opened 2019-03-22
Bug 1533937 - cleanup after tests (azure tables, at least) on push

djmitche:
This just cleans up tables, but could easily be extended to any other kind of resource, e.g., SQS queues, Azure containers.

I guess ideally this would be a crontask, but in the absence of good support for those, I think running this just on pushes is a decent compromise.

Bugzilla Bug: [1533937](https://bugzilla.mozilla.org/show_bug.cgi?id=1533937)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -62,6 +62,11 @@ tasks:
         command: yarn lint
       - name: test:meta
         command: yarn test:meta
+        # only do cleanup on pushes; no sense doing so on pull requests
+      - $if: 'tasks_for == "github-push" && event["ref"] == "refs/heads/master"'
+        then:
+          name: test:cleanup
+          command: yarn test:cleanup
       - name: taskcluster-terraform
         image: 'jonatanblue/gitlab-ci-terraform:latest'
         command: 'cd infrastructure/terraform; terraform fmt -check'
--- a/services/auth/test/cleanup.js
+++ b/services/auth/test/cleanup.js
@@ -1,0 +1,59 @@
+const helper = require('./helper');
+const azure = require('fast-azure-storage');
+
+suite('Test Cleanup', function() {
+  suiteSetup(async function() {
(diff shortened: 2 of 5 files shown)


.taskcluster.yml
```diff
@@ -62,6 +62,10 @@ tasks:
         command: yarn lint
       - name: test:meta
         command: yarn test:meta
+      - $if: 'tasks_for == "github-push"'  # only do cleanup on pushes; no sense doing so on pull requests
```
imbstack: The only thing to note here is that when we push to this repo and make a PR, the actual tasks that are triggered are for the push. The only time PR action is triggered is if there's a PR from another repo. Maybe make this so that it is only on pushes to master or something?

services/auth/test/cleanup.js
```diff
@@ -0,0 +1,51 @@
...
+          continue;
+        }
+
+        console.log(`deleting ${name}`);
+        try {
+          await table.deleteTable(name);
+        } catch (err) {
+          // this is likely because there are parallel runs of this task, so ignore
+          // but log the error..
+          console.log(`..failed: ${err}`);
```
imbstack: I see this failing for months on end for other reasons and we won't notice. Maybe there's some other alert we could have or something?

imbstack: Not blocking landing this, just a nit.

imbstack:
Having this makes me feel a lot better

.taskcluster.yml
djmitche: Oh my, that's pretty clever -- I didn't see that.  Yeah, I'll update to only trigger on pushes to master.

services/auth/test/cleanup.js
djmitche: Good point.  I'll fix it to only ignore 404's.

Issue #480, opened 2019-03-22
creating a client does not show an accessToken


djmitche:
**Describe the bug**
When creating a new client, the UI does not display the client's accessToken

**To Reproduce**
1. Go to Authentication -> Clients
2. enter a clientId, click +
3. See the new client

**Expected behavior**
Should be shown the accessToken as a one-time display.

djmitche:
Note that the "reset accessToken" button does show the access token.

PR #489, opened 2019-03-23
[UI] Show access token when creating a client

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/480.

### DIFF
--- a/ui/src/views/Clients/ViewClient/index.jsx
+++ b/ui/src/views/Clients/ViewClient/index.jsx
@@ -62,6 +62,7 @@ import { THEME } from '../../../utils/constants';
   },
   panelCard: {
     background: theme.palette.warning.dark,
+    marginBottom: theme.spacing.unit,
   },
   clearIcon: {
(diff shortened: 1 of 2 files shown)


ui/src/views/Clients/ViewClient/index.jsx
```diff
@@ -165,35 +163,46 @@ export default class ViewClient extends Component {
...
   render() {
-    const { error, loading, accessToken, accessTokenPanelOpen } = this.state;
-    const { isNewClient, data, classes } = this.props;
+    const { error, loading, accessToken } = this.state;
+    const { isNewClient, data, classes, location } = this.props;
+
+    if (location.state && location.state.accessToken) {
+      const state = { ...location.state };
+
+      delete state.accessToken;
```
djmitche: Isn't `location` supposed to be immutable?

helfi92: It is, hence the `{ ...location.state }`.

PR #494, opened 2019-03-25
Typo fix

petemoore:
Thanks


djmitche:
@petemoore you'll need to run `yarn generate` and amend the commit.  Otherwise looks good :)

PR #500, opened 2019-03-26
Bug 1535755 - put user data in docs, with dev data in READMEs

djmitche:
I didn't try to write too much new documentation, but at least the reference section doesn't talk about how to run tests or how it's deployed in heroku anymore..

Bugzilla Bug: [1535755](https://bugzilla.mozilla.org/show_bug.cgi?id=1535755)


ui/docs/reference/core/index/README.md
```diff
@@ -1,56 +1,83 @@
...
-```
+## Index Hierarchy
+
+Tasks are indexed in a dot (`.`) separated hierarchy called a namespace.
+For example, a task could be indexed with the index path `some-app.<revision>.linux-64.release-build`.
+In this case the following namespaces are created.
+
+ 1. `some-app`,
+ 1. `some-app.<revision>`, and,
+ 1. `some-app.<revision>.linux-64`
```
helfi92: Wow, I didn't know multiple `1.` becomes `1. 2. 3.`.

ui/docs/reference/integrations/login/README.md
```diff
@@ -1,43 +1,8 @@
...
- * Access to the Auth0 management API with the `read:users` scope
-
-### Instances
-
-`taskcluster-login` exists on two instances, stage for pre-deployment validation, and production for actual use.
-
-## Service Owner
-
-Service Owner: dustin@mozilla.com
+See [getting user credentials](getting-user-creds) for more detail about interacting with this service.
```
helfi92: Link is not working for me.

ui/docs/reference/core/index/README.md
djmitche: Markdown is amazing!

ui/docs/reference/integrations/login/README.md
djmitche: Hm, I think none of these relative links are going to work..

helfi92: You can probably fix it in https://github.com/taskcluster/taskcluster/blob/master/ui/src/views/Documentation/components/Anchor/index.jsx#L30 by making the url relative.

djmitche: I think the problem is that the README is rendered at `/docs/reference/integrations/login`, so a relative `getting-user-creds` leads to `/docs/reference/integrations/getting-user-creds` instead of `/docs/reference/integrations/login/getting-user-creds` (which is the intent).

djmitche: Fix pushed..

PR #502, opened 2019-03-28
Bug 1538961 - improve subscriptions support

djmitche:
This ends up being a big refactor, but the goal is to improve subscriptions support.
 * Allow communicating errors that occur in web-server to subscribers
 * Handle errors binding queues better (in a way that doesn't kill connections) and communicate those errors to subscribers
 * Make PulseIterator indicate when a message is delivered or abandoned, and use that to ack/nack messages in pulse
 * Use Queue TTLs to automatically delete orphaned queues

Bugzilla Bug: [1538961](https://bugzilla.mozilla.org/show_bug.cgi?id=1538961)


services/web-server/test/pulseengine_test.js
```diff
@@ -0,0 +1,153 @@
...
+      assert.deepEqual(await pi.next(), {value: 'M1', done: false});
+      assert.deepEqual(sent, ['M1']);
+
+      // an unsent M2 will error out after onError is called
+      engine.onMessage('M2')
+        .then(() => { throw new Error('unexpected success'); })
+        .catch(err => sent.push(err.toString()));
+      assert.deepEqual(sent, ['M1']);
+
+      engine.onError(new Error('uhoh'));
```
helfi92: Should this listener be set up before L117?

services/web-server/src/PulseEngine/Subscription.js
```diff
@@ -0,0 +1,162 @@
...
+          onMessage(message).then(
+            () => channel.ack(amqpMsg),
+            () => channel.nack(amqpMsg));
+        });
+
+        this.consumerTag = consumerTag;
+        this.listening = true;
+      }
+    } catch (err) {
+      debug(`Reconciling subscription ${subscriptionId}: ${err}`);
```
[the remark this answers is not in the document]
djmitche: Oops :)

services/web-server/test/pulseengine_test.js
djmitche: It's not a listener -- it's a callback.  The name is misleading.  I'll change `onError` and `onMessage` to be `handleXxx` isntead.

djmitche:
By the way, when you're testing (with development credentials -- hip-macaw), you can watch the rabbitmq admin UI to see what the backend is doing, and if it's leaving any queues around.  The hip-macaw admin credentials are in passwordstore:
```
├── cloudamqp.com
│   └── hip-macaw
```
and can be used to create a test user via the web UI.  Or if you'd prefer you can just reset the password on the `bug1538961` user.

services/web-server/src/PulseEngine/Subscription.js
```diff
@@ -0,0 +1,162 @@
...
+        const { handleError, handleMessage, subscriptions } = this;
+
+        // declare the queue, with autoDelete and exclusive both set to false so that
+        // the queue will stick around if we need to reconnect, but with a short TTL
+        // so that if we "forget" about the queue, it will be deleted quickly.
+        await channel.assertQueue(queueName, {
+          exclusive: false,
+          durable: true,
+          autoDelete: false,
+          expires: 30000, // 30 seconds, long enough for a reconnect
```
owlishDeveloper: I wonder if this is gonna yield the same error as I got when working on the TTL bug 🤔

services/web-server/src/PulseEngine/PulseIterator.js
```diff
@@ -38,40 +41,65 @@ export default class PulseIterator {
...
+    if (!this.listening) {
+      return Promise.reject(new Error('iterator cancelled'));
     }
+    return new Promise((resolve, reject) => {
+      this.pushQueue = this.pushQueue.push({resolve, reject, value, done: false});
+      this.match();
+    });
   }
 
+  // pull a value from the iterator; the resulting promise resolves when a value
```
owlishDeveloper: (is `iterall` more convenient than native async iterator?)

services/web-server/src/PulseEngine/Subscription.js
djmitche: I hope not -- these queues should not already exist, so the consistency error shouldn't occur.

services/web-server/src/PulseEngine/PulseIterator.js
djmitche: I'll read about `iterall` and see.. thanks!

services/web-server/src/PulseEngine/PulseIterator.js
```diff
@@ -14,22 +14,25 @@ export default class PulseIterator {
     this.listening = true;
     this.subscriptionId = this.pulseEngine.subscribe(
       subscriptions,
-      this.pushValue.bind(this)
+      this.pushValue.bind(this),
+      this.pushError.bind(this),
     );
+    this.pullId = 0;
```
owlishDeveloper: what is this var for?

services/web-server/src/PulseEngine/PulseIterator.js
```diff
@@ -14,22 +14,25 @@ export default class PulseIterator {
...
+    return this.pullValue();
   }
 
   return() {
-    this.emptyQueue();
+    this.stop({done: true});
 
-    return Promise.resolve({ value: undefined, done: true });
+    // return should always succeed, even if an error was introduced
+    return Promise.resolve({done: true});
```
owlishDeveloper: can you tell me more about the change `{ value: undefined, done: true }` ==> `{done: true}`?

services/web-server/src/PulseEngine/PulseIterator.js
djmitche: Debugging leftover!  I'll delete..

services/web-server/src/PulseEngine/PulseIterator.js
owlishDeveloper: (I think on the other end a "tuple" is expected, hence my question)

djmitche: Just saving a bit of code, really - the two should be equivalent.  I think I changed this a few times in hacking on it.  The [iterator protocol docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols) say that the value can be omitted if `done` is `true`.

services/web-server/src/PulseEngine/PulseIterator.js
```diff
@@ -38,40 +44,65 @@ export default class PulseIterator {
...
+  }
+
+  // insert a value into the queue; the resulting promise resolves when the value
+  // is consumed, and rejects if the iterator finishes without consuming the value
   pushValue(value) {
-    if (this.pullQueue.size !== 0) {
-      this.pullQueue.first()({ value, done: false });
-      this.pullQueue = this.pullQueue.shift();
-    } else {
-      this.pushQueue = this.pushQueue.push(value);
```
owlishDeveloper: Interesting, in this case `next()` would not return a promise and would not indicate whether we're done or not. Was it causing any troubles, just out of curiosity?

services/web-server/src/PulseEngine/PulseIterator.js
```diff
@@ -38,40 +44,65 @@ export default class PulseIterator {
...
+
+      this.pullQueue = this.pullQueue.shift();
+      if (!push.done) {
+        this.pushQueue = this.pushQueue.shift();
+      }
+
+      if (push.error) {
+        pull.reject(push.error);
+      } else {
+        pull.resolve({value: push.value, done: push.done});
```
owlishDeveloper: Oh, ok, so it seems we now always return `Promise` of `{value: blahblah, done: blahblah}`

helfi92:
Update: I plan on finishing my review on Monday

services/web-server/src/PulseEngine/PulseIterator.js
```diff
@@ -14,22 +18,24 @@ export default class PulseIterator {
...
       subscriptions,
-      this.pushValue.bind(this)
+      this.pushValue.bind(this),
+      this.pushError.bind(this),
     );
   }
 
   next() {
-    return this.listening ? this.pullValue() : this.return();
+    return this.pullValue();
```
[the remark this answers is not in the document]
djmitche: Yes, `next` must return a value, so it has to pull one.

services/web-server/src/PulseEngine/PulseIterator.js
djmitche: This is the case where pullQueue is empty, so there are no pending `next` calls waiting to return anything.

services/web-server/src/PulseEngine/Subscription.js
```diff
@@ -0,0 +1,162 @@
+import Debug from 'debug';
+
+const debug = Debug('Subscription');
+
+export default class Subscription {
+  constructor({ subscriptionId, handleMessage, handleError, subscriptions }) {
+    this.subscriptionId = subscriptionId;
```
helfi92: The constructor is missing `this.consumerTag`. Also, it'd be great if you could add a comment explaining `consumerTag`.

services/web-server/src/PulseEngine/PulseIterator.js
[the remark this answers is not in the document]
owlishDeveloper: I meant `next` can return the value directly, without having to call another function. Like, the code that's in `pullValue` can be just moved into `next`

services/web-server/src/PulseEngine/PulseIterator.js
owlishDeveloper: why are there two "queues"? I don't understand this whole design

services/web-server/src/PulseEngine/PulseIterator.js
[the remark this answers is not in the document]
djmitche: Ah, makes sense, thanks.

services/web-server/src/PulseEngine/PulseIterator.js
djmitche: The class matches up calls to `next()` and calls to `pushValue()`, either of which can happen at any time.  It does so by keeping a queue of each (`pullQueue` and `pushQueue`, respectively) and matching off items from those queues until one is empty (in the `match` method).

PR #503, opened 2019-03-28
Bug 1538152 - "tell a story" about tasks in queue logging

djmitche:
This should allow us to search for `jsonPayload.Fields="<taskId>"` and find all of the interesting bits about a task.

Bugzilla Bug: [1538152](https://bugzilla.mozilla.org/show_bug.cgi?id=1538152)

### DIFF
--- a/libraries/monitor/README.md
+++ b/libraries/monitor/README.md
@@ -25,7 +25,7 @@ const manager = new MonitorManager({
   serviceName: 'foo',
 });
 manager.setup({
-  mock: cfg.monitor.mock,  // false in production, true in testing
+  mock: cfg.monitor.mock,      // false in production, true in testing
   processName: 'server',       // or otherwise for e.g., periodic tasks
 });
 const monitor = manager.monitor(); // To get a root monitor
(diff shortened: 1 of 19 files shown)


djmitche:
This also would let us search for workerGroup/workerId to see the claiming / reclaiming activity.  I thought about adding logging for declareWorker, too, but that's not used and will probably go away with worker-manager, so it's not worth the effort.

djmitche:
OK, one last force-push and I'm hoping I'll get all green checkmarks.

libraries/monitor/README.md
```diff
@@ -339,3 +339,16 @@ This function will:
  * shut down and flush monitoring
  * call `process.exit` on success or failure
 Note, then, that the function **does not return**.
+
+### Testing
+
+If the `mock` option is true, then all messges logged are available from the `messages` property on the MonitorManager, in the format `{Type, Fields, Logger}`.
+These messages can be cleared with the `reset()` method -- generally done in a Mocha `setup` function.
+This allows tests to assert that a message was logged correctly:
```
imbstack: :+1: good idea. I wasn't able to come up with a nice way to do this.

djmitche: Well, you did really -- it just wasn't documented and I implemented it in a worse way, then discovered your implementation and documented it :)

PR #505, opened 2019-03-29
Dev docs refactor

djmitche:
This involves some light rewriting of the best practices, where they were outdated. Hopefully nothing here is very controversial, but it's not my intent to sneak in any new "rules" so please let me know if I've done something someone might disagree with.


dev-docs/README.md
```diff
@@ -0,0 +1,19 @@
+# Development Documentation
+
+This directory contains general information useful for people working on the Taskcluster codebase.
+It should be used in concert with the READMEs for the services and libraries, and of course examination of the code itself.
+
+The [Taskcluster monorepo](https://github.com/taskcluster/taskcluster) on Github contains the source code for all of the microservices, a collection of supporting libraries, and more.
+
+## RFCs
```
imbstack: Should we also include a link to the retrospectives repo? I see them as a big part of the story of tc.

dev-docs/best-practices/microservices.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Independence
+
+Although they ar developed in the same repository, each service must operate as an independent entity, using only the public interfaces of other services.
+Package imports between services are strictly forbidden.
+Where this seems like a good idea, prefer to create a new [library](libraries) instead.
+
+## Package Mechanics
+
+A service has a `serviceName` and a `projectName`, the latter often having a `taskclsuter-` prefix.
```
imbstack: We should have a meta-test that catches tskclsuter as well.

dev-docs/best-practices/microservices.md
```diff
@@ -0,0 +1,111 @@
...
+        monitor,
+      });
+    },
+  },
+}
+```
+
+#### Error Handling
+
+Do not use `res.status(..)` to return error messages.
```
imbstack: Should we disable using res.status programmatically? Either could be some sort of lint or overriding it somehow.

imbstack: It seems we only use `res.status` in 6 service files at the moment.

dev-docs/best-practices/microservices.md
djmitche: I have myriad ways of misspelling taskcluster :)

dev-docs/best-practices/microservices.md
djmitche: There are legitimate uses (like returning a response code without a body).  It's just *error* responses that should not..

PR #514, opened 2019-03-29
Procs yml support

imbstack:
This switches all process configuration to a `procs.yml` for each service that can leave out the `cd services/...` bit. We later autogenerate that into the `Procfile` that we make for each service. This also simplifies image generation.

Later this should allow us to generate kubernetes config stuff easier!


imbstack:
Now that I think about it, we might be able to do this one step nicer by autogenerating this all from `main.js`. Will look into that later.

imbstack:
No Procfile actually changed beyond newlines and comments, so this should be a noop to deploy. I'll try one service out to make sure.

PR #523, opened 2019-04-01
Bug 1540699 - fix calls to monitor.log.task*

djmitche:
And add tests for them.

Bugzilla Bug: [1540699](https://bugzilla.mozilla.org/show_bug.cgi?id=1540699)


djmitche:
I think the tests for the deadline resolver passed anyway because they saw the effects they expected, and the deadline resolver just logged the error and carried merrily on.

There were no tests for the claim resolver.

PR #526, opened 2019-04-01
Bug 1534713 - better debugging for pulse issues

djmitche:
This doesn't actually fix the bug, but will hopefully help get some better error messages that will help track down the underlying issues.

Bugzilla Bug: [1534713](https://bugzilla.mozilla.org/show_bug.cgi?id=1534713)


libraries/monitor/src/index.js
```diff
@@ -211,8 +211,12 @@ class MonitorManager {
   }
 
   _unhandledRejectionHandler(reason, p) {
-    const err = 'Unhandled Rejection at: Promise ' + p + ' reason: ' + reason;
-    this.rootMonitor.reportError(err);
+    if (reason instanceof Error) {
```
imbstack: What if instead of this we just did

```js
_unhandledRejectionHandler(reason, p) {
  this.rootMonitor.reportError(reason);
  //... stuff about bailOnUnhandled here.
}
```

My reasoning is that `reportError` handles just about anything you throw into it including "error shaped objects" that aren't really `instanceof Error`. Also even [though the node docs say](https://nodejs.org/api/process.html#process_event_unhandledrejection) you can log `p`, this has always been a useless thing to write out afaict.

djmitche: yeah, sounds good.

PR #527, opened 2019-04-01
Bug 1535316 - remove use of http-proxy for testing auth

djmitche:
This works around a bug in http-proxy by .. removing it.

Brian, I'm not sure you'll like the `apis` loader component.  Another approach I tried was to attach the `testServerApi` to the express app after `await exports.load('server')`, but starting the server adds a blanket 404 express handler, which means that any other express handlers added after that never get invoked.  I couldn't see a good way around that.  I'm open to ideas :)

Bugzilla Bug: [1535316](https://bugzilla.mozilla.org/show_bug.cgi?id=1535316)


djmitche:
Would it make more sense to add the `apis` component to every service?

djmitche:
Oh, hey, client uses this too.  Hmm.

djmitche:
Haha, and it has a workaround for this very bug:

https://github.com/taskcluster/taskcluster/blob/5c5050fe3c4c22b67bf0a1d808a17c8086d352c9/clients/client/test/retry_test.js#L152-L154

djmitche:
..and it turns out that's totally unnecessary

PR #528, opened 2019-04-02
Add GCP provider object. Implement listWorkers method

owlishDeveloper:
Bugzilla Bug: [1487699](https://bugzilla.mozilla.org/show_bug.cgi?id=1487699)

Feedback I am looking for and questions I am having:

1. Should I add objects for AWS and packet.net, and work on all three in parallel? For example, after this PR I would make a similar PR (object + `listWorkers` method) for AWS and packet.net. Or is it better to finish GCP object first?

2. Please look at creation of Worker objects - does the information I get into it make sense?

3. What on Earth is `providerData`?

4. Is it even a thing, a worker of capacity > 1? If not, maybe just remove the `capacity` prop from Worker objects?

Nota bene: I am not adding tests at this point, because it's unclear to me whether any redesign is happening, and they are just not too useful at the moment. My messy working branch (can be observed here #431) has some hint on the tests, but I am not writing them at this point

Also, linter here is failing, but if this #529 is approved, it won't be ^__^

### DIFF
--- a/services/worker-manager/src/providers/gcpprovider.js
+++ b/services/worker-manager/src/providers/gcpprovider.js
@@ -46,13 +46,14 @@ class GCPProvider extends Provider {
   async listWorkers({states, workerTypes}) {
     console.log(`Got states: ${states}, and worker types: ${workerTypes}`);
 
-    let filter = workerTypes.map(wt => `(labels.worker-type = ${wt})`).join(' OR ');
+    let wtFilter = workerTypes.map(wt => `labels.worker-type = ${wt}`).join(' OR ');
 
-    const instances = (await this.gcpCompute.getVMs({
-      autoPaginate: false,
-      filter,
-    }))[0]
-      .filter(i => states.includes(this.convertFromGCPStatus(i.metadata.status)));
+    let gcpStates = states.flatMap(this.convertToGCPStatus);
+    let sFilter = gcpStates.map(s => `status = ${s}`).join(' OR ');
+
+    let filter = `(${wtFilter}) AND (${sFilter})`;
+
+    const instances = (await this.gcpCompute.getVMs({filter}))[0];
 
     return instances.map(i => new Worker({
       id: i.id, // i.id same as i.name (must be unique); i.metadata.id is numerical id of the instance
--- a/services/worker-manager/src/providers/gcpprovider.js
+++ b/services/worker-manager/src/providers/gcpprovider.js
@@ -1,0 +1,175 @@
+'use strict';
+
+const {Bid} = require('../bid');
+const {Provider} = require('../provider');
+const Worker = require('../worker');
+const Compute = require('@google-cloud/compute');
+
+class GCPProvider extends Provider {
+
+  /**
+   * Construct a GCP Provider
+   */
+  constructor({id}) {
+    super({id});
+    this.gcpCompute = new Compute();
+  }
+
+  convertToGCPStatus(taskclusterWorkerState) {
+    switch (taskclusterWorkerState) {
+      case Provider.states.booting: return ['STAGING'];
+      case Provider.states.requested: return ['PROVISIONING'];
+      case Provider.states.running: return ['RUNNING'];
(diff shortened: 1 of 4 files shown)


services/worker-manager/src/providers/gcpprovider.js
```diff
@@ -0,0 +1,175 @@
...
+      filter,
+    }))[0]
+      .filter(i => states.includes(this.convertFromGCPStatus(i.metadata.status)));
+
+    return instances.map(i => new Worker({
+      id: i.id, // i.id same as i.name (must be unique); i.metadata.id is numerical id of the instance
+      group: i.zone.id,
+      workerType: i.metadata.labels['worker-type'],
+      state: this.convertFromGCPStatus(i.metadata.status),
+      capacity: 1,
```
owlishDeveloper: also: is it even a thing, a worker of capacity > 1? If not, maybe just remove this prop?

owlishDeveloper:
Linter here is failing, but if this #529 is approved, it won't be ^__^

djmitche:
Workers are told their capacity when they start and use that to decide how many tasks to run in parallel.  When we use huge things like g8.32xlarge or whatever, we run capacity > 1.

services/worker-manager/src/providers/gcpprovider.js
```diff
@@ -0,0 +1,175 @@
...
+   * @param workerTypes Array<String>
+   * @returns {Promise<Array<Worker>>}
+   */
+  async listWorkers({states, workerTypes}) {
+    console.log(`Got states: ${states}, and worker types: ${workerTypes}`);
+
+    let filter = workerTypes.map(wt => `(labels.worker-type = ${wt})`).join(' OR ');
+
+    const instances = (await this.gcpCompute.getVMs({
+      autoPaginate: false,
```
imbstack: What does this do?

services/worker-manager/src/providers/gcpprovider.js
```diff
@@ -0,0 +1,175 @@
...
+   * this is called in provisioner
+   *
+   * @param states Array<String>
+   * @param workerTypes Array<String>
+   * @returns {Promise<Array<Worker>>}
+   */
+  async listWorkers({states, workerTypes}) {
+    console.log(`Got states: ${states}, and worker types: ${workerTypes}`);
+
+    let filter = workerTypes.map(wt => `(labels.worker-type = ${wt})`).join(' OR ');
```
imbstack: Is it possible to include the states in this filter we send to gcp as well?

imbstack:
I think `providerData` is going to be stuff for spawning the instances such as which disks to attach, etc. I'm not 100% sure though.

imbstack:
I think the worker object makes sense at least for now. Let's just focus on gcp at the moment as well.

services/worker-manager/src/providers/gcpprovider.js
owlishDeveloper: oh, I actually have to remove this parameter. It turns off automatic pagination handling. When I started with this, I was hoping that would turn pagination off, but it seems it just switches that to manual mode.

Speaking of pagination, I have to add some code here to account for pagination

services/worker-manager/src/providers/gcpprovider.js
owlishDeveloper: Short answer is no

As you can see, the filter is conveniently a string with a rather simplistic syntax, it doesn't seem to be created for anything more complicated that couple of conditions. I tried experimenting, but it either complains on the syntax or doesn't work. 

Also, putting together that string is obnoxious, hard to write, read and debug. I decided to compromise: request filtering by worker type will produce less results than by status; and then iterate over and filter by status.

owlishDeveloper:
> I'm not 100% sure though.

ok, I will see as I write this

services/worker-manager/src/providers/gcpprovider.js
owlishDeveloper: Actually, I decided to make sure I'm not wrong, and turns out it's actually possible

PR #535, opened 2019-04-03
Loads of updates to tc-lib-iterate

djmitche:
* *error handling*: this drops the complex try-to-crash-the-process stuff and just emits `error` when `maxFailures` is reached, with instructions to the caller to handle that error or let Node crash the process.
* *starting/stopping*: the start and stop methods now wait for the corresponding event to occur, meaning that `i.stop()` might wait for a while while an iteration finishes.  In practice this doesn't mean much, since we don't ever stop an iterator in production, but it's nicer for tests.
* *improve testing* now with fake time, which has been integrated into tc-lib-testing
* drop support for `state` (this is easier to implement with a variable from a closure)


libraries/iterate/src/iterate.js
```diff
@@ -149,27 +155,32 @@ class Iterate extends events.EventEmitter {
...
           this.currentTimeout = setTimeout(resolve, this.waitTime);
         });
+        await Promise.race([stopPromise, waitTimePromise]);
+        this.onStopCall = null;
+        if (!this.keepGoing) {
+          break;
+        }
       }
     }
+    this.emit('stopped');
```
imbstack: This is much easier to read.

libraries/iterate/src/iterate.js
```diff
@@ -222,7 +234,10 @@ class Iterate extends events.EventEmitter {
 
   stop() {
     this.keepGoing = false;
-    debug('stopped');
+    if (this.onStopCall) {
+      this.onStopCall();
+    }
+    return this.stoppedPromise;
```
imbstack: Woo!

libraries/iterate/src/iterate.js
```diff
@@ -218,18 +222,10 @@ class Iterate extends events.EventEmitter {
...
-      this.emit('started');
-      try {
-        await this.iterate();
-      } catch (err) {
-        console.error(err.stack || err);
-      }
-    }, 0);
+    return new Promise(resolve => {
+      this.once('started', resolve);
+      process.nextTick(() => this.iterate());
```
imbstack: Maybe a comment to explain the `nextTick` here?

libraries/iterate/README.md
```diff
@@ -17,7 +17,7 @@ i = new Iterate({
   maxIterationTime: 10000,
   watchdogTime: 5000,
   waitTime: 2000,
-  handler: async (watchdog, state) => {
+  handler: async watchdog => {
```
imbstack: I like this change a lot. I'm just surprised nothing in any service used `state`.

libraries/testing/package.json
```diff
@@ -24,5 +24,5 @@
   "engines": {
     "node": ">=8"
   },
-  "main": "./src/testing.js"
+  "main": "./src/index.js"
```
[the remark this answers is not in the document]
djmitche: I'll add it.  There are a few other libraries which I will clean up by-and-by.

libraries/iterate/src/iterate.js
djmitche: I'm not sure why.. I'll remove it :)

libraries/iterate/README.md
djmitche: There are only four uses of this library, and three of them (in queue) are cookie-cutter)..

imbstack:
@djmitche I'll land this as soon as the merge conflicts are fixed. It isn't letting me do it through the web-ui and I can't push to your fork so I can't do it myself!

imbstack:
Carried on in #556

PR #540, opened 2019-04-03
remove now-redundant-and-incorrect logging about error

djmitche:


djmitche:
Failing queue check is https://bugzilla.mozilla.org/show_bug.cgi?id=1525444

djmitche:
> I'm wondering if this is some hold-over from the sentry days?

I think so, yeah -- I spotted it in a log a while ago and was confused

owlishDeveloper:
I understand this is redundant and incorrect after we've moved to the structured logging?

djmitche:
Yes -- it made sense when still using sentry.

owlishDeveloper:
`this.monitor.reportError` doesn't go to Sentry anymore, goes to regular logs (where the `debug` used to go)?

djmitche:
Yes, and that ends up in the GCP errors display (that's replacing sentry)

djmitche:
@owlishDeveloper is this OK to land?

owlishDeveloper:
Yes!

PR #541, opened 2019-04-03
Bug 1525030 - move all dependencies to the top level

djmitche:
Bugzilla Bug: [1525030](https://bugzilla.mozilla.org/show_bug.cgi?id=1525030)

I used a bunch of `yarn why` for things with different versions, and in all cases the solver had already picked a single version.  The `yarn.lock` diff shows no change in what's actually installed.  So I think this is Super-Safe™!


test/meta_test.js
```diff
@@ -67,10 +68,6 @@ suite('Repo Meta Tests', function() {
 
   test('Dependencies are not missing/unused', async function() {
     const depOptions = {
-      ignoreMatches: [
-        'morgan', // Peer dependency of morgan-debug
-        'ejs', // This dependency is used in web-server (see createApp.js)
-      ],
       specials: [], // don't target webpack
     };
     const root = await depcheck(ROOT_DIR, depOptions);
```
imbstack: I think some of the remainder of this test are now unused.

djmitche: I am not familiar enough with depcheck to know -- please feel free to clean it up more :)

djmitche:
(I'll fix the conflicts ASAP)

PR #545, opened 2019-04-04
Bug 1525444 - fix tests that poll for pulse messages

djmitche:
See bug for details..

Bugzilla Bug: [1525444](https://bugzilla.mozilla.org/show_bug.cgi?id=1525444)

### DIFF
--- a/libraries/testing/README.md
+++ b/libraries/testing/README.md
@@ -313,6 +313,39 @@ This function assumes the following config values:
 
 And assumes that the `exports` argument has a `load` function corresponding to a sticky loader.
 
+withPulse
+---------
+
+This function helps test applications that publish pulse messages.
+It is typically set up in `test/helper.js` like this:
+
+```js
+const {withPulse} = require('taskcluster-lib-testing');
+exports.withPulse = (mock, skipping) => {
+  withPulse({helper, skipping, namespace: 'taskcluster-someservice'});
+};
+```
+
+It assumes that `helper.loader` is set up as a sticky loader, and that there is a loader component named `pulseClient`.
+It replaces this client with a fake version, for which `libPulse.consume()` and `exchanges.publisher()` will produce fake consumers and publishers, respectively.
+The fake client can be identified by having a truthy `isFakeClient` property.
+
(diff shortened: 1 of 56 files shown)


djmitche:
I'm working on those lats two checks.. and finding lots of small bugs to fix.

djmitche:
Finally, all green!

djmitche:
ping, anyone up for reviewing this? :)

It should fix the constant failing of the tc-queue tests in CI.

libraries/testing/README.md
```diff
@@ -312,6 +312,39 @@ This function assumes the following config values:
...
+withPulse
+---------
+
+This function helps test applications that publish pulse messages.
+It is typically set up in `test/helper.js` like this:
+
+```js
+const {withPulse} = require('taskcluster-lib-testing');
+exports.withPulse = (mock, skipping) => {
+  withPulse({helper, skipping, namespace: 'taskcluster-someservice'});
```
imbstack: Nice

libraries/testing/src/with-pulse.js
```diff
@@ -0,0 +1,218 @@
...
+    await helper.load('cfg');
+    client = new FakeClient(namespace);
+    helper.load.inject('pulseClient', client);
+
+    const matchingMessageExists = (exchange, check) =>
+      client.messages.some(message =>
+        (!exchange || message.exchange.endsWith(exchange)) &&
+        (!check || check(message)));
+
+    helper.assertPulseMessage = (exchange, check) => {
```
imbstack: I notice that a lot of the checks we've replace used to have a `deepEqual`. I see that is not possible now with this more generic form but I wonder if we can have some way to at least log the payload of the message or something? Does not block landing this.

djmitche: Yeah, that annoyed me at one point and I fell back to sticking in some `console.logs`.  `_.isEqual` works, but it doesn't give any diff when it fails.  I'm not sure how to fix that..

djmitche:
rebased

Once that's green, I'll merge.

PR #547, opened 2019-04-04
Add login via mozilla auth0

helfi92:
Relates to https://github.com/taskcluster/taskcluster-rfcs/issues/142.

client ID: `t2bF4DuTfGLfdtaC4omAgTG9pU6ZlpGY`
client secret: refer to the password store


djmitche:
@helfi92 do you want me to review this?  I see you removed the review request..

helfi92:
Not yet :)

helfi92:
@djmitche I'm aware that tests are missing, however before I write any tests, I wanted to make sure I'm in the right direction. Thanks.

services/web-server/src/utils/decode.js
```diff
@@ -0,0 +1,4 @@
+// Decoding:
+// 1. Replace ! with %
+// 2. Decode with decodeUriComponent (which converts %21 to !)
+module.exports = str => decodeURIComponent(str.replace(/!/g, '%'));
```
djmitche: This and encode should probably be in a single module, with some description of the encoding they use beyond "encode" and "decode" -- those names are pretty generic!

services/web-server/src/login/handlers/github-oauth2.js
```diff
@@ -0,0 +1,50 @@
...
+    // take a user and attach roles to it
+    // this.addRoles(userProfile, user);
+
+    return user;
+  }
+
+  // exposed method
+  userFromClientId(clientId) {
+    const patternMatch = CLIENT_ID_PATTERN.exec(clientId);
+    const identity = patternMatch && patternMatch[1];
```
djmitche: This bit should probably be captured in a utility function, rather than exporting the regexp and using a numeric index into the capture groups.

services/web-server/src/login/clients/PersonAPI.js
```diff
@@ -0,0 +1,62 @@
+const assert = require('assert');
+const request = require('superagent');
+const Debug = require('debug');
+
+const debug = Debug('PersonAPI');
+const baseUrl = 'https://person.api.sso.mozilla.com/v2';
```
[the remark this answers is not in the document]
helfi92: I don't think it's there yet given the api is not considered stable yet. fwiw, I don't think there's any harm with the current behavior.

services/web-server/src/graphql/Auth.graphql
```diff
@@ -0,0 +1,42 @@
+enum AuthProvider {
```
djmitche: Should this be `IdentityProvider` or `LoginStrategy` instead of a third term?

services/web-server/src/loaders/auth.js
```diff
@@ -0,0 +1,42 @@
...
+  const getCredentials = new DataLoader(queries => {
+    return Promise.all(
+      queries.map(async (provider, accessToken) => {
+        const strategy = strategies[provider];
+
+        if (!strategy) {
+          throw new WebServerError('InputError', `Could not find a strategy for provider ${provider}`);
+        }
+
+        const user = await strategy.userFromRequest(req);
```
djmitche: Passing `req` to this function makes sense if the credentials are attached as headers (as they are for tc-login).  But since they're query parameters (`provider, accessToken`), I think those parameters will need to be passed along, instead.

helfi92: In `mozila-auth0`'s `userFromRequest`, `req` goes through `express-jwt` which I believe this package expects a request object. Once the jwt is verified, `express-jwt` sets the `req.user` property with the user credentials. Would you prefer if we provide both the request and parameters as arguments?

```jsx
const user = await strategy.userFromRequest({
  req,
  parameters: req.body.variables,
});
```

djmitche: I bet that's why the original tc-login implementation expected the JWT in a header instead of in the request body.  That seemed awkward with REST, and seems even more awkward with GraphQL.  I suspect we need to not use `express-jwt`, and instead use a lower-level JWT library to validate and examine the token.  Or we could shift the definition of the GraphQL query to include only the identityProviderId in the query and require that the token be provided in a header for mozilla-auth0 (and maybe for Github too?).

helfi92:
Rebased.

services/web-server/config.yml
```diff
@@ -22,6 +22,12 @@ defaults:
     credentials:
```
djmitche: I think app.loginStrategies can be removed now?

services/web-server/src/utils/verifyJwt.js
```diff
@@ -0,0 +1,39 @@
+const assert = require('assert');
+const jwt = require('jsonwebtoken');
+const jwks = require('jwks-rsa');
+
+module.exports = async ({ token, domain }) => new Promise((resolve, reject) => {
+  assert(token, 'No authorization token was found');
+  assert(domain, 'A domain is required to verify the jwt');
+
+  const client = jwks({
```
djmitche: I think we want to construct this once, not for every call -- it rate-limits itself and caches the value between calls.

ui/src/components/SignInDialog/index.jsx
```diff
@@ -82,6 +84,20 @@ export default class SignInDialog extends Component {
...
+                target="_blank">
+                <ListItemAvatar>
+                  <Avatar>
+                    <LoginVariantIcon />
+                  </Avatar>
+                </ListItemAvatar>
+                <ListItemText primary="Sign in with Auth0" />
+              </ListItem>
+            )}
             {process.env.LOGIN_STRATEGIES.includes('github') && (
```
djmitche: Should this be github-oauth2?

services/web-server/src/utils/verifyJwt.js
```diff
@@ -0,0 +1,39 @@
...
+    });
+  };
+
+  jwt.verify(
+    token,
+    getKey,
+    {
+      // and expect a token issued by auth0
+      issuer: `https://${domain}/`,
+      algorithms: ['RS256'],
```
djmitche: I think we need to [check the audience, too](https://github.com/taskcluster/taskcluster/blob/master/services/login/src/handlers/mozilla-auth0.js#L32)?

djmitche:
By the way, I deployed this on https://dustin.taskcluster-dev.net, and the github logins are hanging at the callback..

helfi92:
> By the way, I deployed this on https://dustin.taskcluster-dev.net, and the github logins are hanging at the callback..

Seems to work for me. I wonder why...

Screen Shot 2019-04-30 at 1 02 26 PM

djmitche:
Weird!  I can't see any errors in the console, and the network traffic looks normal (just loads the page and the favicon).

services/web-server/src/utils/verifyJwt.js
helfi92: I thought so too but I was getting a weird `aud` value. See https://github.com/mozilla-iam/cis/issues/423.

djmitche: Yeah, the audience and scope are part of the way we support third-party logins.  Basically when the frontend (be that tc-ui or treeherder or releng-services) requests an auth0 login, it indicates an additional audience and scope that correspond to an auth0 "API" object which we asked the IAM team to create.  When Auth0 sees this, it presents a consent dialog saying "do you want to grant treeherder access to Taskcluster?"  The result is an access token that has two audiences (treeherder and Taskcluster), and both of those services can rely on it to authenticate the user.  So this is how we support third-party logins.

It's extra work to get the IAM team to set this up, and since we don't want to support this method of third-party logins, it's probably not worth it.

However, we do still need to verify that the access_token was meant for us - that is, that we are the intended audience!  That is done by checking that the audience is equal to the auth0 client id.  Without this, I could steal your TC credentials by:
 * setting up some toy Mozilla site (maybe another Bugzilla dashboard!)
 * configuring auth0 for that site
 * getting you to login
 * taking the resulting access_token and calling tc-web-server's `getCredentials` query with it

services/web-server/src/login/User.js
```diff
@@ -0,0 +1,80 @@
...
+
+  scopes() {
+    let scopes = this.roles.map(role => 'assume:' + role);
+    // the `login-identity:*` role defines what each user gets access to.
+    scopes.push(`assume:login-identity:${this.identity}`);
+
+    return scopes;
+  }
+
+  createCredentials(options) {
```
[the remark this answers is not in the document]
djmitche: #662

PR #555, opened 2019-04-09
rename libraries' index files to index.js

djmitche:
..just another cleanup on the trip home..


djmitche:
(This matches what I wrote in the best-practices a week or so ago)

PR #557, opened 2019-04-09
[Bug 1540838] Make lib-pulse tests actually run

imbstack:
Bugzilla Bug: [1540838](https://bugzilla.mozilla.org/show_bug.cgi?id=1540838)


imbstack:
Quick link to the [logs for lib-pulse tests](https://tools.taskcluster.net/groups/b_XadFz8S3u8BiK_KkfkbQ/tasks/PamwHvyUToGPsGPx4UAEfw/runs/0/logs/public%2Flogs%2Flive.log) showing that they work!

infrastructure/docker-images/build.sh
```diff
@@ -15,3 +17,14 @@ RUN apt-get update && apt-get install -y firefox-esr xvfb
...
 
 docker build -t "taskcluster/browser-test:${node_version}" ${tmpdir}
+
+# Next we do the rabbit-test image
+
+tmpdir=$(mktemp -d)
+trap "cd /; rm -rf ${tmpdir}" EXIT
+cat > ${tmpdir}/Dockerfile <<EOF
+FROM node:${node_version}-stretch
+RUN apt-get update && apt-get install -y rabbitmq-server
```
djmitche: I wonder if we'll run into trouble with RabbitMQ versions here.  We can adjust if that happens..

libraries/testing/src/secrets.js
```diff
@@ -36,7 +36,7 @@ class Secrets {
           // delay loading cfg until we know we need it (allowing this to work with
           // loaders that do not have a `cfg` component); note that this also ensures
           // later calls to load.cfg(..) will work.
-          if (!cfg) {
+          if (!cfg && this.load) {
```
djmitche: This makes mockSuite able to operate without stickyLoader, right?  If so, that should get a docs update in the README.

infrastructure/docker-images/build.sh
imbstack: Yeah, I figure let's do the easiest thing now and build this up to be more complete later.

PR #558, opened 2019-04-11
[WIP][UI] Remove search pattern for clients viewer

### LINKED ISSUE #513: [UI] remove search pattern for clients viewer
On the old tools site, by default we searched for clients whose names have the current user's clientId as a prefix.  This was done because we have a convention of naming users' clients after their own clientId (and tc-login gives them permission to do so).  That convention is now embodied in the client creator, and most users generate clients that way.

The old UI showed the search pattern right above the list.  The new UI shows it far away in the upper-right corner, in a rather dim typeface.  This has now tripped me up a few times, where I don't see the client I'm looking for and go wondering how it got deleted.

Let's just default to showing all clients, like we do for roles.

projectyang:
Fixes #513


djmitche:
It looks (from the addition of `package-lock.json`) like you're using `npm` instead of `yarn`?  Please be sure to squash changes so that the addition of that file doesn't get recorded in the final PR.

PR #561, opened 2019-04-11
Generate Helm Templates

imbstack:
This work will be carried forward by @edunham to a large degree. Getting it up for review so we can talk it through.


edunham:
The `Kind:` annotations are meant to expand to things like https://github.com/taskcluster/taskcluster/blob/master/infrastructure/terraform/modules/deployment/deployment.yaml

edunham:
Regarding the infrastructure/builder bits: 

https://github.com/taskcluster/taskcluster/tree/master/infrastructure/builder 

most of the work needed will probably go into https://github.com/taskcluster/taskcluster/pull/561/files#diff-b4719223275b7b6fcb079a4da72b4acfR42

edunham:
```
101 15:20:38 < edunham> so the generator is to build out the helm templates themselves per-service?
101 15:22:57 < bstack> yeah, that's the general idea. I also tried to have handwritten helm templates and have the generator generate values.json that would then loop over the handwritten templates
101 15:22:59 < bstack> but this way seemed easier
```

edunham:
> 101 15:25:41 < bstack> ah, one more thing. `yarn generate` is run and checked in so it won't have to be run at the time that you run helm later

edunham:
101 15:28:46 < edunham> when you template stuff with yarn, where do you prefer the source values to live?
101 15:29:38 < bstack> I was just pondering that
101 15:29:46 < bstack> I think right next to the code is fine
101 15:29:50 < bstack> in src/
101 15:30:25 < bstack> the templates for each tc service should end up being mostly identical other than env vars I think
101 15:30:37 < bstack> and then there's a few strange services that are quite different
101 15:30:54 < bstack> and things like replicas will be set at `helm time I guess`
101 15:30:54 < edunham> do you have the list of which those are on the top of your head?
101 15:34:34 < dustin> probably https://github.com/taskcluster/taskcluster/blob/master/infrastructure/terraform/web-ui.tfand 
                       https://github.com/taskcluster/taskcluster/blob/master/infrastructure/terraform/references-service.tf
101 15:34:52 < dustin> lol I just noticed https://github.com/taskcluster/taskcluster/blob/master/infrastructure/terraform/references-service.tf#L6
101 15:35:15 < dustin> hm, in fact neither of those is very weird
101 15:35:34 < dustin> we have hidden a lot of the weirdness in the docker image
101 15:36:01 < dustin> (specifically, web-ui runs `yarn build` and then starts nginx, while references runs a shell script then starts nginx, and all of the other services just run a node web server)
101 15:36:54 < edunham> they all take the same image?
101 15:37:22 < dustin> yes
101 15:37:37 < dustin> its entrypoint dispatches to the various startup things based on command-line argument
101 15:37:49 < bstack> ah, so maybe it will be easier than I expected
101 15:37:56 < dustin> docker run <image> auth/web runs the web process for the auth service, etc.

imbstack:
The cloudops riff on this is in #883

PR #562, opened 2019-04-11
Add list worker types endpoint

owlishDeveloper:


services/worker-manager/src/api.js
```diff
@@ -145,3 +145,17 @@ builder.declare({
...
+  method: 'get',
+  route: '/workertypes',
+  name: 'listWorkerTypes',
+  title: 'List All WorkerTypes',
+  stability: APIBuilder.stability.experimental,
+  description: [
+    'TODO',
+  ].join('\n'),
+}, async function(req, res) {
+  await this.WorkerType.scan();
```
owlishDeveloper: actually, this maybe will be another table. Information I want from it:

```json
{
name
pending tasks
pending capacity
running capacity
last active
last resolved
number of failed / exception / unsched
provider name
}
```

It's possible that I will actually need to query some other service for some of that info (we currently query Queue for tasks and things), but I'll still need a `GET /workerTypes` sort of endpoint

imbstack: I think you can get name, pending capacity, running capacity, and provider name from here. The rest will probably require asking the queue?

PR #563, opened 2019-04-11
Bug 1534713 - ignore errors draining a broken channel

djmitche:
This adds a new test that simulates the errors detailed in the bug, and
then ignores them.  They come from trying to cleanly shut down a
connection that has already failed.

Bugzilla Bug: [1534713](https://bugzilla.mozilla.org/show_bug.cgi?id=1534713)


djmitche:
Not related, but the branch point for this PR is before the fix for that landed, so not surprising either.

PR #565, opened 2019-04-12
Bug 1541741 - log API calls and satisfying scopes in the services themselves

djmitche:
Bugzilla Bug: [1541741](https://bugzilla.mozilla.org/show_bug.cgi?id=1541741)


djmitche:
Ok, ready for review :)

I'd like two reviewers for this since it's security-critical.

Per irc discussion, we are willing to accept the risk that the logging shim will fall over and we'll lose important logging data.

libraries/api/src/middleware/logging.js
```diff
@@ -0,0 +1,41 @@
+/**
+ * Log an API request on completion, including information determined
+ * by the `remoteAuthentication` middleware, if present.
+ */
+const logRequest = ({builder, entry, monitor}) => {
+  const apiMethod = `${builder.serviceName}.${builder.apiVersion}.${entry.name}`;
```
imbstack: Isn't `serviceName` available from the logger name in the message that is created? Why duplicate here?

libraries/api/src/middleware/logging.js
```diff
@@ -0,0 +1,41 @@
...
+          hasAuthed: req.hasAuthed,
+          resource: req.originalUrl,
+          method: req.method,
+          clientId: req.hasAuthed ? await req.clientId() : '',
+          satisfyingScopes: req.satisfyingScopes ? req.satisfyingScopes : [],
+          statusCode: res.statusCode,
+          duration: d[0] * 1000 + d[1] / 1000000,
+        });
+      } catch (err) {
+        monitor.reportError(err);
```
imbstack: Should we consider moving the `monitor.log.apiMethod` into a `finally` of this try/catch? I don't want there to be many situations where it doesn't end up getting logged.

libraries/monitor/src/builtins.js
```diff
@@ -39,14 +40,20 @@ module.exports = [
...
-      public: 'Will be true if the endpoint requires no scopes.',
+      duration: 'The duration in ms of the call.',
+      public: 'True if the endpoint requires no scopes.',
       hasAuthed: 'Will be true if the request had req.authorize() called on it.',
-      clientId: 'The clientId that made the request if there was one.',
+      clientId: 'The clientId that made the request, if there was one.',
+      satisfyingScopes: `
+        The set of scopes posessed by the caller that were used to authorize this request, or [] if scopes were not required.
+
+        This set of scopes is intuitively "the minimal set of scopes required to satisfy the scope expression" but is not quite minimal in one sense:
```
imbstack: "minimal set of scopes possessed by the client"?

services/auth/src/signaturevalidator.js
```diff
@@ -349,20 +349,6 @@ const createSignatureValidator = function(options) {
         message: message.toString(),
       };
     }
-    options.monitor.log.signatureValidation({
```
imbstack: I think we might want to still logs this, just without `scopes`?

libraries/api/src/middleware/logging.js
djmitche: Hm, true.  I did want to make it a bit more explicit since we have so many methods named `list` for example, but you're right -- this is redundant.

libraries/api/src/middleware/logging.js
djmitche: The only other bit in the function is a conditional and `process.hrtime`, though.. we'd add more code to handle the case where `duration` wasn't set, and include that in the `finally` block.

libraries/monitor/src/builtins.js
djmitche: It needs to have some reference to the API method.  How about "the minimal set of scopes required to satisfy the API method's authorization requirements"?

services/auth/src/signaturevalidator.js
djmitche: The authenticateHawk call will be logged.. is that enough?  What is this useful for?

djmitche: I ask because this is the third log line per API method call, so it's high-volume..

libraries/api/src/middleware/logging.js
imbstack: Maybe we should just remove the try/catch in that case.

libraries/monitor/src/builtins.js
imbstack: The only problem I have with that is that it makes it seem like something we can know about a given endpoint without knowing the scopeset that we tried to evaluate it with.

services/auth/src/signaturevalidator.js
imbstack: Ah, good point. I guess if we make sure that all of the contents of this message end up in the `apiMethod` message then we can remove this.

The only other consideration is that we know this one will end up in papertrail. Not sure if the data that is in `authenticateHawk`'s `apiMethod` message is enough to replace this?

In [bug 1538010](https://bugzilla.mozilla.org/show_bug.cgi?id=1538010) we have started sending this off to security's setup. We should let them know this is changing or going away depending on what we decide to do.

libraries/scopes/src/expressions.js
```diff
@@ -36,6 +37,64 @@ exports.satisfiesExpression = function(scopeset, expression) {
...
+ * Given a scope expression and set of scopes that satisfies it, return a
+ * subset of those scopes that satisfies the scope expression.  This subset is
+ * not necessarily minimal: in the case of AnyOf, it includes all scopes
+ * satsifying any branch of that alternative.  To do otherwise would return
+ * only one of several possible minimal scope sets.
+ *
+ * Returns undefined if the scopeset does not satisfy the expression.
+ */
+exports.scopesSatisfying = (scopeset, expression) => {
+  let scopes = [];
```
imbstack: What if we called this `used` or something like that?

libraries/scopes/src/expressions.js
```diff
@@ -36,6 +37,64 @@ exports.satisfiesExpression = function(scopeset, expression) {
...
+    if (typeof expr === 'string') {
+      if (scopeset.some(s => patternMatch(s, expr))) {
+        scopes.push(expr);
+        return true;
+      }
+      return false;
+    }
+
+    if (expr.hasOwnProperty('AllOf')) {
+      let scopeIndex = scopes.length;
```
imbstack: `startingIndex`?

Not necessary to change, just makes it easier to understand for me.

libraries/scopes/src/expressions.js
```diff
@@ -36,6 +37,64 @@ exports.satisfiesExpression = function(scopeset, expression) {
...
+
+ * Given a scope expression and set of scopes that satisfies it, return a
+ * subset of those scopes that satisfies the scope expression.  This subset is
+ * not necessarily minimal: in the case of AnyOf, it includes all scopes
+ * satsifying any branch of that alternative.  To do otherwise would return
+ * only one of several possible minimal scope sets.
+ *
+ * Returns undefined if the scopeset does not satisfy the expression.
+ */
+exports.scopesSatisfying = (scopeset, expression) => {
```
imbstack: This looks great. Easy to understand.

libraries/scopes/test/expression_test.js
```diff
@@ -50,8 +51,10 @@ suite(testing.suiteName(), function() {
   });
 
   suite('scope expression satisfaction:', function() {
+    // We want to be confident that `satisfiesExpression` and `scopesSatisfyingExpression` both
+    // accept and reject the same inputs, so they are tested in parallel.
```
imbstack: ++

imbstack:
This looks great to me. Requesting @ajvb as second reviewer for a security person's point of view.

libraries/monitor/src/builtins.js
djmitche: Good point -- I'll add some sentences.

services/auth/src/signaturevalidator.js
djmitche: I'll leave this in for now, and filed [bug 1544942](https://bugzilla.mozilla.org/show_bug.cgi?id=1544942) to turn it off.

PR #569, opened 2019-04-13
Stop using the promise dependency

djmitche:
Since Node 0.12, Promise has been built-in and available as a global.

For a while, we used a Promise library because `Promise.race` was not
available.  It is available now.


djmitche:
Ah, I was thinking `Promise.race`.  It looks like we stopped using Bluebird in [bug 1518564](https://bugzilla.mozilla.org/show_bug.cgi?id=1518564).

PR #589, opened 2019-04-16
Bug 1544807 - improve handling of huge error messages

djmitche:
Bugzilla Bug: [1544807](https://bugzilla.mozilla.org/show_bug.cgi?id=1544807)


djmitche:
If we leave a URL, then we need to somehow store those scopes.. I think this will be useful in the limited cases where people are experimenting and have just a few scopes set up.

djmitche:
OK, updated to always omit the set of scopes, and to give different forms when there are no credentials and when there are insufficient credentials.

PR #596, opened 2019-04-17
only serve /playground in development

djmitche:
It probably doesn't hurt to have the playground out there in production, but it seems like leaving dangling functionality is unnecessary risk.

### DIFF
--- a/services/web-server/src/index.js
+++ b/services/web-server/src/index.js
@@ -153,10 +153,12 @@ const load = loader(
 
         /* eslint-disable no-console */
         console.log(`\n\nWeb server running on port ${cfg.server.port}.`);
-        console.log(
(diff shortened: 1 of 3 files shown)


services/web-server/src/index.js
```diff
@@ -153,10 +153,12 @@ const load = loader(
 
         /* eslint-disable no-console */
         console.log(`\n\nWeb server running on port ${cfg.server.port}.`);
-        console.log(
-          `\nOpen the interactive GraphQL Playground and schema explorer in your browser at:
-        http://localhost:${cfg.server.port}/playground\n`
-        );
+        if (cfg.server.playground) {
```
helfi92: `cfg.app.playground`.

djmitche: good spot - I changed it at the last minute :)

PR #597, opened 2019-04-17
Bug 1541741 - handle API methods that are sometimes public

djmitche:
Bugzilla Bug: [1541741](https://bugzilla.mozilla.org/show_bug.cgi?id=1541741)


djmitche:
It does!

PR #599, opened 2019-04-17
Bug 1538169 - retry yarn operations in a new task on failure

djmitche:
Bugzilla Bug: [1538169](https://bugzilla.mozilla.org/show_bug.cgi?id=1538169)


djmitche:
I think those are covered here -- for example the `lint` task has this setting applied.

PR #466, opened 2019-03-18
ui to display lastfires

Biboswan:
Continue PR of https://github.com/taskcluster/taskcluster-web/pull/413

PR #606, opened 2019-04-18
Change firedBy to enum

Biboswan:
#466

### DIFF
--- a/generated/references.json
+++ b/generated/references.json
@@ -4452,6 +4452,12 @@
                 "type": "string"
               },
               "firedBy": {
+                "enum": [
+                  "schedule",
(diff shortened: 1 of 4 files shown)


Biboswan:
oh i missed something

djmitche:
Ah, and I see now they did not.  Also, this will need a run of `yarn generate`.  Then it should be good to merge :)

Biboswan:
We use firedBy 'test'. Now I remember why I kept it any string. So the fix is to add a `test` enum as well in the schema ?

djmitche:
Probably better to change the tests to use an allowed value.

PR #608, opened 2019-04-18
Show deploy previews against the staging instance

djmitche:
This helps ensure that we are not using production credentials on untested code, nor sharing them with netlify (I *think* netlify has not yet been sec-reviewed?)


djmitche:
This *works* just fine, judging by a click on the preview link.  However, we are currently linking to the "main" netlify deployment and it would be confusing if that suddenly started showing data from the staging site (specifically, not the same tasks!).

Maybe I should give up on this idea, and just disable sign-ins on netlify deployments?

PR #612, opened 2019-04-19
Add payload in Signature in API Docs

### LINKED ISSUE #605: [UI] "signature" in API docs should include payload
**Describe the bug**
The signature of a method that takes a request payload should include the `payload` argument

**To Reproduce**
1. Go to https://docs.taskcluster.net/docs/reference/core/notify/api#email
1. Expand the email method

**Expected behavior**
Should see `email(payload)`.

**Screenshots**
Screenshot_2019-04-18 Api - Taskcluster

arshadkazmi42:
Fixes #605 
-

Adds `payload` as function parameter of signature in api docs

### DIFF
--- a/ui/src/views/Documentation/Reference/Entry.jsx
+++ b/ui/src/views/Documentation/Reference/Entry.jsx
@@ -109,6 +109,10 @@ export default class Entry extends Component {
       ? entry.args.concat(`{${entry.query.join(', ')}}`)
       : entry.args;
 
+    if (entry.input) {
+      parameters.unshift('payload');
+    }
+
     return `${entry.name}(${parameters.join(', ')}) : ${
       entry.output ? 'result' : 'void'
     }`;
--- a/ui/src/views/Documentation/Reference/Entry.jsx
+++ b/ui/src/views/Documentation/Reference/Entry.jsx
@@ -109,8 +109,8 @@ export default class Entry extends Component {
       ? entry.args.concat(`{${entry.query.join(', ')}}`)
       : entry.args;
 
-    if (entry.input) {
-      parameters.unshift('payload');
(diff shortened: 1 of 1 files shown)


djmitche:
By the way -- and this is a tiny detail but worth mentioning -- we and most projects use the imperative tense in git commit messages and PR titles.  So this would be "Add payload in signature in API docs".

ui/src/views/Documentation/Reference/Entry.jsx
```diff
@@ -109,8 +109,8 @@ export default class Entry extends Component {
       ? entry.args.concat(`{${entry.query.join(', ')}}`)
       : entry.args;
```
djmitche: I think this is actually the problem -- this puts the array `entry.args` in `parameters` by reference, and then pushes to it.  So every time this function runs, a new element is added to the end of `entry.args`.  We really shouldn't be modifying `entry.args`.  So I think the fix is to make a copy of the array here.  That can be done with `[...entry.args]`.

ui/src/views/Documentation/Reference/Entry.jsx
```diff
@@ -109,8 +109,8 @@ export default class Entry extends Component {
...
 
-    if (entry.input) {
-      parameters.unshift('payload');
+    if (entry.input && parameters.indexOf('payload') === -1) {
+      parameters.push('payload');
```
djmitche: But `push` instead of `unshift` is the right fix :)

ui/src/views/Documentation/Reference/Entry.jsx
arshadkazmi42: Ah. yes. I completely missed this. Let me fix it right away

PR #616, opened 2019-04-19
Stop requiring assume() in production code

djmitche:
This allows us to drop it as a production dependency.


libraries/loader/src/index.js
```diff
@@ -57,15 +57,16 @@ function renderGraph(componentDirectory, sortedComponents) {
...
+  assert(typeof componentDirectory === 'object');
   if (virtualComponents instanceof Array) {
     virtualComponents = virtualComponents.reduce((acc, k) => {
       acc[k] = null;
       return acc;
     }, {});
   }
   const virtualKeys = Object.keys(virtualComponents);
-  assume(Object.keys(componentDirectory).filter(x => virtualKeys.includes(x))).has.length(0);
+  assert(Object.keys(componentDirectory).every(k => !virtualKeys.includes(k)),
```
imbstack: I am confused as to why this is causing tests to fail. To my brain this looks equivalent.

djmitche: It is -- it's just the test was looking for a different error message.  Sorry about that!

PR #619, opened 2019-04-20
 [Bug 1519730] Replace sqs with pulse

Biboswan:
(https://bugzilla.mozilla.org/show_bug.cgi?id=1519730)

### DIFF
--- a/services/notify/src/exchanges.js
+++ b/services/notify/src/exchanges.js
@@ -82,7 +82,7 @@ exchanges.declare({
   title: 'IRC Notification Messages',
   description: [
     'A message which is to be sent to an irc channel or',
-    'user is published to this exchange'
+    'user is published to this exchange',
   ].join('\n'),
   routingKey: buildCommonRoutingKey(),
   schema: 'notification-message.yml',
--- a/services/notify/src/exchanges.js
+++ b/services/notify/src/exchanges.js
@@ -75,3 +75,18 @@ exchanges.declare({
   routingKeyBuilder: commonRoutingKeyBuilder,
   CCBuilder: commonCCBuilder,
 });
+
+exchanges.declare({
+  exchange: 'irc-notification',
+  name: 'ircNotify',
+  title: 'IRC Notification Messages',
+  description: [
+    'A message which is to be sent to an irc channel or',
+    'user is published to this exchange'
+  ].join('\n'),
+  routingKey: buildCommonRoutingKey(),
+  schema: 'notification-message.yml',
+  messageBuilder: commonMessageBuilder,
+  routingKeyBuilder: commonRoutingKeyBuilder,
+  CCBuilder: commonCCBuilder,
+});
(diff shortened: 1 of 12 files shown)


Biboswan:
oops i forgot to yarn generate

services/notify/src/irc.js
```diff
@@ -70,69 +69,47 @@ class IRCBot {
...
+      queueName: 'irc-notifications',
+    },
+    this.monitor.timedHandler('notification', this.onMessage.bind(this))
+    );
   }
 
-  async notify({channel, user, message}) {
+  async onMessage({payload}) {
+    let {channel, user, message} = payload.message;
+    let retires = 0;
```
imbstack: `retries` rather than `retires`

services/notify/src/notifier.js
```diff
@@ -117,7 +110,8 @@ class Notifier {
     });
   }
 
-  async irc({channel, user, message}) {
+  async irc(msg) {
```
imbstack: Could we just call this `messageRequest` or something? It will make the following function a bit less confusing.

services/notify/src/irc.js
```diff
@@ -70,69 +69,47 @@ class IRCBot {
...
-      // This callback does not ever have an error. If it triggers, we have succeeded
-      // Time this out after 10 seconds to avoid blocking forever
-      await new Promise((accept, reject) => {
-        setTimeout(() => {
-          debug('Timed out joining channel, may be ok. Proceeding.');
-          accept();
-        }, 10000);
-        this.client.join(channel, accept);
-      });
+    while(true) {
```
imbstack: We can actually simplify this I think. Since [pulse consumers](https://github.com/taskcluster/taskcluster/tree/master/libraries/pulse#pulseconsumer) will retry a single time when a handler fails, we can just fail the handler when something goes wrong and hope that the single retry fixes things. This irc stuff is best effort for the most part so once is enough. We can get rid of any explicit retry logic in here including `MAX_RETRIES`. Does that make sense?

imbstack:
Wonderful work so far! Thanks so much for hacking on this.

services/notify/src/irc.js
Biboswan: ah silly typo

services/notify/src/exchanges.js
```diff
@@ -75,3 +75,18 @@ exchanges.declare({
   routingKeyBuilder: commonRoutingKeyBuilder,
   CCBuilder: commonCCBuilder,
 });
+
+exchanges.declare({
```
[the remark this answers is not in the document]
Biboswan: Do you mean `sqs` if message is for `user` or else use `pulse` if message is for `channel`. But then it won't be cloud-agnostic! We can have some scope like thing so that anyone can't just bind to that exchange `irc-notification` ?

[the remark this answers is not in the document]
djmitche: Very little is secret in Taskcluster.  The task that generated the notification isn't secret.  And irc isn't secret.  I'm not too worried about this.  A note in the docs might help :)

"Note that all notifications are transmitted via pulse and should not be considered private."

Biboswan:
@imbstack last commit has the changes

PR #625, opened 2019-04-22
Bug 1545857 - Workers not displaying

helfi92:


helfi92:
> BTW, I have acquired a copy of "Learning GraphQL" so hopefully I can soon be less of a n00b on the topic!

Awesome! Please do share any interesting bits you come across while reading the book :)

PR #628, opened 2019-04-23
[UI] Add fetchPolicy network-only to task view

helfi92:
This will make sure data is up-to-date when user visits an already cached task.


helfi92:
Pulse tests are failing. Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1546388.

PR #632, opened 2019-04-24
Add `list worker types` endpoint to worker manager

owlishDeveloper:
Bugzilla Bug: [1546789](https://bugzilla.mozilla.org/show_bug.cgi?id=1546789)

### DIFF
--- a/services/worker-manager/schemas/v1/workertype-list
+++ b/services/worker-manager/schemas/v1/workertype-list
@@ -1,0 +1,27 @@
+$schema: "/schemas/common/metaschema.json#"
+title: WorkerType Full
+description: A complete workerType definition
+type: object
+properties:
+  data:
+    type: array
+    title: "List of all workertypes"
+    description: List of all workertypes
+    uniqueItems: true
+    items: {$ref: 'workertype-full.json#'}
+
+  continuationToken:
+    type:               string
+    title:              "Continuation Token"
+    description: |
+      Opaque `continuationToken` to be given as query-string option to get the
+      next set of worker-types in the worker-manager.
+      This property is only present if another request is necessary to fetch all
+      results. In practice the next request with a `continuationToken` may not
+      return additional results, but it can. Thus, you can only be sure to have
+      all the results if you've called `listWorkerTypes` with `continuationToken`
+      until you get a result without a `continuationToken`.
+
+additionalProperties: true
+required:
+  - data
--- a/services/worker-manager/src/api.js
+++ b/services/worker-manager/src/api.js
@@ -177,7 +177,7 @@ builder.declare({
   title: 'List All WorkerTypes',
   stability: APIBuilder.stability.experimental,
   description: [
-    'TODO',
+    'Get the list of all the existing workertypes',
   ].join('\n'),
 }, async function(req, res) {
   const { continuationToken } = req.query;
--- a/services/worker-manager/src/api.js
+++ b/services/worker-manager/src/api.js
@@ -176,6 +176,7 @@ builder.declare({
   name: 'listWorkerTypes',
   title: 'List All WorkerTypes',
   stability: APIBuilder.stability.experimental,
+  output: 'workertype-list.yml',
   description: [
     'Get the list of all the existing workertypes',
   ].join('\n'),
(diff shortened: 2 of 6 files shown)


services/worker-manager/src/api.js
```diff
@@ -165,3 +165,32 @@ builder.declare({
...
+  const limit = parseInt(req.query.limit || 100, 10);
+  const scanOptions = {
+    continuation: continuationToken,
+    limit,
+  };
+
+  const data = await this.WorkerType.scan({}, scanOptions);
+
+  if (data.continuation) {
+    data.continuationToken = data.continuation;
```
owlishDeveloper: not sure about this 🤔 your thoughts @imbstack ?

imbstack: Fine with me!

services/worker-manager/src/api.js
```diff
@@ -165,3 +165,32 @@ builder.declare({
...
+  route: '/workertypes',
+  query: {
+    continuationToken: /./,
+    limit: /^[0-9]+$/,
+  },
+  name: 'listWorkerTypes',
+  title: 'List All WorkerTypes',
+  stability: APIBuilder.stability.experimental,
+  description: [
+    'TODO',
```
imbstack: Let's add a description and get this landed!

Only other thing is an output schema. It should just be an array of the workertype-full objects. You can use `$ref` to set that up.

services/worker-manager/src/api.js
```diff
@@ -165,3 +165,32 @@ builder.declare({
...
+builder.declare({
+  method: 'get',
+  route: '/workertypes',
+  query: {
+    continuationToken: /./,
+    limit: /^[0-9]+$/,
+  },
+  name: 'listWorkerTypes',
+  title: 'List All WorkerTypes',
+  stability: APIBuilder.stability.experimental,
```
imbstack: Just need to add the new schema as `output: ...`.

services/worker-manager/schemas/v1/workertype-list
```diff
@@ -0,0 +1,27 @@
+$schema: "/schemas/common/metaschema.json#"
+title: WorkerType Full
```
imbstack: WorkerType List

services/worker-manager/schemas/v1/workertype-list
```diff
@@ -0,0 +1,27 @@
...
+description: A complete workerType definition
```
imbstack: A list of workerType definitions

services/worker-manager/schemas/v1/workertype-list
imbstack: We need to add the `.yml` at the end of this filename.

services/worker-manager/schemas/v1/workertype-list
```diff
@@ -0,0 +1,27 @@
...
+type: object
+properties:
+  data:
```
imbstack: Let's have this key be `workertypes` to keep the same format as the list of clients from auth has for instance.

services/worker-manager/schemas/v1/workertype-list
```diff
@@ -0,0 +1,27 @@
...
+    description: |
+      Opaque `continuationToken` to be given as query-string option to get the
+      next set of worker-types in the worker-manager.
+      This property is only present if another request is necessary to fetch all
+      results. In practice the next request with a `continuationToken` may not
+      return additional results, but it can. Thus, you can only be sure to have
+      all the results if you've called `listWorkerTypes` with `continuationToken`
+      until you get a result without a `continuationToken`.
+
+additionalProperties: true
```
imbstack: This should be false, right?

imbstack:
It occurs to me that this was probably still WIP when I did my followup review and you hadn't requested me to look at it yet. Sorry to jump the gun.

services/worker-manager/schemas/v1/workertype-list
owlishDeveloper: https://github.com/taskcluster/taskcluster/pull/632#discussion_r279103307 
I understand there also will be `continuation` prop

owlishDeveloper:
This PR was never a WIP

PR #634, opened 2019-04-25
Fix: Links in documentation

helfi92:
Relates to https://bugzilla.mozilla.org/show_bug.cgi?id=1546927.

### DIFF
--- a/ui/docs/manual/design/apis/hawk/README.md
+++ b/ui/docs/manual/design/apis/hawk/README.md
@@ -7,7 +7,7 @@ order: 10
 All Taskcluster components use [Hawk](https://github.com/hueniverse/hawk) over
 SSL for authentication and authorize access based on a set of "scopes"
 associated with each client. Credentials and scopes are managed by the
-[authentication](/docs/reference/platform/taskcluster-auth/references/api) component.
+[authentication](/docs/reference/platform/auth/api) component.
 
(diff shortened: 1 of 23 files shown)


ui/docs/manual/design/apis/hawk/README.md
```diff
@@ -7,7 +7,7 @@ order: 10
 All Taskcluster components use [Hawk](https://github.com/hueniverse/hawk) over
 SSL for authentication and authorize access based on a set of "scopes"
 associated with each client. Credentials and scopes are managed by the
-[authentication](/docs/reference/platform/taskcluster-auth/references/api) component.
```
djmitche: I think this transform is the biggest cause of 404s (for my usage anyway).  Maybe we could encode some route translation to do this automatically?

helfi92: Sure. I think we can do something like:

```jsx
const path = require('path');
const re = /(\/docs\/reference\/.*)\/(?:taskcluster-)(.*)(?:\/references)\/(api)/;
const matches = request.uri.match(re);

if (matches) {
  const uri = path.join(matches[1], matches[2], matches[3]);
  request.uri = uri;
  return callback(// ...);
}

// continue with the rest
```

The AWS console seems to be down. I'll do it once it's back up.

PR #638, opened 2019-04-26
Bug 1547000 - don't show diffs for generated files

djmitche:
Note that these files are still shown in a PR, just without the
line-by-line diff -- much like `yarn.lock` is already.

Note, too, that this does not cover files which are *edited* by yarn
generate, such as `package.json` or the READMEs.  Diffs in those files
will appear in PRs as expected.

Bugzilla Bug: [1547000](https://bugzilla.mozilla.org/show_bug.cgi?id=1547000)


djmitche:
Hm, I wonder if this will mean that those files are also ignored for calculating recommended reviewers?  We'll see!

PR #641, opened 2019-04-26
Bug 1542405 - update tc-github handler tests to use new fake pulse support

djmitche:
Bugzilla Bug: [1542405](https://bugzilla.mozilla.org/show_bug.cgi?id=1542405)


owlishDeveloper:
I love that we are now testing the bindings!

PR #642, opened 2019-04-26
Bug 1544942 - replace signatureValidated logs with apiMethod

djmitche:
Bugzilla Bug: [1544942](https://bugzilla.mozilla.org/show_bug.cgi?id=1544942)


libraries/api/test/logging_test.js
```diff
@@ -112,6 +114,7 @@ suite(testing.suiteName(), function() {
         public: false,
         resource: '/api/test/v1/require-some-scopes',
         satisfyingScopes: ['aa', 'bb', 'dd'],
+        sourceIp: '::ffff:127.0.0.1',
```
owlishDeveloper: Maybe save that ip into a var and just use the var everywhere? #nitpicking

PR #643, opened 2019-04-27
[UI] Remove search pattern for clients viewer

### LINKED ISSUE #513: [UI] remove search pattern for clients viewer

projectyang:
Fixes #513 

- Removed loading userId as default search query for logged in users

- Added in search query param in URL

- Logging in and out will not reset search results now

Demo:
if1TWmKz6J

### DIFF
--- a/ui/src/views/Clients/ViewClients/index.jsx
+++ b/ui/src/views/Clients/ViewClients/index.jsx
@@ -40,16 +40,6 @@ import ErrorPanel from '../../../components/ErrorPanel';
   },
 }))
 export default class ViewClients extends PureComponent {
-  componentDidMount() {
-    if (this.props.history.location.search) {
-      this.setState({
-        clientSearch: decodeURIComponent(
-          parse(this.props.history.location.search)['?search']
-        ),
-      });
-    }
-  }
-
   state = {
     clientSearch: '',
     // This needs to be initially undefined in order for defaultValue to work
--- a/ui/src/views/Clients/ViewClients/index.jsx
+++ b/ui/src/views/Clients/ViewClients/index.jsx
@@ -22,9 +22,7 @@ import ErrorPanel from '../../../components/ErrorPanel';
       clientOptions: {
         ...(props.history.location.search
           ? {
-              prefix: decodeURIComponent(
-                parse(props.history.location.search.slice(1)).search
-              ),
+              prefix: parse(props.history.location.search.slice(1)).search,
             }
           : null),
       },
@@ -40,21 +38,13 @@ import ErrorPanel from '../../../components/ErrorPanel';
   },
 }))
 export default class ViewClients extends PureComponent {
-  state = {
-    search: '',
-  };
-
   handleClientSearchSubmit = async search => {
     const {
       data: { refetch },
     } = this.props;
     const searchUri = this.props.history.location.search
-      ? decodeURIComponent(
-          parse(this.props.history.location.search.slice(1)).search
-        )
+      ? parse(this.props.history.location.search.slice(1)).search
       : '';
-
-    this.setState({ search });
 
     await refetch({
       clientOptions: {
@@ -79,6 +69,7 @@ export default class ViewClients extends PureComponent {
   handlePageChange = ({ cursor, previousCursor }) => {
     const {
       data: { fetchMore },
+      history,
     } = this.props;
 
     return fetchMore({
@@ -89,10 +80,10 @@ export default class ViewClients extends PureComponent {
           cursor,
           previousCursor,
         },
-        ...(this.state.search
+        ...(history.location.search
           ? {
               clientOptions: {
-                prefix: this.state.search,
+                prefix: parse(history.location.search.slice(1)).search,
               },
             }
           : null),
--- a/ui/src/views/Clients/ViewClients/index.jsx
+++ b/ui/src/views/Clients/ViewClients/index.jsx
@@ -66,6 +66,8 @@ export default class ViewClients extends PureComponent {
     return null;
   }
 
+  //test
+
   handleClientSearchSubmit = clientSearch => {
     const {
       data: { refetch },
--- a/ui/src/views/Clients/ViewClients/index.jsx
+++ b/ui/src/views/Clients/ViewClients/index.jsx
@@ -113,7 +113,7 @@ export default class ViewClients extends PureComponent {
       description,
       data: { loading, error, clients },
     } = this.props;
-    const { clientSearch, value } = this.state;
+    const { value } = this.state;
 
     return (
       <Dashboard
@@ -124,7 +124,6 @@ export default class ViewClients extends PureComponent {
             disabled={loading}
             onSubmit={this.handleClientSearchSubmit}
             onChange={this.handleClientSearchChange}
-            defaultValue={clientSearch}
             value={value}
             placeholder="Client starts with"
           />
--- a/ui/src/views/Clients/ViewClients/index.jsx
+++ b/ui/src/views/Clients/ViewClients/index.jsx
@@ -12,11 +12,9 @@ import Button from '../../../components/Button';
 import ClientsTable from '../../../components/ClientsTable';
 import { VIEW_CLIENTS_PAGE_SIZE } from '../../../utils/constants';
 import clientsQuery from './clients.graphql';
-import { withAuth } from '../../../utils/Auth';
 import ErrorPanel from '../../../components/ErrorPanel';
 
 @hot(module)
-@withAuth
 @graphql(clientsQuery, {
   options: props => ({
     variables: {
@@ -43,6 +41,14 @@ export default class ViewClients extends PureComponent {
     value: null,
   };
 
+  componentDidMount() {
+    if (this.props.history.location.search) {
+      this.setState({
+        clientSearch: this.props.history.location.search.substring(8),
+      });
+    }
+  }
+
   handleClientSearchSubmit = clientSearch => {
     const {
       data: { refetch },
@@ -59,9 +65,12 @@ export default class ViewClients extends PureComponent {
       },
     });
 
-    this.props.history.push(
-      clientSearch.length > 0 ? `?search=${clientSearch}` : '/auth/clients'
-    );
+    // Prevent searching for the same query as the one already loaded
+    if (clientSearch !== this.props.history.location.search.substring(8)) {
+      this.props.history.push(
+        clientSearch.length > 0 ? `?search=${clientSearch}` : '/auth/clients'
+      );
+    }
   };
 
   handleCreate = () => {
@@ -113,7 +122,7 @@ export default class ViewClients extends PureComponent {
       description,
       data: { loading, error, clients },
     } = this.props;
-    const { value } = this.state;
+    const { clientSearch, value } = this.state;
 
     return (
       <Dashboard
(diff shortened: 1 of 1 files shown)


ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -37,33 +37,16 @@ import ErrorPanel from '../../../components/ErrorPanel';
...
-
-    if (!props.user && state.previousClientId !== '') {
-      return {
-        clientSearch: '',
-        previousClientId: '',
-      };
+  componentDidMount() {
+    if (this.props.history.location.search) {
+      this.setState({
+        clientSearch: this.props.history.location.search.substring(8),
```
helfi92: This could probably be defined up top:

```jsx
state = {
  clientSearch: this.props.history.location.search ? ... : '',
  // ...
};
```

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -12,16 +12,16 @@ import Button from '../../../components/Button';
 import ClientsTable from '../../../components/ClientsTable';
 import { VIEW_CLIENTS_PAGE_SIZE } from '../../../utils/constants';
 import clientsQuery from './clients.graphql';
-import { withAuth } from '../../../utils/Auth';
 import ErrorPanel from '../../../components/ErrorPanel';
 
 @hot(module)
-@withAuth
 @graphql(clientsQuery, {
```
helfi92: Instead of using the `substring` method, let's make use of the `qs` package.

```jsx
import { parse, stringify } from 'qs';
```

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -37,33 +37,16 @@ import ErrorPanel from '../../../components/ErrorPanel';
 export default class ViewClients extends PureComponent {
   state = {
     clientSearch: '',
-    // eslint-disable-next-line react/no-unused-state
-    previousClientId: '',
     // This needs to be initially null in order for the defaultValue to work
     value: null,
```
helfi92: I'm seeing the following error in the console:

Screen Shot 2019-04-29 at 8 52 01 AM


To remove this error, I think we can do `value: undefined` instead. Let's also update the comment above it to say undefined instead of null :)

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -81,6 +64,13 @@ export default class ViewClients extends PureComponent {
         limit: VIEW_CLIENTS_PAGE_SIZE,
       },
     });
+
+    // Prevent searching for the same query as the one already loaded
+    if (clientSearch !== this.props.history.location.search.substring(8)) {
+      this.props.history.push(
+        clientSearch.length > 0 ? `?search=${clientSearch}` : '/auth/clients'
```
helfi92: Let's have the client search term encoded via `encodeURIComponent` before pushing it to the url. To decode, you can use `decodeURIComponent`.

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -37,50 +42,36 @@ import ErrorPanel from '../../../components/ErrorPanel';
...
       data: { refetch },
     } = this.props;
+    const searchUri = this.props.history.location.search
+      ? decodeURIComponent(parse(this.props.history.location.search)['?search'])
+      : '';
 
     this.setState({ clientSearch });
 
-    refetch({
+    await refetch({
```
projectyang: Added in await to prevent scenario where user freezes client by refetching data that's already been fetched:

e.g.
1. Search for 'email'
2. Search another query with incorrect string e.g. 'sadsa asdas' -> without await, code will treat this as valid query and change the url query parameter
3. Search for 'email' again -> now possible even though we initially prevent searching for the same data twice
4. Client freezes

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -146,7 +134,6 @@ export default class ViewClients extends PureComponent {
             disabled={loading}
             onSubmit={this.handleClientSearchSubmit}
             onChange={this.handleClientSearchChange}
-            defaultValue={searchDefaultValue}
             value={value}
```
projectyang: Removed defaultValue along with componentDidMount(). These were initially set to display a filled in searchbar if user reloads web page (e.g. http://localhost:5080/auth/clients?search=email%2Fbastien will have email/bastien in the searchbar).

However, this will be misleading since handleClientSearchSubmit does not take into account defaultValue and will assume empty string.

ui/src/views/Clients/ViewClients/index.jsx
helfi92: > Search another query with incorrect string e.g. 'sadsa asdas' -> without await, code will treat this as valid query and change the url query parameter

That's okay. I think it should be safe to remove the `await` statement.

> Client freezes

Interesting. It seems to happen whether we `await` or not.  I can try to investigate. Any ideas?

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -12,16 +13,20 @@ import Button from '../../../components/Button';
...
-@withAuth
 @graphql(clientsQuery, {
   options: props => ({
     variables: {
       clientOptions: {
-        ...(props.user ? { prefix: props.user.credentials.clientId } : null),
+        ...(props.history.location.search
+          ? {
+              prefix: decodeURIComponent(
+                parse(props.history.location.search)['?search']
```
helfi92: I think something like the following would also work:

https://github.com/taskcluster/taskcluster/blob/0d6377a32c874be65929b36149d13eaed8dfd5d1/ui/src/views/PulseMessages/index.jsx#L38

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -37,50 +42,36 @@ import ErrorPanel from '../../../components/ErrorPanel';
 export default class ViewClients extends PureComponent {
   state = {
     clientSearch: '',
-    // eslint-disable-next-line react/no-unused-state
-    previousClientId: '',
-    // This needs to be initially null in order for the defaultValue to work
-    value: null,
+    // This needs to be initially undefined in order for defaultValue to work
+    value: undefined,
```
helfi92: Now that we don't use `value`, we should be able to remove this completely.

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -146,7 +134,6 @@ export default class ViewClients extends PureComponent {
             disabled={loading}
             onSubmit={this.handleClientSearchSubmit}
             onChange={this.handleClientSearchChange}
```
helfi92: `handleClientSearchChange` can be removed.

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -37,50 +42,36 @@ import ErrorPanel from '../../../components/ErrorPanel';
...
       },
       clientsConnection: {
         limit: VIEW_CLIENTS_PAGE_SIZE,
       },
     });
+
+    if (clientSearch !== searchUri) {
+      this.props.history.push(
+        clientSearch.length > 0
+          ? `?search=${encodeURIComponent(clientSearch)}`
```
helfi92: You can probably use `stringify` instead. Something like

https://github.com/taskcluster/taskcluster/blob/0d6377a32c874be65929b36149d13eaed8dfd5d1/ui/src/views/PulseMessages/index.jsx#L156

ui/src/views/Clients/ViewClients/index.jsx
projectyang: > Interesting. It seems to happen whether we await or not. I can try to investigate. Any ideas?

I believe what's happening is loading is stuck on true. Here's a relevant issue: https://github.com/apollographql/apollo-client/issues/1186

A proposed solution is adding notifyOnNetworkStatusChange: true, which I've tested by adding it to the graphql object at the start. Still doesn't seem to fix it.

Await on the refetch, however, does seem to prevent this. Shouldn't we keep it in?

Just to confirm - how I'm reproducing the issue:

1. Visit http://localhost:5080/auth/clients
2. Search with improper string (e.g. 'sadas sads')
3. Search again with empty string '',

helfi92: If awaiting fixes it, then let's keep it :)

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -36,51 +41,35 @@ import ErrorPanel from '../../../components/ErrorPanel';
 }))
 export default class ViewClients extends PureComponent {
   state = {
-    clientSearch: '',
-    // eslint-disable-next-line react/no-unused-state
-    previousClientId: '',
-    // This needs to be initially null in order for the defaultValue to work
-    value: null,
+    search: '',
```
helfi92: Now that the search value is embedded in the url, it probably doesn't make sense to keep the search value in the state variable.

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -36,51 +41,35 @@ import ErrorPanel from '../../../components/ErrorPanel';
...
       },
       clientsConnection: {
         limit: VIEW_CLIENTS_PAGE_SIZE,
       },
     });
+
+    if (search !== searchUri) {
+      this.props.history.push(
+        search.length > 0 ? `?${stringify({ search })}` : '/auth/clients'
+      );
```
helfi92: Would it be possible to change the url prior to `refetch`ing?

ui/src/views/Clients/ViewClients/index.jsx
```diff
@@ -36,51 +41,35 @@ import ErrorPanel from '../../../components/ErrorPanel';
...
-  }
-
-  handleClientSearchSubmit = clientSearch => {
+  handleClientSearchSubmit = async search => {
     const {
       data: { refetch },
     } = this.props;
+    const searchUri = this.props.history.location.search
+      ? decodeURIComponent(
+          parse(this.props.history.location.search.slice(1)).search
```
helfi92: Do we need `decodeURIComponent` if we're already using `parse`?

ui/src/views/Clients/ViewClients/index.jsx
projectyang: Unfortunately, changing the URL prior to awaiting the refetch will cause loading to be stuck again.

helfi92:
Thanks Jason :-)

helfi92:
@allcontributors[bot] please add @projectyang for code!

helfi92:
@projectyang Do you have any interest in tackling https://github.com/taskcluster/taskcluster/issues/705? The task involves upgrading a package from the alpha to a stable version.

projectyang:
@helfi92 I'd love to help - but I'm leaving on vacation for two weeks tomorrow. :)

helfi92:
No worries :) Enjoy your vacation!

PR #651, opened 2019-04-29
universal schema viewer

### LINKED ISSUE #459: [UI] provide a universal schema viewer
Sometimes I just want to link to a particular schema or sub-schema, rather than finding a documentation page that includes that schema.

Let's add a "Schema Index" to the docs page, that lists all schemas in the references by their `$id` and displays the selected schema.

arshadkazmi42:
Fixes #459


arshadkazmi42:
it looks something like this

Screen Shot 2019-04-29 at 11 44 50 PM

arshadkazmi42:
There is one problem, some of the schema does not have data, so schema table on expanding just shows a loader

ui/src/views/Documentation/Reference/SchemaPanel.jsx
```diff
@@ -0,0 +1,144 @@
...
+
+    this.setState({
+      expanded: !expanded,
+    });
+  };
+
+  render() {
+    const { classes, entry } = this.props;
+    const { expanded } = this.state;
+    const entryHashKey = entry.content.$id ? '$id' : '$schema';
```
djmitche: Are there schemas that don't have `$id`?  There *really* shouldn't be :)

ui/src/views/Documentation/Reference/SchemaPanel.jsx
```diff
@@ -0,0 +1,144 @@
...
+      <ExpansionPanel
+        defaultExpanded={expanded}
+        onChange={this.handlePanelChange(entryHashKey)}
+        CollapseProps={{ unmountOnExit: true }}>
+        <ExpansionPanelSummary
+          classes={{
+            content: classNames(classes.expansionPanelSummary),
+          }}
+          expandIcon={<ExpandMoreIcon />}>
+          {this.renderExchangeExpansionPanelSummary()}
```
djmitche: This page isn't about exchanges.  I suspect these function names were just copy/pasted?  Just `renderExpansionPanelSummary` and `renderExpansionDetails` should be ok.

ui/src/views/Documentation/Reference/Schema.jsx
```diff
@@ -0,0 +1,38 @@
...
+    apiVersion: string.isRequired,
+  };
+
+  render() {
+    const { serviceName } = this.props;
+
+    return (
+      <div>
+        <HeaderWithAnchor>Universal Schema Viewer</HeaderWithAnchor>
+        {references && Boolean(references.length) && (
```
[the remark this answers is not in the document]
arshadkazmi42: If the file does not contain any data, I was not sure how data is loaded in the file, so I have added this check.
Will the file always have the data? If so, then we can remove this check

ui/src/views/Documentation/Reference/SchemaPanel.jsx
arshadkazmi42: Yes, there are a couple of schemas which does not have `$id`

`Exchanges to manage hooks` this is one of the example

ui/src/views/Documentation/Reference/SchemaPanel.jsx
arshadkazmi42: Oh, yes. I was referring the exchange component for reference and copied a couple of functions from there.

Updating this.

arshadkazmi42:
> Can you give an example?

`Task Index API Documentation`
`Exchanges to manage hooks`

These two schemas do not have `$id`  and have that issue also, on expanding its just showing the loader

ui/src/views/Documentation/Reference/Schema.jsx
```diff
@@ -0,0 +1,38 @@
...
+          <Fragment>
+            {references.map(entry => (
```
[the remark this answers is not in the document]
arshadkazmi42: Some schemas do not have `$id`

ui/src/views/Documentation/Reference/Schema.jsx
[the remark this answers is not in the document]
djmitche: Yes, it will always exist.

ui/src/views/Documentation/Reference/SchemaPanel.jsx
djmitche: Ah, that's not a schema -- it's a reference document.  Sorry, I forgot the format of `references.json` was confusing.  It's a list of objects of the form `{filename, content}`, and contains both schemas and references.  The schemas can be identified by having filenames beginning with `schemas/`.  For this view, the references can be ignored.

ui/docs/reference/schema/index.md
```diff
@@ -0,0 +1,9 @@
+---
+order: 1
+inline: true
+title: schema
+---
+
+import Schema from 'taskcluster-ui/views/Documentation/Reference/Schema'
+
+<Schema serviceName="schema" apiVersion="v1" />
```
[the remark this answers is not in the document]
djmitche: Also, i don't think it needs these two props.  At any rate, there's no service named "schema"..

ui/src/views/Documentation/Reference/SchemaPanel.jsx
arshadkazmi42: So, if something does not have `$id`, we will ignore those objects from the view?

ui/src/views/Documentation/Reference/SchemaPanel.jsx
```diff
@@ -0,0 +1,144 @@
...
+    return (
+      <ListItem>
+        <ListItemText
+          primaryTypographyProps={primaryTypographyProps}
+          disableTypography
+          primary={<Typography variant="body1">{headerTitle}</Typography>}
+          secondary={
+            <Fragment>
+              <br />
+              <SchemaTable schema={schema} serviceName={serviceName} />
```
djmitche: You can just pass `serviceName="unknown"` here.  I need to fix the SchemaTable so it doesn't take this prop anymore (#657).

ui/src/views/Documentation/Reference/SchemaPanel.jsx
djmitche: Anything that has a filename beginning with `schemas/` will definitely have an `$id`.  There's another data format that I was thinking of that looks like `{references: [..], schemas: [..]}`, and everything under the `schemas` property there is a schema and thus has an `$id`.  That's why I didn't mention this earlier.  Sorry about that!

ui/src/views/Documentation/Reference/SchemaIndex.jsx
```diff
@@ -0,0 +1,48 @@
...
+import filter from 'lodash/filter';
+import SchemaPanel from './SchemaPanel';
+import HeaderWithAnchor from '../components/HeaderWithAnchor';
+import references from '../../../../../generated/references.json';
+
+@withRouter
+export default class SchemaIndex extends Component {
+  static propTypes = {
+    // the service name to document
+    serviceName: string.isRequired,
```
djmitche: This prop isn't passed in `ui/docs/reference/schema/index.md`.. I think it can be omitted.

ui/src/views/Documentation/Reference/SchemaIndex.jsx
```diff
@@ -0,0 +1,48 @@
...
+  };
+
+  // Filters only schemas, sorted by $id
+  getSortedReferences = () => {
+    const sortedReferences = filter(references, entry =>
+      entry.filename.startsWith('schemas/')
+    );
+
+    return sortBy(sortedReferences, ['content.$id']);
+  };
```
djmitche: :100:

djmitche:
Looking good to me anyway -- @helfi92 will need to take a peek, too.

ui/src/views/Documentation/Reference/SchemaPanel.jsx
```diff
@@ -0,0 +1,136 @@
...
+          expandIcon={<ExpandMoreIcon />}>
+          {this.renderExpansionPanelSummary()}
+        </ExpansionPanelSummary>
+        <ExpansionPanelDetails>
+          {this.renderExpansionDetails()}
+        </ExpansionPanelDetails>
+      </ExpansionPanel>
+    );
+  }
+}
```
helfi92: This file seems really similar to `src/views/Documentation/Reference/Entry.js`. Is there a way to make use of that component instead of this file?

arshadkazmi42: I think we can pass a type flag to Entry, something like `schema` and use SchemaPanel functions for rendering the SchemaPanel related components.

We might have to do some changes in this logic
```
    const isEntryExchange = type === 'topic-exchange';
    const isLogType = type === 'logs';
    const isFunctionType = !isLogType && !isEntryExchange;
    const entryHashKey = isLogType ? 'type' : 'name';
```

arshadkazmi42: Should I change this, to use Entry Component and Remove SchemaPanel component?

helfi92: Yes, please :)

arshadkazmi42:
> Looking better. A couple of changes:
> 
>     * When a user clicks on a panel (e.g., Client Message), the URL currently changes to `http://localhost:5080/docs/reference/schema#/schemas/auth/v1/client-message.json#`. Let's encode the hash so that it becomes `%2Fschemas%2Fauth%2Fv1%2Fclient-message.json%23`.
> 
>     * The reason why we add a hash to the url when a user clicks on a panel is so that they can share it with other people. Clicking on a panel then refreshing the page should have that panel open by default since it's embedded in the url hash.
> 
>     * Rename "Schema" to "Schema Index" in the navigation bar
> 
> 
> Screen Shot 2019-05-01 at 10 24 03 AM

`http://localhost:8081/docs/reference/schema#%2Fschemas%2Fauth%2Fv1%2Fazure-container-response.json%23`

I have added encoding, now url fromats to this, but still on refereshing it does not expand the panel by default, is there something I need to handle for default panel to expand automatically.

arshadkazmi42:
> > Looking better. A couple of changes:
> > ```
> > * When a user clicks on a panel (e.g., Client Message), the URL currently changes to `http://localhost:5080/docs/reference/schema#/schemas/auth/v1/client-message.json#`. Let's encode the hash so that it becomes `%2Fschemas%2Fauth%2Fv1%2Fclient-message.json%23`.
> > 
> > * The reason why we add a hash to the url when a user clicks on a panel is so that they can share it with other people. Clicking on a panel then refreshing the page should have that panel open by default since it's embedded in the url hash.
> > 
> > * Rename "Schema" to "Schema Index" in the navigation bar
> > ```
> > 
> > 
> > Screen Shot 2019-05-01 at 10 24 03 AM
> 
> `http://localhost:8081/docs/reference/schema#%2Fschemas%2Fauth%2Fv1%2Fazure-container-response.json%23`
> 
> I have added encoding, now url fromats to this, but still on refereshing it does not expand the panel by default, is there something I need to handle for default panel to expand automatically.

Found the issue, there was an issue at my end.

ui/src/views/Documentation/Reference/SchemaIndex.jsx
```diff
@@ -0,0 +1,39 @@
+import React, { Component, Fragment } from 'react';
+import { withRouter } from 'react-router-dom';
+import sortBy from 'lodash/sortBy';
+import filter from 'lodash/filter';
```
helfi92: Let's switch to using ramda, which is what we use in the ui directory. You can probably find the sort and filter methods in https://ramdajs.com/docs/.

ui/src/views/Documentation/Reference/SchemaIndex.jsx
```diff
@@ -0,0 +1,39 @@
...
+import Entry from './Entry';
+import HeaderWithAnchor from '../components/HeaderWithAnchor';
+import references from '../../../../../generated/references.json';
+
+@withRouter
```
helfi92: Is this needed?

ui/src/views/Documentation/Reference/SchemaIndex.jsx
```diff
@@ -0,0 +1,39 @@
...
+    const sortedReferences = this.getSortedReferences();
+
+    return (
+      <div>
+        <HeaderWithAnchor>Schema Index</HeaderWithAnchor>
+        <Fragment>
+          {sortedReferences.map(entry => (
+            <Entry
+              key={entry.content.$id}
+              serviceName="unknown"
```
helfi92: Instead of passing `unknown`, we could make `serviceName` not required. You should be able to do that in `static PropTypes`.

ui/src/views/Documentation/Reference/Entry.jsx
```diff
@@ -498,28 +522,53 @@ export default class Entry extends Component {
...
-      history.push(`#${entry[key]}`);
+      history.push(`#${encodeURIComponent(get(entry, key))}`);
     }
 
     this.setState({
       expanded: !expanded,
     });
   };
 
+  getEntryHashKey = (isLogType, isEntrySchema) => {
```
helfi92: We can probably rewrite this method to be:

```jsx
getEntryHashKey = type => {
  switch(type) {
      case '...' : {
        return // ...
      }
      // ...
  }
};
```

ui/src/views/Documentation/Reference/SchemaIndex.jsx
djmitche: For the moment, I want this to show up if it's used to generate a URL (it shouldn't be..).  There's a bug to fix this that's assigned to me.

ui/src/views/Documentation/Reference/SchemaIndex.jsx
arshadkazmi42: No, I don't this its getting used, removing it.

ui/src/views/Documentation/Reference/SchemaIndex.jsx
arshadkazmi42: @helfi92 @djmitche So I should keep it as `unknown` ?

djmitche: Yes -- I'll change it in #657.

ui/src/views/Documentation/Reference/Entry.jsx
```diff
@@ -101,7 +102,9 @@ export default class Entry extends Component {
   state = {
     expanded:
       this.props.entry.name === window.location.hash.slice(1) ||
-      this.props.entry.type === window.location.hash.slice(1),
+      this.props.entry.type === window.location.hash.slice(1) ||
+      encodeURIComponent(get(this.props.entry, 'content.$id')) ===
```
helfi92: `get` is a lodash method. Let's either use `ramda` or plain javascript.

arshadkazmi42: Plain javascript is crashing for `api` and `exchange` pages, so we would need something which can handle exceptions, I will change it to `ramda`

arshadkazmi42: @helfi92 I can't find `get` in `ramda` docs, is it names something else? I went through the document, couldn't find any related to get

helfi92: It's not necessarily named `get`. Perhaps https://ramdajs.com/docs/#prop?

arshadkazmi42:
> Clicking on a panel properly sets up the url but refreshing the page doesn't properly scroll to the container. I can't see the arrow.
> 
> Actual:
> 
> Screen Shot 2019-05-02 at 2 31 57 PM
> 
> Expected (taken from the api references page):
> 
> Screen Shot 2019-05-02 at 2 33 53 PM
> 
> In the second screenshot, I am able to see the start of the panel. I think it's probably because the `scrollToHash` function is not able to find `document.getElementById(an encoded hash)`. 

Yes. right,  I think I missed it while testing in local. 

> To fix this, I think you would need to encode the `id` attribute of the element.

Encoding `id` fixed the issue

arshadkazmi42:
> Is the following expected?
> 
> Screen Shot 2019-05-02 at 3 36 36 PM

I don't think this is expected.
I did some digging on this and found `references.json` file these two specific schema has below values

```
"allOf": [
        {
          "$ref": "./metaschema.json#"
        },
        {
          "$ref": "#/definitions/topLevelMetadata"
        }
      ],
```

and for this reference is going in and executing this code

`https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/SchemaTable/index.jsx#L141-L162`

and its crashing at 

`https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/SchemaTable/index.jsx#L152`

I am exactly sure, how to process this, `allOf` key.

helfi92:
I'll create a separate issue for this. Let's merge this :)

arshadkazmi42:
Some tests are failing. I am checking it in my local, what exactly is breaking

helfi92:
I'm rerunning the tests to see if they pass.

arshadkazmi42:
Everything is passing now.

helfi92:
Awesome work!!!

helfi92:
Filed https://github.com/taskcluster/taskcluster/issues/689.

Issue #654, opened 2019-04-29
[UI] Resetting the store causes issues in the UI


helfi92:
STR:

1. Navigate to https://taskcluster-web.netlify.com/tasks/groups/e_7vMlcZTcys5OQ03EPFSQ while being logged out
2. Login

Expected: 

<img width="1336" alt="Screen Shot 2019-04-29 at 6 53 40 PM" src="https://user-images.githubusercontent.com/3766511/56932030-263f9400-6ab0-11e9-9a51-a2305dd5dc70.png">

Actual: The task status cards shows a loading symbol that never goes away 

<img width="1336" alt="Screen Shot 2019-04-29 at 6 53 03 PM" src="https://user-images.githubusercontent.com/3766511/56932011-12942d80-6ab0-11e9-94ed-c3b393c302b4.png">

The culprit: https://github.com/taskcluster/taskcluster/blob/771c78c6d6606591e51c1576e4555defc31894e9/ui/src/components/SignInDialog/index.jsx#L70

This line causes the store to be cleared and all active queries to be refetched. The only problem is that for many of our views, we depend on `fetchMore` to fetch the full list. Refetching active queries will only fetch the first query.

To fix this issue, we should rely on the `key` prop and the `networkPolicy` property to refetch active queries. Changing the `key` prop of a component in React will re-render that component. This will have a similar effect as `resetStore`.

djmitche:
In "Learning GraphQL", I noticed that the process they use is `cache.readQuery` -> modify value -> `cache.writeQuery`.  Would that help in this case?

helfi92:
> In "Learning GraphQL", I noticed that the process they use is cache.readQuery -> modify value -> cache.writeQuery. Would that help in this case?

We want to re-fetch the full query when a user logs in/out to make sure the state reflects the current user's permissions. `readQuery` and `writeQuery` only interacts with the cache unfortunately.

PR #655, opened 2019-04-29
[UI] Rely on key and fetchPolicy instead of resetStore

helfi92:
`resetStore` causes the store to be cleared and all active queries to be refetched. The only problem is that for many of our views, we depend on`fetchMore` to fetch the full list. Refetching active queries will only fetch the first query. There would probably be other problems with some of our subscriptions.

To fix this issue, we should rely on the key prop and the `networkPolicy` property to refetch active queries. Changing the `key` prop of a component in React will re-render that component. This will have a similar effect as `resetStore` & `clearStore`.

Fixes https://github.com/taskcluster/taskcluster/issues/654.

### DIFF
--- a/ui/src/components/UserMenu/index.jsx
+++ b/ui/src/components/UserMenu/index.jsx
@@ -57,9 +57,6 @@ export default class UserMenu extends Component {
   handleSignOutClick = () => {
     this.handleMenuClose();
     this.props.onUnauthorize();
-    // Since Apollo caches query results, it’s important to get rid of them
-    // when the login state changes.
-    this.props.client.clearStore();
   };
 
   handleMenuClick = e => {
--- a/ui/src/views/Profile/index.jsx
+++ b/ui/src/views/Profile/index.jsx
@@ -14,7 +14,11 @@ import profileQuery from './profile.graphql';
 
 @hot(module)
 @withAuth
-@graphql(profileQuery)
+@graphql(profileQuery, {
+  options: () => ({
+    fetchPolicy: 'network-only',
+  }),
+})
 export default class Profile extends Component {
   render() {
     const {
--- a/ui/README.md
+++ b/ui/README.md
@@ -113,6 +113,12 @@ import BookOpenPageVariantIcon from 'mdi-react/BookOpenPageVariantIcon';
 
 <sup>* We use this library because it provides substantially more icons with minimal file-system headaches.</sup>
 
+## Necessary Practices
+
+1. For views showing secret, user-linked data or are using `client.fetchMore` to load more data
+on page load, graphql queries of that view must include the `fetchPolicy: 'network-only'` option to ensure
+data reflects the user's permissions and is up-to-date when a user logs in/out.
+
 ## Table of Contents
 
 <!-- TOC BEGIN -->
--- a/ui/src/App/index.jsx
+++ b/ui/src/App/index.jsx
@@ -217,7 +217,7 @@ export default class App extends Component {
             <MuiThemeProvider theme={theme}>
               <FontStager />
               <CssBaseline />
-              <Main routes={routes} error={error} />
+              <Main key={auth.user} routes={routes} error={error} />
             </MuiThemeProvider>
           </ToggleThemeContext.Provider>
         </AuthContext.Provider>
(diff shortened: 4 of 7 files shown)


ui/src/components/UserMenu/index.jsx
```diff
@@ -57,9 +57,6 @@ export default class UserMenu extends Component {
   handleSignOutClick = () => {
     this.handleMenuClose();
     this.props.onUnauthorize();
-    // Since Apollo caches query results, it’s important to get rid of them
-    // when the login state changes.
-    this.props.client.clearStore();
```
owlishDeveloper: Would calling `this.props.client.resetStore()` here still be a good idea?

helfi92: Re-fetching the active queries won't work for many of our views where we purposely avoid pagination by fetching all the data via (`fetchMore`) on page load. `this.props.client.resetStore()` would only re-fetch the first page for paginated endpoints.

ui/src/views/Profile/index.jsx
```diff
@@ -14,7 +14,11 @@ import profileQuery from './profile.graphql';
 
 @hot(module)
 @withAuth
-@graphql(profileQuery)
+@graphql(profileQuery, {
+  options: () => ({
+    fetchPolicy: 'network-only',
+  }),
+})
```
djmitche: Is this a necessary practice, then -- views showing secret or user-linked data must have this fetchPolicy?  Is it worth writing that down somewhere?

ui/src/components/UserMenu/index.jsx
djmitche: So if I understand, the combination of clearing the cache here, re-rendering the `Main` component, and using `network-only` on any secret views means that views will not just have their queries re-run, but will re-render from an initial empty state (so resetting to the first page, for example).  Is that about right?  If we're clearing the cache, why is the `network-only` policy required?

ui/src/views/Profile/index.jsx
helfi92: Yes, It probably makes sense to write to document this. I'll push a commit.

ui/src/components/UserMenu/index.jsx
helfi92: > If we're clearing the cache

We're not clearing the cache. We are overwriting the cache.

> Why is the network-only policy required?

Changing the `key` prop when auth changes re-renders the content without clearing the cache. For certain views we want to make sure the graphql query doesn't use a cached response when re-rendering, we add the `network-only` policy.

djmitche: > We're not clearing the cache. We are overwriting the cache.

I don't understand that distinction..

helfi92: Consider the scenario where a user is on https://taskcluster-web.netlify.com/secrets/garbage%2Farmenzg%2FfakeOrg.

Using `clearStore`:
* User is able to see the secret.
* User logs out.
* `clearStore` will cause the store to be cleared.
* The user still sees the secret.

Using `resetStore`:
* User is able to see the secret.
* User logs out.
* `resetStore` will cause the store to be cleared and all active queries to be refetched.
* User sees an error saying they are missing permissions to view the secret. This will work for the secrets view but will fail for the TaskGroup since all of the `fetchMore` queries won't be triggered therefore showing only the first 20 tasks https://github.com/taskcluster/taskcluster/blob/f283feaacb60b06d47c2a24344e857f9a2c7d217/ui/src/views/Tasks/TaskGroup/index.jsx#L52

Using the `key` prop and the `network-only` policy:
* User is able to see the secret.
* User logs out.
* The `key` prop will change. This will reconstruct the view from scratch (i.e., the constructor will run again; this is different than just re-running the render method). Changing the key will have the same effect as reloading the page but in this case the page is the view component.

* The graphql query of that view will see that it has `network-only` as a policy so it will always make a request to the server. The response will overwrite whatever was in the store for that query.
* User sees an error saying they are missing permissions to view the secret.

djmitche: OK, got it -- so `clearStore` and `key` + `network-only` are two different layers of protection here.

PR #664, opened 2019-04-30
HOTFIX: Do not run strategy if config for strategy is undefined

helfi92:


helfi92:
> Shouldn't the service fail to start if it doesn't have config? With this change, it will just silently fail to load that strategy..

Yes, it should. I'm in the process of adding the required config to heroku. Once that's done, I'll revert this PR.

PR #667, opened 2019-04-30
[web-server] Make web-server run straight from source

helfi92:
https://bugzilla.mozilla.org/show_bug.cgi?id=1525797


services/web-server/test/requireContext_test.js
```diff
@@ -0,0 +1,58 @@
+const assert = require('assert');
+const path = require('path');
+const testing = require('taskcluster-lib-testing');
+const requireContext = require('../src/utils/requireContext');
+
+suite(testing.suiteName(), () => {
```
djmitche: omg omg tests! :grinning:

services/web-server/test/requireContext_test.js
```diff
@@ -0,0 +1,58 @@
...
+    const keys = importer.keys();
+
+    assert(keys.length > 0, 'should of had files');
+  });
+
+  test('should throw an exception if the module does not exists', () => {
+    try {
+      requireContext('./cat');
+
+      assert.fail('should of had an error');
```
djmitche: this won't fail this way -- the AssertionError will get caught :)

Often `assert.throws` or `assert.rejects` are the easiest way to test for an exception.

djmitche:
So cool :)

BTW, you can remove the `Web Server` hook in `infrastructure/builder/src/build/monoimage.js` too.

services/web-server/package.json
```diff
@@ -2,13 +2,12 @@
   "name": "taskcluster-web-server",
   "version": "0.0.0-private",
   "private": true,
-  "main": "build/index.js",
+  "main": "./src/index.js",
```
djmitche: Please rename `main.js` to match other services

package.json
```diff
@@ -112,9 +113,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.2.2",
-    "@babel/plugin-proposal-class-properties": "^7.1.0",
-    "@neutrinojs/mocha": "9.0.0-rc.0",
-    "@neutrinojs/node": "9.0.0-rc.0",
+    "@babel/plugin-proposal-class-properties": "^7.4.4",
```
djmitche: Why do these two stick around?  I tried removing them and `ui` still built properly..

helfi92: `@babel/plugin-proposal-class-properties` is being used by https://github.com/taskcluster/taskcluster/blob/6ae3183757756fd9028538436d21669636c9c82c/clients/client-web/.neutrinorc.js#L12 The latter has `@babel/core` specified as a peer dependency in https://github.com/babel/babel/blob/16e17c560b4d6c818b5ededcb0bbe6d6bf671f56/packages/babel-plugin-proposal-class-properties/package.json#L19.

djmitche: Indeed, I forgot about client-web.  It looks like that could be moved to `clients/client-web/package.json`, but I can do that in another PR.

helfi92:
Rebased.

services/web-server/src/utils/requireContext.js
```diff
@@ -0,0 +1,51 @@
+const fs = require('fs');
+const path = require('path');
+const getCallerFile = require('get-caller-file');
+
+require.extensions['.graphql'] = function (module, filename) {
+  module.exports = fs.readFileSync(filename, 'utf8');
+};
+
+// Mock webpack's implementation of require.context
```
owlishDeveloper: Interesting. I was under the impression that this functionality is redundant on the backend, so there's not need to implement it?

helfi92: Without `require.context`, we still need to write something to loop over files in a directory and have them `require`d. `requireContext` does that. I went ahead and made the API look just like webpack's. Another way of writing this utility is to rewrite this function to return the full path of all files under a directory and have the file using this utility responsible of requiring the file. I think both implementations are fine except this one has the added bonus of being able to require the file directly.

owlishDeveloper: Is it possible to create an `index.js` file for the dir, import all the files from the dir there and then import that file instead of writing loops? (at least that is how I've it's usually done)

helfi92: > Is it possible to create an index.js file for the dir, import all the files from the dir there and then import that file

The 3 areas where we use `requireContext` are in `src/loader/index.js`, `src/resolvers/index.js` and `src/graphql/index.js`.

> instead of writing loops?

I'm not sure how to require the files without a loop. What did you have in mind?

owlishDeveloper: > I'm not sure how to require the files without a loop. Did you have something in mind?

*some file somewhere*

```
const loaders = require("./loaders");
```

*loaders/index.js*

```
exports.awesomeLoader = require("./loaders/awesomeLoader.js");
exports.fabulousLoader = require("./loaders/fabulousLoader.js");
exports.amazingLoader = require etc.
```

djmitche: https://github.com/taskcluster/taskcluster/blob/ef46fee21fa8d56ac990378534b53bee56f0ab9b/libraries/api/src/middleware/index.js#L5-L15
This is an example in our codebase.  I don't mind the re-implementation (with tests!) of `importer` here, although this is a pretty simple use-case for that function. It would be nice to have one way we do this throughout the codebase.

djmitche: Maybe this is OK as-is and fodder for a later clean-up?

helfi92:
Rebased.

services/web-server/src/utils/requireContext.js
owlishDeveloper: I wasn't arguing with Hassan here, I was trying to understand our use case. I see the `requireContext` to be rather hard to read and understand (and maybe it's just me) comparing to a series of `require`. I just wanted to understand the reason for using something more complex.

owlishDeveloper: I can see how questions like "why are we doing this instead of that" can come off as criticizing. It's just I am more focused on gaining expertise as a developer and learning from the team rather than criticizing you people :) I am also imagining I should become more productive if I understand our code base better

The current code is absolutely fine with me as long as it works, and especially if there's no explanation for me, we can just end this conversation right here :)

helfi92: Having to manually edit 3 `index.js` whenever we add/remove/rename a file in the `loader` directory for example is a cumbersome. We would usually need to modify 3 `index.js` files because adding a new file for example `loaders/test.js` also entails adding `resolvers/Test.js` and `graphql/Test.graphql`. Minimizing the mental overload of having to glue things together is what I was going for. I hope that clarifies the intent. Also, if you feel the `requireContext` is a bad name, I'm more than happy to change it to something else. I just took the name webpack uses since it's the same API.

helfi92:
Rebased 😅

services/web-server/src/utils/requireContext.js
owlishDeveloper: > I hope that clarifies the intent
It does, thank you!

> Also, if you feel the requireContext is a bad name, I'm more than happy to change it to something else
I seldom have problems with names :) In fact, it's something I don't usually even look at when doing a PR review

Issue #659, opened 2019-04-30
[web-server] Construct jwks once (rate-limiting)


helfi92:
https://github.com/taskcluster/taskcluster/pull/547#discussion_r279836433

PR #673, opened 2019-05-01
[web-server] Construct jwks once

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/659.


helfi92:
Yeah, it's static but I needed a way to propagate that value to the utility function `verifyJwt`. That's one way :)

Issue #453, opened 2019-03-16
[UI] show git version somewhere out of the way


djmitche:
We have `.git-version` in the root of the repository, when taskcluster is deployed.  We currently use that in logging to indicate which version of Taskcluster we're running.  As we get more and more deployments running, it would be helpful to have that visible somewhere in the UI as well so that users can report the version they're running.  @helfi92 what do you think would be most appropriate way to do this?  Random ideas from me:
 * ``console.log(`Taskcluster version ${gitVersion}`)`` on load
 * display it in a small font at the bottom of the navigation drawer
 * on a new "About Taskcluster" page

Also, note that `.git-version` may not always exist (such as when running locally), so the implementation should be able to handle an undefined value.

helfi92:
Out of curiosity, why do we put the version of Taskcluster in a file named `.git-version` instead of something like `.taskcluster-version`?

Having it below the application name sounds good to me:

![Screen Shot 2019-03-18 at 11 48 57 AM](https://user-images.githubusercontent.com/3766511/54543344-e09d9080-4973-11e9-873f-1f79f47ea691.png)

djmitche:
We could change the filename.  I don't think there's any particular reason for that name.

iFlameing:
@djmitche can you assign me this?

djmitche:
I need to invite you to the org first.  Once you've accepted, let me know and I will assign.

djmitche:
Oops, I don't have my Github password with me.  @helfi92 any chance you could invite @iFlameing?  Otherwise I will in the next day.

helfi92:
Invitation sent.

iFlameing:
@djmitche accepted the invitation.

djmitche:
Thanks hassan :)

iFlameing:
@djmitche where I can see the prototype of .git-version file?

djmitche:
You can see what the running docker image looks like with
```
docker run -ti --rm taskcluster/taskcluster-monoimage:SVC-8c70f90225ae bash
```
and you'll see the file in `/app`.

helfi92:
@allcontributors[bot] please add @iFlameing  for code!

iFlameing:
When I run the above command, I successfully pull the image but when I am running the command in bash i.e(
bash-4.4#  /app .
bash: /app: Is a directory
)

I got the following error. what should I do now?

djmitche:
[...]

PR #674, opened 2019-05-01
Adding Version Number of taskcluster app to docs drawer.

iFlameing:
This is related to #453 in which we have to add the version number of taskcluster app to docs drawer.


ui/.neutrinorc.js
```diff
@@ -153,6 +153,12 @@ module.exports = {
           .test(/\.all-contributorsrc$/)
           .use('json-loader')
             .loader('json-loader');
+
```
helfi92: Remove empty spaces. Keep the newline however.

.git-version
```diff
@@ -0,0 +1 @@
+8026691dab4045166c5a1798f3c7661ee622da4e
```
helfi92: This file shouldn't be commited.

ui/src/components/Dashboard/index.jsx
```diff
@@ -188,8 +197,26 @@ export default class Dashboard extends Component {
...
+    this.loadVersionNumber();
+  }
+
+  async loadVersionNumber() {
+    try {
+      const {
+        default: versionNumber,
+      } = await import(`../../../../.git-version`);
+
+      this.setState({ versionNumber });
```
helfi92: Set the version in `componentDidMount`. `loadVersionNumber` should just return the value.

iFlameing:
@helfi92 please! take a look, I made the changes as you suggested.

iFlameing:
@helfi92 I made all the changes as you suggested above. But we are ignoring the error which will be thrown if we don't have .git-version file in root app directory  which I don't understand why? I also intended to add a class to nav-element but I was little bit afraid in changing code. thanks!

helfi92:
>  But we are ignoring the error which will be thrown if we don't have .git-version file in root app directory which I don't understand why?

Good question. I added a comment to the issue to ask Dustin. Let's wait and see what he says.

iFlameing:
@helfi92 rebased the pr :)

PR #681, opened 2019-05-01
Configure login strategies once, in a JSON object

### LINKED ISSUE #672: [web-server] take login strategies config as a single object
We would like to be able to make deployments with no login strategies at all (since the login strategies tend to require a lot of configuration), but have other deployments with one or more strategies enabled.  Currently, we require that both github and auth0 are enabled, and require configuration for both of them, in the form of a half-dozen terraform variables.

Let's shrink that to one terraform variable, and ideally one that configures both tc-ui and tc-web-server.

djmitche:
Fixes #672 

I couldn't get terraform to extract the keys from the config object, so there are two variables (one for tc-web-server, one for tc-ui).


services/web-server/config.yml
```diff
@@ -37,19 +37,22 @@ defaults:
...
+    #   clientSecret: auth0 accessToken corresponding to that ID
+    #   audience: (to be removed)
+    #   scope: (to be removed)
+    #
+    # github:
+    #   # client id and secret for a Github OAuth2 application -- create a
+    #   # fresh one on the GitHub website.
+    #   clientId: !env GITHUB_CLIENT_ID
+    #   clientSecret: !env GITHUB_CLIENT_SECRET
+    strategies: !env:json UI_LOGIN_STRATEGIES
```
helfi92: @djmitche Will the change throw an error in:

https://github.com/taskcluster/taskcluster/blob/5cc3b9714408917abd9ced34af1c90ee5e880396/services/web-server/src/login/strategies/mozilla-auth0.js#L19

https://github.com/taskcluster/taskcluster/blob/b03837fb0b464ea576a5da22851026dd7a0894c3/services/web-server/src/login/strategies/github.js#L15

djmitche: It shouldn't, since `strategies` will be a JSON object with properties from

https://github.com/taskcluster/taskcluster/blob/b03837fb0b464ea576a5da22851026dd7a0894c3/services/web-server/src/index.js#L151

PR #684, opened 2019-05-02
[UI] TaskGroupTable: Add taskGroupId in memoize key

helfi92:
For hook retriggers, tasks will have the same names possibly causing the memoize function to return old data. Adding `taskGroupId` will make sure we are looking at the right tasks.


djmitche:
There's nothing preventing a single task group from having 100 tasks with the same `metadata.name` property.

helfi92:
Agreed. I must have forgotten to include the `taskGroupId` when implementing the `memoize` function.

Issue #662, opened 2019-04-30
[web-server] log credential generation


djmitche:
https://github.com/taskcluster/taskcluster/pull/547/files#r276752600

helfi92:
I assume you want this to be done via `monitor.info()`.

djmitche:
This partly addresses https://bugzilla.mozilla.org/show_bug.cgi?id=1307271 (but that bug also includes logging for changes to roles and clients' scopes, from the auth service)

PR #688, opened 2019-05-03
[web-server] Log credential generation

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/662.

### DIFF
--- a/services/web-server/src/monitor.js
+++ b/services/web-server/src/monitor.js
@@ -12,7 +12,9 @@ manager.register({
   level: 'info',
   description: 'A client has been issued Taskcluster credentials',
   fields: {
-    clientId: 'The client name.',
+    clientId: 'The clientId of the issued credentials',
+    userIdentity: 'The identity of the user to which the credentials were issued',
+    expires: 'Date time when the issued credentials expires.',
   },
 });
 
--- a/services/web-server/src/monitor.js
+++ b/services/web-server/src/monitor.js
@@ -10,10 +10,9 @@ manager.register({
   type: 'create-credentials',
   version: 1,
   level: 'info',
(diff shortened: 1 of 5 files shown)


services/web-server/src/monitor.js
```diff
@@ -4,4 +4,17 @@ const manager = new MonitorManager({
...
 
+manager.register({
+  name: 'createCredentials',
+  title: 'Credentials Created',
+  type: 'create-credentials',
+  version: 1,
+  level: 'info',
+  description: 'A user has been issued Taskcluster credentials',
+  fields: {
+    credentials: 'Taskcluster credentials.',
```
djmitche: I suspect we don't want to log the full set of credentials -- probably just the clientId.

helfi92:
Rebased.

services/web-server/src/monitor.js
```diff
@@ -4,4 +4,16 @@ const manager = new MonitorManager({
...
+manager.register({
+  name: 'createCredentials',
+  title: 'Credentials Created',
+  type: 'create-credentials',
+  version: 1,
+  level: 'info',
+  description: 'A client has been issued Taskcluster credentials',
+  fields: {
+    clientId: 'The client name.',
+  },
```
djmitche: Sorry for the double-drive-by here.  I should have looked harder last time.  I think this is issuing credentials to a *user* - so let's change this around to "A user has been issued Taskcluster credentials", with two fields:

```js
{
  clientId: 'The clientId of the issued credentials',
  userIdentity: 'The identity of the user to which the credentials were issued',
}
```

I think the latter can come from `user.identity`.

I think the last version had an `expires` property, too?  I can imagine the opsec team would want that in order to understand the timeframe of vulnerability from a compromised credential, for example.

PR #693, opened 2019-05-03
Sort tasks topologically

owlishDeveloper:
Bugzilla Bug: [1548633](https://bugzilla.mozilla.org/show_bug.cgi?id=1548633)


owlishDeveloper:
> Otherwise this is all A+

if there's an otherwise, it's not A+ 😆  You're right - we _make calls_ to the queue in the right order now, but we don't await until every call returned. I'll correct this, thank you!

Issue #677, opened 2019-05-01
[Docs] Write a simple crawler that indexes our doc files


helfi92:
Problem: https://taskcluster-web.netlify.com/docs doesn't have a search bar that can search through our docs.

Solution: Let's write a simple crawler that indexes the .md/.mdx files. We would then be able to use the result of that crawler to add search support to our docs.

arshadkazmi42:
I would like to take this up.
Can you give some more details on the issue. What type of crawler is needed here?

helfi92:
Hey @arshadkazmi42 👋

I have an implementation on my local machine of the crawler. I just need to finalize it and then I should be ready to push it. Once the crawler is merged, we will need to hook it up to the frontend. Maybe you can help here :) I should have assigned the issue to myself earlier. Sorry for the trouble. I'll ping you once we are ready with the second part of this task.

Issue #405, opened 2019-03-13
[UI] docs logs deep linking not working


helfi92:
STR:

1. Navigate to https://taskcluster-web.netlify.com/docs/reference/platform/auth/logs#monitor.count (notice the `#monitor.count` at the end of the url)

**Expected**: The `monitor.count` panel is expanded
**Actual**: The `monitor.count` panel is not expanded

ziyadElon:
Hi @helfi92 
The attached url directs to a 404.
Could you please look into this?

helfi92:
Sorry about that. It should work now.

RehabHussein:
@helfi92 Hi Hassan, I'm new to open source contribution - would like to get started on this. Is it OK?

helfi92:
Go for it. I sent you an invitation via email. Let me know once your accept it so I can assign the issue to you.

RehabHussein:
@helfi92 Perfect ! 
I accepted the invitation. Thank you !

helfi92:
How is it going Rehab? Let me know if you had questions.

helfi92:
I can take care of it this time, np. If you're still interested in another problem, let me know and I can assign it to you.

RehabHussein:
@helfi92 I was just stuck in some issues deploying the app :( 
Sorry for being late ! 
Sure I am still interested in participating in task cluster, will find another suitable issue and ping you :)
Thanks Hassan for your understanding !

PR #699, opened 2019-05-06
[Docs] Add crawler to "yarn generate"

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/677.

Features:
* Uses a headless browser to index the SPA.
* Takes around 90 seconds to index. Much better performance than the 10 minutes we were seeing with Algolia.
* Indexes docs headers

To run locally:

`APIFY_LOG_LEVEL=WARNING APIFY_LOCAL_STORAGE_DIR="generated/apify_storage" yarn generate`.

### DIFF
--- a/infrastructure/builder/src/generate/generators/docs-crawler.js
+++ b/infrastructure/builder/src/generate/generators/docs-crawler.js
@@ -1,0 +1,132 @@
+const Apify = require('apify');
+const { promisify } = require('util');
+const { join } = require('path');
+const { sortBy, prop } = require('ramda');
+const { REPO_ROOT, writeJSON, readdir, readJSON } = require('../util');
+
+const rimraf = promisify(require('rimraf'));
+const GENERATED = 'generated';
+const APIFY_STORAGE = 'apify_storage';
+const TEMP_STORAGE_PATH = join(GENERATED, APIFY_STORAGE, 'datasets', 'default');
+// Ignore links outside the <main> element
+const PAGE_CONTENT_SELECTOR = 'main';
+// Select relative links only
+const LINK_SELECTOR = `${PAGE_CONTENT_SELECTOR} a[href^="/"]`;
+
+async function apifyCrawler({ url }) {
+  // Apify.openRequestQueue() is a factory to get a preconfigured RequestQueue instance.
+  // We add our first request to it - the initial page the crawler will visit.
+  const requestQueue = await Apify.openRequestQueue();
+  await requestQueue.addRequest({ url });
+  // Create an instance of the PuppeteerCrawler class - a crawler
+  // that automatically loads the URLs in headless Chrome / Puppeteer.
+  return new Apify.PuppeteerCrawler({
+    requestQueue,
+
+    // Options that are passed to the Apify.launchPuppeteer() function.
+    launchPuppeteerOptions: { headless: true },
+
+    // The number of Requests the crawler should process
+    // 1000 should be enough...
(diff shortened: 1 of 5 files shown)


infrastructure/builder/src/generate/generators/docs-crawler.js
```diff
@@ -0,0 +1,131 @@
...
+
+function removeApifyStorage() {
+  return rimraf(join(REPO_ROOT, GENERATED, APIFY_STORAGE));
+}
+
+exports.tasks = [{
+  title: 'Crawler that indexes the docs site',
+  provides: ['docs-crawler'],
+  run: async (requirements, utils) => {
+    // Prepare a clean state for the crawler
```
helfi92: Running `APIFY_LOG_LEVEL=WARNING yarn generate` suppresses info log messages but adding `process.env.APIFY_LOG_LEVEL = 'WARNING';` here doesn't seem to work. Any idea why this happens?

infrastructure/builder/src/generate/generators/docs-crawler.js
```diff
@@ -0,0 +1,131 @@
...
+    await removeApifyStorage();
+
+    const crawler = await apifyCrawler({ url: 'https://taskcluster-web.netlify.com/docs' });
```
helfi92: The `url` should probably not be hard-coded. This value should correspond to the domain it's being deployed against. Is there an environment variable already set up that has the domain?

infrastructure/builder/src/generate/generators/docs-crawler.js
helfi92: I'm facing a similar issue with setting up `APIFY_LOCAL_STORAGE_DIR="generated/apify_storage"`. The variable doesn't seem to be picked up with set via `process.env`.

infrastructure/builder/src/generate/generators/docs-crawler.js
djmitche: There's no deployment yet -- this is generating code that goes back into the repository, after all.

Also, running a headless browser adds a *really* big and potentially fragile requirement to our contributors for running `yarn generate`.  There's also the potential for different browsers or different versions of those browsers to produce different output, which would lead to unnecessary index changes getting checked in, or spurious failures of `yarn generate` in CI.

That said, to get this working I think you'll need to run tc-ui within the `yarn generate` process, and index the documents from there.

imbstack:
Sorry to butt in, just want to check if we can avoid using a web crawler by building indexes from the stuff that's checked and generated directly. That way we avoid the need for headless browsers and the like. We wouldn't even need an external service necessarily in that case.

helfi92:
> Is there a way to use React's server-side rendering support  to render all of that to some static HTML, and then textify that and  index it?

Good question. I think there might be. I'll look into it.

> I wonder if it would be better to just write some code to generate indexable text for markdown, and for reference files, and for schemas. That could be gradually improved to allow deep-linking to reference entries, searching for reference entries by the text of their schema, etc.

That could be the approach we take if server-side rendering doesn't fully work.

helfi92:
I think it could work with React's server-side rendering support. We could use the [`renderToString`](https://reactjs.org/docs/react-dom-server.html#rendertostring) method to render a React element to its HTML string and the [Static Router](https://reacttraining.com/react-router/web/api/StaticRouter) for the routing.

djmitche:
Applies to #405.

djmitche:
Superseded by #2080

Issue #660, opened 2019-04-30
[web-server] Use the numeric github user ID


helfi92:
https://github.com/taskcluster/taskcluster/pull/547#discussion_r279815834

PR #704, opened 2019-05-06
[web-server; login] Use numeric github user ID rather than username

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/660.

### DIFF
--- a/services/web-server/src/login/strategies/github.js
+++ b/services/web-server/src/login/strategies/github.js
@@ -5,10 +5,10 @@ const { Strategy } = require('passport-github');
 const taskcluster = require('taskcluster-client');
 const User = require('../User');
 const identityFromClientId = require('../../utils/identityFromClientId');
-const { encode, decode } = require('../../utils/codec');
 const login = require('../../utils/login');
 const WebServerError = require('../../utils/WebServerError');
 const tryCatch = require('../../utils/tryCatch');
+const { encode, decode } = require('../../utils/codec');
 const GithubClient = require('../clients/GithubClient');
 
 const debug = Debug('strategies.github');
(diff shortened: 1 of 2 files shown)


helfi92:
Ah, I guess I didn't fully understand the comment

> I think we need to use the numeric github user ID, rather than the username

Makes sense. I'll revise.

services/web-server/src/login/strategies/github.js
```diff
@@ -19,12 +19,24 @@ module.exports = class Github {
...
 
-    user.identity = `${this.identityProviderId}/${encode(userId)}`;
+    if (!username) {
+      return;
+    }
+
+    const [githubErr, githubUser] = await tryCatch(this.githubClient.userFromUsername(username));
+
+    if (githubErr) {
+      debug(`error retrieving user data from Github: ${githubErr}\n${githubErr.stack}`);
```
djmitche: Shouldn't this return?

services/web-server/src/login/strategies/github.js
```diff
@@ -38,14 +50,14 @@ module.exports = class Github {
   }
 
   async userFromToken(accessToken) {
-    const githubClient = new GithubClient();
-    const [githubErr, githubUser] = await tryCatch(githubClient.userFromToken(accessToken));
+    const [githubErr, githubUser] = await tryCatch(this.githubClient.userFromToken(accessToken));
 
     if (githubErr) {
       debug(`error retrieving user data from Github: ${githubErr}\n${githubErr.stack}`);
```
djmitche: Same here actually (should return)

helfi92:
> I think this will need a change to userFromClientId as well -- it currently assumes that the identity is just a userid.

In order to fetch user information, we require the Github username, not the ID. I wasn't able to find a documented on fetching a user using the numeric ID only.

`getUser` requires the github username in order to fetch the user information. This is the reason why a previous commit in this PR also adds

```jsx
async getUser({ userId }) {
  // ...
  const username = userId.split('|')[1];

  // ...

  const [githubErr, githubUser] = await tryCatch(this.githubClient.userFromUsername(username));

  // ...
}
```

Maybe we could just rename `userId` in `getUser` to something else?

djmitche:
Well, that's pretty annoying.  That said, ignoring the id portion of the `userId` would allow the attacks I suggested before, so we need to ensure that value is correct somehow.

An alternative, then, is to look up the user by username and verify a matching userId, returning no user if they don't match.  Then even if someone with userid 999999 takes over the `djmitche` username, the identity `github/28673|djmitche` won't match that person (but github/999999|djmitche` will).  Similarly, if I change my username to `dustin`, `github/28673|djmitche` won't work anymore.

services/web-server/src/login/strategies/github.js
```diff
@@ -22,12 +24,31 @@ module.exports = class Github {
...
     this.identityProviderId = 'github';
+    this.githubClient = new GithubClient();
   }
 
-  async getUser({ userId }) {
+  async getUser({ username, userId }) {
     const user = new User();
 
-    user.identity = `${this.identityProviderId}/${encode(userId)}`;
+    if (!username) {
```
djmitche: Is this just in case something else is calling getUser?

helfi92: Can't think of why this was needed. Let's remove it.

helfi92:
> Looks good! Maybe that /user/ -> /users/ will fix the failure of github logins to work on my laptop?

I hope.

PR #707, opened 2019-05-07
Refactor monitoring to use a singleton manager

djmitche:
With this change, there is a single defaultMonitorManager, which can be
configured at module load time and can be set up *once*, resulting in a
root monitor.  After that, child monitors are generated from that root
monitor.  This resembles the successful azure-entities approach.

Concrete changes:
 * logging cannot be disabled
 * root logger is now named `taskcluster.<serviceName>` (no `.root`)
 * `pretty` setup option has been replaced with `debug`, which causes
   pretty log output to the debug log (and is enabled by default in
   tests)
 * `mock` setup option has been replaced with `fake` because it's a
   fake, not a mock.
 * `monitor.monitor(name)` invocation is now `monitor.childMonitor(name)`
 * taskcluster-lib-testing has a `withMonitor` function that supports
   setting up a monitor for testing services

I apologize for the huge diff.  This required changes to most tests, and also involved a bunch of moving existing code around (for example between the Monitor and MonitorManager classes).  I didn't see any way to make some changes without making all of the changes.


djmitche:
Note that this has also removed the undocumented `monitor` parameter from tc-client constructors.  It seems that wasn't used in our services, and is very unlikely to be used elsewhere.

owlishDeveloper:
wth is tc-cat doing here? also, queue tests are failing

djmitche:
Yeah, I'm looking at queue.  And, tc-cat was enabled for my org -- i disabled it.

djmitche:
Ah, queue fix was easy, fixed in a sec.

owlishDeveloper:
> And, tc-cat was enabled for my org

mm, I'm not following - why would it be commenting in a PR in a taskcluster org? Or was enabled here?

djmitche:
It commented on the push to the branch for this PR, which is in my org.

owlishDeveloper:
> It commented on the push to the branch for this PR, which is in my org.

you essentially repeated the sentence I don't understand :) Can you re-phrase?

djmitche:
I think we sorted the tc-cat question in Slack.

@owlishDeveloper did you want to review this too or is it OK to merge with @imbstack's r+?

owlishDeveloper:
clear for merging!

djmitche:
rebased with no substantive changes

djmitche:
Deployed Everywhere.

PR #708, opened 2019-05-07
Bug 1545864 - Remove support for publishing schemas and references

djmitche:
This will stop publishing to raw-docs, which no longer feeds into anything of interest.  It also stops updating `schemas.taskcluster.net` and `references.taskcluster.net`, which are likely no longer up to date anyway.

If we need to generate clients from tools that require web-accessible references, we can deploy to staging and aim the tool there.

Bugzilla Bug: [1545864](https://bugzilla.mozilla.org/show_bug.cgi?id=1545864)


djmitche:
@petemoore is this OK with you?  Note that we no longer document schemas/references.taskluster.net as authoritative.

petemoore:
> Submitting the task to Taskcluster failed. Details

Do we need to worry about this?

escapewindow:
I'm not sure I know enough context to say definitively. Would the reference file be vendored in? Are there benefits to using tc-client-generator for clients in other languages? I do see potential benefits with having the python client in the monorepo, as long as everyone's ok with that.

djmitche:
(I'm not sure how I made the last few comments at some time in the future...)

> Would the reference file be vendored in?

Not for options 1 or 2.  For option 3, yes, we use a source-generation process that includes the raw references in the repo.

> Are there benefits to using tc-client-generator for clients in other languages?

I'm not sure.  I think the big benefit of tc-client-generator is generating data structures based on json-schema, for languages that need such things (golang and java).

> I do see potential benefits with having the python client in the monorepo, as long as everyone's ok with that.

I'm ok with that option, too.  And I don't think that would have to affect module-ownership.  I'll wait to hear from @petemoore.

petemoore:
I'm also fine with rolling the python client into the monorepo. 👍

PR #715, opened 2019-05-08
Fix SchemaTable to parse refs relatively

### LINKED ISSUE #657: Fix SchemaTable to parse refs relatively
The `readReference` function in `ui/src/components/SchemaTable/index.jsx` currently treats relative `$ref` as relative to id `/schemas/<serviceName>`.  This is the calculation that should be used to find a schema for a reference document's "input" or "output" property, but *not* how `$ref` should be parsed -- it should be parsed using the usual relative-URL joining.

### LINKED ISSUE #689: [Docs] schema is undefined
From https://github.com/taskcluster/taskcluster/pull/651#issuecomment-488938780:

> Is the following expected?
> 
> Screen Shot 2019-05-02 at 3 36 36 PM

I don't think this is expected.
I did some digging on this and found `references.json` file these two specific schema has below values

```
"allOf": [
        {
          "$ref": "./metaschema.json#"
        },
        {
          "$ref": "#/definitions/topLevelMetadata"
        }
      ],
```

and for this reference is going in and executing this code

`https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/SchemaTable/index.jsx#L141-L162`

and its crashing at 

`https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/SchemaTable/index.jsx#L152`

I am exactly sure, how to process this, `allOf` key.

djmitche:
This moves the logic to calculate entry schema references into the Entry
class.  It's still a bit awkward to show both actual reference entries
*and* schemas using the same Entry class, but it gets all the nice
expanding and hash reference behavior without code duplication.

I suppose an alternative would be to factor that behavior out into a ReferenceExpansionPanel class and then use that in both Entry and SchemaPanel.

Fixes #657, fixes #689.  cc: @arshadkazmi42

### DIFF
--- a/ui/src/components/SchemaTable/index.jsx
+++ b/ui/src/components/SchemaTable/index.jsx
@@ -1,14 +1,22 @@
 import React, { Component } from 'react';
 import { withRouter } from 'react-router-dom';
-import { join } from 'path';
 import RefParser from 'json-schema-ref-parser';
 import { string } from 'prop-types';
 import Spinner from '@mozilla-frontend-infra/components/Spinner';
 import { withStyles } from '@material-ui/core/styles';
 import Table from 'react-schema-viewer/lib/SchemaTable';
+import cloneDeep from 'lodash.clonedeep';
+import jsonSchemaDraft06 from 'ajv/lib/refs/json-schema-draft-06.json';
+import jsonSchemaDraft07 from 'ajv/lib/refs/json-schema-draft-07.json';
 import ErrorPanel from '../ErrorPanel';
 import { THEME } from '../../utils/constants';
 import references from '../../../../generated/references.json';
+
+// Local copies of the json-schemas schemas, since TC schemas $refer to these
+const EXTERNAL_SCHEMAS = [jsonSchemaDraft06, jsonSchemaDraft07].reduce(
+  (schemas, schema) => ({ ...schemas, [schema.$id]: schema }),
+  {}
+);
 
 @withRouter
 @withStyles(
@@ -83,19 +91,8 @@ import references from '../../../../generated/references.json';
  */
 export default class SchemaTable extends Component {
   static propTypes = {
+    // The $id of the schema to render
     schema: string.isRequired,
-    /**
-     * The service name in which the entry belongs to.
-     * Required for a reference document (api or exchanges) because
-     * the `input` and `output` property values of the schema are relative URIs.
-     * The service name helps us find the referenced schema given
-     * the relative path.
-     * */
-    serviceName: string,
-  };
-
-  static defaultProps = {
-    serviceName: null,
   };
 
   state = {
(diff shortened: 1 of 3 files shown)


ui/src/components/SchemaTable/index.jsx
```diff
@@ -119,14 +107,18 @@ export default class SchemaTable extends Component {
...
 
-  readReference(schemaPath) {
-    const { serviceName } = this.props;
-    const id = schemaPath.startsWith('/')
-      ? schemaPath
-      : join('/', 'schemas', serviceName, schemaPath);
-    const schemaId = id.endsWith('#') ? id : `${id}#`;
+  readReference(schemaId) {
+    const fullSchemaId =
+      schemaId.indexOf('#') === -1 ? `${schemaId}#` : schemaId;
```
owlishDeveloper: you can also do `schemaId.includes('#') ? ` <=== if you find that's more readable

djmitche: Oh, hm, I meant to preserve existing behavior there (only add a # at the *end*).  I'm not sure why I changed that.. So I'll switch back to the original `endsWith`.

ui/src/components/SchemaTable/index.jsx
```diff
@@ -119,14 +107,18 @@ export default class SchemaTable extends Component {
...
 
-  readReference(schemaPath) {
-    const { serviceName } = this.props;
-    const id = schemaPath.startsWith('/')
-      ? schemaPath
-      : join('/', 'schemas', serviceName, schemaPath);
-    const schemaId = id.endsWith('#') ? id : `${id}#`;
+  readReference(schemaId) {
+    const fullSchemaId =
+      schemaId.endsWith('#') ? schemaId : `${schemaId}#`;
```
helfi92: Linter is not happy here.

djmitche: Lol, of course, I change a few characters and the linter hates me :)

djmitche:
No, that's not expected.  That's referring to a schema that's not on the site (it's on json-schema.org).  I don't think we can handle that *well* at the moment -- not until we get the new schema viewer built -- but we can certainly do better than erroring out :)

djmitche:
Hm, that might not be fixable after all.  We've never tried to render those metaschemas that `$ref` the upstream JSON-schema document.  I can think of a few options:
 * wait until we have the new schema viewer which follows $ref only on click and either
   * refuse to follow external $ref, or
   * follow external refs by fetching that data remotely (noting that they're http URLs)
 * pre-load the http://json-schema.org schemas and allow $refs pointing to those
 
I'll try the last option.

djmitche:
I'm learning to hate json-schema-ref-parser..

djmitche:
OK, that's a good bit better.  Those metaschemas still don't look great, but that's more a limitation of the schema viewer than of this patch.

ui/src/components/SchemaTable/index.jsx
```diff
@@ -1,15 +1,22 @@
...
+import jsonSchemaDraft07 from 'ajv/lib/refs/json-schema-draft-07.json';
 import ErrorPanel from '../ErrorPanel';
 import { THEME } from '../../utils/constants';
 import references from '../../../../generated/references.json';
 
+// Local copies of the json-schemas schemas, since TC schemas $refer to these
+const EXTERNAL_SCHEMAS = [jsonSchemaDraft06, jsonSchemaDraft07].reduce(
+  (schemas, schema) => ({ ...schemas, [schema.$id]: schema })
+);
+
```
helfi92: You'll probably want to add `{}` as the initial value otherwise it won't work.

ui/src/components/SchemaTable/index.jsx
```diff
@@ -119,38 +115,54 @@ export default class SchemaTable extends Component {
...
+    const { protocol, hostname, pathname } = new URL(file.url);
+
+    // since json-schema-ref-parser uses the window's location for relative
+    // URIs, only map those to our local schema list
+    if (
+      protocol !== window.location.protocol ||
+      hostname !== window.location.hostname
+    ) {
+      throw new Error(`External schema ${file.url} not available`);
+    }
```
helfi92: 😍

ui/src/views/Documentation/Reference/Entry.jsx
```diff
@@ -86,12 +86,12 @@ export default class Entry extends Component {
...
-    /** The reference entry. */
-    entry: object.isRequired,
+    /** The reference entry, or {$id, schema} for a schema. */
+    entry: object,
     /** Required when `type` is `topic-exchange`. */
     exchangePrefix: string,
-    /** The service name in which the entry belongs to. */
-    serviceName: string.isRequired,
+    /** The service name to which the entry belongs, or null for a schema. */
+    serviceName: string,
```
helfi92: Provide default values for `serviceName` and`entry` in `static defaultProps`.

ui/src/components/SchemaTable/index.jsx
djmitche: It does, surprisingly!  I guess `{...undefined}` is equivalent to `{}`?  But I'll add the `{}`.

PR #718, opened 2019-05-09
[UI] Load docs in a single chunk

helfi92:
Make switching documentation pages in mobile faster. We don't need to do
this for the rest of the app because most users typically view one
micro service per visit whereas in docs people tend to view multiple pages. This commit enhances the performance in page transitions in docs.

`eager`: Generates no extra chunk. All modules are included in the current chunk and no additional network requests are made. A Promise is still returned but is already resolved. In contrast to a static import, the module isn't executed until the call to import() is made.

### DIFF
--- a/ui/src/views/Documentation/index.jsx
+++ b/ui/src/views/Documentation/index.jsx
@@ -113,13 +113,13 @@ export default class Documentation extends Component {
 
   async readDocFile(path) {
     try {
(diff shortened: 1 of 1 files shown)


helfi92:
> I'm assuming that the error-handling still works OK in this case?

Correct. In fact, once the user opens up the docs, any changes to docs won't result in a missing chunk because the chunk is already loaded on page load and so it wouldn't need to try to send a request when transitioning within docs.

Issue #629, opened 2019-04-24
[Login] Add a .md doc describing the whole login process


helfi92:
For easy access, the markdown file should be linked from the README.

djmitche:
Oh, I don't think this is fixed -- we should have some step-by-step documentation of how login works ("a redirects to b with query parameter c, ..."), as that will help guide understanding, implementation, review, and analysis of that process.

PR #719, opened 2019-05-10
[docs] Document login process

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/629.

### DIFF
--- a/dev-docs/login-process.md
+++ b/dev-docs/login-process.md
@@ -1,0 +1,97 @@
+# Login Process
+
+Making a call to a protected Taskcluster API requires a user to be logged in.
+The full process consists of:
+
+1. Using a login strategy to acquire an access token.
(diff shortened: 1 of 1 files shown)


dev-docs/login-process.md
```diff
@@ -0,0 +1,97 @@
...
+});
+```
+
+## 2. Taskcluster Credentials
+
+Authenticating with a login strategy returns back an access token to the caller.
+To get Taskcluster credentials, the caller must make a request to
+the graphQL `getCredentials` endpoint, including the `accessToken` and `provider`.
+The `accessToken` will be verified against the `provider`.
+The response will contain Taskcluster credentials corresponding to a permanent client.
```
djmitche: It's a temporary client, actually :)

PR #721, opened 2019-05-10
Bug 1550000 - strip all but the first line of multiline errors

djmitche:
..at least, in Fields.stack and in message, which are the two fields
StackDriver is concerned with.

Bugzilla Bug: [1550000](https://bugzilla.mozilla.org/show_bug.cgi?id=1550000)

### DIFF
--- a/libraries/monitor/src/monitor.js
+++ b/libraries/monitor/src/monitor.js
@@ -245,7 +245,13 @@ class Monitor {
       extra = level;
       level = 'err';
     }
-    this.log.errorReport(Object.assign({}, serializeError(err), extra), {level});
+    const serialized = serializeError(err);
+
+    // stackdriver doesn't like multiline prefixes to its stacks, so
+    // pre-process this down to "<FIRST LINE>\n  at ..."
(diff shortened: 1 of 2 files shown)


libraries/monitor/src/monitor.js
```diff
@@ -245,7 +245,13 @@ class Monitor {
       extra = level;
       level = 'err';
     }
-    this.log.errorReport(Object.assign({}, serializeError(err), extra), {level});
+    const serialized = serializeError(err);
+
+    // stackdriver doesn't like multiline prefixes to its stacks, so
+    // pre-process this down to "<FIRST LINE>\n  at ..."
+    serialized.stack = serialized.stack
+      .replace(/^([^\n]*)(?:\n[^\n]*)*?((?:\n {4}at[^\n]+)+)$/s, '$1$2');
```
imbstack: oof

djmitche: Is it too complicated?  I feel like the comment above at least summarizes such that the reader doesn't have to parse the regexp..

PR #722, opened 2019-05-10
Only use babel for clients/client-web

djmitche:


djmitche:
Too bad it doesn't pass tests..

djmitche:
OK, this babel mumbo-jumbo is over my head.  I can't figure out why regeneratorRuntime is already imported in the situation on `master`, but not on this branch.  I also can't figure out why `@babel/preset-env` is automatically required on `master` but must be listed explicitly on this branch.

We can keep installing babel as a devDependency in the root for now.

PR #727, opened 2019-05-10
Worker Table for the worker manager UI

owlishDeveloper:


ui/src/components/WMWorkersTable/index.jsx
```diff
@@ -0,0 +1,257 @@
...
+import DateDistance from '../DateDistance';
+import sort from '../../utils/sort';
+import { WMWorkers } from '../../utils/prop-types';
+
+@withRouter
+export default class WorkerManagerWorkersTable extends Component {
+  static propTypes = {
+    searchTerm: String,
+    workers: WMWorkers.isRequired,
+  };
```
helfi92: For non required `propTypes`, it would be nice to have `defaultProps` defined.

```jsx
static propTypes = {
 // ...
};

static defaultProps = {
  searchTerm: '',
};
```

ui/src/components/WMWorkersTable/index.jsx
```diff
@@ -0,0 +1,257 @@
...
+          <TableCell>
+            <Typography>n/a</Typography>
+          </TableCell>
+        )}
+
+        <TableCell>
+          <TableCellListItem
+            button
+            component={Link}
+            to={`${path}/worker-types/${workerType}/workers/${workerGroup}/${workerId}/recent-errors`}>
```
helfi92: Clicking turns the url into `http://localhost:5080/worker-manager/worker-types/:workerType/worker-types/banana/workers/%F0%9F%A6%80/rust-awesomness/recent-errors`. `worker-types` seems to be duplicated and there is `:workerType`.

ui/src/components/WMWorkersTable/index.jsx
```diff
@@ -0,0 +1,257 @@
...
+            />
+            <AlertIcon size={iconSize} />
+          </TableCellListItem>
+        </TableCell>
+
+        <TableCell>
+          <TableCellListItem
+            button
+            component={Link}
+            to={`${path}/worker-types/${workerType}/workers/${workerGroup}/${workerId}/resources`}>
```
helfi92: I see a similar behavior to the previous comment re: "Click to see errors".

ui/src/utils/prop-types.js
```diff
@@ -180,6 +180,31 @@ export const workers = shape({
   edges: arrayOf(worker),
 });
 
+export const WMWorkers = arrayOf(
+  shape({
```
helfi92: In this file, we could define what a `WMWorker` looks like and then in the component file we could say that it is an array of `WMWorker`. In other words, instead of specifying `arrayOf` in this file, we would do it in `WMWorkersTable/index.jsx `.

services/web-server/src/graphql/WorkerManager.graphql
```diff
@@ -16,6 +16,25 @@ type WorkerManagerWorkerTypeSummary {
...
+
+    latestTaskRun: TaskRun
+
+    quarantineUntil: DateTime
+
+    recentErrors: Int
+
+    providerId: String!
+    workerType: String!
+}
```
helfi92: Do we know which properties will require an extra request and which won't? If `providerId` and `workerType` don't trigger an extra request, they could be brought up below `workerAge` without the newline.

services/web-server/src/loaders/workerManager.js
```diff
@@ -12,6 +12,16 @@ module.exports = ({ workerManager }) => {
...
 
+  // const WMWorkers = new DataLoader(queries =>
+  //   Promise.all(
+  //     queries.map(async (workerType, isQuarantined, filter) => {
+  //       const workers = await workerManager.getWorkers(provisionerId, workerType, isQuarantined);
+  //
+  //       return filter ? sift(filter, workers) : workers;
+  //     })
+  //   )
+  // );
```
helfi92: Do you want me to look at the validity of these comments?

owlishDeveloper: not yet - these are just sketches at the moment. I plan to get them finished when I switch to API, and when I'll clean up this from the mock data

ui/src/components/WMWorkersTable/index.jsx
owlishDeveloper: Oh, that link is supposed to direct you to the list of the recent tasks for the worker. I'll remove the link altogether for now

ui/src/components/WMWorkersTable/index.jsx
owlishDeveloper: yeah, same thing. I'll remove these links altogether for now.

ui/src/utils/prop-types.js
owlishDeveloper: what are the advantages to this method?

services/web-server/src/graphql/WorkerManager.graphql
owlishDeveloper: Not yet (at least, I'm not concentrating on that yet). I organized this with spaces to have the following structure:

`new fields from worker manager`
`data from queue`
`I'm not even sure about this one`
`two fields that "roll over" from the previous table`

It's more for my convenience than for anything. If you like, I can add this to the clean up ticket to remove all the white spaces later

helfi92: Adding it to the clean up ticket works for me.

ui/src/utils/prop-types.js
helfi92: Advantage: If a future view will be responsible to render a single worker manager worker, then it would be able to grab the `WMWorker` structure from here.

PR #728, opened 2019-05-10
Bug 1550483 - move taskcluster-client.py in-tree

djmitche:
Bugzilla Bug: [1550483](https://bugzilla.mozilla.org/show_bug.cgi?id=1550483)


djmitche:
@escapewindow ping for your review on this too, pending your accepting the invite to this repo.

The individual commits are well-isolated and make it easier to see what has changed in the generated code (very little).

djmitche:
I suppose I'd like you to not get tripped up on something silly I've done next time you want to make a change -- could you make sure you can run the tests locally?

When this lands, I'll release a 7.1.0 to check that the release process works (it at least mechanically works, as I uploaded 7.0.1 to test.pypi.org -- https://test.pypi.org/project/taskcluster/)

escapewindow:
`tox` works for me locally, thanks!

djmitche:
It occurs to me that we can't use a pattern in .gitattributes here, since the `taskcluster/` and `taskcluster/aio/` directories both mix generated and non-generated code.  One option is to autogenerate `.gitattributes`, and another is to try to reorganize the code to better segregate generated and non-generated code.

escapewindow:
Yeah, I wondered if we should make a `generated/` directory or something. It's possible we could craft imports in a way that this is seamless to users; also possible we create a one-time breaking change.

djmitche:
We could probably leave compatibility "stubs" at the current location..

Anyway, this can be left for a followup.  Hopefully I'll get a taker from @taskcluster/services-reviewers to review this soon ;)

djmitche:
Followup in [Bug 1551850](https://bugzilla.mozilla.org/show_bug.cgi?id=1551850)

clients/client-py/CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,15 @@
+# Community Participation Guidelines
```
imbstack: We can probably remove this file, right?

clients/client-py/CONTRIBUTING.md
```diff
@@ -0,0 +1,28 @@
+# How to Contribute
```
imbstack: And this one.

djmitche:
Good point.  I think I'm just blind to those files since they're everywhere :)

PR #736, opened 2019-05-13
Google provider for worker-manager

imbstack:
I know I promised to have this ready for review by Monday but there is of course more to it than I expected. There is still some rewiring and fighting with the google api left to do but I have confidence that this plan will work.

Pretty much everything in here other than bits of the google provider itself are how I want them to be, so feel free to peruse and have me fix my mistakes. That way once I work out the remaining issues with google we can quickly land this.

Sorry about the larg-ish dump of code. I don't think my commit history here is clean enough to walk through it one-by-one unless you really want to see into the corners of my brain.


imbstack:
Ok, there's still a few more bits to tweak and the google provider itself needs some tests but I can say that this officially works pretty much! There's a couple of TODOs but I want to get this reviewed starting now so that we can land this soon after those are done.

Please be critical/harsh. I have no pride about what I've made here :anger: 
I suppose as we discussed before let's not be super nitpicky now though either since this doesn't run in production yet.

services/worker-manager/src/api.js
```diff
@@ -195,3 +227,38 @@ builder.declare({
   }
   return res.reply(data);
 });
+
+/*
+ * ************** BELOW HERE LIVE PROVIDER ENDPOINTS **************
+ */
+
+builder.declare({
```
owlishDeveloper: so, each provider will have its own API? What sort of endpoints will there be?

services/worker-manager/src/estimator.js
```diff
@@ -0,0 +1,34 @@
+class Estimator {
```
owlishDeveloper: I wonder if it would be nice to have a comment here explaining what the thing is supposed to do and why?

services/worker-manager/src/api.js
djmitche: I think we'll want to include the providerId in here, so if for example we have two gcp providers configured they're not fighting over `/credentials/google/:name`

services/worker-manager/src/provisioner.js
```diff
@@ -1,25 +1,34 @@
...
 const Iterate = require('taskcluster-lib-iterate');
+const {consume} = require('taskcluster-lib-pulse');
 
 /**
  * Run all provisioning logic
  */
 class Provisioner {
-  constructor({queue, provisionerId, providers, iterateConf, WorkerType, monitor, notify}) {
-    this.queue = queue;
+  constructor({provisionerId, providers, iterateConf, WorkerType, monitor, notify, pulseClient, reference, rootUrl}) {
```
owlishDeveloper: So what is a provider and what is a provisioner and why are they different?

services/worker-manager/README.md
```diff
@@ -2,6 +2,11 @@
 
 The worker manager service manages workers, including interacting with cloud services to create new workers on demand.
 
+It currently includes providers for:
```
owlishDeveloper: I'd prefer to see some more documentation explaining the overall architecture of the thing

services/worker-manager/src/provider.js
```diff
@@ -67,49 +87,46 @@ class Provider {
...
 
-  /**
-   * Terminate all workers managed by this provider.  This method must not
-   * return until the request to terminate all workers is completed.  It does
-   * not need to wait until all workers actually terminate.
+  /*
+   * Called after an iteration of the worker scanner
    */
-  async terminateAllWorkers() {
-    throw new Error('Method Unimplemented!');
```
owlishDeveloper: Should we leave these errors in place? They help making sure all the required methods are implemented when writing another provider

services/worker-manager/src/api.js
owlishDeveloper: what could we use two gcp providers for?

services/worker-manager/src/provisioner.js
owlishDeveloper: oh, ok, provisioner runs all the time and calls methods of the providers it has

services/worker-manager/src/provisioner.js
```diff
@@ -35,20 +44,61 @@ class Provisioner {
...
+    }
     await this.iterate.stop();
     await Promise.all(Object.values(this.providers).map(x => x.terminate()));
   }
 
+  async onMessage({exchange, payload}) {
+    const {name, provider: providerName, previousProvider} = payload;
+    const workerType = await this.WorkerType.load({name});
+    const provider = this.providers[providerName]; // Always have a provider
+    switch (exchange.split('/').pop()) {
```
owlishDeveloper: I wonder if it would be better to have just separate handler functions for separate events?

services/worker-manager/src/provisioner.js
owlishDeveloper: To summarize the above two comments: maybe some documentation is needed here 😆

services/worker-manager/src/provider.js
```diff
@@ -6,42 +9,53 @@ class Provider {
    * a cloud provider for terminating/listing instances. Any provisioning
    * logic should be started in `initiate` below.
    */
-  constructor({id, monitor, notify}) {
-    this.id = id;
+  constructor({
```
owlishDeveloper: so I noticed in `provisioner.js`, line 103, we use `provider.prepare` but I don't see this method here?

services/worker-manager/src/provider_google.js
```diff
@@ -22,86 +65,502 @@ class GoogleProvider extends Provider {
...
+    const accountId = 'taskcluster-workers';
+    this.workerAccountEmail = `${accountId}@${this.project}.iam.gserviceaccount.com`;
+    const accountRef = `projects/${this.project}/serviceAccounts/${this.workerAccountEmail}`;
+    const roleId = 'taskcluster_workers';
+    const roleName =`projects/${this.project}/roles/${roleId}`;
 
-  async terminate() {
-  }
+    // First we set up the service account
+    await this.readModifySet({
```
djmitche: I like this!

services/worker-manager/src/provider_google.js
```diff
@@ -22,86 +65,502 @@ class GoogleProvider extends Provider {
...
+   *
+   * All fields we check in the token are signed by google rather than the
+   * requester so we know that they are not forged arbitrarily. Be careful
+   * when selecting new fields to validate here, they may come from the requester.
+   */
+  async verifyIdToken({token, workerType}) {
+    // This will throw an error if the token is invalid at all
+    let {payload} = await this.oauth2.verifyIdToken({
+      idToken: token,
+      audience: this.rootUrl,
```
djmitche: clever!

services/worker-manager/src/provider_google.js
```diff
@@ -22,86 +65,502 @@ class GoogleProvider extends Provider {
...
+                  }),
+                },
+              ],
+            },
+          },
+        });
+        op = res.data;
+      } catch (err) {
+        if (!err.errors) {
+          throw err;
```
djmitche: do you want to report this error too?

services/worker-manager/src/provider_google.js
```diff
@@ -22,86 +65,502 @@ class GoogleProvider extends Provider {
...
+        }
+        for (const error of err.errors) {
+          await workerType.reportError({
+            kind: 'creation-error',
+            title: 'Instance Creation Error',
+            description: error.message, // TODO: Make sure we clear exposing this with security folks
```
djmitche: I never understood the hesitance to show these error messages..

services/worker-manager/src/provider_google.js
```diff
@@ -22,86 +65,502 @@ class GoogleProvider extends Provider {
...
   }
 
-  async terminateAllWorkers() {
-    throw new Error('Method Unimplemented!');
+  /**
+   * It is important that with the current design we only check on errors
+   * for error reporting. We should not use it to gate further progress of
+   * provisioning due to the fact that we might not succeed in recording
+   * the operation when it actually suceeded.
+   */
```
djmitche: *@djmitche deletes review comment asking what happens in this case*

services/worker-manager/src/provisioner.js
djmitche: I think "there's no such thing as a provisioner anymore" is the most accurate answer.  But we have this property called `provisionerId` that still exists for compatibility.  It now means something like "workerTypeGroup".

services/worker-manager/src/provider_google.js
```diff
@@ -1,17 +1,60 @@
...
+const libUrls = require('taskcluster-lib-urls');
+const uuid = require('uuid');
 const {google} = require('googleapis');
 const {Provider} = require('./provider');
 
 class GoogleProvider extends Provider {
 
-  constructor({id, monitor, notify, project, creds, credsFile}) {
-    super({id, monitor, notify});
+  constructor({
```
owlishDeveloper: This class doesn't seem to be implementing `terminate()`, `validate()`, `createResources(workertype)`, `updateResources(workertype)`, `removeResources(workertype)` and `cleanup()`?

services/worker-manager/src/provisioner.js
owlishDeveloper: > there's no such thing as a provisioner anymore

I do see `class Provisioner` in the code tho. I meant to ask to add documentation for it

services/worker-manager/src/api.js
```diff
@@ -45,37 +46,44 @@ builder.declare({
...
+  if (error) {
+    return res.reportError('InputValidationError', error);
+  }
+
   const now = new Date();
   let workerType;
 
+  const definition = {
+    name,
+    provider: providerName,
```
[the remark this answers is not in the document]
owlishDeveloper: by "provider and provider implementation" do you mean "interface and classes implementing the interface"?

services/worker-manager/src/data.js
```diff
@@ -46,41 +55,171 @@ WorkerType.prototype.serializable = function() {
...
+  });
+};
+
+const Worker = Entity.configure({
+  version: 1,
+  partitionKey: Entity.keys.StringKey('workerType'),
+  rowKey: Entity.keys.StringKey('workerId'),
+  properties: {
+    // The workertype this maps to.
+    workerType: Entity.types.String,
```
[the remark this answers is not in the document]
owlishDeveloper: What is `provisionerId`?

services/worker-manager/src/data.js
```diff
@@ -46,41 +55,171 @@ WorkerType.prototype.serializable = function() {
...
+    created: Entity.types.Date,
+
+    // The time that this worker is no longer needed and
+    // should be deleted
+    expires: Entity.types.Date,
+
+    // A string specifying the state this worker is in
+    // so far as worker-manager knows. This can be any
+    // of the fields defined in the enum below.
+    state: Entity.types.String,
```
owlishDeveloper: Does this include "quarantined" state?

owlishDeveloper: Oh, it doesn't. So where will that information be?

services/worker-manager/src/data.js
```diff
@@ -46,41 +55,171 @@ WorkerType.prototype.serializable = function() {
...
+
+    // Anything a provider may want to remember about this worker
+    providerData: Entity.types.JSON,
```
owlishDeveloper: will this include some info on the resources (like instance IDs and whatnot?)

services/worker-manager/src/data.js
```diff
@@ -46,41 +55,171 @@ WorkerType.prototype.serializable = function() {
...
-
-It includes the extra information:
-
-\`\`\`json
-${yaml.safeDump(extra)}
-\`\`\`
-    `.trim(),
+};
+
+const WorkerTypeError = Entity.configure({
```
owlishDeveloper: I don't see worker errors anywhere?

services/worker-manager/src/provisioner.js
djmitche: Oh, good point!  I thought you were commenting on the `provisionerId` in this line :)

services/worker-manager/src/data.js
djmitche: Would quarantined remain a thing the queue is concerned about?  It basically means "don't give this worker any tasks"

djmitche:
Is can be merge times?

services/worker-manager/src/api.js
imbstack: Ah, making the provider an arg makes sense too I think.

We would have multiple gcp providers for a couple potential reasons:

1. If we end up having a second version of a google provider once we learn more about how this one fits our needs. We would run both for a bit while we transition workers over.
2. Each provider only handles one project. In production we'll actually have multiple projects so that level-3 is isolated from level-1 or servo people are isolated from gecko people, etc.

services/worker-manager/src/estimator.js
imbstack: For sure. I need to add docs all over the place.

services/worker-manager/src/provider.js
imbstack: They would help but not all providers will implement this endpoint. For instance static workers running in our datacenters don't have anything to do for this. I expect we may need to create an endpoint that the UI can call to ask what actions a provider supports. Does that make sense to you? We can make a bug to track that work if so.

services/worker-manager/src/provisioner.js
owlishDeveloper: no it was just a random line I chose to make a generic comment on the file :)

services/worker-manager/src/provisioner.js
imbstack: Ok, sounds good to me!

services/worker-manager/src/provider.js
imbstack: It is unneeded for this provider currently.

services/worker-manager/src/data.js
owlishDeveloper: I don't know. People do want to see that information. What is "quarantined" anyways?

libraries/references/src/validate.js
```diff
@@ -22,6 +22,11 @@ const UNREFERENCED_SCHEMAS = [
   // schemas for an unpublished, deprecated API methods
   {service: 'index', schema: 'v1/list-namespaces-request.json#'},
   {service: 'queue', schema: 'v1/poll-task-urls-response.json#'},
+
+  // schemas for dynamic configs in worker-manager TODO: What to do about these?
+  {service: 'worker-manager', schema: 'v1/config-google.json#'},
+  {service: 'worker-manager', schema: 'v1/config-static.json#'},
+  {service: 'worker-manager', schema: 'v1/config-testing.json#'},
```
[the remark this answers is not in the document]
imbstack: I actually implemented it this way before. The issue is that setting an implementation in the workerType makes it awkward to switch a workertype's provider implementation compared to just switching it in config as would be done now.

Having an `anyOf` without specifying an implementation resulted in very confusing error reports when you would get something wrong.

We could definitely investigate going down this road again.

services/worker-manager/src/data.js
```diff
@@ -9,12 +11,20 @@ const WorkerType = Entity.configure({
...
+    // If a workertype was previously assigned to another provider and no longer is, it will
+    // be added to this field. The provider can then remove any resources created for this
+    // workertype and then remove itself from this field when done
+    previousProviders: Entity.types.JSON,
+
     // A useful human-readable description of what this workertype is for
     description: Entity.types.String,
 
+    // If this is true, a provider should clean up any resources with this and then delete it
+    scheduledForDeletion: Entity.types.Boolean,
```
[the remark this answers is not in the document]
imbstack: Very well could be.

services/worker-manager/src/data.js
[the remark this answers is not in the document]
imbstack: It does not include `provisionerId` at this time. It is all just hardcoded for now.

`provisionerId` is a taskcluster concept that was designed to allow for multiple provisioners to be used. So for instance we could run the old aws-provisioner alongside a gcp-provisioner and they would each handle their own workers. It is part of a `workerType` id. `provisionerId/workerTypeId`.

services/worker-manager/src/data.js
```diff
@@ -46,41 +55,171 @@ WorkerType.prototype.serializable = function() {
...
+const Worker = Entity.configure({
+  version: 1,
+  partitionKey: Entity.keys.StringKey('workerType'),
+  rowKey: Entity.keys.StringKey('workerId'),
+  properties: {
+    // The workertype this maps to.
+    workerType: Entity.types.String,
+
+    // The id of this worker
+    workerId: Entity.types.String,
```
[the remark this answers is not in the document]
imbstack: Not yet.

services/worker-manager/src/provider_google.js
imbstack: All credit to Google mostly.

services/worker-manager/src/provider_google.js
imbstack: Me neither but we were worried before so we can be worried again I guess.

imbstack: I'll include it in the RRA

services/worker-manager/src/provider_google.js
imbstack: Yeah, they ended up being unnecessary for this provider. I figured I would leave them in the base class in case other providers want them.

services/worker-manager/src/data.js
imbstack: Yeah, I believe queue knows about `quarantined` as it only really changes things for the queue.

services/worker-manager/src/data.js
imbstack: Not instanceIds currently. It probably makes sense to store that there rather than reconstructing it from the workerId every time by appending or removing `gcp-`.

services/worker-manager/src/data.js
imbstack: In my current design even an error for a worker would actually be a workerType error since we shouldn't be caring about individual workers in most cases.

imbstack: But I am not 100% convinced my design makes sense so we could definitely add a `WorkerErrors` table if that makes sense.

services/worker-manager/src/data.js
djmitche: It is a state where the worker continues to run, but the queue never assigns any tasks to the worker.  So a quarantined worker is no different from any other worker from worker-manager's perspective.

libraries/references/src/validate.js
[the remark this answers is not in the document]
djmitche: isn't the config in the workerType?

services/worker-manager/src/api.js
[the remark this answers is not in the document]
djmitche: Yes, although from the perspective of users of worker-manager, they can't see the classes and instances.

Examples of providers would be "gce-level-1", "packet-net', and "gce-level-3".  Examples of provider implementations would be "google", "packet", and "ec2".

services/worker-manager/src/data.js
djmitche: We should probably ensure this data structure has a finite maximum size -- that is, not arrays of arbitrary length.  Otherwise we'll eventually hit the size limit for a single Azure entity, leading to great sadness.

services/worker-manager/src/api.js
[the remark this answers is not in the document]
owlishDeveloper: that was “either-or” type of question, and your reply was “yes”, can you explain? Also, I wasn’t thinking about user’s perspective at all... so I didn't understand the answer

PR #758, opened 2019-05-20
Use an explicit project for GCP credentials

djmitche:
This changes the API method signature to require an explicit project_id in the request.  That means that later we can add support for multiple projects without changing the API.

This also supports mock testing with a fake google API implementation.

### DIFF
--- a/services/auth/test/gcp_test.js
+++ b/services/auth/test/gcp_test.js
@@ -1,88 +1,63 @@
+const _ = require('lodash');
 const assert = require('assert');
 const helper = require('./helper');
 const testing = require('taskcluster-lib-testing');
-const slugid = require('slugid');
-const {google} = require('googleapis');
 
-helper.secrets.mockSuite(testing.suiteName(), ['app', 'gcp'], function(mock, skipping) {
-  if (mock) {
-    return;
-  }
-
-  helper.withPulse('mock', skipping);
-  helper.withEntities('mock', skipping);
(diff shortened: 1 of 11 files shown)


services/auth/test/gcp_test.js
```diff
@@ -1,88 +1,63 @@
...
   test('gcpCredentials invalid account', async () => {
     try {
-      await helper.apiClient.gcpCredentials('-', 'invalidserviceaccount@mozilla.com');
-      assert.fail('The call should fail');
+      await helper.apiClient.gcpCredentials(helper.gcpAccount.project_id, 'invalid@mozilla.com');
     } catch (e) {
-      assert.equal(e.statusCode, 404);
+      if (e.statusCode !== 404) {
+        throw e;
+      }
```
walac: Why not a single line with `assert.equal`?

djmitche: In the case where the error is *not* a 404 error, `assert.equal` ends up hiding the error -- especially if the error doesn't have `statusCode`, as then the assertion just says `undefined !== 404`.  That's what prompted me to make this change: I messed up something and caused an error (I don't remember what) and was getting an assertion error that didn't tell me the issue.  Re-throwing that error directly showed me what I needed to fix.

djmitche:
I tried this in staging, where GCP_CREDENTIALS also aren't set, and auth started up.  So, here goes nothing in production!

PR #759, opened 2019-05-20
Stop reclaiming of credentials

imbstack:
I think I lost this check in the shuffle to get this landed


djmitche:
The tc-lib-pulse failure looks like it's just a timing burp -- it's expecting something that occurs every 10ms to occur at least 5 times in 100ms.  I can increase that range a little..

PR #764, opened 2019-05-21
Add github client-related scopes

djmitche:
Without this, temporary credentials don't work for github logins, since they require that the issuer have `auth:create-client:<clientId>`.


djmitche:
(this change is already deployed in staging)

PR #765, opened 2019-05-21
Fix codecov take 2

imbstack:
Ok, so what we've done here is just manually collect all the reports and submit them in one go. That should make reporting more reliable.


imbstack:
Codecov has failed this because coverage "went down" but otherwise it is ready!

Issue #768, opened 2019-05-22
[UI] Automatic Refreshes


helfi92:
From https://github.com/taskcluster/taskcluster/issues/750#issuecomment-494536404:

> The automatic credential refreshing seems to be timer-based, meaning that someone with an open tab will poll for new credentials every 15 minutes.  That is probably a simpler model than what we do in tools (re-request credentials just before making an API call, if necessary) -- good call.  However, I think multiple tabs will all try to renew at the same time.  We [applied some randomization](https://github.com/taskcluster/taskcluster-tools/blob/17302c900904b892bc66f359c683e855beccb9a3/src/auth/AuthController.js#L50-L55) in tc-tools to try to avoid that: in most cases one tab will try to renew first, and the others will learn about it via the resulting LocalStorage event.  This is a good followup fix, too.

PR #770, opened 2019-05-22
[UI] Apply randomization to avoid multiple tabs renewal at same time

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/768.

### DIFF
--- a/ui/src/auth/AuthController.js
+++ b/ui/src/auth/AuthController.js
@@ -67,7 +67,14 @@ export default class AuthController {
     if (user) {
       const expires = new Date(user.expires);
       const now = new Date();
-      const timeout = Math.max(0, expires.getTime() - now.getTime());
+      let timeout = Math.max(0, expires.getTime() - now.getTime());
+
+      // if the timeout is in the future, apply up to a few minutes to it
+      // randomly.  This avoids multiple tabs all trying to renew at the
+      // same time.
+      if (timeout > 0) {
+        timeout += Math.random() * 5 * 60 * 1000;
+      }
 
       this.renewalTimer = window.setTimeout(() => {
         this.renewalTimer = null;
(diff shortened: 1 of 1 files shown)


ui/src/auth/AuthController.js
```diff
@@ -67,7 +67,14 @@ export default class AuthController {
     if (user) {
       const expires = new Date(user.expires);
       const now = new Date();
-      const timeout = Math.max(0, expires.getTime() - now.getTime());
+      let timeout = Math.max(0, expires.getTime() - now.getTime());
+
+      // if the timeout is in the future, apply up to a few minutes to it
+      // randomly.  This avoids multiple tabs all trying to renew at the
+      // same time.
```
djmitche: In https://github.com/taskcluster/taskcluster-tools/blob/17302c900904b892bc66f359c683e855beccb9a3/src/auth/AuthController.js#L47, we renew after "renewAfter", and that's for renewing the auth0 access_token (calling `auth0Renew`), rather than renewing TC credentials at expiration.  So that code is "cheating" a little since it knows that the renewal time is just a suggestion, not an expiration time.

In this case, the credentials actually *expire* at `expires`, so waiting up to five minutes after that will mean up to five minutes of failing API calls -- no bueno.

I bet just subtracting a random time instead of adding it (while ensuring we don't pick a time in the past) would fix this.

helfi92: Makes a lot of sense.

ui/src/auth/AuthController.js
```diff
@@ -67,7 +67,14 @@ export default class AuthController {
...
+      if (timeout > 0) {
+        timeout -= Math.random() * 5 * 60 * 1000;
```
djmitche: > while ensuring we don't pick a time in the past

;)

helfi92: 🤦‍♂️

PR #772, opened 2019-05-22
[UI] Staying signed in

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/768.

### DIFF
--- a/ui/src/auth/AuthController.js
+++ b/ui/src/auth/AuthController.js
@@ -3,6 +3,7 @@ import { snake, upper } from 'change-case';
 import { AUTH_STORE } from '../utils/constants';
 import credentialsQuery from './credentials.graphql';
 import removeKeys from '../utils/removeKeys';
+import UserSession from './UserSession';
 
 /**
  * Controller for authentication-related pieces of the site.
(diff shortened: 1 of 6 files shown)


ui/src/auth/AuthController.js
```diff
@@ -43,14 +43,15 @@ export default class AuthController {
...
-    const user = auth ? JSON.parse(auth) : null;
+    let user = auth ? JSON.parse(auth) : null;
 
     if (user) {
       const expires = new Date(user.expires);
       const now = new Date();
 
       if (expires < now) {
         localStorage.removeItem(AUTH_STORE);
+        user = null;
```
djmitche: Ah, this fixes the bug that is causing me to be unable to replicate the signed-out-after-fifteen-minutes issue.  So, it would mean everyone gets signed out after 15 minutes :)

*Before* this change, AuthController removes the item from LocalStorage, but then `resetRenewalTimer` schedules a call to `this.renew()` with a delay of 0ms, so we immediately renew and reset the value in LocalStorage.

I think we need to distinguish *two* expiration times: expiration of the TC credentials (15 minutes) and expiration of the access_token (might be a lot longer).  If on page load the access_token is still valid but the TC credentials are not, then we need to renew the TC credentials but need not delete the user.  So it probably makes sense to track both of those expiration times on the user class, using distinct property names.

If the access_token is no longer valid, then we need to try to renew it.  For auth0, there's a [hidden-iframe way to do that](https://github.com/taskcluster/taskcluster-tools/blob/17302c900904b892bc66f359c683e855beccb9a3/src/auth/auth0.js#L35-L46).  For GitHub, I'm not sure how best to do it -- maybe just signing the user out in that case.

This PR is about TC credentials expiration.  The access_token renewal can be left to a followup.

Issue #769, opened 2019-05-22
[web-server] Scanning


helfi92:
From https://github.com/taskcluster/taskcluster/issues/750#issuecomment-494536404:

> The client scanner is defined, but not currently set up to run in kubernetes or in Heroku. We probably shouldn't run it in Heroku (since tc-login is still handing out credentials there) but it should be set up to scan in other deployments, and we should test that it works correctly (I was also going to say it should have tests like it does in tc-login, but I see now there are no tests for tc-login's scanner!).

PR #777, opened 2019-05-23
[web-server] Add scanner cron process to web-server login

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/769.


djmitche:
We should probably get some tests for this, too.  If you want to assign that one to me that's OK :)

PR #779, opened 2019-05-23
Bug 1552240 - use providerId and providerType consistently

djmitche:
This matches the terminology we agreed on.  This is a breaking change to the API, but contains migrations for the DB changes so there will be no need to delete any tables.

Bugzilla Bug: [1552240](https://bugzilla.mozilla.org/show_bug.cgi?id=1552240)


services/worker-manager/src/data.js
```diff
@@ -7,18 +7,33 @@ const WorkerType = Entity.configure({
...
+    description: Entity.types.String,
+    scheduledForDeletion: Entity.types.Boolean,
+    created: Entity.types.Date,
+    lastModified: Entity.types.Date,
+    config: Entity.types.JSON,
+    owner: Entity.types.String,
+    emailOnError: Entity.types.Boolean,
+    providerData: Entity.types.JSON,
+  },
+}).configure({
```
imbstack: Honestly I'd prefer we just break things again to keep this file simpler. I doubt many people have deployed a cluster in between the last breaking change and this one.

djmitche: Fair enough

services/worker-manager/README.md
```diff
@@ -2,10 +2,16 @@
...
+## Providers
 
-* Static Workers
-* Google Cloud
+The service configuration defines a number of providers, indexed by `providerId`, along with configuration.
+Each provider has a `providerType` indicating the class that implements the provider.
+
+This service currently includes providers for:
+
+* Static Workers (`static`)
```
owlishDeveloper: @djmitche How will these be mapped onto `providerId`s? For example, if I create a workerpool with `providerId` "banana" - how do we know to which cloud we should send the tasks? (I am working on the worker pool creation ui, and started having this question)

imbstack: My understanding of our plan is that we should have an endpoint on worker-manager that lists providers. It will return data about which implementation a provider uses. Does that make sense?

PR #783, opened 2019-05-23
Use a Providers object to index providers

djmitche:
Subsumes #782.


djmitche:
(I'll fix up the lint before landing... gotta run right now)

PR #802, opened 2019-05-29
disallow an empty argument to childMonitor

djmitche:
I noticed these empty `.childMonitor()` in worker-manager.

### DIFF
--- a/libraries/monitor/src/monitor.js
+++ b/libraries/monitor/src/monitor.js
@@ -81,6 +81,7 @@ class Monitor {
    * Get a prefixed child monitor
    */
   childMonitor(name, metadata = {}) {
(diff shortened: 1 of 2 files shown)


djmitche:
image
:eyes:

imbstack:
For once I think the coverage thing was accurate. [This](https://codecov.io/gh/taskcluster/taskcluster/commit/02af5db645031efa8f5d1f58b6af28f30169f09b) shows that the patch itself was 33% covered. We could change it so that new patches don't need to have high coverage to pass. I'm happy with anything.

djmitche:
Ah, I think that's because the expirations that I updated aren't actually tested -- I have tests for that in another branch.

PR #804, opened 2019-05-29
Add tests for listWorkerTypes

owlishDeveloper:
Bugzilla Bug: [1554799](https://bugzilla.mozilla.org/show_bug.cgi?id=1554799)

NOTA BENE: to be merged after https://github.com/taskcluster/taskcluster/pull/798 to avoid conflicts


services/worker-manager/test/api_test.js
```diff
@@ -208,6 +208,29 @@ helper.secrets.mockSuite(testing.suiteName(), ['taskcluster', 'azure'], function
     throw new Error('delete of non-existent workertype succeeded');
   });
 
+  test('get workertypes', async function() {
```
imbstack: Should we also have a test with >1 workertypes?

owlishDeveloper: I was thinking about this. I decided against because then there probably will be the sorting issue - but we sort on the frontend.

services/worker-manager/src/main.js
```diff
@@ -87,7 +87,7 @@ let load = loader({
   expireErrors: {
     requires: ['cfg', 'WorkerPoolError', 'monitor'],
     setup: ({cfg, WorkerPoolError, monitor}) => {
-      return monitor.childMonitor('expireErrors').oneShot('expire workerPoolErrors', async () => {
```
imbstack: I think removing `'expireErrors'` was a merge typo?

owlishDeveloper: oh, maybe!

djmitche:
> I think removing 'expireErrors' was a merge typo?

We will soon have a test to catch that :)

PR #811, opened 2019-05-30
Bug 1553965 - delete worker pools by setting providerId="null"

djmitche:
With some more tests for expirations (+fixes) and with a tweak to the loader so that it doesn't get confused when a component's value is `undefined`.

Bugzilla Bug: [1553965](https://bugzilla.mozilla.org/show_bug.cgi?id=1553965)

### DIFF
--- a/services/worker-manager/src/api.js
+++ b/services/worker-manager/src/api.js
@@ -71,7 +71,6 @@ builder.declare({
     owner: input.owner,
     emailOnError: input.emailOnError,
     providerData: {},
-    scheduledForDeletion: false,
   };
 
   try {
@@ -103,7 +102,13 @@ builder.declare({
     'worker-manager:provider:<providerId>',
   ]},
   description: [
-    'Given an existing worker pool definition, this will modify it and return the new definition.',
+    'Given an existing worker pool definition, this will modify it and return',
+    'the new definition.',
+    '',
+    'To delete a worker pool, set its `providerId` to `"null-provider"`.',
+    'After any existing workers have exited, a cleanup job will remove the',
+    'worker pool.  During that time, the worker pool can be updated again, such',
+    'as to set its `providerId` to a real provider.',
   ].join('\n'),
 }, async function(req, res) {
   const {workerPoolId} = req.params;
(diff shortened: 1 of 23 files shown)


services/worker-manager/src/api.js
```diff
@@ -173,34 +179,6 @@ builder.declare({
   res.reply(workerPool.serializable());
 });
 
-builder.declare({
```
imbstack: I sorta wonder if we should keep this endpoint and have it just set the `providerId` to `"null"` itself?

services/worker-manager/src/api.js
```diff
@@ -103,7 +102,14 @@ builder.declare({
     'worker-manager:provider:<providerId>',
   ]},
   description: [
-    'Given an existing worker pool definition, this will modify it and return the new definition.',
+    'Given an existing worker pool definition, this will modify it and return',
+    'the new definition.',
+    '',
+    'To delete a worker pool, set its `providerId` to `"null"` and empty its',
```
imbstack: To avoid confusion, what if we called it `"null-provider"`

imbstack: Or `"expire-provider"` or... something like that?

services/worker-manager/src/api.js
```diff
@@ -103,7 +102,14 @@ builder.declare({
...
+    'configuration (note that this is a string and not the JSON value `null`).',
```
imbstack: We could also make `null-config` allow any properties and then you wouldn't need to delete the config?

services/worker-manager/src/api.js
djmitche: I opted to get rid of the endpoint since it's not really deleting anymore -- I think folks might be surprised when after "deleting" a worker pool that pool still appears in the list.

services/worker-manager/src/api.js
djmitche: I like null-provider..

services/worker-manager/src/api.js
djmitche: Oh, that's a good idea.

Issue #806, opened 2019-05-29
[web-server] auth changes


helfi92:
#### TODOs:
* [x] Generate a jwt token when a user uses a strategy (mozilla-auth0, github) in web-server.
* [x] Return a taskcluster token (jwt) and an expiration time when using a strategy to login

Nits:
* [x] Remove `api.sso.mozilla.com` from the `audience` config and `taskcluster-credentials` from `scope`. I don't think we need this anymore.  We will not need `audience` or `scopes` config for mozilla-auth0 at all anymore (no need for an audience at all, and scopes can be hard-coded to `openid profile`)
* [x] Do not decode jwt in [`getPersonApi()`](https://github.com/taskcluster/taskcluster/blob/master/services/web-server/src/login/strategies/mozilla-auth0.js). Use the `expires_in` directly that we get from the `/oauth/token` endpoint.
* [x] Remove `window.opener.location` from [`callback.ejs`](https://github.com/taskcluster/taskcluster/blob/master/services/web-server/src/views/callback.ejs). Use rootUrl instead.

djmitche:
I modified the first nit slightly.

helfi92:
@djmitche Would you recommend the private signing key to be stored as an environment variable?

djmitche:
I think it should be service config -- that's what we do for, for example, crypto keys for Azure tables.

djmitche:
Is this finished?

helfi92:
Indeed.

PR #815, opened 2019-05-30
[web-server] Remove window.opener.location and stop decoding jwt

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/806.

### DIFF
--- a/services/web-server/src/login/strategies/mozilla-auth0.js
+++ b/services/web-server/src/login/strategies/mozilla-auth0.js
@@ -10,6 +10,7 @@ const { encode, decode } = require('../../utils/codec');
 const identityFromClientId = require('../../utils/identityFromClientId');
 const verifyJwt = require('../../utils/verifyJwt');
 const tryCatch = require('../../utils/tryCatch');
+const login = require('../../utils/login');
 
 const debug = Debug('strategies.mozilla-auth0');
 
@@ -47,8 +48,8 @@ module.exports = class MozillaAuth0 {
       });
     const {
       access_token: accessToken,
-      expires_in: expiresIn
-    }  = JSON.parse(res.text);
+      expires_in: expiresIn,
+    } = JSON.parse(res.text);
     const expires = new Date().getTime() + (expiresIn * 1000);
 
     if (!accessToken) {
@@ -185,6 +186,7 @@ module.exports = class MozillaAuth0 {
   useStrategy(app, cfg) {
     const { credentials } = cfg.taskcluster;
     const strategyCfg = cfg.login.strategies['mozilla-auth0'];
+    const loginMiddleware = login(cfg.app.publicUrl);
 
     if (!credentials || !credentials.clientId || !credentials.accessToken) {
       throw new Error(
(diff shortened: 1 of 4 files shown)


services/web-server/src/login/strategies/mozilla-auth0.js
```diff
@@ -240,6 +239,7 @@ module.exports = class MozillaAuth0 {
       (request, response) => {
         response.render('callback', {
           user: request.user,
+          publicUrl: cfg.app.publicUrl,
         });
```
djmitche: Would it be useful to abstract this out to a utility?

services/web-server/src/login/strategies/mozilla-auth0.js
```diff
@@ -46,16 +45,16 @@ module.exports = class MozillaAuth0 {
         client_id: this.clientId,
         client_secret: this.clientSecret,
       });
-    const accessToken = JSON.parse(res.text).access_token;
+    const {
+      access_token: accessToken,
+      expires_in: expiresIn
+    }  = JSON.parse(res.text);
+    const expires = new Date().getTime() + (expiresIn * 1000);
```
djmitche: +1 although maybe not intended for this PR?

services/web-server/src/login/strategies/mozilla-auth0.js
```diff
@@ -240,6 +239,7 @@ module.exports = class MozillaAuth0 {
       (request, response) => {
         response.render('callback', {
           user: request.user,
+          publicUrl: cfg.app.publicUrl,
```
djmitche: I don't think this config exists -- I don't see `publicUrl` in `config.yml` anyway, and I recall getting rid of that configuration a while ago.  I think you'll want `cfg.taskcluster.rootUrl`.

helfi92: `publicUrl` is used for the callback URL https://github.com/taskcluster/taskcluster/blob/6ccceca715cdcad390af530119a3155f8a41a4ee/services/web-server/src/login/strategies/mozilla-auth0.js#L206 The taskcluster-web-server service in heroku has `TASKCLUSTER_ROOT_URL` pointing to `https://taskcluster.net`. If we send a post message to this URL, I think the front end won't be able to intercept the response since it's a different origin.

services/web-server/src/login/strategies/mozilla-auth0.js
helfi92: This PR does both:
* Remove window.opener.location
* Stop decoding jwt

services/web-server/src/login/strategies/mozilla-auth0.js
helfi92: I'm not sure I understand the question. What kind of utility?

services/web-server/src/login/strategies/mozilla-auth0.js
helfi92: >  I don't see publicUrl in config.yml

https://github.com/taskcluster/taskcluster/blob/6ccceca715cdcad390af530119a3155f8a41a4ee/services/web-server/config.yml#L4

djmitche: Whoops, I looked in `services/worker-manager/config.yml` since that's what I was working on.  You're right about publicUrl.

services/web-server/src/login/strategies/mozilla-auth0.js
djmitche: These five lines appear in both strategies, and likely would appear in all strategies.  It's a minor detail.

PR #816, opened 2019-05-30
[web-server] Modify audience and scope fields in mozilla-auth0

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/806.

### DIFF
--- a/services/web-server/README.md
+++ b/services/web-server/README.md
@@ -119,8 +119,7 @@ UI_LOGIN_STRATEGIES='{
     "mozilla-auth0": {
         "domain": "<auth0 subdomain>",
         "clientId": "<clientId from registration of client>",
-        "clientSecret": "<clientSecret from registration of client>",
-        "scope": "openid profile"
+        "clientSecret": "<clientSecret from registration of client>"
     }
 }'
 ```
--- a/services/web-server/README.md
+++ b/services/web-server/README.md
@@ -129,7 +129,6 @@ or in `user-config.yml` as
 * `login.strategies.mozilla-auth0.domain`
 * `login.strategies.mozilla-auth0.clientId`
 * `login.strategies.mozilla-auth0.clientSecret`
-* `login.strategies.mozilla-auth0.scope`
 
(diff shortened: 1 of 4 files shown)


services/web-server/README.md
```diff
@@ -120,8 +120,7 @@ UI_LOGIN_STRATEGIES='{
         "domain": "<auth0 subdomain>",
         "clientId": "<clientId from registration of client>",
         "clientSecret": "<clientSecret from registration of client>",
-        "audience": "api.sso.mozilla.com",
-        "scope": "display:all taskcluster-credentials openid profile"
+        "scope": "openid profile"
```
djmitche: There's no other acceptable value for this field, so I don't think it should be exposed as configuration.

services/web-server/src/login/strategies/mozilla-auth0.js
```diff
@@ -201,7 +200,6 @@ module.exports = class MozillaAuth0 {
           domain: strategyCfg.domain,
           clientID: strategyCfg.clientId,
           clientSecret: strategyCfg.clientSecret,
-          audience: strategyCfg.audience,
           scope: strategyCfg.scope,
```
djmitche: `'openid profile'` can be hard-coded here.

services/web-server/README.md
helfi92: True that!

services/web-server/test/strategyMozillaAuth0_test.js
```diff
@@ -10,7 +10,6 @@ suite(testing.suiteName(), () => {
         strategies: {
           'mozilla-auth0': {
             domain: 'login-test.taskcluster.net',
-            audience: 'login-test.taskcluster.net',
             clientId: 'abcd',
             clientSecret: 'efgh',
             scope: 'taskcluster',
```
djmitche: ..and here

services/web-server/README.md
djmitche: still needs to be removed here..

PR #820, opened 2019-05-31
Handle unknown providerId without crashing

djmitche:

### DIFF
--- a/services/worker-manager/Procfile
+++ b/services/worker-manager/Procfile
@@ -1,7 +1,7 @@
 # GENERATED FILE from `yarn generate`! To update, change procs.yml and regenerate.
 
 web: cd services/worker-manager && node src/main.js server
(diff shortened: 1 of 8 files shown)


imbstack:
Do we want to make codecov not complain when a patch reduces coverage?

djmitche:
It seems like I should probably do a better job of testing this.

But the loss of 1% overall seems related to things in tc-queue getting hit or not hit..

imbstack:
It allows that 1% now, the project check is green. The 41% coverage for this patch is what failed. We can make it allow things like this and not fail.

djmitche:
OK, added some coverage :crossed_fingers:

djmitche:
> 84% of diff hit (target 73%)
:grin:

djmitche:
Re-flagging for review mainly for the change to main.js.

PR #823, opened 2019-05-31
Remove an eslint check with lots of false positives

djmitche:


djmitche:
Should I leave a comment in place so we rememebr?

PR #833, opened 2019-06-03
added isSecretDirty logic

shubhamgupta2956:
Added isSecretDirty logic which checks for unchanged secret


helfi92:
This is awesome. I will review it soon.

helfi92:
I uncovered a small issue that I'll need to fix prior to merging this. I'll try to fix this sometime this week (probably tomorrow).

ui/src/components/SecretForm/index.jsx
```diff
@@ -132,6 +133,7 @@ export default class SecretForm extends Component {
   render() {
     const { classes, isNewSecret, loading } = this.props;
     const { secretName, editorValue, expires, showSecret } = this.state;
+    const isSecretDirty = !equals(secret, this.props.Secret);
```
helfi92: This won't work because the secret values are actually in `this.state.secretName`, `this.state.expires` and `this.state.editorValue`. Also, `this.props.Secret` should be `this.props.secret` (lower case "s").

To fix this, you would need to do something like:

```jsx
const { secret, classes, isNewSecret, loading } = this.props;
const { secretName, editorValue, expires, showSecret } = this.state;
const isSecretDirty =
  secretName !== secret.name ||
  editorValue !== safeDump(secret.secret) ||
  expires !== secret.expires;
```

shubhamgupta2956:
Thanks. I have resolved the issue. One more thing, should I also put the isClientDirty and isRoleDirty logic in client form and role form?

helfi92:
Yeah, that would be really awesome. Thank you!

ui/src/components/SecretForm/index.jsx
```diff
@@ -130,8 +130,12 @@ export default class SecretForm extends Component {
   };
 
   render() {
-    const { classes, isNewSecret, loading } = this.props;
+    const { secret, classes, isNewSecret, loading } = this.props;
     const { secretName, editorValue, expires, showSecret } = this.state;
+    const isSecretDirty =
+      secretName !== secret.name ||
```
helfi92: This would probably not work when creating a new secret (http://localhost:5080/secrets/create). To fix this, we would need to add a check for `isNewSecret`:

```jsx
const isSecretDirty =
  isNewSecret ||
 // same as before
```

PR #843, opened 2019-06-05
Make the worker pools view functional

owlishDeveloper:
Changes:

* Now the form displays actual data it gets from db and pulse. The mock data was cleaned up
* I removed some columns from the table that got in there by mistake (those were people's wishes for workers table, not worker pools)
* Graphql types reflect the latest changes in our worker manager db table

Relevant Bugzilla Bugs: [1554799](https://bugzilla.mozilla.org/show_bug.cgi?id=1554799), [1548068](https://bugzilla.mozilla.org/show_bug.cgi?id=1548068)

The thing is deployed on Brian's cluster: [click](https://taskcluster.imbstack.com/worker-manager/) - please note it was deployed before I rebased on master, so the Add Worker Pool button is missing from the deployed version

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,7 @@
     "taskcluster-lib-scopes": "link:libraries/scopes",
     "taskcluster-lib-urls": "^12.0.0",
     "taskcluster-lib-validate": "link:libraries/validate",
+    "taskcluster-worker-manager": "link:services/worker-manager",
     "thirty-two": "^1.0.2",
     "topo-sort": "^1.0.0",
     "type-is": "^1.6.15",
--- a/package.json
+++ b/package.json
@@ -103,7 +103,6 @@
     "taskcluster-lib-scopes": "link:libraries/scopes",
     "taskcluster-lib-urls": "^12.0.0",
     "taskcluster-lib-validate": "link:libraries/validate",
-    "taskcluster-worker-manager": "link:services/worker-manager",
     "thirty-two": "^1.0.2",
     "topo-sort": "^1.0.0",
     "type-is": "^1.6.15",
--- a/services/web-server/src/graphql/WorkerManager.graphql
+++ b/services/web-server/src/graphql/WorkerManager.graphql
@@ -1,19 +1,23 @@
 type WorkerManagerWorkerPoolSummary {
-    workerPool: String!
+    workerPoolId: String!
     description: String
     owner: String!
-    minCapacity: Int!
-    maxCapacity: Int!
-    requestedCapacity: Int!
-    pendingCapacity: Int!
-    runningCapacity: Int!
-    lastActive: DateTime!
-    lastResolved: DateTime!
-    pendingTasks: Int!
-    failed: Int!
-    exception: Int!
-    unscheduled: Int!
-    provider: String!
+    config: WMWorkerPoolGeneralConfig
+    lastActive: DateTime
+    lastResolved: DateTime
+    pendingTasks: Int
+    failed: Int
+    exception: Int
+    unscheduled: Int
+    providerId: String!
+}
+
+type WMWorkerPoolGeneralConfig {
+    minCapacity: Int
+    maxCapacity: Int
+    requestedCapacity: Int
+    pendingCapacity: Int
+    runningCapacity: Int
 }
 
 type WorkerManagerWorker {
--- a/services/web-server/src/graphql/WorkerManager.graphql
+++ b/services/web-server/src/graphql/WorkerManager.graphql
@@ -3,11 +3,6 @@ type WorkerManagerWorkerPoolSummary {
     description: String
     owner: String!
     config: WMWorkerPoolGeneralConfig
-    lastActive: DateTime
-    lastResolved: DateTime
-    failed: Int
-    exception: Int
-    unscheduled: Int
     providerId: String!
 
     pendingTasks: Int
--- a/services/web-server/src/graphql/WorkerManager.graphql
+++ b/services/web-server/src/graphql/WorkerManager.graphql
@@ -2,18 +2,9 @@ type WorkerManagerWorkerPoolSummary {
     workerPoolId: String!
     description: String
     owner: String!
-    config: WMWorkerPoolGeneralConfig
     providerId: String!
 
     pendingTasks: Int #this value comes from a separate request to queue
-}
-
-type WMWorkerPoolGeneralConfig {
-    minCapacity: Int
-    maxCapacity: Int
-    requestedCapacity: Int
-    pendingCapacity: Int
-    runningCapacity: Int
 }
 
 type WorkerManagerWorker {
--- a/services/web-server/src/graphql/WorkerManager.graphql
+++ b/services/web-server/src/graphql/WorkerManager.graphql
@@ -5,11 +5,12 @@ type WorkerManagerWorkerPoolSummary {
     config: WMWorkerPoolGeneralConfig
     lastActive: DateTime
     lastResolved: DateTime
-    pendingTasks: Int
     failed: Int
     exception: Int
     unscheduled: Int
     providerId: String!
+
+    pendingTasks: Int
 }
 
 type WMWorkerPoolGeneralConfig {
--- a/services/web-server/src/graphql/WorkerManager.graphql
+++ b/services/web-server/src/graphql/WorkerManager.graphql
@@ -5,7 +5,7 @@ type WorkerManagerWorkerPoolSummary {
     config: WMWorkerPoolGeneralConfig
     providerId: String!
 
-    pendingTasks: Int
+    pendingTasks: Int #this value comes from a separate request to queue
 }
 
 type WMWorkerPoolGeneralConfig {
--- a/services/web-server/src/loaders/workerManager.js
+++ b/services/web-server/src/loaders/workerManager.js
@@ -5,7 +5,7 @@ module.exports = ({ workerManager }) => {
   const WorkerManagerWorkerPoolSummaries = new DataLoader(queries =>
     Promise.all(
       queries.map(async ({ filter }) => {
-        const summaries = await workerManager.listWorkerPools();
+        const summaries = (await workerManager.listWorkerPools()).workerPools;
 
         return filter ? sift(filter, summaries) : summaries;
       })
--- a/services/web-server/src/loaders/workerManager.js
+++ b/services/web-server/src/loaders/workerManager.js
@@ -2,15 +2,15 @@ const DataLoader = require('dataloader');
 const sift = require('sift').default;
 
 module.exports = ({ workerManager }) => {
-  // const WorkerManagerWorkerPoolSummaries = new DataLoader(queries =>
-  //   Promise.all(
-  //     queries.map(async (filter) => {
-  //       const summaries = await workerManager.listWorkerPools();
-  //
-  //       return filter ? sift(filter, summaries) : summaries;
-  //     })
-  //   )
-  // );
+  const WorkerManagerWorkerPoolSummaries = new DataLoader(queries =>
+    Promise.all(
+      queries.map(async ({ filter }) => {
+        const summaries = await workerManager.listWorkerPools();
+
+        return filter ? sift(filter, summaries) : summaries;
+      })
+    )
+  );
 
   // const WMWorkers = new DataLoader(queries =>
   //   Promise.all(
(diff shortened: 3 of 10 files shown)


helfi92:
When I try to checkout this branch and load http://localhost:5080/worker-manager, I am presented with an error `getaddrinfo ENOTFOUND worker-manager.taskcluster.net worker-manager.taskcluster.net:443`. Is there a way I can make it work for me locally on my machine?

owlishDeveloper:
@helfi92 it works for me here. Which command gets you that error?

owlishDeveloper:
@helfi92 if you mean see that error in the ui - is your local server connected to azure? where does it get the data from?

owlishDeveloper:
@helfi92 - (sorry, at first I had thought you are not able to start the server or smth) yeah, so I just checked - I get the same error. For me, the reason is my server here is not connected to azure (I didn't supply much creds to the thing here), so I would check your credentials in the server's config, and check the azure you want it to talk to and make sure it has the table (and data if you want to see the data in the ui)

helfi92:
I don't think `web-server/config.yml` asks for azure configs. Do we need to update that?

owlishDeveloper:
> I don't think web-server/config.yml asks for azure configs. Do we need to update that?

I think that would be taskcluster client creds. I am verifying that. I don't think we need to make changes to config - the deployed version seems to be working

owlishDeveloper:
@helfi92 ah hold on - it looks like worker-manager has to be running too, I think that's the problem

owlishDeveloper:
no, that doesn't make much sense... 🤔 let me think. In the development mode it definitely doesn't get some config it gets if you deploy

owlishDeveloper:
@helfi92 how do you typically debug errors seen in the ui? They are not logged in console, there's no errors on the Network tab, and there's no stack trace in the ui

owlishDeveloper:
@helfi92 so the reason is because the build is ranning against production and makes requests to `taskcluster.net` (in this particular case, it's `GET https://worker-manager.taskcluster.net/v1/worker-pools`) - and there's no worker manager in production yet. You can try running your local build against bstack's cluster

services/web-server/src/resolvers/WorkerManager.js
```diff
@@ -1,4 +1,15 @@
+const { splitWorkerPoolId } = require('taskcluster-worker-manager/src/util');
+
 module.exports = {
+  WorkerManagerWorkerPoolSummary: {
+    pendingTasks({ workerPoolId }, args, { loaders }) {
+      const { provisionerId, workerType } = splitWorkerPoolId(workerPoolId);
+      return loaders.pendingTasks.load({
```
owlishDeveloper: this is a request to the queue (`GET /pending/:provisionerId/:workerType`) - in the today's discussion in Slack turned out that in worker manager two parts of the `workerPoolId` have no meaning. My question is: does this request make sense here? And also: should there be a new analogous endpoint for worker manager?

Queue checks up on pulse queues for `pending` messages, but we probably won't be having such queues anymore?

package.json
```diff
@@ -103,6 +103,7 @@
     "taskcluster-lib-scopes": "link:libraries/scopes",
     "taskcluster-lib-urls": "^12.0.0",
     "taskcluster-lib-validate": "link:libraries/validate",
+    "taskcluster-worker-manager": "link:services/worker-manager",
```
imbstack: Why does this need to be added here?

ui/src/views/WorkerManager/WMViewWorkerPools/WMWorkerPools.graphql
```diff
@@ -1,14 +1,9 @@
 query workerPools {
   WorkerManagerWorkerPoolSummaries {
-    workerPool
-    pendingCapacity
-    runningCapacity
+    workerPoolId
+    providerId
+    description
+    owner
     pendingTasks
```
helfi92: I think it would be beneficial to add a newline before `pendingTasks` to warn developers that this will trigger an extra request.

ui/src/utils/prop-types.js
```diff
@@ -254,16 +254,15 @@ export const awsProvisionerWorkerType = shape({
...
+  owner: string,
   pendingTasks: number,
   lastActive: date,
   lastResolved: date,
   failed: number,
   exception: number,
   unscheduled: number,
-  provider: string,
+  providerId: string,
 });
```
helfi92: I think the properties in here should reflect the values in https://github.com/taskcluster/taskcluster/blob/master/services/worker-manager/schemas/v1/worker-pool-full.yml.

services/web-server/src/graphql/WorkerManager.graphql
```diff
@@ -1,19 +1,19 @@
...
-    pendingCapacity: Int!
-    runningCapacity: Int!
-    lastActive: DateTime!
-    lastResolved: DateTime!
-    pendingTasks: Int!
-    failed: Int!
-    exception: Int!
-    unscheduled: Int!
-    provider: String!
+    config: WMWorkerPoolGeneralConfig
```
helfi92: `config` is defined as an object in https://github.com/taskcluster/taskcluster/blob/master/services/worker-manager/schemas/v1/worker-pool-full.yml. Are we sure those properties won't change for a different worker pool? Is it possible we want to show a different table depending on the type of worker pool? Do we foresee a future where a user may want to see a certain value returned in config but this value happens not to be relevant for a different worker pool?

For property values that don't have a well defined structure, I think it makes sense to make the return value a JSON.

```jsx
config: JSON
```

package.json
owlishDeveloper: question for a question: why are other taskcluster dependencies added here?

services/web-server/src/resolvers/WorkerManager.js
owlishDeveloper: @djmitche or @imbstack ^ that is a question for you

helfi92:
>  You can try running your local build against bstack's cluster

Thanks. That did it :)

services/web-server/src/loaders/workerManager.js
```diff
@@ -2,15 +2,15 @@ const DataLoader = require('dataloader');
...
-  //       const summaries = await workerManager.listWorkerPools();
-  //
-  //       return filter ? sift(filter, summaries) : summaries;
-  //     })
-  //   )
-  // );
+  const WorkerManagerWorkerPoolSummaries = new DataLoader(queries =>
+    Promise.all(
+      queries.map(async ({ filter }) => {
+        const summaries = (await workerManager.listWorkerPools()).workerPools;
```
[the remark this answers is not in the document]
owlishDeveloper: no, we can't. I added the parenthesis based on my experience

ui/src/views/WorkerManager/WMViewWorkerPools/WMWorkerPools.graphql
owlishDeveloper: Newline itself doesn't explain anything. I can add a comment to that effect.

If there is no way to automate such newlines, it means it's a matter of taste. Maybe we could omit preferential suggestions in code reviews?

ui/src/utils/prop-types.js
owlishDeveloper: Can you point out the specific problem you see here?

services/web-server/src/graphql/WorkerManager.graphql
owlishDeveloper: > Are we sure those properties won't change for a different worker pool? 

It's not used at the moment anyways, I will remove that property altogether

services/web-server/src/resolvers/WorkerManager.js
djmitche: I think the request does make sense.  Queue is responsible for queues, so it's definitely the thing to ask about pending tasks.  And GraphQL is meant to integrate data from multiple backend services, so I think it makes sense to call both queue and worker-manager in this resolver.

For calling APIs that don't yet support `workerPoolId` or `taskQueueId`, we split it at the `/` and pass `provisionerId` and `workerType` separately, as you've don here.

I just wrote up an [RFC for this](https://github.com/taskcluster/taskcluster-rfcs/pull/145) that should hopefully give some background for that.

services/web-server/src/loaders/workerManager.js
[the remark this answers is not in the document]
helfi92: Interesting. I tried locally and I think it seems to work. Anyways, that was only a nit so I'm okay both ways :)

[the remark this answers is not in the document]
owlishDeveloper: @helfi92 I think my PRs get enough suggestions even without nits :) Why don't we automate formatting we don't like?

ui/src/views/WorkerManager/WMViewWorkerPools/WMWorkerPools.graphql
helfi92: The reason why we are adding a newline instead of a comment is because GraphQL self documents itself based on the comments defined above the properties in the graphql definition file. For example:

Screen Shot 2019-06-06 at 1 10 29 PM





Adding information about an extra request didn't seem like a good place to put it as a comment at the time since that didn't really define the purpose of that property. That being said, removing all of the newlines and replacing them with a comment that would help make it clear to developers which properties would trigger an extra request is also another option. I don't mind which method we take but I think we should  aim for consistency. Thoughts?

services/web-server/src/loaders/workerManager.js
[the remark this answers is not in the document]
helfi92: I think both ways are fantastic. Apologies for bringing that up.

ui/src/utils/prop-types.js
helfi92: I think things like `failed` and `exception` are not part of the worker-pool schema. It's also missing a couple of fields like `config` and `lastModified`.

ui/src/views/WorkerManager/WMViewWorkerPools/WMWorkerPools.graphql
owlishDeveloper: My thoughts are:

* Personally, I don't understand newlines. The reason for the newline never occurred to me until you told me why it's there :) I suspect I'm not alone. I view comments as easier to understand than formatting

* Why don't we automate this? I would like our reviews to be more about finding actual bugs, errors in the logic, security problems etc. Why don't we use a style guide and a linter for the rest?

helfi92: > Personally, I don't understand newlines.

I don't as well. Let's change things to use comments. But let's change it everywhere. Maybe we can create an issue to track this?

> Why don't we automate this?

I wish everything was automated :) I don't have the bandwidth to work on this right now but feel free to create an issue or something for this.

helfi92: >  Why don't we use a style guide and a linter for the rest?

We should. I will update the README to have a styleguide section. We could have the newline thing there as a start. Hope that helps.

package.json
imbstack: the taskcluster bits added here are libraries that are used in the services or other libraries. Services aren't imported by any other code so shouldn't be in here.

services/web-server/src/graphql/WorkerManager.graphql
owlishDeveloper: > Are we sure those properties won't change for a different worker pool? Is it possible we want to show a different table depending on the type of worker pool? Do we foresee a future where a user may want to see a certain value returned in config but this value happens not to be relevant for a different worker pool?

But to clear this up: I believe there was a discussion  about this yesterday in the channel. I believe you will find the background information on this, as well as my opinion on this, there in that discussion. In short: yes, we do foresee and we are aware :)

Another thing to explain, you probably notice the word `general` in the name of the type. That was meant to be a type that has some fields that are common for all the providers.

But anyways, since it's not used anyways, I am just removing that type altogether

ui/src/utils/prop-types.js
owlishDeveloper: I am not clear on the connection between the `prop-types` and the schema. However, I am getting rid of those properties in this PR (I listed the intended changes in the PR description), and I forgot to remove them from here. So I removed them.

ui/src/views/WorkerManager/WMViewWorkerPools/WMWorkerPools.graphql
owlishDeveloper: > Maybe we can create an issue to track this?

Sure, I can create an issue

> I don't have the bandwidth to work on this right now but feel free to create an issue or something for this.

Sure. Although sometimes it seems to me you could actually save yourself a lot of time by automating all these things

package.json
owlishDeveloper: https://github.com/taskcluster/taskcluster/pull/843/files#diff-7b7f521247a2c870baae56e97f99da7eR1 (sorry I was assuming you read the code since you had approved it)

services/web-server/src/resolvers/WorkerManager.js
```diff
@@ -1,4 +1,15 @@
+const { splitWorkerPoolId } = require('taskcluster-worker-manager/src/util');
```
djmitche: Oh, I missed this! [We don't import between microservices](https://github.com/taskcluster/taskcluster/blob/master/dev-docs/best-practices/microservices.md#independence).  I think it would be best to just copy this utility, since it's short and straightforward.  If it were more complex, the fix would be to create a new tc-lib-foo library.

ui/src/utils/prop-types.js
helfi92: The input which we are feeding `WorkerManagerWorkerPoolsTable` comes from the `WorkerManagerWorkerPoolSummaries` graphql query. Right now we are only setting the ones that we are requiring in that specific view but technically we could have another view that uses the same query but fetches a different set of fields. Because of that, we should list all of the fields but have them as optional (i.e., they don't have the `.isRequired` specified).

```jsx
export const WorkerManagerWorkerPoolSummary = shape({
  workerPoolId: string,
  providerId: string,
  description: string,
  created: date,
  lastModified: date,
  config: object,
  owner: string,
  emailOnError: bool,
});
```

owlishDeveloper: oh I see. I'll make it match the schema - the only question is `pendingTasks` field ,as it's not in the schema and come from a different service.... I see in the code, the prop-type for aws provisioner does have that field - I'll do the same

owlishDeveloper:
> Note: Tests are failing with SyntaxError: Unexpected token {

oh! thank you!

Issue #598, opened 2019-04-17
[UI] Auth Client UX


helfi92:
Same as https://github.com/taskcluster/taskcluster/issues/581 but for the auth clients and roles view:

Client Form: https://taskcluster-web.netlify.com/auth/clients/github%2Fhelfi92%2F

Role Form: https://taskcluster-web.netlify.com/auth/roles/github-admin%3A*

shubhamgupta2956:
@helfi92 I want to take this issue. Please assign it to me. Thanks.

helfi92:
I think the only thing left would be to add the snackbar confirmation when saving a role/client. Awesome work so far!

shubhamgupta2956:
I will add it and submit a PR for it.Thanks.

helfi92:
Thank you for your work @shubhamgupta2956. Very nice!

PR #844, opened 2019-06-05
Auth_Client_UX_#598

shubhamgupta2956:
#598

### DIFF
--- a/ui/src/components/ClientForm/index.jsx
+++ b/ui/src/components/ClientForm/index.jsx
@@ -191,10 +191,10 @@ export default class ClientForm extends Component {
       disabled,
     } = this.state;
     const isClientDirty =
-      clientId !== client.ID ||
+      isNewClient ||
       description !== client.description ||
       expires !== client.expires ||
-      scopeText !== client.scopeText;
+      scopeText !== client.scopes.join('\n');
 
     return (
       <Fragment>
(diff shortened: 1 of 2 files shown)


ui/src/components/ClientForm/index.jsx
```diff
@@ -184,6 +190,11 @@ export default class ClientForm extends Component {
       expandedScopes,
       disabled,
     } = this.state;
+    const isClientDirty =
+      clientId !== client.ID ||
+      description !== client.description ||
+      expires !== client.expires ||
+      scopeText !== client.scopeText;
```
helfi92: `clientId` is not editable when viewing an existent client so it should be okay to omit that. We also need to make sure this code still works when a user is creating a new client (i.e., when `client === null`). I think we can do the following change:

```jsx
const isClientDirty =
  isNewClient ||
  description !== client.description ||
  expires !== client.expires ||
  scopeText !== client.scopes.join('\n');
```

Note: It's `client.scopes`, not `client.scopeText`

ui/src/components/RoleForm/index.jsx
```diff
@@ -111,6 +117,10 @@ export default class RoleForm extends Component {
       lastModified,
       expandedScopes,
     } = this.state;
+    const isRoleDirty =
+      roleId !== role.ID ||
+      description !== role.description ||
+      scopeText !== role.scopeText;
```
helfi92:
```jsx
const isRoleDirty =
  isNewRole ||
  description !== role.description ||
  scopeText !== role.scopes.join('\n');
```

PR #846, opened 2019-06-06
[web-server] hotfix: check for user before generating jwt

helfi92:


djmitche:
I'm curious how we got an accessToken for which we couldn't generate creds..

helfi92:
We didn't. `web-server` crashed...

Issue #534, opened 2019-04-03
[UI] Streamline trigger-hook flow


mitchhentges:
From my usage of the Taskcluster UI so far, I have triggered hooks hundreds of times, but I've never supplied a custom `context`. Therefore, I'm making the broad assumption that, perhaps, supplying a `context` is an edge case?

If so, perhaps the hook-triggering UI can be optimized for the case where the `context` isn't edited. I'd love to avoid the think-and-click that happens every time the hook-trigger dialog comes up.

(One method would be to have two buttons: trigger hook, and trigger hook (with options) - though this has downsides in usability and complexity. :man_shrugging: )

What do you think? :thinking:

helfi92:
I think we can have the initial focus be on the "Trigger Hook" button when opening the modal assuming that most hooks are triggered with an empty context. This would allow people to press the enter key right after the trigger hook dialog opens up.

shubhamgupta2956:
@helfi92 I want to take this issue. Please assign it to me. Thanks

shubhamgupta2956:
@helfi92 what I can understand in it is that we want the initial focus on the "TRIGGER HOOK" button in the dialog that opens so that the user just can press enter to trigger hook. But this thing is implemented in the code. Can you please explain me what we want to solve in this issue.

helfi92:
> what I can understand in it is that we want the initial focus on the "TRIGGER HOOK" button in the dialog that opens so that the user just can press enter to trigger hook.

That's right!

> But this thing is implemented in the code. Can you please explain me what we want to solve in this issue. I am a bit confused.

Pressing "enter" might trigger a hook but the user doesn't know that. We want to show the ripple in the button to let the user know that the button is focused:

![Screen Shot 2019-06-25 at 7 06 54 PM](https://user-images.githubusercontent.com/3766511/60139616-8d11be80-977c-11e9-8afd-eccdf768c479.png)

helfi92:
How is it going @shubhamgupta2956? Let me know if you had any questions :)

shubhamgupta2956:
Sorry for inactivity @helfi92.

I had tried to search for the problem. Not sure but it seems that when we open dialog action, it takes it like the dialog box was in the same page and thus disables ripple effect. I am still confused about it and searching for proper solution.

shubhamgupta2956:
@helfi92 am I going in correct direction. Is my assumption correct as I can hardly find any solution for it. Please suggest me in what direction should I go. Thanks.

[...]

Issue #497, opened 2019-03-26
[UI] Hooks list focus state


helfi92:
STR:

1. https://taskcluster-web.netlify.com/hooks
2. Press the tab key many times until the focus is on the first element of the hooks list

**Expected**: The focus effect background color looks like

![Screen Shot 2019-03-26 at 7 55 35 AM](https://user-images.githubusercontent.com/3766511/54995215-9718fb00-4f9c-11e9-9477-6a3fe1a85f90.png)

**Actual**:

![Screen Shot 2019-03-26 at 7 54 40 AM](https://user-images.githubusercontent.com/3766511/54995216-9718fb00-4f9c-11e9-83bc-a553debb9178.png)

VirmaniDivya:
Hi! Can I take up this issue, please?

shubhamgupta2956:
@helfi92 I would like to take this issue.
One more thing, I am having difficulty in running web-services due to which `Network error: Failed to fetch` is coming when I open hooks page in my setup. Also when I start web services with `yarn start` it gives this output:
`$ NODE_ENV=development node src/main.js devServer


No Pulse namespace defined; no Pulse messages will be received.
Error: Unable to use "github" login strategy without taskcluster clientId and accessToken
    at Github.useStrategy (/home/savitar/opensource/taskcluster/services/web-server/src/login/strategies/github.js:84:13)
    at Object.values.forEach.strategy (/home/savitar/opensource/taskcluster/services/web-server/src/servers/createApp.js:39:14)
    at Array.forEach (<anonymous>)
    at module.exports (/home/savitar/opensource/taskcluster/services/web-server/src/servers/createApp.js:38:29)
    at setup (/home/savitar/opensource/taskcluster/services/web-server/src/main.js:105:39)
    at Promise (/home/savitar/opensource/taskcluster/libraries/loader/src/index.js:149:33)
    at new Promise (<anonymous>)
    at loaded.(anonymous function).Promise.all.then.deps (/home/savitar/opensource/taskcluster/libraries/loader/src/index.js:147:18)
    at process._tickCallback (internal/process/next_tick.js:68:7)
    at Function.Module.runMain (internal/modules/cjs/loader.js:757:11)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.`
[...]

PR #850, opened 2019-06-06
snackbar confirmation for client and role form

shubhamgupta2956:
#598


shubhamgupta2956:
Thanks @helfi92. Can you please suggest me what should I start next.

helfi92:
Next, I think something like https://github.com/taskcluster/taskcluster/issues/497 or https://github.com/taskcluster/taskcluster/issues/534 would be interesting. The second issue involves adding the `autoFocus` prop similar to https://github.com/taskcluster/taskcluster/blob/a60f57c3215c46022a3f0d6f59834bb9134359c1/ui/src/components/SignInDialog/CredentialsDialog.jsx#L72 and then checking that it works. Information about what `autoFocus` does can be found in https://material-ui.com/api/text-field/#props. Feel free to comment directly on the issue you're interested most. Thanks!

PR #851, opened 2019-06-06
[web-server] Add initial styleguide

helfi92:
My hope is for this styleguide to stay minimal :)

Relates to https://github.com/taskcluster/taskcluster/pull/843#discussion_r291286012.

### DIFF
--- a/dev-docs/best-practices/microservices.md
+++ b/dev-docs/best-practices/microservices.md
@@ -124,12 +124,23 @@ Instead, use `taskcluster-lib-monitor` as described in its documentation.
 
 ### taskcluster-web-server
 
-#### Contents
+#### GraphQL Schemas
 
-* [Schema Field Definitions](#schema-field-definitions)
-
-#### Schema Field Definitions
-
-* For fields that may trigger an additional request for a client,
+For fields that may trigger an additional request for a client,
 add a comment above the field definition of the form "Note: This field will trigger an additional request."
-This will make sure developers are aware of possible performance penalties.
+This will make sure developers are aware of possible performance penalties. For example:
+
+```graphql
+type Hook {
+  hookGroupId: ID!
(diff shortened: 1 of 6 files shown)


helfi92:
Screen Shot 2019-06-06 at 3 48 49 PM

services/web-server/README.md
```diff
@@ -271,3 +271,16 @@ Variables:
...
+### Contents
+
+* [Schema Field Definitions](#schema-field-definitions)
+ 
+### Schema Field Definitions
+
+* For fields that may trigger an additional request for a client,
+add a comment above the field definition of the form "Note: This field will trigger an additional request."
+This will make sure developers are aware of possible performance penalties.
+
```
djmitche: It would probably make sense to put this information with the other best-practices in `dev-docs` in the root.

dev-docs/best-practices/microservices.md
```diff
@@ -121,3 +121,15 @@ Be friendly and document the errors in the API's `description` property, as they
...
 *Do not use* `taskcluster-lib-stats` or `raven`.
 Instead, use `taskcluster-lib-monitor` as described in its documentation.
+
+### taskcluster-web-server
+
+#### Contents
+
+* [Schema Field Definitions](#schema-field-definitions)
+
+#### Schema Field Definitions
```
djmitche: To fit into the rest of the document, I think this will make sense as a `### GraphQL Schemas` section under `## Implementation`

helfi92: Done but I suspect there may be more changes in the future so perhaps a `taskcluster-web-server` section is more appropriate. Let me know what you think.

djmitche: Sure, that makes senes, too -- whatever fits with the flow of the document.

PR #855, opened 2019-06-07
Use workers for tracking operations rather than workerPools

imbstack:
This should avoid us running into issues with the size of entities. It also makes things simpler!

This also lays the groundwork for some other improvements I want to make.


services/worker-manager/src/providers/google.js
```diff
@@ -40,20 +39,17 @@ class GoogleProvider extends Provider {
...
     this.configSchema = 'config-google';
-    this.fake = fake;
 
     this.instancePermissions = instancePermissions;
     this.project = project;
     this.zonesByRegion = {};
 
-    // TODO: Make fakes be injected
-    if (fake) {
+    if (fakes['google']) {
```
djmitche: `fakes.google` would work here too..

services/worker-manager/src/main.js
```diff
@@ -181,11 +181,17 @@ let load = loader({
     }),
   },
 
+  // This is used in testing to inject provider fakes
+  externalFakes: {
```
djmitche: What does "external" mean here?

imbstack: I was using it to mean "cloud service provider" I guess. Very open to suggestions for better names.

djmitche: fakeCloudApis?

PR #857, opened 2019-06-07
Bug 1557487 - fix /references/references.json, add docs

djmitche:
Bugzilla Bug: [1557487](https://bugzilla.mozilla.org/show_bug.cgi?id=1557487)

Docs, except the last paragraph, were already reviewed in #852.


djmitche:
I tested this by building an image and running the `references/web` process.

PR #863, opened 2019-06-10
Bug 1555490 - implement yarn release

djmitche:
Bugzilla Bug: [1555490](https://bugzilla.mozilla.org/show_bug.cgi?id=1555490)

The first major release would be 14.1.0, since tc-client is already at 14.0.0.

This doesn't handle:
 * making a github release
 * pushing client packages to various packaging things
 * determining appropriate level and constructing changelog


infrastructure/builder/src/generate/generators/node-version.js
```diff
@@ -1,4 +1,4 @@
-const {readFile, modifyFile, modifyJSON} = require('../util');
+const {readRepoFile, modifyRepoFile, modifyRepoJSON} = require('../../utils');
```
imbstack: This is a great move

PR #869, opened 2019-06-13
Bug1558713 - add endpoint to list workers for a given workerpool

owlishDeveloper:
Bugzilla Bug: [1558713](https://bugzilla.mozilla.org/show_bug.cgi?id=1558713)

### DIFF
--- a/services/worker-manager/schemas/v1/worker-full.yml
+++ b/services/worker-manager/schemas/v1/worker-full.yml
@@ -1,0 +1,73 @@
+$schema: "/schemas/common/metaschema.json#"
+title: Worker Full Definition
+description: A complete worker definition.
+type: object
+properties:
+
+  workerPoolId:
+    title: Worker Pool ID
+    type: string
+    # Note that the portion after the `/` is quite severely limited in hopes it works with
+    # most cloud providers
+    pattern: '^[a-zA-Z0-9-_]{1,38}/[a-z]([-a-z0-9]{0,36}[a-z0-9])?$'
+    description: |
+      The ID of this worker pool (of the form `providerId/workerType` for compatibility)
+
+  workerGroup:
+    title: Worker group
+    type: string
+    description: Worker group to which this worker belongs
+    minLength: {$const: identifier-min-length}
+    maxLength: {$const: identifier-max-length}
+    pattern: {$const: identifier-pattern}
+
+  workerId:
+    title: Worker ID
+    type: string
+    description: Worker ID
+    minLength: {$const: identifier-min-length}
+    maxLength: {$const: identifier-max-length}
+    pattern: {$const: identifier-pattern}
+
+  providerId:
+    title: Provider
+    type: string
+    # note that this is typically used as the workerGroup for workers,
+    # so its format should match that for workerGroup.
+    minLength: {$const: identifier-min-length}
+    maxLength: {$const: identifier-max-length}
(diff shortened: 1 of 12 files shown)


services/worker-manager/schemas/v1/worker-full.yml
```diff
@@ -0,0 +1,73 @@
+$schema: "/schemas/common/metaschema.json#"
+title: Worker Full Definition
+description: A complete worker definition.
+type: object
+properties:
+
+  workerPoolId:
```
imbstack: This can also be a `$ref` to `worker-pool-full` I think. Either that or just `$ref` things like `type`/`pattern` and make the description different here.

services/worker-manager/schemas/v1/worker-full.yml
```diff
@@ -0,0 +1,73 @@
...
+    type: string
+    # note that this is typically used as the workerGroup for workers,
+    # so its format should match that for workerGroup.
+    minLength: {$const: identifier-min-length}
+    maxLength: {$const: identifier-max-length}
+    pattern: {$const: identifier-pattern}
+    description: |
+      The provider responsible for managing this worker.
+
+      If this value is `"null-provider"`, then the worker pool is pending deletion
```
imbstack: I don't think a worker will ever have a provider of `null-provider` since it will never create a worker.

services/worker-manager/schemas/v1/worker-full.yml
```diff
@@ -0,0 +1,73 @@
...
+
+  created:
+    title: Created
+    description: Date and time when this worker was created
+    type: string
+    format: date-time
+
+  expires:
+    title: Expires
+    description: Date and time when this worker will expire
```
imbstack: Let's make sure to say here that this is not when the worker stops working but when it is deleted from the db.

services/worker-manager/schemas/v1/worker-full.yml
owlishDeveloper: you mean I should reference one field from another schema? I was looking at the `data.js`, it says `workerPoolId: Entity.types.String`, so it doesn't seem to me like you're asking me to reference the whole object

owlishDeveloper: like `{$ref: "worker-pool-full.json#/properties/workerPoolId"}`?

services/worker-manager/schemas/v1/worker-full.yml
```diff
@@ -0,0 +1,62 @@
...
+
+  providerId:
+    title: Provider
+    type: string
+    # note that this is typically used as the workerGroup for workers,
+    # so its format should match that for workerGroup.
+    minLength: {$const: identifier-min-length}
+    maxLength: {$const: identifier-max-length}
+    pattern: {$const: identifier-pattern}
+    description: The provider responsible for managing this worker.
```
djmitche: This confused me at first, because I thought "wait, the workerPoolId should define the providerId".  But as you know that's wrong -- the worker might have been started by a different provider that has since been changed.

What would you think about adding a bit to the description to help readers avoid my confusion?

owlishDeveloper: oh, that's a good point. I did wonder about the data duplication but didn't ask about it for some reason. Yeah, I'll add that info to the description

PR #871, opened 2019-06-14
Bug 1551607 - use docker volumes for building

djmitche:
This requires a good deal more manual copying back and forth, but means
that the `yarn install` operations run entirely on a docker volume,
rather than over a bind mount.  Bind mounts don't work reliably on
Docker for Mac, so this change allows `yarn build` to operate correclty
on such systems.

Bugzilla Bug: [1551607](https://bugzilla.mozilla.org/show_bug.cgi?id=1551607)

### DIFF
--- a/infrastructure/builder/src/build/index.js
+++ b/infrastructure/builder/src/build/index.js
@@ -55,7 +55,6 @@ class Build {
       return;
     }
     const context = await taskgraph.run({'build-can-start': true});
-    console.log('DONE');
 
     console.log(`Monoimage docker image: ${context['monoimage-docker-image']}`);
     if (!this.cmdOptions.push) {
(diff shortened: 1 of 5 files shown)


djmitche:
I want to test this on my local mac before getting review.

djmitche:
Great success on the mac, on the first try.  It's still godawful slow (870 seconds) but it works.

djmitche:
So, the slowest bit is "Copying /app out of docker volume".  A decent followup would be to try copying a single file (such as a tarball) instead of the fully-unpacked `app` directory, and see if that's faster.  Still, this *works* so I'm not going to touch it.

infrastructure/builder/src/build/index.js
```diff
@@ -55,6 +55,7 @@ class Build {
       return;
     }
     const context = await taskgraph.run({'build-can-start': true});
+    console.log('DONE');
```
imbstack: Is this stray?

djmitche: I was just testing you?  Yeah, thanks :)

PR #872, opened 2019-06-14
Bug 1559471 - use JWTs with HS256 for logins

djmitche:
This also adds configuration to terraform to invent an HMAC secret.

The *generation* of these JWTs is tested to work locally.  I can't test more than that because the `callback.ejs` page has `window.opener = null`, so it doesn't actually call back.  I'll try this in dustin-dev as well to see if it works better there.  @helfi92 is that something you've seen in your local testing of signins?

Bugzilla Bug: [1559471](https://bugzilla.mozilla.org/show_bug.cgi?id=1559471)

### DIFF
--- a/infrastructure/terraform/web-server-service.tf
+++ b/infrastructure/terraform/web-server-service.tf
@@ -10,6 +10,10 @@ resource "random_string" "web_server_access_token" {
   override_special = "_-"
 }
 
+resource "random_string" "web_server_jwt_key" {
+  length = 66
+}
+
 module "web_server_secrets" {
   source            = "modules/service-secrets"
   project_name      = "taskcluster-web-server"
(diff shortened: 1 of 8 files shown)


djmitche:
Hm, so I'm still seeing this in dustin-dev, too -- the "Logging In..." tab stays open with a white background, and if I type `window.opener` in the console it tells me `null`.

djmitche:
I'll file a separate bug for that.  I used `window.open("/login/github")` in the console to open the login screen, and the whole thing worked just fine.  So I think this patch is in good shape.

infrastructure/terraform/web-server-service.tf
```diff
@@ -10,6 +10,10 @@ resource "random_string" "web_server_access_token" {
   override_special = "_-"
 }
 
+resource "random_string" "web_server_jwt_key" {
+  length = 66
```
helfi92: Is there a reason behind a length of 66?

djmitche: I copied it from `auth-service.tf`.  I don't know why I chose 66 there!

djmitche:
Let's move the `window.opener` discussion to #874

PR #897, opened 2019-06-24
Bug 1558345 - move provider-related endpoints to provider source file

djmitche:
Bugzilla Bug: [1558345](https://bugzilla.mozilla.org/show_bug.cgi?id=1558345)

This moves some code around, and then ensures that every provider type implementation is loaded, even if it's not in the config, so that the API definition does not depend on the config.

It also adds an extra check that the provider type is correct.  I wonder if we should pass the providerId in userData and have the worker include that in the reply?


imbstack:
I'm hoping to standardize what gets passed to workers when I do the packet provider. I'll keep that in mind.

PR #898, opened 2019-06-24
Bug 1544917 - limit access to web-server using CORS

djmitche:
This should work OK for local development, PRs, and on https://taskcluster-ui.herokuapp.com.

The intent is to set ADDITIONAL_ALLOWED_CORS_ORIGIN to `/https://deploy-preview-\\d+--taskcluster-web\\.netlify\\.com/` in whatever deployment we choose to aim the deploy previews at (ideally staging, at least soon).

Bugzilla Bug: [1544917](https://bugzilla.mozilla.org/show_bug.cgi?id=1544917)

### DIFF
--- a/services/web-server/config.yml
+++ b/services/web-server/config.yml
@@ -13,6 +13,15 @@ defaults:
     # Port to listen for requests on
     port: !env:number PORT
 
+    # The CORS allowed origins (passed to https://yarnpkg.com/en/package/cors);
+    # typically this is the root URL, but in testing it may have other values.
+    # The legacy root URL is special-cased.  Values beginning with `/` are
+    # treated as regular expressions.  ADDITIONAL_ALLOWED_CORS_ORIGIN is used
+    # to allow UI deploy previews to access a running deployment.
+    allowedCORSOrigins:
+      - !env TASKCLUSTER_ROOT_URL
+      - !env ADDITIONAL_ALLOWED_CORS_ORIGIN
+
   # Configuration of access to other taskcluster components
   taskcluster:
     rootUrl: !env TASKCLUSTER_ROOT_URL
@@ -62,6 +71,12 @@ development:
   server:
     # taskcluster-ui's dev server assumes port 3050
     port: 3050
+    # for testing, allow any origin
+    allowedCORSOrigins: [true]
+
+test:
+  server:
+    port: 63821
 
 production:
   pulse:
--- a/services/web-server/src/servers/createApp.js
+++ b/services/web-server/src/servers/createApp.js
@@ -11,7 +11,18 @@ module.exports = async ({ cfg, strategies }) => {
 
   app.set('view engine', 'ejs');
   app.set('views', 'src/views');
-  app.use(cors());
+
+  const allowedCORSOrigins = cfg.server.allowedCORSOrigins.map(o => {
+    if (typeof(o) === 'string' && o.startsWith('/')) {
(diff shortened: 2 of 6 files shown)


services/web-server/test/server_test.js
```diff
@@ -0,0 +1,45 @@
...
+
+      suiteTeardown(function() {
+        helper.load.restore();
+      });
+
+      helper.withServer(false, () => false);
+
+      test('request', async function() {
+        const res = await request
+          .options(`http://localhost:${helper.serverPort}/graphql`)
```
helfi92: What does `.options` do? I can't seem to find the relevant section in http://visionmedia.github.io/superagent/.

services/web-server/config.yml
```diff
@@ -61,6 +64,12 @@ development:
   server:
     # taskcluster-ui's dev server assumes port 3050
     port: 3050
+    # for testing, allow any origin
+    allowedCORSOrigin: '*'
```
helfi92: `allowedCORSOrigin` doesn't seem to be used. Did you mean `allowedOrigin`?

services/web-server/test/server_test.js
djmitche: It makes an `OPTIONS /graphql` request, similar to what a browser would do for POST (I think..).

djmitche:
Thanks -- it turns out this review was for an old push (I must have missed a `-f` in a force push).  The updated version handles the netlify addresses changes more smoothly.  Want to have another look?

helfi92:
Awesome. I'll look in a bit.

djmitche:
Hm, does that work on master?  That seems like the `loginMiddleware` is malfunctioning somehow (since that's the only "View" involved), and this patch doesn't change it.

I don't have local auth0 credentials set up, but I can successfully login with github, and that uses the same EJS template.

djmitche:
I've pushed a rebased version, too (that uses jwt.key instead of jwt.{public,private}Key)

helfi92:
> I've pushed a rebased version, too (that uses jwt.key instead of jwt.{public,private}Key)

That seemed to have been the issue :)

djmitche:
Lol, great :D

djmitche:
Oh, argh!! I pushed the old version again.  Grr..  Switching between laptop and development host is not going well on this PR.

djmitche:
I also tried to repro with the wrong branch.  Sorry, sorry, sorry.

djmitche:
OK, reproduced.

services/web-server/src/servers/createApp.js
```diff
@@ -9,9 +9,16 @@ const credentials = require('./credentials');
 module.exports = async ({ cfg, strategies }) => {
   const app = express();
 
-  app.set('view engine', 'ejs');
-  app.set('views', 'src/views');
```
djmitche: OK, this was the problem.  Sorry, this PR is a hot mess :(

djmitche:
OK, this revision, d34e36b8f091810bbb5748332b68dae22f374022, allowed me to login to github and to view a bunch of hooks.  I *think* everything is working.  One more look?  The only thing I changed since your [first review yesterday](https://github.com/taskcluster/taskcluster/pull/898#pullrequestreview-255386395) was what I noted above ("OK, this was the problem").

djmitche:
I'll add a changelog in a push to master.

PR #905, opened 2019-06-24
Bug 1560649 - import go client into monorepo

djmitche:
This
 * keeps the existing revision history
 * supports regeneration
 * reads from generated/references.json instead of a running deployment
 * uses go modules

However, it does not
 * <s>run automatically in `yarn generate`</s>
 * support the old import paths
 * support go versions that do not support go modules

To those points:
 * I'm undecided about how best to support this.  It seems like expecting both Node-x.y.z *and* goP.Q is the right way to go, but that's annoying both for developers (who don't use gvm) and in CI (where we'll need to bake and maintain our own docker image (currently we just use the upstream `node:x.y.z` docker image)).  Another option would be to build a `taskcluster-client-go-generator` binary and automatically download that or ask users to download it.  But this might start to seem silly when we also import taskcluster-cli.  We should probably just require go be installed.
 * I don't see a way to easily "redirect" users of the old paths to the new; perhaps it's best to just require that users of the old clients switch to the new path on update?  We could even keep the old repo alive for a while and apply updates to it.
 * I think we can support old versions of go by vendoring (`go mod vendor`).  It'd be nice to avoid that though.  Could we just require go1.12 and leave older go versions behind?

@petemoore  @milescrabill any advice on these three points?

Bugzilla Bug: [1560649](https://bugzilla.mozilla.org/show_bug.cgi?id=1560649)


djmitche:
By the way, I don't think it's practical to do the generation in JS.  While porting jsonschema2go would be possible, the generation process also uses `go fmt` and `goimports` which are not so practical to port.

petemoore:
I'm comfortable with the monorepo taskcluster go client generator (and generated client) to be incompatible with the former go client. There are some things that could be improved in the client that would require breaking changes, so this seems like a good opportunity to do so, if we are going to move the go package to the monorepo and thus change the import path in a breaking way.

For example, I would prefer API methods to return an intermediary object, rather than the results of the API call, e.g. rather than

```
queue := tcqueue.NewFromEnv()
result, err := queue.CreateTask(...)
```

I think it would be nicer to have
```
queue := tcqueue.NewFromEnv()

// support cancellation with explicit context field in API request
ctx, _ := context.WithTimeout(ctx, 3*time.Second)

// execute API query directly
result, err := queue.CreateTask(...).Exec(ctx) 

// get signed url (only available on API calls that support it)
result, err := queue.GetArtifact(...).SignedURL() 

// capture a trace of the http request(s)/response(s) made
result, trace, err := queue.CreateTask(...).Trace(ctx)

// fetch the HTTP request to make the HTTP call (including Authorization header)
// without actually making the request
request, err := queue.CreateTask(...).HTTPRequest() 
```

I don't foresee any reason that the former tool client generator is likely to break in the near future, so I am happy to leave it updating in my crontab, and we could make the new client a v2 rewrite with more bells and whistles.

In other words, I'm fine for you to import taskcluster-client-go into the monorepo, and leave the old version running, but i'm also happy for it to be rewritten (e.g. in node.js) to achieve the above goals, and to leave the old one in place until we have a solid replacement that users can migrate to.

djmitche:
Per chat conversation with @petemoore 
* Import now with the ability to make breaking changes later -- users should be using `go mod` or vendoring to pin versions anyway, so breaking changes are just a major version bump
* Keep the old repo active as suggested, with a note in the README pointing users to the new version (once it lands and is stable)
* Require go1.12 (for `go mod`), but without introducing code changes that would make it incompatible with go1.10 (which is what generic-worker is using right now).  If we need to use the new client with generic-worker and that's still at go1.10, we can use `go mod vendor` and add some tests for go1.10/go1.11.

infrastructure/builder/src/utils/repo.js
```diff
@@ -57,17 +57,24 @@ exports.writeRepoJSON = async (filename, data) => {
...
   const contents = await readFile(
     path.join(REPO_ROOT, filename),
     {encoding: 'utf8'});
   const modified = await modifier(contents);
   await writeFile(filename, modified, {encoding: 'utf8'});
 };
 
+let modifyRepoPromise = Promise.resolve();
+exports.modifyRepoFile = (filename, modifier) => {
+  modifyRepoPromise = modifyRepoPromise.catch(() => {}).then(() => modifyRepoFile(filename, modifier));
```
owlishDeveloper: what is the purpose of `modifyRepoPromise`?

infrastructure/docker-images/build.sh
```diff
@@ -7,6 +7,14 @@ if [ -z "${node_version}" ]; then
     exit 1
 fi
 
+
+go_version=$(<${repo_root}/.go-version)
```
owlishDeveloper: we build two images with one command? If one of the `docker build` fails - I presume we still get the second image? But then if we have to re-run the command the second image gets built again?

infrastructure/builder/src/utils/repo.js
djmitche: It's serializing the promises -- there is a comment up above the function describing it.  I ran into a bug where two things were calling `modifyRepoFile('.taskcluster.yml', ..)` and overwriting each other.  But this is kind of a mess.  Maybe I can find a library for it?  We do this a few other places, too.  I've made a followup bug for that (https://bugzilla.mozilla.org/show_bug.cgi?id=1561668)

infrastructure/docker-images/build.sh
djmitche: This is the third one, actually.  It probably makes sense to split them into separate scripts as, yes, it will rebuild the other images and end up with multiple different images with the same tag.  I'll fix that up in another push and then land.  Thanks!

These builds don't take too long, as they're mostly just based on existing images.

PR #907, opened 2019-06-25
[github] Allow `.taskcluster.yml` to set `schedulerId`.

### LINKED ISSUE #748: feature request: taskcluster-github should allow for a custom schedulerId per repo
Afaict, taskcluster-github hardcodes `taskcluster-github` as the `schedulerId` for tasks. When using [ci-taskgraph](https://hg.mozilla.org/ci/taskgraph), ci-taskgraph creates the graph with the `schedulerId` [`{trust-domain}-level-{level}`](https://hg.mozilla.org/ci/taskgraph/file/tip/src/taskgraph/create.py#l34), which breaks.

I'm currently using [this workaround](https://phabricator.services.mozilla.com/D31542), but @tomprince would prefer we change taskcluster-github to pick up the `taskcluster.ci.config.scheduler-id` as the `schedulerId`, rather than always using `taskcluster-github`.

What do you think?

```
14:39 <aki> have you seen this before? https://www.irccloud.com/pastebin/FHjNzKj7/
14:40 <aki> i suppose the decision task has a taskcluster-github schedulerId
15:01 <tomprince> aki: Yeah, you need to set it explicitly.
15:01 <aki> https://phabricator.services.mozilla.com/D31542 addresses it, though i'm not sure that's the wanted fix
15:02 <tomprince> aki: I think you want to set the scheduler-id in the .taskcluster.yml to mach what taskgraph expects.
15:03 <aki> can do. i think i copied it from the redo PR
15:04 <tomprince> Yeah. Need to think about how to handle things. I think for major projects, we want to give a new trust-domain, which comes along with a scheduler-id. But for tiny things like redo, that might not be worth it.
15:09 <aki> the decision task is still being created with schedulerId taskcluster-github. not sure if that's based on taskcluster/ci/config.yml on master, or if tc-gh overrides it, or what
15:09 <aki> https://github.com/escapewindow/xpitest-manifest/pull/2
15:10 <aki> worked with the taskgraph scheduler_id fix, broken with the config.yml scheduler-id fix
15:17 <tomprince> Hmm. I guess maybe tc-github overrides it. I'd rather get that changed than change taskgraph, if possible, though.
```

tomprince:
Fixes #748.

### DIFF
--- a/changelog/github-scheduler.md
+++ b/changelog/github-scheduler.md
@@ -1,0 +1,3 @@
+level: patch
+---
+With the proper scopes, github repositories can now override the default scheduler. This only works with experimental `checks` status reporting.
--- a/changelog/github-scheduler.md
+++ b/changelog/github-scheduler.md
@@ -1,3 +1,3 @@
 level: patch
 ---
-With the proper scopes, github repositories can now override the default scheduler. This only works with experimental `checks` status reporting.
+With the proper scopes, github repositories can now override the default scheduler. Adding custom schedulerId to the task definition while using github's Statuses API might break the status reporting functionality of tc-github in the case of successful build. Therefore, this only works with experimental `checks` status reporting.
--- a/services/github/src/tc-yaml.js
+++ b/services/github/src/tc-yaml.js
@@ -282,8 +282,8 @@ class VersionOne extends TcYaml {
         taskMap[taskId] = {
           taskId,
           task: {
-            ...taskWithoutTaskId,
             schedulerId: cfg.taskcluster.schedulerId,
+            ...taskWithoutTaskId,
           },
         };
       });
--- a/services/github/test/tc-yaml_test.js
+++ b/services/github/test/tc-yaml_test.js
@@ -91,7 +91,7 @@ suite(testing.suiteName(), function() {
       assume(config.tasks[1].taskId).to.not.equal(config.tasks[1].task.taskGroupId);
     });
 
-    test('compileTasks forces schedulerId, but uses user-supplied taskId/taskGroupId', function() {
+    test('compileTasks uses user-supplied taskId/taskGroupId/schedulerId', function() {
       const config = {
         tasks: [{
           taskId: 'task-1',
@@ -109,7 +109,7 @@ suite(testing.suiteName(), function() {
         task: {
           created: now,
           taskGroupId: 'tgid-1',
-          schedulerId: 'test-sched',
+          schedulerId: 'my-scheduler-id',
           routes: ['statuses-queue'],
         },
       }, {
@@ -117,7 +117,7 @@ suite(testing.suiteName(), function() {
         task: {
           created: now,
           taskGroupId: 'tgid-2',
-          schedulerId: 'test-sched',
+          schedulerId: 'my-scheduler-id',
           routes: ['statuses-queue'],
         },
       }]);
--- a/ui/docs/reference/integrations/github/taskcluster-yml-v1.md
+++ b/ui/docs/reference/integrations/github/taskcluster-yml-v1.md
@@ -89,7 +89,12 @@ The result looks like this:
 }
 ```
 
-Taskcluster-Github will set the schedulerId of each task, as is required for proper status tracking of the resulting task.
+Taskcluster-Github will set the `schedulerId` of each task, as is required for proper status tracking of the resulting task.
+With the change , it became possible to set a custom `schedulerId` in the task definition, provided you have the scopes
+for using that particular `schedulerId`. However, adding custom schedulerId to the task definition while using github's 
+Statuses API can break the status reporting functionality of tc-github in the case of successful build. 
+Checks API implementation in tc-github is currently in experimental state.
+
 
 The `taskId` and `taskGroupId` properties can be set by the JSON-e template,
 but default values are also available.  If the JSON-e rendering produces only


djmitche:
As a reminder, please add a `changelog/` entry for this update.

owlishDeveloper:
Apologies this taking me long to get to. Will review ASAP

owlishDeveloper:
Apologies it took me so long to get to this! 

The queue event `taskGroupResolved` is currently bound to the handlers by `schedulerId`. We set up the handlers and the bindings once during the deploy process, and we grab the `schedulerId` from the github config. If this change is merged as it is, people who change the `schedulerId` in their `taskcluster.yml` won't get the status of the build posted if there are no exceptions or failures in the build.

Relevant code to look at: [click to see the handlers' setup function](https://github.com/taskcluster/taskcluster/blob/master/services/github/src/handlers.js#L81-L192)

owlishDeveloper:
To implement this change (add ability to set `schedulerId` in `.taskcluster.yml` - or elsewhere) we either have to find some other bindings, or we have to make it possible to create the exchanges dynamically (reacting to a push event to a branch and examine the change and then create the exchange if that was a change of `schedulerId`? Have a admin-sort of UI in the tools website to let people set the scheduler  Id for their repo (rather than in `.taskcluster.yml`)? etc. - there're lots of possible ideas. It seems to me like this is something that requires some thinking

tomprince:
Looking at the code you linked, only the commit status method of reporting results is affected by this. The checks reporting code doesn't use the scheduler-id, so wouldn't be impacted.

This could be a little bit of a footgun if it gets set for a project using the status reporting. However, I generally get the impression taskcluster allows you to whatever you want, rather than having strong guard rails. In any case, unless permissions have explicitly been granted to a repo to use a different scheduler, it will fail to create any tasks because of the scopes.

So, given that you can report results using the checks API, and that one need to go out of ones way to be able to specify a non-default scheduler, allowing this seems low risk.

owlishDeveloper:
> Looking at the code you linked, only the commit status method of reporting results is affected by this. The checks reporting code doesn't use the scheduler-id, so wouldn't be impacted.

Yes, that's exactly what I was saying above. Checks are not finished nor deployed to production yet.

>a project using the status reporting.

...which is 99.9999% of projects at the moment

owlishDeveloper:
Can you please remind me what do you need this feature for?

tomprince:
> > Looking at the code you linked, only the commit status method of reporting results is affected by this. The checks reporting code doesn't use the scheduler-id, so wouldn't be impacted.
> 
> Yes, that's exactly what I was saying above. Checks are not finished nor deployed to production yet.

It looks like it is being used by this repo. What do you mean by "not deployed to production"? There may be improvements that can be made, but they are usable today.

> 
> > a project using the status reporting.
> 
> ...which is 99.9999% of projects at the moment

Does it matter how many projects are using it, if the feature is useful to those that do? In any case, I hope to have 90+% of the github projects on the firefox-ci cluster using this by the end of H2, if not sooner.

> Can you please remind me what do you need this feature for?

I want this to be able to use taskgraph with github projects. taskgraph is opinionated about the scheduler-id it uses, and it is reasonable to do so, since it provides us more control who can add and manage tasks in the taskgroup, which is important for stuff that we ship to end-users.

owlishDeveloper:
> What do you mean by "not deployed to production"? There may be improvements that can be made, but they are usable today.

I mean that they haven't been rolled out to users yet. We (tc team) use it to alpha-test the feature. Checks are not currently usable because the status of the tasks created by decision task is not reported. We don't have a decision task in our repo (yet), that's why we are able to use it.

djmitche:
That should probably be in the `taskcluster-yml-v1.md` file, too?

owlishDeveloper:
> That should probably be in the taskcluster-yml-v1.md file, too?

I don't think so because that is part of the task definition which is not covered by that schema

@tomprince can you please add a copy-paste of the phrase about potential caveat of the change from [this comment](https://github.com/taskcluster/taskcluster/pull/907#pullrequestreview-259656796)? I'd like the wording to be clearest possible

tomprince:
> That should probably be in the `taskcluster-yml-v1.md` file, too?

I don't see anything in there describing the structure of tasks. I do see a mention of tc-github setting the scheduler-id (which it will still do if not provided), but I don't know if it is worth updating that, particularly given @owlishDeveloper's concerns about usability with statuses.

owlishDeveloper:
> That should probably be in the taskcluster-yml-v1.md file, too?

Yes, I agree. @tomprince can you please also duplicate the changelog in the said file, so that the users are aware? In case not everybody reads changelogs. Also, in the `taskcluster-yml-v1.md` can you please mention the fact that Checks functionality of the taskcluster-github is currently in alpha state. Let's make sure that in the case of accidental discovery of the fact that there can be a custom `schedulerId` and scopes requests people have some information regarding the caveat.

owlishDeveloper:
> I don't know if it is worth updating that

I think it's worth it, I agree with Dustin here

owlishDeveloper:
> particularly given @owlishDeveloper's concerns about usability with statuses.

(Checks, not Statuses ) That's the primary reason I agree with Dustin and would like to see this mentioned in the docs: I'd like people to have easier access to the information about the caveats: a) that statuses can be broken and b) that using Checks won't make for a great user experience at this point

PR #912, opened 2019-06-25
Switch to c8 for coverage

imbstack:
I believe this will accurately report coverage across the entire monorepo.


djmitche:
/me :clap: with glee at the new shiny and all its promise

imbstack:
+13% coverage!

imbstack:
Ok, now that I look at this again, it is still stripping other monorepo package stuff from the report. It is in the raw coverage data now. This does seem to do a better job at reporting all coverage though?

djmitche:
Interesting that the numbers are so different with this approach, but still ignoring coverage of files outside of the cwd.

Could we autogenerate some content in the top-level package.json that would test each service and library?  Something like `yarn test:auth` or `yarn test:lib-api`?  I *think* if that translated to `mocha services/auth/test/**/*_test.js` then the cwd would be the repo root and we'd capture all of the library usage?

Although, looking at the numbers, it's really only lib-testing that's afflicted by this issue -- otherwise this report looks pretty accurate.  So maybe I could just write some tests for lib-testing.

imbstack:
And that the entire repo shows up in the intermediary reports which is one step closer than I saw with nyc. Maybe there will be a way to get us to having them all the way through.

PR #915, opened 2019-06-25
Update react-vnc-display version

helfi92:


djmitche:
@helfi92 any idea why renovate didn't ping us about this?

helfi92:
A new version of `react-vnc-display` was released yesterday. Renovate pings us at the beginning of the week. We would have received a PR from renovate if we had waited till Monday :)

djmitche:
Ah, thanks :)

PR #916, opened 2019-06-26
Bug 1560668 - do the whole image build process in one Dockerfile

djmitche:
Multi-stage docker builds are pretty neat.  Turns out, we can do the
whole thing in one Dockerfile, just pre-calculating the TC version and
the entrypoints.

This takes some care to avoid pulling in files that shouldn't be in the
image (like, secrets).  It first filters some obvious things out of the
tarball handed to `docker build`, and then the build process uses `git
clone` to ensure that only files recognized by git are transferred to
the generated image.  The use of multi-stage builds ensures that no
detritus from the build steps is included in the built image.

Bugzilla Bug: [1560668](https://bugzilla.mozilla.org/show_bug.cgi?id=1560668)


djmitche:
I've tested this on my loaner Mac.  I have noticed that docker builds will fail if the laptop sleeps while they're running, so I'm occasionally moving the mouse to keep it awake.  It took about 5 minutes on my EC2 instance, and about 10 seconds longer on the Mac.

PR #918, opened 2019-06-26
Fix and doc mozilla auth0 callback

djmitche:


deployment-docs/clouds.md
```diff
@@ -0,0 +1,12 @@
+# Clouds
+
+Taskcluster is a multi-cloud system, and as such expects credentials for several cloud providers.
+The services' Kubernetes deployment can run anywhere; at Mozilla it is deployed in GKE, Google Cloud's Kubernetes offering.
+Task artifacts are stored in Amazon S3, so AWS credentials will be required.
+Services' data is stored in Azure Table Storage, so Azure credentials will be required.
+
+Taskcluster can dynamically provision workers in a variety of clouds.
+You will need appropriate credentials for any clouds you intend to use for workers.
```
edunham: Later it'd be great to go into more detail on what credentials are appropriate, as this is quite non-obvious, but that's likely outside the scope of these changes

deployment-docs/clouds.md
```diff
@@ -0,0 +1,12 @@
+# Clouds
+
+Taskcluster is a multi-cloud system, and as such expects credentials for several cloud providers.
```
edunham: Do we consider rabbitmq one of the cloud providers? If I was looking at these docs first when evaluating taskcluster, I might mistake the following for a complete list of external services you have to either make or rent in order to get a working TC...

djmitche: I don't think so -- and there's a separate section on Pulse, anyway.

deployment-docs/clouds.md
djmitche: Yeah, we need to loop back to that at some point!

deployment-docs/clouds.md
djmitche: (In my mind, "cloud" means a provider of generalized cloud services like compute and storage -- Azure, GCP, AWS, Packet)

PR #920, opened 2019-06-26
Upgrade client-go deps to pinned versions

djmitche:
I'm a bit confused about this.  Importing the library without this PR encounters
```
build github.com/djmitche/goproj: cannot load code.google.com/p/go-uuid/uuid: cannot find module providing package code.google.com/p/go-uuid/uuid
```
which I think occurs because it's pulling in what was the latest *release* of github.com/taskcluster/slugid-go (v1.0.0).  That version was old and dated back to when `code.google.com` still existed.  But `go.mod` says `v1.0.1-0.20190423134720-372cc2dd3685` where `372cc2dd3685` is the current master revision of slugid-go.  So that's confusing.

I released a new version v1.0.1 (just pushed a tag based on master to the repo) and upgraded to that version.  With this change, I no longer get the code.google.com-related error.

This suggests we might be using old version of pulse-go, as well.  We'll see.


djmitche:
Per [bug 1561799](https://bugzilla.mozilla.org/show_bug.cgi?id=1561799), I'll roll in pinned versions of the other libraries, too.

I'd love to hear anyone's ideas about why this installed slugid-go@v1.0.0 instead of the master version, for my own edification -- but I think we're in good shape now anyway.

djmitche:
This seems to work OK from a temporary go project that uses the `replace` statement in `go.mod` to point at this branch.  But, I'm not totally convinced by that because I don't really understand the implications of `replace` (in particular, it seems not to list the indirect dependencies in `go.mod` anymore..)

petemoore:
> Per [bug 1561799](https://bugzilla.mozilla.org/show_bug.cgi?id=1561799), I'll roll in pinned versions of the other libraries, too.
> 
> I'd love to hear anyone's ideas about why this installed slugid-go@v1.0.0 instead of the master version, for my own edification -- but I think we're in good shape now anyway.

I'm not sure either, but if it is working, let's ship it, and if this becomes a problem later we can investigate more deeply.

PR #922, opened 2019-06-26
[Docs] Add UI best practices

helfi92:
There's probably more stuff to add but this is what came to mind when writing this document.

### DIFF
--- a/dev-docs/best-practices/ui.md
+++ b/dev-docs/best-practices/ui.md
@@ -24,6 +24,22 @@ When we have a good reason to not follow the best practices in the UI, we docume
 * Calculate link URLs before the user clicks on them. By doing so, a user will be able to do things like open links in a
 new tab or have the ability to inspect link URLs prior to the clicking.
 
+_Don't_
+
+```jsx
+handleLinkClick = () => {
+   this.props.history.push(...);
+};
+
+<Link onClick={handleLinkClick} />
+```
+
+_Do_
+
+```
+<Link to={...} />
+```
+
 ## Buttons
 * Destructive actions (e.g., a delete operation) should have a confirmation step.
 It’s an added click but will prevent accidental data loss frustration.
@@ -36,6 +52,8 @@ It’s an added click but will prevent accidental data loss frustration.
 ## Spacing
 If you need to add spacing to an element, use the spacing scale defined in the UI as opposed to picking
 a number that looks right to you. This will make sure there is a feel of consistency across the site design.
+The spacing scale for Taskcluster UI is defined
+[here](https://github.com/taskcluster/taskcluster/blob/8b32ca158622af8450626579ad1d4b9d8f1d3a85/ui/src/theme.js#L129-L134)
 
 ## Color
 Don’t depend only on color to communicate information, otherwise people who are color blind will have a hard time
--- a/dev-docs/best-practices/ui.md
+++ b/dev-docs/best-practices/ui.md
@@ -11,7 +11,7 @@ When we have a good reason to not follow the best practices in the UI, we docume
 * [Spacing](#spacing)
 * [Color](#color)
 * [Visual Hierarchy](#visual-hierarchy)
-* [Document Hierarchy](#document-hierarchy)
+* [Document Header Hierarchy](#document-header-hierarchy)
 
 ## Tables
 * Avoid multi-line text in a table - too much information for the given space. Alternatives:
@@ -49,7 +49,7 @@ For example, a destructive button that isn’t considered a primary action in a 
 need to be red and bold. It could have a secondary look. Still want it to be red and bold?
 Hide it behind a menu like a [SpeedDial](https://material-ui.com/components/speed-dial/) component.
 
-## Document Hierarchy
+## Document Header Hierarchy
 Document hierarchy should be respected everywhere, especially for documentation pages.
 There should be at most one `h1` tag in a page and it should be the first header in the document.
 
--- a/dev-docs/best-practices/ui.md
+++ b/dev-docs/best-practices/ui.md
@@ -12,6 +12,7 @@ When we have a good reason to not follow the best practices in the UI, we docume
 * [Color](#color)
 * [Visual Hierarchy](#visual-hierarchy)
 * [Document Header Hierarchy](#document-header-hierarchy)
+* [Tracking Changes](#tracking-changes)
 
 ## Tables
 * Avoid multi-line text in a table - too much information for the given space. Alternatives:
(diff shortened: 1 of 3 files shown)


dev-docs/best-practices/ui.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Visual Hierarchy
+Each button in a view has its own level of importance in that page relative to other actions.
+Most views have a single primary action, a couple that are medium emphasis and some that are rarely used.
+Buttons that are not primary should not have high contrast background colors.
+For example, a destructive button that isn’t considered a primary action in a view destructive does not
+need to be red and bold. It could have a secondary look. Still want it to be red and bold?
+Hide it behind a menu like a [SpeedDial](https://material-ui.com/components/speed-dial/) component.
+
+## Document Hierarchy
```
djmitche: Is this "document header hierarchy"?

helfi92: Indeed.

dev-docs/best-practices/ui.md
```diff
@@ -0,0 +1,68 @@
...
+## Color
+Don’t depend only on color to communicate information, otherwise people who are color blind will have a hard time
+understanding the UI. You can use labels. For things like graphs, you can use different shades of the same color.
+They are easier to differentiate for someone who is color blind since the colors will differ in contrast.
+
+## Visual Hierarchy
+Each button in a view has its own level of importance in that page relative to other actions.
+Most views have a single primary action, a couple that are medium emphasis and some that are rarely used.
+Buttons that are not primary should not have high contrast background colors.
+For example, a destructive button that isn’t considered a primary action in a view destructive does not
```
owlishDeveloper: I tend to disagree with this. Red is universally a color of warning. Destructive buttons should not be easy to click on accidentally

owlishDeveloper: Another reason why I disagree with this: in the new worker manager UI I have buttons in the table. This is an experiment and something new in our UI. Evidence suggests that those buttons are super-easy to miss if they don't have a background color.

owlishDeveloper:
Thank you for writing this up! 👍 I'll review later today

dev-docs/best-practices/ui.md
```diff
@@ -0,0 +1,68 @@
...
+* [Tracking Changes](#tracking-changes)
+
+## Tables
+* Avoid multi-line text in a table - too much information for the given space. Alternatives:
+  * A [Card](https://material-ui.com/components/cards/).
+  * A [Drawer](https://material-ui.com/components/drawers/) that opens up via a button click in the table.
+* Right-align numbers - it’s easier to scan a column of numbers if the decimal part starts in the same location.
+
+## Links
+* Calculate link URLs before the user clicks on them. By doing so, a user will be able to do things like open links in a
```
owlishDeveloper: what do you mean by this? If a link is a template literal, the string gets generated at the render time, doesn't it?

dev-docs/best-practices/ui.md
```diff
@@ -0,0 +1,68 @@
...
+new tab or have the ability to inspect link URLs prior to the clicking.
+
+## Buttons
+* Destructive actions (e.g., a delete operation) should have a confirmation step.
```
owlishDeveloper: Do our current components have confirmation steps everywhere?

owlishDeveloper:
The only note I have here: this hasn't even been merged yet, but you already try to apply these rules to https://github.com/taskcluster/taskcluster/pull/865 . <---- that PR was opened 21 days ago, before this one. I think any rules we add should apply to the PRs that are created _after_

dev-docs/best-practices/ui.md
```diff
@@ -0,0 +1,68 @@
...
+It’s an added click but will prevent accidental data loss frustration.
+
+## Text
+* Don’t center text spanning more than 3 lines. It’s not pretty. Alternatives:
+  * Rewrite the content to be shorter.
+  * Left-align the content if the text direction is left-to-right. Otherwise Right-align.
+
+## Spacing
+If you need to add spacing to an element, use the spacing scale defined in the UI as opposed to picking
```
owlishDeveloper: A link would be helpful

dev-docs/best-practices/ui.md
helfi92: > If a link is a template literal, the string gets generated at the render time, doesn't it?

Yeah, that's fine. What I was trying to say is to avoid links where the `href` attribute is only calculated after the user clicks on the link:

_Bad_

```jsx
handleLinkClick = () => {
   this.props.history.push(...);
};

<Link onClick={handleLinkClick} />
```


_Good_

```jsx
<Link to={...} />
```

I'll add this example to the document.

dev-docs/best-practices/ui.md
helfi92: Our current components don't have this at the moment but they should. I filed https://github.com/taskcluster/taskcluster/issues/957 yesterday to start the process. I plan a tracker issue to track all of the components that require this type of change.

> I see this can affect the PR I currently have in flight

We would need to add this change to the worker manager views. Feel free to ignore this for your current PR though since we don't have this implemented on any of our views. The tracker issue can take care of it.

dev-docs/best-practices/ui.md
helfi92: I understand your concern.

Our site tries to follow the material design spec. The details for a button can be found [here](https://material.io/design/components/buttons.html#usage). The types of buttons are:

1. Text button (low emphasis)
    - Text buttons are typically used for less important actions.
2. Outlined Button (medium emphasis)
    - Outlined buttons are used for more emphasis than text buttons due to the stroke.
3. Contained button (high emphasis)
    - Contained buttons have more emphasis, as they use use a color fill and shadow.
4. Toggle button
    - Toggle buttons group a set of actions using layout and spacing. They’re used less often than other button types.

>  in the new worker manager UI I have buttons in the table. This is an experiment and something new in our UI. I introduced that to make the UI more convenient.

Sure, I don't think I see a problem with this. I think it's a good thing to have but it's not a primary action for that view. Having the background red will make the icon button a contained one giving it more emphasis than it deserves for that specific view. When a user first enters that view, their eyes will automatically gravitate there first which is not what we want to do. Additionally, I think buttons are hard to miss in a table if they occupy their own column.

helfi92:
> The only note I have here: this hasn't even been merged yet, but you already try to apply these rules to #865 . <---- that PR was opened 21 days ago, before this one. I think any rules we add should apply to the PRs that are created after

We always tried to follow these rules or at least should of had (re: confirmation step). The purpose of this document is to make sure more people are aware of them and that we can refer to them when needed. This is an initial guide. I'm sure there will be more stuff written to it.

PR #924, opened 2019-06-27
headings-in-docs-file

### LINKED ISSUE #902: Check that headings in docs files match expectations
In markdown files under `ui/docs`, the rule is:

* if there are any headings (`#`, `##`, etc.), then there should be a single top-level heading (`#`) at the beginning of the file
* if there are no headings besides a single top-level heading, there should be no top-level heading

Let's write a meta-check to verify this.

subhamskm:
Fixes #902

### DIFF
--- a/test/meta_test.js
+++ b/test/meta_test.js
@@ -10,7 +10,7 @@ const glob = require('glob');
 
 const ROOT_DIR = path.join(__dirname, '..');
 
-suite('Repo Meta Tests', function() {
+suite('Repo Meta Tests', function () {
   const packageJsonFile = path.join(ROOT_DIR, 'package.json');
   const uiPackageJsonFile = path.join(ROOT_DIR, 'ui/package.json');
   const packageJson = JSON.parse(fs.readFileSync(packageJsonFile, 'utf8'));
@@ -19,10 +19,10 @@ suite('Repo Meta Tests', function() {
   const taskclusterYmlFile = path.join(ROOT_DIR, '.taskcluster.yml');
   const taskclusterYml = yaml.safeLoad(fs.readFileSync(taskclusterYmlFile, 'utf8'));
 
-  test('All packages in CI', async function() {
+  test('All packages in CI', async function () {
     const configured = taskclusterYml.tasks.in.$let.packages.map(pkg => pkg.name);
 
-    const {stdout} = await exec('yarn workspaces info -s');
+    const { stdout } = await exec('yarn workspaces info -s');
     const existing = Object.keys(JSON.parse(stdout))
       // taskcluster-client is tested separately
       .filter(name => name !== 'taskcluster-client');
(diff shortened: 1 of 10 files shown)


test/meta_test.js
```diff
@@ -119,4 +119,49 @@ suite('Repo Meta Tests', function() {
...
+      if (topLevelHd < 7) {
+        if (hd[topLevelHd].length > 1) {
+          throw new Error(`${filename} does not match the expectations`);
+        }
+      }
+
+      //check if there are no headings besides a single top-level heading
+      if (countLevel == 1) {
+        if (hd[topLevelHd].length == 1)
+          throw new Error(`${filename} does not match the expectations`);
```
djmitche: Could you add some more explanation to these two errors, so the user knows what to fix?

djmitche:
Oh, and please also fix the files that don't match the expectations :)

subhamskm:
> This is looking good. One more thing to fix up: can you arrange this so that it throws a single error if any file does not match, containing all of the errors?

You mean that all filenames and corressponding error in them should be displayed at once?

subhamskm:
> Oh, and please also fix the files that don't match the expectations :)

Do

subhamskm:
I don't know how the pull request got closed. How to reopen it?

imbstack:
Maybe you don't have access to the reopen button. I just clicked it for you!

subhamskm:
> Maybe you don't have access to the reopen button. I just clicked it for you!

Thanks @imbstack!

djmitche:
I see that the lint check is still failing -- can you fix that up?  `yarn lint` will run it locally for you.

djmitche:
Would you like to find something to work on next?

ui/docs/manual/using/administration.md
```diff
@@ -4,8 +4,6 @@ title: Project Administration
 order: 90
 ---
 
-# Project Administration
```
helfi92: Out of curiosity, what's the reason for removing this header? I think this is a top-level heading so there's no need to remove it. This comment applies to the other changes in the docs as well.

djmitche: Oh, I thought that was the rule: if there's only one top-level heading in the file, remove it as it's duplicative of the document title.

helfi92: Ah, I see your point. I think the problem is that not every top-level header share the same name as the filename. For example, a user viewing the login file won't necessarily know that it's docs about the login service. Another thing is that README.md files have no document title. We just use "Documentation" in those instances. Apologies for not bringing that up earlier.

subhamskm:
> Would you like to find something to work on next?

Yes, I would like to.

djmitche:
@subhamskm what do you think about https://bugzilla.mozilla.org/show_bug.cgi?id=1551848?

djmitche:
Or possibly https://bugzilla.mozilla.org/show_bug.cgi?id=1562752

helfi92:
> All of the Markdown files can have a title property in their metadata, including READMEs. So perhaps the files changed here should be updated to include that property? I see that https://taskcluster-ui.herokuapp.com/docs/reference/core/notify lacks it, for example, but https://taskcluster-ui.herokuapp.com/docs/manual/using/administration has it. Would that make more sense, or should we change what this check verifies?

That works but we would probably need to have a check to make sure there are no single `#` in a markdown file since we would rely on the `title` property. I think the easy solution would be to simply remove the test checking for "if there are no headings besides a single top-level heading, there should be no top-level heading" and fixing up the modified markdown files of this PR. I'm okay with both approaches but leaning towards the latter.

djmitche:
I agree with leaning toward the latter.  @subhamskm do you want to make a new PR for that?

subhamskm:
What I have to change?

helfi92:
@subhamskm, I think there are 2 things to change:

1. Remove the test that checks for "if there are no headings besides a single top-level heading, there should be no top-level heading"
2. Undo the changes in ` ui/docs/*`

djmitche:
By the way, this kind of thing happens a lot -- it's much easier to see that an idea is not-quite-right once it's expressed in code, so it's not uncommon for a PR to be followed up quickly with another PR to adjust it.  It can be frustrating, but it's pretty common.

PR #926, opened 2019-06-28
Bug 1558346 - add workerManager.reportWorkerError

djmitche:
Bugzilla Bug: [1558346](https://bugzilla.mozilla.org/show_bug.cgi?id=1558346)

### DIFF
--- a/services/worker-manager/src/api.js
+++ b/services/worker-manager/src/api.js
@@ -16,6 +16,7 @@ let builder = new APIBuilder({
     'WorkerPoolError',
     'providers',
     'publisher',
+    'notify',
   ],
 });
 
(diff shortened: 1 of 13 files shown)


services/worker-manager/src/api.js
```diff
@@ -217,6 +218,53 @@ builder.declare({
...
+
+  const workerPool = await this.WorkerPool.load({workerPoolId}, true);
+  if (!workerPool) {
+    return res.reportError('ResourceNotFound', 'Worker pool does not exist', {});
+  }
+
+  const wpe = await workerPool.reportError({
+    kind: input.kind,
+    title: input.title,
+    description: input.description,
```
imbstack: Since we eventually ended up adding a kludgy thing to remove fields like `password` from logs should we maybe jump the gun and do it here too before someone accidentally logs stuff?

djmitche: It's a string, though.  My hope was to just r- anything that dumps a random batch-o-data into `extra`, in preference for very specific properties.

djmitche: It's the same level of caution we exercise with `res.reportError` for example..

PR #927, opened 2019-06-28
Remove tc-events from ingress (it doesn't exist)

djmitche:


djmitche:
I verified that /api/events is 503'ing in staging, as you'd expect from a nonexistent ingress backend service

PR #928, opened 2019-06-28
Bug 1561708 - put major version in client-go package path

djmitche:
Bugzilla Bug: [1561708](https://bugzilla.mozilla.org/show_bug.cgi?id=1561708)


clients/client-go/README.md
```diff
@@ -1,44 +1,43 @@
...
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/v14/tcawsprovisioner
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/v14/tcec2manager
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/v14/tcgithub
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/v14/tchooks
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/v14/tcindex
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/v14/tclogin
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/v14/tcnotify
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/v14/tcpurgecache
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/v14/tcqueue
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/v14/tcsecrets
```
petemoore: I guess now that we have the monorepo, it makes sense that the services are versioned together with the platform, rather than having individual version numbers, so this seems ok. I just wanted to highlight that until now, services were versioned independently (so e.g. `tcqueue` could have a different version number to `tcsecrets`) but I guess that doesn't make much sense for us now, so this looks fine.

infrastructure/builder/src/release/tasks.js
```diff
@@ -105,7 +105,15 @@ module.exports = ({tasks, cmdOptions}) => {
         contents.replace(/VERSION = .*/, `VERSION = '${requirements['release-version']}'`));
       changed.push(pyclient);
 
-      // go client is versioned only by git tag
+      // the go client requires the major version number in its import path, so
+      // just about every file needs edited.  This matches the full package
```
petemoore:
```suggestion
      // just about every file needs to be edited.  This matches the full package
```

clients/client-go/README.md
djmitche: That could still happen -- a single application could certainly link to v14 and v15 in different places.  But tcsecrets v14 and v15 might be identical if the breaking change did not involve that service.

infrastructure/builder/src/release/tasks.js
djmitche: I was trying to reflect the Pittsburgh manner of speaking :)

clients/client-go/README.md
```diff
@@ -1,44 +1,43 @@
 # Taskcluster Client Go
 
-[![Build Status](https://travis-ci.org/taskcluster/taskcluster/clients/client-go.svg?branch=master)](http://travis-ci.org/taskcluster/taskcluster/clients/client-go)
```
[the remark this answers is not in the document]
djmitche: I did - it no longer builds in travis-ci.

Issue #925, opened 2019-06-28
[UI] Remove aws-provisioner and provisioners


helfi92:
Worker manager UI will be replacing the aws-provisioner and provisioners views. Brace yourself, red diffs incoming!

PR #929, opened 2019-06-28
Remove aws-provisioner and provisioners views

helfi92:
Resolves https://github.com/taskcluster/taskcluster/issues/925.


djmitche:
I looked at this too quickly -- removing AWS-provisioner makes lots of sense, but I don't know if removing the provisioners views did.  Those still work, and are calling queue APIs that at least for the moment still exist.

Certainly replacing links to workerIds with links to tools.taskcluster.net isn't going to work once tools.taskcluster.net is shut down.

Should we maybe revert this and only remove AWS provisioner?

helfi92:
That works. Will revert.

owlishDeveloper:
I thought the plan was to remove these components after worker manager has been deployed to production?

PR #946, opened 2019-07-01
Bug 1558187 - handle connection failures better

djmitche:
amqplib's failure handling is pretty poor -- it seems this particluar
error doesn't match IllegalOperationError.  This patch also adds some
lib-testing support to handle testing for error messages.  And it
sets prefetch=1 in a few places to get more stable test behavior.

Bugzilla Bug: [1558187](https://bugzilla.mozilla.org/show_bug.cgi?id=1558187)

### DIFF
--- a/libraries/pulse/src/consumer.js
+++ b/libraries/pulse/src/consumer.js
@@ -4,6 +4,23 @@ const assert = require('assert');
 const slugid = require('slugid');
 
 /**
+ * Recognize some "expected", ignorable errors due to normal network failures.
+ */
+const isExpectedError = err => {
+  // IllegalOperationError happens when we are draining a broken channel; ignore
+  if (err instanceof amqplib.IllegalOperationError) {
+    return true;
+  }
+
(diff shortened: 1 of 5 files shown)


djmitche:
@imbstack I think you've looked at this code the most..

libraries/pulse/src/consumer.js
```diff
@@ -3,6 +3,23 @@ const amqplib = require('amqplib');
 const assert = require('assert');
 const slugid = require('slugid');
 
+/**
+ * Recognize some "expected", ignorable errors due to normal network failures.
+ */
+const isExpectedError = err => {
```
imbstack: Everything about this function makes me uneasy but it is not your fault at all.

imbstack: I hate that there isn't a better way to do any of this.

djmitche: `amqplib` certainly leaves a great deal to be desired in terms of error handling.

PR #948, opened 2019-07-01
Bug 1555213 - remove previous providers when they are done

djmitche:
This defers calling removeResources until all workers in the pool for a
specific provider are STOPPED.

This brings up some questions that might lead to good followups:

 * how are workers removed from the Worker table?
 * should we consider workers in the STOPPED state but still in the table to be a reason to hold onto the provider's resources?

Bugzilla Bug: [1555213](https://bugzilla.mozilla.org/show_bug.cgi?id=1555213)


djmitche:
@imbstack ping?

PR #950, opened 2019-07-01
Bug 1562751 - request a changelog entry for each PR

djmitche:
Bugzilla Bug: [1562751](https://bugzilla.mozilla.org/show_bug.cgi?id=1562751)


djmitche:
That's a good question -- what is your opinion?  A PR to update the linked best-practice doc would be great.

I do think at least *some* UI changes should be in the changelog!

helfi92:
> I do think at least some UI changes should be in the changelog!

Agree. That works.

>  A PR to update the linked best-practice doc would be great.

Good idea.

PR #961, opened 2019-07-02
No references to tools.taskcluster.net

subhamskm:


subhamskm:
> * The glob pattern only matches files with a dot in the name -- why is that?
Dot occurs in every file name between the name and extension. So I thought it could be used to distinguish between a directory and a file and would return all the filenames. But still directory names were coming in the search result. So I used  `fs.lstatSync(filename).isFile()` to check for filenames only. I will remove the dot in next commit.

subhamskm:
> * the test fails because there are lots of files containing the string.  I had suggested in the issue to make a whitelist -- a list of files where the string is allowed.  To start with, that list can contain all the files that contain the string, and we can make followup PRs to remove them bit by bit (I'll probably do those PRs since they will be 99% thinking about what to write instead and 1% writing that thing).

So I need to store all the filenames in a javascript list. Rest of the code will be same. Did I get you right?

djmitche:
That sounds about right, yes.

subhamskm:
The previous commits are not getting removed. Should I open a new pull request?

djmitche:
I'm not sure what you mean by "not getting removed"?

subhamskm:
> * it has the changes to `headings in docs files match expectations` mixed in

I thought you want me to remove those changes that are mixed in.

djmitche:
Yes, that's what I want -- each PR should contain only the changes related to that PR.

PR #964, opened 2019-07-03
Bug 1558345 - replace credentialsGoogle with registerWorker

djmitche:
Bugzilla Bug: [1558345](https://bugzilla.mozilla.org/show_bug.cgi?id=1558345)


services/worker-manager/src/api.js
```diff
@@ -335,3 +338,79 @@ builder.declare({
...
+  }
+
+  if (worker.providerId !== providerId) {
+    return res.reportError('InputError',
+      `Worker ${workerGroup}/${workerId} does not have provider ${providerId}`, {});
+  }
+
+  const {expires, errorMessage} = await provider.registerWorker({worker, workerPool, workerIdentityProof});
+
+  if (errorMessage) {
```
imbstack: I actually liked the behavior before of throwing errors, catching, logging them, and then returning a generic message. 

This way seems easier to get wrong and has less useful error messages in our logs.

djmitche: It's not a great design, you're right.  I'm looking for a way to distinguish 500's and 400's, basically, with some useful generic messages for the 400's.  If I switched it to use a dedicated error class, would that be good?

imbstack: I wonder if we should intentionally _not_ distinguish 500/400 here. Otherwise a dedicated class definitely makes sense to me!

djmitche: I did that :)

djmitche: Sorry, that comment didn't appear.  One distinction is that 500 is implicitly a server-side issue and will likely be retried, while a 400 is a pretty clear "ya done something wrong" indication.

services/worker-manager/src/providers/google.js
```diff
@@ -208,63 +199,31 @@ class GoogleProvider extends Provider {
 
     // First check to see if the request is coming from the project this provider manages
     if (dat.project_id !== this.project) {
-      const error = new Error(`Invalid project ${dat.project_id} is not ${this.project}`);
-      error.project = dat.project_id;
-      error.validProject = this.project;
-      throw error;
+      throw new RegistrationError('Token project mismatch');
```
owlishDeveloper: I have a question: I'm not sure if this guards against some accidentals or against a malicious action..... if the latter, maybe we shouldn't be as specific in the response? For example, Github throws 404 for authentication/authorization errors.

djmitche: It is the latter.  That's what Brian was suggesting, too.  And it's what we do for bad authentication (401 with "Bad Mac").  I suppose it's pretty unlikely to see any of these errors when things are operating correctly. And I just realized that the most likely error a malicious actor would trigger is an exception from validateIdToken -- which this code handles as a 500.  So, I'll rewrite to always return the same error in any of those cases.

services/worker-manager/src/api.js
```diff
@@ -335,3 +339,83 @@ builder.declare({
...
+  const worker = await this.Worker.load({workerPoolId, workerGroup, workerId}, true);
+  if (!worker) {
+    return res.reportError('ResourceNotFound',
+      `Worker ${workerGroup}/${workerId} in worker pool ${workerPoolId} does not exist`, {});
+  }
+
+  if (worker.providerId !== providerId) {
+    return res.reportError('InputError',
+      `Worker ${workerGroup}/${workerId} does not have provider ${providerId}`, {});
+  }
```
djmitche: Note that these are all publicly verifiable facts, so there's no reason to obscure these messages.

PR #966, opened 2019-07-03
Better documentation of the provider.setup method

djmitche:
This clears up some of the confusion I had in #923, and is based on how the method is called.

I did notice that if `setup` fails, the process startup will fail.  For things like creating IAM users, that could mean that e.g., an EC2 outage would stop the worker-manager from starting up.  Should we try to soften that requirement a little bit?  Ideas how?  I can file a followup bug..

### DIFF
--- a/services/worker-manager/src/providers/provider.js
+++ b/services/worker-manager/src/providers/provider.js
@@ -34,9 +34,10 @@ class Provider {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


djmitche:
https://bugzilla.mozilla.org/show_bug.cgi?id=1563369

PR #973, opened 2019-07-05
[Bug 1530644] Add tags to pulse messages

srfraser:
Addition of tags to these pulse messages will allow listeners to better
classify the tasks that are appearing, especially after we add ~3 fields.

Bugzilla Bug: [1530644](https://bugzilla.mozilla.org/show_bug.cgi?id=1530644)

### DIFF
--- a/services/queue/schemas/v1/task-completed-message.yml
+++ b/services/queue/schemas/v1/task-completed-message.yml
@@ -6,6 +6,7 @@ type:         object
 properties:
   version:    {$const: message-version}
   status:     {$ref: 'task-status.json#'}
+  tags: {$ref: "task.json#/properties/tags"}
   runId:
     description: |
       Id of the run that completed the task
--- a/services/queue/schemas/v1/task-completed-message.yml
+++ b/services/queue/schemas/v1/task-completed-message.yml
@@ -6,15 +6,7 @@ type:         object
 properties:
   version:    {$const: message-version}
   status:     {$ref: 'task-status.json#'}
-  task:
-    description: |
-      Subset of a task definition
-    type:               object
-    properties:
-      tags:     {$ref: "task.json#/properties/tags"}
-    additionalProperties: false
-    required:
-      - tags
+  task: {$ref: 'task-pulse-definition.json#'}
   runId:
     description: |
       Id of the run that completed the task
--- a/services/queue/schemas/v1/task-completed-message.yml
+++ b/services/queue/schemas/v1/task-completed-message.yml
@@ -6,7 +6,15 @@ type:         object
 properties:
   version:    {$const: message-version}
   status:     {$ref: 'task-status.json#'}
-  tags: {$ref: "task.json#/properties/tags"}
+  task:
+    description: |
+      Subset of a task definition
+    type:               object
+    properties:
+      tags:     {$ref: "task.json#/properties/tags"}
+    additionalProperties: false
+    required:
+      - tags
   runId:
     description: |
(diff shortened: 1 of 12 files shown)


services/queue/schemas/v1/task-failed-message.yml
```diff
@@ -6,6 +6,7 @@ type:         object
 properties:
   version:    {$const: message-version}
   status:     {$ref: 'task-status.json#'}
+  tags: {$ref: "task.json#/properties/tags"}
```
petemoore: nit: indenting in all of these `.yml` files is a bit off

srfraser:
> However, all of the other data in this message is inside of the `status` object. I'm not sure why that's the case, but we should follow the pattern rather than break new ground.

I'd assumed that was because the 'status' object contains the result of the `task.status()` call, and it allowed code re-use where interpreting that structure could be done whether the source was an api call or a pulse message.

Are you imagining that underneath status would be a `tags` object that contains all the tags underneath it, or would the contents of tags be merged with those in status? I'd prefer to keep the contents of the tags as flexible as possible, as not every task sets every option, and it makes adding new tags awkward.

djmitche:
Good point regarding the existing status schema.

We discussed this in the team meeting today, with some concern that adding tags is opening the door to adding the whole task definition -- and that would be a 5x or so increase in pulse bandwidth.  So we thought about what the right distinction is between something that belongs in the pulse message and something that doesn't.  What we landed on is, we can include data that is useful for deciding whether to fetch a task definition.  Tags are an easy "yes".  Metadata might OK, too (not in this PR, but thinking forward..).  Task.extra and task.payloard are an easy "no" (just fetch the task).

Since we might want to later add additional fields, @ccooper suggested that we include them in the same structure as a task definition, in a top-level `task` property.  So the message would have the form `{"task": {"tags": ...}, "status": {..}}` (and we can later add `"metadata"` to the `task` object).

So, updating the request:
* move tags into a new `task` object
* make it required
* add tests to ensure that the value is correct

srfraser:
> What we landed on is, we can include data that is useful for deciding whether to fetch a task definition. Tags are an easy "yes". Metadata might OK, too (not in this PR, but thinking forward..). Task.extra and task.payloard are an easy "no" (just fetch the task).

This seems entirely reasonable to me, I too want to avoid large pulse messages.

> Since we might want to later add additional fields, @ccooper suggested that we include them in the same structure as a task definition, in a top-level `task` property. So the message would have the form `{"task": {"tags": ...}, "status": {..}}` (and we can later add `"metadata"` to the `task` object).

+1 here, I like this and will update the PR.

services/queue/schemas/v1/task-completed-message.yml
```diff
@@ -6,6 +6,15 @@ type:         object
...
   status:     {$ref: 'task-status.json#'}
+  task:
+    description: |
+      Subset of a task definition
+    type:               object
+    properties:
+      tags:     {$ref: "task.json#/properties/tags"}
+    additionalProperties: false
+    required:
+      - tags
```
djmitche: Since this hunk is repeated three times, let's pull it out into its own schema file, and also add some context to the description, perhaps:

> Subset of a task definition containing values that are useful for determining whether a message is interesting to the receiver.  Where the full task definition is required for a "interesting" task, the receiver should call `queue.task` to download that definition.

djmitche:
one more `yarn generate` :)

djmitche:
Deployed to production..

Issue #969, opened 2019-07-04
[UI] Redirect taskcluster-web.netlify.com to taskcluster-ui.herokuapp.com


helfi92:
Netlify is only used for preview PRs. Let's have taskcluster-web.netlify.com redirect to the new site but not redirect preview URLs.

arku:
@helfi92 Can we close this as it is done in https://github.com/taskcluster/taskcluster/pull/983 or is there something missing?

helfi92:
It's actually not fixed. Not sure why at the moment. Let me know if you had any ideas. Navigating to https://taskcluster-web.netlify.com/ doesn't redirect to https://taskcluster-ui.herokuapp.com/ for example.

arku:
@helfi92 I submitted a [PR](https://github.com/taskcluster/taskcluster/pull/1041) attempting to fix this issue. Hopefully, it works. Take a look when you get a chance.

arku:
@helfi92 I tested this on an app of mine and the redirect rule works :ok_hand:

helfi92:
w00t 🎉 The redirect works perfectly! Thanks for taking the time to look into it @arku :)

PR #983, opened 2019-07-08
[UI] Redirect netlify master branch to heroku app

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/969.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -13,6 +13,12 @@
   UI_LOGIN_STRATEGY_NAMES = ""
   TASKCLUSTER_ROOT_URL = "https://taskcluster.net"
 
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -13,6 +13,12 @@
   UI_LOGIN_STRATEGY_NAMES = ""
   TASKCLUSTER_ROOT_URL = "https://taskcluster.net"
 
+# Rule for redirecting master branch to the heroku app
+[[redirects]]
+  from = "https://taskcluster-web.netlify.com/*"
```
arku: @helfi92 I don't think we can pass a URL as the `from` value, it has to be a relative path. I could be wrong though.

Issue #908, opened 2019-06-25
[UI] Go to line feature stops working when follow log is first clicked


helfi92:
STR:

1. Navigate to http://localhost:5080/tasks/fRav6OG-Sbij116H7gfqag/runs/0/logs/https%3A%2F%2Fqueue.taskcluster.net%2Fv1%2Ftask%2FfRav6OG-Sbij116H7gfqag%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log
2. Click the follow log button

![Screen Shot 2019-06-25 at 8 52 38 AM](https://user-images.githubusercontent.com/3766511/60099910-7a13d580-970d-11e9-8e26-bf0305b3f8a1.png)

3. Click the "1 2 3" icon button and enter a line number (e.g., 4) then click on "Go To Line".

**Expected:** Log jumps to the selected line number
**Actual:** Log does not jump to the selected line number.

arku:
@helfi92 Do you mind helping me understand how the 'Go to Line' feature works?

helfi92:
Sure. The LazyLog component (aka log viewer) accepts a `scrollToLine` prop which will theoretically jump to that line when that value changes. I think the logic for that can be found [here](https://github.com/mozilla-frontend-infra/react-lazylog/blob/e8951828d6ac4e94d0592b501921de73ced4d990/src/components/LazyLog/index.jsx#L293-L299).

For some reason, when we click the follow log icon, the jump to line feature stops functioning properly. The problem could either be in this repo or in https://github.com/mozilla-frontend-infra/react-lazylog.

Does that help? Let me know if you had more questions.

arku:
@helfi92 Thanks for clearing it up. Gonna take a look at it.

Will get back to you after making some progress.

helfi92:
Awesome! Also thanks for helping other contributors on the other issues :)

arku:
@helfi92 Looks like the issue is on [this line](https://github.com/mozilla-frontend-infra/react-lazylog/blob/e8951828d6ac4e94d0592b501921de73ced4d990/src/utils.js#L20). When the 'Follow Log' icon is clicked, the `follow` property in `Log` component's state is set to `true` and is never reset when `lineNumber` is updated. And it looks like the 'Go to line' functionality isn't working when actually the `scrollToIndex` function always returns `count - 1 - offset` as `follow` is true.

Let me fix it.

arku:
[...]

PR #984, opened 2019-07-08
Fix 'Go to Line' functionality when 'Follow Log' is enabled

arku:
Fixes https://github.com/taskcluster/taskcluster/issues/908

Resets `follow` in `Log` component's state when `lineNumber` is updated from the `Go to Line` dialog.

### DIFF
--- a/ui/src/components/Log/index.jsx
+++ b/ui/src/components/Log/index.jsx
@@ -190,7 +190,7 @@ export default class Log extends Component {
   };
 
   handleLineNumberChange = lineNumber => {
-    this.setState({ lineNumber });
+    this.setState({ lineNumber, follow: null });
   };
 
   handleScroll = ({ scrollTop, scrollHeight, clientHeight }) => {


ui/src/components/Log/index.jsx
```diff
@@ -190,7 +190,7 @@ export default class Log extends Component {
   };
 
   handleLineNumberChange = lineNumber => {
-    this.setState({ lineNumber });
+    this.setState({ lineNumber, follow: null });
```
helfi92: `follow` is a boolean so we should probably just set it to `false` instead of `null`. I'll pull the code down in a bit to test it.

arku: @helfi92 I thought about it but ended up with `null` we are `resetting` the value of `follow` and `null` made sense in that context (The default value of `follow` is `null`).

That being said, I don't have any objections setting it to `false`. Do you want me to?

helfi92: I tend to agree with you. We can keep it as `null` :)

arku:
@helfi92 Good catch. Looks like the search state is managed inside the `SearchBar` component from `react-lazylog`. Not sure why it affects 'Go to Line' though. Let me check.

arku:
@helfi92 [This line](https://github.com/mozilla-frontend-infra/react-lazylog/blob/e8951828d6ac4e94d0592b501921de73ced4d990/src/components/LazyLog/index.jsx#L193) is the cause of the bug, I guess. If you reset the search keyword ('Public' in your example), you can observe the log jumping to the right line.

Can you help me understand the purpose of it?

helfi92:
> @helfi92 [This line](https://github.com/mozilla-frontend-infra/react-lazylog/blob/e8951828d6ac4e94d0592b501921de73ced4d990/src/components/LazyLog/index.jsx#L193) is the cause of the bug, I guess. If you reset the search keyword ('Public' in your example), you can observe the log jumping to the right line.

Thanks for tracking this down 🥇

> Can you help me understand the purpose of it?

When the user types in the input field, the [`onSearch`](https://github.com/mozilla-frontend-infra/react-lazylog/blob/e8951828d6ac4e94d0592b501921de73ced4d990/src/components/SearchBar/index.jsx#L15-L18) handler will be triggered. This will set `isSearch` to true. This is done [here](https://github.com/mozilla-frontend-infra/react-lazylog/blob/e8951828d6ac4e94d0592b501921de73ced4d990/src/components/LazyLog/index.jsx#L378-L393). The boolean flag `isSearch` is used to format lines (highlight matching substrings) of the log [here](https://github.com/mozilla-frontend-infra/react-lazylog/blob/e8951828d6ac4e94d0592b501921de73ced4d990/src/components/LazyLog/index.jsx#L446-L456). The role of `scrollToIndex` that is calculated in the line your shared is explained in https://github.com/bvaughn/react-virtualized/blob/master/docs/List.md. It will make sure that row is visible in the UI. Perhpas we are feeding the component the wrong scroll index.

arku:
@helfi92 Thanks for the explanation.

When a search is active (`isSearching` is `true`) and go to line functionality is used next, the `scrollToIndex` is never updated until `isSearching` is set to `false`. How do we fix this? 🤔

helfi92:
> How do we fix this? 🤔

To be honest, I'm not sure why we have

```jsx
const newScrollToIndex = isSearching
  ? scrollToIndex
  : getScrollIndex({ follow, scrollToLine, count, offset });
```

I tried inserting a `console.log` below this statement and saw that `scrollToIndex` and `getScrollIndex` always returned the same value. Maybe we can simply do

```jsx
const newScrollToIndex = getScrollIndex({ follow, scrollToLine, count, offset });
```

arku:
> To be honest, I'm not sure why we have
> 
> ```js-jsx
> const newScrollToIndex = isSearching
>   ? scrollToIndex
>   : getScrollIndex({ follow, scrollToLine, count, offset });
> ```

@helfi92 Neither do I 😞 

> I tried inserting a console.log below this statement and saw that scrollToIndex and getScrollIndex always returned the same value.

Interesting. They should only be same the first time right (unless you go to the same line)? 🤔 

Initially, the values should be the same (both -1). When you try to go to a line (say, 200), the `scrollIndex` in the state of `Lazylog` component will still be -1 but `getScrollIndex` will return 199. Unless you go to the same line, `scrollIndex` in state will never be equal to the return value of `getScrollIndex`.

Or am I missing something? Happy to be corrected if so.

arku:
@helfi92 I have submitted a [PR](https://github.com/mozilla-frontend-infra/react-lazylog/pull/42) to `react-lazylog` fixing the issue. I used `npm link` to (manually) test the behavior. Take a look when you get a chance.

Thanks!

PR #989, opened 2019-07-08
[Bug 1563545] Elide bewit in logs

imbstack:
Bugzilla Bug: [1563545](https://bugzilla.mozilla.org/show_bug.cgi?id=1563545)


libraries/api/src/middleware/queries.js
```diff
@@ -22,12 +22,14 @@ const queryValidator = ({context, entry}) => {
...
       if (pattern instanceof RegExp) {
         if (!pattern.test(value)) {
+          delete req.query[key];
           errors.push('Query-string parameter: ' + key + '="' + value +
                       '" does not match expression: ' + pattern.toString());
         }
       } else {
         const msg = pattern.call(context, value);
         if (typeof msg === 'string') {
+          delete req.query[key];
```
imbstack: @djmitche just want to call this out in particular as a more possibly breaking change here to make sure you think it is ok.

libraries/api/test/logging_test.js
```diff
@@ -237,4 +259,112 @@ suite(testing.suiteName(), function() {
...
+        apiVersion: 'v1',
+        clientId: '',
+        hasAuthed: false,
+        method: 'GET',
+        public: true,
+        query: {
+          foo: 'abc',
+          bewit: '...',
+        },
+        resource: '/api/test/v1/bewitiful',
```
imbstack: @djmitche actually, while I'm looking at it, we can remove `baseUrl` from here to make this just `/bewitful` to save some bytes since we already have `apiVersion` in the logging method and the rest of this string is useless. Does that make sense to you?

libraries/api/src/middleware/queries.js
djmitche: If I'm reading correctly, this is omitting values from `req.query` in cases where the query was ill-formed, which was already a 400 error and not calling the handler, right?  If so, then I'm not worried.

libraries/api/test/logging_test.js
djmitche: Sounds OK to me.  That should be noted in the changelog :)

PR #998, opened 2019-07-09
Bug 1563341 - create / remove workers in worker-manager

djmitche:
Bugzilla Bug: [1563341](https://bugzilla.mozilla.org/show_bug.cgi?id=1563341)

### DIFF
--- a/clients/client-go/tcworkermanager/types.go
+++ b/clients/client-go/tcworkermanager/types.go
@@ -33,6 +33,10 @@ type (
 		Token string `json:"token"`
 	}
 
+	// Provider-specific information
+	ProviderData struct {
+	}
+
 	// Request body to `registerWorker`.
 	RegisterWorkerRequest struct {
 
@@ -92,6 +96,16 @@ type (
 
 	// (No information is currently required; TBD)
 	StaticProviderType struct {
+	}
+
+	// Request to create a worker
+	WorkerCreationRequest struct {
+
+		// Date and time when this worker will be deleted from the DB
+		Expires tcclient.Time `json:"expires"`
+
+		// Provider-specific information
+		ProviderInfo ProviderData `json:"providerInfo,omitempty"`
 	}
 
 	// A report of an error from a worker.  This will be recorded with kind
--- a/services/worker-manager/schemas/v1/create-worker-request.yml
+++ b/services/worker-manager/schemas/v1/create-worker-request.yml
@@ -1,0 +1,20 @@
+$schema: "/schemas/common/metaschema.json#"
+title: Worker Creation Request
+description: Request to create a worker
+type: object
+properties:
+  expires:
+    title: Expires
+    description: Date and time when this worker will be deleted from the DB
+    type: string
+    format: date-time
+  providerInfo:
+    title: Provider Data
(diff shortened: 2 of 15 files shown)


djmitche:
Could I get a, dunno, 40% review?  The last commit is almost done, and basically defers createWorker and removeWorker to the relevant provider.  I'll submit the PR for review once that's done, but the next step will be to support these methods in the static provider, and will require that a shared secret be given when creating a worker and supplied to registerWorker (details in the bug).

Am I on the wrong track?

services/worker-manager/src/api.js
```diff
@@ -365,6 +365,100 @@ builder.declare({
...
 
+builder.declare({
+  method: 'put',
+  route: '/workers/:workerPoolId:/:workerGroup/:workerId',
+  name: 'createWorker',
+  title: 'Create a Worker',
+  stability: APIBuilder.stability.experimental,
+  input: 'create-worker-request.yml',
+  output: 'worker-full.yml',
+  // note that this pattern relies on workerGroup and workerId not containing `/`
```
imbstack: what happens if they do?

djmitche: The schema prohibits it, so it will be an error -- the comment is just to suggest that if we change the schema, we're gonna have issues here.

djmitche:
100% review plz!  I'll follow up with a PR to add support in the providers.

djmitche:
Oh, thanks!  I looked for "TODO" and "XXX" in the diffs and read through them, but missed the description :)

clients/client-go/tcworkermanager/types.go
```diff
@@ -33,6 +33,10 @@ type (
 		Token string `json:"token"`
 	}
 
+	// Provider-specific information
+	ProviderData struct {
+	}
```
petemoore: This looks like it might not be what you want.

services/worker-manager/schemas/v1/create-worker-request.yml
```diff
@@ -0,0 +1,20 @@
...
+    title: Expires
+    description: Date and time when this worker will be deleted from the DB
+    type: string
+    format: date-time
+  providerInfo:
+    title: Provider Data
+    description: Provider-specific information
+    type: object
+    properties: {}
+    additionalProperties: false
```
petemoore:
```suggestion
    additionalProperties: true
```

clients/client-go/tcworkermanager/types.go
djmitche: It is what I want -- at the moment, no providers support any fields, so nothing is allowed.  In the next PR, when I add static-provider support, it will be an anyOf, similar to 

https://github.com/taskcluster/taskcluster/blob/8bf75678626567ec2fdae3889dff2f6d86235958/services/worker-manager/schemas/v1/register-worker-request.yml#L15-L32

(which gets treated as a `json.RawMessage`)

petemoore: Ah cool, I thought maybe it was a freeform blob, but that makes sense.

Note, that `anyOf` should be `oneOf`.

PR #999, opened 2019-07-09
no references to tools.taskcluster.net

subhamskm:

### DIFF
--- a/test/meta_test.js
+++ b/test/meta_test.js
@@ -173,20 +173,47 @@ suite('Repo Meta Tests', function () {
     const filenames = glob.sync(
       '**/*',
       { cwd: ROOT_DIR });
-
-    let files=[]; //this list stores the files which has references to tools.taskcluster.net
+    const whitelist = [
+      'node_modules/slugid/README.md',
+      'node_modules/taskcluster-lib-urls/src/index.js',
+      'node_modules/taskcluster-lib-urls/src/main/java/org/mozilla/taskcluster/urls/LegacyURLs.java',
+      'node_modules/taskcluster-lib-urls/src/main/java/org/mozilla/taskcluster/urls/URLProvider.java',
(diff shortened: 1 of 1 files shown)


subhamskm:
> Nice job removing the other commits from this PR!
> 
> This now nicely bundles the errors into a single error message. However, it does not yet have the "whitelist" I suggested in the previous PR, which is why the `test meta` check is failing.

I don't know what is a "whitelist". Can you help me create it?

djmitche:
I described it in the second item [here](https://github.com/taskcluster/taskcluster/pull/961#pullrequestreview-257456968).

subhamskm:
I have created a list `files` which contains all the filenames in which the string is present.

subhamskm:
> I have created a list `files` which contains all the filenames in which the string is present.

djmitche:
Not quite -- a [whitelist](https://en.wikipedia.org/wiki/Whitelisting) is a list of "allowed" exceptions to a rule, with anything not in the list being treated as an error.  So what I'd like to see is a list of files where `tools.taskcluster.net` can appear, and the test configured to ignore those files but fail if the string appears in other files (that aren't in the whitelist).  That should make the `test:meta` check pass in this PR and we can merge it.

Then, in later PRs, we can remove `tools.taskcluster.net` from files and, at the same time, remove those files from the whitelist.  Eventually, we might even get to an empty whitelist.

This is a pretty common technique for making gradual changes: the use of a whitelist prevents anyone from adding *new* `tools.taskcluster.net` strings in any files, and gives us time to work on the existing files.

djmitche:
The next bit is to start replacing some of those references, but that's 99% figuring out what they *should* refer to at 1% changing it, so I won't ask you to do so.  I'll reassign the bug to myself for that purpose.  Thanks again!

subhamskm:
Thanks @djmitche for your guidance.

PR #1004, opened 2019-07-11
Document provider implementation a markdown file

djmitche:
This is mostly just moving info out of comments in `provider.js` and into a markdown file.. with more detail.

### DIFF
--- a/services/worker-manager/providers.md
+++ b/services/worker-manager/providers.md
@@ -1,0 +1,117 @@
+# Providers
+
+Providers are implemented as subclasses of `Provider` in `services/worker-manager/src/providers`, with their `providerType` defined in `index.js` in that directory.
+
+## Operation
+
+### Setup
+
+At start-up, the service reads the set of defined providers and creates a class instance for each one.
+The class constructor should set instance properties, but should not do anything that can fail, as such a failure would cause the process to exit, preventing other providers from operating.
+
+The async `setup` method is called after the constructor, and can be used to perform more sophisticated setup operations.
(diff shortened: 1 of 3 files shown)


services/worker-manager/providers.md
```diff
@@ -0,0 +1,117 @@
...
+The worker-manager runs a single "provisioner" process that runs the provisioning loop, where new workers are created to execute queued tasks.
+In each iteration of this loop, each provider is responsible for determining how many instances to start, and starting them.
+
+### Provisioning Workers
+
+The provisioning process begins by calling each provider's `initiate` method.
+This is similar to `setup`, but is not called in the web service.
+
+Similarly, when the provisioning process stops cleanly, it calls each provider's `terminate` method.
+Of course, clean stops are not the norm -- typically the process runs continuously until it crashes or is terminated -- so providers should not rely on `terminate` to clean up resources in production.
```
imbstack: I guess this isn't technically true since services are restarted at least once a day but we don't handle that by cleaning up.

djmitche:
I've removed a bunch of stuff from the service's README.md.  I'll add that back in in the service docs in another PR.

services/worker-manager/providers.md
djmitche: Yeah, even in that case I think the process is just SIGTERM'd, right?  So `terminate` wouldn't be called?

imbstack: Yeah, I believe that's what happens. We could start catching SIGTERM and doing cleanup on it, but we've never needed to.

PR #1005, opened 2019-07-11
Dev cluster from cloudops way

imbstack:
### HEADS UP: This changes the name of the azure account env var (among other things) in most services. updated heroku before landing!

This takes the new way of doing things with helm and brings into the tc repo development flow a bit more.

1. `yarn generate` builds a bunch of helm templates
2. it also makes a schema for config options, an example config file, and defaults
3. adds some niceties around helm deployments and such
4. Moves entrypoint generation out of the build process and checks it in
5. ~some stuff with the dockerfile I think I'll undo~
6. ~add the generated stuff to the gitattributes list of things to hide~

Hey look, it Works :tm: https://taskcluster.imbstack.com/

I still need to fill out 90% of the variables for services other than auth, but at least auth responds to ping and the ui shows up.

Heads up to @djmitche @owlishDeveloper @edunham @sciurus. Thanks to the cloudops folks, we could never have gotten this correctly without your guidance.

Things TODO:
- [X] Make the `helm` stuff in `yarn` scripts a bit more correct. It fails to update properly sometimes at the moment (ANSWER: Helm3 currently has a bug where upgrades don't work. For the shared dev cluster we can either wait for helm3 to fix itself or securely install helm2)
- [X] Figure out what we're going to do with terraform bits. We could replace them all with a script and forgo terraform entirely? (ANSWER: Get rid of them after cloudops takes them over)
- [X] Set up some sort of `yarn dev:init` that creates a config file for you with many of the per-service fields filled out such as rabbitmq passwords and such. The actual work of creating things would either be done by the script or perhaps terraform.
- [x] Set up a single medium-sized dev cluster for the taskcluster team to use. Each developer would have their own namespace there in which to deploy their taskcluster. This should both save us money and operational effort.
- [X] Figure out if this will actually work with the cloudops deployment stuff as-is or if there are things we'll need to tweak. (ANSWER: It seems like this will work with what they have. It will be a bit of work on their side but it can work)
- [x] ~Optionally get this working with minikube. I think the only thing we'd need is to make two ingress files and only apply one in `helm` depending on some sort of variable~
- [X] various cleanups and papercuts. This was made in a bit of a frenzy so lots of things could be renamed or cleared up.
- [X] Add docs on how to use this!

This is a huge diff but a lot of it is autogenerated so you can mostly ignore it. Things of interest:

- `infrastructure/k8s/values.schema.json`: helm 3 supports a schema file for input values natively! ~Right now the schema I've generated is a bit easy-going because our underlying configs don't do a good job of indicating that a config option is optional. I'd like to start where we are now and then tighten over time.~
- `user-config-example.yaml`: This probably needs a new name, but it is a list of every option you can pass into a taskcluster instance. At some point this should grow descriptions.


infrastructure/builder/src/generate/generators/k8s.js
```diff
@@ -0,0 +1,354 @@
...
+        context['deadlineSeconds'] = conf.deadline;
+        break;
+      }
+      default: continue; // We don't do anything with build/heroku-only
+    }
+    const rendered = jsone(templates[tmpl], context);
+
+    // json-e can't create a "naked" string for go templates to use to render an integer.
+    // thankfully this is the only case where this bites us (so far), so we can just do some
+    // post processing
```
djmitche: I'm not sure what this means -- is there an issue with escaping in json-e?

infrastructure/builder/src/generate/generators/k8s.js
```diff
@@ -0,0 +1,354 @@
...
+    };
+    switch (conf['type']) {
+      case 'web': {
+        tmpl = 'deployment';
+        context['needsService'] = true;
+        const rendered = jsone(templates['service'], context);
+        const file = `taskcluster-${name}-service-${proc}.yaml`;
+        ingresses.push({
+          projectName: `taskcluster-${name}`,
+          paths: conf['paths'] || [`/api/${name}/*`], // TODO: This version of config is only for gcp ingress :(
```
djmitche: Could Helm conditionals take care of this?

infrastructure/builder/src/generate/generators/k8s.js
```diff
@@ -0,0 +1,354 @@
...
+          description: 'The docker image containing taskcluster.',
+        },
+        ingressStaticIpName: {
+          type: 'string',
+          description: 'A google static ip object that the ingress can use to maintain the same ip address.',
+        },
+        ingressCertName: {
+          type: 'string',
+          description: 'A google certificate name that the ingress can use to set up tls.',
+        },
```
djmitche: Hm, more GCP-specific stuff..

infrastructure/builder/src/generate/generators/k8s.js
```diff
@@ -0,0 +1,354 @@
...
+      schema.required.push(confName);
+      schema.properties[confName] = {
+        type: 'object',
+        title: `Configuration options for ${cfg.name}`,
+        properties: {
+          procs: {
+            type: 'object',
+            title: 'Process settings for this service',
+            properties: {},
+            required: [],
```
djmitche: additionalProperties: false

infrastructure/builder/src/generate/generators/k8s.js
```diff
@@ -0,0 +1,354 @@
...
+          },
+        },
+        required: ['procs'],
+      };
+
+      // Some services actually duplicate their config env vars in multiple places
+      // so we de-dupe first. We use the variable name for this. If they've asked
+      // for the same variable twice with different types then this is not our fault
```
djmitche: Shouldn't tc-lib-config do this?

infrastructure/builder/src/generate/generators/k8s.js
```diff
@@ -0,0 +1,354 @@
...
+            required: ['memory', 'cpu'],
+            additionalProperties: false,
+          };
+        }
+      });
+    });
+
+    await writeRepoJSON(path.join(CHART_DIR, 'values.schema.json'), schema);
+    await writeRepoYAML(path.join(CHART_DIR, 'variables.yaml'), variablesYAML);
+    await writeRepoYAML('user-config-example.yaml', exampleConfig);
```
djmitche: Maybe this file should be in `deployment-docs/`?

infrastructure/builder/src/generate/generators/k8s.js
```diff
@@ -0,0 +1,354 @@
...
+        },
+      },
+      required: ['rootUrl', 'dockerImage'],
+      aditionalProperties: false,
+    };
+
+    // Something to copy-paste for users
+    const exampleConfig = {
+      rootUrl: '...',
+      dockerImage: '...',
```
djmitche: I liked that this sort of "automatically" selected the TC docker image version corresponding to the tf templates.  Is it possible to do the same here, to avoid the situation where users update the docker image and helm templates out-of-sync?  Maybe with some Helm/Go-Template conditionals to use the "official" docker image if an override isn't given?

infrastructure/builder/templates/k8s/secret.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: ${projectName}
+data:
+  $merge:
+    $map: {$eval: secrets}
+    each(s):
+      ${uppercase(s.key)}: '{{ ${s.val} | toJson | trimAll "\"" | b64enc }}'
```
djmitche: I don't think the `trimAll` will be compatible with `!env:json` config..

services/web-server/procs.yml
```diff
@@ -2,6 +2,11 @@ web:
   type: web
   command: node src/main.js server
   readinessPath: /.well-known/apollo/server-health
+  paths:
+    - '/login'
+    - '/login/*'
+    - '/graphql'
+    - '/graphql/*'
```
djmitche: `/subscription` too.  Also, I think `/graphql/*` is unnecessary.

package.json
```diff
@@ -166,8 +166,13 @@
     "generate": "node infrastructure/builder/src/main.js generate",
     "changelog": "node infrastructure/builder/src/main.js changelog",
     "shared-secrets": "node test/shared-secrets.js",
-    "heroku-prebuild": "echo $SOURCE_VERSION > taskcluster-version",
-    "heroku-postbuild": "if [ $PROCFILE = \"ui/Procfile\" ]; then cd ui && yarn install --frozen-lockfile && yarn build; fi"
+    "heroku-prebuild": "echo {\\\"version\\\"\\: \\\"$SOURCE_VERSION\\\"} > version.json",
+    "heroku-postbuild": "if [ $PROCFILE = \"ui/Procfile\" ]; then cd ui && yarn install --frozen-lockfile && yarn build; fi",
+    "COMMENT": "TODO: allow setting own deployment name below",
+    "dev:install": "helm install -f ./user-config.yaml -f ./infrastructure/k8s/variables.yaml taskcluster ./infrastructure/k8s",
```
sciurus: Change variables.yaml to values.yaml and you won't have to explicitly load it.

user-config-example.yaml
```diff
@@ -0,0 +1,187 @@
...
+  pulse_password: ...
+  pulse_hostname: ...
+  pulse_vhost: ...
+  aws_access_key_id: ...
+  aws_secret_access_key: ...
+  aws_region: ...
+  gcp_credentials: {}
+  test_bucket: ...
+built_in_workers:
+  taskcluster_client_id: ...
```
sciurus: There's no reason for people to customize these from the standard names like  "static/taskcluster/github" , is there?

I think all the `*_client_id` could go in variables.yaml.

infrastructure/builder/src/generate/generators/k8s.js
sciurus: If it's better for you, you could switch back to nginx ingress here and we could add in the gcp ingress on the cloudops side.

We're eventually going to switch away from straight gclb=>taskcluster to gclb=>iprepd-nginx=>taskcluster, which wouldn't make sense to bother supporting here since no one else will want to use iprepd-nginx,

infrastructure/builder/src/generate/generators/k8s.js
imbstack: We discussed this outside of this PR. The issue is that gotmpl uses {{}} to substitute things and yaml sees that as an object.

infrastructure/builder/src/generate/generators/k8s.js
imbstack: Yes! Created bug 1566928 for followup

infrastructure/builder/src/generate/generators/k8s.js
imbstack: I think in this PR we'll only support GKE and then we can come by in a later patch and make this generally useful.

infrastructure/builder/src/generate/generators/k8s.js
imbstack: I've now made it so that `yarn release` updates the default value in `values.yaml` for the helm chart. It should fall back to the most recent release if not overridden.

infrastructure/builder/templates/k8s/secret.yaml
imbstack: It only trims leading/trailing `"`. It seems to work empirically so far. This whole line is very ugly though and I won't be surprised when it doesn't work in some other way.

infrastructure/builder/src/generate/generators/k8s.js
imbstack: It probably should but it was easier to do it this way for now and hopefully lib-config goes away soon.

imbstack:
Ok, this is now "entirely working".

Some services don't run due to small config details that we still need to set up. I have faith that we can track these down in short order -- I am just tired of doing the tracking so hopefully someone can help me there.

There's also some quite manual hacks in the dev cluster deployment stuff. Hopefully someone who understands these services better than I can will help un-hack.

Other than that (and the fact that helm3 doesn't do updates correctly yet) I think this is actually usable!

services/auth/config.yml
```diff
@@ -78,7 +78,7 @@ defaults:
 
   azure:
     # Azure table storage account name
-    accountId:                !env AZURE_ACCOUNT
+    accountId:               !env AZURE_ACCOUNT_ID
```
djmitche: don't forget to update this in Heroku, too (for all services) :)

services/auth/src/data.js
```diff
@@ -118,6 +118,7 @@ Client.syncStaticClients = async function(clients = []) {
   for (const client of clients) {
     assert(typeof client.clientId === 'string', 'expected clientId to be a string');
     assert(typeof client.accessToken === 'string', 'expected accessToken to be a string');
+    assert(client.accessToken.length > 15, 'accessToken must have at least 15 chars');
```
djmitche: The schema specifies 22-66 characters..

package.json
```diff
@@ -169,10 +173,11 @@
     "heroku-prebuild": "echo {\\\"version\\\"\\: \\\"$SOURCE_VERSION\\\"} > version.json",
     "heroku-postbuild": "if [ $PROCFILE = \"ui/Procfile\" ]; then cd ui && yarn install --frozen-lockfile && yarn build; fi",
     "COMMENT": "TODO: allow setting own deployment name below",
-    "dev:install": "helm install -f ./user-config.yaml -f ./infrastructure/k8s/variables.yaml taskcluster ./infrastructure/k8s",
-    "dev:upgrade": "helm upgrade -f ./user-config.yaml -f ./infrastructure/k8s/variables.yaml taskcluster ./infrastructure/k8s --force --recreate-pods",
+    "dev:init": "node infrastructure/builder/src/main.js dev --init",
+    "dev:install": "helm install -f ./user-config.yaml taskcluster ./infrastructure/k8s",
```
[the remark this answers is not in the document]
imbstack: Yeah, let's change it. I just used it at first since it as already in `.gitignore`

infrastructure/builder/src/dev/index.js
```diff
@@ -0,0 +1,45 @@
...
+
+  await commonPrompts({userConfig, prompts, configTmpl});
+  await rabbitPrompts({userConfig, prompts, configTmpl});
+  await azurePrompts({userConfig, prompts, configTmpl});
+
+  let {meta, ...answer} = await inquirer.prompt(prompts);
+  let rabbitUsers = {};
+  if (meta) {
+    rabbitUsers = meta.rabbitAdminPassword;
+    delete meta.rabbitAdminPassword;
```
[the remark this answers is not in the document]
imbstack: Yeah, that's the idea. It definitely all looks weird. I think I'm going to land it without a cleanup and then we can iterate from there.

infrastructure/builder/src/dev/index.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  const prompts = [];
+
+  await commonPrompts({userConfig, prompts, configTmpl});
+  await rabbitPrompts({userConfig, prompts, configTmpl});
+  await azurePrompts({userConfig, prompts, configTmpl});
+
+  let {meta, ...answer} = await inquirer.prompt(prompts);
```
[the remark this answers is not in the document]
imbstack: It is just working around a way that inquirer has to nest answers. I fully commit to owning this code 100% until we get it cleaned up.

infrastructure/builder/src/dev/rabbit.js
```diff
@@ -0,0 +1,86 @@
...
+      }
+      return true;
+    },
+  });
+
+  prompts.push({
+    when: () => !userConfig.pulseVhost,
+    type: 'input',
+    name: 'pulseVhost',
+    default: previous => (previous.meta || {}).deploymentPrefix || (userConfig.meta || {}).deploymentPrefix,
```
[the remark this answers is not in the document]
imbstack: Yeah, it defaults to being the prefix for the rest of the cluster. This is only asked if the value for `pulseVhost` is undefined so that makes sense.

sciurus:
Thanks for adopting my suggestions. I took another look and still don't see any obvious problems with us using this. I did a brief test rendering the k8s templates for github using the helm templates from this PR and it worked fine. 

:ship:

imbstack:
Ok, I have an entirely working dev deployment from this now!

I'm going to update the heroku configs and land this first thing tomorrow.

PR #1007, opened 2019-07-11
Fixed Broken Links

kanurag94:
Fixed Broken links inside contributing.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,18 +3,18 @@
 We welcome pull requests from everyone. We do expect everyone to adhere to the [Mozilla Community Participation Guidelines][participation].
 
 If you're trying to figure out what to work on, here are some places to find suitable projects: 
-* [Good first bugs][https://codetribute.mozilla.org/projects/taskcluster?tag%3Dgood%2520first%2520issue]: these are scoped to make it easy for first-time contributors to get their feet wet with Taskcluster code.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -23,12 +23,12 @@ You can add new people to the list using the [@all-contributors bot][all-contrib
 
 [participation]: https://www.mozilla.org/en-US/about/governance/policies/participation/
 [issues]: ../../issues
-[bugsahoy]: https://www.joshmatthews.net/bugsahoy/?taskcluster=1
-[goodfirstbug]: http://www.joshmatthews.net/bugsahoy/?taskcluster=1&simple=1
+[bugsahoy]: https://codetribute.mozilla.org/projects/taskcluster?tag%3Dbug
+[goodfirstbug]: https://codetribute.mozilla.org/projects/taskcluster?tag%3Dgood%2520first%2520issue
```
imbstack: Interestingly it seems like codetribute has both `good-first-bug` and `good-first-issue`. Should we link to both here?

kanurag94: Oh, thanks for noticing. Yes I think we should link to both and add a describing comment to both. I'll update the PR

kanurag94: Both are the same thing, aren't they? Still we can link to both in case someone likes to explore

PR #1009, opened 2019-07-11
Provide user-oriented docs for providers

djmitche:
This has some open TODO's for @imbstack to answer :)

### DIFF
--- a/ui/docs/reference/core/worker-manager/google.md
+++ b/ui/docs/reference/core/worker-manager/google.md
@@ -1,0 +1,55 @@
+order: 10
+---
+import SchemaTable from 'taskcluster-ui/components/SchemaTable'
+
+# Google Provider Type
+
+Google-based providers create workers dynamically in GCE, using a single GCP project.
+Best practice is to use a dedicated project for each provider that is not shared with other uses.
+This isolates the workers from other GCP activities, protecting both from misuse or abuse.
+
+## Provider Configuration
+
+A google-based provider is be configured in `providers` with the following structure:
+
+```json
+{
+  "myProvider": {
+    "providerType": "google",
+    "project": "<gcp project identifier>",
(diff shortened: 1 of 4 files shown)


ui/docs/reference/core/worker-manager/google.md
```diff
@@ -0,0 +1,42 @@
...
+  },
+  ...
+}
+```
+
+The `project` configuration names the project in which instances will be created.
+It is the project *ID* and not the human-readable project name.
+
+The `instancePermissions` configuration defines the permissions granted to the GCP role assumed by the workers.
+TODO: what does this look like? does everyone need it?
```
imbstack: I can help document this better next week but it is an array of strings where each string is a gcp cloud iam permission string. Most users will probably want the permissions to write to stackdriver at least. At this time I can't think of anything else it would want.

ui/docs/reference/core/worker-manager/google.md
```diff
@@ -0,0 +1,42 @@
...
+
+The GCP credentials are provided either in string form (`creds`) or in a file (`credsFile`).
+In either case, the data is the large JSON object containing a service account's keys.
+TODO: what permissions must this service account possess?
```
imbstack: Ah, of course! It needs the roles:

* `roles/iam.serviceAccountAdmin`
* `roles/iam.roleAdmin`
* `roles/resourcemanager.projectIamAdmin`
* `roles/compute.admin`

Note that these are very powerful so this should probably be a project dedicated to just these workers.

djmitche: Thanks!

ui/docs/reference/core/worker-manager/google.md
djmitche: I summarized that.

PR #1013, opened 2019-07-12
Add a section on conventions to the design chapter

djmitche:
We have a bunch of conventions in place that are not quite "baked in" to the platform, but are assumed by things like the UI. So we should probably put them in one place.


djmitche:
> I wonder if we should do some work at some point to "bake these in" more since the ui does basically rely on them.

The UI degrades gracefully when they're not present.  I think we're OK as we are?

PR #1015, opened 2019-07-13
aws provider: other methods

owlishDeveloper:
You can ignore the first 3 commits - you already reviewed them, no changes there.

I put some questions/things I don't understand in the comments. One big question I have that's absent from the comments is: what is all that `operation` thing in `checkWorker` in google provider? I didn't understand that part at all. What is happening? (I assumed that it was something gcp-specific so I just ignored that part).

10-30% review please! Am I missing anything (functionality, error handling etc)?

### DIFF
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -174,7 +174,7 @@ class AwsProvider extends Provider {
         },
         MaxCount: toSpawn,
         MinCount: toSpawn,
-        ImageId: workerPool.config.imageId,
+        ImageId: workerPool.config.image[this.region],
         InstanceType: workerPool.config.instanceType,
         Placement: {
           AvailabilityZone: this.region,
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -156,9 +156,11 @@ class AwsProvider extends Provider {
       });
     }
 
+    const config = this.chooseConfig(workerPool.config);
+
     const toSpawn = await this.estimator.simple({
       workerPoolId,
-      ...workerPool.config,
+      ...config,
       running: workerPool.providerData[this.providerId].running,
     });
 
@@ -166,14 +168,13 @@ class AwsProvider extends Provider {
 
     try {
       spawned = await this.ec2.runInstances({
+        ...config,
         IamInstanceProfile: {
           Arn: this.instanceProfile.Arn,
           Name: this.instanceProfile.InstanceProfileName,
         },
-        MaxCount: toSpawn,
-        MinCount: toSpawn,
-        ImageId: workerPool.config.image[this.region],
-        InstanceType: workerPool.config.instanceType,
+        MaxCount: config.MaxCount || toSpawn,
+        MinCount: config.MinCount ? Math.min(toSpawn, config.MinCount) : toSpawn,
         Placement: {
           AvailabilityZone: this.region,
         },
@@ -276,6 +277,18 @@ class AwsProvider extends Provider {
         wp.providerData[this.providerId].running = seen;
       });
     }));
+  }
+
+  /**
+   * Method to select instance launch specification. At the moment selects at random
+   *
+   * @param possibleConfigs Array<Object>
+   * @returns Object
+   */
+  chooseConfig(possibleConfigs) {
+    const i = Math.random() * (possibleConfigs.length() - 1);
+
+    return possibleConfigs[i];
   }
 }
 
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -247,12 +247,12 @@ class AwsProvider extends Provider {
       switch (is.InstanceState.Name) {
         case 'pending':
         case 'running':
+        case 'shutting-down': //so that we don't turn on new instances until they're entirely gone
+        case 'stopping':
           this.seen[worker.workerPoolId] += 1;
           return Promise.resolve();
 
-        case 'shutting-down':
         case 'terminated':
-        case 'stopping':
         case 'stopped':
           return worker.modify(w => {w.state = this.Worker.states.STOPPED;});
 
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -1,7 +1,8 @@
-const {Provider} = require('./provider');
+const {ApiError, Provider} = require('./provider');
 const aws = require('aws-sdk');
 const _ = require('lodash');
 const taskcluster = require('taskcluster-client');
+const verify = require('iid-verify');
 
 class AwsProvider extends Provider {
   constructor({
@@ -17,6 +18,7 @@ class AwsProvider extends Provider {
     estimator,
     validator,
     notify,
+    ec2iid_DSA_SHA1_cert,
   }) {
     super({
       providerId,
@@ -33,6 +35,7 @@ class AwsProvider extends Provider {
     this.instancePermissions = instancePermissions;
     this.apiVersion = apiVersion;
     this.region = region;
+    this.ec2iid_DSA_SHA1_cert = ec2iid_DSA_SHA1_cert;
 
     aws.config.update({region});
     this.iam = new aws.IAM({apiVersion});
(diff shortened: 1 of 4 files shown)


services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,342 @@
...
+        state: this.Worker.states.REQUESTED,
+        providerData: { // what do we want to remember, anyways? 🤔
+          project: this.project,
+        },
+      });
+    }));
+  }
+
+  async deprovision({workerPool}) {
+    // should this be implemented in Provider? Looks like it's going to be the same for all
```
djmitche: This is probably wrong for the google provider, too -- I think it should do nothing, so the worker-manager just waits for the instances to terminate.  This changed in https://bugzilla.mozilla.org/show_bug.cgi?id=1555213 and I should have updated the google provider at that time.  I'll make a PR.

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,342 @@
...
+      wt.previousProviderIds = wt.previousProviderIds.filter(p => p !== this.providerId);
+      delete wt.providerData[this.providerId];
+    });
+    // why don't we remove workers from db or terminate instances?
+  }
+
+  async registerWorker({worker, workerPool, workerIdentityProof}) {
+    // check worker's identity -
+    // The way AWS works, workers will be getting temporary creds for accessing AWS APIs
+    // automatically. There's no token back-and-forth
```
djmitche: The worker will need to provide something to worker-manager to prove that it really is this worker -- otherwise anyone can call `workerManager.registerWorker` and get Taskcluster credentials for free.  When John was thinking about this he was looking at using [instance identity documents](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html) for this purpose.

services/worker-manager/src/providers/provider.js
```diff
@@ -61,7 +61,7 @@ class Provider {
   async scanPrepare() {
   }
 
-  async checkWorker({Worker}) {
+  async checkWorker({Worker}) { // todo: it seems we're actually getting worker object
```
djmitche: yep, this should be a lower-case "w"

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,342 @@
...
+    //todo: do something to enable logging to stackdriver
+  }
+
+  async provision({workerPool}) {
+    const {workerPoolId} = workerPool;
+
+    if (!workerPool.providerData[this.providerId] || workerPool.providerData[this.providerId].running === undefined) {
+      await workerPool.modify(wt => {
+        wt.providerData[this.providerId] = wt.providerData[this.providerId] || {};
+        wt.providerData[this.providerId].running = wt.providerData[this.providerId].running || 0; // why zero?
```
djmitche: Absent any other information, we assume there are zero running workers.

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,342 @@
...
+        },
+      });
+    } catch (e) {
+      await workerPool.reportError({
+        kind: 'creation-error',
+        title: 'Instance Creation Error',
+        description: e.message, // TODO: Make sure we clear exposing this with security folks
+        notify: this.notify,
+        WorkerPoolError: this.WorkerPoolError,
+      });
```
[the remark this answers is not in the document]
owlishDeveloper: oh I thought `workerPool.reportError` was a side effect function so returning it would make no difference, and the method would return anyways upon execution of the `catch` block so no reason to put `return` in there. I can add the keyword tho, sure

services/worker-manager/src/providers/aws.js
owlishDeveloper: oh thank you! I'll look into that, I was so confused about this. I had thought this method just helped getting google creds to call the google APIs

services/worker-manager/src/providers/provider.js
owlishDeveloper: this error comes from the google provider. I'll make a PR

services/worker-manager/src/providers/aws.js
owlishDeveloper: oh, I actually was looking at that documentation page on Thursday/Friday. So my understanding is, the worker doesn't need to even perform `curl http://169.254.169.254/latest/dynamic/instance-identity/document` - it's done automatically. Because I create a role with policies in `setup`, and that is attached to each instance through `InstanceProfile` https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html

You are mentioning the Taskcluster credentials though. What is the workflow there? I find myself not knowing which credentials will the workers need and how they usually get them

djmitche: > You are mentioning the Taskcluster credentials though. What is the workflow there? I find myself not knowing which credentials will the workers need and how they usually get them

Details are [in the docs](https://github.com/taskcluster/taskcluster/blob/master/services/worker-manager/providers.md#registering) but `registerWorker` is an API method that is called without any Taskcluster credentials, and returns Taskcluster credentials.  So it has to include *something* to prove that the caller is who they claim to be.  I'm not sure how that will work for AWS.

djmitche: [More docs here](https://docs.taskcluster.net/docs/reference/core/worker-manager/worker-interaction) :)

owlishDeveloper: Thank you!

services/worker-manager/src/providers/aws.js
```diff
@@ -1,17 +1,32 @@
...
-  constructor({WorkerPool, instancePermissions, apiVersion, region}) {
-    super({WorkerPool});
+  constructor({
+    WorkerPool,
+    instancePermissions,
+    apiVersion,
+    region,
+    estimator,
+    notify,
+    project,
```
imbstack: I think `project` is a google-specific thing iiuc how this is being used. Probably here it would be `account`?

services/worker-manager/src/providers/aws.js
```diff
@@ -230,7 +230,56 @@ class AwsProvider extends Provider {
...
+
+    Promise.all(instanceStatuses.map(is => {
+      switch (is.InstanceState.Name) {
+        case 'pending':
+        case 'running':
+          this.seen[worker.workerPoolId] += 1;
+          return Promise.resolve();
+
+        case 'shutting-down':
+        case 'terminated':
```
imbstack: I think we'll also mark it as stopped if it is terminated

services/worker-manager/src/providers/aws.js
owlishDeveloper: oh, right!

services/worker-manager/src/providers/aws.js
owlishDeveloper: these are the statuses that come aws. Or do you mean.. something else?

services/worker-manager/src/providers/aws.js
```diff
@@ -121,6 +135,151 @@ class AwsProvider extends Provider {
...
+    Promise.all(spawned.Instances.map(i => {
+      return this.Worker.create({
+        workerPoolId,
+        providerId: this.providerId,
+        workerGroup: this.providerId,
+        workerId: i.InstanceId,
+        created: new Date(),
+        expires: taskcluster.fromNow('1 week'),
+        state: this.Worker.states.REQUESTED,
+        providerData: { // what do we want to remember, anyways? 🤔
```
owlishDeveloper: @imbstack @djmitche I have one more question^  I guess this is information that either w-m or other services might need about the worker? Any use cases for this data? Or suggestions what to put in there? (maybe I'd just leave for now and add things as the need arises?)

imbstack: It is only for w-m to use itself. It doesn't get returned by any api calls. I use it to remember things about a worker I'll need later to confirm the request was completed, etc.

services/worker-manager/src/providers/aws.js
imbstack: Ah, just that down below you have something like `case 'stopped':` where you mark the worker state as `STOPPED`. We should do that when the aws status is `terminated` as well.

services/worker-manager/src/providers/aws.js
djmitche: I have a bug, or a note somewhere, to change the constructor syntax so that it doesn't mix provider config items (of which project is one) with parameters passed to all providers (estimator, notify, etc.)

services/worker-manager/src/providers/aws.js
owlishDeveloper: yep, that's exactly how this switch is supposed to work

imbstack: Ohhhh, duh. I was not reading that correctly. I wonder if we should keep `shutting-down` and `stopping` in the running state so that we don't turn on new instances until they're entirely gone?

owlishDeveloper: Sounds good to me!

services/worker-manager/src/providers/aws.js
owlishDeveloper: just to make sure I understand: so we shouldn't delete `providerData` from the db?

djmitche: correct

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,342 @@
...
+        case 'stopped':
+          return worker.modify(w => {w.state = this.Worker.states.STOPPED;});
+
+        default:
+          return Promise.reject(`Unknown state: ${is.InstanceState.Name} for ${is.InstanceId}`);
+      }
+    }));
+  }
+
+  // should this be implemented on Provider? Looks like it's going to be the same for all providers
```
owlishDeveloper: @imbstack I found unanswered question^

djmitche: I think once we have a few providers in their final form, we'll find a number of good places to refactor, this being one of them -- but it's ok, and probably wise given time pressures, to wait to make those refactors later.

PR #1017, opened 2019-07-14
Bug 1523807 - bring tc-cli into the monorepo

djmitche:
This incorporates the tc-cli repo into the tree.  It leaves out a few things:
 * `taskcluster status` -- this was the thing "scraping" the ping endpoints, which no longer works with the newer manifests
 * `taskcluster pulse` -- this used the old tc-events API which no longer exists
 * `taskcluster shell` -- this depends on some old libraries from tc-worker that bring in lots of ancient stuff

I'll file bugs to re-implement the shell and pulse functionality.

Bugzilla Bug: [1523807](https://bugzilla.mozilla.org/show_bug.cgi?id=1523807)


infrastructure/builder/src/generate/generators/tc-client-go.js
```diff
@@ -4,12 +4,10 @@ const exec = util.promisify(require('child_process').execFile);
 const {REPO_ROOT, readRepoFile, execCommand} = require('../../utils');
 
 exports.tasks = [{
-  title: 'Generate Taskcluster-Client-Go',
-  requires: ['references-json'],
-  provides: ['target-taskcluster-client-go'],
+  title: 'Check Go Version',
```
owlishDeveloper: (Not related to this PR, but I wonder if this would make a useful meta-test for the repo?)

djmitche: Maybe, but it's a test of the environment (what version of go is installed) rather than of the source code.

PR #1018, opened 2019-07-14
added confirmation dialog client view

shubhamgupta2956:
added confirmation dialog box on delete button for client view.

### DIFF
--- a/ui/src/views/Clients/ViewClient/index.jsx
+++ b/ui/src/views/Clients/ViewClient/index.jsx
@@ -88,12 +88,10 @@ export default class ViewClient extends Component {
   handleDeleteClient = async clientId => {
     this.setState({ dialogError: null, loading: true });
 
-    await this.props.client.mutate({
+    return this.props.client.mutate({
       mutation: deleteClientQuery,
       variables: { clientId },
     });
-
-    this.props.history.push(`/auth/roles`);
   };
 
   handleDialogActionError = error => {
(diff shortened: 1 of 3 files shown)


ui/src/views/Clients/ViewClient/index.jsx
```diff
@@ -84,20 +86,22 @@ export default class ViewClient extends Component {
...
-        mutation: deleteClientQuery,
-        variables: { clientId },
-      });
+    await this.props.client.mutate({
+      mutation: deleteClientQuery,
+      variables: { clientId },
+    });
 
-      this.setState({ error: null, loading: false });
+    this.props.history.push(`/auth/roles`);
```
helfi92: This line should be done inside `handleDialogActionComplete` since it has nothing to do with the asynchronous action.

ui/src/views/Clients/ViewClient/index.jsx
```diff
@@ -84,20 +86,22 @@ export default class ViewClient extends Component {
...
 
-    try {
-      await this.props.client.mutate({
-        mutation: deleteClientQuery,
-        variables: { clientId },
-      });
+    await this.props.client.mutate({
+      mutation: deleteClientQuery,
+      variables: { clientId },
+    });
```
helfi92: Remove the `async-await` stuff in this function and just use return. The `DialogAction` component will take care of running this function asynchronously. I think if you put `return this.props.client.mutate(...);` then the dialog will start showing a spinner above the "Delete Client" button.

shubhamgupta2956:
Sorry for delay @helfi92. I got a bit busy as my college started and I have to shift to my new room. I have changed the things as you have said. I still can't find the spinner. Maybe the action is faster as compared to redirecting. Please tell me about any other changes to be made.

PR #1020, opened 2019-07-14
Fix the tools.taskcluster.net meta-test

djmitche:
The previous version used `glob` over the whole repository and as a
result found instances of tools.taskcluster.net in `node_modules` and
other non-checked-in files.  And, it missed dotfiles like `.taskcluster.yml`.

### DIFF
--- a/test/meta_test.js
+++ b/test/meta_test.js
@@ -169,15 +169,11 @@ suite('Repo Meta Tests', function () {
       throw new Error(errors);
     }
   });
+
   test("no references to tools.taskcluster.net in the repository", async function () {
-    const filenames = glob.sync(
-      '**/*',
-      { cwd: ROOT_DIR });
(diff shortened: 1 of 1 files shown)


test/meta_test.js
```diff
@@ -198,21 +194,33 @@ suite('Repo Meta Tests', function () {
...
+    }
+
+    const files = new Set(res.stdout
+      .split('\n')
+      .map(l => l.slice(0, l.indexOf(':')))
+      .filter(f => f !== '')
+      .filter(f => !whitelist.has(f)));
+
+    if (files.size > 0) {
+      console.log(files);
```
imbstack: stray `console.log`

djmitche: I really should grep for those..

subhamskm:
LGTM

Issue #705, opened 2019-05-06
[UI] Update date package (date-fns)


helfi92:
This task is about updating `date-fns`.

**Current version**: `2.0.0-alpha.7`.
**Wanted version**: `2.0.0-alpha.27`.

You can have access to the docs at https://date-fns.org/docs/Getting-Started. There's a neat feature where you view older doc versions in the top right corner of the page.

![Screen Shot 2019-05-06 at 4 15 51 PM](https://user-images.githubusercontent.com/3766511/57252543-5c35c880-701a-11e9-9d38-d95ff624bdae.png)

Accessing the [releases page](https://github.com/date-fns/date-fns/releases) can be handy to know what changed between versions.

sagalbot:
@helfi92 I have a branch with the `date-fns` imports updated to work with `v1.30.1` (https://github.com/sagalbot/taskcluster/tree/bfa6270069f662e6ff5ca1c661ec7a66b54b814e).

However, it appears that `material-ui-pickers` [only supports `v2` of `date-fns`](https://material-ui-pickers.dev/getting-started/installation#peer-library). This is consistent through `v1` to `v3-beta` of the pickers library.

Can you confirm the requirement of getting `date-fns v1.30.1` in place?

helfi92:
👋 @sagalbot 

Apologies for that. I made a mistake with specifying the versions :( What we want is to actually go to the latest release of v2. In other words `2.0.0-alpha.7` -> `2.0.0-alpha.27`. I'll update the initial comment. Sorry for the trouble.

sagalbot:
@helfi92 no problem, that makes more sense! Should be simpler than the v2 -> v1 stable downgrade. I'll take a look at it this evening.

shubhamgupta2956:
Hi,
I am new to open source and wants to take this issue. I know the basics of js. Please guide me on how to start

helfi92:
Hi @shubhamgupta2956. Let's work on https://github.com/taskcluster/taskcluster/issues/578 first then we can look at other issues. Hope that's okay with you.

helfi92:
@sagalbot How is it going? Let me know if there's anything I can do to help :-)

ruchikabgosain:
Hey @helfi92 are there still changes needed in this one? Let me know i'll be happy to take this up :)

helfi92:
I haven't heard from @sagalbot. Feel free to give it a try.

arku:
[...]

PR #1029, opened 2019-07-15
Upgrade to Material UI v4

### LINKED ISSUE #895: [UI] Migrate to material-ui v4

arku:
Closes #895

I followed the steps described in https://material-ui.com/guides/migration-v3. I couldn't get the stuff mentioned under https://material-ui.com/guides/migration-v3/#list though.

@helfi92 How do we go about testing visual regressions (if there are any)?

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -17,9 +17,10 @@
     "create:fragment-matcher": "dotenv node src/scripts/queryServer"
   },
   "dependencies": {
-    "@material-ui/core": "^3.9.3",
-    "@material-ui/icons": "^3.0.2",
-    "@material-ui/lab": "3.0.0-alpha.30",
+    "@material-ui/core": "^4.2.1",
+    "@material-ui/icons": "^4.2.1",
+    "@material-ui/lab": "^4.0.0-alpha.20",
+    "@material-ui/styles": "^4.2.1",
     "@mdx-js/react": "^1.0.20",
     "@mdx-js/runtime": "^1.0.20",
     "@mozilla-frontend-infra/components": "^2.4.0",
@@ -75,6 +76,7 @@
     "react-apollo": "^2.5.5",
     "react-copy-to-clipboard": "^5.0.1",
     "react-dom": "^16.8.6",
+    "react-event-listener": "^0.6.6",
     "react-fout-stager": "^3.0.0",
     "react-ga": "^2.5.3",
     "react-helmet": "^5.2.1",
--- a/ui/src/views/AwsProvisioner/ViewAwsWorkerType/index.jsx
+++ b/ui/src/views/AwsProvisioner/ViewAwsWorkerType/index.jsx
@@ -32,7 +32,7 @@ import terminateWorkerTypeQuery from './terminateWorkerType.graphql';
     ...theme.mixins.fab,
   },
   spinner: {
-    marginTop: theme.spacing.double,
+    marginTop: theme.spacing(2),
   },
   terminateButton: {
     ...theme.mixins.errorIcon,
@@ -125,7 +125,7 @@ export default class ViewAwsWorkerType extends Component {
     return (
       <Dashboard title={`AWS Provisioner ${workerType}`}>
         <ErrorPanel fixed error={this.state.error || error} />
-        <Tabs fullWidth value={currentTab} onChange={this.handleTabChange}>
+        <Tabs value={currentTab} onChange={this.handleTabChange}>
           <Tab label="Status" />
           <Tab label="Recent Provisioning Errors" />
           <Tab label="EC2 Resources" />
--- a/ui/src/views/Documentation/components/Anchor/index.jsx
+++ b/ui/src/views/Documentation/components/Anchor/index.jsx
@@ -9,29 +9,33 @@ const useStyles = withStyles(theme => ({
     ...theme.mixins.link,
   },
 }));
+// Ref: https://material-ui.com/guides/composition/#caveat-with-refs
+// eslint-disable-next-line react/display-name
+const Anchor = React.forwardRef(
+  ({ classes, href, children, ...props }, ref) => {
+    if (href.startsWith('http')) {
+      return (
+        <a
+          className={classes.link}
+          href={href}
+          {...props}
+          target="_blank"
+          rel="noopener noreferrer"
+          ref={ref}>
+          {children}
+        </a>
+      );
+    }
 
-function Anchor({ classes, href, children, ...props }) {
-  if (href.startsWith('http')) {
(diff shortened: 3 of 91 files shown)


ui/package.json
```diff
@@ -75,6 +76,7 @@
     "react-apollo": "^2.5.5",
     "react-copy-to-clipboard": "^5.0.1",
     "react-dom": "^16.8.6",
+    "react-event-listener": "^0.6.6",
```
arku: @helfi92 I had to install this package (dependency of `material-ui-pickers`) as I was running into `Can't resolve 'react-event-listener'` error when running the app. Ideally, we should migrate from `material-ui-pickers` to `@material-ui/pickers` v3.

ui/src/views/Documentation/components/Anchor/index.jsx
```diff
@@ -9,29 +9,33 @@ const useStyles = withStyles(theme => ({
     ...theme.mixins.link,
   },
 }));
+// Ref: https://material-ui.com/guides/composition/#caveat-with-refs
+// eslint-disable-next-line react/display-name
```
arku: To fix the lint rule, I should use a normal `function` as the callback but I am unable to do so as we have another lint rule which requires arrow functions as callbacks and not normal functions.

helfi92:
Wow, this is great. I will review soon. This is going to be so awesome once merged :)

ui/package.json
```diff
@@ -17,9 +17,10 @@
     "create:fragment-matcher": "dotenv node src/scripts/queryServer"
   },
   "dependencies": {
-    "@material-ui/core": "^3.9.3",
+    "@material-ui/core": "^4.2.0",
     "@material-ui/icons": "^3.0.2",
```
helfi92: We could also bump `@material-ui/icons`.

ui/package.json
```diff
@@ -17,9 +17,10 @@
...
     "@material-ui/lab": "3.0.0-alpha.30",
```
helfi92: `@material-ui/lab` could probably also be bumped to the alpha version of v4. We use a couple of components from there. I hope there isn't anything critical to change. The CHANGELOG should be able to say what changed.

ui/package.json
helfi92: Agreed we should migrate. I think what's blocking us from migrating to the v3 version of the pickers package is https://github.com/taskcluster/taskcluster/issues/705.

ui/src/views/AwsProvisioner/ViewAwsWorkerType/index.jsx
```diff
@@ -125,7 +125,7 @@ export default class ViewAwsWorkerType extends Component {
     return (
       <Dashboard title={`AWS Provisioner ${workerType}`}>
         <ErrorPanel fixed error={this.state.error || error} />
-        <Tabs fullWidth value={currentTab} onChange={this.handleTabChange}>
+        <Tabs value={currentTab} onChange={this.handleTabChange}>
```
helfi92: What was the reason for removing `fullWidth`?

helfi92:
> @helfi92 How do we go about testing visual regressions (if there are any)?

We will have to do this manually for now.

ui/src/views/AwsProvisioner/ViewAwsWorkerType/index.jsx
arku: @helfi92 The deprecated `fullWidth` prop is removed in v4. 

Ref: https://material-ui.com/guides/migration-v3/#tabs

arku:
@helfi92 Thanks for catching the regressions. Let me fix those.

helfi92:
How is it going @arku? Given that this pull-request affects multiples files, the rebase might be painful if we wait too long. There's no rush but just wanted to bring this up so that we can prevent having to do a painful rebase :)

arku:
@helfi92 My bad, forgot this one.

Let me try fixing the regressions.

arku:
@helfi92 I have spent more time on this than I wanted to and it's time to get some help, I guess. Here are my observations regarding the regressions.

### Table row height

Notes from Migration docs:
> [TableRow] Remove the fixed height CSS property. The cell height is computed by the browser using the padding and line-height.

The table rows no longer have a fixed height of 48px. But this is not the only issue. For instance, let's consider the [`HookLastFiredTable`](https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/HookLastFiredTable/index.jsx) component. The [`TableCell`](https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/HookLastFiredTable/index.jsx#L36) component  has a padding of `14px 40px 14px 16px` while the same from v3 has a padding of `4px 56px 4px 24px`. And the [`ListItemText`](https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/HookLastFiredTable/index.jsx#L42) component has a top and bottom margin of `4px` while in v3, it has no margin. I noticed the `div` rendered by the `ListItemText` component has a CSS class of `MuiListItemText-dense` (which should set the margin to 0) but the CSS styles have no reference to it :confused: 

To make the sure the table looks the same as in v3, we have to add styles to the `TableCell` and `ListItemText`. This would mean we have to do the same for all `TableCell`s in all `Table` components and the same for `ListItemText` components. I am not sure if this is the right way to go about it though. What do you think?

### Font size

Notes from Migration docs:

> [Typography] Change the default variant from body2 to body1. A font size of 16px is a better default than 14px. Bootstrap, material.io, and even our documentation use 16px as a default font size. 14px like Ant Design uses is understandable, as Chinese users have a different alphabet. We recommend 12px as the default font size for Japanese.

I have updated the variants for the `Typography` components. I noticed a style of `font-size: 0.875rem;` added to `body` but I don't understand why it is injected. We should figure out why it is injected and remove it maybe?

### Color

A style of `rgba(255, 255, 255, 0.7)` is added to `html` and `body`.

Upon some digging, it looks like the [`textSecondary`](https://github.com/taskcluster/taskcluster/blob/master/ui/src/theme.js#L39) color from our theme is what's used for above style. It should be the `textPrimary` color though.

### Drawer

I noticed a spacing issue in the docs site after the upgrade but I didn't have any luck fixing it :disappointed: 

docs_spacing_issue

Do you mind helping me out with these issues? Once these are addressed, we have to check all the pages manually.

helfi92:
Thank you for the all of the information 🥇 I plan on taking a look as soon as I get the chance.

arku:
@helfi92 Sounds good, thanks!

helfi92:
Continues in https://github.com/taskcluster/taskcluster/pull/1729.

PR #1037, opened 2019-07-15
document google provider's definition of workerId

djmitche:


djmitche:
Hang on, I'll throw in the worker identity proof docs here too.

Issue #1038, opened 2019-07-15
[web-server] Stop generating the JWT token


helfi92:
The secops folks recommend maintaining a session via cookies.

PR #1039, opened 2019-07-15
[web-server] Rely on a session instead of generating a JWT

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/1038.

TODO:
* [ ] Update `dev-docs/login-process.md` after this gets approved


services/web-server/config.yml
```diff
@@ -58,10 +58,8 @@ defaults:
     #   clientId: !env GITHUB_CLIENT_ID
     #   clientSecret: !env GITHUB_CLIENT_SECRET
     strategies: !env:json UI_LOGIN_STRATEGIES
-    jwt:
-      # this should be a random string, and is used as input to the HS256 JWT
-      # signing algorithm
-      key: !env JWT_KEY
+    # this should be a random string, and is used to create the session ID
+    sessionSecret: !env SESSION_SECRET
```
djmitche: This will need to be mentioned in the changelog, since deployers will need to update their config.

services/web-server/src/servers/createApp.js
```diff
@@ -1,13 +1,20 @@
...
 const cors = require('cors');
 const express = require('express');
 const playground = require('graphql-playground-middleware-express').default;
 const passport = require('passport');
+const url = require('url');
 const credentials = require('./credentials');
 
 module.exports = async ({ cfg, strategies }) => {
   const app = express();
+  const store = new MemoryStore({
```
djmitche: Will using a memory store mean that users are logged out every time we restart the web-server process?  Could we create an `AzureStore` that uses an Azure table?  Would that end up hitting Azure for every GraphQL query (even though it's only needed for getCredentials)?

services/web-server/src/servers/createApp.js
```diff
@@ -23,7 +30,27 @@ module.exports = async ({ cfg, strategies }) => {
...
+    unset: 'destroy',
+    // Force a session identifier cookie to be set on every response.
+    // The expiration is reset to the original maxAge,
+    // resetting the expiration countdown.
+    rolling: true,
+    cookie: {
+      secure: url.parse(cfg.app.publicUrl).hostname !== 'localhost',
+      httpOnly: true,
+      // 15 minutes
+      maxAge: 15 * 60 * 5000,
```
djmitche: Would this mean the user is logged out of their session every 15 minutes without activity?

ui/.neutrinorc.js
```diff
@@ -49,6 +49,9 @@ module.exports = {
           '/login': {
             target: 'http://localhost:3050',
           },
+          '/logout': {
```
djmitche: This will need to get added to the k8s ingress, too.  Maybe it's easier to just use `/login/logout`?

services/web-server/src/servers/createApp.js
```diff
@@ -43,8 +70,26 @@ module.exports = async ({ cfg, strategies }) => {
...
-  passport.serializeUser((user, done) => done(null, user));
-  passport.deserializeUser((obj, done) => done(null, obj));
+  passport.serializeUser((user, done) => {
+    const isGithub = user.profile.provider === 'github';
+    const userId = isGithub ? user.profile.id : user.profile.user_id;
+
+    return done(null, {
+      identityProviderId: isGithub ? 'github' : 'mozilla-auth0',
+      userId,
+    });
```
djmitche: Rather than special-case github here (since we'll eventually have other strategies..), could this just include the identity string (from `user.identity`)?

services/web-server/src/servers/createApp.js
helfi92: > Will using a memory store mean that users are logged out every time we restart the web-server process?

Good question. I think it will and that's probably not the best design choice.

> Could we create an AzureStore that uses an Azure table?

Yes. I can look into that.

> Would that end up hitting Azure for every GraphQL query (even though it's only needed for getCredentials)?

Yes since we have `rolling: true` which might cause performance issues. Another way is to probably only regenerate the session when `getCredentials` is called.

services/web-server/src/servers/createApp.js
helfi92: Yeah. We should probably make this time longer. Perhaps 1 week?

djmitche: That makes sense.

services/web-server/src/servers/createApp.js
helfi92: Good idea. Not sure why I didn't think about that :)

services/web-server/src/servers/createApp.js
helfi92: > Could we create an AzureStore that uses an Azure table?

@djmitche, Do we happen to use `AzureStore` somewhere else in the codebase? I'm trying to find some documentation on this so I can understand on how to implement this.

djmitche: I made up the name, so no.  We do use Azure tables, though.  References:
 * https://github.com/taskcluster/azure-entities
 * https://github.com/taskcluster/taskcluster/tree/master/libraries/azure#readme

example:
https://github.com/taskcluster/taskcluster/blob/d135176b954fe411faebccd0f13d051f7ead667c/services/secrets/src/data.js#L5-L15

https://github.com/taskcluster/taskcluster/blob/d135176b954fe411faebccd0f13d051f7ead667c/services/secrets/src/main.js#L37-L51

With that, the `Secret` object defined in `main.js` has methods `load`, `create`, `scan` etc.

djmitche:
I've flagged AJ here too, since this stems from his suggestion that we implement sessions :)

ajvb:
This all looks good to me, two questions:

1. Where is `req.user` getting set? Is passport integrating with express session in some way?
2. Are there tests around the unauthorizedError path? (https://github.com/taskcluster/taskcluster/pull/1039/files#diff-9398efadc01e4bf3317b02efe6061f1eR15)

services/web-server/src/servers/createApp.js
```diff
@@ -23,7 +26,32 @@ module.exports = async ({ cfg, strategies }) => {
...
   app.use(cors({origin: allowedCORSOrigins}));
 
+  // Use MemoryStore locally (default behavior when a store is not set)
+  const store = process.env.NODE_ENV === 'production' ? AzureTablesStoreFactory.create({
+    table: cfg.azure.tableName,
+    storageAccount: cfg.azure.accountId,
+    accessKey: cfg.azure.accountKey,
+    // Delete expired sessions every 1 hour
+    sessionTimeOut: 60,
+  }) : undefined;
```
djmitche: Bad news incoming:

It looks like `connect-azuretables` doesn't seem to support SAS signatures, which is what we use for controlling access to azure resources.  Providing a storage account and access key means that the application can access any table in that account -- including the secrets service's table, for example.  No bueno..

I also suspect that we will be wanting to add API methods to manage sessions soon (sec folks will want a way to revoke all sessions for a user), and this library's use of a private table wouldn't permit that.

I also suspect that we'll be adding additional Azure tables in web-server, to support the third-party logins process (tables of access_tokens, for example).  So using the existing TC azure infrastructure will make sense.

One option here would be to modify the `connect-azuretables` library to accept a function that returns a SAS signature, which is conveniently what [tc-lib-azure](https://github.com/taskcluster/taskcluster/tree/master/libraries/azure#readme) produces and what [azure-entities expects](https://github.com/taskcluster/azure-entities#sas-function).

services/web-server/src/utils/login.js
```diff
@@ -1,5 +1,9 @@
-// A function that returns a middleware that renders callback.ejs
-module.exports = publicUrl => (request, response) => {
+/**
+ * A function that returns a middleware that renders callback.ejs
+ * and saves the session in a table so that the user is not logged out
+ * when the server restarts.
```
djmitche: does it do this second part?

services/web-server/src/utils/regenerateSession.js
```diff
@@ -0,0 +1,13 @@
+// Regenerate a sessionId given a request
+// https://stackoverflow.com/a/26394156/6150432
```
djmitche: Even reading this stackoverflow, I can't quite figure out what this is for.. help?

services/web-server/src/servers/createApp.js
helfi92: Adding API methods to manage sessions and additional Azure tables makes a lot of sense.

> One option here would be to modify the connect-azuretables

Would you suggest we fork `connect-azuretables` or create a new library that wraps `connect-azuretables`?

> to accept a function that returns a SAS signature, which is conveniently what tc-lib-azure produces and what azure-entities expects.

I'll look into it.

djmitche: From what I saw, it can't be wrapped (SAS credentials don't last forever so they need to be "fresh", which is why azure-entities supports a callback to get them).  Forking might be a short-term option but we should probably try to upstream the change.  One risk there is is that it seems a bit out of date -- it's using the v2 Azure libraries while v10 is the latest, for example.  It's not clear to me from the docs whether V2 supports SAS signatures.

helfi92: Hmm, perhaps we shouldn't use that library if it's using a very old version of Azure libraries. The list of compatible stores can be found [here](https://github.com/expressjs/session#compatible-session-stores). We could technically create our own Azure store by implementing the required and recommended methods but I fear it might slow us down. Maybe it would make sense to use [memorystore](https://www.npmjs.com/package/memorystore) and accept that we will be logged out when web-server is deployed which shouldn't happen often after this whole login work is done. Thoughts?

djmitche: That seems like a great plan.  Updating or rewriting the Azure connector might be a good project for a contributor during that time, too.

services/web-server/src/utils/login.js
helfi92: It doesn't anymore. Removing that now.

services/web-server/src/utils/regenerateSession.js
helfi92: Passport adds a `passport` key to `req.session`. When `req.session.regenerate` runs,  `req.session.passport` becomes undefined since `req.session` gets recreated. The callback makes the necessary changes to make passport aware of the current user.

I'll add a comment.

helfi92:
> Where is req.user getting set? Is passport integrating with express session in some way?

Yes, the passport library sets `req.user` after authenticating the user.

> Are there tests around the unauthorizedError path? (https://github.com/taskcluster/taskcluster/pull/1039/files#diff-9398efadc01e4bf3317b02efe6061f1eR15)

There wasn't but now there is.

services/web-server/src/utils/regenerateSession.js
```diff
@@ -0,0 +1,15 @@
+// Regenerate a sessionId given a request
+module.exports = req => new Promise((resolve) => {
+  const temp = req.session.passport;
+
+  req.session.regenerate((err) => {
+    // Passport adds a passport key to req.session. When req.session.regenerate runs,
+    // req.session.passport becomes undefined because req.session is recreated.
+    // The callback makes the necessary changes to make passport aware of the current user.
+    // For more information, see https://stackoverflow.com/a/26394156/6150432
```
djmitche: Thanks!

PR #1044, opened 2019-07-16
Bug 1563341- require static workers be managed manually, identified by secrets

djmitche:
Bugzilla Bug: [1563341](https://bugzilla.mozilla.org/show_bug.cgi?id=1563341)


djmitche:
I don't think that codecov/patch thing is accurate.  At least, clicking on it takes me to some random changes in the queue service.

PR #1045, opened 2019-07-16
Remove reference to tools.taskcluster.net in API tests

arku:

### DIFF
--- a/services/notify/test/api_test.js
+++ b/services/notify/test/api_test.js
@@ -57,7 +57,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['taskcluster', 'aws'], function(m
       address: 'success@simulator.amazonses.com',
       subject: 'Task Z-tDsP4jQ3OUTjN0Q6LNKQ is Complete',
       content: 'Task Z-tDsP4jQ3OUTjN0Q6LNKQ is finished. It took 124 minutes.',
-      link: {text: 'Inspect Task', href: 'https://tools.taskcluster.net/task-inspector/#Z-tDsP4jQ3OUTjN0Q6LNKQ'},
+      link: {text: 'Inspect Task', href: 'https://taskcluster.net/task-inspector/#Z-tDsP4jQ3OUTjN0Q6LNKQ'},
     });
     helper.checkEmails(email => {
       assert.deepEqual(email.delivery.recipients, ['success@simulator.amazonses.com']);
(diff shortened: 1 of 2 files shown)


services/notify/test/api_test.js
```diff
@@ -57,7 +57,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['taskcluster', 'aws'], function(m
       address: 'success@simulator.amazonses.com',
       subject: 'Task Z-tDsP4jQ3OUTjN0Q6LNKQ is Complete',
       content: 'Task Z-tDsP4jQ3OUTjN0Q6LNKQ is finished. It took 124 minutes.',
-      link: {text: 'Inspect Task', href: 'https://tools.taskcluster.net/task-inspector/#Z-tDsP4jQ3OUTjN0Q6LNKQ'},
+      link: {text: 'Inspect Task', href: 'https://taskcluster.net/task-inspector/#Z-tDsP4jQ3OUTjN0Q6LNKQ'},
```
arku: @djmitche The `href` here can be any valid URL. Should we bother changing it to something else?

Btw, if we do want to replace references to tools.taskcluster.net with the new app, can we use taskcluster-ui.herokuapp.com?

djmitche: You're right, it could be anything, since it's just in a test.  In general the example / testing rootUrl is `https://tc.example.com`, so you could use `https://tc.example.com/tasks/Z-tDsP4jQ3OUTjN0Q6LNKQ` here.

djmitche: `taskcluster-ui.herokuapp.com` is a temporary hostname, so we should not add additional references to it in the codebase.

djmitche:
@all-contributors please add @arku for code. :balloon:

helfi92:
I thought they were already in the contributors list. @arku has been doing some amazing work!

services/notify/test/api_test.js
arku: > taskcluster-ui.herokuapp.com is a temporary hostname, so we should not add additional references to it in the codebase.

@djmitche  Right. Will we be using `https://taskcluster.net` in the future to host the app? I noticed it redirects to https://docs.taskcluster.net/docs at the moment. I am asking because I am not sure what hostname to replace the `tools.taskcluster.net` references with.

For instance, In https://docs.taskcluster.net/docs/reference/core/notify/usage#notification-types, we have the following:

> Inspect: https://tools.taskcluster.net/task-inspector/#f0rU3kS7RmG3xSWwbq6Ndw

What should I replace this URL with? The pathname should be `/tasks/f0rU3kS7RmG3xSWwbq6Ndw` but what about the hostname?

djmitche: There won't be any single deployment of Taskcluster anymore.  In examples in tests and documentation, we can use a rootUrl of `https://tc.example.com`.   For links to documentation, we can use `https://docs.taskclsuter.net/docs`.

arku: Thanks @djmitche!

PR #1049, opened 2019-07-17
Support a "category" for API methods 

subhamskm:
Bugzilla Bug: [1551848](https://bugzilla.mozilla.org/show_bug.cgi?id=1551848)

### DIFF
--- a/libraries/api/src/index.js
+++ b/libraries/api/src/index.js
@@ -191,6 +191,7 @@ class APIBuilder {
           input: entry.input,
           output: entry.output,
           description: entry.description,
+          category: entry.category
         };
         if (entry.scopes) {
           retval.scopes = entry.scopes;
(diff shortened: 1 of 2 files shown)


subhamskm:
> This is looking right so far, but the tests are failing -- can you figure out why?

The error is saying that there is no category in references.entries[0]. But I have added category option in entry. I am not getting why it is not able to generate the references file with category option.

djmitche:
Yeah, that's confusing for me, too.  I'd suggest adding some `console.log` to see what that data structure actually looks like at that point, and following the trail from there (for example, if "category" isn't there -- why?  If it is there, why is the schema validation unhappy?).

I tend to use `console.log` but other people love to use debuggers, and running that test case under a debugger is also a good approach.

By the way, you can run a single test for the index service, for example, with
```
cd services/index
mocha -b test/references_test.js
```

subhamskm:
Due to some issues, I want to move this PR to #1082.

PR #1055, opened 2019-07-18
Minor release fixups

djmitche:
see commit messages..

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -3,290 +3,292 @@ reporting: checks-v1
 policy:
   pullRequests: public
 tasks:
-  $let:
-    node: '10.16.0'
-    go: '1.12.6'
-    artifactsDir:
-      public:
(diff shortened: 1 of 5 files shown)


.taskcluster.yml
```diff
@@ -3,290 +3,292 @@ reporting: checks-v1
...
-    node: '10.16.0'
-    go: '1.12.6'
-    artifactsDir:
-      public:
-        # This is not used at the moment other than by the trailer task to upload results. Feel free to bump later
-        expires: {$fromNow: '24 hours'}
-        path: /taskcluster/artifacts
-        type: directory
-  in:
+  $if: 'tasks_for in ["github-push", "github-pull-request"]'
```
imbstack: I thought line 288 already did this basically? 

Also time for a decision task? :laughing:

djmitche: It does, but `allTasks` is already evaluated by that time, and fails because event properties aren't present in that case.

And, yes, please :)

PR #1057, opened 2019-07-18
Add titles for missing docs

### LINKED ISSUE #1056: worker-manager docs not appearing
There are several docs files in the repo under https://github.com/taskcluster/taskcluster/tree/master/ui/docs/reference/core/worker-manager but only "Worker-Manager - Worker Interaction" appears in the sidebar at

https://docs.taskcluster.net/docs/reference/core/worker-manager

Loading https://docs.taskcluster.net/docs/reference/core/worker-manager/google or https://docs.taskcluster.net/docs/reference/core/worker-manager/static shows the expected pages, and they appear in the docs-toc file:
https://github.com/taskcluster/taskcluster/blob/832f7498f5655b03250d3e6ea67f4484196a5f9a/generated/docs-table-of-contents.json#L2376-L2396

djmitche:
If this fixes #1056 then we should probably either fail when generating docs with no title, or default to title = name.

### DIFF
--- a/generated/docs-table-of-contents.json
+++ b/generated/docs-table-of-contents.json
@@ -2361,7 +2361,7 @@
                 "name": "logs",
                 "next": {
                   "path": "reference/core/worker-manager/google",
-                  "title": "google"
+                  "title": "Google Provider Type"
                 },
                 "path": "reference/core/worker-manager/logs",
                 "prev": {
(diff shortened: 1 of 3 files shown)


djmitche:
lol, of course I forgot yarn generate.  Will run and merge.

PR #1060, opened 2019-07-18
Fix up handling of titles in the TOC

djmitche:
This uncovered two more "missing" docs files.  With this change, they will at least appear in the TOC, if with an unexpected name.

### DIFF
--- a/ui/docs/reference/core/worker-manager/README.md
+++ b/ui/docs/reference/core/worker-manager/README.md
@@ -9,9 +9,8 @@ Each provider has a `providerType` indicating the class that implements the prov
 
 The service currently includes providers for:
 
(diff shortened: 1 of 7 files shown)


ui/docs/reference/core/worker-manager/README.md
```diff
@@ -9,9 +9,8 @@ Each provider has a `providerType` indicating the class that implements the prov
 
 The service currently includes providers for:
 
-* Static Workers (`static`)
-* Google Cloud (`google`)
-* Testing (`testing`, only used in the service's unit tests)
```
djmitche: I omitted testing since it's only for .. well, testing.

PR #1068, opened 2019-07-18
Created confirmation dialog before deleting hook

### LINKED ISSUE #957: [UI] Confirmation dialog when deleting a hook
Destructive actions (e.g., a delete operation) should have a confirmation step.
It’s an added click but will prevent accidental data loss frustration.

STR:

1. Navigate to http://localhost:5080/hooks/garbage/abc456
2. Hover/Click on the triple vertical dots icon then click on "Delete Hook"

Screen Shot 2019-07-02 at 3 59 57 PM

**Actual:** The client will try to delete right away

**Expected:** A confirmation dialog shows up using the custom [DialogAction](https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/DialogAction/index.jsx) component.

rbrishabh:
Fixes #957 
@helfi92 

Before: No confirmation dialog on deleting hook.

After:

is this ok

Confirmation dialog box comes first on deleting hook.

Let me know if this needs any changes!
Thank you! :)

### DIFF
--- a/ui/src/components/HookForm/index.jsx
+++ b/ui/src/components/HookForm/index.jsx
@@ -141,24 +141,12 @@ const initialHook = {
 }))
 /** A form to view/edit/create a hook */
 export default class HookForm extends Component {
-  static defaultProps = {
-    isNewHook: false,
-    hook: initialHook,
-    hookLastFires: null,
-    onTriggerHook: null,
-    onCreateHook: null,
-    onUpdateHook: null,
-    onDeleteHook: null,
-    actionLoading: false,
-    dialogError: null,
-  };
-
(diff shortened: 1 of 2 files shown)


rbrishabh:
I saw the changes you did as well.

I thought it was okay to use the same props for dialog close and dialog error of DialogOpen for DialogDeleteActionHook as well.

I understood that now. Thank you again. 

And you're welcome!

I would be around here to contribute more, so if I can be of any service, just tag me! :D

helfi92:
Nice. Are you interested in https://github.com/taskcluster/taskcluster/issues/1073? Comment on the bug directly if you are :)

PR #1082, opened 2019-07-19
Support a category for API methods

subhamskm:
Bugzilla Bug: [1551848](https://bugzilla.mozilla.org/show_bug.cgi?id=1551848)


djmitche:
How is this related to #1049?

djmitche:
Ah I missed the comment there.  It doesn't look like this is complete, though?

djmitche:
Oh, yes, it is complete.  I'll take a look when I'm back Monday if nobody beats me to it.

subhamskm:
Why are the checks like 'yarn generate' and 'taskcluster-lib-api' not running?

imbstack:
@subhamskm I am not sure why the statuses did not get updated (I will look into it now).

I did find the tasks than ran those checks though: https://taskcluster-ui.herokuapp.com/tasks/groups/TWkBgUH_RfmW1i_kmCAXtA

djmitche:
I get the feeling from the name of the branch and the commits that you're working on this without actually having a copy of the repository locally on your system.  That's not easy!  Is there something we could do to help you be able to run `yarn generate` locally?

subhamskm:
I have a copy locally. But I am new to git so I face issues while rebasing and ingoring some commits.

subhamskm:
> I get the feeling from the name of the branch and the commits that you're working on this without actually having a copy of the repository locally on your system. That's not easy! Is there something we could do to help you be able to run `yarn generate` locally?

When I run `yarn generate`  locally:

djmitche:
I'm always happy to help with git issues.  It can be confusing!

Regarding the error -- I see
image
which looks like it's involving a Windows path.  Taskcluster doesn't run or build directly on Windows, unfortunately.  I think it does work under WSL, and certainly works in a Linux VM.  This is the case because when Taskcluster is running in production, it's on Linux, so that's where we test it too.

I'll push a run of `yarn generate` to this PR and we can get it merged, I think.

subhamskm:
Thanks again!

djmitche:
Now the last phase is to show that information in the API references in the UI..

PR #1097, opened 2019-07-22
Possible fix for cookies not being sent in production

helfi92:
Logging in with the mozilla-auth strategy works in localhost but fails in production with error "Authentication is required to generate credentials". This error is coming from `loaders/auth.js`.

The reason why this is happening is because the cookies are not being sent from the browser and so `req.user` is undefined. I figured https://www.apollographql.com/docs/react/recipes/authentication/#cookie might be the fix 🤞

### DIFF
--- a/ui/src/App/index.jsx
+++ b/ui/src/App/index.jsx
@@ -69,6 +69,7 @@ export default class App extends Component {
 
   httpLink = new HttpLink({
     uri: process.env.GRAPHQL_ENDPOINT,
+    credentials: 'same-origin',
   });
 
   wsLink = new WebSocketLink({
--- a/services/web-server/src/servers/createApp.js
+++ b/services/web-server/src/servers/createApp.js
@@ -24,10 +24,10 @@ module.exports = async ({ cfg, strategies }) => {
     }
     return o;
   }).filter(o => o);
-    app.use(cors({
-      origin: allowedCORSOrigins,
-      credentials: true
-    }));
+  app.use(cors({
+    origin: allowedCORSOrigins,
+    credentials: true,
+  }));
 
   app.use(session({
     store: new MemoryStore({
--- a/services/web-server/src/servers/createApp.js
+++ b/services/web-server/src/servers/createApp.js
@@ -24,7 +24,10 @@ module.exports = async ({ cfg, strategies }) => {
     }
     return o;
   }).filter(o => o);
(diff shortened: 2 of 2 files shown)


helfi92:
Ugh, this doesn't seem to have fixed it. STR:

1. Navigate to https://taskcluster-ui.herokuapp.com/
2. Open the developer console and type `window.open('/login/mozilla-auth0')`

ui/src/App/index.jsx
```diff
@@ -69,6 +69,7 @@ export default class App extends Component {
 
   httpLink = new HttpLink({
     uri: process.env.GRAPHQL_ENDPOINT,
+    credentials: 'same-origin',
```
helfi92: From https://www.apollographql.com/docs/react/recipes/authentication/#cookie:

>  You just need to pass the credentials option. e.g. credentials: 'same-origin' as shown below, if your backend server is the same domain or else credentials: 'include' if your backend is a different domain.

Because we're using a proxy to talk to web-server, I figured we can use use `same-origin`. @djmitche Does that make sense to you? or do you think it should be set to `include`.

djmitche: Ugh, for the herokuapp domain it will need to be `include`, since it's the one case where it is *not* the same origin.  Maybe just add a special case here for `rootUrl == 'https://taskcluster.net'`?

arku: @djmitche Just to clarify, are you suggesting we check the domains of the current URL and `rootUrl`?

Something like `window.location.hostname === new URL(process.env.TASKCLUSTER_ROOT_URL).hostname ? 'same-origin' : 'include'` should work, I think.

helfi92: I think you can probably just do `process.env.TASKCLUSTER_ROOT_URL === 'https://taskcluster.net' ? 'include' : 'same-origin'`.

arku: @helfi92 Interesting. What other `TASKCLUSTER_ROOT_URL` do we use in the builds?

helfi92: The site can be deployed by anyone using their own configuration variables. Examples:
* https://taskcluster.imbstack.com/
* https://docs.taskcluster.net/docs
* https://taskcluster-ui.herokuapp.com/

arku: Gotcha.

PR #1100, opened 2019-07-23
Encoding artifact url correctly

### LINKED ISSUE #1079: Artifacts with URL encoded filenames can't be downloaded
My Task Cluster task most naturally produces artifacts identified by URLs, so these URLs are encoded in the filename (well, in the directory part of the filename).  These appear to be uploaded correctly but they cannot be downloaded.  See, for example, https://tools.taskcluster.net/groups/Lp77UnjWSo65lTmcgyA1Pg/tasks/G7fVftFGTg6BMkjg0geGXQ/runs/0/artifacts which has an artifact named like `public/test_info/browsertime-results/https%3A%2F%2Fwww.facebook.com/browsertime.json` that links to https://queue.taskcluster.net/v1/task/G7fVftFGTg6BMkjg0geGXQ/runs/0/artifacts/public/test_info/browsertime-results/https%3A%2F%2Fwww.facebook.com/browsertime.json.  That link results in a 404.  If I manually URL encode further, like https://queue.taskcluster.net/v1/task/G7fVftFGTg6BMkjg0geGXQ/runs/0/artifacts/public/test_info%2Fbrowsertime-results%2Fhttps%253A%252F%252Fwww.facebook.com%2Fbrowsertime.json, I get an access denied message instead.

rbrishabh:
Fixes #1079 

Always encoding artifact url but leaving / un-escaped.

Is this what was required?

### DIFF
--- a/services/web-server/src/loaders/artifacts.js
+++ b/services/web-server/src/loaders/artifacts.js
@@ -29,12 +29,12 @@ module.exports = ({ queue }, isAuthed, rootUrl) => {
           ? urls.api(
             'queue',
             'v1',
-            `task/${taskId}/runs/${runId}/artifacts/${encodeURIComponent(artifact.name)}`
+            `task/${taskId}/runs/${runId}/artifacts/${encodeURI(artifact.name)}`
           )
           : urls.api(
             'queue',
             'v1',
-            `task/${taskId}/artifacts/${encodeURIComponent(artifact.name)}`
+            `task/${taskId}/artifacts/${encodeURI(artifact.name)}`
           ),
       };
     }
--- a/services/web-server/src/loaders/artifacts.js
+++ b/services/web-server/src/loaders/artifacts.js
@@ -29,12 +29,12 @@ module.exports = ({ queue }, isAuthed, rootUrl) => {
           ? urls.api(
             'queue',
             'v1',
-            `task/${taskId}/runs/${runId}/artifacts/${artifact.name}`
+            `task/${taskId}/runs/${runId}/artifacts/${encodeURIComponent(artifact.name)}`
           )
           : urls.api(
             'queue',
             'v1',
-            `task/${taskId}/artifacts/${artifact.name}`
+            `task/${taskId}/artifacts/${encodeURIComponent(artifact.name)}`
           ),
       };
     }


djmitche:
I believe that will over-encode the URL -- the URL will have the form `https://something/api/queue/v1/task/../run/../artifact/../../..` and this will change all of those `/` to `%2F`.  In fact, we want *none* of those `/` to turn to `%2F`.

There are a number of merge commits in the PR here -- could you perhaps squash it into one commit?  And then see why the tests are failing in GitHub.  Once those are fixed, we'll have a deploy preview that we can click on to test it out.

rbrishabh:
I tried to squash all commits. This was my first attempt on squashing. This is what I ended up with. 

I think I messed it up a bit. 

Please let me know if this is correct or not! Thanks!

imbstack:
I don't think you squashed entirely correctly here. Not a problem, we can help. If you do the following on your branch, it should work:

```
git fetch origin
git rebase -i origin/master
```

It will show you a list of commits. Keep the first one as `pick` and change the rest to `squash`. Continue and fix whatever merge conflicts it shows along the way.

You should end up with a single commit containing only the changes you've made here.
Let me know how else we can help!

rbrishabh:
Thank you @imbstack ! That helped me alot!

@djmitche ,  I am using the method ```encodeURI()``` which does not encode any of the ```/```.
```encodeURIComponent()``` encodes the ```/```, ```encodeURI()``` doesn't.

 I still don't understand why you think the url is being over-encoded. Can you explain that once?

djmitche:
I'm just back from vacation.  Now that the commit history is sorted out, I'll have a look.

rbrishabh:
Hahaha, alright! I hope you had a wonderful time! :D

rbrishabh:
> 
> 
> Ah, I found where `url` is created:
> 
> https://github.com/taskcluster/taskcluster/blob/6aa4bd623f222b2d65830601314974eab4d93ec8/services/web-server/src/loaders/artifacts.js#L15-L40
> 
Okay got it! I made some changes here.

> I think _that's_ where we should get the encoding right. URL encoding is a particularly trick thing to do correctly: it can work correctly for all the examples you can think of, but then there's one example you didn't think of. This is mostly because it leaves most characters unchanged, so it's easy to decode one time too many, or encode one time too few, and still have things look OK.
> 
Yep, I understand now that this is definitely tricky.

> In this case, here's a counterexample: imagine you get a `url` from the server that has a bewit in it, and the bewit is already properly urlencoded:
> 
> ```
> >>> encodeURI('https://queue.taskcluster.net/v1/task/foo/run/1/artifact/private/test_info/browsertime-results/https%3A%2F%2Fwww.facebook.com/browsertime.json?bewit=lkajdsf%7Clskajflsad%7Clkajfd')
> <- "https://queue.taskcluster.net/v1/task/foo/run/1/artifact/private/test_info/browsertime-results/https%253A%252F%252Fwww.facebook.com/browsertime.json?bewit=lkajdsf%257Clskajflsad%257Clkajfd"
> ```
> 
> so we've now _over-encoded_ the contents of the bewit -- when the server tries to decode that, it will get `lkajdsf%7Clskajflsad%7Clkajfd` but is expecting `lkajdsf|lskajflsad|lkajfd`.

Thanks for the example. This made it quite clear :)

I pushed another patch in an attempt to solve this issue.
I noticed in [this](https://github.com/taskcluster/taskcluster/issues/1079#issue-470511093) comment that only the artifact name needs to be encoded, so that's what I tried to do. 

To my understanding, the ```/```s in the artifact name need also be encoded.

Tell me if this is heading in the correct direction

Thanks :)

rbrishabh:
> So I think we want to use `encodeURI`, instead.
> 

I love the way you always make sure to explain things such clearly! A great learning experience for me always!

> It's funny how sometimes a lot of discussion comes down to just a few lines! Sometimes software engineering is like that. The hard part was figuring out the solution, and I appreciate your hard work to do so.
> 

Definitely! It's funny but figuring out the solution is where all the fun lies! I appreciate your hardwork to explain things such clearly as well! =)

> So, if you can make that change then I'll get this merged.

I made the changes! Thanks :)

PR #1101, opened 2019-07-23
Fix go mod in release for client-shell

imbstack:
@milescrabill asking for your review for your golang brain.

this is the bit of the services stuff that updates our go clients when we release a new version. The shell library relies on the go-client library. Before this change, trying to cut a release yelled about not finding the version of the go-client library that we wanted. I believe it was because this was updating

```
github.com/taskcluster/taskcluster/clients/client-go/v14 v14.0.0-20190713182157-1b8e18d76d60
```
to 

```
github.com/taskcluster/taskcluster/clients/client-go/v15 v14.0.0-20190713182157-1b8e18d76d60
```

which obviously doesn't make any sense. Now it updates it to 

```
github.com/taskcluster/taskcluster/clients/client-go/v15 v15.0.0
```
Does that make sense to you?

cc @djmitche so you can tell if this was actually correct for the way we cut releases when you're back

### DIFF
--- a/infrastructure/builder/src/release/tasks.js
+++ b/infrastructure/builder/src/release/tasks.js
@@ -141,11 +141,17 @@ module.exports = ({tasks, cmdOptions, baseDir}) => {
         contents.replace(/download\/v[0-9.]*\/taskcluster-/, `download/v${requirements['release-version']}/taskcluster-"`));
       changed.push(shellreadme);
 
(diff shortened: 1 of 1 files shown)


milescrabill:
Yeah, that's what I was thinking as well. I meant to retro-R+ this after @imbstack and I worked through it.

PR #1108, opened 2019-07-25
Use kubectl to interact with dev clusters

imbstack:
This should allow redeploying the cluster without uninstalling first!

It should also be faster and more verbose about what it is doing.

### DIFF
--- a/infrastructure/builder/src/dev/helm.js
+++ b/infrastructure/builder/src/dev/helm.js
@@ -1,56 +1,148 @@
-const child_process = require('child_process');
-const {REPO_ROOT, readRepoYAML} = require('../utils');
+const {REPO_ROOT, readRepoYAML, execCommand} = require('../utils');
+const {TaskGraph} = require('console-taskgraph');
+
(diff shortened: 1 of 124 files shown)


infrastructure/builder/src/dev/helm.js
```diff
@@ -1,56 +1,148 @@
...
+      }
+    },
+  },
+  {
+    title: 'Generate k8s Resources',
+    requires: ['dev-config', 'helm-version'],
+    provides: ['target-templates'],
+    run: async (requirements, utils) => {
+      let command = ['helm', 'template'];
+      if (requirements['helm-version'] === 3) {
```
edunham: does this allow configuring to pick a helm version? if so, could you drop a note about it in the docs somewhere?

imbstack: Doesn't allow configuring, it just works with whatever helm you have installed (2 or 3, who knows about 1). I'll mention in the docs though how this works

PR #1109, opened 2019-07-25
Fix Taskcluster shell linux binary download link

arku:
Removes the stray `"` from the download link.

### DIFF
--- a/infrastructure/builder/src/release/tasks.js
+++ b/infrastructure/builder/src/release/tasks.js
@@ -138,7 +138,7 @@ module.exports = ({tasks, cmdOptions, baseDir}) => {
       const shellreadme = 'clients/client-shell/README.md';
       utils.status({message: `Update ${shellreadme}`});
       await modifyRepoFile(shellreadme, contents =>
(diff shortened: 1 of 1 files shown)


arku:
@djmitche Just curious, is there a way to skip Taskcluster builds with a special commit message (that starts with `[skip ci]` or something like that)?

imbstack:
@arku This actually won't fix this. This line is generated in [infrastructure/builder/src/release/tasks.js](https://github.com/taskcluster/taskcluster/blob/master/infrastructure/builder/src/release/tasks.js#L141)

There's no way to skip building but we like it that way! It's not a waste of resources or anything.

arku:
@imbstack Ah, my bad.

Let me fix it.

arku:
@imbstack I am having a hard time testing my fix (running into `spawn ENOENT` errors). How do I go about it?

imbstack:
@arku that's ok. `yarn release` is really hard to test. This looks correct to me! I'll land it as soon as tests complete.

Thanks for the fix.

arku:
@imbstack Perfect, thanks!

PR #1110, opened 2019-07-26
Fix credentials option in HttpLink

### LINKED ISSUE #1099: [web-server] Cookies not working properly in production
STR:

1. Navigate to https://taskcluster-ui.herokuapp.com/
2. Open the developer console and type window.open('/login/mozilla-auth0')
3. Login

Expected: The user is logged in
Actual: The user is not logged in in production. Works in localhost.

Problem: Cookies are not being set for some reason. https://github.com/taskcluster/taskcluster/pull/1097 was an attempt to fix the issue but that didn't fix it as far as I can tell.

arku:
Relates to #1099

Credits to @djmitche for identifying the issue and the fix.

### DIFF
--- a/ui/src/App/index.jsx
+++ b/ui/src/App/index.jsx
@@ -69,7 +69,10 @@ export default class App extends Component {
 
   httpLink = new HttpLink({
     uri: process.env.GRAPHQL_ENDPOINT,
-    credentials: 'same-origin',
+    credentials:
+      process.env.TASKCLUSTER_ROOT_URL === 'https://taskcluster.net'
+        ? 'include'
+        : 'same-origin',
   });
 
   wsLink = new WebSocketLink({


helfi92:
Merging this made the graphql server stop returning responses even if the user is logged in. Not sure what happened.

arku:
@helfi92 Any idea how to debug this?

helfi92:
That's a really good question. @djmitche What would you recommend the best way for @arku to test things in a "production" environment. I can think of having having them create two heroku apps (ui and web-server) with the right config or set up a deployment via kubernetes which may involve a bunch of configurations for the services. Thoughts?

djmitche:
Both of those options will be pretty difficult without a lot of access to secret values.  It might be easiest to debug this "live".

I assume this stopped working on https://taskcluster-ui.herokuapp.com which talks to https://web-server.taskcluster.net as its backend and has TASKCLUSTER_ROOT_URL = `https://taskcluster.net`, so in effect it translates to setting `credentials: 'include'`.  The docs say that [gets passed directly to `fetch`](https://github.com/github/fetch#sending-cookies) which [treats `include` as meaning "always includes credentials with this request"](https://fetch.spec.whatwg.org/#concept-request-credentials-mode).  With this deployed, what do you see in the network inspector?  That it stops returning responses entirely suggests more than a missing `Cookie` header.

I don't know much about fetch.  I wonder if it cannot control `httpOnly` cookies?  Or if this is interacting with CORS somehow?

PR #1113, opened 2019-07-26
Bug 1552970: Do not set IAM policy before generating a temp. token

walac:
Bugzilla Bug: [1552970](https://bugzilla.mozilla.org/show_bug.cgi?id=1552970)


walac:
This is the second part of the RRA feedback.

PR #1128, opened 2019-07-30
Refactor configuration of GCP to support multiple projects

djmitche:
We're already introducing a major change to the configuration of this
endpoint, so let's make that configuration look the way we want it to
look long-term: allowing specification of multiple projects, each with
their own list of allowed service accounts.

I stopped short of actually *implementing* that support, but it's a
small refactor away.

This also includes some documentation about setting up for testing and
in deployments.


djmitche:
(btw I'd *really* like to land this before the next release)

PR #1133, opened 2019-07-31
require --force to rerun a non-failed, non-exception task

### LINKED ISSUE #1062: [client-shell] Require `--force` to rerun successful or scheduled tasks
From #188 

@escapewindow
For rerun, we specifically want to check for status in ['failed', 'exception'] and require --force to rerun completed tasks.

---

@srfraser  
I wonder whether there is also value in checking a task's dependencies for some operations. For example, if Task C depends on tasks A and B, and we want to rerun C, that doesn't really make sense if A and B didn't complete.

escapewindow:
This bit us again in Fx 69.0b9 this week. This appears to work in my limited testing:

- I can't rerun a green task without `--force`.
- I can rerun the green task with `--force`.
- I can rerun a red task without `--force`.

Fixes #1062.

### DIFF
--- a/changelog/issue-1062.md
+++ b/changelog/issue-1062.md
@@ -1,0 +1,4 @@
+level: minor
+reference: issue #1062
+---
+The taskcluster cli `rerun` action now takes a `--force` option. It will refuse to rerun non-exception, non-failed tasks without `--force`.
--- a/clients/client-shell/cmds/task/actors.go
+++ b/clients/client-shell/cmds/task/actors.go
@@ -46,6 +46,7 @@ func runCancel(credentials *tcclient.Credentials, args []string, out io.Writer, 
 func runRerun(credentials *tcclient.Credentials, args []string, out io.Writer, flagSet *pflag.FlagSet) error {
 	noop, _ := flagSet.GetBool("noop")
 	confirm, _ := flagSet.GetBool("confirm")
+	force, _ := flagSet.GetBool("force")
 
 	q := makeQueue(credentials)
 	taskID := args[0]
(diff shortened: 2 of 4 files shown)


escapewindow:
Looking at the test.

escapewindow:
In theory this should fix it. I'd love to have something like `pytest.mark.parametrize` where I can test the various task states.

escapewindow:
Not sure what's broken at this point.

escapewindow:
For posterity: Once I cloned the test deps in `$GOPATH/src/github.com/`, I was able to test locally with `NO_TEST_SKIP=true go test -v -race ./...` in `clients/client-shell`.

changelog/issue-1062.md
```diff
@@ -0,0 +1,4 @@
+level: minor
+reference: issue #1062
+---
+The taskcluster cli `rerun` action now takes a `--force` option. It will refuse to rerun non-exception, non-failed tasks without `--force`.
```
djmitche: ++!

djmitche:
@escapewindow you shouldn't need to clone any test deps if you're using go1.12.6 -- `go test ./...` should be all it takes, as that will load the deps in `go.mod` for you.

changelog/issue-1062.md
djmitche: Oh, in fact the `#` here is treated as a YAML comment -- that should be `issue 1062`.  I'll fix in master.

escapewindow:
> @escapewindow you shouldn't need to clone any test deps if you're using go1.12.6 -- `go test ./...` should be all it takes, as that will load the deps in `go.mod` for you.

This fails for me on go1.12.7; is that expected? I can try downgrading if needed, or workaround as I have been.

```
akimoz: /src/tc/tc-monorepo/clients/client-shell (rerun-force) [10:21:48]
10528$ go test ./...
# github.com/taskcluster/taskcluster/clients/client-shell/apis
/src/go/src/github.com/taskcluster/taskcluster/clients/client-shell/apis/provider_test.go:11:2: cannot find package "github.com/stretchr/testify/require" in any of:
<snip>
```

djmitche:
I think the difference there isn't the `.6` vs `.7` but building the package outside of $GOPATH.  `gvm use` and `gvm install` both set the GOPATH to some `.foo` directory under your homedir.  The `go` command only uses `go.mod` when it's not under GOPATH.  At least, that's my remedial understanding of the situation :)

Anyway, we pin exact dot versions of Go, so go1.12.6 is a good choice, and I'm glad it's working :)

PR #1136, opened 2019-07-31
Fix changelog references and run 'yarn generate' to catch this in PRs

djmitche:


infrastructure/builder/src/changelog/index.js
```diff
@@ -60,14 +60,15 @@ class ChangeLog {
       }
 
       if (reference) {
+        console.log(reference);
```
imbstack: Stray log

PR #1139, opened 2019-07-31
Set UI_LOGIN_STRATEGY_NAMES

djmitche:


djmitche:
OK, that won't work, but here's a reminder to fix this :)

djmitche:
This is bogus like three different ways :)

PR #1145, opened 2019-08-02
Add provider listing endpoint to worker-manager

imbstack:


services/worker-manager/src/api.js
```diff
@@ -26,6 +26,25 @@ let builder = new APIBuilder({
...
+  method: 'get',
+  route: '/providers',
+  name: 'listProviders',
+  input: undefined,
+  output: 'provider-list.yml',
+  stability: 'stable',
+  category: 'Worker Manager',
+  title: 'List Providers',
+  description: [
+    'Retrieve a list of providers that are available for worker pools.',
```
djmitche: Are we later going to regret not paginating this method?

imbstack: Do we think we'll have more than 1000 providers?

imbstack: Actually.. I think you're right so that we keep with the convention of having a continuation token at least. I'll add it.

djmitche: I can see a case where *another* user has 1000 providers (e.g., a company that has some process for creating a provider per customer).

PR #1147, opened 2019-08-03
AWS provider, pass 3

owlishDeveloper:
My plan at this point would be to deploy this code to my dev cluster and see if it works. 75% review please!

I implemented all the feedback from the previous pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "hashids": "^1.2.2",
     "hawk": "^7.0.10",
     "hsts": "^2.1.0",
-    "iid-verify": "^1.1.0",
     "immutable": "^3.8.2",
     "irc-upd": "^0.10.0",
     "iterall": "^1.2.2",
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "hashids": "^1.2.2",
     "hawk": "^7.0.10",
     "hsts": "^2.1.0",
+    "iid-verify": "^1.1.0",
     "immutable": "^3.8.2",
     "irc-upd": "^0.10.0",
     "iterall": "^1.2.2",
--- a/services/worker-manager/schemas/v1/config-aws.yml
+++ b/services/worker-manager/schemas/v1/config-aws.yml
@@ -1,0 +1,34 @@
+$schema: "/schemas/common/metaschema.json#"
+title: AWS Provider Config
+description: |
+  Defines the shape of the config section of a workerpool if the workerPool.providerId is an AWS provider.
+  Much of these fields are documented [here](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#runInstances-property)
+type: object
+properties:
+  minCapacity:
+    title: Minimum Capacity
+    type: integer
+    minimum: 0
+    description: |
+      The minimum amount of capacity that worker-manager should make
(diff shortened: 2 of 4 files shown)


services/worker-manager/schemas/v1/config-aws.yml
```diff
@@ -0,0 +1,16 @@
...
+uniqueItems: true
+items:
+  type: object
+  description: |
+    An object to be passed in to AWS runInstances method.
+    See the documentation [here](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#runInstances-property)
+    You can configure any properties you need. Note, however, that we disallow passing in arbitrary IAM instance profiles.
+  properties:
+    not:
+      IamInstanceProfile:
```
djmitche: It'd be good to have a test that this value is rejected.  I've not seen "not" used that way before.

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,398 @@
...
+          owner: spawned.OwnerId,
+        },
+      });
+    }));
+  }
+
+  async registerWorker({worker, workerPool, workerIdentityProof}) {
+    const {document, pkcs7} = workerIdentityProof;
+    if (!document || !pkcs7) {
+      throw new ApiError('Token validation error');
```
djmitche: This could probably be "instance Identity Document validation error"

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,398 @@
...
+      validator,
+      notify,
+    });
+    this.configSchema = 'config-aws';
+    this.instancePermissions = instancePermissions;
+    this.apiVersion = apiVersion;
+    this.region = region;
+    this.ec2iid_DSA_SHA1_cert = ec2iid_DSA_SHA1_cert;
+
+    aws.config.update({region});
```
imbstack: Different workerpools exist in different regions so I think we'll have to do this per-request ultimately?

package.json
```diff
@@ -54,6 +54,7 @@
     "hashids": "^1.2.2",
     "hawk": "^7.0.10",
     "hsts": "^2.1.0",
+    "iid-verify": "^1.1.0",
```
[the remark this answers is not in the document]
owlishDeveloper: That's a good question. Currently, I am not able to make it work even on master, I must confess

services/worker-manager/src/providers/aws.js
owlishDeveloper: I was trying to be vague on purpose, but if you think the error message you suggest is better, I can sure add it.

services/worker-manager/src/providers/aws.js
owlishDeveloper: oh I forgot all about this, with all the iid stuff last week, thank you for reminding!

services/worker-manager/src/providers/aws.js
djmitche: The vagueness is definitely good :)

package.json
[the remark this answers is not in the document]
owlishDeveloper: Ok, so after solving the docker trouble, the answer to this is "yes". I was able to successfully build an image from this branch

PR #1162, opened 2019-08-07
Bug 1553953 - limit chars in provisionerId and workerType

djmitche:
This brings these symbols into correspondence with the limits on
workerPoolId, which are designed to be compatible with various cloud
providers.

Bugzilla Bug: [1553953](https://bugzilla.mozilla.org/show_bug.cgi?id=1553953)


djmitche:
I'll wait until the docker-worker fix lands.

PR #1163, opened 2019-08-07
Point from worker-manager to existing workertypes view

djmitche:
Per discussion with @owlishDeveloper and @helfi92

### DIFF
--- a/ui/src/components/Dashboard/SidebarList.jsx
+++ b/ui/src/components/Dashboard/SidebarList.jsx
@@ -69,11 +69,9 @@ export default class SidebarList extends Component {
           AWS Provisioner
         </SidebarListItem>
 
-        {process.env.SHOW_WORKER_MANAGER && (
-          <SidebarListItem to="/worker-manager" icon={<HexagonSlice4 />}>
-            Worker Manager
-          </SidebarListItem>
-        )}
+        <SidebarListItem to="/worker-manager" icon={<HexagonSlice4 />}>
+          Worker Manager
+        </SidebarListItem>
 
         <SidebarListItem to="/provisioners" icon={<MixcloudIcon />}>
           Provisioners
--- a/ui/src/components/WMWorkerPoolsTable/index.jsx
+++ b/ui/src/components/WMWorkerPoolsTable/index.jsx
@@ -23,6 +23,7 @@ import TableCellItem from '../TableCellItem';
 import ErrorPanel from '../ErrorPanel';
 import formatError from '../../utils/formatError';
 import { NULL_PROVIDER } from '../../utils/constants';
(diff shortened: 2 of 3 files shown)


ui/src/components/Dashboard/SidebarList.jsx
```diff
@@ -79,6 +79,10 @@ export default class SidebarList extends Component {
           Provisioners
         </SidebarListItem>
 
+        <SidebarListItem to="/worker-manager" icon={<MixcloudIcon />}>
+          Worker Manager
+        </SidebarListItem>
+
```
helfi92: We would probably just need to remove `process.env.SHOW_WORKER_MANAGER`

https://github.com/taskcluster/taskcluster/blob/d6fddea345e02aa6075dc8cc80d89ddf47c949ce/ui/src/components/Dashboard/SidebarList.jsx#L72-L76

helfi92: We will also need to remove `SHOW_WORKER_MANAGER` from `.neutrinorc.js`.

djmitche: Oh, I didn't know that was a thing!

owlishDeveloper: why to remove that variable? I don't want worker manager to be available in, well, not production, but the current public beta of the UI

ui/src/components/WMWorkerPoolsTable/index.jsx
```diff
@@ -186,9 +190,9 @@ export default class WorkerManagerWorkerPoolsTable extends Component {
                 className={classes.viewWorkersButton}
                 variant="outlined"
                 component={Link}
-                to={`${this.props.match.path}/${encodeURIComponent(
```
owlishDeveloper: `${this.props.match.path}` was used to not hardcode paths

ui/src/components/Dashboard/SidebarList.jsx
djmitche: We're going to have people using it to set up workers in GCP, so it does need to be visible.

ui/src/components/WMWorkerPoolsTable/index.jsx
djmitche: The new path goes all the way back to the root, so nothing of the matched route is preserved.  Is there a better way to generate the path for another view?

owlishDeveloper: So the `/provisioners/...` link leads to a different tool? The easiest thing to do is to hardcode then, in the absence of url constants

djmitche: It does.

PR #1164, opened 2019-08-08
Improve docs for auth.listRoles/auth.listRoles2

djmitche:
This helps the reader (which was me, yesterday) understand why there are
two methods and which one to use.


djmitche:
It was one of those endpoints that just returned an array of data, so we couldn't add a `continuationToken` to the response body.

PR #1168, opened 2019-08-09
Bug 1572764: Do not automatically log HTTP transactions in go client

walac:
By default, only logs method, hostname, port, and response body.
It avoids leaks of secrets contained in the requests.

If the environment variable `TASKCLUSTER_DEBUG` is defined, then the
full request is logged.

Bugzilla Bug: [1572764](https://bugzilla.mozilla.org/show_bug.cgi?id=1572764)

### DIFF
--- a/changelog/bug1572764.md
+++ b/changelog/bug1572764.md
@@ -1,0 +1,8 @@
+level: minor
+reference: bug 1572764
+---
+
+The go client doesn't log the full request in case of an error anymore.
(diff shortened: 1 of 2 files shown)


walac:
> Even if so, why default to logging responses? It seems dangerous in general.
It is only logged when an error happens, in which case the response body contains error details.

djmitche:
Yup, clients can't download artifacts

PR #1177, opened 2019-08-12
[Bug 1572775] Document what needs to be monitored

imbstack:
This automatically generates a table of background processes that should be monitored. We can probably make this document more helpful in the future as well!

cc @edunham and @sciurus 

Bugzilla Bug: [1572775](https://bugzilla.mozilla.org/show_bug.cgi?id=1572775)


dev-docs/monitoring-suggestions.md
```diff
@@ -0,0 +1,46 @@
...
+  "Fields": {
+    "name": "<Name>"
+  }
+}
+```
+
+They will also have Fields for `status`, `duration`, and a serialized `error` if an error occured.
+
+The processes that have `continuous` for their dedaline and schedule run every few minutes and should complete fairly quickly. The rest
+have their schedules and maximum allowed duration defined here.
```
djmitche: It's probably worth including a markdown comment indicating that this file is automatically generated and should not be modified.

libraries/iterate/src/index.js
```diff
@@ -133,19 +139,15 @@ class Iterate extends events.EventEmitter {
...
+        status: iterError ? 'exception': 'success',
+        error: iterError ? serializeError(iterError) : undefined,
+      }, {level: iterError ? 'err' : 'notice'});
 
       if (iterError) {
-        if (this.monitor) {
-          this.monitor.reportError(iterError, 'warning', {
-            consecutiveErrors: failures.length,
-          });
-        }
```
djmitche: What's the rationale for not reporting these errors as errors anymore?  Won't that mean they don't appear in error dashboards like sentry?

libraries/monitor/src/monitor.js
```diff
@@ -181,18 +181,24 @@ class Monitor {
...
     try {
-      try {
-        assert.equal(typeof name, 'string');
-        assert.equal(typeof fn, 'function');
+      assert.equal(typeof name, 'string');
+      assert.equal(typeof fn, 'function');
 
-        await this.timer(name, fn);
-      } catch (err) {
-        this.reportError(err);
```
djmitche: Same question here: why not report these errors?

dev-docs/monitoring-suggestions.md
```diff
@@ -0,0 +1,46 @@
...
+
+| Service        | Name                   | Logger                     | Deadline (seconds) | Schedule    |
+| -------------- | ---------------------- | -------------------------- | ------------------ | ----------- |
+| auth           | purgeExpiredClients    | taskcluster.auth           | 86400              | At 12:00 AM |
```
djmitche: Should the schedule say "UTC"?

infrastructure/builder/src/generate/generators/monitoring.js
```diff
@@ -0,0 +1,69 @@
...
+      `procs-${name}`,
+    ],
+    provides: [`monitoring-hints-${name}`],
+    run: async (requirements, utils) => {
+      const res = [];
+      Object.entries(requirements[`procs-${name}`]).forEach(([proc, ext]) => {
+        if (ext.type === 'cron') {
+          res.push([name, proc, `taskcluster.${name}`, ext.deadline, cronstrue.toString(ext.schedule)]);
+        }
+        if (ext.type === 'background' && ext.subType === 'iterate') {
```
djmitche: What about handler processes?

dev-docs/monitoring-suggestions.md
imbstack: I guess it depends on what k8s thinks about timezones and cron. I'll look that up shortly.

infrastructure/builder/src/generate/generators/monitoring.js
imbstack: How would one monitor that? They could potentially never have an event fire and it would be fine.

libraries/iterate/src/index.js
imbstack: Ah, I had something in my head about wanting every "event" in tc to end up with a single log message but we could duplicate it into an error event too. I think actually this would work with both because it sets the log level to error. Does that make sense or should we duplicate?

infrastructure/builder/src/generate/generators/monitoring.js
djmitche: Good point.  We do automatically reconnect, so there's potential to say something here, but let's leave it for another PR in the indefinite future :)

libraries/iterate/src/index.js
djmitche: I think we should have two events.  Stackdriver would (maybe, who knows) find the error in the output, but maybe something else wouldn't.  For example, I was thinking of connecting to Sentry directly from the `monitor.reportError` method, rather than pulling errors out of the log stream and calling the Sentry API with them.

One option here is to not include the serialized error in the iterate/oneShot output, since it's just intended for monitoring -- that reduces the conceptual duplication a bit.  And it means that an ops person who starts seeing failures has a time-range to go start searching for errors.

imbstack: Could we hook in direct sentry integration at the logging level rather than monitor and have anything with > err be sent to sentry or something? I'm not firmly in either camp yet. Just trying to figure what's best.

Is there a way you'd prefer this to be, @sciurus / @edunham

djmitche: Possibly -- I just don't want to tie us too closely into the stackdriver model.

sciurus: I don't think as written this would end up in stackdriver error reporting, since it doesn't conform to the format at https://cloud.google.com/error-reporting/docs/formatting-error-messages

I do think you want to make sure the serialized error makes it to sentry or stackdriver error reporting, so that you can have grouping.

I think Emily and I are agnostic on how this is implemented internally, and the only thing we know we don't want is to have to run a separate process to tail the logs, extract errors, and submit them somewhere.

dev-docs/monitoring-suggestions.md
imbstack: Everything I've read so far says it is just based on the timezone of the master.

djmitche: Maybe a sentence to that effect?

libraries/loader/src/index.js
```diff
@@ -146,7 +146,7 @@ function loader(componentDirectory, virtualComponents = {}) {
           }
           return new Promise((resolve, reject) => {
             try {
-              resolve(def.setup.call(null, ctx));
+              resolve(def.setup.call(null, ctx, target));
```
djmitche: It'd be great if this was mentioned in the lib-loader README (sorry I didn't say that last review)

PR #1178, opened 2019-08-13
Added confirmation dialog box for denylist page

shubhamgupta2956:
There is still another problem in denylist page
Screenshot from 2019-08-13 20-40-10

### DIFF
--- a/ui/src/components/DenylistForm/index.jsx
+++ b/ui/src/components/DenylistForm/index.jsx
@@ -1,13 +1,15 @@
 import React, { Component, Fragment } from 'react';
-import { bool, func } from 'prop-types';
+import { oneOfType, object, string, func, bool } from 'prop-types';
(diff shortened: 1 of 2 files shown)


ui/src/components/DenylistForm/index.jsx
```diff
@@ -219,11 +259,26 @@ export default class DenylistForm extends Component {
...
+          <DialogAction
+            open={dialogOpen}
+            onSubmit={this.handleAddressDelete}
+            onComplete={onDialogActionComplete}
+            onClose={onDialogActionClose}
+            onError={onDialogActionError}
+            error={dialogError}
+            title="Delete Address?"
+            body={
+              <Typography>This will delete {address.notificationAddress}.</Typography>
```
helfi92: The issue you were having is because `{address}` was an object but React can only render strings.

PR #1180, opened 2019-08-14
[WIP] Add indicator of optional to dev config example

imbstack:
This makes it a bit more clear what is needed or not. All data on what is optional comes from my guesses a few weeks back so I might be wrong here or there.


infrastructure/builder/src/utils/config.js
```diff
@@ -48,14 +48,15 @@ exports.configToSchema = type => {
...
-    case '!env:string': return '...';
-    case '!env:number': return '...';
-    case '!env:bool': return 'true/false';
-    case '!env:json': return {};
-    case '!env:list': return [];
+    case '!env': return `${expl}...`;
+    case '!env:string': return `${expl}...`;
+    case '!env:number': return `${expl}...`;
+    case '!env:bool': return `${expl}true/false`;
+    case '!env:json': return `${expl}{}`;
```
djmitche: this ends up not being valid json.. (`foo: '(optional) {}'`)

imbstack: Yeah, I've given up on it being valid and just being strings that say what should be there.

imbstack: Oh wait shoot, this gets pasted into dev-configs too. ugh

PR #1183, opened 2019-08-14
Highlight filter icon when it's active

### LINKED ISSUE #1181: Log search filter background color
We should try to give a background color to the filter icon when selected.

STR:

1. Navigate to https://taskcluster-ui.herokuapp.com/tasks/Kp5viqUERuKejEkgz8u4dA/runs/0/logs/live/https%3A%2F%2Fqueue.taskcluster.net%2Fv1%2Ftask%2FKp5viqUERuKejEkgz8u4dA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log

Screen Shot 2019-08-14 at 2 19 57 PM

2. Click on the filter button (4 lines icon)

Actual:

Screen Shot 2019-08-14 at 2 20 03 PM

Expected: The filter should have a background similar to the first button in the following screenshot:

Screen Shot 2019-08-14 at 2 22 03 PM

arku:
Fixes #1181 

Screenshots

image

image

### DIFF
--- a/ui/src/components/Log/index.jsx
+++ b/ui/src/components/Log/index.jsx
@@ -14,94 +14,118 @@ import Loading from './Loading';
 import Button from '../Button';
 import SpeedDial from '../SpeedDial';
 import SpeedDialAction from '../SpeedDialAction';
+import { THEME } from '../../utils/constants';
 
 const LINE_NUMBER_MATCH = /L(\d+)-?(\d+)?/;
 const FOLLOW_STORAGE_KEY = 'follow-log';
 
 @withRouter
-@withStyles(theme => ({
-  '@global': {
-    'div.react-lazylog': {
-      backgroundColor: theme.palette.background.default,
-      fontFamily: 'Consolas, Monaco, Andale Mono, Ubuntu Mono, monospace',
-      fontSize: 13,
-      paddingTop: 4,
-      paddingBottom: theme.spacing.unit,
-      color: theme.palette.text.primary,
-      '-webkit-font-smoothing': 'auto',
+@withStyles(theme => {
+  const filterStyles = {
+    borderRadius: 4,
+    padding: '4px 8px',
+    lineHeight: 1,
+    '&.active': {
+      ...theme.mixins.successIcon,
+      fill: THEME.PRIMARY_TEXT_DARK,
     },
-    'div.react-lazylog-searchbar': {
-      backgroundColor: theme.palette.background.default,
-      fontFamily: 'Consolas, Monaco, Andale Mono, Ubuntu Mono, monospace',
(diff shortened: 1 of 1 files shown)


ui/src/components/Log/index.jsx
```diff
@@ -19,89 +19,105 @@ const LINE_NUMBER_MATCH = /L(\d+)-?(\d+)?/;
...
-      paddingBottom: theme.spacing.unit,
-      color: theme.palette.text.primary,
-      '-webkit-font-smoothing': 'auto',
+@withStyles(theme => {
+  const filterStyles = {
+    padding: '4px 8px',
+    lineHeight: 1,
+    '&.active': {
+      ...theme.mixins.successIcon,
+      borderRadius: '4px',
```
helfi92: The border radius could be set for all times, not just when the button is active. Also, we can omit the px and remove the quotations: `borderRadius: 4,`.

ui/src/components/Log/index.jsx
```diff
@@ -19,89 +19,105 @@ const LINE_NUMBER_MATCH = /L(\d+)-?(\d+)?/;
...
+    fabIcon: {
+      ...theme.mixins.fabIcon,
+    },
+    logSpeedDial: {
+      ...theme.mixins.fab,
+      ...theme.mixins.actionButton,
+      bottom: theme.spacing.triple,
+    },
+  };
+})
```
helfi92: Any chance you could give an overview of the changes you made? The diff is a bit hard to follow for some reason 😅

arku:
> Let's make the fill white instead of black when the button is active:
> 
> Screen Shot 2019-08-15 at 10 06 05 AM
> 
> If we could also do the same for the follow log button, that'd be great :)

I left it black to be consistent with the follow log button styles. Let me change them both to white.

ui/src/components/Log/index.jsx
arku: Done.

ui/src/components/Log/index.jsx
arku: Sure. I added a `filterStyles` object [here](https://github.com/taskcluster/taskcluster/pull/1183/files#diff-6af56ca36cff0662454eca418e5191b6R24) for the active filter styles in both dark and light themes. And I updated the follow log button styles [here](https://github.com/taskcluster/taskcluster/pull/1183/files#diff-6af56ca36cff0662454eca418e5191b6R110). 

Does that help?

arku:
@helfi92 I have addressed your comments. Do you mind taking another look?

arku:
@helfi92 It looks much better now 😍 

Thanks!

Issue #749, opened 2019-03-08
switch python client to use pytest


escapewindow:
pytest is a better test framework; let's use it.

melvin2016:
where should i start?

djmitche:
Hi!  Probably a good place to start is to get the repository checked out and get the *current* test suite running.  You'll want to work in the `clients/client-py` subdirectory of this repository -- you can safely ignore everything else in the repository.

Once that's working for you, let's come up with a plan to use pytest, instead of the existing `unittest` approach.  Rewriting one or two of the tests to use Pytest will be a nice proof-of-concept, but we'll want all of them rewritten in order to merge the changes.

helfi92:
@melvin2016 Are you still working on this?

melvin2016:
Sorry for this.. can you take this up

Shubhamchinda:
Hey, can I work on this ?

djmitche:
You sure can!  My comments from June 1 should get you started.

Shubhamchinda:
> You sure can! My comments from June 1 should get you started.

Thanks, I'll look into it and ask you if I get stuck.

Patil2099:
There is no requirements.txt file Adding one would save some time in installing dependencies @djmitche

djmitche:
@Patil2099 that sounds like a good PR!

Patil2099:
Where should I add the text file ?

djmitche:
where you'd expect to find it :)

carpecodeum:
hey can i work on this issue? this would be a good first issue for me.

djmitche:
I believe @Shubhamchinda is already working on it -- sorry!

vishakhanihore:
@djmitche Its good to be closed now I think.

PR #1187, opened 2019-08-15
Switched python client to pytest

Shubhamchinda:
#749 
@djmitche 

Hey, I used `pytest` in file `clients/clients-py/test/test_async.py`. Let me know if I did it correct, then we'll make the changes in other files too.

Thanks a lot!


djmitche:
I'll defer to @taskcluster/python-reviewers to have a look at this.

Shubhamchinda:
> I'll defer to @taskcluster/python-reviewers to have a look at this.

Okay :)

clients/client-py/test/test_async.py
```diff
@@ -1,15 +1,16 @@
...
 
 import asyncio
 
 import base
 import taskcluster.aio.auth as subjectAsync
 
 
-@unittest.skipIf(os.environ.get('NO_TESTS_OVER_WIRE'), "Skipping tests over wire")
+@pytest.mark.skipif(os.environ.get('NO_TESTS_OVER_WIRE'), reason ="Skipping tests over wire")
 class TestAuthenticationAsync(base.TCTest):
```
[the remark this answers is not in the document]
Shubhamchinda: Okay @escapewindow got it! 

One more thing that should be the value of `REAL_ROOT_URL` ?

Thanks a lot! :)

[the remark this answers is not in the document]
escapewindow: No worries.

I'm referring to [these lines](https://github.com/taskcluster/taskcluster/blob/master/clients/client-py/test/base/__init__.py#L18-L22). We should add

```
TEST_ROOT_URL = 'https://tc-tests.example.com'
# rootUrl of a real deployment (that needs no pre-configuration)
REAL_ROOT_URL = 'https://taskcluster.net'
```

to the file, usable without `TCTest`.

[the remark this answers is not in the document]
Shubhamchinda: > ```
> TEST_ROOT_URL = 'https://tc-tests.example.com'
> # rootUrl of a real deployment (that needs no pre-configuration)
> REAL_ROOT_URL = 'https://taskcluster.net'
> ```

Thanks a lot! I'll update the PR by this weekend. :)

Shubhamchinda:
@escapewindow I made the suggested changes, However I got stuck in file `clients/clients-py/test/test_client.py` class `ClientTest(base.TCTest)` is used in other classes as well. 

The file `clients/clients-py/test/test_async.py` is complete I guess. And in file `clients/clients-py/test/test_utils.py` only one test case is failing. Could you please help?

Thanks a lot :)

clients/client-py/test/test_utils.py
```diff
@@ -31,214 +33,188 @@ def dst(self, dt):
...
-        @httmock.all_requests
-        def response_content(url, requet):
-            return {'status_code': 404}
-
-        with httmock.HTTMock(response_content):
-            d = subject.makeSingleHttpRequest('GET', 'http://www.example.com', {}, {})
-            with self.assertRaises(requests.exceptions.RequestException):
-                d.raise_for_status()
 
+def test_success_put_file():
```
Shubhamchinda: @escapewindow 
This test case is failing with the error 
```
def test_success_put_file():
        with mock.patch.object(subject, 'makeSingleHttpRequest') as p:
            class FakeResp:
                status_code = 200
    
                def raise_for_status():
                    pass
    
            p.return_value = FakeResp()
>           subject.putFile('setup.py', 'http://www.example.com', 'text/plain')

test_utils.py:178:
```
---

```
filename = 'setup.py', url = 'http://www.example.com', contentType = 'text/plain'

    def putFile(filename, url, contentType):
>       with open(filename, 'rb') as f:
E       FileNotFoundError: [Errno 2] No such file or directory: 'setup.py'
taskcluster\utils.py:298: FileNotFoundError

Comment (escapewindow) [reply clients/client-py/test/test_utils.py]: Looking.

File: clients/client-py/test/test_async.py [hunk 3]
```diff
@@ -1,63 +1,64 @@
...
-                }, session=session)
-
-                result = await client.testAuthenticate({
-                    'clientScopes': ['test:*'],
-                    'requiredScopes': ['test:xyz'],
-                })
-                self.assertEqual(result, {'scopes': ['test:xyz'], 'clientId': 'tester'})
-
-        loop.run_until_complete(x())
+@pytest.mark.skipif(os.environ.get('NO_TESTS_OVER_WIRE'), reason ="Skipping tests over wire")
```
escapewindow: You want to have this directly above line 15 (the line before the `def test...` line). You also want one before line 39.

clients/client-py/test/test_utils.py
escapewindow: I was able to fix with:

```
diff --git a/clients/client-py/test/test_utils.py b/clients/client-py/test/test_utils.py
index a2bdda8e2..b20150064 100644
--- a/clients/client-py/test/test_utils.py
+++ b/clients/client-py/test/test_utils.py
@@ -171,11 +171,12 @@ def test_success_put_file():
         class FakeResp:
             status_code = 200

-            def raise_for_status():
+            def raise_for_status(*args):
                 pass

         p.return_value = FakeResp()
-        subject.putFile('setup.py', 'http://www.example.com', 'text/plain')
+        path = os.path.join(os.path.dirname(os.path.dirname(__file__)), 'setup.py')
+        subject.putFile(path, 'http://www.example.com', 'text/plain')
         p.assert_called_once_with('put', 'http://www.example.com', mock.ANY, mock.ANY, mock.ANY)
```

The `raise_for_status` change avoids the error `TypeError: raise_for_status() takes 0 positional arguments but 1 was given` (it expects a `self` arg). What I'm doing with `os.path...` is finding the path of the test file (e.g., `client-py/test/test_utils.py`), finding the directory of the directory that holds it (e.g., `client.py/test/..`, which becomes `client-py/`), and then finding `setup.py` in that directory.

escapewindow:
> @escapewindow I made the suggested changes, However I got stuck in file `clients/clients-py/test/test_client.py` as class `ClientTest(base.TCTest)` is used in other classes as well.

`realTimeSleep` is only referenced in one place, so we can make that a top level constant (e.g., `REAL_TIME_SLEEP = time.sleep`)

The rest of it is in `setUp` and `tearDown`. We can handle this with [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html). Essentially, there are function args that will get pre-populated with objects of your choosing. For instance, if you add `pytest-mock` to your test dependencies, you get access to the `mocker` fixture. You can use this by

```
def test_foo(mocker):
    # now you can mock things with the `mocker` object
    mocker.patch.object(object, 'function')
    ...
```

Or the `tmpdir` fixture, which gives you access to a temporary directory that gets cleaned up after each test run.

You can write your own fixtures to do things like we're doing in `setUp` and `tearDown` at the moment. For instance, you could

```
@pytest.yield_fixture(scope='function')
def apiRef():
    keys = [
        base.createTopicExchangeKey('primary_key', constant='primary'),
        base.createTopicExchangeKey('norm1'),
        base.createTopicExchangeKey('norm2'),
        base.createTopicExchangeKey('norm3'),
        base.createTopicExchangeKey('multi_key', multipleWords=True),
    ]
    topicEntry = base.createApiEntryTopicExchange('topicName', 'topicExchange', routingKey=keys)
    entries = [
        base.createApiEntryFunction('no_args_no_input', 0, False),
        base.createApiEntryFunction('two_args_no_input', 2, False),
        base.createApiEntryFunction('no_args_with_input', 0, True),
        base.createApiEntryFunction('two_args_with_input', 2, True),
        base.createApiEntryFunction('NEVER_CALL_ME', 0, False),
        topicEntry
    ]
    apiRef = base.createApiRef(entries=entries)
    yield apiRef  # the test gets this
    # we can do any post-test cleanup here if needed


def test_apiVersion_set_correctly_default(self, apiRef):
    # `apiRef` is already pre-populated. we have our own copy, so
    # we can modify it and the next test run will get a clean copy
```

Does that make sense? It will take a bit of work to see what parts of `setUp` are used, and creating fixtures from them, but once we create the fixtures, they're more useful and flexible.

> The file `clients/clients-py/test/test_async.py` is complete I guess. And in file `clients/clients-py/test/test_utils.py` only one test case is failing. Could you please help?

`test_utils` fix is above. Or below. Somewhere :) I can take a look at `test_async` as well as the rest .

clients/client-py/test/test_client.py
```diff
@@ -645,311 +645,307 @@ def fakeSite(url, request):
...
-
-    def test_permacred_insufficient_scopes(self):
-        """A call with insufficient scopes results in an error"""
-        client = subject.Auth({
-            'rootUrl': self.real_root_url,
+        assert self.gotUrl == 'https://tc-tests.example.com/api/fake/v1/two_args_with_input/a/b'
+        assert json.loads(self.gotRequest.body) == {"x": 1}
+
+
+@pytest.mark.skipif(os.environ.get('NO_TESTS_OVER_WIRE'), reason = "Skipping tests over wire")
```
escapewindow: Same here: this decorator needs to go before every test function. Alternately, you can skip every test in a file if you do something like

```
pytestmark = [
    pytest.mark.skipif(os.environ.get("NO_TESTS_OVER_WIRE"), reason="Skipping tests over wire")
]
```

at the top of the file after the imports.

Shubhamchinda: @escapewindow Got it, Thanks!
I'll update thr PR.

Shubhamchinda:
Hey @escapewindow , Sorry for the late response. 

I updated the PR, Please take a look.

clients/client-py/test/test_client.py
```diff
@@ -618,49 +639,51 @@ def test_builds_surl_keyword(self):
...
+    yield fakeResponse
+
+@pytest.fixture(scope='function')
+def fakeSite(url, request, fakeResponse):
+    gotUrl = urllib.parse.urlunsplit(url)
+    gotRequest = request
+    return fakeResponse
+
+
+def test_no_args_no_input(client, fakeSite):
```
Shubhamchinda: From this test case, the next five test are are failing. I tried to make a fixture for `fakeSite & fakeResponse` but I think the test `test_no_args_no_input` can't resolve the `fakeSite`.

I'm getting this error : 

Screenshot (251)

the `fakeSite` fixture can't resolve `url`, I don't know why. 😕
Could you please help me here?

PS : I think rest of the files are fixed.

Thanks a lot! :)

escapewindow:
👍 looking, will get back to you soon.

clients/client-py/test/test_client.py
```diff
@@ -18,281 +17,306 @@
...
+    yield client
+
+@pytest.yield_fixture(scope='function')
+def patcher(client):
+    patcher = mock.patch.object(client, 'NEVER_CALL_ME')
+    yield patcher
+
+
+def tearDown():
+    time.sleep = REAL_TIME_SLEEP
```
escapewindow: This function doesn't do anything. This belongs in the cleanup section of a yield fixture, or we can use fixtures that clean up after themselves like `mocker`.

clients/client-py/test/test_client.py
```diff
@@ -310,646 +334,626 @@ def raise_for_status(self):
...
 
-    def setUp(self):
-        ClientTest.setUp(self)
-        patcher = mock.patch.object(self.client, 'NEVER_CALL_ME')
-        never_call = patcher.start()
-        never_call.side_effect = AssertionError
-        self.addCleanup(never_call.stop)
+""" This class covers both the _makeApiCall function logic as well as the
+logic involved in setting up the api member functions since these are very
+related things"""
```
escapewindow: I think this comment no longer applies here. We should probably remove it.

clients/client-py/test/test_client.py
escapewindow: I believe I fixed this. I can show you the patch if you want, but I basically:

1. `import functools`
2. change `fakeSite` from a fixture, where all the args and kwargs need to be fixtures, to a function with kwargs we can fill in:

```
def fakeSite(url, request, expected_url=None, expected_body=None):
    if expected_url is not None:
        assert urllib.parse.urlunsplit(url) == expected_url
    if expected_body is not None:
        assert json.loads(request.body) == expected_body
    return ""
```

Then, to use:

```
def test_no_args_no_input(client):
    site = functools.partial(
        fakeSite,
        expected_url='https://tc-tests.example.com/api/fake/v1/no_args_no_input',
        # if we have an expected body, also pass `expected_body={...},`
    )
    with httmock.HTTMock(site):
         client.no_args_no_input()
```

For the `json.loads` values, set `expected_body`. Essentially, we're creating another function based on `fakeSite`, but with `expected_body` and `expected_url` optionally pre-populated.

After I greened this up, we get a bunch of linting errors from flake8... those will need cleaning up too. Luckily that tells us what needs fixing on which line, and how to fix.

clients/client-py/test/base/__init__.py
```diff
@@ -8,6 +8,10 @@
 # Mocks really ought not to overwrite this
 _sleep = time.sleep
 
+TEST_ROOT_URL = "https://tc-tests.example.com"
+# rootUrl of a real deployment (that needs no pre-configuration)
+REAL_ROOT_URL = 'https://taskcluster.net'
+
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 log.addHandler(logging.NullHandler())
```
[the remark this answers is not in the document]
Shubhamchinda: Okay

clients/client-py/test/test_client.py
```diff
@@ -18,281 +17,306 @@
...
+        base.createApiEntryFunction('no_args_no_input', 0, False),
+        base.createApiEntryFunction('two_args_no_input', 2, False),
+        base.createApiEntryFunction('no_args_with_input', 0, True),
+        base.createApiEntryFunction('two_args_with_input', 2, True),
+        base.createApiEntryFunction('NEVER_CALL_ME', 0, False),
+        topicEntry
+    ]
+    apiRef = base.createApiRef(entries=entries)
+    clientClass = subject.createApiClient('testApi', apiRef)
+    client = clientClass({'rootUrl': base.TEST_ROOT_URL})
```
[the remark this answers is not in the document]
Shubhamchinda: > I'm not sure if `client` and `clientClass` are doing anything here. You created fixtures for these below, which is good; that's how the test functions would use `client` and `clientClass`. You can probably remove these two lines.

Yeah, I removed them initially but then all tests seems to have error. Ssince we have fixtures, it shouldn't happen, right? It's a little confusing.

clients/client-py/test/test_client.py
```diff
@@ -18,281 +17,306 @@
...
+    patcher = mock.patch.object(client, 'NEVER_CALL_ME')
+    never_call = patcher.start()
+    never_call.side_effect = AssertionError
+    fakeResponse = ''
+
+    def fakeSite(url, request):
+        gotUrl = urllib.parse.urlunsplit(url)
+        gotRequest = request
+        return fakeResponse
+    fakeSite = fakeSite
```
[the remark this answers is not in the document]
Shubhamchinda: I'll remove it. :)

clients/client-py/test/test_client.py
Shubhamchinda: Sorry, makes sense. I'll remove it too.

Shubhamchinda:
>https://github.com/taskcluster/taskcluster/pull/1187#discussion_r320983578
Thanks a lot @escapewindow , I really appreciate your help. :)
I'll update the changes by this weekend :)

Shubhamchinda:
Hey @escapewindow , I updated the PR. Ready for another look.

I want to know how can I check and fix flake8 errors? 

Thanks :)

djmitche:
You can install flake8 with `pip install flake8`, then run
```
./lint.sh --python python --flake8 flake8
```

clients/client-py/tox.ini
```diff
@@ -33,5 +33,5 @@ extras = test
 usedevelop = true
 
 commands =
-    {envbindir}/nosetests -v --with-noseexclude --with-xunit --rednose --force-color {env:IG:} {env:EX1:} {env:EX2:}
+    {envbindir}/py.test --cov-config .coveragerc --cov=taskcluster --cov-report term-missing -v --with-noseexclude --with-xunit --rednose --force-color {env:IG:} {env:EX1:} {env:EX2:}
```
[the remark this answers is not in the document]
escapewindow: It looks like we still need to make this change? Once tox works, you'll see the linting errors at the end of the tox logs, both in the automated tests and if you run tox locally.

clients/client-py/setup.py
```diff
@@ -9,6 +9,8 @@
 tests_require = [
     'nose',
     'nose-exclude',
+    'pytest',
+    'pytest-cov',
```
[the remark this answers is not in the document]
escapewindow: We'll need to add `pytest-mock` here for the `mocker` fixture.

clients/client-py/test/test_async.py
```diff
@@ -1,63 +1,66 @@
...
+            }, session=session)
+            result = await client.testAuthenticate({
+                'clientScopes': ['test:a'],
+                'requiredScopes': ['test:a'],
+            })
+            assert result == {'scopes': ['test:a'], 'clientId': 'tester'}
+
+    loop.run_until_complete(x())
+
+@pytest.mark.skipif(os.environ.get('NO_TESTS_OVER_WIRE'), reason ="Skipping tests over wire")
```
escapewindow: We no longer need this line, since we get this skipif from lines 12-14.

clients/client-py/test/test_client.py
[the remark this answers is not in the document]
escapewindow: I was able to comment out lines 52-59 and still get green tests.

clients/client-py/test/test_client.py
escapewindow: This is still here?

Shubhamchinda:
> So:
> 
> * tox.ini fix
> * docstring cleanup in test_client.py
> * if we use mocker, we'll need pytest-mock in setup.py test_requires
> * the linting fix list is long. You should be able to get them via running `tox` when you fix `tox.ini`, or via [Dustin's suggestion](https://github.com/taskcluster/taskcluster/pull/1187#issuecomment-533702954)

Hey @escapewindow , I fixed all the changes but I wasn't able to run `tox`, So I fixed the errors with [Dustin's suggestion](https://github.com/taskcluster/taskcluster/pull/1187#issuecomment-533702954).
I tried running tox with `tox` command I get these errors: 
Screenshot (12)

How can I fix this?

Thanks a lot :)

escapewindow:
Hm, I'm not able to replicate that error. I wonder if it's windows only?

If you look at the test results, taskcluster-client-py-3.6 and taskcluster-client-py3.7 are both green, which is good!

I fixed py2.7 by

```diff
diff --git a/clients/client-py/tox.ini b/clients/client-py/tox.ini
index 0b084240f..4e92815b4 100644
--- a/clients/client-py/tox.ini
+++ b/clients/client-py/tox.ini
@@ -6,7 +6,7 @@ envlist =

 [testenv:py27]
 setenv =
-    EX1 = --ignore=test/aio
+    EX1 = --ignore=test/test_async.py
     EX2 = --ignore=taskcluster/aio
     IG = "--ignore=.*async.*"

@@ -17,5 +17,5 @@ extras = test
 usedevelop = true

 commands =
-    {envbindir}/py.test --cov-config .coveragerc --cov=taskcluster --cov-report term-missing -v {env:IG:} {env:EX1:} {env:EX2:} 
+    {envbindir}/py.test --cov-config .coveragerc --cov=taskcluster --cov-report term-missing -v {env:IG:} {env:EX1:} {env:EX2:}
     ./lint.sh --python "{envpython}" --flake8 "{envbindir}/flake8"
```

(I just trimmed trailing whitespace at the end of the py.test commandline. It looks like `test_async.py` is not under `test/aio`, so the first change is the actual fix.) Once you land that, the py2.7 task should go green.

I'm not sure why the worker-manager test broke... might be intermittent. The upload-coverage task failed because the worker-manager task broke.

djmitche:
Azure has been having some ongoing issues that are causing tests that use Azure to be intermittent.  You can safely ignore any failing service tests.  Sorry about that!

djmitche:
(https://bugzilla.mozilla.org/show_bug.cgi?id=1579496)

Shubhamchinda:
> Hm, I'm not able to replicate that error. I wonder if it's windows only?

Maybe yes, I'll try it with wsl.

> I fixed py2.7 by
 Thanks for suggesting the fix :)

> I'm not sure why the worker-manager test broke... might be intermittent. The upload-coverage task failed because the worker-manager task broke.

I think it's not related to our changes! 
Thanks to @djmitche for confirming it in [comment](https://github.com/taskcluster/taskcluster/pull/1187#issuecomment-535008402).

Thanks a lot!

Shubhamchinda:
> 🎉
> Thank you! And much respect to you: this is a significant change.

Thanks to you for being a great support throughout this issue. :)

PR #1188, opened 2019-08-15
Created requirements.txt

Patil2099:
#749

### DIFF
--- a/clients/client-py/requirements.txt
+++ b/clients/client-py/requirements.txt
@@ -6,4 +6,4 @@ mock==3.0.5
 importlib==1.0.4
 six==1.12.0
 hypothesis==4.32.3
-
+mohawk==1.0.0
--- a/clients/client-py/requirements.txt
+++ b/clients/client-py/requirements.txt
@@ -1,0 +1,9 @@
+taskcluster==16.0.0
+pytest==5.0.1
+requests==2.22.0
+httmock==1.3.0
+mock==3.0.5
+importlib==1.0.4
+six==1.12.0
+hypothesis==4.32.3
+


escapewindow:
Thank you for this PR!

First off, what problem are we trying to solve here? The requirements are listed in `setup.py`, both for installation and testing. We automatically install the install requirements on installation of the python client; we automatically install the test requirements on running `tox` to run the tests. Is there another use case you're trying to solve for?

If we want to land requirements file(s):

1. These are the test requirements, not the install requirements. I generally expect to find install requirements at requirements.txt, if any. We could deal with that by having a requirements.txt and a requirements-test.txt. In releng we've dealt with that via [pip-compile-multi](https://pypi.org/project/pip-compile-multi/), meaning we would have a `requirements/base.in` and a `requirements/test.in` that we compile into `requirements/base.txt` and `requirements/test.txt` (or even `requirements/base-py36.txt` etc.). However, two static files work, as long as they're maintained.
2. if we have a requirements.txt (or multiple), then ideally it would be the source of truth. That means we should change `setup.py` to read from the `requirements.txt`, and we would need to update the packaging to include `requirements.txt`. This would be true for `requirements-test.txt` as well.

Patil2099:
> Thank you for this PR!
> 
> First off, what problem are we trying to solve here? The requirements are listed in `setup.py`, both for installation and testing. We automatically install the install requirements on installation of the python client; we automatically install the test requirements on running `tox` to run the tests. Is there another use case you're trying to solve for?
> 
> If we want to land requirements file(s):
> 
> 1. These are the test requirements, not the install requirements. I generally expect to find install requirements at requirements.txt, if any. We could deal with that by having a requirements.txt and a requirements-test.txt. In releng we've dealt with that via [pip-compile-multi](https://pypi.org/project/pip-compile-multi/), meaning we would have a `requirements/base.in` and a `requirements/test.in` that we compile into `requirements/base.txt` and `requirements/test.txt` (or even `requirements/base-py36.txt` etc.). However, two static files work, as long as they're maintained.
> 2. if we have a requirements.txt (or multiple), then ideally it would be the source of truth. That means we should change `setup.py` to read from the `requirements.txt`, and we would need to update the packaging to include `requirements.txt`. This would be true for `requirements-test.txt` as well.

It was for testing.

escapewindow:
Something like `tox --develop --notest` will create virtualenvs in `.tox/py{27,36,37}` with all the test dependencies installed. Does that work for you?

djmitche:
It sounds like agreement is to not do this..

PR #1189, opened 2019-08-15
Bug 1571133 - report worker pool errors to service log

djmitche:
These aren't reported as errors in the service log (so wouldn't go to
StackDriver, for example), but still constitute useful data for service
operators who might, for example, monitor the frequency of such errors
and use that to get alerted of cloud-related outages (with attendant
false-positives from someone misconfiguring their worker pool).

Bugzilla Bug: [1571133](https://bugzilla.mozilla.org/show_bug.cgi?id=1571133)

### DIFF
--- a/services/worker-manager/src/monitor.js
+++ b/services/worker-manager/src/monitor.js
@@ -35,4 +35,25 @@ monitorManager.register({
   },
 });
 
+monitorManager.register({
+  name: 'workerError',
+  type: 'worker-error',
+  title: 'Worker Error',
+  level: 'warning',
+  version: 1,
+  description: `
+    An error was reported regarding the given worker pool.  Such errors are generally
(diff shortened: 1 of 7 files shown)


services/worker-manager/src/monitor.js
```diff
@@ -35,4 +35,25 @@ monitorManager.register({
   },
 });
 
+monitorManager.register({
+  name: 'workerError',
+  type: 'worker-error',
+  title: 'Worker Error',
+  level: 'warning',
```
imbstack: Maybe even just a notice? I'm happy either way.

djmitche: hm, makes sense -- if someone's looking for warnings, this might be a big spew of irrelevant stuff.

PR #1190, opened 2019-08-15
Cleanup configs and make them understandable

imbstack:
As it says on the tin


services/hooks/config.yml
```diff
@@ -2,9 +2,9 @@ defaults:
   # defaults outlining all configuration options and some sane defaults
   app:
     component:        hooks
-    hookTableName:     !env HOOK_TABLE_NAME
-    lastFireTableName: !env LASTFIRE_TABLE_NAME
-    queuesTableName:  !env QUEUE_TABLE_NAME
+    hookTableName: Hooks
+    lastFireTableName: LastFire3
```
djmitche: Will this require a corresponding change to the static client?  And, do I recall that changes to the static clients don't get deployed automatically?

dev-docs/dev-config-example.yml
```diff
@@ -150,9 +150,6 @@ github:
   pulse_username: ...
```
djmitche: note for later, but this file should probably be ignored in .gitattributes and .git/CODEOWNERS?

services/hooks/config.yml
imbstack: Why will changing the table name require updating clients? I think I'm missing something or other here.

djmitche: The hooks client has scope `auth:azure-table:read-write:${azureAccountId}/LastFire`, not `auth:azure-table:read-write:${azureAccountId}/LastFire3`

imbstack: Oh, those clients! haha, I was thinking of every other client for some reason. Yes, good catch.

PR #1199, opened 2019-08-19
Remove use of dot

djmitche:
Dot has an unfixed sec vuln (WS-2019-0043).  That does not affect our
praticular use of dot, but neither do we actually need this feature.


djmitche:
:icecream:

PR #1202, opened 2019-08-19
Add Third Party Login

helfi92:


services/web-server/config.yml
```diff
@@ -44,6 +54,7 @@ defaults:
     vhost: !env PULSE_VHOST
 
   login:
+    registeredClients: !env:json REGISTERED_CLIENTS
```
helfi92: Example of a registered client:

```yaml
- clientId: Treeherder
  scope: ["assume:test"]
  redirectUri:
    - "http://localhost:5000"
  whitelisted: true
  responseType: "code"
```

djmitche:
I'm still in the process of reviewing and will finish tomorrow.  But I did push a bit of a test framework if you want to have a look.  Things mostly work until the getCredentials call :)

ui/src/components/SignInDialog/index.jsx
```diff
@@ -28,18 +29,43 @@ export default class SignInDialog extends Component {
     credentialsDialogOpen: false,
   };
 
+  shouldStartOauth2LoginFlow() {
```
djmitche: Why would we start this from the sign-in dialog?

ui/src/components/SignInDialog/index.jsx
```diff
@@ -28,18 +29,43 @@ export default class SignInDialog extends Component {
...
-    const { onAuthorize, onClose } = this.props;
+    const {
+      shouldStartOauth2LoginFlow,
+      props: { onAuthorize, onClose },
+    } = this;
 
     window.addEventListener(
       'message',
-      function handler(e) {
+      async function handler(e) {
```
djmitche: does addEventListener support async functions?

djmitche:
Some rfc updates in https://github.com/taskcluster/taskcluster-rfcs/pull/150.

services/web-server/src/graphql/Auth.graphql
```diff
@@ -29,4 +29,9 @@ extend type Query {
   # monitor this expiration and refresh the credentials if necessary, by calling
   # this endpoint again, if they have expired.
   getCredentials: Credentials
+  # Return true if the user is logged.
+  #
+  # Sessions are stored in memory so in a case of a server restart,
+  # the front end will need to log the user out
```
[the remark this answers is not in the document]
helfi92: > Even in the case where sessions are in a DB, we'll still need to know if the user has an active session, right?

I was thinking perhaps the UI can log the user out if the session is expired for example.

ui/src/components/SignInDialog/index.jsx
helfi92: I believe so. I think async just doesn't work in constructors.

services/web-server/src/graphql/Auth.graphql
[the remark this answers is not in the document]
helfi92: >    Even in the case where sessions are in a DB, we'll still need to know if the user has an active session, right?

I was thinking perhaps the UI can log the user out if the session is expired for example.

ui/src/components/SignInDialog/index.jsx
djmitche: Hm, [addEventListener docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#The_event_listener_callback) dont't say so, and the [handleEvent](https://developer.mozilla.org/en-US/docs/Web/API/EventListener/handleEvent#Return_value) docs also suggest the opposite.  Anyway, this is a minor issue -- a `.catch(..)` should fix it up.

helfi92: I just tested on the branch master via:

```diff
diff --git a/ui/src/components/SignInDialog/index.jsx b/ui/src/components/SignInDialog/index.jsx
index abe48e58d..eddd0f7de 100644
--- a/ui/src/components/SignInDialog/index.jsx
+++ b/ui/src/components/SignInDialog/index.jsx
@@ -28,16 +28,26 @@ export default class SignInDialog extends Component {
     credentialsDialogOpen: false,
   };
 
+  sleep = () =>
+    new Promise(resolve => {
+      window.setTimeout(resolve, 3000);
+    });
+
   componentDidMount() {
     const { onAuthorize, onClose } = this.props;
+    const { sleep } = this;
 
     window.addEventListener(
       'message',
-      function handler(e) {
+      async function handler(e) {
         if (e.origin !== window.origin || !e.data || e.data.type !== 'login') {
           return;
         }
 
+        console.time('sleep');
+        await sleep();
+        console.timeEnd('sleep');
+
         onAuthorize(UserSession.create(e.data));
         window.removeEventListener('message', handler);
         onClose();
```

The console shows `sleep: 3005ms - timer ended`

helfi92: I think the docs just say that  it won't consider the return value but there's nothing stopping the callback from being asynchronous.

djmitche: The usual concern is around error handling -- in Node an async error would be considered an UnhandledException and that's a big no-no, but it looks like the browser reports that error the same way it would a synchronous error in an event handler.  Sorry for the noise :)

helfi92: No worries :)

helfi92:
OK, I think this is ready for review. I added some deployment docs but it's still missing docs in the access-control chapter.

Question: CI test is complaining about a missing secret `secrets missing and NO_TEST_SKIP is set: taskcluster`. Any idea how to remove this error? Tests pass locally fwiw.

djmitche:
> `secrets missing and NO_TEST_SKIP is set: taskcluster`

I think that's because there's no TC credentials set up for `taskcluster-web-server`'s tests (creds that would let it access Azure tables for testing).  I can add that.

djmitche:
>  I can add that.

I think this is done.

services/web-server/src/entities/AccessToken.js
```diff
@@ -0,0 +1,37 @@
+const assert = require('assert');
+const Entity = require('azure-entities');
+
+const AccessToken = Entity.configure({
```
imbstack: Has access to these tables been added to `infrastructure/tooling/static-clients.yml`? We should also make sure to have a changelog stating that clients need updated in deployments for cloudops

services/web-server/src/servers/oauth2.js
```diff
@@ -0,0 +1,306 @@
...
+      if (!client) {
+        return done(null, false);
+      }
+
+      if (!client.redirectUri.some(uri => uri === redirectURI)) {
+        return done(null, false);
+      }
+
+      return done(null, client, redirectURI);
+    }, async (client, user, scope, done) => {
```
djmitche: This `immediate` argument appears undocumented in oauth2orize (but most of the library is undocumented).  As I understand it, it's called to decide whether to immediately approve the request and return a redirect to the `redirect_uri`.  That's the second argument to `done` here.  Since it's undocumented, a comment explaining that might help the next reviewer :)

services/web-server/src/servers/oauth2.js
```diff
@@ -0,0 +1,306 @@
...
+      // Skip consent form if the client is whitelisted
+      if (client.whitelisted && user && _.isEqual(client.scope.sort(), scope.sort())) {
+        // Resetting the access token is the default behavior for whitelisted clients.
+        // One less click in the UI.
+        await auth.resetAccessToken(user.identity);
```
djmitche: We don't reset access tokens anymore, so I think this can just skip to the `done(null, true, ..)` call.

services/web-server/test/third_party_test.js
```diff
@@ -0,0 +1,478 @@
...
+  const getQuery = (url, sep='?') => {
+    const qmark = url.indexOf(sep);
+    if (qmark === -1) {
+      return new URLSearchParams();
+    }
+    return new URLSearchParams(url.slice(qmark + 1));
+  };
+
+  suite('unit', function() {
+    test('authorization endpoint redirects to the root page if user is not logged in', async function() {
```
djmitche: Is this the intent?  My hunch would be that it would redirect to `/third-party` page in the UI which would ask the user to click the "sign in" button and not display the AuthConsent form until that's done.

djmitche:
https://tools.taskcluster.net/groups/EIJeKJPGQpSVXIxueZVHxA/tasks/A8lIfl4hQkeAXzy6KdjeUw/details looks a lot better regarding the test scopes..

ajvb:
> OAuth2 also has "confidential clients", which can only do the Authorization Code Flow and include their credentials in Authorization: Basic <base64(`${client_id}:${client_secret}`)>. @ajvb I'm going to guess we should require that for all registered clients with `responseType: code`? Should we defer that to a followup?

@djmitche I'm not sure if I understand the question. When you say "should require that", what is `that` in this context? As well, when you say `registered clients with responseType: code` does this mean all clients that are asking to retrieve access tokens and refresh tokens using the code grant system?

---

Also, this PR seems to still be very much in-progress. If I am incorrect in that, let me know. Otherwise, I'd prefer to do a review once the code is nearly ready to be merged.

djmitche:
> @djmitche I'm not sure if I understand the question. When you say "should require that", what is that in this context? As well, when you say registered clients with responseType: code does this mean all clients that are asking to retrieve access tokens and refresh tokens using the code grant system?

Sorry!  Background data:
 * all clients are registered with their responseType ("code" for authorization code flow and "token" of implicit flow)
 * we don't use refresh tokens
 * currently, no clients have a `client_secret`, meaning that they are all "public clients"

The change I'm suggesting is to require that all clients with responseType code must be confidential clients.  That would mean (a) that they are issued a `client_secret` when registered and (b) they must provide that `client_secret` in the form of an `Authorization: Basic` header when calling the `/login/oauth/token` endpoint.

When I wrote the RFC, I mistakenly thought this behavior was required by the protocol.  That is, I thought the authorization code flow required confidential clients.

ui/src/App/Main.jsx
```diff
@@ -73,6 +75,18 @@ export default class Main extends Component {
...
+    // be able to complete the third party login flow.
+    if (user && user.identityProviderId === 'manual' && thirdPartyLogin) {
+      onUnauthorize();
+    }
+
+    if (user && thirdPartyLogin) {
+      window.location.href = `${window.location.origin}/login/oauth/authorize${
+        window.location.search
+      }`;
+    }
```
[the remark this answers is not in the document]
helfi92: If a third party tries to login but the user is not logged in on the site they will be prompted to login. Once logged in, we need to re-initiate the oauth2 login flow since we are dealing with a single page application. I can add a comment to make the intention clearer.

ui/src/components/SignInDialog/index.jsx
```diff
@@ -117,7 +118,10 @@ export default class SignInDialog extends Component {
                 <ListItemText primary="Sign in with GitHub" />
               </ListItem>
             )}
-            <ListItem button onClick={this.handleCredentialsDialogOpen}>
+            <ListItem
+              disabled={thirdPartyLogin}
```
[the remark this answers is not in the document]
helfi92: > I don't think they'd ever see the sign-in menu while at /third-party. Maybe I'm missing something?

The sign-in menu is simply a modal and can be displayed on any view. You can be at `/third-party` and still see the sign-in menu.

> Also, it may make sense to conditionalize this on just having /third-party as the current URL path

👍

ui/docs/manual/access-control/third-party.md
```diff
@@ -7,3 +7,4 @@ When Taskcluster is integrated with other web applications, it is useful for tho
 The third-party sign-in process is being [rewritten to use an OAuth2 flow](https://github.com/taskcluster/taskcluster-rfcs/blob/master/rfcs/0147-third-party-login.md) and will be documented here.
```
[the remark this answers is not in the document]
helfi92: Noted. Unless there is a disagreement, I think we can probably do this in a follow-up.

ajvb:
@djmitche Right, ok that makes sense to me. Yeah, referring back to the spec, it just mentions that `response_type: code` / Authorization Code Grant's are "optimized for confidential clients" - https://tools.ietf.org/html/rfc6749#section-4.1

djmitche:
> That said, most of them aren't working for me. I'm seeing:

It looks like the issue was that I had config in my `user-config.yml`.  Tests work fine for me now.

ui/docs/manual/access-control/third-party.md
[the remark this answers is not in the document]
djmitche: In fact I think you've already done this :)

helfi92:
> I think sign-ins still don't work due to the cookie issues, so there's no way to test the UI directly if I were to, say, deploy this to my dev environment. Would it be good to get that solved so we can try this out?

I think that's a good idea but probably shouldn't stop the review. I was hoping I could have that issue fixed while this is in review. We could however wait before merging this PR until it's tested on a dev environment. For the time being, I would just test locally.

ui/docs/reference/core/web-server/third-party-login.md
```diff
@@ -0,0 +1,77 @@
...
+
+# Third Party Login
+
+The "big picture" here is that a Taskcluster deployment acts as an OAuth2 authorization server and resource server.
+The "resource" that the deployment protects is Taskcluster credentials. Thus a client carries out a standard OAuth2
+authorization transaction, then uses the resulting access_token to request Taskcluster credentials as needed.
+
+The deviations from OAuth2 are as follows:
+* The scope arguments are Taskcluster scopes, with * carrying its usual meaning; and
+* The authorize request has an nonstandard, optional expires query parameter.
```
djmitche: an -> a

ui/src/components/AuthConsent/index.jsx
```diff
@@ -0,0 +1,173 @@
...
+        <input name="clientId" type="hidden" value={clientId} />
+        <div>
+          <CardHeader
+            className={classes.header}
+            component="div"
+            avatar={<img alt="Logo" height={40} src={logo} />}
+            title={registeredClientId}
+            subheader={
+              <span>
+                is requesting access to client ID <code>{clientId}</code>
```
djmitche: @ajvb might have more comments on language here, but I wonder if "..is requesting you Taskcluster credentials with the following details" would be better?

ui/src/components/AuthConsent/index.jsx
```diff
@@ -0,0 +1,173 @@
...
+    ...theme.mixins.fab,
+  },
+  allowIcon: {
+    ...theme.mixins.successIcon,
+  },
+  denySpanProps: {
+    right: theme.spacing.unit * 11,
+  },
+}))
+export default class AuthConsent extends Component {
```
djmitche: This UI is so simple now :D

ui/src/views/ThirdPartyLogin/index.jsx
```diff
@@ -0,0 +1,111 @@
...
+      !(data.currentScopes instanceof Array) ||
+      !query.transactionID
+    ) {
+      return null;
+    }
+
+    return {
+      formData: {
+        ...state.formData,
+        scopes: scopeIntersection(scopes, data.currentScopes),
```
djmitche: As a note for future reviewers, this is not the authoritative place that this scope-intersection operation occurs: `scopeIntersection` is also called in `services/web-server/src/servers/oauth2.js` and then enforced by the use of `authorizedScopes` in the `auth.createClient` call in that file.

djmitche:
Oh, and a `yarn generate` run..

helfi92:
`gvm install go1.12.9` throws

```
Building Go cmd/dist using /Users/haali/.gvm/gos/go1.4.
failed MSpanList_Insert 0x127d000 0x1a46275994d8e 0x0
fatal error: MSpanList_Insert
```

I will reset the last commit https://github.com/taskcluster/taskcluster/pull/1202/commits/e0e03a71f90eb021203f75296add47d4153cc086 since it's probably not a valid `yarn generate` output.

helfi92:
> `gvm install go1.12.9` throws

I was able to fix this with `gvm install go1.12.9 -B`.

services/web-server/procs.yml
```diff
@@ -12,6 +12,16 @@ scanner:
...
   deadline: 86400
   command: node src/main.js scanner
+expire-authorization-codes:
+  type: cron
+  schedule: '0 0 * * *'
+  deadline: 86400
+  command: node src/main.js expire-authorization-codes
+expire-access-tokens:
+  type: cron
+  schedule: '0 0 * * *'
```
ajvb: Why just once a day? Why not much more frequently? Doesn't this mean that access tokens and authorization codes (for the above one) could stay around for nearly 24 hours after they expire?

helfi92: I think this is just used for general memory cleanup. Each entry has an `expires` entry that we look at. https://github.com/taskcluster/taskcluster/pull/1202/files#diff-9e4009a92aeb7e9312ee0f512c302100R157-R159

ajvb: @helfi92 Ah ok cool, yeah that looks good. It might be worth renaming these jobs to `clean-up-expired-[access-tokens/auth-code]` or similar, but that's you call.

djmitche: This is pretty typical of how we do stuff: check the expired timestamp on every request, and then periodically clean up expired stuff.  The updated name is probably clearer :)

changelog/bug-1561905.md
```diff
@@ -0,0 +1,10 @@
+level: major
+reference: bug 1561905
+---
+
+1. Clients need to be updated in deployments.
```
djmitche: +"Static" (I think you mean static clients, right?)

PR #1204, opened 2019-08-19
Bug 1530517 -  Don't rebuild the UI on deployment

helfi92:


ui/src/App/index.jsx
```diff
@@ -68,13 +68,13 @@ export default class App extends Component {
   });
 
   httpLink = new HttpLink({
-    uri: process.env.GRAPHQL_ENDPOINT,
+    uri: window.env.GRAPHQL_ENDPOINT,
```
owlishDeveloper: I cannot google this (I just keep getting results about Windows the OS) - what exactly is `window.env`? Do you have any links to read on it?

ui/src/static/env.prod.js
```diff
@@ -0,0 +1,11 @@
+window.env = {
```
owlishDeveloper: Should these be commited to the codebase? They look like something that would rather be changed on infrastructure level

ui/src/App/index.jsx
helfi92: `window` is the global variable available to the browser: https://developer.mozilla.org/en-US/docs/Web/API/Window.

ui/src/static/env.prod.js
helfi92: I think they should but I could be wrong here. The idea was to allow people to modify these files (i.e., `build/static/env.js`) directly without having to `yarn build`. I think environment variables in the browser are not meant to contain secret values so there should probably be no worries in terms of security.

ui/src/App/index.jsx
owlishDeveloper: oic. I started to suspect that was something custom - that explains why I had trouble finding it :)

ui/src/static/env.prod.js
owlishDeveloper: It's just that it appears that you have to modify the code itself(make changes to these `env` files and commit them) in order to globally configure the application.... The idea of env variables seems to be the opposite of that - so that you could change the global configs at infrastructure level without the need to change the code when you deploy something somewhere.

Or am I misunderstanding how this is going to work?

helfi92: Right now if there is any problem with environment variables, I believe we need to recreate an image which makes the pod startup slow. That's because webpack injects environment variables at build time. What this PR does is define the environment variables at runtime so when something crashes because of misconfiguration for example, we don't need to build an new image. We can just change the `env.js` file directly and have the change be reflected right away. Once we are happy with the new config we can commit them. At least that's how I understand it.

> The idea of env variables seems to be the opposite of that - so that you could change the global configs at infrastructure level without the need to change the code when you deploy something somewhere.

Part 2 of this is to have a `config.yml` that defines these variables then copies them to `env.js` which if I understood correctly, will ease the interaction at the infrastructure level since that's how the rest of our services are configured. I may be wrong...

djmitche: Having spent a lot of time removing things referencing `taskcluster.net` from the codebase, I'm with @owlishDeveloper in being pretty skeptical of adding it back.  Especially since as of some date in the next few months, that domain won't work anymore.

> Once we are happy with the new config we can commit them.

We won't be able to check those in since they'll differ for every deployment.  There aren't even suitable default values that we can use for development, since there's no "main" deployment for which we could specify the TASKCLUSTER_ROOT_URL.

Maybe `yarn build` should not write an `env.js` file at all, and `yarn start` should be set up to write out a copy of the file based on the user's environment before running `neutrino start`, and failing if required fields like TASKCLUSTER_ROOT_URL aren't set.  Then part 2 is just doing a similar thing in the k8s startup process.

One tweak that might make that idea easier is to make all of those variables optional, or provide default values for them, so just running `yarn start` will get you something that works, even if it sometimes generates some bogus URLs (due to a default value for TASKCLUSTER_ROOT_URL).  Default values for the GraphQL endpoints could just be `/graphql` and `/subscription`, I think.  And the others seem to have decent default values already.

ui/src/static/env.js
```diff
@@ -1,8 +1,8 @@
 window.env = {
   APPLICATION_NAME: 'Taskcluster',
   TASKCLUSTER_ROOT_URL: 'https://taskcluster.net',
```
djmitche: The typical testing / fake root URL is `https://tc.example.com` -- that might be a good choice here.

ui/src/static/env.js
```diff
@@ -1,8 +1,8 @@
...
-  GRAPHQL_ENDPOINT: 'http://localhost:5080/graphql',
-  GRAPHQL_SUBSCRIPTION_ENDPOINT: 'ws://localhost:5080/subscription',
+  GRAPHQL_ENDPOINT: '/graphql',
+  GRAPHQL_SUBSCRIPTION_ENDPOINT: '/subscription',
   DOCS_ONLY: 'false',
   UI_LOGIN_STRATEGY_NAMES: 'github mozilla-auth0',
```
djmitche: ..and I think we would want this empty by default?

ui/src/static/env.prod.js
helfi92: > pretty skeptical of adding it back.

I'm not sure I follow. Adding back what?

> Maybe yarn build should not write an env.js file at all

Are you suggesting dropping the idea of `window.env`?

> yarn start should be set up to write out a copy of the file based on the user's environment before running neutrino start

Doesn't this patch already do that in .neutrinorc.js?

```js
// This code runs for both development and production
neutrino.use(copy, {
  patterns: [
    // ...
    {
      context: 'src/static',
      from: process.env.NODE_ENV === 'production' ? 'env.prod.js' : 'env.js',
      to: 'static/env.js'
    },
  ],
});
```

>  Default values for the GraphQL endpoints could just be /graphql and /subscription

Done.

ui/src/static/env.js
helfi92: Sure.

ui/src/static/env.prod.js
djmitche: > > pretty skeptical of adding it back.
> 
> I'm not sure I follow. Adding back what?

Sorry -- adding files containing the string `taskcluster.net`.

> > Maybe yarn build should not write an env.js file at all
> 
> Are you suggesting dropping the idea of `window.env`?

No, I think `window.env` is the right idea -- I'm suggesting not *writing* `env.js` in `yarn build`, and instead always writing it before the server process is started.

> > yarn start should be set up to write out a copy of the file based on the user's environment before running neutrino start
> 
> Doesn't this patch already do that in .neutrinorc.js?

It does, but that's in `yarn build`.  I'm suggesting doing so only for `yarn start`, so the result of `yarn build` has no `env.js` at all.

I'm suggesting this to try to partition the steps of "building" and "running" the UI, even when developing.  So the output of the "building" part has no configuration, and the act of "running" the UI requires the user to supply some configuration.  I realize there's a little fuzziness there since `yarn start` also runs a build.

helfi92: > No, I think window.env is the right idea -- I'm suggesting not writing env.js in yarn build, and instead always writing it before the server process is started.

`yarn build` nukes `build/` and creates a new `build/` with fresh assets. A user who `rm -rf build/` would have a new `build/` directory generated with only one file `env.js` the next time they run `yarn start`. That looks a bit odd to me. Thoughts?

djmitche: Oh, I didn't realize that `yarn start` doesn't write to `build/`.  Does it just hold all that data in memory?  Is there a way to provide dynamic content for the HTTP resource `/static/env.js` during `yarn run` without writing it to disk as part of the build process?

helfi92: That's right. It holds all of that data in memory using [webpack-dev-server](https://github.com/webpack/webpack-dev-server) to support live reloading and fast access to assets.

> Is there a way to provide dynamic content for the HTTP resource /static/env.js during yarn run without writing it to disk as part of the build process?

I can't think of anything but maybe because I don't understand the problem. Did you have something in mind? Would [creating a webpack loader](https://webpack.js.org/contribute/writing-a-loader/) be helpful here?

djmitche: It might -- or maybe before calling `neutrino start`, write out an `env.js` that is in `.gitignore` so that it doesn't get committed.  Then `yarn build` can ensure that file's not present before calling `neutrino build`, so that the file isn't part of the build.

ui/.neutrinorc.js
```diff
@@ -157,12 +159,42 @@ module.exports = {
...
+        };`;
+
+        if (!fs.existsSync(STATIC_DIR)){
+          fs.mkdirSync(STATIC_DIR);
+        }
+
+        if (!fs.existsSync(join(STATIC_DIR, 'env.js'))){
+          fs.writeFileSync(join(STATIC_DIR, 'env.js'), envJs, 'utf8');
+        }
+      }
```
djmitche: let's unlink it here, just so that we never end up accidentally including something in a production build.

helfi92: I'm not sure I understand the issue. Are you suggesting deleting `static/env.js` after L196? In what scenario would we be accidentally including something else?

PR #1206, opened 2019-08-21
Bug 1574656 - add workerConfig to google configs

djmitche:
The `userData` property remains just so that existing configs continue to work after this change.  It is ignored by tc-worker-runner.

I didn't see an easy way to assert this was being added to the metadata, but it seems clear enough.

Bugzilla Bug: [1574656](https://bugzilla.mozilla.org/show_bug.cgi?id=1574656)

### DIFF
--- a/changelog/bug1574656.md
+++ b/changelog/bug1574656.md
@@ -1,0 +1,5 @@
+level: patch
+reference: bug 1574656
+---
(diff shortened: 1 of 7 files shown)


djmitche:
r- no changelog ;)

djmitche:
Oh, also this would be a breaking change since suddenly `workerConfig` is required.  I'll make it optional.

djmitche:
..I'll get this right eventually

PR #1210, opened 2019-08-21
Fix - Date picker text color on light theme not visible

### LINKED ISSUE #1185: Date picker text color on light theme not visible
STR:

1. Navigate to https://taskcluster-ui.herokuapp.com/auth/clients/mozilla-auth0%2Fad|Mozilla-LDAP|mtabara%2F
2. Click on the calendar icon

Screen Shot 2019-08-15 at 9 32 18 AM

**Actual:**

Screen Shot 2019-08-15 at 9 08 26 AM

**Expected:** The date is visible. It works on the dark theme:

Screen Shot 2019-08-15 at 9 08 18 AM

rbrishabh:
Fixes #1185
Hey @helfi92 !

These changes in the two files were my two rather weak attempts to fix this issue.

Yep, that didn't work.

I can't figure these two things out:

 1)   Where should I add the ```className``` prop, if at all it has to be used, and change the color to black?

 2)  How do I only target my changes for light theme only?

Thanks! =)


rbrishabh:
Alright, I'll do that!

helfi92:
https://material-ui-pickers.dev/

When migrating, I think what we want is the [KeyboardDatePicker](https://material-ui-pickers.dev/api/KeyboardDatePicker). This would allow us to remove the `keyboard` prop we were giving the component in https://github.com/taskcluster/taskcluster/blob/bb08a08d5aa050680b22cd6fe227ac0339dd03ad/ui/src/components/DatePicker/index.jsx#L45.

rbrishabh:
Alright. I'll be back home in 2 hours! I'll do that as soon as I'm home.

Thanks for your help!

helfi92:
My pleasure!

rbrishabh:
Ahh!

So ```@material-ui/pickers``` reqiures ```@material-ui/core``` v4 and taskcluster is currently running     ```"@material-ui/core": "^3.9.3",``` [v3](https://github.com/taskcluster/taskcluster/blob/4b229726074684afe69a03ce597c0338734a85fe/ui/package.json#L20).

[Note that this package reqiures @material-ui/core v4. It will not work with the old v3](https://github.com/mui-org/material-ui-pickers#installation)

So @material-ui/core needs to be upgraded as well.

Should I update that as well? I recon that might require changing a lot of other things that might be deprecated in v4 but is being used in taskcluster UI..

ui/src/components/DatePicker/index.jsx
```diff
@@ -15,6 +15,7 @@ import { date } from '../../utils/prop-types';
     '& > div': {
       alignItems: 'center',
     },
+    color: 'black',
```
helfi92: For colors to match the theme, you can use the color palette that we have. See https://github.com/taskcluster/taskcluster/search?q=theme.palette&unscoped_q=theme.palette.

rbrishabh: Alright. That's understood.

But changing the color property from here isn't really helping with fixing this issue. Right? 

Although, I now have the knowledge about theme colors for future use! So thanks!

rbrishabh:
I will update the package to latest version, okay.

Let's see if that fixes the issue!

rbrishabh:
Whoops! I don't know what happened there! Seems like the whole files changed when I only changed a few lines.. let me try and correct it

PR #1219, opened 2019-08-23
Bug 1530517 - Part 2 - build the UI as part of docker image building

djmitche:
Bugzilla Bug: [1530517](https://bugzilla.mozilla.org/show_bug.cgi?id=1530517)

I found in my testing that `yarn start` would always use the config from the *last* run of `yarn start`.  Switching the order seemed to help (0d7ce96).

@imbstack I'm curious for your feedback on the Dockerfile changes.  I know you ordered things in there carefully, and I'm not sure I understood well enough to do the right thing.

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -19,7 +19,7 @@ http://localhost:3050, but otherwise need not be publicly accessible. The
 development server for this repo will proxy requests as necessary to
 http://localhost:3050.
 
-## Environment
+## Development
 
 To get started local development, just:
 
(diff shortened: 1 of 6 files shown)


ui/README.md
```diff
@@ -28,21 +28,42 @@ To get started local development, just:
...
+export APPLICATION_NAME="My UI"
+yarn start
+```
+
+_IMPORTANT_: All configuration for the UI is passed to the browser of every user.
+Do not send secret data!
+
+### Implementation
+
+The frontend reads configuraiton vairables from relative URL `./static/env.js` on page load.
```
helfi92: `s/vairables/variables/`

ui/README.md
```diff
@@ -28,21 +28,42 @@ To get started local development, just:
...
+This file is written during `yarn start`, based on environment variables.
+In a real deployment, this file is writen during the docker-container startup process, based on deployment configuration.
```
helfi92: `s/writen/written/`

ui/generate-env-js.js
```diff
@@ -0,0 +1,43 @@
+const { dirname } = require('path');
+const fs = require('fs');
+
+const ENV_DEFAULTS = {
```
imbstack: It would be neat if this or something like this was tied into `generators/k8s.js` so that we don't hardcode ui config options  in there.

imbstack: Not blocking landing on this, feel free to open a bug an assign to me to make this happen.

djmitche:
> How long does yarn build take after this change?

It adds about 2 minutes for the webpack build on my t2.xlarge + T2-Unlimited instance, plus a few more seconds to rm `ui/src` and `ui/node_modules`.

ui/generate-env-js.js
djmitche: Yeah, I think that's a good next step.  Note, though, that `cd ui; yarn start` requires this information as well.

djmitche: https://bugzilla.mozilla.org/show_bug.cgi?id=1576249

PR #1222, opened 2019-08-23
AWS Provider

owlishDeveloper:
Bugzilla Bug: [1518507](https://bugzilla.mozilla.org/show_bug.cgi?id=1518507)

100% review, please. Please give me feedback on tests because I am proposing to do tests here differently:
* I use `sinon` library to stub things instead of passing in fake aws. The main reason why I resorted to `sinon` is because I actually started on the FakeAws thingy, but I didn't like how it added ternaries to the code. I thought: what if I just stub them 🔪 Let me know if you think it's legit way to test or do you think I'm making a mess for nothing - I suspect it might actually be better to have more ternaries rather than having non-uniform code
* I test the whole `registerWorker` functionality in the api test suite, as opposed to testing provider's `registerWorker` method separately in the provider test-suite. Again, let me know if you think it's not a good idea

More tests are coming. My current plan is:
* To add testcases for `provision()`. My idea for testing that functionality is to assert on the `launchConfig` that we pass into the `runInstances`. Currently I also have some assertions on the contents of our db - not sure how important that is, considering I assert on the `launchConfig`, because I seem to be testing Azure Entities library there, which I trust is already well covered). Let me know your thoughts
* Once I define `chooseConfig`, that will get covered in `provider-aws-test`.
* Add test cases for `registerWorker` - to the `registerWorker` api endpoint tests suite
* Add tests for `checkWorker` functionality - again, to assert that we don't check up on stopped instances; to assert on the parameters that `ec2.describeInstances` is getting; to test the switch at the end.

Let me know if I am missing anything or intend to do something unnecessary!


services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,306 @@
...
+                Value: `wm-${this.providerId}`,
+              }, {
+                Key: 'Owner',
+                Value: workerPool.owner,
+              }],
+          },
+        ],
+      }).promise();
+    } catch (e) {
+      this.monitor.err(`🚨 Error calling AWS API: ${e}`);
```
djmitche: Let's leave emoji out of the official error messages :)

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,306 @@
...
+    if (!workerPool.providerData[this.providerId] || workerPool.providerData[this.providerId].running === undefined) {
+      await workerPool.modify(wt => {
+        wt.providerData[this.providerId] = wt.providerData[this.providerId] || {};
+        wt.providerData[this.providerId].running = wt.providerData[this.providerId].running || 0;
+      });
+    }
+
+    const config = this.chooseConfig({possibleConfigs: workerPool.config.launchConfigs});
+
+    aws.config.update({region: config.region});
```
djmitche: This method can be called in parallel for different workerPools at the same time.  It looks like this call updates the AWS module's global region setting.  I suspect that will cause a problem if different workerPools choose configs in different regions at the same time.  Is there a way to pass the region to the `aws.EC2` constructor?

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,306 @@
...
+    try {
+      spawned = await ec2.runInstances({
+        ...config.launchConfig,
+
+        // please make sure this string is no more than 1024 chars long
+        // https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-add-user-data
+        UserData: JSON.stringify(userData).toString('base64'),
+
+        MaxCount: config.launchConfig.MaxCount ? Math.max(toSpawn, config.launchConfig.MaxCount) : toSpawn,
+        MinCount: config.launchConfig.MinCount ? Math.min(toSpawn, config.launchConfig.MinCount) : toSpawn,
```
djmitche: I think this would mean that in cases where a user sets `MinCount` to, say, 5, it will spawn 5 new instances on *every* provisioning loop.  Am I mis-reading?

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,306 @@
...
+      credentials: this.providerConfig.credentials,
+    });
+
+    const toSpawn = await this.estimator.simple({
+      workerPoolId,
+      minCapacity: config.launchConfig.MinCount,
+      maxCapacity: config.launchConfig.MaxCount,
+      capacityPerInstance: config.capacityPerInstance,
+      running: workerPool.providerData[this.providerId].running,
+    });
```
djmitche: If `toSpawn` is zero should this exit here?

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,306 @@
...
+  /**
+   * This method checks instance identity document authenticity
+   * If it's authentic it checks whether the data in it corresponds to the worker
+   *
+   * @param worker string
+   * @param workerPool string
+   * @param workerIdentityProof {document: string, signature: string}
+   * @returns {Promise<{expires: *}>}
+   */
+  async registerWorker({worker, workerPool, workerIdentityProof}) {
```
djmitche: From `docs/providers.md`: "Providers that wish to limit registration to once per worker should return an error message from this function if the worker is already RUNNING".  We definitely wish to do so with this provider!

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,306 @@
...
+   */
+  verifyInstanceIdentityDocument({document, signature}) {
+    const verifier = crypto.createVerify('sha256');
+
+    verifier.update(document);
+    let result;
+
+    result = verifier.verify(this.ec2iid_RSA_key.toString(), signature.toString(), 'base64');
+
+    return result;
```
djmitche: Could be simplified to just `return verifier.verify(..)`.

services/worker-manager/src/providers/aws.js
```diff
@@ -0,0 +1,306 @@
...
+    const {document, signature} = workerIdentityProof;
+    if (!document || !signature) {
+      throw new ApiError('Token validation error');
+    }
+
+    if (!this.verifyInstanceIdentityDocument({document, signature})) {
+      throw new ApiError('Instance identity document validation error');
+    }
+
+    this.verifyWorkerInstance({document, worker});
```
djmitche: For consistency with `verifyInstanceIdentityDocument`, it might be good to have this also return a boolean that is checked here.  I got confused for a while during review, thinking this was ignoring a return value!

services/worker-manager/src/worker-scanner.js
```diff
@@ -65,6 +65,7 @@ class WorkerScanner {
         seen(worker.providerId, worker.workerPoolId);
         const provider = this.providers.get(worker.providerId);
         if (provider) {
+          console.log('GOT PROVIDER. CHECKING...');
```
djmitche: Debugging..

services/worker-manager/src/providers/aws.js
owlishDeveloper: lol ok

services/worker-manager/src/providers/aws.js
owlishDeveloper: oh wow, that's a good point. Lemme find out

services/worker-manager/src/providers/aws.js
owlishDeveloper: 🤔 that's how I was imagining it was supposed to work..... are you saying it actually should rather look at the demand (`toSpawn`)?

services/worker-manager/src/providers/aws.js
owlishDeveloper: according to my improved understanding - yes!

services/worker-manager/src/providers/aws.js
djmitche: I think the min and max the user specifies are *overall* minimum and maximum number running at any time, counting things that have already been started.  So yes, related to `toSpawn`.

services/worker-manager/src/providers/aws.js
owlishDeveloper: oh that var was for debugging. thank you for catching that!

services/worker-manager/src/worker-scanner.js
owlishDeveloper: 🎣

services/worker-manager/src/providers/aws.js
owlishDeveloper: Yes, that's possible! I'll do that, thanks!

PR #1223, opened 2019-08-23
Docs, tests, fixes, and refactors for web-server login strategies

### LINKED ISSUE #778: [web-server] Add tests to the scanner process

djmitche:
These commits can be considered individually.  Mostly, adding docs and tests, and doing some refactors to simplify things.  I did find a few minor bugs:

 1. We should disable clients when the identity provider says the user is gone.  This was worthy of a changelog entry.
 1. The `github` strategy was returning `undefined` from `userFromIdentity` whenever GitHub returned an error, which would cause the scanner to disable all GitHub users if it happened to run during a GitHub downtime (and those happen a lot!)
 1. The `mozilla-auth0` strategy was not checking the things it appends to identity strings, such as emails or GitHub usernames, allowing users to continue using an old identity even after changing their GitHub username.

Fixes #778.


services/web-server/src/login/strategies/mozilla-auth0.js
```diff
@@ -88,15 +87,23 @@ module.exports = class MozillaAuth0 {
...
     if (encodedUserId.startsWith('github|') || encodedUserId.startsWith('oauth2|firefoxaccounts|')) {
       encodedUserId = encodedUserId.replace(/\|[^|]*$/, '');
     }
 
     const userId = decode(encodedUserId);
-    return this.getUser({ userId });
+    const user = await this.getUser({ userId });
+
+    // catch cases where the calculated identity differs, such as when we add a
+    // GitHub username, and return no user in that case.
```
imbstack: I don't particularly understand what this means.

djmitche: See https://github.com/taskcluster/taskcluster/blob/8be8f81d521d514548b1225fcd59babdeb3c5be6/services/web-server/docs/login-strategies.md#identities

imbstack: yeah but is `userId` not the numeric id? Why do we return no user at all?

I probably just need to read this more in depth.

djmitche: It is.  The decision here is that if you changed your username to `bstack`, your identity would go from `mozilla-auth0/github|127521|imbstack` to `mozilla-auth0/github|127521|bstack`, and that's a different identity.  So at that point `mozilla-auth0/github|127521|imbstack` is no longer a valid identity, and we return no user for it.

Issue #1148, opened 2019-08-03
[UI] Allow pasting an full index to find tasks.


tomprince:
Most ways of interacting with the index don't treat the last component specially, so all components are separated by `.`. It would be nice to paste that form of index both into the UI, and into a URL, and be able to see the task. Currently, you need to edit the index to replace the trailing `.` with `/` to be able to see the index.

helfi92:
Perhaps we could add a search box that takes in a namespace name (e.g., buildbot.branches.mozilla-esr52.linux64-artifact). Would that help?

tomprince:
I often have a full index path, not just the namespace, and would like to be able to use this directly. This is particularly important for namespaces like https://taskcluster-ui.herokuapp.com/tasks/index/gecko.v2.try.revision that have hundreds or thousands of entries in effectively random order.

helfi92:
> I often have a full index path, not just the namespace, and would like to be able to use this directly.

You would be able to use the full index path, for example, `gecko.v2.try.revision.0001b4f6ab73f02e12bc7594ed840bb6789357a7.firefox.decision` in the search box after the search component gets implemented.

arku:
@helfi92 I would like to attempt building this feature.

helfi92:
👋 @arku. Go for it :)

arku:
My bad, totally forgot this task 🙈

Will make some progress tomorrow.

arku:
@helfi92 I am occupied with other stuff that I am unable to address your comments in #1238

Do you mind reassigning this issue to someone to make sure it gets fixed? 

Thanks!

helfi92:
No worries. Thanks for letting us know @arku :)

@Rolikasi are you interested in finishing up this PR?

Rolikasi:
Sure! I want to work on it)

arku:
@Rolikasi Thanks for taking it up!

PR #1238, opened 2019-08-26
Add search to tasks index page

arku:
Implements the feature requested in https://github.com/taskcluster/taskcluster/issues/1148

### DIFF
--- a/ui/src/views/Tasks/TaskIndex/IndexedTask/index.jsx
+++ b/ui/src/views/Tasks/TaskIndex/IndexedTask/index.jsx
@@ -5,6 +5,7 @@ import dotProp from 'dot-prop-immutable';
 import Spinner from '@mozilla-frontend-infra/components/Spinner';
 import Dashboard from '../../../../components/Dashboard';
 import HelpView from '../../../../components/HelpView';
+import Search from '../../../../components/Search';
 import IndexedEntry from '../../../../components/IndexedEntry';
 import { ARTIFACTS_PAGE_SIZE } from '../../../../utils/constants';
 import ErrorPanel from '../../../../components/ErrorPanel';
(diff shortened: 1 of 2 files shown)


ui/src/views/Tasks/TaskIndex/IndexedTask/index.jsx
```diff
@@ -90,6 +91,21 @@ export default class IndexedTask extends Component {
...
 
+  handleTaskNamespaceSearchSubmit = taskNamespaceSearch => {
+    const searchText = taskNamespaceSearch.replace(/\s+/g, '');
+    const namespace = searchText
+      .split('.')
+      .slice(0, -1)
+      .join('.');
+    const taskFromNamespace = searchText.split('.').slice(-1)[0];
+
+    if (namespace && taskFromNamespace) {
```
arku: @helfi92 Is this fine?

Here are the possible ways someone might interact with the search bar and the obtained results.

**No task in the namespace**

If a user just searches for 'gecko', nothing would happen as there is no task identified from the namespace.

**Namespace and task present**

If a user searches for `ci.cache.level-3.docker-images.v2.hg-push.latest`, they will be taken to `/tasks/index/ci.cache.level-3.docker-images.v2.hg-push/latest` which will render the `IndexedTask` component.

But if search for `ci.cache.level-3.docker-images.v2.hg-push`, they will be taken to `/tasks/index/ci.cache.level-3.docker-images.v2/hg-push` which won't work and an error will be rendered.

Should we worry about this? If yes, how do we handle this scenario?

ui/src/views/Tasks/TaskIndex/ListNamespaces/index.jsx
```diff
@@ -138,7 +154,14 @@ export default class ListNamespaces extends Component {
     return (
       <Dashboard
         title="Index Browser"
-        helpView={<HelpView description={description} />}>
+        helpView={<HelpView description={description} />}
+        search={
```
arku: @helfi92  If a user searches for a task that doesn't exist, the `IndexedTask` component will be rendered with the error message. If they want to retry (search again), they have go back to the Index page again. So, I decided to add the search bar here as well.

ui/src/views/Tasks/TaskIndex/IndexedTask/index.jsx
helfi92: No worries!

helfi92:
@arku Let me know when this is ready for review :)

arku:
@helfi92 Will do.

PR #1253, opened 2019-08-27
Allow configuration of providerId based on active providers

### LINKED ISSUE #1249: [worker-manager] Allow selection of providers from the configured providers, and provide a default configuration based on providerType.

djmitche:
Fixes #1249.

That this will give an error after 1000 providers is probably just fine -- but at least, when some deployment gets to that point, they'll get a clear error.  I suspect the fix will be to default to an open text-entry box after, say, 100 providers.


helfi92:
Navigating to `http://localhost:5080/worker-manager/create` shows a blank page. It's probably requiring both graphql queries to have a `name`.

djmitche:
Hm, it works in my testing.  Is there an error in network tools?

helfi92:
I am getting a `TypeError: "data is undefined"`

djmitche:
Huh, that sounds like an issue from lint fixes.  Maybe it didn't reload correctly for me during the lint fixes.  I'll give it another test.

djmitche:
Ugh, and of course that error crashes my browser since React bizarrely produces it in a tight loop.

djmitche:
Ah!  Lint made me remove parens in `!loading && (thing ? thing : thing)` but that changed the meaning!  *shakes fist at lint*

djmitche:
OK, I *think* that both passes prettier's checks and works.

djmitche:
Failed on push -- https://tools.taskcluster.net/groups/Dmsg_5bMQRqz1WDEn2tR3A/tasks/Rg2HRimDQZmGJp-dt2TWxQ/runs/0/logs/public%2Flogs%2Flive.log

I'll take a look.

PR #1260, opened 2019-08-29
Google provider cleanup part 1

imbstack:
This removes the ability for google provider to set up its own service accounts for workers. This both simplifies the code significantly and makes worker-manager much safer to run! Couple of other QOL improvements here as well. Part 2 should be coming soon where we clean up a few last TODOs and add some rate limiting logic.

### DIFF
--- a/services/worker-manager/src/providers/google.js
+++ b/services/worker-manager/src/providers/google.js
@@ -1,3 +1,4 @@
+const assert = require('assert');
 const slugid = require('slugid');
 const _ = require('lodash');
 const taskcluster = require('taskcluster-client');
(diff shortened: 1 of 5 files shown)


services/worker-manager/src/providers/google.js
```diff
@@ -14,39 +14,42 @@ class GoogleProvider extends Provider {
     ...conf
   }) {
     super(conf);
-    let {project, instancePermissions, creds, credsFile} = providerConfig;
+    let {project, creds, workerServiceAccountId} = providerConfig;
     this.configSchema = 'config-google';
 
-    this.instancePermissions = instancePermissions;
+    assert(project, 'Must provide a project to google providers');
```
owlishDeveloper: excellent! I should add some assertions like this to aws provider as well

PR #1261, opened 2019-08-30
Interactive connect CSP

helfi92:
The interactive connect feature doesn't work on Firefox & Chrome but works on Safari. I'm not sure why but probably because safari is more lenient.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src#Syntax

This is a similar fix as https://github.com/taskcluster/taskcluster-tools/pull/607.


helfi92:
Haha, that made me laugh :)

PR #1265, opened 2019-08-30
Bug 1577842 - document aws provider

djmitche:
Bugzilla Bug: [1577842](https://bugzilla.mozilla.org/show_bug.cgi?id=1577842)


djmitche:
There's always followups :)

PR #1275, opened 2019-09-03
Fix - Can't copy/paste labels from task

### LINKED ISSUE #1256: Can't copy/paste label from task
**Describe the bug**
The label and other gray pills for a bug can't be copy/pasted.

image

In general, copy/paste doesn't work for a lot of parts of the UI, which makes debugging hard since I have to carefully retype things or find them in the DOM inspector and copy from there.  If it's possible to globally turn off whatever the UI is doing to prevent the browser from allowing copy/paste, that would be great.

**To Reproduce**
1. Go to https://taskcluster-ui.herokuapp.com/tasks/SDy5lkIuQiuKRxBVw-xnTA

**Expected behavior**
Every piece of text onscreen can be selected and copied

rbrishabh:
Fixes #1256
Hey @helfi92 ! I hope you are doing great :D

So, I created a ChipComponent, and used it over in ```src/views/Tasks/ViewTask```, and it does seem to fix the issue. 

But there are a few things I would like to mention! 
Let me know where to proceed with this!

Thanks!

### DIFF
--- a/ui/src/components/ChipComponent/index.jsx
+++ b/ui/src/components/ChipComponent/index.jsx
@@ -1,0 +1,29 @@
+import React, { Component } from 'react';
+import Chip from '@material-ui/core/Chip';
+import { withStyles } from '@material-ui/core/styles';
+
+@withStyles(() => ({
+  chipLabel: {
+    whiteSpace: 'nowrap',
+    userSelect: 'text',
+    cursor: 'inherit',
+  },
+  chipRoot: {
+    whiteSpace: 'nowrap',
+    userSelect: 'text',
+    cursor: 'inherit',
+  },
+}))
+export default class ChipComponent extends Component {
+  render() {
+    const { classes } = this.props;
+
+    return (
+      <Chip
+        classes={{ label: classes.chipLabel, root: classes.chipRoot }}
+        label={this.props.label}
+        className={this.props.className ? this.props.className : null}
+      />
+    );
+  }
+}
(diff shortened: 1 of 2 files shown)


ui/src/components/ChipComponent/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+  },
+}))
+export default class ChipComponent extends Component {
+  render() {
+    const { classes } = this.props;
+
+    return (
+      <Chip
+        classes={{ label: classes.chipLabel, root: classes.chipRoot }}
+        label={this.props.label}
```
rbrishabh: If I just use ```{...this.props}``` here, I am still not able to copy paste.
So instead, I had to add ```label={this.props.label}``` and ```className={this.props.className}``` after which, I believe, things worked just fine. (Except for a lot of errors inside the console) @helfi92

ui/src/components/ChipComponent/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+        className={this.props.className ? this.props.className : null}
```
rbrishabh: Because of the comment I just wrote up, I couldn't simply replace ChipComponent everywhere. 
One way to export this every where could be by using ternary operators. 
I can create ternary for all possible props for Chips used through the Taskcluster UI (there aren't many - so this option is not too unfeasible)

Other than that, I'm always open to suggestions from you! (Thanks as always ;) )

PR #1285, opened 2019-09-04
upgrade apollo-server-express, graphql, and graphql-validation-complexity

djmitche:
I don't know why renovate hasn't gotten around to this..


djmitche:
I admit I totally chose random things to upgrade, just to see if there was some reason they couldn't be upgraded that would cause renovate to skip them.  Apparently not.

PR #1288, opened 2019-09-05
Upgrade depcheck and eslint

djmitche:
Renovate was skipping eslint due to #894.  However, I can't figure out why it was skipping depcheck.  https://github.com/renovatebot/config-help/issues/350


djmitche:
Lints to fix:

https://tools.taskcluster.net/groups/IGTZJDS3QNuL4VS_d0Q75Q/tasks/DsKKh6-KTEiLuHUW1189ZA/runs/0/logs/public%2Flogs%2Flive_backing.log

services/queue/src/workerinfo.js
```diff
@@ -238,13 +238,9 @@ class WorkerInfo {
...
-      try {
-        result = await worker.modify((entity) => {
-          entity.expires = new Date(expires || entity.expires);
-        });
-      } catch (err) {
-        throw err;
-      }
+      result = await worker.modify((entity) => {
+        entity.expires = new Date(expires || entity.expires);
+      });
```
helfi92: Would this throw is `.modify` fails?

libraries/monitor/src/monitor.js
```diff
@@ -245,7 +245,7 @@ class Monitor {
    *
    */
   reportError(err, level = 'err', extra = {}) {
-    if (err.hasOwnProperty && !(err.hasOwnProperty('stack') || err.hasOwnProperty('message'))) {
+    if (!(Object.prototype.hasOwnProperty.call(err, 'stack') || Object.prototype.hasOwnProperty.call(err, 'message'))) {
```
helfi92: Did you mean to make use of `in` instead of `hasOwnProperty` here? In any case, this looks good to me.

djmitche: This is investigating an "unknown" object that might be a class instance, so I stuck with `hasOwnProperty` to avoid capturing any prototype methods.  The other cases where I used `in` are looking at regular key/value "objects" where that's not likely to be an issue.

services/queue/src/workerinfo.js
djmitche: Yep, it's just a de-dented version of the original code, with the `try { .. } catch (err) { throw err; }` removed.

PR #1292, opened 2019-09-05
Use renovate rangeStrategy: update-lockfile

djmitche:
Per https://github.com/renovatebot/config-help/issues/350 this should
get us updates for packages with versions that fall within the current
semver range.

### DIFF
--- a/package.json
+++ b/package.json
@@ -184,12 +184,12 @@
   },
   "renovate": {
     "extends": [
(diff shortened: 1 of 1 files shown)


djmitche:
That's https://bugzilla.mozilla.org/show_bug.cgi?id=1579085 - I'm working on it.

PR #1293, opened 2019-09-05
Change workflow in reportError

owlishDeveloper:
Bugzilla Bug: [1578902](https://bugzilla.mozilla.org/show_bug.cgi?id=1578902)

### DIFF
--- a/services/worker-manager/src/data.js
+++ b/services/worker-manager/src/data.js
@@ -128,29 +128,29 @@ WorkerPool.prototype.reportError = async function({kind, title, description, ext
       let extraInfo = '';
       if (Object.keys(extra).length) {
         extraInfo = `
-            It includes the extra information:
-            
-            \`\`\`
-            ${yaml.safeDump(extra)}
-            \`\`\`
-                  `.trim();
+It includes the extra information:
+
+\`\`\`
+${yaml.safeDump(extra)}
+\`\`\`
+      `.trim();
       }
 
       await this.notify.email({
         address: this.owner,
         subject: `Taskcluster Worker Manager Error: ${title}`,
         content: `
-            Worker Manager has encountered an error while trying to provision the worker pool ${this.workerPoolId}:
-            
-            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-            
-            ${description}
-            
-            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-            
-            ErrorId: ${errorId}
-            
-            ${extraInfo}
+Worker Manager has encountered an error while trying to provision the worker pool ${this.workerPoolId}:
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+${description}
+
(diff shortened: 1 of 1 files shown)


services/worker-manager/src/data.js
```diff
@@ -123,54 +123,59 @@ WorkerPool.expire = async function(monitor) {
...
+  try {
+    if (this.emailOnError) {
+      let extraInfo = '';
+      if (Object.keys(extra).length) {
+        extraInfo = `
+            It includes the extra information:
+            
+            \`\`\`
+            ${yaml.safeDump(extra)}
+            \`\`\`
```
djmitche: It looks like this got indented?

services/worker-manager/src/data.js
```diff
@@ -123,54 +123,59 @@ WorkerPool.expire = async function(monitor) {
...
+            
+            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+            
+            ${description}
+            
+            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+            
+            ErrorId: ${errorId}
+            
+            ${extraInfo}
```
djmitche: Same here..

services/worker-manager/src/data.js
```diff
@@ -123,54 +123,59 @@ WorkerPool.expire = async function(monitor) {
...
+      `.trim(),
+      });
+    }
 
-ErrorId: ${errorId}
+    await this.monitor.log.workerError({
```
djmitche: This should probably be in the `finally` block, too.

services/worker-manager/src/data.js
owlishDeveloper: oh yeah, I did that, it was hard to read. I am wondering if I can write a tiny function that gets in those variables and returns the string, and just replace the template literal with that function call?

services/worker-manager/src/data.js
owlishDeveloper: I want only one function call to be in the `finally`, the most important one (which, I judge, is the `workerPoolError` thing because it's displayed in the UI. If there are two functions in the `finally`, and the first errors out, the second won't get executed.

owlishDeveloper: Another way to go about this can be orchestrating a series of try-catch blocks, one per function call, if we'd rather consider all three to be equally important

djmitche: I was thinking of just running all three things in parallel - they're all important, after all.

services/worker-manager/src/data.js
djmitche: Yeah, that would be nice.  @imbstack also wrote an un-indenting function, but I've forgotten what it was called.

services/worker-manager/src/data.js
owlishDeveloper: I'm going to merge this as is for now. I'd like to use `Promise.allSettled` here to make sure the first two functions work independently in parallel, and then the third function is executed in `.then()` or `.finally()` of that. I need Node 12 for all this awesomeness tho

PR #1298, opened 2019-09-05
Bug 1579085 - clean up azure containers, too

djmitche:
Bugzilla Bug: [1579085](https://bugzilla.mozilla.org/show_bug.cgi?id=1579085)


djmitche:
Yes, it's been having Azure timeouts a lot recently..

PR #1338, opened 2019-09-10
Fix static provider gotchas

imbstack:
This makes the code match what the schemas require

### DIFF
--- a/services/worker-manager/src/api.js
+++ b/services/worker-manager/src/api.js
@@ -403,6 +403,13 @@ builder.declare({
   return res.reply(data.serializable());
 });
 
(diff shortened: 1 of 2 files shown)


services/worker-manager/src/api.js
```diff
@@ -403,6 +403,13 @@ builder.declare({
   return res.reply(data.serializable());
 });
 
+let cleanCreatePayload = payload => {
+  if (payload.providerInfo && payload.providerInfo.staticSecret) {
+    payload.providerInfo.staticSecret = '(OMITTED)';
+  }
+  return payload;
+};
+
```
djmitche: good call!

PR #1349, opened 2019-09-11
[UI] Fix cookies in production

helfi92:
Deployed in https://hassan.taskcluster-dev.net. You can try logging in with GitHub. auth0 is not configured in web-server so that won't work.

Fixes https://github.com/taskcluster/taskcluster/issues/1099.

### DIFF
--- a/services/web-server/src/servers/createApp.js
+++ b/services/web-server/src/servers/createApp.js
@@ -12,6 +12,8 @@ const credentials = require('./credentials');
 module.exports = async ({ cfg, strategies }) => {
   const app = express();
 
+  // trust first proxy
+  app.set('trust proxy', 1);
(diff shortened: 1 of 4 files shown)


services/web-server/src/servers/createApp.js
```diff
@@ -12,6 +12,8 @@ const credentials = require('./credentials');
 module.exports = async ({ cfg, strategies }) => {
   const app = express();
 
+  // trust first proxy
+  app.set('trust proxy', 1);
```
djmitche: It looks like this unconditionally trusts the proxy?  Did you mean to use `cfg.trustProxy` here?

helfi92: Indeed. Just pushed a fix :)

PR #1351, opened 2019-09-11
Revert "Bug 1580483 - skip the worst offenders of bug 1579496"

imbstack:
In #1350 we tried a fix for these issues. Enabling these again to catch
any issues sooner

Bugzilla Bug: [1579496](https://bugzilla.mozilla.org/show_bug.cgi?id=1579496)


djmitche:
xxxxx

PR #1352, opened 2019-09-11
Revert "[UI] Fix cookies in production"

helfi92:
Reverts taskcluster/taskcluster#1349


helfi92:
Navigating to https://taskcluster-ui.herokuapp.com/provisioners/ showed:

Screen Shot 2019-09-11 at 4 43 04 PM

helfi92:
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials#What_went_wrong explains what went wrong.

I'm not sure why `Access-Control-Allow-Origin` is improperly set to `*` for the POST request but properly set for the OPTIONS request:

OPTIONS:

Screen Shot 2019-09-11 at 5 03 14 PM

POST:

Screen Shot 2019-09-11 at 5 03 22 PM


We set that header in:

https://github.com/taskcluster/taskcluster/blob/ee4ca085ab7c2ae5836621128fe5bd7472d1159d/services/web-server/src/servers/createApp.js#L23

helfi92:
According to https://expressjs.com/en/resources/middleware/cors.html, the default configuration for `cors` is

```json
{
  "origin": "*",
  "methods": "GET,HEAD,PUT,PATCH,POST,DELETE",
  "preflightContinue": false,
  "optionsSuccessStatus": 204
}
```

We overwrite the `origin`key but the `methods` should still accept `POST`...

helfi92:
Wow, it turns out apollo-server-express is overwriting that value. We need to disable `cors` there and rely on the `cors` package in `createApp.js` instead.

PR #1354, opened 2019-09-12
Bug 1580512 - use replace to avoid the need for gopkg.in

djmitche:
This basically replicates the role of gopkg.in, which seems to be having load issues (and seems mostly unmaintained, honestly).  This is going to be a PITA as things get updated, since renovate will not know to update the `replace` lines, but it will at least get our CI working again until either gopkg.in is fixed or useful packages like gopkg.in/yaml move to more stable hosting.

See https://github.com/niemeyer/gopkg/issues/67 for the gopkg.in issue.

Bugzilla Bug: [1580512](https://bugzilla.mozilla.org/show_bug.cgi?id=1580512)

### DIFF
--- a/clients/client-go/go.mod
+++ b/clients/client-go/go.mod
@@ -19,3 +19,8 @@ require (
 	golang.org/x/tools v0.0.0-20190722020823-e377ae9d6386
 	gopkg.in/yaml.v2 v2.2.2 // indirect
 )
+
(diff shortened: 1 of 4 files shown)


djmitche:
I ran this in a loop for 6 hours (oops, sorry github) with no errors.

petemoore:
Awesome, that seems to do the job, thanks for deep diving into the problem and fixing it!

PR #1365, opened 2019-09-16
Bug 1574279 - Assign roles to Github logins

helfi92:


services/web-server/src/login/clients/GithubClient.js
```diff
@@ -1,18 +1,50 @@
...
       debug(`profile for user ${username} not found`);
     }
 
     return body;
   }
+
+  async orgsFromUsername(username) {
+    const { body } = await request
+      .get(`${baseUrl}/users/${username}/orgs`)
+      .set('Authorization', `Bearer ${this.accessToken}`);
```
djmitche: This endpoint only seems to return public org memberships, even with an access token for which I've checked all of the permissions boxes.  I don't know if there's a way around that, but if not we should probably document that clearly: you can only get permissions on an org with public membership.

services/web-server/src/login/clients/GithubClient.js
```diff
@@ -1,18 +1,50 @@
+const assert = require('assert');
 const request = require('superagent');
 const Debug = require('debug');
 
 const debug = Debug('GithubClient');
 const baseUrl = 'https://api.github.com';
 
 module.exports = class GithubClient {
+  constructor({ accessToken }) {
+    assert(accessToken, 'An access token is required to access Github endpoints');
```
djmitche: This is different from the OAuth App's credentials.  So maybe we have to have *three* different sets of GitHub credentials?

services/web-server/src/login/strategies/github.js
```diff
@@ -46,11 +46,35 @@ module.exports = class Github {
...
+    const roles = [].concat(
+      ...(
+        await Promise.all(orgs.map(async ({ login: org }) => {
+          const [reposErr, repos] = await tryCatch(this.githubClient.reposFromOrg(org));
+
+          if (reposErr) {
+            throw reposErr;
+          }
+
+          return repos.map(repo => `github-group:${org}/${repo.name}`);
```
djmitche: If I understand correctly, this will end up issuing roles for every repo in any organization of which the user is a public member.  There are a few issues with that:
 * It's only public org membership
 * It will create scopes for repos to which the user doesn't have write access
 * Many users will be in orgs with hundreds of repos, so the resulting credentials will have hundreds of roles and not fit in an HTTP header

helfi92: > It's only public org membership

Good observation. We should look into adding private ones as well.

> It will create scopes for repos to which the user doesn't have write access

We should fix that :)

> * Many users will be in orgs with hundreds of repos, so the resulting credentials will have hundreds of roles and not fit in an HTTP header

The size of the request is currently set to 128kb but I think we could just increase that to something like 256kb. Let me know if you think that's not the right approach.

https://github.com/taskcluster/taskcluster/blob/b34aacc27c4e96565b598abf7c638e8fa3ce5530/ui/web-ui-nginx-site.conf#L32

djmitche: That nginx config is only for the UI, which only handles GET requests for UI assets.  And https://stackoverflow.com/questions/3326210/can-http-headers-be-too-big-for-browsers suggests Firefox's limit is 10K.

helfi92: You're right. I tested this locally and found that FF currently has it at around 12kb.

helfi92:
> Many users will be in orgs with hundreds of repos, so the resulting credentials will have hundreds of roles and not fit in an HTTP header

Do you have any ideas on that? I can't think of something sensible. I wonder if it would make sense to revisit once we have real users hitting this problem. Most browsers allow header sizes > 250kb. It's unfortunate this issue is only relevant for FF.

djmitche:
> If I recall, the plan here was to get team membership for orgs to which the OAUth app has been authorized. Then users can use GH teams to manage access, creating for example a "ci-devs" group in the "octotools" repo and then adding some scopes to role github-team:octotools/ci-devs.

I think that's the right approach.  I don't think we should do something that we know is wrong and won't work, and then force our users to change later.

services/web-server/src/login/clients/GithubClient.js
helfi92: I think this is just the oauth access token we get after the user login with Github.

services/web-server/src/login/clients/GithubClient.js
helfi92: I've switched to using `user/memberships/orgs` which seems to return private org memberships.

helfi92:
Ready for review.

services/web-server/src/login/strategies/github.js
```diff
@@ -82,8 +105,10 @@ module.exports = class Github {
           clientID: strategyCfg.clientId,
           clientSecret: strategyCfg.clientSecret,
           callbackURL: `${cfg.app.publicUrl}${callback}`,
+          scope: 'repo',
         },
         async (accessToken, refreshToken, profile, next) => {
+          this.githubClient = new GithubClient({ accessToken });
```
djmitche: This is setting a global variable (global to the strategy) based on credentials for a single user.  Since these are async operations that could overlap, that could lead to one user getting another user's credentials!  Let's not do that :)

This leads to a bigger problem: if we need the accessToken from the OAuth transaction to get the user's team memberships, then user scanning won't work.  It's cleverly hidden here by the global variable, but when the scanner runs, it will fail with something like `this.githubClient.listTeams is undefined`.

Solving that might be a bit of a puzzle.  @owlishDeveloper might be able to help!

helfi92: Those are valid points. We could use a `Map` to avoid this but I don't think this would solve the second point re: "scanner failing" since we won't have the user's oauth access token. One idea that comes to mind is to store the access token in an azure table keyed by the user's identity. This would allow the scanner to read public & private team memberships of a user and as far as I can tell, the oauth github token has no expiration. Does this make any sense? @owlishDeveloper @djmitche Did you have something else in mind?

owlishDeveloper: I will need to think about this a bit, and familiarize myself with the code. I'll get back to you asap

helfi92: Let me know if you had any questions or if there's anything that is not making sense. This whole thing is very confusing.

helfi92: owlish and I spoke on Slack and agreed to store the access token in a db.

services/web-server/src/login/strategies/github.js
```diff
@@ -22,12 +22,22 @@ module.exports = class Github {
...
-    const [githubErr, githubUser] = await tryCatch(this.githubClient.userFromUsername(username));
+    const githubEntry = await this.GithubAccessToken.load({ userId: String(userId) }, true);
+
+    if (!githubEntry) {
+      debug(`Github user id ${userId} could not be found in the database.`);
+
+      return;
+    }
+
+    const githubClient = this.clients.get(userId);
```
djmitche: What happens if this isn't set?

services/web-server/src/login/strategies/github.js
```diff
@@ -82,8 +123,15 @@ module.exports = class Github {
...
           clientSecret: strategyCfg.clientSecret,
           callbackURL: `${cfg.app.publicUrl}${callback}`,
+          scope: 'repo',
         },
         async (accessToken, refreshToken, profile, next) => {
+          this.clients.set(Number(profile.id), new GithubClient({ accessToken }));
+          await this.GithubAccessToken.create({
+            userId: profile.id,
+            accessToken,
+            expires: taskcluster.fromNow('10 minutes'),
```
djmitche: 10 minutes doesn't seem a very long time to be logged in.  How long is the access token from GH valid?

services/web-server/src/data/GithubAccessToken.js
```diff
@@ -0,0 +1,38 @@
+const assert = require('assert');
+const Entity = require('azure-entities');
+
+const GithubAccessToken = Entity.configure({
+  version: 1,
+  partitionKey: Entity.keys.ConstantKey('githubAccessTokens'),
+  rowKey: Entity.keys.StringKey('userId'),
+  signEntities: true,
```
djmitche: I suspect encryption is even more important, since these are user credentials..

services/web-server/src/login/strategies/github.js
helfi92: The token from GH never expires. You're right, 10 minutes is not a very long time. I didn't think when setting the expiration time that this will be used by the scanner process. I could change it to something like 1 week.

services/web-server/src/login/strategies/github.js
helfi92: Good catch. This is no bueno. We should make sure it's set.

services/web-server/src/login/strategies/github.js
djmitche: Oh, cool, never is good!  Then we could probably keep them in this table forever, too?

So I guess the "expiration" is when a user revokes access, at which point the access token becomes invalid and the scanner should then disable all of their clients (rather than crash).  Which makes sense: revoke access, lose credentials.

I don't think there's a reason to try to purge revoked tokens from the table -- there is only one row per user, so the table will have a fairly stable size.

services/web-server/src/data/GithubAccessToken.js
helfi92: The access token is encrypted [here](https://github.com/taskcluster/taskcluster/pull/1365/files/f8038b5297b243ef3d300639266709b6817983ec#diff-e87117fe4e535da5efe2d8d264b0be73R13). Did you have something else in mind?

services/web-server/src/login/strategies/github.js
helfi92: Correct. Also, forever sounds good to me :)

> I don't really understand how this will work. Can you outline it?

Let me know if you still don't understand how this will work. I can try to explain.

services/web-server/src/data/GithubAccessToken.js
djmitche: I didn't, I was being dumb, sorry!

services/web-server/src/login/strategies/github.js
```diff
@@ -37,6 +37,12 @@ module.exports = class Github {
     }
 
     const githubClient = this.clients.get(userId);
+
+    if (!githubClient) {
+      debug(`Github client for user id ${userId} could not be found.`);
```
djmitche: I suspect this will be common, since the entry is only added to the map when the user logs in, and it's an in-memory map.  So once web-server restarts, or if the user hits a different process, it won't exist.

services/web-server/src/login/strategies/github.js
```diff
@@ -46,11 +55,42 @@ module.exports = class Github {
...
+    }
+
+    const [userMembershipsOrgsErr, userMembershipsOrgs] = await tryCatch(githubClient.userMembershipsOrgs());
+
+    if (userMembershipsOrgsErr) {
+      throw userMembershipsOrgsErr;
+    }
+
+    const roles = [
+      ...teams.map(({ slug: team }) => `github-team:${team}`),
```
djmitche: I think this needs to be qualified as `${organization.login}/${team}`

owlishDeveloper:
> (with one change)
> 
> @ajvb can you take a look before this lands? In particular, is it OK to store the GitHub access_token we get from users on login forever? I hope so :)

Do we store them forever? From glancing very briefly at the code, it seems like we expire the records in 10 minutes?

helfi92:
> Do we store them forever? From glancing very briefly at the code, it seems like we expire the records in 10 minutes?

It turns out I forgot to remove the unused config variable. Should be good now. Good catch!

ajvb:
> (with one change)
> 
> @ajvb can you take a look before this lands? In particular, is it OK to store the GitHub access_token we get from users on login forever? I hope so :)

Yeah, that makes sense to me.

What is the purpose of `GithubAccessToken.expire` then?

helfi92:
> What is the purpose of GithubAccessToken.expire then?

It's not serving a purpose anymore. It should be removed :)

djmitche:
The code-coverage reduction is weird but unrelated to this issue.

@owlishDeveloper just your review left :D

helfi92:
> Oh, this can happen in a followup if you'd prefer, but while waiting for final review, https://docs.taskcluster.net/docs/reference/core/web-server/github-login-strategy needs updated.

Done.

PR #1370, opened 2019-09-18
Allow tc.yml to produce undefined 'tasks'

djmitche:
If a user writes:
```yaml
tasks:
  $if: ..
  then: [ .. ]
```

then in the case where the conditional is false, the resutling config
will not have a `tasks` property.  Adding `else: []` would fix that, but
let's be nice and accept `undefined` as equivalent to `[]`.

This occurred at https://github.com/servo/webrender/pull/3755#issuecomment-532653876 (I'm the one who changed the `.taskcluster.yml`, so that's why I started digging..)

### DIFF
--- a/services/github/src/tc-yaml.js
+++ b/services/github/src/tc-yaml.js
@@ -150,7 +150,7 @@ class VersionZero extends TcYaml {
 
     // Add common taskGroupId and schedulerId. taskGroupId is always the taskId of the first
     // task in taskcluster.
-    if (config.tasks.length > 0) {
+    if (config.tasks && config.tasks.length > 0) {
       let taskGroupId = config.tasks[0].taskId;
       config.tasks = config.tasks.map((task) => {
         return {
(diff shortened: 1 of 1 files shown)


services/github/src/tc-yaml.js
```diff
@@ -234,7 +234,7 @@ class VersionOne extends TcYaml {
   }
 
   compileTasks(config, cfg, payload, now) {
-    if (config.tasks.length > 0) {
+    if (config.tasks && config.tasks.length > 0) {
```
owlishDeveloper: Should it be `||`?

djmitche: Why?

owlishDeveloper: I must confess I didn’t quite understand from the description what was the bug (edited) 
It seems to me that change guards against an error which would be thrown in case config.tasks is undefined, but that’s not what you’re talking about in the PR description (or maybe I am misunderstanding)

djmitche: Sorry if the description was unclear -- that's exactly what I'm guarding against.  So we should treat the situation when `config.tasks` is `undefined` the same as `[]`.  I think this conditional is correct:
 * `undefined` -> `false && ..` -> false
 * `[]` -> true && false -> false
 * `[{some-task}]` -> true && true -> true

jrmuizel:
Review ping?

services/github/src/tc-yaml.js
owlishDeveloper: got it!

djmitche:
deployed!

djmitche:
This wasn't sufficient -- https://github.com/servo/webrender/pull/3765#issuecomment-536181495

PR #1373, opened 2019-09-19
Bug 1582376 - increase AMQP frame size

djmitche:
I tested this with 128 routes on a single message, no problem.

256 caused a buffer overflow in node, but let's pretend that didn't happen.  The current max for routes in a task is 64, and that seems a good number.

Bugzilla Bug: [1582376](https://bugzilla.mozilla.org/show_bug.cgi?id=1582376)

### DIFF
--- a/changelog/bug1582376.md
+++ b/changelog/bug1582376.md
@@ -1,0 +1,6 @@
+level: patch
+reference: bug 1582376
+---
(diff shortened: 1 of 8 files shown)


La0:
Thanks a lot Dustin for working on that. Nice reactivity :clap:

PR #1374, opened 2019-09-19
aws provider

owlishDeveloper:

### DIFF
--- a/services/worker-manager/schemas/v1/config-aws.yml
+++ b/services/worker-manager/schemas/v1/config-aws.yml
@@ -10,6 +10,12 @@ properties:
     items:
       type: object
       properties:
+        region:
+          title: Region
+          type: string
+          description: |
+            AWS region (not to be confused with AZ) to launch this config in
+            https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions
         launchConfig:
           title: Launch Configuration
           type: object
@@ -45,22 +51,12 @@ properties:
           description: Maximum capacity that should not be exceeded at any given time
       additionalProperties: false
       required:
+        - region
         - launchConfig
         - capacityPerInstance
         - minCapacity
         - maxCapacity
-  regions:
-    title: Regions
-    type: array
-    description: |
-      AWS region (not to be confused with AZ) to launch the instances in. See here:
-      https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions
-      The instances will be uniformly distributed over all of the regions set here.
-    uniqueItems: true
-    items:
-      type: string
 additionalProperties: false
 required:
   - launchConfigs
-  - regions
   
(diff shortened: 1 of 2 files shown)


services/worker-manager/schemas/v1/config-aws.yml
```diff
@@ -13,7 +13,9 @@ properties:
         region:
```
imbstack: I think region will need to move into `launchConfigs`? Or perhaps there will be a new array above this that includes both region and launch configs

services/worker-manager/src/providers/aws.js
```diff
@@ -82,83 +64,108 @@ class AwsProvider extends Provider {
...
 
-    // Make sure we don't get "The same resource type may not be specified more than once in tag specifications" errors
-    const TagSpecifications = config.launchConfig.TagSpecifications ? config.launchConfig.TagSpecifications : [];
-    const instanceTags = [];
-    const otherTagSpecs = [];
-    TagSpecifications.forEach(ts =>
-      ts.ResourceType === 'instance' ? instanceTags.concat(ts.Tags) : otherTagSpecs.push(ts)
-    );
+    const uniqueRegions = new Set(workerPool.config.launchConfigs.map(lc => lc.region));
+    let ec2s = {};
```
imbstack: why don't we make this in the constructor so that they can be cached across iteration loops? And then every time we get a new region we can try to grab the old one or make a new one if we haven't seen it before. No need to make `uniqueRegions` at that point

services/worker-manager/src/providers/aws.js
```diff
@@ -82,83 +64,108 @@ class AwsProvider extends Provider {
...
-        title: 'Instance Creation Error',
-        description: e.message,
-        notify: this.notify,
-        WorkerPoolError: this.WorkerPoolError,
+      const toSpawn = await this.estimator.simple({
+        workerPoolId,
+        minCapacity: config.minCapacity,
+        maxCapacity: config.maxCapacity,
+        capacityPerInstance: config.capacityPerInstance,
+        running: workerPool.providerData[this.providerId].running, // todo: if this is per config now, do we need a config id?
```
imbstack: We'll need to add a `pending` to this as well and we can update pending from within this loop. I'm doing that now in my branch.

services/worker-manager/src/providers/aws.js
```diff
@@ -82,83 +64,108 @@ class AwsProvider extends Provider {
...
-              {
-                Key: 'Name',
-                Value: `${workerPoolId}`,
-              }],
-          },
-        ],
-      }).promise();
-    } catch (e) {
-      this.monitor.err(`Error calling AWS API: ${e}`);
+    for await (let config of workerPool.config.launchConfigs) {
```
imbstack: Isn't this pretty much going to pick the first launchConfig every time? We probably need some randomness or something involved here.

services/worker-manager/src/providers/aws.js
owlishDeveloper: hm, pending is considered `running`, isn't it?

PR #1380, opened 2019-09-22
Bug 1577841 - display worker pool errors

djmitche:
Bugzilla Bug: [1577841](https://bugzilla.mozilla.org/show_bug.cgi?id=1577841)

I'd like to get a first-round r? from owlish to see if this corresponds with expectations, then we can get into the details of UI and implementation.


owlishDeveloper:
This looks good!

It didn't display errors from my table tho (I ran this against my cluster). Is it in accord with the master-plan?

ui/src/components/WMWorkerPoolsTable/index.jsx
```diff
@@ -184,29 +185,37 @@ export default class WorkerManagerWorkerPoolsTable extends Component {
         </TableCell>
 
         <TableCell>
+          <Button
+            className={classes.linksButton}
+            variant="outlined"
+            component={Link}
+            to={`/provisioners/${encodeURIComponent(
```
owlishDeveloper: (this is unrelated to the PR, but from git history I see this change is yours. We don't have provisioners and worker-types anymore, should they be in the url 🤔 should we create a bug to change the url?)

djmitche:
> It didn't display errors from my table tho (I ran this against my cluster). Is it in accord with the master-plan?

It is if there weren't any errors to display.  Otherwise there's a bug!   I tested it against the current production deployment, and a worker pool that I knew had some errors.

ui/src/components/WMWorkerPoolsTable/index.jsx
djmitche: This is the old "provisioners" stuff.  Once the new worker-manager list-of-workers UI is finished, I think well get rid of (or redirect from) `/provisioners`.

djmitche:
ok, @helfi92, please have a look when you can?

ui/src/components/WMWorkerPoolsTable/index.jsx
owlishDeveloper: > > It didn't display errors from my table tho (I ran this against my cluster). Is it in accord with the master-plan?
> 
> It is if there weren't any errors to display. Otherwise there's a bug! I tested it against the current production deployment, and a worker pool that I knew had some errors.

oh, worker manager is in production already? I didn't know that.

I do have an error in my cluster. In the Network tab, I didn't even see the request being made. Lemme try again against the production. It's `https://taskcluster.net`, right?

owlishDeveloper:
> > It didn't display errors from my table tho (I ran this against my cluster). Is it in accord with the master-plan?
> 
> It is if there weren't any errors to display. Otherwise there's a bug! I tested it against the current production deployment, and a worker pool that I knew had some errors.

Ok, so it does work against production. It looks great! I would maybe move the `i` button to the Description column, because it's the description that's truncated. My intuition was to click on or around the description to see the full version.

I would also add the "Jump to the beginning" and "Jump to the end" buttons (re: pagination)

djmitche:
> Ok, so it does work against production. It looks great! I would maybe move the i button to the Description column, because it's the description that's truncated. My intuition was to click on or around the description to see the full version.

This sounds good to me. @helfi92 if you agree I'll do it..

> I would also add the "Jump to the beginning" and "Jump to the end" buttons (re: pagination)

This sounds good too but is probably over my head -- I think it would be a change to ConnectionDataTable that affects all paginated results?

helfi92:
> @helfi92 if you agree I'll do it..

I agree with @owlishDeveloper :)

ui/src/components/WMErrorsTable/index.jsx
```diff
@@ -0,0 +1,249 @@
...
+          onHeaderClick={this.handleHeaderClick}
+          renderRow={this.renderTableRow}
+          headers={['Title', 'Description', 'Reported']}
+          padding="dense"
+          onPageChange={onPageChange}
+        />
+        <Drawer
+          anchor="right"
+          open={Boolean(drawerError)}
+          onClose={this.handleDrawerClose}>
```
helfi92: The drawer is hard to close on mobile:

Screen Shot 2019-09-27 at 10 15 54 AM

We can do something similar to the drawer in the pulse-messages view

https://github.com/taskcluster/taskcluster/blob/816c24710f85ed8da561b1647f6246a32deb64e1/ui/src/views/PulseMessages/index.jsx#L65-L69

Screen Shot 2019-09-27 at 10 16 26 AM

services/web-server/src/graphql/WorkerManager.graphql
```diff
@@ -8,6 +8,25 @@ type WorkerManagerWorkerPoolSummary {
...
 }
 
+type WorkerManagerError {
+    errorId: ID!
+    kind: String!
+    title: String!
+    description: String!
+    reported: DateTime
+    extra: JSON
+}
```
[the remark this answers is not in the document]
djmitche: we don't ask for the workerPoolId since it's implied in the query, so I haven't included it in  the prop-types.

ui/src/components/WMErrorsTable/index.jsx
```diff
@@ -0,0 +1,249 @@
...
+    return (
+      <Fragment>
+        <ConnectionDataTable
+          connection={sortedErrors}
+          pageSize={VIEW_WORKER_POOL_ERRORS_PAGE_SIZE}
+          sortByHeader={sortBy}
+          sortDirection={sortDirection}
+          onHeaderClick={this.handleHeaderClick}
+          renderRow={this.renderTableRow}
+          headers={['Title', 'Description', 'Reported']}
```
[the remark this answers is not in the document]
djmitche: Maybe, but it's called "description" in the API.  I tend to want to keep the API and the UI displaying the same thing.

djmitche:
OK, I think I've got everything :)

ui/src/components/WMErrorsTable/index.jsx
```diff
@@ -0,0 +1,276 @@
...
+            <Fragment>
+              <IconButton
+                onClick={this.handleDrawerClose}
+                className={classes.drawerCloseIcon}>
+                <CloseIcon />
+              </IconButton>
+              <div className={classes.metadataContainer}>
+                <Typography variant="h5" className={classes.headline}>
+                  {drawerError.errorId}
+                </Typography>
```
helfi92: Screen Shot 2019-09-27 at 4 13 26 PM

We can make this better by adding `text-overflow: ellipsis;`, `overflow: hidden;` and then setting a `min-width` to something a bit smaller than the width of the drawer.

Screen Shot 2019-09-27 at 4 11 44 PM

Also we can add `title={drawerError.errorId}` to show a tooltip when the user hovers on it just in case they are interested in seeing the end of the error id.

ui/src/views/WorkerManager/WMViewErrors/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+  render() {
+    const { search } = this.state;
+    const {
+      data: { loading, error, WorkerManagerErrors },
+    } = this.props;
+
+    return (
+      <Dashboard
+        title={`Errors for ${this.props.match.params.workerPoolId}`}
```
helfi92: `decodeURIComponent(this.props.match.params.workerPoolId)`

ui/src/components/WMErrorsTable/index.jsx
```diff
@@ -0,0 +1,276 @@
...
+          </TableCellItem>
+        </TableCell>
+        <TableCell>
+          <IconButton
+            className={classes.infoButton}
+            name={errorId}
+            onClick={this.handleDrawerOpen}>
+            <InformationVariantIcon size={iconSize} />
+          </IconButton>
+          <Typography className={classes.errorDescription}>
```
helfi92: Ah, we can add `title={description}` here as well.

djmitche:
I keep running `yarn lint --fix` in the root directory by mistake!

PR #1382, opened 2019-09-23
Remove babel from top-level package.json

djmitche:
It's only used to release client-web, so just install it in that
directory when that's necessary.


djmitche:
It will yell.

djmitche:
Well, this seems to omit regeneratorRuntime for reasons I don't understand.  Meh.

Issue #1315, opened 2019-09-10
[UI] Remove warning


helfi92:
Navigate to http://localhost:5080/auth/clients/email%2Fbastien%40nextcairn.com%2Ftc-notify-test

![Screen Shot 2019-09-10 at 9 23 28 AM](https://user-images.githubusercontent.com/3766511/64617554-ab1ab280-d3ac-11e9-9b2b-bfa407593ded.png)

Let's try to remove this error from the dev-tools.

MisraAnandita:
Hi I would like to  contribute to this issue.  I am a beginner in open source. I have gone through the Task cluster website and have read the initial details about the tool. However, I am unable to open the link provided above for the issue. Could you help me with this? 
Thank you!

helfi92:
Hi @MisraAnandita 👋

The [README](https://github.com/taskcluster/taskcluster/tree/master/ui#readme) inside the `ui/` directory should have information on how to run the UI locally. Once that's done, the link above should work. Give it a try and let me know.

helfi92:
[Hacking on the UI](https://github.com/taskcluster/taskcluster/blob/master/dev-docs/development-process.md#hacking-on-the-ui) will help as well.

MisraAnandita:
Hi @helfi92 Thank you for your prompt reply. It helped me a lot. I cloned the repository to Github Desktop. Installed node and yarn. Did the dependency installations for yarn. As I set the TASKCLUSTER_ROOT_URL to point to a Taskcluster deployment and then try to run yarn start on the services/web-server folder, I have been getting the following error:
$ NODE_ENV=development node src/main.js devServer
'NODE_ENV' is not recognized as an internal or external command,
operable program or batch file.
Have been trying to debug this issue but nothing seems to work.
If you could help me on this, it would be of huge help.
Also, wanted to add that I am using a windows laptop.
Thank you!

rbrishabh:
Hi @MisraAnandita ! There was some talk regarding this issue in windows in [this](https://github.com/taskcluster/taskcluster/pull/1236#issuecomment-524837219) comment.

There seems to be an issue with running Taskcluster on Windows.

I am working on a PR to fix that.

Meanwhile, you could use WSL for Windows.

This will help you get started:
https://docs.microsoft.com/en-us/windows/wsl/install-win10

Also, you might have to reinstall yarn packages on wsl cli, but you can always try once before that!

Hope this helps!

Patil2099:
[...]

PR #1390, opened 2019-09-24
fix issue #1315, remove warning

Rolikasi:
Implemented button in SpeedDialAction component

### DIFF
--- a/ui/src/components/ClientForm/index.jsx
+++ b/ui/src/components/ClientForm/index.jsx
@@ -434,15 +434,14 @@ export default class ClientForm extends Component {
                 ButtonProps={{ disabled: loading }}
               />
               <SpeedDialAction
+                requiresAuth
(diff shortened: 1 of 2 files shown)


Rolikasi:
It's my first contribution. Is everything ok?

Rolikasi:
I had an error after yarn command in /ui/, may be because of this can't pass 2 tests. I fixed it after commit

Rolikasi:
I tried to return to first commit, made new branch and push contribution. 
Now it's [here](https://github.com/taskcluster/taskcluster/pull/1400) 
I think I should close this PR?
Thank you for help, I know that we cant use "class" in JSX, but I thought that 'classes' is another property in this case.

Rolikasi:
But I don't understand why after commit 2 files was deleted and then added with fixes

PR #1400, opened 2019-09-26
[UI] Remove warning

Rolikasi:
[UI]Fix issue #1315


djmitche:
As a quick note, it's best to name both your commits and pull requests a bit more descriptively.  We have some advice [here](https://github.com/taskcluster/taskcluster/blob/master/dev-docs/best-practices/commits.md).  Looking back in the history, the reader won't know (without looking up what #1315 is) what this change does.

It's OK for this PR, so just a note for next time!

Issue #1156, opened 2019-08-05
Task group without a decision task repeatedly shows an error


djmitche:
When viewing [this](https://taskcluster-ui.herokuapp.com/tasks/groups/B-8evnFQRbuX6yeLO49IRg) task group, I get an error bar:

![image](https://user-images.githubusercontent.com/28673/62472850-9c1d6080-b76d-11e9-977b-0c10fcaf2bdb.png)

After dismissing this, it reappears almost instantly.  It looks like it's looking for the decision task (since it's referring to a taskId matching the taskGroupId).

helfi92:
> After dismissing this, it reappears almost instantly.

This issue probably only happens when the task group is running, re-showing the error whenever an update happens and the listener triggers.

Culprit: When the task event listener triggers, the error object reference changes which makes the ErrorPanel component re-display the error since the `error` prop reference changes.

helfi92:
@djmitche How often is there a task group without a decision task?

djmitche:
In Firefox CI, rarely, but in general, quite often.

PR #1402, opened 2019-09-26
[UI] Ignore error if task group without decision task

helfi92:
Resolves https://github.com/taskcluster/taskcluster/issues/1156.

### DIFF
--- a/ui/src/views/Tasks/TaskGroup/index.jsx
+++ b/ui/src/views/Tasks/TaskGroup/index.jsx
@@ -584,6 +584,16 @@ export default class TaskGroup extends Component {
     }
   };
 
+  shouldIgnoreError(error) {
+    if (!error) {
+      return true;
+    }
+
+    // Outside Firefox CI, quite often we can have a
+    // task group without a decision task.
+    return /does not correspond to a task that exists/.test(error.message);
+  }
+
   render() {
     const {
       groupActions,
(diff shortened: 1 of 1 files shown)


ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -584,6 +584,16 @@ export default class TaskGroup extends Component {
...
   };
 
+  shouldIgnoreError(error) {
+    if (!error) {
+      return true;
+    }
+
+    // Outside Firefox CI, quite often we can have a
+    // task group without a decision task.
+    return /does not correspond to a task that exists/.test(error.message);
```
djmitche: Matching on error messages is always tricky -- if we change that error text, this would fail.  Does web-server forward the error code from the backend?  If not, then this is fine (a regression wouldn't be the end of the world).

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -584,6 +584,16 @@ export default class TaskGroup extends Component {
     }
   };
 
+  shouldIgnoreError(error) {
+    if (!error) {
+      return true;
+    }
+
+    // Outside Firefox CI, quite often we can have a
+    // task group without a decision task.
```
djmitche: Maybe "Task groups do not necessarily have a decision task, so handle task-not-found errors gracefully"?

ui/src/views/Tasks/TaskGroup/index.jsx
helfi92: Yes. I added a patch :)

djmitche: Awesome, much better..

Issue #1240, opened 2019-08-26
smoketest - create a task for built-in/succeed and wait for it to succeed


djmitche:
Add a new test to the `yarn smoketest` by adding a module to `infrastructure/tooling/src/smoketest/checks/`.  In this case, it would create a new task for the `built-in/succeed` worker-type, and wait for that task to be claimed and succeed.

See #1218.

ededals:
Hi,

I would like to try to solve this. Is there some example on how to create a task of built/in succeed type ir should I pretty much go and discover it myself?

djmitche:
You'll probably want to try creating such a task using the [task creator](https://tools.taskcluster.net/task-creator/).  Once you're signed in, if you let me know your clientId as shown at https://tools.taskcluster.net/credentials then I can grant you permission to create a task for that worker pool (specifically, you'll set provisionerId to `built-in` and workerType to `succeed`).

ededals:
My clientId: mozilla-auth0/github|43218607|ededals

djmitche:
You should be all set!

ededals:
Hey @djmitche 
I think I finally understood how to do this, but there is one problem. When I am trying to run the test I receive some exception connected to permissions:

```
(node:7217) UnhandledPromiseRejectionWarning: Error: This request requires Taskcluster credentials that satisfy the following scope expression:
{
  "AllOf": [
    {
      "AnyOf": [
        {
          "AllOf": [
            "queue:scheduler-id:-",
            {
              "AnyOf": [
                "queue:create-task:highest:built-in/succeed",
                "queue:create-task:very-high:built-in/succeed",
                "queue:create-task:high:built-in/succeed",
                "queue:create-task:medium:built-in/succeed",
                "queue:create-task:low:built-in/succeed",
                "queue:create-task:very-low:built-in/succeed",
                "queue:create-task:lowest:built-in/succeed"
              ]
            }
          ]
        },
        {
          "AnyOf": [
            "queue:create-task:built-in/succeed",
            {
              "AllOf": [
                "queue:define-task:built-in/succeed",
                "queue:task-group-id:-/EfEcxJsZSfuTkOGMX3IDzg",
                "queue:schedule-task:-/EfEcxJsZSfuTkOGMX3IDzg/EfEcxJsZSfuTkOGMX3IDzg"
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

[...]

PR #1414, opened 2019-09-30
Built in/succeed smoketest

ededals:
Creates a built-in/succeed smoketest for #1240

### DIFF
--- a/infrastructure/tooling/src/smoketest/checks/succeed.js
+++ b/infrastructure/tooling/src/smoketest/checks/succeed.js
@@ -25,11 +25,8 @@ exports.tasks.push({
     utils.status({message: 'built-in/succeed taskId: ' + taskId});
     let queue = new taskcluster.Queue(taskcluster.fromEnvVars());
     await queue.createTask(taskId, task);
-    let pollForStatus = true;
-    setTimeout(function () {
-      pollForStatus = false;
-    }, 120000);
-    while (pollForStatus){
+    let pollForStatusStart = new Date();
+    while((new Date() - pollForStatusStart) < 120000){
       let status = await queue.status(taskId);
       if (status.status.state === 'pending' || status.status.state === 'running'){
         utils.status({
--- a/infrastructure/tooling/src/smoketest/checks/succeed.js
+++ b/infrastructure/tooling/src/smoketest/checks/succeed.js
@@ -22,25 +22,35 @@ exports.tasks.push({
       payload: {},
     };
     let taskId = taskcluster.slugid();
-    let queue = new taskcluster.Queue({
-      rootUrl: process.env.TASKCLUSTER_ROOT_URL,
-      credentials: {
-        clientId: process.env.TASKCLUSTER_CLIENT_ID,
-        accessToken: process.env.TASKCLUSTER_ACCESS_TOKEN,
-      },
-    });
-    let result = await queue.createTask(taskId, task);
-    while (1){
+    utils.status({message: 'built-in/succeed taskId: ' + taskId});
+    let queue = new taskcluster.Queue(taskcluster.fromEnvVars());
+    await queue.createTask(taskId, task);
+    let pollForStatus = true;
+    setTimeout(function () {
+      pollForStatus = false;
+    }, 12000);
+    while (pollForStatus){
       let status = await queue.status(taskId);
-      if (status.state === 'pending'){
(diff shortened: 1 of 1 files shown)


infrastructure/tooling/src/smoketest/checks/succeed.js
```diff
@@ -0,0 +1,46 @@
...
+      if (status.state === 'pending'){
+        await new Promise(resolve => setTimeout(resolve, 2000));
+      } else {
+        return {
+          ['built-in/succeed']: result.state,
+        };
+
+      }
+    }
+    return result;
```
djmitche: There's a lint error here due to the unreachable return.  When you've got the next revision ready to push, run `yarn lint` to find any details like that.

infrastructure/tooling/src/smoketest/checks/succeed.js
```diff
@@ -0,0 +1,46 @@
...
+      metadata: {
+        name: "Smoketest built-in/succeed",
+        description: "built-in/succeed task created during smoketest",
+        owner: "smoketest@taskcluster.net",
+        source: "https://taskcluster.net",
+      },
+      payload: {},
+    };
+    let taskId = taskcluster.slugid();
+    let queue = new taskcluster.Queue({
```
djmitche: you can use `taskcluster.Queue(taskcluster.fromEnvVars())` as a shorter way to write the same thing.

infrastructure/tooling/src/smoketest/checks/succeed.js
```diff
@@ -0,0 +1,46 @@
...
+      rootUrl: process.env.TASKCLUSTER_ROOT_URL,
+      credentials: {
+        clientId: process.env.TASKCLUSTER_CLIENT_ID,
+        accessToken: process.env.TASKCLUSTER_ACCESS_TOKEN,
+      },
+    });
+    let result = await queue.createTask(taskId, task);
+    while (1){
+      let status = await queue.status(taskId);
```
djmitche: The return value of this call has the shape `{status: { state: .. }}`, so you'd need to look at `status.status.state`, not `status.state`.

infrastructure/tooling/src/smoketest/checks/succeed.js
```diff
@@ -0,0 +1,46 @@
...
+      if (status.state === 'pending'){
```
djmitche: ..so `status.state` is always undefined, so this is never true..

infrastructure/tooling/src/smoketest/checks/succeed.js
```diff
@@ -0,0 +1,46 @@
...
+        await new Promise(resolve => setTimeout(resolve, 2000));
+      } else {
+        return {
```
djmitche: It'd be great to assert the state is `completed` here.

ededals:
Hi, thanks for the comments :) I hope I managed to fix everything. 
One thing I am not sure about is, if using ```utils.skip(...)``` for the failing tests was the right choice.

infrastructure/tooling/src/smoketest/checks/succeed.js
```diff
@@ -0,0 +1,56 @@
...
+      payload: {},
+    };
+    let taskId = taskcluster.slugid();
+    utils.status({message: 'built-in/succeed taskId: ' + taskId});
+    let queue = new taskcluster.Queue(taskcluster.fromEnvVars());
+    await queue.createTask(taskId, task);
+    let pollForStatus = true;
+    setTimeout(function () {
+      pollForStatus = false;
+    }, 12000);
```
djmitche: Needs one more `0` (this is only 12 seconds) :)

infrastructure/tooling/src/smoketest/checks/succeed.js
```diff
@@ -0,0 +1,56 @@
...
+        return utils.skip({
+          [`built-in/succeed`]: status.status.state,
+          reason: 'Task finished with status ' + status.status.state,
+        });
+      }
+    }
+    return utils.skip({
+      [`built-in/succeed`]: 'deadline exceeded',
+      reason: 'Deadline exceeded',
+    });
```
djmitche: Good question.  In both of these cases, I think we can just throw an exception.

ededals:
Yes this seems to bet working better and is also more concise.

PR #1415, opened 2019-09-30
Bug 1584321 - merge static_clients into dev-config.yml automatically

djmitche:
Merging static_clients using Helm is not possible, but we can at least
verify that the dev-config.yml has the correct scopes..

Bugzilla Bug: [1584321](https://bugzilla.mozilla.org/show_bug.cgi?id=1584321)


djmitche:
Oh hang on, I bet I can make this work..

djmitche:
This should help with dev deployments.  @imbstack how does this work with cloudops' deployments?

infrastructure/tooling/src/dev/helm.js
```diff
@@ -22,9 +29,26 @@ const actions = [
       if (!config.meta || !config.meta.deploymentPrefix) {
         throw new Error('Must have configured dev-config.yml to deploy.');
       }
+
+      if (config.auth && config.auth.static_clients) {
+        if (config.auth.static_clients.some(({scopes}) => Boolean(scopes))) {
+          throw new Error('auth.static_clients in `dev-config.yml` should not contain scopes');
```
imbstack: Thanks for this (also just a good opportunity to use the multiline comment)

infrastructure/tooling/src/dev/helm.js
```diff
@@ -22,9 +29,26 @@ const actions = [
...
     },
   },
+  {
+    title: 'Get static clients',
+    requires: ['dev-config'],
+    provides: ['static-clients'],
+    run: async (requirements, utils) => {
+      const ymlFile = await readRepoYAML('infrastructure/tooling/static-clients.yml');
+      const config = requirements['dev-config'];
+      const staticClients = await jsone(ymlFile, {azureAccountId: config.azureAccountId || 'unknown'});
```
imbstack: Do we actually want to fallback to `unknown` here or just bail?

djmitche: I don't know if this gets run in a case where this value isn't yet determined.  If not, I can remove the fallback.

djmitche:
> I think they'll still need to update manually.

We should probably either make copy-pasting the static clients a standard part of the deployment process for cloudops (regardless of changelog), or find a way to automate it.

djmitche:
In favor of #1494..

PR #1437, opened 2019-10-01
[UI] Remove page information from /tasks/index

Rolikasi:
These "first good issues" really help me understand how to navigate in big project's code!)


Rolikasi:
Hmmm, I think I forgot to delete previous commits

PR #1456, opened 2019-10-02
Added tooltip to link

devnoorfatima:


ui/src/components/TaskDetailsCard/index.jsx
```diff
@@ -214,7 +214,9 @@ export default class TaskDetailsCard extends Component {
                   component="a"
                   href={task.metadata.source}
                   target="_blank"
-                  rel="noopener noreferrer">
+                  rel="noopener noreferrer" 
+                  title ="https://hg.mozilla.org/integration/mozilla-inbound/raw-file/8735151820ebeaf7ee06e9f2ed16868c164d85ab/.taskcluster.yml"
```
djmitche: I think the idea of the issue is that it shows the value of the `source` property for the task -- not that it shows this exact title every time :)

package.json
```diff
@@ -167,6 +167,7 @@
     "zurvan": "^0.5.2"
   },
   "scripts": {
+
```
djmitche: can you remove this stray blank line?

devnoorfatima: I removed this line also run the yarn lint --fix in ui.

ui/src/components/TaskDetailsCard/index.jsx
devnoorfatima: I am sorry what do you mean by value of source property for the task.

devnoorfatima: Do you mean the title should appear only when Source is hovered.

djmitche: What you've written will always show the exact same link on hover, for every task.  That's not what we want -- every task will have a different link, and we should show that on hover.

devnoorfatima:
Thanks, You have been really nice and helpful. This is my first time
contributing to open source project so simple things seems difficult to
me.I am not sure which issue to choose next.If you can suggest me some that
I can do.

djmitche:
I think #1487 fits the bill?

helfi92:
@all-contributors please add @devnoorfatima for code.

PR #1461, opened 2019-10-02
Remove redundancy in index JSON schemas

### LINKED ISSUE #1452: Remove redundancy in index JSON schemas
`services/index/schemas/v1/list-tasks-response.yml` contains a subschema that is identical to `services/index/schemas/v1/indexed-task-response.yml`.  Use `$ref` to eliminate the duplication by referring from the former to the latter.  Let's also change the name of the latter from "indexed task response" to just "indexed task".

Rolikasi:
Change a name of indexed task response to indexed task
Use $ref to eliminate the duplication
Closes #1452


Rolikasi:
Is it the right way to rename a file in this project? 
I just removed "-response" from indexed-task-response.yml and pushed changes

djmitche:
In `services/index/src/api.js` there's a reference to the indexed-task-response file, that will need to be changed to point to the new name.

Rolikasi:
Got it! Step by step I'm getting the idea how it works)

Rolikasi:
Can I ask you to assign me to any issue with JSON schema after this one? I wanna understand it better)
I live in UTC+6 time zone and can't contact to you or helfi92 during a day

djmitche:
One last thing here -- you'll need to run `yarn generate` to update the generated references file.  That will require getting Go installed as well.  I'd recommend using https://github.com/moovweb/gvm for that purpose.

Look around in the schemas for other services, and see what other redundancies you can find and fix.

Rolikasi:
Ok. thank you, @djmitche!

djmitche:
Hm, I may have changed my mind..  It looks like the Go client (`clients/client-go`) uses the title of the schema to generate a struct name.  I'm not sure that's a great idea, but we should probably not change the struct names unnecessarily.  So let's undo the title change, and run `yarn generate` again, and we should see the struct names under `clients/client-go` go back to what they were.

Rolikasi:
1) I undo title change in indexed-task.yml the run yarn generate and pushed
2) I realized that I need to change all names changes and run yarn generate
after step 2 i had next error in shell 
error json

djmitche:
Hm, that's not a very helpful error message.  `yarn generate` is running `go generate ./...` in the `clients/client-go` directory.  Do you see anything more useful if you run that command directly?

Rolikasi:
Sorry, that error was after second run yarn generate.
Here is the error that occurs after step 2 in previous comment
err yarn generate

Rolikasi:
OK. This issue becomes more interesting) 

Now I will focus on [bugzilla bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1436421) 

I will return to this issue if we'll figure out how to avoid changes for go client!

Issue #1472, opened 2019-10-03
SyntaxError in ui/docs/manual/using/namespaces.mdx


djmitche:
```
WARNING in ./docs/manual/using/namespaces.mdx
Module build failed (from ./node_modules/mdx-loader/index.js):
SyntaxError: unknown: Unexpected token (245:60)

  243 | <Warning>
  244 | In some cases, projects need additional scopes.
> 245 | To do so, create a role `project:<project>:grants/<grantName` containing the desired scopes.
      |                                                             ^
  246 | This role can then be assumed where necessary, or project admins can use it to grant specific scopes as desired.
  247 | </Warning>
  248 |     </MDXLayout>
```

I'm not sure why mdx-loader considers this a "warning" instead of an error.

@vishakhanihore did you see this issue when you made these changes?

djmitche:
Hm, I actually don't know how to fix this..

djmitche:
The bits I don't know:
 * how to make this error show up in CI
 * how to allow writing "regular" Markdown within `<Warning>..</Warning>`.

vishakhanihore:
I will take a look as soon as I get free from my classesm

vishakhanihore:
> ```
> WARNING in ./docs/manual/using/namespaces.mdx
> Module build failed (from ./node_modules/mdx-loader/index.js):
> SyntaxError: unknown: Unexpected token (245:60)
> 
>   243 | <Warning>
>   244 | In some cases, projects need additional scopes.
> > 245 | To do so, create a role `project:<project>:grants/<grantName` containing the desired scopes.
>       |                                                             ^
>   246 | This role can then be assumed where necessary, or project admins can use it to grant specific scopes as desired.
>   247 | </Warning>
>   248 |     </MDXLayout>
> ```
> 
> I'm not sure why mdx-loader considers this a "warning" instead of an error.
> 
> @vishakhanihore did you see this issue when you made these changes?

@djmitche AFAIK, Such issues didn't come up in my PR #1451 as the CI passed at that time.
BTW is this error coming up in `console` or `CI Build` ?

djmitche:
It's coming up in the terminal where I ran `yarn start`.  It doesn't seem to show up in CI -- I'm not sure why!  Thanks for your notes..

vishakhanihore:
@djmitche My fork is insync with the upstream/master and I still cant see this error in web-server nor ui.

djmitche:
[...]

PR #1473, opened 2019-10-03
Fix syntax error in namespace.mdx

djmitche:
This may not be the best fix -- ideally the content of `<Warning>` would
be treated as Markdown, and in markdown `<` is not a special character
when included in backticks.

Refs #1472.


djmitche:
This is causing consternation among outreachy applicants so I'm going to just merge it :/

Issue #1460, opened 2019-10-02
Client editor should show description in markdown


djmitche:
**Describe the bug**
The client editor should show the description in rendered Markdown

**To Reproduce**
1. Go to a client like https://taskcluster-ui.herokuapp.com/auth/clients/static%2Ftaskcluster%2Froot
1. See the description as a text-entry box, rather than as rendered markdown (so for example `*foo*` instead of *foo*).

**Expected behavior**
A tabbed view showing both a preview and a place to write the description, similar to how hooks are displayed.  See https://taskcluster-ui.herokuapp.com/hooks/hg-push/try for example.

**Screenshots**
![Screenshot_2019-10-02 Client - Taskcluster](https://user-images.githubusercontent.com/28673/66076723-69ed6c80-e52c-11e9-8d04-ada35c1e7b1f.png)
![Screenshot_2019-10-02 Hook - Taskcluster](https://user-images.githubusercontent.com/28673/66076724-69ed6c80-e52c-11e9-86c8-c12cd630e5c2.png)

vishakhanihore:
Can I be assigned this ?

djmitche:
We want to make sure everyone gets a chance to work on an issue, and we don't have too many!  So let's save this for another applicant.  You can take a look at the other issues, not labeled "outreachy", and find one that looks good.  Those are likely to be a bit more challenging.

LeoSilvaGomes:
Can I try to fix this issue?

FahdJamy:
@LeoSilvaGomes are you going to take on this one??

FahdJamy:
Hey @LeoSilvaGomes are you going to take on this or should I take it on?
I completed the first step for submitting a screenshot.

somchi:
i did like to fix this issue

echessa:
Hi @djmitche if no one who has responded before me get's assigned, than I'd like to take this issue.

djmitche:
@LeoSilvaGomes is the first to ask after @vishakhanihore.

I'm sorry things are so busy these days!  Unfortunately we always get a *lot* of contributors, and it's difficult to think of enough good first tasks for all of them!  Likely some folks will "disappear" and we can reassign those issues.  Note that you can also find a larger set of bugs to work on in https://codetribute.mozilla.org/projects/taskcluster.  Not all of them are quite as "quick" as those tagged "outreachy", but perhaps you'll find something there that sparks your interest.

olsenme:
[...]

PR #1474, opened 2019-10-03
[UI] Add markdown of discription view

LeoSilvaGomes:
### Change the description client

- Update the TextField to can preview the markdown way, updated to MarkdownTextArea and putted on context

Issue: #1460

---

Result:

- before fix:

image

- after fix:

image

image

### DIFF
--- a/ui/src/components/ClientForm/index.jsx
+++ b/ui/src/components/ClientForm/index.jsx
@@ -13,6 +13,7 @@ import TextField from '@material-ui/core/TextField';
 import Switch from '@material-ui/core/Switch';
 import FormGroup from '@material-ui/core/FormGroup';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
+import MarkdownTextArea from '@mozilla-frontend-infra/components/MarkdownTextArea';
 import ContentSaveIcon from 'mdi-react/ContentSaveIcon';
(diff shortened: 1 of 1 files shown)


djmitche:
Yes, #1473 should fix it.

LeoSilvaGomes:
@helfi92 Done! :+1:

ui/src/components/ClientForm/index.jsx
```diff
@@ -323,15 +328,13 @@ export default class ClientForm extends Component {
               }
             />
           </ListItem>
-          <ListItem>
-            <TextField
-              label="Description"
-              name="description"
+          <ListItem className={classes.clickDescriptionListItem}>
```
helfi92: What does `click` in `clickDescriptionListItem` mean? What do you think if we omit `click` and instead rename it to `descriptionListItem`.

LeoSilvaGomes: What do you think of changing clientDescriptionListItem? I thought about differentiating this description view from others.

helfi92: Sure. `clientDescriptionListItem` sounds good to me :)

LeoSilvaGomes: done!

Issue #1406, opened 2019-09-30
[UI] Remove error in console


helfi92:
**Reserved** for [outreachy](https://www.outreachy.org/) candidates. In order for this to be assigned to you, make sure to complete the first step of the application process outlined in the [project description](https://www.outreachy.org/december-2019-to-march-2020-internship-round/communities/mozilla/#build-a-component-that-can-display-a-json-schema).

---

**Describe the bug**
Navigating to the hook view shows an error in the console.

**To Reproduce**
1. Go to http://localhost:5080/hooks/garbage/abc456
2. Open the developer tools and find for the following error

![Screen Shot 2019-09-30 at 9 29 19 AM](https://user-images.githubusercontent.com/3766511/65883354-dc135500-e364-11e9-893d-8dcd7764249c.png)

**Expected behavior**
This error does not show up.

katherine95:
I have completed the first step and would like to be assigned this one @helfi92

ajinkabeer:
@helfi92 @djmitche 
I would like to work on this issue. Can you please assign it to me?

djmitche:
@katherine95 it's all yours.  I'll assign to you once you accept the invitation.

katherine95:
Thank you, Dustin. I am experiencing issues enabling the two-factor
authentication so that I can accept the repo invite.

katherine95:
@djmitche I have a [PR](https://github.com/taskcluster/taskcluster/pull/1475) for this issue.

PR #1475, opened 2019-10-03
[UI]Fix taskPriority propType validation error

katherine95:
#### What does this PR do?
This PR closes [Issue](https://github.com/taskcluster/taskcluster/issues/1406) concerning the propType validation console error for `taskPriority` at `/hooks/garbage/abc456`

#### Type of change:
- [x] Bug fix (non-breaking change which fixed an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update.

#### How Has This Been Tested?
- Clone this repository 
- Git checkout to `katherine95:Issue-1406`
- Run `yarn` then `yarn start`
- Navigate to hook view  `/hooks/garbage/abc456` and check the console
- You should see that there is no propType validation error on the console

#### Checklist:
- [x] Navigating to the hook view shows no error in the console.

#### Before
Screenshot 2019-10-03 at 18 21 25

#### After
Screenshot 2019-10-14 at 21 14 14

### DIFF
--- a/ui/src/utils/prop-types.js
+++ b/ui/src/utils/prop-types.js
@@ -126,6 +126,7 @@ export const taskPriority = oneOf([
   'LOW',
   'VERY_LOW',
   'LOWEST',
+  'normal',
 ]);
 
 export const taskActions = shape({
--- a/ui/src/utils/prop-types.js
+++ b/ui/src/utils/prop-types.js
@@ -126,7 +126,7 @@ export const taskPriority = oneOf([
   'LOW',
   'VERY_LOW',
   'LOWEST',
-  'normal',
+  'NORMAL',
 ]);
 
 export const taskActions = shape({
--- a/services/web-server/src/graphql/Tasks.graphql
+++ b/services/web-server/src/graphql/Tasks.graphql
@@ -46,6 +46,7 @@ enum TaskPriority {
   LOW
   VERY_LOW
   LOWEST
+  NORMAL
 }
 
 # Dependency requirement semantics; the task's relation to its dependencies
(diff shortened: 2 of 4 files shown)


ui/src/utils/prop-types.js
```diff
@@ -126,6 +126,7 @@ export const taskPriority = oneOf([
   'LOW',
   'VERY_LOW',
   'LOWEST',
+  'normal',
```
helfi92: Ah, we are indeed missing the normal priority of a task. The convention in graphql is to capitalize constants then in web-server we can map the capital letters to their actual values.

helfi92:
@katherine95 Did you want to finish the PR? :-)

katherine95:
Yes @helfi92. Let me look at the feedback given. Thank for feedback.

ui/src/utils/prop-types.js
katherine95: So what I should do is capitalize on the `normal` constant?, It's not piking it when it's normal. So it seems it's picking data that is already from the db and the taskPriority is in small letters.

helfi92: The code to make it work is on the diff that I posted on my review. Also, whenever we change things in `web-server`, we currently need to restart the process to see the changes. Does that help?

katherine95: Yes that really helped. Thanks @helfi92

katherine95:
@helfi92 please review the new changes.

helfi92:
@all-contributors please add @katherine95 for code.

PR #1443, opened 2019-10-02
Placeholder for '/tasks' and '/tasks/groups'

Soundharyaam:
A unique placeholder has been added for  '/tasks' and '/tasks/groups' routes. 

Resolves #1416

PR #1479, opened 2019-10-03
[UI] Resolve text-overflow inside the search box

Soundharyaam:
**Issue:** 

When the placeholder value has been changed inside the search box, the overflow of text happened inside the search box of the route - `/tasks/groups`, since the number of characters were more in this.

**What did I do to resolve this:**

I have increased the overall width of the search box while in stationary mode. In transition mode, it works like before.

**Before bug fix:**

**After bug fix:**

Related to PR:https://github.com/taskcluster/taskcluster/pull/1443

### DIFF
--- a/ui/src/components/Search/index.jsx
+++ b/ui/src/components/Search/index.jsx
@@ -16,7 +16,7 @@ import { THEME } from '../../utils/constants';
     },
     '& $input': {
       transition: theme.transitions.create('width'),
-      width: 200,
+      width: 230,
       '&:focus': {
         width: 300,
       },


djmitche:
This looks OK to me, so I'll leave it to the expert to review :)

Soundharyaam:
Haha. Sure :) @djmitche

ui/src/components/Search/index.jsx
```diff
@@ -16,7 +16,7 @@ import { THEME } from '../../utils/constants';
     },
     '& $input': {
       transition: theme.transitions.create('width'),
-      width: 200,
+      width: 230,
```
helfi92: On mobile, we can no longer see the book icon (docs)

Screen Shot 2019-10-08 at 12 32 29 PM

It used to be

Screen Shot 2019-10-08 at 12 33 01 PM


What do you think we should do?

Soundharyaam: We can keep the search box `width` as same 200 in mobile-view but can reduce the font-size. I think, this would work.

Soundharyaam:
Hi @helfi92,

I made some tweaks around the code and found something interesting. Even, if we keep the search-box `width` as `200`, in the mobile-view it breaks.

Below is the screenshot for `screen-width:650` and with search-box `width:200`:

Screen Shot 2019-10-09 at 2 42 49 PM

Below is the screenshot for `screen-width:375` and with search-box `width:200`:

Screen Shot 2019-10-09 at 2 43 55 PM

Over here, the last two entities are not showing up.

What should we do for this?

Soundharyaam:
@helfi92 That sounds good. Yeah. I will work on that. So, left corner - triple horizontal lines icon(will open left navbar), Center - Search component, right corner - vertical triple dots icons(will open the other four entities on the right side of the search bar). Is this, what you are saying? If that so, what about the heading of the present section? The user has to know, where he/she is in.

helfi92:
You have the right idea!

> If that so, what about the heading of the present section?

This can be omitted on mobile. The user would be able to know where they are by looking at the URL or by inspecting which drawer from the navigation bar is selected.

Soundharyaam:
Cool, then. That would be a new patch as a whole.

helfi92:
Yes. Feel free to close this PR.

Soundharyaam:
But, this fixes the issue in web view.

helfi92:
> But, this fixes the issue in web view.

Correct. A rebase is needed before we merge :)

Soundharyaam:
Closing this PR cause of the issue with rebasing. 

Created the new one over here: 
https://github.com/taskcluster/taskcluster/pull/1563

Issue #1403, opened 2019-09-26
[UI] Add hover effect to the link icon in /auth/scopes


helfi92:
**Reserved** for [outreachy](https://www.outreachy.org/) candidates. In order for this to be assigned to you, make sure to complete the first step of the application process outlined in the [project description](https://www.outreachy.org/december-2019-to-march-2020-internship-round/communities/mozilla/#build-a-component-that-can-display-a-json-schema).

---

**Describe the bug**
The hover effect on list items is missing for the list of scopes.

**To Reproduce**
1. Go to https://taskcluster-ui.herokuapp.com/auth/scopes
2. Hover your mouse on an item from the list

**Expected behavior**
Hovering adds an effect on the icon icon

**Screenshots**

Actual:

![Screen Shot 2019-09-26 at 4 03 01 PM](https://user-images.githubusercontent.com/3766511/65720988-264fba00-e077-11e9-8144-73d1a849131d.png)

![Screen Shot 2019-09-26 at 4 03 05 PM](https://user-images.githubusercontent.com/3766511/65720989-26e85080-e077-11e9-8a1d-fa8c995cddf2.png)

Expected:

![Screen Shot 2019-09-26 at 4 02 17 PM](https://user-images.githubusercontent.com/3766511/65720953-0d470900-e077-11e9-9ac8-30b664ca0c7d.png)

![Screen Shot 2019-09-26 at 4 02 22 PM](https://user-images.githubusercontent.com/3766511/65720952-0cae7280-e077-11e9-81a0-66f857e6e4da.png)

**Additional context**
The hover effect should be similar to how we do it on the rest of the site. Examples where the hover effect on the link icon works:
* https://taskcluster-ui.herokuapp.com/tasks/groups/Ld0iQrz7SQS-LgkMoIWb_Q
* https://taskcluster-ui.herokuapp.com/provisioners/aws-provisioner-v1
* https://taskcluster-ui.herokuapp.com/worker-manager

Jo-IE:
Hi @helfi92 , I would like to work on this please :)

Rolikasi:
@djmitche , @helfi92 , If nobody assigned to this issue, can I take it? I can't find unassigned UI issues without outreachy label.
Or can you suggest me an issue that will help me better understand how to work with JSON schema?

Jo-IE:
Thank you @djmitche !

djmitche:
@Jo-IE are you still working on this?

[...]

Issue #631, opened 2019-04-24
Data field names are coupled with UI


owlishDeveloper:
The way `DataTable` component is currently set up, the names of the fields in the data and the names of the columns in the table must be the same. `DataTable` takes in the array of strings - which we define, and which later will become the keys of the columns to sort by. We map the names of the columns to the data fields by using `camelCase` function (see my last link) - which removes the spaces from the string we see in the UI.

This seems restrictive and abstraction leak. UI should not be coupled with the data like that.

It's better to have an array of objects like this: `{id: 'workerType', label: 'Name'}`, where the `label` is what will be displayed, and `id` is what corresponds to the data field and what will be used to sort. That way, even if I have a field in my data that is named `workerType`, I have no obligation to name the column `Worker Type` in the UI - I can name it `Name`, or `Your Lovely Worker Types` or absolutely whatever. And that whatever can be changed any time without the need to change graphql code and migrate database

Example of what I am talking about in the `material-ui` documentation: https://material-ui.com/demos/tables/#sorting-amp-selecting and press `Show the source` button

Relevant code to look at:
https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/DataTable/index.jsx#L127-L141 - definition of the header and the sorting label
https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/DataTable/index.jsx#L63 - array of strings comes into the table
https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/AwsProvisionerWorkerTypeTable/index.jsx#L39-L66 - example of the sorting function.
OR
https://github.com/taskcluster/taskcluster/blob/master/ui/src/components/AwsProvisionerWorkerTypeTable/index.jsx#L41 - you can just start reading code from this line

helfi92:
Agreed. Adding a label would make this neater. We currently use a mapping function for that. For example:

https://github.com/taskcluster/taskcluster/blob/a60f57c3215c46022a3f0d6f59834bb9134359c1/ui/src/components/AwsProvisionerWorkerTypeStatus/index.jsx#L122-L131

sudipt1999:
If I am not getting it wrong we are planning to change the type props is being received in the component, 
something to be received as array where each element would look like this

```javascript
{ id: 'ColumnName', label: 'Worker Type' , 'Instance Type': item.instanceType, 'Availability Zones': item.zone, }
```
where id would represent the `column names`.

If that so I would like to try this one then!

helfi92:
[...]

PR #1480, opened 2019-10-03
Authscopes hovereffect

Jo-IE:
Thank you for your help @djmitche !

 I made the changes requested in issue #1403 , please let me know if there are any additional changes I need to make. :)

### DIFF
--- a/changelog/1403.md
+++ b/changelog/1403.md
@@ -1,0 +1,4 @@
+# level: patch
+# reference: issue 1403
+---
+The hover effect on list items in  https://taskcluster-ui.herokuapp.com/auth/scopes is now visible with this patch. This change fixes a bug where the list items did not have the hover effect.
--- a/changelog/1403.md
+++ b/changelog/1403.md
@@ -1,4 +1,0 @@
-# level: patch
-# reference: issue 1403
----
-The hover effect on list items in  https://taskcluster-ui.herokuapp.com/auth/scopes is now visible with this patch. This change fixes a bug where the list items did not have the hover effect.
--- a/ui/src/components/ClientScopesTable/index.jsx
+++ b/ui/src/components/ClientScopesTable/index.jsx
@@ -22,6 +22,7 @@ import sort from '../../utils/sort';
 import Link from '../../utils/Link';
 import { VIEW_CLIENT_SCOPES_INSPECT_SIZE } from '../../utils/constants';
 import { pageInfo, client, scopeExpansionLevel } from '../../utils/prop-types';
+import TableCellItem from '../TableCellItem';
 
 const sorted = pipe(
   rSort((a, b) => sort(a.node.clientId, b.node.clientId)),
@@ -113,8 +114,9 @@ export default class ClientScopesTable extends Component {
       map(node => (
         <TableRow key={node}>
           <TableCell padding="dense">
-            <Link
-              className={classes.tableCell}
+            <TableCellItem
+              button
+              component={Link}
               to={
                 selectedScope
                   ? `/auth/clients/${encodeURIComponent(node)}`
@@ -124,7 +126,7 @@ export default class ClientScopesTable extends Component {
                 <Typography>{node}</Typography>
                 <LinkIcon size={16} />
               </div>
-            </Link>
+            </TableCellItem>
           </TableCell>
         </TableRow>
       ))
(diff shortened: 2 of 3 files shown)


changelog/1403.md
```diff
@@ -0,0 +1,4 @@
+# level: patch
+# reference: issue 1403
```
djmitche: The error from `yarn changelog` is because these have leading `#` on them.  But I think this change does not need a changelog entry -- we only make such entries for changes that users would need to know about when upgrading (such as something that might break).  So I think it's best to remove this file.  But good thinking adding it!

Jo-IE: Thank you for your feedback! I will make those changes.

Jo-IE:
Hi @djmitche I have run `yarn lint` and removed the changelog file but now I am getting the following error. I also merged my fork with the original repo to keep it in sync. I posted this on the IRC channel earlier. Any tips on how to resolve this?

Thanks!


image

djmitche:
Sure, that's something that broke in another push.  It's been fixed now, and if you rebase against master you should not see it anymore.

`git pull --rebase upstream/master`

Jo-IE:
Thanks for the reply! I'll do that

Jo-IE:
Hey @djmitche Is it ok to take another issue (currently looking at #631 ) or should I wait till this one is closed?

djmitche:
#631 looks like a good issue for you to look at, yes.

ui/src/components/ClientScopesTable/index.jsx
```diff
@@ -113,8 +114,9 @@ export default class ClientScopesTable extends Component {
       map(node => (
         <TableRow key={node}>
           <TableCell padding="dense">
-            <Link
-              className={classes.tableCell}
+            <TableCellItem
+              button
+              component={Link}
```
helfi92: I think TableCellItem accepts a `dense ` prop to make the items more compact. Let's give this a try. You could have a look at https://taskcluster-ui.herokuapp.com/tasks/index. I think the implementation and look should be similar.

ui/src/components/RoleScopesTable/index.jsx
```diff
@@ -93,18 +93,20 @@ export default class RoleScopesTable extends Component {
...
               : `/auth/scopes/${encodeURIComponent(item)}`
           }>
-          <ListItemText primary={item} />
+          <ListItemText
+            disableTypography
+            primary={<Typography>{item}</Typography>}
+          />
           <LinkIcon size={iconSize} />
-        </ListItem>
+        </TableCellItem>
```
helfi92: Screen Shot 2019-10-07 at 3 14 12 PM

Observations:
* A horizontal line is showing up in the middle which doesn't happen in the clients tab
* The text is too close to the border

Jo-IE: Thank you for the feedback @helfi92 I will make those changes right away!

Jo-IE:
@helfi92 
Update: I am still making the changes requested but I am having a lot of trouble running the environment locally. When I try to load the `http://localhost:5080/auth/scopes` page, the page loads for a very long time and gives me this error in the terminal:

`HPM] Error occurred while trying to proxy request /graphql from localhost:5080 to http://localhost:3050 (ECONNRESET) (https://nodejs.org/api/errors.html#errors_common_system_errors)`

The page only works intermittently but produces the above errror most of the time and so I will be taking longer than expected to make the changes

helfi92:
> When I try to load the http://localhost:5080/auth/scopes page, the page loads for a very long time and gives me this error in the terminal

This usually happens when the web-server (backend) is not running. Make sure that both the ui and web-server are running in their own terminal.

> I also noticed that the tables in https://taskcluster-ui.herokuapp.com/auth/roles and https://taskcluster-ui.herokuapp.com/auth/clients are lacking the hover effect as well. Do I go ahead and make changes to these tables as well?

I think they were fixed recently :)

Jo-IE:
Thanks for the reply @helfi92 ! I also have this [question](https://gist.github.com/Jo-IE/7a8023091ac397df59970f2975843370) concerning this issue

Jo-IE:
ok thank you I have seen your reply and will look into react-window.

helfi92:
Let me know if you had any questions for me.

Jo-IE:
A new PR was opened #1683

PR #1497, opened 2019-10-05
[Bug 1586145] Fix a handful of google provider issues

imbstack:
Each of the commits here is mostly self-explanatory. I believe after this lands we have fixed pretty much all "known" issues with the provider other than estimation being a bit wonky. That and `removeAll` will be my next steps. Testing this in dev now, will update this PR if something goes wrong.

### DIFF
--- a/services/worker-manager/src/providers/google.js
+++ b/services/worker-manager/src/providers/google.js
@@ -354,58 +354,51 @@ class GoogleProvider extends Provider {
     this.seen[worker.workerPoolId] = this.seen[worker.workerPoolId] || 0;
     this.errors[worker.workerPoolId] = this.errors[worker.workerPoolId] || [];
 
-    let deleteOp = false;
-    if (worker.providerData.operation) {
-      deleteOp = await this.handleOperation({
-        op: worker.providerData.operation,
-        errors: this.errors[worker.workerPoolId],
-      });
-    }
-
-    let res;
     try {
-      res = await this._enqueue('get', () => this.compute.instances.get({
+      const {data} = await this._enqueue('get', () => this.compute.instances.get({
         project: worker.providerData.project,
         zone: worker.providerData.zone,
         instance: worker.workerId,
       }));
+      const {status} = data;
+      if (['PROVISIONING', 'STAGING', 'RUNNING'].includes(status)) {
+        this.seen[worker.workerPoolId] += 1;
+
+        // If the worker will be expired soon but it still exists,
+        // update it to stick around a while longer. If this doesn't happen,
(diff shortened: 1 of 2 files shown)


imbstack:
Looks like everything works in my dev env! 

https://taskcluster.imbstack.com/worker-manager/testing%2Fupdates-to-provider-4/errors

services/worker-manager/src/providers/google.js
```diff
@@ -354,58 +354,51 @@ class GoogleProvider extends Provider {
...
-        zone: worker.providerData.zone,
-        instance: worker.workerId,
-      }));
       await worker.modify(w => {
         w.state = states.STOPPED;
-        if (deleteOp) {
-          delete w.providerData.operation;
-        }
       });
+      return;
```
djmitche: Am I mis-reading the indentation, or is this not neccesary?

services/worker-manager/src/providers/google.js
```diff
@@ -354,58 +354,51 @@ class GoogleProvider extends Provider {
...
-      // If the worker will be expired soon but it still exists,
-      // update it to stick around a while longer. If this doesn't happen,
-      // long-lived instances become orphaned from the provider. We don't update
-      // this on every loop just to avoid the extra work when not needed
-      if (worker.expires < taskcluster.fromNow('1 day')) {
-        await worker.modify(w => {
-          w.expires = taskcluster.fromNow('1 week');
+      if (worker.providerData.operation) {
+        // We only check in on the operation if the worker failed to
+        // start succesfully
```
djmitche: ..and this knows that the start failed because the `compute.instance.get` returned a 404?

services/worker-manager/src/providers/google.js
```diff
@@ -119,7 +122,7 @@ class GoogleProvider extends Provider {
         throw err;
       }
 
-      return await this._enqueue(type, func, tries++);
+      return await this._enqueue(type, func, tries + 1);
```
djmitche: oops!

imbstack: yeah...

services/worker-manager/src/providers/google.js
imbstack: Ah, it was a holdover from the previous version before copy-paste. will remove

services/worker-manager/src/providers/google.js
imbstack: That's my working model, yes. If that proves to not be true later then we can go back to the complicated way. I'm thinking that if the instance was created then the operation that reports the status of creating the instance will not have useful info on a failure.

PR #1514, opened 2019-10-07
Uncouple data from ui

Jo-IE:
My latest commit is SHA 7d4f8c7. My other commits from issue #1403 are showing up I'm guessing because it is a diff against taskcluster/master @djmitche

### DIFF
--- a/ui/src/components/AwsProvisionerWorkerTypeTable/index.jsx
+++ b/ui/src/components/AwsProvisionerWorkerTypeTable/index.jsx
@@ -88,10 +88,18 @@ export default class AwsProvisionerWorkerTypeTable extends Component {
         <DataTable
           items={sortedWorkerTypes}
           headers={[
-            'Worker Type',
-            'Pending Tasks',
-            'Running Capacity',
-            'Pending Capacity',
+            { label: 'Worker Type', id: 'Worker Type', type: 'String' },
+            { label: 'Pending Tasks', id: 'Pending Tasks', type: 'Number' },
+            {
+              label: 'Running Capacity',
+              id: 'Running Capacity',
+              type: 'Number',
+            },
+            {
+              label: 'Pending Capacity',
+              id: 'Pending Capacity',
+              type: 'Number',
+            },
           ]}
           sortByHeader={sortBy}
           sortDirection={sortDirection}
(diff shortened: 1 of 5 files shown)


ui/src/components/AwsProvisionerWorkerTypeTable/index.jsx
```diff
@@ -88,10 +88,18 @@ export default class AwsProvisionerWorkerTypeTable extends Component {
         <DataTable
           items={sortedWorkerTypes}
           headers={[
-            'Worker Type',
-            'Pending Tasks',
-            'Running Capacity',
-            'Pending Capacity',
+            { label: 'Worker Type', id: 'Worker Type', type: 'String' },
```
helfi92: I think we should make the id be the property key (e.g., `workerType`). The type should be whatever `typeof <value>` returns. In our case, `typeof 'some-string' will return `string`.

```
{ label: 'Worker Type', id: 'workerType', type: 'string' },
```

helfi92: Did you test that it actually works in the UI?

Jo-IE: Yes I tested it out and the table displays with the headers like it did before. I'll make the changes and update my PR :)

Question : What is the purpose of the `type` key, I am not sure I fully understand what it does.

helfi92: Good question. The purpose of `type` is so that we can right-align numbers in tables for better user experience. The type will help identify columns with numbers. The right alignment doesn't need to be done here. We can add that later.

Jo-IE: Ok, I get it now. Thank you for the response!

Jo-IE:
Hi @helfi92  I have a lot of commits related to a different issue showing up here. Should I start a new PR with just my relevant commits ? Sorry for the confusion!

helfi92:
The way to fix that is to rebase your branch on the remote master. One way to do that is:

```
git remote add upstream git@github.com:taskcluster/taskcluster.git
git pull upstream master --rebase
git push origin uncouple-data-from-ui --force
```

If that doesn't work, then you can open a new PR. Your choice :)

Jo-IE:
😞 I did that already so I'll just open a new PR.

helfi92:
Continued in https://github.com/taskcluster/taskcluster/pull/1559.

PR #1515, opened 2019-10-07
fix python client issue

somchi:
Bug 1135584: [py client] We should return 'None' when the content length is 0, the HTTP status is 204 and something appropriate for non-json responses

Bugzilla Bug 1135584  https://bugzilla.mozilla.org/show_bug.cgi?id=1135584


djmitche:
Congratulations on your first PR, though!

Issue #1518, opened 2019-10-07
[UI] Profile shows n/a for scopes


djmitche:
**Describe the bug**
The "profile" view should show the user's current (basic and expanded) scopes.  It appears to show "n/a" instead

**To Reproduce**
1. Sign in to https://community-tc.services.mozilla.com/ with GitHub
2. Click your picture and choose "Account"
3. See "n/a"

**Expected behavior**
Shows scopes

**Screenshots**
![Screenshot_2019-10-07 Profile - Community-TC](https://user-images.githubusercontent.com/28673/66341160-ab0ec380-e914-11e9-81c7-7015e113aba5.png)

djmitche:
(note that the signature is not especially secret - it's useless without the accessToken - but I figured I'd obscure it anyway.

djmitche:
This may be related to https://bugzilla.mozilla.org/show_bug.cgi?id=1586874.

ANURADHAJHA99:
@djmitche  Can you please assign the following issue to me? I would like to work on it.

djmitche:
Sure!  I think the required fix is actually much smaller: when no scopes are returned, instead of reading "n/a", it should read "*no scopes*".

ANURADHAJHA99:
@djmitche Okay I will try to fix that. Thank you for helping out.

ANURADHAJHA99:
@djmitche I have solved the issue. Please kindly review it. Thank you for assigning me. This the link to the PR. https://github.com/taskcluster/taskcluster/pull/1520

PR #1519, opened 2019-10-07
Bug 1586874 - add github role scopes to web-server

djmitche:
Bugzilla Bug: [1586874](https://bugzilla.mozilla.org/show_bug.cgi?id=1586874)


djmitche:
The more I think about it, the more I think this fixes #1518.  Given invalid credentials (which lack of scopes in the certificate causes), `auth.currentScopes` will return `[]`.

PR #1521, opened 2019-10-08
Heading change in /tasks/index route

Soundharyaam:
**Issue:**

The heading was displayed as the `Index Browser` in `/tasks/index` route which mismatches the current navbar heading for this route. 

**What did I do to resolve this:**

Hence, I made the heading for this route as `Task Index` to match the navbar.

### DIFF
--- a/clients/client-go/tcworkermanager/tcworkermanager.go
+++ b/clients/client-go/tcworkermanager/tcworkermanager.go
@@ -151,6 +151,22 @@ func (workerManager *WorkerManager) UpdateWorkerPool(workerPoolId string, payloa
 
 // Stability: *** EXPERIMENTAL ***
 //
+// Mark a worker pool for deletion.  This is the same as updating the pool to
+// set its providerId to `"null-provider"`, but does not require scope
+// `worker-manager:provider:null-provider`.
+//
+// Required scopes:
+//   worker-manager:manage-worker-pool:<workerPoolId>
+//
+// See #deleteWorkerPool
+func (workerManager *WorkerManager) DeleteWorkerPool(workerPoolId string) (*WorkerPoolFullDefinition, error) {
+	cd := tcclient.Client(*workerManager)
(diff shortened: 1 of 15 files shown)


vishakhanihore:
There are multiple unwanted changes in the file which you just made to increase the line count. Also there is no issue created for the following PR. I think you didn't read the `Contributions Guidelines`.

Soundharyaam:
Ah. Someone needs a chill-pill. This consists of changes from my previous commit as well. I am already on a talk with @djmitche and @helfi92, regarding these bugs. I am not sure whether you read the last point over here: https://github.com/taskcluster/taskcluster/issues/1460. If not, please take a look. And, yeah, I do read `Contributions Guidelines` from top to bottom. Good luck!

vishakhanihore:
I see a lot of unnecessary changes like splitting a intact tag to 4 lines
and making 4 PRs for changes of 4 lines. I think you need to polish your
Open Source Guide to understand how it works. Remeber Quality>Quantity.
Anyways good luck with your 4 PRs .

Soundharyaam:
We all learn from our mistakes. Don't we? If @djmitche or @helfi92, say something about, I would be looking into further for sure. But, pasting the same comment in the four PR and taking it after, when I comment it down, is not a colloquial way of conveying something. We are all working in an open-source environment where we should support each other in a positive way that makes someone be better but not to deter. Think about it. Looking forward to seeing your contributions :)

djmitche:
Thanks for leaving this discussion here -- this is probably an opportune time to mention the [Mozilla Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/) which govern communication on this and all Mozilla projects.  Nothing here violates those guidelines, but worth mentioning all the same.

Whether to split a change into multiple PRs is a judgement call -- if there's a chance that some of the changes might be accepted and others not, then multiple PRs might make sense.  In this case, they probably could have been combined into one PR, since they'll likely all be accepted or changes requested at the same time.  But clicking "Merge" four times is not the end of the world!

Anyway, all good things to think about for the future.

Soundharyaam:
Closing this PR, cause of the issue with rebasing.

Created the new one over here:
https://github.com/taskcluster/taskcluster/pull/1565

PR #1526, opened 2019-10-08
Adding 'go mod tidy' in 'yarn generate`

### LINKED ISSUE #1122: Run 'go mod tidy' in yarn generate
`go mod tidy` seems to do a good job of cleaning out the cruft that other `go` tools put into `go.mod` and `go.sum`.  Let's run that command in both `clients/client-go` and `clients/client-shell` on every `yarn generate` run.

It will also be useful to add that command to the [renovate config](https://renovatebot.com/docs/configuration-options/#gomod).

vishakhanihore:
Fixes #1122


vishakhanihore:
@djmitche Please have a look

vishakhanihore:
@djmitche `Yarn Generate` is running with 100% success on my local after addition of this commands by me. IDK whats the issue on test cases. It would be great if you can take a look.

djmitche:
Did you commit the result of `yarn generate`?  The check fails when running `yarn generate` results in a difference from the source code that's checked in; see https://tools.taskcluster.net/groups/UlQhqaySTkGEPtH1BNDs7w/tasks/SxAkJeiYQaOUGF95Na6KUA/runs/0/logs/public%2Flogs%2Flive_backing.log

vishakhanihore:
Do I need commit the result of yarn generate too?.

djmitche:
Yes please :)

vishakhanihore:
@djmitche @helfi92 Now only 98% tc are running, From the two tc I added, one of the `go mod tidy` runs but it gets stuck on second also without showing any error on console. Help would be much appreciated.
image

djmitche:
You can avoid the fancy colorful output and just get a regular log-style output by running `yarn changelog | cat` -- that might help at least show where it's getting stuck.  I saw it getting stuck for me once, too.  See what you can find, but don't spend more than an hour or so trying to get it to work.  I can take a look after that.

vishakhanihore:
> You can avoid the fancy colorful output and just get a regular log-style output by running `yarn changelog | cat` -- that might help at least show where it's getting stuck. I saw it getting stuck for me once, too. See what you can find, but don't spend more than an hour or so trying to get it to work. I can take a look after that.

@djmitche the second `go mod tidy` gets stuck at `getting go version` . Can you debug where I am going wrong?

djmitche:
I will flag myself for review and take a look.

djmitche:
So, I think the reason for the hangs might have been that the `go mod tidy` run did not depend on the `go generate` run, so both were happening at the same time.  Maybe -- I'm not sure about that.  I added the dependency and the hangs stopped!  I pushed that change.

The other thing I'm worried about, that may result in us abandoning this entire PR, is that `go mod tidy` might not produce consistent results.  It seems to have produced different results for you and for me, and we'll see if the run in the GitHub checks below produces yet different results.  If that's the case, then we may have to give up -- it's critical that `yarn generate` always produce the same output, not matter where it's run.

djmitche:
Well, that's unfortunate: https://tools.taskcluster.net/groups/Ql9Sz1ckTDKjDg1pgsuW-Q/tasks/TqKRXn4oQ7OgGjPMdiW0rg/runs/0/logs/public%2Flogs%2Flive_backing.log#L4896

It generated different output in the automatic check.  So, I don't think this will work, at least not yet.  I'm hoping that in go1.13 or a later version, the `go mod` stuff is a bit better-behaved.  We'll see.

Sorry!  To be clear, that this didn't work is not your fault!  It just happens sometimes that a thing that seemed a good idea turns out to be impossible..

djmitche:
(and you can still count this as a contribution if you'd like)

vishakhanihore:
> Well, that's unfortunate: https://tools.taskcluster.net/groups/Ql9Sz1ckTDKjDg1pgsuW-Q/tasks/TqKRXn4oQ7OgGjPMdiW0rg/runs/0/logs/public%2Flogs%2Flive_backing.log#L4896
> 
> It generated different output in the automatic check. So, I don't think this will work, at least not yet. I'm hoping that in go1.13 or a later version, the `go mod` stuff is a bit better-behaved. We'll see.
> 
> Sorry! To be clear, that this didn't work is not your fault! It just happens sometimes that a thing that seemed a good idea turns out to be impossible..

@djmitche So we need to wait until the `Go Mod` becomes a bit more mature, now if this is the case than should we close the #1122 too or label it as `later` since it may divert new contributors?


> (and you can still count this as a contribution if you'd like)

Thanks for the consideration 😄

djmitche:
I'll put a comment in that issue.

PR #1528, opened 2019-10-08
aws provider - distribute instances over configs

owlishDeveloper:
Bugzilla Bug: [1578900](https://bugzilla.mozilla.org/show_bug.cgi?id=1578900)

This PR doesn't attempt to incorporate the estimator changes, only distributing the instances over different configs.

### DIFF
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -71,6 +71,7 @@ class AwsProvider extends Provider {
     if (toSpawn === 0) {
       return;
     }
+    const toSpawnPerConfig = Math.ceil(toSpawn / workerPool.config.launchConfigs.length);
 
     const userData = Buffer.from(JSON.stringify({
       rootUrl: this.rootUrl,
@@ -93,6 +94,7 @@ class AwsProvider extends Provider {
     const shuffledConfigs = _.shuffle(workerPool.config.launchConfigs);
 
     let spawned;
+    let toSpawnCounter = toSpawn;
     for await (let config of shuffledConfigs) {
       // Make sure we don't get "The same resource type may not be specified
       // more than once in tag specifications" errors
@@ -109,8 +111,8 @@ class AwsProvider extends Provider {
 
           UserData: userData.toString('base64'), // The string needs to be base64-encoded. See the docs above
 
-          MaxCount: toSpawn,
-          MinCount: toSpawn,
+          MaxCount: Math.min(toSpawnCounter, toSpawnPerConfig),
+          MinCount: Math.min(toSpawnCounter, toSpawnPerConfig),
           TagSpecifications: [
             ...otherTagSpecs,
             {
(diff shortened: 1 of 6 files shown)


services/worker-manager/src/providers/aws.js
```diff
@@ -82,83 +90,90 @@ class AwsProvider extends Provider {
...
+      );
+
+      try {
+        spawned = await this.ec2s[config.region].runInstances({
+          ...config.launchConfig,
+
+          UserData: userData.toString('base64'), // The string needs to be base64-encoded. See the docs above
+
+          MaxCount: toSpawn,
+          MinCount: toSpawn,
```
imbstack: I think the only issue here is that if estimator asks for 4 instances and the user has 3 `launchConfigs` specified, this will spawn 12 instances, right? In my head the way this would work would be looping over `shuffledConfigs` spawning at most N instances at a time until we've spawned `toSpawn` instances. Does that make sense or am I missing something.

owlishDeveloper: ah yes yes

services/worker-manager/src/providers/aws.js
```diff
@@ -82,83 +91,93 @@ class AwsProvider extends Provider {
...
+        ts.ResourceType === 'instance' ? instanceTags.concat(ts.Tags) : otherTagSpecs.push(ts)
+      );
+
+      try {
+        spawned = await this.ec2s[config.region].runInstances({
+          ...config.launchConfig,
+
+          UserData: userData.toString('base64'), // The string needs to be base64-encoded. See the docs above
+
+          MaxCount: Math.min(toSpawnCounter, toSpawnPerConfig),
```
imbstack: Can `toSpawnCounter` end up being negative here I think? Probably we can get away with just passing `toSpawnPerConfig` directly to `MaxCount` and exiting this loop early if `toSpawnCounter <= 0`?

services/worker-manager/src/providers/aws.js
```diff
@@ -45,31 +61,24 @@ class AwsProvider extends Provider {
...
-    });
-
     const toSpawn = await this.estimator.simple({
       workerPoolId,
-      minCapacity: config.minCapacity,
-      maxCapacity: config.maxCapacity,
-      capacityPerInstance: config.capacityPerInstance,
+      minCapacity: workerPool.config.minCapacity,
+      maxCapacity: workerPool.config.maxCapacity,
+      capacityPerInstance: 1, // todo: this will be corrected along with estimator changes (bug 1579554)
```
djmitche: :+1:

services/worker-manager/src/providers/aws.js
owlishDeveloper: So the states where `toSpawnCounter <= 0` are reached when unforeseen happens, like instances shutting themselves down immediately etc. I'm not noticing that affecting the functionality, but it does create error noise. I'll add a check at the beginning of the loop and will be breaking out of the loop immediately if `toSpawnCounter <= 0`

owlishDeveloper:
Confirmed working for test cases:

1. n < m, m%n = 0
2. n < m, m%n != 0
3. n = m
4. n > m

where n is the number of configs and m is the number of instances needed

owlishDeveloper:
By "working" I mean it does distribute the amount of instances evenly over the configs. In case one of the configs produces an error, it still spins up the amount needed but in other configs. The errors are logged.

PR #1540, opened 2019-10-08
Multiple launch configs for google provider

imbstack:
This is inspired by #1528 and thus also hardcodes `capacityPerinstance` for now. I know we don't like landing TODOs but this will allow us to land our two changes now and then in [bug 1587234](https://bugzilla.mozilla.org/show_bug.cgi?id=1587234) we can immediately fix both providers at once.

This should also fix [bug 1581483](https://bugzilla.mozilla.org/show_bug.cgi?id=1581483)

Obviously all currently existing workerpools will need updating but I think we can do that all at once and it won't hurt too bad.


services/worker-manager/src/providers/google.js
```diff
@@ -228,50 +228,40 @@ class GoogleProvider extends Provider {
         wt.providerData[this.providerId].running = wt.providerData[this.providerId].running || 0;
       });
     }
-    const regions = workerPool.config.regions;
 
     const toSpawn = await this.estimator.simple({
       workerPoolId,
       ...workerPool.config,
+      capacityPerInstance: 1, // TODO Hardcoded for now until estimator updates
```
djmitche: I think TODO's are OK, *when* they are connected to a bug, and link to that bug in the source.  Then if I find the TODO still in place, I can figure out whether it was done, or closed as WONTFIX, or remains to be done.

owlishDeveloper: This is the bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1579554

I'll mention it in my todo (in my PR) as well

services/worker-manager/src/providers/google.js
```diff
@@ -288,17 +278,21 @@ class GoogleProvider extends Provider {
                 'https://www.googleapis.com/auth/cloud-platform',
               ],
             }],
+            scheduling: {
+              ...cfg.scheduling || {},
+              automaticRestart: false,
```
djmitche: It's probably worth mentioning this and the other hard-coded things (shutdown?) in the schema.

imbstack: We can't hardcode shutdown unfortunately because the different workers get configured to do that differently. Perhaps we should investigate making worker-runner handle that part in a uniform way?

I'll add notes about the other hard-coded things in the schema

I tested this in my dev cluster with old non-compliant worker pools. It throws an error but otherwise doesn't bust anything. I figure if we land this, then the ci-config changes within a few minutes it will be fine.

PR #1541, opened 2019-10-09
[Bug 1461815] Include versions in the docs paths for api, event reference

Soundharyaam:


djmitche:
Discussion of this patch so far is occurring in the bug, https://bugzilla.mozilla.org/show_bug.cgi?id=1461815

Soundharyaam:
@helfi92 @djmitche Could you help me with this issue?

djmitche:
Sure, what have you tried?  What's happened?  What do you need to know?

Soundharyaam:
@djmitche, As I have written over here: https://bugzilla.mozilla.org/show_bug.cgi?id=1461815

Now, I can able to generate the links with v1 version: https://imgur.com/ykFY3gC, https://imgur.com/CQQJK1e.
But, now in the navbar, it renders like Hooks (v1 / logs). When I click on v1 - it's pointing to http://localhost:5080/docs/reference/core/hooks/v1/api.

I think I have to rethink about the conditions which I have written in `docs-tocs.js`. Could you please let me know, how should I proceed after this?

djmitche:
OK, that sounds like good progress!  I think there are three parts here:
 1. Move the files to `v1/` directories (which you've done for hooks)
 1. Make the table of contents data structure skip the intermediate directories, so that it goes from
    * `/docs/reference/core/hooks`
    * `/docs/reference/core/hooks/v1/api`
    * `/docs/reference/core/hooks/v1/exchanges`
    * `/docs/reference/core/hooks/logs`

    without stopping at `/docs/reference/core/hooks/v1/`.  That's what the bit of code you've edited is doing.  The existing code, when it finds `/docs/reference/some/path/to/README`, copies the content to `/docs/reference/some/path/to` and removes that node's children.  The effect is that we never see "README" in the paths in the table of contents. 
 1. update the way the navbar is generated to display this as "hooks (api: _v1_ / exchanges: _v1_ / logs)" instead of putting each of those on their own lines.

You've made a start to step 2, but it's probably good to finish it up so it works for more than just hooks.  After that, I'd suggest finishing step 1 for all of the services.  Then work on step 3.

djmitche:
@Soundharyaam did you want to finish this up?

djmitche:
(closing as stale)

PR #1542, opened 2019-10-09
[Bug 1436421] Edit hooks in YAML instead of JSON

Rolikasi:
Bugzilla Bug: 1436421
replace usage of json with yaml

### DIFF
--- a/ui/src/components/HookForm/index.jsx
+++ b/ui/src/components/HookForm/index.jsx
@@ -13,6 +13,7 @@ import ListItem from '@material-ui/core/ListItem';
 import ListItemText from '@material-ui/core/ListItemText';
 import TextField from '@material-ui/core/TextField';
 import Switch from '@material-ui/core/Switch';
+import { safeLoad, safeDump } from 'js-yaml';
 import FormGroup from '@material-ui/core/FormGroup';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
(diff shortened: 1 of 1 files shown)


ui/src/components/HookForm/index.jsx
```diff
@@ -67,6 +68,7 @@ const initialHook = {
     additionalProperties: false,
   },
 };
+const safeDumpOpts = { noCompatMode: true, noRefs: true };
```
helfi92: I think we can remove this and use default options.

ui/src/components/HookForm/index.jsx
```diff
@@ -759,6 +760,7 @@ export default class HookForm extends Component {
                     </span>
                   </Typography>
                   <CodeEditor
+                    mode="yaml"
                     lint
                     value={taskInput}
                     onChange={this.handleTaskChange}
```
helfi92: For both editors, let's add (in YAML) to the title. Something like:

```jsx
<Typography variant="subtitle1">
    Task Template <small>(in YAML)</small> *
</Typography>
```

helfi92:
> I've found a bug but I cant fix it. can you please help me @helfi92?

Hm, I wasn't able to replicate after removing the safeDump options. It should be okay.

Rolikasi:
Thank you, @helfi92) 
Feel like superhero!)

PR #1547, opened 2019-10-09
[Bug 1535] Add cross-env to support users on windows

Karskaya:
Add cross-env in devDependencies ui/package.json and change set enviroment variable string to use cross-env in services/web-server

Bugzilla Bug: [1535](https://github.com/taskcluster/taskcluster/issues/1535)


ui/package.json
```diff
@@ -106,6 +106,7 @@
     "@neutrinojs/react": "9.0.0-rc.0",
     "babel-loader": "^8.0.5",
     "chai": "^4.2.0",
+    "cross-env": "^6.0.3",
```
helfi92: We have to run `yarn add -D cross-env` to add the package instead of manually writing it. This will make sure to update `ui/yarn.lock`.

services/web-server/package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "coverage": "c8 yarn test",
     "coverage:report": "c8 yarn test && c8 report --temp-directory ./coverage/tmp --reporter json --report-dir ../../artifacts",
-    "start": "NODE_ENV=development node src/main.js devServer",
+    "start": "cross-env NODE_ENV=development node src/main.js devServer",
```
helfi92: We will need to add `cross-env` with `yarn add -D cross-env -W` otherwise it won't be recognized here.

Karskaya: I've run `yarn add -D cross-env -W`. It changed package.json and yarn.lock. But after commit merge conflict appeared in following place:
`"taskcluster-lib-scopes@link:../libraries/scopes":
  version "0.0.0"
  uid ""`

helfi92: Sometimes running `yarn` will help fix the conflict.

Karskaya:
@helfi92, @djmitche Thank you for reviewing! What's next should I do here?

ui/package.json
helfi92: Ah, we'll also want to use `cross-env` on the start script of the ui:

```
"start": "cross-env GENERATE_ENV_JS=1 webpack-dev-server --mode development",
```

helfi92:
Nice!!!

PR #1559, opened 2019-10-09
Uncouple data from the UI

### LINKED ISSUE #631: Data field names are coupled with UI

Jo-IE:
Array of strings props passed into the Data Tables component caused the
column names to always be the same as the respective string passed in.

Closes #631


Jo-IE:
Thanks for the review @helfi92 

Would it be ok to eliminate this line completely
`const sortByProperty = sortBy;` and change all instances of `sortByProperty` to be just `sortBy` ?

helfi92:
> Would it be ok to eliminate this line completely
const sortByProperty = sortBy; and change all instances of sortByProperty to be just sortBy ?

That works :)

Jo-IE:
I have updated the PR with the requested changes @helfi92

ui/src/components/DataTable/index.jsx
```diff
@@ -132,12 +144,11 @@ export default class DataTable extends Component {
...
                   {headers.map(header => (
-                    <TableCell key={`table-header-${header.label}`}>
+                    <TableCell key={`table-header-${header.id}`}>
                       <TableSortLabel
-                        id={header.label}
-                        active={header.label === sortByHeader}
+                        active={header.id === sortByLabel}
                         direction={sortDirection || 'desc'}
-                        onClick={this.handleHeaderClick}>
+                        onClick={() => this.handleHeaderClick(header)}>
```
helfi92: Hm, I think this looks a bit different than the the diff.

ui/src/components/DataTable/index.jsx
```diff
@@ -40,22 +41,33 @@ export default class DataTable extends Component {
...
     sortDirection: oneOf(['desc', 'asc']),
     /**
-     * A list of header names to use on the table starting from the left.
+     * An list of objects. Each element contains a label
+     * representing a header name to use on the table starting from the left,
+     * an id which is a camel-cased version of the header name used in
+     * the sorting function to sort column contents
+     * and a type, a string identifying the data type of the column contents.
      */
-    headers: arrayOf(string),
```
helfi92: Add a generic comment for headers like `A list of header objects to use on the table starting from the left.` and then add a comment for each property. See other comments.

ui/src/components/DataTable/index.jsx
```diff
@@ -40,22 +41,33 @@ export default class DataTable extends Component {
...
+    headers: arrayOf(
+      shape({
+        id: string,
+        type: string,
```
helfi92: `A string identifying the data type of the column contents (e.g., 'string', 'object', 'number').`

ui/src/components/DataTable/index.jsx
```diff
@@ -40,22 +41,33 @@ export default class DataTable extends Component {
...
+        label: string,
```
helfi92: `A label to use for the column name.`

ui/src/components/DataTable/index.jsx
```diff
@@ -40,22 +41,33 @@ export default class DataTable extends Component {
...
+     * An list of objects. Each element contains a label
+     * representing a header name to use on the table starting from the left,
+     * an id which is a camel-cased version of the header name used in
+     * the sorting function to sort column contents
+     * and a type, a string identifying the data type of the column contents.
      */
-    headers: arrayOf(string),
+    headers: arrayOf(
+      shape({
+        id: string,
```
helfi92: `A identifier for the header. This is used to identify which column was clicked for example.`

ui/src/components/AwsProvisionerWorkerTypeTable/index.jsx
```diff
@@ -82,26 +80,27 @@ export default class AwsProvisionerWorkerTypeTable extends Component {
...
-            },
-            {
-              label: 'Pending Capacity',
-              id: 'pendingCapacity',
-              type: 'number',
-            },
-          ]}
-          sortByHeader={sortBy}
+          headers={headers}
+          sortByLabel={sortBy}
```
helfi92: We'll need to do something similar to every view/component that uses the DataTable component.

ui/src/components/DataTable/index.jsx
Jo-IE: Yes, the reason for that is because when I call the function the way it is in the diff, react throws an error `Uncaught Invariant Violation: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.` so I had to pass a reference of the `handleHeaderClick` function to avoid it being repeatedly called when `render()` is called.

ui/src/components/AwsProvisionerWorkerTypeTable/index.jsx
Jo-IE: ok I 'll get to work on that. :)

ui/src/components/DataTable/index.jsx
helfi92: Interesting. Can you try again, please? I wonder if it was because it needed a hard refresh. It seems to work on my end.

Jo-IE: Ok will do

ui/src/components/AwsProvisionerWorkerTypeTable/index.jsx
Jo-IE: I am still working on making changes to this PR. I will have an update by tomorrow.

ui/src/components/DataTable/index.jsx
Jo-IE: I did a refresh for this but I still got the same error

Jo-IE:
Do I repeat this for the `ConnectionDataTable `component next?

helfi92:
> @helfi92 I have updated the views and components that display the DataTable component. Do I repeat this for the ConnectionDataTable component next?

I think that's a good idea but we should probably do that in a separate PR to avoid a mega PR :) I will review this very soon!

Jo-IE:
@helfi92 Ok noted :)

ui/src/components/AwsProvisionerWorkerTypeStatus/index.jsx
```diff
@@ -112,19 +112,19 @@ export default class AwsProvisionerWorkerTypeStatus extends Component {
...
     const mapping = {
-      'Instance Type': item.instanceType,
-      'Availability Zones': item.zone,
-      'Running Instances': item.running,
-      'Pending Instances': item.pending,
+      instanceType: item.instanceType,
+      availabilityZones: item.zone,
+      runningInstances: item.running,
+      pendingInstances: item.pending,
     };
```
helfi92: We should be able to do this without a `valueFromNode` method. We no longer need a mapping when we have the `id` key.

ui/src/components/DataTable/index.jsx
```diff
@@ -40,22 +41,39 @@ export default class DataTable extends Component {
...
-     * A list of header names to use on the table starting from the left.
+     * A list of header objects to use on the table starting from the left.
      */
-    headers: arrayOf(string),
+    headers: arrayOf(
+      shape({
+        /**
+         * A string identifying the data type of the column contents
+         * (e.g., 'string', 'object', 'number'). */
+        id: string,
```
helfi92: The comment for `id ` seems to belong to `type`.

ui/src/components/DataTable/index.jsx
```diff
@@ -40,22 +41,39 @@ export default class DataTable extends Component {
...
+        /**
+         * An identifier for the header.
+         * This is used to identify, for example, which column was clicked.
+         */
+        type: string,
```
helfi92: The comment for `type ` seems to belong to `id`.

ui/src/components/Ec2ResourcesTable/index.jsx
```diff
@@ -78,19 +78,19 @@ export default class Ec2ResourcesTable extends Component {
...
   valueFromNode(item) {
     const mapping = {
-      'Instance ID': item.id,
+      instanceId: item.id,
       State: item.state,
-      'Availability Zone': item.zone,
-      'Launch Time': item.launch,
+      availabilityZone: item.zone,
+      launchTime: item.launch,
     };
```
helfi92: We should be able to do this without a mapping.

ui/src/components/Ec2ResourcesTable/index.jsx
```diff
@@ -109,22 +109,40 @@ export default class Ec2ResourcesTable extends Component {
...
     const iconSize = 16;
+    const headers = [
+      { label: 'Instance ID', id: 'instanceId', type: 'string' },
+      {
+        label: 'State',
+        id: 'state',
+        type: 'string',
+      },
+      // Blocked by https://bugzilla.mozilla.org/show_bug.cgi?id=1453649
+      // 'Instance Type',
```
helfi92: We are no longer blocked by this bug. We can remove the comment.

ui/src/components/Ec2ResourcesTable/index.jsx
```diff
@@ -109,22 +109,40 @@ export default class Ec2ResourcesTable extends Component {
...
+      {
+        label: 'Availability Zone',
+        id: 'availabilityZone',
+        type: 'string',
+      },
+      // Blocked by https://bugzilla.mozilla.org/show_bug.cgi?id=1453649
+      // 'AMI',
```
helfi92: Same thing here. Comment not needed.

ui/src/components/DataTable/index.jsx
Jo-IE: Sorry about the mixup, I'll get that sorted.

ui/src/components/Ec2ResourcesTable/index.jsx
Jo-IE: Should I also remove the 'Instance Type' and 'AMI' comments? What are those for?

helfi92: Yeah, you can remove them.

ui/src/components/AwsProvisionerWorkerTypeStatus/index.jsx
Jo-IE: After removing the `valueFromNode` method from ui/src/components/AwsProvisionerWorkerTypeStatus/index.jsx and ui/src/components/Ec2ResourcesTable/index.jsx, I discovered that the column contents no longer sort when the header is clicked. In ui/src/components/AwsProvisionerWorkerTypeStatus/index.jsx, I had to change the `id` keys in this way to enable sorting:
```
[
      { label: 'Instance Type', id: 'instanceType', type: 'string' },
      { label: 'Availability Zones', id: 'zone', type: 'string' },
      {
        label: 'Running Instances',
        id: 'running',
        type: 'number',
      },
      {
        label: 'Pending Instances',
        id: 'pending',
        type: 'number',
      },
    ];
```
They are no longer camel cased versions of the labels. Is this ok?

helfi92: That's fine :) You are doing it properly.

Jo-IE:
Hi @helfi92  thanks for the review. Might I ask why we deleted those files? Are those components no longer needed?

helfi92:
Yep, no longer needed. It was a view that was unused.

helfi92:
I usually do an interactive rebase to help me fix confilicts. See the section on interactive rebase in https://www.atlassian.com/git/tutorials/merging-vs-rebasing. I think it might help.

Jo-IE:
ok thanks

Jo-IE:
@helfi92 Hi, I have done the rebase as requested.

PR #1569, opened 2019-10-10
Fix and check go pseudo-versions

### LINKED ISSUE #1492: client-go requires an invalid pseudo-version of streadway/amqp
**Describe the bug**

In module-aware mode, `go get` complains about an invalid pseudo-version.

**To Reproduce**
1. Create a new go module that requires clients-go (`github.com/taskcluster/taskcluster/clients/client-go/v18 v18.0.3`).
2. Run `go get`.
3. See error:
```
go: github.com/taskcluster/taskcluster/clients/client-go/v18@v18.0.3 requires
        github.com/streadway/amqp@v0.0.0-20190827080102-edfb9018d271: invalid pseudo-version: does not match version-control timestamp (2019-08-27T07:21:41Z)
```

**Expected behavior**
No error produced

 - OS: Linux
 - Go: 1.13

**Additional context**
I only started using Go modules recently so maybe I'm doing something wrong...

djmitche:
Fixes #1492.  See that issue for what little of this I understand.


imbstack:
Maybe have the test fail with a link to what to do if this happens again? Not blocking this landing though

djmitche:
I'm not sure I know what to do..

PR #1571, opened 2019-10-10
Update links from tc-github to point to new UI paths

djmitche:
I noticed these in https://github.com/djmitche/taskcluster/runs/255790865

### DIFF
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -321,6 +321,11 @@ class Handlers {
 }
 module.exports = Handlers;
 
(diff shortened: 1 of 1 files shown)


djmitche:
> Will this work on the tools site?

Nope.  Should I add some conditionals on rootUrl or should we just pinky-swear not to deploy tc-github for the next 29 days?

djmitche:
Eh, I decided to add some conditionals in there.  I'm still not planning to deploy to heroku, but at least it's not broken if someone does.

PR #1574, opened 2019-10-10
Bug 1574666 - move this repo to the community deploy ment

djmitche:
Bugzilla Bug: [1574666](https://bugzilla.mozilla.org/show_bug.cgi?id=1574666)

* Use the proj-taskcluster/ci worker pool
* Use a new set of secrets, including Azure credentials (rather than fetching from Auth)

This will be a breaking change: PRs for branches not rebased onto this one will no longer work, because they'll use the "old" worker types and the wrong secrets.  I don't see a great way around that.  We will just need to do some rebasing.  We only have 37 open PRs ;)

Also, this one will fail on the old deployment so expect some X's below - a [push](https://community-tc.services.mozilla.com/tasks/groups/doiuNx_YT9Ovi-5heI7PYA) was green.

### DIFF
--- a/libraries/testing/src/secrets.js
+++ b/libraries/testing/src/secrets.js
@@ -118,6 +118,10 @@ class Secrets {
     const that = this;
     let skipping = false;
 
+    if (secretList.some(n => ! this.secrets[n])) {
+      throw new Error(`Unknown secrets in ${JSON.stringify(secretList)}`);
+    }
+
     suite(`${title} (mock)`, function() {
       suiteSetup(async function() {
         skipping = false;
--- a/libraries/testing/src/secrets.js
+++ b/libraries/testing/src/secrets.js
@@ -1,6 +1,6 @@
 const _ = require('lodash');
 const assert = require('assert');
-const request = require('superagent');
+const taskcluster = require('taskcluster-client');
 const debug = require('debug')('tc-lib-testing:secrets');
 
 class Secrets {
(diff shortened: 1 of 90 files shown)


libraries/testing/src/secrets.js
```diff
@@ -80,15 +80,13 @@ class Secrets {
   async _fetchSecrets() {
     const secrets = {};
 
-    // construct a taskcluster-proxy URL to get the secret.  We can't use the taskcluster-client
-    // as it cannot form URLs that match the proxy right now (https://bugzilla.mozilla.org/show_bug.cgi?id=1460015)
```
imbstack: Hey look, we closed a bug

services/auth/src/aws.js
```diff
@@ -124,6 +124,7 @@ builder.declare({
   }
 
   // For details on the policy see: http://amzn.to/1ETStaL
+  console.log(this.sts);
```
imbstack: Stray debug

services/hooks/test/taskcreator_test.js
```diff
@@ -12,30 +11,17 @@ const testing = require('taskcluster-lib-testing');
...
     this.slow(500);
 
-    /* Note that this requires the following set up in production TC:
-     *  - TC credentials given in cfg.get('taskcluster:credentials') with
-     *    - assume:hook-id:tc-hooks-tests/tc-test-hook
-     *    - auth:azure-table-access:pamplemousse/*
-     *  - a role `hook-id:tc-hooks-tests/tc-test-hook` with scopes
-     *    - queue:create-task:no-provisioner/test-worker
-     *    - project:taskcluster:tests:tc-hooks:scope/required/for/task/1
-     */
```
imbstack: I think this being a big red block confirms that this was a good change to make

services/auth/src/aws.js
djmitche: oh geez

PR #1578, opened 2019-10-11
Bug 1588083 - improve smoketests

djmitche:
Bugzilla Bug: [1588083](https://bugzilla.mozilla.org/show_bug.cgi?id=1588083)

### DIFF
--- a/infrastructure/tooling/src/smoketest/checks/builtin.js
+++ b/infrastructure/tooling/src/smoketest/checks/builtin.js
@@ -1,5 +1,13 @@
 const taskcluster = require('taskcluster-client');
 
+exports.scopeExpression = {
+  AllOf: [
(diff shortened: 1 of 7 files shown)


infrastructure/tooling/src/smoketest/checks/builtin.js
```diff
@@ -1,5 +1,13 @@
 const taskcluster = require('taskcluster-client');
 
+exports.scopeExpression = {
```
imbstack: We'll have to be careful to keep these synchronized with whatever the actual scopes are for various smoketests right? This feels like it will go out of date quickly iiuc

djmitche: Yeah, the best I could do was to make it per-file (so this chunk is just the scopes for the built-in-workers smoketests).

PR #1579, opened 2019-10-11
improve error output for tooling

djmitche:
The key here is `console.error(err.toString())` instead of `console.error(err)`; the latter prints some object-like notation, whereas the former calls the error's stringify function, which assertion errors use to format the nice diff display.


djmitche:
FIL! (Friday I learned..)

Issue #1248, opened 2019-08-26
smoketest - purge a cache


djmitche:
Add a new test to the `yarn smoketest` by adding a module to `infrastructure/tooling/src/smoketest/checks/`.  In this case, it would call the purge-cache service to create a purge, then pretend to be a worker and see that the purge was properly recorded.

See #1218.

ajinkabeer:
Can I try this one out please?
@djmitche

djmitche:
Sure thing!  If you need scopes added to the role, please let me know.

ajinkabeer:
> Sure thing! If you need scopes added to the role, please let me know.

I have started working on this smoketest, I need scopes to make API requests to Purge Cache API and also can I set any random provisionerId and workerType or should I use any existing ones?

@djmitche

djmitche:
Can you tell me the precise scopes you need?

Let's use `built-in` and `succeed` for provisionerId / workerType.

ajinkabeer:
> Can you tell me the precise scopes you need?
> 
> Let's use `built-in` and `succeed` for provisionerId / workerType.

Yes, I need `purge-cache:built-in/succeed:This is a test cache name` as the scope.

djmitche:
I added `purge-cache:built-in/succeed:smoketest-cache` to https://tools.taskcluster.net/auth/roles/project%3Ataskcluster%3Asmoketests

ajinkabeer:
Okay, so that means my cacheName is smoketest-cache right? 
@djmitche

ajinkabeer:
> I added `purge-cache:built-in/succeed:smoketest-cache` to https://tools.taskcluster.net/auth/roles/project%3Ataskcluster%3Asmoketests

Yes, It's working. Thank you.

ajinkabeer:
I am having some doubts regarding this test. As you can see, the response I get is `request : {[object]}`. I want the data inside the object to compare if the cacheName is recorded right? Is this the right approach, can you take a loot at this screenshot please?
 
![Capture](https://user-images.githubusercontent.com/30138596/66657689-e1727880-ec5d-11e9-8baf-025a6181de80.JPG)

djmitche:
It's looking pretty good, yeah.  I'm not sure why the errors are showing up like that, and not spelling out the actual differences.  But if you want to see what the data structure looks like, just add a `console.log` temporarily, and run `yarn smoketest | cat` to avoid the animated, colored output and show the result of console.log.

[...]

PR #1580, opened 2019-10-11
Add purgecache smoketest

ajinkabeer:
This PR which addresses the issue #1248  adds a new test to the yarn smoketest by adding a module to `infrastructure/tooling/src/smoketest/checks/`. It calls the purge-cache service to create a purge, then pretend to be a worker and see that the purge was properly recorded.

### DIFF
--- a/infrastructure/tooling/src/smoketest/checks/purgeCache.js
+++ b/infrastructure/tooling/src/smoketest/checks/purgeCache.js
@@ -10,13 +10,13 @@ exports.tasks.push({
   ],
   run: async () => {
     let purge = new taskcluster.PurgeCache(taskcluster.fromEnvVars());
-    const provisionorId = 'built-in'
-    const workerType = 'succeed'
+    const provisionorId = 'built-in';
+    const workerType = 'succeed';
     const payload = {
-        cacheName:'smoketest-cache'
+      cacheName: 'smoketest-cache',
     };
-    await purge.purgeCache(provisionorId,workerType,payload)
-    const pretendWorker = await purge.purgeRequests(provisionorId,workerType)
-    assert.equal(pretendWorker.requests[0].cacheName,payload.cacheName,"Error")
- }
+    await purge.purgeCache(provisionorId, workerType, payload);
+    const pretendWorker = await purge.purgeRequests(provisionorId, workerType);
+    assert.equal(pretendWorker.requests[0].cacheName, payload.cacheName, "Error");
+  },
 });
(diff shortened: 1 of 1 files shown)


infrastructure/tooling/src/smoketest/checks/purgeCache.js
```diff
@@ -0,0 +1,22 @@
...
+exports.tasks = [];
+exports.tasks.push({
+  title: 'Create purge cache smoke test',
+  requires: [],
+  provides: [
+    'purge-cache',
+  ],
+  run: async () => {
+    let purge = new taskcluster.PurgeCache(taskcluster.fromEnvVars());
+    const provisionorId = 'built-in';
```
djmitche: `provision_e_rId` :)

ajinkabeer: Oh sorry, my mistake!

ajinkabeer: > `provision_e_rId` :)

Fixed it. :)

djmitche:
As a quick note, when pushing additional changes to a PR like this, there are two common approaches:

1. Add a new commit with a subject like "fix spelling error" or sometimes just "fix review comments" if it fixes a bunch of tiny things; or
1. Use `git commit --amend` to merge the additional changes into the first commit, and then `git push -f` to push the updates to the PR.

The first case has the advantage that it's easy for the reviewer to see what you've changed.  The second has the advantage that it makes fewer commits and hides some of the not-quite-right changes.

Different OSS projects might have preferences between those.  In Taskcluster, either one is fine.  One approach to avoid, though, is what you've done here where each commit has the same subject -- that makes it easy to get confused about which one is which!

ajinkabeer:
> As a quick note, when pushing additional changes to a PR like this, there are two common approaches:
> 
> 1. Add a new commit with a subject like "fix spelling error" or sometimes just "fix review comments" if it fixes a bunch of tiny things; or
> 2. Use `git commit --amend` to merge the additional changes into the first commit, and then `git push -f` to push the updates to the PR.
> 
> The first case has the advantage that it's easy for the reviewer to see what you've changed. The second has the advantage that it makes fewer commits and hides some of the not-quite-right changes.
> 
> Different OSS projects might have preferences between those. In Taskcluster, either one is fine. One approach to avoid, though, is what you've done here where each commit has the same subject -- that makes it easy to get confused about which one is which!

Thanks for the note. I usually squash the unwanted commits, I will follow these two common approaches like you mentioned.
@djmitche

PR #1582, opened 2019-10-11
Bug 1587974  - Catch error when retrieving profile

helfi92:
Bugzilla Bug: [https://bugzilla.mozilla.org/show_bug.cgi?id=1587974](https://bugzilla.mozilla.org/show_bug.cgi?id=1587974)

Can't seem to reproduce but I think this is a possible fix.

### DIFF
--- a/services/web-server/src/login/strategies/github.js
+++ b/services/web-server/src/login/strategies/github.js
@@ -13,7 +13,7 @@ const GithubClient = require('../clients/GithubClient');
 const debug = Debug('strategies.github');
 
 module.exports = class Github {
-  constructor({ name, cfg, GithubAccessToken }) {
+  constructor({ name, cfg, GithubAccessToken, monitor }) {
     const strategyCfg = cfg.login.strategies[name];
 
     assert(strategyCfg.clientId, `${name}.clientId is required`);
@@ -23,6 +23,7 @@ module.exports = class Github {
 
     this.identityProviderId = 'github';
     this.GithubAccessToken = GithubAccessToken;
+    this.monitor = monitor;
   }
 
   async getUser({ username, userId }) {
@@ -43,6 +44,7 @@ module.exports = class Github {
       if (githubErr.status === 404) {
         return;
       }
+
       throw githubErr;
     }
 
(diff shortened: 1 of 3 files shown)


services/web-server/src/login/strategies/mozilla-auth0.js
```diff
@@ -62,9 +62,14 @@ module.exports = class MozillaAuth0 {
 
   async getUser({ userId }) {
     const personApi = await this.getPersonApi();
-    const userProfile = await personApi.getProfileFromUserId(userId);
+    const [userProfileErr, userProfile] = await tryCatch(personApi.getProfileFromUserId(userId));
     const user = new User();
 
+    if (userProfileErr) {
+      debug(`error retrieving profile from the PersonAPI`);
```
djmitche: Let's report that error, too.  If "user not found" is treated as an error, then we could check for that condition and report any *other* errors.

services/web-server/src/login/strategies/mozilla-auth0.js
```diff
@@ -192,9 +202,9 @@ module.exports = class MozillaAuth0 {
         // extraParams.id_token has the JSON Web Token
         // profile has all the information from the user
         async (accessToken, refreshToken, extraParams, profile, done) => {
```
djmitche: Ah, so this is where it wasn't handled.  Nice find!

services/web-server/src/login/strategies/mozilla-auth0.js
djmitche: My thinking was to report everything *except* userNotFound -- we might want to see, for example, if a lot of our People API calls are 500'ing, a report in Sentry.  But if a bunch of people get fired and the client scanner doesn't find them when it looks them up, that's a "normal" failure and doesn't need to be reported.

The other issue with the arrangement as written here is that it treats a 500 from the people API as a missing user, which will cause us to disable clients.

So I think the right thing to do here is to propagate this error out of `getUser`, and then down where it says `// it is likely in the service logs` use `monitor.reportError` to put it in the service logs.

helfi92:
> The other issue with the arrangement as written here is that it treats a 500 from the people API as a missing user, which will cause us to disable clients.

That's a really good observation!

services/web-server/src/login/strategies/github.js
```diff
@@ -129,9 +131,16 @@ module.exports = class Github {
...
             accessToken,
           }, true);
-          const user = await this.getUser({ username: profile.username, userId: Number(profile.id) });
-
-          if (!user) {
+          const [userErr, user] = await tryCatch(
+            this.getUser({ username: profile.username, userId: Number(profile.id) })
+          );
+
+          if (userErr || !user) {
```
djmitche: I think we just want `if (userErr)` here -- the `!user` case is "normal operation"

services/web-server/src/login/strategies/mozilla-auth0.js
```diff
@@ -192,9 +193,13 @@ module.exports = class MozillaAuth0 {
...
-          const user = await this.getUser({ userId: profile.user_id });
+          const [userErr, user] = await tryCatch(this.getUser({ userId: profile.user_id }));
 
-          if (!user) {
+          if (userErr || !user) {
```
djmitche: same here

PR #1593, opened 2019-10-12
visually divide API references into categories

### LINKED ISSUE #1577: Visually divide API references into categories
**Reserved** for [outreachy](https://www.outreachy.org/) candidates. In order for this to be assigned to you, make sure to complete the first step of the application process outlined in the [project description](https://www.outreachy.org/december-2019-to-march-2020-internship-round/communities/mozilla/#build-a-component-that-can-display-a-json-schema).

---

In [bug 1551848](https://bugzilla.mozilla.org/show_bug.cgi?id=1551848) and #1049 , @subhamskm  added a `category` to each API method. That category doesn't show up in the [documentation](https://community-tc.services.mozilla.com/docs/reference/platform/auth/api) though.

It would be great to use these categories as small "headings" in the list of API methods, in cases where there is more than one category (such as the Auth service).  In cases where there is only one category, there should be no change.

Mocked up with GIMP:
sketch

JiwoonKim:
**Closes Issue #1577**
visually divide API references into categories
- use categories as small "headings" in the list of API methods for cases where there is more than one category (such as the Auth service). 
- In cases where there is only one category, there should be no change.

**Changes Applied**
1. Added `groupBy()` function which creates an array of API methods grouped by categories
2. Added `GroupEntry` component which uses conditional rendering to add headings and display lists of API methods
   - if category has only one method, displays single entry usng `entry` component
   - if category has more than one methods, displays heading and list of API methods using `entry` component

**Result**
- before
before

- after
after

### DIFF
--- a/ui/src/views/Documentation/Reference/ApiReference.jsx
+++ b/ui/src/views/Documentation/Reference/ApiReference.jsx
@@ -10,22 +10,10 @@ import Anchor from '../components/Anchor';
 import findRefDoc from '../../../utils/findRefDoc';
 
 class GroupEntry extends Component {
-
   render() {
     const { group, serviceName } = this.props;
     const [groupName, listOfEntries] = group;
-    const isGroupHeaderNeeded = (listOfEntries.length > 1) ? true : false;
 
-    if (!isGroupHeaderNeeded) {
-      const singleEntry = listOfEntries[0];
-      return (
-        <Entry
-          type="function"
-          entry={singleEntry}
-          serviceName={serviceName}
-        />
-      )
-    }
     return (
       <div>
         <br />
@@ -55,15 +43,18 @@ export default class ApiReference extends Component {
 
   groupBy = (list, keyGetter) => {
     const map = new Map();
+
     list.forEach(item => {
       const key = keyGetter(item);
       const collection = map.get(key);
+
       if (!collection) {
         map.set(key, [item]);
       } else {
         collection.push(item);
       }
     });
+
     return map;
   };
 
@@ -81,9 +72,9 @@ export default class ApiReference extends Component {
 
     const functionEntries =
       ref.entries && ref.entries.filter(({ type }) => type === 'function');
-
-    const groupedEntries = 
-      Array.from(this.groupBy(functionEntries, entry => entry.category));
+    const groupedEntries = Array.from(
+      this.groupBy(functionEntries, entry => entry.category)
+    );
 
     return (
       <div>
@@ -102,10 +93,10 @@ export default class ApiReference extends Component {
             </Typography>
             <br />
             {groupedEntries.map(group => (
-              <GroupEntry 
+              <GroupEntry
                 key={group[0]}
-                group={group} 
-                serviceName={ref.serviceName} 
+                group={group}
+                serviceName={ref.serviceName}
               />
             ))}
           </Fragment>
--- a/ui/src/views/Documentation/Reference/ApiReference.jsx
+++ b/ui/src/views/Documentation/Reference/ApiReference.jsx
@@ -9,29 +9,6 @@ import HeaderWithAnchor from '../components/HeaderWithAnchor';
 import Anchor from '../components/Anchor';
 import findRefDoc from '../../../utils/findRefDoc';
 
-class GroupEntry extends Component {
-  render() {
-    const { group, serviceName } = this.props;
-    const [groupName, listOfEntries] = group;
-
-    return (
-      <div>
-        <br />
-        <HeaderWithAnchor type="h4">{groupName}</HeaderWithAnchor>
-        <br />
-        {listOfEntries.map(entry => (
-          <Entry
-            key={`${entry.name}-${entry.query}`}
-            type="function"
-            entry={entry}
-            serviceName={serviceName}
-          />
-        ))}
-      </div>
-    );
-  }
-}
-
 @withRouter
 export default class ApiReference extends Component {
   static propTypes = {
@@ -93,11 +70,19 @@ export default class ApiReference extends Component {
             </Typography>
             <br />
             {groupedEntries.map(group => (
-              <GroupEntry
-                key={group[0]}
-                group={group}
-                serviceName={ref.serviceName}
-              />
+              <div key={group[0]}>
+                <HeaderWithAnchor type="h4">{group[0]}</HeaderWithAnchor>
+                <br />
+                {group[1].map(entry => (
+                  <Entry
+                    key={`${entry.name}-${entry.query}`}
+                    type="function"
+                    entry={entry}
+                    serviceName={ref.serviceName}
+                  />
+                ))}
+                <br />
+              </div>
             ))}
           </Fragment>
         )}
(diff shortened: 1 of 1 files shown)


JiwoonKim:
@djmitche @helfi92 Should it be better to separate out the `GroupEntry` component into a separate file? (I'm not sure if I am allowed to create a new file...)

djmitche:
Oh, and please have a look at the CI failures -- i suspect they are related to lint errors that can be fixed with `yarn lint --fix`.

JiwoonKim:
@djmitche Thanks for the feedback! 

> The expected distinction here is whether the entire set of API methods has multiple categories, or just one. What you've implemented is whether the category has multiple entries.

I'm sorry I misunderstood the issue. I think I got a bit confused seeing the mock up GIMP and assumed the issue was about multiple entries. 

Just to make sure I haven't misunderstood the request this time,
> But, looking at the API definitions, I see that all of them have at least two categories, so let's drop this distinction entirely (so, removing the isGroupHeaderNeeded stuff).

So the result UI should display two categories in case of auth api reference page: `ping server` and `auth service`. Is my understanding correct?

djmitche:
Yes :)

imbstack:
Just dropping by to say that this is super cool

ui/src/views/Documentation/Reference/ApiReference.jsx
```diff
@@ -9,6 +9,29 @@ import HeaderWithAnchor from '../components/HeaderWithAnchor';
...
 
+class GroupEntry extends Component {
+  render() {
+    const { group, serviceName } = this.props;
+    const [groupName, listOfEntries] = group;
+
+    return (
+      <div>
+        <br />
+        <Typography> {groupName} </Typography>
```
helfi92: Use the header with anchor component instead to make it linkable :)

```jsx
<HeaderWithAnchor type="h4">{groupName}</HeaderWithAnchor>
```

ui/src/views/Documentation/Reference/ApiReference.jsx
```diff
@@ -9,6 +9,29 @@ import HeaderWithAnchor from '../components/HeaderWithAnchor';
...
+            key={`${entry.name}-${entry.query}`}
+            type="function"
+            entry={entry}
+            serviceName={serviceName}
+          />
+        ))}
+      </div>
+    );
+  }
+}
```
helfi92: We can inline all of this inside the render method of ApiReference. No need to create a new class.

ui/src/views/Documentation/Reference/ApiReference.jsx
```diff
@@ -18,6 +41,23 @@ export default class ApiReference extends Component {
     apiVersion: string.isRequired,
   };
 
+  groupBy = (list, keyGetter) => {
+    const map = new Map();
```
helfi92: Nice usage of Map :)

djmitche: In particular, I'm pleased that Map retains the order in which its keys were inserted, so that we can adjust the order of categories in the display by adjusting the order in the reference files.

JiwoonKim: Thanks for the review! I had fun doing this part because it felt like solving an algorithm problem. It helped me realize that algorithm problems can be applied to real-world solutions and that was so exciting!

ui/src/views/Documentation/Reference/ApiReference.jsx
JiwoonKim: Applied the changes!

ui/src/views/Documentation/Reference/ApiReference.jsx
JiwoonKim: Applied the changes!

JiwoonKim:
> Submitting the task to Taskcluster failed. Details

Hmm.. I think I rebased mychanges on the latest master using `git fetch upstream` where upstream is set to the original taskcluster repository. Maybe I'm missing something?

djmitche:
```
git fetch upstream
git rebase upstream/master
```

JiwoonKim:
@djmitche Thanks! Rebasing has been a bit confusing... but I think I'm getting the hang of it!

ui/src/views/Documentation/Reference/ApiReference.jsx
```diff
@@ -49,13 +69,20 @@ export default class ApiReference extends Component {
...
             <br />
-            {functionEntries.map(entry => (
-              <Entry
-                key={`${entry.name}-${entry.query}`}
-                type="function"
-                entry={entry}
-                serviceName={ref.serviceName}
-              />
+            {groupedEntries.map(group => (
+              <div key={group[0]}>
```
helfi92: Use a `Fragment` instead of a `div`. This will make sure the following style is applied:

https://github.com/taskcluster/taskcluster/blob/dc700ea3b0690967fcf0af585d9b34cf60307a7d/ui/src/views/Documentation/components/HeaderWithAnchor/index.jsx#L12-L14

ui/src/views/Documentation/Reference/ApiReference.jsx
```diff
@@ -49,13 +69,20 @@ export default class ApiReference extends Component {
...
+                <HeaderWithAnchor type="h4">{group[0]}</HeaderWithAnchor>
+                <br />
```
helfi92: No need for a `<br />`.

ui/src/views/Documentation/Reference/ApiReference.jsx
JiwoonKim: Applied the Changes!

ui/src/views/Documentation/Reference/ApiReference.jsx
JiwoonKim: Applied the changes!

djmitche:
@JiwoonKim now that this has landed, one thing you could do is look at all of the API reference pages and see if the categories seem awkward or wrong, and make a patch to change that.  Aside from the Auth service, I think all of the other services basically have two categories (ping and everything else).

JiwoonKim:
> @JiwoonKim now that this has landed, one thing you could do is look at all of the API reference pages and see if the categories seem awkward or wrong, and make a patch to change that. Aside from the Auth service, I think all of the other services basically have two categories (ping and everything else).

Sure, that sounds fun! I'll look into all of the API reference pages and check for the categories. Would I have to compare the API reference pages' methods against the `generated.references.json` file?

djmitche:
Those two things will match.  The source of that information is in `services/*/src/api.js` for each service. The command `git grep category: services/` will find most of the occurrences.

JiwoonKim:
Thanks! @djmitche 
Just one more thing. I think my pull request has been merged into the master, but I'm not seeing any changes on the actual deployed page (not sure which one is the final deployed version...)
- https://docs.taskcluster.net/docs/reference/platform/auth/api 
or 
- https://taskcluster-ui.herokuapp.com/docs/reference/platform/auth/api . 

Is there something I am missing?

djmitche:
I see it now on the second link.  Hopefully the first will update soon, too.  If not, we'll need to take a look at what's wrong.

JiwoonKim:
> BTW it looks like I mis-remembered, and even the Auth service only has two categories. I think that could be divided into "Roles", "Clients", etc. Want to give it a shot, based on reading the names and descriptions of the API methods?

Yep, I'd love to give it a shot! Since this pull request has been closed, should I create a new separate issues and pull request about dividing further api references?

helfi92:
> I would love to further give it a shot on this issue! Since this pull request has been closed, should I create a new separate issue and pull request about further dividing the api references?

Yes, please :)

PR #1595, opened 2019-10-12
Remove old Task Creator link from 'hello-world' tutorial

clrmachado:
Removed [https://tc.example.com/tasks/create] from Hello-World tutorial and added a reference to the Task-Creator tool instead.


clrmachado:
> Can you include the link for current one . Would seem more appropriate.

Hello @vishakhanihore, thanks for the review. 
Sorry, I think I was not very clear on my PR description. I've discussed this change with @djmitche and he mentioned that due to some changes in the project it is still not clear to where will the docs refer. That is why I've removed the link.

vishakhanihore:
Then its Lgtm!

djmitche:
Thanks!

PR #1603, opened 2019-10-13
Tab focus works now.

devnoorfatima:

### DIFF
--- a/ui/src/components/WMWorkerPoolEditor/index.jsx
+++ b/ui/src/components/WMWorkerPoolEditor/index.jsx
@@ -296,7 +296,11 @@ export default class WMWorkerPoolEditor extends Component {
       onDialogActionComplete,
     } = this.props;
     const { workerPool, error, actionLoading, validation } = this.state;
-
+    const options = {
+      extraKeys: {
+        Tab: false,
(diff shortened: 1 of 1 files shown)


ui/src/components/WMWorkerPoolEditor/index.jsx
```diff
@@ -397,6 +402,7 @@ export default class WMWorkerPoolEditor extends Component {
               value={JSON.stringify(workerPool.config, null, 2)}
               onChange={this.handleEditorChange}
               lint
+              options={options}
```
helfi92: Good job on making it work :) I think it would make sense to move this upstream (i.e., in https://github.com/mozilla-frontend-infra/components) since it would make the CodeEditor more accessible. Once we move it there, we could release a new version of `@mozilla-frontend-infra/components` and then bump the version in this repo to receive the new changes. Feel free to create a PR directly there :)

helfi92: How is it going @devnoorfatima?

helfi92:
This was implemented upstream in @mozilla-frontend-infra/components.

PR #1604, opened 2019-10-13
[UI] Invite user to login in /profile when logged out #1581

km-js:
I added the message which the user can see while logged out. Please let me know if we need any other additions to this.

### DIFF
--- a/ui/src/views/Profile/index.jsx
+++ b/ui/src/views/Profile/index.jsx
@@ -37,7 +37,7 @@ export default class Profile extends Component {
       <Dashboard title="Profile">
         {!currentScopes && loading && <Spinner loading />}
         <ErrorPanel fixed error={error} />
-        {!user && (
+        {!user && !loading && (
           <Typography variant="subtitle1">
             Sign in to view your Profile
           </Typography>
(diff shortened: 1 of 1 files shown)


ui/src/views/Profile/index.jsx
```diff
@@ -37,6 +37,9 @@ export default class Profile extends Component {
       <Dashboard title="Profile">
         {!currentScopes && loading && <Spinner loading />}
         <ErrorPanel fixed error={error} />
+        <Typography variant="subtitle1">
+          Sign in to view your Profile
```
helfi92: "Profile" can be lower case.

ui/src/views/Profile/index.jsx
```diff
@@ -37,6 +37,11 @@ export default class Profile extends Component {
       <Dashboard title="Profile">
         {!currentScopes && loading && <Spinner loading />}
         <ErrorPanel fixed error={error} />
+        {!user && (
```
helfi92: `{!user && !loading && (`

ui/src/views/Profile/index.jsx
helfi92: This is still not done. `profile` instead of `Profile`.

km-js: Sorry for that, it's done now!

PR #1605, opened 2019-10-14
Bug 1585102 - improve InsufficientScopes error from tc-github

djmitche:
Bugzilla Bug: [1585102](https://bugzilla.mozilla.org/show_bug.cgi?id=1585102)

The GitHub service now posts a more useful comment to pull requests and commits when an InsufficientScopes error occurs.

### DIFF
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -217,7 +217,35 @@ class Handlers {
       credentials: this.context.cfg.taskcluster.credentials,
     });
     for (const t of tasks) {
-      await scopedQueueClient.createTask(t.taskId, t.task);
+      try {
+        await scopedQueueClient.createTask(t.taskId, t.task);
+      } catch (err) {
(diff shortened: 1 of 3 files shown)


services/github/src/handlers.js
```diff
@@ -217,7 +217,35 @@ class Handlers {
...
+            '```',
+            '',
+            'The expansion of these scopes is not sufficient to create the task and is missing',
+            'scopes satisfying the following expression:',
+            '',
+            '```',
+            JSON.stringify(err.details.unsatisfied, null, 2),
+            '```',
+            '',
+            'The `queue.createTask` call requires scopes satisfying the following expression:',
```
imbstack: Is `queue.createTask` kinda an internal detail they don't care about?

djmitche: I think it's an important breadcumb in trying to figure out what went wrong, if the scopes themselves are not enough.

PR #1644, opened 2019-10-15
[worker-manager] Make AWS provider capacity not be per-region.

tomprince:


tomprince:
This is fixed by #1528

Issue #1398, opened 2019-09-25
"Compare Scopesets" and "Expand Scopesets" should have a linkable URL


djmitche:
**Reserved** for [outreachy](https://www.outreachy.org/) candidates. In order for this to be assigned to you, make sure to complete the first step of the application process outlined in the [project description](https://www.outreachy.org/december-2019-to-march-2020-internship-round/communities/mozilla/#build-a-component-that-can-display-a-json-schema).

---

When working with scopes in a chat conversation, it would be useful to be able to link to these two tools with their values filled in already.  For example, I should be able to enter some scopes into the Scope Expander, click expand, and then copy the URL bar and paste it into a chat.  Currently, the URL does not change.

mkneumann777:
I would like to try and tackle this 'bug.'

djmitche:
All yours!

djmitche:
You can find the scopesets utilities under "Authorization" section of the menu. It might help to read a little bit about scopes in the documentation to understand what these tools do.

As for the UI changes, if you search for task MSjwqraNRdS4Q4i3CSLjuw in the task inspector, you'll see that the URL changes to include the taskId, e.g., `https://localhost:5080/tasks/MSjwqraNRdS4Q4i3CSLjuw`.  We want to do the same kind of thing for these tools, so that I, as a user, can come and type some stuff into the boxes and then easily send a link to another person to show them what I'm seeing.

djmitche:
@mkneumann777 are you still working on this?

mkneumann777:
Yes, I am still working towards solving this issue. I just finished up my Web Developer Bootcamp and started working through a React.js course so I can have an understanding of JSX and React.js as a whole.

mkneumann777:
I have made it through about 25% of my React course at this time. However, at this point, I don't think I'm any closer to solving this issue. From what I currently understand is that I need to get `scopeTextA` & `scopeTextB` values and update the URL params. Then when that URL is executed the application takes said params values and plugs them into the appropriate places.

djmitche:
You've got it so far!

One good technique is to look for other examples of similar functionality.  One such example is the pulse inspector -- https://taskcluster-ui.herokuapp.com/pulse-messages?bindings%5B0%5D%5Bexchange%5D=abc&bindings%5B0%5D%5Bpattern%5D=%23

mkneumann777:
[...]

PR #1661, opened 2019-10-16
Bug #1398 "Compare Scopesets" and "Expand Scopesets" should have a linkable URL

mkneumann777:
I have developed a solution. First I used `history.push() `on the `this.props` object to explicitly return an object with 2 properties within the `if` statement in `handleCompareScopesClick()` method. I also accounted for comma use within scopes as seen in the example:

```
[
  "queue:create-task:aws-provisioner-v1/tutorial",
  "queue:route:index.garbage.*",
]
```

Within the `handleCompareScopesClick()` method:

```
handleCompareScopesClick = async () => {
  ...
  const scopeParamA = scopesA.join('%');
  const scopeParamB = scopesB.join('%');

  this.props.history.push({ 
    pathname: '/auth/scopes/compare',
    search: `?q1=${scopeParamA}&q2=${scopeParamB}`,
  });
...
}
```

This will push a new query string to the Url where each line is delimited by the '%' symbol. Now, in order for a user to share the Url. All the user has to do is share the Url in the address bar. I implemented a lifecycle method of `componentDidMount()` to take the Url parameters and set the state:

The `replace()` method is used to change all occurrences of the '%' within the query string to the newline syntax - '\n' this will place all text in the correct format.

```
componentDidMount() {
  const searchParams = new URLSearchParams(this.props.location.search);
  const scopesA = searchParams.get('qA');
  const scopesB = searchParams.get('qB');

  if (scopesA && scopesB) {
    this.setState(() => ({
      scopeTextA: scopesA.replace(/%/g, '\n'),
      scopeTextA: scopesB.replace(/%/g, '\n'),
    }));
  }
}
```

This will take any query values and pass them into the state properties. All the user has to do at this point is to click the "Compare Scopes" button. *I have also added the same functionality to the Scopeset Expander.*

### DIFF
--- a/ui/src/views/Scopes/ScopesetComparison/index.jsx
+++ b/ui/src/views/Scopes/ScopesetComparison/index.jsx
@@ -12,6 +12,38 @@ import Dashboard from '../../../components/Dashboard/index';
 import Button from '../../../components/Button/index';
 import splitLines from '../../../utils/splitLines';
 
+const getScopesetDiff = (scopesA, scopesB) => {
+  const scopesUnion = scopeUnion(scopesA, scopesB);
+  const scopesetDiff = [];
+
+  scopesUnion.forEach(scope => {
+    const s1 = scopeIntersection([scope], scopesA);
+    const s2 = scopeIntersection([scope], scopesB);
+
+    scopesetDiff.push([s1, s2]);
+  });
+
+  return scopesetDiff;
+};
+
+const getCellColors = scopesetDiff => {
+  const cellColors = [];
+
+  scopesetDiff.forEach(([s1, s2]) => {
+    if (s1.length === 0 && s2.length) {
+      cellColors.push(['', 'greenCell']);
+    } else if (s1.length && s2.length === 0) {
+      cellColors.push(['redCell', '']);
+    } else if (!equals(s1, s2)) {
+      cellColors.push(['yellowCell', 'yellowCell']);
+    } else {
+      cellColors.push(['', '']);
+    }
+  });
+
+  return cellColors;
+};
+
 @hot(module)
 @withStyles(theme => ({
   actionButton: {
@@ -37,10 +69,29 @@ import splitLines from '../../../utils/splitLines';
   },
 }))
 export default class ScopesetComparison extends Component {
-  state = {
-    scopeTextA: '',
-    scopeTextB: '',
-  };
+  constructor(props) {
+    super(props);
+
+    const query = parse(this.props.location.search.slice(1));
+    const { scopesA, scopesB } = query;
+
+    if (scopesA && scopesB) {
+      const scopesetDiff = getScopesetDiff(scopesA, scopesB);
+      const cellColors = getCellColors(scopesetDiff);
+
+      this.state = {
+        scopeTextA: scopesA.join('\n'),
+        scopeTextB: scopesB.join('\n'),
+        scopesetDiff,
+        cellColors,
+      };
+    } else {
+      this.state = {
+        scopeTextA: '',
+        scopeTextB: '',
+      };
+    }
+  }
 
   handleScopesAChange = scopeTextA => {
     this.setState({ scopeTextA });
@@ -50,31 +101,14 @@ export default class ScopesetComparison extends Component {
     this.setState({ scopeTextB });
   };
 
-  componentDidMount() {
-    const query = parse(this.props.location.search.slice(1));
-    const { scopesA, scopesB } = query;
-
-    if (scopesA && scopesB) {
-      const scopesetDiff = this.getScopesetDiff(scopesA, scopesB);
-      const cellColors = this.getCellColors(scopesetDiff);
-
-      this.setState(() => ({
-        scopeTextA: scopesA.join('\n'),
-        scopeTextB: scopesB.join('\n'),
-        scopesetDiff,
-        cellColors,
-      }));
-    }
-  }
-
   handleCompareScopesClick = async () => {
     const { scopeTextA, scopeTextB } = this.state;
 
     if (scopeTextA && scopeTextB) {
       const scopesA = splitLines(scopeTextA);
       const scopesB = splitLines(scopeTextB);
-      const scopesetDiff = this.getScopesetDiff(scopesA, scopesB);
-      const cellColors = this.getCellColors(scopesetDiff);
+      const scopesetDiff = getScopesetDiff(scopesA, scopesB);
+      const cellColors = getCellColors(scopesetDiff);
       const queryObj = { scopesA, scopesB };
       const queryStr = stringify(queryObj);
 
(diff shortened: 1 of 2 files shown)


ui/src/views/Scopes/ScopesetComparison/index.jsx
```diff
@@ -57,6 +71,13 @@ export default class ScopesetComparison extends Component {
       const scopesB = splitLines(scopeTextB);
       const scopesetDiff = this.getScopesetDiff(scopesA, scopesB);
       const cellColors = this.getCellColors(scopesetDiff);
+      const scopeParamA = scopesA.join('%');
+      const scopeParamB = scopesB.join('%');
+
+      this.props.history.push({
+        pathname: '/auth/scopes/compare',
+        search: `?qA=${scopeParamA}&qB=${scopeParamB}`,
+      });
```
helfi92: This logic seems to work nicely, however, it's a bit inconsistent with the way we parse/stringify query parameters in the rest of the site. Let's use the [`qs`](https://yarnpkg.com/en/package/qs) package to simplify things.

```jsx
import { parse, stringify } from 'qs';
```

We use the `qs` package in the following places:

```
src/components/ClientForm/index.jsx
src/utils/isThirdPartyLogin.js
src/views/Clients/ViewClients/index.jsx
src/views/Hooks/ListHooks/index.jsx
src/views/Provisioners/ViewWorkers/index.jsx
src/views/PulseMessages/index.jsx
src/views/ThirdPartyLogin/index.jsx
src/views/WorkerManager/WMViewWorkerPools/index.jsx
```

Feel free to use to look at these files for inspiration.

ui/src/views/Scopes/ScopesetExpander/index.jsx
```diff
@@ -41,8 +41,25 @@ export default class ScopesetExpander extends Component {
...
+    const searchParams = new URLSearchParams(this.props.location.search);
+    const scope = searchParams.get('q');
+
+    if (scope) {
+      this.setState(() => ({
+        scopeText: scope.replace(/%/g, '\n'),
+      }));
+    }
+  }
+
```
helfi92: Same thing here. Let's use the `qs` package.

ui/src/views/Scopes/ScopesetComparison/index.jsx
mkneumann777: I have implemented the `qs` package/lib. I found this quite neat because it allowed me to clean up the code a bit. In addition, I was able to incorporate some of the useful new syntaxes I have been learning about such as Object Destructuring and passing self-named object properties by themselves.

I also added the functionality that was suggested. The `componentDidMount()` will now additionally set the state for the `scopesetDiff` and `cellColors`.

The same `qs` implementation can be seen in both `ScopesetComparision` & `ScopesetExpander`. I had to change some variable naming due to namespace collisions.

ui/src/views/Scopes/ScopesetComparison/index.jsx
```diff
@@ -49,6 +50,23 @@ export default class ScopesetComparison extends Component {
...
+      const cellColors = this.getCellColors(scopesetDiff);
+
+      this.setState(() => ({
+        scopeTextA: scopesA.join('\n'),
+        scopeTextB: scopesB.join('\n'),
+        scopesetDiff,
+        cellColors,
+      }));
+    }
+  }
```
helfi92: The React [docs](https://reactjs.org/docs/react-component.html#componentdidmount) don't suggest calling `this.setState` in `componentDidMount` for performance reasons. We can avoid this by having this logic in the constructor.

```jsx
constructor(props) {
  super(props);
  
  // ...
  
  this.state = {
    // ...
  };
}
```

_Note: We might not be able to call `this.<something>` inside the constructor so we would need to extract things like `getScopesetDiff` and  `getCellColors` to be outside the class. Putting them after the imports lines of this file sounds good to me._

ui/src/views/Scopes/ScopesetComparison/index.jsx
```diff
@@ -99,7 +124,7 @@ export default class ScopesetComparison extends Component {
     const { scopeTextA, scopeTextB, scopesetDiff, cellColors } = this.state;
 
     return (
-      <Dashboard title="Compare Scopes">
+      <Dashboard title="Compare Scopesets">
```
helfi92: I think we can keep whatever the latest master has this value as.

ui/src/views/Scopes/ScopesetExpander/index.jsx
```diff
@@ -41,8 +42,26 @@ export default class ScopesetExpander extends Component {
...
+  componentDidMount() {
+    const query = parse(this.props.location.search.slice(1));
+    const { scopes } = query;
+
+    if (scopes) {
+      this.setState(() => ({
+        scopeText: scopes.join('\n'),
+      }));
+    }
+  }
```
helfi92: Same thing as previous comment about `componentDidMount`.

ui/src/views/Scopes/ScopesetComparison/index.jsx
mkneumann777: I have removed the need for the `componentDidMount()` lifecycle method and replaced it with the constructor method.

ui/src/views/Scopes/ScopesetComparison/index.jsx
mkneumann777: I don't ever recall changing this title, but I have reverted the change.

ui/src/views/Scopes/ScopesetExpander/index.jsx
mkneumann777: I have removed the `componentDidMount()` and replaced it with a `constructor` here as well.

ui/src/views/Scopes/ScopesetExpander/index.jsx
```diff
@@ -59,7 +80,7 @@ export default class ScopesetExpander extends Component {
 
     return (
       <Dashboard
-        title="Expand Scopes"
+        title="Expand Scopesets"
```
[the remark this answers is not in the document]
mkneumann777: I am unsure of how to rebase on the master given the fact that I am working on master.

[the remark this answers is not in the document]
helfi92: You can try doing `git pull git@github.com:taskcluster/taskcluster.git master --rebase`.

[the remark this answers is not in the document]
mkneumann777: I tried your mentioned command but was denied because of permissions. However, I did re-run a modified version of the command: `https://github.com/taskcluster/taskcluster.git master --rebase` and it did pull and apply the commits to the `top,` but I'm not seeing a change.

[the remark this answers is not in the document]
helfi92: Interesting. Maybe being on the "master" branch is not helping. Feel free to close this PR and open a new one on a branch that isn't master. I think that might help.

PR #1668, opened 2019-10-17
[bug 1588625] Add workerConfig to aws provider as well

imbstack:
This works in concert with taskcluster/taskcluster-worker-runner#51.

This was mostly me guessing so feel free to yell that we should do this differently!

### DIFF
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -102,7 +102,7 @@ class AwsProvider extends Provider {
         return await workerPool.reportError({
           kind: 'creation-error',
           title: 'User Data is too long',
-          description: 'Try a shorter workerPoolId and/or a shorter rootUrl',
+          description: 'Try removing some workerConfiguration and consider putting it in a secret',
           notify: this.notify,
           WorkerPoolError: this.WorkerPoolError,
         });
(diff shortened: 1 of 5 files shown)


services/worker-manager/src/providers/aws.js
```diff
@@ -107,6 +89,25 @@ class AwsProvider extends Provider {
...
+        workerGroup: this.providerId,
+        workerConfig: config.launchConfig.workerConfig || {},
+      }));
+      // https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-add-user-data
+      // The raw data should be 16KB maximum
+      if (userData.length > 16384) {
+        return await workerPool.reportError({
+          kind: 'creation-error',
+          title: 'User Data is too long',
+          description: 'Try a shorter workerPoolId and/or a shorter rootUrl',
```
djmitche: Maybe better advice here is "try removing some workerConfiguration and consider putting it in a secret"?

services/worker-manager/src/providers/aws.js
```diff
@@ -107,6 +89,25 @@ class AwsProvider extends Provider {
         ts.ResourceType === 'instance' ? instanceTags.concat(ts.Tags) : otherTagSpecs.push(ts)
       );
 
+      const userData = Buffer.from(JSON.stringify({
```
owlishDeveloper: Why was this moved?

imbstack: Because the workerconfig comes from the launchconfigs so it needs to be inside the loop

Issue #1633, opened 2019-10-15
[UI] Improve user experience of a paginated table (ConnectionDataTable)


helfi92:
**Describe the bug**
**Reserved** for [outreachy](https://www.outreachy.org/) candidates. In order for this to be assigned to you, make sure to complete the first step of the application process outlined in the [project description](https://www.outreachy.org/december-2019-to-march-2020-internship-round/communities/mozilla/#build-a-component-that-can-display-a-json-schema).

---

Knowing when a table is paginated is not straightforward. For example, navigating to 
https://taskcluster-ui.herokuapp.com/auth/clients requires the user to scroll all the way down to know this table is paginated. Another issue is that a user that tries to search for a client using the browser's "CMD + F" page search feature might not find what they're looking for if the search term happens to be in a different page. For big paginated tables, users need to know right away that a table is paginated without having to scroll down to the end.

**To Reproduce**
1. Go to https://taskcluster-ui.herokuapp.com/auth/clients

**Expected behavior**
A user knows when a table is paginated without having to scroll all the way down to the end.

JiwoonKim:
Hi! I've completed all the tasks assigned to me at the moment. May I take this issue?

JiwoonKim:
@djmitche @helfi92 
hmm... I'm trying to get the web-server running but it seems to get this error that I should provide `option.accountId`. The web-server was working fine until I rebased my code on the upstream/master. Maybe I'm missing something?

<img width="700" alt="error" src="https://user-images.githubusercontent.com/29671309/66859406-34ce2900-efc6-11e9-8ca0-f479e642a737.PNG">

helfi92:
It will be fixed once https://github.com/taskcluster/taskcluster/pull/1637#pullrequestreview-302126230 is merged :) Sorry about that.

djmitche:
(I assigned #1246 to @JiwoonKim so no need here)

JiwoonKim:
No problem! solved it after rebasing the code based on #1637 :)

This sounds like an interesting task. 
So I'm thinking a simple solution would be to maybe bring the pages back and forth button (< >), which is at the bottom, to the top of the page? But while this might solve the first problem, I'm thinking it might not help for the second problem for searching a specific client. Should I think of an entirely different way to figure out the ui?

JiwoonKim:
[...]

PR #1680, opened 2019-10-18
[UI] Improve ux of a paginated table (ConnectionDataTable)

JiwoonKim:
**Closes Issue** #1633 
for big paginated tables, users need to know right away that a table is paginated without having to scroll down to the end

**Changes Applied**
add the pagination both on top and bottom

1. create boolean prop `withTopPagination` which allows larger tables like https://taskcluster-ui.herokuapp.com/auth/clients, that take most of the screen, to be rendered with both top and bottom paginations
   - small tables like the artifacts table in https://taskcluster-ui.herokuapp.com/tasks/f2WzQXE5RUSVDjPrHZubCQ remain the same as before, having only one pagination at the bottom
2. apply top pagination feature for clients view table

**Results**
client view table now has pagination on top and bottom

after1


JiwoonKim:
@helfi92 
I've only implemented the top pagination feature for the client view table, for now. 
If the implementation seems good enough, maybe I can also add the top paginations to the following pages as well
- role view table: https://taskcluster-ui.herokuapp.com/auth/roles
- scope view table (on the client tab): https://taskcluster-ui.herokuapp.com/auth/scopes
- secrets view table: https://taskcluster-ui.herokuapp.com/secrets

ui/src/components/ConnectionDataTable/index.jsx
```diff
@@ -87,6 +97,13 @@ export default class ConnectionDataTable extends Component {
      * A list of header names to use on the table starting from the left.
      */
     headers: arrayOf(string),
+    /**
+     * Whether table should display additional pagination on top.
+     * At default false value, table pagination is displayed on bottom.
+     * If indicator is true, table pagination is displayed on both top
+     * and bottom.
+     */
```
helfi92:
```
/**
 * If true, an additional pagination component will be displayed
 * at the top of the table.
 */
```

ui/src/components/ConnectionDataTable/index.jsx
```diff
@@ -150,23 +167,55 @@ export default class ConnectionDataTable extends Component {
     });
   };
 
+  renderConnectionDataTable = (colSpan, count) => {
```
helfi92: Rename to `renderTablePagination`.

ui/src/components/ConnectionDataTable/index.jsx
```diff
@@ -37,6 +46,7 @@ export default class ConnectionDataTable extends Component {
     sortDirection: 'desc',
     headers: null,
     onHeaderClick: null,
+    withTopPagination: false,
```
helfi92: Now that I think about it, let's set to `true` and set `withTopPagination` to false in `TaskRunsCard/index.jsx` only.

ui/src/components/ConnectionDataTable/index.jsx
JiwoonKim: Applied changes!

ui/src/components/ConnectionDataTable/index.jsx
JiwoonKim: Applied Changes!

ui/src/components/ConnectionDataTable/index.jsx
JiwoonKim: Applied changes!

helfi92: Is it possible you forgot to push your changes? :) Also, I think it would be better if we rename the prop to `withoutTopPagination` and keep the default value to `false`.

JiwoonKim: Sorry about that! I've been trying to re-organize the commits and messages before I fully push the changes. It's all implemented now and the commits are also organized :)
Also, applied the changes to use `withoutTopPagination` as well!

PR #1693, opened 2019-10-18
Bug1589449

owlishDeveloper:
Bugzilla Bug: [1589449](https://bugzilla.mozilla.org/show_bug.cgi?id=1589449)


services/worker-manager/src/providers/aws.js
```diff
@@ -254,6 +254,34 @@ class AwsProvider extends Provider {
...
   }
 
+  async removeWorker({worker}) {
+    let result;
+    try {
+      result = await this.ec2s[worker.providerData.region].terminateInstances({
+        InstanceIds: [worker.workerId]
+      }).promise();
+    } catch (e) {
+      this.monitor.err(`Error terminating AWS instance: ${e}`);
```
djmitche: The call to `workerPool.reportError` will log the error as a `workerError` message, so I think this is redundant.

services/worker-manager/src/providers/aws.js
```diff
@@ -254,6 +254,34 @@ class AwsProvider extends Provider {
...
+
+      return await workerPool.reportError({
+        kind: 'termination-error',
+        title: 'Instance Termination Error',
+        description: e.message,
+        notify: this.notify,
+        WorkerPoolError: this.WorkerPoolError,
+      });
```
djmitche: This is awesome.

services/worker-manager/src/providers/aws.js
```diff
@@ -254,6 +254,34 @@ class AwsProvider extends Provider {
...
+
+      throw e;
```
djmitche: This will result in a 500 Internal Server Error to the caller.  Is that the best approach?  Should we maybe just return success?

owlishDeveloper: this is a failure path tho..... return success even from here? how the caller would know the call has fail then?

services/worker-manager/src/providers/aws.js
imbstack: We also don't want to necessarily report an error at all. If the instance is already gone, this should just be a noop. So I think a 404 or whatever is fine.

services/worker-manager/src/providers/aws.js
imbstack: agreed. just do nothing here

services/worker-manager/src/providers/aws.js
```diff
@@ -254,6 +254,39 @@ class AwsProvider extends Provider {
...
+    }
+
+    result.TerminatingInstances.forEach(ti => {
+      if (!ti.InstanceId === worker.workerId || !ti.CurrentState.Name === 'shutting-down') {
```
imbstack: When would the first half of this be untrue? If it can be false, let's have a different error since this error won't actually say what happened. Will either of these ever be true actually?

services/worker-manager/src/providers/aws.js
```diff
@@ -254,6 +252,37 @@ class AwsProvider extends Provider {
...
+    let result;
+    try {
+      result = await this.ec2s[worker.providerData.region].terminateInstances({
+        InstanceIds: [worker.workerId],
+      }).promise();
+    } catch (e) {
+      const workerPool = this.WorkerPool.load({
+        workerPoolId: worker.workerPoolId,
+      });
+      await workerPool.reportError({
```
imbstack: Ah, my comment here became outdated. I think we should check to see if the instance is already gone here and if so, not report an error. The requested state is already true in that case

services/worker-manager/src/providers/aws.js
owlishDeveloper: Actually, I just realized that in the similar case in `provision` method we don't rethrow. So why do it here.... I removed that line

services/worker-manager/src/providers/aws.js
owlishDeveloper: If instance is already gone, the aws request won't error, it will be considered a success

services/worker-manager/src/providers/aws.js
owlishDeveloper: I'd say it's highly unlikely.... In case of a bug I guess

> Will either of these ever be true actually?
I don't know. Seemed like a sensical check

owlishDeveloper: I changed the error message to be one combined "Unknown error"

PR #1700, opened 2019-10-20
[UI] Migrate to latest Neutrino

helfi92:


ui/src/components/PageNavigation/index.jsx
```diff
@@ -51,7 +51,7 @@ export default class PageNavigation extends Component {
     return (
       <Link className={classes.link} to={to}>
         <Button variant="outlined" className={classes.button} {...props}>
-          <Fragment>
+          <>
```
djmitche: GitHub thinks this is invalid for some reason (highlighted in red).  Same in a few other spots.  It looks like that's limited to where `<>` is embedded in another element.  Is `<>` necessary in those cases?

helfi92: Good observation! I've added a rule upstream to disallow `<> </>` for now.

PR #1701, opened 2019-10-20
Bug 1577839 - fix handling of custom tags in aws provider

djmitche:
Bugzilla Bug: [1577839](https://bugzilla.mozilla.org/show_bug.cgi?id=1577839)

This fixes the implementation in two ways:
 * looks for `TagSpecifications` in the launchConfig itself
 * includes the custom instance tags in `instanceTags` (`Array.concat` is a pure function)

It also removes the assertions from the fake `runInstances`.  Because errors from `runInstances` are handled internally, any assertion failures are logged (at the NOTICE level) and do not cause a test failure.

### DIFF
--- a/changelog/bug-1577839.md
+++ b/changelog/bug-1577839.md
@@ -1,0 +1,4 @@
+level: patch
+reference: bug 1577839
+---
+Including TagSpecifications in a launchConfig for the AWS provider now applies those tags to the indicated resources.  Before this version, those tags would be ignored.
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -82,12 +82,16 @@ class AwsProvider extends Provider {
       if (toSpawnCounter <= 0) break; // eslint-disable-line
       // Make sure we don't get "The same resource type may not be specified
       // more than once in tag specifications" errors
-      const TagSpecifications = config.TagSpecifications ? config.TagSpecifications : [];
+      const TagSpecifications = config.launchConfig.TagSpecifications || [];
       const instanceTags = [];
       const otherTagSpecs = [];
-      TagSpecifications.forEach(ts =>
-        ts.ResourceType === 'instance' ? instanceTags.concat(ts.Tags) : otherTagSpecs.push(ts)
-      );
+      TagSpecifications.forEach(ts => {
+        if (ts.ResourceType === 'instance') {
+          ts.Tags.forEach(tag => instanceTags.push(tag));
+        } else {
+          otherTagSpecs.push(ts);
+        }
+      });
 
       const userData = Buffer.from(JSON.stringify({
         rootUrl: this.rootUrl,
--- a/services/worker-manager/test/fake-aws.js
+++ b/services/worker-manager/test/fake-aws.js
@@ -1,46 +1,42 @@
-const assert = require('assert');
-
 module.exports = {
   EC2: {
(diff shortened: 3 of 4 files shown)


djmitche:
As I'm looking at this, I see that the fake `runInstances` no longer uses its arguments, so could probably do with some simplification.  What do you think?

changelog/bug-1577839.md
```diff
@@ -0,0 +1,4 @@
+level: patch
+reference: bug 1577839
+---
+Including TagSpecifications in a launchConfig for the AWS provider now applies those tags to the indicated resources.  Before this version, those tags would be ignored.
```
owlishDeveloper: Can we re-phrase this as "Fixes the bug where TagSpecifications in a launchConfig for the AWS provider would be ignored etc" or something like that, because it wasn't deliberate or me not realizing this needed to be done - my intention _was_ to apply those tags, it's just that because of so many `launchConfigs` I didn't see I was making a mistake

services/worker-manager/src/providers/aws.js
```diff
@@ -82,12 +82,16 @@ class AwsProvider extends Provider {
...
       // more than once in tag specifications" errors
-      const TagSpecifications = config.TagSpecifications ? config.TagSpecifications : [];
+      const TagSpecifications = config.launchConfig.TagSpecifications || [];
       const instanceTags = [];
       const otherTagSpecs = [];
-      TagSpecifications.forEach(ts =>
-        ts.ResourceType === 'instance' ? instanceTags.concat(ts.Tags) : otherTagSpecs.push(ts)
-      );
+      TagSpecifications.forEach(ts => {
+        if (ts.ResourceType === 'instance') {
```
owlishDeveloper: why replacing ternary?

services/worker-manager/test/fake-aws.js
```diff
@@ -1,46 +1,42 @@
...
-      assert.deepStrictEqual(launchConfig,
-        {
-          ...defaultLaunchConfig.launchConfig,
-          MinCount: launchConfig.MinCount, // this is estimator's functionality, no need to test this here
-          MaxCount: launchConfig.MaxCount,
-          TagSpecifications,
-          UserData: Buffer.from(JSON.stringify(UserData)).toString('base64'),
+    runInstances: ({defaultLaunchConfig, TagSpecifications, UserData}) => {
+      const fake = launchConfig => {
+        // note that asserting in this function will not cause a test failure, as the error is logged
```
owlishDeveloper: I just tried it and I saw an assertion failure. I think the main problem was that the `instance tags in launch spec - should merge them with our instance tags` test case wasn't implemented rather than the mock design

owlishDeveloper:
Also, you link to 1577839, but that one is the bug for unit tests. This PR fixes a different bug, as you say, handling the user tags. The bug was introduced along with the schema change. This PR adds only one test case..... Not like I am requiring any changes, just a note that this PR is about a different bug, and adds one test case (which indeed was supposed to be part of 1577839)

services/worker-manager/test/fake-aws.js
owlishDeveloper: The above means, can we at least remove the comment? It doesn't seem to be true to me, like I said above. (I also don't quite see the need to rewrite the mock tbh)

djmitche:
Let's just close this up.  I started by writing the test case, and I thought I encountered some bugs, but perhaps not.  At any rate, it's not critical work now.

PR #1703, opened 2019-10-21
Improve error messages for registerWorker with invalid inputs

djmitche:
This came up in slack today as a 500 and I figured we could do better.

I changed 'Token validation' since there's no token being validated, so it's kind of a misleading error message.

### DIFF
--- a/changelog/V9gQQys8RWiQUxYiQwsQSg.md
+++ b/changelog/V9gQQys8RWiQUxYiQwsQSg.md
@@ -1,0 +1,2 @@
+level: silent
+---
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -196,8 +196,8 @@ class AwsProvider extends Provider {
     }
 
     const {document, signature} = workerIdentityProof;
-    if (!document || !signature) {
-      throw new ApiError('Token validation error');
+    if (!document || !signature || !(typeof document === "string")) {
+      throw new ApiError('Request must include both a document (string) and a signature');
     }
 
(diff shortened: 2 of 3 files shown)


djmitche:
https://console.cloud.google.com/errors/CM2Ol4O1uNyw8wE?time=P30D&project=heroku-logging

changelog/V9gQQys8RWiQUxYiQwsQSg.md
```diff
@@ -0,0 +1,2 @@
+level: silent
```
owlishDeveloper: why "silent" and no description?

services/worker-manager/src/providers/aws.js
```diff
@@ -196,8 +196,8 @@ class AwsProvider extends Provider {
     }
 
     const {document, signature} = workerIdentityProof;
-    if (!document || !signature) {
-      throw new ApiError('Token validation error');
+    if (!document || !signature || !(typeof document === "string")) {
+      throw new ApiError('Request must include both a document (string) and a signature');
```
owlishDeveloper: I thought we wanted to keep the error messages vague on purpose? In case of a malicious action

djmitche: The vagueness was about the validity of the document -- this is just about adherence to the expected form of the request, which is public information anyway.

changelog/V9gQQys8RWiQUxYiQwsQSg.md
djmitche: "silent: Minor bugfixes where the bugs did not impact many users"

PR #1711, opened 2019-10-21
Estimator updates to support `capacityPerInstance`

imbstack:
Bugzilla Bug: [1590175](https://bugzilla.mozilla.org/show_bug.cgi?id=1590175)

This involved some changes to how each provider loops over launch configs. I've tested the google provider in dev but it would be nice if @owlishDeveloper or someone can test the aws provider. I can provide an image for whoever wants it so they can avoid building it again themselves.

### DIFF
--- a/services/worker-manager/src/monitor.js
+++ b/services/worker-manager/src/monitor.js
@@ -29,9 +29,9 @@ monitorManager.register({
     pendingTasks: 'The number of tasks the queue reports are pending for this worker pool',
     minCapacity: 'The minimum amount of capacity that should be running',
     maxCapacity: 'The maximum amount of capacity that should be running',
-    capacityPerInstance: 'Amount of capacity a single instance provides',
-    running: 'Number of currently requested and running instances',
-    desiredSize: 'Number that this estimator thinks we should have',
+    runningCapacity: 'Amount of currently requested and available capacity',
+    desiredCapacity: 'Amount of capacity that this estimator thinks we should have',
+    requestedCapacity: 'Amount of capacity that this estimator thinks we should add',
   },
 });
 
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -66,8 +66,7 @@ class AwsProvider extends Provider {
       workerPoolId,
       minCapacity: workerPool.config.minCapacity,
       maxCapacity: workerPool.config.maxCapacity,
-      capacityPerInstance: 1, // todo: this will be corrected along with estimator changes (bug 1579554)
-      running: workerPool.providerData[this.providerId].running,
+      runningCapacity: workerPool.providerData[this.providerId].running,
     });
     if (toSpawn === 0) {
       return;
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -240,7 +240,7 @@ class AwsProvider extends Provider {
         case 'running':
         case 'shutting-down': //so that we don't turn on new instances until they're entirely gone
         case 'stopping':
-          this.seen[worker.workerPoolId] += worker.providerData.instanceCapacity;
+          this.seen[worker.workerPoolId] += worker.providerData.instanceCapacity || 1;
           return Promise.resolve();
 
         case 'terminated':
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -62,7 +62,7 @@ class AwsProvider extends Provider {
       });
     }
 
-    const toSpawn = await this.estimator.simple({
+    let toSpawn = await this.estimator.simple({
       workerPoolId,
       minCapacity: workerPool.config.minCapacity,
       maxCapacity: workerPool.config.maxCapacity,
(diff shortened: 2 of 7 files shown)


owlishDeveloper:
Is it also this bug? https://bugzilla.mozilla.org/show_bug.cgi?id=1579554

services/worker-manager/src/providers/aws.js
```diff
@@ -62,24 +62,26 @@ class AwsProvider extends Provider {
...
+      runningCapacity: workerPool.providerData[this.providerId].running,
     });
     if (toSpawn === 0) {
       return;
     }
-    const toSpawnPerConfig = Math.ceil(toSpawn / workerPool.config.launchConfigs.length);
 
-    const shuffledConfigs = _.shuffle(workerPool.config.launchConfigs);
+    const cfgs = [];
+    const maxPerConfig = 15;
```
owlishDeveloper: I don't understand this number, can you tell more about it and how you came up with it?

services/worker-manager/src/providers/aws.js
```diff
@@ -62,24 +62,26 @@ class AwsProvider extends Provider {
...
+    while (toSpawn> 0) {
+      const cfg = _.sample(workerPool.config.launchConfigs);
+      const provided = Math.min(toSpawn, maxPerConfig * cfg.capacityPerInstance);
+      cfgs.push([provided, cfg]);
+      toSpawn -= provided;
+    }
```
owlishDeveloper: What is the reason for this change? What was wrong with the previous logic?

services/worker-manager/src/providers/aws.js
```diff
@@ -62,24 +62,26 @@ class AwsProvider extends Provider {
...
 
-    let spawned;
-    let toSpawnCounter = toSpawn;
-    for await (let config of shuffledConfigs) {
-      if (toSpawnCounter <= 0) break; // eslint-disable-line
+    await Promise.all(cfgs.map(async ([toSpawnThisConfig, config]) => {
```
owlishDeveloper: Why replacing `for await`?

services/worker-manager/src/providers/aws.js
imbstack: I just picked a number. We normally have between 5 and 5000 workers in a pool so this felt like a nice chunk. Happy to change it!

services/worker-manager/src/providers/aws.js
imbstack: This way runs all requests concurrently rather than serially so loops will take less time.

services/worker-manager/src/providers/aws.js
imbstack: The old way could not handle different `capacityPerInstance` for different `launchConfigs`.

imbstack:
Would you like to make a PR into this one that changes the aws provider to the way you'd like it? I'm happy either way!

services/worker-manager/src/providers/aws.js
```diff
@@ -108,6 +105,7 @@ class AwsProvider extends Provider {
         });
       }
 
+      let spawned;
```
owlishDeveloper: [copying over from Slack] I’d have `MaxCount: Math.min(toSpawnCounter, Math.ceil(toSpawnPerConfig / config.capacityPerInstance) )`, on line 115, and the same on line 116. 

Btw I am assuming we want to distribute capacity over configs/regions and not instances?

actually the `toSpawnCounter` also needs division in the `runInstances` call because it counts capacity, not instances. All those variables are capacity except the numbers in `runInstances` call that are instances. I think that might work

services/worker-manager/src/providers/aws.js
```diff
@@ -108,14 +105,15 @@ class AwsProvider extends Provider {
...
       try {
         spawned = await this.ec2s[config.region].runInstances({
           ...config.launchConfig,
 
           UserData: userData.toString('base64'), // The string needs to be base64-encoded. See the docs above
 
-          MaxCount: Math.min(toSpawnCounter, toSpawnPerConfig),
-          MinCount: Math.min(toSpawnCounter, toSpawnPerConfig),
+          MaxCount: Math.min(toSpawnCounter, Math.ceil(toSpawnPerConfig / config.capacityPerInstance)),
```
owlishDeveloper: I think we need to divide `toSpawnCounter` as well, because it counts _capacity_, but we are sending the amount of _instances_ we need to `runInstances`...

services/worker-manager/src/providers/aws.js
```diff
@@ -108,14 +105,16 @@ class AwsProvider extends Provider {
...
+      let spawned;
       try {
         spawned = await this.ec2s[config.region].runInstances({
           ...config.launchConfig,
 
           UserData: userData.toString('base64'), // The string needs to be base64-encoded. See the docs above
 
-          MaxCount: Math.min(toSpawnCounter, toSpawnPerConfig),
-          MinCount: Math.min(toSpawnCounter, toSpawnPerConfig),
+          MaxCount: instanceCount,
```
owlishDeveloper: I'm probably not being clear. I meant:

```
MaxCount: Math.min(Math.ceil(toSpawnCounter  / config.capacityPerInstance), Math.ceil(toSpawnPerConfig / config.capacityPerInstance) )
```

in `runInstances`, and the same for `minCount`. Why don't I just test what I am suggesting tho? Maybe it doesn't work

imbstack: The `instanceCount` calculations are equivalent to this afaict.

services/worker-manager/src/monitor.js
```diff
@@ -29,9 +29,9 @@ monitorManager.register({
     pendingTasks: 'The number of tasks the queue reports are pending for this worker pool',
     minCapacity: 'The minimum amount of capacity that should be running',
     maxCapacity: 'The maximum amount of capacity that should be running',
-    capacityPerInstance: 'Amount of capacity a single instance provides',
-    running: 'Number of currently requested and running instances',
-    desiredSize: 'Number that this estimator thinks we should have',
+    runningCapacity: 'Amount of currently requested and available capacity',
+    desiredCapacity: 'Amount of capacity that this estimator thinks we should have',
+    requestedCapacity: 'Amount of capacity that this estimator thinks we should add',
```
djmitche: It might be good to include somewhere in the docs (maybe in `ui/docs/reference/core/worker-manager/README.mdx`?) a definition for "capacity".

services/worker-manager/src/providers/aws.js
```diff
@@ -163,6 +162,7 @@ class AwsProvider extends Provider {
           state: this.Worker.states.REQUESTED,
           providerData: {
             region: config.region,
+            instanceCapacity: config.capacityPerInstance,
```
djmitche: :heart: this is what always tripped up aws-provisioner: it didn't know the capacity of existing instances.  Nice to see that solved here!

services/worker-manager/src/providers/google.js
```diff
@@ -229,23 +229,31 @@ class GoogleProvider extends Provider {
...
+      runningCapacity: workerPool.providerData[this.providerId].running,
     });
 
-    await Promise.all(new Array(toSpawn).fill(null).map(async i => {
+    if (toSpawn === 0) {
+      return; // Nothing to do
+    }
+
+    const cfgs = [];
+    while (toSpawn> 0) {
```
djmitche: whitespace around operator (this has been driving me nuts in contributors' code, too.. I'll try to get eslint to watch for this)

PR #1714, opened 2019-10-21
Implement CLI login

helfi92:
Resolves https://github.com/taskcluster/taskcluster/issues/1713.

Sample URL: http://localhost:5080/auth/clients/create?client_id=foo&callback_url=http%3A%2F%2Flocalhost%3A9500&expires=3%20days&scope%5B%5D=assume%3Abar&description=client%20for%20cli%20usage

### DIFF
--- a/ui/src/views/Clients/ViewClient/index.jsx
+++ b/ui/src/views/Clients/ViewClient/index.jsx
@@ -2,6 +2,7 @@ import { hot } from 'react-hot-loader';
 import React, { Component, Fragment } from 'react';
 import { graphql, withApollo } from 'react-apollo';
 import CopyToClipboard from 'react-copy-to-clipboard';
+import { parse } from 'qs';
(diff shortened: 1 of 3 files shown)


djmitche:
(once it has a changelog..)

ui/src/views/Clients/ViewClient/index.jsx
```diff
@@ -165,8 +168,26 @@ export default class ViewClient extends Component {
     this.setState({ accessToken: null });
   };
 
+  // Only localhost callback_url's are whitelisted.
+  // This tool is not intended for other uses
+  // than setting up credentials via `taskcluster signin`,
+  // which uses this URL format.
+  isWhitelistedCallback = callbackUrl =>
```
imbstack: I think we're trying to get away from calling things white/black list but otherwise this makes sense.

PR #1592, opened 2019-10-12
[ui] Bringing all dependancies inline with each other

vishakhanihore:
fixes #1591

PR #1718, opened 2019-10-22
[ui] Bringing all dependancies inline with each other

### LINKED ISSUE #1591: Various dependancies is `ui` are backlogged or not maintained
**Describe the bug**

Various dependencies is `ui` are backlogged or not maintained. Update them to a new version or a maintained version.

vishakhanihore:
fixes #1591

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -5,7 +5,7 @@
   "license": "MPL-2.0",
   "engines": {
     "node": "10.16.3",
-    "yarn": ">=1.10.0"
+    "yarn": ">=1.17.3"
   },
   "scripts": {
     "build": "webpack --mode production",
--- a/yarn.lock
+++ b/yarn.lock
@@ -97,27 +97,6 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/generator@^7.3.4":
-  version "7.3.4"
-  resolved "https://registry.yarnpkg.com/@babel/generator/-/generator-7.3.4.tgz#9aa48c1989257877a9d971296e5b73bfe72e446e"
-  integrity sha512-8EXhHRFqlVVWXPezBW5keTiQi/rJMQTg/Y9uVCEZ0CAF3PKtCCaVRnp64Ii1ujhkoDhhF1fVsImoN4yJ2uz4Wg==
-  dependencies:
-    "@babel/types" "^7.3.4"
-    jsesc "^2.5.1"
-    lodash "^4.17.11"
-    source-map "^0.5.0"
-    trim-right "^1.0.1"
-
-"@babel/generator@^7.6.0":
-  version "7.6.2"
-  resolved "https://registry.yarnpkg.com/@babel/generator/-/generator-7.6.2.tgz#dac8a3c2df118334c2a29ff3446da1636a8f8c03"
-  integrity sha512-j8iHaIW4gGPnViaIHI7e9t/Hl8qLjERI6DcV9kEpAIDJsAOrcnXqRS7t+QbhL76pwbtqP+QCQLL0z1CyVmtjjQ==
-  dependencies:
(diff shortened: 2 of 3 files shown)


vishakhanihore:
@helfi92 @djmitche Please review it.
I think I messed up with `force push` thats why had to close #1592

vishakhanihore:
@helfi92 @djmitche Please review it. Once done I will squash my commits

yarn.lock
```diff
@@ -119,27 +119,6 @@
     semver "^5.4.1"
```
helfi92: I see changes in `ui/package.json` but not in `package.json`. Because there are no changes in the root package json, any reason we have changes in the root `yarn.lock`?

ui/package.json
```diff
@@ -60,7 +60,7 @@
     "markdown-it-highlightjs": "^3.0.0",
     "markdown-it-link-attributes": "^3.0.0",
     "material-ui-pickers": "2.2.4",
-    "material-ui-treeview": "^3.2.0",
+    "material-ui-treeview": "^3.4.0",
```
helfi92: Ah, the renovate bot recently filed a PR to update this package to `3.5.0` in https://github.com/taskcluster/taskcluster/pull/1716. This line is probably not necessary anymore.

ui/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MPL-2.0",
   "engines": {
     "node": "10.16.3",
-    "yarn": ">=1.10.0"
+    "yarn": ">=1.17.3"
```
helfi92: Looks like the packages in package.json are more or less up-to-date. I still see the benefit of adding `"yarn": ">=1.17.3"` though. This means we probably won't have any changes in `yarn.lock` and `ui/yarn.lock`.

vishakhanihore: I will check again

vishakhanihore:
@helfi92 It looks good now, As you said `Package.json` was already up-to-date but after running yarn upgrade and seeing warnings,  I have added and removed some `peer-dependencies` from `.lock` and the deploy preview is successful too.
I think its ready for review 😅

vishakhanihore:
@helfi92 @djmitche Please review before conflicts due to @renovate-bot

vishakhanihore:
@helfi92 @djmitche Please review before conflicts due to @renovate-bot

vishakhanihore:
> Changing dependencies is risky business, as it's common for an upgrade to have unintended consequences. For example, in #1501 identifies a possible issue upgrading @octokit/rest -- an upgrade that's buried in the 1,111 lines of `yarn.lock` change in this PR.
> 
> That's why we use renovate -- it makes one change per PR, and we can quickly merge those we're confident are not harmful or well-tested, while taking or time to test out more risky changes. Also, if something does go wrong, we can roll back the change fairly easily.
> 
> Regarding the changes to YARN_VERSION and `yarn` in `ui/package.json`, we are actually using yarn-1.19.1 (it's in `.yarn`).
> 
> This PR is "update all the things!" which is quite a broad project. I think a better approach here would be to pick a specific goal -- perhaps harmonizing the yarn versions, or removing one of the peer-dependency warnings -- and then solve that with the minimal change to the lockfile, to limit the chance of undesirable side-effects.

@djmitche Hmm, I think you are correct. Also, I was thinking to drop this issue since the following can be done by `Dependabot and Renovate`, Also as you said making such broad changes may have effects which we may not know off.
Also, not to mention I have added many peer dependencies manually in `yarn.lock` to suppress the warning. May be we can look into it afterwards one by one. 😄

djmitche:
Sounds good.  This kind of work is, in my mind, the least "fun" part about open source, so I don't want you to feel obligated to work on it -- writing code to fix bugs or add features is much more interesting!

Issue #1685, opened 2019-10-18
[UI] Make a worker pool that is "scheduled for deletion" editable


helfi92:
**Reserved** for [outreachy](https://www.outreachy.org/) candidates. In order for this to be assigned to you, make sure to complete the first step of the application process outlined in the [project description](https://www.outreachy.org/december-2019-to-march-2020-internship-round/communities/mozilla/#build-a-component-that-can-display-a-json-schema).

---

**Describe the bug**

**To Reproduce**
1. Go to https://taskcluster-ui.herokuapp.com/worker-manager?include_deleted=true
2. Look for a worker pool that has the "SCHEDULED FOR DELETION" label

![Screen Shot 2019-10-18 at 9 20 57 AM](https://user-images.githubusercontent.com/3766511/67097711-ace44c80-f188-11e9-9f5e-d55cfb45c002.png)

3. Click on the Worker Pool ID (e.g., `gecko-t/linux-large`)

**Actual behavior**
Nothing happens. The text is not a button.

**Expected behavior**
The user is navigated to the edit view of that worker pool.

**Additional context**
This bug was originally filed in https://bugzilla.mozilla.org/show_bug.cgi?id=1581863.

ajinkabeer:
Can I try this one out if there aren't any new outreachy candidates looking for working on issues?

km-js:
Can I try this one??

helfi92:
Resolved in https://github.com/taskcluster/taskcluster/commit/1cdc517bd9948bf86c893ad4346715bfed2929d5.

PR #1720, opened 2019-10-22
[UI] Make a worker pool that is "scheduled for deletion" editable #1685

km-js:
Bugzilla Bug: [1581863](https://bugzilla.mozilla.org/show_bug.cgi?id=1581863)

#1685 

This bug can be resolved by making the worker pool that is "scheduled for deletion" behave like the other worker pools that are editable.

Changes Done:
Removed the conditional operator so that both types of worker pools can work alike.


km-js:
Hey @helfi92, sorry for being late but can you help me a lil bit to understand what exactly we have to do with the provider field. Do we have to set it's initial value to null provider?

helfi92:
Now that I think about it, I think it makes sense for it to not have a provider selected.

helfi92:
Before we merge, we need to remove the `package-lock.json` file.

helfi92:
@all-contributors please add @km-js for code.

PR #1662, opened 2019-10-16
Create role and delete smoketest

ajinkabeer:
This PR addresses the issue #1246. This is the first version of the smoketest which creates a role and deletes it.

PR #1724, opened 2019-10-22
Add parameterized role in smoketest

ajinkabeer:
This PR is a follow up to #1662 which is to accomplish the following tasks. 

- [x] parameterized roles - let's change the test to create a parameterized role (so, one with <..> in its scopes and * at the end of its roleId), and then expand an example of that scope, and verify that the resulting expansion is what we expect.

- [ ]  The assertion should have an explicit list of scopes that it expects, rather than just asserting that it is equal to createdRole.expandedScopes. The rationale is that we want to test that the Auth service is generating the right output. What landed in #1662 just checks that the Auth service generates the same output in two different ways -- but that output might be wrong!

- [ ] clean up: at the end of the test, please use listRoles2, including its pagination feature, to look for roles that this check might have created in the past (based on the pattern of roleId it creates). For any matching roles, look at the lastModified date, and if it is older than 1 hour, remove the role. This will allow us to eventually "clean up" any leftover roles from failed test runs, crashed workers, etc. The "older than 1 hour" part avoids deleting a role that is being used by another check running at exactly the same time (which can happen more often than you'd think!)

Could you please tell me if this is the right way to add parameterized role? and also what exactly is `continuationToken `because when I call `listRoles2`, every role is being listed. Is `continuationToken `something which I can filter out specific role?
@djmitche

### DIFF
--- a/infrastructure/tooling/src/smoketest/checks/role.js
+++ b/infrastructure/tooling/src/smoketest/checks/role.js
@@ -9,19 +9,24 @@ exports.tasks.push({
     'target-roles',
   ],
   run: async () => {
-    let auth = new taskcluster.Auth(taskcluster.fromEnvVars());
-    let randomId = taskcluster.slugid();
-    let roleId = `project:taskcluster:smoketest:${randomId}`;
-    let payload = {
+    const auth = new taskcluster.Auth(taskcluster.fromEnvVars());
+    const randomId = taskcluster.slugid();
+    const roleId = `project:taskcluster:smoketest:${randomId}:*`;
+    const payload = {
       description: 'smoketest test',
-      scopes: ['project:taskcluster:smoketest:*'],
+      scopes: ['project:taskcluster:smoketest:<..>/*'],
     };
-    let expandPayload = {
-      scopes: [`assume:${roleId}`],
+    const expandPayload = {
+      scopes: [`assume:project:taskcluster:smoketest:${randomId}:abc`],
     };
-    let roleCreated = await auth.createRole(roleId, payload);
-    let expandedRole = await auth.expandScopes(expandPayload);
-    assert.deepEqual(roleCreated.expandedScopes, expandedRole.scopes);
+    await auth.createRole(roleId, payload);
+    const expandedRole = await auth.expandScopes(expandPayload);
+    const expectedScopes = {
+      scopes:
+      [ `assume:project:taskcluster:smoketest:${randomId}:abc`,
+        'project:taskcluster:smoketest:abc/*' ],
+    };
+    assert.deepEqual(expandedRole.scopes, expectedScopes.scopes);
     await auth.deleteRole(roleId);
   },
 });


infrastructure/tooling/src/smoketest/checks/role.js
```diff
@@ -9,19 +9,26 @@ exports.tasks.push({
     'target-roles',
   ],
   run: async () => {
-    let auth = new taskcluster.Auth(taskcluster.fromEnvVars());
-    let randomId = taskcluster.slugid();
-    let roleId = `project:taskcluster:smoketest:${randomId}`;
-    let payload = {
+    const auth = new taskcluster.Auth(taskcluster.fromEnvVars());
+    const randomId = taskcluster.slugid();
+    const roleId = `project:taskcluster:smoketest:${randomId}*`;
```
djmitche: Let's put some punctuation in there, just for readability
```suggestion
    const roleId = `project:taskcluster:smoketest:${randomId}:*`;
```

infrastructure/tooling/src/smoketest/checks/role.js
```diff
@@ -9,19 +9,26 @@ exports.tasks.push({
...
+    const payload = {
       description: 'smoketest test',
-      scopes: ['project:taskcluster:smoketest:*'],
+      scopes: ['project:taskcluster:smoketest:<..>/*'],
     };
-    let expandPayload = {
+    const expandPayload = {
       scopes: [`assume:${roleId}`],
```
djmitche: So, here we'll want to expand `assume:project:taskcluster:smoketest:${randomId}:abc`, and see that the `abc` is substituted properly into the resulting scopes.

infrastructure/tooling/src/smoketest/checks/role.js
ajinkabeer: Yes.

infrastructure/tooling/src/smoketest/checks/role.js
ajinkabeer: Yes, I have made the change. Can you take a look please?
@djmitche

infrastructure/tooling/src/smoketest/checks/role.js
```diff
@@ -9,19 +9,24 @@ exports.tasks.push({
...
+    const roleId = `project:taskcluster:smoketest:${randomId}:*`;
+    const payload = {
       description: 'smoketest test',
-      scopes: ['project:taskcluster:smoketest:*'],
+      scopes: ['project:taskcluster:smoketest:<..>/*'],
     };
-    let expandPayload = {
-      scopes: [`assume:${roleId}`],
+    const expandPayload = {
+      scopes: [`assume:${roleId}abc`],
```
djmitche: This ends up as `assume:project:taskclsuter:smoketest:${randomId}:*abc`, which is why the results below have an extra `*` in them.  Rather than using `${roleId}` here, let's use `assume:project:taskclsuter:smoketest:${randomId}:abc` to avoid that issue.

infrastructure/tooling/src/smoketest/checks/role.js
```diff
@@ -9,19 +9,24 @@ exports.tasks.push({
...
     };
-    let roleCreated = await auth.createRole(roleId, payload);
-    let expandedRole = await auth.expandScopes(expandPayload);
-    assert.deepEqual(roleCreated.expandedScopes, expandedRole.scopes);
+    await auth.createRole(roleId, payload);
+    const expandedRole = await auth.expandScopes(expandPayload);
+    const expectedScopes = {
+      scopes:
+      [ `assume:project:taskcluster:smoketest:${randomId}:*abc`,
+        'project:taskcluster:smoketest:*abc/*' ],
```
djmitche: That will result in the first `*` in each of these lines going away

infrastructure/tooling/src/smoketest/checks/role.js
ajinkabeer: Sure, I will change it.

infrastructure/tooling/src/smoketest/checks/role.js
ajinkabeer: Awesome, now its working perfectly! Can you take a look at the changes please?
@djmitche

djmitche:
On to the cleanup work?

ajinkabeer:
> On to the cleanup work?

Thanks. Wow, Awesome. Let's do it!
@djmitche

PR #1737, opened 2019-10-24
Bug 1588834: [aws-provider] Allow specifying additional userdata for AWS instances.

tomprince:
@owlishDeveloper I think OCC (the tool we use for configuring windows instances wants to have some additional data provided to it, that doesn't really belong in the worker config. This does the simplest thing to allow worker-pools to configure the extra data, while preventing the data that worker-manger uses from being overwritten or specified.

I also noticed that the schema for preventing `UserData` and counts didn't work as expected (negative schemas can have surprising behavior). I did not see any easy way to write tests for this, but I did test that the idiom I used does prevent the properties from being specified.

Bugzilla Bug: [1588834](https://bugzilla.mozilla.org/show_bug.cgi?id=1588834)


changelog/1588834.md
```diff
@@ -0,0 +1,4 @@
+level: minor
+reference: bug 1588834 
+---
+* Allow specifying addtional userdata in AWS Provider worker pools.
```
djmitche: [The changelog doc](https://github.com/taskcluster/taskcluster/blob/master/dev-docs/best-practices/changelog.md) says "Write it in the present tense, as if the change had just been deployed."

```suggestion
AWS Provider worker pools now allow specifying additional userdata beyond that generated by the provider itself.
```

owlishDeveloper:
> I also noticed that the schema for preventing UserData and counts didn't work as expected (negative schemas can have surprising behavior)

Nice catch! Can you file a bug with the description of your findings?

owlishDeveloper:
oh it was corrected here. Still, can you describe the behavior you saw? @tomprince

services/worker-manager/schemas/v1/config-aws.yml
```diff
@@ -25,14 +25,27 @@ properties:
             ImageId:
               type: string
               description: The ID of the AMI to launch an instance from.
+            # These properties are not allowed, so we assert that
+            # they do not match the empty schema
+            TagSpecifications: {not: {}}
```
owlishDeveloper: why `TagSpecifications`?

services/worker-manager/schemas/v1/config-aws.yml
```diff
@@ -25,14 +25,27 @@ properties:
...
+            IamInstanceProfile: {not: {}}
```
owlishDeveloper: Can you talk more about what you saw, and how you came up with this solution? (`{not: {}}` instead of `not.required:`)

services/worker-manager/schemas/v1/config-aws.yml
tomprince: Looking at [this code](https://github.com/taskcluster/taskcluster/blob/8a4d49f3e8b134168781d84d4abf481774ffc6ae/services/worker-manager/src/providers/aws.js#L85) it appears that the `TagSpecification` is read from the config, not the `launchConfig`, and `TagSpecifications` is overridden in the `runInstances` call. So this will prevent it being specified and then ignored.

services/worker-manager/schemas/v1/config-aws.yml
tomprince: If you have the schema:
```yaml
not:
    required: [a, b]
```
then `{a: stuff}` and `{b: stuff}` will both validate, but `{'a: stuff, b:stuff}` will not. That is because both the first two do *not* validate against `required: [a, b]` (since in either case they are missing one of the required properties).

The way that `{not: {}}` works is that `{}` will match against any value (including `null`). So, `not: {}` will not match against any value. Therefore, if an optional property has to match `not: {}`, it can't be specified (since the value specified would not match the schema).

services/worker-manager/schemas/v1/config-aws.yml
owlishDeveloper: That's a bug, Dustin and I are both aware of it and there's a fix coming

tomprince: Ok. That seemed like it might have been expected behavior. I'm happy to remove that, or it could be removed along with the fix to the bug?

owlishDeveloper:
@tomprince There are 3 issues here in one PR:
* Fix for `TagSpecifications`
* Fix for `not-required` in json-schema
* Fix for the OCC thing

Can you please make a separate PR for the OCC thing, and a separate PR for the `not-required` thing along with the test case demonstrating the issue? As for the `TagSpecifications` - it's a bug in aws provider that was introduced with a schema change. We know about it, and there's a fix coming, so there's no need to try and fix that.

tomprince:
> separate PR for the `not-required` thing along with the test case demonstrating the issue?

I'm happy to make a separate PR for it, but I have no idea how to write a test for it. I don't see any examples of testing the provider config; it looks like the provider expects a validator to be passed to it, and it isn't clear how to construct one in the tests.

tomprince:
#1743 is the other change; it looks like the merge between the two PRs is trivial.

djmitche:
> I'm happy to make a separate PR for it, but I have no idea how to write a test for it. I don't see any examples of testing the provider config; it looks like the provider expects a validator to be passed to it, and it isn't clear how to construct one in the tests.

I can work on this.  @imbstack mentioned there's an example in the queue.

Issue #1684, opened 2019-10-18
[UI] Add confirmation dialog for delete button


helfi92:
**Reserved** for [outreachy](https://www.outreachy.org/) candidates. In order for this to be assigned to you, make sure to complete the first step of the application process outlined in the [project description](https://www.outreachy.org/december-2019-to-march-2020-internship-round/communities/mozilla/#build-a-component-that-can-display-a-json-schema).

---

**To Reproduce**
1. Go to https://taskcluster-ui.herokuapp.com/worker-manager
2. Click on the garbage icon
3. See error

**Expected behavior**
A confirmation dialog pops asking for confirmation.

![Screen Shot 2019-10-18 at 9 03 10 AM](https://user-images.githubusercontent.com/3766511/67096437-35152280-f186-11e9-9edb-9cf0c2d209a2.png)

**Additional context**
The implementation should be similar to the delete button inside the speed dial in https://taskcluster-ui.herokuapp.com/worker-manager/gce%2Fgecko-1-b-linux. To be able to delete, you would need to login with any credentials (e.g., clientId `test` and accessToken`test`).

naimiii:
can  you please assign me this issue ?

naimiii:
@helfi92  I was debugging the 'deleteRequest' function and got an error for wrong 'clientId' error . Is this a right behavior ? 
![74389760_934055763631215_4553600592953475072_n](https://user-images.githubusercontent.com/28563415/67316055-36957080-f521-11e9-8dec-ac214a413a20.png)

helfi92:
Yes, you won't be able to perform the actual delete action. You can comment it out to make it look like it worked.

naimiii:
@helfi92    so I will create a pop-up that will render before so user will have to confirm deletion just like in the Worker Pool form.

naimiii:
@helfi92  what should the string say in the DialogAction component when deleting from the worker pools table component? should it still join the workerPool.workerPoolId1, workerPool.workerPoolId2 ID's?

helfi92:
Do what you think makes sense then we can adjust things on the PR.

naimiii:
@helfi92  Kindly check the PR of this issue . :)  thanks .

vishakhanihore:
@naima-shaikh Are you still working on this ticket ?

naima-shaikh:
@vishakhanihore  I already fixed it .Just waiting for the review . :(

PR #1742, opened 2019-10-24
Workerpools table confirmation pop up

naima-shk:
Fixes https://github.com/taskcluster/taskcluster/issues/1684.


naimiii:
@helfi92 
73291296_456797885189278_2299681247267389440_n

naimiii:
@helfi92  I made the changes you requested . :) 
76183114_663417277517559_5165857569796784128_n

naimiii:
@helfi Please review it :) thanks .

ui/src/components/WMWorkerPoolsTable/index.jsx
```diff
@@ -57,6 +57,8 @@ export default class WorkerManagerWorkerPoolsTable extends Component {
     sortDirection: 'asc',
     error: null,
     actionLoading: false,
+    dialogOpen: false,
+    workerPool: null,
```
helfi92: Instead of `workerPool`, `dialogOpen`, and `error`, let's try using `dialogState`:

```jsx
dialogState: {
  error: null,  
  open: false,
  title: '',
  confirmText: '',
  item: null,
},
```

Then we a user clicks on delete, we set the dialog state appropriately:

```jsx
this.setState({
  dialogState: {
    // ...
  },
});
```

If there is an error, we could do:

```jsx
this.setState({
  ...this.state.dialogState,
  error,
});
```

ui/src/components/WMWorkerPoolsTable/index.jsx
```diff
@@ -131,17 +131,48 @@ export default class WorkerManagerWorkerPoolsTable extends Component {
...
+  handleDialogActionOpen = workerPool => {
+    this.setState({
+      dialogOpen: true,
+      workerPool,
+    });
+  };
+
+  handleDialogActionError = error => {
+    this.setState({
+      error: formatError(error),
```
helfi92: I think we don't need `formatError` anymore.

naima-shaikh:
@helfi92  please update review :)

ui/src/components/WMWorkerPoolsTable/index.jsx
```diff
@@ -112,36 +118,87 @@ export default class WorkerManagerWorkerPoolsTable extends Component {
...
+  };
+
+  handleDialogActionError = error => {
+    const { item } = this.state.dialogState;
+
+    this.setState({
+      dialogState: {
+        ...this.state.dialogState,
+        error,
+        body: `Could not delete ${item.workerPoolId}. Please try again.`,
```
helfi92: No need to change the body. The error is enough to grab the user's attention.

ui/src/components/WMWorkerPoolsTable/index.jsx
```diff
@@ -112,36 +118,87 @@ export default class WorkerManagerWorkerPoolsTable extends Component {
...
+    this.setState({
+      dialogState: {
+        error: null,
+        open: false,
+        title: '',
+        body: '',
+        confirmText: '',
+        item: null,
+      },
+    });
```
helfi92: No need to remove the title, body, confirmText, item when closing the dialog. We can just set `error: null, open: false`. Also, since we're setting `error: null` here, then no need to set `error: null` in `handleDialogActionOpen`.

ui/src/components/WMWorkerPoolsTable/index.jsx
```diff
@@ -112,36 +118,87 @@ export default class WorkerManagerWorkerPoolsTable extends Component {
...
     } catch (error) {
-      this.setState({ error: formatError(error), actionLoading: false });
+      this.handleDialogActionError(error);
     }
   };
 
+  handleDialogActionOpen = workerPool => {
+    this.setState({
+      dialogState: {
+        error: null,
```
helfi92: I don't think `error` is needed here. See other comment for explanation.

ui/src/components/WMWorkerPoolsTable/index.jsx
```diff
@@ -256,16 +325,28 @@ export default class WorkerManagerWorkerPoolsTable extends Component {
...
+          <DialogAction
+            open={open}
+            onSubmit={handleDeleteClick}
+            onComplete={handleDialogActionComplete}
+            onClose={handleDialogActionClose}
+            onError={handleDialogActionError}
+            error={error}
+            title={title}
+            body={<Typography>{body}</Typography>}
+            confirmText={confirmText}
```
helfi92: STR:
1. Open dialog
2. Try to delete and wait until the error shows up in the dialog
3. Try to delete again

Actual: The error message doesn't move.

Expected: the error is set to null before attempting to delete again. That way the error message goes away for a split second and only shows up again if the action fails. The benefit of doing this is to give feedback to the user that we have cleared this error and that we are trying again.

naima-shaikh:
@ can you please update review ?

naima-shk:
@helfi92  Alright , Thanks :)

PR #1743, opened 2019-10-24
[aws-provider] Disallow entries in worker-pool that are used by the provider.

tomprince:


owlishDeveloper:
Can you please the config that revealed the bug?

tomprince:
I discovered the bug by inspection, when working on #1737 (I don't recall if it was when seeing if it was implemented, or when adjusting the schema for it).

owlishDeveloper:
@tomprince I added a test framework for schemas. (you'll find a sample test in `worker-manager/test/schema_validate_test.js`) Can you please add a test case for this change?

tomprince:
> @tomprince I added a test framework for schemas. (you'll find a sample test in `worker-manager/test/schema_validate_test.js`) Can you please add a test case for this change?

It looks like it doesn't now about providers, so it will accept anything in `config`.

owlishDeveloper:
This was already merged

Issue #1660, opened 2019-10-16
[UI] Error message hard to read in docs


helfi92:
**Describe the bug**
When viewing docs, sometimes error panel are hard to see.

**To Reproduce**
1. Add the following diff

```diff
diff --git a/ui/docs/manual/README.mdx b/ui/docs/manual/README.mdx
index 68a7e270f..a05dac2ca 100644
--- a/ui/docs/manual/README.mdx
+++ b/ui/docs/manual/README.mdx
@@ -36,3 +36,5 @@ describes how these tasks are executed.  The next chapter, [System
 Design](/docs/manual/system-design), provides the details you will need to interact
 with Taskcluster. Finally, [Using Taskcluster](/docs/manual/using) presents a
 collection of common use cases and approaches to satisfy them.
+
+<>
```

2. Go to http://localhost:5080/docs/manual/
3. See error

![Screen Shot 2019-10-16 at 11 42 02 AM](https://user-images.githubusercontent.com/3766511/66935461-fe1a0200-f009-11e9-8d07-02997bd5b776.png)

**Expected behavior**
We can see the full content of the error message.

naimiii:
@helfi92  can i contribute into it please ? I am an outreachy  candidate .

THEozmic:
What's the status of this?

helfi92:
This issue is free to grab. I haven't heard back from @DrFeederino.

PR #1756, opened 2019-10-26
Fix ErrorPanel Visibility Bug

THEozmic:
Issue: https://github.com/taskcluster/taskcluster/issues/1660

### DIFF
--- a/ui/src/components/ErrorPanel/index.jsx
+++ b/ui/src/components/ErrorPanel/index.jsx
@@ -3,7 +3,6 @@ import { withStyles } from '@material-ui/core/styles';
 import classNames from 'classnames';
 import { bool, oneOfType, string, object } from 'prop-types';
 import MuiErrorPanel from '@mozilla-frontend-infra/components/ErrorPanel';
-import { CONTENT_MAX_WIDTH } from '../../utils/constants';
 
 @withStyles(theme => ({
   warning: {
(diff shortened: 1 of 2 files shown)


THEozmic:
@djmitche

djmitche:
Remember most of us do not work on the weekends!  I'm sure your mentor will be along to review this in due time. :)

ui/src/components/ErrorPanel/index.jsx
```diff
@@ -21,14 +20,6 @@ import { CONTENT_MAX_WIDTH } from '../../utils/constants';
...
   },
-  fixed: {
-    position: 'fixed',
-    zIndex: theme.zIndex.drawer - 1,
-    left: '50%',
-    transform: 'translateX(-50%)',
-    width: '92%',
-    maxWidth: CONTENT_MAX_WIDTH,
-  },
 }))
```
helfi92: The pinned error panel (fixed) won't work anymore. For example, if a user scrolls down the page, the error panel will scroll out of the screen instead of always staying visible on the screen.

What I am looking for is a change to this `fixed` class to make it always stay within the boundaries of the `<main/ >` tag without removing the `position: 'fixed'` property.

Screen Shot 2019-10-28 at 8 55 50 AM

THEozmic: Okay. I've made an update. This time, the ErrorPanel is above the drawer and is always fully visible when you scroll.

helfi92:
Apologies for the delay. I [modified](https://github.com/taskcluster/taskcluster/pull/2216) the pull-request to have it center in the main content rather than have it stack on top of the navigation menu. Thank you so much for your help and for getting this started!

Issue #1758, opened 2019-10-26
Update `taskcluster signin` to work with new deployments


djmitche:
The [taskcluster shell client](https://github.com/taskcluster/taskcluster/tree/master/clients/client-shell#readme), available for download [here](https://github.com/taskcluster/taskcluster/releases/tag/v21.2.0), has a `taskcluster signin` command which interacts with a deployment's UI to get Taskcluster credentials on the command line.  In #1714, @helfi92 added support for this to the new UI.  However, the query arguments differ slightly between the old (tools.taskcluster.net) and new (everything else) UIs.

The bug here is to update the `taskcluster signin` implementation to use the new query parameters for all rootUrls *except* `https://taskcluster.net`, for which it should use the old query parameters.

RootUrls to try this against are https://stage.taskcluster.nonprod.cloudops.mozgcp.net or https://community-tc.services.mozilla.com (both will behave the same).

ededals:
Is this one for me? Judging by the #1714 the new query parameters should look like this:
```http://localhost:5080/auth/clients/create?client_id=foo&callback_url=http%3A%2F%2Flocalhost%3A9500&expires=3%20days&scope%5B%5D=assume%3Abar&description=client%20for%20cli%20usage```

I guess i will have to modify signin.go url construction part to match the new format.

Just to double check, is the expected result a succesfull authentification when trying against https://stage.taskcluster.nonprod.cloudops.mozgcp.net ?

djmitche:
> I guess i will have to modify signin.go url construction part to match the new format.

Indeed

> Just to double check, is the expected result a succesfull authentification when trying against https://stage.taskcluster.nonprod.cloudops.mozgcp.net ?

Correct.  Also, it should still work against https://tools.taskcluster.net (with TASKCLUSTER_ROOT_URL=https://taskcluster.net).

helfi92:
@ededals Feel free to create a separate issue to handle the UI bit discussed in https://github.com/taskcluster/taskcluster/pull/1760#pullrequestreview-308029839 if that's more convenient for you. Otherwise, I'm fine keeping the discussion here.

ededals:
[...]

PR #1760, opened 2019-10-26
Rename CLI login client_id to name and add slugId to avoid resetting token

ededals:
Issue #1758

### DIFF
--- a/changelog/issue-1758.md
+++ b/changelog/issue-1758.md
@@ -1,0 +1,4 @@
+level: minor
+reference: issue 1758
+---
+Taskcluster shell client 'signin' command can now interact with the new UI.
--- a/clients/client-shell/cmds/signin/signin.go
+++ b/clients/client-shell/cmds/signin/signin.go
@@ -119,17 +119,16 @@ func cmdSignin(cmd *cobra.Command, _ []string) error {
 	var loginURL string
 	if config.RootURL() == "https://taskcluster.net" {
 		loginURL += libUrls.UI(config.RootURL(), "/auth/clients/new")
-		loginURL += "?name=" + url.QueryEscape(name)
-  		for i := range scopes {
-  			loginURL += "&scope=" + url.QueryEscape(scopes[i])
+		for i := range scopes {
+  			loginURL += "?scope=" + url.QueryEscape(scopes[i])
   		}
   	} else {
 		loginURL += libUrls.UI(config.RootURL(), "/auth/clients/create")
-		loginURL += "?client_id=" + url.QueryEscape(name)
 	  	for i := range scopes {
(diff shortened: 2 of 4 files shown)


djmitche:
Oh, and you'll need to add a changelog entry.  See https://github.com/taskcluster/taskcluster/blob/master/dev-docs/best-practices/changelog.md.  I'd consider this a "minor" update as it adds significant new functionality.

helfi92:
I'm having trouble running this. Something like `go run main.go signin` throws:

```
➜  client-shell git:(master) go run main.go signin
panic: interface conversion: interface {} is nil, not string

goroutine 1 [running]:
github.com/taskcluster/taskcluster/clients/client-shell/config.Setup()
	/Users/haali/Documents/Mozilla/projects/taskcluster/clients/client-shell/config/init.go:53 +0x61f
main.main()
	/Users/haali/Documents/Mozilla/projects/taskcluster/clients/client-shell/main.go:12 +0x22
exit status 2
```

This is my first time running the client-shell locally so it's very possible I'm doing something wrong. @ededals @djmitche Would you have to know why this is happening. I'm running go version `go1.12.9` darwin/amd64 which was installed using gvm. fwiw, `yarn generate` which tells me that go is properly installed.

ededals:
@helfi92 I do not know if this will be of any use, but  did build a binary first by going to shell-client root directory and running ```go build -o taskcluster```. Then after ```./taskcluster signin``` everything seemed to work fine. I am using ```go version go1.13.3 linux/amd64``` by the way

helfi92:
How is it going @ededals? Did you have any questions for me? We should plan on finishing this implementation soon. I can help with something if that's more convenient.

ededals:
Hey @helfi, I would like to try to handle this, as I like how this is turning out, but during weekdays between family and work it is difficult to find time. I would like to do it on weekend. If this is too late, I can modify just signin.go

helfi92:
The weekend is perfect :) All yours! Thanks for your help :)

ededals:
@helfi92 was thinking on how to get login state. What I came up with was to use this query ```ui/src/App/isLoggedIn.graphql``` on client prop. and use it in ```ui/src/views/ViewClient/index.js``` to check if user is logged in. Does this sound right to you? Is there some better way I should cinsider?

helfi92:
> @helfi92 was thinking on how to get login state. What I came up with was to use this query `ui/src/App/isLoggedIn.graphql` and use it in `ui/src/views/ViewClient/index.js` on client prop, to check if user is logged in. Does this sound right to you? Is there some better way I should consider?

That sounds good to me.

ededals:
Hey @helfi92,
 
In the end I decided to use Auth context to check logged in user.
I hope I have not missed anything.

Issue #1246, opened 2019-08-26
smoketest - create a role and expand it


djmitche:
Add a new test to the `yarn smoketest` by adding a module to `infrastructure/tooling/src/smoketest/checks/`.  In this case, it would create a parameterized role in the auth service, then call `auth.expandScopes` to see that the role expands properly.

See #1218.

JiwoonKim:
Hi! I am an Outreachy Applicant and have previously submitted one pull request for one of the `good first issue` labeled issues. I am still waiting for the pull request to be approved, but thought that this is an interesting issue because my last issue was about editing the role UI :)
I understand that the other `good first issue` labeled issues are reserved for other applicants (I hope they all get the opportunity to contribute, I learned a lot! :) )
Was wondering if it would be okay to maybe take one of the harder issues like this. Would it be okay to take this one, please?

djmitche:
Are you still working on #987?

ajinkabeer:
I am not assigned any issues at the moment, and if @JiwoonKim is not currently working on it, can I try this one out please?
@djmitche

djmitche:
Sure!

djmitche:
Whoops, sorry, mixed up the usernames there!

djmitche:
I think we'll want to do something similar to what we did with clients, where we generate a random roleId each time.  So I've added
```
auth:create-role:project:taskcluster:smoketest:*
auth:update-role:project:taskcluster:smoketest:*
auth:delete-role:project:taskcluster:smoketest:*
```
and also, as a collection of scopes that you can give to the role you create,
```
project:taskcluster:smoketest:*
```

One difference from clients is, roles don't "expire".  So we'll need to find a way to make sure that smoketests clean up after themselves.  The first version of the test you write should just be careful to delete the role it creates; we can do a second PR to add some "cleanup" in case a test fails to delete its own role.

ajinkabeer:
> I think we'll want to do something similar to what we did with clients, where we generate a random roleId each time. So I've added
> 
> ```
> auth:create-role:project:taskcluster:smoketest:*
> auth:update-role:project:taskcluster:smoketest:*
> auth:delete-role:project:taskcluster:smoketest:*
> ```
> 
> and also, as a collection of scopes that you can give to the role you create,
> 
> ```
> project:taskcluster:smoketest:*
> ```
> 
[...]

PR #1776, opened 2019-10-28
Add cleanup for role smoketest 

ajinkabeer:
This PR is a follow up to #1724  which is to accomplish the following tasks. The original issue is #1246

- [ ]  clean up: at the end of the test, please use listRoles2, including its pagination feature, to look for roles that this check might have created in the past (based on the pattern of roleId it creates). For any matching roles, look at the lastModified date, and if it is older than 1 hour, remove the role. This will allow us to eventually "clean up" any leftover roles from failed test runs, crashed workers, etc. The "older than 1 hour" part avoids deleting a role that is being used by another check running at exactly the same time (which can happen more often than you'd think!)

I am not sure if this is an efficient way to do it, please guide me if I am taking the wrong approach here. )
@djmitche

### DIFF
--- a/infrastructure/tooling/src/smoketest/checks/role.js
+++ b/infrastructure/tooling/src/smoketest/checks/role.js
@@ -27,6 +27,20 @@ exports.tasks.push({
         'project:taskcluster:smoketest:abc/*' ],
     };
     assert.deepEqual(expandedRole.scopes, expectedScopes.scopes);
-    await auth.deleteRole(roleId);
+    const query = {};
+    const anHourAgo = Date.now() - (1000*60*60);
+    while (1) {
+      const res = await auth.listRoles2();
+      for(let i=0;i<res.roles.length;i++){
+        if(res.roles[i].roleId.includes('project:taskcluster:smoketest:') && res.roles[i].lastModified < anHourAgo){
+          await auth.deleteRole(res.roles[i].roleId);
+        }
+      }
+      if (res.continuationToken) {
+        query.continuationToken = res.continuationToken;
+      } else {
+        break;
+      }
+    }
   },
 });


helfi92:
@all-contributors please add @ajinkabeer for code.

infrastructure/tooling/src/smoketest/checks/role.js
```diff
@@ -27,6 +27,16 @@ exports.tasks.push({
         'project:taskcluster:smoketest:abc/*' ],
     };
     assert.deepEqual(expandedRole.scopes, expectedScopes.scopes);
-    await auth.deleteRole(roleId);
+    const query = {limit: 1};
+    const anHourAgo = Date.now() - (1000*60*60);
+    while (1) {
+      const res = await auth.listRoles2(query);
+      res.roles[0].roleId.includes('project:taskcluster:smoketest:') === true && res.roles[0].lastModified > anHourAgo ? await auth.deleteRole(roleId) : null;
```
djmitche: Please use an `if` statement, rather than a ternary operator(.. `?` .. `:` ..).

Also, the logic is backward here; we want to delete roles *older* than an hour, not younger :)

infrastructure/tooling/src/smoketest/checks/role.js
```diff
@@ -27,6 +27,16 @@ exports.tasks.push({
         'project:taskcluster:smoketest:abc/*' ],
     };
     assert.deepEqual(expandedRole.scopes, expectedScopes.scopes);
-    await auth.deleteRole(roleId);
+    const query = {limit: 1};
```
djmitche: While using `limit: 1` is good for testing the continuation support, it's not very efficient, so now that you're sure the continuation support works, remove the limit.  That will also require looping over the roles returned from `listRoles2`.

infrastructure/tooling/src/smoketest/checks/role.js
ajinkabeer: Yes. I will change the ternary operator to an `if `statement.

infrastructure/tooling/src/smoketest/checks/role.js
ajinkabeer: I am a little confused regarding this part. Can you please check my latest pushed code and tell me the changes that I can make to get the clean up to work correctly?

infrastructure/tooling/src/smoketest/checks/role.js
```diff
@@ -27,6 +27,18 @@ exports.tasks.push({
         'project:taskcluster:smoketest:abc/*' ],
     };
     assert.deepEqual(expandedRole.scopes, expectedScopes.scopes);
-    await auth.deleteRole(roleId);
+    const query = {};
+    const anHourAgo = Date.now() - (1000*60*60);
+    while (1) {
+      const res = await auth.listRoles2();
+      if(res.roles[0].roleId.includes('project:taskcluster:smoketest:') === true && res.roles[0].lastModified < anHourAgo){
```
djmitche: There's no need to compare a value to true -- booleans are booleans :)

ajinkabeer: Yes, exactly :)

ajinkabeer: Can you please take a look now to my changes ? :)

ajinkabeer: Can you please take a look at my changes please :)
@djmitche

infrastructure/tooling/src/smoketest/checks/role.js
```diff
@@ -27,6 +27,20 @@ exports.tasks.push({
         'project:taskcluster:smoketest:abc/*' ],
     };
     assert.deepEqual(expandedRole.scopes, expectedScopes.scopes);
-    await auth.deleteRole(roleId);
+    const query = {};
+    const anHourAgo = Date.now() - (1000*60*60);
+    while (1) {
+      const res = await auth.listRoles2();
+      for(let i=0;i<res.roles.length;i++){
```
djmitche: The more "modern" way to do this in JS is either
```
res.roles.forEach(role => {...});
```
or
```
for (let role of roles) { ... }
```

ajinkabeer: Yes, this approach would've been more clean and modern.

PR #1779, opened 2019-10-28
Fix readme links

djmitche:
Relative links are tricky in the docs, but origin-relative will always work..


helfi92:
Missing a changelog.

djmitche:
Yeah, it'd just be silent so I'll skip it here.

Issue #1773, opened 2019-10-28
[UI] next/prev icon in light theme


helfi92:
**Describe the bug**
The next and previous icon buttons are not showing in the light theme.

**To Reproduce**

1. Go to https://taskcluster-ui.herokuapp.com/tasks/eWxW0D38QgOv9WZNKgCgAg
2. Click on the light bulb icon in the app bar to switch to the light theme
3. Look at the right panel Task Run

**Actual behavior**

![Screen Shot 2019-10-28 at 11 29 44 AM](https://user-images.githubusercontent.com/3766511/67692340-62c44d80-f976-11e9-85e8-984abf508bb4.png)

**Expected behavior**

![Screen Shot 2019-10-28 at 11 30 30 AM](https://user-images.githubusercontent.com/3766511/67692338-62c44d80-f976-11e9-8135-4799d4eed377.png)

aryamanpuri:
@helfi92 Can you assign me. I will look forward to it .

aryamanpuri:
@helfi92 Should I add two function calls on clicking the toggle light / dark theme button .

helfi92:
Hm, I'm not sure I understand the question. Feel free to create a pull-request then we can look at it there.

helfi92:
Resolved in https://github.com/taskcluster/taskcluster/commit/be2876a740821888562fbd57717fadadc42b37b5.

PR #1828, opened 2019-10-29
 Minor update for the next/previous button in the light theme .

aryamanpuri:
I have changed the color of the Left and Right arrow Icons.
@helfi92  Have a look .
#1773

### DIFF
--- a/changelog/issue-1773.md
+++ b/changelog/issue-1773.md
@@ -1,0 +1,4 @@
+
+level: silent
+reference: issue 1773
+---
--- a/ui/src/components/TaskRunsCard/index.jsx
+++ b/ui/src/components/TaskRunsCard/index.jsx
@@ -485,7 +485,7 @@ export default class TaskRunsCard extends Component {
                 onClick={this.handleNext}
                 disabled={run ? selectedRunId === runs.length - 1 : true}>
                 Next
-                <ChevronRightIcon />
+                <ChevronRightIcon color="rgb(149, 27, 27)" />
               </Button>
             }
             backButton={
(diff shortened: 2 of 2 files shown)


changelog/issue-1773.md
```diff
@@ -0,0 +1,4 @@
+
```
vishakhanihore: The format is incorrect, that's why the `changelog` is failing. Please recheck.

ui/src/components/TaskRunsCard/index.jsx
```diff
@@ -485,15 +485,15 @@ export default class TaskRunsCard extends Component {
                 onClick={this.handleNext}
                 disabled={run ? selectedRunId === runs.length - 1 : true}>
                 Next
-                <ChevronRightIcon />
+                <ChevronRightIcon color="rgb(149, 27, 27)" />
```
vishakhanihore: I think adding custom `RGB Codes` is not an optimal solution for the following. You can check other icons implementation and then do the same.

changelog/issue-1773.md
aryamanpuri: Okay , but after removing the first line , it is still failing the yarn  changelog check

ui/src/components/TaskRunsCard/index.jsx
aryamanpuri: Yes, it is not an optimal solution, but the other icons still differ in the color only, it is a minor problem related to the icons not visible in the light/dark theme. So this can be handled by custom RGB codes.

changelog/issue-1773.md
aryamanpuri: @vishakhanihore

vishakhanihore:
Squash your commits and try, should be fixed after I think

helfi92:
Let me know when this is ready for review.

PR #1830, opened 2019-10-30
Update styleguidist version to v9

PaliwalArpit:


PaliwalArpit:
I am not sure why these checks are failing, its just version upgrade that is being done, any idea?

helfi92:
> I am not sure why these checks are failing, its just version upgrade that is being done, any idea?
In ideal case yarn.lock should be in gitignore isn't it?

`yarn.lock` is important to have to ensure consistent installs across machines (https://yarnpkg.com/lang/en/docs/yarn-lock/). `yarn add react-styleguidist` should update yarn.lock.

The problem all of this is not working is because there has been some breaking changes which will require changing our source code. Did you check for the breaking changes in https://github.com/styleguidist/react-styleguidist/releases since v7? Did you find any breaking changes?

PaliwalArpit:
@helfi92 Hey, I can see [PR](https://github.com/taskcluster/taskcluster/pull/1834) for upgrading to v10, there as well we have check failure for taskcluster-ui. Are we skipping v9 all together?
Still i did some digging here inside logs:

Inside log and in code the error we are encountering is in :
`error: Unable to resolve path to module 'react-styleguidist/lib/rsg-components/StyleGuide/StyleGuideRenderer' (import/no-unresolved) at src/styleguide/StyleGuideRenderer.jsx:2:24:

  1 | import React, { Fragment } from 'react';
> 2 | import StyleGuide from 'react-styleguidist/lib/rsg-components/StyleGuide/StyleGuideRenderer';
    |                        ^
  3 | import FontStager from '../components/FontStager';
  4 | 
  5 | const StyleGuideRenderer = props => (`

 Found one possible solution for this as : [https://github.com/styleguidist/react-styleguidist/issues/1051](https://github.com/styleguidist/react-styleguidist/issues/1051)

helfi92:
> @helfi92 Hey, I can see PR for upgrading to v10, there as well we have check failure for taskcluster-ui. Are we skipping v9 all together?

Sure. If it was me, I would probably do the following:

1. Upgrade v7 to v8 locally and check that the styleguide is working
2. Upgrade v8 to v9 locally and check that the styleguide is working
3. Upgrade v9 to v10 locally and check that the styleguide is working
4. Push my changes

To check that the styleguide is working, I would

1. Inside the `ui` directory, I would run

```
yarn styleguide
```

2. Navigate to http://localhost:6060 and make sure it works

helfi92:
> Found one possible solution for this as : styleguidist/react-styleguidist#1051

Sure, if that fixes it, then that's good :)

PaliwalArpit:
@helfi92 What are your thoughts on v9 style guide? Are we keeping this or version upgrade work well for you to v10?
Please let me know what can be done with this PR, I will close otherwise to avoid cluttering.

helfi92:
Upgrading to v10 works for me.

PR #1837, opened 2019-10-30
Fix firing-hooks link

djmitche:
(mostly testing to make sure I've put back the secrets that I deleted..)


djmitche:
Looks like I got all of them!

PR #1856, opened 2019-10-31
[UI] Change "More..." to "See More" and add icon #1822

### LINKED ISSUE #1822: [UI] Change "More..." to "See More" and add icon
**To Reproduce**

1. Go to http://localhost:5080/tasks/eWxW0D38QgOv9WZNKgCgAg
2. Look at the last list item of both cards

#### Actual

Screen Shot 2019-10-29 at 1 32 35 PM

#### Expected

Instead of "More...", I was thinking of changing the text to "See More". On the left side of the text, we can insert an icon from https://materialdesignicons.com/. Surprise me with an icon that you think makes sense :)

_Note: When you open a pull request, please put a screenshot of the new change._

km-js:
Fixes #1822 

Changed the text from `More..` and` Less..` to `See More` and `See Less`.
`
Also, added the plus and minus icons in front of the text.

**Before Changes applied:**

Screenshot_44

**After Changes applied:**

Screenshot_45

### DIFF
--- a/changelog/issue-1822.md
+++ b/changelog/issue-1822.md
@@ -1,0 +1,3 @@
+level: silent
+reference: issue 1822
+---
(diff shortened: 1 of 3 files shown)


helfi92:
For the list item text, you can do `<ListItemText disableTypography primary={<Typography variant="subtitle1" color="textSecondary">See More</Typography>} />`

km-js:
> 
> 
> I thought about it a little bit and I think we should do something more like:
> 
> Screen Shot 2019-11-04 at 11 45 08 AM
> 
> Sorry for not saying this earlier. What you did looks nice too :)

It's ok. It's looking better now :)

PR #1859, opened 2019-11-01
Create an indexed task and find it smoketest

### LINKED ISSUE #1244: smoketest - create an indexed task and find it in the index
Add a new test to the `yarn smoketest` by adding a module to `infrastructure/tooling/src/smoketest/checks/`.  In this case, it would create a task with routes for the index service, and once that task is completed look to see that the task is correctly shown in the index.

See #1218.

Rolikasi:
Add smoketest that creates indexed task and finds it in the index
Closes #1244


Rolikasi:
Is it ok that these smoketest indexed tasks expire in a year?

djmitche:
> Is it ok that these smoketest indexed tasks expire in a year?

An hour would be better :)

infrastructure/tooling/src/smoketest/checks/indexTask.js
```diff
@@ -0,0 +1,48 @@
+const taskcluster = require('taskcluster-client');
+
+exports.tasks = [];
+exports.tasks.push({
+  title: 'create an indexed task and find it in the index',
+  requires: [],
+  provides: [
+    'target-indexTask',
```
djmitche: let's use `-` here, so `target-indexed-task`

infrastructure/tooling/src/smoketest/checks/indexTask.js
```diff
@@ -0,0 +1,48 @@
...
+        owner: "smoketest@taskcluster.net",
+        source: "https://taskcluster.net",
+      },
+      payload: {},
+      routes: [`index.project.taskcluster.smoketest.${randomId}`],
+    };
+    utils.status({message: 'indexTask-find taskId: ' + randomId});
+    await queue.createTask(randomId, task);
+    let index = new taskcluster.Index(taskcluster.fromEnvVars());
+    index.findTask(findIndexedTask);
```
djmitche: This call must be `await`ed

infrastructure/tooling/src/smoketest/checks/indexTask.js
```diff
@@ -0,0 +1,48 @@
...
+    let pollForStatusStart = new Date();
+    while((new Date() - pollForStatusStart) < 120000){
+      let status = await queue.status(randomId);
+      if (status.status.state === 'pending' || status.status.state === 'running'){
+        utils.status({
+          message: 'Current status: ' + status.status.state,
+        });
+        await new Promise(resolve => setTimeout(resolve, 1000));
+      } else if (status.status.state === "completed") {
+        return;
```
djmitche: I think we would want to check here that the task is in the index?

Note that the task is indexed *after* it completes, so it may be necessary to wait another few seconds before `findTask` finds the task.

Rolikasi:
@djmitche kindly review the changes) I think it's checking now that the task is in the index)

infrastructure/tooling/src/smoketest/checks/indexTask.js
```diff
@@ -0,0 +1,50 @@
+const taskcluster = require('taskcluster-client');
+
+exports.tasks = [];
```
djmitche: If you rebase onto the latest master, you'll see this needs a `scopeExpression` giving the scopes it requires.

infrastructure/tooling/src/smoketest/checks/indexTask.js
```diff
@@ -0,0 +1,50 @@
...
+      deadline: taskcluster.fromNowJSON('2 minutes'),
+      expires: taskcluster.fromNowJSON('60 minutes'),
+      metadata: {
+        name: "Smoketest indexTask-find",
+        description: "built-in/succeed task created during smoketest",
+        owner: "smoketest@taskcluster.net",
+        source: "https://taskcluster.net",
+      },
+      payload: {},
+      routes: [`index.project.taskcluster.smoketest.${randomId}`],
```
djmitche:
```suggestion
      routes: [`index.${taskIndex}`],
```

infrastructure/tooling/src/smoketest/checks/indexTask.js
```diff
@@ -0,0 +1,50 @@
...
+exports.tasks.push({
+  title: 'create an indexed task and find it in the index',
+  requires: [],
+  provides: [
+    'target-indexed-task',
+  ],
+  run: async (requirements, utils) => {
+    let queue = new taskcluster.Queue(taskcluster.fromEnvVars());
+    let randomId = taskcluster.slugid();
+    const findIndexedTask='project.taskcluster.smoketest.' + randomId;
```
djmitche: maybe call this variable `taskIndex`?

infrastructure/tooling/src/smoketest/checks/indexTask.js
```diff
@@ -0,0 +1,50 @@
...
+      let finded = await index.findTask(findIndexedTask);
+      if (status.status.state === 'pending' || status.status.state === 'running'){
+        utils.status({
+          message: 'Current status: ' + status.status.state,
+        });
+        await new Promise(resolve => setTimeout(resolve, 1000));
+      } else if (finded) {
+        return;
+      } else {
+        throw new Error('Task finished with status ' + status.status.state);
```
djmitche: I think we'll end up here if we happen to poll *after* the task has completed but before the index service has inserted it into the index.  That's not a failure case -- it's a case where we should wait 1s and try again.

Rolikasi:
@djmitche kindly review the changes)

infrastructure/tooling/src/smoketest/checks/indexTask.js
```diff
@@ -0,0 +1,60 @@
...
+      if (status.status.state === 'pending' || status.status.state === 'running'){
+        utils.status({
+          message: 'Current status: ' + status.status.state,
+        });
+        await new Promise(resolve => setTimeout(resolve, 1000));
+      } else if (finded) {
+        return;
+      } else {
+        await new Promise(resolve => setTimeout(resolve, 1000));
+      }
```
djmitche: This conditional could be refactored a little by moving the sleep outside of the conditional, since it happens in two of the three branches of control flow, while the third branch returns.

infrastructure/tooling/src/smoketest/checks/indexTask.js
```diff
@@ -0,0 +1,60 @@
...
+      routes: [`index.${taskIndex}`],
+    };
+    utils.status({message: 'indexTask-find taskId: ' + randomId});
+    await queue.createTask(randomId, task);
+    let index = new taskcluster.Index(taskcluster.fromEnvVars());
+    await new Promise(resolve => setTimeout(resolve, 5000));
+    let pollForStatusStart = new Date();
+    while((new Date() - pollForStatusStart) < 120000){
+      let status = await queue.status(randomId);
+      let finded = await index.findTask(taskIndex);
```
djmitche: "found" is the typical English word.

infrastructure/tooling/src/smoketest/checks/indexTask.js
```diff
@@ -0,0 +1,60 @@
...
+        owner: "smoketest@taskcluster.net",
+        source: "https://taskcluster.net",
+      },
+      payload: {},
+      routes: [`index.${taskIndex}`],
+    };
+    utils.status({message: 'indexTask-find taskId: ' + randomId});
+    await queue.createTask(randomId, task);
+    let index = new taskcluster.Index(taskcluster.fromEnvVars());
+    await new Promise(resolve => setTimeout(resolve, 5000));
```
djmitche: I commented this out, and the test failed.  That means that something in this test is dependent on timing, and that often leads to intermittent test failures.  Can you remove this and fix the issue causing the test to fail?

Rolikasi:
After weekend update I guess there is no role that give me the necessary scopes to run the smoketests, am I right?

infrastructure/tooling/src/smoketest/checks/indexTask.js
Rolikasi: oups

Rolikasi:
The problem was that I tried to find a task before its state was "completed". I made some changes now I think it works correctly

Rolikasi:
@djmitche kindly review the changes)

infrastructure/tooling/src/smoketest/checks/indexTask.js
```diff
@@ -0,0 +1,58 @@
...
+      let status = await queue.status(randomId);
+      if (status.status.state === 'pending' || status.status.state === 'running'){
+        utils.status({
+          message: 'Current task status: ' + status.status.state,
+        });
+      } else if (status.status.state === 'completed') {
+        await index.findTask(taskIndex);
+        return;
+      } else {
+        await new Promise(resolve => setTimeout(resolve, 1000));
```
djmitche: This will only happen in the case where the task is not pending or running, meaning that the check will busy-loop without any sleeping while the task is pending or running.

Rolikasi:
Thank for the review @djmitche) 
Added try statement to prevent errors when task could not be indexed right after it's completed
Moved sleep out of the condition
Deleted else condition

PR #1869, opened 2019-11-02
Bug 1591591 - use APPLICATION_NAME for GitHub status and check context

djmitche:
Bugzilla Bug: [1591591](https://bugzilla.mozilla.org/show_bug.cgi?id=1591591)

@imbstack is there a better way to phrase this changelog for deployers?  Does this linking of APPLICATION_NAME to github status even make sense?


djmitche:
@owlishDeveloper FYI regarding this minor change in tc-github.

PR #1882, opened 2019-11-04
Bug 1593754 - use pulse.namespace for the pulse namespace

djmitche:
Bugzilla Bug: [1593754](https://bugzilla.mozilla.org/show_bug.cgi?id=1593754)

It turns out this was just a bug where we were using the wrong config value.

### DIFF
--- a/services/web-server/src/main.js
+++ b/services/web-server/src/main.js
@@ -56,7 +56,7 @@ const load = loader(
 
         return new Client({
           monitor: monitor.childMonitor('pulse-client'),
-          namespace: cfg.pulse.username,
+          namespace: cfg.pulse.namespace,
           credentials: pulseCredentials(cfg.pulse),
         });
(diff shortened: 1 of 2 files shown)


services/web-server/src/main.js
```diff
@@ -56,7 +56,7 @@ const load = loader(
 
         return new Client({
           monitor: monitor.childMonitor('pulse-client'),
-          namespace: cfg.pulse.username,
+          namespace: cfg.pulse.namespace,
```
helfi92: Other services hard-code this value. Should we do the same and the remove `namespace` from the config file?

djmitche: I think that will make it harder to run this service locally against a pulse service.  If that's not a problem, then we could hard-code.  But I think this change will be sufficient to work in production.

PR #1883, opened 2019-11-04
Display pattern on hover 

### LINKED ISSUE #1455: Pattern does not have way to view whole if cut off
**Describe the bug**
In the schema viewers of the docs site, if a string has a pattern that is too long, it will be cut off and an ellipsis will be shown. This is great, the only issue is that there's no way I can find to see the entire thing

**To Reproduce**
1. Go to [the listTaskGroup api doc](https://docs.taskcluster.net/docs/reference/platform/queue/api#listTaskGroup)
2. Scroll down to `taskGroupId` in the Response Payload
4. See that `^[A-Za-z0-9_-]{...` has no way to expand to the entire thing.

**Expected behavior**
It would be nice if there was a tooltip on hover or something of the sort.

**Screenshots**

image

 - OS: ubuntu
 - Browser: firefox
 - Version: 71.0a1 (2019-10-01) (64-bit)

Vane237:
Github Bug/Issue: Fixes #1455

### DIFF
--- a/changelog/bug-1593754.md
+++ b/changelog/bug-1593754.md
@@ -1,0 +1,4 @@
+level: patch
+reference: bug 1593754
+---
+The web-server service now uses the correct Pulse namespace to listen for pulse messages.  This fixes one more bug preventing task and task-group UI from dynamically updating.
--- a/services/web-server/src/main.js
+++ b/services/web-server/src/main.js
@@ -56,7 +56,7 @@ const load = loader(
 
         return new Client({
           monitor: monitor.childMonitor('pulse-client'),
-          namespace: cfg.pulse.username,
+          namespace: cfg.pulse.namespace,
           credentials: pulseCredentials(cfg.pulse),
         });
       },


helfi92:
Can you post a screenshot of how that looks?

Vane237:
Screen Shot 2019-11-05 at 15 56 43

ui/src/App/Main.jsx
```diff
@@ -45,6 +45,9 @@ import isLoggedInQuery from './isLoggedIn.graphql';
     ':not(pre) > code': {
       ...theme.mixins.highlight,
     },
+    '.mmJN4cVbficDwBMDQAPWE': {
+      position: 'absolute',
+    },
```
helfi92: Nice find! Seems to work but we should probably add this change in https://github.com/taskcluster/react-schema-viewer so that it works for everyone using that library, not just here. Does this make sense to you? Let me know if you had questions.

Vane237: Yeah that's true

helfi92: > To achieve that, should I instead fork that repository, add the fix and create a PR?

Sounds like a good idea  to me. I would fork then clone the repo.

Vane237: Here's a link to the PR
https://github.com/taskcluster/react-schema-viewer/pull/37

helfi92:
@Vane237 All good for sending a PR updating the package version? You'll have to do it on a separate branch, not master.

Vane237:
> @Vane237 All good for sending a PR updating the package version? You'll have to do it on a separate branch, not master.

Oops.. I already did it on the master branch

PR #1885, opened 2019-11-05
Bug 1593809 - github.bot_username and 404's from github.latest

djmitche:
Bugzilla Bug: [1593809](https://bugzilla.mozilla.org/show_bug.cgi?id=1593809)

This does two linked things:

 * Takes BOT_USERNAME as a config, since it is different per deployment.  From GitHub's perspective, the username is the App name plus `[bot]`, e.g., `cloudopsstage-taskcluster[bot]`.  Partially to simplify configuration and partially because @edunham already set all of the config to names lacking `[bot]`.
 * Handles 404's from the github API method by passing them along to the caller -- both for badges (with a new-repo svg) and for latest.

### DIFF
--- a/changelog/bug-1593809.md
+++ b/changelog/bug-1593809.md
@@ -1,0 +1,5 @@
+level: patch
+reference: bug 1593809
+---
+The taskcluster-github service now correctly uses the `github.bot_username` configuration to look up the latest status for a branch.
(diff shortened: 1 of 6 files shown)


djmitche:
I'm open to just requiring the `[bot]` in the config.  Often being too flexible in what's allowed in a config hurts more than it helps.

PR #1902, opened 2019-11-05
Fix clean up of role smoketest

ajinkabeer:
This PR is a follow up to the Add cleanup for role smoketest #1776 pull request which traces back to the original issue #1246. The issue for not cleaning up was with the time condition.

- Updated the for loop syntax

- Cleans up by deleting the role if the last modified time is 1 hour or more

@djmitche

### DIFF
--- a/infrastructure/tooling/src/smoketest/checks/role.js
+++ b/infrastructure/tooling/src/smoketest/checks/role.js
@@ -13,7 +13,7 @@ exports.tasks.push({
     const randomId = taskcluster.slugid();
     const roleId = `project:taskcluster:smoketest:${randomId}:*`;
     const payload = {
-      description: 'smoketest test',
+      description: 'smoketest for creating a role and expanding it',
       scopes: ['project:taskcluster:smoketest:<..>/*'],
     };
     const expandPayload = {
@@ -31,9 +31,9 @@ exports.tasks.push({
     const anHourAgo = Date.now() - (1000*60*60);
     while (1) {
       const res = await auth.listRoles2();
-      for(let i=0;i<res.roles.length;i++){
-        if(res.roles[i].roleId.includes('project:taskcluster:smoketest:') && res.roles[i].lastModified < anHourAgo){
-          await auth.deleteRole(res.roles[i].roleId);
+      for(let role of res.roles){
+        if(role.roleId.includes('project:taskcluster:smoketest:') && role.lastModified < new Date(anHourAgo)){
+          await auth.deleteRole(role.roleId);
         }
       }
       if (res.continuationToken) {


infrastructure/tooling/src/smoketest/checks/role.js
```diff
@@ -28,12 +28,16 @@ exports.tasks.push({
...
       const res = await auth.listRoles2();
-      for(let i=0;i<res.roles.length;i++){
-        if(res.roles[i].roleId.includes('project:taskcluster:smoketest:') && res.roles[i].lastModified < anHourAgo){
-          await auth.deleteRole(res.roles[i].roleId);
+      for(let index in res.roles){
+        if(res.roles[index].roleId.includes('project:taskcluster:smoketest:')){
+          const lastModified = (new Date(res.roles[index].lastModified)).getHours();
+          if(currentHour - lastModified >= 1){
+            await auth.deleteRole(res.roles[index].roleId);
+          }
```
djmitche: This only compares the hours, so for example something created at 23:00 two days ago would not be deleted if it's currently 7:00.

I think the issues in the previous version were around the types:

```
> res_roles_index_lastModified = '2019-10-29T23:23:31.391Z'
'2019-10-29T23:23:31.391Z'
> anHourAgo = Date.now() - (1000*60*60)
1572979864372
> res_roles_index_lastModified < anHourAgo
false
> new Date(res_roles_index_lastModified) < new Date(anHourAgo)
true
> 
```
so I think you could fix it by just calling `new Date(..)` with both the numeric value used to calculate `anHourAgo` and the string value we get in `res.roles[index].lastModified`.

infrastructure/tooling/src/smoketest/checks/role.js
```diff
@@ -28,12 +28,16 @@ exports.tasks.push({
...
     const query = {};
-    const anHourAgo = Date.now() - (1000*60*60);
+    const date = new Date();
+    const currentHour = date.getHours();
     while (1) {
       const res = await auth.listRoles2();
-      for(let i=0;i<res.roles.length;i++){
-        if(res.roles[i].roleId.includes('project:taskcluster:smoketest:') && res.roles[i].lastModified < anHourAgo){
-          await auth.deleteRole(res.roles[i].roleId);
+      for(let index in res.roles){
```
djmitche: Even better: `for (let role of res.roles)` -- no need for indexing at all!  The construct you've used actually iterates over the properties of `res.roles`, which for an array happen to be the array indexes, but *as strings* and *not in numeric order*.

infrastructure/tooling/src/smoketest/checks/role.js
ajinkabeer: Yes. That is correct. Thank you.

infrastructure/tooling/src/smoketest/checks/role.js
ajinkabeer: This is so much better. Thank you very much!

djmitche:
Nice fix -- thank you!!

ajinkabeer:
> Nice fix -- thank you!!

Really appreciate your help!

PR #1913, opened 2019-11-06
[UI] Using same color for pending and unscheduled is misleading

### LINKED ISSUE #1906: [UI] Using same color for pending and unscheduled is misleading
Two different states should not be represented with the same color. I'm not good at color picking, but grey for unscheduled seems okay. Could we get another color for pending?

km-js:
Fixes #1906

Changes Applied:
Changed the color of Pending status from grey to amber which is a warning color.
I hope that makes sense if not, please suggest me something better.


ui/src/utils/labels.js
```diff
@@ -1,6 +1,6 @@
 export default {
   RUNNING: 'info',
-  PENDING: 'default',
+  PENDING: 'warning',
```
helfi92: We use `warning` for exception already. Also, the filter card color should change as well:

Screen Shot 2019-11-06 at 1 37 10 PM.

km-js: Yeah, but all others are also in use already. What do you suggest in that case?

helfi92: That's a good question. We could probably just use a class to override the color.

`<StatusLabel className={...} />`

km-js: Hm.. But Isn't `StatusLabel `is used in various components?

km-js:
Hey @helfi92 , any reviews on this?

km-js:
> 
> 
> Pending has a yellow background at the top but the status label shows it as grey. Let's fix that.
> 
> Screen Shot 2019-11-12 at 9 26 55 AM

Hey, Can you help a bit on how to do that?
 As pending is coming from the state here:

`<StatusLabel state={taskGroup.status.state} />`

helfi92:
You should be able to add a className:

```jsx
{
  background: `... !important`,
}

<StatusLabel state={taskGroup.status.state} className={...} />
```

km-js:
Won't this change every status's background color to yellow rather than just 'pending'???

helfi92:
> Won't this change every status's background color to yellow rather than just 'pending'???

Yes. You'll need to add to conditionally add this class.

km-js:
Hey @helfi92 , Are there any other changes required here?

Issue #1554, opened 2019-10-09
[web-server] implement an azure session store for express-session


helfi92:
We currently store sessions in memory using [memorystore](https://github.com/roccomuso/memorystore#readme).

Problems:
* Restarting the service clears all user sessions
* Spinning up multiple web-server services for load balancing is not possible since memory sessions belong to a single instance.

One solution would be to implement an azure session store for express-session that is compatible with shared access signatures credentials.

PR #1923, opened 2019-11-07
Add an azure session store for express-session

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/1554.

https://github.com/expressjs/session#session-store-implementation

### DIFF
--- a/services/web-server/src/login/AzureSessionStore.js
+++ b/services/web-server/src/login/AzureSessionStore.js
@@ -1,6 +1,5 @@
 const assert = require('assert');
 const taskcluster = require('taskcluster-client');
-const tryCatch = require('../utils/tryCatch');
 const { NOOP } = require('../utils/constants');
 
 module.exports = function ({ session, SessionStorage, options = {} }) {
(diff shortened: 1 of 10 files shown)


djmitche:
I :heart: the PR title but don't want it to land before tomorrow so I haven't looked at the diff yet.

services/web-server/src/login/AzureSessionStore.js
```diff
@@ -0,0 +1,98 @@
...
+
+  return class AzureSessionStore extends Store {
+    /*
+     * required
+     *
+     * This is used to destroy/delete a session from the store given a session ID (sessionId).
+     * The callback should be called as callback(error) once the session is destroyed.
+     */
+    async destroy(sessionId, callback = NOOP) {
+      await SessionStorage.remove({ sessionId }, true);
```
djmitche: If this fails, it will be an UnhandledPromiseRejection and the callback will never be called.  I feel like we've seen this issue in a few places (oauth2.js comes to mind).  Maybe there's a general solution for de-promisifying a function like this?

services/web-server/test/session_store_test.js
```diff
@@ -0,0 +1,137 @@
+const assert = require('assert');
+const testing = require('taskcluster-lib-testing');
+const session = require('express-session');
+const { promisify } = require('util');
+const helper = require('./helper');
+const AzureSessionStore = require('../src/login/AzureSessionStore');
+
+helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping) {
```
djmitche: woo!!

services/web-server/src/login/AzureSessionStore.js
helfi92: Good observation. The immediate fix that comes to mind is to wrap the method with a try catch method.

PR #1926, opened 2019-11-08
Bug 1588099 - Add simplifyScopeExpression function 

ajinkabeer:
This PR adds a function `simplifyScopeExpression` in `libraries/scopes`.  The scope expression

```
{
  "AllOf": [
    {
      "AllOf": [
        "queue:create-task:highest:built-in/succeed",
        "queue:create-task:highest:built-in/fail",
        "queue:scheduler-id:-"
      ]
    },
    {
      "AllOf": [
        "secrets:get:project/taskcluster/smoketest/*",
        "secrets:set:project/taskcluster/smoketest/*"
      ]
    }
  ]
}
```

Can be simplified to just :

```
{
      "AllOf": [
        "queue:create-task:highest:built-in/succeed",
        "queue:create-task:highest:built-in/fail",
        "queue:scheduler-id:-"
        "secrets:get:project/taskcluster/smoketest/*",
        "secrets:set:project/taskcluster/smoketest/*"
      ]
}
```

@djmitche


libraries/scopes/src/simplify.js
```diff
@@ -0,0 +1,2 @@
+exports.simplifyScopeExpression = (scope) =>
+  scope === 'First test' || 'Second test';
```
djmitche: The first step is to create a function that returns the correct *type* of data but without any logic.  What you've written here returns a boolean or a string, but the function should return a scope expression.

libraries/scopes/test/simplify_test.js
```diff
@@ -0,0 +1,13 @@
+const assert = require('assert');
+const utils = require('../src');
+const testing = require('taskcluster-lib-testing');
+
+suite(testing.suiteName(), function() {
+  test('Simplify Scope Expression test ', function() {
+    assert(utils.simplifyScopeExpression('First test'));
```
djmitche: I think you'll want to use `assert.deepEqual`.  Here's an example test:

```js
test('single scope simplifies to itself', function() {
  assert.equal(simplifyScopeExpression("scope:1"), "scope:1");
});
```
then update the simplifyScopeExpression function to pass that test.  Then add another test.  It's good to start with the "base cases" for any recursive data structure.  From [here](https://github.com/taskcluster/taskcluster/tree/master/libraries/scopes#new-style) we can see that the single scope is one of three options, with the other two being AnyOf and AllOf.  And for the moment we can stick to cases where no simplification is possible.

```js
test('AnyOf simplifies to itself', function() {
  assert.equal(
    simplifyScopeExpression({AnyOf: ["scope:1", "scope:2"]}),
    {AnyOf: ["scope:1", "scope:2"]});
});

test('AllOf simplifies to itself', function() {
  assert.equal(
    simplifyScopeExpression({AllOf: ["scope:1", "scope:2"]}),
    {AllOf: ["scope:1", "scope:2"]});
});
```

Once you've got that working, here are a few more advanced tests to work on:

```js
test('AnyOf list of scopes duplicates are removed', function() {
  assert.equal(
    simplifyScopeExpression({AnyOf: ["scope:1", "scope:2", "scope:1"]}),
    {AnyOf: ["scope:1", "scope:2"]});
});

test('AllOf list of scopes duplicates are removed', function() {
  assert.equal(
    simplifyScopeExpression({AllOf: ["scope:1", "scope:2", "scope:1"]}),
    {AllOf: ["scope:1", "scope:2"]});
});

test('AllOf list of scopes is normalized', function() {
  // note that AnyOf lists of scopes should *not* be normalized this way!
  assert.equal(
    simplifyScopeExpression({AllOf: ["scope:1", "scope:2", "scope:*"]}),
    {AllOf: ["scope:*"]});
});
```

ajinkabeer:
> This would be a good use for a "Draft" PR.

Yes. I did not know about draft PR's. Should we continue with this PR itself?
@djmitche

djmitche:
Also we are realy busy today preparing for a major downtime tomorrow, so I won't be able to review further until Monday or Tuesday.

ajinkabeer:
> Also we are realy busy today preparing for a major downtime tomorrow, so I won't be able to review further until Monday or Tuesday.

That's okay. I hope I can make some progress in the coming days. Thanks for telling me though. :)

djmitche:
Continuing in this PR is fine.

ajinkabeer:
I have updated the ` simplifyScopeExpression` function to  check if it passes the base test cases. I am not sure if this is the right way. So can you please give me your suggestions?

ajinkabeer:
> I think you've switched the tests and the code being tested! The code being tested is doing the assert, whereas the tests should be doing the asserts.

I think that was foolish move from my side. Now I understood the approach. I will change my function to test the advance cases also. :)

ajinkabeer:
> Looking better!
> 
> https://github.com/taskcluster/taskcluster/tree/master/libraries/scopes#sorting-merging-and-normalizing might come in handy for removing duplicates.

Yay! , I will check it out then.

ajinkabeer:
> Looking better!
> 
> https://github.com/taskcluster/taskcluster/tree/master/libraries/scopes#sorting-merging-and-normalizing might come in handy for removing duplicates.

I did look at it. I couldn't get to work so then I thought I will write a function to do it. I don't know if this is a good approach. It does remove duplicate scopes so I pushed it to get your review. What do you think?
@djmitche

libraries/scopes/src/simplify.js
```diff
@@ -0,0 +1,22 @@
...
+  console.log(scopeSet[0]);
+  return scopeSet[0];
+};
+
+function removeDuplicateScopes(scope){
+  if(Object.keys(scope).includes('AnyOf') || Object.keys(scope).includes('AllOf')){
+    const values = (Object.values(scope)[0]);
+    if(ifDuplicateExists(values)){
+      let unique = [...new Set(values)];
+      const uniqueValues = unique.join(', ');
```
djmitche: I'm not sure why the join here?  They should remain distinct strings..

libraries/scopes/test/simplify_test.js
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  test('AllOf simplifies to itself', function() {
+    assert.deepStrictEqual(
+      utils.simplifyScopeExpression({AllOf: ["scope:1", "scope:2"]}),
+      {AllOf: ["scope:1", "scope:2"]});
+  });
+
+  test('AnyOf list of scopes duplicates are removed', function() {
+    assert.notDeepEqual(
```
djmitche: "not"??

libraries/scopes/src/simplify.js
```diff
@@ -0,0 +1,22 @@
...
+  scopeSet.push(scope);
+  removeDuplicateScopes(scope);
+  console.log(scopeSet[0]);
+  return scopeSet[0];
+};
+
+function removeDuplicateScopes(scope){
+  if(Object.keys(scope).includes('AnyOf') || Object.keys(scope).includes('AllOf')){
+    const values = (Object.values(scope)[0]);
+    if(ifDuplicateExists(values)){
```
djmitche: I like the approach of `[...new Set(values)]` to remove duplicates!  However, given a list without duplicates that will just reproduce a list with the same items, so there's no need to test for duplicates beforehand -- just always use the approach.

I suspect you added this because sometimes the list of scopes came out in a different order?  For consistency, you could sort the scopes -- that's a pretty common tactic in TDD.

The normalizeScopeSet function will both sort and remove duplicates.. I'm curious what didn't work with those?

libraries/scopes/src/simplify.js
ajinkabeer: Yes, that is correct. But here are my observations with and without `join()`, without `join()`, the splice function creates a nested array. I will upload the logs here. 

Without `join(()`
Capture1

With join()

libraries/scopes/test/simplify_test.js
ajinkabeer: Uh oh!, that should be `strictDeepEqual`

libraries/scopes/src/simplify.js
ajinkabeer: Initially I tried it without the `join()` which resulted in nested arrays. Presently, this is my output.

Capture1

They are two different strings now. The problem now is nested arrays. Phew!

libraries/scopes/src/simplify.js
ajinkabeer: Yes true. When I added this because the `normalizeScopeSet ` function did not work because I didn't add the sort function. Now its working. The only problem is I have a nested array.

newone

ajinkabeer: Plus, netlify tests are failing because of the following error:

> Module not found: Error: Can't resolve 'taskcluster-lib-scopes' in '/opt/build/repo/libraries/scopes/src'_

libraries/scopes/src/simplify.js
```diff
@@ -0,0 +1,23 @@
+let scopeUtils = require('taskcluster-lib-scopes');
```
djmitche: I suspect this is why netlify is failing: it's a circular requirement.  Instead:
```suggestion
const {normalizeScopeSet, scopeCompare} = require('./normalize');

```

libraries/scopes/src/simplify.js
```diff
@@ -0,0 +1,23 @@
...
+  return scope;
+};
+
+function removeDuplicateScopes(scope){
+  const values = (Object.values(scope)[0]);
+  if(Object.keys(scope).includes('AnyOf') || Object.keys(scope).includes('AllOf')){
+    values.sort(scopeUtils.scopeCompare);
+    scopeUtils.normalizeScopeSet(values);
+    let normalized = scopeUtils.normalizeScopeSet(values);
+    Object.values(scope)[0].splice(0, values.length, normalized);
```
djmitche: `splice` is a dangerous tool to use - it modifies its argument, which means that `removeDuplicates` will modify its input value, rather than returning a fresh new array.  It's generally better to just use assignment here.
```suggestion
    scope[Object.keys(scope)[0]] = normalized;
```

libraries/scopes/src/simplify.js
ajinkabeer: Thanks, I think that should fit it!

libraries/scopes/src/simplify.js
ajinkabeer: Oh I see, thanks for that!
@djmitche

ajinkabeer: I have added all the tests that you mentioned and I would really like to know what do you think please.
@djmitche

libraries/scopes/src/simplify.js
```diff
@@ -0,0 +1,35 @@
+const {normalizeScopeSet, scopeCompare} = require('./normalize');
+
+exports.simplifyScopeExpression = (scope) => {
+  const values = (Object.values(scope)[0]);
+  if(duplicateExists(values)){
+    removeDuplicateScopes(scope);
+  }
```
djmitche: Normalizing will remove duplicates, so there's no need to do this too.

libraries/scopes/src/simplify.js
```diff
@@ -0,0 +1,35 @@
+const {normalizeScopeSet, scopeCompare} = require('./normalize');
+
+exports.simplifyScopeExpression = (scope) => {
```
djmitche:
```suggestion
exports.simplifyScopeExpression = (scopeExpression) => {
```

libraries/scopes/src/simplify.js
ajinkabeer: Yes, exactly. So I can remove the whole function now. Thanks!

libraries/scopes/src/simplify.js
ajinkabeer: Good change!

ajinkabeer:
> Looking good! Here's a few more test cases:
> 
> * `{AnyOf: ["scope1", "scope2", "scope*"]}` should simplify to `{AnyOf: ["scope*"]}`
> * `{AnyOf: ["scope1", {AllOf: ["scope2a", "scope2b, "scope2*", "scope3"]}]}` should simplify to `{AnyOf: ["scope1", {AllOf:["scope2*", "scope3"]}]}`.
> 
> Let's start getting into the original idea of the bug:
> 
> * `{AllOf: [{AllOf: ["scope1", "scope2"]}, {AllOf: ["scope2", "scope3"]}]}` should simplify to `{AllOf: ["scope1", "scope2", "scope3"]}`.
> * `{AllOf: [{AllOf: ["scope1", "scope2"]}, "scope2", "scope3"]}` should simplify to `{AllOf: ["scope1", "scope2", "scope3"]}`.
> * `{AllOf: ["scope0", {AllOf: ["scope1", {AllOf: ["scope2", {AllOf: ["scope3, {AllOf: ["scope4", "scope5"]}]}]}]}]}` should simplify to `{AllOf: ["scope0", "scope1", "scope2", "scope3", "scope4", "scope5"]}`.
> * `{AnyOf: ["scope1"]}` should simplify to `"scope1"`
> * `{AllOf: ["scope1"]}` should simplify to `"scope1"`
> * `{AllOf: ["scope1", {AnyOf: ["scope*"]}]}` should simplify to `"scope*"`
> 
> This is starting to get tough, now, so balance your time between getting each test case to work (without breaking the others) and going back to think about the original problem. The danger is that you'll write code to pass the specific tests you've got so far, but that doesn't implement the full meaning of the function. An extreme case would be something like
> 
> ```js
> exports.simplifyExpression = (scopeExpression) => {
>   if (_.is(scopeExpression, {AllOf: ["scope1", {AnyOf: ["scope*"]}]})) {
>     return "scope*"
>   } else if (..) {
>     return ..
>   } else {
>     throw new Error("unrecognized scopeExpression");
>   }
> }
> ```
> 
> but of course then I can just think up one more perfectly valid test case that this function won't pass. We can keep going like that for a while.
> 
> I think TDD is fun -- and I don't often get a chance to do it. You can see another example of this in https://github.com/taskcluster/json-e where we have about 1000 test cases. Adding a new feature to that library always involves writing a bunch of new test cases before implementing the feature.

Yes, I understand. Now we have to focus back on to our original problem. I think its going to get tough. But I'll surely won't back down from this challenge because this is my first time doing a TDD and I think its one of the efficient ways to add new features and I am really enjoying it right now.

I will definitely checkout the link that you have mentioned and I will continue to work on this problem!. Thanks for the tips. Appreciate it!
:)

ajinkabeer:
Did not finish it yet, but managed to pass two tests based on our original problem. I would like to know what you think please. Its getting tough as we get going! Phew!  :)
@djmitche

libraries/scopes/src/simplify.js
```diff
@@ -0,0 +1,58 @@
+const { normalizeScopeSet, scopeCompare } = require("./normalize");
+
+exports.simplifyScopeExpression = scopeExpression => {
+  let key = Object.keys(scopeExpression);
```
djmitche: What if `scopeExpression` is just a scope, like `"scope1"`?

libraries/scopes/src/simplify.js
```diff
@@ -0,0 +1,58 @@
...
+  if (Object.values(scopeExpression)[0].length === 1) {
+    if (key.includes("AnyOf")) {
+      return scopeExpression.AnyOf[0];
+    } else if (key.includes("AllOf")) {
+      return scopeExpression.AllOf[0];
+    }
+  }
```
djmitche: I think this section handles the case where AnyOf and AllOf have a single option.  But what if they have multiple options that *simplify* to a single option?  This is a place where fixed-point iteration would help -- perhaps a previous round converted `{AnyOf: ["scope1", "scope2", "scope*"]}` to `{AnyOf: ["scope*"]}`, and this round then simplifies that further to `"scope*"`.

libraries/scopes/src/simplify.js
ajinkabeer: Yes, I think my approach here is really flawed. It's better for me to rethink the whole function by taking in the points you said.

djmitche: Yep, "refactoring" is an important part of the TDD process :)

libraries/scopes/src/simplify.js
ajinkabeer: Yes. I understand now.

libraries/scopes/src/simplify.js
djmitche: "Flawed" sounds mean -- it passes the tests!  It's just a step along the way, part of the process :)

ajinkabeer: Oh, :). But I hope someday I will be able to write a good solution to this problem. I'll keep on working.

djmitche:
How's this going?  Note that I reorganized the files in this library a little bit, but that should be an easy rebase when you're done with this work.

ajinkabeer:
> How's this going? Note that I reorganized the files in this library a little bit, but that should be an easy rebase when you're done with this work.

I am re-writing the entire function using recursion. It passes all the base case test-cases including removing duplicates as of now. Now I need to simplify the nested cases. I think I am getting close. 
@djmitche

djmitche:
I ended up completing this in another PR.  Thanks again :D

PR #1938, opened 2019-11-09
Intersect TPL client scopes with *expanded* user scopes

djmitche:


services/web-server/src/servers/oauth2.js
```diff
@@ -52,6 +52,8 @@ module.exports = (cfg, AuthorizationCode, AccessToken, strategies, auth, monitor
     const accessToken = new Buffer.from(taskcluster.slugid()).toString('base64');
     const currentUser = await strategies[user.identityProviderId].userFromIdentity(user.identity);
 
+    const userScopes = await auth.expandScopes(currentUser.scopes());
```
helfi92: Looking at https://github.com/taskcluster/taskcluster/blob/5cf30481fe09b5dbc8b438b144972d545f3339cd/services/web-server/src/login/scanner.js#L53

input and return value seem to be an object.

PR #1941, opened 2019-11-10
Make Task Group status more visible on mobile

### LINKED ISSUE #1908: [UI] long enough title makes it hard to see the status on mobile
Screenshot_20191106-094249

As you can see, one has to constantly scroll to see status.

At first, it might be good if we could avoid wasting so much space on sides, especially on mobile. The "link" item also wastes space on the same direction.

kellim:
Github Bug/Issue: Fixes #1908

The only page I found with the issue is the page that displays a Task Group. Please let me know if I missed another page with the same issue.

I added breakpoints for small devices to remove extra padding on the right and left sides of the table being displayed and made it so the entire status is visible and the name takes up the rest of the space.

Result: 
make-status-more-visible-on-mobile

### DIFF
--- a/changelog/issue-1908.md
+++ b/changelog/issue-1908.md
@@ -1,0 +1,3 @@
+level: silent
+reference: issue 1908
+---
--- a/ui/src/components/TaskGroupTable/index.jsx
+++ b/ui/src/components/TaskGroupTable/index.jsx
@@ -94,23 +94,14 @@ const createSortedTasks = memoize(
     textDecoration: 'none',
     ...theme.mixins.hover,
     ...theme.mixins.listItemButton,
-    [theme.breakpoints.down('sm')]: {
-      paddingLeft: 0,
-      paddingRight: 0,
-    },
   },
   taskGroupName: {
     marginRight: theme.spacing(1),
-    maxWidth: '55vw',
     whiteSpace: 'nowrap',
     overflow: 'hidden',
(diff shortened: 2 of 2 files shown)


kellim:
@helfi92  There are some really long task names that start out the same and are only different at the end, so I'm not sure the users will be able to view enough of the task name without clicking on the link in order to distinguish between the tasks with similar names. Also, the tooltip with the full name that appears upon hovering doesn't work on mobile devices. 

I don't know if all Task Groups have tasks with such long and similar names, but the Task Group I've been looking at as an example is https://taskcluster-ui.herokuapp.com/tasks/groups/VDoqqceETXmxwfUMNdXotw

In this example, there are some tasks with similar names you can't tell apart. If reading the task name is important, it might be better to put the task name on one line with the status under it but I'm not sure if that can be done with flexbox alone and the user would have to scroll more to get to tasks at the bottom.

image

helfi92:
That's a good observation @kellim. That would be a different issue we can tackle in the future maybe by showing the beginning of the title then the end (e.g., `test-andr...proc-8`) Also, users on that view most of the time just want to make sure all tests pass and if there is one that fails for example, they would want to view the logs of it. Clicking on the task would reveal the task name as well as the log.

kellim:
@helfi92 I added your updates and I also made an additional update to tableSecondCell style so the status is vertically centered and it aligns horizontally to flex-end.

PR #1944, opened 2019-11-11
Dependencies not shown when a task doesn't have a decision task

helfi92:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1595418 and https://bugzilla.mozilla.org/show_bug.cgi?id=1595506.

### DIFF
--- a/services/web-server/src/loaders/tasks.js
+++ b/services/web-server/src/loaders/tasks.js
@@ -3,16 +3,15 @@ const sift = require('../utils/sift');
 const fetch = require('../utils/fetch');
 const ConnectionLoader = require('../ConnectionLoader');
 const Task = require('../entities/Task');
+const tryCatch = require('../utils/tryCatch');
 
 module.exports = ({ queue, index }) => {
   const task = new DataLoader(taskIds =>
-    Promise.all(
-      taskIds.map(async taskId => {
-        const task = await queue.task(taskId);
+    Promise.all(taskIds.map(async (taskId) => {
+      const [error, task] = await tryCatch(queue.task(taskId));
 
-        return new Task(taskId, null, task);
-      }),
-    ),
+      return error || new Task(taskId, null, task);
+    })),
   );
   const indexedTask = new DataLoader(indexPaths =>
     Promise.all(indexPaths.map(indexPath => index.findTask(indexPath))),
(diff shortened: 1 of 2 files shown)


services/web-server/src/loaders/tasks.js
```diff
@@ -3,16 +3,15 @@ const sift = require('../utils/sift');
...
-      taskIds.map(async taskId => {
-        const task = await queue.task(taskId);
+    Promise.all(taskIds.map(async (taskId) => {
+      const [error, task] = await tryCatch(queue.task(taskId));
 
-        return new Task(taskId, null, task);
-      }),
-    ),
+      return error || new Task(taskId, null, task);
+    })),
```
helfi92: We would need to do a similar logic to all loaders that use `Promise.all`. I could take care of that later, in a separate PR.

services/web-server/src/loaders/tasks.js
```diff
@@ -3,16 +3,15 @@ const sift = require('../utils/sift');
...
-    Promise.all(
-      taskIds.map(async taskId => {
-        const task = await queue.task(taskId);
+    Promise.all(taskIds.map(async (taskId) => {
+      const [error, task] = await tryCatch(queue.task(taskId));
 
-        return new Task(taskId, null, task);
-      }),
-    ),
+      return error || new Task(taskId, null, task);
```
djmitche: Will returning an Error object do the right thing, or would this need to be thrown?

helfi92: > Will returning an Error object do the right thing

Good question. The dataloader package seems to check for the instance of the object so we are good.

https://github.com/graphql/dataloader/blob/efc2b1a376863daeee88ff8e3bc901bb54fc5356/src/index.js#L293

helfi92:
> I don't really understand how this works. I think I don't understand when there would be multiple taskIds

https://www.apollographql.com/docs/graphql-tools/connectors/#dataloader-and-caching has a good explanation.

services/web-server/src/loaders/tasks.js
helfi92: Good question. The dataloader package seems to check for the instance of the object so we are good.

https://github.com/graphql/dataloader/blob/efc2b1a376863daeee88ff8e3bc901bb54fc5356/src/index.js#L293

PR #1956, opened 2019-11-12
Bug 1595667 - Make TP _first_ login experience smooth

helfi92:
Bugzilla Bug: [1595667](https://bugzilla.mozilla.org/show_bug.cgi?id=1595667).

### DIFF
--- a/changelog/bug-1595667.md
+++ b/changelog/bug-1595667.md
@@ -1,0 +1,4 @@
+level: patch
+reference: bug 1595667
+---
+Taskcluster third-party login  UI now instructs users to sign in to provide credentials to a third party registered client instead of showing them the home page.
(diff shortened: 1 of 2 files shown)


glandium:
I'm not sure this makes things better. IIRC that's what I did the very first time (clicked on signin), and in the end, I still had to go back to the treeherder tab manually.

helfi92:
> I'm not sure this makes things better. IIRC that's what I did the very first time (clicked on signin), and in the end, I still had to go back to the treeherder tab manually.

When the user successfully signs in to the Taskcluster site, we just close the window and allow the browser to decide which tab to show next. In general, this should bring the user back to the treeherder tab since it was the last one used but if the user manually switches tabs while logging in, then the outcome may be different. That's my current understanding.

sarah-clements:
I think if taskcluster didn't open the auth0/github login screen in a new window, this might solve the issue (if this is easily doable).

sarah-clements:
> I think if taskcluster didn't open the auth0/github login screen in a new window, this might solve the issue (if this is easily doable).

Edit: see my comment on the bugzilla bug.

PR #1967, opened 2019-11-13
Add LRU caching to get task method

owlishDeveloper:
Bugzilla Bug: [1595221](https://bugzilla.mozilla.org/show_bug.cgi?id=1595221)

Github Bug/Issue: Fixes #1595221


services/queue/src/main.js
```diff
@@ -431,6 +432,7 @@ let load = loader({
         blobRegion: ctx.cfg.app.blobArtifactRegion,
         publicBlobBucket: ctx.cfg.app.publicBlobArtifactBucket,
         privateBlobBucket: ctx.cfg.app.privateBlobArtifactBucket,
+        LRUcache: new QuickLRU({maxSize: 1000000}),
```
djmitche: Let's make this a configurable value, with a small default (10?).  At a conservative 3k apiece, a million Task objects would occupy 3GB, which is more than we budget for these k8s pods.

services/queue/config.yml
```diff
@@ -100,6 +100,8 @@ defaults:
     useCloudMirror:               !env:bool:optional USE_CLOUD_MIRROR
     cloudMirrorHost:              'cloud-mirror.taskcluster.net'
 
+    taskCacheMaxSize: 5000
```
djmitche: To make this changeable without releasing a new TC, it needs to be an !env:.. variable.  I think that can use `optional` too, with a suitable default value.

imbstack:
I wouldn't say I have any edge cases in mind in particular. It feels like the kind of thing that we should get 100% right is all I mean. I'll look at this today.

services/queue/src/api.js
```diff
@@ -142,19 +143,23 @@ builder.declare({
...
+  const {taskId} = req.params;
+
+  // Load Task entity if it's absent from cache
+  let task;
+  if (this.LRUcache.has(taskId)) {
+    task = this.LRUcache.get(taskId);
+  } else {
+    task = await this.Task.load({taskId}, true);
+    this.LRUcache.set(taskId, task);
+  }
```
imbstack: I think we need this sorta thing in other places where we load tasks such as `claimWork`

djmitche: That's not a bad idea, but let's start small with the caching and see what the impact on our load is.

imbstack: I don't think much of our load is coming from calls to `task` itself? I don't mind releasing this as-is though.

changelog/bug-1595221.md
```diff
@@ -0,0 +1,5 @@
+level: patch
+reference: bug 1595221
+---
+Adds an LRU cache to getTask method, so that we don't have to make too many calls to Azure (tasks are immutable anyways)
+The default value for the cache size is 10. The name of the optional prop in the config.yml is `taskCacheMaxSize`
```
imbstack: It should be `queue.task_cache_max_size` in this case.

imbstack:
Although I do wonder if the queue test failure is "real". I think your code is correct but the test might be asserting something that is broken by this change

owlishDeveloper:
> Although I do wonder if the queue test failure is "real". I think your code is correct but the test might be asserting something that is broken by this change

Which queue test failure?

PR #1970, opened 2019-11-14
Rename "Provisioners" to "Workers" in the sidebar

djmitche:
Provisioners aren't really a thing anymore..


djmitche:
You're right that I should make the names all align -- will do.

This is just a temporary measure until we can build a better UI for this.  So "provisioner ID" can stay around for now.

I'll rename "Workers" to "Worker Manager", too.  That will probably be permanent, with that interface the way to manage worker-manager, and the "Workers" view eventually replaced with a view that mixes data from both queue and worker-manager.

djmitche:
> I'll rename "Workers" to "Worker Manager", too. That will probably be permanent, with that interface the way to manage worker-manager, and the "Workers" view eventually replaced with a view that mixes data from both queue and worker-manager.

Ah, I misunderstood.  I think two "Workers" views, only one of which is in the sidebar, will be OK for a month or two until we develop the new UI.

PR #1971, opened 2019-11-14
Remove 'production' config from auth where it overrides env vars

djmitche:
This allows the possibility of configuring sentry parameters that are different from what we use here at Mozilla.  I'll follow up with similar changes to other services, but this was the important one due to that improvement.

Sentry configs were handled in https://bugzilla.mozilla.org/show_bug.cgi?id=1594575.  @edunham, can you verify my bold assertion in the first paragraph of the changelog is true for the deployments you control? :D


djmitche:
Oh, I meant to wait for @edunham's verification.  If that's *not* the case please let me know ASAP and we can back this out before releasing.

djmitche:
:+1: from edunham :)

PR #1978, opened 2019-11-14
Bug 1596531 - fix notify tests

djmitche:
Bugzilla Bug: [1596531](https://bugzilla.mozilla.org/show_bug.cgi?id=1596531)

This was definitely my fault, being a bit over-zealous cleaning up SQS queues a few days ago.

### DIFF
--- a/services/notify/test/helper.js
+++ b/services/notify/test/helper.js
@@ -141,11 +141,13 @@ exports.withSES = (mock, skipping) => {
           VisibilityTimeout: 30,
           WaitTimeSeconds: 20,
         }).promise();
-        const messages = resp.Messages;
-        await sqs.deleteMessage({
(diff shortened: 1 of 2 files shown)


services/notify/test/helper.js
```diff
@@ -131,6 +131,30 @@ exports.withSES = (mock, skipping) => {
...
+              },
+            },
+          ],
+        };
+        await sns.setQueueAttributes({
+          QueueUrl: emailSQSQueue,
+          Attributes: {
+            Policy,
+          },
+        }).promise();
```
imbstack: What does this do?

djmitche: Makes it work? :)

I'll add a comment..

PR #1982, opened 2019-11-15
Remove support for BaseURL from the Go client

djmitche:
This is a breaking change (hence `level: major`) but for most users shouldn't be a big deal.

When upgraded, generic-worker and taskcluster-proxy may require some changes, but those should be minor and probably improve clarity of those codebases.


djmitche:
RootURL serves the same purpose.

djmitche:
@petemoore where do we go from here?  Taskcluster no longer supports BaseURLs.  How can we get the testing support you need without BaseURLs?  What functionality is lacking in RootURLs?

petemoore:
My position hasn't really changed since https://github.com/taskcluster/generic-worker/pull/123#issuecomment-458464927 - but if you want to remove support, introduce a breaking change, and are prepared to rewrite the client/worker tests that take advantage of the feature, I won't stop you. ;-)

djmitche:
Per Slack discussion, the issue is that generic-worker has e.g., `queueBaseURL` config.  When generic-worker updates to a newer client, it can be refactored to use `queueRootURL` instead, and perhaps remove that from the documented configuration options so that non-test users of generic-worker aren't tempted to configure each rootURL differently.

djmitche:
(Those changes will occur when generic-worker re-vendors the go client, which is already going to bring in several other breaking changes -- it will not happen automatically on landing this change).

PR #1984, opened 2019-11-15
Remove all hard-coded production configuration

djmitche:
This fixes a few places where config parameters were overridden
with hard-coded values, and also removes the last place where we allowed
a configuration parameter to define a table name.

Please squint carefully at these changes!  Very little I changed here can be caught by tests, or even by deployment to a dev environment (where for example it would be easy to miss a table name changing).


imbstack:
This is looking reasonable so far but I am going to put on my careful squinting glasses and give it another pass soon

djmitche:
image

djmitche:
I want to wait a day or two until 23.0.0 is deployed to see if it warrants any patch releases.  Then I'll land this.

PR #1985, opened 2019-11-15
Bug 1577785 - Remove blob and azure artifact types

djmitche:
This also recognizes that the artifact-created Pulse event is only
created for reference and error artifacts.  This has been the case for a
long time, but was not documented.

Bugzilla Bug: [1577785](https://bugzilla.mozilla.org/show_bug.cgi?id=1577785)

:red_circle: :red_circle: :red_circle: :red_circle: :red_circle: :red_circle: :red_circle: :red_circle: :red_circle:


djmitche:
Hold the presses, I found some more stuff to delete

djmitche:
OK, roll the presses...

even more :red_circle: :red_circle: :red_circle: :red_circle: :red_circle: :red_circle: :red_circle: :red_circle:

PR #1976, opened 2019-11-14
[UI] Migrate to material-ui v4

helfi92:
Reverts taskcluster/taskcluster#1974.

PR #1992, opened 2019-11-15
[UI] fix datePickers styles, update material-ui/pickers

Rolikasi:
fix [#1976](https://github.com/taskcluster/taskcluster/pull/1976#issuecomment-554080447)


Rolikasi:
Sure! We need also fix this one:
STR:
1) turn on light theme
2) open calendar
3) hover selected day
4) see invisible numbers

calendar error

helfi92:
Good catch! That'd be awesome :)

Rolikasi:
Done! :)

changelog/issue-895-1.md
```diff
@@ -0,0 +1,3 @@
+level: silent
+reference: issue 895
+---
```
helfi92: Let's remove this changelog even if CI test fail. We'll merge it to the material-ui pr which already has a changelog.

ui/src/theme.js
```diff
@@ -369,6 +386,18 @@ const createTheme = isDarkTheme => {
...
+              backgroundColor: isDarkTheme
+                ? THEME.TEN_PERCENT_WHITE
+                : THEME.TEN_PERCENT_BLACK,
+            },
+          },
+        },
+      },
       MuiDrawer: {
         paper: {
           maxWidth: 500,
```
helfi92: Hovering on a date removes the background color of the date picker. Hovering on a datepicker in https://material-ui-pickers.dev/demo/datepicker does not remove the background color. Is there a way to fix this?

ui/src/theme.js
```diff
@@ -369,6 +386,18 @@ const createTheme = isDarkTheme => {
           },
         },
       },
+      MuiDialogActions: {
+        root: {
+          '& button': {
+            color: isDarkTheme ? 'white' : '#000',
```
helfi92: Instead of using `white` and `#000`, I wonder if it would make more sense to use `THEME.PRIMARY_TEXT_DARK` and `THEME.PRIMARY_TEXT_LIGHT`.

ui/src/theme.js
Rolikasi: Yes, there is)

PR #1994, opened 2019-11-15
Bug 1579496 - actually skip azure tests

djmitche:
Bugzilla Bug: [1579496](https://bugzilla.mozilla.org/show_bug.cgi?id=1579496)


djmitche:
The downside here is, it uses mock AWS for all of the "real" tests.  This is pretty gross.

PR #1996, opened 2019-11-15
Rely on MemoryStore in local development

helfi92:
Resolves https://github.com/taskcluster/taskcluster/issues/1995.


changelog/issue-1995.md
```diff
@@ -0,0 +1,4 @@
+level: patch
```
djmitche: This could probably be silent, since it's a devel-only change.

PR #2001, opened 2019-11-16
[UI]Adding code linking in docs section

vishakhanihore:
Bugzilla Bug: [1591236](https://bugzilla.mozilla.org/show_bug.cgi?id=1591236)

### DIFF
--- a/changelog/bug-1591236.md
+++ b/changelog/bug-1591236.md
@@ -1,0 +1,4 @@
+level: minor
+reference: bug 1591236
+---
+Linking Docs pages to their source codes.
--- a/ui/docs/manual/task-execution/provisioning.mdx
+++ b/ui/docs/manual/task-execution/provisioning.mdx
@@ -4,7 +4,15 @@ title: Provisioning
 order: 40
 ---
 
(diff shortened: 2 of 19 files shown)


vishakhanihore:
@djmitche @helfi92 Please give it a look so that I can use the same in complete docs sections.

changelog/bug-1591236.md
```diff
@@ -0,0 +1,4 @@
+level: minor
+reference: bug 1591236
+---
+Linking Docs pages to their source codes.
```
djmitche: The [best practices](https://github.com/taskcluster/taskcluster/blob/master/dev-docs/best-practices/changelog.md) for our changelogs suggest:
 * `silent`: Changes to tests or documentation
 * Write it in the present tense, as if the change had just been deployed.

Since it's silent, there's no need for text here, but the second bullet point would mean something like "Documentation pages now contain links to their source code".

vishakhanihore: @djmitche You are right, giving the docs link in the status bar would make more sense. But wont this result in two nav-bar components in `ui/src` .

djmitche: I don't think it needs another navbar component -- just different placement.

Issue #2005, opened 2019-11-16
Time duration are given with too little precision


SimonSapin:
**Describe the bug**

Time durations in the UI are always given as an integer number of the largest possible unit, which is often not enough precision.

**To Reproduce**

1. Go to https://community-tc.services.mozilla.com/tasks/IpzWeA6ZSgairHSwGk3s3Q
2. Look under "Resolved"
3. See that this task was resolved "about 1 hour later" (after it was started)
4. Ponder what that could mean. (The answer is [anything from 44 mins 30 secs to 89 mins 30 secs](https://date-fns.org/v2.7.0/docs/formatDistance). So a twofold variation could be made invisible by this approximation.)
5. Decide this is not enough information to answer the question that brought you here.
6. Hover the mouse cursor over "Resolved" to see that the timestamp is `2019-11-15T23:04:14.369Z`, and over "Started" to see `2019-11-15T22:16:19.037Z`
7. Start doing mental subtraction
8. If only we had some kind of machine that is good at doing arithmetic

**Expected behavior**

Time durations should be given with at least two, preferably three significant digits. (Two digits means almost 10% error when the first digit is `1`, which is not great.)

Subjectively, I think the `formatDistance` function in the `date-fns` library is designed for human events where the order of magnitude of how long ago something happened is more important than the actual date or time. It doesn’t seem appropriate for timing computing processes that are sometimes our jobs to optimize.

PR #2015, opened 2019-11-18
[UI] Date distance precision

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/2005.

### DIFF
--- a/changelog/issue-2005.md
+++ b/changelog/issue-2005.md
@@ -1,0 +1,4 @@
+level: patch
+reference: issue 2005
+---
+Taskcluster UI now adds more accuracy when displaying the distance between given dates in words.
--- a/ui/src/components/DateDistance/index.jsx
+++ b/ui/src/components/DateDistance/index.jsx
@@ -1,10 +1,5 @@
 import { Component } from 'react';
-import {
-  isDate,
-  formatDistance,
-  formatDistanceStrict,
-  parseISO,
-} from 'date-fns';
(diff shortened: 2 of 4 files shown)


SimonSapin:
I don’t know what parts of the UI use what part of the code, but https://community-tc.services.mozilla.com/tasks/IpzWeA6ZSgairHSwGk3s3Q uses "about" which looks like a call to `formatDistance` rather than `formatDistanceStrict`. Is this PR modifying the right call?

The cut-off at 120 minutes means that we could still have imprecise "about 2 hours" or "about 3 hours" etc. But I’m ok with declaring that any single task that takes that long is a bug anyway :)

helfi92:
@SimonSapin I added the change to the offset time as well.

Screen Shot 2019-11-18 at 11 20 44 AM

SimonSapin:
Looks good, thanks!

djmitche:
It looks like this test [can fail](https://community-tc.services.mozilla.com/tasks/Wtr2HHYwQ9yp_9Fwdl6yiA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FWtr2HHYwQ9yp_9Fwdl6yiA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log).

helfi92:
I pushed a hotfix yesterday to fix the test. Rebasing on master should do it.

PR #2016, opened 2019-11-18
Bug 1596098 - return 400 when entities are too big

djmitche:
Bugzilla Bug: [1596098](https://bugzilla.mozilla.org/show_bug.cgi?id=1596098)

(draft so that I can see if the bare azure-entities upgrade passes CI)


djmitche:
I think this is uncommon enough that 400 is OK.  That's what we'd do for too many dependencies, for example.

PR #2017, opened 2019-11-18
Bug 1597331 - provide a workerPoolId tag/label on instances

djmitche:
Bugzilla Bug: [1597331](https://bugzilla.mozilla.org/show_bug.cgi?id=XXXXX)

### DIFF
--- a/changelog/bug-1597331.md
+++ b/changelog/bug-1597331.md
@@ -1,0 +1,4 @@
+level: patch
+reference: bug 1597331
+---
+Instances created by the AWS provider now have an explicit `WorkerPoolId` tag.  The Google provider now supplies `created-by` and `owner` tags.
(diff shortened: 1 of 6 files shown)


changelog/bug-1597331.md
```diff
@@ -0,0 +1,4 @@
+level: patch
+reference: bug 1597331
+---
+Instances created by the AWS and Google providers are now tagged/labeled with the workerPoolId.
```
owlishDeveloper: they were tagged with workerPoolId _before_ tho.... what I really see in the change, for aws provider you make it _absolutely clear_ what is workerPoolId, and for google provider you add `owner` and `created by` tags, as well as changing the regex rules for converting those 🤔

#preferential

djmitche: Good point - I'll redraft the changelog and land.

PR #2020, opened 2019-11-18
Fix cli client, so that you can request multiple scopes.

tomprince:


changelog/tc-cli-scopes.md
```diff
@@ -0,0 +1,3 @@
+type: patch
```
djmitche:
```suggestion
level: patch
```

PR #2026, opened 2019-11-19
Wrap async callbacks functions to call done(..)

djmitche:
I like what this does to oauth2.js -- a simple wrapper that allows use of async functions without any trouble.

The unpromisify.js is pretty gnarly, though.  There appears to be no better way to set a function's arity, and oauth2orize will call functions completely differently based on their arity, so we need to match the original arity including the trailing `done` method.

Like elsewhere, there's one spot in here where I made a function `async` even though it doesn't use `await`, because that allows any thrown exceptions to be treated as Promise rejections.

### DIFF
--- a/services/web-server/src/utils/unpromisify.js
+++ b/services/web-server/src/utils/unpromisify.js
@@ -1,7 +1,8 @@
 /**
  * Wrap an async `fn` in such a way that it can act as a function taking a callback
  * as its last argument.  That callback either gets cb(err) for errors, or cb(null, res)
- * or (if returnsArray is true) cb(null, ...res)
+ * or (if returnsArray is true) cb(null, ...res).  The resuting function has the
+ * appropriate arity (one more than the input arity, to account for the 'done' argument)
  */
 module.exports = (fn, {returnsArray} = {}) => {
   // eslint-disable-next-line no-unused-vars
--- a/services/web-server/src/utils/unpromisify.js
+++ b/services/web-server/src/utils/unpromisify.js
@@ -1,0 +1,19 @@
+/**
+ * Wrap an async `fn` in such a way that it can act as a function taking a callback
+ * as its last argument.  That callback either gets cb(err) for errors, or cb(null, res)
+ * or (if returnsArray is true) cb(null, ...res)
+ */
+module.exports = (fn, {returnsArray} = {}) => {
+  // eslint-disable-next-line no-unused-vars
+  const call = (self, args) => {
+    const done = args.pop();
+    Promise.resolve(fn.apply(self, args)).then(
+      returnsArray ? res => done.call(null, null, ...res) : res => done.call(null, null, res),
+      err => done.call(null, err));
+  };
+
+  // preserve arity by generating code..
+  const args = [...new Array(fn.length + 1)].map((_, i) => `arg${i}`);
+  // eslint-disable-next-line no-eval
+  return eval(`const x = function (${args}) { return call(this, [...arguments]); }; x`);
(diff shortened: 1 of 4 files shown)


djmitche:
(the Taskcluster-UI failure is [unrelated](https://github.com/taskcluster/taskcluster/pull/2015#issuecomment-555509287))

services/web-server/src/utils/unpromisify.js
```diff
@@ -0,0 +1,19 @@
...
+    Promise.resolve(fn.apply(self, args)).then(
+      returnsArray ? res => done.call(null, null, ...res) : res => done.call(null, null, res),
+      err => done.call(null, err));
+  };
+
+  // preserve arity by generating code..
+  const args = [...new Array(fn.length + 1)].map((_, i) => `arg${i}`);
+  // eslint-disable-next-line no-eval
+  return eval(`const x = function (${args}) { return call(this, [...arguments]); }; x`);
+};
```
helfi92: > The unpromisify.js is pretty gnarly

Agreed. Any chance we could add tests?

services/web-server/src/utils/unpromisify.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Wrap an async `fn` in such a way that it can act as a function taking a callback
+ * as its last argument.  That callback either gets cb(err) for errors, or cb(null, res)
+ * or (if returnsArray is true) cb(null, ...res)
+ */
+module.exports = (fn, {returnsArray} = {}) => {
```
helfi92: Love the ingenuity :)

services/web-server/src/utils/unpromisify.js
```diff
@@ -0,0 +1,19 @@
...
+    const done = args.pop();
+    Promise.resolve(fn.apply(self, args)).then(
+      returnsArray ? res => done.call(null, null, ...res) : res => done.call(null, null, res),
+      err => done.call(null, err));
+  };
+
+  // preserve arity by generating code..
+  const args = [...new Array(fn.length + 1)].map((_, i) => `arg${i}`);
+  // eslint-disable-next-line no-eval
+  return eval(`const x = function (${args}) { return call(this, [...arguments]); }; x`);
```
helfi92: It looks like the only reason why we are using `eval` is so that we can spread the list of arguments. I think we should be able to avoid this by doing something like:

```js
return (args) => {
  return call(this, [...args]);
}(args);
```

Thoughts?

services/web-server/src/utils/unpromisify.js
djmitche: Good idea..

services/web-server/src/utils/unpromisify.js
djmitche: No, see the comment above about arity.  The code you've written has arity 1.

helfi92: Sorry, I meant to write:

```js
return (...args) => {
  return call(this, args);
}(args);
```

helfi92: Sorry, I meant to write:

```js
return (...args) => {
  return call(this, [...args]);
}(args);
```

djmitche: That has arity 0 :)

helfi92: You're right. My mistake.

djmitche: I wrestled with this stuff for a while writing this patch!

PR #2028, opened 2019-11-19
Add Python client helper for easier usage.

La0:
This is a relatively thin wrapper around the taskcluster library to provide some easier access to secrets and services in different authentication contexts.
We (Mozilla relman/salltt teal) use this class across different projects:
- in [mozilla/code-review](https://github.com/mozilla/code-review/blob/master/tools/code_review_tools/taskcluster.py)
- in [mozilla/code-coverage](https://github.com/mozilla/code-coverage/blob/master/tools/code_coverage_tools/taskcluster.py)
- in [mozilla/libmozevent](https://github.com/mozilla/libmozevent/blob/master/libmozevent/taskcluster.py)

The idea is to instanciante only one `TaskclusterConfig` in the user's project, authenticate once (usually at the beginning of runtime), then access all the Taskcluster services through that wrapper so they all get the same configuration.

It supports Taskcluster authentifications in this order:
* directly provided credentials
* credentials from local configuration (XDG folder : `~/.config/taskcluster/config.json`)
* credentials from environment variables
* `taskclusterProxy`
* no authentication

We use it all over the place, in our bots (Python 3 cli apps) running in taskcluster, in web backend or other services running in Heroku, but also on our own computers while developing.

Here you can see [a sample usage in code review bot](https://github.com/mozilla/code-review/blob/master/bot/code_review_bot/cli.py#L52)

An even simpler usage could be:

```python
from taskcluster.helper import TaskclusterConfig

t = TaskclusterConfig()
t.auth() # will use env or XDG profile

q = t.get_service('queue')
print(q) # would give <taskcluster.generated.queue.Queue object at 0x7fc199174358>

s = t.load_secrets(
    'project/XXX/testdev',
    existing={
        "somekey": "a",
    },
    required=["topsecretkey"],
)
print(s) # would give {"someKey": "a", "topsecretkey": ...}
```

I'm totally open to change any naming you find 

No changelog snippet nor documentation, as i'm not sure you'll approve this addition....

### DIFF
--- a/changelog/pr-2028.md
+++ b/changelog/pr-2028.md
@@ -1,0 +1,4 @@
+level: minor
+reference: issue 2028
+---
+Add a Python client helper to ease integration into customers projects.
--- a/changelog/pr-2028.md
+++ b/changelog/pr-2028.md
@@ -1,4 +1,4 @@
 level: minor
 reference: issue 2028
 ---
-Add a Python client helper to ease integration into customers projects.
+The Taskcluster Python client now has helper classes to ease integration into customers' projects.
--- a/clients/client-py/README.md
+++ b/clients/client-py/README.md
@@ -359,6 +359,9 @@ await hooks.ping()
 Supported environment variables are:
 - `TASKCLUSTER_ROOT_URL` to specify your Taskcluster instance base url. You can either use that variable or instanciate `TaskclusterConfig` with the base url.
 - `TASKCLUSTER_CLIENT_ID` & `TASKCLUSTER_ACCESS_TOKEN` to specify your client credentials instead of providing them to `TaskclusterConfig.auth`
+- `TASKCLUSTER_PROXY_URL` to specify the proxy address used to reach Taskcluster in a task. It defaults to `http://taskcluster` when not specified.
+
+For more details on Taskcluster environment variables, [here is the documentation](https://docs.taskcluster.net/docs/manual/design/env-vars).
 
 ### Loading secrets across multiple authentications
 
--- a/clients/client-py/README.md
+++ b/clients/client-py/README.md
@@ -320,7 +320,7 @@ Generally a project using this library will face different use cases and authent
 
 ### Shared authentication
 
-The class `taskcluster.helper.TaskclusterConfig` is made to be instancied once in your project, usually in a top level module. That singleton is then accessed by different parts of your projects, whenever a Taskcluster service is needed.
+The class `taskcluster.helper.TaskclusterConfig` is made to be instantiated once in your project, usually in a top level module. That singleton is then accessed by different parts of your projects, whenever a Taskcluster service is needed.
 
 Here is a sample usage:
 
--- a/clients/client-py/README.md
+++ b/clients/client-py/README.md
@@ -308,6 +308,155 @@ taskcluster.fromNow("1 year", dateObj=dateObject1);
 # datetime.datetime(2018, 1, 21, 17, 59, 0, 328934)
 ```
 
+## Library helpers to easily integrate Taskcluster in your project
+
+The Python Taskcluster client has a module `taskcluster.helper` with utilities which allows you to easily share authentication options across multiple services in your project.
+
+Generally a project using this library will face different use cases and authentication options:
+
+* No authentication for a new contributor without Taskcluster access,
+* Specific client credentials through environment variables on a developer's computer,
+* Taskcluster Proxy when running inside a task.
+
+### Shared authentication
+
+The class `taskcluster.helper.TaskclusterConfig` is made to be instancied once in your project, usually in a top level module. That singleton is then accessed by different parts of your projects, whenever a Taskcluster service is needed.
+
+Here is a sample usage:
+
+1. in `project/__init__.py`, no call to Taskcluster is made at that point:
+
+```python
+from taskcluster.helper import Taskcluster config
+
+tc = TaskclusterConfig('https://community-tc.services.mozilla.com')
+```
+
+2. in `project/boot.py`, we authenticate on Taskcuster with provided credentials, or environment variables, or taskcluster proxy (in that order):
+
+```python
+from project import tc
+
+tc.auth(client_id='XXX', access_token='YYY')
+```
+
+3. at that point, you can load any service using the authenticated wrapper from anywhere in your code:
+
+```python
+from project import tc
+
+# Synchronous service class
+queue = tc.get_service('queue')
+
+# Asynchronous service class
+hooks = tc.get_service('hooks', use_async=True)
+
+# You can then use these classes as documented below:
+queue.ping()
+await hooks.ping()
+```
+
+Supported environment variables are:
+- `TASKCLUSTER_ROOT_URL` to specify your Taskcluster instance base url. You can either use that variable or instanciate `TaskclusterConfig` with the base url.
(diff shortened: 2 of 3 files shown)


clients/client-py/taskcluster/helper.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Local configuration used to access Taskcluster service and objects
+    """
+
+    def __init__(self, url="https://community-tc.services.mozilla.com"):
+        self.options = None
+        self.secrets = None
+        self.default_url = os.environ.get("TASKCLUSTER_ROOT_URL", url)
+
+    def auth(self, client_id=None, access_token=None, max_retries=12):
```
djmitche: Could you rearrange things so that `get_service` and `load_secrets` automatically call this?

clients/client-py/taskcluster/helper.py
```diff
@@ -0,0 +1,128 @@
...
+         * credentials from local configuration
+         * credentials from environment variables
+         * taskclusterProxy
+         * no authentication
+        """
+        self.options = {"maxRetries": max_retries}
+
+        if client_id is None and access_token is None:
+            # Credentials preference: Use local config from release-services
+            xdg = os.path.expanduser(os.environ.get("XDG_CONFIG_HOME", "~/.config"))
```
djmitche: I'm wary of this -- putting creds in cleartext in ~ suggests that they are long-lived credentials (more than a few days) and that they are fairly powerful (not tailored to one specific operation).  That's kind of dangerous, especially for the Firefox-CI deployment, and not a behavior we'd like to encourage -- that's largely why we removed support for such config files from the clients.  Would you be comfortable removing it here, too?

clients/client-py/taskcluster/helper.py
```diff
@@ -0,0 +1,128 @@
...
+            logger.info("Taskcluster Proxy enabled")
+            self.options["rootUrl"] = "http://taskcluster"
+
+        else:
+            logger.info("No Taskcluster authentication.")
+            self.options["rootUrl"] = self.default_url
+
+    def get_service(self, service_name):
+        """
+        Build a Taskcluster service instance using current authentication
```
djmitche: This specifically returns synchronous services -- could you add support for async as well?

clients/client-py/taskcluster/helper.py
```diff
@@ -0,0 +1,128 @@
...
+
+        service = getattr(_client_importer, service_name.capitalize(), None)
+        assert service is not None, "Invalid Taskcluster service {}".format(
+            service_name
+        )
+        return service(self.options)
+
+    def load_secrets(
+        self, secret_name, prefixes=[], required=[], existing={}, local_secrets=None
+    ):
```
djmitche: I think this is a cool utility -- but perhaps it should be exposed as a function, instead?  `taskcluster.helpers.load_secrets`?  Then it could be used independently of the TaskclusterConfig class.

clients/client-py/taskcluster/helper.py
```diff
@@ -0,0 +1,128 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class TaskclusterConfig(object):
+    """
+    Local configuration used to access Taskcluster service and objects
+    """
+
+    def __init__(self, url="https://community-tc.services.mozilla.com"):
```
djmitche: We have avoided having any default rootUrls.  Callers should be explicit about the deployment they are using.

clients/client-py/taskcluster/helper.py
La0: Works for me, we are rarely using this feature anymore and if needs arise, we can still implement it in the user application

clients/client-py/taskcluster/helper.py
La0: Makes sense, i'll remove that.

clients/client-py/taskcluster/helper.py
La0: I'm moving it in its own function, but adding a shortcut on the `TaskclusterConfig` class too so we can keep our behavior while still using the function independently.

La0:
@djmitche If you're ok with the current structure, i'll add some doc in `clients/client-py/README.md`

djmitche:
Looks good at a skim :)

clients/client-py/taskcluster/helper.py
```diff
@@ -0,0 +1,130 @@
...
+            self.options["credentials"] = {
+                "clientId": client_id,
+                "accessToken": access_token,
+            }
+            self.options["rootUrl"] = self.default_url
+
+        elif "TASK_ID" in os.environ:
+            # Use Taskcluster Proxy when running in a task
+            logger.info("Taskcluster Proxy enabled")
+            self.options["rootUrl"] = "http://taskcluster"
```
SimonSapin: Should this be `os.environ.get("TASKCLUSTER_PROXY_URL")`? https://community-tc.services.mozilla.com/docs/manual/design/env-vars

La0: Good catch ! I'll update the PR

La0:
[Added some documentation](https://github.com/taskcluster/taskcluster/blob/66357ee319f4e7a9720d13fc32a4864da8daab4a/clients/client-py/README.md#library-helpers-to-easily-integrate-taskcluster-in-your-project)

clients/client-py/README.md
```diff
@@ -308,6 +308,155 @@ taskcluster.fromNow("1 year", dateObj=dateObject1);
...
+hooks = tc.get_service('hooks', use_async=True)
+
+# You can then use these classes as documented below:
+queue.ping()
+await hooks.ping()
+```
+
+Supported environment variables are:
+- `TASKCLUSTER_ROOT_URL` to specify your Taskcluster instance base url. You can either use that variable or instanciate `TaskclusterConfig` with the base url.
+- `TASKCLUSTER_CLIENT_ID` & `TASKCLUSTER_ACCESS_TOKEN` to specify your client credentials instead of providing them to `TaskclusterConfig.auth`
```
djmitche: Let's mention TASKCLUSTER_PROXY_URL, too, and link to https://docs.taskcluster.net/docs/manual/design/env-vars

clients/client-py/README.md
```diff
@@ -308,6 +308,155 @@ taskcluster.fromNow("1 year", dateObj=dateObject1);
...
+
+Generally a project using this library will face different use cases and authentication options:
+
+* No authentication for a new contributor without Taskcluster access,
+* Specific client credentials through environment variables on a developer's computer,
+* Taskcluster Proxy when running inside a task.
+
+### Shared authentication
+
+The class `taskcluster.helper.TaskclusterConfig` is made to be instancied once in your project, usually in a top level module. That singleton is then accessed by different parts of your projects, whenever a Taskcluster service is needed.
```
djmitche:
```suggestion
The class `taskcluster.helper.TaskclusterConfig` is made to be instantiated once in your project, usually in a top level module. That singleton is then accessed by different parts of your projects, whenever a Taskcluster service is needed.
```

changelog/pr-2028.md
```diff
@@ -0,0 +1,4 @@
+level: minor
+reference: issue 2028
+---
+Add a Python client helper to ease integration into customers projects.
```
djmitche:
```suggestion
The Taskcluster Python client now has helper classes to ease integration into customers' projects.
```

clients/client-py/README.md
La0: Done :+1:

PR #2034, opened 2019-11-19
Bug 1588096 - only support scope expressions, not array notation

djmitche:
This replaces a few uses of the old `[[scope]]` notation for scope
requirements with scope expressions, and removes support for that
notation.

Bugzilla Bug: [1588096](https://bugzilla.mozilla.org/show_bug.cgi?id=1588096)


djmitche:
Please have a careful look at the changes to auth and web-server.  Those changes continue to pass tests, but please verify the logic.

libraries/scopes/src/satisfaction.js
```diff
@@ -1,13 +1,4 @@
...
-const {validateScopeSets} = require('./validate');
-
-const validateScopePatterns = (scopePatterns) => {
-  assert(scopePatterns instanceof Array && scopePatterns.every((scope) => {
-    return typeof scope === 'string';
-  }), 'scopes must be an array of strings');
-};
-
-const patternMatch = exports.patternMatch = (pattern, scope) => {
+exports.patternMatch = (pattern, scope) => {
```
imbstack: should we just rename this file at this point if it only has `patternMatch` in it?

services/auth/src/signaturevalidator.js
```diff
@@ -88,7 +88,7 @@ const limitClientWithExt = function(credentialName, issuingClientId, accessToken
     }
 
     // Validate certificate scopes are subset of client
-    if (!utils.scopeMatch(scopes, [cert.scopes])) {
+    if (!utils.satisfiesExpression(scopes, {AllOf: cert.scopes})) {
```
imbstack: just a side note that I really like how much clearer the new syntax is to read

imbstack:
Also needs a changelog :)

libraries/scopes/README.md
```diff
@@ -2,7 +2,7 @@
 
 Simple utilities to validate scopes, scope-sets, and scope-expression satisfiability.
 
-For information on scopes, see [the Taskcluster documentation](../../ui/docs/manual/design/apis/hawk/scopes.mdx).
+For information on scopes, see [the Taskcluster documentation](https://docs.taskcluster.net/docs/manual/design/apis/hawk/scopes).
```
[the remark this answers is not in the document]
djmitche: I figure link to the highest-quality stable URL.  This README is read on GitHub, so we can't use a link to the "current deployment", but `docs.taskcluster.net` is stable.  The old link directly to the .mdx file was not high-quality :)

services/auth/src/signaturevalidator.js
djmitche: Definitely, it was a good improvement..

libraries/scopes/src/satisfaction.js
djmitche: Yeah, honestly most of the code could fit in one file at this point..

services/auth/src/api.js
```diff
@@ -844,7 +844,7 @@ builder.declare({
 
       // Check scopes
       const formerRoleScopes = this.resolver.resolve(role.scopes);
-      const scopesAdded = input.scopes.filter(s => !scopeUtils.scopeMatch(formerRoleScopes, [[s]]));
+      const scopesAdded = input.scopes.filter(s => !scopeUtils.satisfiesExpression(formerRoleScopes, s));
```
owlishDeveloper: just to make sure: so we are now sure to expect the new syntax..... is it an array? Not an object like `{AllOf: []}` or `{AnyOf: []}` etc?

djmitche: In this case, `s` is a single scope (a string).

owlishDeveloper: sorry, I meant `input.scopes`

owlishDeveloper: Basically, I just wanted to double check that we don't loop with `Array.prototype.filter` over an object. If this worry is unfounded, r+

djmitche: Good question -- yeah, that's enforced with JSON-schema.

PR #2041, opened 2019-11-20
Put endpoints into stable 🐴

owlishDeveloper:
Bugzilla Bug: [1585157](https://bugzilla.mozilla.org/show_bug.cgi?id=1585157)

I didn't touch the Queue's endpoints because all the experimental ones reference `provisionerId` and `workerType` which are ye olde concepts so my understanding is, we will be actually deprecating those. (Please confirm)


owlishDeveloper:
> Looks good. I think the artifact-listing endpoints in the queue are stable -- we haven't changed those in a long time. But you're right about all of the declare* etc.

Which declare* do you mean? So, I understand your answer we mark queue's endpoints as stable as well?

djmitche:
Let's just add listArtifacts and listLatestArtifacts as stable, and merge.  Leave the other experimental queue endpoints as-is.

owlishDeveloper:
> Let's just add listArtifacts and listLatestArtifacts as stable, and merge. Leave the other experimental queue endpoints as-is.

oh ok

owlishDeveloper:
I also noticed that I missed github. I marked almost everything there as `stable` except for `badge` (it may get changed to work with statuses), `createStatus` (it's part of the current checks implementation which also might get changed) and `gerRepositoryStatus` - this one is not working for some reason currently

PR #2046, opened 2019-11-20
Rely on an env var for canonical_host

helfi92:
From https://github.com/heroku/heroku-buildpack-static#canonical-host, it looks like we can use an environment variable to specify the redirect. I suggest the following:

taskcluster-ui app -> `HOST=community-tc.services.mozilla.com`
taskcluster-net app -> `HOST=docs.taskcluster.net`

Thoughts?

### DIFF
--- a/changelog/OlF1yHM_SoKK81FH4ZiD_g.md
+++ b/changelog/OlF1yHM_SoKK81FH4ZiD_g.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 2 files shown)


djmitche:
It does seem like a big app to deploy to just issue redirects!  And, yes, a 302 seems safer.

https://github.com/kenmickles/heroku-redirect

helfi92:
> https://github.com/kenmickles/heroku-redirect

That's a very good idea. I forked the repo to our Taskcluster org and added it to the taskcluster-ui heroku app.

https://taskcluster-ui.herokuapp.com/ should properly redirect now.

PR #2047, opened 2019-11-20
Bug 1577785 - Remaining cleanup from no-longer-supported artifact types

djmitche:


services/queue/config.yml
```diff
@@ -5,18 +5,10 @@ defaults:
...
-    privateBlobArtifactBucket:    !env PRIVATE_BLOB_ARTIFACT_BUCKET
-    blobArtifactRegion:           !env BLOB_ARTIFACT_REGION
-
-    # S3 buckets where artifacts are stored NOTE: Only legacy S3 type
+    # S3 buckets where artifacts are stored
     publicArtifactBucket:         !env PUBLIC_ARTIFACT_BUCKET
     privateArtifactBucket:        !env PRIVATE_ARTIFACT_BUCKET
 
-    # Azure blob container for artifacts NOTE: Only for legacy Azure Type
-    artifactContainer:            artifacts
```
owlishDeveloper: do we also need to remove lines 144 and 172?

djmitche: Can you link to the lines you mean?  I don't see related config at those lines in either the "before" or "after" versions of this file.

owlishDeveloper: [one](https://github.com/taskcluster/taskcluster/pull/2047/files/417e17564be0535772d47d0c70ea205a957ba734#diff-aef05eaa19eb6dbca13a21a68301bda7R164)

[two](https://github.com/taskcluster/taskcluster/pull/2047/files/417e17564be0535772d47d0c70ea205a957ba734#diff-aef05eaa19eb6dbca13a21a68301bda7R197)

djmitche: Thanks, I don't know how I didn't see those!

PR #2048, opened 2019-11-21
Bug 1518190 - Add `yarn backup`

djmitche:
This implements a simple `yarn backup:run` command based on [taskcluster-backup](https://github.com/taskcluster/taskcluster-backup).  One of the cool bits is that it automatically backs up only the tables and containers used by services.

I added a good bit of machinery to get that information in a stable form (azure.yml) that we can't forget to update when we add a new table (thanks to the meta test).  As a side-effect, this cleans up the static scopes nicely.

I initially imported the taskcluster-backup git history, but the code was rewritten enough that git couldn't find the similarities, so I just dropped the history.

NOTE: `yarn backup:restore` and `yarn backup:verify` still need to be written!  Wander, is that something you'd like to work on?


dev-docs/best-practices/microservices.md
```diff
@@ -144,3 +144,14 @@ type Hook {
...
+
+## Scopes
+
+Any scopes that the service needs in its credentials should be listed in `service/<name>/scopes.yml`.
+The exception is scopes for Azure table access.
+
+## Azure Tables
+
+Any azure tables or containers that a service uses in production should be described in `service/<name>/azure.yml`, under keys `containers` and `tables`.
+The `yarn generate` logic will ensure that appropriate scopes are added to the service's static clients.
```
imbstack: I wonder if we should even take the next step and automatically inject azure table creation into lib-loader based on this :bulb:

services/auth/src/static-scopes.json
```diff
@@ -2,44 +2,40 @@
...
       "assume:repo:github.com/*",
-      "queue:scheduler-id:taskcluster-github",
-      "queue:route:statuses",
-      "queue:route:checks",
       "assume:scheduler-id:taskcluster-github/*",
-      "auth:azure-table-access:${azureAccountId}/TaskclusterGithubBuilds",
-      "auth:azure-table-access:${azureAccountId}/TaskclusterIntegrationOwners",
+      "auth:azure-table:read-write:${azureAccountId}/TaskclusterCheckRuns",
+      "auth:azure-table:read-write:${azureAccountId}/TaskclusterChecksToTasks",
       "auth:azure-table:read-write:${azureAccountId}/TaskclusterGithubBuilds",
```
imbstack: how was this even working before?

ui/docs/manual/deploying/backups.mdx
```diff
@@ -0,0 +1,69 @@
...
+Of course, backups should nonetheless be kept in a secure, non-public bucket.
+
+### Backup Lifecycle
+
+We recommend using the tools provided by AWS S3 to manage the lifecycle of your backups.
+Use versioning to keep multiple versions of each table, and use a lifecycle rule to delete older versions.
+
+## Runing Restores
+
+TBD
```
imbstack: lol

services/auth/src/static-scopes.json
djmitche: the former scope implies the latter, it's just the legacy form.

dev-docs/best-practices/microservices.md
djmitche: Given the desire to move to postgres, I didn't want to over-complicate things here.

djmitche:
TODO:
 * [ ] implement inclusion / exclusion (we can leave which tables to include/exclude up to deployers)
 * [ ] implement restore and verification

infrastructure/tooling/src/main.js
```diff
@@ -158,6 +158,27 @@ program.command('backup:run')
     run(backup, options[0]);
   });
 
+program.command('backup:restore <resource> [destination]')
+  .option('--version-id <VersionId>', 'VersionId of the S3 object to restore (default latest)')
```
imbstack: neat addition. we didn't support this in the old version, right?

djmitche: Right :)

PR #2052, opened 2019-11-21
Add categories

djmitche:
image

This also updates the stability of auth.sentryDSN since we are now supporting that endpoint again :)


changelog/WGyqIUL3Se6bP0_yWYR6Jw.md
```diff
@@ -0,0 +1,2 @@
+level: silent
```
owlishDeveloper: why silent if the difference is in the UI?

clients/client-py/taskcluster/generated/aio/auth.py
```diff
@@ -447,7 +447,7 @@ class Auth(AsyncBaseClient):
         initial team configured for this component. Contact a Sentry admin
         to have the project transferred to a team you have access to if needed
 
-        This method is ``deprecated``
+        This method is ``stable``
```
owlishDeveloper: why do we resurrect it?

changelog/WGyqIUL3Se6bP0_yWYR6Jw.md
djmitche: From https://github.com/taskcluster/taskcluster/blob/master/dev-docs/best-practices/changelog.md#levels
> * silent: Changes to tests or documentation
> * silent: Minor changes or fixes to UI

clients/client-py/taskcluster/generated/aio/auth.py
djmitche: At one point we were going to stop supporting sentry within the platform, when we were using stackdriver for error reporting.  But we reversed course and now report errors directly to sentry.  And, workers use this method to get sentry credentials to report their own errors.  So basically we are no longer planning to remove this method.

changelog/WGyqIUL3Se6bP0_yWYR6Jw.md
owlishDeveloper: I wonder if I'll ever get used to that scale.... somehow everything is 2 tones down in there comparing to how I'd personally classify changes 😄

PR #2056, opened 2019-11-21
Move meta tests into infrastructure/tooling

djmitche:
..and use console-taskgraph for them.  This isn't much more than cleanup.


owlishDeveloper:
can haz no red crossez?

djmitche:
That's a lot of them!  But I'll look.

infrastructure/tooling/src/meta/checks/depcheck.js
```diff
@@ -0,0 +1,83 @@
...
+const util = require('util');
+const depcheck = require('depcheck');
+const exec = util.promisify(require('child_process').exec);
+const _ = require('lodash');
+const {REPO_ROOT} = require('../../utils');
+
+/*
+ * The 'depcheck' tool is async but still blocks for long stretches, perhaps
+ * doing computation.  So, we defer that to a worker thread.
+ */
```
imbstack: I love that our general view on web endpoint timings is `¯\_(ツ)_/¯` but when it comes to our tooling we are going out of our way to make it performant.

infrastructure/tooling/src/meta/checks/speling.js
```diff
@@ -0,0 +1,32 @@
+const util = require('util');
```
imbstack: hahaha, `speling.js`

infrastructure/tooling/src/meta/checks/depcheck.js
djmitche: It's worse than that! I was annoyed that the console-taskgraph spinny things weren't spinning!

Also, this was a chance to play with worker threads in a non-scary context.

PR #2064, opened 2019-11-24
[UI] Show task actions while viewing live log

### LINKED ISSUE #1960: [UI] Show task actions while viewing live log
Would be nice to be able to trigger task actions (e.g. re-run) while viewing the live log, without having to go back to the task viewer first.

Rolikasi:
I made a new component with buttons in `views/Tasks` because it's used only in these views.(or do we need to move it to `src/components`?)
It seems like it works but it's a bit messy
I don't understand why do we need `actionData` and how it works.
I tried to add `className` to new component in Log view, but it is not working, so I changed styles of leftArrow icon

Closes #1960


ui/src/views/Tasks/TaskLog/index.jsx
```diff
@@ -10,13 +10,15 @@ import Link from '../../../utils/Link';
...
+import TaskButtons from '../TaskButtons';
 
+@withApollo
 @hot(module)
 @withStyles(theme => ({
   fab: {
     ...theme.mixins.fab,
     ...theme.mixins.actionButton,
-    bottom: theme.spacing(3),
+    bottom: theme.spacing(2),
```
helfi92: What's the reasoning behind this change?

ui/src/views/Tasks/TaskLog/index.jsx
```diff
@@ -75,6 +77,7 @@ export default class TaskLog extends Component {
             </Link>
           }
         />
+        <TaskButtons data={this.props.data} url={url} rawButton />
```
helfi92: All the props that are defined here should be defined in the component itself in `propTypes` and perhaps `defaultProps` if any of those are not always required.

ui/src/views/Tasks/TaskButtons/index.jsx
```diff
@@ -0,0 +1,751 @@
...
+import formatTaskMutation from '../../../utils/formatTaskMutation';
+import scheduleTaskQuery from './scheduleTask.graphql';
+import rerunTaskQuery from './rerunTask.graphql';
+import cancelTaskQuery from './cancelTask.graphql';
+import purgeWorkerCacheQuery from './purgeWorkerCache.graphql';
+import createTaskQuery from '../createTask.graphql';
+import submitTaskAction from '../submitTaskAction';
+
+@withRouter
+@hot(module)
```
helfi92: `@hot` is not required.

ui/src/views/Tasks/TaskButtons/index.jsx
```diff
@@ -0,0 +1,751 @@
...
+@withApollo
+export default class TaskActionButtons extends Component {
```
helfi92: Maybe we can rename this component to `TaskActionsSpeedDial` and move it to be in `src/components`.

ui/src/views/Tasks/TaskLog/task.graphql
```diff
@@ -1,13 +1,71 @@
...
+    status {
+      state
+    }
+    metadata {
+      name
+    }
   }
-}
\ No newline at end of file
+}
```
helfi92: Could you help me understand why we are now requiring all of these fields now?

helfi92:
@ededals Any plans on finishing this? :-)

helfi92:
Note to self:

* [ ] Create two separate task.graphql and only require what's needed. For example, dependent tasks, artifacts are not require when viewing a task log.
* [ ] Test that the implementation works

helfi92:
Closing for now. Will re-open PR in the future.

PR #2082, opened 2019-11-25
Bug 1597777: additional_allowed_cors_origin is json

edunham:
because https://github.com/taskcluster/taskcluster/blob/master/infrastructure/k8s/templates/taskcluster-web-server-secret.yaml#L17 said so.

and thus there are double backslashes in the actual origin regex in the community cluster, which need to cease being double backslashes by the time that actually gets used as a regex.

Bugzilla Bug: [1597777](https://bugzilla.mozilla.org/show_bug.cgi?id=1597777)

### DIFF
--- a/changelog/bug-1597777.md
+++ b/changelog/bug-1597777.md
@@ -1,0 +1,4 @@
+level: silent 
+reference: bug 1597777
+---
+web server service reading additional_allowed_cors_origin from json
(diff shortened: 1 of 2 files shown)


djmitche:
I think this will result in needing to use quadruple-backslashes, to survive both Helm's de-backslashing and JSON de-backslashing?  Can you give more information on what the error is?

djmitche:
To be clear, the following are equivalent

* `ADDITITIONAL_CORS_ALLOWED_ORIGIN='abc'` using `!env`
* `ADDITITIONAL_CORS_ALLOWED_ORIGIN='"abc"' using `!env:json`

in other words, the change you've made requires *more* escaping of the input value, not less.

djmitche:
Ah, I see the comment in the bug now.  I'll look harder tomorrow morning.

djmitche:
I think #2090 addresses this..

PR #2083, opened 2019-11-25
Add a test for cross-pool creds

imbstack:
This was actually covered before by tests but this one makes it a bit more explicit. This scared the heck of of me for a few minutes.

### DIFF
--- a/services/worker-manager/test/api_test.js
+++ b/services/worker-manager/test/api_test.js
@@ -870,7 +870,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
       providerData: {},
     };
 
-    test('no such workerPool)', async function() {
+    test('no such workerPool', async function() {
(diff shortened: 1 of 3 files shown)


services/worker-manager/test/api_test.js
```diff
@@ -909,6 +909,26 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
...
     });
 
+    test('worker requests across pools', async function() {
+      await helper.WorkerPool.create({
+        ...defaultWorkerPool,
+      });
+      await helper.WorkerPool.create({
+        ...defaultWorkerPool,
+        workerPoolId: 'ff/tt',
+      });
```
owlishDeveloper: why do we have both of them?

owlishDeveloper: (if the default has id of `ff/ee` then it's clear to me tho)

PR #2085, opened 2019-11-25
[bug 1599291] Monitoring for worker pools improvements

imbstack:
Bugzilla Bug: [1599291](https://bugzilla.mozilla.org/show_bug.cgi?id=1599291)

### DIFF
--- a/services/worker-manager/src/estimator.js
+++ b/services/worker-manager/src/estimator.js
@@ -21,7 +21,7 @@ class Estimator {
     // how many extra we want if we do want any
     const requestedCapacity = Math.max(0, desiredCapacity - runningCapacity);
 
-    this.monitor.log.simpleEstimate({
+    const estimatorInfo = {
       workerPoolId,
       pendingTasks,
       minCapacity,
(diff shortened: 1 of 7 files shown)


services/worker-manager/src/estimator.js
```diff
@@ -21,15 +21,26 @@ class Estimator {
...
       minCapacity,
       maxCapacity,
       runningCapacity,
       desiredCapacity,
       requestedCapacity,
-    });
+    }
+
+    let overProvisioned = false;
+    if (runningCapacity > (maxCapacity * 1.25)) {
```
owlishDeveloper: I understand this is arbitrary, but maybe we should add a comment on why and how this number

services/worker-manager/src/estimator.js
```diff
@@ -21,15 +21,26 @@ class Estimator {
...
+      overProvisioned = true;
+    }
+
+    this.monitor.log.simpleEstimate(estimatorInfo, {level: overProvisioned ? 'err' : 'notice'});
```
owlishDeveloper: 👍

PR #2086, opened 2019-11-26
[bug 1599299] Fix google provider tags

imbstack:
Bugzilla Bug: [1599299](https://bugzilla.mozilla.org/show_bug.cgi?id=1599299)

### DIFF
--- a/changelog/bug-1599299.md
+++ b/changelog/bug-1599299.md
@@ -1,0 +1,3 @@
+level: silent
+reference: bug 1599299
+---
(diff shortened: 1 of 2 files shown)


imbstack:
Let's not release a version of taskcluster without this btw. Just a heads up in case we were thinking of releasing tomorrow before I start

imbstack:
It looks like the changelog in https://github.com/taskcluster/taskcluster/commit/545dcbc9ec228be2aa5458385cbb4e56da38fd94 is still correct. I could just make this one a silent changelog.

djmitche:
perfect :)

PR #2087, opened 2019-11-26
[web-server] Use structured logging

helfi92:


helfi92:
If I want to log `monitor.debug` messages temporarily in production, is this something I should be reaching out to the cloudops people?

PR #2095, opened 2019-11-26
Bug 1599122 - spawn a more precise number of instances

djmitche:
Bugzilla Bug: [1599122](https://bugzilla.mozilla.org/show_bug.cgi?id=1599122)

Note that while this addresses an overprovisioning, it is a *limited* overprovisioning -- at most one instance per launchConfig, and there were 10 launchConfigs in the worker pool that led to bug 1599122.  So this cannot (alone) account for 100's of additional instances.

### DIFF
--- a/changelog/bug-1599122.md
+++ b/changelog/bug-1599122.md
@@ -1,0 +1,4 @@
+level: patch
+reference: bug 1599122
+---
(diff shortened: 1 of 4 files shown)


monwork.js
```diff
@@ -0,0 +1,30 @@
+const taskcluster = require('taskcluster-client');
```
imbstack: Whoops :p

djmitche: Haha, `git add -a` is my nemesis.  At least I didn't have creds in that file :)

PR #2097, opened 2019-11-27
[Bug 1587511] (part 1) Terminate workers that don't register 

imbstack:

This is just part one because the bug is for terminating workers if they go idle for a long time and don't try to claim or reclaim work. That will be a bigger project requiring queue<-->w-m interaction.

Sticking this key inside a `lifecycle` field allows us to add more settings like this in the future but if it feels unwieldy, let me know and I will move it up. Also happy to change names, etc. As an example, we could add the `runningTimeout` to account for the 96 hour thing in this bug.

TODO:

- [x] Test this on a dev environment
- [x] Figure out how to do worker tests for aws provider and add lifecycle tests for it. @owlishDeveloper I think you were working on adding these tests before? Is it in a state where you could either share the code with me or make a PR into this one with the tests?

### DIFF
--- a/changelog/bug-1587511.md
+++ b/changelog/bug-1587511.md
@@ -1,0 +1,9 @@
+level: minor
+reference: bug 1587511
+---
+WorkerPools can now be configured to terminate workers that fail to register after some amount of time.
+Both of the google and aws providers now support a `lifecycle` object that for now has a single key
+of `registrationTimeout`. It is optional and if it is provided the value is an integer with the number
+of seconds a worker has to register before it is terminated.
+
+This helps catch misconfigured or broken workers before they become zombies or worse.
--- a/services/worker-manager/test/fixtures/schemas/sample-aws-config.json
+++ b/services/worker-manager/test/fixtures/schemas/sample-aws-config.json
@@ -1,30 +1,30 @@
 {
-    "launchConfigs": [
-      {
-        "launchConfig": {
-          "ImageId": "ami-blabla",
-          "KeyName": "somekey",
-          "SecurityGroupIds": [
-            "sg-yadayada"
-          ],
-          "InstanceType": "m5.xlarge"
-        },
-        "region": "us-west-2",
-        "capacityPerInstance": 1
+  "launchConfigs": [
+    {
+      "launchConfig": {
+        "ImageId": "ami-blabla",
+        "KeyName": "somekey",
+        "SecurityGroupIds": [
+          "sg-yadayada"
+        ],
+        "InstanceType": "m5.xlarge"
       },
-      {
-        "launchConfig": {
-          "ImageId": "ami-yadayada",
-          "KeyName": "anotherkey",
-          "SecurityGroupIds": [
-            "sg-blahblahblah"
-          ],
-          "InstanceType": "m5.xlarge"
-        },
-        "region": "us-east-1",
-        "capacityPerInstance": 1
-      }
-    ],
-    "minCapacity": 1,
-    "maxCapacity": 1
-  }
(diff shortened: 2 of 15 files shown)


changelog/bug-1587511.md
```diff
@@ -0,0 +1,9 @@
+level: minor
+reference: bug 1587511
+---
+WorkerPools can now be configured to terminate workers that fail to register after some amount of time.
+Both of the google and aws providers now support a `lifecycle` object that for now has a single key
+of `registrationTimeout`. It is optional and if it is provided the value is an integer with the number
+of seconds a worker has to register before it is terminated.
+
+This helps catch misconfigured or broken workers before they become zombies or worse.
```
djmitche: What's worse than a zombie?  Werewolf?

services/worker-manager/test/provider_aws_test.js
```diff
@@ -144,6 +148,9 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
         assert.strictEqual(w.workerGroup, providerId, 'Worker group id should be the same as provider id');
         assert.strictEqual(w.state, helper.Worker.states.REQUESTED, 'Worker should be marked as requested');
         assert.strictEqual(w.providerData.region, defaultLaunchConfig.region, 'Region should come from the chosen config');
+        // Check that this is setting times correctly to within a second or so to allow for some time
+        // for the provisioning loop
```
djmitche: more seconds of fuzz might be good to avoid flakiness

services/worker-manager/test/provider_aws_test.js
```diff
@@ -280,6 +287,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
   });
 
   suite('AWS provider - checkWorker', function() {
+    // TODO: Can someone fill these out and then I'll add tests for the lifecycle stuff here as well?
```
djmitche: plz to link to https://bugzilla.mozilla.org/show_bug.cgi?id=1577839

imbstack: I’d like this TODO to be fixed before we land this. It is ok if this PR gets landed after the 24 release I guess

changelog/bug-1587511.md
imbstack: 🧛‍♂️

owlishDeveloper:
I didn't have time to write any tests, @imbstack ! There are only empty shells for some of them in the code.

changelog/bug-1587511.md
owlishDeveloper: 👾

services/worker-manager/test/provider_aws_test.js
imbstack: @owlishDeveloper can we write these tests now so that we know we're not breaking the provider with this change? Would it be better for you or me to write these tests?

changelog/bug-1587511.md
imbstack: :japanese_ogre:

services/worker-manager/test/fixtures/schemas/sample-aws-config.json
```diff
@@ -1,12 +1,4 @@
 {
-  "workerPoolId": "delete/worker",
```
owlishDeveloper: I trust it the indentation is ok after the deletion? It's hard to tell from the github ui

#nonblocking #preferential

imbstack: It seems to be ok!

PR #2098, opened 2019-11-27
Bug 1597922  - Add CSP headers

helfi92:
https://bugzilla.mozilla.org/show_bug.cgi?id=1597922


helfi92:
> Should we add the cache-control headers, too (maybe in a followup)?

Yep.

Issue #2099, opened 2019-11-27
[UI] Add cache-control CSP headers


helfi92:
Let's add similar cache-control headers that are found in https://github.com/taskcluster/taskcluster/blob/679ea9b792b2535fcb204609008d117ffd1e9c52/static.json#L24-L61 to https://github.com/taskcluster/taskcluster/blob/master/ui/web-ui-nginx-site.conf.

helfi92:
Done in https://github.com/taskcluster/taskcluster/pull/2098.

PR #2100, opened 2019-11-27
[UI] Add cache-control CSP headers

aryamanpuri:
Adding up of  cache-control headers in _https://github.com/taskcluster/taskcluster/blob/master/ui/web-ui-nginx-site.conf._
#2099


helfi92:
Oops, forgot to close the issue but it was fixed in https://github.com/taskcluster/taskcluster/pull/2098.

PR #2104, opened 2019-11-28
Bug 1598649 - automate the release process

djmitche:

This will be the first step: automating all parts of the release but still running it locally.  The next step will be to split this up and run most of it in CI, so that `yarn release` tags and pushes, with the rest of the fun happening remotely.

* [x] publish to pypi
* [x] remove debugging commented-out stuff


djmitche:
Note that one thing I missed here is credentials for `docker push`.  I'll get that on the next PR.

PR #2106, opened 2019-11-29
[UI] Add DateTimePicker in UI; fix  clock styles

### LINKED ISSUE #2031: [UI] Expiry times only include date.
Various place in the UI only display dates, where the underlying value also includes times.

The place I've most noticed this is the client page; I usually use CLI clients with expiry <1h, so the lack of being able to see the current expiry time, and adjust it is impactful. I've also noticed that secrets only show the dates There may be other place in the UI that display dates as well.

Manish-Giri:
- Replaced `KeyboardDatePicker` with `KeyboardDateTimePicker` to additionally allow choosing time in pickers across the UI
- Removed `format` prop from three components that render `DatePicker` - `SecretForm`, `ClientForm` and `AuthConsent`, so the new `format` in `DatePicker` works correctly
- Introduced new styles in `theme.js` so that the hour and clock hands display prominently in the pickers, across both dark and light themes

For choosing prominent colors for the clock hands, I started by looking at how the colors were displayed on the [material-ui-pickers demo page](https://material-ui-pickers.dev/demo/datetime-picker). It looks like their clock hand colors match the color of the main toolbar, so I decided to go the same route for the picker here, and chose the color of the main toolbar at `/secrets/create` - `#4177a5`, which is defined as `THEME.SECONDARY` - https://github.com/taskcluster/taskcluster/blob/5e6c2b635b916f495873e9b64fd07a69fb090946/ui/src/utils/constants.js#L52

I used the same color for both dark and light themes, and they stood out quite well, but if required, I can change the color based on theme.

---
### After - 

dark-hour

dark-min

---

### After -

light-hour-noop

light-hour-noppoint

---

I also changed the background-color of `MuiPickersClockPointer.noPoint`, because otherwise, the white circle in the blue looked a bit jarring -

light-hour
 

---

Github Bug/Issue: Fixes #2031

### DIFF
--- a/changelog/issue-2031.md
+++ b/changelog/issue-2031.md
@@ -1,4 +1,4 @@
 level: minor
 reference: issue 2031
 ---
-Taskcluster UI pickers now feature a time component, so for instance, users can now pick a particular time (hour:minute) on a particular day to set as the expiration time.
+Taskcluster UI revamped the date picker component to allow selecting the hour and the minute in addition to the date.
--- a/changelog/issue-2031.md
+++ b/changelog/issue-2031.md
@@ -1,0 +1,4 @@
+level: minor
+reference: issue 2031
+---
+Taskcluster UI pickers now feature a time component, so for instance, users can now pick a particular time (hour:minute) on a particular day to set as the expiration time.
--- a/ui/src/theme.js
+++ b/ui/src/theme.js
@@ -297,7 +297,7 @@ const createTheme = isDarkTheme => {
       },
       MuiPickersToolbarText: {
         toolbarTxt: {
-          color: 'rgba(255, 255, 255, 0.54)',
+          color: THEME.PRIMARY_TEXT_DIMMED,
         },
         toolbarBtnSelected: {
           color: THEME.PRIMARY_TEXT_DARK,
@@ -345,7 +345,7 @@ const createTheme = isDarkTheme => {
         },
         thumb: {
           borderColor: THEME.SECONDARY,
-          backgroundColor: 'rgba(254,255, 255, 0.9)',
+          backgroundColor: THEME.PRIMARY_TEXT_DARK,
         },
         noPoint: {
           backgroundColor: THEME.SECONDARY,
@@ -360,14 +360,9 @@ const createTheme = isDarkTheme => {
         tabs: {
           color: '#fff',
           backgroundColor: THEME.SECONDARY,
-        },
-      },
-      PrivateTabIndicator: {
-        colorPrimary: {
-          backgroundColor: '#fff',
-        },
-        colorSecondary: {
-          backgroundColor: '#000',
+          '& .MuiTabs-indicator': {
+            backgroundColor: isDarkTheme ? '#fff' : '#000',
+          },
         },
       },
       MuiListItem: {
--- a/ui/src/theme.js
+++ b/ui/src/theme.js
@@ -297,7 +297,7 @@ const createTheme = isDarkTheme => {
       },
       MuiPickersToolbarText: {
         toolbarTxt: {
-          color: THEME.PRIMARY_TEXT_DIMMED,
+          color: 'rgba(255, 255, 255, 0.54)',
         },
         toolbarBtnSelected: {
           color: THEME.PRIMARY_TEXT_DARK,
--- a/ui/src/theme.js
+++ b/ui/src/theme.js
@@ -334,6 +334,22 @@ const createTheme = isDarkTheme => {
           },
         },
       },
+      MuiPickersClock: {
+        pin: {
+          backgroundColor: THEME.SECONDARY,
+        }
+      },
+      MuiPickersClockPointer: {
+        pointer: {
+          backgroundColor: THEME.SECONDARY,
+        },
+        thumb: {
+          borderColor: THEME.SECONDARY,
+        },
+        noPoint: {
+          backgroundColor: THEME.SECONDARY,
+        },
+      },
       MuiListItem: {
         root: {
           userSelect: 'text',
--- a/ui/src/theme.js
+++ b/ui/src/theme.js
@@ -358,7 +358,16 @@ const createTheme = isDarkTheme => {
       },
       MuiPickerDTTabs: {
         tabs: {
+          color: '#fff',
           backgroundColor: THEME.SECONDARY,
+        },
+      },
+      PrivateTabIndicator: {
+        colorPrimary: {
+          backgroundColor: '#fff',
+        },
+        colorSecondary: {
+          backgroundColor: '#000',
         },
       },
       MuiListItem: {
--- a/ui/src/theme.js
+++ b/ui/src/theme.js
@@ -297,7 +297,7 @@ const createTheme = isDarkTheme => {
       },
       MuiPickersToolbarText: {
         toolbarTxt: {
-          color: THEME.PRIMARY_TEXT_DARK,
+          color: 'rgba(255, 255, 255, 0.54)',
         },
         toolbarBtnSelected: {
           color: THEME.PRIMARY_TEXT_DARK,
@@ -345,8 +345,19 @@ const createTheme = isDarkTheme => {
         },
         thumb: {
           borderColor: THEME.SECONDARY,
+          backgroundColor: 'rgba(254,255, 255, 0.9)',
         },
         noPoint: {
+          backgroundColor: THEME.SECONDARY,
+        },
+      },
+      MuiPickersClockNumber: {
+        clockNumberSelected: {
+          color: '#fff',
+        },
+      },
+      MuiPickerDTTabs: {
+        tabs: {
           backgroundColor: THEME.SECONDARY,
         },
       },
--- a/ui/src/theme.js
+++ b/ui/src/theme.js
@@ -337,7 +337,7 @@ const createTheme = isDarkTheme => {
       MuiPickersClock: {
         pin: {
           backgroundColor: THEME.SECONDARY,
-        }
+        },
       },
       MuiPickersClockPointer: {
         pointer: {
--- a/ui/src/components/AuthConsent/index.jsx
+++ b/ui/src/components/AuthConsent/index.jsx
@@ -121,7 +121,6 @@ export default class AuthConsent extends Component {
                         name="expires"
                         value={expires}
                         onChange={onExpirationChange}
-                        format="yyyy/MM/dd"
                         maxDate={addYears(new Date(), 1001)}
                       />
                     }
--- a/ui/src/components/ClientForm/index.jsx
+++ b/ui/src/components/ClientForm/index.jsx
@@ -293,7 +293,6 @@ export default class ClientForm extends Component {
                 <DatePicker
                   value={expires}
                   onChange={this.handleExpirationChange}
-                  format="yyyy/MM/dd"
                   maxDate={addYears(new Date(), 1001)}
                 />
               }
--- a/ui/src/components/DatePicker/index.jsx
+++ b/ui/src/components/DatePicker/index.jsx
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import { func } from 'prop-types';
 import classNames from 'classnames';
 import {
-  KeyboardDatePicker,
+  KeyboardDateTimePicker,
   MuiPickersUtilsProvider,
 } from '@material-ui/pickers';
 import { withStyles } from '@material-ui/core/styles';
@@ -39,7 +39,7 @@ export default class DatePicker extends Component {
 
     return (
       <MuiPickersUtilsProvider utils={DateFnsUtils}>
-        <KeyboardDatePicker
+        <KeyboardDateTimePicker
           className={classNames(classes.datePicker, className)}
           showTodayButton
           keyboardIcon={<CalendarIcon />}
@@ -47,6 +47,7 @@ export default class DatePicker extends Component {
           leftArrowIcon={<ChevronLeftIcon />}
           value={value}
           onChange={onChange}
+          format="yyyy/MM/dd hh:mm a"
           {...props}
         />
       </MuiPickersUtilsProvider>
--- a/ui/src/components/SecretForm/index.jsx
+++ b/ui/src/components/SecretForm/index.jsx
@@ -204,7 +204,6 @@ export default class SecretForm extends Component {
               label="Expires"
               value={expires}
               onChange={this.handleExpirationChange}
-              format="yyyy/MM/dd"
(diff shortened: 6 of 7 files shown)


Manish-Giri:
Looks like I need to add a changelog? Not sure why the other 6 tasks failed.

Manish-Giri:
@helfi92 I have fixed your suggestions 2, 3 and 4, but I wasn't able to find this dialog section that has the calendar and time icon in it -

69886902-56534d80-12b2-11ea-9726-76aadc029cb4

I have tried fiddling with various combinations of the picker at `/secrets/create` as well as the demo page at https://material-ui-pickers.dev/demo/datetime-picker, but the dialog box consists of only the date and time in textual form in all cases -

ww

I looked at the [KeyboardDateTimePicker API](https://material-ui-pickers.dev/api/KeyboardDateTimePicker) and the only thing that looked like it could enable the dialogbox is the `dateRangeIcon` prop (of type `ReactNode`), but it is currently not used anywhere in `DatePicker.jsx`. Unless I missed a keyboard shortcut that enables the dialog.

If you could tell me the correct way to enable the dialog box, I can then implement your suggestion 1. Thank You.

helfi92:
I'm not sure I understand the question correctly but to change the background color of the dialog that has the calendar and clock icon, we could do something like:

```diff
diff --git a/ui/src/theme.js b/ui/src/theme.js
index cc7559923..98040be50 100644
--- a/ui/src/theme.js
+++ b/ui/src/theme.js
@@ -291,6 +291,11 @@ const createTheme = isDarkTheme => {
           whiteSpace: 'nowrap',
         },
       },
+      MuiPickerDTTabs: {
+        tabs: {
+          backgroundColor: 'red',
+        },
+      },
       MuiPickersToolbar: {
         toolbar: {
           backgroundColor: THEME.SECONDARY,
```

Screen Shot 2019-12-09 at 3 29 20 PM

---

For the background color as well as the color of the tab indicator, we should have something that looks like:

Screen Shot 2019-12-09 at 3 19 54 PM
Screen Shot 2019-12-09 at 3 20 47 PM

Hope that helps.

Manish-Giri:
@helfi92 Sorry if I wasn't clear enough. What I meant with this - 

> If you could tell me the correct way to enable the dialog box, I can then implement your suggestion 1

was - when I was working on my windows machine on this, I didn't find the section of the dialog box that contains the date and time icons. 

If you notice the screenshots below -

1. *In this screenshot of `/secrets/create`, the dialog box above the actual clock, consists of only the date and time in text*. The section of the dialog that is supposed to consist of the date and time icon (to which you added `background-color: red` in your comment just now) is missing. 
 
Screenshot 2019-12-07 18 00 11 2

2. Same thing with the original [datetime-picker component demo](https://material-ui-pickers.dev/demo/datetime-picker) -

Screenshot 2019-12-07 20 56 04 2

But when I open the same demo on my mac machine, the dialog with the date and time icon is present -

Screenshot_2019-12-09_15_54_58

I'm not sure what happened there, but I'll go back and investigate tonight. Thank You for your help.

helfi92:
wow. That must be an issue with `@material-ui/pickers`. It would be good if you could file an issue in https://github.com/mui-org/material-ui-pickers/issues.

Manish-Giri:
@helfi92 I updated the PR. The latest commit should take care of all your suggestions. 

---

However, that 1 test in `taskcluster-ui` is still failing. Per the task logs, the error message is -
```js
 isDateWithin
    ✔ should return true
    ✖ should return false

...
FAILED TESTS:
  isDateWithin
    ✖ should return false
      Firefox 68.0.0 (Linux 0.0.0)
    expected true to be false
    AssertionError@test/isDateWithin_test.js:169:13
    ./node_modules/chai/lib/chai/assertion.js/module.exports/Assertion.prototype.assert@test/isDateWithin_test.js:478:13
    ./node_modules/chai/lib/chai/core/assertions.js/module.exports/<@test/isDateWithin_test.js:1313:10
    propertyGetter@test/isDateWithin_test.js:8219:29
    proxyGetter@test/isDateWithin_test.js:9729:22
    ./test/isDateWithin_test.js/</</<@test/isDateWithin_test.js:29187:58
    ./test/isDateWithin_test.js/</<@test/isDateWithin_test.js:29186:85
```

On doing some digging, the second test in `isDateWithin_test.js` is failing -
https://github.com/taskcluster/taskcluster/blob/20d15d3ead48f411753037ce8758cf05e7911021/ui/test/isDateWithin_test.js#L19-L29

The `isDateWithin()` function uses `fromNow()` in [fromNow.js](https://github.com/taskcluster/taskcluster/blob/master/ui/src/utils/fromNow.js), which in turn uses [parseTime.js](https://github.com/taskcluster/taskcluster/blob/master/ui/src/utils/parseTime.js). 

Could the error be due to the fact that we replaced DatePicker with DateTimePicker which now has an additional time component? But I'm not sure where these components are being used for the test, (behind the scenes), if that's the case. I didn't find anything in the files I checked.

helfi92:
> Could the error be due to the fact that we replaced DatePicker with DateTimePicker which now has an additional time component?

Nope. That's a known issue https://bugzilla.mozilla.org/show_bug.cgi?id=1598643

Manish-Giri:
All right, I'll start working on these two issues. I'll switch to my mac machine because the dialogbox with the calendar and time icons aren't visible on my windows machine. If I could have tested them, I would have perhaps caught the two issues.

> Aside, did you have time to create an issue upstream to raise the issue you were having on windows?

Yes, I created issue https://github.com/mui-org/material-ui-pickers/issues/1419 last night https://github.com/taskcluster/taskcluster/pull/2106#ref-issue-535508079, but they already closed that issue claiming that if the device window height is small, the tabs won't appear, otherwise the picker would need a scrollbar.

My windows machine is a 15.6" laptop, and the browser is already maximized, not sure how I can increase the height further. I also fiddled in dev tools (as they recommended), but no difference.

helfi92:
Interesting. Thanks for filing the issue.

helfi92:
You'll probably want to rebase your work on master so that it doesn't include commits outside this PR.

Manish-Giri:
@helfi92 I fixed the last two remaining issues. 

I'm not sure what's going with the additional commits that just got pulled in. I had to change machines to get the calendar and time icons to show up, but I think I forgot to do a `git pull --rebase`, which is why this happened. I'll try it now.

ui/src/theme.js
```diff
@@ -298,7 +298,7 @@ const createTheme = isDarkTheme => {
       },
       MuiPickersToolbarText: {
         toolbarTxt: {
-          color: THEME.PRIMARY_TEXT_DARK,
+          color: 'rgba(255, 255, 255, 0.54)',
```
helfi92: This would affect everywhere we use a [material-ui tab](https://material-ui.com/components/tabs/#tabs). We'll want to make this change for the pickers component only.

ui/src/theme.js
```diff
@@ -335,6 +335,42 @@ const createTheme = isDarkTheme => {
...
+          backgroundColor: THEME.SECONDARY,
+        },
+      },
+      MuiPickersClockPointer: {
+        pointer: {
+          backgroundColor: THEME.SECONDARY,
+        },
+        thumb: {
+          borderColor: THEME.SECONDARY,
+          backgroundColor: 'rgba(254,255, 255, 0.9)',
```
helfi92: Was there a reason to hard-code this value?

ui/src/theme.js
Manish-Giri: Sorry, possible noob question - I did some digging, but I couldn't find the class `.MuiPickersToolbarText-toolbarTxt` being used by `Tab`/`Tabs` (from [material-ui tab](https://material-ui.com/components/tabs/#tabs)) anywhere.  I looked at the classes used in these tabs, and created this [gist](https://gist.github.com/Manish-Giri/19e0f03351521753b155e9e049edff7a) which has all the classes used by `.MuiTab` and `.MuiTabs`.

Were you referring to the color change in the `PrivateTabIndicator` class here?
```css
  PrivateTabIndicator: {
        colorPrimary: {
          backgroundColor: '#fff',
        },
        colorSecondary: {
          backgroundColor: '#000',
        },
      },
```

Because, I found out that while this changes the color of the tab-indicator underneath the calendar-time dialog, as you had shown in your [comment](https://github.com/taskcluster/taskcluster/pull/2106#issuecomment-563424157), this also changes the color of tab-indicator in other places, like in `http://localhost:5080/auth/scopes`, as shown below -

Screenshot 2019-12-20 21 38 55

Is this what you meant when you said we should be making the change for the Picker component only?

ui/src/theme.js
Manish-Giri: Um, the `254` in the red-component is actually a typo. It should be `255`. I'll address this in my next commit. 

As for the opacity values (`0.9` and `0.54`), I found these values from the material-ui pickers.dev component demos. They use `0.9` for picker text which is currently selected, while everything else is dimmed out using `0.54`.

helfi92: Ah, ok. We can use `PRIMARY_TEXT_DARK`:

https://github.com/taskcluster/taskcluster/blob/0afbbc46a1cba2d6905c76b5f82392380255f0df/ui/src/utils/constants.js#L48

ui/src/theme.js
helfi92: Oh, that's very possible. My mistake.

Manish-Giri: @helfi92 um, so should I try to explore if the color of the TabIndicator can be restricted to the picker component only? For reference, here's how the tabindicator currently looks on https://firefox-ci-tc.services.mozilla.com/auth/scopes -

Screen Shot 2019-12-23 at 8 22 42 PM

But there might be other places too using this tabindicator, which would be affected by this change.

ui/src/theme.js
Manish-Giri: Got it. I'll make this change.

ui/src/theme.js
helfi92: Yes, please :)

Manish-Giri: @helfi92 So, I came across an observation while fiddling with the `PrivateTabIndicator` class - the `span` that has this class also has the `MuiTabs-indicator` class -

Screenshot 2019-12-24 20 30 53

If I target this class instead of the `PrivateTabIndicator` class, then, while the color of the tab indicator in the picker component remains exactly the way we want it (white in dark theme and black in light theme), this behavior also replicates in other places like `/auth/scopes`, thereby fixing the issue I had pointed out in my comment [here](https://github.com/taskcluster/taskcluster/pull/2106#discussion_r360624987).

In other words, whereas all tab indicators **currently** have the same color (`#4177A5`) irrespective of the theme, as shown in [this comment](https://github.com/taskcluster/taskcluster/pull/2106#discussion_r361039726), with this change, the color should be as follows -

`auth/scopes` in Dark theme -
Screenshot 2019-12-24 21 12 58

`auth/scopes` in Light theme -
Screenshot 2019-12-24 21 12 49

I wanted to check with you if this change is acceptable. I happened to come across this behavior while I was actually looking to target just the `PrivateTabIndicator` class in the picker component, using descendant/child class selectors, but that didn't work out (I'll go into more details on that if you want me to).

helfi92: What do you think of the following change?

```jsx
      MuiPickerDTTabs: {
        tabs: {
          // ...
          '& .MuiTabs-indicator': {
            backgroundColor: isDarkTheme ? '#fff' : '#000',
          },
        },
      },
```

The diff would be:

```diff
diff --git a/ui/src/theme.js b/ui/src/theme.js
index a7f054b8a..2ef178f92 100644
--- a/ui/src/theme.js
+++ b/ui/src/theme.js
@@ -360,14 +360,9 @@ const createTheme = isDarkTheme => {
         tabs: {
           color: '#fff',
           backgroundColor: THEME.SECONDARY,
-        },
-      },
-      PrivateTabIndicator: {
-        colorPrimary: {
-          backgroundColor: '#fff',
-        },
-        colorSecondary: {
-          backgroundColor: '#000',
+          '& .MuiTabs-indicator': {
+            backgroundColor: isDarkTheme ? '#fff' : '#000',
+          },
         },
       },
       MuiListItem: {

```

Manish-Giri: Thank You for the confirmation. I also had something similar, only I didn't think of using `.MuiPickerDTTabs-tabs` as the parent selector!

Screenshot 2019-12-27 18 33 47

I'll make the change.

Manish-Giri:
@helfi92 I updated the PR. Along with the last few changes you had asked for, I also refactored out the hardcoded value for dimming-out unselected text into a new constant in `utils/constants.js`, so if someone were to replicate the behavior in another component, they wouldn't need to hardcode the value again -

Screenshot 2019-12-27 19 01 34

(I hope that's okay!)

ui/src/theme.js
```diff
@@ -297,7 +297,7 @@ const createTheme = isDarkTheme => {
       },
       MuiPickersToolbarText: {
         toolbarTxt: {
-          color: THEME.PRIMARY_TEXT_DARK,
+          color: THEME.PRIMARY_TEXT_DIMMED,
```
helfi92: We won't be able to use `THEME.PRIMARY_TEXT_DIMMED` on the light theme outside the toolbar. This is because the toolbar has a fixed background color. I think it's okay to hard-code the value here. We're also doing this in the theme file so it shouldn't be too bad.

changelog/issue-2031.md
```diff
@@ -0,0 +1,4 @@
+level: minor
+reference: issue 2031
+---
+Taskcluster UI pickers now feature a time component, so for instance, users can now pick a particular time (hour:minute) on a particular day to set as the expiration time.
```
helfi92: Taskcluster UI revamped the date picker component to allow selecting the hour and the minute in addition to the date.

ui/src/theme.js
Manish-Giri: Got it. I updated the PR.

changelog/issue-2031.md
Manish-Giri: Done

Manish-Giri:
@helfi92 Thank You for being so patent, and guiding me along every step of the way! I learnt a lot 💯

helfi92:
My pleasure! Do you want to work on another issue? :-) Happy to help.

Manish-Giri:
I'm working on the deep search fix for #2123 and #2124 currently, but if you are aware of another issue that I can contribute to, I'll be happy to take it on too :)

helfi92:
Oh, that's true. We can worry about next steps after these issues are merged. Looking forward to seeing what you come up with!

PR #2111, opened 2019-11-30
[UI] Enable form save button on edit if logged out

### LINKED ISSUE #1961: [UI] Form "save" button
In the old UI, whenever a user wanted to edit a form, they had to click "Edit" first, perform their change, then click "Save". Clicking on "Edit" made it clear that they had to click on "Save" to finish the transaction.

In the new UI, the edit/view/create modes are done in a single view. In order to inform users to not forget to save their changes, we presently animate the save button whenever the form becomes "dirty" (i.e., changed). One problem is that we don't animate this button when the user is logged out. Consider the following scenario to understand the problem:

1. Navigate to https://firefox-ci-tc.services.mozilla.com and make sure you are logged out
2. Navigate to https://firefox-ci-tc.services.mozilla.com/hooks/hg-push/ash
3. Delete or add a schedule

**Expected:** The save button becomes enabled
**Actual:** The save button is disabled with a tooltip "Save Hook (Auth Required)". A user that is logged out will not receive the feedback that they need to save the hook

One promising option is to not disable the save button when the user is logged out so that changes to the hook will immediately animate the save button from a disabled state to the bright green color. We would still disable action buttons that require user authentications to avoid having to do an extra request.

Forms that will need this change:

* [x] ui/src/components/HookForm/index.jsx
* [x] ui/src/components/SecretForm/index.jsx
* [x] ui/src/components/ClientForm/index.jsx
* [x] ui/src/components/RoleForm/index.jsx
* [ ] ~ui/src/views/Quickstart/index.jsx~
* [x] Worker pool form

mmbis:
- Remove requiresAuth attribute from save button when editing
- Add isWorkerPoolDirty variable to serve as condition when checking for changes to enable/disable worked pool editor form save button

Github Bug/Issue: Fixes #1961

### DIFF
--- a/ui/src/components/ClientForm/index.jsx
+++ b/ui/src/components/ClientForm/index.jsx
@@ -385,7 +385,6 @@ export default class ClientForm extends Component {
         ) : (
           <Fragment>
             <Button
-              requiresAuth
               tooltipOpen
               variant="round"
               onClick={this.handleSaveClient}
--- a/ui/src/components/HookForm/index.jsx
+++ b/ui/src/components/HookForm/index.jsx
@@ -813,7 +813,6 @@ export default class HookForm extends Component {
                 ),
               }}
               tooltipProps={{ title: 'Save Hook' }}
-              requiresAuth
               classes={{ root: classes.successIcon }}
               variant="round"
               disabled={!this.validHook() || actionLoading || !isHookDirty}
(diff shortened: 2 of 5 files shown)


helfi92:
@mmbis Do you plan on finishing this? :-)

aryamanpuri:
@helfi92 Can I finish this up?

helfi92:
Continued in https://github.com/taskcluster/taskcluster/pull/2251.

PR #2126, opened 2019-12-03
[bug 1599122] Make provisioning behave

imbstack:
Bugzilla Bug: [1599122](https://bugzilla.mozilla.org/show_bug.cgi?id=1599122)

This is part 1 of our plan to make provisioning do the right thing. This moves scanning into the provisioning loop and removes the `running` count from this. It should ensure that we never over-provision. It will however potentially take a very long time to loop so we have to do part 2 which is to only scan a subset of the workers on every loop.

As a side-effect of this change, I needed to move `capacity` to the top-level of the worker datatype, but I actually like it more there anyway.


entrypoint
```diff
@@ -53,7 +53,6 @@ web-server/cleanup-expire-access-tokens) cd services/web-server && exec node src
 web-server/write-docs) cd services/web-server && exec node src/main.js writeDocs;;
 worker-manager/web) cd services/worker-manager && exec node src/main.js server;;
 worker-manager/provisioner) cd services/worker-manager && exec node src/main.js runProvisioner;;
-worker-manager/workerscanner) cd services/worker-manager && exec node src/main.js workerScanner;;
```
djmitche: Just to check -- this doesn't warrant a major release because removal of the k8s deployment is included in the Helm scripts -- right?

services/worker-manager/schemas/v1/worker-full.yml
```diff
@@ -53,6 +53,12 @@ properties:
     type: string
     enum: ["requested", "running", "stopped"]
 
+  capacity:
+    title: Worker Capacity
+    description: Number of tasks this worker can handle at once
```
djmitche: Since this isn't required in `createWorker`, it should probably specify a default value in the docs.

services/worker-manager/src/providers/aws.js
```diff
@@ -81,21 +81,20 @@ class AwsProvider extends Provider {
...
   }
 
-  async provision({workerPool}) {
+  async provision({workerPool, existingCapacity}) {
     const {workerPoolId} = workerPool;
 
-    if (!workerPool.providerData[this.providerId] || workerPool.providerData[this.providerId].running === undefined) {
+    if (!workerPool.providerData[this.providerId]) {
       await workerPool.modify(wt => {
         wt.providerData[this.providerId] = wt.providerData[this.providerId] || {};
```
djmitche: Speaking of eventual consistency, this is an unconditional modification and could end up deleting data some other process added at this time.  It might be better to repeat the conditional inside the `modify` callback:

```suggestion
        if (!workerPool.providerData[this.providerId]) {
          wt.providerData[this.providerId] = wt.providerData[this.providerId] || {};
        }
```

(I don't see this causing any issues, and it's unrelated to this PR, so no need to accept this change - it's just an easy way to show what I meant)

services/worker-manager/src/provisioner.js
```diff
@@ -96,26 +104,79 @@ class Provisioner {
    * Run a single provisioning iteration
    */
   async provision() {
+    if (this.provisioningLoopAlive) {
+      this.monitor.notice('loop-interference', {});
+      return;
+    }
+    this.provisioningLoopAlive = true;
```
djmitche: You'll probably want the rest of this function in a try/finally, otherwise this value may get set to `false` permanently.

services/worker-manager/src/provisioner.js
```diff
@@ -96,26 +104,79 @@ class Provisioner {
...
+      }
+    };
+
+    // Check the workers
+    await this.providers.forAll(p => p.scanPrepare());
+    await this.Worker.scan({
+      state: Entity.op.notEqual(this.Worker.states.STOPPED),
+    }, {
+      handler: async worker => {
+        seen(worker.providerId, worker.workerPoolId, worker.capacity);
```
djmitche: Oh, this is really neat!  If we change a worker pool's provider, the new provider will only provision capacity not already supported by workers from the old provider.  So it'd be possible to, for example, transition `gecko-t/t-linux-xlarge` from one AWS account to another with no downtime and no over-provisioning.  How cool is that?

services/worker-manager/src/provisioner.js
```diff
@@ -96,26 +104,79 @@ class Provisioner {
...
+        const provider = this.providers.get(worker.providerId);
+        if (provider) {
+          try {
+            await provider.checkWorker({worker});
+          } catch (err) {
+            this.monitor.reportError(err); // Just report it and move on so this doesn't block other providers
```
djmitche: But notably we've already counted its capacity!

One concern here: if checkWorker recognized that the worker is STOPPED, its capacity will still be counted.  That would be fixed on the next iteration, though.  Is that a problem?

services/worker-manager/src/provisioner.js
```diff
@@ -96,26 +104,79 @@ class Provisioner {
...
     await this.providers.forAll(p => p.prepare());
 
+    // track the providerIds seen for each worker pool, so they can be removed
+    // from the list of previous provider IDs
+    const providersByPool = new Map();
+    const seen = (providerId, workerPoolId, workerCapacity) => {
+      const v = providersByPool.get(workerPoolId);
+      if (v) {
+        v.providers.add(providerId);
+        v.count += workerCapacity;
```
djmitche: `s/count/capacity/g`?

services/worker-manager/src/provisioner.js
```diff
@@ -128,6 +189,21 @@ class Provisioner {
           } catch (err) {
             this.monitor.reportError(err, {providerId: pId}); // Just report this and move on
           }
+
+          if (!providerByPool.providers.has(pId)) {
```
djmitche: This took me a while to figure out, so maybe a comment?
```suggestion
          // If this provider no longer has any workers, ask it to remove its resources
          // and remove it from the previousProviderIds list
          if (!providerByPool.providers.has(pId)) {
```

entrypoint
imbstack: At least in our dev env helm seems to remove the old deployment. I will check with cloudops to make sure their stuff works the same way.

services/worker-manager/schemas/v1/worker-full.yml
imbstack: Ah shoot, I'll make it required actually.

services/worker-manager/src/provisioner.js
imbstack: Ah, totally! Good catch

services/worker-manager/schemas/v1/worker-full.yml
djmitche: Making it required is a breaking change to the REST API, though.  That's fine, just upgrade the changelog to major :)

services/worker-manager/src/provisioner.js
imbstack: I think it is not a problem. Just handling it on the next loop is simpler and if the loops are happening quickly enough as we plan, this is fine.

entrypoint
imbstack: I've added a note to the changelog that they should verify this but it doesn't really feel like a breaking change.  We will probably get bitten by this at this rate :p

services/worker-manager/schemas/v1/worker-full.yml
```diff
@@ -46,13 +46,31 @@ properties:
     type: string
     format: date-time
 
+  lastModified:
```
owlishDeveloper: This is out of scope of this PR, just something I remember looking at these: I wonder if we could open up some of the currently hidden useful fields in worker types? Especially now that we don't have direct access to the DB but need to debug things. Like some of the potentially useful `providerData`?

imbstack: `providerData` is intended to be used for internal-only information and sometimes might contain secrets etc. Perhaps we want another field called `tags` or something like that where we can define a format for data we want the ui to show?

services/worker-manager/schemas/v1/worker-full.yml
```diff
@@ -46,13 +46,31 @@ properties:
...
+    title: Last Modified
+    description: Date and time when this worker last changed state
+    type: string
+    format: date-time
+
+  lastChecked:
+    title: Last Checked
+    description: Date and time when the state of this worker was verified with a cloud api
```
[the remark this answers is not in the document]
djmitche: (^^ a suggestion to add a sentence to the description; I can see from the code that it's just set to the createWorker time)

services/worker-manager/schemas/v1/worker-full.yml
owlishDeveloper: It's just that in that provider data we have stuff that's useful for debugging..... but I guess ease of debugging can be achieved in other ways as well

imbstack:
I think I'll squash this and land it after tests pass

PR #2134, opened 2019-12-04
Prefer URL provided as argument over the TASKCLUSTER_ROOT_URL environment variable

marco-c:


La0:
To avoid dumb fixes like https://github.com/mozilla/code-review/pull/322

Issue #2130, opened 2019-12-03
[web-server] Add graphql tests


helfi92:
**Describe the bug**
Whenever we get new updates from the renovate-bot to bump package versions related to the web-server service, we are [not confident](https://github.com/taskcluster/taskcluster/pull/2067#issuecomment-558659636) 100% that things will still work. We should look into adding some tests to tests graphql queries, mutations, and subscriptions.

MooseJ:
@helfi92 So I started working on this, but then realized that I'm not 100% what you wanted. Are you looking for tests to be created for every GraphQL query, mutation, and subscription?

helfi92:
> @helfi92 So I started working on this, but then realized that I'm not 100% what you wanted. Are you looking for tests to be created for every GraphQL query, mutation, and subscription?

You have the right idea :) We can start with "Roles" first.

helfi92:
@MooseJ How do you feel about giving the following a try https://github.com/taskcluster/taskcluster/blob/de2c662e1780f884204b94c167d0a54d61f5d05d/services/web-server/src/graphql/Tasks.graphql#L376-L379

It would be our first subscription test.  Once we have a subscription tested, we would be able to merge https://github.com/taskcluster/taskcluster/pull/2067

MooseJ:
@helfi92 Yeah I can give that shot!

helfi92:
Thank you so much!

helfi92:
@MooseJ Happy new year! How is it going? :-)

MooseJ:
@helfi92 Hey sorry, I've been going a bit slower with this because of the holidays. I have an idea of what the tests are going to be, but this was a more difficult than I thought it would be 😄I also think I have an idea of what I need to do, but I'm going to ask you some questions later because I feel like I'm going to get stuck. I should have a better idea of things by the end of the weekend

MooseJ:
@helfi92 

So, I am currently getting this error from my test results:
```js
{
  "graphQLErrors": [
    {
      "message": "Cannot destructure property `tasksSubscriptions` of 'undefined' or 'null'.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "tasksSubscriptions"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "TypeError: Cannot destructure property `tasksSubscriptions` of 'undefined' or 'null'.",
[...]

PR #2139, opened 2019-12-06
[web-server] Replace Promise.all in loaders with try/catch

### LINKED ISSUE #2125: [web-server]  Replace Promise.all in loaders with try/catch
**Describe the bug**
If an error is thrown in Promise.all, the whole thing is rejected. We should gracefully catch errors in graphql loaders.

**Taskcluster version**
23.0.0

**Additional context**
We would need to find all `Promise.all` calls in `src/loaders` and replace them with try catch statement. The change should be similar to https://github.com/taskcluster/taskcluster/pull/1944/files.

MooseJ:
Github Bug/Issue: Fixes #2125

I wasn't sure if _everything_ in the `promise.all`'s should be wrapped with try catches, so I did just to be safe. Let me know what you think!

### DIFF
--- a/services/web-server/src/loaders/auth.js
+++ b/services/web-server/src/loaders/auth.js
@@ -13,7 +13,7 @@ module.exports = (clients, isAuthed, rootUrl, monitor, strategies, req, cfg) => 
 
         // If req.user is falsy for one query, it will be falsy for the rest.
         // Instead of returning an error in the try catch block below for every query,
-        // we just want to throw an unauthorized error 
+        // we just want to throw an unauthorized error
         if (!req.user) {
           throw unauthorizedError;
         }
--- a/services/web-server/src/loaders/auth.js
+++ b/services/web-server/src/loaders/auth.js
@@ -15,22 +15,33 @@ module.exports = (clients, isAuthed, rootUrl, monitor, strategies, req, cfg) => 
           throw unauthorizedError;
         }
 
-        const credsResponse = await generateCredentials({
-          cfg,
-          strategy: strategies[req.user.identityProviderId],
-          identity: req.user.identity,
-          monitor,
-        });
+        try { 
+          const credsResponse = await generateCredentials({
+            cfg,
+            strategy: strategies[req.user.identityProviderId],
+            identity: req.user.identity,
+            monitor,
+          });
 
-        await regenerateSession(req);
+          await regenerateSession(req);
 
-        return credsResponse;
+          return credsResponse;
+        } catch (err) {
+          return err;
+        }
       }),
     );
   });
+  
   const isLoggedIn = new DataLoader(queries =>
     Promise.all(
-      queries.map(() => Boolean(req.user)),
+      queries.map(() => {
+        try {
+          Boolean(req.user)
+        } catch (err) {
+          return err;
+        }
+      }),
     ),
   );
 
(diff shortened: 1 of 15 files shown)


services/web-server/src/loaders/auth.js
```diff
@@ -15,22 +15,33 @@ module.exports = (clients, isAuthed, rootUrl, monitor, strategies, req, cfg) =>
           throw unauthorizedError;
```
MooseJ: I left the above code unwrapped since it seemed like the throw unauthorizedError was intentional and probably be handled by the upstream caller. Also if req.user is undefined for one query it will be undefined for the others, so they would all fail anyways.

services/web-server/src/loaders/auth.js
```diff
@@ -15,22 +15,33 @@ module.exports = (clients, isAuthed, rootUrl, monitor, strategies, req, cfg) =>
...
   const isLoggedIn = new DataLoader(queries =>
     Promise.all(
-      queries.map(() => Boolean(req.user)),
+      queries.map(() => {
+        try {
+          return Boolean(req.user);
+        } catch (err) {
+          return err;
+        }
+      }),
```
helfi92: No need to wrap this in a try catch statement since I don't think `Boolean(req.user)` will ever throw an error.

services/web-server/src/loaders/auth.js
helfi92: That makes a lot of sense. A comment that says what you just commented here would probably help for posterity :)

services/web-server/src/loaders/tasks.js
```diff
@@ -6,16 +6,26 @@ const Task = require('../entities/Task');
...
+        }
+      }),
+    ),
   );
   const indexedTask = new DataLoader(indexPaths =>
-    Promise.all(indexPaths.map(indexPath => index.findTask(indexPath))),
+    Promise.all(
+      indexPaths.map(indexPath => {
+        try {
+          return index.findTask(indexPath);
```
helfi92: Hm, I'm not sure if this will throw since we're not awaiting. Perhaps this can be solved with:

```jsx
try {
  return await index.findTask(indexPath);
} catch (err) {
  return err;
}
```

MooseJ: Should some other calls that aren't being awaited be awaited as well. Like the one's in `github.js` and `client.js` for example

helfi92: Good catch! Yes, please.

helfi92:
@MooseJ Are you interested in writing tests for some of these services? I'm written some in https://github.com/taskcluster/taskcluster/tree/master/services/web-server/test/graphql and https://github.com/taskcluster/taskcluster/tree/master/services/web-server/test/loaders. Let me know if you're interested in that or whether you're looking for something else.

MooseJ:
@helfi92 Yeah I'd be willing to work on tests! I won't be able to start until Sunday though. The graphql tests should be tied to this issue right? https://github.com/taskcluster/taskcluster/issues/2130

helfi92:
> @helfi92 Yeah I'd be willing to work on tests! I won't be able to start until Sunday though.

Awesome!

> The graphql tests should be tied to this issue right? #2130

Correct :)

PR #2140, opened 2019-12-06
Bug 1587511 pt 1 tests

owlishDeveloper:
Bugzilla Bug: [1577839](https://bugzilla.mozilla.org/show_bug.cgi?id=1577839)

I'd like some feedback on the way I stub `EC2.describeInstanceStatus`, especially from @djmitche (I know bstack approves)

I disabled "comma-dangle" for the file because I'd like to format long function arguments on every line, and you don't need a comma after the last function argument or parameter

I used Dustin's idea of `fake.call` in the `EC2.runInstances` stub. Re: [this comment in the old Dustin's PR](https://github.com/taskcluster/taskcluster/pull/1701/files#diff-71648e88330ea7c4c8837b3d2fadf21eR5) - the bug in the fixture was that in my eternal wisdom I instantiated the stub and passed in default `TagSpecifications`, so that was the only thing the stub could assert on. In case of user-defined tags, we needed to assert on the whole entirety: the default tags + the user tags. Assertion itself _did_ work tho 😛😄 I agree, however, that assertions are better done in one place instead of in two places, so the `fake.call` was brought from the above PR.

### DIFF
--- a/services/worker-manager/test/provider_aws_test.js
+++ b/services/worker-manager/test/provider_aws_test.js
@@ -1,3 +1,5 @@
+/* eslint-disable comma-dangle */
+
 const {ApiError} = require('../src/providers/provider');
 const assert = require('assert');
 const sinon = require('sinon');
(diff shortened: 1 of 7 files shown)


services/worker-manager/test/provider_aws_test.js
```diff
@@ -156,14 +126,100 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
...
+            ResourceType: 'instance',
+            Tags: [
+              {Key: 'CreatedBy', Value: 'taskcluster-wm-aws'},
+              {Key: 'Owner', Value: 'whatever@example.com'},
+              {Key: 'ManagedBy', Value: 'taskcluster'},
+              {Key: 'Name', Value: 'foo/bar'},
+              {Key: "WorkerPoolId", Value: "foo/bar"},
+            ],
+          },
+        ],
```
djmitche: Is this an example of the comma it would complain about?  I'm just curious :)

djmitche:
> I'd like some feedback on the way I stub EC2.describeInstanceStatus, especially from @djmitche (I know bstack approves)

I approve too :)

services/worker-manager/test/provider_aws_test.js
owlishDeveloper: [this is an example](https://github.com/taskcluster/taskcluster/pull/2140/files#diff-7137f60d74d4c486beca74a3f5ae881aR211) of the comma it would complain about. I have several of those in the file.

owlishDeveloper:
> Awesome, ty! I think I'd prefer to not change the eslint rules there but minor nit that we can just agree to disagree on. 🚢

In response to this, I decided to be more deliberate and disable that rule on specific lines. That way when people are adding tests in the future they will get legitimate errors if they miss a trailing comma where it's actually needed. Also, disabling each line makes it very obvious that I looked at each error, know what it is and disable it on purpose. And if people find my way of formatting function parameters unreadable, they can just refactor

PR #2141, opened 2019-12-06
[github] Add a flexible route for handling task status reporting.

tomprince:
Bugzilla Bug: [1533235](https://bugzilla.mozilla.org/show_bug.cgi?id=1533235)

@owlishDeveloper  We've talked a few times about having a more flexible route that can be used to report checks from tasks. This is a very quick sketch of what ain implementation for it might look like.


imbstack:
This makes sense to me as a way forward.

PR #2142, opened 2019-12-06
[web-server] Add preliminary graphql tests

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/2130.


helfi92:
I was having trouble setting up tests for subscriptions. If you have any idea, let me know :)

djmitche:
image

services/web-server/test/graphql_test.js
```diff
@@ -0,0 +1,95 @@
...
+      source: "https://github.com/taskcluster/taskcluster",
+    },
+    tags: {
+      purpose: "taskcluster-testing",
+    },
+    extra: {},
+  }, options);
+
+  suite('graphql', function() {
+    test('query works', async function() {
```
djmitche: It might be good to have the suite named "Queue" for example, and the test named "create and retrieve task".  Then when we add more tests they can follow that same pattern.

PR #2145, opened 2019-12-09
[web-server] Add another test case

helfi92:


services/web-server/test/graphql/queue_graphql_test.js
```diff
@@ -1,15 +1,14 @@
 const assert = require('assert');
-const taskcluster = require('taskcluster-client');
+const taskcluster = require('../../../../clients/client/src');
```
djmitche: This shouldn't need to be a relative path -- this and taskcluster-lib-testing (below) are defined in the root `package.json`.  In my testing, the PR still works with this and the tc-lib-testing change reverted.

helfi92: Webstorm must have changed the URL for some odd reason...

PR #2146, opened 2019-12-09
Bug 1595749 (comment 4) - Fix identityFromProfile

helfi92:
Bugzilla Bug: [1595749](https://bugzilla.mozilla.org/show_bug.cgi?id=1595749)

Looks like we didn't change identityFromProfile after switching to the CIS API (v2 of the PersonAPI)

Here's the [profile](https://github.com/mozilla-iam/cis/blob/master/well-known-endpoint/tpl/profile.schema) schema.

### DIFF
--- a/changelog/bug-1595749.md
+++ b/changelog/bug-1595749.md
@@ -1,0 +1,4 @@
+level: patch
+reference: bug 1595749
+---
+Taskcluster login now properly handles an edge case where a couple of users were hitting which prevented them to login.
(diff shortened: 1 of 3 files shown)


helfi92:
> Are all users using this new format now, or do we need to remain compatible with the old format, as well?

`identityFromProfile` didn't work well after moving away from using the Auth0 management API to the CIS API (we never used the PersonAPI, we just called it PersonAPI since it's technically v2 of that API; I plan on renaming it to CIS soon). The profile argument in `identityFromProfile` comes from the CIS API. In the days of taskcluster-login, we used to read the profile from the auth0 management API. That being said, all users are using this format now.

djmitche:
Awesome, thanks for the details!

PR #2148, opened 2019-12-09
Bug 1602464 - fix 'yarn build' now that yarn version is upgraded

djmitche:
Bugzilla Bug: [1602464](https://bugzilla.mozilla.org/show_bug.cgi?id=1602464)


djmitche:
I think that's unrelated -- at a guess, you need to run `gclould auth login` again?

djmitche:
(intentionally merging without changelog as this would be silent anyway)

Issue #2127, opened 2019-12-03
[UI] Clicking on the copy icon should show feedback


helfi92:
**Describe the bug**
Clicking on the copy icon does not show any kind of feedback that whatever the user was trying to copy has now been copied to the clipboard.

**To Reproduce**

1. Go to https://firefox-ci-tc.services.mozilla.com/tasks/UBaqTi4kQ0elIUsYW2YdKg
2. Click on the list item with the copy icon

![Screen Shot 2019-12-03 at 2 43 28 PM](https://user-images.githubusercontent.com/3766511/70083932-58334e80-15db-11ea-8669-afbba9d2afb2.png)

**Actual behavior**
Nothing happens.

**Expected behavior**
A feedback is shown to the user that it was copied.

**Taskcluster version**
23.0.0

**Additional context**

Whenever a user tries to copy something after clicking on a list item or the copy icon, we can replace the copy icon with a check icon (e.g., ✔️) for a short duration. This will make sure the user knows that whatever they were trying to copy has been copied to the clipboard. This will avoid having users click on the button multiple times. In order to do this, we can create a custom component that augments it with this feature.

aryamanpuri:
@helfi92 Can I give this a try?
Thinking of changing the icon to the text "copied" .

aryamanpuri:
@helfi92 How to setup locally as the task is already submitted and creating new task locally is giving Unknown server error ?

helfi92:
Does http://localhost:5080/tasks/fFaGg92IRWa5Rmj69FMDNQ work?

aryamanpuri:
@helfi92 No

aryamanpuri:
@helfi92 Yes it worked

aryamanpuri:
>  https://firefox-ci-tc.services.mozilla.com/tasks/UBaqTi4kQ0elIUsYW2YdKg

Yes it worked

helfi92:
Resolved in https://github.com/taskcluster/taskcluster/commit/22772ba5c17aaeaab6a192245600db20afa5a197

PR #2151, opened 2019-12-10
[UI] On Clicking the copy icon it gives feedback .

aryamanpuri:
#2127

### DIFF
--- a/ui/src/components/TaskDetailsCard/index.jsx
+++ b/ui/src/components/TaskDetailsCard/index.jsx
@@ -15,6 +15,8 @@ import ListItemText from '@material-ui/core/ListItemText';
 import ChevronUpIcon from 'mdi-react/ChevronUpIcon';
 import ChevronDownIcon from 'mdi-react/ChevronDownIcon';
 import ContentCopyIcon from 'mdi-react/ContentCopyIcon';
+import CheckIcon from 'mdi-react/CheckIcon';
+import { green } from '@material-ui/core/colors';
 import LinkIcon from 'mdi-react/LinkIcon';
 import OpenInNewIcon from 'mdi-react/OpenInNewIcon';
 import DateDistance from '../DateDistance';
@@ -58,6 +60,11 @@ import Link from '../../utils/Link';
   listItemText: {
     paddingLeft: theme.spacing(2),
   },
+  checkIcon: {
+    '& .mdi-icon': {
+      fill: green,
+    },
+  },
 }))
 /**
  * Render information in a card layout about a task.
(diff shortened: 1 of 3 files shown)


ui/src/components/TaskDetailsCard/index.jsx
```diff
@@ -91,6 +98,42 @@ export default class TaskDetailsCard extends Component {
...
+  handleCreatedCopied = () => {
+    if (this.state.Createdcopied) {
+      this.setState({ Createdcopied: false });
+    } else {
+      this.setState({ Createdcopied: true });
+      setTimeout(() => {
+        this.setState({ Createdcopied: false });
+      }, 2000);
+    }
   };
```
helfi92: Having to write all of this logic in all of the places we use the copy feature will make us prone for errors. Instead, I think we should create a custom component that handles of this logic. This will make sure we don't re-write the same code multiple times.

aryamanpuri: @helfi92 So I should write the whole function or just the timeout function in another component? And one more thing, in another component how to make it generalize for all the copy buttons as they are independent of others?
_Is there is this.setstate({ ${``}copied :false}) kind of thing in  react ?_

helfi92: No need to do `${``}copied :false`. Since it will be in its own component, we just need to worry of a single state at a time. Perhpas just `this.setState({ copied: false });` for example.

aryamanpuri: So I should write all the functions in another component, not a single one?

helfi92: We want to create a component (e.g., ui/src/components/<new-component>.jsx) then use it in `TaskDetailsCard`, `TaskRunCard`. The component can be used multiple times in the same file.

aryamanpuri: What should be there in the render of the new component?

helfi92: I would have to think about it. What did you try?

aryamanpuri: I haven't tried it yet  , bit busy with semester end exam .what should be in render I was confused about..

aryamanpuri: @helfi92 what are your about it ?

aryamanpuri: @helfi92 What are your views about this?

helfi92: I looked into it. I wrote something. I'll /cc you in the pull-request.

helfi92: https://github.com/taskcluster/taskcluster/pull/2186#issuecomment-566572404

PR #2156, opened 2019-12-11
[tc-lib-postgres] Store versions in an array rather than Map

helfi92:


helfi92:
`yarn test` shows:

```
database_test.js
    - (TEST_DB_URL is not set)
```

Also, not sure why rebasing is not working...

```
git pull https://github.com/helfi92/taskcluster.git tc-lib-postgres --rebase
From https://github.com/helfi92/taskcluster
 * branch                tc-lib-postgres -> FETCH_HEAD
Current branch replace-map-array is up to date.
```

djmitche:
BTW I'd be totally happy for you to just push these changes to the original branch.  We'll do the review stage later.

libraries/postgres/src/Schema.js
```diff
@@ -39,14 +37,14 @@ class Schema{
 
       const version = yaml.safeLoad(fs.readFileSync(filename));
 
-      versions.set(version.version, version);
+      versions[version.version - 1] = version;
```
djmitche: We could avoid the `-1` by putting a "dummy" version 0 in the array.  I don't know if that would make things clearer or not :)

PR #2165, opened 2019-12-11
[UI] Improve /auth/scopes experience

helfi92:


ui/src/views/Scopes/ViewScope/index.jsx
```diff
@@ -79,17 +78,15 @@ export default class ViewScope extends Component {
...
     return (
       <Dashboard
-        title={selectedScope}
+        title={`Scope ${selectedScope}`}
+        disableTitleFormatting
         search={
           <Search
             onSubmit={this.handleSearchSubmit}
-            placeholder="Result contains"
+            placeholder="Roles/Clients contains"
```
djmitche: Maybe singular here? ("Role/Client contains")

PR #2168, opened 2019-12-12
Bug 1603391 - make worker pool config property an anyOf of possible provider configs

petemoore:
Bugzilla Bug: [1603391](https://bugzilla.mozilla.org/show_bug.cgi?id=1603391)

### DIFF
--- a/changelog/bug-1603391.md
+++ b/changelog/bug-1603391.md
@@ -1,3 +1,4 @@
-level: silent
+level: major
 reference: bug 1603391
 ---
+Worker Pool `config` properties now must match one (or more) of the provider json schemas. Previously, config object of a Worker Pool definition was not validated for conformance to any of the provider json schemas. This resulted in limited information being provided in the createWorkerPool api reference documentation, and no go types being generated in the go client to represent the various config provider data structures.
--- a/changelog/bug-1603391.md
+++ b/changelog/bug-1603391.md
@@ -1,0 +1,3 @@
+level: silent
+reference: bug 1603391
+---
--- a/services/worker-manager/schemas/v1/config-google.yml
+++ b/services/worker-manager/schemas/v1/config-google.yml
@@ -2,7 +2,7 @@ $schema: "/schemas/common/metaschema.json#"
 title: Google Provider Config
 description: |
   'The config section of a workertype is using a provider with implementation google must match this.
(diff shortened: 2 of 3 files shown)


petemoore:
Note, this change is untested. Is there an automated deployment of this PR which would allow me to manually test that the docs are updated as expected?

changelog/bug-1603391.md
```diff
@@ -0,0 +1,3 @@
+level: silent
+reference: bug 1603391
```
djmitche: This certainly warrants mention as an API change.  If that's a breaking change, then `major`.

imbstack:
iirc the reasoning to not do this is that the error it throws for invalid configs is ultra-confusing and actually somewhat misleading. If you can confirm that I'm wrong here though I'd love to have this as well.

petemoore:
> iirc the reasoning to not do this is that the error it throws for invalid configs is ultra-confusing and actually somewhat misleading. If you can confirm that I'm wrong here though I'd love to have this as well.

@imbstack 
The problem was that `oneOf` matched multiple subschemas, I've fixed that in the commit above by using `anyOf` instead which does what we really want.

The `yarn generate` failure is due to the schema change generating additional code in the generated go client, which is a good thing!

So I think I just need to update the changelog to make this is a major change, and we should be good to go!

imbstack:
Just to update here:

We discussed this in-channel yesterday and we're going to hold off on this until we can make sure that the error message sent to users who misconfigure something is usable.

petemoore:
We had a quick chat in channel too about possibly having a different top level key based on the cloud the provider is for, e.g. instead of having a single key `config` instead having e.g. keys `aws`, `gcp`, `<other-cloud-foo>`. This would help a few things:

1) It makes it easier in future to be able to have a worker pool that can run in multiple clouds, since a worker pool definition could at some point allow multiple cloud configurations, since they would now have different keys
2) It makes the shape of the config section a little more deterministic - from the key name, you know the subschema of what can sit underneath it, so validation is simpler and easier to reason about
3) No more `anyOf`/`oneOf` semantics that make it difficult to map to go types in the go client

We also agreed (@imbstack and I) that it probably warrants a separate RFC in our RFC repo...

imbstack:
This will be followed up by an rfc as mentioned in the last comment when we get around to it. closing the PR for now

PR #2173, opened 2019-12-12
[tc-lib-postgres] Do not allow capital letters for db methods

helfi92:
Rather than rely on capital letters, I thought about disallowing them. Using snake case for method names would align better with method names. Capital letters should probably just be SQL commands. Let me know if you disagree. Also, this is not tested.

### DIFF
--- a/libraries/postgres/test/database_test.js
+++ b/libraries/postgres/test/database_test.js
@@ -125,14 +125,10 @@ dbSuite(path.basename(__filename), function() {
       }],
     });
 
-    try {
-      await Database.upgrade({schema, runUpgrades: true, readDbUrl: this.dbUrl, writeDbUrl: this.dbUrl});
-      assert.fail('should have failed');
-    } catch (e) {
-      assert(e);
-    }
-    finally {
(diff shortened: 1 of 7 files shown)


djmitche:
Snake case was what I meant -- my notes weren't very clear.  Basically I was saying we should be *testing* with things that aren't all lowercase, so that tests find issues with omitted `"` characters.

But prohibiting is OK too :)

libraries/postgres/test/database_test.js
```diff
@@ -104,4 +104,35 @@ dbSuite(path.basename(__filename), function() {
...
+        },
+      }],
+    });
+
+    try {
+      await Database.upgrade({schema, runUpgrades: true, readDbUrl: this.dbUrl, writeDbUrl: this.dbUrl});
+      assert.fail('should have failed');
+    } catch (e) {
+      assert(e);
+    }
```
djmitche: `assert.rejects` is useful here -- otherwise, while I think this is right, it's very easy to get this construct wrong.

PR #2174, opened 2019-12-12
Bug 1436478 - Switch Taskcluster to Postgres

djmitche:
(draft PR to try to run the test for this library in CI)


djmitche:
Seems to have worked -- the error is legit :)

infrastructure/tooling/src/generate/generators/db-schema.js
```diff
@@ -10,6 +10,6 @@ exports.tasks = [{
   run: async (requirements, utils) => {
     const schema = Schema.fromDbDirectory(path.join(REPO_ROOT, 'db'));
 
-    writeRepoJSON('generated/db-schema.json', JSON.parse(schema.asSerializable()));
+    writeRepoJSON('generated/db-schema.json', schema.asSerializable());
```
helfi92: How do we guarantee that the schema won't change between `yarn generate` if we remove `stringify(...)` from the `asSerializable` method?

helfi92: Nevermind, I just saw that the following commit uses `Map.prototype.values` to keep track of insertion order. Smart!

djmitche: Yep, and also, `writeRepoJson` uses `json-stable-stringify`, so it should still be stable.

helfi92: Will the data still be readable? Previously, we used `{ space: 2 }`.

helfi92: Ah, ok!

libraries/entities/src/index.js
```diff
@@ -0,0 +1,210 @@
...
+
+    Object.entries(context).forEach(([key, value]) => {
+      assert(this.context.includes(key), `context key ${key} was not declared in Entity.configure`);
+      ctx[key] = value;
+    });
+
+    return ctx;
+  }
+
+  _doCondition(conditions) {
```
djmitche: I was thinking a little more about this while walking the dog.  Looking at uses of `load`, `query` and `scan`, I see the following kinds of conditions:

 * things that are used to generate partitionKey / rowKey
 * datestamps (`expires: Entity.op.greaterThanOrEqual(new Date())`)
 * problematic cases
   * `storageType: 'blob'` in `services/queue/src/api.js` -- we can in fact delete this code, as we no longer support blobs
   * `resolveTask` in `services/queue/src/dependencytracker.js` -- we can change this around to just fetch all dependencies of the task (omitting the condition) and check the property with an `if` statement in the JS code.  In this particular case, that won't be especially inefficient.

So, I think we can eliminate the two problematic cases and *only* build conditions for the first two cases -- that is, the only WHERE clause elements we should ever build dynamically are for datestamp properties, where we are given a Date object in JS and can properly convert that to a SQL string without risk of SQL injection.

helfi92:
@djmitche Any chance you could look at tc-lib-entities and provide a review, please? Thanks!

libraries/entities/test/ascendingintegerkey_test.js
```diff
@@ -71,11 +71,11 @@ helper.dbSuite(path.basename(__filename), function() {
       const createdEntity = await TestTable.create(entry);
 
       // TODO: Fix this
```
djmitche: TODONE :)

libraries/entities/test/entity_query_test.js
```diff
@@ -216,53 +200,17 @@ helper.dbSuite(path.basename(__filename), function() {
       });
     });
 
-    test('Filter by active === true', async function() {
+    test.only('Filter by active === false throws an error', async function() {
```
djmitche: stray `.only`..

libraries/entities/src/index.js
```diff
@@ -0,0 +1,633 @@
...
+    try {
+      rk = this.__rowKey.exactFromConditions(conditions);
+    } catch (e) {
+      if (options.matchRow === 'exact') {
+        assert(rk, 'conditions should provide enough constraints for constructions of the row key');
+      }
+    }
+
+    if (pk) {
+      condition.push(`partition_key = '${pk}'`);
```
djmitche: This (and `'${rk}'` below) is substituting an arbitrary value into SQL.  I think we'll need to change the stored function to take an `rk` and `pk` argument and use placeholders to pass those into the resulting SQL statement.

helfi92: Done!

helfi92:
Done with the suggested changes.

libraries/entities/test/db/versions/0001.yml
```diff
@@ -100,17 +100,28 @@ methods:
...
         page_offset integer;
       begin
-        if condition is not null then
-          sql := sql || ' where ' || condition;
+        if test_entities_scan.pk is not null or test_entities_scan.rk is not null or condition is not null then
+          sql := sql || ' where ';
         end if;
 
+        if test_entities_scan.pk is not null then
+          partition_key_var := 'partition_key = ' || quote_literal(test_entities_scan.pk);
```
djmitche: I definitely trust Pg's own `quote_literal` to do the correct thing :)

djmitche:
Oh, and some lint bits :)

helfi92:
wow, pagination is so fast on postgres...

Kapture 2020-03-16 at 13 16 19

infrastructure/tooling/src/importer/importer.js
```diff
@@ -73,75 +73,65 @@ const importer = async options => {
...
         {
-          name: `${tableName}-11/14`,
+          name: `${tableName}-11/12`,
           filter: `PartitionKey ${Table.Operators.GreaterThanOrEqual} ${Table.Operators.string("n")} ${Table.Operators.And} PartitionKey ${Table.Operators.LessThan} ${Table.Operators.string("u")} `,
           title: `[n, u[`,
         },
         {
-          name: `${tableName}-12/14`,
+          name: `${tableName}-12/12`,
           filter: `PartitionKey ${Table.Operators.GreaterThanOrEqual} ${Table.Operators.string("u")} ${Table.Operators.And} PartitionKey ${Table.Operators.LessThanOrEqual} ${Table.Operators.string("z")} `,
```
djmitche, line 132: I think you want to remove the `z` part of this and just do u to infinity.

helfi92: Z is covered in 7/12

PR #2176, opened 2019-12-13
[tc-lib-postgres] Add service name for each method

helfi92:


libraries/postgres/src/Database.js
```diff
@@ -24,11 +25,17 @@ class Database {
...
-    schema.allMethods().forEach(({ name, mode}) => {
+    schema.allMethods().forEach(({ name, mode, serviceName: procServiceName }) => {
       // ensure that quoting of identifiers is correct
       assert(!/.*[A-Z].*/.test(name), 'db procedure methods should not have capital letters');
 
       this.procs[name] = async (...args) => {
+        if (serviceName !== procServiceName && mode === WRITE) {
+          throw new Error(
+            `${serviceName} is not allowed to call any methods that do not belong to this service and which have mode=WRITE`,
+          );
```
djmitche: Nice!

libraries/postgres/test/database_test.js
```diff
@@ -94,7 +95,7 @@ dbSuite(path.basename(__filename), function() {
   });
 
   test('setup creates JS methods that can be called', async function() {
-    await Database.upgrade({schema, runUpgrades: true, readDbUrl: this.dbUrl, writeDbUrl: this.dbUrl});
+    await Database.upgrade({schema, runUpgrades: true, readDbUrl: this.dbUrl, writeDbUrl: this.dbUrl, serviceName: 'service-1'});
```
djmitche: This is a bit awkward, though -- an upgrade doesn't happen in the context of a single service.

PR #2177, opened 2019-12-13
[tc-lib-postgres] Assert that methods names are unique

helfi92:


libraries/postgres/test/database_test.js
```diff
@@ -101,6 +101,45 @@ dbSuite(path.basename(__filename), function() {
...
+          },
+          testdata: {
+            mode: 'read',
+            serviceName: 'service-2',
+            args: '',
+            returns: 'void',
+            body: `begin
+              insert into testing values (1, 2), (3, 4);
+            end`,
+          },
```
djmitche: I'm curious how this test case works: doesn't that create a JS object with just one property?

helfi92: oo, you're right. I actually haven't tested this. I was relying on CI to run the tests for me. It's probably a good time for me to read the README and set up the database.

helfi92:
@djmitche It turns out that `safeLoad` disallows duplicate key fields which we use in the `fromDbDirectory` method.

djmitche:
Awesome :)

PR #2178, opened 2019-12-13
Bug 1537922 - test and fix when publishing clientCreated fails

djmitche:
Bugzilla Bug: [1537922](https://bugzilla.mozilla.org/show_bug.cgi?id=1537922)

Write a test, find a bug!

### DIFF
--- a/services/auth/test/client_test.js
+++ b/services/auth/test/client_test.js
@@ -1,9 +1,10 @@
-const assert = require('assert');
+const assert = require('assert').strict;
 const helper = require('./helper');
 const _ = require('lodash');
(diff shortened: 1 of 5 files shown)


services/auth/test/client_test.js
```diff
@@ -1,9 +1,10 @@
-const assert = require('assert');
+const assert = require('assert').strict;
```
imbstack: Oh, can we just make this import `strict` everywhere rather than updating all calls?

djmitche: That's my understanding from the docs anyway, yeah.

PR #2191, opened 2019-12-18
[db] Display stored procedures in README

helfi92:


db/README.md
```diff
@@ -6,6 +6,16 @@ This directory defines the Taskcluster database:
...
 
+## Stored Procedures
+
+<!-- SP BEGIN -->
+### secrets
+
+| Name | Mode | Args | Returns |
+| --- | --- | --- | --- |
+| get_secret | Symbol(read) | name text | table (secret text) |,| get_secret_with_expires | Symbol(read) | name text | table (secret text, expires timestamp) |,| list_secrets | Symbol(read) |  | table (name text) |,| list_secrets_with_expires | Symbol(read) |  | table (name text, expires timestamp) |,| remove_secret | Symbol(write) | name text | void |,| remove_secret | Symbol(write) | name text | void |,| set_secret | Symbol(write) | name text, secret text | void |,| set_secret_with_expires | Symbol(write) | name text, secret text, expires timestamp | void |
+<!-- SP END -->
```
djmitche: It looks like this is `,`-separated instead of newlines?

Also, maybe this is a good time to just replace the Symbol(..) with regular strings :)

Finally, should we include a description field with each of these?

PR #2192, opened 2019-12-18
[tc-lib-postgres] Generate JSON version of the schema

helfi92:


libraries/postgres/src/Schema.js
```diff
@@ -12,7 +13,7 @@ class Schema{
    * the Postgres DO statment; that is usually 'BEGIN stmt; stmt; .. END'.
    */
   // TODO: Make sure that procedure argument values don't change
-  constructor(versions, access) {
+  constructor(versions = schemaJson.versions, access = schemaJson.access) {
```
djmitche: I have a vague preference that reference to the generated bits occurs in the `db/` directory, and that this library remains agnostic to it.  So services would call a method in `taskcluster-db` to construct a new Database, and that would automatically load a schema from the generated schema.

package.json
```diff
@@ -129,6 +129,7 @@
     "console-taskgraph": "^1.6.0",
     "cronstrue": "^1.82.0",
     "cross-env": "^6.0.3",
+    "deep-sort-object": "^1.0.2",
```
djmitche: Is this to avoid spurious changes in the generated JSON?  We have `json-stable-stringify` for that purpose already, so perhaps we can just reuse that..

PR #2195, opened 2019-12-18
Bug 1596171 - do not omit schedulerId when cloning a task

djmitche:
Bugzilla Bug: [1596171](https://bugzilla.mozilla.org/show_bug.cgi?id=1596171)

I will admit I didn't think too hard about this change.  WCPGW?

### DIFF
--- a/changelog/bug-1596171.md
+++ b/changelog/bug-1596171.md
@@ -1,0 +1,4 @@
+level: patch
+reference: bug 1596171
+---
+The "Edit" button in the task view no loner removes the schedulerId
--- a/ui/src/views/Tasks/ViewTask/index.jsx
+++ b/ui/src/views/Tasks/ViewTask/index.jsx
@@ -293,7 +293,6 @@ export default class ViewTask extends Component {
         ...TASK_ADDED_FIELDS,
         'routes',
         'taskGroupId',
-        'schedulerId',
         'priority',
         'dependencies',
         'requires',


tomprince:
Given that there is (now) a scheduler-id that is used conventionally for tasks from the UI, would it make sense to set that by default, instead?

djmitche:
Maybe, but it feels weird to not only *omit* a field but to actually change it to another field that may or may not be useful in the particular deployment in use.

It does seem, in the absence of other information, like "the schedulerId the original task has" is the best default for schedulerId.

tomprince:
`-` was implicitly the scheduler-id for tasks created in the UI. But that is opaque. My suggestion in [Bug 1596171](https://bugzilla.mozilla.org/show_bug.cgi?id=1596171) was to change that implicit convention with an explicit one. I think changing that to an explicit default with a reasonable name is sensible.

For example, if somebody takes a test task from mozilla-central, and edits it, they would likely be able to create the task (as testers are available on level-1), but the scheduler-id would be gecko-3, which they would not have permissions for.

djmitche:
I'm almost convinced by that argument :)

@imbstack, @owlishDeveloper what do you think?

imbstack:
I'm convinced by it so long as we make this official and documented, etc. Perhaps there are other fields that we can define this way, etc.

djmitche:
Hm, so we would need some kind of convention in the docs for how to clone a task.  I like that, too.

PR #2196, opened 2019-12-18
Bug 1602985 - don't check for dirty repo when releasing py client

djmitche:
..and fix the log filenames in 'yarn release'

Bugzilla Bug: [1602985](https://bugzilla.mozilla.org/show_bug.cgi?id=1602985)

### DIFF
--- a/clients/client-py/release.sh
+++ b/clients/client-py/release.sh
@@ -25,6 +25,8 @@ if [ -n "$modified" ]; then
   echo "There are changes in the local tree.  This probably means"
   echo "you'll do something unintentional.  For safety's sake, please"
   echo 'revert or stash them!'
+  echo ''
+  echo -- "$modified"
   exit 66
 fi
 
--- a/changelog/bug-1602985.md
+++ b/changelog/bug-1602985.md
@@ -1,0 +1,3 @@
+level: silent
+reference: bug 1602985
+---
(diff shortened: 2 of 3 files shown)


clients/client-py/release.sh
```diff
@@ -17,17 +17,6 @@ cd "$(dirname "${0}")"
...
-# continuing. git stash can help here! Untracked files can make it into release
-# so let's make sure we have none of them either.
-modified="$(git status --porcelain)"
-if [ -n "$modified" ]; then
-  echo "There are changes in the local tree.  This probably means"
-  echo "you'll do something unintentional.  For safety's sake, please"
-  echo 'revert or stash them!'
-  exit 66
-fi
-
```
imbstack: I'm confused about why this would be getting triggered. The guard here actually seems like something we want?

djmitche: I'm a bit confused, too.  It might be that there's some temp file generated by a process occurring at the same time.  At any rate, this is now operating in the controlled setting of a release process (and eventually in CI), so I don't think the guard is actually that useful.

djmitche: I could find out if this was easier to test, but unfortunately I don't know how to test except by making more releases..

imbstack: Let's get it to log what `git status` says and then land this.

PR #2197, opened 2019-12-19
Bug 1600966 - azure-provider initial add

milescrabill:
- adds initial azure-provider for worker-manager

I've done some testing on my dev cluster. We spin up VMs with everything they need to go from w-m => w-r => docker-worker, though the image we're using has some issues and isn't claiming tasks.

- known issues:
  - `_removeWorker()` doesn't properly detach IPs from NICs before deletion which can cause an error

- todo:
  - [x] `registerWorker()` doesn't verify attested data / worker identity
  - [x] scan for and clean up leftover resources (NICs, IPs, and Disks are not deleted when instances are)
    - when workers are removed we delete relevant resources
  - [ ] verify that instances are running as Spot (when they are requested as such)


djmitche:
(I've flagged myself for review so it goes on my list.. it's not blocking and if owlish/bstack approve, go ahead and land)

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,379 @@
...
+          workerConfig: cfg.workerConfig || {},
+        })).toString('base64');
+
+        virtualMachine = await this._enqueue('query', () => this.computeClient.virtualMachines.createOrUpdate(resourceGroupName, virtualMachineName, {
+          ...cfg,
+          osProfile: {
+            adminUsername: 'worker',
+            ...cfg.osProfile,
+            // throw this away
+            adminPassword: slugid.nice(),
```
imbstack: Let's do 2-4 slugids just to make me feel warm and fuzzy

imbstack: May as well give a bit more context for why we're throwing it away too.

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,379 @@
...
+          providerId: this.providerId,
+          workerGroup: this.providerId,
+          rootUrl: this.rootUrl,
+          workerConfig: cfg.workerConfig || {},
+        })).toString('base64');
+
+        virtualMachine = await this._enqueue('query', () => this.computeClient.virtualMachines.createOrUpdate(resourceGroupName, virtualMachineName, {
+          ...cfg,
+          osProfile: {
+            adminUsername: 'worker',
```
imbstack: I think @grenade was thinking this should be `Administrator` or something like that?

imbstack: Let's move it below the `...cfg.osProfile` as well so that people can't overwrite it.

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,379 @@
...
+            networkInterfaces: [
+              {
+                id: networkInterface.id,
+                primary: true,
+              },
+            ],
+          },
+          priority: cfg.priority || "Spot",
+          evictionPolicy: cfg.evictionPolicy || "Deallocate",
+          billingProfile: {
```
imbstack: Will it yell at us if this is specified on non-spot instances?

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,379 @@
...
+            ...cfg.networkProfile,
+            networkInterfaces: [
+              {
+                id: networkInterface.id,
+                primary: true,
+              },
+            ],
+          },
+          priority: cfg.priority || "Spot",
+          evictionPolicy: cfg.evictionPolicy || "Deallocate",
```
imbstack: Let's get rid of the defaults.

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,379 @@
...
+          tags: {
+            ...cfg.tags || {},
+            'created-by': `taskcluster-wm-${this.providerId}`.replace(/[^a-zA-Z0-9-]/g, '-'),
+            'managed-by': 'taskcluster',
+            'worker-pool-id': workerPoolId.replace(/[^a-zA-Z0-9-]/g, '-').toLowerCase(),
+            'owner': workerPool.owner.replace(/[^a-zA-Z0-9-]/g, '-').toLowerCase(),
+          },
+        }));
+      } catch (err) {
+        await this._removeWorker({...providerData});
```
imbstack: Let's add a comment that this is here to clean up the resources like nics, etc and that if they don't exist yet it is fine.

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,379 @@
...
+      w.lastModified = new Date();
+      w.state = this.Worker.states.RUNNING;
+    });
+    // assume for the moment that workers self-terminate before 96 hours
+    return {expires: taskcluster.fromNow('96 hours')};
+  }
+
+  async scanPrepare() {
+    this.seen = {};
+    this.errors = {};
```
imbstack: I think that `this.errors` is functionally unused in this provider. Let's rip it out from here and its other uses below. It is just used in the google provider to get reports from operations.

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,379 @@
...
+        // If the worker will be expired soon but it still exists,
+        // update it to stick around a while longer. If this doesn't happen,
+        // long-lived instances become orphaned from the provider. We don't update
+        // this on every loop just to avoid the extra work when not needed
+        if (worker.expires < taskcluster.fromNow('1 day')) {
+          await worker.modify(w => {
+            w.expires = taskcluster.fromNow('1 week');
+          });
+        }
+      } else if (['Failed', 'Canceled', 'Deleting', 'Deallocating', 'Stopped', 'Deallocated'].includes(provisioningState)) {
```
imbstack: Are there more states than this or the possibility of them adding more? We could maybe at least report an error on unknown states

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,379 @@
...
+          providerId: this.providerId,
+          workerId: worker.workerId,
+        });
+        state = states.STOPPED;
+      }
+    } catch (err) {
+      if (err.code !== 'ResourceNotFound') {
+        throw err;
+      }
+      this.monitor.log.workerStopped({
```
imbstack: We probably want to `removeWorker` in this case too?

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,379 @@
...
+              privateIPAllocationMethod: 'Dynamic',
+              subnet: {
+                id: cfg.subnetId,
+              },
+              publicIPAddress: {
+                id: ipAddress.id,
+              },
+            },
+          ],
+        }));
```
djmitche: It might make sense to run these two operations in parallel with `Promise.all(..)`, just to speed up provisioning a bit.

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,379 @@
...
+      try {
+        await promise;
+      } catch (err) {
+        if (err.code === 404) {
+          return; // Nothing to do, it is already gone
+        }
+        throw err;
+      }
+    };
+    ignoreNotFound(this._enqueue('query', () => this.computeClient.virtualMachines.deleteMethod(
```
djmitche: missing await (here and below)

services/worker-manager/test/fixtures/schemas/sample-azure-config.json
```diff
@@ -0,0 +1,27 @@
...
+    {
+      "capacityPerInstance": 1,
+      "location": "westus",
+      "subnetId": "/subscriptions/{subscription-id}/resourceGroups/{resource-group-name}/providers/Microsoft.Network/virtualNetworks/{vnet-name}/subnets/{subnet-name}",
+      "hardwareProfile": {
+        "vmSize": "Basic_A1"
+      },
+      "storageProfile": {
+        "imageReference": {
+          "id": "/subscriptions/{subscription-id}/resourceGroups/{resource-group-name}/providers/Microsoft.Compute/images/{image-name}"
```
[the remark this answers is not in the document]
milescrabill: Those were just my notation for users having to replace those values. I'm switched them to be "your-subscription-id" to avoid confusion.

services/worker-manager/src/providers/azure.js
milescrabill: We have to have the public IP address created before we can create the NIC, unfortunately

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,447 @@
...
+    const error = () => new ApiError('Signature validation error');
+
+    // workerIdentityProof is a signed message
+
+    // We need to check that:
+    // 1. The embedded document was signed with the private key corresponding to the
+    //    embedded public key
+    // 2. The embedded public key has a proper certificate chain back to a trusted CA
+    // 3. The embedded message contains the vmId that matches the worker making the
+    //    worker making the request
```
ajvb: looks like `worker making the` is repeated twice.

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,447 @@
...
+      workerPoolId: workerPool.workerPoolId,
+      providerId: this.providerId,
+      workerId: worker.workerId,
+    });
+    await worker.modify(w => {
+      w.lastModified = new Date();
+      w.state = this.Worker.states.RUNNING;
+    });
+
+    // assume for the moment that workers self-terminate before 96 hours
```
ajvb: What does this assumption imply exactly? If the instance doesn't self-terminate within 96 hours, what does that mean?

services/worker-manager/test/provider_azure_test.js
```diff
@@ -0,0 +1,301 @@
...
+    await provider.scanPrepare();
+    await provider.checkWorker({worker});
+    await provider.scanCleanup();
+    assert(!fakeAzure.deleteVMStub.called);
+    assert(!fakeAzure.deleteDiskStub.called);
+    assert(!fakeAzure.deleteIPStub.called);
+    assert(!fakeAzure.deleteNICStub.called);
+  });
+
+  suite('registerWorker', function() {
```
ajvb: Similar to the AWS tests (https://github.com/taskcluster/taskcluster/blob/master/services/worker-manager/test/provider_aws_test.js#L257) it would be nice to see some failure tests with bad azure signatures, incorrect vmId, and maybe bad encoding and bad chain of trust.

If possible, should aim for near 100% test coverage of this function.

ui/docs/reference/core/worker-manager/azure.mdx
```diff
@@ -0,0 +1,54 @@
...
+## Tags
+
+The provider starts workers with the following labels.
+
+ * `worker-pool-id` - the worker's workerPoolId
+ * `managed-by` - `taskcluster`
+ * `created-by` - `taskcluster-wm-<providerId>`
+ * `owner` - the worker pool's owner
+
+Note that values are lower-cased and disallowed characters are replaced with `-`, to comply with with [Azure requirements](https://docs.microsoft.com/en-us/azure/architecture/best-practices/resource-naming).
```
helfi92: We could probably use the `Warning` component:

```
import SchemaTable from 'taskcluster-ui/components/SchemaTable';
import Warning from 'taskcluster-ui/views/Documentation/components/Warning';

// ...

<Warning>
  ....
</Warning>
```

services/worker-manager/src/providers/azure.js
milescrabill: We're giving out temporary credentials, so the credentials will expire and be unusable after 96 hours. This makes sense for spot instances in GCP and AWS because they are expected to be short lived... but I've realized that Azure does not guarantee any kind of shutdown timing, so this could be an issue.

services/worker-manager/schemas/v1/register-worker-request.yml
```diff
@@ -52,6 +52,16 @@ properties:
               curl http://169.254.169.254/latest/dynamic/instance-identity/signature on the instance
         additionalProperties: false
         required: [document, signature]
+      - title: azure provider type
+        type: object
+        properties:
+          document:
+            title: Document
+            description: |
```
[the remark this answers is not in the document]
imbstack: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1608258 for the aws portion

services/worker-manager/schemas/v1/config-azure.yml
```diff
@@ -0,0 +1,157 @@
...
+            maxPrice:
+              title: Max Price
+              type: number
+              description: the maximum price to pay for a spot virtual machine in US Dollars
+        hardwareProfile:
+          title: Hardware Profile
+          type: object
+          description: https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/createorupdate#hardwareprofile
+          additionalProperties: true
+          required: [vmSize]
```
imbstack: let's just do this the

```yaml
foo: 
  - bar
```
way to match up with how we do it everywhere else even if it uses more lines

services/worker-manager/src/providers/azure-ca-certs/README.md
```diff
@@ -0,0 +1,7 @@
+### About
```
imbstack: Let's add a link to the bug about doing this in a long-term way into here.

services/worker-manager/test/api_test.js
```diff
@@ -17,13 +17,14 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
 
   test('list providers', async function() {
     const {providers} = await helper.workerManager.listProviders();
-    assert.deepStrictEqual(providers.sort(), [
+    assert.deepStrictEqual(providers, [
```
imbstack: Why no sort? If the default ordering changes in a different node version or something this test will fail I think?

ui/docs/manual/deploying/workers.mdx
```diff
@@ -164,3 +164,100 @@ If this service-linked role is not set up properly, worker creation will fail wi
...
+    "apiRateLimits": {
+      "get": {"interval": 1000, "intervalCap": 20000},
+      "query": {"interval": 1000, "intervalCap": 40000},
+      "list": {"interval": 1000, "intervalCap": 20000},
+    },
+  }
+}
+```
+
+`clientId` and `secret` are properties of Azure application secrets. See [here](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal)
```
imbstack: Thanks for adding this

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,453 @@
...
+        }
+        // If we don't want to do anything special here, just throw and let the
+        // calling code figure out what to do
+        throw err;
+      },
+    });
+    this._enqueue = cloud.enqueue.bind(cloud);
+
+    // load microsoft intermediate certs from disk
+    // TODO: we should download the intermediate certs,
```
imbstack: add reference to bug here

imbstack: Maybe a note here as well as to why there are 1/2/4/5 and no 3? Otherwise it looks like a typo

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,453 @@
...
+          workerGroup: this.providerId,
+          rootUrl: this.rootUrl,
+          workerConfig: cfg.workerConfig || {},
+        })).toString('base64');
+
+        virtualMachine = await this._enqueue('query', () => this.computeClient.virtualMachines.createOrUpdate(resourceGroupName, virtualMachineName, {
+          ...cfg,
+          osProfile: {
+            ...cfg.osProfile,
+            adminUsername: 'worker',
```
imbstack: I think I remember @grenade wanting this to be `admin`?

services/worker-manager/src/providers/azure.js
imbstack: It isn't a security issue however, just a usability one. The creds do expire and the worker is just useless at that point

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,453 @@
...
+    this.monitor.log.scanSeen({providerId: this.providerId, seen: this.seen});
+    await Promise.all(Object.entries(this.seen).map(async ([workerPoolId, seen]) => {
+      const workerPool = await this.WorkerPool.load({
+        workerPoolId,
+      }, true);
+
+      if (!workerPool) {
+        return; // In this case, the workertype has been deleted so we can just move on
+      }
+    }));
```
imbstack: Is this copy-pasta from somewhere? It seems like an unnecessary thing to do.

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,453 @@
...
+  }
+
+  async _removeWorker({resourceGroupName, location, vm, ip, nic, disk}) {
```
imbstack: Are any of these able to be done in parallel?

services/worker-manager/test/api_test.js
milescrabill: The sort doesn't work because the items are objects and they aren't being sorted by key - so order matters either way here

services/worker-manager/src/providers/azure.js
milescrabill: Unfortunately we can't use (quite a few) values for `adminUsername`. Grenade recommends setting this to a throwaway value [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1604650#c4) - I'll change to a `slugid.nice()` call. It was convenient for testing to have this be a known value for password reset purposes but it's still easy to find.

services/worker-manager/src/providers/azure.js
milescrabill: The issue that I've run into is that they're used by each other, i.e. an IP cannot be deleted while it is attached to a NIC, so we delete the NIC first. I'll group the other three.

services/worker-manager/src/providers/azure.js
milescrabill: Yeah, this is pasta from the google provider. I wasn't totally sure of the purpose there, but we send a monitor message for `scanSeen` there, this replicates that

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,456 @@
...
+    } catch (err) {
+      this.monitor.warning('Error verifying PKCS#7 message signature');
+      throw error();
+    }
+
+    // verify that the embedded certificate has proper chain of trust
+    try {
+      assert(forge.pki.verifyCertificateChain(this.caStore, [crt]), true);
+    } catch (err) {
+      this.monitor.warning('Error verifying certificate chain', {error: err.toString()});
```
imbstack: I'm not sure we need to do all of this `monitor.warning` stuff since we throw the error anyway right away and it will have a reference to where the error happened.

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,447 @@
...
+    }
+
+    let content, crt, pem, sig;
+    // get message content, signing certificate, and signature
+    try {
+      // in testing, message.content is empty, so we access the raw ASN1 structure
+      content = message.rawCapture.content.value[0].value;
+      // convert to pem for convenience
+      // there may be more than one certificate, we only look at the first one
+      crt = message.certificates[0];
```
ajvb: Are the remaining certs in here the intermediate(s)? Or do the intermediates need to be fetched as in the example from the docs? (https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#validating-that-the-vm-is-running-in-azure)

services/worker-manager/src/providers/azure.js
```diff
@@ -0,0 +1,447 @@
...
+      verifier.update(Buffer.from(content));
+      assert(verifier.verify(pem, sig, 'binary'), true);
+    } catch (err) {
+      this.monitor.warning('Error verifying PKCS#7 message signature');
+      throw error();
+    }
+
+    // verify that the embedded certificate has proper chain of trust
+    try {
+      assert(forge.pki.verifyCertificateChain(this.caStore, [crt]), true);
```
ajvb: can this pass in `message.certificates` instead of `[crt]`?

services/worker-manager/src/providers/azure.js
ajvb: Ok totally, sounds good.

services/worker-manager/src/providers/azure.js
imbstack: I was wrong about this

services/worker-manager/src/providers/azure.js
imbstack: I'll give it a try now. I'm not sure what the other certs are or if they'll pass this.

services/worker-manager/src/providers/azure.js
djmitche: I think MS only provides the one certificate; the format would allow several, though.

services/worker-manager/src/providers/azure.js
djmitche: I think that would introduce a vulnerability where an attacker could produce a document with `certificates[0]` set to a self-signed certificate that's used to sign the document, and `certificates[1]` set to a certificate signed by the MS intermediaries (for example one copied from a legitimate document).  Then this code would (I think, I mean, node-forge is undocumented, yay cryptographers and their inability to relate to humans) find certificates[1] to be valid and succeed.

I think we want to assume that there's only one certificate in the document.  It might even be worth *requiring* that and failing verification if there are >1?

imbstack: I can confirm the `message.certificates` version works in my dev env. I think deciding which way we want to go on this is the only major remaining blocker to landing this.

The other is more tests of this verification

services/worker-manager/src/providers/azure.js
milescrabill: The intermediates aren't bundled, if there were more certs here they would in theory be additional signing certs for the message. The intermediate cert URL is available via the authorityInfoAccess location extension in the cert.

+1 to what Dustin said, they sign with one cert from what I've seen

services/worker-manager/src/providers/azure.js
imbstack: assertion added in latest code

services/worker-manager/test/provider_azure_test.js
```diff
@@ -0,0 +1,377 @@
...
+      const worker = await helper.Worker.create({
+        ...defaultWorker,
+      });
+      // this file is a version of `azure_signature_good` where vmId has been edited in the message
+      const document = fs.readFileSync(path.resolve(__dirname, 'fixtures/azure_message_bad')).toString();
+      const workerIdentityProof = {document};
+      await assert.rejects(() =>
+        provider.registerWorker({workerPool, worker, workerIdentityProof}),
+      /Signature validation error/);
+      assert(monitorManager.messages[0].Fields.error.includes('Only 8, 16, 24, or 32 bits supported: 264'));
```
imbstack: This and the next test are passing but seem to fail with a failure that is not testing the thing we intended? I need to figure out  more about `'Only 8, 16, 24, or 32 bits supported: 264'` before we ship this I think

imbstack: I think this is because we just arbitarily futzed with a valid message to get an invalid one. We probably have to actually encode a real message with an invalid cert to get this test to be acccurate. Does that seem true, @milescrabill?

I'm pretty sure this is _working_ right now but just that the test is not testing it.

PR #2199, opened 2019-12-20
[UI] Prevent form resetting value when creating a new client

### LINKED ISSUE #2187: [UI] Form resetting when creating a new client
**Describe the bug**
Creating a new client while missing scopes will show an error but then clear the form. We should not clear the form when there is an error.

**To Reproduce**

1. Go to http://localhost:5080/auth/clients/create and make sure you are not logged in
2. Fill the form with any values
2. Click on the save button icon on the lower right
3. Close the error panel

**Expected behavior**
The form values are still there.

**Actual behavior**
The form values have been cleared out

**Taskcluster version**
24.1.1

**Additional context**
The reason why this is happening is because the key prop is changing when an error happens:

https://github.com/taskcluster/taskcluster/blob/caec22bccd3481e7d0f02c091ce1bc8eba3263a7/ui/src/views/Clients/ViewClient/index.jsx#L361

When a `key` prop changes in React, it re-creates a whole component, hence losing all of the internal state. One way to fix this issue is to make sure that `JSON.stringify(initialClient)` doesn't change on subsequent renders. We can probably do it with https://github.com/caiogondim/fast-memoize.js#custom-serializer (we use this in a couple of places in the repo).

ss77995ss:
### Related Ticket
* Fixes #2187 

### Description
* When we click `save` button, it would trigger rerender multiple times since `GraphQL` would update `error` and `loading` props. This makes `initialClient`'s `expires` update. Therefore, when `key` props changed, the component `ClientForm` would create a new instance and reset form's value.
* To prevent the resetting, use `memoize` function from `fast-memoize` to cache function `getClientFormKey` return value and use `serializer` to set the `cache key`.

### DIFF
--- a/changelog/issue-2187.md
+++ b/changelog/issue-2187.md
@@ -1,0 +1,5 @@
+level: silent
+reference: issue 2187
+---
+
+Prevent resetting value when creating a new client
--- a/ui/src/views/Clients/ViewClient/index.jsx
+++ b/ui/src/views/Clients/ViewClient/index.jsx
@@ -103,10 +103,9 @@ export default class ViewClient extends Component {
 
   getClientFormKey = memoize(initialClient => JSON.stringify(initialClient), {
     serializer: initialClient => {
-      // Use expires as cache key
-      const serializeKey = JSON.stringify(omit(['expires'], initialClient));
-
-      return serializeKey;
+      // expires changes on every render so it's best to keep
+      // it out of the caching key
+      return JSON.stringify(omit(['expires'], initialClient));
     },
   });
 
--- a/ui/src/views/Clients/ViewClient/index.jsx
+++ b/ui/src/views/Clients/ViewClient/index.jsx
@@ -2,6 +2,8 @@ import { hot } from 'react-hot-loader';
 import React, { Component, Fragment } from 'react';
 import { graphql, withApollo } from 'react-apollo';
 import { parse } from 'qs';
+import memoize from 'fast-memoize';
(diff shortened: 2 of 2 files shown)


changelog/issue-2187.md
```diff
@@ -0,0 +1,5 @@
+level: minor
```
helfi92:
```suggestion
level: silent
```

ui/src/views/Clients/ViewClient/index.jsx
```diff
@@ -99,6 +100,15 @@ export default class ViewClient extends Component {
     },
   };
 
+  getClientFormKey = memoize(initialClient => JSON.stringify(initialClient), {
+    serializer: (initialClient, query) => {
+      // Use expires as cache key
+      const serializeKey = query ? query.expires : 'no-expires';
```
helfi92: Is there a way to use the initial client as the serialize key?

ss77995ss: Maybe use `initialClient` but omit `expires`? Since `expires` will be different every time component rerender

helfi92: Yep, that's what I had in mind.

ss77995ss: OK, I see! Let me update it

ss77995ss: Updated!

PR #2209, opened 2019-12-23
[UI] Revamp the Warning component in docs

helfi92:
#### Dark Theme

Before             |  After
:-------------------------:|:-------------------------:
Screen Shot 2019-12-23 at 12 55 07 PM | Screen Shot 2019-12-23 at 12 56 01 PM

#### Light Theme

Before             |  After
:-------------------------:|:-------------------------:
Screen Shot 2019-12-23 at 12 55 23 PM  |  Screen Shot 2019-12-23 at 12 55 48 PM

### DIFF
--- a/changelog/CVgKPSmcRsqCpRkeHIDwoQ.md
+++ b/changelog/CVgKPSmcRsqCpRkeHIDwoQ.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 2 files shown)


helfi92:
Oh, material-ui just introduced the Alert component in [v4.8.3](https://github.com/mui-org/material-ui/releases/tag/v4.8.3). Maybe we can start using that instead.

Once we upgrade material-ui to v4.8.3, we could even aim to deprecate usage `@mozilla-frontend-infra/components/ErrorPanel` and rely on the Alert component instead since it's basically the same thing.

djmitche:
@helfi92 this looks good, though -- do you want to land it in the interim?

PR #2217, opened 2019-12-26
Remove some TODOs

djmitche:
This has some relatively low-impact changes to remove TODO's from the monorepo.

There are still a few:

```
clients/client-web/.neutrinorc.js:              // TODO: Override using tap() so as to not have to repeat the
clients/client-web/.neutrinorc.js:        // TODO: Remove this when fixed in @neutrinojs/library.
infrastructure/tooling/src/dev/aws.js:  // TODO: Add both blob buckets
infrastructure/tooling/src/dev/aws.js:  // TODO: Also set up auth/notify aws stuff
infrastructure/tooling/src/dev/rabbit.js:        if (!['hooks', 'auth'].includes(service)) { // These services hardcode namespace TODO: fix?
infrastructure/tooling/src/dev/taskcluster.js:  // TODO: Figure out what any of these should be set to
infrastructure/tooling/src/dev/taskcluster.js:  //TODO: These github things can/should be questions in this setup thing
infrastructure/tooling/src/dev/taskcluster.js:  // TODO: These can/should be generated by dev:init aws setup
infrastructure/tooling/src/dev/taskcluster.js:  // TODO: This eventually should just build these from rootUrl itself probably
infrastructure/tooling/src/generate/generators/k8s.js:          paths: conf['paths'] || [`/api/${name}/*`], // TODO: This version of config is only for gcp ingress :(
infrastructure/tooling/src/generate/generators/k8s.js:        // TODO: iirc google doesn't set the headers that we need to trust proxy so we don't set this, let's fix it
infrastructure/tooling/src/generate/generators/k8s.js:        // TODO: In config.ymls somehow mark fields as "required" or "optional" and then assert
infrastructure/tooling/src/generate/generators/k8s.js:            cpu: '50m', // TODO: revisit these defaults
infrastructure/tooling/src/generate/generators/k8s.js:            cpu: '50m', // TODO: revisit these defaults
services/web-server/src/graphql/Provisioners.graphql:# TODO: Provisioner mutations
ui/src/components/ProvisionerDetailsTable/index.jsx:  // TODO: Action not working.
ui/src/components/TaskActionForm/index.jsx:    // TODO: Replace with taskAction
ui/src/views/Provisioners/ViewWorker/index.jsx:    // TODO: Action not working.
ui/src/views/Provisioners/ViewWorkers/index.jsx:  // TODO: Action not working
```

That breaks down to:
 * Neutrino config for client-web
 * `yarn dev`
 * Worker Actions support in the UI.

### DIFF
--- a/services/auth/test/stories_test.js
+++ b/services/auth/test/stories_test.js
@@ -77,6 +77,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure', 'gcp'], function(mock, s
             'auth:reset-access-token:test-users/charlene/*',
             'assume:test-role:role1',
             'assume:test-role:role2',
+            'scope3a',
           ],
         }),
       });
@@ -160,8 +161,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure', 'gcp'], function(mock, s
       });
     });
 
-    // TODO: bug 1242473
-    test.skip('charlene replaces role3 with one of its constituent scopes', async () => {
+    test('charlene replaces role3 with one of its constituent scopes', async () => {
       await charlene.updateClient('test-users/charlene/travis-tests', {
         description: 'Permacred created by test',
         expires: taskcluster.fromNow('3 hours'),
--- a/changelog/P-IpBKPgQyCIbt7HWVq6gA.md
+++ b/changelog/P-IpBKPgQyCIbt7HWVq6gA.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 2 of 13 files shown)


services/auth/test/stories_test.js
```diff
@@ -160,8 +161,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure', 'gcp'], function(mock, s
       });
     });
 
-    // TODO: bug 1242473
```
imbstack: I wonder if an enterprising person could cook up a test that checks the status of `TODO: bug XXXX` type comments and throw if they are closed or something. It would only run on master or such.

djmitche: If we were going to be enterprising, I'd kind of prefer to just get rid of them, and add a meta-check that no further TODO's are added.  I'm not terribly motivated beyond the effort that went into this PR, though.

PR #2220, opened 2019-12-28
Make worker manager form invalid if owner not provided

### LINKED ISSUE #2201: [UI] Worker manager required owner field
STR:

1. Navigate to http://localhost:5080/worker-manager/create
2. Fill out all the fields except the owner field

Screen Shot 2019-12-20 at 2 00 03 PM

**Actual:** The save button (diskette icon) is enabled.
**Expected:** The save button is disabled since the owner field is required but empty

Possibly also relates to https://github.com/taskcluster/taskcluster/issues/2200.

theprover97:
Save WM button is enabled only when all fields are filled, including owner field.

Fixes https://github.com/taskcluster/taskcluster/issues/2201.

### DIFF
--- a/ui/src/components/WMWorkerPoolEditor/index.jsx
+++ b/ui/src/components/WMWorkerPoolEditor/index.jsx
@@ -133,6 +133,7 @@ export default class WMWorkerPoolEditor extends Component {
       },
       owner: {
         error: null,
+        check: false,
         message: null,
       },
     },
@@ -151,6 +152,7 @@ export default class WMWorkerPoolEditor extends Component {
           ...this.state.validation,
           owner: {
             error: !validity.valid,
+            check: true,
             message: validationMessage,
           },
         },
@@ -186,7 +188,7 @@ export default class WMWorkerPoolEditor extends Component {
     return (
       !this.state.invalidProviderConfig &&
       !Object.values(validation).some(({ error }) => Boolean(error)) &&
-      providerId !== ''
+      providerId !== '' && this.state.validation.owner.check
     );
   }
 
--- a/ui/src/components/WMWorkerPoolEditor/index.jsx
+++ b/ui/src/components/WMWorkerPoolEditor/index.jsx
@@ -133,7 +133,6 @@ export default class WMWorkerPoolEditor extends Component {
       },
       owner: {
         error: null,
-        check: false,
         message: null,
       },
     },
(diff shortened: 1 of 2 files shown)


ui/src/components/WMWorkerPoolEditor/index.jsx
```diff
@@ -186,7 +188,7 @@ export default class WMWorkerPoolEditor extends Component {
     return (
       !this.state.invalidProviderConfig &&
       !Object.values(validation).some(({ error }) => Boolean(error)) &&
-      providerId !== ''
+      providerId !== '' && this.state.validation.owner.check
```
helfi92: We can probably fix this without having to introduce a new state `check`. Here's one way:

```diff
diff --git a/ui/src/components/WMWorkerPoolEditor/index.jsx b/ui/src/components/WMWorkerPoolEditor/index.jsx
index 26f865850..f7d9485d5 100644
--- a/ui/src/components/WMWorkerPoolEditor/index.jsx
+++ b/ui/src/components/WMWorkerPoolEditor/index.jsx
@@ -182,14 +182,16 @@ export default class WMWorkerPoolEditor extends Component {
 
   isValid() {
     const {
-      workerPool: { providerId },
+      workerPool: { workerPoolId1, workerPoolId2, owner },
       validation,
     } = this.state;
 
     return (
       !this.state.invalidProviderConfig &&
       !Object.values(validation).some(({ error }) => Boolean(error)) &&
-      providerId !== ''
+      workerPoolId1 !== '' &&
+      workerPoolId2 !== '' &&
+      owner !== ''
     );
   }
```

helfi92:
Is it possible to rename the pull request to something more meaningful. Right now it just says `Fixed #2201` which doesn't tell us much about the content of this PR. What do you think with "Make worker manager form invalid if owner not provided"?

ui/src/components/WMWorkerPoolEditor/index.jsx
theprover97: I have resolved the issue . Thanks for the feedback!

theprover97:
Is it necessary to write changelog even though the PR introduces no new features? As here we are only correcting an element's behaviour.

djmitche:
That's what the `silent` level is for.  `yarn changelog --silent` should do the trick.

PR #2228, opened 2019-12-31
Bug 1598649 - split into 'yarn release' and 'yarn release:publish'

djmitche:

Next step will be to perform the `yarn release:publish` portion in CI on pushes to tags.

### DIFF
--- a/infrastructure/tooling/src/utils/command.js
+++ b/infrastructure/tooling/src/utils/command.js
@@ -42,7 +42,7 @@ exports.execCommand = async ({
       } else {
         output = '...\n' + chunk.toString();
       }
-      callback(null, output);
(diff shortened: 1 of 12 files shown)


infrastructure/tooling/src/utils/command.js
```diff
@@ -42,7 +42,7 @@ exports.execCommand = async ({
       } else {
         output = '...\n' + chunk.toString();
       }
-      callback(null, output);
+      callback(null, chunk);
```
imbstack: huh, how did this even work before?

djmitche: It just output a lot of `...` to the stream, but if nothing went wrong you didn't really notice.

PR #2230, opened 2020-01-02
Bug 1598649 - run 'yarn release:publish' on tags

djmitche:

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -288,6 +288,63 @@ tasks:
                 description: ${job.name}
                 owner: ${owner}
                 source: ${repo.url}
+          releaseTasks:
+          - taskId: {$eval: 'as_slugid("release")'}
+            provisionerId: proj-taskcluster
+            workerType: release
+            created: {$fromNow: ''}
+            deadline: {$fromNow: '3 hours'}
+            extra:
+              notify:
+                email:
+                  subject: 'Taskcluster Release Status'
+            routes:
+            - notify.email.taskcluster-notifications@mozilla.com.on-any
+            - notify.irc-channel.#taskcluster-bots.on-any
+            scopes:
+            - secrets:get:project/taskcluster/release
+            payload:
+              features:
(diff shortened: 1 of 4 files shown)


djmitche:
This *mostly* ran, with some hacks, in [this task](https://community-tc.services.mozilla.com/tasks/STzeTrA6RSWhvQeMpiWf_Q/runs/0/logs/live/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FSTzeTrA6RSWhvQeMpiWf_Q%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log).  I didn't give it the "real" release credentials, because I didn't want to do a real release!

I'm reasonably confident that, given real credentials, it would work -- we've done at least one release supplying those values manually via env vars.

I'd like to land this PR, then do a "real" patch release immediately, and then fix up whatever fails and try again.  Hopefully only a few (dozen?) iterations of that pattern will be necessary.

.taskcluster.yml
```diff
@@ -288,10 +288,73 @@ tasks:
...
+                    go version
+
+                    # set up gox
+                    go get github.com/mitchellh/gox
+
+                    # set up Node, using nvm (generic-worker does not support .xz, so we can't use mounts)
+                    curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.2/install.sh | bash
+                    . ~/.nvm/nvm.sh
+                    nvm install ${node}
+                    npm install --global yarn
```
imbstack: I can't find it right now but I feel like I've seen lots of bug reports in yarn over the years based on being installed with npm rather than their recommended way. I don't think we need to be _particularly_ worried about the cautionary note in [their docs](https://yarnpkg.com/en/docs/install#alternatives-stable), but we may as well do it their recommended way.

djmitche: trying it now..

PR #2233, opened 2020-01-03
[tc-lib-entities] entity create, delete, load, update, scan

helfi92:


libraries/entities/src/index.js
```diff
@@ -23,7 +23,26 @@ class Entity {
...
+    const documentId = this.calculateId(properties);
+
+    return this.db.procs[`${this.tableName}_update`](documentId, properties, 1);
+  }
+
+  load(properties) {
+    const documentId = this.calculateId(properties);
+
+    return this.db.procs[`${this.tableName}_load`](documentId);
   }
```
helfi92: Questions:

1. If a user tries to update an entry that doesn't exist, where do we throw the error? I wasn't sure if we should handle errors here or inside the stored procedure in `0001.yml`.
2. Should we rely on the implementation of the stored procedure to update the etag when an entry is modified (current behavior) or is this something we should take care of inside the entity class?

djmitche: 1. The errors will need some specific JS behavior (`err.code = "ResourceNotFound"`).  So the stored procedure should return enough information for the JS to construct an appropriate error.
2. It seems like it would be easiest to do so in the stored procedure, since you get to check within a transaction.  We'll probably need to use some locking or the like to ensure serializability.

libraries/entities/src/index.js
```diff
@@ -23,7 +26,54 @@ class Entity {
...
+      filter,
+      limit = 1000,
+      page = 1,
+    } = options;
+
+    if (filter) {
+      assert(typeof filter === 'function', 'filter should be a function');
+    }
+
+    const result = await this.db._withClient(READ, client => {
```
djmitche: We'll need to avoid using raw clients for this library, and stick to stored procedures.  Otherwise, the code using the raw clients will stop working when we migrate away from the entities tables, as the tables will no longer exist. So, scanning and querying will need to be done in a stored procedure.

libraries/entities/src/index.js
```diff
@@ -23,7 +26,54 @@ class Entity {
...
+      return client.query(`select * from ${this.tableName}`);
+    });
+
+    const filteredRows = filter ? result.rows.filter(filter) : result.rows;
```
djmitche: We probably can't get away with doing client-side filtering -- that will have the database sending the whole table to us every time.  We'll need to find a way to translate at least the scan conditions related to the row and partition keys into a `WHERE` clause.

libraries/entities/src/index.js
helfi92: You make a good point. If anything in the db changes, queries made with the raw client may break.

> the raw clients will stop working when we migrate away from the entities tables, as the tables will no longer exist.

In that case, would it make sense to have our stored procedures not have as a prefix the table name? Right now, `create`, `delete`, `update`, and `load` all do ``` return this.db.procs[`${this.tableName}_xxxx`](...); ```

djmitche: Yeah, maybe, if they can take the table name as a parameter.  I can imagine a few different approaches.  We can talk about that this afternoon.

PR #2250, opened 2020-01-09
Bug 1608176 - allow `client.SignedURL(..)` to take a full URL again

djmitche:
Bugzilla Bug: [1608176](https://bugzilla.mozilla.org/show_bug.cgi?id=1608176)

This is functionality we had long, long ago, and on which [taskcluster-proxy depended](https://github.com/taskcluster/taskcluster-proxy/pull/46).  It's easy enough to re-introduce it!

I also added some tests for bewit support.  Happily, they passed :)

### DIFF
--- a/clients/client-go/codegenerator/model/api.go
+++ b/clients/client-go/codegenerator/model/api.go
@@ -359,8 +359,10 @@ func (entry *APIEntry) generateDirectMethod(apiName string) string {
 
 func (entry *APIEntry) generateSignedURLMethod(apiName string) string {
 	// if no required scopes, no reason to provide a signed url
-	// method, since no auth is required, so unsigned url already works
-	if entry.Scopes.Type == "" {
+	// method, since no auth is required, so unsigned url already works,
+	// except for TestAuthenticateGet, which can be usefully used to test
+	// with signed URLs
+	if entry.Scopes.Type == "" && entry.MethodName != "TestAuthenticateGet" {
 		return ""
 	}
 	comment := "// Returns a signed URL for " + entry.MethodName + ", valid for the specified duration.\n"
--- a/clients/client-go/creds_test.go
+++ b/clients/client-go/creds_test.go
@@ -2,6 +2,7 @@ package tcclient_test
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"reflect"
 	"testing"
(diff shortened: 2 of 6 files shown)


clients/client-go/codegenerator/model/api.go
```diff
@@ -359,8 +359,10 @@ func (entry *APIEntry) generateDirectMethod(apiName string) string {
 
 func (entry *APIEntry) generateSignedURLMethod(apiName string) string {
 	// if no required scopes, no reason to provide a signed url
-	// method, since no auth is required, so unsigned url already works
-	if entry.Scopes.Type == "" {
+	// method, since no auth is required, so unsigned url already works,
+	// except for TestAuthenticateGet, which can be usefully used to test
+	// with signed URLs
+	if entry.Scopes.Type == "" && entry.MethodName != "TestAuthenticateGet" {
```
petemoore: I guess this is ok for now, it feels a little hacky though. I'm not sure of an alternative simple solution that would achieve the same, without requiring more fields in the API definition, which also feels like overkill.

I would still advocate for our official CI tests to use real credentials, and then we wouldn't need to introduce this workaround.

clients/client-go/creds_test.go
```diff
@@ -122,6 +124,114 @@ func Test_NamedTempCred(t *testing.T) {
 		"jimmy", []string{"scope:1", "scope:2"})
 }
 
+func Test_PermaCred_Bewit(t *testing.T) {
+	rootURL := os.Getenv("TASKCLUSTER_ROOT_URL")
+	if rootURL == "" {
+		t.Skip("Cannot run test, TASKCLUSTER_ROOT_URL is not set to a non-empty string")
+	}
```
petemoore: I think it would be better to require an opt-out environment variable to indicate that the test shouldn't run, rather than an opt-in. The danger of opt-in is that a user doesn't know that `TASKCLUSTER_ROOT_URL` needs to be set, or worse, in our CI, for some reason, `TASKCLUSTER_ROOT_URL` doesn't get set when the task runs due to an accidental bug, and we don't notice that certain unit tests no longer run. On the other hand, if we require env var `SKIP_INTEGRATION_TESTS` to be set in order to skip this test and others that require a taskcluster deployment to talk to, then we can be sure that the CI will never skip tests we didn't intend it to, and users know that they are not running all tests because they had to explicitly enable the option.

clients/client-go/http.go
```diff
@@ -259,15 +259,21 @@ func (client *Client) APICall(payload interface{}, method, route string, result
...
-// SignedURL creates a signed URL using the given Client, where route is the
-// url path relative to the RootURL stored in the Client, query is the set of
-// query string parameters, if any, and duration is the amount of time that the
-// signed URL should remain valid for.
+// SignedURL creates a signed URL using the given Client, where route is
+// either a url path relative to `<RootURL>/api/<serviceName>/<apiVersion>` or
+// a fully qualified URL.  query is the set of query string parameters, if any,
+// and duration is the amount of time that the signed URL should remain valid
+// for.  The full-URL form permits signing URLs that are not even on the given
+// RootURL.
```
petemoore: It would be good to mention in the last sentence why signing foreign URLs is useful/desirable.

clients/client-go/codegenerator/model/api.go
djmitche: I don't think the assumption that an API method without scopes doesn't need a `_SignedURL` method is a valid assumption.  Rather than change it wholesale, I made this exception.  So I suppose we agree that this is a workaround, but differ on what it's working around.  Can we just leave it as a compromise?

clients/client-go/http.go
djmitche: I'm not sure it is useful / desirable -- I just added this to avoid dropping the feature from tc-proxy.

clients/client-go/creds_test.go
djmitche: I just copied this from the existing tests -- we could change that in a followup, if you'd like to file an issue or bug.

djmitche: (the monorepo uses `NO_TEST_SKIP` as an indicator that skipped tests should fail)

PR #2254, opened 2020-01-11
Added update sub-command in client-shell

### LINKED ISSUE #1963: [client-shell] Making it easier to update taskcluster cli client

or

DantrazTrev:
Github Bug/Issue: Fixes #1963 
Added update subcommand in client-shell which returns the url to the release if version isn't uptodate


clients/client-shell/cmds/version/version.go
```diff
@@ -18,13 +47,69 @@ var (
...
+
+    //Check if taskcluster is already up to date
+
+	if R.Name == VersionNumber {
+       
+        fmt.Println("taskcluster version  %s is already updated\n", VersionNumber)
+    
+    
+    } else {
+    fmt.Println("taskcluster curent version  %s is older than the latest %s \n", VersionNumber,R.Name)
```
srfraser: `Println` doesn't take string formatting in this way (and the CI job is reporting this at the link below. `fmt.Printf` should be fine. 
Nitpick: "current"


 https://community-tc.services.mozilla.com/tasks/Qii0N7nbRwWnd_5TRy8_CQ/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FQii0N7nbRwWnd_5TRy8_CQ%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L197-198

clients/client-shell/cmds/version/version.go
```diff
@@ -18,13 +47,69 @@ var (
...
+    fmt.Println("taskcluster curent version  %s is older than the latest %s \n", VersionNumber,R.Name)	
+    if runtime.GOOS == "linux" {
+    fmt.Println(R.Asslist[1].Download)
+  
+
+    }
+    if runtime.GOOS == "darwin"{
+    fmt.Println(R.Asslist[0].Download)
+                                }
+        }
```
srfraser: Could you run `go fmt` over the code, please? Thanks!

clients/client-shell/cmds/version/version.go
```diff
@@ -18,13 +47,69 @@ var (
...
+    R := Release{}
+    if err := json.Unmarshal([]byte(s), &R); err != nil {
+        fmt.Errorf(err.Error())
+    }
+
+    //Check if taskcluster is already up to date
+
+	if R.Name == VersionNumber {
+       
+        fmt.Println("taskcluster version  %s is already updated\n", VersionNumber)
```
srfraser: `Printf` here, too, I believe. "is already up to date" or "has no update available" reads more smoothly, imho. Thanks!

clients/client-shell/cmds/version/version.go
```diff
@@ -18,13 +47,69 @@ var (
...
+    //Create an object for the struct to parse the json data into given structure
+    
+    R := Release{}
+    if err := json.Unmarshal([]byte(s), &R); err != nil {
+        fmt.Errorf(err.Error())
+    }
+
+    //Check if taskcluster is already up to date
+
+	if R.Name == VersionNumber {
```
srfraser: When running a test just now it seems that `VersionNumber` is 24.1.9 (for me) and `R.Name` is "v24.1.10", so with the "v" prefix I'm not sure this check will ever be true.

clients/client-shell/cmds/version/version.go
DantrazTrev: Done

clients/client-shell/cmds/version/version.go
DantrazTrev: Changed

clients/client-shell/cmds/version/version.go
DantrazTrev: Done

clients/client-shell/cmds/version/version.go
DantrazTrev: Thanks

clients/client-shell/cmds/version/version.go
```diff
@@ -2,12 +2,40 @@
...
+// Useful structure for assets[] in the returned json data 
+type Assets struct {
+    Download string                  `json:"browser_download_url"`
+   }
+  
+   // Useful structure in the returned json data 
+  
+  type Release struct {
+      Name string                    `json:"name"`
+      Asslist []Assets                `json:"assets"`
```
djmitche: It's usually best to make the struct name mirror the JSON, name, with proper case, so here `Assets`.

clients/client-shell/cmds/version/version.go
```diff
@@ -23,8 +51,63 @@ var (
...
+    //Check if taskcluster is already up to date
+
+	if R.Name == "v"+VersionNumber {
+       
+        fmt.Printf("taskcluster is already updated")
+    
+    
+    } else {
+    if runtime.GOOS == "linux" {
+    fmt.Printf(R.Asslist[1].Download)
```
djmitche: These assets may not always be in this order.  It will be better to find the asset with the matching filename for GOARCH and GOOS.

PR #2262, opened 2020-01-13
Bug 1608828 - migrate generic-worker into monorepo

petemoore:
Bugzilla Bug: [1608828](https://bugzilla.mozilla.org/show_bug.cgi?id=1608828)

This initial PR integrates the generic-worker repository wholesale into the monorepo under /workers/generic-worker. The build scripts, CI tasks, package imports are updated as necessary, but the project is not restructured in any way. Future PRs can restructure or reorganise code as appropriate.

This is initially a draft pull request until I have everything working.

Git history from the generic-worker repository has been preserved.


petemoore:
Hi @djmitche, I've reduced this to 16 commits on top of the generic worker repository. I also rewrote the generic-worker history to remove the `/vendor` subfolder, which was very large and produced a lot of noise. With go modules, we don't have the `/vendor` folder, so I preferred to remove it rather than import it all, only for it to get deleted later (and it led to quite spurious contribution statistics).

It is probably easier to browse through the commits in sequence than view the complete diff, since that includes the entire generic-worker codebase.

Note, I haven't tackled updating the monorepo release process to publish generic-worker binaries, or integrated the build with `yarn` yet, but I figured these things can be done incrementally once this PR lands. This gets generic-worker building in the CI, although due to [bug 1613498](https://bugzil.la/1613498) we should keep a close eye on CI results before merging PRs, since the PR interface may make it appear that all CI tasks have passed when there are worker failures.

So despite not having all the i's dotted and t's crossed yet, I think this should be in good enough shape for a first pass.

Note, in the end I closed mozilla/community-tc-config#195 in favour of an [alternative approach](https://github.com/golang/go/issues/27089#issuecomment-415329050).

workers/generic-worker/gwconfig/config.go
```diff
@@ -29,7 +28,7 @@ type (
 
 	PublicConfig struct {
 		PublicEngineConfig
-		AuthBaseURL                    string                 `json:"authBaseURL"`
+		AuthRootURL                    string                 `json:"authRootURL"`
```
djmitche: I think this will require a README update as well, and it would be good to mention that the *RootURL configs are only for use in testing and development.

PR #2264, opened 2020-01-13
Title Formatting

aryamanpuri:
TItle Formatting for https://firefox-ci-tc.services.mozilla.com/quickstart

### DIFF
--- a/ui/src/components/Dashboard/index.jsx
+++ b/ui/src/components/Dashboard/index.jsx
@@ -167,7 +167,7 @@ export default class Dashboard extends Component {
     search: null,
     helpView: null,
     docs: false,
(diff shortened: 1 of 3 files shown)


ui/src/components/Dashboard/index.jsx
```diff
@@ -167,7 +167,7 @@ export default class Dashboard extends Component {
     search: null,
     helpView: null,
     docs: false,
-    disableTitleFormatting: false,
+    disableTitleFormatting: true,
```
helfi92: Doing this will affect all views. Let's do it on the quickstart view only.

aryamanpuri: Done

PR #2272, opened 2020-01-14
Check the head user when checking collaborators in PR

owlishDeveloper:
Bugzilla Bug: [1606874](https://bugzilla.mozilla.org/show_bug.cgi?id=1606874)

Github Bug/Issue: Fixes #XXXX

### DIFF
--- a/services/github/test/handler_test.js
+++ b/services/github/test/handler_test.js
@@ -217,6 +217,16 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
         handlers.handlerRejected = reject;
       });
 
+      let body = {
+        pull_request: {
+          head: {
+            user: {
+              login: user,
+            },
+          },
+        },
+      };
+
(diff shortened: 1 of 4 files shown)


services/github/test/handler_test.js
```diff
@@ -314,7 +325,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
       assert.equal(build.state, 'pending');
     });
 
-    test('valid pull_request (user is collaborator) creates a taskGroup', async function() {
+    test.only('valid pull_request (user is collaborator) creates a taskGroup', async function() {
```
imbstack: Stray `only`

djmitche:
Once this lands, let's get a point release out, too.

services/github/test/handler_test.js
owlishDeveloper: oopsies

PR #2274, opened 2020-01-16
Skip netlify builds when change is not in ui/

helfi92:


helfi92:
Hm, that doesn't seem to be working. From the Netlify site:

> For sites created before October 3rd, 2019
>
> If you would like to use this feature for a site that was created before October 3rd, 2019, you'll > need to re-link your respository to make this feature available for your site.

https://docs.netlify.com/configure-builds/file-based-configuration/#ignore-builds

PR #2277, opened 2020-01-16
Bug 1437193 replytwice

lcontant:
Bugzilla Bug: [1437193](https://bugzilla.mozilla.org/show_bug.cgi?id=1437193)


libraries/api/test/validate_test.js
```diff
@@ -129,6 +127,44 @@ suite(testing.suiteName(), function() {
...
+    method: 'get',
+    route: '/test-double-send',
+    name: 'testDoubleSend',
+    output: 'test-schema.yml',
+    category: 'API Library',
+    title: 'Test End-Point',
+    description: 'place to call to trigger a double send',
+  }, function(req, res) {
+    res.status(400).json('bad req');
+    res.status(200).reply({value: 1});
```
djmitche: what's the difference between this and `/test-double-json-send`?  Maybe this case should use `res.send(..)` instead?

libraries/api/src/middleware/schema.js
```diff
@@ -64,6 +64,9 @@ const validateSchemas = ({validator, absoluteSchemas, rootUrl, serviceName, entr
     // Add a reply method sending JSON replies, this will always reply with HTTP
     // code 200... errors should be sent with res.json(code, json)
     res.reply = (json) => {
+      if (res.headersSent) {
+        monitor.reportError(new Error('called send twice'));
```
djmitche: I think you'll want to `return` here, otherwise Express throws its *own* error ("Cannot set headers after they are sent to the client").

libraries/api/test/validate_test.js
```diff
@@ -296,4 +332,40 @@ suite(testing.suiteName(), function() {
...
+  test('calling send twice with reportError triggers an Error', () => {
+
+    const url = u('/test-double-error-send');
+    return request.get(url).then((res) => {
+      assert(false, 'Error didn\'t trigger');
+    }).catch(() => {
+      assert(true, 'Error was triggered');
+      setTimeout( () => {
+        assert(helper.monitorManager.messages[0].Fields.message === 'called send twice');
+      }, 100);
```
djmitche: This is a subtle bug!  The test here is synchronous, so as soon as the test function returns, Mocha marks it as passed and starts the next test. Then 100ms later, this setTimeout callback fires, and checks that the message exists because of the *next* test, not this one!

See the note above as to why this test case doesn't fail without the setTimeout.

libraries/api/test/validate_test.js
```diff
@@ -129,6 +127,44 @@ suite(testing.suiteName(), function() {
...
+    method: 'get',
+    route: '/test-double-error-send',
+    name: 'testDoubleErrorSend',
+    output: 'test-schema.yml',
+    category: 'API Library',
+    title: 'Test End-Point',
+    description: 'place to call to trigger a double send',
+  }, function(req, res) {
+    res.reportError('InputError', 'uhoh', {});
+    res.status(200).reply({value: 1});
```
djmitche: `res.reportError` actually throws an exception, so in this case, the `res.status` line never runs -- which is why the "called send twice" error is never logged for this API method (but see the note below for why the test case appeared to pass).

If you switch the order of these two lines, it will make it a more useful test case.

libraries/api/src/middleware/schema.js
djmitche: Also, let's use a more verbose error message: "API method implementation called res.send twice"

lcontant:
Would we need to caputre res.send for this to be a comprehensive error? 
The exemple I'm running into is res.json when called twice.

djmitche:
I'm not sure what you mean by "capture".  Can you show me what you're seeing?

lcontant:
I think I mis used the work capture. I mean I'm trying to detect when res.send is being called from anywhere so that I can know if it's being called a second time. A good example of this is for this test : 
```
res.status(400).json('bad req'); 
res.status(200).json('{value: 1}');
```
Where it doesn't seem possible to detect the call and throw the new exception.

djmitche:
Ah, I understand.  Since both `res.send` and `res.json` are methods provided by Express, there's no (good) way to intercept those calls and check for double-calls.  I think that's OK.  Let's limit this project to the cases where the second call made is `res.reply`.

lcontant:
Everything seems fine. It's a good idea to add the return when throwing the error.

djmitche:
Awesome, thanks for the check!

PR #2279, opened 2020-01-17
Added update sub-command in client-shell

### LINKED ISSUE #1963: [client-shell] Making it easier to update taskcluster cli client

DantrazTrev:
Github Bug/Issue: Fixes #1963 
Added the update command and the test for the command


clients/client-shell/cmds/version/version.go
```diff
@@ -18,13 +46,68 @@ var (
 
 	// VersionNumber is a formatted string with the version information. This is
 	// filled in by `yarn release`
-	VersionNumber = "24.2.0"
+	VersionNumber = "24.1.10"
```
djmitche: Please revert this change -- 24.2.0 is the correct current value.

clients/client-shell/cmds/version/version.go
```diff
@@ -18,13 +46,68 @@ var (
...
 	root.Command.AddCommand(Command)
+	root.Command.AddCommand(Updcommand)
 }
 
+
 func printVersion(cmd *cobra.Command, _ []string) {
 	fmt.Fprintf(cmd.OutOrStdout(), "taskcluster version %s\n", VersionNumber)
 }
+
+func check(cmd *cobra.Command, _ []string) {
```
djmitche: This should probably be named `update` since it corresponds to the `update` command..

clients/client-shell/cmds/version/version.go
```diff
@@ -18,13 +46,68 @@ var (
...
+    response, err := http.Get ("https://api.github.com/repos/taskcluster/taskcluster/releases/latest") 
+
+    if err != nil {
+        fmt.Print(err.Error())
+        
+    }
+
+    // Read the whole response body and check for any errors
+
+
```
djmitche: This diff has quite a few more newlines than other files in in client-shell.  Try to follow the existing style a little more closely, with fewer newlines.  There's plenty of room for minor variations, so long as the new code "blends in" to the existing.

clients/client-shell/cmds/version/version.go
```diff
@@ -18,13 +46,68 @@ var (
...
+    
+    
+    } else {
+    if runtime.GOOS == "linux" {
+    fmt.Printf(R.Asslist[1].Download)
+  
+
+    }
+    if runtime.GOOS == "darwin"{
+    fmt.Printf(R.Asslist[0].Download)
```
djmitche: See comments on #2254.

djmitche:
@srfraser followed up on this in another PR, which has landed.  Thanks for getting this started!

PR #2290, opened 2020-01-22
Add, but do not yet use, tc-lib-postgres

djmitche:
Bugzilla Bug: [1436478](https://bugzilla.mozilla.org/show_bug.cgi?id=1436478)

This lands the library itself, but nothing is using the library yet, and no deploy-time configuration is required.  The intent is to get the library reviewed and landed as a self-contained entity and then follow up with later stages of the implementation.  The README for tc-lib-postgres summarizes its functionality.

These commits were extracted (via git-filter-branch) from a larger branch where we're working on later functionality.  I'd like to keep them (via a merge commit) to maintain both authorship and a history of the development and wrong-turns we took in building this library.


djmitche:
It'd be super-cool to get this reviewed so I can make further changes :D

libraries/postgres/test/helper.js
```diff
@@ -0,0 +1,40 @@
...
+if (dbUrl) {
+  exports.dbSuite = (...args) => {
+    suite(...args.slice(0, -1), function() {
+      suiteSetup('setup database', function() {
+        exports.dbUrl = dbUrl;
+      });
+      setup('clear database', async function() {
+        await clearDb(dbUrl);
+      });
+      args[args.length - 1].call(this);
```
milescrabill: Am I right in thinking that `...args.slice(0, -1)` essentially serves as `args[0]` here?
I'm a little less sure about `args[args.length - 1].call(this)`, but I suppose that runs the suite itself?

libraries/postgres/src/Database.js
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  /**
+   * Private constructor (use Database.setup and Database.upgrade instead)
+   */
+  constructor({urlsByMode, statementTimeout}) {
+    const makePool = dbUrl => {
+      const pool = new Pool({connectionString: dbUrl});
+      // ignore errors from *idle* connections
+      pool.on('error', client => {});
```
imbstack: Will we really want to throw away 100% of errors from idle connections 100% of the time? I'm wondering if we'll hide some larger issue here by accident.

libraries/postgres/src/Database.js
```diff
@@ -0,0 +1,292 @@
...
+      pool.on('connect', async client => {
+        if (statementTimeout) {
+          await client.query(`set statement_timeout = ${statementTimeout}`);
+        }
+        // unconditionally apply a timeout for idle transactions. we should
+        // never be idle in a transaction (well, for more than few ms beteween
+        // statements), so this is set quite low.  Holding a transaction open
+        // can make locks pile up and also prevent vacuuming of tables, both
+        // leading to performance issues.
+        await client.query('set idle_in_transaction_session_timeout = 200');
```
imbstack: Should this be config even if it defaults to a low value?

I'm envisioning a situation in which we realize this needs to be a few ms higher to stop the trees from burning but it requires a full release to do so.

libraries/postgres/src/Database.js
```diff
@@ -0,0 +1,292 @@
...
+      this.pools[mode] = makePool(urlsByMode[mode]);
+    }
+
+    this.fns = {};
+  }
+
+  /**
+   * Run cb with a client.
+   *
+   * This is for INTERNAL USE ONLY.  All external access to the DB should be
```
imbstack: maybe "LIBRARY USE ONLY" I interpreted this as internal to TC. I know that is nonsensical but still

libraries/postgres/src/Database.js
```diff
@@ -0,0 +1,292 @@
...
+    const client = await pool.connect();
+    try {
+      try {
+        return await cb(client);
+      } catch (err) {
+        // show hints or details from this error in the debug log, to help
+        // debugging issues..
+        for (let p of ['hint', 'detail', 'where', 'code']) {
+          if (err[p]) {
+            err.message += `\n${p.toUpperCase()}: ${err[p]}`;
```
imbstack: Why not structured output?

libraries/postgres/src/util.js
```diff
@@ -0,0 +1,10 @@
+exports.dollarQuote = str => {
+  let i = '';
+  while (true) {
+    const quote = `$${i}$`;
+    if (str.indexOf(quote) === -1) {
+      return quote + str + quote;
+    }
+    i = i + '-';
+  }
+};
```
imbstack: This is like a tongue twister.

libraries/postgres/test/mocha.opts
```diff
@@ -0,0 +1,3 @@
+--ui tdd
```
imbstack: iirc we use `.mocharc.yml` in the monorepo root to configure this now. This file is unnecessary.

libraries/postgres/src/Version.js
```diff
@@ -0,0 +1,59 @@
...
+    return new Version(
+      content.version,
+      content.migrationScript,
+      objMap(content.methods,
+        ([name, meth]) => [name, Method.fromYamlFile(name, meth, filename)]),
+    );
+  }
+
+  static fromSerializable(serializable) {
+    for (let k of Object.keys(serializable)) {
```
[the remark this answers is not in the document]
djmitche: Only that the error messages aren't very helpful in that case.  That is more concise though :)

libraries/postgres/test/helper.js
djmitche: Not quite, the [negative index is from the end of the array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) so `slice(0, -1)` gets all but the last element.  And yes, `args[args.length - 1].call(this)` calls that last argument, which is the function defining the suite.

infrastructure/docker-images/build-node-and-pg.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# install postgres
+apt-get update
+apt-get install -y curl ca-certificates gnupg
+curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
+echo "deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main" > /etc/apt/sources.list.d/pgdg.list
+apt-get update
+apt-get install -y postgresql-$pg_version
+
+# allow postgres to connect locally with no auth -- this is for testing!
```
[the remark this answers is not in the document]
djmitche: This image is just for CI, so I think if we get to the point of putting "real" data in there we're pretty heavy into Doing It Wrong already, and having no password on the DB is not really the issue.

libraries/postgres/src/util.js
djmitche: It's definitely a weird quoting mechanism..

djmitche:
I'm going to land this, but if there's more discussion I'll be happy to make followups.

PR #2295, opened 2020-01-23
Convince 'yarn test:meta' that it is not running in a terminal

djmitche:
This avoids console-taskgraph outputting animation, which makes for a very large and repetitive logfile.


djmitche:
The changelog check is OK with PRs that only change whitelisted files.  `.taskcluster.yml` is one of those.

PR #2304, opened 2020-01-29
[Bug 1611266] - azure-provider: generate valid resource names

milescrabill:
Bugzilla Bug: [1611266](https://bugzilla.mozilla.org/show_bug.cgi?id=1611266)

Quick fix to azure-provider to generate valid names for azure resources by limiting generated names to alphanumeric

### DIFF
--- a/changelog/bug-1611266.md
+++ b/changelog/bug-1611266.md
@@ -1,0 +1,5 @@
+level: patch
+reference: bug 1611266
+---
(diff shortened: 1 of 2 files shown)


milescrabill:
> This will produce names with varying length -- is that OK?

Yes, as long as they're under the limit that is fine. We could throw even more slugids in to get closer to guaranteeing we generate sufficient characters, but I think this is ok.

PR #2305, opened 2020-01-30
Client shell update

### LINKED ISSUE #1963: [client-shell] Making it easier to update taskcluster cli client

srfraser:
Github Bug/Issue: Fixes #1963 

Based on work by @DantrazTrev, I was unable to update that PR as I'm not a maintainer here so I've submitted this variant. Feel free to adapt/copy back into the original PR if desired.

### DIFF
--- a/changelog/issue-1963.md
+++ b/changelog/issue-1963.md
@@ -1,0 +1,3 @@
+level: silent
+reference: issue 1963
+---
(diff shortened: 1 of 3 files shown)


srfraser:
> Looks good to me! I see that the tests hit the GitHub API, but that's probably fine -- at least in CI, the tests already hit GitHub to download the source, after all!

Thanks! The mocking I was investigating would have avoided that call, but the trade-off would have been extra configuration complexity (and a lot of tests to deal with potential api failures for that one call)

PR #2311, opened 2020-02-04
Improve deployment docs: pulse, schema

djmitche:
I set out to just doc the Pulse setup a little better, but decided to take a swing at including the values schema in the docs, too.


djmitche:
@edunham I'd love your feedback, too.  If you click on the netlify deploy in the checks below, you should be able to see the resulting documentation "live".

PR #2313, opened 2020-02-05
Docs: Fix invalid SID

edunham:
Per https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_grammar.html: "For IAM policies, basic alphanumeric characters (A-Z,a-z,0-9) are the only allowed characters in the Sid value."

This patch represents the change I had to make to successfully create the policy in AWS.

Tangentially related to https://bugzilla.mozilla.org/show_bug.cgi?id=1609276 in that I had to fix this to make progress on it. No changelog entry as it's literally just fixing a typo in the docs.

### DIFF
--- a/changelog/pull-2313.md
+++ b/changelog/pull-2313.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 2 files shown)


edunham:
```
$ yarn changelog --silent
Usage: yarn [options]

yarn: error: no such option: --silent
```
apparently I've gotten worse at yarn stuff since last time I patched typos around here, so I faked it and amended the commit.

PR #2314, opened 2020-02-05
Bug 1584208 - Improve client documentation

djmitche:
Bugzilla Bug: [1584208](https://bugzilla.mozilla.org/show_bug.cgi?id=1584208)

In particular to the bug, this references TASKCLUSTER_PROXY_URL, the environment-variables docs, and a new page about clients in general.

But it also rewrites all of the clients' READMEs to contain similar documentation, organized in a bit more of a logical fashion (for example, the constructor options for the Python client were distributed rather randomly over the README).


djmitche:
@helfi92 do we have a way to redirect /docs/manual/using/integration/libraries to /docs/manual/using/api?

PR #2315, opened 2020-02-06
[Bug 1587511] Allow worker-manager to terminate old workers

imbstack:
This should allow misconfigured workers to be reaped among other things. Right now nothing in here actually supports re-registering but it should be easy to add to providers. Just wanted to validate this approach before moving ahead.

This was finished last week at all-hands and I haven't looked at it since then so consider this a 90% review and I'll make sure it is up to good quality today as I get it paged back into my head.

### DIFF
--- a/services/worker-manager/schemas/v1/worker-lifecycle.yml
+++ b/services/worker-manager/schemas/v1/worker-lifecycle.yml
@@ -15,13 +15,13 @@ properties:
       been requsted from the cloud provider. After this
       timeout, worker-manager will terminate the instance,
       assuming it is broken.
-  checkinTimeout:
+  reregisterTimeout:
     title: Checkin Timeout
     type: integer
     default: 345600
     description: |
-      If specified, workers in this pool must re-register via `checkin()`
-      within `checkinTimeout` seconds to get new credentials. If the
+      If specified, workers in this pool must re-register via `reregister()`
+      within `reregisterTimeout` seconds to get new credentials. If the
       worker has not done so, it will be terminated. This defaults
       to 96 hours.
 additionalProperties: false
--- a/services/worker-manager/schemas/v1/worker-lifecycle.yml
+++ b/services/worker-manager/schemas/v1/worker-lifecycle.yml
@@ -10,10 +10,19 @@ properties:
     title: Registration Timeout
     type: integer
     description: |
-      Each worker in this pool has `timeout` seconds to
+      Each worker in this pool has `registrationTimeout` seconds to
       register itself with worker-manager after it has
       been requsted from the cloud provider. After this
       timeout, worker-manager will terminate the instance,
-      assuming it is broken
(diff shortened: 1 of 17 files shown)


services/worker-manager/src/providers/aws.js
```diff
@@ -271,8 +269,12 @@ class AwsProvider extends Provider {
       w.state = this.Worker.states.RUNNING;
     });
 
-    // return object that has expires field
-    return {expires: taskcluster.fromNow('96 hours')};
+    let expires = taskcluster.fromNow('96 hours');
+    if (worker.providerData.reregistrationDeadline) {
+      expires = new Date(worker.providerData.reregistrationDeadline);
```
djmitche: I suspect the providerData should contain the time delta, rather than an absolute time, as reregister will need to calculate a new deadline.

services/worker-manager/schemas/v1/worker-lifecycle.yml
```diff
@@ -10,10 +10,19 @@ properties:
...
       timeout, worker-manager will terminate the instance,
-      assuming it is broken
+      assuming it is broken.
+  reregisterTimeout:
+    title: Checkin Timeout
+    type: integer
+    default: 345600
+    description: |
+      If specified, workers in this pool must re-register via `reregister()`
+      within `reregisterTimeout` seconds to get new credentials. If the
```
djmitche: within .. seconds .. *of what*?

services/worker-manager/src/api.js
```diff
@@ -586,11 +633,29 @@ builder.declare({
...
+  return taskcluster.createTemporaryCredentials({
+    clientId: `worker/${worker.providerId}/${worker.workerPoolId}/${worker.workerGroup}/${worker.workerId}`,
+    scopes: [
+      `assume:worker-type:${worker.workerPoolId}`, // deprecated role
+      `assume:worker-pool:${worker.workerPoolId}`,
+      `assume:worker-id:${worker.workerGroup}/${worker.workerId}`,
+      `queue:worker-id:${worker.workerGroup}/${worker.workerId}`,
+      `secrets:get:worker-type:${worker.workerPoolId}`, // deprecated secret name
+      `secrets:get:worker-pool:${worker.workerPoolId}`,
+      `queue:claim-work:${worker.workerPoolId}`,
```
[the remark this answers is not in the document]
imbstack: I was thinking we would land all of this to get the "misconfigured things are terminated" stuff and then go back and add reregistering in a followup since that bit is less important.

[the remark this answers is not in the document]
djmitche: Sounds good

services/worker-manager/src/providers/aws.js
imbstack: Will do this when we add back reregister!

PR #2320, opened 2020-02-07
Avoid using tc-lib-scopes in tc-ui

### LINKED ISSUE #2318: UI uses tc-lib-scopes, which now requires lodash and json-stable-stringify
https://app.netlify.com/sites/taskcluster-web/deploys/5e3c9c8c64d9e2000ae68341

djmitche:
This includes a simple in-file implementation of flatten in terms of
reduce, and switches to fast-json-stable-stringify (which was already an
indirect requirement in ui/) for all stable stringification.

This should fix the netlify build errors.  We'll need to land this ASAP so that we can ship a release.

Github Bug/Issue: Fixes #2318

### DIFF
--- a/changelog/issue-2318.md
+++ b/changelog/issue-2318.md
@@ -1,0 +1,3 @@
+level: silent
+reference: issue 2318
+---
--- a/ui/package.json
+++ b/ui/package.json
@@ -87,7 +87,6 @@
     "resolve-pathname": "^3.0.0",
     "shebang-loader": "^0.0.1",
     "subscriptions-transport-ws": "^0.9.16",
-    "taskcluster-lib-scopes": "link:../libraries/scopes",
     "taskcluster-lib-urls": "^12.0.0",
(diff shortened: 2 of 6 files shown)


djmitche:
On pushing, it turns out that fast-json-stable-stringify doesn't do pretty output, and we need pretty output for `yarn generate`.  So, the updated patch uses both `fast-json-stable-stringify` and `json-stable-stringify`.  it turns out that both of these are already included for other reasons, so this doesn't end up adding an dependencies to TC.

djmitche:
OK, change of plans -- I pulled the `scopeIntersection` function out of tc-lib-scopes and put it in a ui utility.  I think this is a better approach that avoids cross-linking the two codebases.

djmitche:
Oh, ugh, that didn't work either.  I missed that tc-lib-scopes is used in a few places in tc-ui.  I think #2321 is the right fix here.

PR #2324, opened 2020-02-07
tc-lib-postgres support for reading from external files

djmitche:


djmitche:
Oh, my research did not turn up any reasonable SQL linters.  There's one in Perl which would be annoying to use (because perl) and who knows if we actually like the syntax is enforces.

PR #2329, opened 2020-02-10
Introduce taskcluster-lib-azure

### LINKED ISSUE #2292: figure out how queueing will work with entities
Plan is to introduce a tc-lib-msgqueue to support this.

djmitche:
This is a temporary library for simulating the Azure Queue API using
Postgres.  Fixes #2292.


libraries/azqueue/test/db/versions/0001.yml
```diff
@@ -0,0 +1,153 @@
...
+          returning m1.message_id, m1.message_text, m1.pop_receipt;
+      end
+
+  azure_queue_delete:
+    description: |
+      Delete the message identified by the given `queue_name`, `message_id` and
+      `pop_receipt`.
+    mode: write
+    serviceName: test-azqueue
+    args: queue_name text, message_id uuid, pop_receipt uuid
```
imbstack: Silly question but why do we need anything other than `message_id`

libraries/azqueue/test/slow_test.js
```diff
@@ -0,0 +1,177 @@
...
+  // index.  That index does two things: selects the specific queue (avoiding
+  // one huge queue slowing down other queries), and gets messages in the
+  // correct FIFO order (avoiding a sort).  We still scan over those messages
+  // sequentally to skip any that are invisible or expired.
+  //
+  // Room for improvement: this currently scans rows looking at visibility
+  // and expiration; ideally postgres could do that with an index.  The
+  // planner does not do so automatically, probably because of the LIMIT: combining
+  // two indexes would require scanning both and creating a bitmap, when in many
+  // cases only a partial index scan will be required to satisfy the LIMIT.
```
imbstack: Does this answer my question from above about indices?

imbstack: Can we rely on the planner always Doing the Right Thing :tm: ?

libraries/azqueue/test/db/versions/0001.yml
```diff
@@ -0,0 +1,153 @@
...
+    #   for update -- locks returned rows
+    #   skip locked -- skips locked rows
+    #
+    # The first ensures that no two concurrent transactions return the
+    # same message, while the second ensures that concurrent transactions
+    # do not serialize themselves but instead immediately return distinct
+    # messages, if they are available.
+    body: |-
+      begin
+        raise info 'oh dear %s', azure_queue_get.visible::text;
```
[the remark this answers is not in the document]
djmitche: haha, yes, it is :)

libraries/azqueue/test/db/versions/0001.yml
```diff
@@ -0,0 +1,153 @@
+version: 1
+migrationScript: |-
+  begin
+    create extension pgcrypto;
```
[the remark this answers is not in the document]
djmitche: Good point.. migration scripts are run once, so this is probably OK, but another alternative would be to run this before every upgrade and just ignore the errors.

libraries/azqueue/test/db/versions/0001.yml
```diff
@@ -0,0 +1,153 @@
...
+      message_id uuid not null primary key,
+      queue_name text not null,
+      message_text text not null,
+      inserted timestamp not null,
+      visible timestamp not null, -- visible after this time
+      expires timestamp not null,  -- expired after this time
+      pop_receipt uuid -- null means not popped
+    );
+    -- 'get' operations sort by inserted within a queue
+    create index azure_queue_messages_inserted on azure_queue_messages(queue_name, inserted);
```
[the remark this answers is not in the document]
djmitche: In postgres, `select count(*)` has basically the same performance as `select *` (it loads each row to check its visibility in the current transaction).  The count operation is `select count(*) from azure_queue_messages msgs where msgs.queue_name = azure_queue_count.queue_name`, the where clause for which is a prefix of this index, so it will use the index to scan the rows for that queue name.

libraries/azqueue/test/db/versions/0001.yml
djmitche: True, since that's the primary key.  This is reflecting the existing behavior of the Azure Queues API.  In that context, `pop_receipt` is a bearer proof that the caller is the one who popped the message from the queue, and I suppose queue_name helps to direct the request to the appropriate server.  In the postgres context, both function as secondary checks after the primary key index is used to find the row.

libraries/azqueue/test/slow_test.js
djmitche: > Does this answer my question from above about indices?

indirectly

> Can we rely on the planner always Doing the Right Thing tm ?

Not necessarily, but it usually has reasons for doing what it does.  Sometimes those reasons are wrong (bad heuristics).

djmitche:
Summarizing for myself:
 * [ ] enable extensions before running upgrades, ignoring errors
 * [ ] EXPLAIN the count operation to check my assertions above
 * [ ] remove "oh dear"

libraries/azqueue/test/db/versions/0001.yml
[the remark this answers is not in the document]
djmitche:
```
2020-02-11 22:28:44.572 UTC [64] LOG:  duration: 0.204 ms  plan:
        Query Text: SELECT (select count(*)
          from azure_queue_messages msgs
          where msgs.queue_name = azure_queue_count.queue_name)
        Result  (cost=11.30..11.31 rows=1 width=8) (actual time=0.193..0.194 rows=1 loops=1)
          InitPlan 1 (returns $0)
            ->  Aggregate  (cost=11.29..11.30 rows=1 width=8) (actual time=0.188..0.190 rows=1 loops=1)
                  ->  Bitmap Heap Scan on azure_queue_messages msgs  (cost=4.17..11.28 rows=3 width=0) (actual time=0.020..0.107 rows=100 loops=1)
                        Recheck Cond: (queue_name = 'queue-1-DEF'::text)
                        Heap Blocks: exact=3
                        ->  Bitmap Index Scan on azure_queue_messages_inserted  (cost=0.00..4.17 rows=3 width=0) (actual time=0.013..0.013 rows=100 loops=1)
                              Index Cond: (queue_name = 'queue-1-DEF'::text)
```

That's an index scan looking for the queue name, and then a scan of the table based on the results of that index scan -- as expected.

djmitche:
I think I want to block this on #2330 landing, at which point I can put the stored functions right into `db/`, and write unit tests and fakes for them at that time.

djmitche:
..in fact this will get moved around a bit, and some better tests, once #2330 has landed, so I'm going to close this without merging and plan to re-open a new PR at that time.  The "core" code will be about the same.

PR #2330, opened 2020-02-10
Make notify depend (lightly) on postgres

### LINKED ISSUE #2328: merge tc-lib-postgres with an empty or trivial db
(maybe a getWidget api function?) and get that deployed before putting any data in it.  A chance to practice deploying and see if tc-lib-postgres "works" in production without risking production data.

djmitche:
This introduces the `db/` directory and our very first DB version, which introduces a temporary widgets table.  It also shows off the testing arrangement for DB versions and functions.

With this change in place, a DB will be required to deploy Taskcluster.  Without it, or if it is misconfigured, the notify service will fail to start.

The intent is to allow setup of the DB deployment process in a low-impact fashion (this is "phase 0.5" of the TC postgres project).  If something goes wrong, having the notify service down for a few minutes is not a tree-closing event, as it will "catch up" on its queued pulse messages once it starts.

Github Bug/Issue: Fixes #2328


djmitche:
Hm, I don't know why netlify would have failed there.  This patch doesn't modify anything related, and the upstream revision's build succeeded.

libraries/postgres/src/util.js
```diff
@@ -24,3 +25,39 @@ exports.loadSql = (value, dir) => {
 
   return fs.readFileSync(pathname, 'utf8');
 };
+
+/**
+ * Annotate a postgres error message with additional useful information,
+ * including some additional attributes usually hidden as error properties and,
+ * for the common case of a syntax error, a nice caret highlighting the erro
```
imbstack: erro

.taskcluster.yml
```diff
@@ -73,6 +73,10 @@ tasks:
         - name: taskcluster-hooks
         - name: taskcluster-index
         - name: taskcluster-notify
+          image: taskcluster/node-and-postgres:node${node}-pg11
+          command: >-
+            ./db/test-setup.sh &&
+            yarn workspace taskcluster-notify coverage:report
```
imbstack: Why not just make all services do this db setup stuff now so that we don't have to expand out the `command` for this. They'll all need to do it soon anyway.

db/test-setup.sh
```diff
@@ -1,12 +1,24 @@
 #! /bin/bash
 
-if [ -n "${TASK_ID}" ]; then
+if [ -z "${TASK_ID}" ]; then
```
imbstack: lol

db/test/versions/0001_test.js
```diff
@@ -0,0 +1,23 @@
...
+const {UNDEFINED_TABLE} = require('taskcluster-lib-postgres');
+
+suite(testing.suiteName(), function() {
+  helper.withDbForVersion();
+
+  test('widgets table created', async function() {
+    await helper.withDbClient(async client => {
+      await assert.rejects(
+        () => client.query('select * from widgets'),
+        err => err.code === UNDEFINED_TABLE);
```
imbstack: Heck yes. I like this example for how we're going to structure these tests.

services/notify/src/api.js
```diff
@@ -215,3 +216,26 @@ builder.declare({
     continuationToken: query.continuation || undefined,
   });
 });
+
+builder.declare({
```
imbstack: If this is going into production don't we want to scope protect this? I know it is just a throwaway table but it seems nice to not let arbitrary access. We could also `noPublish` this maybe?

djmitche: Good point!  It turns out noPublish means that the method isn't even in the generated API clients, which wasn't very useful.  I don't know why that exists at all tbh.

djmitche:
I also added some checks to help us remember to do all the things in the checklist.

djmitche:
I'll hold off on landing this until @edunham has the deployment system ready, so that we don't end up with a released version of Taskcluster that we can't deploy.

djmitche:
latest estimate is end of the week for being able to deploy this to stage

djmitche:
This worked OK in staging!  Just going to wait until the automation is complete to deploy it everywhere, then land it.

PR #2331, opened 2020-02-10
Bug 1611266 - azure-provider must ensure adminPassword meets requirements

milescrabill:
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1611266


milescrabill:
@ajvb I feel pretty confident in the [generate-password](https://github.com/brendanashworth/generate-password/) library looking at the source, please let us know if you have any concerns

PR #2333, opened 2020-02-11
[Bug 1587511] (part 1) Terminate workers after a timeout

imbstack:
This just implements the deadline and worker termination phases of this work. Going to test this in my dev env tomorrow as well!

Supersedes: #2315

### DIFF
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -272,7 +272,7 @@ class AwsProvider extends Provider {
     await worker.modify(w => {
       w.lastModified = new Date();
       w.state = this.Worker.states.RUNNING;
-      w.providerData.terminateAfter = expires;
+      w.providerData.terminateAfter = expires.getTime();
     });
 
     return {expires};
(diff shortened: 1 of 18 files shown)


services/worker-manager/src/providers/provider.js
```diff
@@ -85,6 +85,20 @@ class Provider {
...
+
+  interpretLifecycle({lifecycle}) {
+    let registrationExpiry, reregisterDeadline = null;
+    if (lifecycle){
+      const {registrationTimeout, reregisterTimeout} = lifecycle;
+      if (registrationTimeout) {
+        registrationExpiry = Date.now() + registrationTimeout * 1000;
+      }
+      if (reregisterTimeout) {
+        reregisterDeadline = Date.now() + reregisterTimeout * 1000;
```
djmitche: The schema docs say this is the time since the `register()` call, but here it's being calculated when the instance is created. (this is the reason for the r-)

services/worker-manager/src/providers/aws.js
```diff
@@ -260,6 +258,11 @@ class AwsProvider extends Provider {
       throw new ApiError('Instance validation error');
     }
 
+    let expires = taskcluster.fromNow('96 hours');
+    if (worker.providerData.reregistrationTimeout) {
+      expires = new Date(Date.now() + worker.providerData.reregistrationTimeout);
```
djmitche: I think this needs a `*1000`?

PR #2340, opened 2020-02-12
Bug 1608185 - wait for protocol initialization

djmitche:
Bugzilla Bug: [1608185](https://bugzilla.mozilla.org/show_bug.cgi?id=1608185)

### DIFF
--- a/workers/generic-worker/usage.go
+++ b/workers/generic-worker/usage.go
@@ -32,6 +32,7 @@ and reports back results to the queue.
 
   Usage:
     generic-worker run                      [--config         CONFIG-FILE]
+                                            [--with-worker-runner]
                                             [--worker-runner-protocol-pipe PIPE]
                                             [--configure-for-aws | --configure-for-gcp | --configure-for-azure]` + installServiceSummary() + `
     generic-worker show-payload-schema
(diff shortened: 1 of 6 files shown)


workers/generic-worker/usage.go
```diff
@@ -32,15 +32,19 @@ and reports back results to the queue.
...
     generic-worker show-payload-schema
     generic-worker new-ed25519-keypair      --file ED25519-PRIVATE-KEY-FILE` + customTargetsSummary() + `
     generic-worker --help
     generic-worker --version
 
   Targets:
-    run                                     Runs the generic-worker.
+    run                                     Runs the generic-worker.  Pass --with-worker-runner if
+											running under that service, otherwise generic-worker will
+											not communicate with worker-runner.
```
petemoore: Something gone awry with the formatting here.

djmitche: Hm, it's just tabs vs. spaces.  I thought Go preferred tabs?

djmitche:
The test failures look like modifications to `go.mod`.  What would cause those?

djmitche:
Ah, got it..

workers/generic-worker/usage.go
```diff
@@ -32,15 +32,19 @@ and reports back results to the queue.
 
   Usage:
     generic-worker run                      [--config         CONFIG-FILE]
+                                            [--with-worker-runner]
                                             [--worker-runner-protocol-pipe PIPE]
                                             [--configure-for-aws | --configure-for-gcp | --configure-for-azure]` + installServiceSummary() + `
+
```
petemoore: I think this blank line broke docopt which indirectly caused the Windows failures. The broken tests on windows are calling a custom windows target which is defined in `customTargetsSummary()` on line 38. docopt doesn't see those options due to the blank line here, so returns the usage text.

djmitche:
alles grün!

PR #2346, opened 2020-02-13
Nest generic-worker README under 'Workers'

djmitche:
Should we remove any of the sub-directories that are showing up here?


djmitche:
The question was mostly for @petemoore -- for example I think the imagesets directory has been replaced with that in community-tc-config?

PR #2349, opened 2020-02-13
when a tooling execCommand has no logfile, include output in error me…

djmitche:
This just helps find the error instead of setting "see undefined for details"

### DIFF
--- a/infrastructure/tooling/src/utils/command.js
+++ b/infrastructure/tooling/src/utils/command.js
@@ -58,7 +58,8 @@ exports.execCommand = async ({
       if (code === 0 || ignoreReturn) {
         resolve(output);
       } else {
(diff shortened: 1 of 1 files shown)


infrastructure/tooling/src/utils/command.js
```diff
@@ -58,7 +58,9 @@ exports.execCommand = async ({
       if (code === 0 || ignoreReturn) {
         resolve(output);
       } else {
-        reject(new Error(`Nonzero exit status ${code}; see ${logfile} for details`));
+        utils.status({message: output});
```
imbstack: Stray status or do we always want to show it?

djmitche: Yeah, I think that doesn't work because the status is hidden once the task fails.  I'll remove.

PR #2354, opened 2020-02-14
Bug 1615631 - include generic-worker artifacts in GH release

djmitche:
This uses `build.sh` to build the artifacts, running it in such a way
that it skips the tests (which should already have passed) and simply
builds the artifacts.

Bugzilla Bug: [1615631](https://bugzilla.mozilla.org/show_bug.cgi?id=1615631)


djmitche:
For the client-shell binaries, we specify the variants to build right in the publish code (it's just two arches).  I could change this around so that it does the same, rather than using build.sh.

djmitche:
I'd like to see if this is how @petemoore would like this implemented before merging.

workers/generic-worker/build.sh
```diff
@@ -81,7 +93,7 @@ else
   esac
 fi
 
-find "${GOPATH}/bin" -name 'generic-worker*'
+find $OUTPUT_DIR -name 'generic-worker*'
```
petemoore: There should be double quotes around `$OUTPUT_DIR` in case the checkout directory contains spaces (which it could if developers specify an output directory).

PR #2356, opened 2020-02-14
Golang release fixes

djmitche:
This fixes some issues that came up in the v25.0.0 release.  See the commit comments for details.


djmitche:
Whoops, one of the tests assumes roles are arranged in a particular way in the Taskcluster deployment.  Which means it's testing the auth service's role expansion.  I'll delete it.

PR #2357, opened 2020-02-14
Allow running go test ./... without rootURL

djmitche:
Aside from the `./workers/generic-worker` package (which requires a build tag), this makes all of the go packages pass or skip when running a bare `go test`.  When running under CI, `NO_TEST_SKIP` is set, and these tests will fail if not given the proper credentials.

This follows on to #2356 and contains its commits.  I can rebase once that lands.

### DIFF
--- a/clients/client-go/codegenerator/model/model.go
+++ b/clients/client-go/codegenerator/model/model.go
@@ -160,7 +160,9 @@ func FormatSourceAndSave(sourceFile string, sourceCode []byte) {
 	formattedContent = regexp.MustCompile(`(?:[ \t]*//\n)?[ \t]*// See http://127.0.0.1:.*\n`).ReplaceAll(formattedContent, []byte(""))
 
 	// goimports often gets confused about versions based on whatever it finds
-	// in GOPATH, so reset the TC version to the appropriate value
+	// in GOPATH, so reset the TC version to the appropriate value.  Note that
+	// the last argument here will be updated to the current version by `yarn
+	// release`, so this will always substitute the correct version.
 	formattedContent = regexp.MustCompile(`github.com/taskcluster/taskcluster/v[0-9]+/`).ReplaceAll(formattedContent, []byte("github.com/taskcluster/taskcluster/v25/"))
 
 	// only perform general format, if that worked...
--- a/clients/client-go/integrationtest/integration_test.go
+++ b/clients/client-go/integrationtest/integration_test.go
@@ -1,192 +1,38 @@
 package integrationtest
 
 import (
-	"encoding/json"
 	"testing"
-	"time"
 
-	"github.com/taskcluster/slugid-go/slugid"
+	"github.com/stretchr/testify/require"
 	tcclient "github.com/taskcluster/taskcluster/v25/clients/client-go"
-	"github.com/taskcluster/taskcluster/v25/clients/client-go/tcindex"
-	"github.com/taskcluster/taskcluster/v25/clients/client-go/tcqueue"
-	"github.com/taskcluster/taskcluster/v25/internal/jsontest"
+	"github.com/taskcluster/taskcluster/v25/clients/client-go/tcauth"
+	"github.com/taskcluster/taskcluster/v25/clients/client-go/tcsecrets"
+	"github.com/taskcluster/taskcluster/v25/internal/testrooturl"
 )
 
-// This is a silly test that looks for the garbage index namespace, to somewhat
-// validate that the index client was generated correctly.
-//
-// Note, no credentials are needed.
-func TestGarbageNamespaces(t *testing.T) {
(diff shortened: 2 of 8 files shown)


clients/client-go/codegenerator/model/model.go
```diff
@@ -159,6 +159,10 @@ func FormatSourceAndSave(sourceFile string, sourceCode []byte) {
 	// remove links based on the schema server
 	formattedContent = regexp.MustCompile(`(?:[ \t]*//\n)?[ \t]*// See http://127.0.0.1:.*\n`).ReplaceAll(formattedContent, []byte(""))
 
+	// goimports often gets confused about versions based on whatever it finds
+	// in GOPATH, so reset the TC version to the appropriate value
+	formattedContent = regexp.MustCompile(`github.com/taskcluster/taskcluster/v[0-9]+/`).ReplaceAll(formattedContent, []byte("github.com/taskcluster/taskcluster/v25/"))
```
petemoore: Maybe we should add an extra comment that the seemingly hardcoded version on the RHS gets automatically updated by the ``/(github.com\/taskcluster\/taskcluster\/v)\d+/g, `$1${major}`)`` global find/replace operation at release time.

clients/client-go/integrationtest/doc.go
```diff
@@ -1,2 +0,0 @@
-// Package integrationtest stores all the integration tests that run against the taskcluster cluster client
-package integrationtest
```
petemoore: Did you mean to delete this file? I see the text has a mistake, but its purpose is to provide an overview of the package for godoc. See [The Go Blog - Godoc: documenting Go code](https://blog.golang.org/godoc-documenting-go-code):

> ... That package uses another convention for packages that need large amounts of introductory documentation: the package comment is placed in its own file, `doc.go`, which contains only those comments and a package clause.

clients/client-go/integrationtest/integration_test.go
```diff
@@ -1,192 +0,0 @@
...
-//
-// Note, no credentials are needed.
-func TestGarbageNamespaces(t *testing.T) {
-	Index := tcindex.NewFromEnv()
-	n, err := Index.ListNamespaces("garbage", "", "")
-	if err != nil {
-		t.Fatalf("%v\n", err)
-	}
-	t.Logf("Namespaces: %#v", n.Namespaces)
-}
```
petemoore: I think this was a useful test. Did you mean to delete it? It ensures that after we've built the client with the code generation, that it can perform at least a simple API request, to list index namespaces.

clients/client-go/integrationtest/integration_test.go
```diff
@@ -1,192 +0,0 @@
...
-	tempCreds, err := permaCreds.CreateTemporaryCredentials(30*time.Second, "queue:cancel-task:"+td.SchedulerID+"/"+td.TaskGroupID+"/"+taskID)
-	if err != nil {
-		t.Fatalf("Exception thrown generating temporary credentials!\n\n%s\n\n", err)
-	}
-	myQueue = tcqueue.New(tempCreds, rootURL)
-	_, err = myQueue.CancelTask(taskID)
-	if err != nil {
-		t.Fatalf("Exception thrown cancelling task with temporary credentials!\n\n%s\n\n", err)
-	}
-}
```
petemoore: What is the reason to delete this test?

djmitche:
I had some comments in the commit message regarding removal of the integration tests:

    One of these never ran, as it requires real (not temporary) credentials.
    It also requires scopes that no longer exist, and involves a complicated
    and fragile process of setting up a task.
    
    The other runs without authentication and accesses an index path that
    may not exist (and doesn't on the rootURL I happen to be using for
    testing).
    
    In any case, with generic-worker and other Go apps in-tree, we get
    adequate integration testing of this library through those consumers.

So that's my rationale overall.  Basically, both of these would need to be rewritten.

I'll bring back the integrationtest package and add something similar to the namespace check but without relying on specific data in the upstream deployment.

djmitche:
> We should keep git history when we migrate repos into the monorepo, such as taskcluster-base-go, to be fair to the original authors.

You're right - I should have done so.  This merged in #2356, though.  I don't think this can be fixed now without rewriting history.

petemoore:
I think this is probably fine, but I'd like to see #2374 land first, and then merge it in here, and check everything is ok, since #2374 is quite an overhaul, and maybe these two PR's might collide. We should also get better test/lint coverage of this PR once the changes from #2374 are merged in here. Thanks!

djmitche:
force-pushed, so if the g-w tasks in https://community-tc.services.mozilla.com/tasks/groups/BoaHKFLQRsqDtvfaVcdg-Q succeed, then I think we're in good shape

djmitche:
Oh, possibly.  I'll fix that up on master.

PR #2358, opened 2020-02-14
Bug 1606874 - check that the person who filed a PR is a collaborator

djmitche:
Before this change, the collaborator check was performed on the owner of
the "head" repo.

Bugzilla Bug: [1606874](https://bugzilla.mozilla.org/show_bug.cgi?id=1606874)

### DIFF
--- a/changelog/bug-1606874.md
+++ b/changelog/bug-1606874.md
@@ -1,0 +1,4 @@
+level: patch
+reference: bug 1606874
+---
+The Taskcluster-GitHub service now checks that the person who *filed* a pull request is a collaborator and the repo from which the changes are being pulled belongs to a collaborator or is the usptream repository.
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -630,19 +630,42 @@ async function jobHandler(message) {
     if (!defaultBranchYml) { return; }
 
     if (this.getRepoPolicy(defaultBranchYml).startsWith('collaborators')) {
-      let login = message.payload.body.pull_request.head.user.login;
+      // There are four usernames associated with a PR action:
+      //  - pull_request.user.login -- the user who opened the PR
+      //  - pull_request.head.user.login -- the username or org name for the repo from which changes are pulled
+      //  - pull_request.base.user.login -- the username or org name for the repo into which changes will merge
+      //  - sender.login -- the user who clicked the button to trigger this action
+      //
+      // The "collaborators" and "collaborators_quiet" policies require:
+      //  - pull_request.user.login is a collaborator; AND
+      //  - pull_request.head.user.login is
+      //    - a collaborator OR
+      //    - the same as pull_request.base.user.login
+      //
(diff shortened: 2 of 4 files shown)


djmitche:
(yarn generate will fail here -- that's fixed in #2356)

changelog/bug-1606874.md
```diff
@@ -0,0 +1,4 @@
+level: minor
```
owlishDeveloper: shouldn't it be a patch?

owlishDeveloper:
also, `yarn generate`

djmitche:
> (yarn generate will fail here -- that's fixed in #2356)

changelog/bug-1606874.md
djmitche: good point

owlishDeveloper:
> > (yarn generate will fail here -- that's fixed in #2356)

merging

djmitche:
Updated based on the discussion in the bug.

services/github/src/handlers.js
```diff
@@ -630,19 +630,42 @@ async function jobHandler(message) {
...
+      //  - sender.login -- the user who clicked the button to trigger this action
+      //
+      // The "collaborators" and "collaborators_quiet" policies require:
+      //  - pull_request.user.login is a collaborator; AND
+      //  - pull_request.head.user.login is
+      //    - a collaborator OR
+      //    - the same as pull_request.base.user.login
+      //
+      // Meaning that the PR must have been opened by a collaborator and be merging code from a collaborator
+      // or from the repo against which the PR is filed.
```
owlishDeveloper: so here we're relying on the repo's policies about who can push branches to it?

services/github/src/handlers.js
```diff
@@ -630,19 +630,42 @@ async function jobHandler(message) {
...
+          }
+          return false; // 404 -> false
+        }));
+      };
+
+      const evt = message.payload.body;
+      const opener = evt.pull_request.user.login;
+      const openerIsCollaborator = await isCollaborator(opener);
+      const head = evt.pull_request.head.user.login;
+      const headIsCollaborator = head === opener ? openerIsCollaborator : await isCollaborator(head);
```
owlishDeveloper: why `opener ?` check? We already used `opener` by this point, so `openerIsCollaborator && await ...` should be enough?

services/github/src/handlers.js
djmitche: Yes -- there was some discussion of that in the bug and it's not clear it's even possible to do better.

services/github/src/handlers.js
djmitche: It's a little bit of an optimization to call the GitHub API less often.  The reasoning is that the common case is that the head and the opener are the same user, so in that case (`head === opener`) we just use the result of the `isCollaborator` call for that user (`openerIsCollaborator`).  Otherwise, we call the GH API again for the head user.

PR #2359, opened 2020-02-14
Bug 1615742 - support staging releases

djmitche:
staging-release


djmitche:
Oh, right, we don't want PRs to run on the release worker. Harrrumph.

PR #2383, opened 2020-02-18
Merge taskcluster-base-go

djmitche:
In #2356 I failed to keep history when importing this library.  This adds the history.


djmitche:
Yeah, sorry I missed it in the first place.  I started out just copying one function and then when I decided to pull all the code I didn't re-consider the strategy.

PR #2384, opened 2020-02-19
Bug 1616649: azure-provider avoid blocking operations

milescrabill:
Bugzilla Bug: [1616649](https://bugzilla.mozilla.org/show_bug.cgi?id=1616649)

note: consulted with @ajvb about changing workerId to key off of generated names, got a 👍

I have tested this in my dev environment end-to-end, I used a custom build of worker-runner and was able to successfully register. Due to unrelated issues with the azure docker-worker VM tasks weren't claimed, but everything else works!*

\* to a reasonable observable degree of confidence

### DIFF
--- a/services/worker-manager/src/providers/azure.js
+++ b/services/worker-manager/src/providers/azure.js
@@ -17,6 +17,14 @@ const msRestAzure = require('@azure/ms-rest-azure-js');
 const {ApiError, Provider} = require('./provider');
 const {CloudAPI} = require('./cloudapi');
 
+// Azure provisioning and VM power states
+// see here: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/states-lifecycle
+// same for linux: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/states-lifecycle
+const successPowerStates = new Set(['PowerState/running', 'PowerState/starting']);
+const failPowerStates = new Set(['PowerState/stopping', 'PowerState/stopped', 'PowerState/deallocating', 'PowerState/deallocated']);
+const successProvisioningStates = new Set(['Succeeded', 'Creating', 'Updating']);
+const failProvisioningStates = new Set(['Failed', 'Deleting', 'Canceled', 'Deallocating']);
+
 // only use alphanumeric characters for convenience
 function nicerId() {
   return (slugid.nice() + slugid.nice() + slugid.nice()).toLowerCase().replace(/[^A-Za-z0-9]/g, '');
@@ -581,9 +589,16 @@ class AzureProvider extends Provider {
         worker.providerData.resourceGroupName,
         worker.providerData.vm.name,
       ));
-      const successStates = ['Creating', 'Updating', 'Starting', 'Running', 'Succeeded'];
-      const failStates = ['Failed', 'Canceled', 'Deleting', 'Deallocating', 'Stopped', 'Deallocated'];
-      if (successStates.includes(provisioningState)) {
+      // lets us get power states for the VM
+      const instanceView = await this._enqueue('get', () => this.computeClient.virtualMachines.instanceView(
+        worker.providerData.resourceGroupName,
+        worker.providerData.vm.name,
+      ));
+      const powerStates = instanceView.statuses.map(i => i.code);
+      if (successProvisioningStates.has(provisioningState) &&
+          // fairly lame check, succeeds if we've ever been starting/running
+          _.some(powerStates, (v) => { return successPowerStates.has(v); })
+      ) {
         this.seen[worker.workerPoolId] += worker.capacity || 1;
 
         // If the worker will be expired soon but it still exists,
@@ -608,7 +623,10 @@ class AzureProvider extends Provider {
             w.providerData.vm.vmId = vmId;
           });
         }
-      } else if (failStates.includes(provisioningState)) {
+      } else if (failProvisioningStates.has(provisioningState) ||
+                // if the VM has ever been in a failing power state
+                _.some(powerStates, (v) => { return failPowerStates.has(v); })
+      ) {
         state = await this.removeWorker({worker});
         this.monitor.log.workerStopped({
           workerPoolId: worker.workerPoolId,
@@ -618,7 +636,7 @@ class AzureProvider extends Provider {
       } else {
         await worker.workerPool.reportError({
           kind: 'creation-error',
-          title: 'Encountered unknown VM provisioningState',
+          title: 'Encountered unknown VM provisioningState or powerStates',
           description: `Unknown provisioningState ${provisioningState}`,
         });
       }
--- a/services/worker-manager/src/providers/azure.js
+++ b/services/worker-manager/src/providers/azure.js
@@ -148,11 +148,8 @@ class AzureProvider extends Provider {
     // Create "empty" workers to provision in provisionResources loop
     await Promise.all(cfgs.map(async cfg => {
       // This must be unique to currently existing instances and match [a-z]([-a-z0-9]*[a-z0-9])?
-      // The lost entropy from downcasing, etc should be ok due to the fact that
-      // only running instances need not be identical. We do not use this name to identify
-      // workers in taskcluster.
       const poolName = workerPoolId.replace(/[\/_]/g, '-').slice(0, 38);
-      const virtualMachineName = `vm-${poolName}-${nicerId()}`.slice(0, 38);
+      const virtualMachineName = `vm-${poolName}-${nicerId()}-${nicerId()}`.slice(0, 64);
       // Windows computer name cannot be more than 15 characters long, be entirely numeric,
       // or contain the following characters: ` ~ ! @ # $ % ^ & * ( ) = + _ [ ] { } \\ | ; : . " , < > / ?
       const computerName = nicerId().slice(0, 15);
@@ -396,7 +393,8 @@ class AzureProvider extends Provider {
   async handleOperation({op, errors}) {
     let req, resp;
     try {
-      // TODO: we could be smarter about getting operation status
+      // NB: we don't respect azure's Retry-After header, we assume our iteration
+      // will wait long enough, and we keep trying
       // see here: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations
       req = new msRestJS.WebResource(op, 'GET');
       // sendLongRunningRequest polls until finished but this is just reading
@@ -404,13 +402,19 @@ class AzureProvider extends Provider {
       // it's ok if we hit an error here, that will trigger resource teardown
       resp = await this._enqueue('opRead', () => this.restClient.sendLongRunningRequest(req));
     } catch (err) {
-      // Rest API has different error semantics than the SDK
-      if (_.has(resp, 'status') && resp.status === 404) {
-        // operation not found because it has either expired or does not exist
-        // nothing more to do
-        return false;
-      }
-      throw err;
+      errors.push({
+        kind: 'operation-error',
+        title: 'Operation Error',
+        description: err.message,
+        notify: this.notify,
+        WorkerPoolError: this.WorkerPoolError,
+      });
+    }
+    // Rest API has different error semantics than the SDK
+    if (resp.status === 404) {
+      // operation not found because it has either expired or does not exist
+      // nothing more to do
+      return false;
     }
 
     let body = resp.parsedBody;
@@ -440,8 +444,12 @@ class AzureProvider extends Provider {
    * provisionResource generically provisions individual resources
    * Handles cases where:
    *  we have not yet created a resource and need to create one,
+   *    * we have no id, get request for name 404s, no operation
    *  we have requested a resource but it is not ready,
+   *    * we have no id, get request for name 404s, we have an operation
    *  we have a resource ready to go
+   *    * we have an id, we short circuit return
+   *    * OR we have no id, get request for name succeeds, we set id
    *
    * worker: the worker for which the resource is being provisioned
    * client: the Azure SDK client for the resource
@@ -583,6 +591,7 @@ class AzureProvider extends Provider {
         title: 'VM Creation Error',
         description: err.message,
       });
+      return await this.removeWorker({worker});
     }
   }
 
@@ -604,7 +613,7 @@ class AzureProvider extends Provider {
       const powerStates = instanceView.statuses.map(i => i.code);
       if (successProvisioningStates.has(provisioningState) &&
           // fairly lame check, succeeds if we've ever been starting/running
-          _.some(powerStates, (v) => { return successPowerStates.has(v); })
+          _.some(powerStates, v => successPowerStates.has(v))
       ) {
         this.seen[worker.workerPoolId] += worker.capacity || 1;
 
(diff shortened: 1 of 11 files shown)


services/worker-manager/src/providers/azure.js
```diff
@@ -387,19 +367,223 @@ class AzureProvider extends Provider {
...
+   * Checks the status of ongoing Azure operations
+   * Returns true if the operation is in progress, false otherwise
+   *
+   * op: a URL for tracking the ongoing status of an Azure operation
+   * errors: a list that will have any errors found for that operation appended to it
+   */
+  async handleOperation({op, errors}) {
+    let req, resp;
+    try {
+      // TODO: we could be smarter about getting operation status
```
imbstack: let's get a bug on file for this and add the bugid to this TODO if we're not going to change it before we land this

services/worker-manager/src/providers/azure.js
```diff
@@ -387,19 +367,223 @@ class AzureProvider extends Provider {
...
+      // see here: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations
+      req = new msRestJS.WebResource(op, 'GET');
+      // sendLongRunningRequest polls until finished but this is just reading
+      // the status of an operation so it shouldn't block long
+      // it's ok if we hit an error here, that will trigger resource teardown
+      resp = await this._enqueue('opRead', () => this.restClient.sendLongRunningRequest(req));
+    } catch (err) {
+      // Rest API has different error semantics than the SDK
```
imbstack: hooray

services/worker-manager/src/providers/azure.js
```diff
@@ -451,49 +635,163 @@ class AzureProvider extends Provider {
...
-    ]);
-    // the public IP cannot be deleted as long as it is attached
-    // can become orphaned if we try to delete before the NIC is deleted
-    await ignoreNotFound(this._enqueue('query', () => this.networkClient.publicIPAddresses.deleteMethod(
-      resourceGroupName,
-      ip.name,
-    )));
+    }
+
+    // FIXME: possible resource leak
```
imbstack: I think it's possible we'll just need to add the background scanner for all resources in a cloud to catch all leaks

djmitche: [Belt and Suspenders](https://en.wiktionary.org/wiki/belt_and_suspenders)

milescrabill: I've added more exhaustive checking (we ensure the resource is gone before moving on)

services/worker-manager/test/fake-azure.js
```diff
@@ -77,45 +87,43 @@ class FakeAzure {
...
 
     const azureError = new Error("something went wrong");
 
-    const getVMStub = sinon.stub();
-    // first call returns provisioningState Creating, second returns Deallocated
-    getVMStub.onCall(0).returns(instanceData);
-    getVMStub.onCall(1).returns({...instanceData, provisioningState: 'Deallocated'});
+    this.getVMStub = sinon.stub();
+    // first call returns provisioningState Creating, second returns Failed
+    this.getVMStub.onCall(0).returns({...instanceData, provisioningState: 'Succeeded'});
```
djmitche: Comment says "Creating" but the code says "Succeeded"??

services/worker-manager/src/providers/azure.js
```diff
@@ -65,16 +86,16 @@ class AzureProvider extends Provider {
...
-      types: ['get', 'query', 'list'],
+      types: ['query', 'get', 'list', 'opRead'],
       apiRateLimits,
       intervalDefault: 100 * 1000, // Intervals are enforced every 100 seconds
       intervalCapDefault: 2000, // The calls we make are all limited 20/sec so 20 * 100 are allowed
       monitor: this.monitor,
       providerId: this.providerId,
       errorHandler: ({err, tries}) => {
-        if (err.code === 429) { // too many requests
+        if (err.statusCode === 429) { // too many requests
```
djmitche: This seems like an important fix!

services/worker-manager/src/providers/azure.js
```diff
@@ -123,12 +145,12 @@ class AzureProvider extends Provider {
       toSpawn -= cfg.capacityPerInstance;
     }
 
+    // Create "empty" workers to provision in provisionResources loop
     await Promise.all(cfgs.map(async cfg => {
       // This must be unique to currently existing instances and match [a-z]([-a-z0-9]*[a-z0-9])?
       // The lost entropy from downcasing, etc should be ok due to the fact that
       // only running instances need not be identical. We do not use this name to identify
       // workers in taskcluster.
```
djmitche: Is this comment still true?  It seems this value is now used for workerId..

services/worker-manager/src/providers/azure.js
djmitche: bump ^^ :)

services/worker-manager/src/providers/azure.js
```diff
@@ -387,19 +383,229 @@ class AzureProvider extends Provider {
...
+      // TODO: we could be smarter about getting operation status
+      // see here: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations
+      req = new msRestJS.WebResource(op, 'GET');
+      // sendLongRunningRequest polls until finished but this is just reading
+      // the status of an operation so it shouldn't block long
+      // it's ok if we hit an error here, that will trigger resource teardown
+      resp = await this._enqueue('opRead', () => this.restClient.sendLongRunningRequest(req));
+    } catch (err) {
+      // Rest API has different error semantics than the SDK
+      if (_.has(resp, 'status') && resp.status === 404) {
```
djmitche: I suspect this can never occur -- `resp` is only set *after* the `_enqueue`d operation has completed successfully, meaning that no error would be thrown.  Should this analysis of `resp` happen *outside* of the catch?

services/worker-manager/src/providers/azure.js
```diff
@@ -387,19 +383,229 @@ class AzureProvider extends Provider {
...
+      try {
+        let resource = await this._enqueue('query', () => client.get(
+          worker.providerData.resourceGroupName,
+          typeData.name,
+        ));
+        // we found the resource
+        await worker.modify(w => {
+          w.providerData[resourceType].id = resource.id;
+          modifyFn(w, resource);
+        });
```
djmitche: do you walso want to update `typeData.id` here, or does that just get seen on the next go-round?

services/worker-manager/src/providers/azure.js
```diff
@@ -387,19 +383,229 @@ class AzureProvider extends Provider {
...
+
+    return false;
+  }
+
+  /**
+   * provisionResource generically provisions individual resources
+   * Handles cases where:
+   *  we have not yet created a resource and need to create one,
+   *  we have requested a resource but it is not ready,
+   *  we have a resource ready to go
```
djmitche: I think these cases are identified by different values of `typeData.id`, `.name`, and `.operation`, but I'm not quite clear which is which -- could that be described here?

services/worker-manager/src/providers/azure.js
```diff
@@ -387,19 +383,229 @@ class AzureProvider extends Provider {
...
+        resourceConfig: workerConfigWithSecrets(worker.providerData.vm.config),
+        modifyFn: () => {},
+      });
+      if (!worker.providerData.vm.id) {
+        return worker.state;
+      }
+      return this.Worker.states.RUNNING;
+    } catch (err) {
+      // we create multiple resources in order to provision a VM
+      // if we catch an error we want to deprovision those resources
```
djmitche: ..but this block only reports the error, and doesn't de-provision?

services/worker-manager/src/providers/azure.js
```diff
@@ -387,19 +383,229 @@ class AzureProvider extends Provider {
...
-      if (['Creating', 'Updating', 'Starting', 'Running', 'Succeeded'].includes(provisioningState)) {
+      // lets us get power states for the VM
+      const instanceView = await this._enqueue('get', () => this.computeClient.virtualMachines.instanceView(
+        worker.providerData.resourceGroupName,
+        worker.providerData.vm.name,
+      ));
+      const powerStates = instanceView.statuses.map(i => i.code);
+      if (successProvisioningStates.has(provisioningState) &&
+          // fairly lame check, succeeds if we've ever been starting/running
+          _.some(powerStates, (v) => { return successPowerStates.has(v); })
```
djmitche: (nit) shorter syntax for the function is `v => successPowerStates.has(v)`

services/worker-manager/src/providers/azure.js
```diff
@@ -412,35 +618,49 @@ class AzureProvider extends Provider {
...
         }
         if (worker.providerData.terminateAfter && worker.providerData.terminateAfter < Date.now()) {
-          await this.removeWorker({worker});
+          state = await this.removeWorker({worker});
+        }
+
+        // vm has successfully provisioned, we need to set id and vmId
+        // id is the fully qualified azure resource ID
+        // vmId is a uuid, we use it for registering workers
+        if (!worker.providerData.vm.id || !worker.providerData.vm.id) {
```
djmitche:
```suggestion
        if (!worker.providerData.vm.id || !worker.providerData.vm.vmId) {
```

services/worker-manager/src/providers/azure.js
```diff
@@ -451,49 +671,158 @@ class AzureProvider extends Provider {
...
+      }
+    }));
   }
 
-  async _removeWorker({resourceGroupName, location, vm, ip, nic, disk}) {
-    // clean up related resources, continue on errors
-    // delete VM, IP, NIC, and disk
-    const ignoreNotFound = async (promise) => {
+  /*
+   * removeResource attempts to delete a resource and verify deletion
```
djmitche: what does the return value mean?

services/worker-manager/src/providers/azure.js
```diff
@@ -451,49 +671,158 @@ class AzureProvider extends Provider {
...
-    ]);
-    // the public IP cannot be deleted as long as it is attached
-    // can become orphaned if we try to delete before the NIC is deleted
-    await ignoreNotFound(this._enqueue('query', () => this.networkClient.publicIPAddresses.deleteMethod(
-      resourceGroupName,
-      ip.name,
-    )));
+    }
+
+    // NB: possible resource leak if we don't require `return true`
```
djmitche: ..`return true` where?  I don't see it below here..

services/worker-manager/src/providers/azure.js
```diff
@@ -451,49 +671,158 @@ class AzureProvider extends Provider {
...
+      // change to stopped
+      state = states.STOPPED;
+      await worker.modify(w => {
+        const now = new Date();
+        w.lastModified = now;
+        w.lastChecked = now;
+        w.state = state;
+      });
+    } catch (err) {
+      // if this is called directly and not via checkWorker may not exist
```
djmitche: .. is this the case when this is called by the REST API?  In that case, what happens to the errors?

services/worker-manager/src/providers/azure.js
milescrabill: Good catch, that second part is not true. According to [this](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules) VM names can be up to 64 characters long, so I've bumped that up and added another nicerId() for entropy.

services/worker-manager/src/providers/azure.js
milescrabill: Yes! Fixed.

services/worker-manager/src/providers/azure.js
milescrabill: My impression is that `typeData.id` refers to the same object (?) and so it is effectively getting updated here.

services/worker-manager/src/providers/azure.js
milescrabill: Fixed!

services/worker-manager/src/providers/azure.js
milescrabill: Added a bit more detail.

services/worker-manager/src/providers/azure.js
milescrabill: From this function, i.e. what I mean by that is that the caller should check the return value to make sure we return true, otherwise the deletion might not have finished. Just awaiting the function might have resources in a still deleting state, even though the `id` has been set to `false`.

For example, we check `vmDeleted`:

```
let vmDeleted = await this.removeResource({
  worker,
  client: this.computeClient.virtualMachines,
  resourceType: 'vm',
});
if (!vmDeleted || worker.providerData.vm.id) {
  return state;
}
```

services/worker-manager/src/providers/azure.js
milescrabill: The only case where this has been necessary that I've seen is in the tests. I haven't tested the REST API for deleteWorker() - this comment creates those things, i.e. handles the case where they don't exist.

services/worker-manager/src/providers/azure.js
milescrabill: added a more detailed comment, returning `true` means we're positive the resource to delete is gone. if we return false we've made a request to delete that hasn't finished yet.

services/worker-manager/test/fake-azure.js
milescrabill: Fixed the comment. Essentially, I hadn't differentiated between `provisioningStates` and `powerStates`, and now we are.

services/worker-manager/src/providers/azure.js
milescrabill: I've downgraded this to an `NB` because I'm pretty sure it's inconsequential

services/worker-manager/src/providers/azure.js
djmitche: It's true, but workerId is limited to 38 characters -- https://community-tc.services.mozilla.com/schemas/worker-manager/v1/worker-full.json

services/worker-manager/src/providers/azure.js
djmitche: Not necessarily -- the `modify` function might have to reload the worker object, and iirc ends up rewriting the properties anyway.

services/worker-manager/src/providers/azure.js
```diff
@@ -148,8 +148,9 @@ class AzureProvider extends Provider {
     // Create "empty" workers to provision in provisionResources loop
     await Promise.all(cfgs.map(async cfg => {
       // This must be unique to currently existing instances and match [a-z]([-a-z0-9]*[a-z0-9])?
+      // 38 chars is workerPoolId / workerId limit
       const poolName = workerPoolId.replace(/[\/_]/g, '-').slice(0, 38);
-      const virtualMachineName = `vm-${poolName}-${nicerId()}-${nicerId()}`.slice(0, 64);
+      const virtualMachineName = `vm-${poolName}-${nicerId()}-${nicerId()}`.slice(0, 38);
```
djmitche: For a 36-char workerPoolId, this would leave one character of entropy for the machine name.  Let's address that in a followup, but ensuring a minimum of entropy and truncating the poolName to accomodate it

PR #2387, opened 2020-02-20
No bug - delete unused directory

petemoore:
No bug - just deletes an unused directory.


djmitche:
I was going to do this in my docs patch :)

djmitche:
(missing changelog is fine here)

PR #2389, opened 2020-02-20
Bug 1616900 - wait for proto initialization in tests

djmitche:
This avoids a subtle race condition where `runnerProto` is not
considered initialized until some time *after* the `hello` message is
received, as the function handling that message must run.

`go test -race` missed this because the use of a Mutex avoids an actual
data race.  The bug here is at a higher level than that.

Bugzilla Bug: [1616900](https://bugzilla.mozilla.org/show_bug.cgi?id=1616900)

### DIFF
--- a/changelog/bug-1616900.md
+++ b/changelog/bug-1616900.md
@@ -1,0 +1,3 @@
+level: silent
+reference: bug 1616900
+---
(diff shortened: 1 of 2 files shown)


djmitche:
It's mostly subtleties around the tests, vs anything that would be important in production use, so that's fine :)

PR #2390, opened 2020-02-21
[Bug 1522154] Add support for matrix notifications

imbstack:
This fairly closely mirrors the irc support already built in to taskcluster. It supports a bit richer of an api because matrix supports a slightly richer api as well. I've been testing this on my dev env and it seems to work. Probably worth doing a short security review before we land this for real.

As for our next steps to deploy this if it looks good:

> bstack: we are hoping to get the taskcluster notifications for completed builds and such things cut over from irc before it is shut down. is there a recommended way to set up clients here for more "official" bots like this rather than something running on a dev server somewhere?

> mhoye: I'm working up the bots doc right now. Short version is, get a service account (bugzilla -> infra & ops -> matrix)  and make a note of it on the bots page.

I have a bot account that I accidentally grandfathered in during a configuration mishap called `@taskcluster-community:mozilla.org`. Perhaps we will just transfer ownership of that to cloudops or something.

Bugzilla Bug: [1522154](https://bugzilla.mozilla.org/show_bug.cgi?id=1522154)


services/notify/test/api_test.js
```diff
@@ -185,6 +187,39 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure', 'aws'], function(mock, s
...
+      assert.equal(helper.matrixClient.sendEvent.callCount, 1);
+      assert.equal(helper.matrixClient.sendEvent.args[0][0], '!rejected:baz.com');
+    }
+  });
+
+  test('matrix (denylisted)', async function() {
+    await helper.apiClient.addDenylistAddress({notificationType: 'matrix-room', notificationAddress: '!foo:baz.com'});
+    try {
+      await helper.apiClient.matrix({body: 'Does this work?', roomId: '!foo:baz.com'});
+      throw new Error('should have failed');
```
djmitche: `assert.rejects` can be helpful here :)

ui/docs/manual/deploying/matrix.mdx
```diff
@@ -0,0 +1,32 @@
...
+which we'll copy here for posterity:
+
+0. In a private/incognito browser window, open Riot.
+0. Log in to the account you want to get the access token for, such as the bot's account.
+0. Click on the bot's name in the top left corner then "Settings".
+0. (Optional) Set your bot's display name and avatar.
+0. Click the "Help & About" tab (left side of the dialog).
+0. Scroll to the bottom and click the <click to reveal> part of Access Token: <click to reveal>.
+0. Copy your access token to a safe place, like the bot's configuration file.
+0. Do not log out. Instead, just close the window. If you used a private browsing session, you should be able to still use Riot for your own account. Logging out deletes the access token from the server, making the bot unable to use it.
```
ajvb: non-blocking: `Logging out deletes the access token from the server, making the bot unable to use it.` - why is that?

PR #2391, opened 2020-02-21
Add docs on how to set up a working github app for development

owlishDeveloper:
If these docs were here before, I'd set up my app way faster. So adding them!

### DIFF
--- a/dev-docs/development-process.md
+++ b/dev-docs/development-process.md
@@ -183,6 +183,31 @@ errorConfig:
 
 in order to report your errors to a sentry project.
 
+### Setting up a taskcluster app in your Development Cluster
+
+You will need:
+1. Development cluster up and running (see above)
+2. A github app created and installed for a testing repository.
+
+To set up a taskcluster-github app:
+0. In the settings of the github app that you created, at the very bottom of the General tab, you will find Generate Private Key button. 
+Press it to generate the private key.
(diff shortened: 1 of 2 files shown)


dev-docs/development-process.md
```diff
@@ -183,6 +183,31 @@ errorConfig:
...
+You will need:
+1. Development cluster up and running (see above)
+2. A github app created and installed for a testing repository.
+
+To set up a taskcluster-github app:
+0. In the settings of the github app that you created, at the very bottom of the General tab, you will find Generate Private Key button. 
+Press it to generate the private key.
+1. In your `dev-config.yml`, in the `github` section, add `github_private_pem` - you can copy-paste the contents of the 
+PEM file you have obtained in the previous step. Be careful to remove any newlines from the encrypted part,
+and the necessary newlines after the header and before the footer should be replaced with `\n`, so the whole thing is a string of one long line.
```
djmitche: maybe an example would help here?

dev-docs/development-process.md
```diff
@@ -183,6 +183,31 @@ errorConfig:
 
 in order to report your errors to a sentry project.
 
+### Setting up a taskcluster app in your Development Cluster
```
helfi92: Should this be "Taskcluster Github app", rather than taskcluster app?

PR #2407, opened 2020-02-24
Bug 1600071: Fix overprovisioning for workers that are slow to boot up

walac:
For workers that are booting, worker-manager considers them as "ready"
and possibly running tasks, and it ignores the fact that they will
eventually claim the current pending tasks. This is not a problem if the
instances of these workers boot relatively quickly.

If the instances take a long time to get ready (like baremetal
instances), worker-manager will keep spawning more instances while there
are pending tasks and until we reach the maximum capacity, causing an
overprovisioning problem.

We fix this by subtracting the number of instances in the REQUESTED
state from the total number to spawn.

### DIFF
--- a/services/worker-manager/src/providers/provider.js
+++ b/services/worker-manager/src/providers/provider.js
@@ -47,7 +47,7 @@ class Provider {
   async prepare() {
   }
 
-  async provision({workerPool, existingCapacity}) {
+  async provision({workerPool, workerInfo}) {
   }
 
   async deprovision({workerPool}) {
--- a/services/worker-manager/src/provisioner.js
+++ b/services/worker-manager/src/provisioner.js
@@ -118,13 +118,18 @@ class Provisioner {
       const providersByPool = new Map();
       const seen = worker => {
         const v = providersByPool.get(worker.workerPoolId);
+        const isRequested = worker.state === this.Worker.states.REQUESTED;
+        // compute the number of instances that have not yet called "registerWorker"
+        const requestedCapacity = isRequested ? worker.capacity : 0;
         if (v) {
           v.providers.add(worker.providerId);
-          v.capacity += worker.capacity;
+          v.existingCapacity += worker.capacity;
+          v.requestedCapacity += requestedCapacity;
         } else {
           providersByPool.set(worker.workerPoolId, {
             providers: new Set([worker.providerId]),
-            capacity: worker.capacity,
+            existingCapacity: worker.capacity,
+            requestedCapacity,
           });
         }
       };
@@ -158,10 +163,18 @@ class Provisioner {
           }
           poolsByProvider.get(providerId).add(workerPoolId);
 
-          const providerByPool = providersByPool.get(workerPoolId) || {providers: new Set(), capacity: 0};
+          const providerByPool = providersByPool.get(workerPoolId) || {
+            providers: new Set(),
+            existingCapacity: 0,
+            requestedCapacity: 0,
+          };
 
           try {
-            await provider.provision({workerPool, existingCapacity: providerByPool.capacity});
+            const workerInfo = {
+              existingCapacity: providerByPool.capacity,
+              requestedCapacity: providerByPool.requested,
+            };
+            await provider.provision({workerPool, workerInfo});
           } catch (err) {
             this.monitor.reportError(err, {providerId: workerPool.providerId}); // Just report this and move on
           }
--- a/services/worker-manager/test/provisioner_test.js
+++ b/services/worker-manager/test/provisioner_test.js
@@ -48,7 +48,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
                 msg => msg.Type === 'test-provision' && msg.Fields.workerPoolId === wt.workerPoolId), {
                 Logger: `taskcluster.worker-manager.provider.${pId}`,
                 Type: 'test-provision',
-                Fields: {workerPoolId: wt.workerPoolId, existingCapacity: wt.existingCapacity},
+                Fields: {workerPoolId: wt.workerPoolId, workerInfo: {}},
                 Severity: LEVELS.notice,
               });
           }));
@@ -171,6 +171,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
       workers: [
         {
           workerPoolId: 'ff/ee',
+          existingCapacity: 0,
           workerGroup: 'whatever',
           workerId: 'testing-OLD',
           providerId: 'testing1',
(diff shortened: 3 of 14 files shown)


walac:
The PR in the current state is more proof of concept than a true patch. It feels kind of hacky to put some estimation logic inside the provider, not to mention it only works for the AWS provider. I believe it makes more sense to move this logic inside the estimator.

walac:
> It looks like you've identified the issue here. Let me explain it in my own words to make sure I understand..

You got it! The right solution would involve the worker-manager knowing how many tasks each worker is currently running, but as it is not currently possible, I think this a good enough solution.

services/worker-manager/src/provisioner.js
```diff
@@ -136,6 +137,16 @@ class Provisioner {
...
 
+      // compute the number of instances that have not yet called "registerWorker"
+      await this.Worker.scan({
+        state: Entity.op.equal(this.Worker.states.REQUESTED),
+      }, {
+        handler: worker => {
+          const v = providersByPool.get(worker.workerPoolId);
+          v.requested += worker.capacity;
+        },
+      });
```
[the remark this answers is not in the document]
walac: I thought about that, may I read the `state` property as a normal Javascript object field or does it require a special invocation? This question is because I implemented a second scan.

services/worker-manager/src/provisioner.js
```diff
@@ -158,10 +169,14 @@ class Provisioner {
...
 
-          const providerByPool = providersByPool.get(workerPoolId) || {providers: new Set(), capacity: 0};
+          const providerByPool = providersByPool.get(workerPoolId) || {providers: new Set(), capacity: 0, requested: 0};
 
           try {
-            await provider.provision({workerPool, existingCapacity: providerByPool.capacity});
+            await provider.provision({
+              workerPool,
+              existingCapacity: providerByPool.capacity,
+              requestedCapacity: providerByPool.requested,
```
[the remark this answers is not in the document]
walac: That makes sense, I will implement that.

services/worker-manager/src/provisioner.js
```diff
@@ -125,6 +125,7 @@ class Provisioner {
           providersByPool.set(worker.workerPoolId, {
             providers: new Set([worker.providerId]),
             capacity: worker.capacity,
+            requested: 0,
```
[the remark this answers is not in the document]
walac: Ok

walac:
I updated the code based on @djmitche's feedback, but I don't understand why I needed to make the changed I made in `test/provisioner_test.js`

changelog/CZIWw6xmQ6Wl3LgvXaaljA.md
```diff
@@ -0,0 +1,3 @@
+level: patch
```
djmitche:
```
reference: bug 1600071
```

services/worker-manager/src/providers/provider.js
```diff
@@ -47,7 +47,7 @@ class Provider {
   async prepare() {
   }
 
-  async provision({workerPool, existingCapacity}) {
+  async provision({workerPool, workerInfo}) {
```
djmitche: There's some docs for this in `services/worker-manager/docs/providers.md` that should be updated, too.

Also, the parent class serves as a kind of documentation:
https://github.com/taskcluster/taskcluster/blob/3d4c26818fbdf5b2322f2a1286146a359f0b76d9/services/worker-manager/src/providers/provider.js#L50-L51

services/worker-manager/src/provisioner.js
```diff
@@ -158,10 +163,14 @@ class Provisioner {
           }
           poolsByProvider.get(providerId).add(workerPoolId);
 
-          const providerByPool = providersByPool.get(workerPoolId) || {providers: new Set(), capacity: 0};
+          const providerByPool = providersByPool.get(workerPoolId) || {providers: new Set(), capacity: 0, requested: 0};
```
djmitche:
```suggestion
          const providerByPool = providersByPool.get(workerPoolId) || {providers: new Set(), existingCapacity: 0, requestedCapacity: 0};
```

services/worker-manager/test/provisioner_test.js
```diff
@@ -48,7 +48,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
                 msg => msg.Type === 'test-provision' && msg.Fields.workerPoolId === wt.workerPoolId), {
                 Logger: `taskcluster.worker-manager.provider.${pId}`,
                 Type: 'test-provision',
-                Fields: {workerPoolId: wt.workerPoolId, existingCapacity: wt.existingCapacity},
+                Fields: {workerPoolId: wt.workerPoolId},
```
djmitche: This message is coming from here:
https://github.com/taskcluster/taskcluster/blob/3d4c26818fbdf5b2322f2a1286146a359f0b76d9/services/worker-manager/src/providers/testing.js#L29-L31
so you'll want to update that as well.

services/worker-manager/test/provisioner_test.js
```diff
@@ -86,6 +86,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
       workers: [
         {
           workerPoolId: 'ff/ee',
+          existingCapacity: 0,
```
djmitche: I don't think this belongs here, as this is a row to be added to the Workers table, and that has no "existingCapacity" column.  Have I misunderstood?

services/worker-manager/test/provisioner_test.js
```diff
@@ -429,7 +432,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
           msg => msg.Type === 'test-provision' && msg.Fields.workerPoolId === workerPool.workerPoolId), {
           Logger: `taskcluster.worker-manager.provider.${workerPool.input.providerId}`,
           Type: 'test-provision',
-          Fields: {workerPoolId: workerPool.workerPoolId, existingCapacity: 0},
+          Fields: {workerPoolId: workerPool.workerPoolId},
```
djmitche: The update to `testing.js` will fix this too (and the next few test cases)

services/worker-manager/src/provisioner.js
```diff
@@ -158,10 +163,18 @@ class Provisioner {
           }
           poolsByProvider.get(providerId).add(workerPoolId);
 
-          const providerByPool = providersByPool.get(workerPoolId) || {providers: new Set(), capacity: 0};
+          const providerByPool = providersByPool.get(workerPoolId) || {
+            providers: new Set(),
```
owlishDeveloper: how come this `Set` is empty?

djmitche: It's the default value if there's no workers found for that workerPoolId.

PR #2412, opened 2020-02-25
[Bug 1473155] Use new schema viewer component in SchemaTable

JiwoonKim:
Bugzilla Bug: [1473155](https://bugzilla.mozilla.org/show_bug.cgi?id=1473155)
replacing `react-schema-viewer` to `material-ui-json-schema-viewer` component

**Screenshot Results**
- **Before**

image

- **After**

  - **Default (Shrunk $ref)**
image

  - **Expanded $ref**
 image


helfi92:
Tests are failing. They're probably linting issues in the ui directory.

helfi92:
>  Could that description be expanded fully?

From my understanding (@JiwoonKim please correct me if I say something that's not correct), in order to have the row height in the left panel correspond to the height in the right panel, we need to know the exact number of lines or height a row in the right panel will take. Making the description fully expanded would make it a bit hard to do. Also, external factors like resizing the window would add to the problem since the row height would change. An alternative would have been to have the description in one line but the issue with that is that the user won't be able to properly view multi line descriptions and will also need to scroll the right panel to view the description which is not the best UX. Hover or click (for mobile users) was the middle ground we ended up with.

JiwoonKim:
> > Could that description be expanded fully?
> 
> From my understanding (@JiwoonKim please correct me if I say something that's not correct), in order to have the row height in the left panel correspond to the height in the right panel, we need to know the exact number of lines or height a row in the right panel will take. Making the description fully expanded would make it a bit hard to do. Also, external factors like resizing the window would add to the problem since the row height would change. An alternative would have been to have the description in one line but the issue with that is that the user won't be able to properly view multi line descriptions and will also need to scroll the right panel to view the description which is not the best UX. Hover or click (for mobile users) was the middle ground we ended up with.

Yes, @helfi92 is correct. For the left panel and right panel to have matching lines and rows, we need to calculate the number of lines or height of the row in advance, which is quite complicated. So instead of multi-line descriptions (which would require calculating the number of lines a description would need and also update when resizing the window), we decided that it would be best to use the tooltip instead.

djmitche:
Makes sense, thanks!

djmitche:
:cake: :balloon:

JiwoonKim:
> Before we merge, could you please update to 1.0.3? Thanks!

Updated to 1.0.3!

helfi92:
Filed https://github.com/taskcluster/taskcluster/issues/2441.

PR #2413, opened 2020-02-25
Bug 1617685 - do not block queue startup on EC2 Region loading

djmitche:
Amazingly, AWS has on several occasions failed to successfully host
their own list of IP address blocks.  With this change, the resolver
tries to load the list, but fails "softly": startup is allowed to
continue, and the resolver will just return "null" for all IPs until the
load is successful.

Bugzilla Bug: [1617685](https://bugzilla.mozilla.org/show_bug.cgi?id=1617685)


services/queue/src/ec2regionresolver.js
```diff
@@ -9,19 +9,59 @@ const AWS_IP_RANGES_URL = 'https://ip-ranges.amazonaws.com/ip-ranges.json';
...
-  async loadIpRanges() {
-    // Get IP ranges from AWS with really stupid retry logic
-    let {body} = await request.get(AWS_IP_RANGES_URL).catch(() => {
-      return request.get(AWS_IP_RANGES_URL).catch(() => {
-        return request.get(AWS_IP_RANGES_URL);
-      });
+  /**
+   * Start running the resolver.  This will return immediately while loading the
+   * list of IP ranges.  Until that load is complete, the resolver will return
+   * null (meaning not an EC2 region) for all IPs.
```
imbstack: The implication being that nothing will use cloudmirror while this is happening? That seems like a great default but let's add that note to this comment.

services/queue/src/ec2regionresolver.js
```diff
@@ -9,19 +9,59 @@ const AWS_IP_RANGES_URL = 'https://ip-ranges.amazonaws.com/ip-ranges.json';
...
+          await this._loadIpRanges();
+          return resolve();
+        } catch (err) {
+          this.monitor.warning(`Failed to download AWS IP ranges (retrying): ${err}`);
+          setTimeout(tryLoad, delay);
+          delay *= 2;
+        }
+      };
+      tryLoad();
     });
```
imbstack: I think we should throw an error after some max time. It can be very long if we want, perhaps 10 minutes even? Still would be nice to crash eventually since potentially this is a very costly issue.

Another option would be to change the warning to an error after 10 attempts or something?

services/queue/src/ec2regionresolver.js
djmitche: We never use cloudmirror (it's turned off).  But when the region data is in place we do hand out an S3 URL for requests from within the same region, and a CDN URL for everything else.

services/queue/src/ec2regionresolver.js
djmitche: I'm not sure what we would do armed with that information.  Crashing isn't great as it just expands the failure radius -- from a few too-expensive artifact URLs to a total failure of the queue service.

One option would be to cache a copy of the region information in the source, and default to that until we can successfully download the most up-to-date region information.  I can try that out..

imbstack: If it is really just a handful of too-expensive calls maybe we should just remove this feature for now?

djmitche: The feature is useful when cloudfront isn't down.. I just mean that during the time cloudfront is down, we would rather make a few too-expensive calls than crash.

imbstack:
other than tests and changelog of course :p

djmitche:
auth failure is just azure being azure

Issue #2408, opened 2020-02-24
could not terminate livelog exposure: close tcp [::]:xxxxx: use of closed network connection


djmitche:
https://community-tc.services.mozilla.com/tasks/SKpYULcgTwmvtVklJwy25g/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FSKpYULcgTwmvtVklJwy25g%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log
```
2020/02/24 18:29:36Z Created dir: /task_1582568020/taskcluster/workers/generic-worker/testdata/TestRunAfterUserCreation/task_1582568020/generic-worker
2020/02/24 18:29:36Z WARNING: Not purging previous task directories/users since config setting cleanUpTaskDirs is false
--- PASS: TestRunAfterUserCreation (0.05s)
FAIL
FAIL	github.com/taskcluster/taskcluster/v25/workers/generic-worker	159.888s
```

The failure occurred on the landing of https://github.com/taskcluster/taskcluster/pull/2389 but the landing after that one succeeded.

@petemoore can you take a look?

petemoore:
The real issue was little disguised, further up the log:

```
--- FAIL: TestPublicDirectoryArtifact (2.13s)
    helper_test.go:230: Scheduled task Z84gjoj3QVilvG9FmEYJ4A
    helper_test.go:243: Worker exited with exit code 0 as required.
    helper_test.go:319: Task Z84gjoj3QVilvG9FmEYJ4A resolved as completed/completed as required.
    artifacts_broken_on_docker_test.go:42: Was expecting 3 artifacts, but got 2
```

When go tests fail, I typically open the raw log and search for `FAIL:` and that usually finds it (as the failure can be a little buried).

The cause of that intermittent failure seems to be a problem the test had with livelogging. If you go to https://community-tc.services.mozilla.com/tasks/Z84gjoj3QVilvG9FmEYJ4A#artifacts you will see indeed there was no live log. :-/

petemoore:
Looks like several failed for the same reason:
```
--- FAIL: TestPublicDirectoryArtifact (2.13s)
--- FAIL: TestConflictingFileArtifactsInPayload (2.31s)
--- FAIL: TestFileArtifactTwiceInPayload (1.51s)
[...]

PR #2414, opened 2020-02-25
Avoid closing the listener to which an https.Server was attached

djmitche:
In #2408 I initially thought perhaps we were closing the exposure twice, so this also adds a commit to set `l.exposure` to nil after closing it, avoiding that particular situation.

Github Bug/Issue: Refs #2408


djmitche:
There's a benefit to keeping the changelog at a readable length, too, especially given the broad audience for it (users, administrators, deployers, and worker-deployers).  This only ever generated a warning, and only when using the local exposer, which is only used for testing.

PR #2417, opened 2020-02-26
Bug 1616922 - adapt TestFileArtifactTwiceInPayload to disallow same artifact twice

petemoore:
A [test failure](https://community-tc.services.mozilla.com/tasks/groups/PoP3IcGlTHyVUYov269yyw) crept in for `TestFileArtifactTwiceInPayload` in #2399 but went unnoticed due to [bug 1533235](https://bugzilla.mozilla.org/show_bug.cgi?id=1533235).

The failure is due to a slight change in behaviour introduced in #2399 that disallows duplicate artifacts being entered in a task payload. Previously it was allowed, as no conflict of information ensues from having identical entries. It was a [permissive approach](https://en.wikipedia.org/wiki/Robustness_principle) but I'm also ok with us being more restrictive, as it may help avoid bugs in downstream code that parse task payloads and assume that entries are unique.

### DIFF
--- a/changelog/aeki4OImR8iFYCwyo-Mr1A.md
+++ b/changelog/aeki4OImR8iFYCwyo-Mr1A.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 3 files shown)


djmitche:
(also needs a changelog)

PR #2423, opened 2020-02-27
Add a FreeBSD simple generic worker.

verm:
Add support for a simple generic worker on FreeBSD.

`build.sh -a` works fine for me with this patch and to build on FreeBSD:

  `GOOS=freebsd GOARCH=amd64 ./build.sh`


workers/generic-worker/runtime/runtime_freebsd.go
```diff
@@ -0,0 +1,8 @@
+// +build darwin linux freebsd
+
```
petemoore: This new file shouldn't be needed, as `OSUser` struct is already defined in `workers/generic-worker/runtime/runtime_posix.go` and has `freebsd` added as a build constraint.

workers/generic-worker/simple_posix.go
```diff
@@ -1,4 +1,4 @@
-// +build darwin,simple linux,simple
+// +build darwin,simple linux,simple,simple freebsd
```
petemoore:
```suggestion
// +build darwin,simple linux,simple freebsd,simple
```

workers/generic-worker/diskspace_posix.go
```diff
@@ -14,7 +14,7 @@ func freeDiskSpaceBytes(dir string) (uint64, error) {
 		return 0, err
 	}
 	// Available blocks * size per block = available space in bytes
-	b := stat.Bavail * uint64(stat.Bsize)
+	b := uint64(stat.Bavail) * uint64(stat.Bsize)
```
petemoore: What is the reason for this change?

verm: Without this change I get this error:

`./diskspace_posix.go:17:19: invalid operation: stat.Bavail * uint64(stat.Bsize) (mismatched types int64 and uint64)`

workers/generic-worker/runtime/runtime_freebsd.go
verm: You're right!  There must have been an artifact laying around as I went ahead and did an `rm -r`, rebuilt and got the error.  It's been a while since I've worked with Go forgive my rustiness.

workers/generic-worker/chain_of_trust_posix.go
```diff
@@ -1,4 +1,4 @@
-// +build multiuser,linux multiuser,darwin
+// +build multiuser,linux multiuser,darwin,freebsd
```
petemoore: This change isn't needed - chain of trust isn't supported in simple engine.

workers/generic-worker/diskspace_posix.go
petemoore: Fair enough, I guess this must be an inconsistency in implementation across platforms. Then this should be fine.

workers/generic-worker/chain_of_trust_posix.go
verm: Okay thank you for the clear explanation.  I looked at my test script and wasn't building multiuser.  I've changed it now to just use `build.sh -a` now for testing before submission sorry for the trouble.

PR #2406, opened 2020-02-24
Bug 1558240 - separate '=== Task Finished ===' from log output

djmitche:
Bugzilla Bug: [1558240](https://bugzilla.mozilla.org/show_bug.cgi?id=1558240)

PR #2424, opened 2020-02-27
Revert "Bug 1558240 - separate '=== Task Finished ===' from log output"

petemoore:
Reverts taskcluster/taskcluster#2406

[Bug 1558240](https://bugzil.la/1558240) was raised for docker-worker, so unfortunately this fix doesn't apply. Although generic-worker logs a similar line, it does not log it directly after the command output (it logs the exit status of the command after the command output). Also note that in generic-worker there can be several commands, so a fix would need to occur once per command rather than once per task. A unit test would have demonstrated this. 😜  Also note each log line is prefixed by a timestamp, so the fix would have left a trailing timestamp at the end of the command output too.

I'm also not sure a new line should be added, in the same way that if a command executed from a bash session completes without outputting a trailing newline character, the new bash prompt will appear directly after the output, without a new line. This informs the user that no trailing newline was present. It reflects what the command actually output, rather than what the shell thinks the command should have output. If a new line should be present, I think it is more appropriate for the chosen command to be modified to produce it, since the logger should be agnostic to the command output.

In any case, since the fix didn't work, and was for the wrong worker implementation, I'm rolling back.

### DIFF
--- a/changelog/bug-1558240.md
+++ b/changelog/bug-1558240.md
@@ -1,0 +1,4 @@
+level: patch
+reference: bug 1558240
+---
+The generic-worker logging change that appeared in v25.4.0 has been reverted.
--- a/workers/generic-worker/artifacts_multiuser_test.go
+++ b/workers/generic-worker/artifacts_multiuser_test.go
@@ -72,7 +72,6 @@ func TestChainOfTrustUpload(t *testing.T) {
 			Extracts: []string{
 				"hello world!",
 				"goodbye world!",
-				"",
 				"=== Task Finished ===",
 				"Exit Code: 0",
 			},
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -1153,7 +1153,7 @@ func (task *TaskRun) Run() (err *ExecutionErrors) {
 	started := time.Now()
 	defer func() {
 		finished := time.Now()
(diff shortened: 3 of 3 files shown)


petemoore:
> Fair enough. This does mess up the log parsing, but I suppose that's a bug in the task.

I'm open to us implementing a change if it is warranted, but I think this particular issue was for a docker-worker log. A similar approach could be used in generic-worker if really wanted, but then it needs to happen in a different place (in the exit status report) since that is what comes immediately after the command output.

I think if we did introduce an automatic line ending, we should log something to say that we added it, in a similar way to how`diff` would report it, e.g.:

```
pmoore@Petes-iMac:~ $ echo hello > A
pmoore@Petes-iMac:~ $ echo -n hello > B
pmoore@Petes-iMac:~ $ diff A B
1c1
< hello
---
> hello
\ No newline at end of file
```

So I could see us adding a new line, if we logged an extra message to say that we had added it

```
<task command output>....
[taskcluster xxxxxx] ^^^ no new-line character at end of command output
[taskcluster xxxxxx] .... usual log lines ....
```

I think though this may only serve to complicate the issue, so my preference is still that if the parser requires the command to have a trailing new line, that either the command is adjusted to have the new line, or the parser is adjusted not to require it...

PR #2425, opened 2020-02-27
Implement "shutdown" message

walac:
The "shutdown" message is implemented for AWS, Azure and GCP providers.
When received, worker-runner will call the worker-worker manager
RemoveWorker API. If it fails, it will fall back to the system
shutdown.

For Linux, the "system shutdown" is the execution of the command
"shutdown -h now". For Windows, it will call the Win32 API
InitiateSystemShutdownEx.

The static and standalone providers don't implement the "shutdown"
message.

### DIFF
--- a/tools/taskcluster-worker-runner/protocol/protocol.go
+++ b/tools/taskcluster-worker-runner/protocol/protocol.go
@@ -1,6 +1,9 @@
 package protocol
 
-import "sync"
+import (
+	"log"
+	"sync"
+)
 
 type MessageCallback func(msg Message)
 
@@ -12,6 +15,9 @@ type Protocol struct {
 	// to avoid finding the empty set at startup)
 	Capabilities *Capabilities
 
+	// Capabilities of the remote peer
+	RemoteCapabilities *Capabilities
+
 	// callbacks per message type
 	callbacks map[string][]MessageCallback
 
@@ -22,10 +28,11 @@ type Protocol struct {
 
 func NewProtocol(transport Transport) *Protocol {
 	return &Protocol{
-		transport:    transport,
-		Capabilities: EmptyCapabilities(),
-		callbacks:    make(map[string][]MessageCallback),
-		initialized:  false,
+		transport:          transport,
+		Capabilities:       EmptyCapabilities(),
+		RemoteCapabilities: EmptyCapabilities(),
+		callbacks:          make(map[string][]MessageCallback),
+		initialized:        false,
 		initializedCond: sync.Cond{
 			L: &sync.Mutex{},
 		},
@@ -55,11 +62,7 @@ func listOfStrings(val interface{}) []string {
 func (prot *Protocol) Start(asWorker bool) {
 	if asWorker {
 		prot.Register("welcome", func(msg Message) {
-			caps := FullCapabilities()
-			otherCaps := FromCapabilitiesList(listOfStrings(msg.Properties["capabilities"]))
-			caps.LimitTo(otherCaps)
-			prot.Capabilities = caps
-
+			prot.RemoteCapabilities = FromCapabilitiesList(listOfStrings(msg.Properties["capabilities"]))
 			prot.Send(Message{
 				Type: "hello",
 				Properties: map[string]interface{}{
@@ -70,15 +73,14 @@ func (prot *Protocol) Start(asWorker bool) {
 		})
 	} else {
 		prot.Register("hello", func(msg Message) {
-			prot.Capabilities = FromCapabilitiesList(listOfStrings(msg.Properties["capabilities"]))
+			prot.RemoteCapabilities = FromCapabilitiesList(listOfStrings(msg.Properties["capabilities"]))
 			prot.SetInitialized()
 		})
 
-		fullCaps := FullCapabilities()
 		prot.Send(Message{
 			Type: "welcome",
 			Properties: map[string]interface{}{
-				"capabilities": fullCaps.List(),
+				"capabilities": prot.Capabilities.List(),
 			},
 		})
 	}
@@ -124,9 +126,13 @@ func (prot *Protocol) recvLoop() {
 		if !ok {
 			return
 		}
-		callbacks := prot.callbacks[msg.Type]
-		for _, cb := range callbacks {
-			cb(msg)
+		callbacks, ok := prot.callbacks[msg.Type]
+		if ok {
+			for _, cb := range callbacks {
+				cb(msg)
+			}
+		} else {
+			log.Printf("No callback registered for message %s\n", msg.Type)
 		}
 	}
 }
--- a/tools/taskcluster-worker-runner/provider/aws/awsprovider.go
+++ b/tools/taskcluster-worker-runner/provider/aws/awsprovider.go
@@ -132,9 +132,18 @@ func (p *AWSProvider) checkTerminationTime() {
 	}
 }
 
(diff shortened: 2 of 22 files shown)


tools/taskcluster-worker-runner/provider/azure/azure.go
```diff
@@ -152,6 +160,12 @@ func (p *AzureProvider) checkTerminationTime() bool {
 }
 
 func (p *AzureProvider) WorkerStarted() error {
+	p.proto.Register("shutdown", func(msg protocol.Message) {
+		if err := provider.RemoveWorker(p.workerManager, &p.workerInfo); err != nil {
+			log.Printf("Shutdown error: %v\n", err)
+		}
+	})
```
djmitche: Need to add the Capability here :)

tools/taskcluster-worker-runner/protocol/protocol.go
```diff
@@ -124,9 +127,13 @@ func (prot *Protocol) recvLoop() {
...
-		callbacks := prot.callbacks[msg.Type]
-		for _, cb := range callbacks {
-			cb(msg)
+		callbacks, ok := prot.callbacks[msg.Type]
+		if ok {
+			for _, cb := range callbacks {
+				cb(msg)
+			}
+		} else {
+			log.Printf("No callback registered for message %s\n", msg.Type)
```
djmitche: nice :)

tools/taskcluster-worker-runner/provider/aws/awsprovider.go
```diff
@@ -53,6 +55,8 @@ func (p *AWSProvider) ConfigureRun(state *run.State) error {
 		return fmt.Errorf("Could not create worker manager client: %v", err)
 	}
 
+	p.workerManager = wm
```
djmitche: This is an un-authenticated client, so it won't be able to call RemoveWorker.  I think you'll need to re-invoke `p.workerManagerClientFactory` when shutting down, passing it the worker credentials.

tools/taskcluster-worker-runner/provider/provider/common.go
```diff
@@ -40,3 +46,19 @@ func RegisterWorker(state *run.State, wm tc.WorkerManager, workerPoolID, provide
...
+func RemoveWorker(wc tc.WorkerManager, wi *WorkerInfo) error {
+	err := wc.RemoveWorker(wi.WorkerPoolID, wi.WorkerGroup, wi.WorkerID)
+	if err != nil {
+		log.Printf("Error removing the worker: %v\n", err)
+		log.Printf("Falling back to system shutdown")
+		if err = Shutdown(); err != nil {
+			log.Printf("Error shutting down the worker: %v\n", err)
+			return err
+		}
+	}
```
djmitche: Would it make sense to always do both?

tools/taskcluster-worker-runner/provider/aws/awsprovider.go
walac: How can it call `RegisterWorker` then?

tools/taskcluster-worker-runner/provider/provider/common.go
walac: Initially, I thought about that. But in the current state, if we see any instance in the "stopped", we know something went wrong with `RemoveWorker`, and can dig into the logs to investigate it.

tools/taskcluster-worker-runner/provider/aws/awsprovider.go
djmitche: With this many moving parts, I hardly think that qualifies as stupid :)

tools/taskcluster-worker-runner/provider/provider/common.go
djmitche: Makes sense!

tools/taskcluster-worker-runner/provider/aws/awsprovider.go
walac: Done

tools/taskcluster-worker-runner/protocol/protocol.go
```diff
@@ -12,6 +15,9 @@ type Protocol struct {
 	// to avoid finding the empty set at startup)
 	Capabilities *Capabilities
 
+	// Capabilities of the remote peer
+	RemoteCapabilities *Capabilities
```
djmitche: I like splitting this into "our" capabilities and the remote system's capabilities.  However, this loses the effect of intersecting the two sets of capabilities to find out what both ends have in common.  So for example a worker using this protocol would unconditionally list `"shutdown"` in its capabilities, and call `proto.Capable("shutdown")` to find out whether it's supported, which would return true.  Even if the worker-runner it was talking to did not support shutdown.

Maybe `proto.Capable` should check both sets of capabilities and && the results?

tools/taskcluster-worker-runner/protocol/protocol.go
```diff
@@ -22,10 +28,11 @@ type Protocol struct {
 
 func NewProtocol(transport Transport) *Protocol {
 	return &Protocol{
-		transport:    transport,
-		Capabilities: EmptyCapabilities(),
-		callbacks:    make(map[string][]MessageCallback),
-		initialized:  false,
+		transport:          transport,
+		Capabilities:       FullCapabilities(),
```
djmitche: I think some of the misunderstanding in the last review here was caused by this default to all capabilities (the default used to be in the `welcome` callback, below, but was defaulting to FullCapabilities).  As it stands, `Capabilities` will always contain `"shutdown"` because of this default, regardless of the `Add` calls in the providers.

Would it make more sense to default to *no* capabilities and then require other code to call `Add` for the relevant capabilities before starting the protocol?  Or maybe requiring that the capabilities be passed to `NewProtocol`?

walac: I don't understand, doesn't `FullCapabilities` return the capabilities listed in `knownCapabilities`? How could `shutdown` always be added?

tools/taskcluster-worker-runner/protocol/protocol.go
walac: Makes sense.

tools/taskcluster-worker-runner/protocol/protocol.go
djmitche: Both of those names are intended to mean "all of the capabilities that exist".  My mistake was thinking that any application using this library would naturally support all of those capabilities -- that's definitely not the case, as here where *sometimes* tc-worker-runner doesn't support `"shutdown"`.

So I'm thinking that the simplest to understand would be to default to no capabilities and require the user of the library to "Add" every capability they do support to the protocol, before starting it.

walac: Ok, so can I get rid of `knownCapabilities`?

djmitche: Yeah.  I think I meant to add a check in Add that the capability was in knownCapabilities, but I forgot :)

tools/taskcluster-worker-runner/protocol/protocol.go
walac: Hrm, I gave a further thought on this. Not necessarily an unsupported capability of the worker will prevent worker-runner from making it available. `shutdown` is such an example. I believe this level of graining check should be a layer upper.

PR #2426, opened 2020-02-27
Bug 1559434 - encode slashes in pulse passwords

djmitche:
Bugzilla Bug: [1559434](https://bugzilla.mozilla.org/show_bug.cgi?id=1559434)


imbstack:
although it looks like maybe our tests are enforcing incorrect behavior?

djmitche:
will check

PR #2429, opened 2020-03-01
Add support for DB downgrades

djmitche:
This adds support for `yarn db:downgrade <version>`.  It's documented with not one, not two, but *three* Warning components!  I strongly suspect that this will rarely, if ever, be used, and I think the warnings are sensible in that context.  However, since it's seldom used, we'll need to be careful to test it thoroughly!

I know that work on using tc-lib-entities is in progress, but note that this fixes an issue where `db/test/*_test.js` weren't running, and that identifies some missing fake functions.  And note, too, that `db/test/versions/0002_test.js` will need to test downgrading as well.

One cool thing that gives me a good bit of confidence here is that this runs a full upgrade / downgrade cycle -- twice, in fact -- thus verifying that all downgrade scripts get run and that at least for the no-data case they do the right thing.


db/versions/0001.yml
```diff
@@ -6,6 +6,11 @@ migrationScript: |-
     );
     grant select, insert, update, delete on widgets to $db_user_prefix$_notify;
   end
+downgradeScript: |-
+  begin
+    revoke select, insert, update, delete on widgets from $db_user_prefix$_notify;
+    drop table widgets;
```
helfi92: Would it be sensible to have a test that makes sure the number of grants/revokes in the migration script is equal to the number of revokes/grants?

libraries/postgres/README.md
```diff
@@ -87,6 +87,18 @@ For example, if an upgrade factors a single table into two tables, then a functi
 A consequence of this design is that "functions are forever" -- an upgrade can never delete a stored function.
 At worst, when a feature is removed, a stored function can be rewritten to return an empty result or perform no action.
 
+#### Downgrades
+
+It is sometimes necessary to roll back a deployment of Taskcluster services, due to an unexpected issue.
+In most cases, the database compatibility model meanas that this can be done without changing the database itself: by design, old code can run against the new database.
```
helfi92: `s/meanas/means/`

libraries/postgres/src/Database.js
```diff
@@ -120,6 +120,50 @@ class Database {
     }
   }
 
+  /**
+   * Downgrade this database to the given version and define functions for all
+   * of the methods in that version.  Note that this does not remove unknown
+   * functions.
```
helfi92: What unknown functions is this referring to?

djmitche: For example if a new stored function is defined in version 7, then downgrading to version 6 won't delete it.

db/versions/0001.yml
djmitche: I think a more reliable thing might be to validate that after downgrading to version 0, all grants are gone.  I'll see if I can cook that up.

djmitche:
The taskcluster-db tests are failing because there are no fake versions of the tc-lib-entities functions -- that test script is not running on the tc-lib-postgres branch, so that test failure's not visible there.

PR #2436, opened 2020-03-02
Try upgrading go dependencies

petemoore:
Renovate bot sends us occasional update requests (#2432 and #2433 are outstanding at the moment), but doesn't run `go mod tidy` afterwards so we always get failures. So I thought I'd see if I could tackle upgrading all dependencies in one go. Just for the fun of it.


djmitche:
It's worth checking to see if that's a renovate option, and filing a bug with renovate to see if they can add that support if it's not already present.  We'll be getting updates from renovate at least every Monday, and probably more often.

go.mod
```diff
@@ -41,20 +41,14 @@ require (
...
-	github.com/xeipuuv/gojsonschema v1.1.0
-	golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550
-	golang.org/x/net v0.0.0-20191002035440-2ec189313ef0
-	golang.org/x/sys v0.0.0-20200302083256-062a44052db1
-	golang.org/x/tools v0.0.0-20200219161401-5fb17a1e7b9b
+	github.com/ulikunitz/xz v0.5.7 // indirect
+	github.com/xeipuuv/gojsonschema v1.2.0
+	golang.org/x/crypto v0.0.0-20200221231518-2aa609cf4a9d
+	golang.org/x/net v0.0.0-20200301022130-244492dfa37a
+	golang.org/x/sys v0.0.0-20200302150141-5c8b2ff67527
```
djmitche: This caused issues in #2440, so I'm guessing you don't want to upgrade it?

petemoore: Let's close #2440 in favour of this PR, since #2440 is more systematic.

petemoore:
> This looks good aside from x/sys.

I think the `x/sys` change here is fine, if we don't land #2440.

> I have another PR that will exempt go.mod/go.sum from the changelog requirement, saving you having to make a silent changelog.

That is a great idea, many thanks.

> Can you link the Renovate issue here, too?

https://github.com/renovatebot/renovate/issues/5631

Hexcles:
FWIW, even though we did enable `gomodTidy`, we still find Renovate sometimes miss updates. I'm not yet sure what's going on, but one theory is that based on its [docs](https://docs.renovatebot.com/golang/), it's not actually using `go get -u` but parsing `go.mod` instead.

djmitche:
Thanks -- good to know. At this point we'll get good notice of any issues with `go.sum` needing update in the `yarn generate` task, and can both notice the issue and easily add a commit to the PR to update it.

PR #2438, opened 2020-03-02
Bug 1618983 - support reregistrationTimeout in static provider

djmitche:
Bugzilla Bug: [1618983](https://bugzilla.mozilla.org/show_bug.cgi?id=1618983)

This will allow static workers to run for 31 days, if configured that way.  And if they reboot in that time, then they can run forever.

### DIFF
--- a/services/worker-manager/src/providers/static.js
+++ b/services/worker-manager/src/providers/static.js
@@ -1,3 +1,4 @@
+const taskcluster = require('taskcluster-client');
 const {ApiError, Provider} = require('./provider');
 const {Worker} = require('../data');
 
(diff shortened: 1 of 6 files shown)


services/worker-manager/src/providers/static.js
```diff
@@ -55,11 +56,21 @@ class StaticProvider extends Provider {
...
+    }
+
+    if (staticSecret !== worker.providerData.staticSecret) {
+      throw new ApiError('bad staticSecret in workerIdentityProof');
+    }
+
+    let expires = taskcluster.fromNow('96 hours');
+    const {reregistrationTimeout} = Provider.interpretLifecycle(workerPool.config);
+    if (reregistrationTimeout) {
+      expires = new Date(Date.now() + reregistrationTimeout);
```
walac:
```suggestion
    let expires;
    const {reregistrationTimeout} = Provider.interpretLifecycle(workerPool.config);
    if (reregistrationTimeout) {
      expires = new Date(Date.now() + reregistrationTimeout);
    } else {
      expires = taskcluster.fromNow('96 hours');
    }
```

PR #2439, opened 2020-03-03
Bug 1596177 - remove support for legacy priority-free scopes

djmitche:
Bugzilla Bug: [1596177](https://bugzilla.mozilla.org/show_bug.cgi?id=1596177)

@tomprince checked, and I verified, that there are no such scopes in any role in either deployment, so I *think* this is safe to land.  But, we'll find out!


services/queue/src/api.js
```diff
@@ -775,29 +752,12 @@ builder.declare({
...
-          {
-            AllOf: [
-              'queue:define-task:<provisionerId>/<workerType>',
-              'queue:task-group-id:<schedulerId>/<taskGroupId>',
-            ],
-          },
-        ]},
+        for: 'priority',
+        in: 'priorities',
+        each: 'queue:create-task:<priority>:<provisionerId>/<workerType>',
```
imbstack: While we're poking around in here are we ready to do bug 1455632 at this point? It would be kinda neat if we enforced `APIBuilder.stability.deprecated` needing a bug number associated with it somehow.

djmitche: I suspect that yes, we could remove defineTask as well, which would close that bug up (we want to keep rerunTask).

And yeah, that sounds like a good contributor project.

imbstack: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1619808

djmitche:
The generic-worker failures are due to `go.sum` issues due to the master revision on which this PR is based.

PR #2440, opened 2020-03-03
Revert "Update golang.org/x/sys commit hash to 062a440"

petemoore:
Reverts taskcluster/taskcluster#2430

There were failures due to this change in the CI.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,7 +46,7 @@ require (
 	github.com/xeipuuv/gojsonschema v1.1.0
 	golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550
 	golang.org/x/net v0.0.0-20191002035440-2ec189313ef0
-	golang.org/x/sys v0.0.0-20200302083256-062a44052db1
+	golang.org/x/sys v0.0.0-20191218084908-4a24b4065292
 	golang.org/x/tools v0.0.0-20200219161401-5fb17a1e7b9b
 	gopkg.in/tylerb/graceful.v1 v1.2.15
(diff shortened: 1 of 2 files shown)


petemoore:
> What were the failures?

See the [CI results](https://community-tc.services.mozilla.com/tasks/groups/VwyqhY4wTiaQ1E0fuMg64g) of #2430. I believe you didn't check the task graph results which we currently need to do, due to [bug 1533235](https://bugzilla.mozilla.org/show_bug.cgi?id=1533235).

The specific failure caused by #2430 was due to the issue we discussed yesterday in https://github.com/taskcluster/taskcluster/pull/2436#issue-382479787 (`go mod tidy` isn't run by renovate). I haven't yet reported this upstream, I will do that later today.

However, I suspect this may be resolved in the general case (i.e. no renovate changes needed) when we upgrade to go 1.14 which came out a couple of days ago. Under go 1.14 we can use a separate mod file for dev tools. See https://github.com/golang/go/issues/30515#issuecomment-581982071 for context.

petemoore:
Let's close this PR since it is superseded by #2436 in any case (and it will only cause conflicts).

djmitche:
In cases where renovate doesn't make all the necessary changes (for example, it doesn't run `pipupdate` for tc-admin), we typically add a commit to the PR to make those additional changes, and then merge the resulting two-commit PR.  That's easier to remember to do when there's a red X!

We could add a `go mod tidy` run to `yarn generate`, making such an update pretty easy (`git ghpull -f $PR; yarn generate; git push`).

PR #2444, opened 2020-03-03
Improve web server login logging

djmitche:


services/web-server/src/login/strategies/mozilla-auth0.js
```diff
@@ -64,7 +64,8 @@ module.exports = class MozillaAuth0 {
     const user = new User();
 
     if (!userProfile) {
-      this.monitor.debug('User profile not found', {
+      this.monitor.warning({
```
imbstack: Should we make these documented log messages?

djmitche: I can if you'd like -- I'm ambivalent :)

djmitche: Please do so in a followup if you're inclined :)

PR #2446, opened 2020-03-03
run 'go mod tidy' in 'yarn generate'

djmitche:
Since I keep making this mistake :)


djmitche:
Hopefully this helps -- note that https://community-tc.services.mozilla.com/tasks/groups/GEYJQkqBRayUkpGcbzBvkg failed on go.sum changes without any actual golang changes involved (aside from string changes in the go client).

PR #2447, opened 2020-03-03
Add generic-worker support tiers to the documentation

djmitche:
As discussed in the weekly meeting.

### DIFF
--- a/ui/docs/reference/workers/generic-worker/support-tiers.mdx
+++ b/ui/docs/reference/workers/generic-worker/support-tiers.mdx
@@ -1,0 +1,44 @@
+---
+title: Support Tiers
+order: 35
+---
+
+# Support Tiers
+
+While Generic-Worker can run on a wide variety of platforms, the level of support for those platforms varies.
+Support levels are categorized into tiers, each bringing a different set of guarantees.
+This model is based on [Rust's support tiers](https://forge.rust-lang.org/release/platform-support.html).
+
+## Tier 1
+
+Tier one platforms are "guaranteed to work".
+That means that
+* They are tested in Taskcluster's continuous integration and tests must pass for a change to land.
+* Binaries are provided with each Taskcluster release.
+
+The Tier-1 platforms are:
+
+* OSX (GOOS darwin, GOARCH amd64)
+* 64-bit Linux (GOOS linux, GOARCH amd64)
+* 64-bit Linux for ARM (GOOS linux, GOARCH arm64)
+* 32-bit Linux for ARM (GOOS linux, GOARCH arm)
+* 64-bit Windows (GOOS windows, GOARCH amd64)
+* 32-bit Windows (GOOS windows, GOARCH 386)
+
+## Tier 2
+
+Tier 2 platforms are "guaranteed to build".
+That means that
+* The are built (generally cross-compiled) in Taskcluster's continuous integration
+* Tests are not run automatically
+* Binaries are not provided with each Taskcluster release
+
+There are currently no Tier-2 platforms.
+
+## Tier 3
+
+Tier 3 platforms are supported, but that support is not at all guaranteed.
+
+The Tier-3 platforms are:
+
+* FreeBSD (GOOS freebsd, GOARCH amd64)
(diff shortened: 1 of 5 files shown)


ui/docs/reference/workers/generic-worker/support-tiers.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+# Support Tiers
+
+While Generic-Worker can run on a wide variety of platforms, the level of support for those platforms varies.
+Support levels are categorized into tiers, each bringing a different set of guarantees.
+This model is based on [Rust's support tiers](https://forge.rust-lang.org/release/platform-support.html).
+
+## Tier 1
+
+Tier one platforms are "guaranteed to worl".
```
imbstack: "guaranteed to worl" sounds like these platforms are guaranteed to spiral out of control. Maybe we should keep the typo

djmitche: 'take 'em for a spin!"

ui/docs/reference/workers/generic-worker/support-tiers.mdx
```diff
@@ -0,0 +1,44 @@
...
+* Binaries are provided with each Taskcluster release.
+
+The Tier-1 platforms are:
+
+* OSX (GOOS darwin, GOARCH amd64)
+* 64-bit Linux (GOOS linux, GOARCH amd64)
+* 64-bit Linux for ARM (GOOS linux, GOARCH arm64)
+* 32-bit Linux for ARM (GOOS linux, GOARCH arm)
+* 64-bit Windows (GOOS windows, GOARCH amd64)
+* 32-bit Windows (GOOS windows, GOARCH 386)
```
petemoore: I think it is more meaningful to use the binary names since it covers the intended engine (different engines may have different support tiers) and also makes it foolproof to map it back to an actual binary release.

```suggestion
* generic-worker-multiuser-darwin-amd64
* generic-worker-multiuser-linux-amd64
* generic-worker-multiuser-windows-386
* generic-worker-multiuser-windows-amd64
* generic-worker-simple-darwin-amd64
* generic-worker-simple-linux-amd64
* generic-worker-simple-linux-arm
* generic-worker-simple-linux-arm64
```

ui/docs/reference/workers/generic-worker/support-tiers.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+## Tier 2
+
+Tier 2 platforms are "guaranteed to build".
+That means that
+* The are built (generally cross-compiled) in Taskcluster's continuous integration
+* Tests are not run automatically
+* Binaries are not provided with each Taskcluster release
+
+There are currently no Tier-2 platforms.
```
petemoore: We currently build docker engine, but it doesn't pass all the tests (most of which have been disabled), so we should probably make it tier 2.

```suggestion
* generic-worker-docker-linux-amd64
```

ui/docs/reference/workers/generic-worker/support-tiers.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+## Tier 3
+
+Tier 3 platforms are supported, but that support is not at all guaranteed.
+
+The Tier-3 platforms are:
+
+* FreeBSD (GOOS freebsd, GOARCH amd64)
```
petemoore: Notably, here is the exception to the rule, there is currently no release called `generic-worker-multiuser-freebsd-amd64`, however, I think it is reasonably straightforward to recognise what it would be for, and if we do start releasing it, this is what we will call it, so I think using just the target name is ok. It is also what it will be called if you run `build.sh -a` locally.

```suggestion
* generic-worker-multiuser-freebsd-amd64
```

workers/generic-worker/build.sh
```diff
@@ -65,6 +65,8 @@ function install {
   GOOS="${2}" GOARCH="${3}" CGO_ENABLED=0 go build -o "$OUTPUT_DIR/generic-worker-${1}-${2}-${3}" -ldflags "-X main.revision=$(git rev-parse HEAD)" -tags "${1}" -v .
 }
 
+# NOTE: when changing this, also update
+# ui/docs/reference/workers/generic-worker/support-tiers.mdx
```
petemoore: Love it, thanks for adding this!

PR #2459, opened 2020-03-05
Bug 1616998 - worker-runner finishing touches

petemoore:
Bugzilla Bug: [1616998](https://bugzilla.mozilla.org/show_bug.cgi?id=1616998)

Work in progress. Will flag for review when ready.


petemoore:
Not quite ready for review yet, but I'm marking as such to see if that is the reason that the worker tasks aren't running the CI...

petemoore:
> This looks good in general, but I think the docs move is only partially complete -- `tools/worker-runner/README.md` now has dangling links to docs, and the reference bits don't look quite right:
> image
> 
> Let's land this as-is, but perhaps fix that up in a followup? I think it would make sense to move the README content to the docs, too, and having a README.mdx should help the docs menu render correctly.

Many thanks. OK I'll land this now, as you suggest, and continue with a follow up. 👍

PR #2460, opened 2020-03-05
Check for SQLi in calls to stored functions

### LINKED ISSUE #2455: Test for sqli in the js methods calling stored functions
https://github.com/taskcluster/taskcluster/blob/8a0efa84a93673a1daf012ef4e96a6be5a9f71c3/libraries/postgres/src/Database.js#L48-L66 is the only bit of dynamic SQL that's called in running services, and it's designed to avoid the possibility of SQLi.

Let's add some tests with "weird" strings that double-check it.

djmitche:
Github Bug/Issue: Fixes #2455


djmitche:
I'm open to other ideas for tests to add.  This addresses the inputs to the JS function.  Is there anything else?

PR #2464, opened 2020-03-05
bug 1591476: registerWorker() returns workerConfig

milescrabill:
Bugzilla Bug: [1591476](https://bugzilla.mozilla.org/show_bug.cgi?id=1591476)

### DIFF
--- a/changelog/bug-1591476.md
+++ b/changelog/bug-1591476.md
@@ -1,6 +1,4 @@
 level: patch
 reference: bug 1591476
 ---
-worker-manager's registerWorker() returns workerConfig, worker-runner azure fetches workerConfig from registerWorker()
-
-Note: other providers do not yet fetch workerConfig from registerWorker()
+worker-manager's `registerWorker()` now returns worker config, and worker-runner (for Azure and static providers, others coming soon) merges that configuration with other configuration sources.  This allows worker pools to include configuration for static workers, and allows Azure workers to fetch their config without referencing the non-functional customData instance metadata.
--- a/changelog/bug-1591476.md
+++ b/changelog/bug-1591476.md
@@ -1,0 +1,6 @@
+level: patch
+reference: bug 1591476
+---
+worker-manager's registerWorker() returns workerConfig, worker-runner azure fetches workerConfig from registerWorker()
+
+Note: other providers do not yet fetch workerConfig from registerWorker()
--- a/services/worker-manager/schemas/v1/register-worker-response.yml
+++ b/services/worker-manager/schemas/v1/register-worker-response.yml
@@ -36,7 +36,18 @@ properties:
     required:
       - accessToken
       - clientId
+  workerConfig:
+    type: object
+    title: Worker Config
+    additionalProperties: true
+    description: |
+      This value is supplied unchanged as the `workerConfig` property of the `taskcluster` instance metadata attribute.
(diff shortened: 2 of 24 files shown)


services/worker-manager/schemas/v1/register-worker-response.yml
```diff
@@ -36,7 +36,18 @@ properties:
...
+  workerConfig:
+    type: object
+    title: Worker Config
+    additionalProperties: true
+    description: |
+      This value is supplied unchanged as the `workerConfig` property of the `taskcluster` instance metadata attribute.
+      The expectation is that the worker will merge this information with configuration from other sources,
+      and this is precisely what [taskcluster-worker-runner](https://github.com/taskcluster/taskcluster-worker-runner) does.
+      This property must not be used for secret configuration, as it is visible both in the worker pool configuration and in the worker instance's metadata.
+      Instead, put secret configuration in the [secrets service](https://github.com/taskcluster/taskcluster-worker-runner#secrets).
```
djmitche: I think this needs updated, since in this case it's not in the metadata.  Same for the similar property in `config-static.yml`

services/worker-manager/schemas/v1/register-worker-response.yml
```diff
@@ -36,7 +36,18 @@ properties:
     required:
       - accessToken
       - clientId
```
djmitche: Add `workerConfig` here.  It's good that worker-runner is resilient to the field not being present, but we can let consumers written after this lands just assume it's present (and catch bugs where it's omitted).

tools/taskcluster-worker-runner/provider/aws/awsprovider.go
```diff
@@ -58,7 +58,7 @@ func (p *AWSProvider) ConfigureRun(state *run.State) error {
 		"signature": interface{}(instanceIdentityDocumentSignature),
 	}
 
-	err = provider.RegisterWorker(
```
djmitche: A comment here indicating that this temporarily ignores the config would be good (pointing to this bug # for followup?)

tools/taskcluster-worker-runner/provider/google/google.go
```diff
@@ -49,7 +49,7 @@ func (p *GoogleProvider) ConfigureRun(state *run.State) error {
 
 	workerIdentityProofMap := map[string]interface{}{"token": interface{}(proofToken)}
 
-	err = provider.RegisterWorker(state, wm, userData.WorkerPoolID, userData.ProviderID, userData.WorkerGroup, workerID, workerIdentityProofMap)
```
djmitche: same here (comment indicating followup)

changelog/bug-1591476.md
```diff
@@ -0,0 +1,6 @@
+level: patch
+reference: bug 1591476
+---
+worker-manager's registerWorker() returns workerConfig, worker-runner azure fetches workerConfig from registerWorker()
```
djmitche:
```suggestion
worker-manager's `registerWorker()` now returns worker config, and worker-runner (for Azure and static providers, others coming soon) merges that configuration with other configuration sources.  This allows worker pools to include configuration for static workers, and allows Azure workers to fetch their config without referencing the non-functional customData instance metadata.
```

djmitche:
(schema updates will need a `yarn generate` run)

PR #2466, opened 2020-03-05
Bug 1547731 - Add structured logging for web-server requests 

helfi92:
Bugzilla Bug: [1547731](https://bugzilla.mozilla.org/show_bug.cgi?id=1547731)

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1547731

### DIFF
--- a/services/web-server/src/main.js
+++ b/services/web-server/src/main.js
@@ -11,6 +11,7 @@ const { ApolloServer } = require('apollo-server-express');
 const { sasCredentials } = require('taskcluster-lib-azure');
 const taskcluster = require('taskcluster-client');
 const { Auth } = require('taskcluster-client');
+const slugid = require('slugid');
 const monitorManager = require('./monitor');
 const createApp = require('./servers/createApp');
 const formatError = require('./servers/formatError');
@@ -118,13 +119,30 @@ const load = loader(
     },
 
     httpServer: {
-      requires: ['app', 'schema', 'context'],
-      setup: ({ app, schema, context }) => {
+      requires: ['app', 'schema', 'context', 'monitor'],
+      setup: ({ app, schema, context, monitor }) => {
+        const logger = monitor.childMonitor('httpServer');
+        const monitorPlugin = {
+          // Fires whenever a GraphQL request is received from a client.
+          requestDidStart(requestContext) {
+            const { request } = requestContext;
+
+            if (request.operationName !== "IntrospectionQuery") {
+              logger.log.requestReceived({
+                operationName: request.operationName,
+                variables: request.variables,
+                query: request.query,
+                requestId: slugid.v4(),
+              });
+            }
+          },
+        };
         const server = new ApolloServer({
           schema,
           context,
           formatError,
           tracing: true,
+          plugins: [monitorPlugin],
         });
         const httpServer = createServer(app);
 
--- a/services/web-server/src/main.js
+++ b/services/web-server/src/main.js
@@ -130,7 +130,6 @@ const load = loader(
             if (request.operationName !== "IntrospectionQuery") {
               logger.log.requestReceived({
                 operationName: request.operationName,
-                variables: request.variables,
                 query: request.query,
                 requestId: slugid.v4(),
               });
--- a/services/web-server/src/main.js
+++ b/services/web-server/src/main.js
@@ -11,7 +11,6 @@ const { ApolloServer } = require('apollo-server-express');
 const { sasCredentials } = require('taskcluster-lib-azure');
 const taskcluster = require('taskcluster-client');
 const { Auth } = require('taskcluster-client');
-const slugid = require('slugid');
 const monitorManager = require('./monitor');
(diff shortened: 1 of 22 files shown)


services/web-server/src/main.js
```diff
@@ -118,13 +119,30 @@ const load = loader(
...
+          // Fires whenever a GraphQL request is received from a client.
+          requestDidStart(requestContext) {
+            const { request } = requestContext;
+
+            if (request.operationName !== "IntrospectionQuery") {
+              logger.log.requestReceived({
+                operationName: request.operationName,
+                variables: request.variables,
+                query: request.query,
+                requestId: slugid.v4(),
```
djmitche: Is there a way to thread this value through so that we can access it when making REST API requests?  (We can figure out how to pass it to API requests in a followup)

services/web-server/src/monitor.js
```diff
@@ -46,4 +46,21 @@ monitorManager.register({
...
+  title: 'Request Received',
+  type: 'request-received',
+  version: 1,
+  level: 'info',
+  description: 'A GraphQL request has been received.',
+  fields: {
+    operationName: `
+      The name of the graphql query performed. If the operation is anonymous
+      (i.e., the operation is query { ... } instead of query NamedQuery { ... })
+      , then operationName is null.`,
```
djmitche: I don't see `query` described here -- does that contain the query content?

services/web-server/src/monitor.js
```diff
@@ -46,4 +46,21 @@ monitorManager.register({
...
+    variables: 'Object that contains all of the variables passed to the GraphQL query.',
```
djmitche: This might contain secret data for a few queries -- notably login, and also getting secrets.  I'm not sure if it makes sense to try to denylist those specific variables (and risk missing some) or try to write a (huge!) allowlist, or just not log variables at all.  In the REST API, we have a [cleanPayload](https://github.com/taskcluster/taskcluster/tree/master/libraries/api#declaring-methods) function that we use to clean payloads for calls that contain secrets.  Maybe we could do something vaguely similar here, and then only define that function for login and secret?

services/web-server/src/monitor.js
helfi92: Added.

services/web-server/src/monitor.js
helfi92: Oh, you're absolutely right. I forgot to think about this. I think we can just avoid logging the variables at all. Forgetting to update this list in the future would put us in a bad position.

services/web-server/src/main.js
helfi92: Yes. We can have the UI send a request-id header that each graphql loader would have access to. Something like:

```diff
diff --git a/services/web-server/src/loaders/workerManager.js b/services/web-server/src/loaders/workerManager.js
index 62b031286..a7973c3f9 100644
--- a/services/web-server/src/loaders/workerManager.js
+++ b/services/web-server/src/loaders/workerManager.js
@@ -2,7 +2,8 @@ const DataLoader = require('dataloader');
 const sift = require('../utils/sift');
 const ConnectionLoader = require('../ConnectionLoader');
 
-module.exports = ({ workerManager }) => {
+module.exports = ({ workerManager }, isAuthed, rootUrl, monitor, strategies, cfg) => {
+  console.log('loader has acccess to request id: ', cfg.headers['request-id']);
   const WorkerManagerWorkerPoolSummaries = new DataLoader(queries =>
     Promise.all(
       queries.map(async ({ filter }) => {
diff --git a/ui/src/App/index.jsx b/ui/src/App/index.jsx
index 1ec445fb3..dddae58ac 100644
--- a/ui/src/App/index.jsx
+++ b/ui/src/App/index.jsx
@@ -29,6 +29,7 @@ import theme from '../theme';
 import introspectionQueryResultData from '../fragments/fragmentTypes.json';
 import { route } from '../utils/prop-types';
 import AuthController from '../auth/AuthController';
+import { v4 } from '../utils/slugid';
 
 const absoluteUrl = (url, overrides = {}) =>
   Object.assign(new URL(url, window.location), overrides).toString();
@@ -103,6 +104,7 @@ export default class App extends Component {
           headers: {
             ...headers,
             Authorization: `Bearer ${btoa(JSON.stringify(user.credentials))}`,
+            'request-id': v4(),
           },
         };
       }),
```

Let me know if this sounds good to you. I could open another PR that sends the request-id to each REST API request.

djmitche: I think it would be better to generate the request-id here in requestDidStart, and then pass it along to loaders (rather than trusting the HTTP client to send an appropriately-random requestId)

helfi92: That works too.

helfi92: I didn't know of a way to send the request id to the context when the request id is generated in the plugin (i.e., requestdidStart). Instead, I started generating the requestId inside the context in `createContext.js` which gets called on each request.

PR #2468, opened 2020-03-05
Add SQLi tests for the scan method

### LINKED ISSUE #2457: test entities scan methods for SQLi
The tc-lib-entities scan method takes a condition which is substituted into the resulting SQL query.  We've been careful to only allow datestamps there, a format which doesn't allow arbitrary text.  Let's verify that with some tests.

djmitche:
This adds tests and some comments about the dangers of the tc-lib-entities `scan` method.

Happily, no issues found :)

Github Bug/Issue: Fixes #2457

### DIFF
--- a/libraries/entities/test/entity_scan_test.js
+++ b/libraries/entities/test/entity_scan_test.js
@@ -100,7 +100,7 @@ helper.dbSuite(path.basename(__filename), function() {
       assert.equal(result.entries.length, 1);
       assert.deepEqual(result.entries[0].taskId, '9');
     });
-    test('retrieve documents (with date condition)', async function() {
+    test('retrieve documents (with equality date condition)', async function() {
       db = await helper.withDb({ schema, serviceName });
       const TestTable = configuredTestTable.setup({ tableName: 'test_entities', db, serviceName });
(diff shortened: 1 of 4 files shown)


libraries/entities/test/entity_scan_test.js
```diff
@@ -302,4 +318,114 @@ helper.dbSuite(path.basename(__filename), function() {
...
+      // ----------------------------------------
+      //  ' or 1=1; drop table test_entities; --
+      //
+      // Furthermore, replacing the usage of `quote_literal` with a vulnerable equivalent results
+      // in the following from this test, indicating a successful injection:
+      //
+      //   error: cannot open multi-query plan as cursor
+      //   WHERE: PL/pgSQL function test_entities_scan(text,text,text,integer,integer) line 31 at RETURN QUERY
+      //   CODE: 42P11
+      const res = await TestTable.scan({taskId: "' or 1=1; drop table test_entities; --"});
```
helfi92: What does `--` do?

djmitche: It's a comment.  I got that string from some "hax0r your way into websites!" guide.  I think the idea is just to comment out any additional SQL that the server might add after the user input.

PR #2477, opened 2020-03-09
Add output.text from an artifact

owlishDeveloper:
Bugzilla Bug: [1552323](https://bugzilla.mozilla.org/show_bug.cgi?id=1552323)
as well as the Finix feature request.

The general idea is that if people want custom task results in checkruns, they need:

* Arrange for an artifact with markdown-formatted output in `.md` format
* Put the artifact's name in the task definition's `extra.github`
* Taskcluster-github will post the contents of the artifact in the check run

### DIFF
--- a/changelog/bug-1552323.md
+++ b/changelog/bug-1552323.md
@@ -1,5 +1,5 @@
 level: minor
 reference: bug 1552323
 ---
-Fixes bug 1552323 and issue https://github.com/mozilla-mobile/fenix/issues/6760
-Adds ability to customize checks output in taskcluster-github Checks feature
+Adds ability to customize checks output in taskcluster-github Checks feature.
+Apart from the bug mentioned, fixes the issue https://github.com/mozilla-mobile/fenix/issues/6760
--- a/changelog/bug-1552323.md
+++ b/changelog/bug-1552323.md
@@ -1,0 +1,5 @@
+level: minor
+reference: bug 1552323
+---
+Fixes bug 1552323 and issue https://github.com/mozilla-mobile/fenix/issues/6760
+Adds ability to customize checks output in taskcluster-github Checks feature
--- a/services/github/src/constants.js
+++ b/services/github/src/constants.js
@@ -15,4 +15,5 @@ module.exports = {
     'intermittent-task': 'neutral', // queue status means: will be retried
   },
   CHECKRUN_TEXT: 'View task in Taskcluster',
+  CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME: 'customCheckRunText.md',
 };
--- a/services/github/src/constants.js
+++ b/services/github/src/constants.js
@@ -15,5 +15,5 @@ module.exports = {
     'intermittent-task': 'neutral', // queue status means: will be retried
   },
   CHECKRUN_TEXT: 'View task in Taskcluster',
-  CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME: 'customCheckRunText.md',
+  CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME: 'public/customCheckRunText.md',
 };
--- a/services/github/src/constants.js
+++ b/services/github/src/constants.js
@@ -1,0 +1,18 @@
+module.exports = {
+  CONCLUSIONS: { // maps status communicated by the queue service to github checkrun conclusions
+    /*eslint-disable quote-props*/
+    'completed': 'success',
+    'failed': 'failure',
+    'exception': 'failure',
+    'deadline-exceeded': 'timed_out',
+    'canceled': 'cancelled',
+    'superseded': 'neutral', // queue status means: is not relevant anymore
+    'claim-expired': 'failure',
+    'worker-shutdown': 'neutral', // queue status means: will be retried
+    'malformed-payload': 'action_required', // github status means "correct your task definition"
+    'resource-unavailable': 'failure',
+    'internal-error': 'failure',
+    'intermittent-task': 'neutral', // queue status means: will be retried
+  },
+  CHECKRUN_TEXT: 'View task in Taskcluster',
+};
--- a/services/github/src/constants.js
+++ b/services/github/src/constants.js
@@ -15,5 +15,5 @@ module.exports = {
     'intermittent-task': 'neutral', // queue status means: will be retried
   },
   CHECKRUN_TEXT: 'View task in Taskcluster',
-  CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME: 'public/customCheckRunText.md',
+  CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME: 'public/github/customCheckRunText.md',
 };
--- a/services/github/src/constants.js
+++ b/services/github/src/constants.js
@@ -1,6 +1,6 @@
 module.exports = {
   CONCLUSIONS: { // maps status communicated by the queue service to github checkrun conclusions
-    /*eslint-disable quote-props*/
+    /*eslint quote-props: ["error", "consistent-as-needed"]*/
     'completed': 'success',
     'failed': 'failure',
     'exception': 'failure',
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -491,7 +491,6 @@ async function statusHandler(message) {
   );
   try {
     const taskDefinition = await this.queueClient.task(taskId);
-    debug(`Result status. Got task build from DB and task definition for ${taskId} from Queue service`);
     const {customCheckRun} = (taskDefinition.extra && taskDefinition.extra.github) || {};
 
     let customCheckRunText = '';
@@ -551,9 +550,9 @@ async function statusHandler(message) {
         output: {
           title: `${this.context.cfg.app.statusContext} (${eventType.split('.')[0]})`,
           summary: `${taskDefinition.metadata.description}`,
-          text: `[${CHECKRUN_TEXT}](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n${customCheckRunText || ''}`,
+          text: `[${CHECKRUN_TEXT}](${taskGroupUI(this.context.cfg.taskcluster.rootUrl, taskGroupId)})\n${customCheckRunText || ''}`,
         },
-        details_url: taskGroupUI(this.context.cfg.taskcluster.rootUrl, taskGroupId),
+        details_url: taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId),
       });
 
       await this.context.CheckRuns.create({
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -496,21 +496,12 @@ async function statusHandler(message) {
 
     let customCheckRunText = '';
 
-    const scopedQueueClient = this.queueClient.use({
-      authorizedScopes: taskDefinition.scopes,
-      credentials: this.context.cfg.taskcluster.credentials,
-    });
     const url = this.queueClient
-      .buildUrl(scopedQueueClient.getArtifact, taskId, runId, textArtifactName);
+      .buildUrl(this.queueClient.getArtifact, taskId, runId, textArtifactName);
 
     let res = await utils.throttleRequest({url, method: 'GET'});
 
     if (res.status >= 400 && res.status !== 404) {
-      /*
-        Some possible errors:
-          - insufficient scopes: statusCode 403
-          - no artifact entity in DB, no artifact in S3: statusCode 404
-      */
       await this.createExceptionComment({
         debug,
         instGithub,
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -3,22 +3,7 @@ const libUrls = require('taskcluster-lib-urls');
 const yaml = require('js-yaml');
 const assert = require('assert');
 const {consume} = require('taskcluster-lib-pulse');
-
-const CONCLUSIONS = { // maps status communicated by the queue service to github checkrun conclusions
-  /*eslint-disable quote-props*/
-  'completed': 'success',
-  'failed': 'failure',
-  'exception': 'failure',
-  'deadline-exceeded': 'timed_out',
-  'canceled': 'cancelled',
-  'superseded': 'neutral', // queue status means: is not relevant anymore
-  'claim-expired': 'failure',
-  'worker-shutdown': 'neutral', // queue status means: will be retried
-  'malformed-payload': 'action_required', // github status means "correct your task definition"
-  'resource-unavailable': 'failure',
-  'internal-error': 'failure',
-  'intermittent-task': 'neutral', // queue status means: will be retried
-};
+const {CONCLUSIONS, CHECKRUN_TEXT} = require('./constants');
 
 /**
  * Create handlers
@@ -523,7 +508,7 @@ async function statusHandler(message) {
         output: {
           title: `${this.context.cfg.app.statusContext} (${eventType.split('.')[0]})`,
           summary: `${taskDefinition.metadata.description}`,
-          text: `[View task in Taskcluster](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n${customCheckRunText || ''}`,
+          text: `[${CHECKRUN_TEXT}](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n${customCheckRunText || ''}`,
         },
       });
     } else {
@@ -535,7 +520,7 @@ async function statusHandler(message) {
         output: {
           title: `${this.context.cfg.app.statusContext} (${eventType.split('.')[0]})`,
           summary: `${taskDefinition.metadata.description}`,
-          text: `[View task in Taskcluster](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n${customCheckRunText || ''}`,
+          text: `[${CHECKRUN_TEXT}](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n${customCheckRunText || ''}`,
         },
         details_url: taskGroupUI(this.context.cfg.taskcluster.rootUrl, taskGroupId),
       });
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -507,9 +507,8 @@ async function statusHandler(message) {
       if (res.status >= 400) {
         /*
           Some possible errors:
-            - insufficient scopes: statusCode 403, code InsufficientScopes
-            - no artifact entity in DB: statusCode 404, code ResourceNotFound
-            - no artifact in S3: statusCode 404, code UnknownError
+            - insufficient scopes: statusCode 403
+            - no artifact entity in DB, no artifact in S3: statusCode 404
         */
         await this.createExceptionComment({
           debug,
@@ -517,12 +516,12 @@ async function statusHandler(message) {
           organization,
           repository,
           sha,
-          error: new Error(res.response.error.text),
+          error: res.response.error,
         });
+        await this.monitor.reportError(res.response.error);
 
-        await this.monitor.reportError(res);
       } else if (res.status >= 200 && res.status < 300) {
-        customCheckRunText = res.text;
+        customCheckRunText = res.text.toString();
       }
     }
 
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -749,7 +749,7 @@ async function jobHandler(message) {
     if (err.code !== 'EntityAlreadyExists') {
       throw err;
     }
-    let build = await this.Builds.load({
+    let build = await this.context.Builds.load({
       taskGroupId,
     });
     assert.equal(build.state, groupState, `State for ${organization}/${repository}@${sha}
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -504,17 +504,29 @@ async function statusHandler(message) {
     `Attempting to update status of the checkrun for ${organization}/${repository}@${sha} (${taskState.conclusion})`,
   );
   try {
+    const taskDefinition = await this.queueClient.task(taskId);
+    debug(`Result status. Got task build from DB and task definition for ${taskId} from Queue service`);
+    const {checkRunArtifactName} = taskDefinition.extra.github;
+
+    let checkRunArtifact = '';
+    if (checkRunArtifactName) {
+      checkRunArtifact = (await this.queueClient.getArtifact(taskId, runId, checkRunArtifactName)).toString();
+    }
+
+
     if (checkRun) {
       await instGithub.checks.update({
         ...taskState,
         owner: organization,
         repo: repository,
         check_run_id: checkRun.checkRunId,
+        output: {
+          title: `${this.context.cfg.app.statusContext} (${eventType.split('.')[0]})`,
+          summary: `${taskDefinition.metadata.description}`,
+          text: `[View task in Taskcluster](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n${checkRunArtifact || ''}`,
+        },
       });
     } else {
-      const taskDefinition = await this.queueClient.task(taskId);
-      debug(`Result status. Got task build from DB and task definition for ${taskId} from Queue service`);
-
       const checkRun = await instGithub.checks.create({
         owner: organization,
         repo: repository,
@@ -523,9 +535,9 @@ async function statusHandler(message) {
         output: {
           title: `${this.context.cfg.app.statusContext} (${eventType.split('.')[0]})`,
           summary: `${taskDefinition.metadata.description}`,
-          text: `[Task group](${taskGroupUI(this.context.cfg.taskcluster.rootUrl, taskGroupId)})`,
+          text: `[View task in Taskcluster](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n${checkRunArtifact || ''}`,
         },
-        details_url: taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId),
+        details_url: taskGroupUI(this.context.cfg.taskcluster.rootUrl, taskGroupId),
       });
 
       await this.context.CheckRuns.create({
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -4,6 +4,7 @@ const yaml = require('js-yaml');
 const assert = require('assert');
 const {consume} = require('taskcluster-lib-pulse');
 const {CONCLUSIONS, CHECKRUN_TEXT} = require('./constants');
+const utils = require('./utils');
 
 /**
  * Create handlers
@@ -227,7 +228,7 @@ class Handlers {
     }
     let body = [
       '<details>\n',
-      '<summary>Submitting the task to Taskcluster failed. Details</summary>',
+      '<summary>Uh oh! Looks like an error! Details</summary>',
       '',
       errorBody, // already in Markdown..
       '',
@@ -495,14 +496,35 @@ async function statusHandler(message) {
 
     let customCheckRunText = '';
     if (customCheckRun && customCheckRun.textArtifactName) {
-      console.log('🧿', taskDefinition.scopes);
       const scopedQueueClient = this.queueClient.use({
         authorizedScopes: taskDefinition.scopes,
         credentials: this.context.cfg.taskcluster.credentials,
       });
-      customCheckRunText = (await scopedQueueClient.getArtifact(taskId, runId, customCheckRun.textArtifactName)).toString();
-    }
 
+      const url = this.queueClient.buildUrl(scopedQueueClient.getArtifact, taskId, runId, customCheckRun.textArtifactName);
+      let res = await utils.throttleRequest({url, method: 'GET'});
+
+      if (res.status >= 400) {
+        /*
+          Some possible errors:
+            - insufficient scopes: statusCode 403, code InsufficientScopes
+            - no artifact entity in DB: statusCode 404, code ResourceNotFound
+            - no artifact in S3: statusCode 404, code UnknownError
+        */
+        await this.createExceptionComment({
+          debug,
+          instGithub,
+          organization,
+          repository,
+          sha,
+          error: new Error(res.response.error.text),
+        });
+
+        await this.monitor.reportError(res);
+      } else if (res.status >= 200 && res.status < 300) {
+        customCheckRunText = res.text;
+      }
+    }
 
     if (checkRun) {
       await instGithub.checks.update({
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -495,7 +495,12 @@ async function statusHandler(message) {
 
     let customCheckRunText = '';
     if (customCheckRun && customCheckRun.textArtifactName) {
-      customCheckRunText = (await this.queueClient.getArtifact(taskId, runId, customCheckRun.textArtifactName)).toString();
+      console.log('🧿', taskDefinition.scopes);
+      const scopedQueueClient = this.queueClient.use({
+        authorizedScopes: taskDefinition.scopes,
+        credentials: this.context.cfg.taskcluster.credentials,
+      });
+      customCheckRunText = (await scopedQueueClient.getArtifact(taskId, runId, customCheckRun.textArtifactName)).toString();
     }
 
 
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -520,7 +520,10 @@ async function statusHandler(message) {
           sha,
           error: res.response.error,
         });
-        await this.monitor.reportError(res.response.error);
+
+        if (res.status < 500) {
+          await this.monitor.reportError(res.response.error);
+        }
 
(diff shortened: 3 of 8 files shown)


owlishDeveloper:
@tomprince can you take a look as well?

services/github/src/handlers.js
```diff
@@ -489,17 +489,29 @@ async function statusHandler(message) {
     `Attempting to update status of the checkrun for ${organization}/${repository}@${sha} (${taskState.conclusion})`,
   );
   try {
+    const taskDefinition = await this.queueClient.task(taskId);
+    debug(`Result status. Got task build from DB and task definition for ${taskId} from Queue service`);
+    const {checkRunArtifactName} = taskDefinition.extra.github;
+
+    let checkRunArtifact = '';
+    if (checkRunArtifactName) {
+      checkRunArtifact = (await this.queueClient.getArtifact(taskId, runId, checkRunArtifactName)).toString();
```
imbstack: We should not crash the server if they have configured a `checkRunArtifactName` but the artifact did not actually get uploaded. What happens in that case with this code?

services/github/src/handlers.js
```diff
@@ -489,17 +489,29 @@ async function statusHandler(message) {
...
     if (checkRun) {
       await instGithub.checks.update({
         ...taskState,
         owner: organization,
         repo: repository,
         check_run_id: checkRun.checkRunId,
+        output: {
+          title: `${this.context.cfg.app.statusContext} (${eventType.split('.')[0]})`,
+          summary: `${taskDefinition.metadata.description}`,
+          text: `[View task in Taskcluster](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n${checkRunArtifact}`,
```
imbstack: what does the check run artifact being on a newline here do? is it just the text name of the artifact?

services/github/src/handlers.js
owlishDeveloper: In a non-draft PR there will be try-catch of some sort :)

services/github/src/handlers.js
owlishDeveloper: no, it's not the name, it's meant to be the contents of the artifact. (I might wanna add some logic to make sure it's a string)

owlishDeveloper: I know that `getArtifact` returns you the actual file. My initial thinking was it might be a good idea to avoid this and try [something like this](https://github.com/taskcluster/taskcluster/pull/2450/files#diff-41e7ecfbb590e5f3bbddd300fbf410dbR971) instead, but Dustin advised to use `getArtifact`, as the size of these artifacts is probably fine to hold in memory.

So, as I said in initial comment to this PR, I am expecting an artifact that contains a string, and I will get the string from there (probably takes more than `toString()` call, but that's something I was hoping to hear in the feedback)

tomprince:
This looks good to me. I chatted with @owlishDeveloper about some bikeshedding and corner cases.

owlishDeveloper:
Also CC @tomprince

services/github/src/handlers.js
```diff
@@ -242,7 +228,7 @@ class Handlers {
     }
     let body = [
       '<details>\n',
-      '<summary>Submitting the task to Taskcluster failed. Details</summary>',
+      '<summary>Uh oh! Looks like an error! Details</summary>',
```
djmitche: Why this change?

ui/docs/reference/integrations/github/checks.mdx
```diff
@@ -0,0 +1,43 @@
...
+            customCheckRun:
+                textArtifactName: <artifact_name>
+    ```
+ to the definition of the task for which you want customize the check.
+* If it's OK with you that the artifact is public, make sure to start the name of your artifact with
+ `public/`, like this: `public/<file_name>` both when you create the artifact and reference it in your
+ task definition. In this case, you will not need any scopes.
+* If you don't want the artifact to be public, name your artifact simply `<file_name>` both when creating it
+ and referencing it in your task definition. In this case, you will need a `queue:get-artifact:<file_name>`
+ scope. Add it to the `scopes` section of the same task.
```
djmitche: It doesn't look like this is supported as currently written.

ui/docs/reference/integrations/github/checks.mdx
```diff
@@ -0,0 +1,43 @@
+# Taskcluster-Github Checks
+
+Taskcluster-Github supports [Github's Checks API](https://developer.github.com/v3/checks/).
+At the moment, the feature is in beta. If you would like to try it out:
+* Make sure you use `.taskcluster.yml` v1
+* Add `reporting: checks-v1` line at the root level of your `.taskcluster.yml`
+* You will need `queue:route:checks` scope
```
djmitche: What does this mean?  Do I as a user need the scope?

ui/docs/reference/integrations/github/checks.mdx
```diff
@@ -0,0 +1,43 @@
...
+ (for the above example, you arrange the processing of your logs in the way you want, 
+ format the result the way you want using Markdown and save the result in an artifact)
+* Add something like 
+    ```YML
+    extra:
+        github:
+            customCheckRun:
+                textArtifactName: <artifact_name>
+    ```
+ to the definition of the task for which you want customize the check.
```
djmitche: Is there a default value for this artifact name?

services/github/src/utils.js
```diff
@@ -0,0 +1,33 @@
+const request = require('superagent');
+const util = require('util');
+
+const setTimeoutPromise = util.promisify(setTimeout);
+
+const throttleRequest = async ({url, method, delay = 10, response = {status: 0}, attempt = 1}) => {
```
djmitche: This is a useful function -- I just needed something like it for the performance-testing framework.  Making this into a utility function in `taskcluster-client` might make a good followup and/or contributor issue.

services/github/src/handlers.js
```diff
@@ -504,17 +490,56 @@ async function statusHandler(message) {
     `Attempting to update status of the checkrun for ${organization}/${repository}@${sha} (${taskState.conclusion})`,
   );
   try {
+    const taskDefinition = await this.queueClient.task(taskId);
+    debug(`Result status. Got task build from DB and task definition for ${taskId} from Queue service`);
```
imbstack: let's not add any new `debug` to service code. This can be a `monitor.debug` or registered message if we want to keep it.

services/github/src/utils.js
```diff
@@ -0,0 +1,33 @@
...
+  let res;
+  try {
+    res = await request(method, url);
+  } catch (e) {
+    if (e.status >= 400 && e.status < 500) {
+      return e;
+    }
+    if (e.status >= 500) {
+      const newDelay = 2 ** attempt * 100;
+      /* eslint-disable comma-dangle */
```
imbstack: Let's not disable these except for exceptional cases. If we don't like the settings, we can change them.

services/github/src/handlers.js
```diff
@@ -504,17 +490,56 @@ async function statusHandler(message) {
...
+            - no artifact entity in DB, no artifact in S3: statusCode 404
+        */
+        await this.createExceptionComment({
+          debug,
+          instGithub,
+          organization,
+          repository,
+          sha,
+          error: res.response.error,
+        });
```
imbstack: Even with the retries we can still get a 500 here. Probably makes sense to only report 400s directly and instead just say "internal error" for 500s? I could be convinced differently.

services/github/src/handlers.js
```diff
@@ -523,9 +548,9 @@ async function statusHandler(message) {
         output: {
           title: `${this.context.cfg.app.statusContext} (${eventType.split('.')[0]})`,
           summary: `${taskDefinition.metadata.description}`,
-          text: `[Task group](${taskGroupUI(this.context.cfg.taskcluster.rootUrl, taskGroupId)})`,
+          text: `[${CHECKRUN_TEXT}](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n${customCheckRunText || ''}`,
         },
-        details_url: taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId),
+        details_url: taskGroupUI(this.context.cfg.taskcluster.rootUrl, taskGroupId),
```
imbstack: why send details to the group instead of the task now? it feels like that should still be the task as well?

services/github/test/handler_test.js
```diff
@@ -738,6 +765,48 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
       assert(monitorManager.messages.some(({Type, Severity}) => Type === 'monitor.error' && Severity === LEVELS.err));
       monitorManager.reset();
     });
+
+    test('successfully adds custom check run text from an artifact', async function () {
```
imbstack: let's also have a test that tasks without custom check run stuff don't throw an error. it seems like we've made all tasks have this metadata in these tests now.

services/github/src/handlers.js
```diff
@@ -504,17 +490,56 @@ async function statusHandler(message) {
...
+        */
+        await this.createExceptionComment({
+          debug,
+          instGithub,
+          organization,
+          repository,
+          sha,
+          error: res.response.error,
+        });
+        await this.monitor.reportError(res.response.error);
```
imbstack: If this is a 400, we don't need to log it as an error here. The service is operating correctly and the user has made a mistake.

ui/docs/reference/integrations/github/checks.mdx
```diff
@@ -0,0 +1,43 @@
...
+ task definition. In this case, you will not need any scopes.
+* If you don't want the artifact to be public, name your artifact simply `<file_name>` both when creating it
+ and referencing it in your task definition. In this case, you will need a `queue:get-artifact:<file_name>`
+ scope. Add it to the `scopes` section of the same task.
+ 
+### Debugging Errors in Custom Output
+ 
+* If you see a `Cannot GET blahblah (403)` kind of error, it's likely that your artifact is private and you
+  forgot to add the scope to the `scopes` section of your task definition.
+* If you see a `Cannot GET blahblah (404)` kind of error, it means the artifact wasn't created successfully.
```
imbstack: it could also mean that they've specified the wrong name for the artifact in `task.extra`

services/github/src/handlers.js
owlishDeveloper: Because after this PR is merged it will be reporting the types of errors that do not lead to failing the submission of the task into Taskcluster. In fact, I am calling this from _final status_ handler, so _obviously_ the task had been submitted already. Instead writing another 98% similar function, I opted for changing the message by making it more generic.

Does that answer your question?

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: Can you explain?

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: I'm not sure I entirely understand your question, but I'll try my best to answer :) So, it's the queue client who will need the scope, and the way I was imagining this, the scope has to be provided by the user in the `scopes` section of the same task that provides the artifact name.

Can you talk more about your concern and your suggestions here?

services/github/src/utils.js
owlishDeveloper: Yes, I was actually thinking: it would be good to have high-level `utils` thing that all the services could utilize, now that we have a monorepo

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: not sure what to answer (I think the answer is obvious from the code), can you explain your concern and offer suggestions?

changelog/bug-1552323.md
```diff
@@ -0,0 +1,5 @@
+level: minor
+reference: bug 1552323
+---
+Fixes bug 1552323 and issue https://github.com/mozilla-mobile/fenix/issues/6760
```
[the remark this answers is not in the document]
owlishDeveloper: I only see present tense

services/github/src/handlers.js
owlishDeveloper: I thought after Dustin's improvements to logging this particular `debug` is actually `monitor` under the hood? I am looking at the code [here](https://github.com/taskcluster/taskcluster/blob/master/services/github/src/handlers.js#L458) and [here](https://github.com/taskcluster/taskcluster/blob/master/services/github/src/handlers.js#L792)

services/github/src/utils.js
owlishDeveloper: This is an exceptional case tho. There's no need in a comma after the last function argument. Or can you explain a bit more what you meant?

services/github/src/handlers.js
owlishDeveloper: My rationale for this kind of reporting was these are the things the users can fix or, in case of some force-majeure just an explanation why their custom input didn't appear. If you mean the appearance of the error, they all will have the same appearance: `Cannot GET <url to the artifact> (<errorcode>)`

services/github/src/handlers.js
owlishDeveloper: I want to switch the two URLs. Right now, the most useful URL (the one that leads straight to the task) appears in small gray font on the bottom of the checkrun :) And the less useful URL (to the taskgroup) that requires additional clicking to get to the task appears as a proper larger link. I am noticing that many users don't even realize that that small gray thing is _there_. So I decided to switch them

services/github/test/handler_test.js
owlishDeveloper: > it seems like we've made all tasks have this metadata in these tests now.

Can you show me in the code where you see that?

ui/docs/reference/integrations/github/checks.mdx
djmitche: There is no support for private artifacts.

ui/docs/reference/integrations/github/checks.mdx
djmitche: Please add that information to the documentation.

ui/docs/reference/integrations/github/checks.mdx
djmitche: We had discussed having a default value in previous reviews, but I don't see it added here.  Please add one.

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: Oh. That was in the `getArtifact` and `createArtifact` documentation. (from the queue code it seemed to me the artifact would be saved in the private bucket for the account (I was looking [here](https://github.com/taskcluster/taskcluster/blob/master/services/queue/src/artifacts.js#L176) ) So, what happens in practice if your artifact name doesn't have `public/` prefix?

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: oh, wait, you were asking about `queue:route:checks` scope - it's the user's github client will need it. I think people need to contact releng team for the scope at this point. I will add this info there (I was hesitant to add it because today it's releng team and tomorrow and it might be somebody else)

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: but if there's a default name, how would tc-gh know if it needs to try and download the artifact? I currently do this check:

`if (customCheckRun && customCheckRun.textArtifactName)`

I'd avoid making unnecessary requests to S3 because we are billed per request. Do you have suggestions?

ui/docs/reference/integrations/github/checks.mdx
djmitche: It's the repository role (`repo:github.com/org/repo:action`) that needs the scope.  I think that's a clearer term for it than "github client".

I agree that we should not mention releng here.

ui/docs/reference/integrations/github/checks.mdx
djmitche: The request for an artifact first goes to the queue, which doesn't bill per request, so I think it's safe to make that request in all cases.

ui/docs/reference/integrations/github/checks.mdx
djmitche: As the PR is written, tc-github will try to fetch the artifact and fail because it does not use authentication to fetch it.

Note that I'm not saying this must support private artifacts -- I'd be just as happy to see the documentation say that the artifact must be a public artifact.

services/github/src/handlers.js
djmitche: Got it, yeah.  Would information on the stage at which the error occurred help the user (or us) later?  Maybe this function could take an extra parameter to give that information?

owlishDeveloper: I don't know. It seems to me that is obvious from the task status and the error message 🤔

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: Do you mean [this auth here](https://github.com/taskcluster/taskcluster/pull/2477/files#diff-41e7ecfbb590e5f3bbddd300fbf410dbR499-R505) will not work?

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: Sounds good, I'll add that info

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: Oooh, I see, I see. Yes, that makes sense, thank you - I'll add the default name then!

owlishDeveloper: (do you think it's OK to _always_ hit the API tho? Because the Queue will then query the database as well....)

services/github/src/handlers.js
owlishDeveloper: If you still convinced we should do that, I can file a bug to do that, because it seems a bit out of scope here

services/github/src/handlers.js
owlishDeveloper: Ok, I'll thought better of it - I'll just remove that log

services/github/src/utils.js
owlishDeveloper: Nevermind, I reformatted the code another way

services/github/src/handlers.js
owlishDeveloper: Nevermind, switched them back. It doesn't matter, and it's out of scope of this PR anyways

services/github/test/handler_test.js
owlishDeveloper: ...because to me it seems what you're asking for is covered by the pre-existing test-cases, so I'm not sure what to do. It seems to me I am using a separate taskid for the tests for this functionality, but I could be wrong

ui/docs/reference/integrations/github/checks.mdx
djmitche: Yes, it's fine

ui/docs/reference/integrations/github/checks.mdx
djmitche: Right.

owlishDeveloper: >  There's the potential here to leak private artifacts, so it will need some careful thought.

Do you have suggestions on how to improve the current auth?

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: Can you expand a bit on that and talk some more about your reasons?

djmitche: APIs are made to be called, so not calling API methods is much lower priority than user experience.  In this case, we are adding one call to `getArtifact` for each call to `createTask`, which is a much heavier-weight API call, so the marginal increase in load on queue will be very small.

That said, I had asked in an earlier review what your plans are to support images and other attributes.  Do those plans involve additional artifacts, additional properties in task.extra, and so on?  Would it make be easier for users if those were all combined into a single artifact?

ui/docs/reference/integrations/github/checks.mdx
djmitche: `buildSignedURL` will build a signed URL that can be used for private artifacts.

services/github/test/handler_test.js
owlishDeveloper: Just to confirm: I am implementing Dustin's suggestion on default name, and `Checks API: result status handler` _definitely_ doesn't have that metadata.

services/github/src/handlers.js
imbstack: removing it is fine. just want to comment to say you were right here about the usage of `debug`

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: How can you combine image and text in a single artifact?

owlishDeveloper: Also, that seems like an incremental improvement to me. I don't see why we can't implement that functionality in the future when it's needed - it won't be expensive to change this code. Fenix doesn't need images right now - this PR solves their problem as it is. Just doesn't seem to be a blocking issue to me

changelog/bug-1552323.md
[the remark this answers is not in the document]
imbstack: Ah sorry, you are right about the tense. This is still a bit wonky though. I'm not sure about the technical term for this but I think the best-practices doc wants these more like, "Github integration users can now add custom text artifacts to check runs which fixes https://github.com/mozilla-mobile/fenix/issues/6760" or something along those lines

services/github/src/handlers.js
owlishDeveloper: Created a bug for this https://bugzilla.mozilla.org/show_bug.cgi?id=1624932

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: https://bugzilla.mozilla.org/show_bug.cgi?id=1624934 <---- created a bug to follow up on your suggestion

services/github/src/constants.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  CONCLUSIONS: { // maps status communicated by the queue service to github checkrun conclusions
+    /*eslint quote-props: ["error", "consistent-as-needed"]*/
```
imbstack: ah sorry, I meant that we should add this to our global eslint config. feel free to do it in a followup bug though.

services/github/test/handler_test.js
imbstack: ah, I see the switch-case handling what I am talking about. was that there before? if so I apologize for missing it

services/github/src/handlers.js
imbstack: The only issue is that the user cannot fix a 500

ui/docs/reference/integrations/github/checks.mdx
djmitche: > Just doesn't seem to be a blocking issue to me

Thinking ahead is valuable, even when it's not blocking.  If the improvement is introducing a different approach (say, a JSON artifact containing image and text), then you will need to support both the approach defined here and the new approach in parallel, until the approach defined here can be deprecated and removed.  It's possible that a small amount of thinking-ahead now can save a bunch of extra work later.  That's why I'm bringing it up now.

ui/docs/reference/integrations/github/checks.mdx
tomprince: It is not clear to me that that the scopes of the task are the right thing to consult, when checking to see if the artifact should be used. The main reason is that existing tasks that have scopes for private artifacts shouldn't implicitly allow tc-github to use those artifacts for check.

My inclination would be to not support private artifacts in the initial implementation.

services/github/src/constants.js
```diff
@@ -0,0 +1,19 @@
...
+    'superseded': 'neutral', // queue status means: is not relevant anymore
+    'claim-expired': 'failure',
+    'worker-shutdown': 'neutral', // queue status means: will be retried
+    'malformed-payload': 'action_required', // github status means "correct your task definition"
+    'resource-unavailable': 'failure',
+    'internal-error': 'failure',
+    'intermittent-task': 'neutral', // queue status means: will be retried
+  },
+  CHECKRUN_TEXT: 'View task in Taskcluster',
+  CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME: 'customCheckRunText.md',
```
tomprince: I'm guessing this probably should have a path associated to it (e.g. `public/gituhb/`)?

owlishDeveloper: public by default?

tomprince: I think so, yes. I don't think workers support private logs currently, and I expect this to be used mostly for summarizing stuff from them, so having it be public seems fine. I also think all the repos actively used are public (or could be), so the checks will be visible anyway. (As I said elsewhere, I think the initial implementation should only support public artifacts)

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: I asked Fenix folks what their needs are. as far as I am aware _at the moment_, not supporting private artifacts in the initial implementation should be fine

owlishDeveloper: update: yeah, they confirmed they do not see the use for private artifacts at the moment. Ok, I will go ahead and follow @tomprince 's advice then because it makes perfect sense to me. We can discuss security of this in more detail down the road and I can always add this back in.

services/github/src/constants.js
owlishDeveloper: Gotcha. Like I mentioned above, your suggestion on supporting public artifacts only made sense to me (I made sure with the Fenix folks - they confirmed it's fine for their needs), so agreed.

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: As far as I know, image is binary data, so it cannot be directly inserted into JSON. So if we want to think about images and other things, we needs to think some more - which doesn't seem to be needed right now, as the need at hand does not include images.

owlishDeveloper:
@djmitche it's ready for another look!

services/github/src/constants.js
```diff
@@ -0,0 +1,19 @@
...
+    'superseded': 'neutral', // queue status means: is not relevant anymore
+    'claim-expired': 'failure',
+    'worker-shutdown': 'neutral', // queue status means: will be retried
+    'malformed-payload': 'action_required', // github status means "correct your task definition"
+    'resource-unavailable': 'failure',
+    'internal-error': 'failure',
+    'intermittent-task': 'neutral', // queue status means: will be retried
+  },
+  CHECKRUN_TEXT: 'View task in Taskcluster',
+  CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME: 'public/customCheckRunText.md',
```
djmitche: Tom had suggested `public/github/customCheckRunText.md` here, and I think that makes sense as a parallel to `public/logs/`, our conventional place for task logfiles.

ui/docs/reference/integrations/github/checks.mdx
```diff
@@ -0,0 +1,44 @@
...
+ format the result the way you want using Markdown and save the result in an artifact)
+* Add something like 
+    ```YML
+    extra:
+        github:
+            customCheckRun:
+                textArtifactName: <artifact_name>
+    ```
+ to the definition of the task for which you want customize the check. If you don't want to bother with that,
+ you can just make sure you name your artifact `customCheckRunText.md`, and tc-gh would fetch that automatically.
```
djmitche: I think we would like to encourage users to use the default artifact name, with the `task.extra` option as an alternative when that default name is not appropriate for their use-case.  That would mean putting the "create an artifact named XYZ" first, followed by "if you would like to use a different artifact name, add something like ..".

Also, the default artifact name should match `CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME` exactly (including `public/` as that's part of the artifact name).

ui/docs/reference/integrations/github/checks.mdx
```diff
@@ -0,0 +1,44 @@
...
+ Please note that at the moment we do not support private artifacts for this functionality, so do make sure
+ to add `public/` prefix to your artifact name when using `Queue.createArtifact` endpoint
+ (like this: `public/<file_name>`).
```
djmitche: The user probably won't be calling `createArtifact` themselves -- they'll be defining this as an artifact that the worker will upload for them.  So mentioning `createArtifact` here might confuse users.

ui/docs/reference/integrations/github/checks.mdx
```diff
@@ -0,0 +1,44 @@
...
+* You will need `queue:route:checks` scope for your `repo:github.com/org/repo:<action>` role.
+* If you are using a decision task, make sure you add `route.checks` to the `routes` of your tasks
+
+## Custom Output in Checks
+
+If you need to add some sort of custom output to your checks (for example, you would like to process
+your logs and add some results to the check's body):
+* You need to arrange for an artifact with the custom text (plain or Markdown) you'd like to add to your check 
+ (for the above example, you arrange the processing of your logs in the way you want, 
+ format the result the way you want using Markdown and save the result in an artifact)
```
djmitche: which example above?

services/github/src/constants.js
owlishDeveloper: good point, I missed that from his comment, thanks

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: lines 12-13, `(for example, you would like to process your logs and add some results to the check's body)`

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: when I was writing this, I was thinking more of the admin type of audience(such as Tom here)who would actually be defining scripts and tasks. Those also might bethe developers themselves.

I'm not sure what's the best way of phrasing this, do you have ideas?

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: I removed the "above example" tho - I agree it might be confusing. Thanks!

ui/docs/reference/integrations/github/checks.mdx
owlishDeveloper: Actually, your idea is expressed in the previous comment, I think.

djmitche: It might be better to write it in a less instructional style (since instructional style is fairly specific to the audience) and more of a reference style.  Something like

When a task completes, the Taskcluster-GitHub service can optionally include task-specific information in the check run displayed in the GitHub checks tab.  To do so, it looks for an artifact on the task named `public/github/customCheckRunText.md` and uses the artifact content as the check run text.  The artifact name can be customized by setting `task.extra.github.customCheckRun.textArtifactName` in the task definition.  For example:
```
payload:
  ...
  artifacts:
    public/github/customCheckRunText.md:
      type: file
      path: checkrun.md
```
or, if for example all build results are in `public/results`:
```
payload:
  ...
  artifacts:
    public/results:
      type: directory
      path: results
task:
  extra:
    github:
      customCheckRun:
        textArtifactName: public/results/checkrun.md
```

services/github/src/constants.js
tomprince: If we have a default, I feel strongly that it should be scoped to `.../github/...`

services/github/src/handlers.js
```diff
@@ -504,17 +490,47 @@ async function statusHandler(message) {
...
+    const {textArtifactName} = customCheckRun || CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME;
+
+    let customCheckRunText = '';
+
+    const url = this.queueClient
+      .buildUrl(this.queueClient.getArtifact, taskId, runId, textArtifactName);
+
+    let res = await utils.throttleRequest({url, method: 'GET'});
+
+    if (res.status >= 400 && res.status !== 404) {
```
tomprince: (follow-up) I wonder if it would be useful if an artifact is explicitly specified, but not found, if mention should be made of it? On the other hand, I can imagine the default artifact being used most of the time.

services/github/src/handlers.js
```diff
@@ -504,17 +490,47 @@ async function statusHandler(message) {
...
     if (checkRun) {
       await instGithub.checks.update({
         ...taskState,
         owner: organization,
         repo: repository,
         check_run_id: checkRun.checkRunId,
+        output: {
+          title: `${this.context.cfg.app.statusContext} (${eventType.split('.')[0]})`,
+          summary: `${taskDefinition.metadata.description}`,
+          text: `[${CHECKRUN_TEXT}](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n${customCheckRunText || ''}`,
```
tomprince: I wonder if the task should be able to completely override this? (in any case, this is something we can iterate after this lands)

services/github/src/handlers.js
```diff
@@ -504,17 +490,47 @@ async function statusHandler(message) {
...
+      customCheckRunText = res.text.toString();
+    }
+
     if (checkRun) {
       await instGithub.checks.update({
         ...taskState,
         owner: organization,
         repo: repository,
         check_run_id: checkRun.checkRunId,
+        output: {
```
tomprince: It would probably make sense to move at least `output` outside the `if`, so the text construction isn't duplicated.

PR #2482, opened 2020-03-10
Bug 1621270 - retry GitHub artifact uploads

djmitche:
Bugzilla Bug: [1621270](https://bugzilla.mozilla.org/show_bug.cgi?id=1621270)


djmitche:
I tested this by copy/pasting the code into a temp file and stubbing out the call to octokit.  A staging release wouldn't actually upload things.

PR #2497, opened 2020-03-11
Update jsone

owlishDeveloper:
Bugzilla Bug: [1621630](https://bugzilla.mozilla.org/show_bug.cgi?id=1621630)


tomprince:
It looks like the bugnumber is probably missing a digit.

djmitche:
fixed :)

djmitche:
(merged so we don't accidentally make another release without short-circuiting)

PR #2500, opened 2020-03-12
add checks route to gw decision-task subtasks

djmitche:
I don't think this will work, but let's try it :)


djmitche:
Oh hey it does!

djmitche:
(Some of those g-w tasks will fail due to https://bugzilla.mozilla.org/show_bug.cgi?id=1622052)

djmitche:
A downside here is that some of these workers are *very* backed up (especially the raspberry pi) so we'll not see green checkmarks for some time.  I suppose it's better to merge something with a few known yellow dots left, than to totally miss failures that aren't visible here.

PR #2501, opened 2020-03-12
Bug 1622052 - distinguish local and remote proto capabilities

djmitche:
This makes capabilities a private property of Protocol, with methods to
manipulate it.  It also prohibits adding a capability after
protocol.Start is called.

This change simplifies the interface, but makes the tests a bit more
verbose!

Bugzilla Bug: [1622052](https://bugzilla.mozilla.org/show_bug.cgi?id=1622052)

I've cherry-picked the try-checks thing so that I can see the g-w results in the github checks on this PR.  If #2500 doesn't land as-is, I'll remove that commit before merging this PR.

### DIFF
--- a/tools/taskcluster-worker-runner/protocol/protocol.go
+++ b/tools/taskcluster-worker-runner/protocol/protocol.go
@@ -11,15 +11,16 @@ type Protocol struct {
 	// transport over which this protocol is running
 	transport Transport
 
-	// current set of agreed capabilities (but call WaitUntilInitialized first
-	// to avoid finding the empty set at startup)
-	Capabilities *Capabilities
-
-	// Capabilities of the remote peer
-	RemoteCapabilities *Capabilities
+	// local and remote capabilities
+	localCapabilities  *capabilities
+	remoteCapabilities *capabilities
 
 	// callbacks per message type
 	callbacks map[string][]MessageCallback
+
+	// tracking for whether Start has been called yet
+	started      bool
+	startedMutex sync.Mutex
 
 	// tracking for whether this protocol is intialized
 	initialized     bool
@@ -29,8 +30,8 @@ type Protocol struct {
 func NewProtocol(transport Transport) *Protocol {
 	return &Protocol{
 		transport:          transport,
-		Capabilities:       EmptyCapabilities(),
-		RemoteCapabilities: EmptyCapabilities(),
+		localCapabilities:  EmptyCapabilities(),
+		remoteCapabilities: EmptyCapabilities(),
 		callbacks:          make(map[string][]MessageCallback),
 		initialized:        false,
 		initializedCond: sync.Cond{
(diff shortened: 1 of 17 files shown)


tools/taskcluster-worker-runner/protocol/protocol.go
```diff
@@ -107,10 +108,19 @@ func (prot *Protocol) WaitUntilInitialized() {
...
 }
 
-// Check if a capability is supported, after waiting for initialization.
+// Add the given capability to the local capabilities
+func (prot *Protocol) AddCapability(c string) {
+	if prot.started {
+		panic("Cannot AddCapability after protocol is started")
+	}
+	prot.localCapabilities.Add(c)
+}
```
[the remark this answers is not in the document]
djmitche: Nice catch!

tools/taskcluster-worker-runner/protocol/protocol.go
```diff
@@ -11,16 +11,17 @@ type Protocol struct {
...
+	// local and remote capabilities
+	localCapabilities  *capabilities
+	remoteCapabilities *capabilities
 
 	// callbacks per message type
 	callbacks map[string][]MessageCallback
 
+	// tracking for whether Start has been called yet
+	started      bool
+	startedMutex sync.Mutex
```
walac: Better to use `RWMutex` to avoid contention.

tools/taskcluster-worker-runner/protocol/protocol.go
```diff
@@ -62,28 +63,31 @@ func listOfStrings(val interface{}) []string {
...
 
 		prot.Send(Message{
 			Type: "welcome",
 			Properties: map[string]interface{}{
-				"capabilities": prot.Capabilities.List(),
+				"capabilities": prot.localCapabilities.List(),
 			},
 		})
 	}
+	prot.startedMutex.Lock()
```
walac: You should lock at the beginning of the function. If not, the race condition persists, as you can add a new capability in the middle of the capability negotiation.

tools/taskcluster-worker-runner/protocol/protocol.go
djmitche: There's only a tiny critical section -- why does it matter?

tools/taskcluster-worker-runner/protocol/protocol.go
djmitche: The code above here just registers callbacks -- the start only occurs in start of the receiving goroutine in the last line (after the unlock).

tools/taskcluster-worker-runner/protocol/protocol.go
djmitche: The mutex isn't protecting the start process -- it's just protecting access to the boolean variable.

walac: Ah, got it, maybe it would be better to use an atomic variable, but it doesn't matter much.

tools/taskcluster-worker-runner/protocol/protocol.go
walac: Ah, in this case, the `mutex` doesn't do much.

djmitche: It does make `go test -race` happy, at least :)

PR #2505, opened 2020-03-13
Integrate tc-lib-azure into DB versions

djmitche:
This moves the postgres code for azqueue into `db/` along with the associated fakes, tests, etc.

Refs #2292


db/test/versions/0003_test.js
```diff
@@ -0,0 +1,25 @@
+const assert = require('assert').strict;
+const helper = require('../helper');
+const testing = require('taskcluster-lib-testing');
+const {UNDEFINED_TABLE} = require('taskcluster-lib-postgres');
+
+suite(testing.suiteName(), function() {
+  helper.withDbForVersion();
+
+  test('azure_queue_messages table created', async function() {
```
helfi92: Should we also test that downgrading removes this table?

djmitche: Good point -- I wrote this before downgrading was supported!

PR #2506, opened 2020-03-13
Check DB user permissions more thoroughly

### LINKED ISSUE #2498: Ensure that per-service postgres users are not superusers
In `db:ugprade`, we check that per-table permissions correspond to those in `db/access.yml`, but we don't look at whether a user is a superuser.  We should.

djmitche:
Thanks to @edunham for recommending this!

Fixes #2498

### DIFF
--- a/libraries/postgres/src/Database.js
+++ b/libraries/postgres/src/Database.js
@@ -250,6 +250,23 @@ class Database {
         }
       }
 
+      // look for unexpected granted roles
+      const roleRes = await client.query(`
+        select
+          r.rolname as role,
+          u.rolname as user
+        from
+          pg_catalog.pg_roles as r,
+          pg_catalog.pg_roles as u,
+          pg_catalog.pg_auth_members
+        where
+         r.oid = roleid and
+         u.oid = member and
+         u.rolname like $1`, [usernamePattern]);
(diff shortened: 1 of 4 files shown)


libraries/postgres/src/Database.js
```diff
@@ -221,6 +222,51 @@ class Database {
         }
       }
 
+      // look for unexpected user attributes
+      const badAttrs = {
+        'superuser': 'rolsuper',
+        'createrole': 'rolcreaterole',
+        'createdb': 'rolcreatedb',
+        'replication': 'rolreplication',
+      };
```
imbstack: just out of curiousity, why do we do `badAttrs` rather than confirm that we only have `goodAttrs`. Do we want to allow for arbitrary roles in general?

djmitche: These are separate from roles.  These are different columns, in a join, so *conceivably* we could use `select *` to get all of the columns available, and then fail if there are any unknown columns with a `true` value.  I don't know if such columns would be added in a major or minor version of pg, though.  It runs the risk of spurious failures if pg adds a new benign attribute like `login`.  I'm not sure it's worth painting all the way into that corner..

PR #2508, opened 2020-03-14
Upgrade acorn, minimist

djmitche:
This doesn't get all of the uses of minimist, but happily it is used
only for CLI interfaces to things like mkdirp that we do not use in
production.


djmitche:
Whoops, I duped myself.  #2509 has a better description :)

PR #2516, opened 2020-03-16
Add NewRelic support

### LINKED ISSUE #2507: Add newrelic support to TC, especially for DB access
This will likely be necessary to drill down into DB performance issues.

djmitche:
This adds NewRelic in an optional way.

The config is a little awkward, but was the cleanest approach I could come up with.

Github Bug/Issue: Fixes #2507


djmitche:
I'll add a changelog.  Oops!

PR #2517, opened 2020-03-17
assign worker reviewers when go.mod/go.sum are modified

djmitche:
Noticed this in #2512


djmitche:
(bug filed for the intermittent test failure)

PR #2521, opened 2020-03-17
Disable tasks on pools with insufficient capacity, linux-arm to tier 2

djmitche:
The worker pools behind these tasks are single workers or otherwise
unable to keep up with demand.  So, rather than merging PRs with yellow
dots or eventually red X's when the tasks time out, let's disable them.

Of these platforms, Windows 10 is a tier-1 platform, but that platform's
testing is covered by the Windows 2012 worker pool.  And linux-arm is a
tier-1 platform, but in the larger Mozilla scope is tier-2 at this
point.

### DIFF
--- a/workers/generic-worker/gw-decision-task/tasks.yml
+++ b/workers/generic-worker/gw-decision-task/tasks.yml
@@ -45,9 +45,6 @@ Tasks:
     - WorkerPool: 'proj-taskcluster/gw-ci-macos'
       Env:
         ENGINE: 'simple'
-    - WorkerPool: 'proj-taskcluster/gw-ci-raspbian-stretch'
-      Env:
-        ENGINE: 'simple'
     - WorkerPool: 'proj-taskcluster/gw-ci-ubuntu-18-04'
       Env:
         ENGINE: 'multiuser'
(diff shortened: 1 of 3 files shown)


workers/generic-worker/gw-decision-task/tasks.yml
```diff
@@ -57,37 +54,36 @@ Tasks:
...
         GW_SKIP_Z_DRIVE_TESTS: 'true'
-#####################################################################
-#   #############################################################   #
-#   # DISABLING WINDOWS 7 UNTIL WE HAVE WINDOWS 7 WORKERS AGAIN #   #
-#   #############################################################   #
-#   - WorkerPool: 'proj-taskcluster/gw-ci-windows7-386'             #
-#     Env:                                                          #
-#       ENGINE: 'multiuser'                                         #
-#####################################################################
+# The following are disabled due to insufficient capacity:
```
imbstack: Should we track this with a bug referenced here or just remove?

djmitche: I think that the platforms are tier-2 is probably sufficient notice for now..

PR #2525, opened 2020-03-17
Rewrite services tests to work with postgres

helfi92:


libraries/testing/src/with-db.js
```diff
@@ -99,7 +99,8 @@ module.exports.withDb = (mock, skipping, helper, serviceName) => {
       return;
     }
     if (helper.db) {
-      await helper.db.close();
+      // TODO: Uncommenting this throws Error: Cannot use a pool after calling end on the pool for the last test
+      // await helper.db.close();
```
helfi92: Do we need this line? It seems to be causing issues when running the tests. Here's the full error stack:

```
Error: Cannot use a pool after calling end on the pool
      at Pool.connect (/Users/hassanali/Documents/Mozilla/projects/taskcluster/node_modules/pg-pool/index.js:148:19)
      at Database._withClient (/Users/hassanali/Documents/Mozilla/projects/taskcluster/libraries/postgres/src/Database.js:378:31)
      at Object.fns.<computed> [as secrets_entities_scan] (/Users/hassanali/Documents/Mozilla/projects/taskcluster/libraries/postgres/src/Database.js:48:32)
      at fetchResults (/Users/hassanali/Documents/Mozilla/projects/taskcluster/libraries/entities/src/index.js:487:67)
      at Function.scan (/Users/hassanali/Documents/Mozilla/projects/taskcluster/libraries/entities/src/index.js:514:25)
      at cleanup (/Users/hassanali/Documents/Mozilla/projects/taskcluster/libraries/testing/src/with-entity.js:71:25)
      at Context.<anonymous> (/Users/hassanali/Documents/Mozilla/projects/taskcluster/libraries/testing/src/with-entity.js:100:11)
      at runMicrotasks (<anonymous>)
```

djmitche: It looks like that's coming from a cleanup call in `with-entities.js`?  It seems like that's indicating a bug somewhere else, with something running that shouldn't be...

PR #2526, opened 2020-03-17
remove unused parameters from 'yarn dev:init'

djmitche:


djmitche:
(this would be a silent, bug-free changelog anyway)

PR #2515, opened 2020-03-16
[postgres] Add tc-lib-entities and db v2/

helfi92:
Extracted tc-lib-entities and version 2 of the postgres database from https://github.com/taskcluster/taskcluster/pull/2174.

https://hassan.taskcluster-dev.net/ is running without azure-entities. it’s using tc-lib-entities which is 0% azure, 100% postgres!

Note: This doesn't change any of the services to use tc-lib-entities just yet. This will not impact anything in production. They're all still using azure-entities.

PR #2528, opened 2020-03-17
Set err.name for tc-lib-entities errors.

djmitche:
Some services look at err.name, and some look at err.code, for the error
messages from Azure entities.  Let's support both.

This will need to be cherry-picked into master once #2515 lands.


djmitche:
@helfi92 do you want to cherry-pick this into #2515 now?

helfi92:
Sounds good.

helfi92:
Done.

PR #2537, opened 2020-03-20
Use UTC with Postgres

### LINKED ISSUE #2536: azqueue tests fail on a machine with non-UTC timezone
Running the azqueue tests on a laptop that's not set to UTC fails.  Something's wrong with the conversion of Date objects into timezonetz columns in the DB.

djmitche:
This fixes failures of tests to run on machines not in UTC.

Actually, only the first commit is required, but the second is something else I tried to fix this, and is a good idea anyway -- it makes very clear to anyone else using the database the precise time that the timestamps indicate, without guessing "UTC? Mountain View?"

I added a card to https://github.com/taskcluster/taskcluster/projects/3 to track checking for and rejecting TIMESTAMP columns, so I don't make the same mistake again.

The entities tables do not have any timestamp(tz) columns, so no change needed.

Github Bug/Issue: Fixes #2536

### DIFF
--- a/db/versions/0004.yml
+++ b/db/versions/0004.yml
@@ -1,0 +1,19 @@
+version: 4
+# in version 3, this table was incorrectly set up with timestamp columns
+# lacking timezones, causing issues when the server and client timezones do not
+# agree.
+migrationScript: |-
+  begin
+    alter table azure_queue_messages
+      alter column inserted type timestamptz,
+      alter column visible type timestamptz,
+      alter column expires type timestamptz;
+  end
+downgradeScript: |-
+  begin
+    alter table azure_queue_messages
+      alter column inserted type timestamp,
+      alter column visible type timestamp,
+      alter column expires type timestamp;
+  end
+methods: []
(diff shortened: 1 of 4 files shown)


db/versions/0004.yml
```diff
@@ -0,0 +1,19 @@
+version: 4
+# in version 3, this table was incorrectly set up with timestamp columns
+# lacking timezones, causing issues when the server and client timezones do not
+# agree.
+migrationScript: |-
+  begin
+    alter table azure_queue_messages
+      alter column inserted type timestamptz,
+      alter column visible type timestamptz,
+      alter column expires type timestamptz;
```
helfi92: Do we need to set the timestamp globally? Some tables in entities have `expires` as a column. I wonder if a scan with a date condition would create issues for the entities.

djmitche: I think that's OK because it's in a JSON object that we both encode and decode in Node.

djmitche:
I cherry-picked this into #2548 so @helfi92 please have another look there.

djmitche:
Correction -- I only cherry-picked one of these two commits to that PR.  Let's review it here.

PR #2548, opened 2020-03-23
Record table structure in db/tables.yml

### LINKED ISSUE #2541: summarize table columns automatically in db docs
This will work a lot like `access.yml`: a file containing the expected DB configuration (with comments and descriptions) that is compared at test time with the actual DB structure, failing if they do not match.

djmitche:
Github Bug/Issue: Fixes #2541

As an added benefit, this adds a check for any stray timezone-less timestamps.  We can easily add additional assertions about the schema later.


libraries/postgres/src/Relations.js
```diff
@@ -0,0 +1,53 @@
...
+   * Load relations from a file
+   */
+  static fromYamlFileContent(content, filename) {
+    Relations._check(content, filename);
+    const relations = new Relations(content);
+    return relations;
+  }
+
+  /**
+   * Load relatoins from a serialized representation
```
helfi92: `s/relatoins/relations/`

libraries/postgres/src/Relations.js
```diff
@@ -0,0 +1,53 @@
+const {isPlainObject} = require('lodash');
+const assert = require('assert').strict;
+
+class Relations {
```
helfi92: What are relations? A comment here perhaps would help.

djmitche: Will do -- I'm expecting that this will be useful for a `views.yml` too, someday, and "relation" is just a general term for anything with rows and columns, such as tables or views.  I'll add a comment.

djmitche:
I have somehow deeply ingrained the idea that DB changes don't need a changelog, from time working on the development branch.  I'll add one and merge.

PR #2530, opened 2020-03-18
Rewrite services tests to work with postgres

helfi92:
Resolves https://github.com/taskcluster/taskcluster/issues/2558.

Services to cover:
* [x] secrets
* [x] purge-cache
* [x] auth
* [x] github
* [x] hooks
* [x] notify
* [x] index
* [x] queue
* [x] web-server
* [x] worker-manager

TODO:
* [x] Fix tests in taskcluster-queue. (1 failing - `listArtifacts, listLatestArtifacts (continuationToken)`)
* [x] Fix tests in taskcluster-worker-manager
* [x] Fix tests in taskcluster-auth
* [x] File a bug re:tc-worker-manager test (fake function isn't there)

PR #2553, opened 2020-03-23
Fix ordering of queue pops when numberOfMessages > 1

djmitche:
When getting multiple messages from the queue in the same call, Azure
Queues returns those messages in the order they were received.  This
updates the Postgres implementation to do the same.

### DIFF
--- a/changelog/issue-2553.md
+++ b/changelog/issue-2553.md
@@ -1,0 +1,4 @@
+level: patch
+reference: issue 2553
+---
+The taskcluster-lib-azqueue library now returns "batches" of messages in the order they were inserted.
--- a/db/versions/0005.yml
+++ b/db/versions/0005.yml
@@ -1,0 +1,45 @@
+version: 5
+migrationScript: |-
+  begin
+  end
+downgradeScript: |-
+  begin
+  end
+methods:
+  azure_queue_get:
+    description: |
+      Get up to `count` messages from the given queue, setting the `visible`
+      column of each to the given value.  Returns a `message_id` and
+      `pop_receipt` for each one, for use with `azure_queue_delete` and
+      `azure_queue_update`.
+    mode: write
+    serviceName: queue
+    args: queue_name text, visible timestamp, count integer
+    returns: table (message_id uuid, message_text text, pop_receipt uuid)
+    # This updates the implementation in version 3 by sorting the results, as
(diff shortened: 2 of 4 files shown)


djmitche:
Hm, I can't reproduce that error - is the latest pushed to #2530?

helfi92:
I just pushed the code. It seems to work in CI...but not locally.

helfi92:
https://community-tc.services.mozilla.com/tasks/JVwyovgeRMG7PXOEhcKGiA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FJVwyovgeRMG7PXOEhcKGiA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log seems to be failing on a test that also uses `claimTask`. Any chance this would also be coming from azqueue?

djmitche:
> I just pushed the code. It seems to work in CI...but not locally.

Hm, I still passing for me there. Does it fail every time for you?

That second error seems less likely to be due to azqueue..

helfi92:
> Hm, I still passing for me there. Does it fail every time for you?

Yeah.

> That second error seems less likely to be due to azqueue..

Which error? There's only one error in the link shared above.

djmitche:
This one:
> https://community-tc.services.mozilla.com/tasks/JVwyovgeRMG7PXOEhcKGiA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FJVwyovgeRMG7PXOEhcKGiA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log seems to be failing on a test that also uses claimTask. Any chance this would also be coming from azqueue?

djmitche:
Ah, I can replicate on my laptop, though, so this must be another timezone issue.  Indeed, merging #2537 seems to fix it.  I'll rebase this onto master since the version files now conflict.

djmitche:
Also, I removed the empty `0004_test.js` file, this version doesn't have a way to test it directly (reading the function body back out from `pg_catalog` and checking that it matches some regexp seems pretty silly).  Maybe someday we'll add a meta-check that all DB versions have test scripts, and can add an empty file there with a comment explaining why tests aren't required.  Let's wait and see how that works out.

PR #2570, opened 2020-03-25
merge from tc-lib-postgres branch

djmitche:
This merges some changes that have been made on the tc-lib-postgres branch, and fixes for some merge errors, back to master.

Notably, this changes the name of all of the tables created in DB version 2.  That's sort of "cheating", but we're hoping that nobody ran `db:upgrade` on v28.1.0, or they will need to drop and reconstruct their DB.  We really shouldn't cheat like this again :)

Hopefully, we can land this before the next TC release and just pretend 28.1.0 never happened.


djmitche:
looks like everything but the coverage upload has passed..

djmitche:
Oh, I misread that, sorry.

The tc-db tests are failing, but they're failing the same way in tc-lib-postgres -- more "fun" from re-re-naming the tables.

djmitche:
OK, I updated this to incorporate #2572, with the idea of not changing a released DB version.  However, this still includes some changes to version 2, changing the `page` parameter of the scan methods to `page_offset`, with corresponding code.  That'd be a problem if a DB were upgraded using TC 28.1.0 (which has the `page` parameter), then subsequently upgraded to TC 29 (which expects the `page_offset` parameter, but would have no reason to re-run the version 2 migration).

So, I think this and the tc-lib-postgres branch require further updates to move those stored-function changes into a new DB version.  So, revoking the review request here and on #2572.

PR #2572, opened 2020-03-25
revert table names in tc-lib-postgres to match master

djmitche:
These table names were released in 28.1.0, and we should avoid changing
database versions that have been released.  The table names changed due
to different versions of change-case being used, so this uses a table of
special cases to avoid the issue entirely.


djmitche:
The "Allow deprecated methods.." commit here is not necessary for this particular PR, but will be someday when we want to actually deprecate a method.

djmitche:
Ignore the previous comment - I'll pull that out into its own PR.

PR #2574, opened 2020-03-26
Hook tables 'Queues' and 'LastFire3' are also signed

djmitche:
Why they're signed, I don't know, but we shouldn't run the importer on them when doing development imports :)


helfi92:
Ah, are they all covered by tc-api-importer?

djmitche:
Queues and LastFire3 will get left empty -- they can't be imported by API.  When we do the production import, they'll get pulled over.

PR #2575, opened 2020-03-26
Bug 1551846 - Add API endpoint to get TC version

helfi92:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1551846:

>  same endpoints for the taskcluster-ui service, using nginx config

Could you elaborate how this would work, please? As of right now the endpoint will be exposed for each server. Should the nginx forward the call to something like the auth server? What would be the URI to connect to the auth server?


libraries/app/test/app_test.js
```diff
@@ -25,6 +27,19 @@ suite(testing.suiteName(), function() {
...
+            try {
+              const taskclusterVersion = fs.readFileSync(taskclusterVersionFile).toString().trim();
+
+              res.header('Content-Type', 'text/plain');
+              res.send(taskclusterVersion);
+            } catch (err) {
+              res.header('Content-Type', 'text/plain');
+              res.status(500).send('Could not locate the version file');
+            }
+          });
```
djmitche: The test below tests *this* implementation, and not real the one in `index.js`.

I suspect that the real implementation could be tested using `mock-fs`.  There are some uses of that package in tc-lib-references if you want to check them out.

libraries/app/test/app_test.js
```diff
@@ -79,6 +94,13 @@ suite(testing.suiteName(), function() {
       assert.equal(res.headers['content-type'], 'text/plain; charset=utf-8');
     });
 
+    test('/__version__', async function() {
+      const res = await request.get('http://localhost:1459/api/test/v1/__version__');
```
djmitche: https://github.com/mozilla-services/Dockerflow/#containerized-app-requirements specifies that the path be `/__version__`, not `/api/test/v1/__version__`.

ui/web-ui-nginx-site.conf
```diff
@@ -50,6 +50,7 @@ http {
 
       location @rewrites {
         rewrite ^\/docs.*$ /docs.html last;
+        rewrite ^\/__version__$ /app/version.json last;
         rewrite ^.+$ /index.html last;
```
helfi92: What should the `__lbheartbeat__` and `__heartbeat__`  endpoints redirect to? Unlike `__version__`, they don't point to a json file but rather are endpoints exposed by each service.

libraries/app/src/index.js
```diff
@@ -124,6 +127,29 @@ const app = async (options) => {
...
+    } catch (err) {
+      res.header('Content-Type', 'application/json');
+      res.status(500).send({ error: 'Not found' });
+    }
+  });
+
+  app.use('/__heartbeat__', (req, res) => {
+    res.header('Content-Type', 'application/json');
+    res.status(200).send({});
+  });
```
helfi92: Should the heartbeat endpoint replicate the logic in ping.js? https://github.com/taskcluster/taskcluster/blob/4fadfb8527ec4fecb50145f679f9641ed23eca1c/infrastructure/tooling/src/smoketest/checks/ping.js#L40-L63

ui/web-ui-nginx-site.conf
djmitche: It sounds like they should just always return a 200 status and an empty JSON body.  We can go back later and check on DB connectivity, etc.

libraries/app/src/index.js
djmitche: It kind of serves the same purpose, but I don't think that uptime information is especially useful.  Just returning an empty body is probably enough.

ui/web-ui-nginx-site.conf
helfi92: Ah great. The current state of the PR does just that.

changelog/bug-1551846.md
```diff
@@ -0,0 +1,4 @@
+level: minor
+reference: bug 1551846
+---
+taskcluster-lib-app now includes endpoints /__version__, /__heartbeat__, and /__lbheartbeat__ to be compatible with Dockerflow requirements.
```
djmitche: (nit: add backticks for markdown quoting)

ui/.neutrinorc.js
```diff
@@ -152,12 +152,6 @@ module.exports = {
           .test(/\.all-contributorsrc$/)
           .use('json-loader')
             .loader('json-loader');
-
-      neutrino.config.module
-        .rule('taskcluster-version')
-          .test(/taskcluster-version$/)
-          .use('raw-loader')
-            .loader('raw-loader');
```
djmitche: oh, that's a nice simplification!

ui/web-ui-nginx-site.conf
djmitche: I think this will need `__(lb)heartbeat__` too?

helfi92:
> Looking good. There's infrastructure/references/nginx.conf too, which can (I think) use pretty much the same nginx config lines.

There isn't any location directive in `infrastructure/references/nginx.conf`. It looks like it's only used for references based on what I'm reading in https://github.com/taskcluster/taskcluster/blob/master/infrastructure/references/README.md. Hm, do we still need to add these endpoints there? If yes, why is that?

djmitche:
> If yes, why is that?

My understanding of the docker-flow is that each container should respond to these `__...__` endpoints individually, so that would include the references container.

If that's difficult we could definitely leave it for a follow-up PR.  But, would just adding the same `location` directives to that file do the trick?

helfi92:
> If that's difficult we could definitely leave it for a follow-up PR. But, would just adding the same `location` directives to that file do the trick?

I was just wondering why it was needed. Yes, adding the same `location` directives should do the trick. I'm on it.

helfi92:
```
➜  taskcluster git:(tc-lib-app-version-endpoint) curl localhost/__version__
{"version": "v28.1.0-52-g94dda914d", "commit": "94dda914d7bed7ec7f58dd48e3494b8b96074251", "source": "https://github.com/taskcluster/taskcluster", "build": "NONE"}
➜  taskcluster git:(tc-lib-app-version-endpoint) curl localhost/__heartbeat__
{}%
➜  taskcluster git:(tc-lib-app-version-endpoint) curl localhost/__lbheartbeat__
{}%
```

djmitche:
> ```
> {}%
> ```
Do you have something cool in your shell config that adds those %'s?

helfi92:
> I tested this by building and running
> 
> ```shell
> docker run -e TASKCLUSTER_ROOT_URL=https://dustin.taskcluster-dev.net  -p 80:80 taskcluster/taskcluster:v28.1.0-51-g6202ee732 references/web
> ```
> 
> and then
> 
> ```
> curl localhost/__heartbeat__
> ```

This helped me to test things on my machine. Thank you for the instructions.

I find myself rebuilding the docker image to test the changes. Do you have a more efficient way to do this? I guess I am looking for a way to modify the same docker image without having to build another one, at least when testing things.

djmitche:
> I find myself rebuilding the docker image to test the changes. Do you have a more efficient way to do this? I guess I am looking for a way to modify the same docker image without having to build another one, at least when testing things.

There's not really a way to modify a docker image, but what I did do was to open a shell in a container running the image (`docker run -ti ... bash`) and then edit the nginx config files (vim is installed) and test it by running the command from `/app/entrypoint` within that shell.  So for example for the UI:
```
$ docker run -ti -e TRUST_PROXY=true -e FORCE_SSL=false -e NODE_ENV=production -e PORT=80 -e AZURE_ACCOUNT_ID=123 -e TASKCLUSTER_ROOT_URL=https://dustin.taskcluster-dev.net  -p 80:80 taskcluster/taskcluster:v28.1.0-52-g94dda914d bash
bash-5.0# vi /app/ui/web-ui-nginx-site.conf 
bash-5.0# nginx -c /app/ui/web-ui-nginx-site.conf -g 'daemon off;'
```

PR #2576, opened 2020-03-26
Allow deprecated methods without repeating body, etc.

djmitche:
This came up in another context, and isn't needed immediately, but will make it a lot easier to deprecate methods later, without having to repeat the method bodies.


libraries/postgres/src/Method.js
```diff
@@ -67,10 +70,20 @@ class Method {
   }
 
   checkUpdateFrom(name, existing, version) {
-    assert.equal(existing.mode, this.mode, `method ${name} changed mode in version ${version.version}`);
-    assert.equal(existing.serviceName, this.serviceName, `method ${name} changed serviceName in version ${version.version}`);
-    assert.equal(existing.args, this.args, `method ${name} changed args in version ${version.version}`);
-    assert.equal(existing.returns, this.returns, `method ${name} changed returns in version ${version.version}`);
+    // these fields may be undefined if the methdod is deprecated, in which case no change
```
helfi92: `s/methdod/method/`

Issue #2555, opened 2020-03-23
[tc-lib-azqueue] pending counts should exclude expired


djmitche:
We only expire every hour, so we should exclude queued messages with expires < now when counting pending tasks.

PR #2578, opened 2020-03-27
[Bug 1618333] Add support for audience to changelogs

imbstack:
This adds support for audiences to changelogs with results that look like this:

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### GENERAL

▶ [patch] [bug 1618333](http://bugzil.la/1618333)
Changelog entries now are categorized by the audience that they are useful for

### ADMINS

▶ [patch] [bug 1622943](http://bugzil.la/1622943)
The maximum value for a worker's `lifecycle.reregistrationTimeout` is now 30 days.  Values greater than this cannot be represented in the worker's temporary credentials anyway.

### DEVELOPERS

▶ [minor] [bug 1616998](http://bugzil.la/1616998)
taskcluster-worker-runner has been renamed to worker-runner and its docs have been added to the reference section of the docs portal.

▶ [patch] [#2555](https://github.com/taskcluster/taskcluster/issues/2555)
The azure-queue emulation library now omits expired messages from its counts.  The visible effect is that pending counts for queues no longer include tasks past their deadline.

▶ [patch] [#2553](https://github.com/taskcluster/taskcluster/issues/2553)
The taskcluster-lib-azqueue library now returns "batches" of messages in the order they were inserted.

### OTHER

▶ Additional change not described here: [#2541](https://github.com/taskcluster/taskcluster/issues/2541).

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Bugzilla Bug: [1618333](https://bugzilla.mozilla.org/show_bug.cgi?id=1618333)


dev-docs/best-practices/changelog.md
```diff
@@ -22,19 +22,34 @@ Each file is a markdown file with the following format:
...
+
+Taskcluster has several groups of people who interact with it:
+
+__users:__ people creating tasks (for testing or CI)
+
+__admins:__ people administering the service (managing scopes and roles)
+
+__deployers:__ people deploying the service (running TC services, setting service config values)
+
+__developers:__ people who are working on TC itself, making code changes, etc
```
djmitche: There's a similar list in the [docs best-practices](https://github.com/taskcluster/taskcluster/blob/master/dev-docs/best-practices/docs.md#audience), added at @ajvb's suggestion, but it has five items.  Notably, I think "worker deployers" is an important category to add to the changelog, as worker upgrades tend to happen on a different cadence than services upgrades.  I'd love to have a less clunky name than "worker deployers" :)

imbstack: I'll just go with the clunky name. I've updated the changelog docs to use the docs-docs audiences and added a note the the docs-docs audiences to please update the changelog audiences when you change those.

imbstack:
I think if we default to general they'll all end up in general. I've updated it to allow defaulting to general for `--silent` changelogs which should get this out of the way for 90% of cases where it isn't actually useful.

imbstack:
Going to request another reviewer since @djmitche is PTO this week!

dev-docs/best-practices/changelog.md
```diff
@@ -22,19 +22,35 @@ Each file is a markdown file with the following format:
...
+audience: ..
 # (optional) a reference to the bug or issue tracking this change ("bug xxxxxxx" or "issue xxx")
 reference:  ...
 ---
 <Snippet Text>
 ```
 
+## Audiences
+
+Taskcluster has several groups of people who interact with it. They are defined in [the docs on docs](/dev-docs/best-practices/docs.md).
```
helfi92: docs on docs

infrastructure/tooling/src/changelog/index.js
```diff
@@ -132,16 +145,36 @@ class ChangeLog {
...
-        '▶ ' + levelLabels[level] +
-        (reference ? reference : '') + '\n' +
-        body.trim()
-      ))
-      .join('\n\n') + silentSuffix;
+      `\n\n### OTHER\n\n▶ Additional change${silentCount === 1 ? '' : 's'} not described here: ${silentLinks}.`;
+
+    // These snippets are already sorted in level-order so when we insert
+    // them here they remain in order. no need to re-sort
+    const categorizedSnippets = this.snippets.reduce((acc, {audience, ...rest}) => {
```
helfi92: What is snippet here?

dev-docs/best-practices/changelog.md
imbstack: haha, I wasn't sure what else to call the best practices docs about writing documentation. maybe I should call it "best practices docs about writing documentation" that seems like a better choice

infrastructure/tooling/src/changelog/index.js
imbstack: snippets are the text we write describing the changelog. I'll try to come up with a better name or add a comment

PR #2579, opened 2020-03-27
Avoid OOM when running the importer

helfi92:


infrastructure/tooling/src/importer/importer.js
```diff
@@ -25,14 +26,37 @@ const importer = async options => {
...
+            azureCreds: credentials.azure,
+            tableName,
+            utils,
+            tableParams: tableParameters,
+            rowsProcessed,
+          });
+          await writeToPostgres(tableName, entities, db, ALLOWED_TABLES);
+
+          if (tableParams.nextPartitionKey && tableParams.nextRowKey) {
+            await importTable(tableParams, count);
```
djmitche: I wonder if this kind of recursion will end up keeping `entities` around on the call stack until the entire process is done.  I don't know how smart Node's optimization is.  Could this instead `return importTable(tableParams, count)`?

helfi92: Good observation. Done!

PR #2596, opened 2020-04-01
Request worker termination when the worker process exits

walac:
If the worker process panics and exits, the instance stays in a zombie
state.

To mitigate the situation, when the worker exits, we request a worker
termination to worker-manager.


walac:
The netlify failures seem unrelated.

PR #2597, opened 2020-04-01
Log DB info once a minute

sciurus:
Stackdriver aggregates log-based metrics into 1 minute buckets. By
outputting this log info once a minute, we can ensure every process is
represented in every bucket. Without doing this, metrics derived from it
would be much harder to interpret, since the processes whose data was
represented in each bucket would vary depending on when they were
started.

### DIFF
--- a/libraries/postgres/src/Database.js
+++ b/libraries/postgres/src/Database.js
@@ -560,7 +560,7 @@ class Database {
   }
 
   /**
-   * Periodically monitor pool size, producing a measure every 5m.  These values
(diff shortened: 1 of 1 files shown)


helfi92:
Tests are failing. Perhaps a rebase on master would fix the issue.

sciurus:
That's odd. I fetched TC master branch and committed on top of 2d56854d9efbb911ecf0db8f24e45ad36b0f2eef for this. That commit is still the tip of master.

helfi92:
Reran the failed tasks and now it's passing.

PR #2600, opened 2020-04-03
Fix notify monitor is required error

helfi92:
From slack:

> hmm, what we're seeing is the error at https://github.com/taskcluster/taskcluster/blob/master/libraries/postgres/src/Database.js#L58
> from https://github.com/taskcluster/taskcluster/commit/979653f465104abb27709e7532128a9cf5963dbc
> which is new in v28.2.0. we were on v28.1.0 before

_untested_ but should work.

### DIFF
--- a/services/notify/src/main.js
+++ b/services/notify/src/main.js
@@ -68,9 +68,10 @@ const load = loader({
   },
 
   db: {
-    requires: ['process', 'cfg'],
+    requires: ['process', 'cfg', 'monitor'],
     setup: ({process, cfg}) => tcdb.setup({
       serviceName: 'notify',
+      monitor: monitor.childMonitor('notify'),
       ...cfg.postgres,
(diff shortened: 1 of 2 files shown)


services/notify/src/main.js
```diff
@@ -68,9 +68,10 @@ const load = loader({
   },
 
   db: {
-    requires: ['process', 'cfg'],
+    requires: ['process', 'cfg', 'monitor'],
     setup: ({process, cfg}) => tcdb.setup({
       serviceName: 'notify',
+      monitor: monitor.childMonitor('notify'),
```
imbstack, line 74: hmm, I think this should be `notify.childMonitor('db')`. I think I r+'d a patch where we did this with service name in the other services. could we update them all to be like this in this PR?

helfi92: Sounds good. Filed https://github.com/taskcluster/taskcluster/issues/2601 so we don't forget.

Issue #2601, opened 2020-04-03
Rename child monitors


helfi92:
From https://github.com/taskcluster/taskcluster/pull/2600#discussion_r403182785:

> hmm, I think this should be notify.childMonitor('db'). I think I r+'d a patch where we did this with service name in the other services. could we update them all to be like this in this PR?

helfi92:
@imbstack Do you remember where that was? I wasn't able to to find what you were referring to with `git grep childMonitor`.

djmitche:
For example, `services/queue/src/main.js`
```
  pulseClient: {
    requires: ['cfg', 'monitor'],
    setup: ({cfg, monitor}) => {
      return new pulse.Client({
        namespace: 'taskcluster-queue',
        monitor: monitor.childMonitor('pulse-client'),                                                   
        credentials: pulse.pulseCredentials(cfg.pulse),
      }); 
    },  
  },  
```

I'll fix this up in #2522.

Issue #2522, opened 2020-03-17
Add monitoring for DB activity


djmitche:
* DB pool client counts
 * function invocation counts (aggregate?)

djmitche:
WIP in `issue2522` on my repo.

helfi92:
Completed in https://github.com/taskcluster/taskcluster/commit/979653f465104abb27709e7532128a9cf5963dbc.

djmitche:
Not quite done yet -- still need to rebase and land https://github.com/taskcluster/taskcluster/pull/2571/commits/a5072f023182373398c02091328bb1ece3f67ac7 from #2571.

djmitche:
And note to self, #2600.

PR #2608, opened 2020-04-06
Set up monitors for all TC services' `db` property

djmitche:
This is the other half of #2522.  The first commit here is cherry-picked from master, and the second updated to include #2601.  We'll count #2601 fixed when this merges back to master.


djmitche:
Hm, I wonder why none of those checks are complete in GitHub - they are all complete in TC.

PR #2609, opened 2020-04-06
Update handlers.js

owlishDeveloper:
https://sentry.prod.mozaws.net/operations/taskcluster-community/issues/7766271/?query=is:unresolved

### DIFF
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -503,7 +503,7 @@ async function statusHandler(message) {
     try {
       const url = this.queueClient.buildUrl(this.queueClient.getArtifact, taskId, runId, textArtifactName);
       const res = await utils.throttleRequest({url, method: 'GET'});
-      
+
       if (res.status >= 400 && res.status !== 404) {
         await this.createExceptionComment({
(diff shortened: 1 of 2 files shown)


services/github/src/handlers.js
```diff
@@ -492,7 +492,7 @@ async function statusHandler(message) {
   try {
     const taskDefinition = await this.queueClient.task(taskId);
     const {customCheckRun} = (taskDefinition.extra && taskDefinition.extra.github) || {};
-    const {textArtifactName} = customCheckRun || CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME;
```
djmitche, line 495: This misses the case where `customCheckRun` is undefined, such as when there's no `task.extra`.

owlishDeveloper: Isn't that taken care of on the previous line? `const {customCheckRun} = (taskDefinition.extra && taskDefinition.extra.github) || {};`

owlishDeveloper: Checked - it doesn't. Now it should be better

PR #2613, opened 2020-04-06
Bug 1627806 - more info on connection counts

djmitche:
Bugzilla Bug: [1627806](https://bugzilla.mozilla.org/show_bug.cgi?id=1627806)


djmitche:
(oops, no changelog needed, but I won't waste CI cycles making a silent changelog)

PR #2614, opened 2020-04-06
Bug 1627807 - add ability to import roles

djmitche:
Roles are in a container in Azure.  We're transitioning to storing them in a single row in postgres, due to some complex inter-dependencies between roles requiring atomic updates.

Bugzilla Bug: [1627807](https://bugzilla.mozilla.org/show_bug.cgi?id=1627807)

### DIFF
--- a/infrastructure/tooling/src/importer/util.js
+++ b/infrastructure/tooling/src/importer/util.js
@@ -47,6 +47,9 @@ exports.ALLOWED_TABLES = [
   'TaskclusterIntegrationOwners',
   'TaskclusterChecksToTasks',
   'TaskclusterCheckRuns',
+
+  // (actually a continer, but that's OK)
+  'Roles',
 ];
 
 // For certain tables, we would like to import faster to make sure we don't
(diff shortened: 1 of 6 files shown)


djmitche:
Oh, and I tested this in my dev environment and the auth service was able to read my "fake" roles entity just fine.

infrastructure/tooling/src/importer/util.js
```diff
@@ -47,6 +47,9 @@ exports.ALLOWED_TABLES = [
   'TaskclusterIntegrationOwners',
   'TaskclusterChecksToTasks',
   'TaskclusterCheckRuns',
+
+  // (actually a continer, but that's OK)
+  'Roles'
```
helfi92, line 52: Putting it here will call `importTable` which will try to import it as an azure table in addition to `importRoles`. Was that intended?

djmitche: Oh, no, it wasn't -- I will add a special case for that.

djmitche: It looks like the loop that creates tasks is over a variable created from the `tables` properties in `services/*/azure.yml`.  So I think this is actually OK.

djmitche:
I did.  It weirdly only imported 11 rows for QueueTasks but LOTS of rows for QueueTaskGroups, but that seems to be the state of the the dustindev Azure storage account so I guess it's OK :)

I'll look at the CI failures before merging.

PR #2618, opened 2020-04-07
Set a lower default for cpu requests in dev environments.

djmitche:
Dev environments are almost always idle, using basically none of their
CPU request but causing us to provision k8s nodes for the default amount
(50m). That's wasting resources.  Let's dial it back by default, and
anyone who needs more CPU can increase the values in their config.


djmitche:
I filed a bug for that -- in fact, to do the opposite and adopt the settings that they have determined as our defaults.

PR #2619, opened 2020-04-08
Bug 1525419 - escape env variable contents on Windows properly

petemoore:
Bugzilla Bug: [1525419](https://bugzilla.mozilla.org/show_bug.cgi?id=1525419)

This should fix the escaping of special characters in environment variables on Windows (except newline/linefeed characters which is tracked in [bug 1495760](https://bugzilla.mozilla.org/show_bug.cgi?id=1495760)).

We'll need to wait for the CI to complete - I haven't manually tested on Windows.

### DIFF
--- a/changelog/bug-1525419.md
+++ b/changelog/bug-1525419.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: bug 1525419
+---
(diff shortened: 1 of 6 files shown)


petemoore:
> (and needs a changelog)

d'oh - looks like I did a `git add .` from the `workers/generic-worker` directory 🤦‍♂

PR #2625, opened 2020-04-08
Run gofmt / go lint at the top level, like for JS

djmitche:
This makes it easy to see when tests pass but there are some lints to fix, without having to open up logs and search for issues.

This PR also removes the check added in #1492, which is no longer necessary.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -92,6 +92,9 @@ tasks:
         others:
         - name: lint
           command: yarn lint
+        - name: lint:go
+          image: taskcluster/node-and-go:node${node}-go${go.release}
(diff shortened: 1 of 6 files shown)


djmitche:
Hm, that netlify issue is vexing :)

helfi92:
> Hm, that netlify issue is vexing :)

Maybe it's because we're running on a old build image of netlify that's causing issues? Not sure. We can try updating to the new one.

djmitche:
How do we do that?

helfi92:
https://app.netlify.com/sites/taskcluster-web/settings/deploys

You should be an "owner" now.

PR #2626, opened 2020-04-08
Move websocktunnel into the monorepo

### LINKED ISSUE #2562: [websocktunnel-monorepo] migrate websocktunnel to the monorepo

djmitche:
More code!  This brings the history of websocktunnel along, but I did run `git-filter-branch` to strip out the `vendor/` directory to save space.  I think that's OK since it's not original content anyone created, and is redundant to `Gopkg.toml` anyway.

This also centralizes management of the TC version in `internal.Version`.

Github Bug/Issue: Fixes #2562


imbstack:
this has a bunch of conflicts now. should I wait until it is rebased to review or just do it now?

djmitche:
I'll rebase, just a minute..

djmitche:
(I merged instead since rebase won't work with all that history)

tools/websocktunnel/.taskcluster.yml
```diff
@@ -0,0 +1,120 @@
+version: 1
```
imbstack, line 1: we could rm this file

djmitche: thanks, nice spot

PR #2628, opened 2020-04-09
Add opt-out for Matrix notifications being "notices" rather than messages

SimonSapin:
Riot appears not to trigger a desktop notification when a notice contains my username and a "watched keyword". This is configurable in Riot’s settings, but sending a normal message instead would make it easier for people to be notified of failures they are interested in.

This is a boolean because Matrix’s other message types do not seem relevant here: emote, image, file, audio, location, video.

This commit is a first pass at adding this, however it is done entirely "blind" since my laptop does not have the required Node version, so I can’t easily run tests or regenerate generated files. I’d appreciate someone taking over.

### DIFF
--- a/services/notify/schemas/v1/matrix-request.yml
+++ b/services/notify/schemas/v1/matrix-request.yml
@@ -21,6 +21,12 @@ properties:
     description: |
       Text that will be rendered by matrix clients that support the given
       format in that format. For instance, `<h1>Header Text</h1>`.
+  notice:
+    type: boolean
+    default: true
+    description: |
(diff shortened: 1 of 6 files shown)


imbstack:
I'll be happy to take this over and get it finished! Thanks for figuring out what the issue was.

imbstack:
I'm going to take this as a starting point and work on it in my own branch. Let's add you as a contributor before I forget since this code will make it into master via the other PR.

@all-contributors please add @SimonSapin for code.

imbstack:
@all-contributors please add @SimonSapin for code.

PR #2629, opened 2020-04-09
Fix command in netlify after changing build image

helfi92:


ui/src/views/Tasks/TaskIndex/ListNamespaces/index.jsx
```diff
@@ -232,7 +232,7 @@ export default class ListNamespaces extends Component {
             </Fragment>
           )}
           {!loading && !hasNamespaces && !hasIndexedTasks && isSinglePath && (
-            <Typography variant="body2">No items for this page.</Typography>
```
djmitche, line 235: Accidentally included?

helfi92: Not meant to be merged but rather just to force a netlify trigger.

PR #2633, opened 2020-04-09
Refactor Pg Importer to display and log its status

### LINKED ISSUE #2631: Pg Importer Improvements
* give more runtime metrics, maybe using a display similar to the perf-tester.  
  * how full that 5-chunk buffer is
  * rows-per-second for each concurrent copy operation, with total
  * number of active Pg connections
  * regular logging to a file of some sort so we can go back and figure what happened at a particular time
* be resilient to connection errors
* configurability
  * concurrency
  * number of pg connections
  * tables to copy

djmitche:
Github Bug/Issue: Fixes #2631

Running this locally suggests that Azure is the rate-limiting factor, not postgres, in getting these imports done.  We'll see what it looks like on the big import..


infrastructure/tooling/src/importer/importer.js
```diff
@@ -1,49 +1,149 @@
...
+  }
 
-      await writeToPostgres(tableName, entities, db, ALLOWED_TABLES);
+  async run() {
+    await this.truncateTable(this.tableName);
+    this.rowsProcessed(0);
 
-      rowsProcessed = count;
+    const BATCH_SIZE = 1000;
+    const BUFFER_SIZE = 2000;
```
helfi92, line 96: I think the buffer size was set to 5. Looking at the latest CPU usage from the db, we were hitting 40-60%. Do we really want to up that number to 2000?

infrastructure/tooling/src/importer/importer.js
```diff
@@ -1,49 +1,149 @@
...
+
+  // support truncating each table exactly once
+  const truncated = {};
+  const truncateTable = tableName => {
+    if (!truncated[tableName]) {
+      truncated[tableName] = (async () => {
+        const pgTable = postgresTableName(tableName);
+        await db._withClient('admin', async client => {
+          await client.query(`truncate ${pgTable}`);
+        });
```
helfi92, line 47: Should this be wrapped a in a try catch statement to account for the case where we are not able to connect the db?

infrastructure/tooling/src/importer/util.js
```diff
@@ -137,20 +96,17 @@ exports.readAzureTableInChunks = async function* ({azureCreds, tableName, utils,
       const results = await table.queryEntities(tableName, tableParams);
```
helfi92, line 96: Would this throw if we are not able to connect to azure? Should we handle this case?

infrastructure/tooling/src/importer/util.js
```diff
@@ -1,35 +1,24 @@
...
-};
-
-// number of table (or table-chunk) imports to run in parallel
-exports.CONCURRENCY = 30;
-
 // tables that are allowed to migrate
 exports.ALLOWED_TABLES = [
-  // 'Clients',
-  // 'Hooks',
+  'Clients',
```
helfi92, line 12: Most imports will likely run on a dev deployment to make sure they work before running it on the actual deployment. Should we have a flag that includes/excludes signing/encrypted tables?

infrastructure/tooling/src/importer/config.js
```diff
@@ -0,0 +1,34 @@
+const chalk = require('chalk');
+
+class Config {
+  constructor({monitor, order}) {
+    const env = (name, dfl, convert = x=>x) => {
```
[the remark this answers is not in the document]
djmitche: Huh, it worked even without that.  Weird!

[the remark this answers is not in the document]
djmitche: Default.  Vim highlights `default` as a keyword, which made me want to avoid using it, but it seems to work.

infrastructure/tooling/src/importer/importer.js
djmitche: That's now a number of rows, not a number of pages, so it's actually a reduction in buffer size from 5 pages = 5000 rows.  The buffer never seems to fill (that's the buffer % in the output), so I don't think the size is particularly important.

infrastructure/tooling/src/importer/util.js
djmitche: Yeah, I think there are a number of ways this script could fail and adding handling for all of those feels like it's not a good use of time -- it's just adding code that can't really be tested and will probably just fail in a different way if something does go wrong.  Or it will mask errors and we'll wait for hours for the import to complete when it's sitting there hitting the same error over and over.  We've had good luck so far, aside from moments when we deliberately killed the pg server, so I think we're OK where we are..

infrastructure/tooling/src/importer/util.js
djmitche: sounds good

infrastructure/tooling/src/importer/config.js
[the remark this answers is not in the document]
djmitche: Oh, actually eslint doesn't like it :)

djmitche:
piping to cat won't work -- you'll just need to output it to the terminal, and I think you might need a big enough terminal or you'll get that kind of thing happening

djmitche:
Rebased.  I don't think it hurts to merge this and then do a test import from master.  But we should definitely test it before the TCW!

PR #2634, opened 2020-04-09
Add test to tc-lib-entities for modify

helfi92:


djmitche:
Haha, and now that we're back on `master` we need changelogs again :)

PR #2617, opened 2020-04-07
[UI] Search terms with empty table results 

rohanharikr:
Github Bug/Issue: Fixes #2607

@helfi92 Thanks a ton for the help!

Issue #2124, opened 2019-12-03
[UI] Deep link search in /worker-manager


helfi92:
**To Reproduce**

1. Go to https://firefox-ci-tc.services.mozilla.com/worker-manager
2. Type "gecko-t/" in the search bar located in the app bar and then press enter.
3. Click on any worker pool id
4. Click back

**Expected behavior**
The list only contains worker pools that start with "gecko-t/".

**Taskcluster version**
23.0.0

**Additional context**
In order to fix this issue, we would need to use a query parameter (e.g., search). An example would be `https://firefox-ci-tc.services.mozilla.com/worker-manager?search=gecko-t%2F`.

helfi92:
Assigned to @Manish-Giri.

helfi92:
@Manish-Giri Are you still interested in finishing this up?

gaborszekely:
@helfi92 Could not replicate this bug.

Manish-Giri:
@helfi92 Yes, I am. Sorry for the delay, I had to step away for a bit, I'll get back on both the issues right away.

helfi92:
> @helfi92 Could not replicate this bug.

I just tried and was able to replicate. What exactly were you not able to replicate?

> @helfi92 Yes, I am. Sorry for the delay, I had to step away for a bit, I'll get back on both the issues right away.

Awesome. Thanks!

Manish-Giri:
@helfi92 Hello! I've been meaning to get your advice on an issue with `/worker-manager`. Whenever I access this page on my local machine, I get this error - 

![Screenshot 2020-06-26 02 02 47](https://user-images.githubusercontent.com/11348778/85826419-8da7b100-b752-11ea-876c-2ee8b63f12ac.png)

I looked around the Network tab, and found this response which causes the error -

![Screenshot 2020-06-26 02 03 38](https://user-images.githubusercontent.com/11348778/85826519-b7f96e80-b752-11ea-918a-8921c3bb5df4.png)

I just wanted to check with you, in case there is some additional setup I should be running to fix this. I already have postgres running in Docker, `web-server` on port 3050, and finally the UI at 5080.

helfi92:
Sorry for the trouble!

@imbstack Should @Manish-Giri use your development environment to continue his UI work or is there another workaround? Thanks.

PR #2655, opened 2020-04-14
[UI] Deep Link search in /secrets

### LINKED ISSUE #2123: [UI] Deep link search in /secrets
**To Reproduce**

1. Go to https://firefox-ci-tc.services.mozilla.com/secrets/
2. Type "project/" in the search bar located in the app bar and then press enter.
3. Click on any secret
4. Click back

**Expected behavior**
The list of secrets only contains secrets that start with "project/".

**Taskcluster version**
23.0.0

**Additional context**
In order to fix this issue, we would need to use a query parameter (e.g., search). An example would be `https://firefox-ci-tc.services.mozilla.com/secrets?search=project%2F`.

Manish-Giri:
@helfi92 I noticed the intended behavior for `/secrets` already exists in `/hooks`, and can be seen here (as an example) - https://firefox-ci-tc.services.mozilla.com/hooks?search=project-

So I figured it might be worth a try to replicate the functionality already present in `ListHooks/index.jsx`. 

This change uses the `qs` package to parse out the search term from the query string, and pushes it onto `history`. Thus, once you do the following steps -
1. Visit `/secrets` and search for an item, example `project/`
2. Click on a secret
3. Go back

The browser URL now contains the search term you searched for -

Screenshot 2020-04-13 23 34 30

whereas it is currently just `/secrets` after you go back -

Screenshot 2020-04-13 22 08 03

However, the page still lists all the available results at `/secrets`. 

I wasn't quite sure about why this behavior at `/secrets` is different from `/hooks`, so I figured it would be nice to get your thoughts on this, if possible, before I investigate further. I wonder if the `refetch()` in the Submit Handler from [here](https://github.com/taskcluster/taskcluster/blob/master/ui/src/views/Secrets/ViewSecrets/index.jsx#L46-L53) makes any difference. However, the [docs](https://www.apollographql.com/docs/react/data/queries/#refetching) from Apollo GraphQL mention this just refreshes the query results in response to a user action, and as far as I can see, the `filter` still lists the correct search string -

Would be really great if I could some thoughts/tips about the next steps!

Screenshot 2020-04-13 23 33 02

Github Bug/Issue: Fixes #2123

### DIFF
--- a/ui/src/views/Secrets/ViewSecrets/index.jsx
+++ b/ui/src/views/Secrets/ViewSecrets/index.jsx
@@ -33,10 +33,6 @@ import secretsQuery from './secrets.graphql';
   },
 }))
 export default class ViewSecrets extends Component {
-  state = {
-    secretSearch: '',
-  };
-
   handleSecretSearchSubmit = async secretSearch => {
     const {
       data: { refetch },
@@ -51,9 +47,14 @@ export default class ViewSecrets extends Component {
         : null,
     });
 
-    this.props.history.push(
-      `/secrets${secretSearch ? `?search=${secretSearch}` : ''}`
-    );
+    const query = qs.parse(window.location.search.slice(1));
+
+    this.props.history.push({
+      search: qs.stringify({
+        ...query,
+        search: secretSearch,
+      }),
+    });
   };
 
   handleCreate = () => {
@@ -64,6 +65,8 @@ export default class ViewSecrets extends Component {
     const {
       data: { fetchMore },
     } = this.props;
+    const query = qs.parse(this.props.location.search.slice(1));
+    const secretSearch = query.search;
 
     return fetchMore({
       query: secretsQuery,
@@ -73,10 +76,10 @@ export default class ViewSecrets extends Component {
           cursor,
           previousCursor,
         },
-        filter: this.state.secretSearch
+        filter: secretSearch
           ? {
               name: {
-                $regex: escapeStringRegexp(this.state.secretSearch),
+                $regex: escapeStringRegexp(secretSearch),
                 $options: 'i',
               },
             }
@@ -102,7 +105,8 @@ export default class ViewSecrets extends Component {
       description,
       data: { loading, error, secrets },
     } = this.props;
-    const { searchTerm } = this.state;
+    const query = qs.parse(this.props.location.search.slice(1));
+    const secretSearch = query.search;
 
     return (
       <Dashboard
@@ -111,6 +115,7 @@ export default class ViewSecrets extends Component {
         search={
           <Search
             disabled={loading}
+            defaultValue={secretSearch}
             onSubmit={this.handleSecretSearchSubmit}
             placeholder="Secret contains"
           />
--- a/ui/src/views/Secrets/ViewSecrets/index.jsx
+++ b/ui/src/views/Secrets/ViewSecrets/index.jsx
@@ -18,14 +18,21 @@ import secretsQuery from './secrets.graphql';
 
 @hot(module)
 @graphql(secretsQuery, {
-  options: () => ({
-    fetchPolicy: 'network-only',
-    variables: {
-      secretsConnection: {
-        limit: VIEW_SECRETS_PAGE_SIZE,
+  options: props => {
+    const { search } = qs.parse(props.location.search.slice(1));
+
+    return {
+      fetchPolicy: 'network-only',
+      variables: {
+        secretsConnection: {
+          limit: VIEW_SECRETS_PAGE_SIZE,
+        },
+        filter: search
+          ? { name: { $regex: escapeStringRegexp(search), $options: 'i' } }
+          : null,
       },
-    },
-  }),
+    };
+  },
 })
 @withStyles(theme => ({
   plusIconSpan: {
@@ -125,7 +132,7 @@ export default class ViewSecrets extends Component {
           <ErrorPanel fixed error={error} />
           {secrets && (
             <SecretsTable
-              searchTerm={searchTerm}
+              searchTerm={secretSearch}
               onPageChange={this.handlePageChange}
               secretsConnection={secrets}
             />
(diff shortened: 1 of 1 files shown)


ui/src/views/Secrets/ViewSecrets/index.jsx
```diff
@@ -36,6 +37,18 @@ export default class ViewSecrets extends Component {
...
+    const searchString = qs.parse(nextProps.location.search, {
+      ignoreQueryPrefix: true,
+    }).search;
+
+    if (prevState.secretSearch !== searchString) {
+      return { secretSearch: searchString };
+    }
+
+    return null;
+  }
```
helfi92, line 50: I just created https://github.com/taskcluster/taskcluster/pull/2660/commits/bd9db1a9905a204a816379468d401394e6ef740c to move away from this kind of logic 😅 See how there's a simpler way to do this without relying on internal state.

ui/src/views/Secrets/ViewSecrets/index.jsx
```diff
@@ -51,6 +64,10 @@ export default class ViewSecrets extends Component {
         ? { name: { $regex: escapeStringRegexp(secretSearch), $options: 'i' } }
         : null,
     });
+
+    this.props.history.push(
+      `/secrets${secretSearch ? `?search=${secretSearch}` : ''}`
```
helfi92, line 69: Let's encode the search term before pushing it to history similar to how it's done in https://github.com/taskcluster/taskcluster/pull/2660/commits/bd9db1a9905a204a816379468d401394e6ef740c.

helfi92: On a second thought, there's no need to encode if we use the `stringify` function from the `qs` package. https://github.com/taskcluster/taskcluster/blob/978b8dc93e54d5ec34cfbd06cad6c88796f53f5f/ui/src/views/Hooks/ListHooks/index.jsx#L36-L45

Manish-Giri:
@helfi92 Sorry for the delay. I have been trying to explore React Router, especially `history` for a while now, to figure out what the possible cause might be, for the incorrect behavior of the back button after a search, but couldn't come up with a concrete reason.

Essentially, in the latest commit, I made the following changes -
1. Replaced internal state with the search being derived from `window.location.search`, following the approach in your recent commit, as you pointed out.
2. Also updated the PageChangeHandler accordingly, in addition to the Submit Handler, to not rely on `state`.
3. Added a `defaultValue` prop to the `Search` component in the `render()` function (I was testing to see if this made a difference...).

But, even though the search results correctly display the search term, this behavior doesn't persist after pressing the back button (displayed in screenshots below) -

1. Search for `cia/` -

Screenshot 2020-04-26 05 12 34

2. Click on one of the search results and go back -

Screenshot 2020-04-26 05 13 02

helfi92:
@Manish-Giri Let me know when you want me to review :-)

Manish-Giri:
@helfi92 Thank You! I think I understand your suggested changes. I’ll fix this right away and re-submit.

helfi92:
Awesome! Thanks @Manish-Giri :-)

Manish-Giri:
@helfi92 Sorry for the delay, once again :/

The search works perfectly now - only secrets that include the search term that was searched for originally, are now shown, when you click on the searched secret, and go back. Here is a screenshot for search of`project/` -  only secrets with `project/` are now shown when you go back -

Screenshot 2020-05-18 04 49 55

However, while rebasing with the latest `master`, I noticed there was some work done on [this file](https://github.com/taskcluster/taskcluster/pull/2617/files) as part of #2617, which included maintaining a `state` in the `ViewSecrets` component. Since we were trying to eliminate the need for maintaining a state in the component, I removed `state`, while keeping other changes from that PR. 
The only thing I was a bit confused about is this line here - 
https://github.com/taskcluster/taskcluster/blob/278a92188b830699b7a75be0badeb96106f5d490/ui/src/views/Secrets/ViewSecrets/index.jsx#L101

which, I think, attempts to destructure `searchTerm` from `state` and then pass it on as a prop to the `SecretsTable` component -
https://github.com/taskcluster/taskcluster/blob/278a92188b830699b7a75be0badeb96106f5d490/ui/src/views/Secrets/ViewSecrets/index.jsx#L119

However, `state` only contains a `secretSearch` property, as far as I could tell, so not sure what's going on there. I assume this is meant to capture the search term, if so, I simply used the existing search term we already obtained from `qs.parse` and then pass it on as the `searchTerm` prop. 

```
const query = qs.parse(this.props.location.search.slice(1));
const secretSearch = query.search;
```
If that is not the case, I can make changes accordingly. 

Thank You!

Manish-Giri:
@helfi92 I rebased it with the latest `master`. I didn't know if the PR needed a changelog, so I didn't add it in, looks like the missing changelog might be one of the errors. If you think it needs the changelog, please let me know an appropriate changelog message and I'll add it in. Thank You!

Manish-Giri:
@helfi92 Thank You so much for all your help with this PR. Should I now start looking at #2124 (fix search in `/worker-manager`), since the underlying issue is the same?

As a side note, I wonder if it would be possible to include me in the [Taskcluster contributors list on the repo](https://github.com/taskcluster/taskcluster#contributors). Sorry if this is a restricted list, but I figured it wouldn't hurt to ask, it would just mean a lot to me to have been able to contribute to Taskcluster 'officially'!

helfi92:
> @helfi92 Thank You so much for all your help with this PR. Should I now start looking at #2124 (fix search in `/worker-manager`), since the underlying issue is the same?

That would be so awesome. I'll go ahead and assign you to the issue now. Thanks for helping as always!

> 
> As a side note, I wonder if it would be possible to include me in the [Taskcluster contributors list on the repo](https://github.com/taskcluster/taskcluster#contributors). Sorry if this is a restricted list, but I figured it wouldn't hurt to ask, it would just mean a lot to me to have been able to contribute to Taskcluster 'officially'!

Of course! You should definitely be on that list. I'm on it :)

helfi92:
@all-contributors please add @Manish-Giri for code.

PR #2664, opened 2020-04-14
Add a SiteSpecific component and use it in a few places

### LINKED ISSUE #2395: [docs] Provide a way to link to site-specific docs
The docs site is currently identical for every deployment.  But deployments are different, and it would make sense to have some links to specific documentation for each deployment.

Let's add a UI configuration parameter, similar to `ui.banner_message`, that gives some links that are substituted into the docs, if they are set.

I would like to allow multiple links, since there are several places users may want to substitute their own docs.  At least:
 * A top-level link on https://docs.taskcluster.net/docs
 * A link describing how to get credentials, for use in https://docs.taskcluster.net/docs/tutorial (useful for #2394)

We will also need to build a React component that can display content if the link exists, and do nothing if the link does not exist.  Something like

```
<IfSiteSpecificDoc doc="deployment-info">
  <h2>Information About This Deployment</h2>
  <p>You can find out more about this deployment of Taskcluster <SiteSpecificDocLink doc="deployment-info">here</SiteSpecificDocLink>.</p>
</IfSiteSpecificDoc>
```

djmitche:
The component has a distinctive color that will help it stand out as
site-specific.  The text of the component should also suggest this.

The component reads values from deployment config.  If the value isn't
set, then the component does not appear.

I addressed a few instances of root URL in the docs, plus the github app
since it came up in [Bug
1629623](https://bugzilla.mozilla.org/show_bug.cgi?id=1629623).

Github Bug/Issue: Fixes #2395


ui/docs/manual/deploying/ui.mdx
```diff
@@ -0,0 +1,87 @@
...
+```
+
+These values are also used occasionally elsewhere in the UI.
+
+## Login Strategies
+
+The taskcluster-ui and taskcluster-web-server services work together to support configurable login strategies for the Taskcluster UI.
+These are configured with the `web_server.ui_login_strategies` and `ui.ui_login_strategy_names` Helm variables.
+
+The first, `web_server.ui_login_strategies`, is a JSON string containing configuration for the desired strategies.
```
imbstack, line 34: maybe add `defined below`

ui/src/components/SiteSpecific/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+        err.code = 'site-specific-variable-not-set';
+        throw err;
+      }
+
+      return result;
+    });
+  } catch (err) {
+    if (err.code === 'site-specific-variable-not-set') {
+      // ignore the whole card..
+      return null;
```
imbstack, line 61: better to ignore the whole thing or just render `UNKNOWN` in place of var?

djmitche: That's kind of the whole point -- otherwise we'd have these confusing purple cards with "Click on UNDEFINED to .."

djmitche: We *might* run into cases where we want some kind of "else" functionality, though.  I'm not quite sure how to write that in JSX, but something like

```jsx
<SiteSpecific else={"The list of providers is different for each deployment.  Consult your deployment documentation or ask your Taskcluster administrator for the list of providers for your deployment."}>
  The providers defined on this deployment are documented [here](%worker_manager_provider_docs_url%).
</SiteSpecific>
```

We can add that when it becomes necessary.

PR #2666, opened 2020-04-15
Add a function to generate links in clients-go readme

### LINKED ISSUE #2268: automatically generate the links in clients/client-go/README.md
#2267 identifies some bad links in this file.  Rather than try to maintain them automatically, let's make this a part of the code-generation process in `clients/client-go`.

To solve this, first get this repository out and get set up to where `yarn generate` will run without error.  Then look at the [code-generation process](https://github.com/taskcluster/taskcluster/tree/master/clients/client-go#generating) in `clients/client-go`.  Update that to automatically fix the list of links in the README.md file.

ajinkabeer:
Fixes #2268 

This PR addresses issue #2268, which is to generate godoc links in README.me file when `yarn generate` is run in the directory `client/client-go`.

This is the first time I am writing a piece of code in Golang. I was able to generate the links in the readme file when ``yarn generate`` is run. I would like to get some feedback and nudge me in the right direction if this is not what we need to achieve, please. 

@djmitche

### DIFF
--- a/clients/client-go/README.md
+++ b/clients/client-go/README.md
@@ -13,6 +13,46 @@ the API methods.
 
 For a general guide to using Taskcluster clients, see [Calling Taskcluster APIs](https://docs.taskcluster.net/docs/manual/using/api).
 
+This library provides the following packages to interface with Taskcluster:
+
+### HTTP APIs
+
+<!--HTTP-API-start-->
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcauthevents
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcgithubevents
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tchooksevents
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcnotifyevents
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcqueueevents
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcworkermanagerevents
+ <!--HTTP-API-end-->
+
+### AMQP APIs
+
+<!--AMQP-API-start-->
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcauth
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcgithub
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tchooks
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcindex
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcnotify
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcpurgecache
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcqueue
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcsecrets
+
+* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/tcworkermanager
+ <!--AMQP-API-end-->
+
 ### Setup
 
 Before invoking API methods, create a client object corresponding to the service you wish to communicate with.
--- a/clients/client-go/codegenerator/model/model.go
+++ b/clients/client-go/codegenerator/model/model.go
@@ -56,6 +56,28 @@ type APIDefinition struct {
 	SchemaURL      string
 }
 
+func GenerateGodocLinkInReadme(amqpLinks string, httpLinks string) {
+
+	path := `../../README.md`
+	formattedContent, err := ioutil.ReadFile(path)
+	if err != nil {
+		panic(err)
+	}
+
+	httpAPI := `<!--HTTP-API-start-->` +
+		amqpLinks +
+		` <!--HTTP-API-end-->`
+
+	AmqpAPI := `<!--AMQP-API-start-->` +
+		httpLinks +
+		` <!--AMQP-API-end-->`
+
+	formattedContent = regexp.MustCompile(`(<!--(AMQP-API-start:?(\w*?):?(\w*?)?)-->)([\s\S]*?)(<!--AMQP-API-end-->)`).ReplaceAll(formattedContent, []byte(AmqpAPI))
+	exitOnFail(ioutil.WriteFile(path, formattedContent, 0644))
+	formattedContent = regexp.MustCompile(`(<!--(HTTP-API-start:?(\w*?):?(\w*?)?)-->)([\s\S]*?)(<!--HTTP-API-end-->)`).ReplaceAll(formattedContent, []byte(httpAPI))
+	exitOnFail(ioutil.WriteFile(path, formattedContent, 0644))
+}
+
 func exitOnFail(err error) {
 	if err != nil {
 		fmt.Printf("%v\n%T\n", err, err)
@@ -178,6 +200,7 @@ type APIDefinitions []*APIDefinition
 // and writes them out as go code.
 func (apiDefs APIDefinitions) GenerateCode(goOutputDir string) {
 	for i := range apiDefs {
+
 		apiDefs[i].PackageName = "tc" + strings.ToLower(apiDefs[i].Data.Name())
 		// Used throughout docs, and also methods that use the class, we need a
 		// variable name to be used when referencing the go type. It should not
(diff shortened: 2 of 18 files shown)


ajinkabeer:
Thanks for the review., I just saw it after I pushed the latest changes. Yes, so we don't need to modify the badge right now.  Currently, so where should we add the `HTTP` and `AMQP ` sections? 

<!--HTTP-SECTION-START>
{links}
<!--HTTP-SECTION-END>

If I am thinking correctly, we should add the links between comments, right? and also are we hardcoding these links in `model.go`?

djmitche:
A little further down in the file, after the introduction.  Take a look at where they were in the commit I linked.

clients/client-go/README.md
```diff
@@ -641,7 +657,7 @@ func main() {
 	fmt.Println("")
 	fmt.Printf("curl -L https://queue.taskcluster.net/v1/task/%v/runs/0/artifacts/public/logs/live.log | gunzip\n", taskID)
 }
-```
+````
```
djmitche, line 660: There are a lot of unrelated changes in this README -- and this one in particular looks incorrect.  Why are they here?

ajinkabeer: Oh, sorry. It must be a typo from my side.

ajinkabeer: Is there a way to log the `apiDefs` inside `GenerateCode `. I tried ` fmt.Println(apiDefs[i])`. It did not work.

djmitche:
> Is there a way to log the `apiDefs` inside `GenerateCode `. I tried ` fmt.Println(apiDefs[i])`. It did not work.

If you're running `yarn generate`, then that output can be a bit hard to find in the noise.  If you change into the `clients/client-go` directory and run `go generate ./...`, you should see your logging (and a bunch of other logging too).  I find it a little easier to use `log.Printf` and `log.Println` than the `fmt` functions.

ajinkabeer:
Thanks a lot!. I will try those.

ajinkabeer:
Hey, I managed to use the `packageName` from `apiDefs ` to generate links in `README`. However, I was not able to run `yarn go:lint`. I tried installing it in several ways but couldn't. I think the link is broken on `golangci-lint'`s install script or something. Right now when I run ``yarn lint:go``, it does not succeed because there's no `golangci-lint'` installed. Looking to find a solution for that.

djmitche:
Yeah, that seems to be a pretty common issue with go tools.  It looks like there's only [one lint error](https://community-tc.services.mozilla.com/tasks/MdCI2l7lRK2umroY86544Q/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FMdCI2l7lRK2umroY86544Q%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L44):
```
clients/client-go/codegenerator/model/model.go:76:18: Error return value of `ioutil.WriteFile` is not checked (errcheck)
	ioutil.WriteFile(path, formattedContent, 0644)
	                ^
```
so maybe it's easiest to just fix that and not try to run it locally?

djmitche:
Looking at the [yarn generate logs](https://community-tc.services.mozilla.com/tasks/UCjDbhixSKyShpHvy0p8tQ/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FUCjDbhixSKyShpHvy0p8tQ%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L6248) it looks like there's an extra space before `<!--AMQP-API-end-->`.  Also, it looks like it's removing a newline from every service file.

You can avoid the `yarn generate` check failing by committing the results of running `yarn generate` by the way.

changelog/issue-#2268.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
+level: patch
```
djmitche, line 2: Since this is just a change to documentation, I think `silent` here will do

changelog/issue-#2268.md
```diff
@@ -0,0 +1,5 @@
...
+reference: issue #2268
```
djmitche, line 3: remove the `#` character here

changelog/issue-#2268.md
ajinkabeer: Alright. I will update it.

changelog/issue-#2268.md
ajinkabeer: Got it. Thanks.

ajinkabeer:
> Looking at the [yarn generate logs](https://community-tc.services.mozilla.com/tasks/UCjDbhixSKyShpHvy0p8tQ/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FUCjDbhixSKyShpHvy0p8tQ%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L6248) it looks like there's an extra space before `<!--AMQP-API-end-->`. Also, it looks like it's removing a newline from every service file.
> 
> You can avoid the `yarn generate` check failing by committing the results of running `yarn generate` by the way.

I think ``yarn generate`` generates many files in other directories as file. So should I include them as well or just the README.md file with the updated links? and also the diff in README.md changed. Don't know how it happened though :(

djmitche:
> I think yarn generate generates many files in other directories. So should I include them as well or just the README.md file with the updated links? and also the diff in README.md changed. Don't know how it happened though :(

It does, but those files should already exist and be exactly the same as what `yarn generate` will generate.  We take the tactic of checking in our generated code.

> Plus, I will also have to change the comment tags for HTTP API's . Now it's something that doesn't make any sense.

The generation should not remove those tags, but just replace the content between them.

ajinkabeer:
I am trying to insert the links between them, but it always replaces it. Do you think it is because of my regex expression?

djmitche:
> I am trying to insert the links between them, but it always replaces it. Do you think it is because of my regex expression?

I ran `yarn generate` on the latest version of your push and it seems to have worked: the doc links are updated and the markers are still there.

There's one small thing to fix, though.  It looks like the README has bene converted to "DOS style newlines", which means that the code to generate the links in the top-level README can't figure out the title anymore, which causes this:

```diff
diff --git a/README.md b/README.md
index a433e74bb..96da95044 100644
--- a/README.md
+++ b/README.md
@@ -34,17 +34,17 @@
 This repository is used to develop, build, and release the Taskcluster services.
 It is not possible to run a full Taskcluster deployment directly from this repository, although individual services can be run for development purposes.
 
 ## Table of Contents
 
 <!-- TOC BEGIN -->
 * [Clients](clients#readme)
     * [Taskcluster Client for JS](clients/client#readme)
-    * [Taskcluster Client for Go](clients/client-go#readme)
+    * [clients/client-go](clients/client-go#readme)
     * [Taskcluster Client for Python](clients/client-py#readme)
     * [Taskcluster Client for Shell](clients/client-shell#readme)
     * [Taskcluster Client for Web](clients/client-web#readme)
 * [Taskcluster Database](db#readme)
 * [Development Documentation](dev-docs#readme)
 * [Infrastructure](infrastructure#readme)
     * [Docker Images](infrastructure/docker-images#readme)
     * [References](infrastructure/references#readme)
```

It might be good to squash your changes in this PR into a single commit, then remove all the changes to `clients/client-go/README.md` in that commit, then run yarn generate and commit the results -- without opening the README in an editor again :)

ajinkabeer:
>  It might be good to squash your changes in this PR into a single commit, then remove all the changes to clients/client-go/README.md in that commit, then run yarn generate and commit the results -- without opening the README in an editor again :)

Alright. I can squash it into a single commit. So I should restore the `README `to the original version without the marker tags or with marker tags and then run `yarn generate`?

djmitche:
Yes, sorry, please include the markers too :)

ajinkabeer:
> There's one small thing to fix, though. It looks like the README has bene converted to "DOS style newlines", which means that the code to generate the links in the top-level README can't figure out the title anymore, which causes this:

Squashed the commits and restored the `README`. However, the top-level README still can't figure out the title. Phew!

djmitche:
The README is still showing DOS-style newlines.  That may be invisible in your editor, especially if it's a Windows editor.  But tools like notepad++ can produce normal newlines.  Alternately, there's a linux utility called `dos2unix` that may be able to fix the problem.  I could fix it locally with:

```shell
$ dos2unix clients/client-go/README.md
$ yarn generate
$ git commit --amend -a
```

Those newlines are the cause of the issues in the top-level README.

djmitche:
Oops, still the lint to fix --

```
clients/client-go/codegenerator/model/model.go:76:18: Error return value of `ioutil.WriteFile` is not checked (errcheck)
	ioutil.WriteFile(path, formattedContent, 0644)
```

ajinkabeer:
> Oops, still the lint to fix --

Pushed the latest changes. Fingers crossed for this one :)

ajinkabeer:
> Looks good -- and the lint check passed :)

Thanks for your help :)

djmitche:
What would you like to do next?  There are some more golang bugs at https://codetribute.mozilla.org/languages/go but we also have plenty of JS bugs.

ajinkabeer:
> What would you like to do next? There are some more golang bugs at https://codetribute.mozilla.org/languages/go but we also have plenty of JS bugs.

This is the first time I have ever written a piece of code in Go and now I am getting the hang of it. I will look into those bugs too. 

> NOTE: Taskcluster currently doesn't show up on pkg.go.dev. I don't know why, but if you can figure that out that'd be cool :)

Before that I guess I will look into this if it sounds good to you.

djmitche:
https://bugzilla.mozilla.org/show_bug.cgi?id=1620350 might be the least complex of those go bugs, but I don't think it's especially simple, either.

I totally forgot about the pkg.go.dev question -- that's a good next step!

ajinkabeer:
> https://bugzilla.mozilla.org/show_bug.cgi?id=1620350 might be the least complex of those go bugs, but I don't think it's especially simple, either.

Thanks. I will keep that in mind.

> I totally forgot about the pkg.go.dev question -- that's a good next step!

Cool :)

clients/client-go/codegenerator/model/model.go
```diff
@@ -221,10 +244,24 @@ func (apiDefs APIDefinitions) GenerateCode(goOutputDir string) {
 //
 // This package was generated from the schema defined at
 // ` + apiDefs[i].URL + `
-
```
petemoore, line 224: This was intentional, to keep this code comment separate from the package docs.

PR #2667, opened 2020-04-15
Bug 1629168 - [websocktunnel] fix race condition in stopping copy ops

djmitche:
Bugzilla Bug: [1629168](https://bugzilla.mozilla.org/show_bug.cgi?id=1629168)

### DIFF
--- a/tools/websocktunnel/wsproxy/stopper.go
+++ b/tools/websocktunnel/wsproxy/stopper.go
@@ -1,0 +1,46 @@
+package wsproxy
+
+import "sync"
+
+// stopper manages stopping a fleet of goroutines.  It is basically
+// a protected boolean with functionality to set, check, and wait.
+type stopper struct {
+	cond    sync.Cond
+	stopped bool
+}
+
+func newStopper() *stopper {
+	return &stopper{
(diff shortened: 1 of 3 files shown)


djmitche:
Hm, I can't reproduce that `yarn generate` failure..

djmitche:
let's retry those..

djmitche:
The tools-1.13.7 failure here is from https://bugzilla.mozilla.org/show_bug.cgi?id=1630124 so not a concern here.

tools/websocktunnel/wsproxy/stopper.go
```diff
@@ -0,0 +1,46 @@
+package wsproxy
+
+import "sync"
+
+// stopper manages stopping a fleet of goroutines.  It is basically
+// a protected boolean with functionality to set, check, and wait.
```
imbstack, line 6: Is this a common thing we do in our go code? Worth moving something like this to `internal`? (not blocking landing this on that)

djmitche: I don't know.. this is the first time I've built something specifically like this.  If we see it again, yeah, it would probably make a good utility.

PR #2673, opened 2020-04-15
Bug 1630124 - fix deadlock in livelog

djmitche:
Bugzilla Bug: [1630124](https://bugzilla.mozilla.org/show_bug.cgi?id=XXXXX)

The second commit here fixes a deadlock.  The bug was an oversight on my part in a modification I made while migrating to the monorepo -- so a new, never-released bug.

In trying to reproduce this I filled my `/tmp`, so I added some support for deleting temp files in tests.

I'm not confident this is going to fix bug 1630124, as the deadlock was in code that handles the situation where a consumer has more than 100 log output "events" buffered.  If it doesn't deadlock, then the effect is to cut off the consumer early, which should cause the test to fail.  But, I can't reproduce that behavior locally, and the test is set up such that it produces data more slowly than it consumes it (ChunkReader has a 20ns delay between chunks, while we use `ioutil.ReadAll` to consume).  So I'd like to land this and see if it still fails in CI, and if so continue to debug from there.

### DIFF
--- a/tools/livelog/writer/stream_handle.go
+++ b/tools/livelog/writer/stream_handle.go
@@ -39,22 +39,26 @@ func newStreamHandle(stream *Stream, start, stop int64) StreamHandle {
 }
 
 func (self *StreamHandle) writeEvent(event *Event, w io.Writer) (int64, error) {
-	// Note that while we need to trim buffers here we don't need to exclude any
-	// since the stream type will only dispatch events which apply to handles
-	// with valid offsets...
+	// We may receive events that were generated while we were catching up
+	// in the backing log, so we may need to ignore some or all of the bytes
+	// in this event..
 
 	eventEndOffset := event.Offset + event.Length
+	if self.Offset >= eventEndOffset {
+		return 0, nil
+	}
 
-	startOffset := self.Offset - event.Offset
-	var endOffset int64
+	// calculate the slice of the event's bytes that we need..
+	startInEvent := self.Offset - event.Offset
+	var endInEvent int64
 	if eventEndOffset > self.Stop {
-		endOffset = eventEndOffset - self.Stop
+		endInEvent = eventEndOffset - self.Stop
 	} else {
-		endOffset = event.Length
+		endInEvent = event.Length
 	}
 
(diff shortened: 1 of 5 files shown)


djmitche:
Well, that was quick..
https://community-tc.services.mozilla.com/tasks/DOKj__PaT7SsCe0Q-OYqGA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FDOKj__PaT7SsCe0Q-OYqGA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log
```
2020/04/15 19:54:56 Removing handle with 99 pending writes
```

djmitche:
OK, that was a bunch of fixes.  The commits are worth looking at individually, and the one removing the double buffering is worth looking at with whitespace changes omitted, as most of it is re-indentation.

Ultimately this test depends on timing, so if it keeps failing then maybe we could rewrite it to do some kind of coordination between writing to the PUT port and reading on the GET port.

tools/livelog/writer/stream_handle.go
```diff
@@ -117,42 +112,20 @@ func (self *StreamHandle) WriteTo(target io.Writer) (n int64, err error) {
 	}
 
 	for event := range self.events {
-		pendingBuf := 0
-		self.pendingEvents[0] = event
-
-		// Build a list of all the pointers to the events we need to update. This
-		// has the very important effect of emptying the channel which may be
-		// building up very quickly.
```
imbstack, line 121: I guess this wasn't actually important?

djmitche: As I understood it, jlal saw that the channel's buffer was filling up and causing the producer to stop sending events to this handler.  So the solution was to add an additional buffer *after* the channel, of the same size.  The effect was to write a lot of confusing lines of code to double the channel buffer size.  I reduced this to the more efficient 

```diff
-const EVENT_BUFFER_SIZE = 100
+const EVENT_BUFFER_SIZE = 200
```

PR #2675, opened 2020-04-15
Remove auth azure tests

### LINKED ISSUE #2674: *cleanup* - remove the azure table/container cleanup in the auth test suite

helfi92:
Resolves #2674.


services/auth/test/azure_test.js
```diff
@@ -1,307 +0,0 @@
...
-  test('azureTables', async function() {
-    // First make sure the table exists
-    await helper.apiClient.azureTableSAS(
-      testaccount,
-      'TestTable',
-      'read-write',
-    );
-    let extra = {};
-    do {
-      const result = await helper.apiClient.azureTables(testaccount, extra);
```
djmitche, line 40: This method and the others tested here still exist, so I think we need to keep this file :)

djmitche:
oh, and yarn generate

PR #2680, opened 2020-04-16
Bug 1591476 - always get config from worker-runner via registerWorker

djmitche:
Bugzilla Bug: [1591476](https://bugzilla.mozilla.org/show_bug.cgi?id=1591476)


djmitche:
I added a spot of w-m docs.  The worker-runner docs look OK.

PR #2682, opened 2020-04-16
Make migrationScript and downgradeScript optional

djmitche:
We have had two revisions now with no upgrade/downgrade scripts, so this seems a reasonable tweak.

### DIFF
--- a/libraries/postgres/src/Version.js
+++ b/libraries/postgres/src/Version.js
@@ -4,6 +4,7 @@ const Method = require('./Method');
 const { loadSql } = require('./util');
 
 const objMap = (obj, fn) => Object.fromEntries(Object.entries(obj).map(fn));
+const ALLOWED_KEYS = ['version', 'migrationScript', 'downgradeScript', 'methods'];
 
 class Version {
   /**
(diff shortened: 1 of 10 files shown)


libraries/postgres/src/Version.js
```diff
@@ -4,6 +4,7 @@ const Method = require('./Method');
 const { loadSql } = require('./util');
 
 const objMap = (obj, fn) => Object.fromEntries(Object.entries(obj).map(fn));
+const ALLOWED_KEYS = ['version', 'migrationScript', 'downgradeScript', 'methods'];
```
imbstack, line 7: argh, you always catch my issues before I get to r- you

djmitche:
Oh, good, point - I could check that if migrationScript is set then downgradeScript must be set.

helfi92:
> Oh, good, point - I could check that if migrationScript is set then downgradeScript must be set.

That's a really good idea!

libraries/postgres/src/Version.js
djmitche: Python's `set` takes an iterator as an arg to its constructor, so I always forget that the JS `Set` doesn't..

PR #2686, opened 2020-04-17
Bug 1630958 - include json-stable-stringify as a production dependency

djmitche:
Bugzilla Bug: [1630958](https://bugzilla.mozilla.org/show_bug.cgi?id=1630958)


djmitche:
See the bug for comments on why we need both json-stable-stringify and fast-json-stable-stringify.

PR #2689, opened 2020-04-17
Bug 1630023 - minor fixes

djmitche:
Bugzilla Bug: [1630023](https://bugzilla.mozilla.org/show_bug.cgi?id=1630023)

A few minor fixes I made while working on this bug, that can be merged sooner than the full patch will be ready..

### DIFF
--- a/libraries/postgres/src/Schema.js
+++ b/libraries/postgres/src/Schema.js
@@ -18,7 +18,7 @@ class Schema {
    */
   static fromDbDirectory(directory) {
     const dentries = fs.readdirSync(path.join(directory, 'versions'));
-    let versions = new Array(dentries.length);
+    let versions = [];
 
     dentries.forEach(dentry => {
       if (dentry.startsWith('.')) {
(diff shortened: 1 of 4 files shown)


libraries/postgres/src/util.js
```diff
@@ -9,7 +9,7 @@ exports.dollarQuote = str => {
     if (str.indexOf(quote) === -1) {
       return quote + str + quote;
     }
-    i = i + '-';
+    i = i + 'x';
```
helfi92, line 12: What's happening here? :)

libraries/postgres/src/Schema.js
```diff
@@ -18,7 +18,7 @@ class Schema {
    */
   static fromDbDirectory(directory) {
     const dentries = fs.readdirSync(path.join(directory, 'versions'));
-    let versions = new Array(dentries.length);
+    let versions = [];
```
helfi92, line 21: What bug is this fixing?

libraries/postgres/src/util.js
djmitche: It turns out Postgres wants `$<tag>$`, and apparently `-` is not a valid character in the tag.  So I used `x` instead.  I noticed this when quoting a string that contained `$$`.

libraries/postgres/src/Schema.js
djmitche: I put some `.sql` files in `db/versions`, and this function got all confused since it counts the entries in the directory and assumes every one is a DB version.  The new version just extends the array as necessary for each version file it finds.

PR #2691, opened 2020-04-18
Update dependency checking to account for production not having devDependencies

imbstack:
This is faster, more correct, and (imho) simpler than using the depcheck library in a hacky way. Under the hood they were doing basically this anyway. I think we were doing something wrong with the old way as well; there were lots of unused dependencies that this found.

There's a bit of an open question in there with how I'm calculating prod vs dev files. If we can improve that I'd be happy but also happy to leave it how it is or give up on trying to do that bit.


infrastructure/tooling/src/meta/checks/depcheck.js
```diff
@@ -40,38 +39,98 @@ if (isMainThread) {
...
+    // are available in dev so we concat
+    const rootPkg = require(`${REPO_ROOT}/package.json`);
+    const deps = Object.keys(rootPkg.dependencies);
+    const devDeps = Object.keys(rootPkg.devDependencies).concat(deps);
+    const specials = rootPkg.metatests.specialImports;
+
+    status("listing files");
+    // HELLO REVIEWERS: Ideas for a nicer way we can go about doing this?
+    let prodFiles = await gitLsFiles({patterns: ['services/*/src/**.js', 'libraries/*/src/**.js', 'db/src/**.js', 'services/prelude.js']});
+    prodFiles = prodFiles.filter(f => !f.startsWith('libraries/testing') && f !== 'services/queue/src/load-test.js');
```
djmitche, line 98: We could delete that load-test.js file, really

package.json
```diff
@@ -25,23 +25,19 @@
...
     "app-root-dir": "^1.0.2",
-    "auth0": "^2.14.0",
     "aws-sdk": "^2.394.0",
     "body-parser": "1.19.0",
     "body-parser-graphql": "^1.0.0",
     "chalk": "^2.4.2",
     "change-case": "^4.1.1",
-    "compression": "^1.7.2",
-    "content-security-policy": "^0.3.0",
+    "compression": "^1.7.2", "content-security-policy": "^0.3.0",
```
djmitche, line 33: missing newline?

PR #2706, opened 2020-04-21
[Bug 1631638] Make the runaway provisioning error a bit less flaky

imbstack:
This is a quite hacky fix to an already hacky alert. If you don't like this approach
feel free to r- it and my feelings will not be hurt. I see this whole part of w-m
getting a pretty good pass shortly with postgres support and cost-savings work
so I mostly want this to stop scaring me every couple weeks for now.

This also passes through tags into sentry reports that we were dropping before

Bugzilla Bug: [1631638](https://bugzilla.mozilla.org/show_bug.cgi?id=1631638)

### DIFF
--- a/libraries/monitor/src/plugins/sentry.js
+++ b/libraries/monitor/src/plugins/sentry.js
@@ -1,5 +1,16 @@
 const Sentry = require('@sentry/node');
 
+const tcToSentryLevel = {
+  emerg: 'fatal',
+  alert: 'error',
+  crit: 'error',
+  err: 'error',
+  warning: 'warning',
+  notice: 'info',
+  info: 'info',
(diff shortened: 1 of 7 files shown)


libraries/monitor/src/plugins/sentry.js
```diff
@@ -15,7 +26,17 @@ class SentryReporter {
...
+    Sentry.configureScope(scope => {
+      if (level) {
+        scope.setLevel(tcToSentryLevel[level] || 'error');
+      }
+      if (extra) {
+        Object.entries(extra).forEach(([k, v]) => {
+          scope.setTag(k, v);
+        });
+      }
+    });
```
djmitche, line 39: NICE!

PR #2715, opened 2020-04-23
Use mozilla's go hawk implementation

petemoore:
I spotted that mozilla maintains its own fork of the tent.io hawk go implementation, so I figured we should probably use it to get the fixes that aren't present upstream.

Were the fixes in the mozilla fork rejected upstream, or just ignored?

CC @jvehent

### DIFF
--- a/changelog/Sd6bb2nOSAKVu2thsZQs2g.md
+++ b/changelog/Sd6bb2nOSAKVu2thsZQs2g.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+---
+The go tools and libraries now use mozilla's fork of the (archived) CoreOS fork
+of the tent.io implementation, instead of the tent.io implementation.
--- a/clients/client-go/http.go
+++ b/clients/client-go/http.go
@@ -19,7 +19,7 @@ import (
 	"github.com/cenkalti/backoff/v3"
 	"github.com/taskcluster/httpbackoff/v3"
 	tcurls "github.com/taskcluster/taskcluster-lib-urls"
-	hawk "github.com/tent/hawk-go"
+	"go.mozilla.org/hawk"
 )
 
 var debug = false
--- a/clients/client-shell/client/credentials.go
+++ b/clients/client-shell/client/credentials.go
@@ -12,7 +12,7 @@ import (
 	"net/url"
 	"strings"
 
-	"github.com/tent/hawk-go"
+	"go.mozilla.org/hawk"
 
 	got "github.com/taskcluster/go-got"
 	tcclient "github.com/taskcluster/taskcluster/v29/clients/client-go"
--- a/go.mod
+++ b/go.mod
@@ -42,9 +42,9 @@ require (
 	github.com/taskcluster/slugid-go v1.1.0
 	github.com/taskcluster/stateless-dns-go v1.0.6
 	github.com/taskcluster/taskcluster-lib-urls v13.0.0+incompatible
-	github.com/tent/hawk-go v0.0.0-20161026210932-d341ea318957
 	github.com/ulikunitz/xz v0.5.7 // indirect
 	github.com/xeipuuv/gojsonschema v1.2.0
+	go.mozilla.org/hawk v0.0.0-20200327135946-18943cc03d32
 	golang.org/x/crypto v0.0.0-20200420201142-3c4aac89819a
 	golang.org/x/net v0.0.0-20200421231249-e086a090c8fd
 	golang.org/x/sys v0.0.0-20200420163511-1957bb5e6d1f
@@ -52,5 +52,4 @@ require (
 	gopkg.in/tylerb/graceful.v1 v1.2.15
 	gopkg.in/yaml.v2 v2.2.8
 	gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c
-	launchpad.net/gocheck v0.0.0-20140225173054-000000000087 // indirect
 )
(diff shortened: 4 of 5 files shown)


petemoore:
> Any idea what the mozilla fork adds?

I suspect the most important change is:

* https://github.com/mozilla-services/hawk-go/commit/efc4d5fa14cf020d4aa0797e3cddc72f82b8b895

This is the CoreOS change:

* https://github.com/mozilla-services/hawk-go/commit/e2691d10fc256cc09e895f3417a51db81e638012

And other than test changes, this seems to be the other change:
* https://github.com/mozilla-services/hawk-go/pull/1

It looks like the CoreOS team tried to get this landed upstream in https://github.com/tent/hawk-go/pull/7 but there has been no response from tent.io.

Now I look, I don't see any open or closed PRs for the mozilla changes, I'm not sure why PR's weren't created, although that might be because we also wanted https://github.com/tent/hawk-go/pull/7 which seems to have not been answered.

petemoore:
Strange, we have hawk errors like this one:

```
2020/04/23 10:44:50Z WORKER EXCEPTION due to response code 401 from Queue when uploading artifact &main.RedirectArtifact{BaseArtifact:(*main.BaseArtifact)(0xc0002609f0), URL:"http://12.34.56.78:33261/log/QSQxMFekR9mta9vo1eXJFw", ContentType:"text/plain; charset=utf-8"} with CreateArtifact payload {"contentType":"text/plain; charset=utf-8","expires":"2020-04-23T11:00:19.936Z","storageType":"reference","url":"http://12.34.56.78:33261/log/QSQxMFekR9mta9vo1eXJFw"} - HTTP response body: {
  "code": "AuthenticationFailed",
  "message": "Unauthorized: Bad mac\n\n---\n\n* method:     createArtifact\n* errorCode:  AuthenticationFailed\n* statusCode: 401\n* time:       2020-04-23T10:44:49.970Z",
  "requestInfo": {
    "method": "createArtifact",
    "params": {
      "0": "public/logs/live.log",
      "taskId": "OOYxWyvbQzWWSe3uL3y2tw",
      "runId": "0",
      "name": "public/logs/live.log"
    },
    "payload": {
      "contentType": "text/plain; charset=utf-8",
      "expires": "2020-04-23T11:00:19.936Z",
      "storageType": "reference",
      "url": "http://12.34.56.78:33261/log/QSQxMFekR9mta9vo1eXJFw"
    },
    "time": "2020-04-23T10:44:49.970Z"
  }
}
```

petemoore:
Weird, we're still getting `Unauthorized: Bad mac` - so something [here](https://github.com/tent/hawk-go/compare/master...mozilla-services:master) must be affecting the client side mac generation. I'll investigate.

petemoore:
So I found the issue. It is actually the CoreOS The RequestURI is different. Example of before vs after:

```
2020/04/23 14:48:58Z Old RequestURI: /api/queue/v1/task/EtL3ywFLTRWTn1Wng5NaWQ/runs/0/artifacts/public%2Flogs%2Flive_backing.log
2020/04/23 14:48:58Z New RequestURI: /api/queue/v1/task/EtL3ywFLTRWTn1Wng5NaWQ/runs/0/artifacts/public/logs/live_backing.log
```

This change is due to [this line](https://github.com/coreos/hawk-go/commit/e2691d10fc256cc09e895f3417a51db81e638012#diff-bd1123e4aca92ed920c6fca9853a1c76R285) in the single commit added by the Core OS fork, that the mozilla-service's fork is based on. In other words, I suspect the Core OS patch may be invalid, since the artifact name in the RequestURI should be urlencoded, and it looks like the single Core OS patch URL decodes the request URI. However, I haven't read the spec or the original implementation to know which is correct. Certainly our server side implementation expects the requestURI to be in its raw form.

petemoore:
I'm not a node.js expert, but looking [here](https://github.com/hapijs/hawk/blob/25ab0702a534ab5ed1c12735b1b9275e17a80f63/lib/crypto.js#L68) it looks like if the request path starts with a leading `/`, it is not altered in any way, which is consistent with the server behaviour we see of not url-decoding `/api/queue/v1/task/EtL3ywFLTRWTn1Wng5NaWQ/runs/0/artifacts/public%2Flogs%2Flive_backing.log`.

Therefore I'm inclined to believe that the fix in https://github.com/coreos/hawk-go/commit/e2691d10fc256cc09e895f3417a51db81e638012#diff-bd1123e4aca92ed920c6fca9853a1c76R285 to align the server and client behaviour in the go client incorrectly adjusted the client behaviour, when it should have adjusted the server behaviour, to be consistent with the node.js implementation.

@djmitche As someone more intimately familiar with hawk, and a solid understanding of node.js, do you agree with the above paragraph? If so, I can make an upstream PR to CoreOS to revert the handling of request URI in the client, and adapt the request URI handling in the server to match. Thanks!

djmitche:
> I'm not a node.js expert, but looking here it looks like if the request path starts with a leading /, it is not altered in any way, which is consistent with the server behaviour we see of not url-decoding /api/queue/v1/task/EtL3ywFLTRWTn1Wng5NaWQ/runs/0/artifacts/public%2Flogs%2Flive_backing.log.

Close -- it actually doesn't matter whether the request path starts with `/` -- the conditional is there to accept a URL in `resource` and extract just the path portion of that value. So by line 64 `resource` is just a path and thus always begins with `/`.  And you're correct, no encoding takes place.

The link to the coreos change looks like it's not *intended* to change behavior around encoding, and perhaps that was an undetected side-effect.  My read is that `url.RequestURI()` uses `url.Opaque` when that value is set, and the commit is trying to replicate `url.ReuqestURI()`'s behavior otherwise, but not look at `url.Opaque`.  But on line 270 it uses `URL.Path` instead of `URL.RawPath`.  I assume that's just an oversight.  A PR would probably be helpful (but note that the CoreOS library is archived and no longer maintained).  It would be good to add some tests, too, to demonstrate the behavior!  The Rust implementation has a [few](https://github.com/taskcluster/rust-hawk/blob/master/src/request.rs#L533-L547) [examples](https://github.com/taskcluster/rust-hawk/blob/master/src/request.rs#L657-L678).

djmitche:
It would definitely be worth leaving a note about this issue inhttps://github.com/tent/hawk-go/pull/7!

petemoore:
This is blocked on https://github.com/mozilla-services/hawk-go/issues/12.

imbstack:
I'm going to convert this to a draft for now since we're blocked!

djmitche:
Note: we just adopted https://github.com/mozilla/hawk, so it makes sense to finish this up so that Mozilla hosts a working Hawk implementation in all Go too.

petemoore:

PR #2716, opened 2020-04-23
Bug 1630019 - Port purge_cache to use normalized DB tables

helfi92:


djmitche:
I'll make a PR for those first two.  I added the [NEEDED??] note because that function is not used and I can't remember why I used it.  I'll probably drop it unless I remember :)

helfi92:
Sounds good!

db/tables.yml
```diff
@@ -87,11 +87,12 @@ denylisted_notification_entities:
...
   etag: uuid
-cache_purges_entities:
-  partition_key: text not null
-  row_key: text not null
-  value: jsonb not null
-  version: integer not null
+cache_purges:
+  provisioner_id: text not null
+  worker_type: text not null
+  cache_name: text not null
```
imbstack, line 93: TIL `text` is not worse performance than `varchar` in postgres. So that's not a reason to use `varchar` here, but I want to make sure that we're actively deciding to limit `worker_type` length at the api schema level rather than the db schema level here? That seems like a reasonable trade-off to make -- is that written down in our best-practices, etc? Just want to know for when I start doing work like this.

db/versions/0009-downgrade.sql
```diff
@@ -0,0 +1,32 @@
...
+      'PartitionKey', encode_string_key(provisioner_id),
+      'RowKey', encode_string_key(worker_type),
+      'provisionerId', provisioner_id,
+      'workerType', worker_type,
+      'cacheName', cache_name,
+      'before', before,
+      'expires', expires) as value,
+    1 as version,
+    etag
+  from cache_purges;
```
imbstack: this is super cool to see in practice

db/versions/0009-migration.sql
```diff
@@ -0,0 +1,29 @@
...
+      (value ->> 'expires')::timestamptz as expires,
+      etag
+    from cache_purges_entities;
+  alter table cache_purges add primary key (provisioner_id, worker_type, cache_name);
+  alter table cache_purges
+    alter column provisioner_id set not null,
+    alter column worker_type set not null,
+    alter column cache_name set not null,
+    alter column before set not null,
+    alter column expires set not null;
```
imbstack, line 24: so if we do have null data in there now, this will fail and then we can roll-back from there? I ask because it seems like we're seeing some schema validation errors on output these days which hints at maybe we have missing data somewhere. I'm not worried and 100% confident that this will be fine, just want to understand the process we would go through.

helfi92: > so if we do have null data in there now, this will fail and then we can roll-back from there? 

Yes. It will fail without consequences other than the time spent migrating.

> I ask because it seems like we're seeing some schema validation errors on output these days which hints at maybe we have missing data somewhere

Interesting! Is this for a known or unknown subset of tables?

helfi92:
> Do I understand correctly that this is migrating data to the new table with an actual schema while maintaining the lib-entities methods so that our usage in the js stays the same? Part 2 of this will be to add new queries that make use of the format of the new table?

Correct!

db/tables.yml
helfi92: I think things like the length of `worker_type` could possibly change in the future. Any changes to the db in the future (i.e, altering the column to accept something different) will hang the db until all of the changes have been done. `alter` afaict is not a very fast operation so for large tables this would possibly need a downtime. Agreed, we should include that in the best practices if the team agrees with this.

db/versions/0009-migration.sql
imbstack: https://sentry.prod.mozaws.net/operations/taskcluster-firefox-ci/issues/7969678/?query=is:unresolved%20validation in particular is what I was thinking of. no real ideas if it is caused by db issues or something else.

I think we shouldn't worry about it too much especially since this will fail gracefully.

db/tables.yml
imbstack: Ok, seems like you+dustin are in agreement about what to do here and so I'm convinced too! Adding it to best-practices at some point makes sense but obviously not blocking landing this.

db/test/versions/0009_test.js
```diff
@@ -0,0 +1,742 @@
...
+
+  test('cache_purges table created', async function() {
+    await helper.assertNoTable('cache_purges');
+    await helper.upgradeTo(THIS_VERSION);
+    await helper.assertTable('cache_purges');
+    await helper.downgradeTo(PREV_VERSION);
+    await helper.assertNoTable('cache_purges');
+  });
+
+  test('cache purges data migrated on upgrade and downgrade', async function() {
```
djmitche, line 101: I think this can all be replaced with a call to the util function in https://github.com/djmitche/taskcluster/commit/70949f259ed237c07d9d0d5afafa31b2c1cb32bd :)

db/versions/0009.yml
```diff
@@ -212,3 +212,92 @@ methods:
...
+        from cache_purges
+        order by cache_purges.provisioner_id, cache_purges.worker_type, cache_purges.cache_name
+        limit case
+          when (page_size_in is not null and page_size_in > 0) then page_size_in
+          else null
+        end
+        offset case
+          when (page_offset_in is not null and page_offset_in > 0) then page_offset_in
+          else 0
+        end;
```
djmitche, line 256: Since you're defining an earlier version than my #2748 is, do you want to factor these out into a utility function (defined in the migration script) like you suggested today?  Then I'll modify #2748 to use those utility functions.

services/purge-cache/src/api.js
```diff
@@ -57,31 +58,7 @@ builder.declare({
...
-      cacheName,
-    });
-
-    await cb.modify(cachePurge => {
-      cachePurge.before = new Date();
-      cachePurge.expires = taskcluster.fromNow('1 day');
-    });
-  }
-
+  await this.db.fns.purge_cache(provisionerId, workerType, cacheName, new Date(), taskcluster.fromNow('1 day'));
```
djmitche, line 61: So much simpler!

db/versions/0009.yml
```diff
@@ -212,3 +212,92 @@ methods:
...
+          new_etag
+        ) on conflict (provisioner_id, worker_type, cache_name) do
+        update
+        set (before, expires, etag) = (purge_cache.bef, exp, new_etag)
+        where cache_purges.provisioner_id = purge_cache.prov_id and cache_purges.worker_type = purge_cache.wt and cache_purges.cache_name = purge_cache.c_name;
+      end
+  cache_purges_expires:
+    description: |-
+      Expire cachePurges that are past their expiration.
+      Returns a promise that all expired cachePurges have been deleted.
```
djmitche, line 288: This isn't JS, it doesn't return a promise ;)

Generally, the descriptions here need a little bit of tweaking to describe the DB functions, rather than the API functions that call them.

db/test/fns/purge_cache_test.js
```diff
@@ -0,0 +1,174 @@
...
+    compare(caches[0], samples[0]);
+
+    offset += 1;
+    caches = await db.fns.all_purge_requests(size, offset);
+    assert.equal(caches.length, 1);
+    compare(caches[0], samples[1]);
+  });
+
+  helper.dbTest('purge_cache (create)', async function(db, isFake) {
+    const dbFns = isFake ? helper.fakeDb['purge_cache'] : db.fns;
```
djmitche, line 69: I don't think this is necessary -- `db.fns` should work for both fake and real calls.

db/test/versions/0009_test.js
```diff
@@ -23,720 +22,72 @@ const CachePurgeEntity = Entity.configure({
...
+    dbVersion: THIS_VERSION,
+    serviceName: 'purge_cache',
+    entityTableName: 'cache_purges_entities',
+    newTableName: 'cache_purges',
+    EntityClass: CachePurgeEntity,
+    samples: {
+      pptt: {
+        provisionerId: 'foo',
+        workerType: 'bar',
+        cacheName: 'gamma',
```
djmitche, line 52: A sample containing some encodable characters would be good

djmitche:
Ah, looks like some CI issues, too.

db/versions/0009-migration.sql
```diff
@@ -0,0 +1,57 @@
...
+  drop table cache_purges_entities;
+  grant select, insert, update, delete on cache_purges to $db_user_prefix$_purge_cache;
+
+  -- Given a page size it returns the limit to use on a paginated db function.
+  create or replace function get_page_limit(page_size integer) RETURNS integer
+  as $$
+      begin
+        return case
+          when (page_size is not null and page_size > 0) then page_size
+          else null
```
imbstack, line 38: Will this mean we default to returning entire tables if `page_size` is not specified?

db/versions/0009.yml
```diff
@@ -0,0 +1,288 @@
...
+          cache_name_in,
+          before_in,
+          expires_in,
+          new_etag
+        ) on conflict (provisioner_id, worker_type, cache_name) do
+        update
+        set (before, expires, etag) = (before_in, expires_in, new_etag)
+        where cache_purges.provisioner_id = provisioner_id_in and cache_purges.worker_type = worker_type_in and cache_purges.cache_name = cache_name_in;
+      end
+  cache_purges_expires:
```
imbstack, line 270: Should this be something more like `expire_purge_caches`?

db/versions/0009.yml
```diff
@@ -0,0 +1,288 @@
...
+            when (size is not null and size > 0) then size + 1
+            else null
+          end
+          offset case
+            when (size is not null and size > 0 and page is not null and page > 0) then page
+            else 0
+          end;
+        end if;
+      end
+  purge_requests:
```
imbstack, line 215: How are we deciding which queries to paginate and which not to? I admit that >1000 cache purges for a workertype seems unlikely but this is a more general question

libraries/postgres/src/util.js
```diff
@@ -2,6 +2,16 @@ const fs = require('fs');
 const path = require('path');
 const {SYNTAX_ERROR} = require('./constants');
 
+exports.ignorePgErrors = async (promise, ...codes) => {
```
imbstack, line 5: this function returning the result on success and nothing on ignored failures is a bit weird to me. Could it make sense to not return anything in either case since nobody should use the results from this anyway without a check of what the results were?

db/versions/0009-migration.sql
djmitche: Yes, that's the intent, as it means that internally we can query the whole table.

db/fns.md
```diff
@@ -86,11 +86,15 @@
 
 | Name | Mode | Arguments | Returns | Description |
 | --- | --- | --- | --- | --- |
+| all_purge_requests | read | page_size_in integer, page_offset_in integer | table (provisioner_id text, worker_type text, cache_name text, before timestamptz) | View all active purge requests. |
```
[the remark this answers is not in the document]
djmitche: #2572

db/versions/0009.yml
djmitche: I think the easy answer is "where the REST API paginates".  There may be circumstances where that's different, but as an 80% answer it's probably good.  In this case, [`purgeCache.purgeRequests`](https://docs.taskcluster.net/docs/reference/core/purge-cache/api#purgeRequests) is not paginated

libraries/postgres/src/util.js
djmitche: Good call.  #2753.

db/versions/0009.yml
djmitche: #2754 (I checked with @helfi92..)

PR #2719, opened 2020-04-24
Tc-proxy comes to monorepo

owlishDeveloper:


owlishDeveloper:
@djmitche pls take a look now

owlishDeveloper:
I don't understand what's the deal with changelog 🤔

I wish TC pushed a commit into the branch after it ran `go mod tidy, go fmt etc` 🤕

go.mod
```diff
@@ -42,8 +42,10 @@ require (
 	github.com/taskcluster/slugid-go v1.1.0
 	github.com/taskcluster/stateless-dns-go v1.0.6
 	github.com/taskcluster/taskcluster-lib-urls v13.0.0+incompatible
+	github.com/taskcluster/taskcluster-proxy v5.1.0+incompatible
```
petemoore, line 45: It looks like something may still be referencing the original taskcluster-proxy repo.

tools/taskcluster-proxy/README.md
```diff
@@ -0,0 +1,247 @@
...
+* When used by generic-worker, the taskcluster-proxy runs as a separate
+  native executable on the host.
+
+## Download binary release
+
+You do __not__ need to install go (golang) to run the proxy. It is shipped as a
+native executable.
+
+See [releases page](https://github.com/taskcluster/taskcluster-proxy/releases)
+and choose a download that matches your platform.
```
petemoore, line 18: I think this can point to the monorepo releases now.

tools/taskcluster-proxy/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Download source and install via `go get`
+
+Alternatively you can build and install from source. For this it is recommended
+you install the latest version of go (golang) first.
+
+```sh
+go get github.com/taskcluster/taskcluster-proxy
+```
```
petemoore, line 27: I think this should be `github.com/taskcluster/taskcluster/v<latest>/tools/taskcluster-proxy` where `<latest>` is the latest release. We should check this `README.md` will get updated with a find/replace regexp like other files in the repo when a new release is made, so this reference gets updated automatically.

tools/taskcluster-proxy/build.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash -e
+
+cd "$(dirname "${0}")"
+
+# Support go 1 release 1.9 or higher
+GO_MAJOR_VERSION=1
+MIN_GO_MINOR_VERSION=9
```
petemoore, line 7: I think the comment (line 5) and `MIN_GO_MINOR_VERSION` (line 7) should probably be updated to require go1.13.

tools/taskcluster-proxy/README.md
owlishDeveloper: > We should check this README.md will get updated with a find/replace regexp like other files in the repo when a new release is made, so this reference gets updated automatically.

Where do I check that?

djmitche: The regexp is run over all of `tools/`, so it should be fine.

changelog/bug-1632708.md
```diff
@@ -0,0 +1,3 @@
+audience: general
+level: silent
+---
```
djmitche, line 3: It looks like the issue here was that it wanted a newline after the `---`
```
diff --git a/changelog/bug-1632708.md b/changelog/bug-1632708.md
index 1252e8ece..8006e8b3f 100644
--- a/changelog/bug-1632708.md
+++ b/changelog/bug-1632708.md
@@ -1,3 +1,3 @@
 audience: general
 level: silent
----
\ No newline at end of file
+---
```
fixed it.  I'll make a PR to be more lenient about that!

djmitche:
I made a staging release [here](https://github.com/taskcluster/taskcluster/runs/624235722).

djmitche:
It looks like everything's here!  If the staging push succeeds, I think this is good to go.  I filed #2734 as a followup, but that's more of a generic-worker change than a tc-proxy change.

@petemoore do we still need all of
```
echo "Building tc-proxy:"
build windows 386 
build windows amd64
build darwin 386 
build darwin amd64
build linux 386 
build linux amd64
build linux arm 
build linux arm64
```
in `build.sh`?

djmitche:
Aww, staging release failed :(

petemoore:
> It looks like everything's here! If the staging push succeeds, I think this is good to go. I filed #2734 as a followup, but that's more of a generic-worker change than a tc-proxy change.
> 
> @petemoore do we still need all of
> 
> ```
> echo "Building tc-proxy:"
> build windows 386 
> build windows amd64
> build darwin 386 
> build darwin amd64
> build linux 386 
> build linux amd64
> build linux arm 
> build linux arm64
> ```
> 
> in `build.sh`?

The platforms covered by generic-worker/docker-worker are:

* windows/386
* windows/amd64
* darwin/amd64
* linux/amd64
* linux/arm
* linux/arm64

So we could probably drop the following platforms for taskcluster-proxy:

* darwin/386
* linux/386

since we don't have any official releases for workers on those platforms. I suspect script-worker doesn't use taskcluster-proxy.

petemoore:
I suspect the failure for 39af04af54978408a2fdc338b1c71d49aeeefb3c was due to the absence of 70789fe1a588af7ae3247027ff405b08d77fd6d5, so I've merged from master to pick up #2709 (where that fix was applied).

PR #2733, opened 2020-04-27
Bug 1633582 - Google provider labeling VM disks

milescrabill:
Bugzilla Bug: [1633582](https://bugzilla.mozilla.org/show_bug.cgi?id=1633582)

### DIFF
--- a/services/worker-manager/src/providers/google.js
+++ b/services/worker-manager/src/providers/google.js
@@ -245,7 +245,7 @@ class GoogleProvider extends Provider {
             name: instanceName,
             labels: {
               ...cfg.labels || {},
-              labels,
+              ...labels,
             },
             description: cfg.description || workerPool.description,
             disks: {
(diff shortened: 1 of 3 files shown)


services/worker-manager/src/providers/google.js
```diff
@@ -240,12 +245,13 @@ class GoogleProvider extends Provider {
             name: instanceName,
             labels: {
               ...cfg.labels || {},
-              'created-by': `taskcluster-wm-${this.providerId}`.replace(/[^a-zA-Z0-9-]/g, '-'),
-              'managed-by': 'taskcluster',
-              'worker-pool-id': workerPoolId.replace(/[^a-zA-Z0-9-]/g, '-').toLowerCase(),
-              'owner': workerPool.owner.replace(/[^a-zA-Z0-9-]/g, '-').toLowerCase(),
+              labels,
```
djmitche, line 248: itym `...labels`?

services/worker-manager/src/providers/google.js
```diff
@@ -240,12 +245,13 @@ class GoogleProvider extends Provider {
...
             },
             description: cfg.description || workerPool.description,
+            disks: {
+              ...cfg.disks || {},
+              labels,
```
djmitche, line 253: (but this looks correct)

milescrabill:
What would be a good way to test this?

djmitche:
Ah, I thought there was already some testing for labels where these new labels could just be added to the list.  I don't see that for any of the providers, in fact..

djmitche:
So I think this is fine as-is.

PR #2736, opened 2020-04-28
Issue 2734 - use monorepo livelog/taskcluster-proxy in generic-worker CI

### LINKED ISSUE #2734: Use latest tc-proxy in generic-worker
Generic-worker uses a fixed version of tc-proxy in its CI (`workers/generic-worker/gw-decision-task/tasks.yml`).

It also a few links to the old repo in the docs -- let's fix that up too.

petemoore:
Github Bug/Issue: Fixes #2734

### DIFF
--- a/workers/generic-worker/livelog/livelog.go
+++ b/workers/generic-worker/livelog/livelog.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"log"
 	"net"
 	"net/http"
 	"os"
@@ -33,6 +34,7 @@ type LiveLog struct {
 	LogWriter io.WriteCloser
 	mutex     sync.Mutex
 	command   *exec.Cmd
+	done      chan (struct{})
 }
 
 // New starts a livelog OS process using the executable specified, and returns
@@ -46,6 +48,7 @@ func New(liveLogExecutable string, putPort, getPort uint16) (*LiveLog, error) {
 		command: exec.Command(liveLogExecutable),
 		PUTPort: putPort,
 		GETPort: getPort,
+		done:    make(chan (struct{})),
 	}
 	l.setRequestURLs()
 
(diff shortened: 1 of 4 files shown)


petemoore:
Previously, we got the failure logs only if `livelog` exited _before_ we were able to connect to the TCP port of the HTTP PUT interface.

I've updated `workers/generic-worker/livelog/livelog.go` so that we now get failure logs from `livelog` if the `livelog` process exits at _any time_ before we try to kill it.

workers/generic-worker/livelog/livelog.go
```diff
@@ -92,6 +95,13 @@ func New(liveLogExecutable string, putPort, getPort uint16) (*LiveLog, error) {
 
 	select {
 	case err := <-inputStreamConnectionResult:
+		go func() {
+			select {
+			case <-l.done:
+			case pr := <-putResult:
+				log.Printf("WARNING: Livelog failure: error '%v' and livelog output:\n%s", pr.e, pr.b)
+			}
+		}()
```
djmitche, line 104: Launching this error-reporting in a goroutine seems weird.  Is the idea to return immediately, but as soon as the livelog process exits, display its output, even if that's later?

petemoore: The constructor needs to return immediately, so the go routine is launched to await either the closure of the `done` channel (which signifies that the logging has completed and `livelog` is about to be terminated) or for the `livelog` process to complete. If the `livelog` process completes before the `done` channel is closed, it means the livelog exited before we intentionally killed it. The output and the error returned by the `livelog` command are written to the `putResult` channel (line 83), so this new go routine's purpose is to capture the output and error of `livelog` if it completes before the `done` channel is closed, and report them to the standard error of the worker process.

This is how I discovered that the reason `livelog` was terminating early was due to the hardcoded port you added, which is fixed above in `workers/generic-worker/livelog.go`.

PR #2738, opened 2020-04-28
Bug 1629807 - Add state token in login

helfi92:
This will need to best tested on an instance before merging.


djmitche:
I'll fire this up in dustin-dev

djmitche:
wow, `yarn build` found the image you built 3 hours ago!

djmitche:
Oh, I don't have Auth0 set up.  Let's ship this and see what happens in staging.  It's easy to revert!

PR #2748, opened 2020-04-30
Bug 1630023 - migrate wmworker_pools_entities table

djmitche:
Bugzilla Bug: [1630023](https://bugzilla.mozilla.org/show_bug.cgi?id=1630023)


djmitche:
This is currently rebased on #2716, and has NOTES.txt as a backup of the guidelines we're working on in https://pad.mozilla.org/p/db-migration, but is otherwise -- I think -- ready to go.

db/versions/0010.yml
```diff
@@ -0,0 +1,379 @@
...
+      begin
+        return query
+        select
+          wmworker_pools_entities_load.partition_key,
+          'workerPool' as row_key,
+          entity_buf_encode(
+            entity_buf_encode(
+              entity_buf_encode(
+                jsonb_build_object(
+                  'PartitionKey', encode_string_key(worker_pool_id),
```
helfi92, line 21: Should this be using `wmworker_pools_entities_load.partition_key,` instead? I guess they both refer to the same thing.

db/versions/0010.yml
```diff
@@ -0,0 +1,379 @@
...
+                'providerData', provider_data::text),
+              'previousProviderIds', previous_provider_ids::text) as value,
+            1 as version,
+          worker_pools.etag as etag
+          from worker_pools
+          where
+            partition_key is null or
+            worker_pool_id = decode_string_key(partition_key)
+          order by worker_pool_id
+          limit size + 1
```
helfi92, line 188: What would happen if size is not defined?

db/versions/0010.yml
```diff
@@ -0,0 +1,379 @@
...
+        from worker_pools
+        order by worker_pool_id
+        limit case
+            when (page_size_in is not null and page_size_in > 0) then page_size_in
+            else NULL
+          end
+        offset case
+            when (page_offset_in is not null and page_offset_in > 0) then page_offset_in
+            else 0
+          end;
```
helfi92, line 259: We can make use of `get_page_limit` and `get_page_offset`.

db/versions/0010.yml
```diff
@@ -0,0 +1,379 @@
...
+        from worker_pools
+        where worker_pools.worker_pool_id = worker_pool_id_in
+        -- lock this row for the duration of this transaction..
+        for update
+        into existing;
+
+        -- update previous_provider_ids, if the provider_id has changed
+        if existing.provider_id <> provider_id_in then
+          -- remove both provider IDs to avoid duplicates, then re-add existing.provider_id
+          existing.previous_provider_ids = (existing.previous_provider_ids - provider_id_in - existing.provider_id) || jsonb_build_array(existing.provider_id);
```
helfi92, line 289: What does the minus symbol mean in `existing.previous_provider_ids - provider_id_in - existing.provider_id`?

db/versions/0010.yml
```diff
@@ -0,0 +1,379 @@
...
+          end;
+      end
+  update_worker_pool:
+    description: |-
+      Update API-accessible columns on an existig worker pool.  All fields are
+      overridden, but if the provider_id changes, then the existing provider_id
+      is added to previous_provider_ids.  The return value contains values
+      required for an API response and previous_provider_id (singular) containing
+      the provider_id found before the update.  If no such worker pool exists,
+      the return value is an empty set.
```
helfi92, line 268: With azure, changes to the primary key are not permitted. Are we changing the way the API handles things?

db/versions/0010.yml
```diff
@@ -0,0 +1,379 @@
...
+        existing record;
+      begin
+        select
+          worker_pools.provider_id,
+          worker_pools.previous_provider_ids
+        from worker_pools
+        where worker_pools.worker_pool_id = worker_pool_id_in
+        -- lock this row for the duration of this transaction..
+        for update
+        into existing;
```
helfi92, line 284: Is this basically the same thing as using update on conflict?

db/test/versions/0010_test.js
```diff
@@ -0,0 +1,114 @@
...
+    previousProviderIds: Entity.types.JSON,
+    description: Entity.types.String,
+    created: Entity.types.Date,
+    lastModified: Entity.types.Date,
+    config: Entity.types.JSON,
+    owner: Entity.types.String,
+    emailOnError: Entity.types.Boolean,
+    providerData: Entity.types.JSON,
+  },
+  context: [],
```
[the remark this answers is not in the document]
djmitche: That's outside of the database-specific bits, and covered by the unit tests for the service itself, I think.

db/versions/0010.yml
djmitche: Good point -- that would be more efficient.

db/versions/0010.yml
djmitche: Fortunately tc-lib-entities always sets size, so it doesn't matter (I think it would be `limit NULL` which means no limit, anyway)
https://github.com/taskcluster/taskcluster/blob/54b7a8d276b358a4b3c832a34759c8393d7c6fde/libraries/entities/src/index.js#L463-L467

db/versions/0010.yml
djmitche: It removes the item on the right from the array on the left.

db/versions/0010.yml
djmitche: No, that's still not allowed - the update operation won't change the `worker_pool_id` column.

db/versions/0010.yml
djmitche: No, because in this case we're not doing an insert but an update -- but I could probably accomplish this with a single update query.  I'll give it a shot.

djmitche: Hm, https://www.postgresql.org/docs/9.4/dml-returning.html says
> In an UPDATE, the data available to RETURNING is the new content of the modified row.
so I think we wouldn't be able to return the `previous_provider_id` without selecting it first.  However, I think I can turn the update and final select into a single `update .. returning`.

db/versions/0010.yml
djmitche: Thanks!

db/versions/0010.yml
djmitche: (note that this was the only change I made to the DB migration step; the remainder were to the service migration step)

djmitche:
..updated based on review comments so far..

services/worker-manager/src/api.js
```diff
@@ -267,10 +252,7 @@ builder.declare({
 builder.declare({
   method: 'get',
   route: '/worker-pools',
-  query: {
-    continuationToken: /./,
-    limit: /^[0-9]+$/,
-  },
+  query: paginateResults.query,
```
helfi92, line 255: Nice!

services/worker-manager/src/providers/aws.js
```diff
@@ -137,12 +137,11 @@ class AwsProvider extends Provider {
...
       // The raw data should be 16KB maximum
       if (userData.length > 16384) {
-        return await workerPool.reportError({
+        return await this.reportError({
+          workerPool,
           kind: 'creation-error',
           title: 'User Data is too long',
           description: 'Try removing some workerConfiguration and consider putting it in a secret',
-          notify: this.notify,
-          WorkerPoolError: this.WorkerPoolError,
```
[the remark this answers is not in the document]
djmitche: This is now calling a method on the parent class, which uses the `notify` and `WorkerPoolError` instance properties for this.  In fact, the parameters were never needed here, and were ignored.

services/worker-manager/src/providers/google.js
```diff
@@ -167,10 +168,6 @@ class GoogleProvider extends Provider {
   }
 
   async removeResources({workerPool}) {
-    // remove any remaining providerData when we are no longer responsible for this workerPool
-    await workerPool.modify(wt => {
-      delete wt.providerData[this.providerId];
-    });
```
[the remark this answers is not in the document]
djmitche: None of the four provider implementations actually use workerpools' providerData.  But google was the only one to try to remove that providerData in `removeResources`.  Because of how I structured `update_worker_pool_provider_data`, it's actually impossible to *remove* it, so I think if we were putting any data here then `removeResources` would just set it to `{}`.  But that's what we do in the setup, so this seemed redundant.

tl;dr: unused but maybe useful someday, so I made all providers look the same in the simplest way possible

services/worker-manager/src/provisioner.js
```diff
@@ -161,76 +165,73 @@ class Provisioner {
       const poolsByProvider = new Map();
 
       // Now for each worker pool we ask the providers to do stuff
-      await this.WorkerPool.scan({}, {
-        handler: async workerPool => {
-          const {providerId, previousProviderIds, workerPoolId} = workerPool;
-          const provider = this.providers.get(providerId);
+      const workerPools = (await this.db.fns.get_worker_pools(null, null)).map(row => WorkerPool.fromDb(row));
```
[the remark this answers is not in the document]
djmitche: It could, but I bet that would be thousands of worker pools.  I think this is something we could double back on later, perhaps when we've added pagination based on keys (as you've suggested) and written some kind of postgres-to-async-iterator utility.

djmitche:
(now with updated `yarn generate`)

services/worker-manager/src/provisioner.js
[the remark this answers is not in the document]
helfi92: >  I think this is something we could double back on later, perhaps when we've added pagination based on keys

I'm not sure I understand. Doesn't `get_worker_pools` already support pagination?

> and written some kind of postgres-to-async-iterator utility

Could you remind me what was the benefit of adding this utility, please?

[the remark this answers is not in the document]
djmitche: It does, but using that pagination is a bit complicated -- enough that it would need additional testing.  I'd like to write a utility similar to paginateResults but for usage within the services that builds an AsyncIterator from a paginated DB call.

PR #2751, opened 2020-05-01
Ensure all deployments have unique container names

sciurus:
In order to track cpu and memory utilization per cronjob or deployment,
we need each one to have a unique container name. This is because
Stackdriver does not tag those metrics based on their metadata, but does
tag them based on the container name. Currently, it looks like all the
cronjobs have unique names but many of the deployments share container
names. This commit changes the deployment naming to match the cronjob
naming.

https://bugzilla.mozilla.org/show_bug.cgi?id=1616858


sciurus:
Should be good to land now @imbstack

PR #2754, opened 2020-05-01
Rename stored db function to expire_cache_purges

djmitche:
This function only just landed and certainly hasn't been released, so
there is no harm in changing it in-place.


djmitche:
That poor mac just works all the time..

PR #2755, opened 2020-05-01
[client-web] Remove external packages from webpack external

helfi92:
Move these packages instead to the list of peer-dependencies. It has the
same effect. webpack [seems to have an issue](https://github.com/webpack/webpack/issues/4824) with externals when the
build target is umd.

### DIFF
--- a/changelog/issue-1536.md
+++ b/changelog/issue-1536.md
@@ -1,4 +1,4 @@
-audience: developers
+audience: users 
 level: patch
 reference: issue 1536
 ---
--- a/changelog/issue-1536.md
+++ b/changelog/issue-1536.md
@@ -1,0 +1,5 @@
+audience: developers
+level: patch
+reference: issue 1536
+---
+taskcluster-client-web no longer shows the 'hawk is undefined' regression error.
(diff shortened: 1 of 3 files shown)


helfi92:
Actually, let me move them to peer dependencies...

changelog/issue-1536.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
```
djmitche, line 1: I think client users are "users" :)

clients/client-web/package.json
```diff
@@ -34,9 +34,11 @@
     "webpack-cli": "^3.2.3"
   },
   "dependencies": {
+    "taskcluster-lib-urls": "^12.0.0"
```
djmitche, line 37: Why is this one OK as a dependency while the others aren't?

helfi92: They're all ok as dependencies but treeherder has asked in the past to trim down the package size. They asked for the consumer to install their own `crypto-js`, `query-string`, `hawk` since they already use it in their app to do other stuff with it.

PR #2757, opened 2020-05-01
Bug 1633882 - use updated call to uploadReleaseAssert

djmitche:
See https://octokit.github.io/rest.js/v17#repos-upload-release-asset

### DIFF
--- a/infrastructure/tooling/src/publish/tasks.js
+++ b/infrastructure/tooling/src/publish/tasks.js
@@ -500,7 +500,6 @@ module.exports = ({tasks, cmdOptions, credentials, baseDir, logsDir}) => {
         draft: false,
         prerelease: false,
       });
-      const {upload_url} = release.data;
 
       const files = requirements['client-shell-artifacts']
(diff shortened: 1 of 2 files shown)


infrastructure/tooling/src/publish/tasks.js
```diff
@@ -510,21 +509,23 @@ module.exports = ({tasks, cmdOptions, credentials, baseDir, logsDir}) => {
...
 
         /* Artifact uploads to GitHub seem to fail.. a lot.  So we retry each
          * one a few times with some delay, in hopes of getting lucky.  */
         let retries = 5;
         while (retries-- > 0) {
           try {
             await octokit.repos.uploadReleaseAsset({
-              url: upload_url,
+              owner: 'taskcluster',
+              repo: 'taskcluster',
```
imbstack, line 521: do we hardcode that we are `taskcluster/taskcluster` anywhere else?

djmitche: A few lines earlier, among other places :)

PR #2768, opened 2020-05-04
Bug 1561668 - use p-synchronize to synchronize async functions

djmitche:
Bugzilla Bug: [1561668](https://bugzilla.mozilla.org/show_bug.cgi?id=1561668)


djmitche:
This diff looks a lot better if you turn off whitespace display..

Issue #2783, opened 2020-05-07
generic-worker tests are connecting to parent livelog, crashing it


djmitche:
https://community-tc.services.mozilla.com/tasks/ayBq3Tp8SiawxTzH6psUBA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FayBq3Tp8SiawxTzH6psUBA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log
```
2020/05/07 21:30:57Z Creating task feature Mounts/Caches...
2020/05/07 21:30:57Z Creating task feature Supersede...
2020/05/07 21:30:57Z Starting task feature Live Log...
2020/05/07 21:30:57Z WARNING: Livelog failure: error 'exit status 1' and livelog output:
2020/05/07 21:30:57 input server listening... 
2020/05/07 21:30:57 listen tcp :60022: bind: address already in use
2020/05/07 21:30:58Z Starting task feature OS Groups...
[taskcluster 2020-05-07T21:30:59.394Z] Uploading redirect artifact public/logs/live.log to URL https://community-tc.services.mozilla.com/api/queue/v1/task/ayBq3Tp8SiawxTzH6psUBA/runs/0/artifacts/public/logs/live_backing.log with mime type "text/plain; charset=utf-8" and expiry 2021-05-08T21:29:33.022Z
[taskcluster:error] exit status 141
```

141 is from EPIPE, I think.

djmitche:
(This started in #2736

djmitche:
Ah, #2736 changed `internalGETPort` from 60099 to 60022.  The LiveLogGETPort and internalGETPort are two different ports, but when using websocktunnel LiveLogGETPort is ignored, so I *think* this will be OK in production and only causes issues in testing.

The failure here is, I think, when the g-w under test tries to start its own livelog server using the default ports.  So the livelog process tries to bind to 60022 and gets an error, but at the same time the g-w under test connects to 60023 and makes a PUT request, but that port is for the livelog of the CI task, and causes that livelog to crash.

djmitche:
Ah, now I remember our conversations about this a week ago -- this never worked, we were just obscuring and ignoring the errors before.  So we need to find a way to invent three new ports for the generic-worker under test.

djmitche:
[...]

PR #2785, opened 2020-05-08
Bug 1636292 - fix importing of multiple-buffer base64 encodings

djmitche:
Bugzilla Bug: [1636292](https://bugzilla.mozilla.org/show_bug.cgi?id=1636292)

This was really pretty silly of me.  I assumed that entities would concatenate the buffers and then base64-decode, when in fact it base64-decodes, then concatenates.

This is a draft because, unfortunately, it's "cheating" to fix this in an older DB version, and for example staging won't re-run the version-8 migration script so it won't see a benefit from this change.  How should we handle this?  @helfi92 what do you think?


djmitche:
That makes sense!

edunham:
I can manually re-run arbitrary migrations for you to undo problems caused by a bad release to staging.

djmitche:
To do so would require downgrading from version 9 to 8 after rolling services back, too.  We want to avoid relying on extraordinary actions by ops folks if possible!

the generic-worker failures here are #2783

db/test/versions/huge_bufs.js
```diff
@@ -0,0 +1,2207 @@
+// Data from a real DB.  See 0008_test.js
```
helfi92, line 1: nit: I would probably put this inside `versions/fixtures` to keep things clean in that directory.

PR #2789, opened 2020-05-08
Generate a documented release tarball for docker-worker

### LINKED ISSUE #2788: create a release asset for docker-worker tarball
Like in [docker-worker-deploy](https://github.com/taskcluster/docker-worker-deploy/blob/master/deploy/bin/build#L32).  This should have a precise definition and a documented set of steps as to how to install this on an image.  This is the hand-off from TC as a development platform to users using the worker.  We *also* provide a tool to accomplish it (monopacker + skeleton), but use of that tool is optional.

djmitche:
Github Bug/Issue: Fixes #2788

There's a staging push [here](https://github.com/taskcluster/taskcluster/runs/657783821)

I snuck in an unrelated commit here to avoid uploading packges to npm / pypi until the GH release is done.

### DIFF
--- a/workers/docker-worker/release.sh
+++ b/workers/docker-worker/release.sh
@@ -1,30 +1,54 @@
-#!/bin/bash -e
+#! /bin/bash
 
-echo "Releasing docker-worker on Github"
+set -ex
 
-if [ "$DOCKER_WORKER_GITHUB_TOKEN" == "" ]; then
-  echo "You need to define the environment variable DOCKER_WORKER_GITHUB_TOKEN" >&2
-  echo "with a valid Github personal token." >&2
-  echo "If you intended to only deploy AMIs, then run the deploy.sh script." >&2
-  exit 1
-fi
+OUTPUT=docker-worker-x64.tgz
+while getopts "o:" opt; do
+    case "${opt}" in
+        o)  OUTPUT=$OPTARG
+            ;;
+    esac
+done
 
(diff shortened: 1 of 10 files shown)


workers/docker-worker/release.sh
```diff
@@ -1,30 +1,46 @@
...
+for f in src schemas .npmignore package.json yarn.lock config.yml bin-utils; do
+    cp -r $PWD/$f $DW_ROOT
+done
 
-if [ "$current_branch_head_sha" != "$remote_branch_head_sha" ]; then
-  echo "$branch and $remote/$branch mismatch!" >&2
-  exit 1
-fi
+# Install Node
+NODE_VERSION=12.11.0
```
imbstack, line 32: Once this lands can we stick a pointer to this in the bug for making d-w use the same node as services?

workers/docker-worker/release.sh
```diff
@@ -1,30 +1,46 @@
...
+mkdir $DW_ROOT/node
+curl https://nodejs.org/dist/v13.14.0/node-v13.14.0-linux-x64.tar.xz | tar -C $DW_ROOT/node --strip-components=1 -xJf -
```
imbstack, line 34: maybe we should say in the docs that this tarball is only for linux x64 and not arm or other platforms

djmitche:
One more look?  The g-w failures are just due to the livelog issue we fixed..

PR #2790, opened 2020-05-08
bug 1631824 - fix azure-provider to track and delete all of a VM's disks

milescrabill:
Bugzilla Bug: [1631824](https://bugzilla.mozilla.org/show_bug.cgi?id=1631824)

- [x] tests for new multiple disks deletion in the new testing infratructure
- [x] tested in my dev env

### DIFF
--- a/services/worker-manager/src/providers/azure.js
+++ b/services/worker-manager/src/providers/azure.js
@@ -632,7 +632,7 @@ class AzureProvider extends Provider {
           name: vm.storageProfile.osDisk.name,
           id: true,
         });
-        for (let disk of vm.storageProfile.dataDisks) {
+        for (let disk of vm.storageProfile.dataDisks || []) {
           disks.push(
             {
               name: disk.name,
@@ -885,8 +885,9 @@ class AzureProvider extends Provider {
           resource = w.providerData[resourceType];
         }
         resource.id = false;
-        if (_.has(deleteRequest, 'azureAsyncOperationHeaderValue')) {
-          resource.operation = deleteRequest.getPollState().azureAsyncOperationHeaderValue;
+        let pollState = deleteRequest.getPollState();
+        if (_.has(pollState, 'azureAsyncOperationHeaderValue')) {
+          resource.operation = pollState.azureAsyncOperationHeaderValue;
         }
       });
     }
--- a/services/worker-manager/src/providers/azure.js
+++ b/services/worker-manager/src/providers/azure.js
@@ -52,13 +52,14 @@ function generateAdminPassword() {
 
 function workerConfigWithSecrets(cfg) {
   assert(_.has(cfg, 'osProfile'));
+  let newCfg = _.cloneDeep(cfg);
   // Windows admin user name cannot be more than 20 characters long, be empty,
   // end with a period(.), or contain the following characters: \\ / \" [ ] : | < > + = ; , ? * @.
-  cfg.osProfile.adminUsername = nicerId().slice(0, 20);
+  newCfg.osProfile.adminUsername = nicerId().slice(0, 20);
   // we have to set a password, but we never want it to be used, so we throw it away
   // a legitimate user who needs access can reset the password
-  cfg.osProfile.adminPassword = generateAdminPassword();
-  return cfg;
+  newCfg.osProfile.adminPassword = generateAdminPassword();
+  return newCfg;
 }
 
 class AzureProvider extends Provider {
@@ -148,7 +149,6 @@ class AzureProvider extends Provider {
       const computerName = nicerId().slice(0, 15);
       const ipAddressName = `pip-${nicerId()}`.slice(0, 24);
       const networkInterfaceName = `nic-${nicerId()}`.slice(0, 24);
-      const diskName = `disk-${nicerId()}`.slice(0, 24);
 
       // workerGroup is the azure location; this is a required field in the config
       const workerGroup = cfg.location;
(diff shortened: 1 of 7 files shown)


services/worker-manager/src/providers/azure.js
```diff
@@ -52,13 +52,14 @@ function generateAdminPassword() {
 
 function workerConfigWithSecrets(cfg) {
   assert(_.has(cfg, 'osProfile'));
+  let newCfg = _.cloneDeep(cfg);
```
milescrabill, line 55: This bit is thrown in to ensure that we don't save `adminPassword` and `adminUsername` when we save parts of `cfg` into the VM `providerData.config`

services/worker-manager/test/fakes/schemas/azure-vm.yml
```diff
@@ -36,9 +36,18 @@ properties:
...
+          type: object
+          properties:
+            name: string
+            testProperty: {}
+          additionalProperties: false
+          required: [name]
       testProperty: {}
     additionalProperties: false
-    required: [osDisk]
+    required: [osDisk, dataDisks]
```
djmitche, line 50: dataDisks isn't required, right?

services/worker-manager/test/provider_azure_test.js
```diff
@@ -528,143 +540,184 @@ helper.secrets.mockSuite(testing.suiteName(), ['db'], function(mock, skipping) {
...
           default:
-            assert(false, `invalid expectation ${resourceType}: ${expectations[resourceType]}`);
+            if (index !== undefined) {
+              assert(false, `invalid expectation ${resourceType} ${index}: ${expectation}`);
+            } else {
+              assert(false, `invalid expectation ${resourceType}: ${expectation}`);
+            }
+        }
+      };
+      for (let resourceType of ['ip', 'vm', 'nic', 'disks']) {
```
djmitche, line 568: Nice refactor here :)

services/worker-manager/test/provider_azure_test.js
```diff
@@ -101,6 +101,10 @@ helper.secrets.mockSuite(testing.suiteName(), ['db'], function(mock, skipping) {
             hardwareProfile: {
               vmSize: 'Basic_A2',
             },
+            storageProfile: {
+              osDisk: {name: "default_os_disk"},
+              dataDisks: [{name: "default_data_disk"}],
```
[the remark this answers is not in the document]
milescrabill: - `storageProfile` is required in the schema
- added tests for the rest, given that we're disallowing naming disks

services/worker-manager/test/provider_azure_test.js
```diff
@@ -519,7 +530,8 @@ helper.secrets.mockSuite(testing.suiteName(), ['db'], function(mock, skipping) {
       ipName = worker.providerData.ip.name;
       nicName = worker.providerData.nic.name;
       vmName = worker.providerData.vm.name;
-      diskName = worker.providerData.disk.name;
+      disk0Name = worker.providerData.disks[0].name;
+      disk1Name = worker.providerData.disks[1].name;
```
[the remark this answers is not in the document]
milescrabill: I've changed things to bake in assumptions about multiple resource names, so disks automatically becomes `disks0`, `disks1`, etc.

services/worker-manager/test/fakes/azure.js
```diff
@@ -237,6 +237,8 @@ class VMResourceManager extends ResourceManager {
       provisioningState: 'Succeeded',
       // Azure generates uuids for vmIds, but we'll use something recognizable
       vmId: `vmid/${request.name}`,
+      // Fills in custom options from request
+      ...request.parameters,
```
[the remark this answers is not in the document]
milescrabill: I added handling for the disks properties (which is what we care about) in `VMResourceManager._requestToResource`

services/worker-manager/src/providers/azure.js
```diff
@@ -165,6 +165,17 @@ class AzureProvider extends Provider {
         workerConfig: cfg.workerConfig || {},
       })).toString('base64');
 
+      // Disallow users from naming diss
+      // required
+      let osDisk = {..._.omit(cfg.storageProfile.osDisk, ['name'])};
```
djmitche, line 170: We should probably throw an error in this case. Let's leave that to a followup (I think there are ways to disallow a property in json-schema, so that's probably the better fix).

services/worker-manager/test/fakes/azure.js
```diff
@@ -232,13 +233,21 @@ class VMResourceManager extends ResourceManager {
   // Subclass Overrides
 
   _requestToResource(request) {
+    let dataDisks = [];
+    for (let i = 0; i < request.parameters.storageProfile.dataDisks.length; i++) {
+      dataDisks.push({name: slugid.nice()});
```
djmitche, line 238: I kind of want these disks to also appear in the disk manager, but I see that for the tests as they are this doesn't matter (because we test provisioning and removing separately).  Maybe a comment to help someone later wondering why they don't exist?

services/worker-manager/src/providers/azure.js
```diff
@@ -620,13 +626,28 @@ class AzureProvider extends Provider {
...
 
       // VM
-      // NB: if we ever need disk id for anything, we could get it here
+      let vmModifyFunc = (w, vm) => {
+        let disks = [];
+        disks.push({
+          name: vm.storageProfile.osDisk.name,
+          id: true,
+        });
+        for (let disk of vm.storageProfile.dataDisks) {
```
djmitche, line 635: It might be good to allow the case where this doesn't exist with `|| []` here

services/worker-manager/src/providers/azure.js
```diff
@@ -837,8 +871,16 @@ class AzureProvider extends Provider {
...
-        w.providerData[resourceType].id = false;
-        w.providerData[resourceType].operation = deleteRequest.getPollState().azureAsyncOperationHeaderValue;
+        let resource;
+        if (index !== undefined) {
+          resource = w.providerData[resourceType][index];
+        } else {
+          resource = w.providerData[resourceType];
+        }
+        resource.id = false;
+        if (_.has(deleteRequest, 'azureAsyncOperationHeaderValue')) {
```
djmitche, line 881: This is looking for a property of the `getPollState()` return value, but on `deleteRequest` itself.  I suspect this will never be true..

PR #2799, opened 2020-05-11
Bug 1636174 - Add GITHUB_* env variables to docker-worker tasks

petemoore:
Bugzilla Bug: [1636174](https://bugzilla.mozilla.org/show_bug.cgi?id=1636174)


petemoore:
The [taskcluster-client-shell-1.13.7 failure](
https://community-tc.services.mozilla.com/tasks/QfiyeBHJTuaDuEiXEK2j1w/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FQfiyeBHJTuaDuEiXEK2j1w%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L183-195) looks unrelated to our docker-worker changes:

```
--- FAIL: TestUpdateCommand (0.46s)
    version_test.go:36: 
        	Error Trace:	version_test.go:36
        	Error:      	"No update available for linux
        	            	" does not contain "taskcluster"
        	Test:       	TestUpdateCommand
        	Messages:   	Update command not returning a valid output
FAIL
FAIL	github.com/taskcluster/taskcluster/v29/clients/client-shell/cmds/version	0.475s
?   	github.com/taskcluster/taskcluster/v29/clients/client-shell/codegen	[no test files]
?   	github.com/taskcluster/taskcluster/v29/clients/client-shell/codegen/cmd/gen-services	[no test files]
?   	github.com/taskcluster/taskcluster/v29/clients/client-shell/config	[no test files]
FAIL
```

petemoore:
The ["Docker-worker tests - 2" failure](https://community-tc.services.mozilla.com/tasks/E_J3cXM7Th2sCjSb-2DXzg/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FE_J3cXM7Th2sCjSb-2DXzg%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L1693-1706) is curious:

```
  1) docker image with manifest.json file
       docker manifest:
     Error: Run 0 was already claimed by another worker.
---
* method:     claimTask
* errorCode:  RequestConflict
* statusCode: 409
* time:       2020-05-13T22:28:06.154Z
      at /home/tester/git_checkout/workers/docker-worker/node_modules/taskcluster-client/src/client.js:362:21
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
      at async Context.<anonymous> (test/integration/docker_manifest_test.js:38:5)
```

It suggests that there were two workers competing for the claim of this task. However, each test-spawned worker gets a unique `workerType`, so that if multiple tests are running concurrently potentially on different machines, the tasks will only be available for the worker that the task was created for.

We can see that the task was claimed by a [test worker](https://community-tc.services.mozilla.com/api/queue/v1/task/i5--XQrAQOKMPtV85LSVBQ/status):

```
        "workerGroup": "random-local-worker",
        "workerId": "dummy-worker-M91-K6eXS",
```

and indeed that it was created for a test worker type:

```
    "provisionerId": "null-provisioner",
    "workerType": "test-5lzvgp7qaapojplnvbu1w-a",
```

 The [test code](https://github.com/taskcluster/taskcluster/blob/8b28741a287824708290ea9285950b5eb9f5577c/workers/docker-worker/test/integration/docker_manifest_test.js#L13-L41) also seems sane in that a single worker is created, a single task is claimed:

```js
    let worker = new TestWorker(DockerWorker, 'docker-worker', 'docker-worker');
    await worker.launch();

    const imageTaskId = slugid.v4();
    const taskDef = worker.TaskFactory.create({
      schedulerId: 'docker-worker-tests',
      taskGroupId: imageTaskId,
      payload: {
        image: 'taskcluster/test-ubuntu',
        artifacts: {
          'public/image.tar.zst': {
            type: 'file',
            expires: expires(),
            path: '/image.tar.zst',
          },
        },
        maxRunTime: 5 * 60,
      },
    });
    taskDef.provisionerId = 'null-provisioner';

    debug(`Creating image task ${imageTaskId}`);
    // create an artifact image with a manifest.json file
    await worker.queue.createTask(imageTaskId, taskDef);

    await worker.queue.claimTask(imageTaskId, 0, {
      workerGroup: 'random-local-worker',
      workerId: 'docker-worker',
    });
```

The possible causes I can think of are:
* Maybe the generated test `workerType` is not guaranteed to be unique, so the task was claimed by a test running on a different machine
* A genuine bug whereby `claimWork` gets called automatically when a worker is instantiated, and races with the explicit `claimWork` call in the test code

I suspect this is an intermittent issue, so have pushed an empty commit. But at the moment I'm not sure why this test should be intermittent.

On the bright side - it is a single failing test out of _all_ the docker worker tests, and the previous failures were due to genuine bugs in docker-worker, not a problem with the tests, so the tests seem to be serving us very well.

petemoore:
So looking more closely, I don't see how the test can really work, as it looks like launching a worker causes it to enter the `queue.claimWork` loop automatically, execute the tasks, upload artifacts and resolve them. However, the test has explicit an call to `claimTask`.

Furthermore, this test is using `claimTask` rather than `claimWork`. I don't think this task can ever have worked, so I will look at the test output from the old docker-worker repository - it looks like it was using an incarnation of `worker.launch()` that presumably just instantiated the worker without running it, at some point in the past.

petemoore:
So the [last time it passed](https://community-tc.services.mozilla.com/tasks/TtGf9laUSHe0zu3yDop_tg/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FTtGf9laUSHe0zu3yDop_tg%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L1788-1790) on docker-worker repo, we can see the task was generated for workerType `docker-worker` rather than e.g. `test-5lzvgp7qaapojplnvbu1w-a` as it was in the [recent failure](https://community-tc.services.mozilla.com/api/queue/v1/task/i5--XQrAQOKMPtV85LSVBQ/status) mentioned above. But indeed it was also failing on the docker-worker repo since April 22.

petemoore:
OK, looking more closely at the CI history in docker-worker, this seems to have been an intermittent test going back much further.

So it seems, strangely, that when the test passes, the task is generated for workerType `docker-worker` (which is passed in via the constructor of `TestWorker`) and when it fails, this value is not used, but the default worker type name is used. It looks like the `TestWorker.TaskFactory.workerType` is correctly set to the value passed in the TestWorker constructor, so I'm not sure why when the test fails, the task is generated for TestWorker.workerTypeName() instead. I think I'm pretty close now to solving the issue, since I can trace the code to see where the workerType property comes from when the task is created.

petemoore:
Oh wow, I see the task creation is done using https://github.com/lightsofapollo/object-factory

petemoore:
TIL https://github.com/lightsofapollo/task-factory/blob/master/task.js

helfi92:
Continued in https://github.com/taskcluster/taskcluster/pull/2832.

PR #2806, opened 2020-05-11
Bug 1637104 - use static binaries in utility docker images

djmitche:
Bugzilla Bug: [1637104](https://bugzilla.mozilla.org/show_bug.cgi?id=1637104)

### DIFF
--- a/infrastructure/tooling/src/publish/tasks.js
+++ b/infrastructure/tooling/src/publish/tasks.js
@@ -246,7 +246,7 @@ module.exports = ({tasks, cmdOptions, credentials, baseDir, logsDir}) => {
         dir: REPO_ROOT,
         logfile: path.join(logsDir, '/websocktunnel-build.log'),
         utils,
-        env: process.env,
+        env: {CGO_ENABLED: '0', ...process.env},
       });
 
       utils.step({title: 'Building Docker Image'});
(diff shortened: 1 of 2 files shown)


infrastructure/tooling/src/publish/tasks.js
```diff
@@ -246,7 +246,7 @@ module.exports = ({tasks, cmdOptions, credentials, baseDir, logsDir}) => {
         dir: REPO_ROOT,
         logfile: path.join(logsDir, '/websocktunnel-build.log'),
         utils,
-        env: process.env,
+        env: {CGO_ENABLED: '0', ...process.env},
```
imbstack, line 249: I hadn't heard of `CGO_ENABLED` before so I poked around a bit and most of the [articles I found referencing it](https://dominik.honnef.co/posts/2015/06/statically_compiled_go_programs__always__even_with_cgo__using_musl/) mention that you also need to tag the build with `netgo`. Is that true here? If not ignore me

petemoore: I don't think using the `netgo` build constraint is required. Go can be configured to use system DNS resolution, or use a go DNS resolver internally. If cgo isn't available, and there is no DNS resolution available, the internal resolver will be used instead.

Looking at the source, my bet is that if `cgo` isn't available, `preferGo()` will return `true` if `netGo` hasn't been explicitly set.

https://github.com/golang/go/blob/go1.13.7/src/net/conf.go#L130-L132

In any case, `CGO_ENABLED` is the big hammer that disables `cgo`, so if the `net` package required `cgo`, the build would fail.

There is some additional context in https://go-review.googlesource.com/c/go/+/11584/.

djmitche:
Thanks Pete!

PR #2807, opened 2020-05-11
Bug 1636189 - Docker-Flow stuff for websocktunnel

### LINKED ISSUE #2781: Please add build URL to version.json in Taskcluster Docker image
Version.json currently looks like this: 

```
$ docker run --rm --entrypoint=cat taskcluster/taskcluster:v29.3.0 version.json
{"version": "v29.3.0", "commit": "8d0600004fcaff7c1661e650bc48e424e7d409de", "source": "https://github.com/taskcluster/taskcluster", "build": "NONE"}
```

Cloudops would like to use this version information to validate Docker images when deploying them to Mozilla-run clusters. However, to validate a Docker image we also need the `build` parameter to contain the URL of the build which created the image, in order to pass the validation checks at https://github.com/mozilla-services/cloudops-infra-deploylib/blob/master/deploylib/docker.py#L207-L219 (basically checking that the image came from a build we trust). 

Please add the URL of the build that created the Docker image to the image's version.json. Thanks!

(relatedly, https://bugzilla.mozilla.org/show_bug.cgi?id=1636193)

djmitche:
Bugzilla Bug: [1636189](https://bugzilla.mozilla.org/show_bug.cgi?id=1636189)

In the process, I fixed #2781!

### DIFF
--- a/infrastructure/tooling/src/publish/tasks.js
+++ b/infrastructure/tooling/src/publish/tasks.js
@@ -13,6 +13,7 @@ const {
   pyClientRelease,
   readRepoFile,
   dockerPush,
+  dockerFlowVersion,
   REPO_ROOT,
 } = require('../utils');
 
@@ -24,15 +25,19 @@ module.exports = ({tasks, cmdOptions, credentials, baseDir, logsDir}) => {
   ensureTask(tasks, {
     title: 'Get release version',
     requires: [],
-    provides: ['release-version'],
+    provides: ['release-version', 'docker-flow-version'],
     run: async (requirements, utils) => {
       if (cmdOptions.staging) {
         return {
           'release-version': '9999.99.99',
+          'docker-flow-version': dockerFlowVersion({
+            gitDescription: 'v9999.99.99',
+            revision: '9999999999999999999999999999999999999999',
+          }),
         };
       }
 
-      const {gitDescription} = await gitDescribe({
+      const {gitDescription, revision} = await gitDescribe({
         dir: REPO_ROOT,
         utils,
       });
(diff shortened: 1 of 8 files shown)


infrastructure/tooling/src/publish/tasks.js
```diff
@@ -24,15 +25,16 @@ module.exports = ({tasks, cmdOptions, credentials, baseDir, logsDir}) => {
   ensureTask(tasks, {
     title: 'Get release version',
     requires: [],
-    provides: ['release-version'],
+    provides: ['release-version', 'docker-flow-version'],
     run: async (requirements, utils) => {
       if (cmdOptions.staging) {
         return {
           'release-version': '9999.99.99',
+          'release-revision': '999999999',
```
imbstack, line 33: I think this should be `docker-flow-version` but also be a JSON'd object

djmitche: Good point!

djmitche:
> Or maybe we should avoid committing to doing it right?

The DF spec is still vague in spots, and not versioned, so for the moment I think this seems best.

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1637229 in hopes of eventually doing some release validation.

djmitche:
Revised and with another [staging release](https://community-tc.services.mozilla.com/tasks/Kh9r54KkSz-nsb3197VY7g) running.

djmitche:
Another [staging release](https://github.com/taskcluster/taskcluster/runs/667149992)

djmitche:
Well, that staging run succeeded in building these images, but failed with #2808.

PR #2809, opened 2020-05-12
Bug 1637302: Compress artifact uploads

catlee:
Bugzilla Bug: [1637302](https://bugzilla.mozilla.org/show_bug.cgi?id=1637302)

### DIFF
--- a/workers/docker-worker/src/lib/features/artifacts.js
+++ b/workers/docker-worker/src/lib/features/artifacts.js
@@ -132,7 +132,6 @@ class Artifacts {
 
       let headers = {
         'content-type': mime.lookup(header.name),
-        'content-length': header.size,
       };
 
       // Check file extension
--- a/workers/docker-worker/src/lib/features/artifacts.js
+++ b/workers/docker-worker/src/lib/features/artifacts.js
@@ -135,6 +135,17 @@ class Artifacts {
         'content-length': header.size,
       };
 
+      // Check file extension
+      let entryExtension = header.name.split('.').slice(-1)[0];
+      let skipCompressionExtensions = ['jpg', 'jpeg', 'png', 'gif', 'webp',
+        '7z', 'zip', 'gz', 'tgz', 'bz2', 'tbz', 'whl', 'xz', 'zst', 'swf',
+        'flv', 'woff', 'woff2'];
+      let compress = false;
+      if (!skipCompressionExtensions.includes(entryExtension)) {
+        headers['content-encoding'] = 'gzip';
+        compress = true;
+      }
+
       try {
         let region = taskHandler.runtime.region;
         assert.ok(region);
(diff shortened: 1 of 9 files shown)


imbstack:
@catlee If you rebase on master and push, the tests are working for docker-worker now.

Also I'll set you up with scopes to run the tests locally. I'll ping you when it is all set up.

workers/docker-worker/src/lib/upload_to_s3.js
```diff
@@ -40,12 +42,21 @@ module.exports = async function uploadToS3 (
       await tmp.writeFile(source);
     } else {
       let stream = fs.createWriteStream(tmp.path);
-      await pipe(source, stream);
+      // TODO: Do we care if this is larger than the original?
```
catlee, line 45: I suspect we don't care. I noticed this in tests where the original file is very small (5 bytes) and the resulting compressed file is 25 bytes.

workers/docker-worker/src/lib/upload_to_s3.js
```diff
@@ -40,12 +42,21 @@ module.exports = async function uploadToS3 (
...
+        debug(`compressed ${gzip.bytesWritten} bytes`);
+      } else {
+        await pipe(source, stream);
+      }
     }
-    debug(`wrote source file to ${tmp.path} for ${artifactName}`);
-
     let stat = await fs.stat(tmp.path);
     size = stat.size;
+    httpsHeaders['content-length'] = size;
```
catlee, line 57: Is this safe to do? Can we stop setting content-length in the callers?

workers/docker-worker/src/lib/features/artifacts.js
```diff
@@ -143,7 +154,7 @@ class Artifacts {
         let start = process.hrtime();
 
         let {digest, size} = await uploadToS3(taskHandler.queue, taskId, runId, stream,
-          entryName, expiry, headers);
+          entryName, expiry, headers, null, null, compress);
```
catlee, line 157: I wasn't sure if there was a more idiomatic way of passing in the compress flag without these extra null parameters.

catlee:
TODO: write some tests for this behaviour

workers/docker-worker/src/lib/upload_to_s3.js
djmitche: Yes, it's safe.

workers/docker-worker/test/integration/file_artifact_test.js
```diff
@@ -337,4 +339,52 @@ suite('artifact extraction tests', () => {
 
     assert.ok(retry);
   });
+
+  test('automatic gzip compression', async () => {
```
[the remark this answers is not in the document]
djmitche: This covers both the compresed and not-compressed cases.

djmitche:
> This covers both the compresed and not-compressed cases.

(I got distracted and didn't finish that comment)

.. so I think these tests are sufficient.

PR #2812, opened 2020-05-13
allow for ssh:// source urls

### LINKED ISSUE #2811: allow for ssh:// source urls
I'm trying to figure out how to detect if a source url is in a private repo or a public repo, so I don't end up trying to search for a private repo every 404 [during CoT verification]. I thought it might be useful to use `ssh://git@github.com:org/reponame` source urls for private repos. The schema prevents that currently.

Filing so I have an issue to track my PR against :)

escapewindow:
Github Bug/Issue: Fixes #2811


escapewindow:
@ajvb can you think of any reason why we shouldn't do this?

PR #2818, opened 2020-05-13
adding logger in [client-shell]

### LINKED ISSUE #1061: [client-shell] Add -v / --verbose
This option should interface with the standard `log` library.  If the option is not given, the logs should be directed to a null output.  If the option is given, direct that output to stderr.

vnzongzna:
Github Bug/Issue: Fixes #1061

### DIFF
--- a/changelog/issue-1061.md
+++ b/changelog/issue-1061.md
@@ -1,5 +1,5 @@
 audience: users
-level: minor
+level: patch
 reference: issue 1061
 ---
 In client-shell added flag --verbose/-v for getting log to stderr for all the commands.
--- a/changelog/issue-1061.md
+++ b/changelog/issue-1061.md
@@ -1,0 +1,5 @@
+audience: users
+level: minor
+reference: issue 1061
+---
+In client-shell added flag --verbose/-v for getting log to stderr for all the commands.
--- a/clients/client-shell/cmds/root/log.go
+++ b/clients/client-shell/cmds/root/log.go
@@ -1,0 +1,28 @@
+package root
+
+import (
+	"os"
+
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	// Logger is the default log.Logger of the commands
+	Logger = &logrus.Logger{
+		Out:       os.Stderr,
+		Formatter: new(logrus.TextFormatter),
+		Hooks:     make(logrus.LevelHooks),
+		Level:     logrus.ErrorLevel,
+	}
+)
+
+// setup log output based on --verbose flag
+func setUpLogs(enable bool) {
+	Logger.SetFormatter(&logrus.TextFormatter{
+		TimestampFormat: "02-01-2006 15:04:05",
+		FullTimestamp:   true,
+	})
+	if enable {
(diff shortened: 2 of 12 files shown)


clients/client-shell/cmds/signin/signin.go
```diff
@@ -84,7 +86,7 @@ func cmdSignin(cmd *cobra.Command, _ []string) error {
 			fmt.Fprintln(cmd.OutOrStdout(), "export TASKCLUSTER_ACCESS_TOKEN='"+qs.Get("accessToken")+"'")
 			fmt.Fprintln(cmd.OutOrStdout(), "export TASKCLUSTER_ROOT_URL='"+rootURL+"'")
 		}
-		fmt.Fprintln(cmd.OutOrStderr(), "Credentials output as environment variables")
+		log.Infoln("Credentials output as environment variables")
```
djmitche, line 89: I think we'll want the user to see this even without `-v`.  What if we made the default level "Info" and changed most of the `Infoln` to `Debugln` here?

clients/client-shell/cmds/root/log.go
```diff
@@ -0,0 +1,28 @@
...
+		Hooks:     make(logrus.LevelHooks),
+		Level:     logrus.ErrorLevel,
+	}
+)
+
+// setup log output based on --verbose flag
+func setUpLogs(enable bool) {
+	Logger.SetFormatter(&logrus.TextFormatter{
+		TimestampFormat: "02-01-2006 15:04:05",
+		FullTimestamp:   true,
```
djmitche, line 23: For a command-line tool, I think timestamps are not required.  Can this be configured to omit them?

changelog/issue-1061.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: minor
```
djmitche, line 2: https://github.com/taskcluster/taskcluster/blob/master/dev-docs/best-practices/changelog.md#levels suggests this would be `patch` since it is not "something that would motivate a user or operator to adopt the upgrade".

clients/client-shell/cmds/signin/signin.go
vnzongzna: This makes sense. This way all the current `fmt.Fprintln` would still be visible and for further logging, we can use `log.Debug`

clients/client-shell/cmds/root/log.go
vnzongzna: Yes, we can omit it, by default it will show the time(in seconds) since the start of the program along with each log

clients/client-shell/cmds/signin/signin.go
djmitche: In fact I think most of the stuff that you've changed to use `log` in this PR is at the right level -- it's just this particular line that I suspect users will miss if it's not seen.

vnzongzna: We can change the default level to `InfoLevel` since that comes in precedence as compared to `DebugLevel`. That way all the current `Infoln` will be visible but further debug messages will need to written with `Debugln` and only visible with `-v` on.

PR #2819, opened 2020-05-13
Return correct dependency value for docker-worker artifact

djmitche:
This is what caused the 29.5.0 release to succeed but not do anything.


djmitche:
That's https://github.com/djmitche/console-taskgraph/issues/14.

PR #2824, opened 2020-05-14
Bug 1638002 - use region / location for workerGroup in dynamically-provisioned workers

djmitche:
Bugzilla Bug: [1638002](https://bugzilla.mozilla.org/show_bug.cgi?id=1638002)

Note that for AWS this is the best we can get, as it's possible to configure EC2 launchConfigs to allow EC2 to select the AZ.  For Azure, location is as specific as it gets.  But for Google, we could potentially use `cfg.zone` as that's a required configuration value.

### DIFF
--- a/services/worker-manager/test/provider_google_test.js
+++ b/services/worker-manager/test/provider_google_test.js
@@ -101,6 +101,8 @@ helper.secrets.mockSuite(testing.suiteName(), ['db'], function(mock, skipping) {
       name: 'foo',
       zone: 'whatever/a',
     });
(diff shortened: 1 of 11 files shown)


services/worker-manager/test/provider_google_test.js
```diff
@@ -101,6 +101,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['db'], function(mock, skipping) {
       name: 'foo',
       zone: 'whatever/a',
     });
+    // TODO: check that the metadata in the compute.instance.insert call has correct properties
```
imbstack, line 104: add a bug for this?

PR #2825, opened 2020-05-14
Bug 1637982 - disallow attempts to complete non-running runs

djmitche:
Bugzilla Bug: [1637982](https://bugzilla.mozilla.org/show_bug.cgi?id=1637982)

This bug has existed since the queue was written!


djmitche:
I briefly considered removing these legacy scopes (that have been "legacy" since 2015) but honestly I don't think this is a good time to be shipping a major, potentially breaking release.

Issue #2427, opened 2020-02-28
"Unauthorized: Bad mac" when viewing private log artifacts.


jschwartzentruber:
**Describe the bug**

When a task generates private `.log` artifacts, those logs are not viewable from the online log viewer.

**To Reproduce**

1. Create a task with private `.log` artifacts.
2. Once task has completed, click on the private log under Artifacts.
3. See error message.

**Expected behavior**

Private logs should be viewable using the log viewer.

**Taskcluster version**

v25.4.0

**Screenshots**

![image](https://user-images.githubusercontent.com/4673461/75563393-474a7b80-5a18-11ea-9a5d-ce6179790c03.png)

**Platform (please complete the following information):**

 - OS: Linux
 - Browser: Firefox
 - Version: 75.0a1 (nightly 2020-02-27)

**Additional context**

The task is: https://community-tc.services.mozilla.com/tasks/NRWtkio3QtizFc4Ccvuo-g

@djmitche [indicated on Matrix](https://chat.mozilla.org/#/room/#taskcluster:mozilla.org/$ROl5hL9xVmmEHPrWI_GsWGV6OEZ5tvEzgy5GmO9j9b8) this is probably the bewit being generated incorrectly.

The full response to the failing request is:
```
{
  "code": "AuthenticationFailed",
  "message": "Unauthorized: Bad mac\n\n---\n\n* method:     getArtifact\n* errorCode:  AuthenticationFailed\n* statusCode: 401\n* time:       2020-02-28T02:28:07.513Z",
  "requestInfo": {
    "method": "getArtifact",
    "params": {
      "0": "project/fuzzing/private/logs/live.log",
      "taskId": "NRWtkio3QtizFc4Ccvuo-g",
      "runId": "0",
      "name": "project/fuzzing/private/logs/live.log"
    },
    "payload": {},
    "time": "2020-02-28T02:28:07.513Z"
  }
}
```

djmitche:
You can reproduce this by copying the URL from the browser and running `curl -v <url>`.

I had a look at the bewit when I ran this.  I don't see anything wrong.  It's a base64-encoded string containing `<clientId>\<expiry>\<accessToken>\<ext>`.  For me, the clientId was `github/28673|djmitche`, the expiry was 30 minutes from the time of the request, and the `ext` contained a certificate with a reasonable set of scopes and appropriate `start` and `expiry` properties.

[...]

PR #2826, opened 2020-05-14
[Bug 1638047] - limit log viewer to public text/plain *.log files

escapewindow:
Bugzilla Bug: [1638047](https://bugzilla.mozilla.org/show_bug.cgi?id=1638047)

### DIFF
--- a/services/web-server/src/entities/Artifact.js
+++ b/services/web-server/src/entities/Artifact.js
@@ -3,7 +3,8 @@ module.exports = class Artifact {
     Object.assign(this, data);
     this.taskId = taskId;
     this.isPublic = /^public\//.test(this.name);
-    this.isLog = /^text\/plain*/.test(this.contentType);
+    // bug 1638047 - limit log viewer to public text/plain *.log files
+    this.isLog = /^text\/plain.*/.test(this.contentType) && /\.log$/.test(this.name) && this.isPublic;
 
     if (runId) {
       this.runId = runId;
--- a/services/web-server/src/entities/Artifact.js
+++ b/services/web-server/src/entities/Artifact.js
@@ -4,7 +4,7 @@ module.exports = class Artifact {
     this.taskId = taskId;
     this.isPublic = /^public\//.test(this.name);
     // bug 1638047 - limit log viewer to public text/plain *.log files
-    this.isLog = /^text\/plain.*/.test(this.contentType) && /\.log$/.test(this.name) && this.isPublic;
+    this.isLog = /^text\/plain.*/.test(this.contentType) && /\.log$/.test(this.name);
 
     if (runId) {
       this.runId = runId;
--- a/changelog/bug-1638047.md
+++ b/changelog/bug-1638047.md
@@ -2,4 +2,4 @@ audience: general
 level: patch
 reference: bug 1638047
 ---
-This release fixes a bug where the web UI opens the log viewer for any `text/plain` artifacts, which breaks for private artifacts. The web UI will now only use the log viewer for public `text/plain` `*.log` files.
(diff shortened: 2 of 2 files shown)


services/web-server/src/entities/Artifact.js
```diff
@@ -3,7 +3,8 @@ module.exports = class Artifact {
     Object.assign(this, data);
     this.taskId = taskId;
     this.isPublic = /^public\//.test(this.name);
-    this.isLog = /^text\/plain*/.test(this.contentType);
```
escapewindow, line 6: I wasn't sure if this `/^text\/plain*/` was a bug -- that means we can have 0 to infinite `n` characters, matching `text/plain` or `text/plainnnnnnnn`, no? Intentional?

helfi92: That was probably not intentional :)

services/web-server/src/entities/Artifact.js
```diff
@@ -3,7 +3,8 @@ module.exports = class Artifact {
...
+    // bug 1638047 - limit log viewer to public text/plain *.log files
+    this.isLog = /^text\/plain.*/.test(this.contentType) && /\.log$/.test(this.name) && this.isPublic;
```
helfi92, line 7: No need to add the last check `&& this.isPublic` to define what is a log. We do a bunch of checks with `this.isPublic` in https://github.com/taskcluster/taskcluster/blob/master/services/web-server/src/loaders/artifacts.js.

escapewindow: Hm. Per [bug 1638047](https://bugzilla.mozilla.org/show_bug.cgi?id=1638047), we currently treat private `text/plain` artifacts as logs, which breaks. Are you saying there are already checks to not do that?

helfi92: Adding `&& /\.log$/` doesn't fix this? My earlier comment was only referring to `&& this.isPublic`.

escapewindow: Won't the same problem occur when we upload a private artifact named `*.log` ?

escapewindow: https://firefox-ci-tc.services.mozilla.com/tasks/c1cPlHmTSmeKDfm81l30rQ#artifacts has a file named `releng/adhoc/test.log`, and it's marked as a log file. I'm logged in, with scopes to read that private artifact, but clicking on it gives me a 401 and saving it gives me an html page rather than the text logfile.

helfi92: Regarding the private artifact, I think this may be a different issue similar to https://github.com/taskcluster/taskcluster/issues/2427.

escapewindow: :+1: Do we want to leave it broken until #2427 is fixed? Or is it better to treat it as a standard text/plain artifact that we can view until #2427 is fixed?

helfi92: > Do we want to leave it broken until #2427 is fixed?

Sounds good to me.

> Or is it better to treat it as a standard text/plain artifact that we can view until #2427 is fixed?

I'm not sure adding `&& this.isPublic` would fix the issue given the issue is that the signed url doesn't look to be properly created. The problem seems to be the generated url rather than the log viewer. I suspect something is wrong when generating the bewit.

escapewindow: The bewit works for me for non-log private artifacts. I've gone ahead and removed the `&& this.isPublic` check though, just to get the fix sooner for other private text/plain artifacts.

helfi92:
The lint:go failure is unrelated to this PR.

escapewindow:
Thanks!

PR #2830, opened 2020-05-14
Truncate multiple tables at once in testing

djmitche:
This should save a bit of time by not setting up a new postgres
connection for each table.

### DIFF
--- a/libraries/testing/src/with-db.js
+++ b/libraries/testing/src/with-db.js
@@ -37,11 +37,13 @@ const resetDb = async ({testDbUrl}) => {
   }
 };
 
-const resetTable = async ({testDbUrl, tableName}) => {
+const resetTables = async ({testDbUrl, tableNames}) => {
   const client = new Client({connectionString: testDbUrl});
   await client.connect();
   try {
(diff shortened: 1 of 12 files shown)


libraries/testing/src/with-db.js
```diff
@@ -37,11 +37,13 @@ const resetDb = async ({testDbUrl}) => {
...
 
-const resetTable = async ({testDbUrl, tableName}) => {
+const resetTables = async ({testDbUrl, tableNames}) => {
   const client = new Client({connectionString: testDbUrl});
   await client.connect();
   try {
-    await ignorePgErrors(client.query(`truncate ${tableName}`), UNDEFINED_TABLE);
+    for (let tableName of tableNames) {
+      await ignorePgErrors(client.query(`truncate ${tableName}`), UNDEFINED_TABLE);
+    }
```
imbstack, line 46: do I understand correctly the call to `client.connect()` is the slow bit right? Doing these requests serially inside this connection is different than setting up a connection every time?

helfi92: That's my current understanding as well.

PR #2831, opened 2020-05-14
Remove the test notify.updateWidgets method

djmitche:
This was added to test deployment of postgres in a low-risk fashion.  It
is no longer required.


djmitche:
> Code LGTM. I assume the failure in the test has nothing to do with this :)

No, auth just fails sometimes because it uses a fixed port number and sometimes that port is already in use.

PR #2836, opened 2020-05-14
Misc worker docs changes

djmitche:


djmitche:
lint:go is just because this is an old base revision and we updated the linter in the docker image.

PR #2840, opened 2020-05-15
Bug 1636164 - use the same version of node for docker-worker as every…

djmitche:
Bugzilla Bug: [1636164](https://bugzilla.mozilla.org/show_bug.cgi?id=1636164)

This doesn't update the ci image.  It calls `make -c git` when there's no Makefile present, and then tries to install a "custom git plugin".  Anyone have any idea what that is?  That image is also using Ubuntu 14.04.  Hopefully the work on a decision task will make that unnecessary.


djmitche:
I'm confused why they're failing -- how is that related to this patch?

petemoore:
> I'm confused why they're failing -- how is that related to this patch?

@djmitche I wonder if this is related to #2852.

djmitche:
I suspect this is related to the docker image in use -- https://hub.docker.com/layers/taskcluster/worker-ci/0.4.4/images/sha256-e4357e0966e0265a3a45294d5902040b9de0303c3acb3271ee8ed81671b5461c?context=explore was pushed a week ago.  If anything, I would have expected to see an error about an incompatible node version, since engineStrict is set.

As I noted in the first comment, I was unable to rebuild this image, so I didn't.  How were you able to build it?

djmitche:
> As I noted in the first comment, I was unable to rebuild this image, so I didn't. How were you able to build it?

This had some "Makefile" magic to try to build manpages to embed in the image.  I just removed that and it worked fine.

djmitche:
Also, I've kept the `@sha256:..` bits for these images, but image integrity will be solved in general when we move to taskgraph.

djmitche:
Pete has deferred to @helfi92

workers/docker-worker/test/images/ci/VERSION
```diff
@@ -1 +0,0 @@
-0.4.4
```
helfi92, line 1: How is this specified now?

djmitche: It's similar to the other node-and-xxx docker images: `taskcluster/node-and-docker:node12.16.3`. Which is a little janky since it includes a copy of docker as well, but we can add that version to the tag when we change it -- I think there's more work to do there on checking docker-worker's docker compatibility.

djmitche: Oh, that jankiness will be addressed with the decision task and automatically building images.

PR #2843, opened 2020-05-15
Include payload schema in docker-worker docs

djmitche:
This also includes some minor edits to the docs, and puts the
docker-worker payload into YAML format.


djmitche:
Oops! I was hoping to find a test that covered this code but did not..

djmitche:
It looks like the tests do cover that code, since they all timed out.

I'll see how hard it would be to add a unit test.

djmitche:
Oh, and the JSON to YAML was done with node and js-yaml so it should be airtight.

Issue #2844, opened 2020-05-15
Run all or most API services in a mock configuration


djmitche:
Investigate the possibility of a general mock-services thing against which workers, etc. can run.  How difficult would this be?

Think about using [supertest](https://github.com/visionmedia/supertest) to avoid the need to make a real HTTP connections.

Ideally this would scale to something that can be exposed for testing go things against, too.

djmitche:
Ramblings..
---
* services require cwd for loading their config (easy fix)
* services need to be able to find each other (new discovery scheme?)
* lots of dependencies need to be faked (other clouds, artifacts, pulse, etc.)
* need some API for configuring fakes and making assertions (resetting to default state, injecting data, etc.)

Incidental benefit should be an easy setup for tests within a service, eliminating some of the chaos that is `test/helper.js`

Could we put everything in one big loader?
 * parameters e.g., monitor(serviceName)
 * ability to use parameters in dependencies so e.g., db(serviceName) can get config from cfg(serviceName)
 * define by calling some registration function, so libraries can do some things directly (pulseClient, db, etc.)

Tests already run in "mock" and "real" mode, so either mocking everything or using the "real" everything.  So maybe "mock" becomes a parameter that's passed around from loader component to loader component, too?

djmitche:
So I'm working on a "One Loader To Rule Them All" kind of thing.  It adds "parameters" to loader components, one of which is serviceName, so it's possible to load `api{"serviceName":"queue"}` and `api{"serviceName":"secrets"}` in the same process, for example.  And I'm setting things up so that basically each library registers its components, leaving `main.js` very sparse.  Hopefully I can get to a spot where there's a single entry point to all taskcluster services in `services/main.js`.  I want to be careful not to `require` the whole world for all services, but that's not turning out to be pretty easy so far.

This is going to make it much easier to get information about all services (references, for example).

I need to think a little more about supporting testing, but I'm anticipating having a `mock` parameter that is set to true when testing and passed around everywhere.  That can "do the right thing" in each component.  We can set that parameter in unit tests, making helper.js quite a bit simpler.

But the coolest bit is, we can gather all of the services' APIs together in a single process, with `mock=true`, and serve them in a single Express app.  We can then access that either with supertest or by serving it via HTTP.

I actually don't see a great benefit to supertest in this particular project: we want to build a mock that can be called, whereas supertest is mostly about writing `expect` to define what is expected in response.  It might be useful for writing service unit tests, but not in this project.

[...]

PR #2853, opened 2020-05-18
Remove services' dependency on cwd

djmitche:
All services can now be run from the root of the monorepo.  The only
libraries that used the service directory were tc-lib-validate and
tc-lib-config.  Both now take a serviceName and calculate the path
appropriately.

This is a (small!) step toward running all services in the same process.

I tested this by running it in my dev environment.  Everything seems to start up just fine.

Github Bug/Issue: Refs #2844


imbstack:
looks like maybe .tc.yml needs updated to handle this?

djmitche:
Note to self: lib-monitor also has a soft dependency on appRootDir.

djmitche:
OK, I did a good bit more cleanup, so this should be ready for another look.

PR #2854, opened 2020-05-18
Mock services in generic-worker tests

petemoore:
Tests typically take several minutes to run, depending on platform and engine combination. By running tests against mocks, developers can have some confidence about the validity of their changes reasonably quickly (tests take between 1 and 2 minutes when running against mocks).

Note there are several caveats here:
* Mocks are not real services - when a test passes against a mock, there is no guarantee it will pass when running against the real service
* Each platform and engine operate differently, so when running tests against mocks locally, typically only the simple/docker engines can be tested against the current OS (multiuser engine creates OS users, which is not ideal for developer environments)
* Maintaining mocks is burdensome and error prone - it would be much better if the tests would deploy the actual queue/secrets/worker manager/purge cache and auth services locally, and only mock the services that we don't have the source code for in the same repo (i.e. gcp/aws/azure metadata mocks, s3 mocks)
* The abstraction of having mocks increases code complexity
* Changes to code require that the CI runs anyway, since we need test coverage of all OSes (Windows/macOS/Linux) - therefore having the facility to run against mocks doesn't mean you can have adequate test coverage without pushing your changes and checking the CI results; and for users of the CI, the process is no different whether the CI use mocks or real services - you only know if your change is ok once the CI tasks have finished
* In order to keep current level of coverage, CI will need to continue running against real services, otherwise we won't always know if things are broken.

So the long and short of it is, yes this introduces mocks, but no they aren't a silver bullet. They are a maintenance overhead, and they don't change the experience greatly for developers. Now they can run the tests locally against mocks, but this coverage is only for their OS, and won't catch as many problems as the tests when they run against a real deployment, which they should do in the CI. They could already run tests locally against a real deployment of taskcluster, if they obtained the single required scope to do so.

For now, the CI will run against the mocks, and I have filed #2994 to make sure the CI runs against real services in future, but tests will run by default against mocks. For now, the tests only run against mocks, but I don't want to invest further effort at this time, as this is not in scope for the current sprint.


workers/generic-worker/tc/factory.go
```diff
@@ -0,0 +1,41 @@
...
+type ServiceFactory interface {
+	Auth(creds *tcclient.Credentials, rootURL string) Auth
+	Queue(creds *tcclient.Credentials, rootURL string) Queue
+	PurgeCache(creds *tcclient.Credentials, rootURL string) PurgeCache
+	Secrets(creds *tcclient.Credentials, rootURL string) Secrets
+	WorkerManager(creds *tcclient.Credentials, rootURL string) WorkerManager
+}
+
+type ClientFactory struct {
+}
```
djmitche, line 21: I like this pattern!  Could this `tc` package go in `internal/` and be used by worker-runner as well?

Also, note that if you like to drop support for talking to worker-manager directly from g-w (since that's now more correctly done by worker-runner), you can save having to mock secrets and worker-manager.  (although worker-runner will need them mocked, but that's already implemented)

petemoore:
The failures above are intermittent docker-worker failures, unrelated to the PR.

I haven't managed to clean up the commit history yet, but it is ready for a review.

I would have preferred to not skip the tests that I have skipped, but unfortunately I ran out of time. I'll create an issue for a follow up activity to reenable those tests.

In short, I have the implementation working with interfaces, and the mocks implemented, but tests that execute tasks that themselves call taskcluster APIs do not currently work, except for worker manager API endpoints and secrets API endpoints, which service the HTTP request.

Our current client generation code generates stubs which take go types as inputs and translate them into HTTP requests. My idea for a future sprint is that our code gen also generates adapters in the opposite direction - http handlers that service incoming HTTP requests, and service them by calling a configurable implementation that implements the same interfaces that the client implements. In other words, with code gen in the opposite direction, we have a way to implement mocks (or even real services) using native go types, and have the HTTP translation automated in both directions so that clients and servers can use the same underlying go type representations, and have the static type enforcement at compile time for server implementations. But that is all for another day.... It will also mean we can use the same stubs to service docker-worker tests, or even tests of any of the other services, as we will then have general purpose stubbing capabilities across all of our services.

petemoore:
Will rewrite the git history when I land. :-)

djmitche:
https://bugzilla.mozilla.org/show_bug.cgi?id=1640267
https://bugzilla.mozilla.org/show_bug.cgi?id=1640911

workers/generic-worker/aws_helper_test.go
```diff
@@ -1,29 +1,25 @@
 package main
```
djmitche, line 1: I think we should remove the `--configure-for-aws`/`gcp`/`azure` functionality.  For Azure it doesn't work anymore anyway (customdata is no longer in the instance metadata), and we may have broken the others as well. The supported way to run generic-worker in a cloud is with worker-runner, although it's still supported to run generic-worker *without* worker-runner in a standalone fashion.

It's already updated in this PR so we can just delete the code it in a followup, but if in revising this PR the configure-for-cloud functionality gives you any trouble, I'd suggest preferring to remove rather than fix.

workers/generic-worker/main_broken_on_docker_test.go
```diff
@@ -3,17 +3,35 @@
...
+		////////////////////////////////////////////////////////
+		t.Skip()
+		////////////////////////////////////////////////////////
+		////////////////////////////////////////////////////////
+		////////////////////////////////////////////////////////
+		////////////////////////////////////////////////////////
+		////////////////////////////////////////////////////////
+		////////////////////////////////////////////////////////
+		////////////////////////////////////////////////////////
+		////////////////////////////////////////////////////////
```
djmitche, line 33: ok, there are a lot of these so I'll stop flagging them :)

workers/generic-worker/taskcluster_proxy_broken_on_docker_test.go
```diff
@@ -9,14 +9,42 @@ import (
...
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+
+	// The following nonsense is to fool the linter that these go identifiers
+	// are used, since they are needed by tests that are currently skipped
```
djmitche, line 23: This is a good reason to delete, rather than skip, the tests, and point to the existing tests in version control history in the followup bugs.

workers/generic-worker/tc/factory.go
djmitche: bump :)

workers/generic-worker/tc/services.go
```diff
@@ -0,0 +1,51 @@
...
+	ListArtifacts(taskId, runId, continuationToken, limit string) (*tcqueue.ListArtifactsResponse, error)
+	ReclaimTask(taskId, runId string) (*tcqueue.TaskReclaimResponse, error)
+	ReportCompleted(taskId, runId string) (*tcqueue.TaskStatusResponse, error)
+	ReportException(taskId, runId string, payload *tcqueue.TaskExceptionRequest) (*tcqueue.TaskStatusResponse, error)
+	ReportFailed(taskId, runId string) (*tcqueue.TaskStatusResponse, error)
+	Status(taskId string) (*tcqueue.TaskStatusResponse, error)
+	Task(taskId string) (*tcqueue.TaskDefinitionResponse, error)
+}
+
+type Artifacts interface {
```
djmitche, line 48: This interface is not like the others, in that it's not for a TC service.  Some comments explaining what it is would help.

workers/generic-worker/tc/services.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/taskcluster/taskcluster/v30/clients/client-go/tcauth"
+	"github.com/taskcluster/taskcluster/v30/clients/client-go/tcpurgecache"
+	"github.com/taskcluster/taskcluster/v30/clients/client-go/tcqueue"
+	"github.com/taskcluster/taskcluster/v30/clients/client-go/tcsecrets"
+	"github.com/taskcluster/taskcluster/v30/clients/client-go/tcworkermanager"
+	"github.com/taskcluster/taskcluster/v30/workers/generic-worker/tclog"
+)
+
+type Auth interface {
```
djmitche, line 15: This seems to be the same pattern used in the worker-runner tests.  It would be good to have similar explanatory comments as to what the interfaces are for.
```
// An interface containing the functions required of WorkerManager, allowing
// use of fakes that also match this interface.
```

workers/generic-worker/tclog/tclog.go
```diff
@@ -0,0 +1,21 @@
...
+
+type Logger interface {
+	Log(prefix, message string)
+	Info(message string)
+	Infof(format string, v ...interface{})
+	Warn(message string)
+	Warnf(format string, v ...interface{})
+	Error(message string)
+	Errorf(format string, v ...interface{})
+}
```
djmitche, line 11: It looks like this is only used for `Artifacts.GetLatest`?  Is the plan to use it more extensively?  Some commentary on what it's for and what is expected to satisfy the interface (I would not have guessed a task!) would be helpful to the reader and anyone using or modifying this later.

workers/generic-worker/README.md
```diff
@@ -8,7 +8,7 @@ For documentation of the worker from a user's perspective, see the [online docum
 
 To see a full description of all the config options available to you, run `generic-worker --help`:
 ```
-generic-worker (multiuser engine) 28.2.2
+generic-worker (multiuser engine) 30.0.1
```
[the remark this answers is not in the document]
petemoore: Created #2981

workers/generic-worker/aws_helper_test.go
petemoore: Let's hold off deleting it until we have the new deploymentId support in worker-runner. At that point, I think we'll have feature parity.

workers/generic-worker/tc/factory.go
petemoore: Let's move the tc package in a follow up PR - this PR is already getting pretty large and dragging on.

workers/generic-worker/tc/services.go
petemoore: I've added as a package comment.

workers/generic-worker/tclog/tclog.go
petemoore: That was a bit of a mess. There was a utility method to download from a URL which was used by the mounts feature and also by the mock framework. I had pulled it out, but not finished documenting it or cleaning it up. In a cleanup since, I've removed the artifact interface that I previously had, that was using it, so now it is only used by mounts feature, and since it has mounts feature specific logging in, I've removed the tclog package entirely, it isn't needed.

workers/generic-worker/main_broken_on_docker_test.go
petemoore: No longer skipping any tests. :-)

workers/generic-worker/taskcluster_proxy_broken_on_docker_test.go
petemoore: Fixed tests. :-)

workers/generic-worker/tc/services.go
petemoore: I've removed it. :-)

workers/generic-worker/artifacts_multiuser_test.go
```diff
@@ -222,6 +239,24 @@ func TestChainOfTrustUpload(t *testing.T) {
...
+	////////////////////////////////////////////////////////
+	t.Skip()
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
+	////////////////////////////////////////////////////////
```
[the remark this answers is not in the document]
petemoore: Test fixed.

workers/generic-worker/artifacts_multiuser_test.go
```diff
@@ -18,6 +18,23 @@ import (
...
```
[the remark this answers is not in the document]
petemoore: Test fixed.

petemoore:
> This is a very close match for the mocks in worker-runner. It should be pretty quick to merge the two into something in `internal/` so that both applications can use it. I think that should happen in this PR as otherwise the two sets of mocks will slowly drift apart and the planned merging will become progressively more difficult.

I'm not sure `internal` would be the right place - that would mean that the mocks would be unavailable to consumers of taskcluster, and probably the mocks are one of the useful libraries that have value to people that write tools that interface with taskcluster. Perhaps we could have a libraries folder, that included the mocks package. Then people could import the package, and use it to run tests of their own tools against our mocks.

I much prefer not throwing worker-runner into the mix in this PR - if I merge the mocks of these two tools in this PR, I will be introducing more noise with potential changes to worker-runner and generic-worker at the same time, whereas this current PR has quite clean scope. If the change should be relatively small, it should still be small after the PR lands, so I think two PRs, one after the other, would be a better approach, to keep the logical steps separate. If you're really keen for this to be done now, let's raise it with @imbstack to see where it fits in relation to other items in the current sprint backlog. I can also review, if you want to do it yourself, and Brian is ok with that.

Of course I agree there is no sense in maintaining two sets of mocks, my preference would actually be to have no mocks at all, and start the real services up. Having three implementations to maintain would be even worse than having two. 😆

PR #2855, opened 2020-05-18
Refactor tc-lib-monitor to not register a global serviceName

djmitche:
This changes how registration is done, from registering on a singleton
that is specialized to the serviceName, to registering via a static
method and only specifying the service name at setup (and when
generating references).

This is a pretty big patch, but is fairly formulaic:
 * services change how they set up tc-lib-monitor in main.js
 * most services drop `monitor.js` in favor of putting log types in the files that call them
 * libraries change how they register their log types
 * tests change how they assert that messages are logged, by looking at `monitor.manager.messages`

Github Bug/Issue: Refs #2844


djmitche:
That's a good point -- I can return to putting all of those declarations in monitor.js.

Issue #2845, opened 2020-05-15
do not require special scopes to run docker-worker tests


djmitche:
Using something like that in #2844, or perhaps a much more limited set of fakes.

djmitche:
It's become pretty clear that getting #2844 to the point where it can support docker-worker tests is a pretty massive project, and not going to be completed any time soon.

djmitche:
I'm going to get this to the point where running `yarn test` works out of the box, but skips most everything.  Then I'll work on #2856.  Then I'll come back to this and write fakes.

djmitche:
I put a few good hours into trying to get this to work with fakes.  I don't think it's possible without the near-certainty of introducing new bugs.

The integration tests currently spawn a new worker process, with which they communicate by dropping files on-disk and by communicating with the production TC services (e.g., to monitor for a task completing).  So, we couldn't do this with the usual fake objects -- it would have to be a full HTTP API against which we could run the worker process.  And having done so, we'd still end up with the failure-as-timeout behavior we see now.  We would also need to find a way to indicate that the spawned process should not expect to find `/dev/video*`, etc., and to fake their presence for the tests that expect it.

I looked at the possibility of running the test worker in-process, but the startup process for docker-worker is extremely convoluted and mixes a lot of modifications to global state (`taskcluster.config(..)`, event listeners on `process`, etc.) with a dense web of interdependent "managers" all pointing to one another.  There's no way to encapsulate that into a single state, nor to do dependency injection safely.   

If this was code we wanted to keep, the right step would be to take the whole thing to bits and put it back together in a better-isolated, testable way -- but that's likely to take a long time and introduce new bugs along the way, so I don't think it's worthwhile.

I see this as an object lesson in the risks of relying on integration tests alone: the resulting balancing act between tests and code can become so fragile that it's impossible to make changes efficiently or with confidence in their correctness.

PR #2857, opened 2020-05-19
Skip d-w tests by default requiring creds or docker

djmitche:
Github Bug/Issue: Refs #2845 

This is just the first phase: make `yarn test` work out of the box.  While we could just require docker, I want to have a way for users to opt-in to using Docker, as the tests do remove images at least, and may do other destructive stuff that might be surprising.

With this change, locally and without creds:

```
  25 passing (2s)                                                                              
  115 pending                                    
```


djmitche:
```
Error: Cannot find module 'debug'
Require stack:
- /home/tester/git_checkout/libraries/testing/src/schemas.js
- /home/tester/git_checkout/libraries/testing/src/index.js
- /home/tester/git_checkout/workers/docker-worker/test/garbage_collection_test.js
- /home/tester/git_checkout/workers/docker-worker/node_modules/mocha/lib/mocha.js
- /home/tester/git_checkout/workers/docker-worker/node_modules/mocha/lib/cli/one-and-dones.js
```
This is because libraries/testing is outside of the node package that we've `yarn install`'d..

(closing and re-opening to get some more ci workers so I can see what kernel they're running, not in the hopes the result will be different)

djmitche:
I gave up on phase 2 here.    So this fixes #2845.

PR #2858, opened 2020-05-19
[Bug 1638370] Fetch vmId in registerWorker for azure provider if it does not exist yet

imbstack:
This should ensure we always have the vmId for registerWorker if the
vm exists. 

This is safe to do because we are using the vm name to fetch the vmId
which is what we are doing in checkWorker anyway. Even if someone passes
in a different worker to `registerWorker` we'll get a different vmId than
the one they can sign and provide as proof.

Bugzilla Bug: [1638370](https://bugzilla.mozilla.org/show_bug.cgi?id=1638370)

### DIFF
--- a/changelog/bug-1638370.md
+++ b/changelog/bug-1638370.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: bug 1638370
+---
+Azure provider no longer has a race condition between `registerWorker` and `checkWorker`.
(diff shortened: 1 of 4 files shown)


imbstack:
I think it's actually harder than that. That bit of azure provider is where we detect that a worker has terminated itself so that we can mark it as stopped. If we never mark a `RUNNING` worker as `STOPPED` they live forever.

I've asked relops to tell us if workers that are in a failed state due to agent issues and workers that are terminated by the worker itself can be distinguished from the provisioning/power states.

PR #2860, opened 2020-05-19
[Bug 1636167] Use decision task for CI

imbstack:
Todos before this is ready to land:
- [x] add generic-worker tasks
- [x] ~add go-mod-tidy~ (part of g-w tasks)
- [x] add worker-runner tests
- [x] add internal-go-libraries
- [x] add all client tests
- [x] add db test
- [x] add meta tests
- [x] tools-1.13.7
- [x] upload-coverage
- [x] yarn build/changelog/generate
- [x] db:upgrade
- [x] various things about release, check to see what is left not running on master on a normal push that needs added to this list as well
- [x] make sure NO_TEST_SKIP is set where it is supposed to

Bugzilla Bug: [1636167](https://bugzilla.mozilla.org/show_bug.cgi?id=1636167)

Below I've added a stream-of-consciousness list of TODOs for after this lands that we can turn into cards or bugs or something.

- this will probably land with gw-decision-task intact, we should move its logic into the normal decision task
- perhaps merge most things into a couple high-level kinds with `tasks_from`
- bring back caches
- images are not being built in the taskgraph way here, let's do that
- even within our manually created images, worker-ci is a one-off, let's undo that
- (nice to have) generate d-w test chunks
- switch to run-task
- this seems to remove the ability to rerun by making actions for it that retrigger instead but we don’t support actions here yet, let's support actions
- it appears support for generic-workers url mounts doesn't allow sha256 to be passed. we should add support for that and bring it back
- I have a feeling that a lot of these transforms, etc could be a lot simpler if we take another pass knowing how this all fits together


taskcluster/src/loader/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+
+from pathlib2 import Path
+
+from taskgraph.util.templates import merge
+
+
+logger = logging.getLogger(__name__)
+
+
+def services_and_libraries_loader(kind, path, config, parameters, loaded_tasks):
```
djmitche, line 11: This is pretty cool :)

imbstack:
Ok, this is ready for a 80% review I think! I'll resolve the conflict tomorrow which I think is just the changes to docker-worker test image mentioned by @djmitche above.

@petemoore I'm wondering what our best plan to get g-w integrated here is since it has its own decision task already. would it be easiest to trigger that decision task from this one and let it spawn tasks or convert its logic into this decision task?

imbstack:
The queue failure is intermittent I think.

Bonus points if somebody who is up before me makes a PR into this branch either adding generic worker tasks or one of the TODOs at the bottom of the description above.

djmitche:
Failing test is "Check expire doesn't drop table".  I have no idea what that ever could have meant -- did we expire artifacts by dropping the Artifacts azure table??

imbstack:
I can't believe it but I think this is ready to land assuming the TODO list is acceptable. Please feel free to land this without me if it looks good to save me from merge conflicts when I get back because this touches a lot of stuff.

I'll add change the TODOs into issues for us to do later if we decide to land this as well.

imbstack:
Block landing this on mozilla/community-tc-config#277

PR #2864, opened 2020-05-19
Fail with a useful error message when audio/video loopback are not available

djmitche:
This seems a common error mode with docker-worker, and until today I didn't realize it could be configured around!

This should at least give better logging..


djmitche:
https://bugzilla.mozilla.org/show_bug.cgi?id=1639582 for the intermittency in g-w.

PR #2866, opened 2020-05-20
Tests cover both having and not having publicIP in config

petemoore:
I spotted that when no public IP is provided, that the worker can crash (it happened for me when running generic-worker locally) - this PR is to see if the tests already capture this, when `publicIP` config isn't provided in tests.

### DIFF
--- a/workers/generic-worker/helper_test.go
+++ b/workers/generic-worker/helper_test.go
@@ -404,7 +404,6 @@ func GWTest(t *testing.T) *Test {
 			NumberOfTasksToRun:        1,
 			PrivateIP:                 net.ParseIP("87.65.43.21"),
 			ProvisionerID:             "test-provisioner",
-			PublicIP:                  net.ParseIP("12.34.56.78"),
 			Region:                    "test-worker-group",
 			// should be enough for tests, and travis-ci.org CI environments don't
 			// have a lot of free disk
--- a/workers/generic-worker/mounts_broken_on_docker_test.go
+++ b/workers/generic-worker/mounts_broken_on_docker_test.go
@@ -5,6 +5,7 @@ package main
 import (
 	"encoding/json"
 	"io/ioutil"
+	"net"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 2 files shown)


petemoore:
Note, this is a great example of an integration test failure that wouldn't have happened if the tests had run against a mock!

petemoore:
The cause is that the payload is invalid for the `queue.createArtifact` API call that tries to redirect the livelog artifact to the running livelog. The payload of that API request looked like this:

```json
    "payload": {
      "contentType": "text/plain; charset=utf-8",
      "expires": "2020-05-20T17:55:37.523Z",
      "storageType": "reference",
      "url": "http://<nil>:41629/log/HK0dhkS7SYuPAH2C0i86OQ"
    }
```

petemoore:
The redirection should be to websocktunnel, I'm not sure why the redirection url looks like a stateless dns type url.

djmitche:
Stateless DNS has been removed.

petemoore:
> Stateless DNS has been removed.

Yeah, that is what I was thinking. I wonder if some part got accidentally left in. I can take a look soon.

djmitche:
It's a URL for the third kind of expose, local.  When I added expose support, publicIP was still required.  We should probably add a new "null" expose that doesn't actually expose anything, and use that in the case where publicIP isn't set. And then reflect that into the livelog support by not uploading a livelog artifact at all.

djmitche:
I'm not sure this is necessary to solve now -- I think publicIP is still technically required, right?  When we change that, this functionality will also need to be changed.

petemoore:
> I'm not sure this is necessary to solve now -- I think publicIP is still technically required, right? 

Unfortunately we've already changed that in #2709 - it is now optional... :-/

petemoore:
I've fixed the merge conflict and pushed again. Not sure if the test will catch the problem this time, as we now run against mocks. So the work isn't done here yet, if the tests all pass, we'll need to add or modify an existing test to demonstrate the issue, and then fix the issue too.

petemoore:
> We should probably add a new "null" expose that doesn't actually expose anything, and use that in the case where publicIP isn't set.

@djmitche If websocktunnel doesn't require the worker to have a public IP, shouldn't that be used? Or do you mean if neither websocktunnel is configured nor a public IP?

djmitche:
> @djmitche If websocktunnel doesn't require the worker to have a public IP, shouldn't that be used? Or do you mean if neither websocktunnel nor a public IP is configured?

The latter is correct -- so the options would be don't expose anything, expose via public IP, or expose via websocktunnel.

petemoore:

PR #2868, opened 2020-05-20
Minor changes to generic-worker usage docs

petemoore:
Just a generic-worker usage docs fix up.


djmitche:
Oh, can you change the PR title just for historical

PR #2874, opened 2020-05-22
Do not throw error when taskVolumeBindings is null

helfi92:


djmitche:
oh, and changelog :)

helfi92:
> I wonder if we should look at the other Object.keys(..) for similar errors?

Yes, can be done in a separate PR. I can take care of it.

> oh, and changelog :)

Done.

changelog/WRN2apJpQVqjcU9cRkanBw.md
```diff
@@ -0,0 +1,4 @@
+audience: users
+level: patch
+---
+Fix docker worker not working in the latest release of Taskcluster
```
djmitche, line 4: Since users see the exception, it might be good to include some keywords here, too, especially `taskVolumeBindings`.

PR #2895, opened 2020-05-22
Stop loading the `azure` secret

djmitche:
The auth service still needs an account for its tests, but other
services do not, so moved this credential into tc-auth's secret.

I also deleted the pamplemousse storage account and re-created it, so auth's tests might fail.. Hopefully not.


djmitche:
All of the failures but changelog here are due to changing credentials.

PR #2907, opened 2020-05-26
Fix web-server relative path causing sign-ins to fail

djmitche:
I was able to replicate the bug in my dev environment, and will verify the fix while this runs in CI.

### DIFF
--- a/services/web-server/src/servers/createApp.js
+++ b/services/web-server/src/servers/createApp.js
@@ -1,4 +1,5 @@
 const bodyParser = require('body-parser');
+const path = require('path');
 const bodyParserGraphql = require('body-parser-graphql');
 const session = require('express-session');
 const compression = require('compression');
@@ -19,7 +20,7 @@ module.exports = async ({ cfg, strategies, AuthorizationCode, AccessToken, auth,
 
   app.set('trust proxy', cfg.server.trustProxy);
   app.set('view engine', 'ejs');
-  app.set('views', 'src/views');
+  app.set('views', path.resolve(path.join(__dirname, '../views')));
 
   const allowedCORSOrigins = cfg.server.allowedCORSOrigins.map(o => {
     if (typeof(o) === 'string' && o.startsWith('/')) {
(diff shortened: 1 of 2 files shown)


services/web-server/src/servers/createApp.js
```diff
@@ -19,7 +19,7 @@ module.exports = async ({ cfg, strategies, AuthorizationCode, AccessToken, auth,
 
   app.set('trust proxy', cfg.server.trustProxy);
   app.set('view engine', 'ejs');
-  app.set('views', 'src/views');
+  app.set('views', __dirname + '../views');
```
helfi92, line 22: Does `__dirname` end in a slash? To avoid errors like "path/to/directory../views", should we use `path.join`?

djmitche: Good catch.  Probably `path.resolve`, really.

djmitche:
That works in dev.  I think both worker failures are intermittents.

services/web-server/src/servers/createApp.js
```diff
@@ -19,7 +20,7 @@ module.exports = async ({ cfg, strategies, AuthorizationCode, AccessToken, auth,
 
   app.set('trust proxy', cfg.server.trustProxy);
   app.set('view engine', 'ejs');
-  app.set('views', 'src/views');
+  app.set('views', path.resolve(path.join(__dirname, '../views')));
```
helfi92, line 23: What's the difference in the output between

* `path.join(__dirname, '../views')`
* `path.resolve(path.join(__dirname, '../views'))`

I'm trying to see what is the benefit of adding path.resolve here other than just in case `../views` changes to an absolute path in the future.

djmitche: My understanding is that it changes `/app/services/web-server/src/servers/../views` to  `/app/services/web-server/src/views`, "resolving" the `..`.  I added it from a general sense that it's nicer to see paths without `..` in logs and things like that, but it's probably immaterial here.

helfi92: That makes a lot of sense. Thanks for elaborating.

PR #2909, opened 2020-05-26
Refactor fakes for AWS and Azure providers

### LINKED ISSUE #2827: Centralize and improve testing of providers
Let's build some testing infrastructure around worker-manager providers.

The rough plan is to rewrite the fakes to not use sinon and to assert that the resource-creation requests match a JSON schema, which we can hand-craft based on reading the docs (and only including fields we use).  And then build a framework to do a lot of small tests that run the provisioner with a given input and assert something about its output.  So I could just add a simple test that given a launchConfig with region X, the metadata has worker-group=X.

This should help catch cases like the recent issues with mis-structuring the label data, and also make it a lot easier to write one new test for each change to the provisioner implementation.

djmitche:
This also factors out some of the provisioning tests into a helper
function that reduces the repetitive setup.

The test for tags was broken (it used the spread operator in the
arguments to `assert.deepEqual` and because the first two elements of
the array it spread were equal, the assert passed, without actually
comparing to the set of tags given).

Github Bug/Issue: Fixes #2827


djmitche:
#2827 is about all three providers, but I've put this up for review after just AWS and Azure, in order to unblock @milescrabill.  The first comment on the PR addresses the AWS changes.  As for Azure, I made one (big) commit that refactors the tests but doesn't change the production code, then a few changes to the production code.  Those should get some special attention, since they may have unintended consequences just like fetchAndUpdateVm did.

Miles, I'm happy to rebase this on top of your work, if you'd like.

djmitche:
pulse issues are #2913.

djmitche:
> Should I assign myself a bug to move google provider into this world?

Let's file one and put it on the backlog..

djmitche:
filed #2921

djmitche:
(the change I just pushed removes the now-unused `fake-azure.js`)

services/worker-manager/src/providers/azure.js
```diff
@@ -621,6 +632,8 @@ class AzureProvider extends Provider {
       if (!worker.providerData.vm.id) {
         return worker.state;
       }
+      // XXX note that this doesn't actually change the state to RUNNING
+      // (that happens in registerWorker)
       return this.Worker.states.RUNNING;
```
milescrabill, line 637: Yikes, this comment is correct and we shouldn't be doing this here. We should probably outright return `worker.state` and refactor to check a `providerData` field that indicates the worker's current state, i.e. at this point we are in state "all resources are provisioned but not yet registered"

services/worker-manager/test/fakes/schemas/azure-vm.yml
```diff
@@ -0,0 +1,74 @@
...
+        type: string
+        pattern: '^[a-zA-Z0-9+/]+=*$' # base64
+      testProperty: {}
+    additionalProperties: false
+    required: [computerName, adminUsername, adminPassword, customData]
+  # https://docs.microsoft.com/en-us/javascript/api/@azure/arm-compute/storageprofile?view=azure-node-latest
+  storageProfile:
+    type: object
+    properties:
+      osDisk:
```
milescrabill, line 32: We'll add datadisks here for the multi disk deletion PR.

djmitche: Yep, and we can add more stuff as we add features for it.

services/worker-manager/src/providers/azure.js
djmitche: Yeah, that can happen in one of the followups I made a card for in https://github.com/taskcluster/taskcluster/projects/7.  The "XXX" is just to highlight that this is surprising :)

PR #2914, opened 2020-05-28
Put all log fields directly under Fields, instead of meta

djmitche:
This was bugging me while I was working on worker-manager, as half of the fields were "hidden" inside `meta`.


libraries/monitor/src/logger.js
```diff
@@ -139,7 +139,7 @@ class Logger {
     elideSecrets(fields);
 
     if (this.metadata) {
-      fields.meta = this.metadata;
+      Object.assign(fields, this.metadata, fields);
```
imbstack, line 142: doesn't this mean that we can't override metadata at a childMonitor level? Why the second `fields`?

djmitche: > doesn't this mean that we can't override metadata at a childMonitor level?

This is handled in the `childMonitor` function:
https://github.com/taskcluster/taskcluster/blob/d9ee68fdeb464e6f14eeea430c5a1dab4758f749/libraries/monitor/src/monitor.js#L93-L96

> Why the second `fields`?

The idea was to make fields override metadata, but it doesn't work, probably due to order of operations:
```
> let fields = {a: 10, b: 20};
undefined
> let metadata = {b: 300, c: 300};
undefined
> Object.assign(fields, metadata, fields);
{ a: 10, b: 300, c: 300 }
```
I'll just use `...`.

imbstack:
Oh, I think the metadata thing is mentioned in the readme too, that can probably be updated now

djmitche:
>  * `metadata` - an object of extra fields to attach to any logs generated by this object. If one field is `traceId`, it is bumped to the top-level of the log.                                                                                                                                                                                                                                 

Which is what it now does, so I think that's OK as-is.  But I added a bit about the `metadata` arg to `childMonitor`.  And fixed lint.  I'll merge when it's green (or as green as things get these days).

PR #2929, opened 2020-05-29
Prevent onConnected from occuring before a message

### LINKED ISSUE #2913: Intermittent tc-lib-pulse failures
https://community-tc.services.mozilla.com/tasks/frPWCpblTGSc5cdWawYkEA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FfrPWCpblTGSc5cdWawYkEA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log
```
  1) libraries/pulse/test/consumer_test.js (real)
       PulseConsumer
         consume messages ephemerally:
     Unknown assertation failure occured, assumed `[ 'msg', 'connected', 'msg', 'msg', 'msg', 'connected' ]` to deeply equal `[ 'connected', 'msg', 'msg', 'msg', 'msg', 'connected' ]`
  /taskcluster/libraries/pulse/test/consumer_test.js
                                           v
      255.       // four messages, reconnected, and then saw nothing.
      256.       assume(numbers).to.deeply.equal(['connected', 'msg', 'msg', 'msg', 'msg', 'connected']);
      257. 
                                           ^
```

djmitche:
This avoids a race condition where a consumer's onConnected callback
could be called *after* the first message is delivered.

This commit also fixes `consumer_test.js` to only start publishing
messages once connected.  This avoids the risk that the messages would
all be published before the consumer began; since this is an ephemeral
consumer, that would mean it receives no messages.

Github Bug/Issue: Fixes #2913

### DIFF
--- a/changelog/issue-2913.md
+++ b/changelog/issue-2913.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 2913
+---
--- a/libraries/pulse/src/consumer.js
+++ b/libraries/pulse/src/consumer.js
@@ -183,7 +183,10 @@ class PulseConsumer {
       // connection (better safe than sorry)
       channel.on('error', () => conn.failed());
 
-      const consumer = await channel.consume(queueName, async (msg) => {
+      // NOTE: channel.consume is not async!  In fact, await'ing it can
+      // result in a message arriving before the onConnected callback is
+      // invoked.
+      const consumer = channel.consume(queueName, async (msg) => {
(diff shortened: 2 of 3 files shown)


djmitche:
The amqplib API is truly misshapen, but I think we're pretty close.  Ultimately this wasn't a big deal -- I suspect in any production use of ephemeral queues that it wouldn't have mattered if handleMessage was called before onConnection.

PR #2953, opened 2020-06-01
Support new-credentials message in docker-worker

### LINKED ISSUE #2890: Add a new-creds message to the worker/runner protocol, use it in g-w and d-w
This is needed because temp creds last 30 days at most so we have to swap them out at reregister

djmitche:
This does some refactoring in docker-worker.  It doesn't touch generic-worker, because I want to wait until #2854 lands as that will conflict pretty badly.

Note that worker-runner doesn't send this message yet, as the reregisterWorker call doesn't exist yet.

Github Bug/Issue: refs #2890

### DIFF
--- a/changelog/bug-1640267.md
+++ b/changelog/bug-1640267.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: bug 1640267
+---
--- a/changelog/issue-2890.md
+++ b/changelog/issue-2890.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 2890
+---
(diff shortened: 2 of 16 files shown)


djmitche:
> Without knowing too much about the docker-worker codebase this looks good. It seems like there is a combination of queue changes (to be able to mock the client) and new-credentials protocol changes here.

The queue and taskqueue changes here are actually not to support mocking, but to support using the most up-to-date credentials.  The change in the superseding test is really just a refactoring of the existing mocking in that file.

> The new-credentials message actions seem simple enough, are protocol responses tested in the worker?

They aren't -- I don't have any effective way to write tests for docker-worker.  I can test it manually.

PR #2954, opened 2020-06-01
Bug 1607605 - Support graceful termination in generic-worker

djmitche:
This adds support for responding to the `graceful-termination` protocol
message.

### DIFF
--- a/workers/generic-worker/graceful/graceful.go
+++ b/workers/generic-worker/graceful/graceful.go
@@ -1,0 +1,56 @@
+package graceful
+
+import "sync"
+
+type GracefulTerminationFunc func(bool)
+
+var (
+	// Mutex for access to other vars
+	m sync.Mutex
+
+	// True if a graceful termination has been requestd
+	terminationRequested bool
+
+	// pending callback for graceful termination (there can be only one)
+	callback GracefulTerminationFunc
+)
+
+// Return true if graceful termination has been requested
+func TerminationRequested() bool {
+	m.Lock()
+	defer m.Unlock()
+
+	return terminationRequested
+}
+
+// Set up to call the given function (in a goroutine) when a termination
+// request is received.  Returns a function which, when called, will remove
+// the callback.  Only one callback can be installed at a time.
+func OnTerminationRequest(f GracefulTerminationFunc) func() {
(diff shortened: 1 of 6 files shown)


workers/generic-worker/graceful/graceful.go
```diff
@@ -0,0 +1,30 @@
+package graceful
```
petemoore, line 1: Let's add a package comment here to say this package isn't thread safe, and isn't intended to be. This can act as a warning for us if we start using it for other things. But since we only have a single calling go routine for now, this shouldn't be a problem.

workers/generic-worker/main.go
```diff
@@ -551,6 +553,11 @@ func RunWorker() (exitCode ExitCode) {
 				log.Printf("No task claimed. Idle for %v%v.%v", idleTime, remainingIdleTimeText, remainingTaskCountText)
 			}
 		}
+
+		if graceful.TerminationRequested {
```
petemoore, line 557: I've just realised, this probably isn't safe. This is the main go routine, that is reading `graceful.TerminationRequested`, which could be happening in parallel to the same variable being updated by the worker runner protocol go routine. So I think we'll need some synchronisation primitives around the updates to the exported variables of the package, or to signal events via channels, if we want to be a little more idiomatic (although that comes at the cost of a little extra complexity).

djmitche: Good point!  I added a mutex.

workers/generic-worker/graceful/graceful.go
```diff
@@ -0,0 +1,53 @@
...
+	// True if a graceful termination has been requestd
+	terminationRequested bool
+
+	// pending callback for graceful termination (there can be only one)
+	callback GracefulTerminationFunc
+)
+
+// Return true if graceful termination has been requested
+func TerminationRequested() bool {
+	return terminationRequested
```
petemoore, line 20: We'll need to lock the mutex for reads too. Note, you could use a [`sync.RWMutex`](https://golang.org/pkg/sync/#RWMutex) for marginal gain (reads won't block other reads), although the `sync.Mutex` is also fine.

Concurrency is tricky.

Issue #2951, opened 2020-06-01
[docker-worker] intermittent file_artifact_test.js timeout


djmitche:
https://community-tc.services.mozilla.com/tasks/cCpAzB_gTVumPN_J3hrwsA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FcCpAzB_gTVumPN_J3hrwsA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log
```
  1) workers/docker-worker/test/integration/file_artifact_test.js (real)
       upload retry:
     Error: Timeout of 200000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/tester/git_checkout/workers/docker-worker/test/integration/file_artifact_test.js)
      at listOnTimeout (internal/timers.js:549:17)
      at processTimers (internal/timers.js:492:7)
```

djmitche:
https://community-tc.services.mozilla.com/tasks/FEAh5K_8RwWePqRYsbMeTw/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FFEAh5K_8RwWePqRYsbMeTw%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log

djmitche:
This seems better?? weird.

PR #2960, opened 2020-06-02
#2951 - skip intermittent test 'file_artifact_test.js'

djmitche:
Github Bug/Issue: Refs #2951

### DIFF
--- a/workers/docker-worker/test/integration/file_artifact_test.js
+++ b/workers/docker-worker/test/integration/file_artifact_test.js
@@ -273,7 +273,8 @@ helper.secrets.mockSuite(suiteName(), ['docker', 'ci-creds'], function(mock, ski
     assert.equal(result.artifacts['public/my-missing.txt'].storageType, 'error');
   });
 
-  test('upload retry', async () => {
+  // intermittent https://github.com/taskcluster/taskcluster/issues/2951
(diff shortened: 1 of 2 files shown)


workers/docker-worker/test/integration/file_artifact_test.js
```diff
@@ -273,7 +273,8 @@ helper.secrets.mockSuite(suiteName(), ['docker', 'ci-creds'], function(mock, ski
     assert.equal(result.artifacts['public/my-missing.txt'].storageType, 'error');
   });
 
-  test('upload retry', async () => {
+  // intermittent https://github.com/taskcluster/taskcluster/issues/2951
+  test.skip('upload retry', async () => {
```
petemoore, line 277: It isn't clear to me if the test is faulty, or it has found a genuine issue. Looking at the test, I can imagine it could be intermittently blocking itself through some side effect of the synchronisation mechanics, but I don't have expertise in synchronisation in node.js, so I'm not well placed to make a judgement.

If it isn't immediately obvious what the problem is, I suspect it is reasonable to disable the test. Although it doesn't give me a good feeling in general, especially if it is going to be disabled for a while. However, I also suspect the retry mechanics of uploads don't see a lot of maintenance, so there is a good chance it has always been a flaky test.

PR #2964, opened 2020-06-02
remove Procfiles

djmitche:
The only thing still running in Heroku is docs, which runs with PROCFILE=ui/Procfile.  That file doesn't exist, but it seems to be fine :shrug: 

So I think this won't break docs, but we can find out when it lands.


helfi92:
I think the docs site added https://github.com/taskcluster/heroku-buildpack-multi-procfile to figure out which command to run the docs site.

djmitche:
Oh! I can actually deploy from this branch to verify..

djmitche:
Yup, works fine.

PR #2965, opened 2020-06-02
refactored __init__.py and basestring changed to six.text_type

### LINKED ISSUE #2924: Clean up decision task hacks
Things noted by @tomprince:

1. `TransformSequence` in `taskcluster/src/transforms/__init__.py` is not necessary
2. We should also change `basestring` to `six.text_type` wherever used
3. Consider using level-1 vs level-3 to keep decision task stuff on its own workers instead of a `decision` worker.

090max:
Github Bug/Issue: Fixes #2924


imbstack:
I believe this is failing because you need to update the path to the code in all of the `taskcluster/ci` files. They all reference `src` when you want `taskgraph_taskcluster` now. In fact, it will be easier if we just do the `six` types in this PR and make another one for the code moving. The `six` parts look correct here! Thanks again.

imbstack:
Closing this for now but feel free to come back to it again and reopen it, @090max!

PR #2966, opened 2020-06-02
[wip] bug 1639279: add UI for showing dependent tasks

### LINKED ISSUE #2910: Display dependent tasks in the tasks view
https://bugzilla.mozilla.org/show_bug.cgi?id=1639279

**Describe the bug**
When viewing a task we display the list of dependencies but not the tasks that depend on the task we're currently viewing.

Having a section in the task view that displays the tasks that depend on it can give us a better idea about when a given task would run or what other tasks will be affected if it fails or if you cancel it. A better understanding of the relationship between tasks would help in the CI cost cutting work.

The implementation will be similar to the way we display the list of dependencies. For example:

Screen Shot 2020-05-27 at 10 09 51 AM

**Taskcluster version**
29.6.0

**Context**
An endpoint that talks to /queue/v1/task/<task-id>/dependents might need to be added to the web-server service first before being able to do the UI additions.

Unlike the list of dependencies, the /dependents endpoint is paginated so we would need to properly handle that in the UI.

mozbhearsum:
Bugzilla Bug: [1639279](https://bugzilla.mozilla.org/show_bug.cgi?id=1639279)

Github Bug/Issue: Fixes #2910

This doesn't quite fix the above bug/issue, but it's close. I'm struggling to get the graphql bits working correctly -- I'm currently getting this error: ```"Cannot query field \"tasks\" on type \"TasksConnection\"."```.

### DIFF
--- a/ui/src/components/TaskDetailsCard/index.jsx
+++ b/ui/src/components/TaskDetailsCard/index.jsx
@@ -22,7 +22,6 @@ import ConnectionDataTable from '../ConnectionDataTable';
 import CopyToClipboardListItem from '../CopyToClipboardListItem';
 import DateDistance from '../DateDistance';
 import StatusLabel from '../StatusLabel';
-import TableCellItem from '../TableCellItem';
 import { DEPENDENTS_PAGE_SIZE } from '../../utils/constants';
 import { pageInfo, task } from '../../utils/prop-types';
 import urls from '../../utils/urls';
@@ -69,6 +68,26 @@ import Link from '../../utils/Link';
   },
   payload: {
     whiteSpace: 'break-spaces',
+  },
+  dependentsStatusAndNameContainer: {
+    display: 'flex',
+  },
+  dependentsStatus: {
+    marginRight: theme.spacing(2),
+  },
+  dependentsName: {
+    display: 'inline-flex',
+    flexBasis: '50%',
+    flexGrow: 1,
+  },
+  dependentsTableRow: {
+    cursor: 'pointer',
+  },
+  dependentsLink: {
+    textDecoration: 'none',
+    display: 'flex',
+    justifyContent: 'space-between',
+    verticalAlign: 'middle',
   },
 }))
 /**
@@ -307,6 +326,7 @@ export default class TaskDetailsCard extends Component {
                             title="View Task">
                             <StatusLabel state={task.status.state} />
                             <ListItemText
+                              primaryTypographyProps={{ variant: 'body2' }}
                               className={classes.listItemText}
                               primary={task.metadata.name}
                             />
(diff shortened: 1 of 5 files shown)


ui/src/views/Tasks/ViewTask/task.graphql
```diff
@@ -68,4 +68,16 @@ query Task($taskId: ID!, $artifactsConnection: PageConnection, $taskActionsFilte
...
+
+  dependents(taskId: $taskId, connection: $dependentsConnection, filter: $dependentsFilter) {
+    taskId
+    tasks {
+      status {
+        state
+      }
+      task {
+        task
+      }
```
helfi92, line 80: To fix the issue you are seeing, we'll want to specify the fields that we want the backend to return to us (e.g., task name, and state). Also, since this is a paginated endpoint, we'll want to receive pagination metadata in order to be able to fetch next/previous pages. Maybe something like the following will work:


```graphql
dependents(taskId: $taskId, connection: $dependentsConnection, filter: $dependentsFilter) {
  pageInfo {
    hasNextPage
    hasPreviousPage
    cursor
    previousCursor
    nextCursor
  }

  edges {
    node {
      taskId
      status {
        state
      }
      metadata {
        name
      }
    }
  }
}
```

ui/src/views/Tasks/ViewTask/index.jsx
```diff
@@ -107,6 +107,9 @@ const getCachesFromTask = task =>
       artifactsConnection: {
         limit: ARTIFACTS_PAGE_SIZE,
       },
+      dependentsConnection: {
+        limit: 1000,
+      },
```
helfi92, line 112: In order to keep the task view minimal, let's show a subset of the dependents at a time (e.g., 5 at a time) with the possibility of changing pages. The `ConnectionDataTable` component might be helpful to use. It renders a paginated table.

ui/src/components/TaskDetailsCard/index.jsx
```diff
@@ -93,6 +94,19 @@ export default class TaskDetailsCard extends Component {
...
+        tasks: arrayOf(
+          shape({
+            status: shape({
+              state: string,
+            }),
+            task,
+          })
+        ),
+      })
+    ),
```
helfi92, line 109: `utils/prop-types` exports a bunch of props. We can probably just do `dependents: arrayOf(task)` where task is exported from that file.

mozbhearsum:
This is mostly working/done now. The only thing that I'm aware of that's broken is page switching. I don't have time to debug it at the moment, but I'll try to follow up in the next couple of weeks.

ahal:
*gentle nudge* :)

mozbhearsum:
I think this is ready to go now -- it turns out the pagination issue was a wrongly named parameter in `handleDependentsPageChange`. I believe this is also an issue in `handleArtifactsPageChange` (which I copy/pasted as a starting point), but is rarely or never hit because the page size is so high there. Regardless, I fixed it.

helfi92:
This will be released in `v36.0.1` (FirefoxCI is currently running v34.0.1)

PR #2967, opened 2020-06-02
 [Bug 1634376] add 404 error test for the listWorkersForWorkerPool api

aditya2499:


services/worker-manager/test/api_test.js
```diff
@@ -501,7 +516,7 @@ helper.secrets.mockSuite(testing.suiteName(), ['db'], function(mock, skipping) {
   });
 
   test('get many workers for a given worker pool', async function () {
-    const workerPoolId = 'apple/apple';
+    //const workerPoolId = 'apple/apple';
```
djmitche, line 519: Please remove this line and other's you've commented out

services/worker-manager/src/api.js
```diff
@@ -383,7 +383,7 @@ builder.declare({
     limit: parseInt(req.query.limit || 100, 10),
     matchPartition: 'exact',
   };
-
+
```
djmitche, line 386: Extra whitespace (and at one point below).  I expect the lint check will complain about this.  You can run lint locally with `yarn lint`.

services/worker-manager/src/api.js
```diff
@@ -565,6 +565,13 @@ builder.declare({
     limit: parseInt(req.query.limit || 100, 10),
     matchPartition: 'exact',
   };
+  const {workerPoolId} = req.params;
+  const workerPool = await WorkerPool.get(this.db,workerPoolId);
+
+  if(!workerPool){
```
djmitche, line 571: Space between `){`

services/worker-manager/test/api_test.js
```diff
@@ -470,9 +470,22 @@ helper.secrets.mockSuite(testing.suiteName(), ['db'], function(mock, skipping) {
...
+       try{
+        await helper.workerManager.listWorkersForWorkerPool(workerPoolId);
+       }  catch(err){
+            if(err.code !== 'ResourceNotFound'){
+               throw err;
+            }
+            return;
+          }
+        throw new Error('get of workers from non-existing pool suceeded');   
+  });
```
djmitche, line 485: I think this can be more neatly with `assert.rejects`, of which there are a few exampls in this file.

services/worker-manager/test/api_test.js
```diff
@@ -548,8 +565,8 @@ helper.secrets.mockSuite(testing.suiteName(), ['db'], function(mock, skipping) {
   });
 
   test('get workers for a given worker pool - no workers', async function () {
-    const workerPoolId = 'r/r';
-
+    //const workerPoolId = 'r/r';
+    await createWorkerPool({providerId : 'google'});
```
djmitche, line 569: Does this not succeed if you just pass `{}` without a providerId?

aditya2499: Yes.
I have made all the necessary changes.

djmitche:
Nice work!!

djmitche:
By the way, it looks like the name/email you're using to make git commits doesn't correspond to your GitHub user, so GitHub isn't able to tell that you made the commits.

```
From: = <=>
```

You can fix that with 
```
$ git config --global user.name "Some Name"
$ git config --global user.email someemail@example.com
```

aditya2499:
I have fixed it now, Do check.

djmitche:
Great!  It will only show up the next time you make a commit and push it to GitHub, so I can't check until then (but I'm looking forward to that!)

PR #2968, opened 2020-06-02
Bug 1632929 - use target_commitish to get SHA for releases

djmitche:
Bugzilla Bug: [1632929](https://bugzilla.mozilla.org/show_bug.cgi?id=1632929)


djmitche:
Looks like there's a race condition in `yarn generate` :(

PR #2971, opened 2020-06-03
docker-worker: only become idle when claimTasks returns nothing

### LINKED ISSUE #2969: docker-worker will terminate if its idle timeout is shorter than the time it takes to call claimWork
[Setting the `afterIdleSeconds` value to 1](https://github.com/mozilla/community-tc-config/commit/2e8f6fa3a711b66ea610e06f889aa1a1ba8428fe#diff-4083ba8a0fdfa215bc0f2edbab4b7b3dR424) seems to result in workers terminating during the call to `claimWork`.

Generic-worker does not have this issue.

djmitche:
Github Bug/Issue: Fixes #2969

### DIFF
--- a/changelog/issue-2969.md
+++ b/changelog/issue-2969.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 2969
+---
+Docker-worker now only considers itself idle if its call to `queue.claimWork` returns no tasks.  This prevents the situation where a very short `afterIdleSeconds` causes the worker to shut down *while* calling `claimWork`.
--- a/workers/docker-worker/src/lib/task_listener.js
+++ b/workers/docker-worker/src/lib/task_listener.js
@@ -141,9 +141,18 @@ class TaskListener extends EventEmitter {
 
     let claims = await this.taskQueue.claimWork(availableCapacity);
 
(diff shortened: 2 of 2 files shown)


djmitche:
I don't know if we have coverage.  I don't know how to find out.  Do you?

djmitche:
I don't see any tests with the name "idle", so I"m guessing not.

This isn't something that can be unit tested without refactoring docker-worker, which would be likely to introduce more bugs.  I think the best we can do here is to test this by hand, unless you know a way to write a test for this.

djmitche:
Thinking on this more, I think the best use of time is to just not solve this bug.  It only occurs when afterIdleSeconds is too short, and we've already stopped doing that.

Changing docker-worker is too hard without decent tests, and writing decent tests is too hard without changing docker-worker :(

djmitche:
Oh hey, there *are* tests for this!  And they passed!
https://github.com/taskcluster/taskcluster/blob/77ab4d7b2328d02eb4f6cfee13e9e403e187868a/workers/docker-worker/test/integration/shutdown_test.js#L45-L106

djmitche:
> Let's double check we've got coverage for this scenario, other than that, it looks good to land!

✅

PR #2979, opened 2020-06-04
correctly include metadata in fields

djmitche:
I noticed this while working on the google provider tests.  Apparently I failed to notice this when I tested 2f57cefb4dc698f680041d4c70c4eb5a3ee5e85d manually!  That commit hasn't been released yet, so this just fixes the fix.


djmitche:
I'll probably need 2-3 more patches to this line to get this right, but we'll see :(

PR #2980, opened 2020-06-04
Bug 1639713: /dev/shm should be a bind mount, not a device mount

jschwartzentruber:
Bugzilla Bug: [1639713](https://bugzilla.mozilla.org/show_bug.cgi?id=1639713)

Capability devices normally use the equivalent of `--device` for mounting devices from the host into the container. `/dev/shm` is different since it is a tmpfs volume, and it should be mounted using a bind mount equivalent to `--volume`.

Note that a test already exists for this, but I'm not sure if those are run, since it should have exhibited the same error reported in bugzilla.

### DIFF
--- a/workers/docker-worker/src/lib/task.js
+++ b/workers/docker-worker/src/lib/task.js
@@ -138,6 +138,7 @@ async function buildDeviceBindings(devices, expandedScopes) {
   }
 
   let deviceBindings = [];
+  let bindMounts = [];
   for (let deviceType of Object.keys(devices || {})) {
     let device = devices[deviceType];
     device.mountPoints.forEach((mountPoint) => {
@@ -149,9 +150,12 @@ async function buildDeviceBindings(devices, expandedScopes) {
         },
       );
     });
+    if (device.binds) {
+      bindMounts = _.union(bindMounts, device.binds);
+    }
   }
 
-  return deviceBindings;
+  return [deviceBindings, bindMounts];
 }
 
 class Reclaimer {
@@ -365,11 +369,6 @@ class Task extends EventEmitter {
     });
     let expandedScopes = (await auth.expandScopes({scopes: this.task.scopes})).scopes;
 
-    if (this.options.devices) {
-      let bindings = await buildDeviceBindings(this.options.devices, expandedScopes);
-      procConfig.create.HostConfig['Devices'] = bindings;
-    }
-
     if (linkInfo.links) {
       procConfig.create.HostConfig.Links = linkInfo.links.map(link => {
         return link.name + ':' + link.alias;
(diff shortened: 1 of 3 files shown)


workers/docker-worker/src/lib/task.js
```diff
@@ -149,9 +150,12 @@ async function buildDeviceBindings(devices, expandedScopes) {
         },
       );
     });
+    if (device.binds) {
+      bindMounts = _.union(bindMounts, device.binds);
+    }
   }
 
-  return deviceBindings;
+  return [deviceBindings, bindMounts];
```
djmitche, line 158: Let's use a container object instead of a tuple, for clarity in `dockerConfig`

```suggestion
  return {deviceBindings, bindMounts};
```

workers/docker-worker/src/lib/task.js
```diff
@@ -385,6 +384,12 @@ class Task extends EventEmitter {
       return binding;
     });
 
+    if (this.options.devices) {
+      let bindings = await buildDeviceBindings(this.options.devices, expandedScopes);
+      procConfig.create.HostConfig['Devices'] = bindings[0];
+      binds = _.union(binds, bindings[1]);
```
djmitche, line 390:
```suggestion
      procConfig.create.HostConfig['Devices'] = bindings.deviceBindings;
      binds = _.union(binds, bindings.bindMounts);
```

workers/docker-worker/src/lib/task.js
jschwartzentruber: Thanks! I was not aware that you could omit the property name like this.

jschwartzentruber:
> When this is done we can build an image and run it on a test workerType, and then you can run a test task against it.

@djmitche how do I go about this?

It seems to be working now on some pools, but not others:
- working: https://community-tc.services.mozilla.com/tasks/RwvHwzFTSHWoG6qP1CRCgQ
- not working: https://community-tc.services.mozilla.com/tasks/YW8KmVhCTSq7h3lVT3cEyg

djmitche:
It's surprising that it's working, as those are both running a docker-worker version before this change landed (30.0.2, seen in the top of the task log).  This was released in 30.1.1.

I can get a GCP image built and shared with you to install.

djmitche:
`docker-worker-gcp-community-googlecompute-2020-06-12t17-25-46z`

jschwartzentruber:
Looks good, this task group is using that image: https://community-tc.services.mozilla.com/tasks/groups/DzN0Tb7lQciLC9s4Y8GuIw

djmitche:
Awesome!  We'll ship it everywhere in https://github.com/mozilla/community-tc-config/pull/293

PR #2985, opened 2020-06-05
Refactor Google provider tests

### LINKED ISSUE #2885: google provider might be hiding user errors?
Reported by @djmitche

Once handleOperation is called for an instance, the instance state is set to STOPPED, which means that checkWorker won't be called again. So this comment

    // Let's check back in on the next provisioning iteration if unfinished
    if (operation.status !== 'DONE') {
      return true;
    }

is incorrect. However, handleOperation is only called when the instance is not found, so perhaps the status is never not DONE?

I'm not sure if the comment should be updated, or if there's an error here that means we will be missing error messages that might be useful to the user.

### LINKED ISSUE #2921: Refactor Google provider tests to use newer fakes
A follow-on to #2827, which addressed AWS and Azure.

djmitche:
Github Bug/Issue: Fixes #2921

This fixes #2885 in the process.


djmitche:
There are two refactoring commits here -- one refactors just the fakes, and then the second refactors the tests.

services/worker-manager/src/providers/google.js
```diff
@@ -413,11 +404,15 @@ class GoogleProvider extends Provider {
...
         // We only check in on the operation if the worker failed to
         // start succesfully
-        await this.handleOperation({
+        if (await this.handleOperation({
           op: worker.providerData.operation,
           errors: this.errors[worker.workerPoolId],
           monitor,
-        });
+        })) {
+          monitor.debug('operation still running');
```
imbstack, line 412: I wonder if google can have operations that never complete? Maybe we set a max-retries to check in before we just mark it as failed?

services/worker-manager/test/fakes/schemas/google-instance.yml
```diff
@@ -0,0 +1,89 @@
+# This is a partial schema for google.compute.instances.insert.  Feel free to add additional
+# properties here if necessary, consulting with the relevant bits of Azure documentation.
```
imbstack, line 2: Google documentation

services/worker-manager/test/fakes/schemas/google-instance.yml
```diff
@@ -0,0 +1,89 @@
...
+# additional configuration from the user is correctly passed to the endpoint.
+#
+# https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert
+type: object
+definitions:
+  # https://cloud.google.com/resource-manager/docs/creating-managing-labels
+  labels:
+    type: object
+    propertyNames:
+      pattern: '^[a-z0-9_-]*$'
```
imbstack, line 14: Is this the thing with the confusing set of allowed chars we talked about yesterday?

So this is the things they accept but without the things that we don't allow? (non-ascii, etc)

services/worker-manager/src/providers/google.js
djmitche: https://cloud.google.com/compute/docs/api/how-tos/api-requests-responses#%20%20%20%20%20%20%20%20%20%20handling_api_responses

> Your request is not complete until the Operation resource's status returns as DONE. This can take some time depending on the nature of your request. Then, after the operation's status returns as DONE, you can check to see if the operation succeeded and whether there were any errors.

That suggests strongly that it will eventually be DONE, so I'm inclined to assume that's true and if we find out otherwise we can figure out if it's a Google bug or a provider bug.  Since they suggest waiting for multiples of 2 minutes with WAIT, I'm not even sure what we'd set as a max wait time.

services/worker-manager/test/fakes/schemas/google-instance.yml
djmitche: Yeah, I figure since we're not injecting any unicode in the tests, this is enough.  It will catch anything where we accidentally include a `/` for example.

PR #2993, opened 2020-06-06
Allow deprecated functions

imbstack:
As part of working on the existing-capacity stuff I found that deprecating db functions didn't actually work. This splits those changes out into a PR by itself.

### DIFF
--- a/libraries/postgres/src/Database.js
+++ b/libraries/postgres/src/Database.js
@@ -396,7 +396,11 @@ class Database {
           await client.query(`DO ${dollarQuote(migrationScript)}`);
         }
         showProgress('..defining methods');
-        for (let [methodName, { args, body, returns}] of Object.entries(version.methods)) {
+        for (let [methodName, {deprecated, args, body, returns}] of Object.entries(version.methods)) {
+          // We don't need to do anything for deprecated methods
+          if (deprecated) {
+            continue;
+          }
           await client.query(`create or replace function
           "${methodName}"(${args})
           returns ${returns}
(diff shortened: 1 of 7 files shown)


libraries/postgres/src/Database.js
```diff
@@ -396,7 +397,11 @@ class Database {
           await client.query(`DO ${dollarQuote(migrationScript)}`);
         }
         showProgress('..defining methods');
-        for (let [methodName, { args, body, returns}] of Object.entries(version.methods)) {
+        for (let [methodName, {deprecated, args, body, returns}] of Object.entries(version.methods)) {
+          // We don't need to do anything for deprecated methods
```
djmitche, line 401: I think in most cases we would want to update deprecated methods -- the method should be deprecated, but old code will still call it so it will also need an updated implementation that continues to function when the DB is upgraded.

In the "two step" process of upgrading from entities tables, it happens we don't want to do this -- the firs step updates the implementation, and the second step deprecates the old.  But in general I don't think this will be the case.

libraries/postgres/src/Schema.js
```diff
@@ -113,7 +113,14 @@ class Schema {
   allMethods() {
     const map = this.versions.reduce(
       (acc, version) => {
-        Object.entries(version.methods).forEach(([name, method]) => acc.set(name, method));
+        Object.entries(version.methods).forEach(([name, method]) => {
+          if (method.deprecated) {
+            const content = acc.get(name);
+            assert(content, 'Cannot deprecate new functions');
```
djmitche, line 119: (similarly here)

djmitche:
Also, once this lands someone can go back and deprecate the wmworkers, cach_purges, and wmwoker_pools entities methods (and add a note to do so to the migration guide). I'll be happy to do this..

PR #3006, opened 2020-06-08
Check generic-worker revision number at publish

djmitche:
I noticed that `build.sh` was using `github.com/taskcluster/taskcluster/v30/workers/generic-worker.revision` with `-ldflags -X` for running tests, but `main.revision` when building.  The built artifacts for 30.1.0 have the correct revision, so it seems either path works, but it highlighted that the test is only checking that the invocation used when running tests, and not the invocation used to build release binaries.

So, this moves the test to verify that the revision is included in the release binaries (in fact, just the linux/amd64 binary, but they are all built with the same invocation, so that's enough).


workers/generic-worker/build.sh
```diff
@@ -60,13 +60,25 @@ fi
...
+  if $PUBLISH; then
+      if [ "${2}" = "linux" -a "${3}" = "amd64" ]; then
+          vers=$("$OUTPUT_DIR/generic-worker-${1}-${2}-${3}" --version \
+              | grep -E 'revision: https://github.com/taskcluster/taskcluster/commits/[a-z0-9]{40}')
+          if [ -z "$vers" ]; then
+              echo "The --version option does not output a proper revision link"
+              exit 1
+          fi
+      fi
+  fi
```
petemoore, line 81: Since a developer can break this at any time, we should find out immediately, not only when a release is made, so let's lose this outer `if` statement. Note, if you had done this, you would have noticed that you had broken it yourself. 😉 

This version is a little bit simpler, and would catch the issue as soon as it breaks:

```suggestion
  if [ "${2}" == "$(go env GOOS)" ] && [ "${3}" == "$(go env GOARCH)" ]; then
    if ! "$OUTPUT_DIR/generic-worker-${1}-${2}-${3}" --version | \
    grep -Eq 'revision: https://github.com/taskcluster/taskcluster/commits/[a-z0-9]{40}'; then
      echo "The --version option does not output a proper revision link"
      exit 1
    fi
  fi
```

workers/generic-worker/build.sh
```diff
@@ -107,10 +119,10 @@ CGO_ENABLED=0 go get github.com/taskcluster/livelog
...
+  CGO_ENABLED=1 GORACE="history_size=7" /usr/bin/sudo "GOPATH=$GOPATH" "GW_TESTS_RUN_AS_CURRENT_USER=" "TASKCLUSTER_CERTIFICATE=$TASKCLUSTER_CERTIFICATE" "TASKCLUSTER_ACCESS_TOKEN=$TASKCLUSTER_ACCESS_TOKEN" "TASKCLUSTER_CLIENT_ID=$TASKCLUSTER_CLIENT_ID" "TASKCLUSTER_ROOT_URL=$TASKCLUSTER_ROOT_URL" $(which go) test -v -tags multiuser -race -timeout 1h ./...
   MYGOHOSTOS="$(go env GOHOSTOS)"
   if [ "${MYGOHOSTOS}" == "linux" ] || [ "${MYGOHOSTOS}" == "darwin" ]; then
-    CGO_ENABLED=1 GORACE="history_size=7" go test -v -tags docker -ldflags "-X github.com/taskcluster/taskcluster/v30/workers/generic-worker.revision=$(git rev-parse HEAD)" -race -timeout 1h ./...
+    CGO_ENABLED=1 GORACE="history_size=7" go test -v -tags docker -race -timeout 1h ./...
```
petemoore, line 125: I don't see any reason to remove the version number from the test binary - in fact it means the code that is being tested is less like production code. Why would we intentionally make the test binary behave differently to the build binaries? Any tests that indirectly require the revision number to be set (even if we don't explicitly test for the revision number) would operate incorrectly.

workers/generic-worker/main_test.go
```diff
@@ -41,15 +41,6 @@ func TestIdleWithoutCrash(t *testing.T) {
...
 
-func TestRevisionNumberStored(t *testing.T) {
-	if !regexp.MustCompile("^[0-9a-f]{40}$").MatchString(revision) {
-		t.Fatalf("Git revision could not be determined - got '%v' but expected to match regular expression '^[0-9a-f](40)$'\n"+
-			"Did you specify `-ldflags \"-X github.com/taskcluster/generic-worker.revision=<GIT REVISION>\"` in your go test command?\n"+
-			"Try using build.sh / build.cmd in root directory of generic-worker source code repository.", revision)
-	}
-	t.Logf("Git revision successfully retrieved: %v", revision)
-}
-
```
petemoore, line 52: Yes, this test tests that the test binary was built correctly, not the release binary - but that is important for knowing that the test environment accurately reflects a release binary. In other words, this test is not useful for validating that the release has the revision number (the addition to `build.sh` to do this was a great idea to solve that problem, many thanks) - but it is still useful for ensuring the test binary works like the release binary. So please keep it.

petemoore:
> I noticed that `build.sh` was using `github.com/taskcluster/taskcluster/v30/workers/generic-worker.revision` with `-ldflags -X` for running tests, but `main.revision` when building.

That is expected and correct behaviour: test binary symbols for members of the `main` package are relocated by the go build system, unlike in non-test binaries.

> The built artifacts for 30.1.0 have the correct revision, so it seems either path works, but it highlighted that the test is only checking that the invocation used when running tests, and not the invocation used to build release binaries.

No, this is wrong. Only a fully qualified `main` path works for test binaries, and only `main` works for non-test binaries (a quick test demonstrates this).

> So, this moves the test to verify that the revision is included in the release binaries (in fact, just the linux/amd64 binary, but they are all built with the same invocation, so that's enough).

An additional test against release binaries is a good suggestion, but deleting the test against the test binary is a bad idea. It is still useful to ensure that the test binary and release binary act consistently. Also there is no reason to _only_ run the test on release, we can run it on build to detect if it breaks during development.

workers/generic-worker/build.sh
```diff
@@ -107,10 +119,10 @@ CGO_ENABLED=0 go get github.com/taskcluster/livelog
 
 if $TEST; then
   go get github.com/taskcluster/taskcluster/v30/tools/taskcluster-proxy
-  CGO_ENABLED=1 GORACE="history_size=7" /usr/bin/sudo "GOPATH=$GOPATH" "GW_TESTS_RUN_AS_CURRENT_USER=" "TASKCLUSTER_CERTIFICATE=$TASKCLUSTER_CERTIFICATE" "TASKCLUSTER_ACCESS_TOKEN=$TASKCLUSTER_ACCESS_TOKEN" "TASKCLUSTER_CLIENT_ID=$TASKCLUSTER_CLIENT_ID" "TASKCLUSTER_ROOT_URL=$TASKCLUSTER_ROOT_URL" $(which go) test -v -tags multiuser -ldflags "-X github.com/taskcluster/taskcluster/v30/workers/generic-worker.revision=$(git rev-parse HEAD)" -race -timeout 1h ./...
```
petemoore, line 110: See comment below - also no reason to remove the revision number here.

PR #3008, opened 2020-06-08
[worker-manager] Avoid loading all workers in memory

helfi92:


helfi92:
Agreed. I pushed a patch to strictly use `Worker.getWorkers`. Thanks for the review.

helfi92:
I probably forgot to fix the failing tests. One moment :)

helfi92:
Done!

PR #3009, opened 2020-06-08
Add support for the 'new-credentials' message in generic-worker

### LINKED ISSUE #2890: Add a new-creds message to the worker/runner protocol, use it in g-w and d-w

djmitche:
This comes with tests for graceful-termination, too.  /cc @milescrabill 

Github Bug/Issue: Fixes #2890


djmitche:
I didn't see where generic-worker was continually reloading its configuration -- is there a way to disable that in the `--with-worker-runner` case?

Issue #2887, opened 2020-05-22
[**report-error**] support reporting errors from workers to worker-manager (requires **g-w-protocol**  and **protocol-startup**)


imbstack:
Discussion of what to do with this is  in https://bugzilla.mozilla.org/show_bug.cgi?id=1562285

imbstack:
(requires **g-w-protocol**  and **protocol-startup**)

djmitche:
I believe this is no longer blocked, as generic-worker now uses the protocol and it is initialized properly at startup.

milescrabill:
The outstanding bit here is the docker-worker portion.

lotas:
Looks like worker-runner is already doing this, g-w not yet? @petemoore do you think it should be?

PR #3010, opened 2020-06-08
issue 2887: support error-report message in worker-runner

milescrabill:
working towards #2887, adds support to worker-runner for a new `error-report` message, adds it to `workerproto` readme

### DIFF
--- a/changelog/issue-2887.md
+++ b/changelog/issue-2887.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 2887
+---
--- a/internal/workerproto/README.md
+++ b/internal/workerproto/README.md
@@ -59,6 +59,8 @@ This message type, sent from the worker, contains an error report intended for t
 ~{"type": "error-report", "kind": "critical", "title": "a serious error", "description": "bad stuff has happened!", "extra": {"number of failures": 1}}
 ```
 
+See [here](https://docs.taskcluster.net/docs/reference/core/worker-manager/api#reportWorkerError) for documentation on worker-manager's `reportWorkerError` API.
+
 ### shutdown
 
 When worker-runner receives this message, and the provider supports it, it will invoke the function `RemoveWorker` from
--- a/internal/workerproto/README.md
+++ b/internal/workerproto/README.md
@@ -51,6 +51,14 @@ Before the connection is initialized, the connection's capabilities are unknown,
 
 The following sections describe the defined message types, each under a heading giving the corresponding capability.
 
+### error-report
+
+This message type, sent from the worker, contains an error report intended for the worker-manager to display among its worker pool errors.
+
+```
+~{"type": "error-report", "kind": "critical", "title": "a serious error", "description": "bad stuff has happened!", "extra": {"number of failures": 1}}
+```
+
 ### shutdown
 
 When worker-runner receives this message, and the provider supports it, it will invoke the function `RemoveWorker` from
--- a/tools/worker-runner/errorreport/errorreport.go
+++ b/tools/worker-runner/errorreport/errorreport.go
@@ -1,0 +1,56 @@
+package errorreport
+
+import (
+	"encoding/json"
+	"log"
+	"reflect"
+
+	"github.com/taskcluster/taskcluster/v30/clients/client-go/tcworkermanager"
+	"github.com/taskcluster/taskcluster/v30/internal/workerproto"
+	"github.com/taskcluster/taskcluster/v30/tools/worker-runner/provider/provider"
+	"github.com/taskcluster/taskcluster/v30/tools/worker-runner/run"
+	"github.com/taskcluster/taskcluster/v30/tools/worker-runner/tc"
+)
+
+func HandleMessage(msg workerproto.Message, factory tc.WorkerManagerClientFactory, state *run.State) {
+	validate := func(things map[string]interface{}, key, expectedType string) bool {
+		if _, ok := things[key]; !ok {
+			return false
+		}
+		if realType := reflect.TypeOf(things[key]); expectedType != realType.String() {
+			log.Printf("got %s with type %#v, expected %s", key, realType.String(), expectedType)
+			return false
+		}
+		return true
+	}
+
+	if !validate(msg.Properties, "description", "string") {
+		log.Printf("Error processing error-report message, missing description or not string")
+	}
+	if !validate(msg.Properties, "kind", "string") {
+		log.Printf("Error processing error-report message, missing kind or not string")
+	}
+	if !validate(msg.Properties, "title", "string") {
+		log.Printf("Error processing error-report message, missing title or not string")
+	}
+	if !validate(msg.Properties, "extra", "map[string]interface {}") {
+		log.Printf("Error processing error-report message, missing extra or not map[string]interface{}")
+	}
+
+	extra := msg.Properties["extra"].(map[string]interface{})
+	extraMsg, err := json.Marshal(extra)
+	if err != nil {
+		log.Printf("Error processing error-report message, could not marshal extra")
+	}
+
+	errorReport := tcworkermanager.WorkerErrorReport{
+		Description: msg.Properties["description"].(string),
+		Kind:        msg.Properties["kind"].(string),
+		Extra:       extraMsg,
+		Title:       msg.Properties["title"].(string),
+	}
+	err = provider.ReportWorkerError(state, factory, &errorReport)
+	if err != nil {
+		log.Printf("Error reporting worker error: %v\n", err)
+	}
+}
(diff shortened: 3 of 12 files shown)


tools/worker-runner/tc/fakeworkermanager.go
```diff
@@ -50,13 +51,39 @@ func (wm *FakeWorkerManager) RemoveWorker(workerPoolID, workerGroup, workerID st
...
+
 // Get the single registration that has occurred, or an error if there are not
 // exactly one.  This resets the list of registrations in the process.
 func FakeWorkerManagerRegistration() (*tcworkermanager.RegisterWorkerRequest, error) {
 	if len(wmRegistrations) == 0 {
 		return nil, fmt.Errorf("No registerWorker calls")
 	} else if len(wmRegistrations) == 1 {
-		return wmRegistrations[0], nil
+		req := wmRegistrations[0]
+		wmRegistrations = []*tcworkermanager.RegisterWorkerRequest{}
```
milescrabill, line 85: `wmRegistrations` wasn't getting reset between tests, although the comment above notes that it is intended to be.

internal/workerproto/README.md
```diff
@@ -51,6 +51,14 @@ Before the connection is initialized, the connection's capabilities are unknown,
 
 The following sections describe the defined message types, each under a heading giving the corresponding capability.
 
+### error-report
+
+This message type, sent from the worker, contains an error report intended for the worker-manager to display among its worker pool errors.
+
+```
+~{"type": "error-report", "kind": "critical", "title": "a serious error", "description": "bad stuff has happened!", "extra": {"number of failures": 1}}
+```
```
djmitche, line 60: Maybe point to the worker-manager API docs here for definitions of these fields?

tools/worker-runner/tc/fakeworkermanager.go
djmitche: Sorry!

tools/worker-runner/errorreport/errorreport.go
```diff
@@ -0,0 +1,56 @@
...
+		log.Printf("Error processing error-report message, could not marshal extra")
+	}
+
+	errorReport := tcworkermanager.WorkerErrorReport{
+		Description: msg.Properties["description"].(string),
+		Kind:        msg.Properties["kind"].(string),
+		Extra:       extraMsg,
+		Title:       msg.Properties["title"].(string),
+	}
+	err = provider.ReportWorkerError(state, factory, &errorReport)
```
djmitche, line 52: Why this call back into the `provider/provider` package?  It seems like `errorreport` is the right place for this functionality!

tools/worker-runner/provider/provider/common.go
```diff
@@ -77,3 +78,15 @@ func RemoveWorker(state *run.State, factory tc.WorkerManagerClientFactory) error
...
 }
+
+// ReportWorkerError will send a worker error report to worker-manager
+func ReportWorkerError(state *run.State, factory tc.WorkerManagerClientFactory, payload *tcworkermanager.WorkerErrorReport) error {
+	wc, err := factory(state.RootURL, &state.Credentials)
+	if err != nil {
+		return errors.Wrap(err, "error instanciating worker-manager client")
+	}
+	if _, err = wc.ReportWorkerError(state.WorkerPoolID, payload); err != nil {
+		log.Printf("Error reporting worker error: %v\n", err)
```
djmitche, line 89: If this error happens, it's probably got a common cause with the report from the worker, so perhaps we should log the payload as well?

tools/worker-runner/errorreport/errorreport.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		log.Printf("Error reporting worker error: %v\n", err)
```
djmitche, line 54: I like that this is not considered a fatal error to worker-runner.

tools/worker-runner/errorreport/errorreport_test.go
```diff
@@ -0,0 +1,89 @@
...
+			Properties: map[string]interface{}{
+				"description": description,
+				"extra":       extra,
+				"kind":        kind,
+				"title":       title,
+			},
+		})
+
+		// delay in between sending message and checking outcome
+		time.Sleep(5 * time.Millisecond)
```
djmitche, line 46: Instead of assuming this is long enough, it'll be less intermittent to poll in a loop until there's more than zero reports.  To help the failure case to happen quickly, it also makes sense to put a relatively short timeout on that -- something I just added for my test in https://github.com/taskcluster/taskcluster/pull/3009/commits/5ec9281d4a504e621cc73cf19fda2b1e702ac10c.

tools/worker-runner/errorreport/errorreport_test.go
```diff
@@ -0,0 +1,89 @@
...
+			Properties: map[string]interface{}{
+				"description": description,
+				"extra":       extra,
+				"kind":        kind,
+				"title":       title,
+			},
+		})
+
+		// delay in between sending message and checking outcome
+		time.Sleep(1 * time.Millisecond)
```
djmitche, line 83: Same here :)

changelog/issue-2887.md
```diff
@@ -0,0 +1,4 @@
+audience: general
+level: silent
```
djmitche, line 2: I guess we'll have `patch` level changelogs when the workers implement this?

djmitche:
> The issue I see with setting the protocol where others are set is that we need the `workerManagerClientFactory` and `state` in order to make requests to worker-manager. Is there a clean way to get those without hooking into the provider?

For `workerManagerClientFactory`, you can create one of those yourself in the errorreporter (there are so many `r`'s in that name!!).  In the providers, we have a public constructor and a private constructor, with the private constructor taking the factory as an argument.  The only reason for the argument is to allow injecting dependencies in tests.

https://github.com/taskcluster/taskcluster/blob/a418f400c6780507921376858a724c3f429bce4f/tools/worker-runner/provider/aws/awsprovider.go#L167-L169
https://github.com/taskcluster/taskcluster/blob/a418f400c6780507921376858a724c3f429bce4f/tools/worker-runner/provider/aws/awsprovider.go#L191-L202

As for state, I think you can pass that to an errorreporter setup function, similar to
https://github.com/taskcluster/taskcluster/blob/a418f400c6780507921376858a724c3f429bce4f/tools/worker-runner/runner/run.go#L158-L159
(whether to use an `errorreporter.Setup` that sets some globals or create an object with `erroreporter.New` is up to you)

tools/worker-runner/errorreport/errorreport.go
```diff
@@ -0,0 +1,86 @@
...
+		log.Printf("Error processing error-report message, missing description or not string")
+	}
+	if !validate(msg.Properties, "kind", "string") {
+		log.Printf("Error processing error-report message, missing kind or not string")
+	}
+	if !validate(msg.Properties, "title", "string") {
+		log.Printf("Error processing error-report message, missing title or not string")
+	}
+	if !validate(msg.Properties, "extra", "map[string]interface {}") {
+		log.Printf("Error processing error-report message, missing extra or not map[string]interface{}")
```
djmitche, line 55: I suspect if any of these validations fail, then `WorkerErrorReport` will fail, too -- but perhaps that's OK?

tools/worker-runner/errorreport/errorreport_test.go
```diff
@@ -0,0 +1,102 @@
...
+				"description": description,
+				"extra":       extra,
+				"kind":        kind,
+				"title":       title,
+			},
+		})
+
+		var reports []*tcworkermanager.WorkerErrorReport
+		var err error
+		func() {
```
djmitche, line 52: Why is this in a goroutine?  Wouldn't it fail while a subsequent test is running?

BTW I added some flushing support in 40caa12290 which will hopefully make this easier.

tools/worker-runner/errorreport/errorreport_test.go
```diff
@@ -0,0 +1,102 @@
...
+			Properties: map[string]interface{}{
+				"description": description,
+				"extra":       extra,
+				"kind":        kind,
+				"title":       title,
+			},
+		})
+
+		// note that without a delay we may not have waited
+		// long enough for potential calls to finish
```
djmitche, line 96: ⚠️

tools/worker-runner/errorreport/errorreport_test.go
milescrabill: It's an anonymous function that gets called, it isn't a goroutine (no `go`), so it runs inline

djmitche: Oh, I did miss that!  But, I am still unclear on the reason to do so..

milescrabill: I did it so that I could neatly return or panic from that function rather than break the loop

tools/worker-runner/errorreport/errorreport.go
milescrabill: I'm making these `return`

tools/worker-runner/errorreport/errorreport_test.go
djmitche: Ah, I see.. neat!

PR #3016, opened 2020-06-09
Add support for reregister in worker-manager

### LINKED ISSUE #3012: Add support for reregister endpoint in worker-runner

This is one part of #2893

helfi92:
Resolves #3012.


db/src/fakes/worker_manager.js
```diff
@@ -455,6 +501,10 @@ class FakeWorkerManager {
       filteredWorkerKeys.length).map(key => this.workers.get(key));
   }
 
+  async get_workers_2(worker_pool_id, worker_group, worker_id, state, page_size, page_offset) {
+    return this.get_workers(worker_pool_id, worker_group, worker_id, state, page_size, page_offset);
+  }
+
```
djmitche, line 541: Could this method be omitted?  There's no reason to need to get more than one worker secret at a time..

db/versions/0013.yml
```diff
@@ -0,0 +1,166 @@
+version: 13
+migrationScript: 0013-migration.sql
+downgradeScript: 0013-downgrade.sql
+methods:
+  get_worker_2:
+    description: Get an existig worker. The returned table will have one or (if no such worker is defined) zero rows.
```
djmitche: Lol, this typo is multiplying quickly :)

```suggestion
    description: Get an existing worker. The returned table will have one or (if no such worker is defined) zero rows.
```

changelog/issue-2962.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
+level: silent
+reference: issue 2962
+---
+taskcluster-lib-postgres now supports `cryptoKeys` which allow for encrypting columns of a table.
```
djmitche, line 5: @helfi92 we'll need another changelog for the table change, with type `major` due to the newly-required config for the worker-manager service.

services/worker-manager/schemas/v1/reregister-worker-response.yml
```diff
@@ -0,0 +1,49 @@
...
+    additionalProperties: false
+    required:
+      - accessToken
+      - clientId
+  secret:
+    title: Secret
+    type: string
+    pattern: "^[a-zA-Z0-9_-]{44}$"
+    description: |
+      The next secret value needed to reregister the worker (in `reregisterWorker).
```
djmitche, line 44: Let's add a page to the worker-manager docs that describes this overall process, and refer to it from here.

services/worker-manager/src/data.js
```diff
@@ -246,7 +248,7 @@ class Worker {
   // UNIQUE_VIOLATION when those checks fail.
   async create(db) {
```
djmitche, line 252: It looks like a worker is never created with a non-null secret, so perhaps there's no need for a `create_worker_2` function either?

services/worker-manager/src/main.js
```diff
@@ -46,6 +46,13 @@ let load = loader({
       serviceName: 'worker_manager',
       monitor: monitor.childMonitor('db'),
       statementTimeout: process === 'server' ? 30000 : 0,
+      cryptoKeys: [
```
djmitche, line 49: I think this should be done differently -- @imbstack ?

helfi92:
Rebased and changed bumped the db version number to 14.

services/worker-manager/src/providers/aws.js
```diff
@@ -298,14 +299,20 @@ class AwsProvider extends Provider {
       workerId: worker.workerId,
     });
     monitor.debug('setting state to RUNNING');
+    const secret = `${slug.nice()}${slug.nice()}`;
     await worker.update(this.db, worker => {
       worker.lastModified = new Date();
       worker.providerData.terminateAfter = expires.getTime();
       worker.state = Worker.states.RUNNING;
+      worker.secret = this.db.encrypt({ value: Buffer.from(secret, 'utf8') });
```
djmitche, line 307: The secret for *re*registration is generated in `api.js`.  Unless there's some reason for the functionality to differ per provider, I think it will make sense to do the same for registration, rather than implementing it in each provider.

services/worker-manager/config.yml
```diff
@@ -19,6 +19,7 @@ defaults:
   postgres:
     readDbUrl: !env READ_DB_URL
     writeDbUrl: !env WRITE_DB_URL
+    cryptoKey: !env PG_CRYPTO_KEY
```
djmitche, line 22: The deployment docs specify `db_crypto_key`, and that it is in JSON format

services/worker-manager/config.yml
```diff
@@ -46,6 +47,8 @@ test:
     accountId: "pamplemousse"
   app:
     errorsExpirationDelay: '-1 hour'
+  postgres:
+    cryptoKey: "aSdtIGJzdGFjayEgaGVsbG8gZnV0dXJlIHBlcnNvbgo="
```
djmitche, line 51: I think this needs to be in object format?

services/worker-manager/test/api_test.js
```diff
@@ -7,6 +8,9 @@ const fs = require('fs');
 const path = require('path');
 
 helper.secrets.mockSuite(testing.suiteName(), ['db'], function(mock, skipping) {
+  if (mock) {
+    return;
+  }
```
djmitche, line 13: Was this temporary?

services/worker-manager/test/api_test.js
```diff
@@ -1125,6 +1138,178 @@ helper.secrets.mockSuite(testing.suiteName(), ['db'], function(mock, skipping) {
...
+      });
+
+      // default is 96 hours when reregistrationTimeout is not specified.
+      assert(new Date(secondResponse.expires) - new Date() > 95 * 1000 * 60 * 60);
+      assert(new Date(secondResponse.expires) - new Date() < 96 * 1000 * 60 * 60);
+      assert.equal(firstResponse.credentials.clientId, secondResponse.credentials.clientId);
+      assert.notStrictEqual(firstResponse.secret, secondResponse.secret);
+    });
+
+    test('works with registrationTimeout', async function() {
```
djmitche, line 1197:
```suggestion
    test('works with reregistrationTimeout', async function() {
```

services/worker-manager/test/api_test.js
```diff
@@ -1125,6 +1138,178 @@ helper.secrets.mockSuite(testing.suiteName(), ['db'], function(mock, skipping) {
...
+
+    suiteTeardown(function() {
+      helper.load.restore();
+    });
+
+    const defaultRegisterWorker = {
+      workerPoolId, providerId, workerGroup, workerId, workerIdentityProof,
+    };
+
+    test('works without registrationTimeout', async function() {
```
djmitche, line 1167: These test cases look similar -- it might be helpful to factor them out a bit so that the reader sees only the things that differ from case to case.

ui/docs/manual/deploying/database.mdx
```diff
@@ -99,12 +99,30 @@ For example:
...
+{
+  "id": "a unique identifier",
+  "algo": "aes-256",
+  "key": "32 bytes of base64 string"
+}
+```
+
+The last element of the array will be the current key and all encryption will be performed with it. Other keys are used in decryption of values that were encrypted with the
+older key. This allows you to rotate your encryption keys. At the current time, Taskcluster implicitly treats `azure_crypto_key` as the oldest element of this list, with id `azure`, to maintain compatibility with existing configurations.
+from older table formats.
```
djmitche, line 124: @imbstack now that I read this in context, maybe it makes sense to add a paragraph here about upgrading from `azure_crypto_key`?

services/worker-manager/schemas/v1/reregister-worker-response.yml
helfi92: How would you refer to it from here when the root url could differ?

djmitche: When this is rendered as markdown, it could be an origin-relative URL (`/docs/..`), but for the moment maybe it's best to just point to `https://docs.taskcluster.net/docs/..`

helfi92: done.

services/worker-manager/src/api.js
```diff
@@ -609,8 +609,10 @@ builder.declare({
   }
 
   let expires, workerConfig;
+  const secret = `${slug.nice()}${slug.nice()}`;
   try {
-    const reg = await provider.registerWorker({worker, workerPool, workerIdentityProof});
+    const encryptedSecret = this.db.encrypt({ value: Buffer.from(secret, 'utf8') });
+    const reg = await provider.registerWorker({worker, workerPool, workerIdentityProof, secret: encryptedSecret });
```
djmitche, line 615: This is a little awkward.  It does save calling the `updateWorker` db function twice, but requires that all of the providers know to write this secret to the DB.  It works as-is, but might be worth making the second call to `updateWorker` for the conceptual clarity of not having to have every provider handle this correctly (including the static provider..).

services/worker-manager/src/providers/aws.js
```diff
@@ -302,10 +297,15 @@ class AwsProvider extends Provider {
...
       worker.state = Worker.states.RUNNING;
+      worker.secret = secret;
     });
 
     const workerConfig = worker.providerData.workerConfig || {};
-    return {expires, workerConfig};
+    return {
+      expires,
+      workerConfig,
+      secret,
```
djmitche, line 307: This (and for other providers) doesn't need to return the secret anymore.

services/worker-manager/src/providers/aws.js
```diff
@@ -260,13 +260,8 @@ class AwsProvider extends Provider {
...
    * This method checks instance identity document authenticity
    * If it's authentic it checks whether the data in it corresponds to the worker
-   *
-   * @param worker string
-   * @param workerPool string
-   * @param workerIdentityProof {document: string, signature: string}
-   * @returns {Promise<{expires: *}>}
    */
-  async registerWorker({worker, workerPool, workerIdentityProof}) {
+  async registerWorker({worker, workerPool, workerIdentityProof, secret}) {
```
djmitche, line 264: For future readers it might help to call this encryptedSecret, so we can see that it's already encrypted.

services/worker-manager/src/providers/static.js
```diff
@@ -76,7 +77,12 @@ class StaticProvider extends Provider {
       expires = taskcluster.fromNow('96 hours');
     }
     const workerConfig = worker.providerData.workerConfig || {};
-    return {expires, workerConfig};
+    const secret = `${slug.nice()}${slug.nice()}`;
```
djmitche, line 80: Looks like this provider didn't get updated to handle a secret argument?

ui/docs/reference/core/worker-manager/worker-interaction.mdx
```diff
@@ -48,6 +48,13 @@ It is the responsibility of the worker to finish its work before the credentials
 
 Depending on the provider, `registerWorker` may be called multiple times for the same worker, or may only be called once.
 
+## Reregistration
+
+The worker may reregister itself to renew its credentials with the worker manager by calling [`worker.reregister`](/docs/reference/core/worker-manager/api#reregisterWorker)
```
djmitche, line 53:
```suggestion
The worker may reregister itself to renew its credentials with the worker manager by calling [`workerManager.reregisterWorker`](/docs/reference/core/worker-manager/api#reregisterWorker)
```

djmitche: Let's mention the secret here, and that each secret can only be used once.

djmitche:
Oh, needs a `yarn generate` again :)

services/worker-manager/src/api.js
```diff
@@ -609,8 +609,13 @@ builder.declare({
   }
 
   let expires, workerConfig;
+  const secret = `${slug.nice()}${slug.nice()}`;
   try {
-    const reg = await provider.registerWorker({worker, workerPool, workerIdentityProof});
+    const encryptedSecret = this.db.encrypt({ value: Buffer.from(secret, 'utf8') });
+    await worker.update(this.db, worker => {
+      worker.secret = encryptedSecret;
+    });
```
imbstack, line 617: This isn't a huge issue but doesn't this allow me to write whatever I want into the secret column of any worker I want since we haven't verified the identity of the client yet at this point? I think doing this after `provider.registerWorker` makes more sense?

services/worker-manager/src/util.js
```diff
@@ -23,3 +24,24 @@ const joinWorkerPoolId = (provisionerId, workerType) => {
 };
 exports.joinWorkerPoolId = joinWorkerPoolId;
 exports.MAX_MODIFY_ATTEMPTS = 5;
+
+const createCredentials = (worker, expires, cfg) => {
```
imbstack, line 28: The comment above this in `registerWorker` explains the importance of using only values that we've generated/stored for this interpolation. I think it would be worth mentioning that here too for future uses

services/worker-manager/test/api_test.js
```diff
@@ -1,4 +1,5 @@
 const taskcluster = require('taskcluster-client');
+const slug = require('slugid');
```
imbstack, line 2: Could we update whatever test in here asserts that you can fetch a worker via the api to assert that it does _not_ include the field `secret`? As you've written this now it will be fine, I just worry about refactors in the future accidentally exposing this field.

ui/docs/reference/core/worker-manager/worker-interaction.mdx
```diff
@@ -48,6 +48,16 @@ It is the responsibility of the worker to finish its work before the credentials
 
 Depending on the provider, `registerWorker` may be called multiple times for the same worker, or may only be called once.
 
+## Reregistration
+
+The worker may reregister itself to renew its credentials with the worker manager by calling
+[`workerManager.reregister`](/docs/reference/core/worker-manager/api#reregisterWorker) and
+supplying it with a one time usage `secret`. The secret value will have been either last configured
+in `registerWorker` (in the case of a newly registerd worker) or `reregisterWorker`.
```
imbstack, line 56: registerd

ui/docs/reference/core/worker-manager/worker-interaction.mdx
```diff
@@ -48,6 +48,16 @@ It is the responsibility of the worker to finish its work before the credentials
 
 Depending on the provider, `registerWorker` may be called multiple times for the same worker, or may only be called once.
 
+## Reregistration
+
+The worker may reregister itself to renew its credentials with the worker manager by calling
+[`workerManager.reregister`](/docs/reference/core/worker-manager/api#reregisterWorker) and
+supplying it with a one time usage `secret`. The secret value will have been either last configured
```
imbstack, line 55: "configured" here implies to me that it is a user-supplied value. Let's rephrase this to say that worker-manager generates these values.

services/worker-manager/schemas/v1/reregister-worker-request.yml
```diff
@@ -0,0 +1,19 @@
...
+    title: Secret
+    type: string
+    pattern: "^[a-zA-Z0-9_-]{44}$"
+    description: |
+      The secret value that was last configured in `registerWorker` (in the case of a newly registerd worker) or
+      `reregisterWorker`.
+      For more information, refer to https://docs.taskcluster.net/docs/reference/core/worker-manager#reregistration.
+additionalProperties: false
+required:
+  - secret
```
imbstack, line 19: Aren't all of these fields required?

ui/docs/reference/core/worker-manager/worker-interaction.mdx
```diff
@@ -48,6 +48,16 @@ It is the responsibility of the worker to finish its work before the credentials
...
+in `registerWorker` (in the case of a newly registerd worker) or `reregisterWorker`.
+The expiration time of the new credentials can be controlled via `reregistrationTimeout` which gets set when
+calling [`workerManager.registerWorker`](/docs/reference/core/worker-manager/api#registerWorker).
+The default behavior, when `reregistrationTimeout` is omitted, is to extend the credentials by 96 hours.
```
imbstack, line 59: The value of this actually comes from workerpool configuration right? The call to `reregisterWorker` can't change this field iiuc.

services/worker-manager/config.yml
```diff
@@ -19,6 +19,7 @@ defaults:
   postgres:
     readDbUrl: !env READ_DB_URL
     writeDbUrl: !env WRITE_DB_URL
+    dbCryptoKeys: !env:json:optional DB_CRYPTO_KEYS
```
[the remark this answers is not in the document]
helfi92: Makes sense. Do you also want me to change `DB_CRYPTO_KEYS` to `CRYPTO_KEYS`? I originally picked dbCryptoKeys to align myself with https://github.com/taskcluster/taskcluster/blob/master/ui/docs/manual/deploying/database.mdx#supporting-encrypted-columns.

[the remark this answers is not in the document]
imbstack: Ah, I see. The env var makes sense to have the `DB` prefix since it isn't nested. But I think having the parallel with the env var here is fine if you'd like. Leave it or change it, either is fine with me!

ui/docs/reference/core/worker-manager/worker-interaction.mdx
helfi92: Correct. I went ahead and rephrased it to say that it comes from the worker pool configuration. Happy to change it to something else.

services/worker-manager/src/api.js
helfi92: Agreed. Makes more sense.

PR #3037, opened 2020-06-11
Add currentCapacity to ui for workerpools

### LINKED ISSUE #2882: Update workerpool page in UI to be more useful
This includes making pending count, existing capacity, workers, and errors for a workerpool visible on the ui page for a workerpool without going back up to the list of workerpools

This depends on adding existing capacity to workerpools as a top-level field in w-m api (#2883). That blocks in-turn on making all static pools register/reregister (#2926). That blocks in turn on supporting reregistration (#2893).

imbstack:
Github Bug/Issue: Fixes #2882 

As it says on the tin! Doesn't fix all of the issues in the issue but covers the most helpful ones IMHO

### DIFF
--- a/services/web-server/src/graphql/WorkerManager.graphql
+++ b/services/web-server/src/graphql/WorkerManager.graphql
@@ -5,6 +5,7 @@ type WorkerManagerWorkerPoolSummary {
     providerId: String!
     config: JSON
     emailOnError: Boolean!
+    currentCapacity: Int!
     pendingTasks: Int #this value comes from a separate request to queue
 }
 
--- a/ui/src/components/WMWorkerPoolEditor/index.jsx
+++ b/ui/src/components/WMWorkerPoolEditor/index.jsx
@@ -5,10 +5,15 @@ import List from '@material-ui/core/List';
 import ListSubheader from '@material-ui/core/ListSubheader';
 import ListItem from '@material-ui/core/ListItem';
 import ListItemText from '@material-ui/core/ListItemText';
+import Chip from '@material-ui/core/Chip';
+import Avatar from '@material-ui/core/Avatar';
 import Typography from '@material-ui/core/Typography';
 import MenuItem from '@material-ui/core/MenuItem';
+import WorkerIcon from 'mdi-react/WorkerIcon';
+import MessageAlertIcon from 'mdi-react/MessageAlertIcon';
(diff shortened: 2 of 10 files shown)


services/web-server/src/graphql/WorkerManager.graphql
```diff
@@ -5,6 +5,7 @@ type WorkerManagerWorkerPoolSummary {
     providerId: String!
     config: JSON
     emailOnError: Boolean!
+    currentCapacity: Int!
```
helfi92, line 8: Navigating to http://localhost:5080/worker-manager shows a red banner with text `Cannot return null for non-nullable field WorkerManagerWorkerPoolSummary.currentCapacity.`

Is this value not always defined?

imbstack: This will only work once the services with `currentCapacity` are deployed. It works when I run it against my dev deployment

helfi92: Makes sense. Thank you!

imbstack:
Sweet, thanks for the pointers. I've updated it to look like that and added a couple links like the issue wanted. Feel free to have me solve this in a different way though!

ui/src/components/WMWorkerPoolEditor/index.jsx
```diff
@@ -315,10 +332,57 @@ export default class WMWorkerPoolEditor extends Component {
...
+                label="Pending Tasks"
+              />
+            </li>
+            <li>
+              <Chip
+                size="small"
+                className={classes.metadata}
+                icon={<WorkerIcon />}
+                label="Workers (Queue View)"
+                component="a"
```
helfi92, line 364: We have a custom component "Link" that allows prefetching the view before the user clicks on it. It will help make the transition much faster. We would need to also change the `href` to `to`.

helfi92: See the diff :)

imbstack:
Oh interesting. The capacity work has been merged for a while. Does it need to be deployed to production too before we land this?

helfi92:
> Oh interesting. The capacity work has been merged for a while. Does it need to be deployed to production too before we land this?

Ah, yes, sorry. Ship it :)

PR #3041, opened 2020-06-12
Support reregistration in worker-runner

### LINKED ISSUE #3012: Add support for reregister endpoint in worker-runner

djmitche:
Github Bug/Issue: Fixes #3012


djmitche:
This will need a few rebases, but they should be relatively straightforward.  Let's do a round of review commit-by-commit, starting with "Add locking".

petemoore:
> This will need a few rebases, but they should be relatively straightforward. Let's do a round of review commit-by-commit, starting with "Add locking".

Is there value in splitting this PR into a series of smaller ones?

djmitche:
> Is there value in splitting this PR into a series of smaller ones?

Good idea -- most of these commits are refactors..

djmitche:
Actually only the first commit (a4e18c8) could be broken out.  The reason for the long rebase is that this PR (with the exception of a4e18c8) depends on @helfi92's work which depends on @imbstack's encrypted-column work which we can't land until it gets security review.

tools/worker-runner/registration/registration.go
```diff
@@ -96,22 +98,81 @@ func (reg *RegistrationManager) WorkerStarted() error {
...
 		}
 	})
 
 	return nil
 }
 
+// Request new credentials and send them to the worker, falling back to requesting
+// a worker termination if anything fails.  Called with reg.state locked
+func (reg *RegistrationManager) reregisterWorker() {
+	log.Println("Taskcluster Credentials are expiring in 30s; re-registering")
```
imbstack, line 119: I wonder if we should give a bit more headroom than 30s in the case of retries, etc.

tools/worker-runner/registration/registration.go
```diff
@@ -96,22 +98,81 @@ func (reg *RegistrationManager) WorkerStarted() error {
...
+		log.Printf("Could not create worker-manager client: %v", err)
+		reg.terminateWorker()
+		return
+	}
+
+	res, err := wm.ReregisterWorker(&tcworkermanager.ReregisterWorkerRequest{
+		//WorkerPoolID:        reg.state.WorkerPoolID,
+		//ProviderID:          reg.state.ProviderID,
+		//WorkerGroup:         reg.state.WorkerGroup,
+		//WorkerID:            reg.state.WorkerID,
```
imbstack, line 132: I haven't looked at the worker-manager patch yet but I'm assuming these shouldn't be commented out?

tools/worker-runner/run/state.go
```diff
@@ -98,3 +102,61 @@ func (state *State) CheckProviderResults() error {
...
+	}
+	err = ioutil.WriteFile(filename, encoded, 0700)
+	if err != nil {
+		return err
+	}
+
+	// This file contains secrets, so ensure that this is really only
+	// accessible to the file owner (and having just created the file, that
+	// should be the current user).
+	err = perms.MakePrivateToOwner(filename)
```
imbstack, line 122: Should we be doing this in a different order? Create empty file -> set permissions -> append content? This leaves it potentially readable for a bit and longer if we crash, right?

tools/worker-runner/registration/registration.go
djmitche: Yeah, I'll catch that on rebase, thanks!

tools/worker-runner/registration/registration.go
djmitche: Good point -- and we want to leave some leeway for restarts, etc. too.  I'll make a more sophisticated check.

tools/worker-runner/run/state.go
djmitche: Yeah, let's defer that to a followup though -- this is just moving existing code.  #3067.

djmitche:
Oh, another issue: this never resets the timer after reregistering!

djmitche:
Rebased on top of master -- I can rebase onto master again once @helfi92's work lands.

services/worker-manager/schemas/v1/reregister-worker-request.yml
```diff
@@ -0,0 +1,19 @@
...
+  workerGroup: {$ref: "worker-full.json#/properties/workerGroup"}
+  workerId: {$ref: "worker-full.json#/properties/workerId"}
+  secret:
+    title: Secret
+    type: string
+    pattern: "^[a-zA-Z0-9_-]{44}$"
+    description: |
+      The secret value that was last configured in `registerWorker` (in the case of a newly registerd worker) or
+      `reregisterWorker`.
+      For more information, refer to https://docs.taskcluster.net/docs/reference/core/worker-manager#reregistration.
```
petemoore, line 16: I didn't realise we still maintain `docs.taskcluster.net`.

djmitche: It's automatically deployed from master.

PR #3045, opened 2020-06-12
Document that paths are absolute in docker-worker artifacts

djmitche:


petemoore:
@djmitche Have you tested with a relative path, to be sure they don't work?

djmitche:
No, I'm not sure what it would be relative to. I added this because in chat I was pretty sure the paths were absolute but didn't find a statement here in the schema.  It is documented as a full path in /docs/manual/using/artifacts.

djmitche:
I don't know what's up with the Python failures!

```
AttributeError: module 'virtualenv.activation.python' has no attribute 'PestythonActivator'
```

I don't see anything related in the diff, and

```
lamport ~/p/taskcluster2 [d-w-path-absolute] $ git grep Pestython
lamport ~/p/taskcluster2 [d-w-path-absolute] $ git grep Pest
lamport ~/p/taskcluster2 [d-w-path-absolute] $ git grep -i Pestython
```

djmitche:
#3046.

PR #3068, opened 2020-06-15
issue 2887: support error-report message in generic-worker

milescrabill:
Issue: part of #3010

### DIFF
--- a/workers/generic-worker/errorreport/errorreport.go
+++ b/workers/generic-worker/errorreport/errorreport.go
@@ -1,28 +1,32 @@
 package errorreport
 
 import (
-	"github.com/taskcluster/taskcluster/v30/internal/workerproto"
+	"fmt"
+
+	"github.com/taskcluster/taskcluster/v34/internal/workerproto"
 )
 
-func Send(proto *workerproto.Protocol, err error, debugInfo map[string]string) {
+func Send(proto *workerproto.Protocol, message interface{}, debugInfo map[string]string) {
+	if !proto.Capable("error-report") {
+		return
+	}
+
 	title := "generic-worker error"
-	description := err.Error()
+	description := fmt.Sprintf("%s", message)
 	// could support differentiating for panics
-	kind := "error"
+	kind := "worker-error"
 	// convert debugInfo from map[string]string to map[string]interface{}
 	extra := map[string]interface{}{}
 	for k, v := range debugInfo {
 		extra[k] = v
 	}
-	if proto.Capable("error-report") {
-		proto.Send(workerproto.Message{
-			Type: "error-report",
-			Properties: map[string]interface{}{
-				"description": description,
-				"kind":        kind,
-				"title":       title,
-				"extra":       extra,
-			},
-		})
-	}
+	proto.Send(workerproto.Message{
+		Type: "error-report",
+		Properties: map[string]interface{}{
+			"description": description,
+			"kind":        kind,
+			"title":       title,
+			"extra":       extra,
+		},
+	})
 }
--- a/workers/generic-worker/errorreport/errorreport.go
+++ b/workers/generic-worker/errorreport/errorreport.go
@@ -1,0 +1,28 @@
+package errorreport
+
+import (
+	"github.com/taskcluster/taskcluster/v30/internal/workerproto"
+)
+
+func Send(proto *workerproto.Protocol, err error, debugInfo map[string]string) {
+	title := "generic-worker error"
+	description := err.Error()
+	// could support differentiating for panics
+	kind := "error"
+	// convert debugInfo from map[string]string to map[string]interface{}
+	extra := map[string]interface{}{}
+	for k, v := range debugInfo {
+		extra[k] = v
+	}
+	if proto.Capable("error-report") {
+		proto.Send(workerproto.Message{
+			Type: "error-report",
+			Properties: map[string]interface{}{
+				"description": description,
+				"kind":        kind,
+				"title":       title,
+				"extra":       extra,
+			},
+		})
+	}
+}
--- a/workers/generic-worker/errorreport/errorreport_test.go
+++ b/workers/generic-worker/errorreport/errorreport_test.go
@@ -1,0 +1,88 @@
+package errorreport
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/taskcluster/taskcluster/v30/internal/workerproto"
+	workerProtoTesting "github.com/taskcluster/taskcluster/v30/internal/workerproto/testing"
+)
+
+func setupProtocols() (*workerproto.Protocol, *workerproto.Protocol) {
+	workerTransp, runnerTransp := workerProtoTesting.NewLocalTransportPair()
+	workerProto := workerproto.NewProtocol(workerTransp)
+	workerProto.AddCapability("error-report")
+	workerProto.Start(true)
+
+	runnerProto := workerproto.NewProtocol(runnerTransp)
+	runnerProto.AddCapability("error-report")
+	return workerProto, runnerProto
+}
+
+func TestSendEmptyExtra(t *testing.T) {
+	errorReported := false
+
+	workerProto, runnerProto := setupProtocols()
+
+	runnerProto.Register("error-report", func(msg workerproto.Message) {
+		assert.Contains(t, msg.Properties, "title")
+		assert.Equal(t, "generic-worker error", msg.Properties["title"])
+		assert.Contains(t, msg.Properties, "kind")
+		assert.Equal(t, "error", msg.Properties["kind"])
+		assert.Contains(t, msg.Properties, "description")
+		assert.Equal(t, "test error", msg.Properties["description"].(string))
+		assert.Contains(t, msg.Properties, "extra")
+		assert.Equal(t, msg.Properties["extra"], map[string]interface{}{})
+		errorReported = true
+	})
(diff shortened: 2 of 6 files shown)


workers/generic-worker/errorreport/errorreport.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"github.com/taskcluster/taskcluster/v30/internal/workerproto"
+)
+
+func Send(proto *workerproto.Protocol, err error, debugInfo map[string]string) {
+	title := "generic-worker error"
+	description := err.Error()
+	// could support differentiating for panics
+	kind := "error"
```
imbstack, line 11: maybe we call this `worker-error` or something to differentiate it from errors that come from w-m itself (in addition to how you've done it with the title)? Feel free to leave the same though.

workers/generic-worker/errorreport/errorreport_test.go
```diff
@@ -0,0 +1,101 @@
...
+	lock.Lock()
+	for i := 0; i < 200; i++ {
+		if !errorReported {
+			time.Sleep(time.Millisecond * 10)
+		} else {
+			return
+		}
+	}
+	lock.Unlock()
+	t.Fatalf("No error-report was received")
```
petemoore, line 61: Since you've acquired the lock on line 52, the callback must have completed, so no need to have the waiting `for` loop on line 53. The lock should be released independently of whether the test passes or not, so it is probably simplest to release it immediately after acquiring it.

Nit: the only way the test can fail is if it hangs waiting to acquire the lock on line 52, which would happen when the test eventually times out. Perhaps a context with a deadline would provide quicker feedback, although maybe not worth changing now.

```suggestion
	lock.Unlock()
	assert.True(t, errorReported, "No error-report was received")
```

workers/generic-worker/errorreport/errorreport.go
```diff
@@ -0,0 +1,28 @@
+package errorreport
+
+import (
+	"github.com/taskcluster/taskcluster/v30/internal/workerproto"
+)
+
+func Send(proto *workerproto.Protocol, err error, debugInfo map[string]string) {
```
petemoore, line 7: Is it worth returning an `error` to indicate if the message could be sent or not? Even if it isn't evaluated by the receiver.

workers/generic-worker/errorreport/errorreport.go
```diff
@@ -0,0 +1,28 @@
...
+	title := "generic-worker error"
+	description := err.Error()
```
petemoore, line 9: A `panic` can send more than just an error, so it might be better to receive `message interface{}` 
rather than `err error` and then format it as a string with e.g. `description := fmt.Sprintf("%s", message)`. This would allow you to call `errorreport.Send` for any panic, rather than just panics that report the `error` type.

workers/generic-worker/errorreport/errorreport_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// unlocked in callback
+	lock.Lock()
+	// report test error
+	Send(workerProto, fmt.Errorf("test error"), map[string]string{
+		"kangaroo": "pouches",
+		"monster":  "trucks",
+	})
+
+	lock.Lock()
```
petemoore, line 91: Same pattern here as above; you should probably unlock immediately afterwards, and skip the `for` loop.

workers/generic-worker/main.go
```diff
@@ -371,6 +388,9 @@ func HandleCrash(r interface{}) {
 	log.Print(string(debug.Stack()))
 	log.Print(" *********** PANIC occurred! *********** ")
 	log.Printf("%v", r)
+	if err, ok := r.(error); ok && err != nil {
+		errorreport.Send(WorkerRunnerProtocol, err, debugInfo)
+	}
```
petemoore, line 393: See above - here I think we can send anything through if we change Send function to receive `interface{}` instead of `error`.

workers/generic-worker/errorreport/errorreport.go
```diff
@@ -0,0 +1,28 @@
...
+	// could support differentiating for panics
+	kind := "worker-error"
+	// convert debugInfo from map[string]string to map[string]interface{}
+	extra := map[string]interface{}{}
+	for k, v := range debugInfo {
+		extra[k] = v
```
petemoore, line 15: Note, some of the fields may have empty values (e.g. `instanceType` is an optional config setting) so some of the properties may have empty values. Does worker manager handle this graciously, or should we remove properties with empty values here? I suspect keeping them is better (as you have done), so that it is explicit that the value was unset, rather than just not included, but we should definitely decide if this is what we want.

workers/generic-worker/errorreport/errorreport.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+	if proto.Capable("error-report") {
```
petemoore, line 17: nit: it would be more efficient to check this condition at the start of the function, e.g. at the start have instead:
```go
	if !proto.Capable("error-report") {
		return
	}
```
and then here have one less level of nesting without the `if` statement.

workers/generic-worker/main.go
```diff
@@ -1249,5 +1269,6 @@ func exitOnError(exitCode ExitCode, err error, logMessage string, args ...interf
 	log.Printf(logMessage, args...)
 	log.Printf("Root cause: %v", err)
 	log.Printf("%#v (%T)", err, err)
+	errorreport.Send(WorkerRunnerProtocol, err, debugInfo)
```
petemoore, line 1272: This error report should include `exitCode`, `err`, `logMessage` and `args`, not just `err`, since a lot of important context is lost if you only include `err`.

workers/generic-worker/errorreport/errorreport.go
petemoore: Oh wait - [`(workerproto.Transport).Send`](https://godoc.org/github.com/taskcluster/taskcluster/internal/workerproto#Transport) doesn't return an `error`?!?

petemoore: @djmitche Why does `PipeTransport` [silently fail](https://github.com/taskcluster/taskcluster/blob/ae8e898abac9821c48b2bf04f5f56561bdd7b690/internal/workerproto/pipe.go#L46-L49) without informing the caller? How can the worker know that its messages are processed?

petemoore: Having reflected on this a little more, I realised the `Send` method is essentially just writing to standard out, so we probably don't need to care too much about an error there, in the same way that we typically don't check the error returned by `fmt.Println(string)`. Note `fmt.Println` does actually return an error, but we just typically ignore it.

workers/generic-worker/errorreport/errorreport_test.go
```diff
@@ -0,0 +1,86 @@
...
+	assert.True(t, errorReported, "No error-report was received")
+	lock.Unlock()
```
petemoore, line 85: The Unlock should be before the assert statement, otherwise if the test fails, the lock will never get released.

workers/generic-worker/errorreport/errorreport_test.go
```diff
@@ -0,0 +1,86 @@
...
+	runnerProto.WaitUntilInitialized()
+
+	// unlocked in callback
+	lock.Lock()
+	// report test error
+	Send(workerProto, fmt.Errorf("test error"), nil)
+
+	lock.Lock()
+	assert.True(t, errorReported, "No error-report was received")
+	lock.Unlock()
```
petemoore, line 53: `Unlock` needs to happen before `assert` here

workers/generic-worker/errorreport/errorreport_test.go
```diff
@@ -24,6 +24,7 @@ func setupProtocols() (*workerproto.Protocol, *workerproto.Protocol) {
 func TestSendEmptyExtra(t *testing.T) {
 	errorReported := false
 	lock := sync.Mutex{}
+	defer lock.Unlock()
```
petemoore, line 27: If the routine was to prematurely exit any time after line 27 (`defer lock.Unlock()`), but _before_ the mutex was locked on line 52, we would have a runtime exception due to trying to unlock an unlocked mutex.

There are only two real possible solutions, and they both would be at line 51 where the mutex is locked:

First option:

```
lock.Lock()
defer lock.Unlock()
```

or second option:

```
lock.Lock()
lock.Unlock()
```

I prefer the first of these because it is simpler, and there is no reason for the lock to be held while the test failure condition is checked and reported.

workers/generic-worker/errorreport/errorreport_test.go
```diff
@@ -50,12 +51,12 @@ func TestSendEmptyExtra(t *testing.T) {
 
 	lock.Lock()
 	assert.True(t, errorReported, "No error-report was received")
-	lock.Unlock()
 }
 
 func TestSendExtra(t *testing.T) {
 	errorReported := false
 	lock := sync.Mutex{}
+	defer lock.Unlock()
```
petemoore, line 59: Same here, this is the wrong place to add this defer statement, since any failure after this line, but before line 83 would lead to a runtime exception.

workers/generic-worker/errorreport/errorreport_test.go
milescrabill: We have to do the second to avoid linter errors. We could disable them, but I think this is fine.

djmitche:
This is approved -- should we get it landed?

PR #3069, opened 2020-06-15
Bug 1638921 - ensure crontasks do not run concurrently

djmitche:
Bugzilla Bug: [1638921](https://bugzilla.mozilla.org/show_bug.cgi?id=1638921)


djmitche:
That sounds like a good idea.  We're doing some of them, and others aren't relevant, so it shouldn't be too bad.

PR #3073, opened 2020-06-16
[worker-runner]  Support calling reregisterWorker when credentials expire 

### LINKED ISSUE #3012: Add support for reregister endpoint in worker-runner

djmitche:
Already reviewed by @imbstack and @pmoore in #3041.

Github Bug/Issue: Fixes #3012


djmitche:
CI's good, so landing on the basis of reviews on #3041.

PR #3077, opened 2020-06-16
use 'dbCryptoKeys' everywhere, fix typo

djmitche:
This fixes a typo in `services/worker-manager/src/main.js` (`crytoKeys`), and additionally calls the value `dbCryptoKeys` throughout, rather than using different names in the configuration and the source code.


djmitche:
confirmed working in dev

PR #3078, opened 2020-06-16
Grant the worker-manager:reregister-worker:* scope to worker-manager …

djmitche:
…itself


djmitche:
force-pushed with a `yarn generate` run..

PR #3090, opened 2020-06-17
use a sync.Cond to wait for creds to expire in tests

djmitche:


tools/worker-runner/registration/registration_test.go
```diff
@@ -112,6 +115,8 @@ func TestCredsExpirationNewCredentials(t *testing.T) {
 
 	fmt.Printf("setting fake expires to %s\n", expires)
 	reg := new(&runnercfg, &state, tc.FakeWorkerManagerClientFactory)
+	reg.credsExpireCond = sync.NewCond(&sync.Mutex{})
+	reg.credsExpireCond.L.Lock()
```
imbstack, line 119: do I read this correctly that this lock will happen after the lock in `reg.credsExpireTimer`?

djmitche: No, this one should happen first.  The timer isn't started until `reg.WorkerStarted`.

The lock is basically held for the duration of the test, except that the test itself releases the lock implicitly by calling `Wait`, and the Timer callback locks, notifies, and unlocks it during that time.

djmitche:
(queue failure here is an S3 issue)

PR #3085, opened 2020-06-17
Only shut down between claimWork calls.

djmitche:
This avoids the issue where the worker starts shutting down while
waiting for a claimWork call that eventually returns tasks (which
otherwise leaves those tasks timing out to claim-expired).

This substantially refactors ShutdownManager and TaskListener to match.

If this passes CI, I'll refactor the "host.getTerminationTime" thing a bit (it doesn't even return a time! and it doesn't distinguish graceful from immediate shutdown)

Github Bug/Issue: Fixes #3080

PR #3091, opened 2020-06-18
Support graceful shutdown in docker-worker

djmitche:
This was a quick refactor away from #3085.  The first two commits are from that PR, so please consider only the last commit here.


djmitche:
Yes, those tests exit.

djmitche:
They both exist and exit ;)

PR #3095, opened 2020-06-18
Bug 1635455: support staging releases

djmitche:
Bugzilla Bug: [1635455](https://bugzilla.mozilla.org/show_bug.cgi?id=1635455)

### DIFF
--- a/infrastructure/tooling/src/publish/index.js
+++ b/infrastructure/tooling/src/publish/index.js
@@ -25,8 +25,8 @@ class Publish {
 
   async generateTasks() {
     // try loading the secret into process.env
-    if (process.env.TASKCLUSTER_PROXY_URL && !this.cmdOptions.staging) {
-      const secretName = "project/taskcluster/release";
+    if (process.env.TASKCLUSTER_PROXY_URL) {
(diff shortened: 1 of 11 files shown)


infrastructure/tooling/src/publish/index.js
```diff
@@ -37,21 +37,22 @@ class Publish {
...
-        if (!process.env[e]) {
-          throw new Error(`$${e} is required (unless --no-push)`);
-        }
-      });
+    const expectedVars = [];
+    expectedVars.push('GH_TOKEN');
+    if (!this.cmdOptions.staging) {
+      expectedVars.push('NPM_TOKEN');
+      expectedVars.push('PYPI_USERNAME');
+      expectedVars.push('PYPI_PASSWORD');
```
imbstack, line 45: iirc there are ways to do staging releases for npm/pypi as well? We don't have to do it now but maybe in the space future we can stage releases to all of these or something.

djmitche: For pypi, yes -- there's a `test.pypi.org`.  Maybe NPM, too?  But honestly pushing to docker hub would probably be the most useful..

PR #3103, opened 2020-06-22
Migrate denylisted_notification_entities to denylisted_notification

### LINKED ISSUE #2934: Port the denylisted_notification_entities table to use a normal DB table
This is just denylisted_notification_entities.  Migrating this should be easy.

https://bugzilla.mozilla.org/show_bug.cgi?id=1630018

petemoore:
Github Bug/Issue: Fixes #2934


db/access.yml
```diff
@@ -26,6 +26,7 @@ index:
 notify:
   tables:
     denylisted_notification_entities: write
```
helfi92, line 28: I think we can remove the old table name from here.

petemoore: @helfi92 Could it be needed if the db gets downgraded?

helfi92: If a downgrade is necessary, I believe we will revert the PR first before running the downgrade script. At that point, this line will have been re-added.

db/versions/0015.yml
```diff
@@ -0,0 +1,104 @@
...
+      begin
+        if not condition is null then
+          raise exception 'no condition support in denylisted_notification_entities_scan';
+        else
+          return query select
+            encode_string_key(denylisted_notification.notification_type) as partition_key,
+            encode_string_key(denylisted_notification.notification_address) as row_key,
+            jsonb_build_object(
+              'PartitionKey', encode_string_key(denylisted_notification.notification_type),
+              'RowKey', notification_address,
```
helfi92, line 85: This should be encoded.

petemoore: Thanks @helfi92!

I based it on [this line](https://github.com/taskcluster/taskcluster/blob/6be5615aea7570ecaaa78ad62ba110f1ca1eaf49/db/versions/0009.yml#L192) - was that wrong?

helfi92: Yeah, looks like it was wrong. The good news is that this method is not used in production.

db/test/versions/0015_test.js
```diff
@@ -0,0 +1,54 @@
+const _ = require('lodash');
+const helper = require('../helper');
+const testing = require('taskcluster-lib-testing');
+const Entity = require('taskcluster-lib-entities');
+
+const THIS_VERSION = 15;
```
helfi92, line 6: Neat trick to get the version number from the file name: https://github.com/taskcluster/taskcluster/pull/3108

db/versions/0015-downgrade.sql
```diff
@@ -0,0 +1,29 @@
...
+  -- lock this table before reading from it, to prevent loss of concurrent
+  -- updates when the table is dropped.
+  lock table denylisted_notification;
+
+  create table denylisted_notification_entities(
+    partition_key text, row_key text,
+    value jsonb not null,
+    version integer not null,
+    etag uuid default public.gen_random_uuid());
+  alter table cache_purges_entities add primary key (partition_key, row_key);
```
helfi92, line 11: Rename table name.

petemoore:
The _create and _load functions seem to be ok:

In local testing, here I clear out the `denylisted_notification` table:

```
postgres=# delete from denylisted_notification;
DELETE 1
```

Then I recreate the `_create` function:

```
postgres=# create or replace function                                                                          
          "denylisted_notification_entities_create"(pk text, rk text, properties jsonb, overwrite boolean, version integer)
          returns uuid
          as $$declare
  new_row denylisted_notification%ROWTYPE;
begin
  select
    (properties ->> 'notificationType')::text as notification_type,
    (properties ->> 'notificationAddress')::text as notification_address,
    public.gen_random_uuid() as etag
  into new_row;
  if overwrite then
    raise exception 'overwrite not implemented';
  else
    execute 'insert into denylisted_notification select $1.*' using new_row;
  end if;
  return new_row.etag;
end$$
          language plpgsql
;
CREATE FUNCTION
```

Then I recreate the `_load` function:

```
postgres=# create or replace function                                                                                                     "denylisted_notification_entities_load"(partition_key text, row_key text)
          returns table (partition_key_out text, row_key_out text, value jsonb, version integer, etag uuid)
          as $$begin
  return query
  select
    denylisted_notification_entities_load.partition_key,
    denylisted_notification_entities_load.row_key,
    jsonb_build_object(
      'PartitionKey', denylisted_notification_entities_load.partition_key,
      'RowKey', denylisted_notification_entities_load.row_key,
      'notificationType', notification_type,
      'notificationAddress', notification_address) as value,
    1 as version,
    denylisted_notification.etag as etag
  from denylisted_notification
  where
    denylisted_notification.notification_type = decode_string_key(denylisted_notification_entities_load.partition_key) and denylisted_notification.notification_address = decode_string_key(denylisted_notification_entities_load.row_key);
end$$
          language plpgsql
;
CREATE FUNCTION
```

Then I check that `_load` doesn't return anything:

```
postgres=# select * from "denylisted_notification_entities_load"('email', 'success!40simulator.amazonses.com');
 partition_key_out | row_key_out | value | version | etag 
-------------------+-------------+-------+---------+------
(0 rows)
```

Then I insert data using the new `_create` function:

```
postgres=# select * from "denylisted_notification_entities_create"('email', 'success!40simulator.amazonses.com', '{"RowKey": "success!40simulator.amazonses.com", "PartitionKey": "email", "notificationType": "email", "notificationAddress": "success@simulator.amazonses.com"}', false, 1);
 denylisted_notification_entities_create 
-----------------------------------------
 2d73f809-ce6a-4752-a1be-d6a54dca07ad
(1 row)
```

Then I check the `_load` function returns the data I just inserted:

```
postgres=# select * from "denylisted_notification_entities_load"('email', 'success!40simulator.amazonses.com');
 partition_key_out |            row_key_out            |                                                                              value                                                                              | version |                 etag                 
-------------------+-----------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------+---------+--------------------------------------
 email             | success!40simulator.amazonses.com | {"RowKey": "success!40simulator.amazonses.com", "PartitionKey": "email", "notificationType": "email", "notificationAddress": "success@simulator.amazonses.com"} |       1 | 2d73f809-ce6a-4752-a1be-d6a54dca07ad
(1 row)
```

The [failing test](https://github.com/taskcluster/taskcluster/blob/da484e22d232201a675923637a39ad88f0e736e4/services/notify/test/api_test.js#L88-L109) should add an entry to the denylisted_notification table, and then send an email notification that uses the email address that it added to the entry in the denylisted_notification table, so expects the notification not to be sent - however this fails, suggesting either the insert into the database fails, or it isn't returned when the table is subsequently queried. So maybe the `_scan` method is used to return all entries rather than the `_load` method just for the given email address - I'll check the `_scan` method implementation now too...

petemoore:
Ah, I see that test `does not send notifications to denylisted email address` is actually passing - so that was a red herring. However, the next test `email without link` is failing for me, and that looks like it is because the entry in the new table isn't purged between tests - I'll double check if the `_remove` function is broken - that is undoubtedly the cause!

petemoore:
It looks like between tests, the database cleanup executes `truncate denylisted_notification_entities` rather than `truncate denylisted_notification` which is why the entry doesn't get cleaned up.

db/test/versions/0015_test.js
petemoore: Thanks, I've added it.

db/versions/0017.yml
```diff
@@ -102,3 +102,63 @@ methods:
...
+    description: List all denylist addresses for the taskcluster-notify service.
+    mode: read
+    serviceName: notify
+    args: page_size_in integer, page_offset_in integer
+    returns: table (notification_type text, notification_address text)
+    body: |-
+      begin
+        return query select denylisted_notification.notification_type, denylisted_notification.notification_address
+        from denylisted_notification
+        order by 1, 2
```
helfi92, line 148: What does `order by 1, 2` mean?.

libraries/postgres/src/Database.js
```diff
@@ -391,9 +391,10 @@ class Database {
         FROM pg_database 
         WHERE datname = current_database()`);
       const collation = res.rows[0].collation;
-      if (collation !== 'en_US.UTF8') {
+      if (collation !== 'en_US.utf8' && collation !== 'en_US.UTF8') {
         throw new Error(
-          `Postgres database must have default collation en_US.UTF8; this database is using ${collation}.`);
+          `Postgres database must have default collation en_US.utf8 or en_US.UTF8; this database is using ${collation}.`,
+        );
```
helfi92, line 397: This should probably be left out of the PR. Rebasing on master might do the trick.

services/notify/src/api.js
```diff
@@ -173,7 +172,7 @@ builder.declare({
 
   await req.authorize(req.body);
   try {
-    await this.DenylistedNotification.create(address);
+    await this.db.fns.add_denylist_address(address);
   } catch (e) {
     if (e.name !== 'EntityAlreadyExistsError') {
```
helfi92, line 177: Now that we no longer are using tc-lib-entities, we should change the errors to be compatible with postgres. For example, this line should probably be:

```
const { UNIQUE_VIOLATION } = require('taskcluster-lib-postgres');

// ...

if (e.code !== UNIQUE_VIOLATION) {
```

A test that asserts this specific error is thrown when the row already exists might be good as well.

services/notify/src/api.js
```diff
@@ -202,7 +201,7 @@ builder.declare({
 
   await req.authorize(req.body);
   try {
-    await this.DenylistedNotification.remove(address);
+    await this.db.fns.delete_denylist_address(address);
   } catch (e) {
     if (e.name !== 'ResourceNotFoundError') {
```
helfi92, line 206: If row does not exist, we could have `delete_denylist_address` throw error code `P0002` with error message 'no such row' then this line could be changed to:

```
if (e.code !== 'P0002') {
```

https://github.com/taskcluster/taskcluster/search?q=P0002&unscoped_q=P0002

services/notify/test/api_test.js
```diff
@@ -4,8 +4,12 @@ const helper = require('./helper');
 const testing = require('taskcluster-lib-testing');
 
 helper.secrets.mockSuite(testing.suiteName(), ['db', 'aws'], function(mock, skipping) {
+  // Temporary workaround for
+  //   https://github.com/taskcluster/taskcluster/issues/3131
+  if (mock) {
+    return;
+  }
```
helfi92, line 11: No need for this condition. I think as long as TEST_DB_URL is defined as an environment variable, the services will use the local db to run its tests.

db/versions/0017.yml
```diff
@@ -102,3 +102,63 @@ methods:
...
+    args: address_in jsonb
+    returns: void
+    body: |-
+      begin
+        insert into denylisted_notification(notification_type, notification_address, etag)
+        values (
+          (address_in ->> 'notificationType')::text,
+          (address_in ->> 'notificationAddress')::text,
+          public.gen_random_uuid()
+        ) on conflict do nothing;
```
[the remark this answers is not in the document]
petemoore: If there is a conflict, it means an entry being has the same notification address and notification type, but a different etag. Since there are no other fields, keeping the existing entry is sufficient, and therefore no change is required. Since the operation should be idempotent, we can avoid throwing an error.

db/versions/0017.yml
petemoore: It means to order by the first column (`denylisted_notification.notification_type`), and then the second column (`denylisted_notification.notification_address`).

libraries/postgres/src/Database.js
petemoore: Ah, has an equivalent fix landed on master?

services/notify/src/api.js
petemoore: Ah, nice spot, thanks!

services/notify/src/api.js
petemoore: Ah yes, good point!

services/notify/test/api_test.js
petemoore: My experience when running tests locally was that the mocks were still run, but I can try again tomorrow. Once I added this, the tests just ran against the real database.

db/test/versions/0017_test.js
```diff
@@ -0,0 +1,54 @@
...
+    entityTableName: 'denylisted_notification_entities',
+    newTableName: 'denylisted_notification',
+    EntityClass: DenylistedNotification,
+    samples: {},
+    loadConditions: [],
+    scanConditions: [],
+    notFoundConditions: [],
+    notImplemented: [],
+    modifications: [],
+  });
```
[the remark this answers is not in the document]
petemoore: Would this not be covered by the tests in services/notify/test/api_test.js?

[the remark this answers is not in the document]
helfi92: The tests in `testEntityTable` will give us confidence that most of the edge cases are covered. Not all services have good coverage.

libraries/postgres/src/Database.js
helfi92: Ah. Forget what I said. I just realized it was simply a change to an error message :)

services/notify/test/api_test.js
helfi92: I just tried locally and I can confirm that mocks are not run when TEST_DB_URL is specified.

db/versions/0017.yml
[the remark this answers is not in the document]
helfi92: From what I can see, the source code expects an error to be thrown when the row already exists. It's surrounded by a try catch statement and currently checks for `EntityAlreadyExistsError`.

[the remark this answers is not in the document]
djmitche: True, but in that case it just returns -- the overall effect of the SQL + JS is to ignore conflicts.  This is the idempotency that Pete refers to above.  I think implementing that idempotency at the DB level is fine, and means that the try/catch can be removed in the JS code that calls the DB function.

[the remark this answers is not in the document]
djmitche: It's worth putting a sentence in the function's `description` to indicate that, though.

services/notify/src/api.js
djmitche: It looks like the JS catches that error just to ignore it, much like adding a row, in order to achieve idempotency.  So I think the same logic applies: if the delete just ignores the case when the row doesn't exist, then that's simpler SQL and simpler JS.

services/notify/test/api_test.js
petemoore: I tried removing, but then I get CI failures, so I've moved it into a separate third commit. I think when we tackle #3131 we can also revert this third commit that just takes care of skipping fake db tests.

djmitche: Sounds fine..

petemoore: I've rebased now, and it looks like the third commit is no longer needed after all, so maybe I was missing something before. In any case I'll drop the third commit now...

services/notify/src/api.js
petemoore: This code disappeared in the end due to https://github.com/taskcluster/taskcluster/pull/3103#discussion_r447998772

db/versions/0017.yml
```diff
@@ -0,0 +1,166 @@
...
+      end
+  delete_denylist_address:
+    description: |-
+      Delete a denylist address for the taskcluster-notify service.
+      Returns number of rows deleted (0 or 1).
+    mode: write
+    serviceName: notify
+    args: address_in jsonb
```
djmitche, line 128: This is sort of a shortcut to named arguments.  But when we add named arguments, it will need to be called as `db.fns.delete_denylist_address({address_in: {notificationType: 'abc', notificationAddress: 'def'}})`.  Is that desirable?

db/versions/0017.yml
```diff
@@ -0,0 +1,166 @@
...
+      begin
+        delete from denylisted_notification where
+        denylisted_notification.notification_type = (address_in ->> 'notificationType')::text and
+        denylisted_notification.notification_address = (address_in ->> 'notificationAddress')::text;
+        if found then
+          return 1;
+        end if;
+        return 0;
+      end
+  all_denylist_address:
```
djmitche, line 140: The singular here reads awkwardly to me.  Maybe pluralize it?

db/versions/0017-migration.sql
```diff
@@ -0,0 +1,25 @@
+begin
+  -- lock this table before reading from it, to prevent loss of concurrent
+  -- updates when the table is dropped.  Note that this may lead to concurrent
+  -- updates failing; the important thing is that they not succeed without
+  -- taking effect.  Failed updates will be retried.
+  lock table denylisted_notification_entities;
+
+  create table denylisted_notification
```
djmitche, line 8: Most (all?) of our non-entities tables have plural names -- `github_builds`, `tasks`, `worker_pools`.  For consistency, let's do the same with this table.

libraries/postgres/src/Database.js
djmitche: I wonder what's behind this difference in case, but ultimately I think the two mean the same thing, so this looks good.

services/notify/src/denier.js
```diff
@@ -12,7 +12,9 @@ class Denier {
     }
 
     const address = {notificationType, notificationAddress};
-    return !!await this.DenylistedNotification.load(address, true);
+    let p = await this.db.fns.exists_denylist_address(address);
```
djmitche, line 15: If this `p` is for promise, after the `await` it's no longer a promise but an array of rows.  No need to change anything, just a note.

libraries/postgres/src/Database.js
petemoore: Yeah, I'm not sure. The test failed for me locally using the postgres docker image because the value was `en_US.utf8` instead of the expected `en_US.UTF8`. If you don't have the same issue with the postgres docker container, perhaps something in my environment is influencing it. I have `LC_CTYPE=UTF-8` in my environment, but that is already capitalised. In any case, I agree it should be safe to assume `en_US.utf8` and `en_US.UTF8` refer to the same encoding. I think the official name includes the hyphen (`UTF-8`) so I think neither of the values are technically correct. But I digress... :-)

db/versions/0017.yml
petemoore: I've changed it, but in general I'm not sure there is value in mapping all fields to unique database columns, unless those columns join to other columns in database queries. Having the data structures exploded means the database and the application need to keep track of the data structures, rather than them just being managed by the json schemas, so leads to additional boilerplate and introduces risk of divergence between the database model and the jsonschema definitions.

db/versions/0017.yml
petemoore: ok

services/notify/src/denier.js
petemoore: I've given it a more descriptive name now.

db/versions/0017-migration.sql
petemoore: ok

db/versions/0017.yml
petemoore: OK, I've changed it.

services/notify/src/main.js
```diff
@@ -136,9 +125,9 @@ const load = loader({
   },
 
   denier: {
-    requires: ['cfg', 'DenylistedNotification'],
-    setup: ({cfg, DenylistedNotification}) =>
-      new Denier({DenylistedNotification, emailBlacklist: cfg.app.emailBlacklist}),
+    requires: ['cfg', 'db'],
+    setup: ({cfg, db}) =>
+      new Denier({emailBlacklist: cfg.app.emailBlacklist, db: db}),
```
helfi92, line 130: This could have been written more briefly with `new Denier({emailBlacklist: cfg.app.emailBlacklist, db}),`. There's nothing wrong with ^ however so no need to make any changes.

helfi92:
Filed https://github.com/taskcluster/taskcluster/issues/3147. I'm not able to currently release a new TC version.

Issue #3089, opened 2020-06-17
Port the namespaces_entities table to use a normal DB table


helfi92:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1626314#c26, `namespaces_entities` has about 1592344 rows.

helfi92:
Folded in https://github.com/taskcluster/taskcluster/issues/2935.

Issue #2935, opened 2020-05-29
port the index service to use normal DB table


djmitche:
These tables can be quite large, so consider finding ways to handle them efficiently or do the migration in a TCW, perhaps with #2932.

https://bugzilla.mozilla.org/show_bug.cgi?id=1630017

helfi92:
There are two tables to migrate for this service. From https://bugzilla.mozilla.org/show_bug.cgi?id=1626314#c26:
* indexed_tasks_entities ~7755715 rows
* namespaces_entities  ~1592344 rows

helfi92:
We probably don't need to consider namespaces_entities a "large" table. Assuming a migration of 40k rows per second, the full upgrade for that table would take ~40 seconds. As long as the hang is <1 min, we should be fine to migrate this outside of a TCW.

PR #3105, opened 2020-06-23
Migrate index service to postgres phase 2

helfi92:
Resolves https://github.com/taskcluster/taskcluster/issues/2935.
Resolves https://github.com/taskcluster/taskcluster/issues/3089.


services/index/src/data.js
```diff
@@ -1,198 +1,623 @@
...
-    parent: Entity.types.String,
-    name: Entity.types.String,
-    expires: Entity.types.Date,
-  },
-});
+const make404 = () => makeError('Resource not found', 'ResourceNotFound', 404);
 
-// Export Namespace
-exports.Namespace = Namespace;
+class IndexedTask {
```
imbstack, line 18: As I understand it we only need these classes if they're getting passed around a lot and not just CRUD'ed inside api endpoints. Could we get away with not having these in index?

db/versions/0017-downgrade.sql
```diff
@@ -0,0 +1,64 @@
+begin
+  -- lock this table before reading from it, to prevent loss of concurrent
+  -- updates when the table is dropped.
+  lock table indexed_tasks;
+
+  create table indexed_tasks_entities(
+    partition_key text, row_key text,
+    value jsonb not null,
+    version integer not null,
+    etag uuid default public.gen_random_uuid());
```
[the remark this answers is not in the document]
helfi92: It does not. We should add a `not null` here. Good observation!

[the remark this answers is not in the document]
djmitche: None of the entities tables have `not null` for etag.  I'm not sure why, but the downgrade needs to restore the existing state so this is correct as-is.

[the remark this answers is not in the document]
helfi92: Ah that's right. I thought they were `not null`. Thanks!

services/index/src/data.js
helfi92: Done.

db/versions/0017.yml
```diff
@@ -0,0 +1,699 @@
...
+            when (size is not null and size > 0 and page is not null and page > 0) then page
+            else 0
+          end;
+        end if;
+      end
+  get_indexed_tasks:
+    description: |-
+      Get existing indexed tasks filtered by the optional arguments,
+      ordered by the `namespace` and `name`.
+      If the pagination arguments are both NULL, all rows are returned.
```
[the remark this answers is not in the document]
djmitche: Sorry, I'm confused -- that last one is equivalent to `get_indexed_task`, right?

db/versions/0017.yml
```diff
@@ -0,0 +1,699 @@
...
+
+        raise exception 'no such row' using errcode = 'P0002';
+      end
+  update_indexed_task:
+    serviceName: index
+    description: |-
+      Update an indexed task.
+      Returns the up-to-date indexed task row that have the same namespace and name.
+      If the etag argument is empty then the update will overwrite the matched row.
+      Else, the function will fail if the etag is out of date. This is useful for concurency handling.
```
djmitche, line 627: I only see this called in one place, and it seems that the JS doesn't catch this error.  Is the eventual consistency necessary here?

services/index/src/api.js
```diff
@@ -57,35 +54,31 @@ builder.declare({
   let namespace = indexPath.join('.');
 
   // Load indexed task
-  let tasks;
   try {
-    tasks = await this.IndexedTask.query({
-      namespace: namespace,
-      name: name,
-      expires: Entity.op.greaterThan(new Date()),
-    });
```
djmitche, line 66: I think this only used `query` here instead of `load` to validate that the row had not expired.  There can only ever be one row.  So, I think this could be translated to a call to `get_indexed_task`.  No pagination required..

Also, the updated code should check the expiration in JS.

services/index/src/api.js
```diff
@@ -57,35 +54,31 @@ builder.declare({
...
+      { query: req.query },
+    );
+    if (!rows.length) {
+      return res.reportError('ResourceNotFound', 'Indexed task not found', {});
+    }
+    const task = rows[0];
+    return res.reply(helpers.taskUtils.serialize(task));
   } catch (err) {
     // Re-throw the error, if it's not a 404
     if (err.code !== 'ResourceNotFound') {
```
djmitche, line 70: I don't think anything in `getIndexedTasks` would throw this error -- and the not-found case is handled above (`if (!rows.length)`).

services/index/src/helpers.js
```diff
@@ -1,111 +1,360 @@
...
+const makeError = (message, code, statusCode) => {
+  const err = new Error(message);
+  err.code = code;
+  err.name = `${code}Error`;
+  err.statusCode = statusCode;
+  return err;
+};
+
+exports.make404 = () => makeError('Resource not found', 'ResourceNotFound', 404);
+exports.makeMaxModifyAttempts = () => makeError('MAX_MODIFY_ATTEMPTS exhausted, check for congestion', 'EntityWriteCongestionError', 409);
```
djmitche, line 20: It looks like this and MAX_MODIFY_ATTEMPTS are unused

services/index/src/helpers.js
```diff
@@ -1,111 +1,360 @@
...
+          expires,
+        ).then(async function() {
+          const etag = (await db.fns.create_indexed_task(
+            namespace,
+            name,
+            input.rank,
+            input.taskId,
+            input.data,
+            expires,
+          ))[0].create_indexed_task;
```
djmitche, line 132: It looks like all of this is kind of an elaborate "upsert" operation, with complications that we only want to replace an existing entity if its rank is lower, and that we need to ensure the namespace exists if it does not already.  I suspect all of that could be done more simply directly in the DB, but maybe that's a bit of a heavy lift for now -- we could leave that to a follow-up.

services/index/src/helpers.js
```diff
@@ -1,111 +1,360 @@
...
+  // `rows` field set to an empty array.
+  //
+  // If a handler is supplied, then it will invoke the handler
+  // on every item of the scan.
+  async getIndexedTasks(
+    db,
+    { namespace, name },
+    {
+      query,
+      handler,
```
djmitche, line 161: It looks like the handler option is never used..

services/index/src/helpers.js
```diff
@@ -1,111 +1,360 @@
...
+  // Call db.get_index_namespaces with named arguments.
+  // You can use this in two ways: with a handler or without a handler.
+  // In the latter case you'll get a list of up to 1000 entries and a
+  // continuation token.
+  // The response will be of the form { rows, continationToken }.
+  // If there are no namespaces to show, the response will have the
+  // `rows` field set to an empty array.
+  //
+  // If a handler is supplied, then it will invoke the handler
+  // on every item of the scan.
```
djmitche, line 246: This handler also appears unused.

services/index/src/helpers.js
```diff
@@ -1,111 +1,360 @@
...
+    if (!(namespace instanceof Array)) {
+      namespace = namespace.split('.');
+    }
+    // Find parent and folder name
+    let name = namespace.pop() || '';
+    let parent = namespace.join('.');
+
+    // Load namespace, to check if it exists and if we should update expires
+    return db.fns.get_index_namespace(parent, name)
+      .then(async function(f) {
```
djmitche, line 320: Mixing `.then` and `async` is a little scary -- perhaps convert the whole thing to `async`?

db/versions/0017.yml
```diff
@@ -0,0 +1,699 @@
...
+          indexed_tasks.etag
+        from indexed_tasks
+        where
+          indexed_tasks.namespace = namespace_in and
+          indexed_tasks.name = name_in and
+          indexed_tasks.expires > now()
+        into new_row;
+        if found then
+          return query select new_row.namespace, new_row.name, new_row.rank, new_row.task_id, new_row.data, new_row.expires, new_row.etag;
+          return;
```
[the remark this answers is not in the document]
helfi92: > Why the double select here?

I'm not sure how to write it in a single select and make sure a "no such row error" is thrown when the row is not there.

> And since this returns a table (and thus an array of rows in JS) anyway, why not just return an empty set on not found?

The reason for that were:
* less refactoring of the index service to reduce the chance of introducing a bug
* we throw a 'no such row' error in `update_{worker|indexed_task}` when the row doesn't exist rather than return an empty row so I figured there shouldn't be an issue to extend this to this db function.

Thoughts?

db/versions/0017.yml
helfi92: Probably not. I will remove it.

services/index/src/api.js
helfi92: > So, I think this could be translated to a call to get_indexed_task. No pagination required..

Sounds good, makes sense.

> Also, the updated code should check the expiration in JS.

Why should it check the expiration in JS? All of the invocations of load and query for the indexed task have `expires: Entity.op.greaterThan(new Date()),` in them. Having the check in the JS will require some refactoring which may introduce bugs. Thoughts?

db/versions/0017.yml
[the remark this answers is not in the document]
helfi92: > Sorry, I'm confused -- that last one is equivalent to get_indexed_task, right?

Right.

db/versions/0017.yml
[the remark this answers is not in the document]
djmitche: For the other `get_` methods, we've just returned the select and included a check on the length of the returned value:
https://github.com/taskcluster/taskcluster/blob/6c871ced2f2ebec3751c368b7e2cc459a799f3c2/services/worker-manager/src/data.js#L45-L47
That ends up with a simpler `get_` function, and the JS is a little shorter too (no `try/catch/err.code===`).

services/index/src/api.js
djmitche: It's a pretty minor refactor -- just adding an `if` statement, right?  And this is well-covered by tests, at least the normal (non-expired) case.  Getting an expired row is definitely the unusual case here.

To be honest, even with Azure I'm not sure why we used `.query` instead of `.load` with a check of the expires value.

helfi92:
Continued in https://github.com/taskcluster/taskcluster/pull/3141 and https://github.com/taskcluster/taskcluster/pull/3150.

Issue #2932, opened 2020-05-29
port the queue's task-tracking tables to use normal DB tables


djmitche:
This involves 7 tables:

* queue_task_group_active_sets_entities:
* queue_task_requirement_entities:
* queue_task_groups_entities:
* queue_tasks_entities:
* ~queue_artifacts_entities:~ #3112 
* queue_task_group_members_entities:
* queue_task_dependency_entities:

The queue relies on the details of Azure's semantics for correctness around these tables, so care will be required in ensuring that the replacements make the same guarantees.

Some of these tables are quite large.  Consider ways to make the migration not block operations for too long, or timing the deployment of the migration for a TCW.

https://bugzilla.mozilla.org/show_bug.cgi?id=1630020

helfi92:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1626314#c26, we can identify the small tables:

* queue_provisioner_entities: 38
* queue_worker_type_entities: 133
* queue_task_group_active_sets: 24818
* queue_worker_entities: 90135
* queue_task_groups_entities: 469297
* queue_task_group_members_entities: > 1M
* queue_task_requirement: 837635
* queue_task_dependency_entities: > 1M
* queue_tasks_entities: > 1M
* queue_artifacts_entities: > 1M

helfi92:
It looks like `queue_task_group_active_sets_entities` may be handled separately since we don't need to consider ways to make this table not block operations.

helfi92:
I took out `queue_task_group_active_sets_entities` from this list just in case we do want to do the migration of the large tables during the TCW in which case let's not squeeze in more work than is necessary.

Filed https://github.com/taskcluster/taskcluster/issues/3084 to port `queue_task_group_active_sets_entities`.

helfi92:
I went ahead and took out `queue_task_groups_entities` and `queue_task_requirement_entities` from this list. The migration time for these tables will most probably be acceptable.

Filed https://github.com/taskcluster/taskcluster/issues/3084 to port `queue_task_group_active_sets_entities`.
Filed https://github.com/taskcluster/taskcluster/issues/3087 to port `queue_task_requirement_entities`.

djmitche:
Oops, this is the bigger tables so should wait until nearer the TCW.

helfi92:
I think it's fine to work on this. We just won't merge until nearer the TCW.

djmitche:
[...]

PR #3107, opened 2020-06-23
add uuid_to_slugid and slugid_to_uuid db functions

djmitche:
This is prep for #2932 but will be useful for other migrations, too.

### DIFF
--- a/changelog/U1wFteWOS-aP0VkZF3C1hQ.md
+++ b/changelog/U1wFteWOS-aP0VkZF3C1hQ.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
--- a/db/test/versions/0016_test.js
+++ b/db/test/versions/0016_test.js
@@ -1,0 +1,56 @@
+const _ = require('lodash');
+const helper = require('../helper');
+const testing = require('taskcluster-lib-testing');
+const assert = require('assert').strict;
+const slugid = require('slugid');
+
+suite(testing.suiteName(), function() {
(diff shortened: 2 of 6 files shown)


petemoore:
Would these functions just be used during the migration, and when the migration complete, everything uses the slugid form, and we would delete the functions?

djmitche:
Yes, the idea is to store everything as slugid strings and eventually these functions will not be necessary.

db/versions/0015-downgrade.sql
```diff
@@ -0,0 +1,4 @@
+begin
+  drop function uuid_to_slugid(uuid text);
+  drop function slugid_to_uuid(uuid text);
```
helfi92, line 3: Argument name for the second function should be `slugid`.

db/test/versions/0015_test.js
```diff
@@ -0,0 +1,56 @@
+const _ = require('lodash');
+const helper = require('../helper');
+const testing = require('taskcluster-lib-testing');
+const assert = require('assert').strict;
+const slugid = require('slugid');
+
+suite(testing.suiteName(), function() {
+  const THIS_VERSION = 15;
```
helfi92, line 8: ` const THIS_VERSION = parseInt(/.*\/0*(\d+)_test\.js/.exec(__filename)[1]);`

djmitche: I'll rebase, now that a 15 has landed :)

PR #3109, opened 2020-06-24
Add sha512 db function to replace Entity.keys.HashKey

helfi92:


djmitche:
oops, looks like it needs ye olde `yarn generate` before merging :)

Issue #2937, opened 2020-05-29
port the github service's tables to use normal DB tables 


djmitche:
https://bugzilla.mozilla.org/show_bug.cgi?id=1630015

helfi92:
Tables to migrate:
* taskcluster_github_builds_entities: 23677 
* taskcluster_integration_owners_entities: 13
* taskcluster_checks_to_tasks_entities: 678358
* taskcluster_check_runs_entities: 678364

The number of rows are an estimation. They were taken from https://bugzilla.mozilla.org/show_bug.cgi?id=1626314#c26. It looks like they are all relatively small tables.

helfi92:
Feel free to break this into multiple PRs if that makes it easier.

PR #3113, opened 2020-06-25
Migrate github builds table

imbstack:
Github Bug/Issue:  #2937 (fixes 1/4 of this)


db/test/versions/0015_test.js
```diff
@@ -0,0 +1,172 @@
...
+    eventType: Entity.types.String,
+    eventId: Entity.types.String,
+  },
+  migrate: function(item) {
+    // Delete Id because it was mistakenly set in the last migration
+    delete item.Id; // Just in case the migrations are chained
+    item.eventId = item.eventId || 'Unknown';
+    return item;
+  },
+});
```
djmitche, line 99: You can shorten this to a single configure call with the final set of properties.  The `version` and `migrate` fields are ignored anyway.

db/versions/0015.yml
```diff
@@ -0,0 +1,186 @@
...
+              'updated', updated,
+              'installationId', installation_id,
+              'eventType', event_type,
+              'eventId', event_id) as value,
+              1 as version,
+          github_builds.etag as etag
+          from github_builds
+          where
+            partition_key is null or
+            task_group_id = decode_string_key(task_group_id)
```
djmitche, line 100:
```suggestion
            task_group_id = decode_string_key(partition_key)
```
It looks like no calls to `.scan` or `.query` pass a partition key, so it's not too surprising this didn't cause a test failure, and wouldn't cause a failure in production either.  You could verify by adding a check that partition_key is null in this function.

imbstack: Made this change in my branch

changelog/issue-2937.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
+level: silent
```
helfi92, line 2: I think this kind of migration would be important for deployers to be aware of. Should wee also make the level `minor` rather than `silent`?

db/src/fakes/github.js
```diff
@@ -18,6 +19,7 @@ class FakeGithub {
     this.taskclusterIntegrationOwners = new Map();
     this.taskclusterChecksToTasks = new Map();
     this.taskclusterCheckRuns = new Map();
+    this.github_builds = new Map();
```
helfi92, line 22: Let's use camel case to align with JS conventions and the other names in this file.

db/versions/0017-migration.sql
```diff
@@ -0,0 +1,42 @@
...
+  -- taking effect.  Failed updates will be retried.
+  lock table taskcluster_github_builds_entities;
+
+  create table github_builds
+  as
+    select
+      (value ->> 'organization')::text as organization,
+      (value ->> 'repository')::text as repository,
+      (value ->> 'sha')::text as sha,
+      (value ->> 'taskGroupId')::text as task_group_id,
```
helfi92, line 15: I think Dustin recommended moving away from azure's way of storing things as uuids but rather store things as slug ids. We'll want to modify the migration and downgrade script to make sure we have the proper slugid -> uuid and uuid -> slugid mappings.

db/versions/0017.yml
```diff
@@ -0,0 +1,287 @@
...
+              'eventId', event_id) as value,
+              1 as version,
+          github_builds.etag as etag
+          from github_builds
+          where
+            partition_key is null or
+            task_group_id = decode_string_key(partition_key)
+          order by task_group_id
+          limit size + 1
+          offset page;
```
helfi92, line 103: If size or offset are not defined then something may go wrong here. One way to fix that is to do:

```
        limit case
          when (size is not null and size > 0) then size + 1
          else null
        end
        offset case
          when (page is not null and page > 0) then page
          else 0
        end;
```

helfi92:
@djmitche Did you want to take a peek at this?

db/versions/0017-migration.sql
```diff
@@ -0,0 +1,42 @@
...
+    alter column task_group_id set not null,
+    alter column state set not null,
+    alter column created set not null,
+    alter column updated set not null,
+    alter column installation_id set not null,
+    alter column event_type set not null,
+    alter column event_id set not null,
+    alter column etag set not null,
+    alter column etag set default public.gen_random_uuid();
+
```
djmitche, line 38: Since `get_github_builds` and `github.builds(..)` can filter by any of organization, repository, and sha, it probably makes sense to add indexes for each of those columns -- otherwise we'll be doing a full table scan for any request there.  The table is only 37240 rows at the moment, but doesn't expire, so that could get pretty expensive.

db/src/fakes/github.js
```diff
@@ -373,6 +375,92 @@ class FakeGithub {
...
+
+  async create_github_build(organization, repository, sha, task_group_id, state,
+    created, updated, installation_id, event_type, event_id) {
+    assert.equal(typeof organization, 'string');
+    assert.equal(typeof repository, 'string');
+    assert.equal(typeof sha, 'string');
+    assert.equal(typeof task_group_id, 'string');
+    assert.equal(typeof state, 'string');
+    assert(created.toTimeString); // duck type a date
+    assert(updated.toTimeString); // duck type a date
```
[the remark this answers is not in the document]
imbstack: It does but if this is used in `zurvan` it will not. At least that's my memory from how azure-entities does it.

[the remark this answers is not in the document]
djmitche: ah, good point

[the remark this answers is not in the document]
helfi92: What's `zurvan`?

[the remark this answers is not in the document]
djmitche: It's a utility for faking the node clocks -- `new Date()`, `Date.now`, `setTimeout`, etc.  It's not compatible with a real DB, though, so we won't be using it much.

db/versions/0017-migration.sql
djmitche: Did this get missed?

imbstack: Ah yeah

imbstack: I added an index on `(org, repo, sha)` since although you're allowed to just search for a sha or a repo without an org, you generally won't want to.

djmitche: Makes sense.  We should disallow that, then -- either in the db fn or in the API method.  We could do that in a followup if you'd prefer.

imbstack:
Added some limiting to the api endpoint

PR #3114, opened 2020-06-25
add a 'toDbVersion' helper, other db testing helpers

djmitche:
pulled out of #2932

edit: whoops, I failed at `git add -p`.  This has a few more things in it.

### DIFF
--- a/db/test/helper.js
+++ b/db/test/helper.js
@@ -16,6 +16,7 @@ assert(exports.dbUrl, "TEST_DB_URL must be set to run db/ tests - see dev-docs/d
  * - helper.withDbClient(fn) to call fn with a pg client.
  * - helper.upgradeTo(v) to upgrade to the given version.
  * - helper.downgradeTo(v) to downgrade to the given version.
+ * - helper.toDbVersion(v) to upgrade or downgrade as necessary to the given version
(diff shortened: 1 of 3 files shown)


djmitche:
* add a 'toDbVersion' helper function to upgrade/downgrade as necessary
* testEntityTable now supports a second customTests argument to add
      additional non-automated tests that still get to use the sample data
* suites are more particular about DB versions, for more accurate tests
      when using `mocha --grep`

db/test/helper.js
```diff
@@ -271,9 +288,10 @@ exports.testEntityTable = ({
   // it is treated as a collection of modifier functions to run in parallel to check
   // support for concurrent modifications.
   modifications,
-}) => {
+  // customTests can define additional test cases in the usual Mocha style.  The
+  // parameter is true if the tests are run for THIS_VERSION, otherwise for PREV_VERSION.
+}, customTests = (isThisVersion) => {}) => {
```
helfi92, line 293: `isThisVersion` looks like it's currently unused.

djmitche: It's used in test cases where customTests is given (#3115).  It's here just to illustrate.

djmitche:
> It would probably be a good idea to add a tiny section about this in the migration document.

Good point.  The uuid functions, too.

PR #3117, opened 2020-06-26
Some fixes to purge cache DB migration

petemoore:
Some things I found as I was going through the purge cache migration.


db/src/fakes/purge_cache.js
```diff
@@ -101,7 +101,7 @@ class FakePurgeCache {
       });
   }
 
-  expire_cache_purges(exp) {
+  async expire_cache_purges(exp) {
```
helfi92, line 104: `async` is only needed for when the body of the function uses `await`. No need for this change.

db/versions/0009.yml
```diff
@@ -189,7 +189,7 @@ methods:
             encode_string_key(cache_purges.cache_name) as row_key,
             jsonb_build_object(
               'PartitionKey', encode_composite_key(cache_purges.provisioner_id, cache_purges.worker_type),
-              'RowKey', cache_name,
+              'RowKey', encode_string_key(cache_name),
```
helfi92, line 192: Thanks :)

db/src/fakes/purge_cache.js
helfi92: It doesn't hurt to keep them so I guess we can keep it.

db/versions/0009.yml
djmitche: Note that typically we can't fix old migrations, as they've already run and won't run again.  It looks like we got away with it in this case (I have a sneaking suspicion that the PartitionKey and RowKey JSON properties are never used?), and fixing this will help anyone referring to this version to not duplicate the bug.

PR #3118, opened 2020-06-26
Issue #3116 - Enforce en_US.UTF8

### LINKED ISSUE #3116: Ensure postgres is using a fixed collation
This seems to be different in CI and locally.

It can be checked with
```
SELECT datcollate AS collation
FROM pg_database 
WHERE datname = current_database();
```

so let's check it.

djmitche:
This implements the "enforce" option from #3116.

Github Bug/Issue: Fixes #3116

### DIFF
--- a/changelog/issue-3116.md
+++ b/changelog/issue-3116.md
@@ -1,0 +1,6 @@
+audience: deployers
+level: patch
+reference: issue 3116
+---
+The db upgrade and downgrade scripts now verify that the default database collation is `en_US.UTF8`.  No other collation is allowed.
+Unfortunately, changing the default collation requires dumping and re-creating the database.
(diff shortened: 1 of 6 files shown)


djmitche:
It turns out the docker image didn't even support en_US.UTF8 (!).  So I built a new image and pushed it and killed all of the running workers, so this PR should be using that new image.

helfi92:
Interesting. Out of curiosity, why didn't it support en_US.UTF8? Also, what change did you do to make it support it?

djmitche:
> why didn't it support en_US.UTF8?

It looks like the default install of postgres on Debian doesn't support it, and that's all the docker image creation was doing.

> Also, what change did you do to make it support it?

I looked at the image history for the official postgres image, and it has a few commands to install `locales` and generate an en_US.UTF8 locale.  I copied those and put them in the docker-image generation script in this PR.

Still working on getting it to actually pass, but I think I'm almost there.

djmitche:
Affected tests are all green..

PR #3119, opened 2020-06-26
Prioritize decision task above other tasks

djmitche:


djmitche:
I thought all of the tasks were the same priority, but in fact decision was "lowest" while everything else was "highest"!

PR #3123, opened 2020-06-26
Allow undefined properties in per-request context

djmitche:
This gives a much clearer error for things like `ctx.Foo`.


helfi92:
> This gives a much clearer error for things like ctx.Foo.

Which `ctx.Foo` errors?

djmitche:
There's a bug here, actually, that's causing the queue failures.

djmitche:
That's better!

Issue #3030, opened 2020-06-11
Port wmworker_pool_errors_entities table to use a normal DB table


helfi92:
This table currently uses azure-entities emulation to store its data in the DB. This bug entails writing a new DB version that will smoothly migrate the last table of the service to use a more typical DB table with nicely-named columns and easy joins.

PR #3134, opened 2020-06-29
issue 3030 - port worker pool errors to postgres phase 2

milescrabill:
Github Issue: parts 1 and 2 of #3030


db/tables.yml
```diff
@@ -125,12 +125,15 @@ session_storage_table_entities:
...
-  version: integer not null
-  etag: uuid
+worker_pool_errors:
+  error_id: text not null
+  worker_pool_id: text not null
+  reported: timestamp with time zone not null
+  title: text not null
+  kind: text not null
+  description: text not null
+  extra: jsonb
```
helfi92, line 135: This should probably be `json not null`.

db/versions/0017-migration.sql
```diff
@@ -0,0 +1,34 @@
...
+  -- lock this table before reading from it, to prevent loss of concurrent
+  -- updates when the table is dropped.  Note that this may lead to concurrent
+  -- updates failing; the important thing is that they not succeed without
+  -- taking effect.  Failed updates will be retried.
+  lock table wmworker_pool_errors_entities;
+
+  create table worker_pool_errors
+  as
+    select
+      (value ->> 'errorId')::text as error_id,
```
helfi92, line 11: `errorId` is defined as a SlugId in data.js. I think Dustin recommended moving away from azure's way of storing things as uuids but rather store things as slug ids. We'll want to modify the migration and downgrade script to make sure we have the proper slugid -> uuid and uuid -> slugid mappings. Dustin created two db functions that we can use `slugid_to_uuid` and `uuid_to_slugid`. We'll want to use them in both the migration and downgrade script. You can refer to Dustin's PR for the inspiration: https://github.com/taskcluster/taskcluster/pull/3115.

db/versions/0017.yml
```diff
@@ -0,0 +1,242 @@
...
+              'extra', extra::text) as value,
+            1 as version,
+          worker_pool_errors.etag as etag
+          from worker_pool_errors
+          where
+            partition_key is null or
+            error_id = decode_string_key(partition_key)
+          order by error_id
+          limit size + 1
+          offset page;
```
helfi92, line 167: If size or offset are not defined then something may go wrong here. One way to fix that is to do:

```
        limit case
          when (size is not null and size > 0) then size + 1
          else null
        end
        offset case
          when (page is not null and page > 0) then page
          else 0
        end;
```

db/versions/0017.yml
```diff
@@ -0,0 +1,242 @@
...
+    returns: table (partition_key_out text, row_key_out text, value jsonb, version integer, etag uuid)
+    body: |-
+      begin
+        return query
+        select
+          wmworker_pool_errors_entities_load.partition_key,
+          wmworker_pool_errors_entities_load.row_key,
+          entity_buf_encode(
+            jsonb_build_object(
+              'PartitionKey', encode_string_key(error_id),
```
helfi92, line 19: This should be `worker_pool_id` rather than `error_id`. This comment applies to the other places in this file where the partition key is specified.

db/versions/0017.yml
```diff
@@ -0,0 +1,242 @@
...
+    serviceName: worker_manager
+    description: See taskcluster-lib-entities
+    mode: write
+    args: partition_key text, row_key text
+    returns: table (etag uuid)
+    body: |-
+      begin
+        return query delete from worker_pool_errors
+        where
+          worker_pool_errors.error_id = decode_string_key(wmworker_pool_errors_entities_remove.partition_key)
```
helfi92, line 71: I think the `where` clause should also specify the partition key `worker_pool_id`. Same thing with the other where clauses in this file.

db/tables.yml
djmitche: *jsonb

db/versions/0018.yml
```diff
@@ -0,0 +1,276 @@
...
+      begin
+        return query
+        select
+          wmworker_pool_errors_entities_load.partition_key,
+          wmworker_pool_errors_entities_load.row_key,
+          entity_buf_encode(
+            jsonb_build_object(
+              'PartitionKey', encode_string_key(worker_pool_id),
+              'RowKey', encode_string_key(error_id),
+              'errorId', error_id,
```
helfi92, line 21: It looks like https://github.com/taskcluster/taskcluster/pull/3134#discussion_r447647401 was missed.

services/worker-manager/src/main.js
```diff
@@ -189,10 +176,10 @@ let load = loader({
   },
 
   providers: {
-    requires: ['cfg', 'monitor', 'notify', 'db', 'estimator', 'WorkerPoolError', 'schemaset'],
-    setup: async ({cfg, monitor, notify, db, estimator, WorkerPoolError, schemaset}) =>
+    requires: ['cfg', 'monitor', 'notify', 'db', 'estimator', 'schemaset'],
+    setup: async ({cfg, monitor, notify, db, estimator, schemaset}) =>
       new Providers().setup({
-        cfg, monitor, notify, db, estimator, WorkerPoolError,
+        cfg, monitor, notify, db, estimator, Worker, WorkerPoolError,
```
helfi92, line 182: Did you mean to remove `WorkerPoolError` rather than add `Worker`?

services/worker-manager/src/data.js
```diff
@@ -134,56 +133,157 @@ class WorkerPool {
...
+  // If there are no workers to show, the response will have the
+  // `rows` field set to an empty array.
+  //
+  // If a handler is supplied, then it will invoke the handler
+  // on every item of the scan.
+  static async getWorkerPoolErrors(
+    db,
+    {
+      query,
+      handler,
```
helfi92, line 192: The handler is not used so no need to have this.

services/worker-manager/src/data.js
```diff
@@ -134,56 +133,157 @@ class WorkerPool {
...
-    workerPoolId: this.workerPoolId,
-    errorId: this.errorId,
-    reported: this.reported.toJSON(),
-    kind: this.kind,
-    title: this.title,
-    description: this.description,
-    extra: this.extra,
-  };
-};
+class WorkerPoolError {
```
helfi92, line 136: I would double check with @imbstack but I suspect we wouldn't want to have a WorkerPoolError class but rather just access the db directly. https://github.com/taskcluster/taskcluster/pull/3105#discussion_r445239721

helfi92: @imbstack @djmitche Can you confirm?

imbstack: I only did a 5-minute scan but it looks like we pretty much only create/read these so yeah I think we can get away with just the db directly. This is not a strongly held position however!

helfi92: It would take some time to rewrite to have it not rely on the class. Given the rest of the tables of this service use a class to handle interactions with the db, maybe it's fine to have it reflect the same format for consistency. For most remaining services it will probably make sense to avoid having a `data.js` file.

djmitche: Either option sounds reasonable, or even keeping what's written but making an issue to switch it later.

djmitche:
@helfi92 has made good comments here so I'll defer to that.

services/worker-manager/src/main.js
milescrabill: Removed both of these here.

services/worker-manager/src/data.js
milescrabill: Removed the handler portion.

services/worker-manager/src/data.js
milescrabill: I left in the class wrapper in to match the rest of worker-manager - I can create an issue to change that if we think that's the right approach.

helfi92:
@milescrabill Is this ready to review?

milescrabill:
Yep, put you back on it. Sorry for the delay!

db/versions/0025.yml
```diff
@@ -0,0 +1,278 @@
...
+            else 0
+          end;
+      end
+  create_worker_pool_error:
+    description: |-
+      Create a new worker pool error.  Raises UNIQUE_VIOLATION if the error already exists.
+    mode: write
+    serviceName: worker_manager
+    args: error_id_in text, worker_pool_id_in text, reported_in timestamptz, kind_in text, title_in text, description_in text, extra_in jsonb
+    returns: uuid
```
helfi92, line 188: The etag is not used in the service so probably best to either return nothing (i.e., `void`) or a table set of length 1 and without the etag.

helfi92: Actually, it's fine for now. We can always remove this later :)

helfi92:
Let's rebase on master and take the next available db version.

db/versions/0025.yml
djmitche: @imbstack asked a similar question in #3115.  Am I correct that the general approach is to keep etags around where that's not difficult, even if they're not strictly necessary?

helfi92: Yes, keep them around but not return them in the new db functions. Thoughts?

djmitche: Sounds good!

djmitche: #3209

djmitche:
(rebased and renumbered)

PR #3135, opened 2020-06-30
Queue task + artifact tables, phase 2, step 1

djmitche:
This contains the DB migrations for the task-related tables.  It's extracted from #3115, which will be rebased to contain only step 2.

Github Bug/Issue: Refs #2932.


djmitche:
In #3143, @helfi92 asked me to cherry-pick the changes in here.  There was some rebasing to do, too :)

djmitche:
So, I think the last three commits here need review, but @helfi92's commit is already r+'d

db/versions/0021-migration.sql
```diff
@@ -0,0 +1,68 @@
...
+    alter column etag set default public.gen_random_uuid();
+
+  -- the task itself contains an array of dependent tasks, so querying by
+  -- dependent_task_id is unusual, so primary key begins with required_task_id.
+  alter table task_dependencies add primary key (required_task_id, dependent_task_id);
+
+  -- queue's isBlocked indexes by dependent_task_id, looking only for unsatisfied
+  -- entries.  This index then takes the place of the queue_task_requirement_entities
+  -- table and that these queries can be handled by accessing only the index.
+  create index task_dependencies_dependent_task_id_idx on task_dependencies (dependent_task_id)
```
helfi92, line 59: Isn't `dependent_task_id` already indexed via the primary key?

db/versions/0019-migration.sql
```diff
@@ -0,0 +1,126 @@
...
+    on
+      tasks_entities.task_id = active_tasks.task_id and
+      tasks_entities.task_group_id = active_tasks.task_group_id;
+
+  alter table tasks add primary key (task_id);
+  -- this index servces the purpose of queue_task_group_members_entities
+  create index tasks_task_group_id_idx on tasks (task_group_id);
+  -- this index servces the purpose of queue_task_group_active_sets_entities
+  create index tasks_task_group_id_unresolved_idx on tasks (task_group_id)
+    where not ever_resolved;
```
helfi92, line 88: Should indexes be dropped during a downgrade?

db/versions/0019-migration.sql
```diff
@@ -0,0 +1,126 @@
...
+    alter column deadline set not null,
+    alter column expires set not null,
+    alter column scopes set not null,
+    alter column payload set not null,
+    alter column metadata set not null,
+    alter column tags set not null,
+    alter column extra set not null,
+    alter column runs set not null,
+    -- note that taken_until is nullable
+    alter column ever_resolved set not null,
```
helfi92, line 112: How is that comment relevant to this line?

db/versions/0019-migration.sql
djmitche: They are automatically dropped when the table they depend on is dropped (confirmed experimentally)

db/versions/0021-migration.sql
djmitche: It is, but as the second field, meaning that it can only be used when the required_task_id is known.

db/versions/0018-downgrade.sql
```diff
@@ -0,0 +1,31 @@
...
+    version integer not null,
+    etag uuid default public.gen_random_uuid());
+  alter table queue_task_groups_entities add primary key (partition_key, row_key);
+
+  insert into queue_task_groups_entities
+  select
+    task_group_id as partition_key,
+    'task-group' as row_key,
+    jsonb_build_object(
+      'PartitionKey', task_group_id,
```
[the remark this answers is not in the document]
djmitche: slugids are unchanged under that transformation, so I left it out to hopefully shave off some time.

db/versions/0019-migration.sql
djmitche: It's because the column isn't named in this list.  I'll update the comment.

djmitche:
I added 0ca6a7c to show some debugging information during the individual steps of the migrations.

djmitche:
That debugging info is available in stackdriver with
```
resource.labels.database_id="taskcluster-dev:dustindev"
textPayload:"LOG:  TIMING"
```

services/queue/test/dependency_test.js
```diff
@@ -579,14 +579,8 @@ helper.secrets.mockSuite(testing.suiteName(), ['aws'], function(mock, skipping)
...
-    debug('### expire task-requirement');
-    await helper.runExpiration('expire-task-requirement');
-    const r5 = await TaskDependency.load({taskId: taskIdA, dependentTaskId: taskIdA}, true);
-    const r6 = await TaskRequirement.load({taskId: taskIdA, requiredTaskId: taskIdA}, true);
-    assert(r5, 'Expected TaskDependency');
-    assert(!r6, 'Did not expect TaskRequirement');
-
-    debug('### expire task-dependency');
+    debug('### expire task-requirement/task-dependency');
+    await helper.runExpiration('expire-task-requirement'); // no longer does anything..
```
imbstack, line 583: maybe link to this in a bug and reference it here so that we remove it later?

djmitche: It gets removed in step 2

djmitche:
docker-worker failure is #3181

djmitche:
OK, as long as that's green, I'll merge and release.

PR #3138, opened 2020-06-30
Support signing postgres table rows

### LINKED ISSUE #3082: Add support for data signing
Some of our tables that are still using taskcluster-lib-entities use signing to protect the integrity of the data. In order to successfully move away from taskcluster-lib-entities and migrate those tables to postgres phase 2, we would need to add inherent support for signing.

Tables affected:
* clients_entities - auth
* hooks_entities - hooks
* last_fire_3_entities - hooks
* queues_entities - hooks
* secrets_entities - secrets
* access_token_table_entities - web_server
* authorization_codes_table_entities - web_server
* github_access_token_table_entities - web_server
* session_storage_table_entities - web_server

djmitche:
Github Bug/Issue: Fixes #3082 

This has some decent tests, but I haven't used it to actually migrate a signed table, so it might need some modification when that time comes.

### DIFF
--- a/libraries/postgres/README.md
+++ b/libraries/postgres/README.md
@@ -315,8 +315,6 @@ The `ser` property, short for "serialization", is an integener defining the colu
 When a table migration occurs that adds a column to the signature, a new `ser` must be introduced, with the old version continuing to verify the old data.
 Note that removing a signed column is impossible without dropping all rows from the table, as otherwise existing signatures would become invalid.
 
-Care must be taken when constructing a serialization for data previously signed by taskcluster-lib-entities, as the serialization is done at the byte level and involves carefully formatted inputs for each column.
-
 ### Signing and Verification API
 
 Both signing and verification expect a row object an an array of functions for serialization of the row, indexed by `ser`:
@@ -343,6 +341,22 @@ const valid = db.verifySignature({
   signature: rows[0].signature,
   signingSerializations,
 });
+```
+
+The `azureEntitiesSerialization` utility function performs part of the serialization operation used by azure-entities.
+It requires as input an object mapping property names to the *hashed* version of their values, where the hashes are defined by each entity value type.
+The function sorts the properties and encodes them using the 32-bit run-length-encoding introduced by azure-entities.
+
+```javascript
+const signingSerializations = [
+  row => azureEntitiesSerialization({
+    // strings hash as themselves
+    id: row.id,
+    name: row.name,
+    // hash value of an integer column is the string form of the integer
+    count: row.count.toString(),
+  }),
+];
 ```
 
 ### Updating Tables With New Keys or New Serializations
--- a/libraries/postgres/README.md
+++ b/libraries/postgres/README.md
@@ -26,6 +26,8 @@ const db = Database.setup({
   poolSize: ..., // optional, default 5
   azureCryptoKey: ..., // optional, only required to read lib-entities encrypted data
   dbCryptoKeys: ..., // optional, only required if encrypting columns, usually from cfg.postgres.dbCryptoKeys
+  azureSigningKey: ..., // optional, only required to read lib-entities signed data
+  dbSigningKeys: ..., // optional, only required if signing rows, usually from cfg.postgres.dbSigningKeys
 });
 ```
 
@@ -35,7 +37,7 @@ The `monitor` is a taskcluster-lib-monitor instance, used to report database met
 if `statementTimeout` is set, then it is treated as a timeout (in milliseconds) after which a statement will be aborted.
 This is typically used in web processes to abort statements running longer than 30s, after which time the HTTP client has likely given up.
 
-The `azureCryptoKey`, and `dbCryptoKeys` parameters are explained below in "Secret Data" and "Encryption".
+The `azureSigningKey`, `dbSigningKeys`, `azureCryptoKey`, and `dbCryptoKeys` parameters are explained below in "Secret Data", "Encryption", and "Signing".
 
 The `poolSize` parameter specifies the maximum number of Postgres clients in each pool of clients, with two pools (read and write) in use.
 DB function calls made when there are no clients available will be queued and wait until a client is avaliable.
@@ -256,7 +258,7 @@ Both take a parameter named `value` that will either be encrypted and built into
 or pulled out of that format and decryped.
 
 To enable encryption/decryption, provide `Database.setup` with at least one of `azureCryptoKey` or `dbCryptoKeys`. The first
-of which is the key that `lib-entities` used for encryption. The latter is structured like
+of which is the key that `lib-entities` used for encryption. The latter is an array of keys with properties `id`, `algo`, and `key`:
 
 ```javascript
 [
@@ -277,7 +279,7 @@ Once the database has been set up, encryption and decryption can be used as foll
 ```javascript
 await db.fns.create_widget(widgetId, db.encrypt({value: Buffer.from(widgetCode, 'utf8')}));
 
-const rows = db.fns.get_widget(widgetId);
+const rows = await db.fns.get_widget(widgetId);
 console.log(db.decrypt({value: rows[0].widget_code}).toString('utf8'));
 ```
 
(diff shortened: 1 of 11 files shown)


djmitche:
OK, I think that passes tests now -- I had missed a bunch of places in `db/` that needed updated.

libraries/postgres/src/Database.js
```diff
@@ -48,26 +48,48 @@ MonitorManager.register({
   },
 });
 
+/** Fixed time comparison of two buffers */
```
petemoore, line 51: This seems like a reasonable implementation, but why not just use [crypto.timingSafeEqual(a, b)](https://nodejs.org/api/crypto.html#crypto_crypto_timingsafeequal_a_b)?

libraries/postgres/src/Database.js
```diff
@@ -721,6 +744,49 @@ class Database {
...
+   */
+  verifySignature({row, signature, signingSerializations}) {
+    const sig = this.calculateSignature({
+      row,
+      kid: signature.kid,
+      v: signature.v,
+      ser: signature.ser,
+      signingSerializations,
+    });
+    return fixedTimeComparison(Buffer.from(signature.sig, 'base64'), Buffer.from(sig, 'base64'));
```
helfi92, line 773: I'm not sure I understand everything with fixed time comparison. Do you have a link to suggest that I can use to understand this better? Thanks!

libraries/postgres/README.md
```diff
@@ -300,6 +302,75 @@ The binary payload is derived by first base64-decoding each `__bufN_val` propert
...
+## Signing
+
+Taskcluster signs database rows, generally including all columns in the row.
+The signature takes the form of a "signature container", similar to a "crypto container", a JSONB object with properties `kid`, `v`, and `ser`.
+The `kid` field references signing keys (a distinct set from crypto keys), and the versions in `v` are numbered distinctly from crypto containers.
+This container is typically stored in a `signature` column.
+
+The versions `v` encompass algorithm and encoding, and are described below.
+
+The `ser` property, short for "serialization", is an integener defining the columns that are included in the serialized input to the signature algorithm.
```
helfi92, line 314: integener

libraries/postgres/README.md
```diff
@@ -300,6 +302,75 @@ The binary payload is derived by first base64-decoding each `__bufN_val` propert
...
+When a table migration occurs that adds a column to the signature, a new `ser` must be introduced, with the old version continuing to verify the old data.
```
helfi92, line 315: Is there a test that would fail if someone forgets to do this or is that something each service needs to test individually?

libraries/postgres/README.md
```diff
@@ -300,6 +302,75 @@ The binary payload is derived by first base64-decoding each `__bufN_val` propert
...
+const signingSerializations = [
+  row => azureEntitiesSerialization({
+    // strings hash as themselves
+    id: row.id,
+    name: row.name,
+    // hash value of an integer column is the string form of the integer
+    count: row.count.toString(),
+  }),
+];
+```
```
helfi92, line 360: Any chance we could add an example that includes the signature container?

ui/docs/manual/deploying/database.mdx
```diff
@@ -125,15 +125,40 @@ is a JSON array where each element is an object of the form
...
+}
+```
+
+The key is used as input to HMAC, so it should be long enough to have
+sufficient entropy but otherwise can contain anything.  The same function given
+above to generate encryption keys is suitable.
+
+As with encryption, the last element in this array is used to sign new and
+updated table rows.  If `azure_signing_key` is given, it is treated as if it
+was the first entry in the `db_signing_keys` array, with id "azure".
```
helfi92, line 160: Why do we need `azure_signing_key` when we could just put it in `db_signing_keys`?

libraries/postgres/src/Database.js
petemoore: https://severalnines.com/database-blog/utilizing-encryption-strengthen-postgresql-database-security is a good overview of postgres's extensive security features, and https://research.kudelskisecurity.com/2013/12/13/timing-attacks-part-1/ is a specific article about timing attacks.

djmitche: To the first link, note that we are specifically *not* doing any crypto on the database itself.  All encryption and signing takes place in Node, with keys and plaintext never visible on the DB server.

In this context I doubt a timing attack would be practical, since the timing of the comparison would be miniscule copmared to the DB query, network traffic, multiple layers of HTTP proxying, calls out to auth.authenticateHawk, and k8s and virutalization overhead.  But it's a best practice and it's easy, so we do it.

libraries/postgres/src/Database.js
djmitche: > There is already [crypto.timingSafeEqual(a, b)](https://nodejs.org/api/crypto.html#crypto_crypto_timingsafeequal_a_b) and you already `require` the `crypto` package, so there is no reason to implement this yourself.

That is a "new" addition (Node 6.6.0) -- this function was copied from azure-entities, and wasn't my handiwork.  We've been using it for years.  Good spot!

> Also note, all changes like these that affect security of the platform should be flagged for security peer review.

Yep, just like the crypto patch, I'm hoping we can get a quick look from Aj based on the fact that this is a quick forward-port of the existing approach.  I pinged on #3082 but I'll send an email now.

djmitche:
Now that I think about this, I realize that there's a downgrade vulnerability in our process for adding columns to the signature.

Imagine we have ser=0 that covers columns A and B, and introduce a column C and ser=1 that covers A, B, and C.  In order to control the value of C on a row, an attacker with DB access can copy an old (ser=0) signature for that row (perhaps from a backup, or perhaps this attacker has been sitting on our DB for months) and then set C to an arbitrary value.  Signature validation will succeed since it only looks at A and B for ser=0, which the attacker has not changed.

Probably the fix is to ensure that validation for old serializations checks that any newly-added columns are NULL.  So in the example here, when we land the patch to define ser=1, we'd also patch the implementation of ser=0 to ensure that C is NULL.  And, of course, C being NULL would have to have some innocuous meaning.

I'll think about how to enforce this invariant at the code level, rather than expecting developers to remember the invariant.  And I'll make the other changes suggested above.

@ajvb if you'd like to look before I make those changes, that should be fine.  I'll make the substantive changes in a new commit, so you can just give that commit a look when it's pushed.

libraries/postgres/README.md
djmitche: We could do so if we require that signatures always cover the entire row, and not just a subset of the columns.  That's always been the case for Azure, so seems reasonable to enforce here.  Thanks!

ui/docs/manual/deploying/database.mdx
djmitche: This matches what we've done with crypto keys.  The idea is that services are running right now with `azure_signing_key` and that doesn't need to change, but next time we rotate keys ops folks can add the *new* key to `db_signing_keys` and leave the old key in `azure_signing_key` until it is no longer needed.

libraries/postgres/src/Database.js
petemoore: Unfortunately that is (understandably) a common misconception. To quote the conclusion of the above cited study:

Screenshot 2020-07-03 at 17 35 46

So you are right that a lot of effort would be needed to launch such an attack successfully. But the key result of this paper is that it isn't infeasible, and such attacks have been successful against other targets in the past, since even a considerable amount of external noise can be filtered out if enough data is collected.

djmitche: Yep, timing attacks are the basis of meltdown and spectre, too.

libraries/postgres/src/Database.js
ajvb: Can we switch to using `crypto.timingSafeEqual(a, b)` in this PR? If not, can we have a migration plan?

libraries/postgres/src/Database.js
```diff
@@ -721,6 +744,49 @@ class Database {
...
+  }
+
+  /**
+   * Helper for sign and verifySignature
+   */
+  calculateSignature({row, kid, v, ser, signingSerializations}) {
```
ajvb, line 779: Similar to https://github.com/taskcluster/taskcluster/pull/3138/files#r449564731 why not use [Sign](https://nodejs.org/api/crypto.html#crypto_class_sign) and [Verify](https://nodejs.org/api/crypto.html#crypto_class_verify) from the Node crypto lib here?

libraries/postgres/src/Database.js
djmitche: Yes, that's part of my followups.  I forgot to work on that yesterday -- sorry!

libraries/postgres/src/Database.js
djmitche: If that produces matching output, that's totally fine -- but it looks like it uses asymmetric-key encryption, where the existing signing is based on HMAC, so symmetric.  We need to maintain compatibility with existing signatures.  We could migrate to asymmetric-key later using the `algo` parameter to in the signature container, but since we'd always carry the public and private keys around together, I'm not sure there's much advantage (and there's cost to deploying two much larger keys).

@ajvb I think the "our own implementation" characterization is referring to the fixed-time comparison, yes?  The use of HMAC here seems pretty standard and non-clever.

libraries/postgres/README.md
djmitche: For now, I'll include the requirement that all rows be considered in the docs.  For anything we're porting from Azure, that's already the case.  I bumped this improvement out to #3180.  We can do that if we decide to keep signing support.

djmitche:
OK, this is ready for review now.  The only code change I made was to use the provided constant-time comparison.  The rest of the changes are to docs only.  They're all in the last few commits in the PR.  I also rebased to master to resolve conflicts.

djmitche:
We have decided not to support signing with postgres, as it does not deliver much value to us.

PR #3142, opened 2020-07-02
Remove fake DB functions and the tests that use them

### LINKED ISSUE #3131: Remove fake db functions
Now that we've moved away from Azure and are using Postgres, the time it takes to run tests via a local postgres db vs fake db functions is not much of a big difference. Rather than having us duplicate the db business logic in two places, let's just run our tests with the postgres db. This will consist of the following changes:
* Delete `db/src/fakes`
* Remove the automation that runs the tests twice
* Update the guideline docs to reflect the new changes to step 2 of the migration

djmitche:
Github Bug/Issue: Fixes #3131


djmitche:
Ah, one thing to note here is that we can't use "fake time" with a real DB.  So I'll need to remove some uses of zurvan.

djmitche:
Seventeenth time's the charm :)

djmitche:
..now with the fakes actually removed!

djmitche:
Last push was just a whitespace lint, so I'm confident this will be green.

PR #3121, opened 2020-06-26
Port queue_artifacts_entities to postgres phase 2

helfi92:
Resolves https://github.com/taskcluster/taskcluster/issues/3112.

Issue #3112, opened 2020-06-24
port the queue's artifacts table to use normal DB tables


djmitche:
* queue_artifacts_entities

This table has 217 million rows!

Broken out from #2932, as this can be done largely separately from that issue.

PR #3143, opened 2020-07-02
Queue artifacts table, phase 2, step 1

helfi92:
This contains the DB migrations for the queue artifacts table.  It's extracted from #3121, which will be rebased to contain only step 2.

Github Bug/Issue: Refs https://github.com/taskcluster/taskcluster/issues/3112.


db/test/versions/0017_test.js
```diff
@@ -0,0 +1,118 @@
...
+          present: false,
+          expires: fromNow('1 day'),
+        }]))),
+    },
+    loadConditions: [
+      {condition: {taskId: 'J6CQMUHpTyGrWDgfp17J8g', runId: 5, name: 'name/name' }, expectedSample: 'pptt'},
+      {condition: {taskId: taskIds[1], runId: 1, name: 'name-1' }, expectedSample: 'samp1'},
+    ],
+    scanConditions: [
+      // expected is ordered by the hashed namespace
```
djmitche, line 94: ??

db/test/versions/0017_test.js
```diff
@@ -0,0 +1,118 @@
...
+      {condition: {}, expectedSamples: ['samp4', 'pptt', 'samp2', 'samp1', 'samp0', 'samp3']},
+      {condition: null, expectedSamples: ['samp4', 'pptt', 'samp2', 'samp1', 'samp0', 'samp3']},
```
djmitche, line 96: Let's get a test here to simulate the query in https://github.com/taskcluster/taskcluster/blob/e587b9ec8b9c074d4c4f55167c9186d546b1b5cb/services/queue/src/artifacts.js#L602-L605

db/versions/0017.yml
```diff
@@ -0,0 +1,231 @@
...
+              'storageType', storage_type,
+              'contentType', content_type,
+              'present', present,
+              'expires', expires),
+            'details', details::text) as value,
+          1 as version,
+          queue_artifacts.etag as etag
+        from queue_artifacts
+        where
+          queue_artifacts.task_id = decoded_composite_key[1] and queue_artifacts.run_id::text = decoded_composite_key[2] and queue_artifacts.name = decode_string_key(queue_artifacts_entities_load.row_key);
```
djmitche, line 36:
```
postgres=# explain select * from queue_artifacts where task_id='x' and run_id::text = '25';
                                         QUERY PLAN                                          
---------------------------------------------------------------------------------------------
 Index Scan using queue_artifacts_pkey on queue_artifacts  (cost=0.29..8.31 rows=1 width=52)
   Index Cond: (task_id = 'x'::text)
   Filter: ((run_id)::text = '25'::text)
```
In general, any time there's anything but a bare column name in a where clause, there's a risk of a table scan.  In this case, it's not so bad because postgres can use the index to find the task, and then just scan through the runs (of which there are probably 1 and at most 50).  Postgres doesn't know that these are all stringified integers, so it can't translate `run_id::text = key` into `run_id = key::int`.  I think explicitly casting `decoded_composite_key[2]::int` is the fix here (and in all the while clauses).

```
postgres=# explain select * from queue_artifacts where task_id='x' and run_id = '25'::int;
                                         QUERY PLAN                                          
---------------------------------------------------------------------------------------------
 Index Scan using queue_artifacts_pkey on queue_artifacts  (cost=0.29..8.30 rows=1 width=52)
   Index Cond: ((task_id = 'x'::text) AND (run_id = 25))
```

helfi92: That makes a lot of sense. Thank you for both explain queries. It makes it easier to spot in the future.

helfi92:
Added the suggested changes. If everything looks good to you, feel free to go ahead and cherry pick the commit to your PR.

djmitche:
cherry-picked / rebased into #3135.

PR #3152, opened 2020-07-03
Queue artifacts table, phase 2, step 2

helfi92:


services/queue/src/utils.js
```diff
@@ -0,0 +1,133 @@
...
+  },
+  //  Get all expired artifacts using a handler function.
+  // This is to avoid loading all epxired queue artifacts in memory.
+  async getExpiredArtifacts(db, expires, handler) {
+    assert(!handler || handler instanceof Function,
+      'If options.handler is given it must be a function');
+
+    const fetchResults = async (continuation) => {
+      const query = continuation ? { continuationToken: continuation } : {};
+      const {continuationToken, rows} = await paginateResults({
```
djmitche, line 45: paginateResults is for APIs.. it's just abstracting over the pagination in a way that obscures what's really happening here.  I think this function could just loop with adjustments to `page_offset` on each iteration, until it gets zero results.

services/queue/src/utils.js
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Remove underlying artifact and entity representing it.
+   *
+   * This method will remove both the artifact in the db and underlying artifact.
+   * But the artifact in the db will not be deleted if there is an error
+   * deleting the underlying artifact.
+   */
+  async expire({ db, publicBucket, privateBucket, ignoreError, monitor, expires }) {
+    let count = 0;
+    await exports.artifactUtils.getExpiredArtifacts(db, expires, async (artifact) => {
```
djmitche, line 85: This seems to be the only use of `getExpiredArtifacts`, so perhaps that utility could be combined here, or incorporated as a nested function?

I think this is going to miss expiring artifacts:

* call get_queue_artifacts with a page size of 1000
* call S3 for each of those 1000
* call delete_artifact for each of those 1000
* call get_queue_artifacts with a page size of 1000, offset 1000

..at this point, we've missed 1000 expired artifacts since the offset skips 1000 non-deleted artifacts.

This actually isn't the end of the world -- those artifacts would hopefully just get picked up in the next run of the expiration job -- but might be confusing when we wonder why there's a small tail of expired artifacts in the DB.

I suspect the fix would be to decrement the offset by 1 for every call to `delete_queue_artifact`.  In the success case, that will end up calling `get_queue_artifacts` with offset = 0 every time, until it returns no results, but in the failure case it will not re-try deleting S3 objects.

Separately, it seems like we should be taking a closer look at these errors instead of just ignoring them.  I'll file a followup for that.

djmitche: (#3183)

services/queue/src/artifacts.js
```diff
@@ -245,10 +246,12 @@ builder.declare({
...
-    await artifact.modify((artifact) => {
-      artifact.expires = expires;
-      artifact.details = details;
-    });
+    artifact = artifactUtils.fromDbRows(
+      await this.db.fns.update_queue_artifact(
+        artifact.taskId, artifact.runId, artifact.name, details,
+        expires, artifact.etag
+      )
+    );
```
djmitche, line 254: Since this passes `etag`, it should handle conflicts.  I'm not sure it needs to handle the etag, though?  I think this is just for idempotency so we can assume that this call will override a previous call.

And, with that change, you can drop etag support for `update_queue_artifact`, which simplifies its implementation quite a bit.

services/queue/src/utils.js
```diff
@@ -0,0 +1,112 @@
...
+      // When underlying artifact is deleted (if any underlying artifact exists)
+      // we delete the artifact Entity.
+      try {
+        await deleted;
+
+        // Delete entity, if underlying resource was successfully deleted
+        await db.fns.delete_queue_artifact(artifact.taskId, artifact.runId, artifact.name);
+        // to avoid having the offset skip an expired artifact when
+        // we delete an artifact from the table.
+        pageOffset -= 1;
```
djmitche, line 95: nice :)

PR #3161, opened 2020-07-06
Migrate web_server.github_access_token_table_entities -> web_server.github_access_tokens

### LINKED ISSUE #3148: port github_access_token_table_entities to use normal DB tables
This only involves doing step 2 of the migration. It's okay if people are logged out.

petemoore:
Github Bug/Issue: Fixes #3148


db/versions/0023.yml
```diff
@@ -0,0 +1,41 @@
...
+migrationScript: 0023-migration.sql
+downgradeScript: 0023-downgrade.sql
+methods:
+  add_github_accesstoken:
+    description: |-
+      Sets the access token for user_id_in to access_token_in.
+
+      If no access token is currently set for user_id_in, a new row is
+      inserted, otherwise the existing row is updated with a new access token
+      and etag, regardless of whether the access token has changed or not.
```
petemoore, line 11: Just spotted my own mistake:

```suggestion
      inserted, otherwise the existing row's access token is updated to
      access_token_in, and a new etag is assigned regardless of whether the
      access token changed or not.
```

changelog/issue-3148.md
```diff
@@ -0,0 +1,6 @@
+audience: deployers
+level: major
```
djmitche, line 2: I think this can be minor, since it does not require a TCW

db/versions/0023-migration.sql
```diff
@@ -0,0 +1,25 @@
...
+  -- updates when the table is dropped.  Note that this may lead to concurrent
+  -- updates failing; the important thing is that they not succeed without
+  -- taking effect.  Failed updates will be retried.
+  lock table github_access_token_table_entities;
+
+  create table github_access_tokens
+  as
+    select
+      (value ->> 'userId')::text as user_id,
+      (value ->> 'accessToken')::text as access_token,
```
djmitche, line 12: This is encrypted, so you'll need to wrap this in a crypto container on upgrade.  I think we'd also considered just dropping this table on upgrade?  The column type for an encrypted column is JSONB.

db/versions/0023-downgrade.sql
```diff
@@ -0,0 +1,30 @@
...
+    partition_key text,
+    row_key text,
+    value jsonb not null,
+    version integer not null,
+    etag uuid default public.gen_random_uuid());
+  alter table github_access_token_table_entities add primary key (partition_key, row_key);
+  grant select, insert, update, delete on github_access_token_table_entities to $db_user_prefix$_web_server;
+
+  insert into github_access_token_table_entities
+  select
```
djmitche, line 16: This table is signed, and on downgrade we will have lost the signatures, so I think we need to just drop the data when downgrading.  That will be the case for all signed tables :(

db/versions/0023.yml
```diff
@@ -0,0 +1,42 @@
+version: 23
+migrationScript: 0023-migration.sql
+downgradeScript: 0023-downgrade.sql
+methods:
+  add_github_accesstoken:
```
djmitche, line 5: The table puts a `_` in `access_token`, but this does not.  Any particular reason for the distinction?

services/web-server/src/login/strategies/github.js
```diff
@@ -19,23 +19,25 @@ module.exports = class Github {
...
+    if (accessToken.length === 0) {
       this.monitor.debug('Github user id could not be found in the database.', {
         userId,
       });
 
       return;
     }
 
-    const githubClient = new GithubClient({ accessToken: githubEntry.accessToken });
+    const githubClient = new GithubClient({ accessToken: accessToken[0]["access_token"] });
```
djmitche, line 40: This will need to be decrypted

services/web-server/src/login/strategies/github.js
```diff
@@ -127,10 +131,10 @@ module.exports = class Github {
           scope: 'repo',
         },
         async (accessToken, refreshToken, profile, next) => {
-          await this.GithubAccessToken.create({
-            userId: profile.id,
+          await this.db.fns.add_github_accesstoken(
+            profile.id,
             accessToken,
```
djmitche, line 136: ...and encrypted here

services/web-server/test/login_strategies_github_test.js
```diff
@@ -5,17 +5,15 @@ const Github = require('../src/login/strategies/github');
...
-    const makeUser = (options) => {
-      return helper.GithubAccessToken.create({
-        userId: '99',
-        accessToken: 'qwerty',
-        ...options,
-      }, true);
+    const makeUser = (userId) => {
+      return helper.db.fns.add_github_accesstoken(
+        userId,
+        'qwerty', // access token
```
djmitche, line 15: encrypted here too :)

services/web-server/test/login_strategies_github_test.js
```diff
@@ -54,31 +52,31 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
+      await makeUser(userId);
       const user = await strategy.userFromIdentity(`github/${userId}|a!2Fc`);
       assert(user, "returned undefined");
       assert.equal(user.identity, `github/${userId}|a!2Fc`);
     });
 
     test('userFromIdentity with unknown user', async function() {
       const strategy = getStrategy();
-      await makeUser();
+      await makeUser(99);
```
djmitche, line 71: should be a string here, too

changelog/issue-3148.md
petemoore: 👍

db/versions/0023-migration.sql
petemoore: Good spot. I've removed the data migration now.

db/versions/0023-downgrade.sql
petemoore: Done!

db/versions/0023.yml
petemoore: Ah quite right - I've embedded a `_` to make it consistent now.

services/web-server/src/login/strategies/github.js
petemoore: Oh yes, whoops! I've done that now.

services/web-server/src/login/strategies/github.js
petemoore: Indeed, I totally forgot about encryption, good that you checked so carefully! ❤️ 
Done now.

services/web-server/test/login_strategies_github_test.js
petemoore: 👍

services/web-server/test/login_strategies_github_test.js
petemoore: Oh yes, thanks. Done!

db/test/fns/web_server_test.js
```diff
@@ -0,0 +1,43 @@
+const assert = require('assert').strict;
+const helper = require('../helper');
+const testing = require('taskcluster-lib-testing');
+
+suite.only(testing.suiteName(), function() {
```
djmitche, line 5:
```suggestion
suite(testing.suiteName(), function() {
```

db/test/versions/0023_test.js
```diff
@@ -0,0 +1,67 @@
...
+  signEntities: true,
+  properties: {
+    // A unique user ID assigned to the user
+    userId: Entity.types.String,
+    // The Github Oauth access token
+    accessToken: Entity.types.EncryptedText,
+  },
+});
+
+suite(testing.suiteName(), function() {
```
djmitche, line 23: This begins to fail after the `.only` is removed, partly due to the absence of a cryptoKey.  But it will also fail because the data's being dropped on upgrade/downgrade.  I bet that could be worked-around by adding a `data-migration` flag to `notImplemented` and just omitting those tests when it's present.

petemoore: Thanks for the tips.

> This begins to fail after the `.only` is removed, partly due to the absence of a cryptoKey.

I see that we set a test crypto key for web server tests [here](https://github.com/taskcluster/taskcluster/blob/426506f034750f7930fdd63e166aef4811889f63/services/web-server/config.yml#L118) - I'm thinking about setting one for the db tests too. Does that seem reasonable to you?

db/test/helper.js
```diff
@@ -406,11 +412,15 @@ exports.testEntityTable = ({
         await exports.upgradeTo(dbVersion);
       });
 
-      for (let {condition, expectedSample} of loadConditions) {
-        test(`load ${JSON.stringify(condition)}`, async function() {
-          const entity = await Entity.load(condition);
-          assert.deepEqual(entity._properties, samples[expectedSample]);
-        });
+      if (notImplemented.includes('entity-functions')) {
+        test('database upgrade', async function() {});
```
helfi92, line 416: Why is this called `named database upgrade` when the test is empty? Would it make more sense to simply have the if statement without the `else`?

```
if (!notImplemented.includes('entity-functions')) {
  // ...
}
```

db/test/helper.js
```diff
@@ -421,30 +431,43 @@ exports.testEntityTable = ({
...
+          await Promise.all(Object.values(samples).map(
+            sample => {
+              return Entity.create(sample);
+            }));
+        }
       });
 
-      makeTests();
+      if (notImplemented.includes('entity-functions')) {
+        test('database setup', async function() {});
```
helfi92, line 444: Same as previous comment. The test talk about a database upgrade but nothing happens in the test?

db/test/helper.js
```diff
@@ -421,30 +431,43 @@ exports.testEntityTable = ({
...
+      } else {
+        makeTests();
+      }
       customTests.call(this, true);
     });
 
     suite(`db version ${dbVersion} -> ${prevVersion} preserves data`, function() {
```
helfi92, line 451: Should this test suite even bother to run if we're skipping step 1? Same comment applies to the first suite `db version ${prevVersion}`.

db/versions/0023.yml
```diff
@@ -0,0 +1,43 @@
+version: 23
+migrationScript: 0023-migration.sql
+downgradeScript: 0023-downgrade.sql
+methods:
+  add_github_access_token:
+    description: |-
+      Sets the encrypted access token for user_id_in to
+      encrypted_access_token_in.
```
helfi92, line 8: In the future we might render the description somewhere where markdown compatible. We can probably wrap column names with back ticks in the description.

db/versions/0023.yml
```diff
@@ -0,0 +1,43 @@
...
+      inserted, otherwise the existing row's encrypted access token is updated
+      to encrypted_access_token_in, and a new etag is assigned regardless of
+      whether the access token changed or not.
+    mode: write
+    serviceName: web_server
+    args: user_id_in text, encrypted_access_token_in jsonb
+    returns: void
+    body: |-
+      declare
+        new_etag uuid := public.gen_random_uuid();
```
helfi92, line 20: If etag is not being used, then maybe we should delete this column.

db/test/helper.js
petemoore: I did this so that the test setup would be executed. Without a test, I think the test setup steps would never run, so this ensures at least one test is generated so that the test setup steps are executed as they are useful for testing the database scripts too even if data isn't migrated.

db/test/helper.js
petemoore: Same as above.

db/test/helper.js
petemoore: Same as above.

db/versions/0023.yml
petemoore: Sure I'll add backticks.

db/test/helper.js
helfi92: Ah, I see. That makes sense. Thanks.

db/versions/0029-migration.sql
```diff
@@ -0,0 +1,12 @@
+begin
+
+  create table github_access_tokens(
+    user_id text not null,
+    encrypted_access_token jsonb not null);
```
djmitche, line 5: I like the `encrypted_` prefix.. let's try to keep doing that!

services/web-server/src/main.js
```diff
@@ -154,18 +153,14 @@ const load = loader(
...
     strategies: {
-      requires: ['cfg', 'GithubAccessToken', 'monitor'],
-      setup: ({ cfg, GithubAccessToken, monitor }) => {
+      requires: ['cfg', 'monitor', 'db'],
+      setup: ({ cfg, monitor, db }) => {
         const strategies = {};
 
         Object.keys(cfg.login.strategies || {}).forEach((name) => {
           const Strategy = require('./login/strategies/' + name);
           const options = { name, cfg, monitor };
```
djmitche, line 162:
```suggestion
          const options = { name, cfg, monitor, db };
```

Issue #2936, opened 2020-05-29
 port the hooks service's tables to use normal DB tables


djmitche:
https://bugzilla.mozilla.org/show_bug.cgi?id=1630016

Complications:
 * some tables are encrypted and signed

helfi92:
The signed tables are:
* hooks_entities
* last_fire_3_entities

Until https://github.com/taskcluster/taskcluster/issues/3082 is resolved, 2/3 of the tables are blocked.

PR #3165, opened 2020-07-07
Port last_fire_3_entities to postgres phase 2

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/2936.

My understand is that it's okay to skip step 1 of the migration since it's okay to lose this data.


helfi92:
Continued in https://github.com/taskcluster/taskcluster/pull/3179.

PR #3167, opened 2020-07-07
Support target versions for DB upgrades

djmitche:


infrastructure/tooling/src/main.js
```diff
@@ -147,6 +147,7 @@ program.command('dev:init')
 
 program.command('dev:db:upgrade')
   .description('Run `yarn db:upgrade` for a development environment')
+  .option('--db-version <v>', 'Downgrade to this DB version (optional, defaults to latest)')
```
helfi92, line 150: Wouldn't this defeat the purpose of having a dev:db:downgrade?

helfi92:
Tests seem to be failing.

infrastructure/tooling/src/main.js
djmitche: Whoops, that was copy-paste -- should be 
```suggestion
  .option('--db-version <v>', 'Upgrade to this DB version (optional, defaults to latest)')
```

djmitche:
Should be fixed.. let's see what the tests say

PR #3171, opened 2020-07-07
Support docker images generated by newer docker versions.

### LINKED ISSUE #2973: [docker-worker] Support new container image formats
Currently, docker-worker only supports loading images from tasks or artifacts in the [v1 image format](https://github.com/moby/moby/blob/master/image/spec/v1.md). While this is the format generated by our docker-in-docker tasks, it is no longer well supported by tools, and is hard to generate with current tools. In particular, I've need to workaround this limitation in [Bug 1626058](https://bugzilla.mozilla.org/show_bug.cgi?id=1626058), by using a second tool to convert to a v1 supporting image.

This limitation is based on the code [here](https://github.com/taskcluster/taskcluster/blob/9ff7e551a64758910d60edc375490d8b5f5f13ef/workers/docker-worker/src/lib/docker/artifact_image.js#L246) that mutates the image before loading it with docker, to ensure the image name that is loaded is controlled by docker-worker.

tomprince:

Github Bug/Issue: Fixes #2973.

### DIFF
--- a/workers/docker-worker/test/integration/docker_manifest_test.js
+++ b/workers/docker-worker/test/integration/docker_manifest_test.js
@@ -83,7 +83,10 @@ helper.secrets.mockSuite(suiteName(), ['docker', 'ci-creds'], function(mock, ski
           '-o',
           '/image.tar.zst',
           '-L',
-          'https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/U9nuixDnT_6eGJ-OsKMLCQ/runs/0/artifacts/public/image.tar.zst',
+          // This image is the busybox image build from https://hg.mozilla.org/try/rev/b76a6f5b3a26211da5ea1fa6a86329f921a302b7
+          // It is a docker image built with kaniko based on `FROM busybox`, with the workaround we have been using for
+          // https://github.com/taskcluster/taskcluster/issues/2973 turned off.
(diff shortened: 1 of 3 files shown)


workers/docker-worker/test/integration/docker_manifest_test.js
```diff
@@ -65,4 +65,46 @@ helper.secrets.mockSuite(suiteName(), ['docker', 'ci-creds'], function(mock, ski
...
+            expires: expires(),
+            path: '/image.tar.zst',
+          },
+        },
+        command: [
+          'curl',
+          '-o',
+          '/image.tar.zst',
+          '-L',
+          'https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/U9nuixDnT_6eGJ-OsKMLCQ/runs/0/artifacts/public/image.tar.zst',
```
tomprince, line 86: This probably wants to be copied to https://s3-us-west-2.amazonaws.com/docker-worker-manifest-test/ before landing, since the artifact referenced here will expire in < month.

djmitche: I'll push a change with the new URL.  Can you add some comments as to how the image was created?

djmitche: Oh, I can't push to your repo.  The new URL is `https://s3-us-west-2.amazonaws.com/docker-worker-manifest-test/image-v1_2.tar.zst`.

PR #3179, opened 2020-07-08
Migrate queue_entities & last_fire_3

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/2936.


changelog/issue-2936.md
```diff
@@ -0,0 +1,5 @@
+audience: deployers
+level: minor
+reference: issue 2936
+---
+Hooks last fires are now stored in a relational table `hooks_last_fires`.
```
djmitche, line 5: *queues and last fires?

db/test/versions/0019_test.js
```diff
@@ -0,0 +1,92 @@
...
+const THIS_VERSION = parseInt(/.*\/0*(\d+)_test\.js/.exec(__filename)[1]);
+const PREV_VERSION = THIS_VERSION - 1;
+
+// (adapted from services/hooks/src/data.js)
+const QueuesEntity = Entity.configure({
+  version: 1,
+  partitionKey: Entity.keys.StringKey('hookGroupId'),
+  rowKey: Entity.keys.StringKey('hookId'),
+  // TODO: ...
+  // signEntities: true,
```
djmitche, line 16: TBD :)

db/versions/0018.yml
```diff
@@ -0,0 +1,110 @@
+version: 18
+migrationScript: 0018-migration.sql
+downgradeScript: 0018-downgrade.sql
+methods:
```
djmitche, line 4: It looks like this leaves the existing entities functions in place, using the now-dropped entities table.  What is the rationale?  If we need to roll back the service deployment, must we also roll back the DB?

db/versions/0018.yml
```diff
@@ -0,0 +1,110 @@
...
+          hooks_last_fires.etag
+        from hooks_last_fires
+        where
+          hooks_last_fires.hook_group_id = hook_group_id_in and
+          hooks_last_fires.hook_id = hook_id_in and
+          hooks_last_fires.task_id = task_id_in;
+      end
+  get_last_fires:
+    description: |-
+      Get hooks last fires filtered by the optional arguments,
```
djmitche, line 49:
```suggestion
      Get hooks last fires filtered by the hook_id and hook_group_id arguments,
```

db/versions/0018.yml
```diff
@@ -0,0 +1,110 @@
...
+        where
+          hooks_last_fires.hook_group_id = hook_group_id_in and
+          hooks_last_fires.hook_id = hook_id_in
+        order by hook_group_id, hook_id, task_id
+        limit get_page_limit(page_size_in)
+        offset get_page_offset(page_offset_in);
+      end
+  delete_last_fires:
+    description: |-
+      Delete last fires that matche a given `hook_group_id` and `hook_id`.
```
djmitche, line 79:
```suggestion
      Delete last fires that match a given `hook_group_id` and `hook_id`.
```

db/versions/0019.yml
```diff
@@ -0,0 +1,273 @@
...
+          hooks_queues.bindings,
+          hooks_queues.etag
+        from hooks_queues
+        order by hook_group_id, hook_id
+        limit get_page_limit(page_size_in)
+        offset get_page_offset(page_offset_in);
+      end
+  update_hooks_queue_bindings:
+    description: |-
+      Update bindings of a hooks queue. If no such worker pool exists,
```
djmitche, line 250:
```suggestion
      Update bindings of a hooks queue. If no such queue exists,
```

services/hooks/src/utils.js
```diff
@@ -0,0 +1,133 @@
...
+      taskId: row.task_id,
+      taskCreateTime: row.task_create_time,
+      result: row.result,
+      error: row.error,
+      etag: row.etag,
+    };
+  },
+  //  Get last Fires using a handler function.
+  // This is also to avoid loading all rows in memory.
+  async getLastFires(db, {hookGroupId, hookId}, handler) {
```
djmitche, line 27: #3184

services/hooks/src/utils.js
```diff
@@ -0,0 +1,133 @@
...
+        return await handleResults(await fetchResults(res.continuationToken));
+      }
+    };
+
+    results = await handleResults(results);
+
+    return results;
+  },
+  definition(lastFire) {
+    return Promise.resolve({
```
djmitche, line 64: This could be async with just `remove`.  Or, does it need to be async at all?

services/hooks/src/utils.js
```diff
@@ -0,0 +1,133 @@
...
+    });
+  },
+};
+
+exports.lastFireUtils = lastFireUtils;
+
+const queueUtils = {
+  // Create a single instance, or undefined, from a set of rows containing zero
+  // or one elements.
+  fromDbRows(rows) {
```
djmitche, line 81: Looks like this is never called (and its friend in lastFireUtils)

services/hooks/src/utils.js
```diff
@@ -0,0 +1,133 @@
...
+    if (rows.length === 1) {
+      return exports.queueUtils.fromDb(rows[0]);
+    }
+  },
+  // Create a single instance from a DB row
+  fromDb(row) {
```
djmitche, line 87: Also never called

services/hooks/src/utils.js
```diff
@@ -0,0 +1,133 @@
...
+const lastFireUtils = {
+  // Create a single instance, or undefined, from a set of rows containing zero
+  // or one elements.
+  fromDbRows(rows) {
+    if (rows.length === 1) {
+      return exports.lastFireUtils.fromDb(rows[0]);
+    }
+  },
+  // Create a single instance from a DB row
+  fromDb(row) {
```
djmitche, line 13: This is only called from getLastFires, and the result is only used within that handler function.  All it does is change capitalization.  So updating the capitalization in the handler funcs would allow dropping this function.

services/hooks/src/listeners.js
```diff
@@ -203,9 +204,9 @@ class HookListeners {
    */
   async _reconcileConsumers() {
     let queues = [];
-    await this.Queues.scan({}, {
-      limit: 1000,
-      handler: (queue) => queues.push(queue)});
+    await queueUtils.getHooksQueues(this.db, queue => {
+      queues.push(queue);
+    });
```
djmitche, line 209: Since this just loads the entire table into memory, this could all be simplified to an un-paginated call to `db.fns.get_hooks_queues`.  The queues table is generally pretty small (strictly smaller than the list of hooks), and it's been this way for a long time, so I think this is OK for now, and makes the rest of this reconciliation simpler.

services/hooks/test/listeners_test.js
```diff
@@ -41,16 +42,17 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
   };
 
   const assertQueueEntities = async (...queues) => {
     const exp = queues.reduce(
       (acc, {hookId, bindings}) => Object.assign(acc, {[`${hookGroupId}/${hookId}`]: bindings}), {});
     const got = {};
-    await helper.Queues.scan({}, {
-      handler: ({hookId, bindings}) => got[`${hookGroupId}/${hookId}`] = bindings,
+
+    await queueUtils.getHooksQueues(helper.db, ({ hookId, bindings }) => {
```
djmitche, line 54: This, too, can be replaced with a call to `db.fns.get_hooks_queues`.

db/versions/0018.yml
helfi92: Given that it's okay to lose data, I suppose we can rewrite them to return an empty set.

djmitche: That works.  On the other hand, you've already written the migration to copy the data over..

PR #3187, opened 2020-07-09
Issue 3169 - merge workerTypeMetadata from provider with value in worker pool config

### LINKED ISSUE #3169: generic-worker workerTypeMetadata in worker pool definitions not in task log headers
**Describe the bug**
By specifying `workerTypeMetadata` generic-worker config setting in worker pool definitions, it should be possible to add information to all task log headers that run on a worker pool.

We have `workerTypeMetadata` specified in the worker pool definitions, like [this one](https://community-tc.services.mozilla.com/worker-manager/proj-taskcluster%2Fgw-ci-windows2012r2-amd64) however, it isn't making its way into task log headers, such as [here](https://community-tc.services.mozilla.com/tasks/TlFuJUi5Rfi48RuyY6sD_Q/runs/0/logs/live/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FTlFuJUi5Rfi48RuyY6sD_Q%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L2-23).

Either this is getting swallowed by worker-runner, or it could be a generic-worker bug. I think the latter is reasonably unlikely as this code hasn't really changed since it was created, so the probably cause at this point is that worker-runner is swallowing it.

I did notice that there is special treatment of `workerTypeMetadata` in the worker runner code. I couldn't see any tests for it though.

**Taskcluster version**
32.0.0 for both services and workers.

petemoore:
Github Bug/Issue: Fixes #3169 

I thought this might be a trivial fix, but I haven't understood worker-runner well enough yet to get the test working, so I'm parking for now.


djmitche:
This should be ready to go now.

djmitche:
worker-manager is an intermittent

PR #3190, opened 2020-07-09
Fix worker.remove calls in static provisioner

### LINKED ISSUE #3189: static provisioner calls worker.remove, which does not exist
Its `removeWorker` method appears to be untested.

djmitche:
Github Bug/Issue: Fixes #3189 

This fixes removeWorker and adds updateWorker, as described in #3189.


services/worker-manager/src/api.js
```diff
@@ -474,6 +474,67 @@ builder.declare({
   return res.reply(worker.serializable());
 });
 
+builder.declare({
+  method: 'post',
+  route: '/workers/:workerPoolId:/:workerGroup/:workerId',
+  name: 'updateWorker',
+  title: 'Update an existing Worker',
+  category: 'Workers',
+  stability: APIBuilder.stability.stable,
```
petemoore, line 483: Is this really `stable` on the first release?

services/worker-manager/src/api.js
```diff
@@ -474,6 +474,67 @@ builder.declare({
...
+  input: 'create-worker-request.yml',
+  output: 'worker-full.yml',
+  // note that this pattern relies on workerGroup and workerId not containing `/`
+  scopes: 'worker-manager:update-worker:<workerPoolId>/<workerGroup>/<workerId>',
+  cleanPayload: cleanCreatePayload,
+  description: [
+    'Update an existing worker.  The precise behavior of this method depends',
+    'on the provider implementing the given worker pool.  Some providers',
+    'do not support updating workers at all, and will return a 400 error.',
+  ].join('\n'),
```
petemoore, line 493: Can the description provide a bit of context why this was added, and the different types of implementations one might expect? Or maybe link to provider docs for more information. The current wording leaves a lot of open questions.

services/worker-manager/docs/providers.md
```diff
@@ -36,16 +36,21 @@ Any other error will result in a 500 error response and an error reported via ta
...
 
+The default implementation of all three methods throws an ApiError that will result in a 500 error to the caller.
+
 The provider's `createWorker` method is called with `workerPool` (an instance of the WorkerPool Azure entity class), `workerGroup`, `workerId`, and `input`.
 The `input` matches the `create-worker-request.yml` schema, and that schema can be adjusted to allow provider-specific parameters.
 The return value should be an instance of the Worker azure entity class.
 [Idempotency](../../dev-docs/idempotency.md) of this method is the responsibilty of the provider.
 
+The provider's `updateWorker` method is called with `{workerPool, worker, input}` where `workerPool` is the WorkerPool instance, `worker` is the Worker instance and `input` is the same as for `createWorker`.
+The return value should be an instance of the Worker azure entity class.
```
petemoore, line 52: Is the purpose of `updateWorker` to refresh configuration on the worker, when a change is made to the worker pool definition? If so, I think we could include this information in the api docs too. Perhaps `refreshWorkerPoolConfig` or `syncWorkerPoolConfiguration` might be a better name?

services/worker-manager/test/api_test.js
```diff
@@ -727,6 +727,48 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
+      await createWorker({});
+      const worker = await helper.workerManager.updateWorker(
+        workerPoolId,
+        workerGroup,
+        workerId, {
+          expires: taskcluster.fromNow('1 hour'),
+          capacity: 2,
+        });
+      assert.deepEqual(worker.capacity, 2);
+    });
```
petemoore, line 770: Should we test any race conditions here, such as a parallel worker termination and worker update?

services/worker-manager/src/api.js
djmitche: We really haven't paid much attention to these levels.  We have three:
 * "experimental" is typically used for "don't use this, we might kill it".  For example, the worker-info information in the queue has this stability level, and we do intend to remove it.  Ideally we wouldn't use this level very often!
 * "stable" is typically used for "normal" API methods.  Almost all methods have this stability, except `reregisterWorker`, and I'm not sure why that's the case.
 * "deprecated" is used for methods we'd rather users didn't call, but which are still used.

So I think "stable" is the correct choice here (and IMHO `reregisterWorker` should be "stable" too).

services/worker-manager/src/api.js
djmitche: This matches the createWorker call.  I can provide a pointer to the documentation for each provider type.

services/worker-manager/docs/providers.md
djmitche: No, that's not the purpose.

services/worker-manager/test/api_test.js
djmitche: I'm not sure what you have in mind here.  What's the failure case?

services/worker-manager/src/api.js
helfi92: It's possible that I set `reregisterWorker` to `experimental`. I always thought new API methods shouldn't be labeled as "stable" from the get go but I do understand your reasoning. I'm fine both ways.

services/worker-manager/src/api.js
helfi92: I agree. Seeing examples of scenarios where one would call this method will be helpful in general.

services/worker-manager/src/api.js
```diff
@@ -474,6 +474,67 @@ builder.declare({
...
+}, async function(req, res) {
+  // TODO: test
```
helfi92, line 495: TODO.

services/worker-manager/src/api.js
```diff
@@ -474,6 +474,67 @@ builder.declare({
...
+  const { workerPoolId, workerGroup, workerId } = req.params;
```
helfi92, line 496: Should there be a call somewhere to `req.authorize` to check scope permissions?

services/worker-manager/src/api.js
```diff
@@ -474,6 +474,67 @@ builder.declare({
...
+  const workerPool = await WorkerPool.get(this.db, workerPoolId);
+  if (!workerPool) {
+    return res.reportError('ResourceNotFound',
+      `Worker pool ${workerPoolId} does not exist`, {});
+  }
+
+  // NOTE: unlike in createWorker, we allow epxires to be in the past,
```
helfi92, line 503: epxires

services/worker-manager/src/api.js
```diff
@@ -474,6 +474,67 @@ builder.declare({
...
+  // as a way of "immediately expiring" a worker.
+
+  const provider = this.providers.get(workerPool.providerId);
```
helfi92, line 506: Should this use `await`?

djmitche: Not in this case -- it's just a wrapper around `Map.get`.  But, good eye :)

services/worker-manager/docs/providers.md
petemoore: Can you explain what the purpose is?

services/worker-manager/test/api_test.js
petemoore: I haven't really understood what the `updateWorker` call is for. I thought it was about refreshing the config on a running worker, but above you state that it isn't that. My suggestion was based on this misunderstanding, so it may not apply.

services/worker-manager/src/api.js
djmitche: See [the tc-lib-api docs](https://github.com/taskcluster/taskcluster/tree/main/libraries/api#scopes) -- `req.authorize` is only reqiured when the scope expression requires information that's not in `req.params`.  And if it's missed, then the API method fails very loudly.

services/worker-manager/src/api.js
petemoore: > We really haven't paid much attention to these levels. We have three:
> 
>     * "experimental" is typically used for "don't use this, we might kill it".  For example, the worker-info information in the queue has this stability level, and we do intend to remove it.  Ideally we wouldn't use this level very often!
> 
>     * "stable" is typically used for "normal" API methods.  Almost all methods have this stability, except `reregisterWorker`, and I'm not sure why that's the case.
> 
>     * "deprecated" is used for methods we'd rather users didn't call, but which are still used.
> 
> 
> So I think "stable" is the correct choice here (and IMHO `reregisterWorker` should be "stable" too).

The docs in https://github.com/taskcluster/taskcluster/tree/main/libraries/api#stability-levels suggest that `stable` is reserved for the situations when we are confident the API won't change. I'm not sure we benefit from making this a `stable` endpoint as we made need to tweak it after the first release. Perhaps we can mark it as `experimental` for the first release, and make it `stable` once we are sure it is fully suitable and requires no further modification?

djmitche:
I think I got everything.  I added a sentence to the method docs, but to reiterate, `updateWorker` is intended to allow updating the properties set when a static worker is created.  The most useful is `staticSecret`, for cases where a secret needs to be rotated.

services/worker-manager/docs/providers.md
petemoore: Ah, I think I've understood - it is like createWorker, but for when the worker is already defined.

services/worker-manager/src/api.js
djmitche: OK

djmitche:
I think I've addressed @petemoore's comments, and we won't overlap for a week, so I'll get this landed now.  We can address any necessary followups, of course.

PR #3200, opened 2020-07-13
Update go dependencies

petemoore:
An alternative for #3195 and #3197.

Those PRs have merge conflicts, and this is a more general upgrade that upgrades all our non-fixed dependencies.


go.mod
```diff
@@ -33,11 +33,11 @@ require (
 	github.com/pierrec/lz4 v2.5.2+incompatible // indirect
 	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
 	github.com/pkg/errors v0.9.1
-	github.com/sirupsen/logrus v1.5.0
+	github.com/sirupsen/logrus v1.6.0
 	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
-	github.com/streadway/amqp v0.0.0-20200108173154-1c71cc93ed71
-	github.com/stretchr/testify v1.5.1
+	github.com/streadway/amqp v1.0.0
```
djmitche, line 39: This looks like a breaking change, but if tests pass, let's do it.

petemoore: It is true that the taskcluster workers and services don't use the go amqp library, but we do create bindings in the go client for the event APIs that taskcluster provides. There could be people using them. See e.g. https://godoc.org/github.com/taskcluster/taskcluster-client-go/tcqueueevents#example-package--TaskclusterSniffer

petemoore: Also worth a read is the package docs for https://godoc.org/github.com/taskcluster/pulse-go/pulse

djmitche: We could probably find out if anyone's using them :)

PR #3202, opened 2020-07-13
roles table phase 2

### LINKED ISSUE #2938: Port the roles_entities table to use a normal DB table
https://bugzilla.mozilla.org/show_bug.cgi?id=1630014

Complications:
 * Roles must be validated and updated as a complete set of roles; concurrent operations to, for example, remove one role and add another role can lead to an inconsistent set of roles due to use of role parameterization.

djmitche:
Github Bug/Issue: Fixes #2938.

~I feel like I cheated here -- the resulting table is very similar to the entities table, albiet with an actual JSONB value rather than the `__buf` nonsense.~

I will take another crack at normalizing this into a one-row-per-role thing.  The reason I didn't do that to start with was kind of lame: it means storing the same etag over and over.


djmitche:
OK, well, that took all day, but I think I like the result.

db/versions/0023.yml
```diff
@@ -0,0 +1,138 @@
...
+    args: partition_key text, row_key text
+    returns: table (partition_key_out text, row_key_out text, value jsonb, version integer, etag uuid)
+    body: |-
+      begin
+        -- if no roles, return canned "empty" value, as the below expression will return NULL
+        perform 1 from roles limit 1;
+        if not found then
+          return query
+          select
+            'role',
```
helfi92, line 18: Should these have a named columns? (e.g., `'name' as partition_key`)

djmitche: I *think* this is the case where postgres ignores the aliases, so it's easier to leave them off than to include them and potentially have a mismatch when something is added or removed.  I'll double-check.

djmitche: Confirmed, adding `as foo` to one of these didn't cause any errors, so postgres is only looking at the order of the `select`ed columns.

PR #3203, opened 2020-07-13
update link to generic-worker and docker-worker;

mraethel:
Update link pointing to the archived generic-worker/docker-worker directories.


mraethel:
Thanks for the hint, the changes should now point to generic-worker/docker-worker in the reference section of the documentation.

PR #3205, opened 2020-07-14
Migrate web_server.session_storage_table_entities -> web_server.sessions

### LINKED ISSUE #3204: Port session_storage_table_entities to postgres phase 2
This only involves doing step 2 of the migration.

petemoore:
Github Bug/Issue: Fixes #3204


petemoore:
I'm going to have to park this now, as I am going on PTO.

I've implemented db methods for upserting a session (which is used for both creating a session, but also for "touching" one), for removing a session, and for fetching a session.

The only remaining pieces are getting the final tests working, and maybe a couple of extra tests might be useful for the db methods.

Other than that, there shouldn't be much more to do here.

Thanks!

petemoore:
The above 7 test failures seem to be the same issue, and I think it is due to the way the `AzureSessionStore.get` method handles failure cases in `services/web-server/src/login/AzureSessionStore.js` or the way the test calls it in `services/web-server/test/session_store_test.js`. I've run out of time to look into it, but I suspect the way the `get` method returns when a row can't be found is not the way the test expects it to, so it probably needs a little adjusting.

helfi92:
Continues in https://github.com/taskcluster/taskcluster/pull/3233.

PR #3207, opened 2020-07-15
downgrade v32 to node 12.18.0

djmitche:
Experimenting with v32.0.0


djmitche:
built in `taskcluster/taskcluster:v32.0.0-1-gac67d356d`

PR #3215, opened 2020-07-17
upgrade lodash 4

djmitche:


djmitche:
I don't think this is responsible for a failure on generic-worker :)

Issue #2933, opened 2020-05-29
port the queue's worker tables to use normal DB tables


djmitche:
This involves

* queue_worker_entities:
* queue_worker_type_entities:
* queue_provisioner_entities:

These tables might be quite large, and the information they contain is of a historical nature only.  It may make sense to drop some or all of that data during the transition (e.g., workers that haven't been seen in more than a day)

https://bugzilla.mozilla.org/show_bug.cgi?id=1630020

helfi92:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1626314#c26:
* queue_worker_entities: ~90135 rows
* queue_worker_type_entities: ~133 rows
* queue_provisioner_entities: ~38 rows

It will take 2/3 seconds to upgrade these 3 tables so probably no need to drop some or all of the data during the transition.

PR #3218, opened 2020-07-17
issue 2933: postgres phase 2 for queue_workers, queue_worker_types, queue_provisioners

milescrabill:
see: #2933

some notes:

- there was some struggling with the `update` functions and arrays, `JSON.stringify()` seemed like a reasonable option combined with defensive `||`'ing, wondering if some of that can be simplified
- I left workerinfo pretty intact, I was thinking that some of the constraints around update frequency probably aren't as relevant with postgres - thoughts?
- kind of scary that there's a quick and dirty way to remove all of these tables, but I see the purpose


db/versions/0037.yml
```diff
@@ -0,0 +1,333 @@
...
+  remove_queue_provisioners_table:
+    description: |-
+      Removes the queue provisioners table in case of backouts.
+    mode: write
+    serviceName: queue
+    args: []
+    returns: void
+    body: |-
+      begin
+        drop table queue_provisioners;
```
helfi92, line 333: Hm, I wonder if dropping the table would create issues. Truncating the table would probably make more sense here as well as for the other `remove_` functions defined in this PR.

services/queue/src/api.js
```diff
@@ -1746,15 +1731,11 @@ builder.declare({
...
-      workerType,
-      expires: Entity.op.greaterThan(new Date()),
-    }, true),
-    this.Provisioner.load({provisionerId}, true),
+    WorkerType.get(this.db, provisionerId, workerType),
+    Provisioner.get(this.db, provisionerId),
   ]);
 
-  if (!wType || !provisioner) {
+  if (!wType || wType.expires < new Date() || !provisioner) {
```
helfi92, line 1738: If we modify the get function to also take expires as an argument then there shouldn't be a need for this additional logic in the JS. I think we are trying to move away from adding additional business logic outside the db.

services/queue/src/workerinfo.js
```diff
@@ -1,53 +1,37 @@
...
   /**
-   * Mark a value as seen, only actually updating the Azure table row
+   * Mark a value as seen, only actually updating the postgres table row
    * occasionally. This takes the approach of updating the row immediately on
    * the first call, then not updating until `updateFrequency` has elapsed.
    * Thus the `expires` value on a row may be out-of-date by `updateFrequency`.
    *
    * Note that the cache is never purged of outdated entries; this assumes that
-   * the process is restarted on a daily basis, so there is not too much time
+   * the process is restarted on a daily 267985
```
djmitche, line 33: tapped the ol' yubikey?

db/versions/0037.yml
djmitche: I think we can just get rid of all this `remove_.._table` stuff.

db/test/versions/0036_test.js
```diff
@@ -0,0 +1,94 @@
...
+          stability: `some/stability${i}`,
+        }]))),
+    },
+    loadConditions: [
+      {condition: {provisionerId: 'aaa/provisioner', workerType: 'aaa/workertype'}, expectedSample: 'aabbccddeeff'},
+      {condition: {provisionerId: 'some/provisioner1', workerType: 'some/workertype1'}, expectedSample: 'samp1'},
+    ],
+    scanConditions: [
+      {condition: {}, expectedSamples: ['aabbccddeeff', 'samp0', 'samp1', 'samp2', 'samp3', 'samp4']},
+      {condition: null, expectedSamples: ['aabbccddeeff', 'samp0', 'samp1', 'samp2', 'samp3', 'samp4']},
```
djmitche, line 76: This should probably test scanning by `{provisionerId}`, too.

db/test/versions/0035_test.js
```diff
@@ -0,0 +1,110 @@
...
+          firstClaim: new Date(2),
+        }]))),
+    },
+    loadConditions: [
+      {condition: {provisionerId: 'aaa/provisioner', workerType: 'aaa/workertype', workerGroup: 'aaa/workergroup', workerId: 'aaa/workerid'}, expectedSample: 'aabbccddeeff'},
+      {condition: {provisionerId: 'some/provisioner1', workerType: 'some/workertype1', workerGroup: 'some/workergroup1', workerId: 'some/workerid1'}, expectedSample: 'samp1'},
+    ],
+    scanConditions: [
+      {condition: {}, expectedSamples: ['aabbccddeeff', 'samp0', 'samp1', 'samp2', 'samp3', 'samp4']},
+      {condition: null, expectedSamples: ['aabbccddeeff', 'samp0', 'samp1', 'samp2', 'samp3', 'samp4']},
```
djmitche, line 92: This should probably have a check for `{provisionerId, workerType, workerGroup}` and `{provisionerId, workerType}` too.

db/versions/0037.yml
```diff
@@ -15,14 +15,16 @@ methods:
...
+            entity_buf_encode(
+              jsonb_build_object(
+                'PartitionKey', encode_string_key(provisioner_id),
+                'RowKey', 'provisioner',
+                'provisionerId', provisioner_id,
+                'expires', expires,
+                'lastDateActive', last_date_active,
+                'stability', stability),
+              'description', description::text),
+            'actions', actions::text) as value,
```
djmitche, line 27: It looks like the lack of `actions` in the first commit is causing the tests to fail on that commit (I went back in my local copy to run them).  Even with `db/` checked out from the full PR, the tests in that commit still fail, suggesting that the queue will fail during the time between upgrading to this DB version 37 and deploying the new version of the JS services.

services/queue/src/data.js
```diff
@@ -189,320 +190,430 @@ class Task {
...
+  // Get a provisioner from the DB, or undefined
+  static async get(db, provisionerId) {
+    return Provisioner.fromDbRows(await db.fns.get_queue_provisioner(provisionerId));
+  }
+
+  // Call db.get_queue_provisioners.
+  // The response will be of the form { rows, continationToken }.
+  // If there are no provisioners to show, the response will have the
+  // `rows` field set to an empty array.
+  static async getProvisioners(
```
djmitche, line 242: It would be nice to replace this (and similar things for the other two classes) with `paginatedIterator` and `paginateResults` as in #3279.  But that's fine to leave as a followup.

services/queue/src/api.js
milescrabill: I've changed the `get_` functions, singular and plural, to accept `expires`. In one case, we're doing more complicated logic (checking quarantine and expired status), my impression was that would need either a special case db function or add significant complexity to the current ones.

djmitche: I think we can fix this up later, if necessary.  This API may not stick around for along if we start mixing queue and worker-manager information about workers.

db/test/versions/0035_test.js
djmitche: I was wrong about the first of these -- sorry!

djmitche:
I had a look and didn't see anything very concerning, except that step 1 wasn't passing as-is.  I fixed that up (but please pull and verify -- it looks like github lost that result when I rebased).

Other than that, I rebased and added a bunch of small commits to spruce things up.

helfi92:
I pulled 7f1a20a and ran the tests for queue. They passed. Let me know if I pulled the wrong commit. I will continue the review now.

services/queue/src/api.js
```diff
@@ -923,12 +920,7 @@ builder.declare({
     workerType,
   });
 
-  const worker = await this.Worker.load({
-    provisionerId,
-    workerType,
-    workerGroup,
-    workerId,
-  }, true);
+  const worker = await Worker.get(this.db, provisionerId, workerType, workerGroup, workerId, new Date());
```
helfi92, line 923: The previous query did not consider `expires`. Should it be set  to `null`? This comment applies to L1031 as well.

services/queue/src/api.js
```diff
@@ -1036,12 +1028,7 @@ builder.declare({
     );
   }
 
-  const worker = await this.Worker.load({
-    provisionerId: task.provisionerId,
-    workerType: task.workerType,
-    workerGroup,
-    workerId,
-  }, true);
+  const worker = await Worker.get(this.db, null, null, workerGroup, workerId, new Date());
```
helfi92, line 1031: `provisionerId` and `workerType` are both specified in the old query but they are set to `null` here. Should they be set?

services/queue/src/api.js
```diff
@@ -1745,13 +1731,10 @@ builder.declare({
...
+  const expires = new Date();
   const [wType, provisioner] = await Promise.all([
-    this.WorkerType.load({
-      provisionerId,
-      workerType,
-      expires: Entity.op.greaterThan(new Date()),
-    }, true),
-    this.Provisioner.load({provisionerId}, true),
+    WorkerType.get(this.db, provisionerId, workerType, expires),
+    Provisioner.get(this.db, provisionerId, expires),
```
helfi92, line 1737: There must have been a reason for adding `expires`. I will stop commenting on this :)

djmitche:
Yes, that's the right commit

services/queue/src/api.js
djmitche: They should -- workerGroup+workerId is only defined to be unique within the context of a workerPool.  Also, omitting those parameters makes this a table scan.

services/queue/src/api.js
imbstack: I think we've started to try to filter out expired results from the api so it makes sense to do here?

imbstack:
failure is an unrelated flake and the remaining tests are all worker tests. landing!

PR #3219, opened 2020-07-18
Test for collation by behavior, rather than name

djmitche:
I don't really know what's going on here, but my docker image suddenly started spelling this `en_US.UTF-8`.

```
     Error: Postgres database must have default collation en_US.utf8 or en_US.UTF8; this database is using en_US.UTF-8.
```

The [docs](https://www.postgresql.org/docs/11/collation.html) do say "By design, ICU will accept almost any string as a locale name and match it to the closest locale it can provide, using the fallback procedure described in its documentation."   But `en_US.UTF8` is a libc collation:

```
postgres=# \dOS+
...
 pg_catalog | en_US                  | en_US.utf8       | en_US.utf8       | libc     | yes            | 
 pg_catalog | en_US.utf8             | en_US.utf8       | en_US.utf8       | libc     | yes            | 
```

also, on the same DB that gave the error above, I don't see `en_US.UTF-8`.

```
postgres=# SELECT datcollate AS collation
postgres-#         FROM pg_database 
postgres-#         WHERE datname = current_database();
  collation  
-------------
 en_US.UTF-8
(1 row)
postgres=# SELECT * FROM pg_collation where collname ilike '%UTF%';
  collname  | collnamespace | collowner | collprovider | collencoding | collcollate | collctype  | collversion 
------------+---------------+-----------+--------------+--------------+-------------+------------+-------------
 C.UTF-8    |            11 |        10 | c            |            6 | C.UTF-8     | C.UTF-8    | 
 en_US.utf8 |            11 |        10 | c            |            6 | en_US.utf8  | en_US.utf8 | 
(2 rows)
```

something's weird here.  So let's just test that it gets the necessary bits right.  We really don't care about sorting of `ß`, for example, just of the ascii characters that we allow in identifiers and most critically in slugids.

### DIFF
--- a/changelog/QwlDxmO1Qwm6c2h6tIYTpw.md
+++ b/changelog/QwlDxmO1Qwm6c2h6tIYTpw.md
@@ -1,0 +1,4 @@
+audience: developers
+level: patch
+---
+The taskcluster-lib-postgres library now allows any Postgres collation that sorts ASCII characters correctly.
--- a/libraries/postgres/src/Database.js
+++ b/libraries/postgres/src/Database.js
@@ -387,15 +387,20 @@ class Database {
 
   async _checkDbSettings() {
     await this._withClient('admin', async client => {
-      const res = await client.query(`
-        SELECT datcollate AS collation
-        FROM pg_database 
-        WHERE datname = current_database()`);
-      const collation = res.rows[0].collation;
-      if (collation !== 'en_US.utf8' && collation !== 'en_US.UTF8') {
-        throw new Error(
-          `Postgres database must have default collation en_US.utf8 or en_US.UTF8; this database is using ${collation}.`,
-        );
+      // check the DB collation by its behavior, rather than by name, as names seem to vary.
+      for (const pair of ['aA', 'ab', 'Ab', 'aB', 'AB', '0a', '0A']) {
+        const res = await client.query(`select 1 where $1 >= $2`, [pair[0], pair[1]]);
+        if (res.rows.length > 0) {
+          const res = await client.query(`
+            SELECT datcollate AS collation
+            FROM pg_database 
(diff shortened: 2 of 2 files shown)


changelog/QwlDxmO1Qwm6c2h6tIYTpw.md
```diff
@@ -0,0 +1,4 @@
+audience: deployers
+level: patch
+---
+The taskcluster-li-postgres library now allows any postgres collation that sorts ASCII characters correctly.
```
helfi92, line 4: `s/li/lib/`

libraries/postgres/src/Database.js
```diff
@@ -387,15 +387,20 @@ class Database {
 
   async _checkDbSettings() {
     await this._withClient('admin', async client => {
-      const res = await client.query(`
-        SELECT datcollate AS collation
-        FROM pg_database 
-        WHERE datname = current_database()`);
-      const collation = res.rows[0].collation;
-      if (collation !== 'en_US.utf8' && collation !== 'en_US.UTF8') {
```
helfi92, line 395: That's definitely a weird behavior. Would it make more sense to just do `collation.toLowerCase()` and then only compare it with  `en_us.utf8` and `en_us.utf-8`? That seems a bit less complicated (i.e., creating an array and running an additional query) and more explicit.

djmitche: My first attempt at this stripped punctuation and lower-cased, and compared to `enusutf8` (hence the branch name), but that is just a guess at what is and is not acceptable.  I think the updated patch is a bit more precise about what we want: that the collation order the things we care about in the order we expect.  Which I think actually just means "not C" -- `sp_ES.utf16` would probably be fine.

PR #3221, opened 2020-07-18
docker-worker deps update

djmitche:
Mostly, this updates lodash

### DIFF
--- a/workers/docker-worker/package.json
+++ b/workers/docker-worker/package.json
@@ -80,7 +80,6 @@
     "ws": "^7.1.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


workers/docker-worker/package.json
```diff
@@ -80,7 +80,6 @@
     "ws": "^7.1.2"
   },
   "devDependencies": {
-    "github": "^13.1.0",
```
helfi92, line 83: Why was this removed?

djmitche: It's unused -- never require'd.  And it's an ancient version, which was pulling in old deps.

PR #3223, opened 2020-07-18
Fix DB errors about digest(text, unknown) not being found

### LINKED ISSUE #3222: digest function not found
Shortly after upgrading to version 0018, which is the first to use the `sha512` function, we see lots of

image

in the logs (sorry for the image, yay stackdriver).

The `digest(text, text)` function is in pgcrypto, and the `sha512` function is defined in version 0015 and is right out of the [docs](https://www.postgresql.org/docs/11/pgcrypto.html#id-1.11.7.34.5).  The index service is correctly finding index entries, so this function has to be working somehow.

djmitche:
Github Bug/Issue: Fixes #3222


djmitche:
@sciurus if this is causing issues, or is a concern, running the migration script in this PR outside of a migration will not cause issues (because we use `create or replace function`).

Issue #3216, opened 2020-07-17
post-azure-entities cleanup


djmitche:
* remove tc-lib-entities
* remove `services/*/azure.yml`
* remove azure permissoins from `services/*/scopes.yml`
* remove configuration for table names from `services/*/config.yml`

.. add yours here!

djmitche:
* make azure_crypto_key optional

djmitche:
* remove withEntity from tc-lib-testing

djmitche:
I'm going to wait to remove tc-lib-entities until 36.0.0 is released.  I'm not sure why, maybe just because I'm lazy..

djmitche:
(I want to think more about how to have checks that methods are only removed after a major version bump)

PR #3234, opened 2020-07-20
Postgres Phase 2 for auth's clients table (steps 1 and 2)

### LINKED ISSUE #3083: Port the clients_entities table to use a normal DB table
Clients are signed and encrypted.

djmitche:
Github Bug/Issue: Fixes #3083

Note that step 1 doesn't pass, because this table is signed.  We have no way to return results from the entity functions such that tc-lib-entities will successfully validate the signature.


helfi92:
> Note that step 1 doesn't pass, because this table is signed. We have no way to return results from the entity functions such that tc-lib-entities will successfully validate the signature.

You can comment out `signEntities: true,` and then run the tests. I think this should do it.

djmitche:
True, but that wouldn't work in production, unless we land a version with that option disabled and release it before landing this change.  Hmm, maybe that's not a bad idea.

helfi92:
Step 1 and 2 are done in the same PR. Step 1 won't run in production since the step 2 commit removes the entity from data.js isn't it?

djmitche:
> Step 1 and 2 are done in the same PR. Step 1 won't run in production since the step 2 commit removes the entity from data.js isn't it?

It's true, but step 1 will land when automation runs `yarn db:upgrade`, and step 2 won't land until the services are upgraded.

djmitche:
@helfi92 #3243 might be a way to address that.

djmitche:
I'm revising this to rely on v35.0.0 to be already deployed, disabling signing on this table.

djmitche:
TODO: adding some more tests for step 1.

djmitche:
This now has a clean step 1 and step 2, where as long as the services are updated to v35.0.0, they will continue to be able to read clients after the DB upgrade.

Testing turned up a bunch of bugs, now fixed.

db/test/fns/auth_test.js
```diff
@@ -1,54 +1,283 @@
...
+        'Some client...',
+        db.encrypt({ value: Buffer.from('sekrit', 'utf8') }),
+        expires,
+        false,
+        JSON.stringify(['scope1']),
+        false,
+      );
+      const [client1] = await db.fns.get_client('some-client');
+
+      // wait long enough for the last-date-uesd to change
```
imbstack, line 249: uesd

db/versions/0041-downgrade.sql
```diff
@@ -0,0 +1,45 @@
...
+        'lastRotated', last_rotated,
+        'deleteOnExpiration', delete_on_expiration
+      )::text) as value,
+    1 as version,
+    etag
+  from clients;
+
+  revoke select, insert, update, delete on clients from $db_user_prefix$_hooks;
+  drop table clients;
+  grant select, insert, update, delete on clients_entities to $db_user_prefix$_hooks;
```
imbstack, line 43: `$db_user_prefix$_hooks` copy-pasta

services/auth/src/api.js
```diff
@@ -265,48 +261,40 @@ builder.declare({
...
 
     if (client.description !== input.description ||
         client.expires.getTime() !== new Date(input.expires).getTime() ||
         !_.isEqual(client.scopes, scopes) ||
-        client.disabled !== 0 ||
+        client.disabled ||
         created < Date.now() - 15 * 60 * 1000) {
       return res.reportError('RequestConflict',
         'client with same clientId already exists, possibly an issue with retry logic or idempotency',
         {});
```
imbstack, line 296: this is fine but part of me feels like this could be an upsert with all of the logic here in the db instead. would that even be desirable?

services/auth/src/static-clients.js
```diff
@@ -0,0 +1,144 @@
...
+    } else {
+      return client;
+    }
+  });
+
+  // substitute the azureAccountId into the scopes
+  clients = clients.map(client => ({
+    ...client,
+    scopes: client.scopes.map(sc => sc.replace(/\${azureAccountId}/g, azureAccountId)),
+  }));
```
imbstack, line 69: do we still need this bit?

djmitche: I'll fix it in #3216.

djmitche:
Everything's green but meta-generate b/c I forgot.. pushing now.

djmitche:
Oh, looks like I also forgot to supply dbCryptoKey to the DB.

PR #3233, opened 2020-07-20
Migrate web_server.session_storage_table_entities -> web_server.sessions 

helfi92:
Resolves #3204.

PR #3236, opened 2020-07-20
Expire session storage

helfi92:
It's not like we had millions in there but turns out we never expired them.


helfi92:
Pushed a [commit](https://github.com/taskcluster/taskcluster/pull/3233/commits/30d0c0f55ee35da1a9f4d728171b3b58c31ad8d3) that would have this fixed to the session storage migration PR.

PR #3239, opened 2020-07-20
Use reported date for ordering of worker pool errors

### LINKED ISSUE #3238: Worker Pool Errors should be sorted by date
**Describe the bug**
Currently worker pool errors are sorted in the UI by date, but this only applies on a per-page basis. We're really returning chunks of errors by errorId (I think) and then sorting the chunk by date. I'm not sure if this is fixed in the v2 migration we just landed. If not, let's fix it there now.

**To Reproduce**

1. Go to https://firefox-ci-tc.services.mozilla.com/worker-manager/gecko-t%2Ft-win10-64-gpu-s/errors
2. Click on the next page button
3. See that the most recent error is always from more recently than the bottom of the previous page

**Expected behavior**
We should sort these by recency since that's the only useful way to view them.

**Taskcluster version**
33.0.0-1-g4c4ad4262

**Screenshots**

 - OS: [e.g. windows]

**Additional context**
This made debugging https://bugzilla.mozilla.org/show_bug.cgi?id=1645994 difficult

imbstack:
Going to skip changelog since if we land this before we make a release the earlier changelog for this table is just as valid

Github Bug/Issue: Fixes #3238


db/versions/0031.yml
```diff
@@ -238,7 +238,7 @@ methods:
           worker_pool_errors.description,
           worker_pool_errors.extra
         from worker_pool_errors
-        order by error_id
+        order by worker_pool_errors.reported desc
         limit get_page_limit(page_size_in)
         offset get_page_offset(page_offset_in);
```
djmitche, line 243: ...why is there no where clause here?

djmitche: whoops, I meant to look into that before submitting :)

Filed #3240.

imbstack:
If this is green in the morning, can we land this and then @milescrabill can land the other updates on top of it!

djmitche:
green -> merging as directed

PR #3241, opened 2020-07-21
Migrate session, authorization code, access token tables

### LINKED ISSUE #2930: port authorization_codes_table_entities & access_token_table_entities to use normal DB tables
This service currently uses azure-entities emulation to store its data in the DB. This bug entails writing a new DB version that will smoothly migrate the service to use more typical DB tables with nicely-named columns and easy joins.

https://bugzilla.mozilla.org/show_bug.cgi?id=1630022

Complications:
 * tables are encrypted and signed

### LINKED ISSUE #3148: port github_access_token_table_entities to use normal DB tables

helfi92:
Fixes #3148.
Fixes #2930.


changelog/issue-3204.md
```diff
@@ -0,0 +1,17 @@
+audience: deployers
+level: major 
+reference: issue 3148
```
djmitche, line 3: It may be worth noting here that users will also be signed out of Taskcluster when this upgrade runs and (if I'm not mistaken) unable to sign in until the services upgrade has been completed.  Or we could wait until the next major release to land this (in addition to reimplementations of all of the entities functions), and remove the signing in the current release.  Which would you prefer?

The GithubAccessToken table is only used for the Github login strategy, so won't affect firefox-ci (and I think the risk is OK for community).  So we don't need to revert that PR, at least!

db/versions/0033.yml
```diff
@@ -0,0 +1,71 @@
+version: 33
+migrationScript: 0033-migration.sql
+downgradeScript: 0033-downgrade.sql
+methods:
```
djmitche, line 4: This should probably stub out the entities methods so that they fail more naturally, such as returning nothing for the read methods and throwing an error with a useful message for create (similar for the other migrations).

db/versions/0033.yml
```diff
@@ -0,0 +1,71 @@
...
+    returns: void
+    body: |-
+      begin
+        delete from sessions
+        where
+          sessions.hashed_session_id = hashed_session_id_in;
+      end
+  expire_sessions:
+    description: |-
+      Expire sessions that come before the current time.
```
djmitche, line 55:
```suggestion
      Delete sessions that expire before the current time.
```

services/web-server/test/session_store_test.js
```diff
@@ -119,13 +118,10 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
         store.touch('foo', { cookie: { maxAge: 300 } }, done);
       });
     });
-    test('it should error when touching the session', async function () {
```
djmitche, line 122: Was this intended to test that touching a nonexistent session fails?  Is that an important characteristic?

services/web-server/test/session_store_test.js
```diff
@@ -119,13 +118,10 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
+    test('it should not error when touching the session', async function () {
```
imbstack, line 121: perhaps this means I didn't read this review closely enough but why does this change from should-error to should-not-error?

services/web-server/test/session_store_test.js
helfi92: No, I think that was just testing the error message.

services/web-server/test/session_store_test.js
helfi92: The reason I removed this was because we were using `session_add` in the `touch` method which will either create or update an existent entry. Now that I look closer, the touch method should be using `session_load` rather than `session_add`. I'll update this. Thanks for flagging this. This could have been a bug.

services/web-server/test/session_store_test.js
helfi92: I retract my comment. I need to fix this.

djmitche:
^^ but some red X's to fix..

helfi92:
I can rebase this once https://github.com/taskcluster/taskcluster/pull/3249 is merged.

djmitche:
Renumbered and rebased and green.

PR #3243, opened 2020-07-22
Stop signing entities

djmitche:
This disables signing on the JS side for all (remaining) tables.

The issue this addresses is this: when a signed table is migrated, it's impossible to calculate signatures from the new table that will satisfy old code still calling the entities functions.  So any code still calling those functions will immediately fail until a newer version of that code is deployed.

If we land this change and ship a TC release, then once that release is deployed, there will be no code expecting valid signatures, and we can freely update the DB to a version that no longer returns valid signatures without causing an outage.


djmitche:
I wonder if we should remove the migration of the hooks service's tables in master, and instead set `signedEntities: false` for that service as well, then re-land the migrations after cutting a release.  Thoughts?

helfi92:
> I wonder if we should remove the migration of the hooks service's tables in master, and instead set signedEntities: false for that service as well, then re-land the migrations after cutting a release. Thoughts?

I would be fine if things like hooks don't work for a minute especially when we have retries already setup for actions. I'm also fine if we revert the PR, add the patch, and then merge it back. I'll let you decide given you have more knowledge about the hooks service in general and its usage.

When I originally brought up the issue in https://github.com/taskcluster/taskcluster/pull/3234#issuecomment-662606281, it was mainly to see the CI tests passing to get that initial confidence of the db migration step.

djmitche:
I think we're *probably* OK with Hooks, since the DB upgrade is pretty quick.  If we do need to roll back the services deployed, we'll need to also roll back the DB at least as far as all of the hooks tables, but that's probably OK.

I'll fix

```
      AssertionError [ERR_ASSERTION]: Don't specify options.signingKey when entities aren't signed!
```

I'm surprised that didn't cause issues for auth!

djmitche:
Reflecting on this a bit more, I think we should back out the hooks change to avoid extra risk.  Then we can upgrade everything in each deployment (db and services) to v35 before moving on to v36 that contains the signed rows.  This is not important for web-server since its signed rows can just be deleted.

So v35 would contain:
 * fix for #2340
 * this PR
 * backed-out hooks tables

and v36 would contain:
 * hooks tables
 * auth.clients

helfi92:
Sounds good to me.

djmitche:
OK, I included the reverts in here as well (one of which required a bunch of renumbering..).  I'll do the git grunt-work to re-land these later.

services/web-server/src/data/AuthorizationCode.js
```diff
@@ -5,7 +5,7 @@ const AuthorizationCode = Entity.configure({
   version: 1,
   partitionKey: Entity.keys.ConstantKey('authorizationCodes'),
   rowKey: Entity.keys.StringKey('code'),
-  signEntities: true,
+  signEntities: false,
```
helfi92, line 8: This will create a conflict with the existing web-server migration PR. If changing this is not necessary for this service since we don't care about old data then maybe we can avoid changing it here.

djmitche: Good spot, thank you -- I meant to remove changes for web-server.  So many things going on :(

changelog/H4wb9nDhSrekI1HnI5LCDw.md
```diff
@@ -0,0 +1,4 @@
+audience: deployers
+level: minor
```
djmitche, line 2: I'll change this to major.

PR #3246, opened 2020-07-22
issue 3240: support listing workerPoolErrors by workerPoolId

### LINKED ISSUE #3240: listWorkerPoolErrors should only return errors for that workerPool
https://github.com/taskcluster/taskcluster/blob/994a671a840bad712a2d63ab888698b20b7d8c6a/services/worker-manager/src/api.js#L326-L347

there is no `errorId` param to this function.  And getWorkerPoolErrors only takes two parameters:

https://github.com/taskcluster/taskcluster/blob/994a671a840bad712a2d63ab888698b20b7d8c6a/services/worker-manager/src/data.js#L188-L193

And `db.fns.get_worker_pool_errors` always returns the whole table, rather than a particular pool's errors:

https://github.com/taskcluster/taskcluster/blob/994a671a840bad712a2d63ab888698b20b7d8c6a/db/versions/0031.yml#L227

let's get that fixed up before the next release!

milescrabill:
Github Issue: Fixes #3240


db/versions/0035.yml
```diff
@@ -0,0 +1,33 @@
...
+          worker_pool_errors.worker_pool_id,
+          worker_pool_errors.reported,
+          worker_pool_errors.kind,
+          worker_pool_errors.title,
+          worker_pool_errors.description,
+          worker_pool_errors.extra
+        from worker_pool_errors
+        where
+          (worker_pool_errors.worker_pool_id = worker_pool_id_in or worker_pool_id_in is null) and
+          (worker_pool_errors.error_id = error_id_in or error_id_in is null)
```
imbstack, line 29: Interesting, I feel like we've generally been making two functions `get_worker_pool_error` and `get_worker_pool_errors` but I can't see why not to just do it the way you've done it here with one function. I actually like it more tbh

djmitche:
(just waiting for green and I'll push the button)

PR #3250, opened 2020-07-24
[db] Add support for named arguments

### LINKED ISSUE #2928: try to make db fns take named rather than positional arguments
Right now db fns take ordered arguments due to how postgres takes arguments. This is mostly fine but many of our fns have 7 or even more arguments and this gets unwieldy with chances of transposition high. Theoretically tests should catch cases where we do this by accident but it would be nice if we just ensured it at the library interface level by taking named arguments. This could also allow us to write better error messages.

If this ends up being a lot of work, let's cancel it and live with positional args however if it is relatively straightforward it feels like an easy win now before we have a lot of uses of db fns!

helfi92:
Fixes #2928.

### DIFF
--- a/changelog/issue-2928.md
+++ b/changelog/issue-2928.md
@@ -1,5 +1,5 @@
 audience: developers
-level: minor
+level: patch 
 reference: issue 2928
 ---
 taskcluster-lib-postgres now allows calling stored functions with named
--- a/changelog/issue-2928.md
+++ b/changelog/issue-2928.md
@@ -1,0 +1,6 @@
+audience: developers
+level: minor
+reference: issue 2928
+---
+taskcluster-lib-postgres now allows calling stored functions with named
+arguments.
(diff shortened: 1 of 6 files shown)


libraries/postgres/README.md
```diff
@@ -300,6 +300,11 @@ The binary payload is derived by first base64-decoding each `__bufN_val` propert
 That binary payload, in turn, is defined in the [`EncryptedBaseType` constructor](https://github.com/taskcluster/azure-entities/blob/c6f63e3553c71f0859a5d6338ce5e7c7eb8c9671/src/entitytypes.js#L716-L725).
 It is the concatenation of a 128-bit (16-byte) random initialization vector (`iv`) and the ciphertext produced by aes-256 in CBC mode.
 
+## Named Arguments with Stored Functions
+
+This library supports calling stored functions with named arguments.
+For this to work, a stored function should have all of this arguments end with "_in".
```
djmitche, line 306:
```suggestion
For this to work, a stored function should have all of its arguments end with "_in".
```

djmitche: An example of what this looks like on the JS side would be good, too.

changelog/issue-2928.md
```diff
@@ -0,0 +1,6 @@
+audience: developers
+level: minor
```
djmitche, line 2: probably patch since this is internal-only

PR #3253, opened 2020-07-24
[db] Allow updating signature of a db method

### LINKED ISSUE #2970: allow using the same name for postgres functions that have changed signature
Currently to update the signature of a function, you have to change the name of the function so that the old and new one can both be stored in the db for migration purposes. This is good but a bit of a leaky abstraction. We should instead have a `version` field of functions that needs to be incremented when signatures change. We can enforce with a test that this happens. Then, in js code, we always can use the latest version of the function.

helfi92:
Fixes #2970.

### DIFF
--- a/changelog/issue-2970.md
+++ b/changelog/issue-2970.md
@@ -1,0 +1,6 @@
+audience: developers
+level: patch
+reference: issue 2970
+---
+The DB schema now allows using the same name for postgres functions that have
+changed signature.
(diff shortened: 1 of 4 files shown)


imbstack:
was a consensus formed on this in slack? happy to review whenever if it's been decided how to move forward

helfi92:
Closing this. See https://github.com/taskcluster/taskcluster/issues/2970#issuecomment-664454709 for reason.

PR #3254, opened 2020-07-24
Fix CI test

helfi92:
I may have merged without paying attention to CI.

### DIFF
--- a/libraries/postgres/test/database_test.js
+++ b/libraries/postgres/test/database_test.js
@@ -820,8 +820,8 @@ helper.dbSuite(path.basename(__filename), function() {
         async () => {
           await db.fns.add_from_numbers({ a: 1, b: 2 });
         },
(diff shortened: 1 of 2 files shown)


libraries/postgres/test/database_test.js
```diff
@@ -820,8 +820,8 @@ helper.dbSuite(path.basename(__filename), function() {
         async () => {
           await db.fns.add_from_numbers({ a: 1, b: 2 });
         },
-        /function add_from_numbers(unknown) does not exist/
-      )
+        UNDEFINED_FUNCTION,
```
imbstack, line 823: I thought this needed to be a regex if it was trying to find a string

helfi92: You're absolutely right. This was a false positive. Fixed!

Issue #3278, opened 2020-07-27
Support pagination by primary key instead of size/offset


djmitche:
size/offset is OK for the UI, but problematic for internal scans.  If rows are inserted, it can cause the same row to be processed twice, while deletion of rows can cause rows to be skipped.  It's also not very performant for the DB, as it must read and discard `offset` rows before it starts returning anything.

The size/offset was a simple approach and got us pretty far with minimal downside, but let's do better.

The preferred approach is to store the "greatest" primary key from the last query, and on the next query filter for rows greater than that value.

This approach doesn't double-count any rows, nor skip any rows that are present when the query occurs.  It also allows postgres to use the primary key's index to quickly find the first row for each page.

djmitche:
Fixed in #4034.  As suggested there, the support for index-based pagination is now in place, but I don't think there's a call for rewriting all uses of offset/limit-based pagination.  Let's do it as-needed.

PR #3279, opened 2020-07-27
Introduce `paginatedIterator` and use it

### LINKED ISSUE #3184: make a general utility for paginating through a table
We are adding a lot of things that query a table and call a handler for each row:

```javascript
  await lastFireUtils.getLastFires(
    this.db,
    { hookGroupId: req.params.hookGroupId, hookId: req.params.hookId },
    item => {
      if (item.taskCreateTime > latest.taskCreateTime) {
        latest = item;
      }
    },
  );
```

Let's write a general utility for doing so, similar to `paginatedResults` but designed for internal use.  That should help remove some of the code otherwise appearing in `data.js` and/or `fooUtils.js`.

djmitche:
This function calls paginated DB functions and translates the results
into an async iterator of rows.

Github Bug/Issue: Fixes #3184

I changed most custom pagination functions to use this.  The one I didn't change is the function scanning for expiration of artifacts, which has a workaround for #3278 in place.  So let's wait until that issue is fixed to change that function.

### DIFF
--- a/libraries/postgres/README.md
+++ b/libraries/postgres/README.md
@@ -234,6 +234,29 @@ table_name:
 Column types are a "stripped down" version of the full Postgres type definition, including only a simple type name and if necessary the suffix `not null`.
 Primary keys, constraints, defaults, sequences, and so on are not included.
 
+## Pagination
+
+Database functions which return many rows typically have `page_size int, page_offset int` as their last two arguments, and return a page of the given size at the given offset.
+The taskcluster-lib-api function `paginatedResults` is useful for translating such paginated results into an API response.
+For other users in Taskcluster services, this library provides `paginatedIterator` to convert paginated results into an async iterator.
+
+```javascript
+const {paginatedIterator} = require('taskcluster-lib-postgres');
+
+const doTheThings = async () => {
+  for await (let row of paginatedIterator({
+    fetch: async (size, offset) => db.fns.get_widgets(..., size, offset),
+    size: 1000, // optional, defaults to 1000
+  })) {
+    // ..do something with `row`
+  }
+}
+```
+
+It's important to remember that each paginated fetch is a different transaction, and so the content of the DB may change from call to call.
+An "offset" into a set of results that is rapidly changing size is not accurate, and can easily return the same row twice or skip a row.
+For UI purposes, this is not a big deal, but may cause issues for other uses.
+
 ## Encryption
 
 As described above, secret data is stored in a "crypto container".
--- a/libraries/postgres/src/util.js
+++ b/libraries/postgres/src/util.js
@@ -76,3 +76,40 @@ exports.annotateError = (query, err) => {
     }
   }
 };
+
+/**
+ * Call the given fetch method with a given size and offset to fetch data, and
+ * return an async iterator that will yield each row in turn.
+ */
+exports.paginatedIterator = ({fetch, size = 1000}) => {
+  return {
+    [Symbol.asyncIterator]() {
+      let offset = 0;
+      let done = false;
+      let rows = [];
+
+      const next = async () => {
+        if (rows.length > 0) {
+          return {value: rows.shift(), done: false};
+        }
+        if (done) {
+          // If this iterator has already "finished", just return that status
+          // without calling out to the DB again (and potentialyl returning more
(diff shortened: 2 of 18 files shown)


libraries/postgres/README.md
```diff
@@ -234,6 +234,29 @@ table_name:
...
+The taskcluster-lib-api function `paginatedResults` is useful for translating such paginated results into an API response.
+For other users in Taskcluster services, this library provides `paginatedIterator` to convert paginated results into an async iterator.
+
+```javascript
+const {paginatedIterator} = require('taskcluster-lib-postgres');
+
+const doTheThings = async () => {
+  for await (let row of paginatedResults({
+    fetch: async (size, offset) => db.fns.get_widgets(..., size, offset),
+    size: 1000, // optional, defaults to 500
```
helfi92, line 249: Did you mean 1000 as the default rather than 500?

libraries/postgres/README.md
```diff
@@ -234,6 +234,29 @@ table_name:
...
+
+Database functions which return many rows typically have `page_size int, page_offset int` as their last two arguments, and return a page of the given size at the given offset.
+The taskcluster-lib-api function `paginatedResults` is useful for translating such paginated results into an API response.
+For other users in Taskcluster services, this library provides `paginatedIterator` to convert paginated results into an async iterator.
+
+```javascript
+const {paginatedIterator} = require('taskcluster-lib-postgres');
+
+const doTheThings = async () => {
+  for await (let row of paginatedResults({
```
helfi92, line 247: Should this be paginatedIterator?

libraries/postgres/src/util.js
```diff
@@ -76,3 +76,37 @@ exports.annotateError = (query, err) => {
...
+      let offset = 0;
+      let done = false;
+      let rows = [];
+
+      const next = async () => {
+        if (rows.length > 0) {
+          return {value: rows.shift(), done: false};
+        }
+        if (done) {
+          return {done};
```
helfi92, line 96: It looks like we return `{ done: true }` `if rows.length === 0`. Does L96 ever run?

services/hooks/src/api.js
```diff
@@ -125,27 +125,24 @@ builder.declare({
...
-      }
-    },
-  );
+  let latest = {task_create_time: new Date(1970, 1, 1)};
+  const fetch = (size, offset) => this.db.fns.get_last_fires(hookGroupId, hookId, size, offset);
+  for await (let item of paginatedIterator({fetch})) {
+    if (item.task_create_time > latest.task_create_time) {
+      latest = item;
+    }
+  }
```
helfi92, line 134: Love this!

services/worker-manager/src/api.js
```diff
@@ -340,17 +337,20 @@ builder.declare({
...
-  );
-  const result = {
-    workerPoolErrors: errors.map(e => e.serializable()),
-    continuationToken,
-  };
+  const {continuationToken, rows} = await paginateResults({
+    query: req.query,
+    fetch: (size, offset) => this.db.fns.get_worker_pool_errors_for_worker_pool(
+      errorId || null,
+      workerPoolId || null,
```
helfi92, line 344: Reminder that we can use named arguments :)

libraries/postgres/src/util.js
djmitche: I think it's allowed to call `iter.next()` after it has returned `done`, so this case saves an additional DB fetch in that situation.

services/worker-manager/src/api.js
djmitche: I was mostly trying to forward-port what we already had.  I'll fix this call.

libraries/postgres/src/util.js
helfi92: Got it. A comment there would be helpful for posterity.

services/worker-manager/src/api.js
helfi92: Ah. Sorry about that. I retract my comment.

libraries/postgres/src/util.js
djmitche: done, and I'll push with that change.

PR #3280, opened 2020-07-27
Secrets migration phase 2

### LINKED ISSUE #2931: Port the secrets_entities table to use a normal DB table
This service currently uses azure-entities emulation to store its data in the DB. This bug entails writing a new DB version that will smoothly migrate the service to use more typical DB tables with nicely-named columns and easy joins.

https://bugzilla.mozilla.org/show_bug.cgi?id=1630021

Complications:
 * tables are encrypted and signed

imbstack:
This actually simplifies the secrets service yet further. I'm happier and happier with this db transition we've made every day.

Github Bug/Issue: Fixes #2931


changelog/issue-2931.md
```diff
@@ -0,0 +1,11 @@
+audience: deployers
+level: major
+reference: issue 2931
+---
+The `secrets_entities` table has been migrated to a relational table `secrets`. The secrets service now requires an additional environment variable `DB_CRYPTO_KEYS` to be set which is a JSON array where each element in an obejct of the form:
```
djmitche, line 5:
```suggestion
The `secrets_entities` table has been migrated to a relational table `secrets`. The secrets service now requires an additional environment variable `DB_CRYPTO_KEYS` to be set which is a JSON array where each element in an object of the form:
```
@helfi92 and I were talking about this in #3299 -- my understanding was that we wanted to continue to support AZURE_CRYPTO_KEYS, with the option to start using DB_CRYPTO_KEYS when it's time to rotate a secret.  This changelog suggests that AZURE_CRYPTO_KEYS support is being dropped, and also doesn't mention that if existing keys are to remain readable, then the DB_CRYPTO_KEYS properties must have very specific values.

db/test/versions/0041_test.js
```diff
@@ -0,0 +1,78 @@
...
+    loadConditions: [
+      {condition: {name: 'secret/1'}, expectedSample: 'simple1'},
+    ],
+    scanConditions: [
+      {condition: {}, expectedSamples: _.range(10).map(i => `simple${i}`)},
+    ],
+    notFoundConditions: [
+      {condition: {name: 'notathing'}},
+    ],
+    notImplemented: ['modifications'],
```
djmitche, line 65: Not implemented, but there are modifications below?

db/versions/0041.yml
```diff
@@ -0,0 +1,199 @@
...
+      end
+  secrets_entities_modify:
+    serviceName: secrets
+    description: See taskcluster-lib-entities
+    mode: write
+    args: partition_key text, row_key text, properties jsonb, version integer, old_etag uuid
+    returns: table (etag uuid)
+    body: |-
+      begin
+        raise exception 'not implemented';
```
djmitche, line 71: AIUI this is not implemented because the only modify occurs on conflict, and that's handled as an upsert in `secrets_entities_create` -- is that right?

imbstack: yeah. secrets never called `entity.modify` actually. always just overwrote

PR #3291, opened 2020-07-28
Add a Rust client

djmitche:
(@walac got a nice start on a rust client -- let's keep a pointer to it in a PR!)

## TODO

* [x] test it in TC
* [x] check formatting in TC
* [x] update client version on release
* [x] use tokio to sleep (since that's what reqwest is using) rather than async-std
* [x] add more tests, specifically for use with credentials
* [x] generate code (https://github.com/carllerche/codegen?) for API methods (to be decided: do we generate structs from schemas as well? serde_json handles unstructured JSON pretty well..)
* [x] support creating temporary credentials
* [x] write and generate a README
* [x] fail on 4xx
* [x] Do not use `failure` (it's deprecated) -- can we just use `reqwest.Error`? (at any rate, support the next item)
* [x] Support finding the response code in a failed request (404, 409, etc.)
* [x] bewit support
* [x] test signing requests with a body
* [x] test bewits
* [x] define `host`, `port`, and hawk `Credentials` once when creating a new Client
* [x] add support for `ext`
* [x] ~~`from_now`, other utilities provided by clients~~ (not needed in Rust, just like Go)
* [x] check that redirects are not followed
* [x] ignore clients/client-rust/target in eslint (otherwise it finds lint issues in rustdoc)
* [x] ensure that redirect response bodies are returned (#4058)
* [x] test paths and query args containing escapable characters
* [x] test calls, signed URLs with temporary credentials
* [x] add a builder pattern for creating Clients
* [x] move authorized_scopes to client, and include them in ext and test them 
* [x] use a specific version of rust and include that in the dev docs, etc.

Followups
* [ ] test request timeouts (httptest appears not to support this; maybe open a raw socket?)
* [ ] test signed URLs with authorized scopes
* [ ] ~~add Debug implementation for Client (that does not reveal key information)~~ (turns out there's not much useful info left)
* [ ] push to crates.io on publish (and to a staging crate on staging publish)


djmitche:
In particular, it might be nice to pull out some of the preliminaries here into other rust crates -- anyone want to own the `slugid` crate?

walac:
Notice this is heavily based on the Golang implementation. iirc, the only missing part is write the code generator.

djmitche:
Awesome, @walac, and good to hear from you!

djmitche:
The code generation went pretty well.  Like the JS and Python clients, this treats the payloads and responses as arbitrary JSON.  Serde makes it pretty easy to work with this, as seen in the tests.  In fact, many of those `.unwrap()` would turn into `?` making it even cleaner.

clients/client-rust/README.md
```diff
@@ -0,0 +1,17 @@
+# Taskcluster Client for Rust
+
+This library is a complete interface to Taskcluster in Python.  It provides
```
walac, line 3: nit: s/Python/Rust/

djmitche: Lol, good spot :)

djmitche:
I think this is ready for review.

If you want to start with the user perspective, run `cargo doc -p taskcluster --open` in the `clients/client-rust` directory and you'll get to see the docs.  Then, look at the implementation of those types and functions.  And finally peruse the tests :)

clients/client-rust/src/lib.rs
```diff
@@ -0,0 +1,275 @@
+/*!
+# Taskcluster Client for Rust
+
+For a general guide to using Taskcluster clients, see [Calling Taskcluster
+APIs](https://docs.taskcluster.net/docs/manual/using/api).
+
+This client is a convenience wrapper around `reqwest` that provides named functions for each API
```
jwhitlock, line 7: ``reqwest`` is a better name for SEO than Python's ``requests``, and so much better than go's ``got``. +1 to Rust.

jwhitlock (approved):
My rust is weak, so I'm glad @walac is giving feedback. The generated code and documentation looks good, and the tests look complete. Thanks @djmitche!

clients/client-rust/src/lib.rs
djmitche: but so hard to type!

djmitche:
@imbstack do you want to have a look as well?

clients/client-rust/src/credentials.rs
```diff
@@ -0,0 +1,376 @@
...
+        duration: Duration,
+        scopes: impl IntoIterator<Item = impl AsRef<str>>,
+    ) -> Result<Credentials, Error> {
+        self.create_temp_creds_inner(Some(temp_client_id), duration, scopes)
+    }
+
+    /// Similar to `create_named_temp_creds`, but creating unnamed credentials.  This approach is
+    /// still supported but users are encouraged to create named credentials when possible to
+    /// support auditability.
+    pub fn create_temp_creds(
```
imbstack, line 142: would `create_temp_creds` and `create_unnamed_temp_creds` better indicate the desired one? I'm happy either way

imbstack (approved):
is there a way to move the autogenerated files to path where we can hide them in reviews?

This is great though!

clients/client-rust/src/client.rs
```diff
@@ -0,0 +1,791 @@
...
+        self
+    }
+
+    /// Set the path_prefix; this will be included between the root URL and the path given to
+    /// `request`, `make_url`, and `make_signed_url`.  This is typically used when building a
+    /// client that will address a single service, such as `api/queue/v1/`.  The path prefix
+    /// must end with a `/` character.
+    pub fn path_prefix<S: Into<String>>(mut self, path_prefix: S) -> Self {
+        let path_prefix = path_prefix.into();
+        assert!(path_prefix.ends_with('/'));
```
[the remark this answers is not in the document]
djmitche: Good question.  I'll change that in a followup.

djmitche:
> is there a way to move the autogenerated files to path where we can hide them in reviews?

Yeah, I'll do that in a followup, too

PR #3299, opened 2020-07-28
pass azureCryptoKeys to db setup

djmitche:
This was missed for these two services.  @imbstack just pointed out I missed it in #3234, so I'm confident it will be remembered in the secrets service :)

This causes a minor issue for v35.0.0, which included the github access tokens table.  It won't start without DB_CRYPTO_KEYS set.  I updated the relase notes in the GitHub release appropriately.  It was a major release, so that's probably enough.

### DIFF
--- a/services/hooks/src/main.js
+++ b/services/hooks/src/main.js
@@ -45,6 +45,7 @@ const load = loader({
       serviceName: 'hooks',
       monitor: monitor.childMonitor('db'),
       statementTimeout: process === 'server' ? 30000 : 0,
+      azureCryptoKey: cfg.azure.cryptoKey,
       dbCryptoKeys: cfg.postgres.dbCryptoKeys,
     }),
   },
--- a/changelog/UqMeg67CQvmStJVJND1veg.md
+++ b/changelog/UqMeg67CQvmStJVJND1veg.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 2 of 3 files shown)


services/hooks/src/main.js
```diff
@@ -45,6 +45,7 @@ const load = loader({
       serviceName: 'hooks',
       monitor: monitor.childMonitor('db'),
       statementTimeout: process === 'server' ? 30000 : 0,
+      azureCryptoKey: cfg.azure.cryptoKey,
       dbCryptoKeys: cfg.postgres.dbCryptoKeys,
```
helfi92, line 49: These have `dbCryptoKeys` already and I think the major changelog  mentions adding `dbCryptoKeys`. Do we still need `azureCryptoKey`?

djmitche: The idea was to continue allowing azureCryptoKey in addition to dbCryptoKeys, rather than requiring deployers to change that setting at the same time they deploy the new version. Otherwise, we should remove the ignored configuration parameter.  As it stands, the service will start up with AZURE_CRYPTO_KEY set, but fail to encrypt/decrypt, leaving deployers wondering what they've done wrong (and they haven't done anything wrong).

helfi92: > As it stands, the service will start up with AZURE_CRYPTO_KEY set, but fail to encrypt/decrypt

Why is that?

djmitche: As it stands (meaning: without this patch), the value of AZURE_CRYPTO_KEY is never used.

helfi92: Ah, ok. Gotcha. Thanks!

djmitche: If and when we do our deployment-config-cleanup, we can remove the AZURE_CRYPTO_KEY config from all services at once.

PR #3302, opened 2020-07-29
Include a full DB schema dump in the generated code.

### LINKED ISSUE #3289: DB structure dump in `yarn generate`
Let's include a full DB structure dump in SQL format in the source, as output of `yarn generate`.

This requires a DB for `yarn generate`, but at this point we also require a DB for all `yarn test`, so that's OK.

This probably should not contain functions, but let's spruce up the `db/fns.md` in the process.

djmitche:
@helfi92 and I paired on this, with him navigating. @imbstack we'd like your feedback!

We looked at removing `db/tables.sql`, but @helfi92 it is part of the tests: when upgraded to the latest version, the tables and columns in the DB are compared to `tables.yml` and any discrepancies flagged as errors.  We think that's a good thing to keep. A possible followup would be to generate `tables.yml` automatically in `yarn generate`, but that is a little bit hard right now since the upgrade process assumes that `tables.yml` exists and is correct.

Github Bug/Issue: Fixes #3289.


djmitche:
Oh, potential issues with this:
 - running `yarn generate` now requires a configured `TEST_DB_URL`
 - running `yarn generate` now requires that `pg_dump` be installed (which, if just using a Docker image for the DB, it might not be)

djmitche:
OK, so it looks like we need to build a docker image with both postgres and go installed.

infrastructure/docker-images/build-ci-image.sh
```diff
@@ -40,11 +47,17 @@ echo 'host all all ::1/128 trust' >> /etc/postgresql/$pg_version/main/pg_hba.con
...
+ENV PATH \$GOPATH/bin:/usr/local/go/bin:\$PATH
+RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b \$GOPATH/bin v1.27.0
 COPY pginstall /pginstall
 RUN bash /pginstall && rm /pginstall
 ENV TEST_DB_URL=postgresql://postgres@localhost/postgres
 EOF
 
-docker build -t "taskcluster/node-and-postgres:node${node_version}-pg${pg_version}" ${tmpdir}
-[ -n "$DOCKER_PUSH" ] && docker push "taskcluster/node-and-postgres:node${node_version}-pg${pg_version}"
+docker build -t "taskcluster/ci-image:node${node_version}-pg${pg_version}-go${go_version}" ${tmpdir}
```
imbstack, line 62: ah, this new image makes a lot of sense

djmitche:
Note that the pending DB changes will need a description on rebase.

PR #3306, opened 2020-07-30
[generic-worker] publicIP vs. publicIp

djmitche:
I noticed this while working on #3169 (and it's based on the commit from that issue because it introduces tests for this functionality).

I verified that generic-worker will start without `publicIP` and that it will accept and ignore `publicIp`, so basically we've been not setting this parameter for a long time.  I think that hasn't mattered because we're using websocktunnel for livelog.


djmitche:
(This might have been what @petemoore was exploring in 37b8689921d5c78b4bc2f6294f3ff102b2ff0dae)

changelog/K2p0pAFmQAGlJbiiPTk-Ig.md
```diff
@@ -0,0 +1,4 @@
+audience: worker-deployers
+level: patch
+---
+Worker-runner now correctly sets the `publicIP` configuration for generic-worker (previously it set `publicIP`, which is ignored).
```
petemoore, line 4:
```suggestion
Worker-runner now correctly sets the `publicIP` configuration for generic-worker (previously it set `publicIp`, which is ignored).
```

djmitche: I swear I double-checked that!

PR #3312, opened 2020-07-30
Use livelog and tc-proxy images with versions matching the worker

### LINKED ISSUE #3308: docker-worker pulls old version of livelog, taskcluster-proxy
monopacker saves the most recent version of livelog on the AMI, but docker-worker just uses hard-coded versions:

https://github.com/taskcluster/taskcluster/blob/a53f88c1a5872eec3180f35abb969dbc02dca248/workers/docker-worker/config.yml#L64-L65

We should find a way to use the appropriate version -- either passing that through from monopacker, or always using the same version as docker-worker's version.

This is probably causing slow startup, as those images need to be pulled.  They are tiny, but still..

djmitche:
Note that previously we had been using very old images -- livelog 4 is
from January 2017, and taskcluster-proxy 5.1.0 is from January 2019.
So, this may cause issues with newly-built workers.

This also drops a few bits of config that are no longer used --
relengapi proxy and balrog proxy are long gone.

Github Bug/Issue: Fixes #3308


djmitche:
(i pushed the v35.0.0 tags of the two images, having forgotten they used to be 35.0.0 (without the v)

PR #3316, opened 2020-07-31
generate READMEs before reading them to make TOCs

djmitche:
Fixes an intermittent failure of generation


djmitche:
Ah, that's because generate was wrong on master -- I pushed a fix to master already.  I just checked a merge of this PR and generate passes on the merge.

Issue #2992, opened 2020-06-05
private artifact links in ui now go nowhere


imbstack:
**Describe the bug**
The link for private artifacts appears to go right back to the ui home.

**To Reproduce**

1. Go to https://community-tc.services.mozilla.com/tasks/a3TRMvMERZuRwCGHb9dpcA/runs/0#artifacts
2. Click on any of the private artifacts

**Expected behavior**
I _think_ I should be linked to the artifact? If my understanding is wrong please feel free to close. The [runs in the api](https://community-tc.services.mozilla.com/api/queue/v1/task/a3TRMvMERZuRwCGHb9dpcA/runs/0/artifacts) appear correct to me.

**Taskcluster version**
30.0.2

**Platform (please complete the following information):**

 - OS: Linux
 - Browser if applicable: Firefox
 - Version if applicable: 79.0a1 (2020-06-05) (64-bit)

helfi92:
There's two underlying issues here:

1. The UI should be better at showing that login is required when clicking on this link.
2. The UI is probably failing to send the authorization header when accessing the artifact URL.

helfi92:
OK, I think I may have figured out the issue. Looking into a possible fix now.

helfi92:
The fix for this is in v36.0.1.

PR #3317, opened 2020-07-31
Private artifact links in UI

helfi92:
`public/logs/live_backing.log` is reachable via `https://community.taskcluster-artifacts.net/a3TRMvMERZuRwCGHb9dpcA/0/public/logs/live_backing.log` but in order to access a private log, it looks like the artifact name is encoded rather than nested (e.g., `../debug-logs%2Flivelog-docker-push.log?bewit=...` rather than `.../debug-logs/livelog-docker-push.log?bewit=...`)?

The log viewer grabs the URL to render from the UI which means the URL is encoded (e.g., `https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2Fa3TRMvMERZuRwCGHb9dpcA%2Fruns%2F0%2Fartifacts%2Fdebug-logs%2Ftaskcluster-proxy-docker-build.log%3Fbewit%3D...`). Upon decoding it, `%2F` becomes `/` so then it will try to access the artifact `debug-logs/taskcluster-proxy-docker-build.log` which doesn't  exist.

It looks like when artifacts are stored, some are encoded (e.g., `debug-logs%2Flivelog-docker-push.log`) and some aren't (e.g., `public/logs/live_backing.log
`). The current PR is a hack to fix this but probably not ideal.

Refers to https://github.com/taskcluster/taskcluster/issues/2992.

---

If this is the right direction we want to take, should this be a major bump? Should we bother adding redirects from the old URL to the new one?

### DIFF
--- a/changelog/issue-2992.md
+++ b/changelog/issue-2992.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 2992
+---
+Private artifacts are now accessable via the UI.
--- a/ui/src/components/TaskRunsCard/index.jsx
+++ b/ui/src/components/TaskRunsCard/index.jsx
@@ -172,7 +172,16 @@ export default class TaskRunsCard extends Component {
     const { taskId, runId } = this.getCurrentRun();
 
     if (isLog) {
-      const encoded = encodeURIComponent(url);
+      // react-router decodes its params (via `decodeURI`)
+      // so something like '%252F' and '%2F' in the URL both become '%2F'
+      // we should be able to remove `encodeURI` once
+      // https://github.com/ReactTraining/history/issues/745 has been resolved.
+      //
+      // Context: In order to create the URIs for private artifacts we use
+      // `queue.externalBuildSignedUrl`. Internally, that method uses
+      // `decodeURIComponent` so when an artifact name has a slash in it
+      // the URI will have '`%2F'.
+      const encoded = encodeURI(encodeURIComponent(url));
 
       return this.isLiveLog()
         ? `/tasks/${taskId}/runs/${runId}/logs/live/${encoded}`


djmitche:
The URLs handled by the API and UI should be API method URLs, so `https://community.taskcluster-artifacts.net` should never appear.  The queue.getArtifact method will *redirect* to a URL like that.

The API method will accept either `%2F` or `/` in artfact names.  It's the same API method for public and private artifacts.  So

```
  https://rooturl/api/queue/v1/task/a3TRMvMERZuRwCGHb9dpcA/runs/0/artifacts/some/private/thing
  https://rooturl/api/queue/v1/task/a3TRMvMERZuRwCGHb9dpcA/runs/0/artifacts/some%2Fprivate%2Fthing
```

are both treated equivalently.

I don't understand from the issue description what is generating the encoded or not-encoded URLs.  Is that something in web-server that is providing these artifact URLs?

djmitche:
Ah:
https://github.com/taskcluster/taskcluster/blob/df72c8b67797325c423514cb14bf1a57e187e282/services/web-server/src/loaders/artifacts.js#L25-L55

So for public artifacts, that code uses `encodeURI` (which does not encode `/`), while for private artifacts it uses `externalBuildUrl` which uses `encodeURIComponent`:
https://github.com/taskcluster/taskcluster/blob/df72c8b67797325c423514cb14bf1a57e187e282/clients/client/src/client.js#L566

So it would return
* public -- artifact.url == `https://community-tc.services.mozilla.com/api/queue/v1/task/a3TRMvMERZuRwCGHb9dpcA/runs/0/artifacts/public/logs/live.log`
* private -- artifact.url == `https://community-tc.services.mozilla.com/api/queue/v1/task/a3TRMvMERZuRwCGHb9dpcA/runs/0/artifacts/debug-logs%2Fdocker-push.log?bewit=Z2l0aHV...`

and, critically, the bewit signs those URLs exactly as they are.  Looking in the graphql response confirms those URLs.

Looking at the task in #2992, I see that the UI is generating links to itself of the form:

* public - https://community-tc.services.mozilla.com/tasks/a3TRMvMERZuRwCGHb9dpcA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2Fa3TRMvMERZuRwCGHb9dpcA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log
* private - https://community-tc.services.mozilla.com/tasks/a3TRMvMERZuRwCGHb9dpcA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2Fa3TRMvMERZuRwCGHb9dpcA%2Fruns%2F0%2Fartifacts%2Fdebug-logs%2Fdocker-push.log%3Fbewit%3DZ2l0aHV...

the parameters of which decode to

* https://community-tc.services.mozilla.com/api/queue/v1/task/a3TRMvMERZuRwCGHb9dpcA/runs/0/artifacts/public/logs/live.log
* https://community-tc.services.mozilla.com/api/queue/v1/task/a3TRMvMERZuRwCGHb9dpcA/runs/0/artifacts/debug-logs/docker-push.log?bewit=Z2l0aHV...

the second of which has over-decoded the %2F. Or perhaps more accurately the issue is that the UI did not correctly encode `...acts/debug-logs%2Fdocker-push.log?be...` to `..acts/debug-logs%252Fdocker-push.log?be...`.

https://github.com/taskcluster/taskcluster/blob/df72c8b67797325c423514cb14bf1a57e187e282/ui/src/components/TaskRunsCard/index.jsx#L175-L179
given the confirmed URL from web-server for a private artifact in the "So it would return.." above, 

```js
> encodeURIComponent("https://community-tc.services.mozilla.com/api/queue/v1/task/a3TRMvMERZuRw
CGHb9dpcA/runs/0/artifacts/debug-logs%2Fdocker-push.log?bewit=Z2l0aHV")
'https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2Fa3TRMvMERZuRwCGHb9dpcA%2Fruns%2F0%2Fartifacts%2Fdebug-logs%252Fdocker-push.log%3Fbewit%3DZ2l0aHV'
```
yet that's not the link we see.  What's missing?  Does `<Link>` do some kind of encoding/decoding?

If we were to refactor this, I think it would be to pass only the artifact name in the URL, rather than an entire sub-URL, and to generate the links using `buildUrl` / `buildSignedUrl` in the UI, rather than in the web-server process.  But hopefully that is not necessary!

helfi92:
It turns out react-router does decode its params so we need to wrap the url. The fix is a one liner :)

djmitche:
I think that generate error is from the upstream version this is based on..

PR #3323, opened 2020-08-01
Avoid generic-worker gzipping zstandard-compressed files

glandium:
This matches docker-worker per https://github.com/taskcluster/taskcluster/blob/067b1e076d804dd4342a6b47d29f943746841710/workers/docker-worker/src/lib/features/artifacts.js#L140


tomprince:
I've incorporated these changes into #3325.

petemoore:
> I've incorporated these changes into #3325.

Thanks Tom (and Mike)! Then I'll close here, as that PR landed.

PR #3326, opened 2020-08-02
[docker-worker] Allow disabling features at the worker level.

tomprince:
Bugzilla Bug: [1623749](https://bugzilla.mozilla.org/show_bug.cgi?id=1623749)

### DIFF
--- a/changelog/disable-features.md
+++ b/changelog/disable-features.md
@@ -1,0 +1,5 @@
+audience: users
+level: minor
+reference: bug 1623749
+---
+Docker-worker now allows features to be disabled in the worker config.
--- a/workers/docker-worker/src/lib/task.js
+++ b/workers/docker-worker/src/lib/task.js
@@ -56,7 +56,7 @@ Convert the feature flags into a state handler.
 @param {Object} task definition.
 @param {Monitor} monitor object implementing record/measure methods
 */
-function buildStateHandlers(task, monitor) {
+function buildStateHandlers(runtime, task, monitor) {
   let handlers = [];
   let featureFlags = task.payload.features || {};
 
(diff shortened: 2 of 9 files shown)


workers/docker-worker/src/lib/task.js
```diff
@@ -68,9 +68,15 @@ function buildStateHandlers(task, monitor) {
     throw new Error(`${diff.join()} ${diff.length > 1 ? 'are' : 'is'} not part of valid features`);
   }
 
+  diff = _.keys(featureFlags).filter(x => !runtime.features[x].enabled);
+
+  if (diff.length) {
+    throw new Error(`${diff.join()} ${diff.length > 1 ? 'are' : 'is'} not enabled on this worker`);
+  }
```
tomprince, line 75: I think the failure mode here isn't great (though I haven't tested it), as live log is not enabled at this point, and won't be if this check fails. I'm not sure of a good way around that without adding special handling for livelog, which I suspect is tricky.

tomprince:
The test failure appears related to #3312. (see [here](https://community-tc.services.mozilla.com/tasks/ZN6a3vLeTha9gi8B4YnEIw/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FZN6a3vLeTha9gi8B4YnEIw%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L22216))

changelog/disable-features.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: minor
+reference: bug 1623749
+---
+Allow disabling docker-worker features at the in the worker config.
```
djmitche, line 5:
```suggestion
Docker-worker now allows features to be disabled in the worker config.
```

workers/docker-worker/src/lib/task.js
djmitche: Sounds like a decent followup to file.  One idea might be to record this somewhere and report it as a warningwhen beginning the livelog.

tomprince: The issues is that livelog is itself a feature, and if this check fails, it won't be started. I suspect the fix is to special case livelog, so that it gets started even if this check fails.

djmitche: Ah, I remember now that "live log" in this case means "any logging at all", rather than just "the ability to see logs while the task is executing."  That's definitely a bummer.

Another option might be to throw an invalid-payload error here.

Anyway, I still think it's fine to leave this as a followup.  If someone does create a task that would have failed with a log, then at worst we dig through papertrail and see this message.

tomprince: #3341

PR #3331, opened 2020-08-03
Handle db function deprecation better

### LINKED ISSUE #3328: Support deprecating db methods and removing after a major revision
We recently introduced a little twist to the compatibility matrix: service upgrades must jump by at most one major version at a time.  So going from v50.1.2 to v52.0.5 is not guaranteed not to break.

That means that if a method is deprecated in v50.1.2, then the method needs to remain functional throughout the v51.x.y series, as a deployment might upgrade from v50.1.2 to v51.x.y an expect things to work.  But that function can be removed completely in v52.

So basically, such functions need to wait for two major versions before they can be removed.  Let's find a way to ensure that, either with a test or in a way that's obvious in reviews.  That might be as simple as changing

```yaml
deprecated: true
```
to
```yaml
deprecated: until v52
```
where v52 is two major versions ahead of the current version when this change lands.

djmitche:
Github Bug/Issue: Fixes #3328 

I don't love this solution, but anything more "automatic" would miss some corner cases (like where we dropped support for github logins) and make for weird interactions between "yarn generate" and "yarn release".


db/README.md
```diff
@@ -33,6 +33,8 @@ The script runs in a single transaction.
 The following checklist summarizes what needs to be written to modify the database.
 
 * [ ] new version file in `db/versions` that updates all impacted stored functions
+  * All DB functions must continue to function for two major Taskcluster versions after they are used.  See [db/fns.md](db/fns.md) to figure out which existing functions you must re-implement.
+  * For any DB functions that are being deprecated in this version, add `deprecated: until vNN` in the re-implementation of that function, giving a major version two greater than the current version.
```
imbstack, line 37: `deprecated until` sounds like it will become undeprecated at that version. maybe there's a better word here

djmitche: Yeah, I don't like it either.  What about
```yaml
deprecated: v33
```
and that translates into "(support guaranteed until v35)" by adding 2?

milescrabill: I like `deprecated by:  vXX` or `deprecated as of: vXX`

djmitche: Thanks! I think I'll go with
```yaml
deprecated-by: vXX
```

djmitche:
OK, thinking about this a bit more, it's super-confusing what "this" version is when a change lands.  But `yarn generate` knows!  So I think I'll just switch back to `deprecated: true` which is very easy to write, and have `yarn generate` generate the appropriate output in `db/fns.md`.  And I'll generate that file on `yarn release` too, just like we do for `db/versions/README.md`.  That might be by just running `yarn generate` as part of `yarn release`.

djmitche:
OK, well, that took more time than I expected.  But I am happy with the result, as it's exactly what we need in order to determine how long to support functions.

The inline test suite is a little funky but hopefully OK for development tooling.  I did test this with `yarn release --no-push` and observed it doing the things I wanted it to do.

db/versions/0033.yml
```diff
@@ -113,6 +117,7 @@ methods:
         end if;
       end
   queues_entities_scan:
+    deprecated: true
```
helfi92, line 120: I was expecting to see `deprecated-by` rather than `deprecated` per https://github.com/taskcluster/taskcluster/pull/3331#discussion_r465276574. Was a commit not pushed for this?

db/README.md
helfi92, line 37: We no longer use `until`.

db/versions/0033.yml
djmitche: I changed the whole approach, so the DB version file no longer has a TC version in it.

db/README.md
djmitche: Oops, I did miss changing that back.

djmitche: (fixed now)

djmitche:
oh, tests still failing .. plz hold

djmitche:
OK, the testing kind of forced me to go all the way here, so this removes tc-lib-entities support as well.

There was some dicsussion in slack about this, and the commit message on the third-to-last commit records the summary for posterity.

ui/docs/reference/workers/generic-worker/README.mdx
```diff
@@ -1,31 +0,0 @@
-# Generic Worker
```
imbstack, line 1: why remove this?

djmitche: Huh, that definitely wasn't intentional.  I wonder if I ctrl-C'd a `yarn generate` run while it was writing that file?  Anyway, I'll restore that file and land.

djmitche:
I'll also fix
```
Error: Depenencies for prod that should be dev! Move ["change-case","json-stable-stringify"] from dependencies to devDependencies in package.json (while executing task Dependencies are used and installed)
    at main (/taskcluster/infrastructure/tooling/src/meta/checks/depcheck.js:124:13)
```

PR #3349, opened 2020-08-05
[generic-worker] Don't require mounted task artifacts to be present in dependencies.

tomprince:
This is annoying when using the edit-task functionality in the UI, as it drops
dependencies when you edit a task, and generic worker forces you to re-add
them.

### DIFF
--- a/changelog/no-mount-depedencies.md
+++ b/changelog/no-mount-depedencies.md
@@ -1,0 +1,5 @@
+audience: users
+level: minor
+---
+Generic worker no longer requires tasks from which artifacts are mounted to be
+dependencies of the running task.
--- a/workers/generic-worker/mounts.go
+++ b/workers/generic-worker/mounts.go
@@ -313,16 +313,6 @@ func (taskMount *TaskMount) initReferencedTaskIDs() {
 			}
 		}
 	}
-	taskDependencies := map[string]bool{}
-	for _, taskID := range taskMount.task.Definition.Dependencies {
-		taskDependencies[taskID] = true
-	}
-	for taskID := range taskMount.referencedTaskIDs {
-		if !taskDependencies[taskID] {
-			taskMount.payloadError = fmt.Errorf("[mounts] task.dependencies needs to include %v since one or more of its artifacts are mounted", taskID)
-			return
-		}
-	}
 }
 
 // Here the order is important. We want to delete file caches before we delete
--- a/workers/generic-worker/mounts_test.go
+++ b/workers/generic-worker/mounts_test.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/taskcluster/slugid-go/slugid"
 	"github.com/taskcluster/taskcluster/v36/workers/generic-worker/gwconfig"
 )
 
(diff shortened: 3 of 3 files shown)


petemoore:
Sorry Tom, the correct fix is for the task editor _not_ to drop the dependencies.

tomprince:
>  the correct fix is for the task editor not to drop the dependencies.

I disagree. Tasks can create artifacts before they are complete, so there is no reason this feature could not be used with a task that is not yet finished. I don't have a particular use case for that, but there is no reason for generic worker to prevent that.

petemoore:
The problem with this, is that it introduces a race condition, since if you do not wait for the task to complete, you cannot be sure the artifact is ready. The problem with generic-worker not enforcing the requirement, is that it can lead to intermittent failures if the task is ready _most of the time_. If a task is to create an artifact that another task depends on, and then wishes to do other work, it probably makes sense for that task to be split into smaller tasks. Like you say, it is difficult to construct a reasonable use case where you would intentionally wish to depend on an artifact from a task that is running concurrently to the current task, is not resolved, but you are sure the artifact is available. For this reason, I prefer to keep the robust design that protects against intermittent race failures. I don't see any disadvantage of the task inspector not removing task dependencies when a task is edited, and think it is much more intuitive for the user, that when they edit a task, it is not modified by the front end in unexpected ways. And this also solves your use case, I believe.

So I still favour fixing the task editor not to remove task dependencies.

tomprince:
I understand the desire to have generic-worker check that the arguments are reasonable. But I think that it causes more friction than it is worth.

I think there are two main cases to consider:
- In automation:
  Automation either tends to get changed infrequently, or is using some library like taskgraph that can take care of ensuring that appropriate dependencies are set. Having generic-worker check dependencies against mounts only helps in the case that things are hand-crafted, which I think is uncommon enough to not be worth worrying about.
- Interactive use:
  When someone is working iteratively on a task, the likely tasks that somebody is going to put in to mounts are likely to be already complete (either because they are editing a task that has already run, or they are changing a task to have a mount point at a new task, that they've already seen complete. In this case, requiring the dependencies be specified is busy work (and the failure case is the task fails immediately, and the user can just edit and re-run again).

petemoore:
Tom, let's meet up to discuss.

I think both the use cases you cite are handled by the fix I proposed (for the task inspector not to drop dependencies). I think the only reason this ever became an issue was because of the unexpected behaviour of the task inspector, and even independently from this issue, it dropping task dependencies is quite arguably a poor design choice, so this is something that should be fixed in any case. When that is fixed, this issue disappears, and we avoid the potential intermittent failures that could arise from the race condition that would exist if we relaxed generic-worker. Another thing to consider is that generic-worker/taskcluster serve more projects than just firefox-ci that is using taskgraph. There are plenty of projects that don't use taskgraph, and this generic-worker feature will help people correct their task scheduling machinery.

But in order to avoid deadlock, let's meet to discuss.

djmitche:
Closing this as stale, and not likely to hear more from Tom.  "unexpected behaviour of the task inspector" is addressed in https://github.com/taskcluster/taskcluster/issues/3497.

PR #3350, opened 2020-08-05
issue 3168: support scaling ratio in worker-manager

### LINKED ISSUE #3168: Support scalingRatio for workerpools
This bug originally read:
> All worker types will inevitably spin up more workers than are needed, I refer to that as overprovisioning.  gecko-t/t-win7-32 is overprovisioned about 4x more than other windows test workers.  I would like to adjust our metrics/heuristics which determine when we create new workers and how many we create.  I am fine with slightly longer queues.

I'm going to take this as a starting point for a bug to support scalingRatio again as a simple tool to help admins dial in their efficiency-to-cost ratios.

Scaling ratio is a number in the range `(0, 1]` where `1.0` would mean "spawn 1 worker per pending task" and `0.5` would mean "spawn 1 worker for ever 2 pending tasks".

I'm not sure this is exactly the formulation we'll use (or if we'll use scaling ratio for sure) by the time we get to the sprint but I'm writing it this way for now.

milescrabill:
Issue: Fixes #3168

### DIFF
--- a/changelog/issue-3168.md
+++ b/changelog/issue-3168.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: minor
+reference: issue 3168
+---
+The worker-manager now supports a `scalingRatio` that determines how many workers to spawn per pending task.
--- a/changelog/issue-3168.md
+++ b/changelog/issue-3168.md
@@ -2,4 +2,6 @@ audience: worker-deployers
 level: minor
 reference: issue 3168
 ---
-The worker-manager now supports a `scalingRatio` that determines how many workers to spawn per pending task.
+The worker-manager now supports a `scalingRatio` that determines how much worker capacity to spawn per pending task.
+The `scalingRatio` is a ratio of worker capacity to pending tasks - a ratio of 1.0 means that 1 capacity will be added for each pending task.
+
--- a/services/worker-manager/schemas/v1/config-aws.yml
+++ b/services/worker-manager/schemas/v1/config-aws.yml
@@ -75,6 +75,13 @@ properties:
     title: Maximum Capacity
     type: integer
     description: Maximum capacity that should not be exceeded at any given time for the entire workerpool
+  scalingRatio:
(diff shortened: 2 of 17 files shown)


services/worker-manager/src/estimator.js
```diff
@@ -6,23 +6,27 @@ class Estimator {
     this.monitor = monitor;
   }
 
-  async simple({workerPoolId, minCapacity, maxCapacity, workerInfo}) {
+  async simple({workerPoolId, minCapacity, maxCapacity, scalingRatio, workerInfo}) {
```
imbstack, line 9: maybe we just default this to `1.0` here?

milescrabill: oops, I meant to add that default!

changelog/issue-3168.md
```diff
@@ -0,0 +1,5 @@
+audience: worker-deployers
+level: minor
+reference: issue 3168
+---
+The worker-manager now supports a `scalingRatio` that determines how many workers to spawn per pending task.
```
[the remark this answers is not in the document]
djmitche: You can also link to the schemas from the changelog via docs.taskcluster.net.

Unfortunately, this detail is, if I remember correctly, hidden in a popover in the schema viewer, so it's going to be a bit hard to find :(

services/worker-manager/schemas/v1/config-aws.yml
```diff
@@ -75,6 +75,13 @@ properties:
     title: Maximum Capacity
     type: integer
     description: Maximum capacity that should not be exceeded at any given time for the entire workerpool
+  scalingRatio:
+    title: Scaling Ratio
+    type: number
+    minimum: 0.0
+    maximum: 1.0
+    description: |
+      The scaling ratio is a ratio of worker capacity to pending tasks - a ratio of 1.0 means that 1 capacity will be added for each pending task
```
djmitche, line 84: This is a really concise description :+1:

PR #3351, opened 2020-08-05
[DO-NOT-MERGE] [generic-worker] Add support for mounting `.tar.zstd` files.

tomprince:
The new version of the package we use is vulnerable to CVE-2019-10743.


petemoore:
Note, we pinned to archiver 2.1 due to mholt/archiver#213.

djmitche:
@petemoore what should we do with this?

petemoore:
Let's test the latest archiver to see if it still exhibits the same issue (there is a test case in the linked issue). If it is still present, I guess we should either make a fix, find an alternative implementation, or remove support for `.zstd` files if we decide our client is genuinely vulnerable (maybe we are lucky and it is vulnerable for some edge case which cannot be exploited in our setup).

petemoore:
Ah looking closer at https://nvd.nist.gov/vuln/detail/CVE-2019-10743 I see this isn't related to `zstd` files, but to all the archive formats, so we should look at this independently of whether we add zstd support or not (looks like we don't currently have it, but this PR was intending to add it).

djmitche:
@petemoore is this close enough that it could be fixed up and merged to add zstd support, separately from the sec issue with archiver?  Or, should we close it as a won't-fix?

petemoore:
> @petemoore is this close enough that it could be fixed up and merged to add zstd support, separately from the sec issue with archiver? Or, should we close it as a won't-fix?

zstd support was added in archiver 3.2.0 in https://github.com/mholt/archiver/commit/f26e4d80350125b0521aece992718733ed2b532d but v3 of archiver is broken for us due to https://github.com/mholt/archiver/commit/f26e4d80350125b0521aece992718733ed2b532d so we pinned to version 2. So unfortunately, there isn't a version of archiver with zstd support that isn't broken for us.

djmitche:
Cool, I moved it to #3540

PR #3352, opened 2020-08-06
issue 3033: AWS provider updates expires timestamp, move updating expires to worker-scanner

### LINKED ISSUE #3033: aws provider should bump `worker.expires`
The Azure and Google providers both update a worker's `expires` timestamp when it is set to expire in less than a day.  The AWS provider does not.

(in fact, probably this should happen in `worker-scanner.js` for all workers, so that an active worker never expires)

milescrabill:
Github Issue: Fixes #3033

Note: first commit is just doing this for AWS provider, second commit is doing it for all providers in worker-scanner

I was wondering whether there is a reason the `static` provider doesn't update `expires` as well - is there a reason it shouldn't?

### DIFF
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -323,6 +323,17 @@ class AwsProvider extends Provider {
         switch (is.InstanceState.Name) {
           case 'pending':
           case 'running':
+            // If the worker will be expired soon but it still exists,
+            // update it to stick around a while longer. If this doesn't happen,
+            // long-lived instances become orphaned from the provider. We don't update
+            // this on every loop just to avoid the extra work when not needed
+            if (worker.expires < taskcluster.fromNow('1 day')) {
+              await worker.update(this.db, worker => {
+                worker.expires = taskcluster.fromNow('1 week');
+              });
+            }
+            this.seen[worker.workerPoolId] += worker.capacity || 1;
+            break;
           case 'shutting-down': //so that we don't turn on new instances until they're entirely gone
           case 'stopping':
             this.seen[worker.workerPoolId] += worker.capacity || 1;
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -323,17 +323,6 @@ class AwsProvider extends Provider {
         switch (is.InstanceState.Name) {
           case 'pending':
           case 'running':
-            // If the worker will be expired soon but it still exists,
-            // update it to stick around a while longer. If this doesn't happen,
-            // long-lived instances become orphaned from the provider. We don't update
-            // this on every loop just to avoid the extra work when not needed
-            if (worker.expires < taskcluster.fromNow('1 day')) {
-              await worker.update(this.db, worker => {
-                worker.expires = taskcluster.fromNow('1 week');
-              });
-            }
-            this.seen[worker.workerPoolId] += worker.capacity || 1;
-            break;
           case 'shutting-down': //so that we don't turn on new instances until they're entirely gone
           case 'stopping':
             this.seen[worker.workerPoolId] += worker.capacity || 1;
--- a/services/worker-manager/src/worker-scanner.js
+++ b/services/worker-manager/src/worker-scanner.js
@@ -1,4 +1,5 @@
 const Iterate = require('taskcluster-lib-iterate');
+const taskcluster = require('taskcluster-client');
 const {paginatedIterator} = require('taskcluster-lib-postgres');
 const { Worker } = require('./data');
 
@@ -62,6 +63,16 @@ class WorkerScanner {
         } else {
           this.monitor.info(
             `Worker ${worker.workerGroup}/${worker.workerId} has unknown providerId ${worker.providerId} (ignoring)`);
+        }
+
+        // If the worker will be expired soon but it still exists,
+        // update it to stick around a while longer. If this doesn't happen,
+        // long-lived instances become orphaned from the provider. We don't update
+        // this on every loop just to avoid the extra work when not needed
+        if (worker.expires < taskcluster.fromNow('1 day')) {
+          await worker.update(this.db, worker => {
+            worker.expires = taskcluster.fromNow('1 week');
+          });
         }
       }
     }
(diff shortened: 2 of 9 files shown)


services/worker-manager/test/provider_aws_test.js
```diff
@@ -497,6 +497,24 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
       assert.strictEqual(provider.seen[worker.workerPoolId], 0);
     });
 
+    test('update expiration for a long-running worker', async function() {
+      const expires = taskcluster.fromNow('-1 week');
+      fake.rgn('us-west-2').instanceStatuses['i-123'] = 'terminated';
```
petemoore: What is the reason to define the instance as terminated here?

services/worker-manager/test/worker_scanner_test.js
```diff
@@ -78,6 +81,8 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
         workerId: 'testing-123',
       });
       assert(worker.providerData.checked);
+      // expires wasn't updated
```
petemoore, line 84: maybe worth a comment to explain expires should only be updated if less than a day, and expires2 is not in 3 days

services/worker-manager/test/worker_scanner_test.js
```diff
@@ -117,16 +122,18 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
         workerId: "testing-123",
       });
       assert(worker1.providerData.checked);
+      assert(worker1.expires > expires);
```
petemoore, line 125: here a comment could be useful that the expiry should be updated since expiry was less than a day

services/worker-manager/test/worker_scanner_test.js
```diff
@@ -162,12 +169,14 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
         workerId: 'testing-123',
       });
       assert(worker1.providerData.checked);
+      assert(worker1.expires > expires);
```
petemoore, line 172: here too

services/worker-manager/test/provider_aws_test.js
helfi92: I had the same question :)

services/worker-manager/src/worker-scanner.js
```diff
@@ -63,6 +64,16 @@ class WorkerScanner {
...
+
+        // If the worker will be expired soon but it still exists,
+        // update it to stick around a while longer. If this doesn't happen,
+        // long-lived instances become orphaned from the provider. We don't update
+        // this on every loop just to avoid the extra work when not needed
+        if (worker.expires < taskcluster.fromNow('1 day')) {
+          await worker.update(this.db, worker => {
+            worker.expires = taskcluster.fromNow('1 week');
+          });
+        }
```
djmitche, line 76: Now that I look at this logic, it's not doing what we want -- we'd like to have every active worker have an expiration at least 1 week in the future.  This logic lets expirations range from 1 to 7 days in the future.  I think we'd want them to range from 7 to 8 days in the future, instead (thus updating them once a day on average).

That should be a quick change to this logic, and updates to the tests -- do you want to just make that adjustment here, or file a followup?

services/worker-manager/src/providers/aws.js
```diff
@@ -323,6 +323,17 @@ class AwsProvider extends Provider {
...
+            // If the worker will be expired soon but it still exists,
+            // update it to stick around a while longer. If this doesn't happen,
+            // long-lived instances become orphaned from the provider. We don't update
+            // this on every loop just to avoid the extra work when not needed
+            if (worker.expires < taskcluster.fromNow('1 day')) {
+              await worker.update(this.db, worker => {
+                worker.expires = taskcluster.fromNow('1 week');
+              });
+            }
+            this.seen[worker.workerPoolId] += worker.capacity || 1;
```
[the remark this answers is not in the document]
milescrabill: that line is there because of the `break` statement before the next case, prior to adding this logic for `pending` and `running` instances there was a fallthrough that includes that line.

Also, that commit is just there to show what it would look like to just do the AWS part of this and won't be landed.

services/worker-manager/src/worker-scanner.js
milescrabill: I updated that logic and the tests.

services/worker-manager/src/worker-scanner.js
```diff
@@ -63,6 +64,16 @@ class WorkerScanner {
...
             `Worker ${worker.workerGroup}/${worker.workerId} has unknown providerId ${worker.providerId} (ignoring)`);
         }
+
+        // If the worker will be expired soon but it still exists,
+        // update it to stick around a while longer. If this doesn't happen,
+        // long-lived instances become orphaned from the provider. We don't update
+        // this on every loop just to avoid the extra work when not needed
+        if (worker.expires < taskcluster.fromNow('1 week')) {
+          await worker.update(this.db, worker => {
+            worker.expires = taskcluster.fromNow('1 week');
```
djmitche, line 74: (Somehow I commented on the commit and not the PR)

Change this to `8 days` and I think this is good to go.  The reason is that we only want to occasionally bump the expires value, rather than on every scan -- so we bump it high enough that the next scan (in fact, the next day's worth of scans) won't satisfy the `< 1 week` condition.

services/worker-manager/test/provider_aws_test.js
milescrabill: Ah, this is from the intermediate commit, I think this was a copy paste or commit fail, it should have been made running (and was at some point!?) - the instance status had to be defined.

djmitche:
:shipit:

petemoore:
> Overall I really like it, but I haven't fully understood the rationale of why we set an expiry on running workers, if we always bump it to be in the future. If I understand correctly, that means running workers will never expire, so I'm not sure what the purpose of having the field is.

@djmitche or @milescrabill I think I might have guessed the answer to this - is the `expiry` field for determining how long a worker appears in the list of workers of a given worker pool? In other words, the worker should still be listed as a worker of the worker pool for around a week after it is no longer active? Previously I was thinking this was related to credentials expiration, but probably that was a false assumption.

djmitche:
Yes, that's exactly it.  Sorry for the confusion!

PR #3353, opened 2020-08-06
curl to download binary on mac

### LINKED ISSUE #3178: [client-shell] MacOSX taskcluster cli tool isn't signed/notarized
**Describe the bug**

When downloading from the [client-shell readme](https://github.com/taskcluster/taskcluster/releases/download/v32.0.0/taskcluster-darwin-amd64), running the binary on recent MacOSX versions gets a warning about being unable to verify the developer, and requires the extra steps mentioned in [Apple's support](https://support.apple.com/en-gb/guide/mac-help/mh40616/mac) in order to run.

**To Reproduce**

Download and run the cli tool, without having done so before on that installation

**Expected behavior**

The binary should be trusted.

MacOSX 10.15.*

So the question for me is: What's the balance between this trust set-up and getting the binary trusted? I suspect this is a signing/notarisation step, and it may well be that it's more effort than we need given the scale of the current audience.

anubhabsen:
Github Bug/Issue: Fixes #3178

### DIFF
--- a/changelog/issue-3178.md
+++ b/changelog/issue-3178.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3178
+---
--- a/clients/client-shell/README.md
+++ b/clients/client-shell/README.md
@@ -13,8 +13,15 @@ scheduled runs.
 
 ### Installation
 
-To install, download the `taskcluster` binary for the latest release your
-platform, and run it!  On POSIX platforms you will need to `chmod +x` of
+To install, Linux users should download the `taskcluster` binary for the latest release your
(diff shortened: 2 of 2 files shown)


petemoore:
Above client-nodejs failure looks like an intermittent. The one and only mac workers seems to have stopped taking jobs, I'll investigate. The change looks fine to me, but I'll let @djmitche / @milescrabill perform a formal review.

djmitche:
The nodejs issue is due to using fixed port numbers in the test -- sometimes those conflict with other things, since they are in the range of ports from which the kernel selects randomly.

Issue #3270, opened 2020-07-27
Estimate worker config prices for GCP


imbstack:
Take as input an array of worker configs and give as output a price per unit time of the config.

imbstack:
From @helfi92's work in https://github.com/taskcluster/taskcluster/pull/3357 we have this:

```js
const { google } = require('googleapis');

const compute = google.compute('v1');
const cloudBilling = google.cloudbilling('v1');
const computeEngineServiceId = '6F81-5844-456A';
const regionNames = {
  'asia-east1': 'Asia Pacific (Taiwan)',
  'asia-east2': 'Asia Pacific (Hong Kong)',
  'asia-northeast1': 'Asia Pacific (Tokyo)',
  'asia-south1': 'Asia Pacific (Mumbai)',
  'asia-southeast1': 'Asia Pacific (Singapore)',
  'australia-southeast1': 'Asia Pacific (Sydney)',
  'europe-north1': 'EU (Finland)',
  'europe-west1': 'EU (Belgium)',
  'europe-west2': 'EU (London)',
  'europe-west3': 'EU (Frankfurt)',
  'europe-west4': 'EU (Netherlands)',
  'northamerica-northeast1': 'Canada (Montréal)',
  'southamerica-east1': 'South America (São Paulo)',
  'us-central1': 'US Central (Iowa)',
  'us-east1': 'US East (South Carolina)',
  'us-east4': 'US East (Northern Virginia)',
  'us-west1': 'US West (Oregon)',
  'us-west2': 'US West (Los Angeles)',
};
const unsupportedInstanceTypes = ['n1-ultramem-40', 'n1-ultramem-80', 'n1-megamem-96', 'n1-ultramem-160'];

async function getSkus(request) {
  let skus = [];
  return new Promise((resolve) => {
    const handlePage = function(err, response) {
      if (err) {
        console.error(err);
        return;
      }

      const itemsPage = response.data.skus;
      if (!itemsPage) {
        return;
      }
      skus = skus.concat(itemsPage);

      if (response.data.nextPageToken) {
        request.pageToken = response.data.nextPageToken;
        cloudBilling.services.skus.list(request, handlePage);
      } else {
        return resolve(skus);
      }
    };
    cloudBilling.services.skus.list(request, handlePage);
  });
}

function priceInUsd(pricingInfos) {
  if (pricingInfos.length !== 1) {
    throw new Error('pricing info not parsable');
  }

  const [pricingInfo] = pricingInfos;
[...]

PR #3357, opened 2020-08-06
Estimate worker config prices for GCP

helfi92:
Fixes https://github.com/taskcluster/taskcluster/issues/3270.

Script is based on https://github.com/banzaicloud/cloudinfo/blob/f6371cb4bf3be02a3842be58ee5deaa1799ed292/internal/cloudinfo/providers/google/cloudinfo.go.

https://banzaicloud.com/blog/cloud-provider-cloudinfo/ has very good documentation on how this whole thing works.

### DIFF
--- a/services/worker-manager/src/googleEstimator.js
+++ b/services/worker-manager/src/googleEstimator.js
@@ -1,0 +1,282 @@
+const { google } = require('googleapis');
+
+const compute = google.compute('v1');
+const cloudBilling = google.cloudbilling('v1');
+const computeEngineServiceId = '6F81-5844-456A';
+const regionNames = {
+  'asia-east1': 'Asia Pacific (Taiwan)',
+  'asia-east2': 'Asia Pacific (Hong Kong)',
+  'asia-northeast1': 'Asia Pacific (Tokyo)',
+  'asia-south1': 'Asia Pacific (Mumbai)',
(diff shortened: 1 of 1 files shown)


services/worker-manager/src/googleEstimator.js
```diff
@@ -0,0 +1,282 @@
...
+        return resolve(machineTypes);
+      }
+    };
+    compute.machineTypes.list(request, handlePage);
+  });
+}
+
+async function estimatePrice(request, gcpRegion, gcpInstanceType) {
+  const regions = await getRegions(request);
+  const pricePerRegion = await getPrice(request);
```
helfi92, line 215: Ah, could probably use `Promise.all` here.

imbstack:
I copy-pasted this code to the issue and will use it from there when we merge it all in!

services/worker-manager/src/googleEstimator.js
```diff
@@ -0,0 +1,282 @@
...
+  const zonesInRegions = {};
+  const allPrices = {};
+
+  for (let r of Object.keys(regions)) {
+    const zones = await getZones(request, r);
+    const machineTypes = await getMachineTypes({ ...request, zone: zones[0]});
```
helfi92, line 221: also here, we can use `Promise.all`

PR #3371, opened 2020-08-09
Replace use of hasPrefixedScopes in docker-worker with scope expressions

### LINKED ISSUE #3344: Replace use of hasPrefixedScopes in docker-worker with scope expressions
@tomprince did something similar in #3327, but left `buildVolumeBindings` un-changed.  That's the last use of `hasPrefixedScopes`, so fixing it will allow us to remove this function's definition.

Why?
 * taskcluster-lib-scopes allows generation of better error messages for the user
 * `hasPrefixedScopes` is only used in docker-worker, so not well-tested and well-used, and poorly-tested security-sensitive code is bad
 * removing code generally simplifies a codebase -- less code to read, less code to have bugs

Note that running docker-worker's tests locally can be difficult.  It's OK to make what you think the change should look like and create a draft PR, then watch the test results on that PR.

bhushan-borole:
Fixes #3344

### DIFF
--- a/changelog/issue-3344.md
+++ b/changelog/issue-3344.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3344
+---
--- a/workers/docker-worker/src/lib/task.js
+++ b/workers/docker-worker/src/lib/task.js
@@ -13,7 +13,6 @@ const os = require('os');
 const features = require('./features');
 const getHostname = require('./util/hostname');
 const { fmtLog, fmtErrorLog } = require('./log');
-const { hasPrefixedScopes } = require('./util/scopes');
 const scopes = require('./scopes');
 const { validatePayload } = require('./util/validate_schema');
 const waitForEvent = require('./wait_for_event');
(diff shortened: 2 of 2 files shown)


bhushan-borole:
Hey @djmitche @helfi92.
Please take a look, and help me figure out what's wrong

bhushan-borole:
That command helped me a lot to get the actual source of the error.
But I am still facing problems, how to fix it. I looked into the `hasPrefixedScopes` function, I have implemented it in the same way, I know the `scopeExpression` is invalid, but why that I am wondering

djmitche:
I would address that in two ways:

First, looking at the function raising the error (and finding its definition based on the filename in the exception):
https://github.com/taskcluster/taskcluster/blob/46bbb339f253f0996071c2eabd62410784159302/workers/docker-worker/src/lib/scopes/expressions.js#L114-L152

It looks like this is happening because it's seeing a thing that is neither a string nor `{AnyOf}` nor `{AllOf}`.  So that helps me stare more closely at the source code to find the error (hint: `Allof` instead of `AllOf` :thinking:).

If that hadn't worked, my next step would be to add
```
console.log('BAD SCOPE EXPRESSION:');
console.log(scopeExpression);
```
and push to the PR and then try to find that output in the logfile.  That would help me figure out if what I *think* the value of that variable is matches reality.  But in this case the first technique worked :)

bhushan-borole:
Yes, that was a typo I did.
I changed `Allof` to `AllOf` and all the test cases passed, except one.
The error is `ERROR: Pull Request 3371 does not modify any files in 'changelog/'`

bhushan-borole:
> > The error is ERROR: Pull Request 3371 does not modify any files in 'changelog/'
> 
> That one's easy to fix -- run `yarn changelog --silent --issue 3344` and check in the results :)
> 
> Otherwise, this just needs updates to the documentation and it's ready to merge!

I am getting error module `js-yaml` not installed, I installed it via npm, but it still gives me the same error.

bhushan-borole:
> > The error is ERROR: Pull Request 3371 does not modify any files in 'changelog/'
> 
> That one's easy to fix -- run `yarn changelog --silent --issue 3344` and check in the results :)
> 
> Otherwise, this just needs updates to the documentation and it's ready to merge!

This worked !!
Thank youu

bhushan-borole:
Thank you @djmitche

PR #3414, opened 2020-08-18
Minify yarn requirements

djmitche:
I just set up this `yarn-minify` library.  The idea here will be to run it on all `yarn.locks`, making it easy to update indirect dependencies when possible by just running `yarn generate`.

For the moment, I've ignored all of the minifications.  Once this lands we can start whittling down those ignore lists using CI to make sure we haven't broken anything too badly.


djmitche:
Oh, shoot, I didn't run `yarn generate` after rebasing.  I'll have a PR shortly.

PR #3420, opened 2020-08-19
Babel yarn-minification

djmitche:
I think this is almost all dev-dependencies (except babel-runtime), so should be relatively low risk..


djmitche:
(oh, and the first commit fixes the error from merging #3414 after having landed some other renovate updates)

PR #3431, opened 2020-08-21
Check postgres version in lib-postgres

### LINKED ISSUE #3039: Check Postgres version in lib-postgres
Since we require 11.x, let's verify that on connection in taskcluster-lib-postgres

https://www.postgresql.org/docs/11/runtime-config-preset.html#GUC-SERVER-VERSION-NUM

Refs #3021

bhushan-borole:
Github Bug/Issue: Fixes #3039


petemoore (approved):
Looks ok to me, thanks @bhushan-borole!

@djmitche Can you give this a look too?

My only thought is whether we might want to include the minimum postgres version in some external file that we read in, so we have a config file we can update if we want to change it, rather than needing to change it inline.

libraries/postgres/src/Database.js
```diff
@@ -386,6 +386,18 @@ class Database {
...
 
+  async _checkVersion({db, schema}) {
+    await db._withClient('admin', async client => {
+      const version = await client.query(`
+        SELECT current_setting('server_version_num');
+      `);
+      const ver = version.rows[0].current_setting;
+      if (ver < 110000) {
+        throw new Error("Postgres version is less than 11. Please upgrade to 11.x or above.");
+      }
```
djmitche, line 397: I think we'll want to check for and prohibit >= 120000 as well.  At the moment, we don't test on such a version so we do not want to run it there.

bhushan-borole:
Thank you for reviewing @petemoore @djmitche 
But everything you just said, was a bouncer for me XD.
Can you simplify it for me please?

djmitche:
Please make sure that running Postgres 12 or higher also results in an error.

bhushan-borole:
> Please make sure that running Postgres 12 or higher also results in an error.

Oh I thought prohibit is some kind of variable, I will make the changes
Thanks!

libraries/postgres/src/Database.js
```diff
@@ -386,6 +386,21 @@ class Database {
...
+  async _checkVersion({db, schema}) {
+    await db._withClient('admin', async client => {
+      const version = await client.query(`
+        SELECT current_setting('server_version_num');
+      `);
+      const ver = version.rows[0].current_setting;
+      if (ver < 110000) {
+        throw new Error("Postgres version is less than 11. Please upgrade to 11.x");
+      }
+      if (ver > 120000) {
```
petemoore, line 398:
```suggestion
      if (ver >= 120000) {
```

petemoore (changes requested):

libraries/postgres/src/Database.js
bhushan-borole: Fixed

libraries/postgres/src/Database.js
bhushan-borole: Ahh missed this.
Fixed it

petemoore (commented):
> Makes sense to me so far! Probably the next step is to call this function somewhere. Possibly as part of `_checkDbSettings`?

I think this is the only part left to do still. Thanks @bhushan-borole!

bhushan-borole:
> > Makes sense to me so far! Probably the next step is to call this function somewhere. Possibly as part of `_checkDbSettings`?
> 
> I think this is the only part left to do still. Thanks @bhushan-borole!

I can do this part as well.
Can you please guide me through it?

djmitche:
Sure!  Right now the new function you have added isn't called anywhere.  Have a look through `Database.js` to see where other check functions get called, and add a call to your function.

bhushan-borole:
> Sure! Right now the new function you have added isn't called anywhere. Have a look through `Database.js` to see where other check functions get called, and add a call to your function.

Sure, I will take a look.

bhushan-borole:
@djmitche 
I am thinking it will be called in `upgrade` function

djmitche:
> I am thinking it will be called in `upgrade` function

Sounds good

bhushan-borole:
> > I am thinking it will be called in `upgrade` function
> 
> Sounds good

Below this line, right?
https://github.com/taskcluster/taskcluster/blob/002d3fa751cbae31d595881e71f094d5d6621467/libraries/postgres/src/Database.js#L179

djmitche:
Make the change and let's take a look :)

bhushan-borole:
Why is it saying that function is not defined?

bhushan-borole:
I am sorry for so much commits, I totally forgot to rebase

libraries/postgres/src/Database.js
```diff
@@ -186,7 +186,9 @@ class Database {
         showProgress('...checking permissions');
         await Database._checkPermissions({ db, schema, usernamePrefix });
         showProgress('...checking table columns');
-        await Database._checkTableColumns({ db, schema, usernamePrefix });
+        await Database._checkTableColumns({db, schema, usernamePrefix});
+        showProgress('...checking version');
+        await Databse._checkVersion({db, schema});
```
djmitche: The comment above this section indicates this is only run at the end of the update.  There are some calls to `this._checkXxx` at the top of this function which run on *every* update, before it starts, and that might be a better place to put this.

djmitche: Also, [it looks like the test failed](https://community-tc.services.mozilla.com/tasks/PG2wotIHRCW32vasmU4vIA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FPG2wotIHRCW32vasmU4vIA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L569)

bhushan-borole: The error message shows, function does not exist, but it does

bhushan-borole: I forgot to make it static XD

djmitche:
@petemoore if you're happy with this, please merge :)

petemoore (approved):
Looks awesome, thanks!

bhushan-borole:
Thank you
@petemoore @djmitche

helfi92:
Excellent work @bhushan-borole!

PR #3432, opened 2020-08-21
Handle error from malformed GitHub check artifacts

### LINKED ISSUE #3354: Handle error when GitHub check artifacts are not valid JSON
https://github.com/taskcluster/taskcluster/blob/849fedfd567d56c6a45b48dba03984ab4283c418/services/github/src/handlers.js#L558-L565

This is trying to parse an artifact provided by a task, and on failure calling `monitor.reportError`.  But, that kind of error reporting is intended for "Quick, something's wrong with Taskcluster!" not "Oops, something's wrong with a task", which is the case here.

Instead of calling reportError in this situation, let's report the error to the person looking at the check run in GitHub.  How could we do that?

rohitgeddam:
Add a test for when the json returned is not a valid json

Github Bug/Issue: Fixes #3354


services/github/src/handlers.js
```diff
@@ -561,7 +561,20 @@ async function statusHandler(message) {
           customCheckRunAnnotations = json;
         }
       } catch (e) {
-        await this.monitor.reportError(e);
+        if (e instanceof SyntaxError) {
+          let errorMessage = "returned text is not valid JSON";
```
djmitche, line 565: Let's add a bit more information here, including at least the artifact name and the taskId, so that whoever sees this comment has the tools they need to figure out what's wrong.

rohitgeddam: Does this error message sound correct 
`Returned text is not valid JSON. Artifact Name - ${annotationsArtifactName}, TaskId - ${taskId}`

services/github/src/handlers.js
```diff
@@ -561,7 +561,20 @@ async function statusHandler(message) {
           customCheckRunAnnotations = json;
         }
       } catch (e) {
-        await this.monitor.reportError(e);
+        if (e instanceof SyntaxError) {
```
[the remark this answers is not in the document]
rohitgeddam: thankYou

services/github/src/handlers.js
djmitche: I think "returned" is a little vague there.  How about "Custom annotations artifact ${annotationsArtifactName} on task ${taskId} does not contain valid JSON."

services/github/test/handler_test.js
```diff
@@ -882,6 +882,26 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
+        .onSecondCall()
+        .returns({ status: 200, text: "{{{invalid json!!" });
+      await simulateExchangeMessage({
+        taskGroupId: TASKGROUPID,
+        exchange: 'exchange/taskcluster-queue/v1/task-completed',
+        routingKey: 'route.checks',
+        taskId: CUSTOM_CHECKRUN_TASKID,
+        reasonResolved: 'completed',
+        state: 'completed',
+      });
```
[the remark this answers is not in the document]
rohitgeddam: I included these in the test. also the tests are passing now !

changelog/issue-3354.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
+level: patch
```
djmitche, line 2: I changed this to `patch` since users may want to know when this is fixed.

PR #3449, opened 2020-08-26
add comment about the tests for auth's gcp support

djmitche:
I mixed this up when rotating credentials, so let's write it down somewhere.


petemoore (approved):
Thanks!

PR #3474, opened 2020-09-07
Bug 1563191 - include required json schema in task log if payload non-conformant

petemoore:
Bugzilla Bug: [1563191](https://bugzilla.mozilla.org/show_bug.cgi?id=1563191)

### DIFF
--- a/changelog/bug-1563191.md
+++ b/changelog/bug-1563191.md
@@ -1,0 +1,5 @@
+audience: users
+level: minor
+reference: bug 1563191
+---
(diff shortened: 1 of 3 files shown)


petemoore:
Note, I added the changelog in the second commit, but for some reason the CI passed on the first commit without the changelog - so something seems up there.

djmitche (approved):

djmitche:
Changelog is implemented in https://github.com/taskcluster/taskcluster/blob/main/infrastructure/tooling/src/changelog/index.js

PR #3479, opened 2020-09-09
Upgrade yargs-parser for GHSA-p9pc-299p-vxgp

djmitche:
Note that this is only used for mocha, so not a big risk anyway.


petemoore (approved):
Looks reasonable, thanks!

helfi92 (approved):

Issue #3056, opened 2020-06-15
[azure provider] simplify the logic in checkWorker


djmitche:
* step creation/deletion state machine forward
* change worker.state to RUNNING or STOPPING or STOPPED based on VM state and current worker state
* handle lifetime expirations
* handle extending worker.expires

This is all in there, but inter-mingled.  Let's factor it out into nice functions.

helfi92:
Resolved in https://github.com/taskcluster/taskcluster/commit/1d49719b8d84efd44cf5658c37d92b9e358d1b72.

Issue #3058, opened 2020-06-15
[azure provider] state machine for creating / deleting resources


djmitche:
Move the state machine that handles creation / deletion of Azure resources into its own class with its own state tracking and separate testing.

This lets us ensure this implementation is really bulletproof, while relying on it in the rest of the provider to simply "return a VM or fail trying" and "delete all resources associated with this VM".

djmitche:
The big advantage to this refactor would be more easily testable code and secondarily more readable/modifyable code (say, if we find a need to create different resources later).  The downside is introducing new bugs :)

There are a few "weird" things about this process that make it not a "normal" state machine:

 * state is both in the Azure API and in the `workers` DB table
 * some things can happen in parallel (disks) while others must be serialized (ip -> nic -> vm).
 * there are no "events" that transition from state to state -- we just poll the API

So, I think implementing this as a "normal" state machine with functions for each "state transition" would be difficult and not result in something especially readable or testable.

The current implementation of `provisionResources` reads something like

```
provision a NIC
if not done, return
provision an IP
if not done, return
provision a VM
if not done, return
on error, switch to removing the worker
```

which is a pretty good summary of what's happening.  There are some callbacks and functions to modify the worker's `provisionerData` that don't read too clearly.

Another structure I could think of to handle this would be to set up nested function calls and use an exception to signal that execution is blocked and should be retried on the next poll.  Something like

```
fn provisionResources():
  try:
    createVm()
  on error = Blocked:
    return
  on any other error:
    removeWorker

fn createVm():
  if providerData.vm is complete:
    return vm
  if providerData.vm indicates creation is in progress:
    if Azure API says creation is complete:
      get information on disks and store it in providerData.vm
      return vm
    else:
      raise blocked
  nic = createNic()
  call Azure API to create the VM with 'nic', store the operation in providerData.vm
  raise blocked

fn createNic():
  if providerData.nic is complete:
    return nic
  if providerData.nic indicates creation is in progress:
    if Azure API says creation is complete:
      return nic
    else:
      raise blocked
  ip = createIp()
  call Azure API to create the NIC with that IP, store the operation in providerData.nic
  raise blocked

fn createIp():
  if providerData.ip is complete:
    return ip
  if providerData.ip indicates creation is in progress:
    if Azure API says creation is complete:
      return ip
    else:
      raise blocked
  call Azure API to create the IP, store the operation in providerData.ip
  raise blocked
```

[...]

PR #3480, opened 2020-09-09
Simplify the logic in checkWorker

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/3056.

>     * step creation/deletion state machine forward
> 
>     * change worker.state to RUNNING or STOPPING or STOPPED based on VM state and current worker state
> 
>     * handle lifetime expirations
> 
>     * handle extending worker.expires
> 
> 
> This is all in there, but inter-mingled. Let's factor it out into nice functions.

Where is  the code for some of the bullet points raised in https://github.com/taskcluster/taskcluster/issues/3056#issue-638982059? I was only able to find the second one in `checkWorker`.

### DIFF
--- a/services/worker-manager/src/providers/azure.js
+++ b/services/worker-manager/src/providers/azure.js
@@ -688,6 +688,23 @@ class AzureProvider extends Provider {
   }
 
   async checkWorker({ worker }) {
+    const changeNotFoundWorkerState = async (state) => {
+      // VM has not been found, so it is either
+      // 1. still being created
+      // 2. already removed but other resources may need to be deleted
+      // 3. deleted outside of provider actions, should start removal
+      if (state === states.REQUESTED) {
+        // GETs and updates workers that have not registered every loop
+        return this.provisionResources({ worker, monitor });
+      } else if (state === states.STOPPING) {
+        // continuing to stop
+        return this.removeWorker({ worker, reason: 'continuing removal' });
+      } else {
+        // VM in unknown state not found, deleted outside provider
+        return this.removeWorker({ worker, reason: `vm in ${state} not found` });
+      }
+    };
+
     const monitor = this.workerMonitor({
       worker,
       extra: {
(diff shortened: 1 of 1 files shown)


services/worker-manager/src/providers/azure.js
```diff
@@ -688,6 +688,23 @@ class AzureProvider extends Provider {
...
 
   async checkWorker({ worker }) {
+    const changeNotFoundWorkerState = async (state) => {
+      // VM has not been found, so it is either
+      // 1. still being created
+      // 2. already removed but other resources may need to be deleted
+      // 3. deleted outside of provider actions, should start removal
+      if (state === states.REQUESTED) {
+        // GETs and updates workers that have not registered every loop
+        return this.provisionResources({ worker, monitor });
```
djmitche, line 698: await :)

djmitche (commented):
>     * step creation/deletion state machine forward

In `provisionResources` and `removeWorker`, respectively.

>     * change worker.state to RUNNING or STOPPING or STOPPED based on VM state and current worker state
> 
>     * handle lifetime expirations

look for `registrationTimeout` and `terminateAfter`

>     * handle extending worker.expires

I think this got fixed in #3352.

helfi92:
> > ```
> > * step creation/deletion state machine forward
> > ```
> 
> In `provisionResources` and `removeWorker`, respectively.
> 

Having creation be in `provisionResources` and deletion in `removeWorker` seems like the right place to me. They are already abstracted away from `checkWorker`. What did you have in mind?

> > ```
> > * change worker.state to RUNNING or STOPPING or STOPPED based on VM state and current worker state
> > 
> > * handle lifetime expirations
> > ```
> 
> look for `registrationTimeout` and `terminateAfter`
> 

The logic for `terminateAfter` in checkWorker is simply one line. The logic for `reregistrationTimeout` in `registerWorker` is also one line. What simplification did you have in mind?

> > ```
> > * handle extending worker.expires
> > ```
> 
> I think this got fixed in #3352.

djmitche:
Overall I think what I had in mind is a more "readable" approach for these, so that the structure is immediately apparent without mentally parsing each line.  `checkWorker` is almost 100 lines long!  Its core is `try { if ( .. ) { .. } else { .. } } catch { .. }`, but it's not clear what that means and every time I look at the function I have to remind myself how it works. How could that be made so that it reads as more of a pseudocode for what it does, maybe in 10-20 lines?

As for `provisionResources` and `removeWorker`, perhaps those could have different names that indicate that they "drive a thing to completion" and that they are sort of mirror images of one another.  Also, `removeWorker` is a method called from outside the class, but `provisionResources` is not, and that asymmetry is a bit weird.  Any refactoring of the body of those functions is probably best left to #3058, though.

helfi92:
Most of the refactor was done in Pete's PR. Closing for now.

PR #3482, opened 2020-09-09
Contributors update

ccooper:
Move miles, owlish, and walac to former-staff


djmitche:
(I wonder if we should make that part of `yarn generate`..)

djmitche:
```
[15:59:56] Dependencies are used and installed: fail: Error: Extra development dependencies! Remove ["all-contributors-cli"] from devDependencies in package.json (or add to metatests.specialImports if required)
```

Oh, that's why :)

Issue #3347, opened 2020-08-05
Allow azure provider to ignore failed provisioning states


djmitche:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1646061, win7 instances don't have a VM agent and thus transition into `powerStates = ['ProvisioningState/failed/OSProvisioningTimedOut', 'PowerState/running']`.

Let's add an option to an Azure worker pool config, `ignoreFailedProvisioningState`, that will ignore `ProvisioningState/failed*` power states, treating them as "instance is running just fine"  Per #3346, re/registrationTimeout acts as a backstop to this, where if an instance really has failed, it will eventually timeout and be killed.

PR #3489, opened 2020-09-10
Issue 3346 - no longer call fetchVmInfo to determine provisioningState

### LINKED ISSUE #3346: Update azure instance state recognition
https://bugzilla.mozilla.org/show_bug.cgi?id=1646061 shows an instance with powerStates `['ProvisioningState/creating']` which should probably be considered transitional.  Also, it seems that `vmInfo.provisioningState` is redundant with the power states?

Now that we have registrationTimeout and reregistrationTimeout, the state recognition can be a little more conservative about killing instances -- if it misses an instance that has failed to start up, that instance's registrationTimeout will expire and it will be terminated anyway.  And if it misses an instance that has failed after startup, reregistrationTimeout will expire and it will be terminated.  So the worker scan's responsibilities are:

 * in state REQUESTED
   * detect failed startup conditions and remove the worker
 * in state RUNNING
   * detect instance stopping or deallocation and remove the worker

Since the [documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/states-lifecycle) on lifecycles is next to useless, let's just look for known, observed situations, and treat everything else as OK (with debug logging for anything that's not known and observed to be good).

petemoore:
Github Bug/Issue: Fixes #3346

In progress, not ready for review yet.

### DIFF
--- a/services/worker-manager/src/providers/azure.js
+++ b/services/worker-manager/src/providers/azure.js
@@ -21,25 +21,15 @@ const { CloudAPI } = require('./cloudapi');
 // Azure provisioning and VM power states
 // see here: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/states-lifecycle
 // same for linux: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/states-lifecycle
-const successPowerStates = new Set([
-  'PowerState/running',
-  'PowerState/starting',
-]);
-const successProvisioningPowerStates = new Set([
-  'ProvisioningState/succeeded',
-  'ProvisioningState/creating',
-  'ProvisioningState/updating',
-]);
 const failPowerStates = new Set([
   'PowerState/stopping',
   'PowerState/stopped',
   'PowerState/deallocating',
   'PowerState/deallocated',
-  'ProvisioningState/failed',
-  'ProvisioningState/deleting',
-  'ProvisioningState/canceled',
-  'ProvisioningState/deallocating',
 ]);
+
+// Provisioning states for any resource type.  See, for example,
+// https://docs.microsoft.com/en-us/rest/api/virtualnetwork/networkinterfaces/createorupdate#provisioningstate
 const failProvisioningStates = new Set(['Failed', 'Deleting', 'Canceled', 'Deallocating']);
 
 // only use alphanumeric characters for convenience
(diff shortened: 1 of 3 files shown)


services/worker-manager/src/providers/azure.js
```diff
@@ -707,47 +713,60 @@ class AzureProvider extends Provider {
...
-        if (workerPool) {
-          await this.reportError({
-            workerPool,
-            kind: 'creation-error',
-            title: 'Encountered unknown VM provisioningState or powerStates',
-            description: `Unknown provisioningState ${provisioningState} or powerStates: ${powerStates.join(', ')}`,
-          });
+
+      // count this worker as having been seen for later logging
+      this.seen[worker.workerPoolId] += worker.capacity || 1;
```
djmitche, line 728: Note that this count _was_ used for provisioning decisions, but that's no longer the case -- the provisioning loop now counts workers directly.  This is only used in the `logSeen` call in this file.  So that it now counts VMs while they are starting up is not going to cause any issues.

services/worker-manager/src/providers/azure.js
```diff
@@ -707,47 +713,60 @@ class AzureProvider extends Provider {
...
+        // It's possible for a newly-requested VM to be running (PowerState/running), but have failed
+        // provisioning.  In this case the VM isn't doing any work, but billing continues.  So, we want
+        // to catch this case and also consider it failed.  These state codes have the form
+        // `ProvisioningState/failed/<SomeCode>`.
+        let failedProvisioningCodes = powerStates
+          .filter(state => state.startsWith('ProvisioningState/failed/'))
+          .map(state => state.split('/')[2]);
+
+        // any failed-provisioning code is treated as a failure
+        if (failedProvisioningCodes.length > 0) {
```
djmitche, line 753: This is leaving room for #3347.

djmitche:
Pete, you're the author and so can't be flagged for review, but please review anyway :)

djmitche:
(I rebased and fixed lint in your commit)

imbstack (approved):

services/worker-manager/src/providers/azure.js
```diff
@@ -707,47 +713,60 @@ class AzureProvider extends Provider {
...
+        // If the worker has not checked in recently enough, we consider it failed regardless of the Azure lifecycle
+        isFailed = true;
+        reason = 'terminateAfter time exceeded';
+      }
+
+      if (isFailed) {
+        // On failure, call `removeWorker`, which works to remove the resources associated with the VM
+        // and eventually mark it as STOPPED
+        state = await this.removeWorker({ worker, reason });
+      }
```
helfi92, line 769: Is it possible that a piece of code ends up going in multiple of these if statements? Right now, the `reason` will always reflect the last matched condition. I wonder if we should rather be using `else if` instead. Thoughts?

helfi92 (approved):
LGTM, just one minor concern.

services/worker-manager/src/providers/azure.js
djmitche: Each is preceded by `!isFailed &&` which aims to stick with the *first* matched condition.  Using else-if is hard with nested if's:
```
if (..) {
  ..
  if (..) {
    isFailed = true;
  }
} else if (..) {
}
```
isn't quite what we want.

djmitche:
Hassan's looking at more refactors of the same function, so we'll get this merged now instead of waiting until tomorrow.

services/worker-manager/src/providers/azure.js
djmitche: Also, come to think of it, the ordering of the checks isn't that important -- if it's both later than terminateAfter *and* the VM is in some bad state, then a removeWorker call with either reason would be fine.

PR #3496, opened 2020-09-11
#3347 - optionally ignore failed provisioning states

### LINKED ISSUE #3347: Allow azure provider to ignore failed provisioning states

djmitche:
Github Bug/Issue: Fixes #3347

### DIFF
--- a/changelog/issue-3347.md
+++ b/changelog/issue-3347.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: minor
+reference: issue 3347
+---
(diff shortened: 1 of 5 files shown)


petemoore (approved):
Looks perfect, many thanks.

Issue #3502, opened 2020-09-14
authenticateHawk response has non-array scopes


djmitche:
https://sentry.prod.mozaws.net/operations/taskcluster-firefox-ci/issues/9581113/activity/?referrer=alert_email
```
Output schema validation error: 
Schema Validation Failed!
Rejecting Schema: https://firefox-ci-tc.services.mozilla.com/schemas/auth/v1/authenticate-hawk-response.json#
Errors:
  * data.scopes should be array
  ..(some other complaints about not matching the error arm of the `oneOf`)..
```
see more comment on the sentry issue.

djmitche:
Searching by traceId (SO COOL!):

![image](https://user-images.githubusercontent.com/28673/93089711-b3fe2e00-f669-11ea-8f70-7247942c04c7.png)

So, it looks like the `authenticateHawk` call was from `queue.claimWork`, was retried after just 200ms, and the retry was successful.

We do zillions of `authenticateHawk` calls, and this is the first time we've seen this issue.

djmitche:
I think I see the issue -- we use `scopes === null` in a place where `scopes` is `undefined`.

djmitche:
Hm, I don't think that's it.  Looking more closely, `this._clients` has clients with undefined `scopes`, but `this._clientCache` is filled in a loop that sets `scopes` to null.

djmitche:
BTW this occurred in v36.0.0, after 3fd060c69c265ecbdf1ef6e6690906dd453c2f5b and 9edca22ec89533160f9cd2cfde57942634489649 modified how the auth service interacts with the DB such that it uses the DB functions directly.  But, `scopes` is calculated and not retrieved from the DB, so I'm not sure how that would be related.  And a look at the diffs didn't reveal any smoking guns.

djmitche:
Hm, I don't think that PR's putting debugging info at the right place.  `queue.claimWork` is called with temporary credentials, so the returned scopes are from `cert.scopes`, not from the client object.

djmitche:
[...]

PR #3504, opened 2020-09-14
#3502 - get more data when scopes is not an array

djmitche:
If my hunch is correct that `client.scopes` is being set to a non-array and then returned from the clientLoader, then this will show what type it has at that point.  If we see schema-validation error again without this assertion failing, then my hunch is incorrect.  Divide and conquer!

Github Bug/Issue: Refs #3502

### DIFF
--- a/changelog/issue-3502.md
+++ b/changelog/issue-3502.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3502
+---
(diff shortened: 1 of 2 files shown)


petemoore (approved):
Fingers crossed! Certainly can't do any harm. Hopefully it will bring us a step closer.

Given the number of requests we process, and the infrequency of this issue, I wonder whether this was just a bit flipping hardware failure etc. Hopefully we'll find out!

djmitche:
> Given the number of requests we process, and the infrequency of this issue, I wonder whether this was just a bit flipping hardware failure etc. Hopefully we'll find out!

Yeah, I've wondered that, too.

PR #3509, opened 2020-09-15
Get all non stopped workers

### LINKED ISSUE #3036: db method to get all non-stopped workers
`services/worker-manager/src/worker-scanner.js` needs to scan *non-stopped* workers to see what they're up to.  Right now, it scans *all* workers and just ignores those with a 'stopped' status, but that's pretty slow.  Let's write a function that returns just the necessary columns, and add an index on `state` while we're at it.

helfi92:
Github Bug/Issue: Fixes #3036.

Pair programmed with Pete. Thanks Pete!


db/test/fns/worker_manager_test.js
```diff
@@ -464,6 +464,49 @@ suite(testing.suiteName(), function() {
...
+          expires: now,
+          state,
+        });
+        i++;
+      }
+
+      const rows = await db.fns.get_non_stopped_workers(null, null, null, null, null);
+
+      assert.equal(rows.length, 6);
+      console.log(rows);
```
djmitche, line 490: I'm not the only one leaving console.log's around :)

db/versions/0047.yml
```diff
@@ -0,0 +1,39 @@
+version: 47
+description: add get_non_stopped_workers db function
```
djmitche, line 2: Up to you as it's a matter of preference, but `get_active_workers` might be a less awkward name..

djmitche (approved):
Looks good, with the unnecessary conditional removed

db/versions/0047.yml
imbstack: I actually like `non_stopped` I think. `active` feels like maybe or maybe not it will include `stopping`?

imbstack (approved):
thanks!

djmitche (approved):

PR #3510, opened 2020-09-15
fix formatting in reportException schema

djmitche:


petemoore (approved):
Nice job, indeed with bullet points is much better. :-)

PR #3519, opened 2020-09-16
Fix gw-workers

### LINKED ISSUE #3518: gw-workers utility is broken since we moved secrets from tc-secrets to password store
**Describe the bug**
The gw-workers tool no longer works.

We should:

* move it to tools directory
* rename it to `workers` instead of `gw-workers` (in anticipation of having it working for all worker implementations)
* not call AWS/GCP directly, but interface with worker manager to get workers
* use secrets from local password store, rather than from taskcluster secrets service

petemoore:
Github Bug/Issue: Fixes #3518

Currently Work In Progress!


djmitche:
This has some dependencies (such as aws-sdk) that aren't required otherwise.  And, it's fairly specific to how we deploy for community-tc.  So, in the process of fixing this, let's move it out to another repository (maybe `taskcluster/community-tc-utils`?)

djmitche:
I moved this WIP to https://github.com/taskcluster/community-tc-utils/pull/1 in a new repository.

Issue #3035, opened 2020-06-11
Write a workers db method that lists workers without provider_data


djmitche:
The provider_data column is large and unnecessary for the listWorkersXxx API methods, and causes queries to be slow.  So let's write a method that can be called by those API methods that does not include it (or anything else that's not needed by the API)

Issue #3032, opened 2020-06-11
Calls to get_workers on firefox-ci are really slow


helfi92:
Reported by @sciurus on slack:

```
   total_minutes    |  average_time_ms  | calls  |                             query                             
--------------------+-------------------+--------+---------------------------------------------------------------
   7.48474414346667 |  6325.13589588733 |     71 | select * from "get_workers"($1,$2,$3,$4,$5,$6)
   3.00127360881668 |  0.61625264030567 | 292212 | select * from "azure_queue_get"($1,$2,$3)
   1.07801060408333 |  1.15940052063168 |  55788 | select * from "queue_tasks_entities_load"($1,$2)
  0.568329448783334 |  144.490537826271 |    236 | select * from "queue_worker_entities_scan"($1,$2,$3,$4,$5)
      0.17861192615 |  3.40862454484733 |   3144 | select * from "queue_artifacts_entities_scan"($1,$2,$3,$4,$5)
  0.123711863366666 | 0.354255323915429 |  20953 | select * from "queue_artifacts_entities_load"($1,$2)
 0.0543825321166666 |  4.64146789046941 |    703 | select * from "queue_tasks_entities_scan"($1,$2,$3,$4,$5)
 0.0465551093666667 | 0.219358140568557 |  12734 | select * from "queue_worker_entities_load"($1,$2)
 0.0436504890999999 | 0.527073726303077 |   4969 | select * from "queue_worker_entities_modify"($1,$2,$3,$4,$5)
       0.0429933757 |  1.02813971383021 |   2509 | select * from "indexed_tasks_entities_scan"($1,$2,$3,$4,$5)
```

generating temp files:

```
temp_blks_written | average_temp_blks_written | temp_blks_read | average_temp_blks_read | calls |                           query                            
-------------------+---------------------------+----------------+------------------------+-------+------------------------------------------------------------
          20193269 |                    126207 |       18002237 |                 112513 |   160 | select * from "get_workers"($1,$2,$3,$4,$5,$6)
```

the query they replaced (at least i think that's what this is) generated some too, but less on average

```
 temp_blks_written | average_temp_blks_written | temp_blks_read | average_temp_blks_read |  calls   |                           query                            
-------------------+---------------------------+----------------+------------------------+----------+------------------------------------------------------------
       12857630591 |                       813 |    12464574081 |                    789 | 15797803 | select * from "wmworkers_entities_scan"($1,$2,$3,$4,$5)
```

[...]

PR #3520, opened 2020-09-16
Improve performance in worker manager

helfi92:
Relates to #3032.
Fixes https://github.com/taskcluster/taskcluster/issues/3035.


petemoore (changes requested):
* I think you're missing the changelog
* For some reason, the decision task broke, so I think none of the taskcluster CI tasks ran. Can you check the cause, and if it was an intermittent, maybe just rerun? The error said something about invalid json, so it may be a genuine failure that needs fixing, or maybe just needs rebasing on `main` etc, I'm not sure...
* Is `get_workers` used anywhere now? If not, can we remove it?

I'm a little worried that the number of database functions we have to maintain is increasingly rapidly, and also that we have so much copy/paste between similar functions, that we end up having a high maintenance cost to keep everything in sync when changes are made due to the duplication. This isn't really something for this PR, but a general observation that this new model may have some scaling issues, so something we should keep a close eye on...

helfi92:
>     * I think you're missing the changelog

Done.

>     * For some reason, the decision task broke, so I think none of the taskcluster CI tasks ran. Can you check the cause, and if it was an intermittent, maybe just rerun? The error said something about invalid json, so it may be a genuine failure that needs fixing, or maybe just needs rebasing on `main` etc, I'm not sure...
>       **UPDATE: cause appears to be #3521**
> 
>     * Is `get_workers` used anywhere now? If not, can we remove it?
> 

It doesn't look like it's being used anymore. We can probably deprecate it? I can create an issue to deprecate the method and then we can do it in a follow up if that's fine with you.

> I'm a little worried that the number of database functions we have to maintain is increasingly rapidly, and also that we have so much copy/paste between similar functions, that we end up having a high maintenance cost to keep everything in sync when changes are made due to the duplication. This isn't really something for this PR, but a general observation that this new model may have some scaling issues, so something we should keep a close eye on...

I agree that we need to keep an eye on the db functions that we create. It was a trade-off we decided to pick by not allowing direct access to the db.

djmitche:
> It doesn't look like it's being used anymore. We can probably deprecate it? I can create an issue to deprecate the method and then we can do it in a follow up if that's fine with you.

Why couldn't it be deprecated in this PR?

helfi92:
> Why couldn't it be deprecated in this PR?

It could always be deprecated in this PR but since deprecating get_workers is not linked with improving performance in worker manager, I figured we can do it in a separate PR.

helfi92:
Hm, the tests use get_workers and provider_data heavily. Maybe we shouldn't deprecate that function after all?

helfi92:
Here's an example (one of many) where it is used in tests https://github.com/taskcluster/taskcluster/blob/e17a2f2372add8f60f16e7a25c984d689bf5628a/services/worker-manager/test/provider_aws_test.js#L144-L152

Having to call get_worker for each element returned from `get_workers` seems unnecessary and would make our tests run slower. Thoughts?

djmitche:
> Having to call get_worker for each element returned from get_workers seems unnecessary and would make our tests run slower. Thoughts?

An extra query or two in tests won't add too much time, and I believe the workerId(s) are known so the tests could potentially be changed to just call `get_worker` without `get_workers`.

To pete's point on the proliferation of DB functions, it feels weird to me to leave a function around just to support testing.  But leaving it around to a followup in the interests of finishing the task at hand quickly is OK, I think.  Removing the old function would make a good-first-bug, too.

helfi92:
> To pete's point on the proliferation of DB functions, it feels weird to me to leave a function around just to support testing.

I totally agree.

>  But leaving it around to a followup in the interests of finishing the task at hand quickly is OK, I think. Removing the old function would make a good-first-bug, too.

👍

db/versions/0048.yml
```diff
@@ -0,0 +1,44 @@
+version: 48
+description: Add an index on state in the workers table
+migrationScript: |-
+  begin
+    create index workers_state_idx on workers (state);
+  end
+downgradeScript: |-
+  begin
```
djmitche, line 8: should probably drop the index here, otherwise an upgrade -> downgrade -> upgrade again cycle will fail with an error that the index already exists.

djmitche (changes requested):
Looking good otherwise.

petemoore (approved):

db/versions/0048.yml
helfi92: Out of curiosity, do we know why we never dropped an index in `downgradeScript` before?

djmitche: I suspect previous indexes have been introduced at the same time as the table they index, and dropping a table automatically drops associated indexes.  I think this is the first time that we've added an index in a version different from that where the table was added.  But, that's just a hunch -- I didn't check.

helfi92: That would make sense. Thanks!

services/worker-manager/src/provisioner.js
```diff
@@ -143,12 +143,10 @@ class Provisioner {
       };
 
       // Check the state of workers (state is updated by worker-scanner)
-      const fetch = async (size, offset) => await this.db.fns.get_workers(null, null, null, null, size, offset);
+      const fetch = async (size, offset) => await this.db.fns.get_non_stopped_workers(null, null, null, size, offset);
```
djmitche, line 146: It seems this got missed in #3509 / 7b7713c6e01833a76a4747c87c2588075bfcca1c?

djmitche (approved):
Looks good.  Nice catch in using `get_non_stopped_workers` :)

services/worker-manager/src/provisioner.js
helfi92: That PR was trying to tackle https://github.com/taskcluster/taskcluster/issues/3036 which only targeted the call in worker-scanner.

PR #3544, opened 2020-09-21
upgrade to newly-released team-owned packages

djmitche:
This upgrades dockerode-process (which drops the unnecessary dockerode requirement, and with it the outdated dockerode-promise requirement), as well as docker-exec-websocket-server (where the major version bump is because it drops the client functionality, which docker-worker does not use).


imbstack (approved):

imbstack (changes requested):
ooh, I take it back. test failures look related?

petemoore (approved):
Looks sane.

PR #3559, opened 2020-09-23
Fix broken link in dev-docs README.md

47analogy:
Observed that the Best Practices link in the dev-docs README.md was landing on a 404 page. 
Made a slight edit so that it now lands in the best-practices directory.


djmitche (approved):
Thank you!!

PR #3582, opened 2020-09-25
Update worker-manager, built-in, and purge-cache to use taskQueueId/workerPoolId internally

### LINKED ISSUE #3579: use taskQueueId internally to services other than queue
For all services except queue, the impact of provisionerId / workerType is pretty minor.  This bug covers refactoring those to use taskQueueId internally, converting back and forth to provisionerId/workerType as necessary to keep external interfaces unchanged.

djmitche:
Worker-manager was built around workerPoolId, so it only mentioned provisionerId/workerType parenthetically in its logging.

The built-in workers had a workerType in `config.yml`, but it didn't come from an env var so it was as good as hard-coded.  So I just hard-coded it into a taskQueueId.  Similarly for the worker group.

Turns out the only non-queue service that has a provisioner_id / worker_type column in the DB is purge-cache, so I added a DB version to migrate those to a single task_queue_id column.

With all of this in place, the remaining changes for these services will be to remove the conversions to/from provisionerId/workerType once the queue's APIs and their own APIs refer directly to taskQueueId.

Github Bug/Issue: Fixes #3579


djmitche:
(I tested `db` and `purge-cache` manually on the "phase 1" commit and both passed)

helfi92 (approved):
LGTM, well done!

djmitche:
I realized this weekend while I was afk that purge-cache is really about workers, so that should probably be `worker_pool_id` in the DB and not `task_queue_id`.  A quick bit of search-and-replace should fix that up.

djmitche:
(and yes, I think the test failures were unrelated, but let's see what happens now)

helfi92 (approved):

PR #3593, opened 2020-09-25
Upgrade googleapis to latest, to remove dependency on vulnerable node-forge

djmitche:
https://github.com/googleapis/google-api-nodejs-client/releases is not exactly helpful regarding what "breaking" changes have occurred.  @imbstack do you have an idea about how best to test this upgrade?

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "fast-json-stable-stringify": "^2.0.0",
     "generate-password": "^1.5.0",
     "get-caller-file": "^2.0.5",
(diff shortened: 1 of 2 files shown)


imbstack (approved):
whoops, sorry I've been missing this

I think the best plan is to run this in my dev env and see what happens. I can do that now and land it if nothing blows up

djmitche:
Sounds good, thanks!

imbstack:
This works afaict!

PR #3594, opened 2020-09-28
Do not try to update CoT hashes when a task is aborted

### LINKED ISSUE #3525: Docker-worker fails to kill the CoT feature
CoT-enabled tasks appear to fail when they get terminated by the cloud provider (exception/worker-shutdown):

`Sep 17 20:34:43 ip-10-143-24-199 docker-worker: 2020/09/17 20:34:43 {"type":"error killing states","source":"top","provisionerId":"gecko-3","workerId":"i-0d8c2ea5d12f1f195","workerGroup":"us-west-1","workerType":"b-linux","workerNodeType":"c5.4xlarge","error":"Could not kill states properly. Error: Error: Error calling 'killed' for chainOfTrust : TypeError: Cannot set property 'public/logs/certified.log' of undefined\n at ChainOfTrust.killed (/home/ubuntu/docker-worker/src/lib/features/chain_of_trust.js:71:54)\n at async Promise.all (index 1)\n at Task.abortRun (/home/ubuntu/docker-worker/src/lib/task.js:502:7)\n at Task.run (/home/ubuntu/docker-worker/src/lib/task.js:968:14)\n at Task.start (/home/ubuntu/docker-worker/src/lib/task.js:700:17)\n at TaskListener.runTaskset (/home/ubuntu/docker-worker/src/lib/task_listener.js:524:9)\n at async Promise.all (index 0)\n at /home/ubuntu/docker-worker/src/lib/states.js:58:15\n at processTicksAndRejections (internal/process/task_queues.js:93:5)\n at Task.abortRun (/home/ubuntu/docker-`

djmitche:
Github Bug/Issue: Fixes #3525

### DIFF
--- a/workers/docker-worker/src/lib/features/chain_of_trust.js
+++ b/workers/docker-worker/src/lib/features/chain_of_trust.js
@@ -41,7 +41,7 @@ class ChainOfTrust {
   }
 
   async killed(task) {
-    if (task.isCanceled()) {return;}
+    if (task.isCanceled() || task.isAborted()) {return;}
 
     // Ensure the stream is completely written prior to uploading the temp file.
     await streamClosed(this.stream);
--- a/changelog/issue-3525.md
+++ b/changelog/issue-3525.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: silent
+reference: issue 3525
+---
+When a docker-worker task is aborted, the chain-of-trust feature will correctly shut down and not log an error.


workers/docker-worker/src/lib/features/chain_of_trust.js
```diff
@@ -41,7 +41,7 @@ class ChainOfTrust {
   }
 
   async killed(task) {
-    if (task.isCanceled()) {return;}
+    if (task.isCanceled() || task.isAborted()) {return;}
```
petemoore, line 44: Currently if a task is aborted (e.g. because it is not completed within `maxRunTime`), the certified log and other chain of trust artifacts will not be uploaded, so it will make verifying their integrity more difficult. Is this definitely desirable?

@escapewindow Do you care whether timed-out tasks lose their certified logs and chain of trust certificate etc, or do you only care about chain of trust artifacts for successful tasks?

petemoore: I see the issue in #3525 now, and realise this fix was motivated by failing worker-shutdowns rather than task timeouts (which I believe are also considered task abortions).

It looks like the error message in #3525 is due to [this line](https://github.com/taskcluster/taskcluster/blob/5a25a717299e9eae61d7ed0935ceb028b8319a26/workers/docker-worker/src/lib/features/artifacts.js#L207) not having run at the time the task is aborted, so I wonder if that initialisation just needs to happen sooner?

escapewindow: > With this change, if a task is aborted (e.g. because it is not completed within `maxRunTime`), the certified log and other chain of trust artifacts will not be uploaded, which would make verifying their integrity more difficult. Is this definitely desirable?
> 
> @escapewindow Do you care whether timed-out tasks lose their certified logs and chain of trust certificate etc, or do you only care about chain of trust artifacts for successful tasks?

Currently, and for the foreseeable future, we only care about CoT for successful tasks. I don't know if we can trust artifacts from unsuccessful tasks for release purposes in general.

djmitche: My logic was that if we're not doing CoT for cancelled tasks, then aborted tasks do not need it either.  And the current behavior is for an aborted task to fail with the error described in #3525 and without CoT artifacts, so this isn't changing current behavior -- just avoiding logging an unnecessary error.

petemoore (approved):
In that case, I think we can go with this version for now. It still feels like fixing the race condition would be the better long-term fix, but since we don't have a strong use case for signed artifacts in failed tasks, let's go with this.

PR #3597, opened 2020-09-28
Add default content-type to tc-proxy requests

### LINKED ISSUE #3521: taskcluster-proxy fails with docker-worker 37.1.0
https://community-tc.services.mozilla.com/tasks/UxWAGXW5Q7WWKY3GeQqdEA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FUxWAGXW5Q7WWKY3GeQqdEA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log
```
[task 2020-09-16T17:44:14.566Z] ---
[task 2020-09-16T17:44:14.566Z] 
[task 2020-09-16T17:44:14.566Z] * method:     createTask
[task 2020-09-16T17:44:14.566Z] * errorCode:  MalformedPayload
[task 2020-09-16T17:44:14.566Z] * statusCode: 400
[task 2020-09-16T17:44:14.566Z] * time:       2020-09-16T17:44:14.395Z
[task 2020-09-16T17:44:14.585Z] 2020-09-16 17:44:14,585 - ERROR - Payload must be JSON with content-type: application/json got content-type: null
[task 2020-09-16T17:44:14.585Z] 
[task 2020-09-16T17:44:14.585Z] ---
[task 2020-09-16T17:44:14.585Z] 
[task 2020-09-16T17:44:14.585Z] * method:     createTask
[task 2020-09-16T17:44:14.585Z] * errorCode:  MalformedPayload
[task 2020-09-16T17:44:14.585Z] * statusCode: 400
[task 2020-09-16T17:44:14.585Z] * time:       2020-09-16T17:44:14.554Z
[task 2020-09-16T17:44:14.591Z] Traceback (most recent call last):
[task 2020-09-16T17:44:14.591Z]   File "/builds/worker/.local/lib/python2.7/site-packages/taskgraph/main.py", line 481, in main
[task 2020-09-16T17:44:14.592Z]     args.command(vars(args))
[task 2020-09-16T17:44:14.592Z]   File "/builds/worker/.local/lib/python2.7/site-packages/taskgraph/main.py", line 321, in decision
[task 2020-09-16T17:44:14.592Z]     taskgraph_decision(options)
[task 2020-09-16T17:44:14.592Z]   File "/builds/worker/.local/lib/python2.7/site-packages/taskgraph/decision.py", line 113, in taskgraph_decision
[task 2020-09-16T17:44:14.592Z]     create_tasks(tgg.graph_config, tgg.morphed_task_graph, tgg.label_to_taskid, tgg.parameters)
[task 2020-09-16T17:44:14.592Z]   File "/builds/worker/.local/lib/python2.7/site-packages/taskgraph/create.py", line 107, in create_tasks
[task 2020-09-16T17:44:14.592Z]     f.result()
[task 2020-09-16T17:44:14.592Z]   File "/usr/local/lib/python2.7/dist-packages/concurrent/futures/_base.py", line 455, in result
[task 2020-09-16T17:44:14.592Z]     return self.__get_result()
[task 2020-09-16T17:44:14.592Z]   File "/usr/local/lib/python2.7/dist-packages/concurrent/futures/thread.py", line 63, in run
[task 2020-09-16T17:44:14.595Z]     result = self.fn(*self.args, **self.kwargs)
[task 2020-09-16T17:44:14.595Z]   File "/builds/worker/.local/lib/python2.7/site-packages/taskgraph/create.py", line 133, in create_task
[task 2020-09-16T17:44:14.595Z]     res.raise_for_status()
[task 2020-09-16T17:44:14.595Z]   File "/usr/local/lib/python2.7/dist-packages/requests/models.py", line 940, in raise_for_status
[task 2020-09-16T17:44:14.595Z]     raise HTTPError(http_error_msg, response=self)
[task 2020-09-16T17:44:14.595Z] HTTPError: 400 Client Error: Bad Request for url: http://taskcluster/queue/v1/task/Oz-P-HHAQaOW7fYaE4wk7Q
```

with these workers.  Presumably taskcluster-proxy is stripping content-type or something like that?

imbstack:
This seems to be the easiest way to restore what we believe legacy behavior was. Still pretty amateurish at go so please feel free to trash this review!

Github Bug/Issue: Fixes #3521

### DIFF
--- a/changelog/issue-3521.md
+++ b/changelog/issue-3521.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 3521
+---
+Taskcluster proxy now automatically adds a `Content-Type: application/json` header if none is specified. This is to match legacy behavior and should be
+unnoticed by users.
--- a/tools/taskcluster-proxy/authorization_test.go
+++ b/tools/taskcluster-proxy/authorization_test.go
@@ -545,3 +545,55 @@ func TestGetResponseBody(t *testing.T) {
 	testWithPermCreds(t, test("tester"), 200)
 	testWithTempCreds(t, test("test:temp-cred-issuer"), 200, "test:authenticate-get")
 }
+
+func TestRequestHeaders(t *testing.T) {
+	test := func(name string, scopes []string) IntegrationTest {
+		return func(t *testing.T, creds *tcclient.Credentials) *httptest.ResponseRecorder {
+			fmt.Printf("-- %s --\n", name)
+
+			server := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+				if req.Header.Get("content-type") != "application/json" {
+					t.Errorf("Missing or incorrect request header: %s", req.Header.Get("content-type"))
+				}
+				res.Header().Set("content-type", "application/json")
+				res.Write([]byte(`{"key": "this needs to be at least 14 bytes long. see above"}`))
+			}))
+			defer server.Close()
+
+			routes := NewRoutes(
+				tcclient.Client{
+					Authenticate: true,
+					RootURL:      server.URL,
+					Credentials: &tcclient.Credentials{
+						ClientID:    creds.ClientID,
+						AccessToken: creds.AccessToken,
+						Certificate: creds.Certificate,
+					},
+				},
+			)
+			if len(scopes) > 0 {
+				routes.Credentials.AuthorizedScopes = scopes
+			}
+
+			req, err := http.NewRequest(
+				"GET", "http://localhost:60024/api/auth/v1/whatever/",
+				new(bytes.Buffer),
+			)
+			if err != nil {
+				log.Fatal(err)
+			}
+			res := httptest.NewRecorder()
+
+			// Function to test
+			routes.APIHandler(res, req)
+			return res
+		}
+	}
+	// We always just return 200 from the test server but we try this with all of the different
+	// cases to try to make sure we send correct headers in all cases
+	testWithPermCreds(t, test("Test with perm creds without authorizedScopes", []string{}), 200)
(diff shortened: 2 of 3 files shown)


petemoore (commented):
Hey Brian,

I checked the full history of `taskcluster-proxy` in the monorepo, and `taskcluster-proxy` in its original repository, and this appears never to have been the behaviour of `taskcluster-proxy` in the past.

TLDR; I think this is the wrong place to make the fix.

I think whatever broke, broke somewhere else, and this fix may end up causing us more pain than it solves. For example, it looks like this will set the `Content-Type` header for requests that have no body, such as GET requests, which could introduce adverse affects in middleware or caching proxies etc. I think passing through only the headers that are sent, plus the Authorization header, is the correct approach, and I suspect whatever was calling taskcluster-proxy is the thing that broke, rather than taskcluster-proxy itself.

I realise this might sound like I'm splitting hairs, but it feels like this change is just to mask problems elsewhere and could lead to unexpected results. For example, calling code that doesn't set the header when it runs through the proxy, wouldn't work if it was pointed at the taskcluster endpoint directly. The clients should all set the `Content-Type` header when it is required, so if it is getting missed somewhere, we have a different problem that this change would mask.

I think the next steps are to do a deep dive to understand what is not passing the Content-Type header in a request to the proxy (whether it is something in a task, something in docker-worker, or something in a taskcluster client) and hopefully we can isolate the component that introduces the issue, and find the commit that broke it, and hopefully find the smoking gun! ☁️ 🔫

jwhitlock:
I agree with @petemoore, setting this header for all proxied requests feels like it covers up the underlying issue and may cause more.

I found a possible change in taskgraph that may set the header (see [my long comment on the bug](https://github.com/taskcluster/taskcluster/issues/3521#issuecomment-700972939)), but the code suggests it would _never_ be set, rather than maybe sometimes unset, which leaves me confused how the code ever worked.

PR #3599, opened 2020-09-29
Issue 3598 - fix db collation check

### LINKED ISSUE #3598: DB collation tests are invalid
**Describe the bug**
In #3219 we introduced db collation tests, that don't accurately capture if the DB collation is correct or not.

They check, for example, that `'a' >= 'A'` - which is true for case sensitive comparison _and_ case insensitive comparison.

Instead the check should be `>` rather than `>=`. For a case-insensitive collation, the first character is always considered less than or equal to the second character, whereas, for a case-sensitive collation, there are entries in the list where the first character is considered higher than the second character.

To recap, the __desired__ collation is case-insensitive ordering (i.e. __doesn't care__ about case).

The pairs checked are [here](https://github.com/taskcluster/taskcluster/pull/3219/files#diff-57f9f48e1fee6178bc651b965dd71599R391). With `>=` changed to `>` they would work correctly as follows:

* `'a' > 'A'` => nondesirable ordering (e.g. case sensitive)
* `'a' > 'b'` => nondesirable ordering (really bad ordering - not even case sensitive ordering)
* `'A' > 'b'` => nondesirable ordering (really bad ordering - not even case sensitive ordering)
* `'a' > 'B'` => nondesirable ordering (e.g. case sensitive)
* `'A' > 'B'` => nondesirable ordering (really bad ordering - not even case sensitive ordering)
* `'0' > 'a'` => nondesirable ordering (some weird ordering for numbers to come after Latin letters)
* `'0' > 'A'` => nondesirable ordering (some weird ordering for numbers to come after Latin letters)

If _none_ of these results return `true`, there is a _reasonable_ chance the collation is suitable, and the ordering is correct.

petemoore:
Github Bug/Issue: Fixes #3598

### DIFF
--- a/libraries/postgres/src/Database.js
+++ b/libraries/postgres/src/Database.js
@@ -423,7 +423,7 @@ class Database {
     await this._withClient('admin', async client => {
       // check the DB collation by its behavior, rather than by name, as names seem to vary.
       for (const pair of ['aA', 'ab', 'Ab', 'aB', 'AB', '0a', '0A']) {
-        const res = await client.query(`select 1 where $1 >= $2`, [pair[0], pair[1]]);
+        const res = await client.query(`select 1 where $1 > $2`, [pair[0], pair[1]]);
         if (res.rows.length > 0) {
           const res = await client.query(`
             SELECT datcollate AS collation
@@ -432,7 +432,7 @@ class Database {
           const collation = res.rows[0].collation;
           throw new Error([
             'Postgres database must have default collation en_US.utf8 (and in particular be case-insensitive for ASCII letters);',
-            `this database is using ${collation}, and sorts '${pair[0]}' >= '${pair[1]}'.`,
+            `this database is using ${collation}, and sorts '${pair[0]}' > '${pair[1]}'.`,
           ].join(' '));
         }
       }
--- a/changelog/issue-3598.md
+++ b/changelog/issue-3598.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3598
+---


djmitche (approved):
Nice fix!

My math professor in college once went on a bit of a rant about how a supposedly "advanced" calculus class apparently didn't understand things as basic as inequalities.  I think he even threw chalk at us during the rant.  Apparently he was right, about me at least.

libraries/postgres/src/Database.js
```diff
@@ -423,7 +423,7 @@ class Database {
     await this._withClient('admin', async client => {
       // check the DB collation by its behavior, rather than by name, as names seem to vary.
       for (const pair of ['aA', 'ab', 'Ab', 'aB', 'AB', '0a', '0A']) {
-        const res = await client.query(`select 1 where $1 >= $2`, [pair[0], pair[1]]);
+        const res = await client.query(`select 1 where $1 > $2`, [pair[0], pair[1]]);
```
helfi92, line 426: I'm not sure I understand. Wouldn't changing '>=' to '>' only make a difference if we are comparing the same character (e.g 'a' > 'a'). None of the pairs have the same character as far as I can see.

In other words, running `select 1 where 'a' > 'A'` and `select 1 where 'a' >= 'A'` both returns `1`. How would that solve ricky's issue?

petemoore: > In other words, running `select 1 where 'a' > 'A'` and `select 1 where 'a' >= 'A'` both returns `1`. How would that solve @ricky26's issue?

For a _correctly_ configured database: `select 1 where 'a' > 'A'` should not return any results, whereas `select 1 where 'a' >= 'A'` should return `1` since `'a'` and `'A'` are considered equal when it comes to ordering in a case-insensitive database.

For an _incorrectly_ configured database: `select 1 where 'a' > 'A'` may return `1` (for example if it has case-sensitive ordering), and `select 1 where 'a' >= 'A'` can also return `1` for the same reason.

Since we only want to return results for incorrectly configured databases, we therefore use the first form (`>`) which does _not_ return `1` for a correctly configured database. 

I put some context in #3598 that goes into a bit more detail.

PR #3600, opened 2020-09-29
Fix error message when postgres collation is bad

djmitche:
The sorting is not "case-insensitive", as that would only be a partial
order.  The important thing for sorting slugids is that different cases
of the same letter be sorted together, which is not what the C collation
does.


petemoore (approved):
Thank you for clarifying this. Indeed I hadn't realised a total ordering was required, but for consistency between systems that totally makes sense, so it is now clear why it is necessary that e.g. 'a' < 'A' rather than them being considered equivalent. 👍

helfi92 (approved):

Issue #3583, opened 2020-09-25
go jobs timing out on macos


djmitche:
```
        [taskcluster 2020-09-25T20:48:42.367Z] Executing command 4: go run 'check-env.go' TASKCLUSTER_WORKER_LOCATION '{"cloud":"9","biscuits":"free"}' RUN_ID 0 TASKCLUSTER_ROOT_URL http://localhost:13243 STRANGE_VAR '()%!^"<>&|%3 r %!4 %~4RS %3 %PATH% %% "  tt'
        [taskcluster:error] Aborting task...
        [taskcluster:warn 2020-09-25T20:51:42.352Z] no such process
``

There are a few other examples that fail in the midst of go downloading dependencies.  Perhaps some hosting site is down or something like that?

djmitche:
https://community-tc.services.mozilla.com/tasks/B2l4_VeQRaOOmdUaHswbOw/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FB2l4_VeQRaOOmdUaHswbOw%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log

djmitche:
Running `go build -tags simple -v -ldflags '-X main.revision=d46cff70827aa7d1d88b1899a1432f4127ce51ca' ./...` locally after `go clean -modcache -cache -testcache` seems to succeed in 15s or so.  Could it be a dirty cache on that worker?

djmitche:
I think this was the issue for which @petemoore just shut down the Apple Silicon mac.

PR #3601, opened 2020-09-29
Sign public artifact urls

ricky26:
Related to taskcluster/taskcluster-rfcs#166.

Add the `signPublicArtifactURLs` configuration variable to the queue service. This should implement the linked RFC entirely.

This is my first contribution to Taskcluster, so please be thorough!


services/queue/src/api.js
```diff
@@ -93,6 +93,7 @@ let builder = new APIBuilder({
   context: [
     'db', // Database instance
     'taskGroupExpiresExtension', // Time delay before expiring a task-group
+    'signPublicArtifactURLs', // Whether to use AWS signed URLs for public artifacts
```
imbstack, line 96: for better or worse it seems like our standard in js is `Urls` rather than `URLs` so let's stick with that here.

services/queue/src/artifacts.js
```diff
@@ -356,7 +356,11 @@ let replyWithArtifact = async function(taskId, runId, name, req, res) {
         skipCache = true;
       }
 
-      if (skipCache && skipCDN) {
+      if (this.signPublicArtifactURLs) {
```
imbstack, line 359: I think this might read a bit better if we swapped the preexisting `if`/`else if` since we don't actually use any cloudmirror/`skipCache`/`skipCDN` logic if `signPublicArtifactURLs` is true. I'm imagining as follows (I think this expresses the same logic, feel free to push back if not):

```js
if (this.signPublicArtifactURLs || bucket === this.privateBucket.bucket) {
  url = await this.privateBucket.createSignedGetUrl(/*...*/);
} else if (bucket === this.publicBucket.bucket) {
  // no diff here
}
```

imbstack (commented):
This is looking really good! I think the `lint-nodejs` failure can be fixed with `yarn lint --fix` and the `meta-generate` error can be fixed by running `yarn generate` at the top-level of the repo. It looks like maybe you hand-edited some of the generated files? We should do a better job at indicating that they're generated. The machinery should take anything you do in `services/queue/config.yml` and automatically update everything else.

I think I saw there's some questions about how to do testing correctly here. I'll look into that next and report back! Thanks for this patch.

services/queue/src/artifacts.js
ricky26: I nearly went this way but I figured I'd have to factor out the bucket object as its own variable (anything else seemed too messy). That would end up having much more impact on the code.

I do agree with the readability argument though, I'll have a go.

djmitche (commented):
The testing approach, which we discussed a bit in matrix, seems great.  I think we don't have a lot of cases where we want to test APIs with different configs.  For the rest I'll just +1 imbstack :)

ricky26:
I'm not sure what happened with the generic-worker build... It looks like it timed out after 60m but the last logs are it still fetching Go dependencies. Is it possible it's something to do with my PR coming from a fork?

djmitche:
No, it's been doing that on and off for a while.  I think there's some go dependency that's hosted on a slow/broken git server somewhere and it just causes the `go get` process to hang.  It's #3583.

services/queue/src/artifacts.js
ricky26: I think I've addressed this now - though I'm not amazingly happy about picking the bucket object. There's technically an edge-case where the specified bucket in the artifact isn't either of the public/private buckets which would be handled differently in this case (but I don't think that's possible to get into without deployment config changes).

imbstack:
@all-contributors please add @ricky26  for code.

PR #3602, opened 2020-09-29
Fix optimistic concurrency in w-m workers

imbstack:
Ok, here's the actual issue (really it is 3 non-issues that interact in a weird way):

1. `update_worker` will default to overwriting rather than conflicting if the input doesn't specify `etag_in`.
1. `get_non_stopped_workers` didn't return the `etag`.
1. `Worker.update` tries to write all fields of a loaded worker when it is called.

By themselves these are fine but in this case where worker-scanner loaded a worker _after_ it was requested but _before_ it registered, the eventual write from the scanner would overwrite the updates from `registerWorker`. Any thoughts on what we can do to avoid this sort of situation going forward? My thinking on the issue is that this is a result of us needing to mash up the results of a db query and a js Object that represents an entity (`Worker` in this case). When we need to do that sort of thing should we be adding a lot of `assert`s to make sure that all fields are represented at all times?

Bugzilla Bug: [1668111](https://bugzilla.mozilla.org/show_bug.cgi?id=1668111)


imbstack:
Ah wait, theoretically `worker.update` should be handling this case (although I think this change will be nice for logging purposes). Still trying to hunt down what is happening here in particular.

djmitche:
> Ah wait, theoretically worker.update should be handling this case (although I think this change will be nice for logging purposes). Still trying to hunt down what is happening here in particular.

Worker.update runs the passed callback "atomically", avoiding the case where two such callbacks run concurrently.  But if that callback unconditionally sets the state, then it will happily set it to "requested" atomically after another call has set it to "running" atomically.  So guarding the changes with `if (worker.state === ..)` makes sense.

I agree that something's funny here, though, since you noted in the bug that the lastModified time is from the registerWorker call, while the state (`stopped`) is from the worker scanner.

djmitche:
> Any thoughts on what we can do to avoid this sort of situation going forward?

Thoughts?  Sure!

> update_worker will default to overwriting rather than conflicting if the input doesn't specify etag_in.

> Worker.update tries to write all fields of a loaded worker when it is called.

It seems like this db function is too feature-ful -- it is either conditional or unconditional, and it can either overwrite all fields or just some of them if NULLs are passed.  I think we should try to make DB functions single-featured.  In this case, that might mean that it's always conditional on etag (so etag=null fails) and that all fields must be supplied, as that fits nicely with the `Worker.update` implementation which would otherwise have to figure out which fields changed.  Another option would be to add purpose-specific functions like `update_worker_state`, `update_worker_provisioner_data` that only address certain fields (but still with etags to make the operations sequential).  Then `Worker.update` would be replaced with more purpose-specific methods, too.

The second option is similar to what's been done with task runs.  All of those functions (`rerun_task`, `cancel_task`, `schedule_task`, etc.) have a big bunch of guards at the beginning for the current state of the task, thereby avoiding the need for an etag.  That could work for `update_worker_state`, if the DB function knows which state transitions are valid and which are not.  So in the case of bug 1668111 it would reject a state transition from `running` to `requested`.

I think `provisioner_data` will always need to be guarded by an etag since it's a free-form data structure that gets modified piece-by-piece (especially in the Azure provider).

> get_non_stopped_workers didn't return the etag.

I agree that asserts should be in place to ensure that a worker row always has the appropriate form (or one of several appropriate forms -- perhaps "with provisionerData" and "without provisionerData"?)  I wrote code to do this for tasks -- so a Task could either be with (`get_task_with_status`) or without (`get_task`) status.  In the end, it was easier to just always include the status, so I removed that code.

Anyway, this is just answering the call for thoughts, not commenting on this PR.

db/test/versions/0051_test.js
```diff
@@ -0,0 +1,6 @@
+const testing = require('taskcluster-lib-testing');
+
+suite(testing.suiteName(), function() {
+  // this version only creates a method, which is tested in unit tests,
+  // so there is nothing to test on upgrade
+});
```
djmitche, line 6: Is this annoying, or helpful as forcing you to consider whether the revision needs tests?

services/worker-manager/test/data_test.js
```diff
@@ -0,0 +1,68 @@
...
+      workerId: 'i-123',
+      providerId: 'testing',
+      capacity: 1,
+      providerData: { terminateAfter: origTerminateAfter }, // use a "normal" registrationTimeout value
+    });
+    await w.create(helper.db);
+
+    // Now loop over the existing workers as we do in scanworker
+    // we do an initial comparison to ensure they make sense up to this point
+    // etag isn't set in the create call so we delete it
```
djmitche, line 39: feels like this should get fixed, too..

djmitche (changes requested):
The previous version of this(?) PR also changed `checkWorker`'s calls to `Worker.update` to only try to update the state when it has actually changed.  I think that's still important, as otherwise the checkWorker call can still set the state back to `requested` after `registerWorker` sets it to `running`.  It's worth checking that the other providers do that correctly, too.

db/test/versions/0051_test.js
imbstack: I don't mind either way. I just copy-paste the last method-only `_test.js`!

djmitche (approved):
Yay!

helfi92 (approved):
🎉

services/worker-manager/src/providers/aws.js
```diff
@@ -362,8 +362,10 @@ class AwsProvider extends Provider {
     monitor.debug(`setting state to ${state}`);
     const now = new Date();
     await worker.update(this.db, worker => {
-      worker.state = state;
-      worker.lastModified = worker.state !== state ? now : null;
+      if (state !== undefined) {
```
helfi92, line 365: When would state ever be `undefined`?

imbstack: If it hasn't changed state it will still be the same value as `let state;`

PR #3605, opened 2020-10-01
Don't enforce but log application/json content type requirement

petemoore:
In order to capture when Content-Type header is not set on incoming requests, but allow it at the same time, this changes the error response to an internal logging call, in order that we can determine which API calls currently do not explicitly set the Content-Type header correctly, that previously succeeded by using the taskcluster-proxy that set it implicitly.

After we've determined all the sources and fixed them, we can enforce the application/json content-type again.

CC @djmitche  @jwhitlock @imbstack

### DIFF
--- a/libraries/api/src/middleware/schema.js
+++ b/libraries/api/src/middleware/schema.js
@@ -46,7 +46,7 @@ const validateSchemas = ({ validator, absoluteSchemas, rootUrl, serviceName, ent
     if (input) {
       if (!typeis(req, 'application/json')) {
         debug(
-          'Payload must have content-type application/json not ' + req.headers['content-type'] || 'none'
+          'Payload must have content-type application/json not ' + req.headers['content-type'] || 'none',
         );
       }
       const error = validator(req.body, input);
--- a/libraries/api/src/middleware/schema.js
+++ b/libraries/api/src/middleware/schema.js
@@ -45,12 +45,9 @@ const validateSchemas = ({ validator, absoluteSchemas, rootUrl, serviceName, ent
     // If input schema is defined we need to validate the input
     if (input) {
       if (!typeis(req, 'application/json')) {
-        return res.reportError(
-          'MalformedPayload',
-          'Payload must be JSON with content-type: application/json ' +
-          'got content-type: {{contentType}}', {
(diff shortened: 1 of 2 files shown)


libraries/api/src/middleware/schema.js
```diff
@@ -45,12 +45,9 @@ const validateSchemas = ({ validator, absoluteSchemas, rootUrl, serviceName, ent
     // If input schema is defined we need to validate the input
     if (input) {
       if (!typeis(req, 'application/json')) {
```
jwhitlock, line 47: This may be out-of-scope, but we're not following the [type-is documented API](https://www.npmjs.com/package/type-is). The second argument should be a list:

```suggestion
      if (!typeis(req, ['application/json'])) {
```

or even ``['json']``. Peeking at the code, it [detects and converts to an array](https://github.com/jshttp/type-is/blob/c1f4388c71c8a01f79934e68f630ca4a15fffcd6/index.js#L55-L61), but I'm wary of using undocumented behavior.

libraries/api/src/middleware/schema.js
```diff
@@ -45,12 +45,9 @@ const validateSchemas = ({ validator, absoluteSchemas, rootUrl, serviceName, ent
...
-        return res.reportError(
-          'MalformedPayload',
-          'Payload must be JSON with content-type: application/json ' +
-          'got content-type: {{contentType}}', {
-            contentType: req.headers['content-type'] || null,
-          });
+        debug(
```
jwhitlock, line 48: As discussed in Slack, we need to log the request URL or at least the task ID in order to hunt down misbehaving clients in production, and `debug` won't add this to the production logs. I'm not familiar with the service code to know what to use, and if we want to log in central logs or send as a message to Sentry

petemoore:
@jwhitlock Apologies, I had forgotten to mark this as a draft pull request, so it wasn't quite ready. Many thanks for reviewing though, and nice catch about the `type-is` API.

djmitche:
Deferring to #3675 

My concern with the approach here is that it introduces the possibility of *new* (ab)uses by calling the API without a content type *apart* from via the taskcluster-proxy.

PR #3610, opened 2020-10-02
Issue 3606 - Implement Slack Notifications

### LINKED ISSUE #3606: Support slack notifications
Background is in [bug 1522154](https://bugzilla.mozilla.org/show_bug.cgi?id=1522154). This should work roughly the same as matrix does currently.

ricky26:
Github Bug/Issue: Fixes #3606

This PR implements Slack support in the notify service. I've tried to follow existing conventions for the other notify backends but I probably need some more guidance as this is the first time I've touched this area.

I'm not sure if I've missed any documentation I should've updated, I tried to find all the places the existing backends were mentioned.

To give you an idea of what the notification looks like in Slack, I've taken a screenshot from my testing instance.
image


services/notify/schemas/v1/slack-request.yml
```diff
@@ -0,0 +1,32 @@
...
+    description: |
+      The unique Slack channel ID, such as `C123456GZ`.
+      In the app, this is the last section of the 'copy link' URL for a channel.
+  text:
+    type: string
+    description: |
+      The main message text. If no blocks are included, this is used as the
+      message text, otherwise this is used as alternative text and the blocks
+      are used.
+  blocks:
```
ricky26, line 21: I was a little stuck for these two - there is technically a specification for them but it would probably be bigger than the schema so far.

services/notify/src/handler.js
```diff
@@ -132,6 +132,58 @@ class Handler {
             throw err;
           }
         }
+        case 'slack-channel': {
+          const channelId = route.slice(2, route.length - 1).join('.');
+
+          const emojiMap = {
```
imbstack, line 138: this would make @owlishDeveloper very happy to see I expect!

services/notify/src/handler.js
```diff
@@ -132,6 +132,58 @@ class Handler {
...
+          if (_.has(task, 'extra.notify.slackText')) {
+            text = this.renderMessage(task.extra.notify.slackText, { task, status, taskId });
+          }
+
+          // This uses Slack blocks format, see https://api.slack.com/messaging/composing/layouts.
+          let blocks = [
+            {
+              type: 'section',
+              text: {
+                type: 'mrkdwn',
```
imbstack, line 157: I'm left wondering why they felt the need to abbreviate this :man_shrugging:

services/notify/src/notifier.js
```diff
@@ -157,6 +158,27 @@ class Notifier {
     this.markSent(roomId, format, formattedBody, body, msgtype);
     this.monitor.log.matrix({ dest: roomId });
   }
+
+  async slack({ channelId, text, blocks, attachments }) {
+    if (!this._slack) {
+      debug('Slack is not configured.');
```
imbstack, line 164: maybe we could use a `this.monitor.warning` here just to make answering "why didn't I get my slack messages" questions easier in the future

imbstack (commented):
I think the only other place to add documentation in `ui/docs/reference/core/notify/usage.mdx` which has bits in there about the `task.extra` fields as well

Thanks, we've wanted this for a looong time and this looks great.

services/notify/src/notifier.js
ricky26: That's a good idea, this did seem like a bit of a weak logging mechanism when I wrote it. I'll sort that out.

services/notify/src/handler.js
ricky26: I was wondering whether this might be considered a little out of place for the base implementation. I copied this from our old CI system and I find the emojis so good for noticing what's going on at a glance. I'm happy it seems to have gone down well! :)

imbstack (approved):
sweet. we'll land this and make a release shortly!

PR #3615, opened 2020-10-05
RFC 165

petemoore:
See [RFC 165](https://github.com/taskcluster/taskcluster-rfcs/blob/main/rfcs/0165-Anonymous-scopes.md)


libraries/api/src/index.js
```diff
@@ -98,6 +96,12 @@ class APIBuilder {
    *   ].join('\n'),
    *   cleanPayload: payload => payload,           // function to 'clean' the payload for
    *                                               // error messages (e.g., remove secrets)
+   *   implicitServiceScopes: false,               // defaults to true
+   *                                               // when true, the single scope
+   *                                               // 'service:<serviceName>:<name>'
```
djmitche, line 101: I would like to avoid this scope pattern.  It differs from our existing `<service>:<action>:<args>` pattern (and uses camelCase instead of dash-case), leaving us either to explain why we have two half-implemented scope patterns, or migrate to one or the other.  Instead, let's define a scope for each otherwise-anonymous API method that matches other scopes for that service.  There aren't that many, and I'm sure in almost every case the appropriate scope is pretty easy to discern.  Where it's not easy to discern, `<service>:<dash-cased-action>` is a good fallback.

libraries/api/src/index.js
```diff
@@ -132,6 +140,14 @@ class APIBuilder {
     if (options.scopes && !ScopeExpressionTemplate.validate(options.scopes)) {
       throw new Error(`Invalid scope expression template: ${JSON.stringify(options.scopes, null, 2)}`);
     }
+    if (options.implicitServiceScopes) {
+      let serviceScope = `service:${this.serviceName}:${options.name}`;
+      if (options.scopes) {
+        options.scopes = { AllOf: [options.scopes, serviceScope] };
```
djmitche, line 146: This would add additional required scopes, e.g., getting a secret would require both `secrets:get:<secret>` and `service:secrets:get`.  For already-scope-protected API endpoints, there's no need to add such additional scopes.

libraries/api/src/index.js
petemoore: I agree, good call.

libraries/api/src/index.js
petemoore: That's a good point. Hand-tailored scope expressions are indeed better, and less magical, so the scope expressions are now all manually configured.

petemoore:
> @helfi92 and I noticed this is the `authorize()` code that gets activated for this test (`scopeExpression` is `null`):
> 
> https://github.com/taskcluster/taskcluster/blob/d8a98cf3e625a7b54a7f9a44e0de1afc60f41744/libraries/api/src/middleware/auth.js#L247-L249
> 
> Other branches appear to use `next()` or `return next(err)`. Should this branch use `next()` as well?

I don't think so, that seems to happen outside the `authorize` function by the function that calls it. See further down where `authorize` gets called...

https://github.com/taskcluster/taskcluster/blob/742709c17343f1125fd6f9354534ba96d3a7fa94/libraries/api/src/middleware/auth.js#L301-L305

petemoore:
@helfi92 @jwhitlock Everything is working now, but the git history was a complete mess. At the moment all the commits are from me, which is a disservice to the help you provided, so let's catch up later in the week to decide how we want to split this up. It wasn't very easy to keep the original commits because they were intermediate changes in files which were often later changed in different/incompatible ways etc, so perhaps we can just split it up a logically, and then each have some of the commits. I don't mind how we do it, I just wanted to have something that I could put out for review today that could be split up into discrete contained pieces.

@imbstack I think @djmitche is out most of this week, so it would be great if you could take a look at the changes and see what you think as primary reviewer. I'm happy to video conference if you want to step through the changes or discuss anything that isn't clear. There is only one open point that concerns me a little, and that is the workaround of needing to add https://github.com/taskcluster/taskcluster/blob/543d96b3840bfc0af62d69b766cad33251f5a1d3/libraries/api/test/logging_test.js#L69-L70 in order to get [this test](https://github.com/taskcluster/taskcluster/blob/543d96b3840bfc0af62d69b766cad33251f5a1d3/libraries/api/test/logging_test.js#L169-L200) to pass.

Thanks everyone for your help and support!

@ricky26 Please feel free to also review the changes. Many thanks!

ricky26:
I've been sneakily watching from the sidelines since I saw the draft appear. It looks good to me (at least for the bits I know anything about)! Thanks for looking into this, very exciting! 😄

libraries/api/src/middleware/auth.js
```diff
@@ -246,6 +233,7 @@ const remoteAuthentication = ({ signatureValidator, entry }) => {
         // occurs with getArtifact for a public artifact, for example)
         if (!scopeExpression) {
           req.public = true;
+          req.hasAuthed = true;
```
imbstack, line 236: have they really authemticated in this case? I feel like this should still be false unless I'm misremembering the meaning of this flag.

libraries/api/test/logging_test.js
```diff
@@ -45,9 +45,10 @@ suite(testing.suiteName(), function() {
     method: 'get',
     route: '/require-no-scopes',
     name: 'requireNoScopes',
-    title: 'Requre no scopse',
+    title: 'Require no scopes',
```
imbstack, line 48: wow, that was quite the typo-laden string before!

services/secrets/src/api.js
```diff
@@ -101,6 +101,7 @@ builder.declare({
   method: 'get',
   route: '/secrets',
   name: 'list',
+  scopes: 'secrets:list-secrets',
```
imbstack, line 104: I thought this should be `service:secrets:list-secrets`?

imbstack (commented):
Thank you for breaking this up into nice commits! I'm making my way through still. 

One thing I'm wondering is if we should have some sort of test that asserts that all methods have the proper scope name? aka method-name and service-name are the proper ones and they come after a `service:`?

changelog/issue-3615.md
```diff
@@ -0,0 +1,5 @@
+audience: general
+level: major
+reference: issue 3615
+---
+RFC 165 - anonymous API calls are now granted the `assume:anonymous` scope, in order that public API methods can be guarded from public view in private taskcluster deployments. Additionally, the scope expressions of all API customer-facing API methods now include `service:<serviceName>:<methodName>`. Since `service:*` is granted to the anonymous role in Community Taskcluster and Firefox CI Taskcluster, there should be no behavioural change to these taskcluster deployments. However, administrators of private taskcluster deployments can adjust the `anonymous` role in order to further restrict access to specific API methods.
```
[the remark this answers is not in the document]
petemoore: Thanks for this. During the implementation, we decided to not generically guard endpoints with `service:<serviceName>:<methodName>` (see https://github.com/taskcluster/taskcluster/pull/3615#discussion_r499785380) so I'll adapt your improved version to accommodate the variations from the original RFC (and I'll update the RFC too to reflect what we actually implemented, in a separate PR).

libraries/api/src/middleware/auth.js
petemoore: This is a good question. The docs defined `hasAuthed` as _Will be true if the request had `req.authorize()` called on it._ In this case, `req.authorize` has been called, but there again, those docs were written when the `no-auth` case was handled outside of `req.authorize` so this could probably be cleaned up.

The places where `noAuth`'s value is tested are:

1) in `libraries/api/src/middleware/handle.js` where an error is thrown if the API call is not public, and `hasAuthed` is `false`
2) in `libraries/api/src/middleware/logging.js` to only retrieve `clientId` and `expires` if `hasAuthed` is true
3) in `libraries/api/src/middleware/schema.js` where an error is thrown if the API call is not public, and `hasAuthed` is `false` (this seems to be a duplication of point 1)

So pre-RFC-165, there was a simple relationship between whether a method was public, and whether it was necessary for the client's authorization credentials to be authorized: namely, if a method wasn't public (i.e. the rendered scope template resulted in a scopeExpression containing one of more scopes), then the client request had to contain hawk authentication, and be authorized against the required scopes. However, post-RFC-165, it is possible for a non-authorized client request to call a guarded API, so long as the required scopes for the API are satisfied by the scopes of the anonymous role.

So if we were to consider `hasAuthed` to mean whether the full set of scopes available to the API request have been determined, then it would still fulfill its function for 1) and 3) above. For 2) I think it is sufficient to use a `clientId` if one is returned, and the same for `expires`, and not depend on the `hasAuthed` property. We can possibly remove the check in 1 or 3 if we decide they are equivalent (although I would be a little anxious to do this directly in this PR where there is already a lot of other auth change).

services/secrets/src/api.js
petemoore: See https://github.com/taskcluster/taskcluster/pull/3615#discussion_r499785380

libraries/api/test/logging_test.js
djmitche:
```
8026691dab (Dustin J. Mitchell 2019-04-12 21:49:35 +0000  30)     title: 'Requre some scopse',
```
classic!

petemoore:
So I think I've fixed things and "tightened the belt" a little with the internal checks that auth is successful when `res.reply()` is called. Previously, the `res.reply` method threw an error if the request wasn't public (i.e. requires one or more scopes) and authentication wasn't successful. This would catch API handler bugs, such as:

```
try {
  await req.authorize({...});
} catch (err) {
  // not handled
}
res.reply({});
```

In this example, even though `req.authorize` has thrown an error, `res.reply` is still called, which is not allowed. Instead this should be implemented as something like this:

```
try {
  await req.authorize({...});
  return res.reply({});
} catch (err) {
  // special handling here...
  return return res.reportError(...);
}
```

So the checks in the middleware were in place to catch issues like these. However, checking that the authentication was successful does not catch cases where authentication is successful, but authorization fails (e.g. due to missing scopes). I therefore tightened the checks to ensure that if an API call is not public, that not only authentication has to be successful, but authorization (scope validation) too.

After this change a couple of tests failed, which were actually calling `res.reply({})` after `req.authorize` had thrown an error, so I fixed those tests to instead call `res.reportError(...)` instead when authorization fails.

libraries/api/src/middleware/auth.js
petemoore: I've fixed this up now in later commits.

helfi92:
Refer to https://github.com/taskcluster/taskcluster/issues/3698 for the load testing.

services/auth/src/api.js
```diff
@@ -923,6 +930,8 @@ builder.declare({
   method: 'get',
   route: '/scopes/current',
   name: 'currentScopes',
+  // Not sure if this needs to be scope free or not
```
djmitche, line 933: Let's answer this question.. (I'm guessing from the green checkmarks that it's OK?)

changelog/issue-3615.md
```diff
@@ -0,0 +1,64 @@
...
+  - `github:get-repository:*`
+  - `github:latest-status:*`
+  - `github:list-builds`
+  - `hooks:get:*`
+  - `hooks:list-hook-groups`
+  - `hooks:list-hooks:*`
+  - `hooks:list-last-fires:*`
+  - `hooks:status:*`
+  - `hooks:trigger-hook:*`
+  - `index:find-task:*`
```
djmitche, line 31: what about `index:list-namespaces*`?  Are there other scopes omitted here?

services/index/src/api.js
```diff
@@ -69,6 +70,11 @@ builder.declare({
   route: '/namespaces/:namespace?',
   query: paginateResults.query,
   name: 'listNamespaces',
+  scopes: {
+    if: 'namespaceGiven',
+    then: 'index:list-namespaces:<namespace>',
+    else: 'index:list-namespaces',
```
djmitche, line 76: This would mean that the anonymous scopes need `index:list-namespaces` and `index:list-namespaces:*`, or `index:list-namespaces*`.  Is that the best option?

services/index/src/api.js
```diff
@@ -103,6 +113,11 @@ builder.declare({
   method: 'post',
   route: '/namespaces/:namespace?',
   name: 'listNamespacesPost',
+  scopes: {
+    if: 'namespaceGiven',
+    then: 'index:list-namespaces-post:<namespace>',
+    else: 'index:list-namespaces-post',
```
djmitche, line 119: This is just a different way of calling the same method, so this can be `index:list-namespaces..`

services/hooks/src/api.js
```diff
@@ -534,6 +538,7 @@ builder.declare({
   route: '/hooks/:hookGroupId/:hookId/trigger/:token',
   name: 'triggerHookWithToken',
   input: 'trigger-hook.yml',
+  scopes: 'hooks:trigger-hook:<hookGroupId>/<hookId>',
```
djmitche, line 541: This API method is intended to be used as a webhook, so I think `scopes: null` would make sense.

services/hooks/src/api.js
```diff
@@ -29,6 +29,7 @@ builder.declare({
   method: 'get',
   route: '/hooks',
   name: 'listHookGroups',
+  scopes: 'hooks:list-hook-groups',
```
djmitche, line 32: Maybe just `hooks:list-hooks:` (following whatever pattern is selected for `index:list-namespaces..`)

services/github/src/api.js
```diff
@@ -187,6 +187,7 @@ builder.declare({
   method: 'post',
   route: '/github',
   name: 'githubWebHookConsumer',
+  scopes: 'github:consume-webhook',
```
djmitche, line 190: This is a webhook called by GitHub, and is protected by its own access control (`X-Hub-Signature` or something like that) so I think `scopes: null` will make sense.

services/index/src/api.js
```diff
@@ -140,6 +159,11 @@ builder.declare({
   route: '/tasks/:namespace?',
   query: paginateResults.query,
   name: 'listTasks',
+  scopes: {
+    if: 'namespaceGiven',
+    then: 'index:list-tasks:<namespace>',
+    else: 'index:list-tasks',
+  },
```
djmitche, line 166: Similar comment here to `index:list-namespaces..`.

services/index/src/api.js
```diff
@@ -174,6 +202,11 @@ builder.declare({
   method: 'post',
   route: '/tasks/:namespace?',
   name: 'listTasksPost',
+  scopes: {
+    if: 'namespaceGiven',
+    then: 'index:list-tasks-post:<namespace>',
+    else: 'index:list-tasks-post',
+  },
```
djmitche, line 209: Similar to `index:list-namespaces-post`

djmitche (commented):
I've mostly been apart from this effort, so I did not give this a line-by-line, in-depth look.  I made some comments on some of the selected scopes, below.

Also, I realize we never really documented the meanings of `auth-success` and `auth-failed`, but now that there's also `no-auth`, perhaps those should be defined somewhere, maybe in the schema?

Finally, the "Deployment" section of the manual should get a nice page documenting how the `anonymous` role works and how to set it up for various scenarios.  The changelog can point to that doc for details.

libraries/testing/src/fakeauth.js
```diff
@@ -38,6 +38,13 @@ exports.start = function(clients, { rootUrl } = {}) {
           scopes = clients[clientId];
           ext = bewitParts[3] || '';
         }
+      } else {
+        return {
+          status: 'no-auth',
+          scheme: 'none',
+          scopes: anonymousScopes || [],
```
imbstack, line 45: should this default to `[assume:anonymous]` just to be parallel to the actual implementation?

services/auth/src/signaturevalidator.js
```diff
@@ -330,16 +332,23 @@ const createSignatureValidator = function(options) {
...
-      result = {
+      result = result || {
         status: 'auth-success',
         scheme: 'hawk',
         expires: credentials.expires,
         scopes: credentials.scopes,
         clientId: credentials.clientId,
       };
-      if (attributes.hash) {
+      if (attributes && attributes.hash) {
```
imbstack, line 351: this isn't 100% identical to the old logic. It is probably fine to make this change but is there a reason we did?

imbstack (commented):
looking good! I agree with everything dustin said with docs and defining `auth-success` `auth-failed`, and `no-auth`.

I have attempted to give the logic itself a thorough review and can't find anything wrong so far.

services/index/src/api.js
helfi92: Per our conversation on Slack, I will go ahead and change this to only require one scope `index:list-namespaces:<namespace>'. A user/service would need `index:list-namespaces:*` to view all namespaces.

changelog/issue-3615.md
helfi92: Good catch. It looks like the scopes that had `if ... else` were not caught by the grep command.

changelog/issue-3615.md
```diff
@@ -9,55 +9,13 @@ unauthenticated API calls are now granted the scope `assume:anonymous`.
...
-  - `secrets:list-secrets`
-  - `worker-manager:get-worker-pool:*`
-  - `worker-manager:get-worker:*`
-  - `worker-manager:list-providers`
-  - `worker-manager:list-worker-pool-errors:*`
-  - `worker-manager:list-worker-pools`
-  - `worker-manager:list-workers:*`
-
+* To maintain current behavior, the `anonymous` will need to be granted to some scopes. Refer to
+`the [anonymous scope section](https://docs.taskcluster.net/docs/manual/design/apis/hawk/authorized-scopes) in the docs.
```
djmitche, line 13:
```suggestion
* To maintain current behavior, some scopes will need to be granted to the `anonymous`role. Refer to
`the [anonymous role section](https://docs.taskcluster.net/docs/manual/deploying/anonymous-role) in the docs.
```

services/index/src/api.js
```diff
@@ -202,11 +190,7 @@ builder.declare({
   method: 'post',
   route: '/tasks/:namespace?',
   name: 'listTasksPost',
-  scopes: {
-    if: 'namespaceGiven',
-    then: 'index:list-tasks-post:<namespace>',
-    else: 'index:list-tasks-post',
-  },
+  scopes: 'index:list-tasks-post:<namespace>',
```
djmitche, line 193:
```suggestion
  scopes: 'index:list-tasks:<namespace>',
```

ui/docs/manual/deploying/anonymous-role.mdx
```diff
@@ -0,0 +1,56 @@
...
+  - `github:list-builds`
+  - `hooks:get:*`
+  - `hooks:list-hooks:*`
+  - `hooks:list-last-fires:*`
+  - `hooks:status:*`
+  - `index:find-task:*`
+  - `index:list-namespaces:*`
+  - `index:list-tasks:*`
+  - `index:list-tasks-post:*`,
+  - `purge-cache:all-purge-requests`
```
djmitche, line 35:
```suggestion
  - `index:list-tasks:*`
  - `purge-cache:all-purge-requests`
```

ui/docs/manual/deploying/anonymous-role.mdx
```diff
@@ -0,0 +1,56 @@
...
+# Anonymous Role
+
+A single scope `assume:anonymous` will be granted for all calls (with or without credentials).
+The `anonymous` role is aimed at private Taskcluster installations. With this, it will
+be possible to prevent any useful access to the Taskcluster deployment without credentials
+allowing for greater administrator control over "public" endpoints.
+
+## Public endpoints
+
+All unauthenticated API calls are granted the scope `assume:anonymous`.
```
djmitche, line 14:
```suggestion
All API calls are granted the scope `assume:anonymous`.
```

ui/docs/manual/deploying/anonymous-role.mdx
```diff
@@ -0,0 +1,56 @@
...
+In order to have all unauthenticated API calls public, grant the `anonymous` role the following scopes:
```
djmitche, line 15:
```suggestion
In order to allow unauthenticated API, grant the relevant scopes to the `anonymous` role.
A "public" deployment of Taskcluster, allowing unauthenticated read-only access, would grant the following scopes:
```

ui/docs/manual/deploying/anonymous-role.mdx
```diff
@@ -0,0 +1,56 @@
...
+---
+
+# Anonymous Role
+
+A single scope `assume:anonymous` will be granted for all calls (with or without credentials).
+The `anonymous` role is aimed at private Taskcluster installations. With this, it will
+be possible to prevent any useful access to the Taskcluster deployment without credentials
+allowing for greater administrator control over "public" endpoints.
+
+## Public endpoints
```
djmitche, line 12:
```suggestion
## Public Deployments
```

ui/docs/manual/deploying/anonymous-role.mdx
```diff
@@ -0,0 +1,56 @@
+---
+title: Anonymous Role
+---
+
+# Anonymous Role
+
+A single scope `assume:anonymous` will be granted for all calls (with or without credentials).
+The `anonymous` role is aimed at private Taskcluster installations. With this, it will
+be possible to prevent any useful access to the Taskcluster deployment without credentials
+allowing for greater administrator control over "public" endpoints.
```
djmitche, line 10: Maybe add a `## Private Deployments` here, for symmetry, suggesting leaving the role empty.

services/auth/src/signaturevalidator.js
helfi92: Without a check for `attributes`, it was generating the following error:

```
  services/auth/test/client_test.js (mock)
    ✓ ping
    ✓ auth.client (root credentials)
    1) auth.client (no credentials)


  2 passing (2s)
  18 pending
  1 failing

  1) services/auth/test/client_test.js (mock)
       auth.client (no credentials):
     Error: Cannot read property 'hash' of undefined

---

* method:     client
* errorCode:  AuthenticationFailed
* statusCode: 401
* time:       2020-10-21T20:17:47.023Z
```

libraries/testing/src/fakeauth.js
helfi92: I don't remember why this was done but perhaps it was done to allow calling this function with no scopes at all since some endpoints actually don't require any scopes.

libraries/api/src/middleware/logging.js
```diff
@@ -24,7 +24,7 @@ MonitorManager.register({
     statusCode: 'The http status code that the endpoint resolved with.',
     duration: 'The duration in ms of the call.',
     public: 'True if the endpoint requires no scopes.',
-    hasAuthed: 'Will be true if the request had req.authorize() called on it.',
+    hasAuthed: 'Will be `true` if the request has successfully authenticated (`auth-success` or `no-auth`) but `false` if req.authenticate() has not been called (for example when no scopes are required), or if the request has invalid credentials (`auth-failed`). It does *not* imply authorization (i.e. scope satisfaction not guaranteed).',
```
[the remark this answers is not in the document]
helfi92: Done.

libraries/api/src/expressions.js
```diff
@@ -137,10 +137,13 @@ const render = (compiledTemplate, params) => {
...
-        if (typeof params[value] !== 'string' && typeof params[value] !== 'number') {
-          throw new Error(`Scope expression template expected parameter '${value}' to be a string or number`);
+        switch (typeof params[value]) {
+          case 'string': return params[value];
+          case 'number': return params[value].toString();
+          case 'undefined': return '';  // translate undefined to an empty string
+          default:
+            throw new Error(`Scope expression template expected parameter '${value}' to be a string, number, or undefined`);
         }
-        return params[value];
```
helfi92, line 143: This change was required because Dustin and I were not able to have an optional parameter be an empty string. `path-to-regexp` doesn't allow this.

tc-lib-api is applying `req.authorize` because it sees that all of the necessary parameters are present for a route like

```
  route: '/namespaces/:namespace?',
  query: paginateResults.query,
  name: 'listNamespaces',
  scopes: 'index:list-namespaces:<namespace>',
```

In the past that wasn't an issue because we didn't require scopes for this endpoint when namespace is not defined.

The error we were  seeing was `Error: Scope expression template expected parameter 'namespace' to be a string or number`.

/cc @imbstack

libraries/testing/src/fakeauth.js
imbstack: I'm thinking we should have it do what auth actually does. (that is if I understand auth's implementation correctly.

services/auth/src/signaturevalidator.js
```diff
@@ -137,7 +137,7 @@ const limitClientWithExt = function(credentialName, issuingClientId, accessToken
       res.expires = cert_expires;
     }
 
-    res.scopes = scopes = expandScopes(cert.scopes);
+    res.scopes = scopes = cert.scopes;
```
imbstack, line 140: iiuc we don't expand here anymore because we always do it to the result of this function anyway. We can remove `expandScopes` from the arguments of this function and where it is called in that case I believe

services/auth/src/signaturevalidator.js
```diff
@@ -291,8 +293,8 @@ const createSignatureValidator = function(options) {
 
         credentials = authResult.credentials;
         attributes = authResult.artifacts; // Hawk uses "artifacts" and "attributes"
-      } else {
-        // If there is no authorization header we'll attempt a login with bewit
+      } else if (/^\/.*[\?&]bewit\=/.test(req.resource)) {
```
imbstack, line 296: just curious why using a regex instead of `req.query`? Have we disabled query parsing maybe? Let's comment if so.

libraries/api/src/expressions.js
```diff
@@ -137,10 +137,13 @@ const render = (compiledTemplate, params) => {
   if (ctmpl instanceof Array) {
     return ctmpl.map((value, i) => {
       if (i % 2 === 1) {
-        if (typeof params[value] !== 'string' && typeof params[value] !== 'number') {
-          throw new Error(`Scope expression template expected parameter '${value}' to be a string or number`);
+        switch (typeof params[value]) {
```
imbstack, line 140: This used to guard against forgetting a param. The empty string substitution might work but feels dangerous to me. Do we have access to the `?` at the end of `/namespaces/:namespace?` here? We could only allow `undefined` if it is optional. Maybe another choice would be to subsitute a placeholder string or something? /me waves hands a bit

Also happy to be shown why this isn't actually a problem. We could just add the explanation to this as a comment if so.

libraries/api/src/expressions.js
imbstack: This used to guard against forgetting a param. The empty string substitution might work but feels dangerous to me. Do we have access to the ? at the end of /namespaces/:namespace? here? We could only allow undefined if it is optional. Maybe another choice would be to subsitute a placeholder string or something? /me waves hands a bit

Also happy to be shown why this isn't actually a problem. We could just add the explanation to this as a comment if so.

services/index/src/api.js
```diff
@@ -120,6 +127,10 @@ builder.declare({
   ].join('\n'),
 }, async function(req, res) {
   let namespace = req.params.namespace || '';
+  await req.authorize({
+    namespace,
+    namespaceGiven: namespace !== '',
```
imbstack, line 132: it looks like `namespaceGiven` isn't actually used in the scope expression. Am I understanding this correctly?

ui/docs/manual/deploying/anonymous-role.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+## Public Deployments
+
+All API calls are granted the scope `assume:anonymous`.
+In order to allow unauthenticated API, grant the relevant scopes to the `anonymous` role.
+A "public" deployment of Taskcluster, allowing unauthenticated read-only access, would grant the following scopes:
+
+  - `auth:current-scopes`
```
imbstack, line 18: Now that I think about it, this list will grow/shrink over time. How can we make sure we keep it up-to-date? Should we be generating this list somehow?

imbstack (commented):
just a few more questions. I've gone through the code and poked at the tests for auth locally today and all checks out so far

services/auth/src/signaturevalidator.js
helfi92: it looks like it. logging `req` shows:

```
{
  method: 'GET',
  resource: '/?bewit=...',
  host: 'test.taskcluster.net',
  port: 443,
  sourceIp: '127.0.0.1'
}
```

i'll put a comment.

ui/docs/manual/deploying/anonymous-role.mdx
helfi92: At the moment there isn't anything automated to keep it up to date but it's definitely something we should think about. I'll go ahead and create an issue for that assuming it's fine to do this in a follow up.

helfi92: Filed https://github.com/taskcluster/taskcluster/issues/3737.

libraries/api/src/expressions.js
helfi92: > Do we have access to the ? at the end of /namespaces/:namespace? 

Maybe there's a way but nothing comes to mind. I looked at the path-to-regexpx [README](https://github.com/pillarjs/path-to-regexp/blob/master/Readme.md) and didn't find a way to do this.

services/auth/src/signaturevalidator.js
djmitche: This regexp is, I think, copied and reduced from one a few lines later:

https://github.com/taskcluster/taskcluster/blob/f79e42cd338091b677292855bb0ada9fe5e1ac87/services/auth/src/signaturevalidator.js#L304-L307

I think that code parses `req.resource` rather than `req.query` in an effort to do precisely the same thing that the hawk library does.  Switching that around would run the risk of changing something with URL escaping, for example.

In the version before this PR, we just called `hawk.uri.authenticate` whenever there was no authorization header, either ending up with scopes or no-auth.  Now, we need to distinguish those two cases into three, so we need to see if there's a `bewit` query arg.  I'm guessing someone copy-pasted and modified the regexp to do so.  I'm don't totally understand why we need to distinguish three cases, but assuming that's required, this way of doing so makes sense in context.

libraries/api/src/expressions.js
djmitche: Good point, @imbstack - this actually loses some safety.

Another option would be to fix it here:
https://github.com/taskcluster/taskcluster/blob/f79e42cd338091b677292855bb0ada9fe5e1ac87/libraries/api/src/middleware/auth.js#L191-L198
https://github.com/taskcluster/taskcluster/blob/f79e42cd338091b677292855bb0ada9fe5e1ac87/libraries/api/src/utils.js#L2-L14

If that could somehow recognize these "optional" parameters and not return them when calculating useUrlParams, that might achieve what's needed here: tc-lib-api would see that `namespace` isn't always avaliable, so it wouldn't automatically call `req.authorize`, and would leave it to the API method, which can apply the `req.params.namespace || ''` logic.

services/auth/src/signaturevalidator.js
imbstack: Sounds good to both of you. I'm convinced this one is not an issue!

libraries/api/src/expressions.js
imbstack: Awesome. Something like that plus a decision on the fakeauth thing and I think we're all set to land :)

libraries/api/src/middleware/handle.js
```diff
@@ -12,7 +12,7 @@ const callHandler = ({ entry, context, monitor }) => {
     Promise.resolve(null).then(() => {
       return entry.handler.call(req.tcContext, req, res);
     }).then(() => {
-      if (!req.public && !req.hasAuthed) {
+      if (!req.public && !req.satisfyingScopes) {
```
imbstack, line 15: Now that I look at this, won't this throw an error in sentry every time a request is made that doesn't have enough scopes? Checking `authenticated` here was the actual intent iirc.

libraries/api/test/auth_test.js
```diff
@@ -320,8 +324,7 @@ suite(testing.suiteName(), function() {
           } else {
             assert(err.message.match(/This request requires Taskcluster/));
           }
-
-          return res.reply({});
+          return res.reportError('InsufficientScopes', 'Insufficient scopes as intended');
```
imbstack, line 327: Could this change mean that another test in here that asserts a 403 is becoming a false negative? It worries me a bit to change this

services/index/src/api.js
```diff
@@ -84,6 +86,9 @@ builder.declare({
   ].join('\n'),
 }, async function(req, res) {
   let namespace = req.params.namespace || '';
+  await req.authorize({
```
imbstack, line 89: Since the only param in here is in the route we don't need the `req.authorize` anymore, right? Or is it required here due to the optional nature of `namespace`?

imbstack (commented):
Sorry to keep finding little things to question. I got burned so badly the last time I changed this code that I am very afraid of changing it!

This really is looking good though, just a few more questions.

services/index/src/api.js
helfi92: Agreed. I can remove that.

libraries/api/test/auth_test.js
helfi92: Change reverted.

changelog/issue-3615.md
```diff
@@ -0,0 +1,22 @@
...
+---
+RFC 165 has been implemented, allowing for greater administrator control over
+"public" endpoints. Previously these were guarded by no scopes and could be
+accessed by anyone with no way to limit this. In this release all
+unauthenticated API calls are now granted the scope `assume:anonymous`.
+Additionally, most previously unprotected endpoints are now guarded by at
+least one scope, to enable the following:
+
+* To maintain current behavior, some scopes will need to be granted to the `anonymous`role. Refer to
+`the [anonymous role section](https://docs.taskcluster.net/docs/manual/deploying/anonymous-role) in the docs.
```
ricky26, line 13:
```suggestion
  the [anonymous role section](https://docs.taskcluster.net/docs/manual/deploying/anonymous-role) in the docs.
```
It looks like a rogue backtick has gotten in here.

helfi92: Good catch!

libraries/testing/src/fakeauth.js
helfi92: Changing this to `scopes: ['assume:anonymous']` seems to have broken the tests.

libraries/api/src/middleware/auth.js
```diff
@@ -192,7 +179,8 @@ const remoteAuthentication = ({ signatureValidator, entry }) => {
     scopeTemplate = new ScopeExpressionTemplate(entry.scopes);
     // Write route parameters into {[param]: ''}
     // if these are valid parameters, then we can parameterize using req.params
-    let [, params] = utils.cleanRouteAndParams(entry.route);
+    let [, params, optionalParams] = utils.cleanRouteAndParams(entry.route);
+    params = params.filter(param => !optionalParams.includes(param));
```
imbstack, line 183: Let's amend the comment about this to add something like:

> We can only decide to `useUrlParams` if all params are _required_ params. Otherwise if they are not provided the scope checking will fail. This means all endpoints with optional params that get included in the scope expression must call `req.authorize`.

In addition this means (I think) that we'll be automatically authorizing fewer places potentially so we need to make sure we're calling `req.authorize` everywhere that we have optional params?

imbstack (commented):
Ok, so the tests are failing with the change to fakeauth because it was set up "correctly" before. I would undo that change. It is faking having the anonymous role by taking the scopes it would provide as that argument I asked about removing.

In addition, the change to libraries/api/src/middleware/handle.js is explained in https://github.com/taskcluster/taskcluster/pull/3615#issuecomment-710924891 so we can undo that change as well

For the future, when I ask a question in a review I really am asking! No need to change something just because I asked. Oftentimes I am wrong! A good explanation of what I'm missing is always worth it.

imbstack:
Ok, most of the comments I made above are un-breaking things I asked you to break and some docs improvements! 

I'd say the last thing to do here is to come up with a testing plan for this in stage. Since pretty much every endpoint is touched in subtle ways I'd love it if we just split up the work between everyone on the team and try to call each endpoint in a few different ways:

1. in a situation where we have creds that should succesfully work
2. in a situation where we have creds that should be rejected
3. in a situation where we have creds that are valid but don't have enough scopes
4. in a situation where we have no creds
5. ??? maybe more

Obviously the way each endpoint should respond is different for each one but I think we could work our way through them in a relatively short amount of time early next week!

services/index/src/api.js
```diff
@@ -86,9 +86,6 @@ builder.declare({
   ].join('\n'),
 }, async function(req, res) {
   let namespace = req.params.namespace || '';
-  await req.authorize({
```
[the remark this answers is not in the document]
helfi92: Removing this line seems to have caused tests to fail. We can look into that next week.

djmitche:
I had a concern about temp credentials, since #3704 should not be necessary -- `assume:anonymous` should be implicitly included everywhere, regardless of what the credentials look like (permanent, temporary, temporary with authorizedScopes).  I modified the signature validator tests to check this, in two ways: I added a test for this specific case, and then I added a blanket assertion that all of the responses from signatureValidator in the test suite include `assume:anonymous` and it is expanded.

Good news is, all of that is the case!  And now we have more tests!

libraries/api/src/middleware/auth.js
djmitche: This is worth checking manually (it should be quick).  However, we'll get failures if `req.authorize` isn't called either manually or automatically.  For example, commenting out the call in `notify.email()` gives an error running that service's tests.  The error is confusing (send is called twice), but that's known.  The error does get raised, which is what we need here.

djmitche:
I deployed this in my dev env without the anonymous role defined, and at least clicking through the UI I can confirm that it doesn't let me do anything, with reasonable error messages.

djmitche:
Adding that anonymous role, I see the expected behavior both when signed in (including seeing the anonymous role and scopes in my profile) and when signed out.

djmitche:
I added some tests for the auth service, ensuring InsufficientScopes from `auth.client`, `auth.role`, and the other "read-only" methods.

libraries/api/src/utils.js
```diff
@@ -2,13 +2,17 @@
...
   // Find parameters for entry
   const params = [];
+  const optionalParams = [];
   // Note: express uses the NPM module path-to-regexp for parsing routes
   // when modifying this to support more complicated routes it can be
   // beneficial lookup the source of this module:
   // https://github.com/component/path-to-regexp/blob/0.1.x/index.js
   route = route.replace(/\/:(\w+)(\(.*?\))?\??/g, (match, param) => {
     params.push(param);
+    if (match.endsWith('?')) {
```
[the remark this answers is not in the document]
djmitche: 1e28e39f6b5779d20ad8539177c4843a8cbbcc68

djmitche:
I think what's left is:

 * [ ] testing plan as outlined by @imbstack above
 * [ ] test 403 from `github.get` (similar to auth tests, above)
 * [ ] test 403 from `github.builds`
 * [ ] test 403 from `github.badge`
 * [ ] test 403 from `github.repository`
 * [ ] test 403 from `github.latest`
 * [ ] test 403 from `hooks.listHookGroups`
 * [ ] test 403 from `hooks.listHoos`
 * [ ] test 403 from `hooks.hook`
 * [ ] test 403 from `hooks.getHookStatus`
 * [ ] test 403 from `hooks.listLastFires`
 * [ ] test 403 from `index.findTask`
 * [ ] test 403 from `index.listNamespaces`
 * [ ] test 403 from `index.listNamespacesPost`
 * [ ] test 403 from `index.listTasks`
 * [ ] test 403 from `index.listTasksPost`
 * [ ] test 403 from `purgeCache.allPurgeRequests`
 * [ ] test 403 from `purgeCache.purgeRequests`
 * [ ] test 403 from `queue.task`
 * [ ] test 403 from `queue.status`
 * [ ] test 403 from `queue.listTaskGroup`
 * [ ] test 403 from `queue.listDependentTasks`
 * [ ] test 403 from `queue.listProvisioners`
 * [ ] test 403 from `queue.getProvisioner`
 * [ ] test 403 from `queue.pendingTasks`
 * [ ] test 403 from `queue.listWorkerTypes`
 * [ ] test 403 from `queue.getWorkerType`
 * [ ] test 403 from `queue.listWorkers`
 * [ ] test 403 from `queue.getWorker`
 * [ ] test 403 from `queue.listArtifacts`
 * [ ] test 403 from `queue.listLatestArtifacts`
 * [ ] test 403 from `secrets.listSecrets`
 * [ ] test 403 from `workerManager.listProviders`
 * [ ] test 403 from `workerManager.workerPool`
 * [ ] test 403 from `workerManager.listWorkerPools`
 * [ ] test 403 from `workerManager.listWorkerPoolErrors`
 * [ ] test 403 from `workerManager.listWorkersForWorkerGroup`
 * [ ] test 403 from `workerManager.worker`
 * [ ] test 403 from `workerManager.listWorkersForWorkerPool`

imbstack:
I have done my portion of the 403-test list. To close out the "testing plan" point I suggest we split the services into thirds and each person can try out each endpoint that makes sense with the list of things in https://github.com/taskcluster/taskcluster/pull/3615#issuecomment-715629108.

To get to that point all we need to do is land this and get it on stage!

imbstack (approved):

djmitche:
Sounds good to me.  Could we instead do that testing in a dev env?  And while I'm at it, who is taking the initiative to get the `anonymous` role set up for firefox-ci / staging?

djmitche (approved):
r+ for @imbstack's commits

djmitche:
https://pad.mozilla.org/p/rfc165-testing for planning the testing in dev.

Issue #3538, opened 2020-09-21
Deprecate db function get_workers


helfi92:
**Describe the bug**
From https://github.com/taskcluster/taskcluster/pull/3520, we replaced  `get_workers` with `get_workers_without_provider_data`. We'll need to do some cleanup work:
* deprecate `get_workers` (see https://github.com/taskcluster/taskcluster/tree/main/db on how to do that)
* remove usage of `get_workers` in tests

djmitche:
@alopezz is looking at working on this.

PR #3618, opened 2020-10-05
Fix link to fns.md in db/README.md

alopezz:
Just found that a link was broken on that README as I was reading up with the intention of tackling #3538 .


imbstack (commented):
Thanks this looks great. I think the failing test will just be because we need a changelog entry for this (even though we probably shouldn't require one here we don't catch all cases where it can be avoided). Easiest fix will be to just do `yarn changelog --silent` from the top-level of the repo and push again!

djmitche (approved):

helfi92 (approved):

PR #3620, opened 2020-10-05
Deprecate db function `get_workers`

### LINKED ISSUE #3538: Deprecate db function get_workers

alopezz:
Github Bug/Issue: Fixes #3538.

Work-in-progress.

Is this what is intended by the first point in the issue (the deprecation part)?


db/versions/0052.yml
```diff
@@ -0,0 +1,17 @@
...
+      ordered by `worker_pool_id`, `worker_group`, and  `worker_id`.
+      If the pagination arguments are both NULL, all rows are returned.
+      Otherwise, page_size rows are returned at offset page_offset.
+    mode: read
+    serviceName: worker_manager
+    args: worker_pool_id_in text, worker_group_in text, worker_id_in text, state_in text, page_size_in integer, page_offset_in integer
+    returns: table(worker_pool_id text, worker_group text, worker_id text, provider_id text, created timestamptz, expires timestamptz, state text, provider_data jsonb, capacity integer, last_modified timestamptz, last_checked timestamptz)
+    deprecated: true
+    body: |-
+      begin
```
djmitche, line 16: Close, but it needs a copy of the same body, as a deprecated method can still be called for two major Taskcluster releases.

djmitche (changes requested):
You'll also need to run `yarn generate` to update a few places that list functions and DB versions.

Also, this will cause worker-manager tests to fail, because worker-manager tests are now calling a deprecated method.  So those tests will need to be updated in the same PR.

alopezz:
> You'll also need to run `yarn generate` to update a few places that list functions and DB versions.

Thanks for the pointer! Note taken on having a copy of the body of the deprecated function as well.

> Also, this will cause worker-manager tests to fail, because worker-manager tests are now calling a deprecated method. So those tests will need to be updated in the same PR.

I thought as much! I was trying to get early feedback on the first part to make sure I had that right before trying to remove tests. I'm basing my understanding on the comments on PR #3520, but I still need to figure out some more things about the code. Is it preferable that I convert the PR to draft for now?

djmitche:
Draft sounds good, yeah.

alopezz:
I've changed the affected tests to use `get_workers_without_provider_data` instead of `get_workers`, and then using `get_worker_2` (via `Worker.get`) to get the necessary `providerData`.

Is there a better way? The only improvement I can think of is to move the statement that I repeated everywhere, such as
https://github.com/taskcluster/taskcluster/blob/03ef4111da3d6dd2de68b6e0ad2e9b2560349966/services/worker-manager/test/provider_aws_test.js#L125-L130
to `helper.js` at https://github.com/taskcluster/taskcluster/blob/03ef4111da3d6dd2de68b6e0ad2e9b2560349966/services/worker-manager/test/helper.js#L245-L247

Also I removed a couple of tests that referred to the deprecated function in db/test/fns/worker_manager_test.js. Is this the right thing to do here?

Finally, regarding the changelog, this is the way I guess I would have to do it, after reading the docs. Is it right?

```
yarn changelog --issue 3538 --minor --developers
```

djmitche:
That helper function sounds like a good idea.

And, `yarn changelog --issue 3538 --patch --developers` in this case, I think.

I'll have a deeper look tomorrow.  Thanks for working on this!

alopezz:
Regarding the level in the changelog, I thought it was minor because of [this](https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/changelog.md#levels):

> minor: Any change that adds a new version to the database

Though in this case there are no schema changes so I'd see why this may not apply here.

djmitche:
> Regarding the level in the changelog, I thought it was minor because of [this](https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/changelog.md#levels):
> 
> > minor: Any change that adds a new version to the database
> 
> Though in this case there are no schema changes so I'd see why this may not apply here.

Ah, good point.  Let's leave it at minor then.  You're right that this is a particularly minimal DB version that will have zero impact as measured on the db server, but it's best to be consistent in following best-practices!

Thanks for calling me out on that!

djmitche (commented):
I'm happy to merge this once it has a changelog.

djmitche (approved):
Nice work -- thank you!  What's next?

djmitche:
@all-contributors please add @alopezz for code

PR #3621, opened 2020-10-06
Add scopes to endpoints

helfi92:
Relates to [RFC 165](https://github.com/taskcluster/taskcluster-rfcs/blob/main/rfcs/0165-Anonymous-scopes.md).

This covers the following services:
* index
* queue
* purge-cache
* hooks


services/queue/src/api.js
```diff
@@ -120,6 +120,7 @@ builder.declare({
   method: 'get',
   route: '/task/:taskId',
   name: 'task',
+  scopes: 'queue:task:<taskId>',
```
petemoore, line 123: I think I would prefer `queue:get-task:<taskId>`.

services/queue/src/api.js
```diff
@@ -195,6 +197,7 @@ builder.declare({
   route: '/task-group/:taskGroupId/list',
   query: paginateResults.query,
   name: 'listTaskGroup',
+  scopes: 'queue:task-group:<taskGroupId>',
```
petemoore, line 200: Maybe `queue:list-task-group:<taskGrouId>`?

services/queue/src/api.js
```diff
@@ -1802,6 +1810,7 @@ builder.declare({
     quarantined: /^(true|false)$/,
   },
   name: 'listWorkers',
+  scopes: 'queue:listWorkers:<provisionerId>/<workerType>',
```
petemoore, line 1813: `queue:list-workers:<provisionerId>/<workerType>`

petemoore (changes requested):
Note, the failing tests will also need to be fixed.

petemoore:
@helfi92 Note, we'll need to fold this into #3615 rather than merge directly into `main` branch, since without anonymous scopes implemented, the previously public APIs would no longer be public...

Can you close this PR and open a separate PR with the same changes against the `rfc165` branch?

Issue #3580, opened 2020-09-25
allow taskQueueId in task definitions, return it along with provisionerId/workerType


djmitche:

alopezz:
I'll probably do this as the logical follow up to the work I've been doing for #3578.

Issue #3139, opened 2020-06-30
Support one-off maintenance jobs running in k8s


djmitche:
For encryption, we have an outline of a process for automatically re-keying tables, but no implementation of it.  Right now, if a deployer adds a new encryption key for a service, the service will dutifully use it for new and updated columns, but existing columns will remain as they are until they are deleted or updated.

The only encrypted column right now is the worker secret, but soon there will be more.  Let's invent a convenient way to either work this into an existing expiration crontask or create a purpose-specific crontask.

djmitche:
Maybe we should define some one-off jobs to do this sort of online DB maintenance.

We could use it for online DB migrations, too.  That could have three parts:
 * A "regular" DB migration from `yarn db:upgrade` that creates new tables, sets up functions, etc.
 * An online job for the migration that must complete before the next "regular" DB migration can start
 * The next "regular" DB migration checks that the online job is complete and fails with a useful error message if it is not.

So then a deployer would see the equivalent of

```
Running DB migration 0198..
DB migration 0198 complete
Running DB migration 0199
FAILED: Maintenance job 'online-0198' on the 'auth' service must run to completion first
```

The deployer then runs the `taskcluster-auth-maintenance' job with argument `online-0198` (or maybe just `kubectl job run taskcluster-auth-maint-online-0189`) and waits until it completes, at which point the DB migration script can run again and will proceed.

More simply, to re-key the DB, we'd define jobs like `taskcluster-secrets-maint-rekey-secrets` that just scan the table and update the encrypted values to use the latest key.

@sciurus is this similar to anything you've encountered with other tools?  I don't mind re-inventing the wheel, but it's easier to do if I can look at someone else's bicycle while I'm doing it..

djmitche:
For the re-keying jobs at least, it's critical that the job have access to the service's secrets.

sciurus:
I wouldn't think this is worth doing for online migrations alone, but it's a good approach for one-off jobs and could be reused for online migrations too.

From ops' perceptive, the main things we'd have to figure out is where/how these jobs live in the taskcluster repo. They'd need to be separate from the main [helm chart](https://github.com/taskcluster/taskcluster/tree/main/infrastructure/k8s) since we wouldn't want to deploy them each time. It's likely easiest for us to have our existing tooling run them if each job is its own helm chart, but if that's too clunky we could change deploylib to support a different structure.

djmitche:
[...]

PR #3641, opened 2020-10-12
use taskQueueId internally to the queue

### LINKED ISSUE #3578: use taskQueueId internally to the queue
This bug covers refactoring the queue to use taskQueueId internally, converting back and forth to provisionerId/workerType as necessary to keep external interfaces unchanged.

alopezz:
Github Bug/Issue: Fixes #3578 

I've changed all references to `povisionerId/workerType` inside queue related tables in the DB to use `taskQueueId`.
I still have work to do here, to use the new DB schema from the queue service.


alopezz:
I'm moving forward with this, but I've found a place where I'm getting stuck, which is the listWorkerTypes API endpoint:
https://github.com/taskcluster/taskcluster/blob/d8a98cf3e625a7b54a7f9a44e0de1afc60f41744/services/queue/src/api.js#L1678-L1686

This one is supposed to give a list of worker types for the given provisioner id, but if internally we are making references to provisioner id and worker type disappear, it is not straightforward to maintain this endpoint the way it is, as the associated table has no longer knowledge of provisioner ids. Am I missing something? @djmitche I'd be grateful if I could get some clarification here, I'm probably missing something.

djmitche:
Nice work so far :)

I think we'll need to "fake" both listProvisioners and listWorkerTypes.  In both cases, they should probably do a full scan of the task_queues table, splitting the taskQueueId on `/` and filtering down to the appropriate simulated value.

These API methods are "experimental" and we will probably introduce a replacement for them at some point, although what that looks like is a little vague right now.  So, emulating their existing behavior for now is a good fix.

alopezz:
> Nice work so far :)
Thank you for the encouragement :)

> I think we'll need to "fake" both listProvisioners and listWorkerTypes. In both cases, they should probably do a full scan of the task_queues table, splitting the taskQueueId on `/` and filtering down to the appropriate simulated value.

I did think of doing that but wasn't sure about it. Thanks for clearing this up, I'll do it this way.
>

alopezz:
I've made all the necessary to make the tests pass, and this should be (I hope) pretty close to being mergeable. I do have a couple of things that I'd like to clear up.

What is the purpose of the following test?
https://github.com/taskcluster/taskcluster/blob/d8a98cf3e625a7b54a7f9a44e0de1afc60f41744/db/test/fns/queue_test.js#L1097-L1101
Considering the same file has the following test, that seems to be the correct one, going by the arguments that are passed to the `get_queue_worker_types` function.

https://github.com/taskcluster/taskcluster/blob/d8a98cf3e625a7b54a7f9a44e0de1afc60f41744/db/test/fns/queue_test.js#L1080-L1084

I've commented out the one I see as questionable, and I'll remove it completely or restore it depending on the feedback I get.

The other thing is that I wasn't always sure where to make the seam (the conversions back and forth between the split and the joint id). I've tended to make most of the conversions within `api.js`, but I have done at least one exception to avoid having to touch too many files (for `Task.status()`).

djmitche:
Sorry to be slow to respond -- I've been away (from the keyboard) at a conference yesterday and today.

To the first question, it looks like that test is nonsense.  You can just delete it.

I'll try to get a look at the patch soon, if @helfi92 doesn't beat me to it, to answer the question about seam placement.

services/queue/src/data.js
```diff
@@ -168,6 +164,8 @@ class Task {
         };
       }),
     };
+    addSplitFields(status);
```
djmitche, line 167: I see that `definition()` doesn't add the split fields, while `status()` does.  It'd be good to be consistent, either way.  I have a mild preference for both functions returning `taskQueueId` and isolating all of the splitting into `api.js`.

services/queue/src/queueservice.js
```diff
@@ -388,8 +385,7 @@ class QueueService {
     // Construct queue name prefix (add priority later)
     let namePrefix = [
       this.prefix, // prefix all queues
-      hashId(provisionerId), // hash of provisionerId
-      hashId(workerType), // hash of workerType
```
djmitche, line 392: This will change the name of the pending queues, meaning that tasks pending when the upgrade is made will never start.  I think this will have to split the taskQueueId and hash individually.  Moving that to a utility function would make it easy to keep compatibility when we eventually drop the requirement that a taskQueueId contain `/`.

services/queue/src/utils.js
```diff
@@ -109,3 +110,35 @@ const artifactUtils = {
...
+exports.joinTaskQueueId = joinTaskQueueId;
+
+/**
+ * Add the provisionerId, workerType to an object that has a
+ * taskQueueId to maintain public interface compatibility
+ */
+const addSplitFields = (obj) => {
+  const { provisionerId, workerType } = splitTaskQueueId(obj.taskQueueId);
+  obj.provisionerId = provisionerId;
+  obj.workerType = workerType;
```
djmitche, line 142: Maybe rename this to `useSplitFields` and have it also `delete obj.taskQueueId`?  And perhaps a mirror-image `fromSplitFields` that takes an API input with `provisionerId` and `workerType` and replaces those with `taskQueueId`?

djmitche (changes requested):
First, this looks fantastic. This was a huge refactor to undertake, and you've done great work with it!

Aside from the line comments below, where things caught my eye, I mostly have higher-level concerns.

First, we learned recently that performing `alter` operations on several tables leaves the earlier tables locked while performing the later alters.  In [that case](https://github.com/taskcluster/taskcluster/blob/5e31f1a03bfcc685149a53231660ab25e5e5be1c/db/versions/0049.yml#L5-L25) it was simply dropping a column, but one of those tables was already locked and the upgrade hung waiting for that lock -- and holding lots more locks, causing an outage.  So the lesson we learned was to try to split "alter" operations on many tables up into different DB versions.  In this case, I'd suggest splitting the PR up into two PRs: one for worker info (provisioners / worker types / workers) and one for tasks.  Having read through this PR, I think that such a split will be pretty quick, and will certainly help to compartmentalize the reviews.  But if it is hard to split, it's fine to keep it all in a single PR (with two DB revisions).

For the worker-info changes, it seems that the "provisioners" bit has yet to be done.  The existing architecture defines a hierarchy (provisioners, worker types, workers) and I think the new model defines a shorter hierarchy (task queues, workers).  The `listProvisioners`, `getProvisioner`, `listWorkerTypes`, and `getWorkerType` can be marked as deprecated and emulated from this new hierarchy.  It would make sense to add `listTaskQueues` and `getTaskQueue`, too (without an `action` property).  The existing `declareProvisioner` and `declareWorkerType` methods can be deprecated and stubbed out to do nothing, as to my knowledge they are unused.  And, I don't think we need a new `declareTaskQueue` method.  We can make a followup issue to rewrite the UI to use the new methods (I'm not sure this is worth the effort).

For the task changes, we have also learned a painful lesson: deployments of Taskcluster have millions of tasks, and as-written the migration would take tens of minutes, causing a service outage.  We've had some ideas about how to solve this kind of problem in #3139.  I think an [alter table .. add column .. default null](https://stackoverflow.com/questions/47709593/postgresql-add-column-default-null-locks-and-performance) can run quickly regardless of number of rows, so maybe that's the right approach.  That just leaves the question of how to update all the other rows without blocking other activity.  What do you think?

It seems that there's a little bit of #3580 here, in that the queue's schemas have changed to add taskQueueId.  Let's keep this PR free of API changes to the task-related methods, just to help compartmentalize what we need to think about.

Finally, I see that `db/test/fns/queue_test.js` was rewritten to test the new `_tqid` functions.  However, the existing functions have also been rewritten and need tests as well.  In fact, in most cases those tests shouldn't change at all, except to replace `db.fns` with `db.deprecatedFns`.  So `queue_test.js`'s diff should mostly *add* tests.

services/queue/src/data.js
alopezz: Yeah, that's exactly what my earlier question about "seams" was referring to. It's just that I found `status()` was called in more places than just `api.js` which was what led me to treat it differently. I think I agree with your preference here despite meaning more code overall.

services/queue/src/queueservice.js
alopezz: Note taken! I was not aware of the implications of this.

services/queue/src/utils.js
alopezz: Looks like a good idea!

alopezz:
Thank you for the thorough review!
 
> First, we learned recently that performing `alter` operations on several tables leaves the earlier tables locked while performing the later alters. In [that case](https://github.com/taskcluster/taskcluster/blob/5e31f1a03bfcc685149a53231660ab25e5e5be1c/db/versions/0049.yml#L5-L25) it was simply dropping a column, but one of those tables was already locked and the upgrade hung waiting for that lock -- and holding lots more locks, causing an outage. So the lesson we learned was to try to split "alter" operations on many tables up into different DB versions. In this case, I'd suggest splitting the PR up into two PRs: one for worker info (provisioners / worker types / workers) and one for tasks. Having read through this PR, I think that such a split will be pretty quick, and will certainly help to compartmentalize the reviews. But if it is hard to split, it's fine to keep it all in a single PR (with two DB revisions).

I happened to come upon references to this today and I thought that my changes may be falling into that as well, so this comment isn't unexpected. I will split the migrations as you say. Splitting into two PRs may not be hard, though I do fear that I may be lacking git skills to comfortably do so from this point. I'll see.

> For the worker-info changes, it seems that the "provisioners" bit has yet to be done. The existing architecture defines a hierarchy (provisioners, worker types, workers) and I think the new model defines a shorter hierarchy (task queues, workers). The `listProvisioners`, `getProvisioner`, `listWorkerTypes`, and `getWorkerType` can be marked as deprecated and emulated from this new hierarchy. It would make sense to add `listTaskQueues` and `getTaskQueue`, too (without an `action` property). The existing `declareProvisioner` and `declareWorkerType` methods can be deprecated and stubbed out to do nothing, as to my knowledge they are unused. And, I don't think we need a new `declareTaskQueue` method. We can make a followup issue to rewrite the UI to use the new methods (I'm not sure this is worth the effort).

Does this mean the `queue_provisioners` table has to be dropped altogether (along with the Provisioner class in `data.js`, etc.)? I may have to go find you at chat.mozilla.org for more detailed feedback on this part.

> For the task changes, we have also learned a painful lesson: deployments of Taskcluster have millions of tasks, and as-written the migration would take tens of minutes, causing a service outage. We've had some ideas about how to solve this kind of problem in #3139. I think an [alter table .. add column .. default null](https://stackoverflow.com/questions/47709593/postgresql-add-column-default-null-locks-and-performance) can run quickly regardless of number of rows, so maybe that's the right approach. That just leaves the question of how to update all the other rows without blocking other activity. What do you think?

I will look into this and think about it, though this may well be a bit far above my skills (tbh I have limited DB / SQL experience).

> It seems that there's a little bit of #3580 here, in that the queue's schemas have changed to add taskQueueId. Let's keep this PR free of API changes to the task-related methods, just to help compartmentalize what we need to think about.

My bad! Initially I was deleting the `taskQueueId` in the responses; but then I read the RFC again and I saw that at some point all three fields had to be returned. Only later I found that this was split in a separate issue, and decided to wait for feedback. I'll restore schemas and delete `taskQueueId` where necessary.

> Finally, I see that `db/test/fns/queue_test.js` was rewritten to test the new `_tqid` functions. However, the existing functions have also been rewritten and need tests as well. In fact, in most cases those tests shouldn't change at all, except to replace `db.fns` with `db.deprecatedFns`. So `queue_test.js`'s diff should mostly _add_ tests.

Actually I hadn't yet figured out how to call deprecated db functions and that's part of the reason I didn't keep tests for them. I followed your related work on #3582 as a reference (https://github.com/taskcluster/taskcluster/pull/3582/files#diff-f7fe03da59a6f3469999aa814ac2159fed3aeb01923981a418319803b5db8af7), where the deprecated functions were not tested anymore, which was the reason I didn't look further into it.

djmitche:
> Does this mean the queue_provisioners table has to be dropped altogether (along with the Provisioner class in data.js, etc.)? I may have to go find you at chat.mozilla.org for more detailed feedback on this part.

Yes, it does mean that.  There are no attached actions that I know of in any deployment (turns out that wasn't a great idea, I guess), so that shouldn't represent any great loss of data.  And, feel free to ping me in #taskcluster :)

We can talk about solving the gradual-upgrade problem, too.  Maybe we can implement #3139 first, before working on that PR.

For a while, we didn't have a way to test deprecated functions, and were relying on CI running on intermediate commits.  Now we have `db.deprecatedFns` to make that easier..

djmitche:
> For the task changes, we have also learned a painful lesson: deployments of Taskcluster have millions of tasks, and as-written the migration would take tens of minutes, causing a service outage. ...

I want to set some expectations here -- #3139 is probably a nontrivial amount of work, and not necessarily something we (the TC team at Mozilla) have time to work on immediately.  So this part of the PR might end up waiting for a bit until that gets solved, either by someone on our team or someone outside.  That said, we will need #3139 for other things too, eventually, so it probably won't sit around forever!

alopezz:
I've been busier than expected this last week so I've barely done any progress here. From your last comment I gather that I should definitely reduce the scope of the current PR and leave tasks for a follow-up for when #3139 makes the migration possible.

djmitche:
That sounds like a good plan -- thanks for the update!

alopezz:
I've force-pushed to this PR's branch removing all the task-related changes but otherwise leaving it (I think) in the same state of progress that it was. So this PR will from now on be limited to worker info.

After this I will address the rest of the points raised in the previous review.

djmitche:
That change sounds fine and good (sorry for the delay in replying)

alopezz:
This is taking long, but I think I'm getting closer to finishing.

I've implemented dropping the `queue_provisioners` table and reimplemented the related functions to use the `task_queues` function to infer the information. I'm not sure if creating and updating should become noop's, on this first step I've had them create special entries in `task_queues` without the equivalent of a `worker_type_id` because it allowed me to reuse the existing test suite and overall change less code, but it's probably not such a good idea.

Again, for convenience (to get things working with minimal change), in this step I haven't deleted the `Provisioner` class yet and I have used the reimplementation of the db functions to try to emulate what the affected methods would do. I understand that this is not what this should look like and that those should be emulated directly in code by using the `TaskQueue` class and deleting the `Provisioner` class. Also, I assume all those db `queue_provisioners` functions that I've reimplemented should be marked as deprecated. Implementing these changes would be my next step.

djmitche:
> This is taking long, but I think I'm getting closer to finishing.

Don't feel rushed -- it's a big project :)

> I've implemented dropping the `queue_provisioners` table and reimplemented the related functions to use the `task_queues` function to infer the information. I'm not sure if creating and updating should become noop's, on this first step I've had them create special entries in `task_queues` without the equivalent of a `worker_type_id` because it allowed me to reuse the existing test suite and overall change less code, but it's probably not such a good idea.

I agree, I think making those functions no-ops is a good final approach.

> Again, for convenience (to get things working with minimal change), in this step I haven't deleted the `Provisioner` class yet and I have used the reimplementation of the db functions to try to emulate what the affected methods would do. I understand that this is not what this should look like and that those should be emulated directly in code by using the `TaskQueue` class and deleting the `Provisioner` class. Also, I assume all those db `queue_provisioners` functions that I've reimplemented should be marked as deprecated. Implementing these changes would be my next step.

Sounds great!

alopezz:
Another stretch of work is done, having deprecated and stubbed all database functions as required. I've ended up leaving the Provisioner class in `data.js`, trimming it to remove anything that would required the now dropped `queue_provisioners` table. I also expect that I may still have to do some cleanup work.

This issue ended up being much more work and changes than I anticipated! But hopefully we are getting closer now. I understand that a review on your part may require a bit of time as well as the diff is going to be quite large.

> For the worker-info changes, it seems that the "provisioners" bit has yet to be done. The existing architecture defines a hierarchy (provisioners, worker types, workers) and I think the new model defines a shorter hierarchy (task queues, workers). The `listProvisioners`, `getProvisioner`, `listWorkerTypes`, and `getWorkerType` can be marked as deprecated and emulated from this new hierarchy. It would make sense to add `listTaskQueues` and `getTaskQueue`, too (without an `action` property). The existing `declareProvisioner` and `declareWorkerType` methods can be deprecated and stubbed out to do nothing, as to my knowledge they are unused. And, I don't think we need a new `declareTaskQueue` method. We can make a followup issue to rewrite the UI to use the new methods (I'm not sure this is worth the effort).

Should I include these changes inside this PR? I thought that the API had to remain unchanged as much as possible for this part. I guess the deprecation part is simple (just change the parameter in the endpoint function) but I'm not sure what are the consequences. Adding new functions is probably not too hard but I feel like it may be adding too much to an already quite large PR.

djmitche:
> Should I include these changes inside this PR? I thought that the API had to remain unchanged as much as possible for this part. I guess the deprecation part is simple (just change the parameter in the endpoint function) but I'm not sure what are the consequences.

The API surface needs to remain unchanged (meaning, the parameters, accepted schema values, response schemas), but the behavior can change.  The idea of suggesting this was that it's actually the easier option (just do nothing, instead of figuring out how to do something with a `provisionerId` parameter that no longer makes sense).

> > And, I don't think we need a new declareTaskQueue method. We can make a followup issue to rewrite the UI to use the new methods (I'm not sure this is worth the effort).
> 
> ... Adding new functions is probably not too hard but I feel like it may be adding too much to an already quite large PR.

Just to avoid confusion, I was suggesting *not* adding a new function.  Hopefully that's easier :)

I'll have a look at what you've got.  I agree that this is the kind of change that is conceptually simple but ends up being a huge diff.

alopezz:
As I said, there are still things to clean up; just as I was looking again, I realize that there are calls to methods that don't exist (i.e. inside `declareProvisioner`. That one slipped by because I removed the tests that called that. I guess I'll have to stub this one out and leave at least one test that at least checks that the method can be called and returns something reasonable.

To clarify what I meant when I responded to that quote, when I said adding new functions I was referring to:

> It would make sense to add `listTaskQueues` and `getTaskQueue`, too

And what I also wanted to clarify is whether, within this PR, I should change the `stability` property to `APIBuilder.stability.deprecated` for all those methods that you said should be deprecated.

djmitche:
Ah, sorry, I did misunderstand.  Adding those methods later is OK too (I'd be happy to do that after this PR lands, if that helps save you some time).  And yes, please change the stability to deprecated.

db/test/fns/queue_test.js
```diff
@@ -1144,126 +1132,374 @@ suite(testing.suiteName(), function() {
...
-      assert.equal(res[0].description, 'new_desc');
+      assert.deepEqual(res[0].expires, expires);
+      assert.deepEqual(res[0].last_date_active, lastDateActive);
+    });
+
+    helper.dbTest('expire_queue_provisioners return 0', async function(db) {
+      // expire_queue_provisioners is now a no-op, and returns 0 to be consistent
+      let res = await db.deprecatedFns.expire_queue_provisioners(new Date());
+      assert.equal(res[0].expire_queue_provisioners, 0);
+    });
```
djmitche, line 1249: 💯 nice job adding new tests as well as modifying the rest!

djmitche (changes requested):
That's a big PR, but nice work, and I think you made the right decisions where there were judgement calls to me.  Nice work!

The major requested change is the `queueservice.js` issue below.  I will get an estimate of the number of rows in `queue_workers` and see how long it might take to run that `update` statement.  If it's too long, I'd be happy to update this PR to use the new online migration support to accomplish it.

alopezz:
> As I said, there are still things to clean up; just as I was looking again, I realize that there are calls to methods that don't exist (i.e. inside `declareProvisioner`). That one slipped by because I removed the tests that called it. I guess I'll have to stub this one out and leave perhaps one test that at least checks that the method can be called and returns something reasonable.

I fixed this part. I tried to come up with something sensible to replace this behavior, which by my interpretation means basically the equivalent of `getProvisioner`.

> Ah, sorry, I did misunderstand. Adding those methods later is OK too (I'd be happy to do that after this PR lands, if that helps save you some time). And yes, please change the stability to deprecated.

I don't mind adding those methods myself, but I'd rather do that in a follow-up. I've now changed the stability as needed.

I'm trying to correct the hashes in the queue names (keeping the split hashes) but tests start failing in `dependency_test.js`. I'm looking into it.

alopezz:
> I'm trying to correct the hashes in the queue names (keeping the split hashes) but tests start failing in `dependency_test.js`. I'm looking into it.

It's changed and working now. I don't know why I had a failing test when I first tried, though.

db/versions/0053.yml
```diff
@@ -0,0 +1,646 @@
...
+    update task_queues set task_queue_id = provisioner_id || '/' || worker_type;
+    alter table task_queues drop column provisioner_id, drop column worker_type;
+    alter table task_queues alter column task_queue_id set not null;
+    alter table task_queues add primary key (task_queue_id);
+
+    lock table queue_workers;
+
+    alter table queue_workers drop constraint queue_workers_pkey;
+    alter table queue_workers add column task_queue_id text;
+    update queue_workers set task_queue_id = provisioner_id || '/' || worker_type;
```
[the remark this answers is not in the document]
djmitche: There are about 100k rows in this table.  In my dev environment, I can do a million-row table in a few seconds.  So this shouldn't be an issue.

djmitche (approved):
Looks great!

This was a huge project and required a lot of careful thought and attention to detail.  _Thank You!_

I'm going to get this merged now and test it out a bit in a development environment, but I don't expect any issues.

djmitche:
I know we broke half of this project out for later, but please don't feel obliged to work on that if it's not of interest or not something you can work on right now.  Please let us know if there's anything we can do to help you get more involved!

djmitche:
https://dustin.taskcluster-dev.net/provisioners is running this patch and looks 👌🏽

PR #3653, opened 2020-10-13
Fix timing issue in auth logging tests

imbstack:
After poking at this for a while I believe this is actually a result of the logging that we're asserting here happens _after_ the response is sent. This means that the `await` on the response isn't actually ensuring that the log has been written. This poll felt like the easiest way to fix it. 

The test is still failing but in a different and more fixable way now. Feel free to either land this or a better fix. I can help more tomorrow on reviewing and determining if this test should actually pass or fail, etc

### DIFF
--- a/libraries/api/test/logging_test.js
+++ b/libraries/api/test/logging_test.js
@@ -192,7 +192,7 @@ suite(testing.suiteName(), function() {
           clientId: 'client-with-aa-bb-dd',
           hasAuthed: true,
           method: 'GET',
-          public: false,
+          public: true,
           query: {
             private: "0",
           },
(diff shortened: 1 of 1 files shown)


libraries/api/test/logging_test.js
```diff
@@ -173,30 +179,32 @@ suite(testing.suiteName(), function() {
...
+      assert.deepEqual(helper.monitorManager.messages[0], {
+        Type: 'monitor.apiMethod',
+        Severity: LEVELS.notice,
+        Fields: {
+          name: 'sometimesRequireNoScopes',
+          apiVersion: 'v1',
+          clientId: 'client-with-aa-bb-dd',
+          hasAuthed: true,
+          method: 'GET',
+          public: false,
```
helfi92, line 195: This test is failing but passes if we set `public` to `true`. I believe it should be `true` since there are no scopes required. @petemoore can you confirm?

petemoore: That's right - `public` implies that the rendered scope template for the current request has no scopes.

Issue #3632, opened 2020-10-08
db v48 -> v49 hung on a lock


helfi92:
**Describe the bug**
Migration from v48 to v49 hung on a lock and was locking some other tables.

From Slack:

```
taskcluster=>   select relation::regclass, * from pg_locks where not granted and mode='AccessExclusiveLock';
-[ RECORD 1 ]------+--------------------
relation           | queue_artifacts
locktype           | relation
database           | 16415
relation           | 91613110
page               | 
tuple              | 
virtualxid         | 
transactionid      | 
classid            | 
objid              | 
objsubid           | 
virtualtransaction | 12/23931639
pid                | 1060149
mode               | AccessExclusiveLock
granted            | f
fastpath           | f
```

so it can't get lock on `queue_artifacts`.

helfi92:
I wonder if the fix would be to simply get a lock on the tables affected before running this migration. Thoughts?

djmitche:
I think the outage comes from locking table A, then waiting for a lock on table B: while waiting, table A is unusable.

I believe that the actual alter operation is very quick regardless of number of rows (I think it just marks the column as hidden in the table metadata), so it might be best to do this table-by-table: lock table A, alter table A, unlock table A, lock table B, ..

I'm not sure if there's a way to release a lock within a transaction, so that might require multiple transactions -- not something we currently support in a single DB version script.

djmitche:
https://www.postgresql.org/docs/8.1/sql-lock.html
> (There is no UNLOCK TABLE command; locks are always released at transaction end.)

That is the kind of simple, declarative statement I love to read in documentation.  Unfortunately, it doesn't say what I would like to say.

It seems we are "out of the woods" on this in that the upgrade has finished successfully.  So perhaps the lesson to learn is that many `alter` operations which will require a full table lock should not be bundled into a single db version, but spread out over multiple versions, especially isolating frequently-accessed tables in their own db versions.

PR #3654, opened 2020-10-14
Rewrite DB v49

helfi92:
From https://github.com/taskcluster/taskcluster/issues/3632, the agreement here is that we will migrate a single table at a time to avoid having the migration script hang.

Relates to https://github.com/taskcluster/taskcluster/issues/3632.


djmitche:
49's already been released, so it can't be rewritten.  I think it's just a learning experience at this point.

helfi92:
Are we planning to do the migration during a tree closure window in that case?

djmitche:
@sciurus can confirm, but I think the migration is complete now.  Once the lock was cleared by killing the crontask, it went quickly.

sciurus:
Right, this migration is done in all environments, so this PR is not needed.

It looks in this PR like you changed the migrations to do each table in its own transaction, which I agree is the correct approach for migrations that require an exclusive lock.

helfi92:
Ah, great to hear. Thanks folks.

PR #3661, opened 2020-10-16
Issue 3639 - Make sure taskDefined messages are marked unscheduled

### LINKED ISSUE #3639: taskDefined messages have the task state set to 'pending' when there are 0 dependencies
**Describe the bug**
The pulse message `task-defined` has `payload.status.state` set to `'pending'` instead of `'unscheduled'` if the task has 0 dependencies.

**To Reproduce**

1. Go to the Pulse Messages view.
2. Add a new listener for `exchange/taskcluster-queue/v1/task-defined` with `#`.
3. Start listening.
4. Create a new task, with a valid definition.
5. Download the pulse messages.
6. Notice that the `payload.status.state` is `'pending'`.

**Expected behavior**
All tasks should be `'unscheduled'` for the `task-defined` event.

**Taskcluster version**
v37.3.0

**Additional Information**
This seems to be just that the task state is fetched _after_ scheduling the task in `createTask`.

ricky26:
This change prefetches an initial empty status which is used in taskDefined messages. This means that taskDefined messages always has a state of 'unscheduled'.

Github Bug/Issue: Fixes #3639

This ended up being slightly more wiry than I hoped because

### DIFF
--- a/services/queue/src/api.js
+++ b/services/queue/src/api.js
@@ -535,6 +535,12 @@ builder.declare({
   );
 
   let task = Task.fromApi(taskId, taskDef);
+
+  // Fetch the status of the task before creation, so that `taskDefined` messages
+  // have a default status. This can't be run after create, since create is
+  // idempotent and does a DB fetch.
+  let initialStatus = task.status();
+
   try {
     await task.create(this.db);
   } catch (err) {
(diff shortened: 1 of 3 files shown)


services/queue/src/api.js
```diff
@@ -535,6 +535,12 @@ builder.declare({
   );
 
   let task = Task.fromApi(taskId, taskDef);
+
+  // Fetch the status of the task before creation, so that `taskDefined` messages
+  // have a default status. This can't be run after create, since create is
+  // idempotent and does a DB fetch.
```
djmitche, line 541: nice catch!

djmitche (approved):
This looks good, and maintains the at-least-once message semantics.

PR #3662, opened 2020-10-16
Issue 3640 - More notify states

### LINKED ISSUE #3640: Support more task states in notify
At the moment, notify only supports reporting on states which are resolutions. It would be useful to be able to report on other task life-cycle states.

My use case is for deployments. We currently run our backend deployments as tasks. It would be nice to show 'running' as well as the terminal states so people are aware the environment might be in flux.

I spoke to @imbstack about this. It sounds like there might be some concerns about performance impact of adding all states. My use case doesn't need anything more than 'running' though.

ricky26:
This change adds support for all task lifecycle states to notify. It also creates two new pseudostates: 'on-resolved' to replace 'on-any' and 'on-transition' which covers all states.

'on-any' is now deprecated.

Github Bug/Issue: Fixes #3640

### DIFF
--- a/services/notify/src/handler.js
+++ b/services/notify/src/handler.js
@@ -26,12 +26,24 @@ class Handler {
 
     this.pulseClient = pulseClient;
     this.bindings = [
+      queueEvents.taskDefined(`route.${routePrefix}.#.on-defined.#`),
+      queueEvents.taskDefined(`route.${routePrefix}.#.on-transition.#`),
+      queueEvents.taskPending(`route.${routePrefix}.#.on-pending.#`),
+      queueEvents.taskPending(`route.${routePrefix}.#.on-transition.#`),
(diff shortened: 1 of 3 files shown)


ricky26:
CC @imbstack, since I can't request a review.

services/notify/src/handler.js
```diff
@@ -57,6 +69,18 @@ class Handler {
     }
   }
 
+  shouldNotifyOnDecider(decider, state) {
```
djmitche, line 72: I like the phrase "decider"!

djmitche (commented):
This looks good to me, but I'd like @imbstack's review.

imbstack (approved):
heck yes thanks again!

imbstack:
Ah, I forgot we should add docs for this. I'll add them in a separate PR

ricky26:
Easy on the trigger! (If you want, I'm happy to have a stab at those tomorrow.)

PR #3671, opened 2020-10-19
Restore existing tc-proxy behavior of adding content-type header when it is missing

djmitche:
Github Bug/Issue: Refs #3521.

Once this is done, we can park #3521 until we want to remove this support.

### DIFF
--- a/changelog/issue-3521.md
+++ b/changelog/issue-3521.md
@@ -2,4 +2,6 @@ audience: users
 level: minor
 reference: issue 3521
 ---
-Taskcluster-proxy now adds a `Content-Type` header to proxied requests lacking one.  While this behavior is not desirable, it matches the behavior of older versions and real tasks depend on it.
+Taskcluster-proxy now adds a `Content-Type` header to proxied requests lacking one.  While this behavior is not desirable, it matches the behavior of older versions and real tasks depend on it.  A future version of Taskcluster will drop this behavior.
+
+When this occurs, the worker will log a message containing the string "Adding missing Content-Type header".  Use this logging to find tasks that fail to include the `Content-Type` header and adjust accordingly.
--- a/changelog/issue-3521.md
+++ b/changelog/issue-3521.md
@@ -1,0 +1,5 @@
+audience: users
+level: minor
+reference: issue 3521
+---
+Taskcluster-proxy now adds a `Content-Type` header to proxied requests lacking one.  While this behavior is not desirable, it matches the behavior of older versions and real tasks depend on it.
--- a/tools/taskcluster-proxy/authorization_test.go
+++ b/tools/taskcluster-proxy/authorization_test.go
@@ -3,7 +3,6 @@ package main
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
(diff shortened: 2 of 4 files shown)


changelog/issue-3521.md
```diff
@@ -0,0 +1,7 @@
+audience: users
+level: minor
+reference: issue 3521
+---
+Taskcluster-proxy now adds a `content-type` header to proxied requests lacking one.  While this behavior is not desirable, it matches the behavior of older versions and real tasks depend on it.  A future version of Taskcluster will drop this behavior.
+
+When this occurs, the worker will log a message containing the string "Adding missing content-type header".  Use this logging to find tasks that fail to include the `content-type` header and adjust accordingly.
```
jwhitlock, line 7: The message has different capitalization:

```
=== RUN   TestAPICallPOST/Test_with_perm_creds_with_authorizedScopes
2020/10/19 19:07:51 Proxying http://localhost:60024/auth/v1/test-authenticate | POST | https://community-tc.services.mozilla.com/api/auth/v1/test-authenticate
2020/10/19 19:07:51 Adding missing Content-Type header (#3521)
```

tools/taskcluster-proxy/authorization_test.go
```diff
@@ -196,14 +195,20 @@ func TestBewit(t *testing.T) {
 		}
 	}
 
-	testWithPermCreds(t, test(false, 200), 303)
-	testWithTempCreds(t, test(false, 200), 303, "test:authenticate-get")
+	t.Run("perm creds, no authorized scopes",
```
jwhitlock, line 198: 👍 I like the verbose test names

changelog/issue-3521.md
```diff
@@ -0,0 +1,7 @@
+audience: users
+level: minor
+reference: issue 3521
+---
+Taskcluster-proxy now adds a `content-type` header to proxied requests lacking one.  While this behavior is not desirable, it matches the behavior of older versions and real tasks depend on it.  A future version of Taskcluster will drop this behavior.
```
jwhitlock, line 5: Nit: ``Content-Type`` to match other capitalization

tools/taskcluster-proxy/authorization_test.go
```diff
@@ -338,7 +352,9 @@ func TestAPICallPOST(t *testing.T) {
 				"http://localhost:60024/auth/v1/test-authenticate",
 				bytes.NewBufferString(`{"clientScopes": ["test:*", "auth:create-client:test:*"], "requiredScopes": ["test:authenticate-post"]}`),
 			)
-			req.Header["content-type"] = []string{"application/json"}
+			if sendContentType {
+				req.Header["content-type"] = []string{"application/json"}
```
jwhitlock, line 356: Nit: ``Content-Type`` to match other capitalization

jwhitlock (approved):
As mentioned in https://github.com/taskcluster/taskcluster/issues/3521#issuecomment-701498701, here's the 5.1.0 original logic:

https://github.com/taskcluster/taskcluster/blob/ea9ceb74f7f3127cad707087579310539ac9cdf1/clients/client-go/http.go#L138-L140

The old version always set the header if the body was non-empty. This version is an improvement, setting the header only if not already set.

I wasn't able to run this locally, but I looked at the CI output and the new log line appears in tests. I hope a chattier proxy doesn't cause ops issues, and that there is enough context to identify the clients that need updating.

djmitche:
> I wasn't able to run this locally, but I looked at the CI output and the new log line appears in tests. I hope a chattier proxy doesn't cause ops issues, and that there is enough context to identify the clients that need updating.

This is a good question.  Workers already log *voluminously*, so this will be pretty tiny in comparison.

djmitche:
I'll fix the capitalization issues and land.

PR #3686, opened 2020-10-19
Remove support for running generic-worker in clouds without worker-runner

### LINKED ISSUE #3015: Remove worker-manager support from generic-worker
It should rely on worker-runner to talk to worker-manager.

djmitche:
Github Bug/Issue: Fixes #3015.


djmitche:
woo, all green!

imbstack (approved):
this is genuinely exciting.

This review seemed pretty straightforward and you went about this in a methodical way so I don't have any worries here

djmitche:
I think this has hewed closely to what we, including @petemoore, agreed when #3015 was filed.  If there's anything you'd like to change here, @petemoore, I'm happy to make a followup when you're back.

Issue #3670, opened 2020-10-19
dev-init needs fixing after landing node-pg 8, not using azure


djmitche:
`infrastructure/tooling/src/dev/util.js` adds `?ssl=1` to the end of the URL.  As of node-pg 8, that tries to verify the certs, but for cloud SQL that doesn't make sense.

Also `infrastructure/tooling/src/dev/azure.js` adds several azure storage account keys, when in reality those aren't needed and `yarn dev:apply` will complain that they exist.  Probably most of that file could be deleted (except the encryption key parts).

Finally, `yarn dev:init` seems to create top-level forceSSL, etc. parameters as objects, which are not valid by the schema.

PR #3703, opened 2020-10-20
Remove SSL=1

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/3670.


djmitche (commented):
I think we actually want to keep the ssl, just configure it to not try to verify the certs.  Funny enough, I just wrote a commit for this 10 minutes ago -- https://github.com/taskcluster/taskcluster/commit/e2d8e37c9d02c556241442b89d4ce45300ba69e3

PR #3691, opened 2020-10-20
Add assume:anonymous to web-server scopes

helfi92:
A follow up on https://github.com/taskcluster/taskcluster/commit/e895ea7eaa91e329521dc870e2068b7eb094d205.

PR #3616, opened 2020-10-05
Add anonymous scope in web-server

helfi92:
Github Bug/Issue: Fixes #3591

PR #3704, opened 2020-10-20
Add anonymous scopes to services

helfi92:
Relates to https://github.com/taskcluster/taskcluster/pull/3615.


services/github/scopes.yml
```diff
@@ -3,3 +3,4 @@
 - queue:route:statuses
 - queue:route:checks
 - assume:scheduler-id:taskcluster-github/*
+- assume:anonymous
```
jwhitlock, line 6: Nit: git will complain without a newline at the end of the file, here and other places.

jwhitlock (approved):
Thanks @helfi92. This looks good. I don't understand the significance of adding this scope, so maybe @imbstack should +1 before merging.

imbstack (commented):
In the context of [this line](https://github.com/taskcluster/taskcluster/pull/3615/files#diff-af49473981bf7d5a1e2b6650234a5dc66582eba373c8dca4cbcc082064f45fbcR245) why do we need to add this? Do the static clients not get that scope added?

djmitche (changes requested):
I don't think this is necessary, based onhttps://github.com/taskcluster/taskcluster/pull/3615#issuecomment-715636601.

I suspect this PR was created because of #3691, with the idea that if one service needs this scope, they all do.

I think the only reason that #3691 was required was that #3616 landed before the rest of the RFC#165 work in #3615.  In fact, I think that both PRs become redundant after #3615 lands and can be reverted.  If #3615 is correct, then we *never* need to specify `assume:anonymous` anywhere, because it's implicitly included everywhere (including in `auth.currentScopes` which the frontend uses for the Profile view).

So to be clear, my thinking is:
 - close this without landing it
 - include a revert of #3616 and #3691 in #3615.

djmitche:
--> #3801.

PR #3705, opened 2020-10-20
Fix up `yarn dev:init`

### LINKED ISSUE #3670: dev-init needs fixing after landing node-pg 8, not using azure

djmitche:
Github Bug/Issue: Fixes #3670


helfi92 (approved):
LGTM

PR #3749, opened 2020-10-22
Support online DB migrations

### LINKED ISSUE #3733: Allow DB migrations to have multiple transactions, repeat until done
Breaking off from https://github.com/taskcluster/taskcluster/issues/3139#issuecomment-713823241:
---
> For 2: Incorporate online migration into the `db:migrate` process, such that any `db:migrate` run that introduces a need for an online migration goes on to execute it immediately. From a deployer's perspective, this means no changes to existing processes. Looked at from the TC developer's perspective and squinting a bit, this basically means that a migration is no longer limited to a single database transaction.

djmitche:
Github Bug/Issue: Fixes #3733


djmitche:
@alopezz, @imbstack I'd love a bit of feedback on this.  Don't take the body of the functions too seriously (the syntax isn't quite right), but hopefully the README and the example are illustrative of how this would work.  Can I get a 10% review that this is conceptually on-track?

alopezz:
For what it's worth, the approach looks reasonable to me and should be able to avoid outages. It does make writing these migrations harder, but I'm not in a position to suggest any better.

djmitche:
That's true, but hopefully "only as hard as it needs to be and no harder" :)

djmitche:
I've added a bit of WIP on `dbVersionTest` that can run a few scenarios of migrations and downgrades automatically.  This is similar to the utility we had for the entities functions, in that it creates a suite of tests for you with a single call.

For the WIP, I modified 0050 (modifying cache_purges to have `worker_pool_id` instead of `provisioner_id`/`worker_type`) to use online migration and downgrade.  I don't intend to land that, but it was helpful to have something to which to apply `dbVersionTest`!  I'll also slim that down to an example for the tc-lib-postgres docs.

libraries/postgres/src/util.js
```diff
@@ -113,3 +113,59 @@ exports.paginatedIterator = ({ fetch, size = 1000 }) => {
...
+    if (this.history.length < 2) {
+      return NaN;
+    }
+
+    const [first, last] = [this.history[0], this.history[this.history.length - 1]];
+    return (last[1] - first[1]) / (last[0] - first[0]);
+  }
+
+  // return the time the value will reach the end, or undefined if this cannot
+  // be calculated.
```
djmitche, line 153: This is currently unused, but at some point we will probably want to add this.  I didn't add it here because it requires that an online migration be able to quickly determine how much work remains, and that's not always possible without a table scan.  If that's possible, we can add this functionality in a later PR.  I'm sure it would make operators happy to know if they have time to grab lunch before the migration completes!

imbstack (approved):
I'm convinced! It's nice to have this machinery in place.

db/test/helper.js
```diff
@@ -223,3 +231,216 @@ exports.assertNoTableColumn = async (table, column) => {
...
+        });
+        await withCheckpoints('down', {
+          start: async () => await concurrentCheck(client),
+          preOnline: async () => await concurrentCheck(client),
+          midOnline: async () => await concurrentCheck(client),
+          done: async () => await concurrentCheck(client),
+        });
+        await startCheck(client);
+        assert(sawMigrationBatches, 'did not see multiple batches of online migration');
+        assert(sawDowngradeBatches, 'did not see multiple batches of online downgrade');
```
alopezz, line 398: @djmitche I'm working on the online migration for the tasks table for #3578, and trying to use this helper for the migration tests, but I'm not sure I understand this check. Since I'm trying to test only the online upgrade part, I set

```js
helper.dbVersionTest({
    version: THIS_VERSION,
    onlineMigration: true,
    onlineDowngrade: false,
...
```

But all these tests that assert unconditionally on both `sawMigrationBatches` and `sawDowngradeBatches` fail, and with reason I think, since I've not implemented an online downgrade for this db version.

I think I would expect these lines to be like lines 360-365:

```js
if (onlineMigration) {
    assert(sawMigrationBatches, 'did not see multiple batches of online migration');
}
if (onlineDowngrade) {
     assert(sawDowngradeBatches, 'did not see multiple batches of online downgrade');
}
```

What am I missing here?
(Also, I don't know if adding a comment here is an appropriate venue for this)

djmitche:
I think I'm the one that missed something there.  You're right, that's exactly what they should look like -- please mix that change into your PR :)

djmitche:
(and sorry for the delay -- US holidays and all that)

PR #3756, opened 2020-10-23
update all deps of client-web

djmitche:
`rm yarn.lock && yarn` -- just for client-web, which doesn't have a lot of dependencies and is pretty simple and well-tested.

This addresses the old minimist version, which was deeply buried in a karma dependency.  And it will hopefully save some renovate PRs.


petemoore (approved):
Seems reasonable, and tests are passing, so good enough for me.

PR #3770, opened 2020-10-25
Use environment markers for client-py

### LINKED ISSUE #3730: Python client does not list 'aiohttp' dependency
**Describe the bug**
The Python client depends on `aiohttp` but does not list it in its requirements:
https://github.com/taskcluster/taskcluster/blob/main/clients/client-py/setup.py#L26

This is causing errors in `mozci` where we depend on the Python client.

djmitche:
This uses environment markers to specify packages specific to python
versions, rather than conditionals in setup.py.  Suggested by @ahal.

Fixes #3730.

### DIFF
--- a/changelog/issue-3730.md
+++ b/changelog/issue-3730.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3730
+---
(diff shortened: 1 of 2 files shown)


imbstack (approved):
this is new to me!

djmitche:
Same, but I like it!

PR #3774, opened 2020-10-26
rename ExpansionPanel to Accordion

djmitche:
See #3731.


petemoore:
The change looks fine to me, but I'd prefer Hassan to review as I don't have a good knowledge of React components.

helfi92 (approved):

PR #3778, opened 2020-10-26
fix links from g-w installing doc

djmitche:
Thanks to @bhearsum for bringing this up.


petemoore (approved):
Thanks @bhearsum and @djmitche!

PR #3781, opened 2020-10-27
Issue 3779 - generic-worker: reasonable default for simple and docker engines tasksDir

### LINKED ISSUE #3779: generic-worker: On Mac, in simple mode, the default tasksDir is /Users
**Describe the bug**
The default `tasksDir` on MacOS in simple mode is `/Users`, which is inappropriate, since generic-worker won't create users in this case.

It also doesn't seem to attempt to create `tasksDir` if it does not exist.

**To Reproduce**

1. Download and configure generic-worker with enough information to run.
2. Notice that generic-worker creates `task_XXX` directories in `/Users`.
3. Set `tasksDir` to `"tasks"`.
4. Notice that generic-worker fails to start with a permissions error.
5. `mkdir tasks`.
6. Notice that generic-worker now runs as expected.

**Expected behavior**
generic-worker should have reasonable defaults (particularly defaulting to `/Users` in simple mode could be very surprising).

**Taskcluster version**
v37.3.0

- OS: MacOS 10.15.7

petemoore:
Github Bug/Issue: Fixes #3779 

Note, in the code it looks as if the directory is created when it does not exist. I wonder if the failure to create the directory in #3779 was due to the user not having write permission for the parent folder?

### DIFF
--- a/changelog/issue-3779.md
+++ b/changelog/issue-3779.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: major
+reference: issue 3779
+---
+Generic-worker simple/docker engine now have a default tasks directory of `tasks`, relative to the working directory. This is a breaking change from previous behaviour.
--- a/workers/generic-worker/simple_docker.go
+++ b/workers/generic-worker/simple_docker.go
@@ -97,9 +97,8 @@ func RenameCrossDevice(oldpath, newpath string) (err error) {
 }
 
 func defaultTasksDir() string {
-	// assume all user home directories are all in same folder, i.e. the parent
-	// folder of the current user's home folder
-	return filepath.Dir(os.Getenv("HOME"))
+	// Issue 3779; default tasks directory is `tasks` relative to working directory
+	return "tasks"
 }
 
 func rebootBetweenTasks() bool {


petemoore:
CC @ricky26

changelog/issue-3779.md
```diff
@@ -0,0 +1,4 @@
+audience: worker-deployers
+level: silent
+reference: issue 3779
+---
```
djmitche, line 4: Is it possible existing users are relying on this default?  Perhaps this should come with a note?

djmitche (approved):

workers/generic-worker/simple_docker.go
```diff
@@ -97,9 +97,8 @@ func RenameCrossDevice(oldpath, newpath string) (err error) {
 }
 
 func defaultTasksDir() string {
-	// assume all user home directories are all in same folder, i.e. the parent
-	// folder of the current user's home folder
-	return filepath.Dir(os.Getenv("HOME"))
+	// ~/tasks seems like a reasonable default for simple and docker engine
+	return filepath.Join(os.Getenv("HOME"), "tasks")
```
ricky26, line 101: My personal preference here would be to prefer just `"tasks"`, i.e. relative to the working directory. (This is how `cachesDir` defaults too.)

ricky26:
I had a quick look to see if the `TasksDir` is created. It looks like it isn't but the individual task directories are created with `MkdirAll`. I have a horrible feeling that this happened too late and that I still got permission issues without creating it manually.

(Is it possible that it tries to list the directory to check for old tasks?)

changelog/issue-3779.md
petemoore: Good point, I'll add one. Thanks!

workers/generic-worker/simple_docker.go
petemoore: That's a very good point - I'll change it as suggested. Thanks!

petemoore:
> I had a quick look to see if the `TasksDir` is created. It looks like it isn't but the individual task directories are created with `MkdirAll`. I have a horrible feeling that this happened too late and that I still got permission issues without creating it manually.
> 
> (Is it possible that it tries to list the directory to check for old tasks?)

Ahhh that sounds like it could be true indeed, I should be able to reproduce, add a test to catch the problem, and then fix it... Thanks Ricky!

djmitche:
> I had a quick look to see if the `TasksDir` is created. It looks like it isn't but the individual task directories are created with `MkdirAll`. I have a horrible feeling that this happened too late and that I still got permission issues without creating it manually.

If this is the case, we can address in a followup.

djmitche (approved):

ricky26 (approved):
:+1:

PR #3801, opened 2020-10-28
Revert web server anonymous scopes

djmitche:
As described in #3704, the anonymous scopes are implicitly included everywhere, so there's no need to do so explicitly.


imbstack (approved):

djmitche:
I'll generate and then land.

PR #3810, opened 2020-10-29
set up k8s/templates once, instead of for every service

### LINKED ISSUE #3738: race in yarn generate
Yarn generate sometimes fails with
```
[Error: EEXIST: file already exists, mkdir '/taskcluster/infrastructure/k8s/templates'] {
  errno: -17,
  code: 'EEXIST',
  syscall: 'mkdir',
  path: '/taskcluster/infrastructure/k8s/templates'
}
```

it looks like that's from multiple parallel mkdir's called at the same time?

djmitche:
Github Bug/Issue: Fixes #3738

### DIFF
--- a/changelog/issue-3738.md
+++ b/changelog/issue-3738.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3738
+---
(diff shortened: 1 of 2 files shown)


imbstack (approved):
oh, nice catch

PR #3825, opened 2020-10-30
Create an empty object service

### LINKED ISSUE #3713: Create a skeleton `services/object` with an empty API
This should be a few minutes work copy/pasting another service, followed by some futzing to get the new service listed in all the right places.  Don't forget the documentation!

djmitche:
Look, new service!

I think this will be a major upgrade, since it will break deployment, so perhaps we should land this on a branch for most of the first week of November, in case there are other changes we want to make in v38.

Github Bug/Issue: Fixes #3713


djmitche:
looks like the py36 failure was an intermittent (a bunch of 502's from the community deployment)

petemoore (approved):
Looks good to me, nice job!

djmitche:
I'll merge it to the `object-service` branch.  We can develop against that branch until 38.x is deployed everywhere (hopefully this week)

PR #3835, opened 2020-11-02
update integration tests to expect anonymous role

djmitche:
Github Bug/Issue: Fixes #3834

PR #3841, opened 2020-11-03
Fix TestTcProxy test failure message

petemoore:
Test failure log message fix from #3835.


djmitche (approved):
Whoops, I just pushed a fix to add a word, sorry!  Yours is better :)

Issue #3831, opened 2020-11-02
The Log viewer URI decodes the artifact URL incorrectly if a bewit is specified


ricky26:
**Describe the bug**
If a bewit is specified for a log artifact, the log viewer incorrectly decodes the artifact URL multiple times, destroying the artifact name and resulting in a 401 message.

**To Reproduce**

1. Make sure that `queue:get-artifact:public/*` is not granted to anonymous.
2. Run a task and wait for it to complete.
3. Click on the backing artifact log link.
4. (Chrome only) Notice that the page loads correctly, press F5.
5. Notice the 401 error.

**Expected behavior**
The log should be displayed correctly.

**Taskcluster version**
v38.0.1

**Screenshots**
https://i.snipboard.io/n0GZL6.jpg

**Platform (please complete the following information):**
 - OS: Windows 10
 - Chrome: 86
 - Firefox: 82.0.2

**Additional context**
The error that yields the 401 is 'Bad mac' which is caused by the URL being decoded one time too many.

djmitche:
With #3844, I get

![image](https://user-images.githubusercontent.com/28673/98023295-b3198b00-1dd4-11eb-9e69-054a672f7180.png)

(awesome!) and in the console:

```
Unhandled GraphQL subscription error Error: GraphQL error: this.innerReconcileSubscriptions is not a function
    ApolloError bundle.esm.js:63
    makeObservable bundle.esm.js:1481
    next Observable.js:322
    notifySubscription Observable.js:135
    onNotify Observable.js:179
    next Observable.js:235
    next bundle.esm.js:866
    next bundle.esm.js:866
    notifySubscription Observable.js:135
    onNotify Observable.js:179
    next Observable.js:235
    opId client.js:163
    processReceivedData client.js:532
    onmessage client.js:485
    connect client.js:483
    executeOperation client.js:261
    subscribe client.js:150
    setContext bundle.esm.js:13
    promise callback*setContext/</< bundle.esm.js:12
    Subscription Observable.js:197
    subscribe Observable.js:279
    multiplex bundle.esm.js:864
    Subscription Observable.js:197
    subscribe Observable.js:279
    map Observable.js:319
    Subscription Observable.js:197
    subscribe Observable.js:279
    subscribeToMore bundle.esm.js:264
    obsSubscribeToMore react-hooks.esm.js:92
    TaskGroup index.jsx:238
    render index.jsx:636
    React 8
    unstable_runWithPriority scheduler.development.js:653
[...]

PR #3844, opened 2020-11-03
Fix "e.artifacts is null" and improve UI error handling

### LINKED ISSUE #3839: e.artifacts is null
**Describe the bug**
Viewing a task gives "e.artifacts is null"

**To Reproduce**

1. Go to https://community-tc.services.mozilla.com/tasks/bOh0SMjqT_uOGdJax_ZKkQ

**Taskcluster version**
38.0.1

Traceback suggests `getLivelogArtifactFromRun` is to blame.

djmitche:
Github Bug/Issue: Fixes #3839

This at least *changes* how #3831 behaves, so perhaps the fix for that issue should be based on this one.

### DIFF
--- a/ui/src/components/ErrorPanel/index.jsx
+++ b/ui/src/components/ErrorPanel/index.jsx
@@ -97,25 +97,57 @@ export default class ErrorPanel extends Component {
     } = this.props;
     const { error } = this.state;
     const hasWarning = Boolean(props.warning);
-    const errorMessage =
-      error && error.graphQLErrors && error.graphQLErrors[0]
-        ? error.graphQLErrors[0].message
-        : error;
+    const message = [];
+
+    if (!error) {
+      return null;
+    }
+
+    // handle GraphQLErrors as a special case
+    if (error.graphQLErrors && error.graphQLErrors.length > 0) {
+      const errors = error.graphQLErrors;
+
+      // Log extensions for all errors received
+      errors.forEach(err => {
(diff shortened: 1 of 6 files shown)


ui/src/components/ErrorPanel/index.jsx
```diff
@@ -97,25 +97,57 @@ export default class ErrorPanel extends Component {
...
+        }
+      });
+
+      // construct a markdown summary of all of the errors
+      if (errors.length > 1) {
+        message.push('Multiple errors from web-server:');
+
+        errors.forEach(err => {
+          message.push(`* ${err.message}`);
+        });
```
helfi92, line 124: In the past we would only show the first error to avoid having a very massive error panel show up when the user is missing a lot of scopes. Would it still make sense to only show the first error or perhaps put a limit on how many messages we can render from this array (e.g., only show 2/3 messages at most)?

helfi92 (commented):
Just one minor concern, otherwise this is awesome!

ui/src/components/ErrorPanel/index.jsx
djmitche: Good point.  I'll limit it to 3 with an "and N additional errors".

djmitche:
I'll merge if that's green when I get back from ☕ but feel free to click merge sooner than that :)

helfi92 (approved):

PR #3845, opened 2020-11-03
Set up PulseEngine before setting onConnected hook

djmitche:
This fixes an (apparently harmless?) error that appears in the UI when
viewing the TaskGroup page, as noted in #3831.


petemoore (approved):
Nice spot, that the `connected` method calls the `innerReconcileSubscriptions` method.

PR #3847, opened 2020-11-03
Update d-w and scope resolution

### LINKED ISSUE #3843: Task using indexed-image fails: "could not find task associated with namespace"
**Describe the bug**

A task created using an indexed-image fails with:

    [taskcluster:error] Pulling docker image has failed.
    [taskcluster:error] Error: Could not find a task associated with "project.fuzzing.config.master" namespace. Supplied credentials do not satisfy authorizedScopes; credentials have scopes:
    ```
    assume:worker-type:proj-fuzzing/linux-pool1
    assume:worker-pool:proj-fuzzing/linux-pool1
    assume:worker-id:us-east1/7625143841217902745
    queue:worker-id:us-east1/7625143841217902745
    secrets:get:worker-type:proj-fuzzing/linux-pool1
    secrets:get:worker-pool:proj-fuzzing/linux-pool1
    queue:claim-work:proj-fuzzing/linux-pool1
    worker-manager:remove-worker:proj-fuzzing/linux-pool1/us-east1/7625143841217902745
    worker-manager:reregister-worker:proj-fuzzing/linux-pool1/us-east1/7625143841217902745
    ```
    authorizedScopes are:
    ```
    docker-worker:capability:device:loopbackAudio
    docker-worker:capability:device:hostSharedMemory
    secrets:get:project/fuzzing/credstash-aws-auth
    queue:scheduler-id:-
    queue:cancel-task:-/*
    queue:create-task:highest:proj-fuzzing/linux-pool1
    secrets:get:project/fuzzing/decision
    ```
    ---
    * method:     findTask
    * errorCode:  AuthenticationFailed
    * statusCode: 401
    * time:       2020-11-03T12:02:23.268Z

**To Reproduce**

The tasks are created by a scheduled hook: https://community-tc.services.mozilla.com/hooks/project-fuzzing/linux-pool1

**Expected behavior**

Tasks should be able to find the public docker image specified. The namespace called out is indexed here: https://community-tc.services.mozilla.com/tasks/index/project.fuzzing.config/master and the image is accessible from the UI.

**Taskcluster version**
38.0.1
worker version 30.1.1

imbstack:
Basic summary:

1.  d-w was using its own creds instead of the task's creds to talk to index service for indexed images. That wasn't a problem before rfc 165 since it was public before.
2. We need to expand scopes _before_ we limit them with `authorizedScopes` because a client that is granted a scope that is in `authorizedScopes` by virtue of it being in `assume:anonymous` must have it at that time. Not doing this results in the error seen in the linked bug. I believe to keep the behavior we've decided for rfc 165 we actually have to apply `assume:anonymous` and expand scopes _again_ after limiting to `authorizedScopes` because otherwise anything limited in that fashion will not have those scopes. This is a bit of a more confusing situation however, happy to debate/change/etc.

Github Bug/Issue: Fixes #3843

### DIFF
--- a/services/auth/src/signaturevalidator.js
+++ b/services/auth/src/signaturevalidator.js
@@ -235,13 +235,21 @@ const createSignatureValidator = function(options) {
     ({ clientId, expires, accessToken, scopes } = await options.clientLoader(issuingClientId));
 
     // apply restrictions based on the ext field
+    // Implicitly grant all clients the anonymous role _before_ we check
+    // `authorizedScopes`. If we didn't do this first then the check that
+    // the client has a superset of `authorizedScopes` can fail if the scope
+    // is provided by it being in `anonymous`
     if (ext) {
+      scopes = utils.mergeScopeSets(scopes, ['assume:anonymous']);
+      scopes = options.expandScopes(scopes);
       ({ scopes, expires, accessToken } = limitClientWithExt(
         credentialName, issuingClientId, accessToken,
         scopes, expires, ext));
     }
 
-    // Implicitly grant all clients the anonymous role
+    // Implicitly grant all clients the anonymous role a second time.
+    // This is required to add it back in if it has been removed by
+    // it (and the scopes it grants) _not_ being included in authorizedScopes.
     scopes = utils.mergeScopeSets(scopes, ['assume:anonymous']);
     scopes = options.expandScopes(scopes);
 
--- a/workers/docker-worker/src/lib/docker/indexed_image.js
+++ b/workers/docker-worker/src/lib/docker/indexed_image.js
@@ -49,7 +49,7 @@ module.exports = class IndexedImage extends ArtifactImage {
 
     const index = new taskcluster.Index({
       rootUrl: this.runtime.rootUrl,
-      credentials: this.runtime.taskcluster,
+      credentials: this.task.claim.credentials,
       authorizedScopes: this.taskScopes,
     });
 
--- a/changelog/issue-3843.md
+++ b/changelog/issue-3843.md
@@ -1,0 +1,10 @@
+audience: general
+level: patch
+reference: issue 3843
+---
+Two bugs were fixed that together made it so that tasks could not use indexed images.
+
+First is that docker-worker now correctly uses the task's credentials rather than
+its own to query the index.
+Second is that scopes are now expanded prior to limiting them with `authorizedScopes`
+in addition to afterward.
--- a/services/auth/test/signaturevalidator_test.js
+++ b/services/auth/test/signaturevalidator_test.js
@@ -1,6 +1,6 @@
 const hawk = require('hawk');
 const _ = require('lodash');
-const assume = require('assume');
+const assert = require('assert');
 const slugid = require('slugid');
 const crypto = require('crypto');
 const taskcluster = require('taskcluster-client');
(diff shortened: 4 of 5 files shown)


imbstack:
This is still a draft but I wanted to see what your thoughts were!

workers/docker-worker/src/lib/docker/indexed_image.js
```diff
@@ -49,7 +49,7 @@ module.exports = class IndexedImage extends ArtifactImage {
 
     const index = new taskcluster.Index({
       rootUrl: this.runtime.rootUrl,
-      credentials: this.runtime.taskcluster,
+      credentials: this.task.claim.credentials,
```
petemoore, line 52: nice spot!

services/auth/src/signaturevalidator.js
```diff
@@ -234,14 +234,23 @@ const createSignatureValidator = function(options) {
     let accessToken, scopes, expires;
     ({ clientId, expires, accessToken, scopes } = await options.clientLoader(issuingClientId));
 
+    // Implicitly grant all clients the anonymous role _before_ we check
+    // `authorizedScopes`. If we didn't do this first then the check that
+    // the client has a superset of `authorizedScopes` can fail if the scope
+    // is provided by it being in `anonymous`
+    scopes = utils.mergeScopeSets(scopes, ['assume:anonymous']);
+    scopes = options.expandScopes(scopes);
+
```
petemoore, line 243: These new lines (237-243) can be moved inside the `if (ext) {` condition below, so that the merge/expand only happens twice if there is an `ext`.

An alternative solution would be for the `utils.satisfiesExpression` function to consider anonymous scopes _only if satisfaction fails without them_. This would probably have some performance benefit and hopefully isolate the scope satisfaction code to a single place, especially if e.g. workers are using the same validation libraries. However, this injects the concept of anonymous scopes at a pretty low level of the conceptual stack, and there is an argument that it is a higher level concern whether a request carries anonymous scopes or not, and shouldn't be considered at a lexical evaluation level. Let's discuss this afternoon.

petemoore:
Note, there were some test failures, e.g.:

```
=== RUN   Test_AuthorizedScopes
--- FAIL: Test_AuthorizedScopes (0.08s)
    creds_test.go:71: got unexpected scopes []string{"scope:1", "scope:3"}
=== RUN   Test_TempCredWithAuthorizedScopes
--- FAIL: Test_TempCredWithAuthorizedScopes (0.07s)
    creds_test.go:71: got unexpected scopes []string{"scope:1"}
=== RUN   Test_NamedTempCredWithAuthorizedScopes
--- FAIL: Test_NamedTempCredWithAuthorizedScopes (0.07s)
    creds_test.go:71: got unexpected scopes []string{"scope:1"}
```

services/auth/src/signaturevalidator.js
djmitche: I agree that we can avoid doing this twice.

This is already really difficult to reason about, and that's *with* a firm definition of "satisfaction" that does not involve role expansion. I think blurring those two things will make this even harder to reason about.

djmitche:
The test failures are integration tests that assume production is running 38.x or higher, which it's not right now :)

djmitche (approved):
Looks good!

helfi92 (approved):

helfi92:
Ah, some tests are failing

imbstack:
We confirmed that the failing tests are due to how community deployment is currently configured.

PR #3853, opened 2020-11-04
remove lerna.json

djmitche:
This seems to be distracting renovate, and now that we've imported everything I think we don't actually use lerna anymore.  Please yell if I'm wrong!

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,0 @@
-{
-  "version": "0.0.0",
-  "npmClient": "yarn",
-  "useWorkspaces": true
-}


helfi92 (approved):
LGTM. Tests are failing.

djmitche:
I think those are just due to having been on a rolled-back version of TC at the time.  I'll try re-running.

djmitche:
Oh, we haven't rolled forward yet.  Those are unrelated to this PR.  And, this doesn't need a changelog.

Issue #3717, opened 2020-10-20
[object service] define a db table for objects


djmitche:

djmitche:
I've been tossing around what this needs to look like, having lots sight of the "MVP" approach we're taking for now.

For the moment, working on a branch, it really doesn't matter.  Just build it :)

helfi92:
An `object_service` table with columns `name`:string , and `data`:jsonb is what you had in mind for the "MVP" approach?

djmitche:
That sounds good.  Probably `objects` is a good table name.

PR #3855, opened 2020-11-04
Define a db table for objects

helfi92:
Resolves https://github.com/taskcluster/taskcluster/issues/3717.


djmitche (approved):
Looks good if it passes the (non-currently-broken) tests :D

Issue #3712, opened 2020-10-20
[object service] implement an expiration cronjob


djmitche:

helfi92:
How do we decide which object to expire? Do we need an `expires` field in the `objects` table that we can read from?

djmitche:
Indeed, let's include that in the schema, and set it on upload.

PR #3859, opened 2020-11-04
Add expiration cronjob to object service

helfi92:
Github Bug/Issue: Fixes https://github.com/taskcluster/taskcluster/issues/3712


services/object/procs.yml
```diff
@@ -2,6 +2,11 @@ web:
   type: web
   command: node services/object/src/main.js server
   defaultReplicas: 0
+cleanup-expire-objects:
```
djmitche, line 5: Judging by the table in `monitoring.mdx`, we usually name this `expire`, so let's do the same here.

services/object/procs.yml
```diff
@@ -2,6 +2,11 @@ web:
...
+  type: cron
+  schedule: '0 0 * * *'
+  deadline: 86400
+  command: node services/object/src/main cleanup-expire-objects
```
djmitche, line 9: Is there an easy way to make sure this doesn't run on production deployments, similar to the `defaultReplicas: 0` for the web proc?

djmitche (approved):
Looks good!

services/object/procs.yml
helfi92: We can probably add an additional field `ignore` to ignore the cron command if `ignore` is set to true. Would you like me to add this?

PR #3860, opened 2020-11-04
Generate correct env:list helm schema

imbstack:

Github Bug/Issue: Fixes #XXXX


djmitche (approved):

djmitche (approved):

helfi92 (approved):

imbstack:
Actually now that I think about this, we need to update `ui/docs/reference/integrations/github/checks.mdx` for this (which I've done in the most recent revision on this. Does the new behavior make sense? I guess it is fine. It is just more permissive than before.

PR #3866, opened 2020-11-05
Fix Livelog TLS Support

### LINKED ISSUE #3865: Livelog is crashing on 38.0.3 images
Seems that we switched to `scratch` from `busybox` and livelog made some assumptions about directories existing, etc. Just making this bug to track fixing it.

imbstack:
- Use busybox for livelog image in order to get /tmp back
- Support TLS in livelog again

This feels like a sorta hacky way to make it work but the whole thing is a hack so whatever.

Github Bug/Issue: Fixes #3865

### DIFF
--- a/tools/livelog/main.go
+++ b/tools/livelog/main.go
@@ -168,7 +168,11 @@ func main() {
 
 	runServer = func(server *http.Server, addr, crtFile, keyFile string) error {
 		server.Addr = addr
-		return server.ListenAndServe()
+		if crtFile != "" && keyFile != "" {
+			return server.ListenAndServeTLS(crtFile, keyFile)
+		} else {
+			return server.ListenAndServe()
+		}
 	}
 
 	serve(putAddr, getAddr)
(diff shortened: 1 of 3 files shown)


tools/livelog/main.go
```diff
@@ -168,7 +168,11 @@ func main() {
 
 	runServer = func(server *http.Server, addr, crtFile, keyFile string) error {
 		server.Addr = addr
-		return server.ListenAndServe()
+		if crtFile != "" && keyFile != "" {
+			return server.ListenAndServeTLS(crtFile, keyFile)
+		} else {
+			return server.ListenAndServe()
+		}
```
djmitche, line 175: Looks like I blew this up in https://github.com/taskcluster/taskcluster/commit/fa6d273984 -- sorry about that!

I think we didn't notice until now because (a) generic-worker doesn't use the TLS support in livelog and (b) docker-worker was stuck on an ancient version.

djmitche (approved):

PR #3869, opened 2020-11-06
Rewrite browser-side auth to fetch credentials on-demand

### LINKED ISSUE #3836: ext.certificate.expiry < now
This used to happen occasionally if you loaded a page that required scopes right away after not having loaded anything for a while (so the credentials had expired).

Now everything requires credentials, so this happens a lot.

image

To reproduce:
* Visit a page such as the worker list
* Don't visit any pages for >15m (I think this is the timeout?)
* Visit a page such as the worker list

djmitche:
The previous solution was an attempt to renew credentials on a schedule,
synchronized across tabs.  It had some pretty bad race conditions, one
of which was between page-load and renewal when the existing creds had
expired.

This switches to a simpler model, where credentials are fetched if
necessary (if they're expired) before making a GraphQL query.

Github Bug/Issue: Fixes #3836

### DIFF
--- a/changelog/issue-3836.md
+++ b/changelog/issue-3836.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 3836
+---
+The web UI no longer fails with "ext.certificate.expiry < now".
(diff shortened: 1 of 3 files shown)


helfi92 (approved):
LGTM, thanks!

djmitche:
Hopefully this is enough to release a new 38.x and we can upgrade again :)

PR #3871, opened 2020-11-06
Work around a history bug in parsing URLs

### LINKED ISSUE #3831: The Log viewer URI decodes the artifact URL incorrectly if a bewit is specified

djmitche:
There are two commits here -- the first fixes my error in #3844, and the second fixes #3831.

I tested this fix on both Firefox and Chome.

Github Bug/Issue: Fixes #3831

### DIFF
--- a/changelog/issue-3831.md
+++ b/changelog/issue-3831.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 3831
+---
+This version fixes an issue introduced in v38.0.0 which would cause the log viewer to display 401 errors.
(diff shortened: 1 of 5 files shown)


helfi92 (approved):
💯

PR #3872, opened 2020-11-06
Add simple upload API

### LINKED ISSUE #3715: [object service] add a simple, temporary upload API
This is just a way to make an API call to upload some data.  We will make no promises about API compatibility, and this doesn't have to be especially featureful.  This is basically a way to get some data into a test deployment and exercise backends, etc.

helfi92:
Github Bug/Issue: Fixes #3715


services/object/test/helper.js
```diff
@@ -36,10 +36,16 @@ exports.withBackends = (mock, skipping) => {
...
-    exports.load.cfg('backendMap', []);
+    exports.load.cfg('backends', {
+      pub: { backendType: 'test' },
+    });
+    exports.load.cfg('backendMap', [
+      // not anchored, so can appear anywhere in the name
+      { backendId: 'pub', when: { name: { regexp: 'public/.*' } } },
+      // anchored to the beginning of the string
+      { backendId: 'pub', when: { name: { regexp: '^pubdocs/.*' } } },
+    ]);
```
djmitche, line 48: Let's use a simpler default that just uses a 'testBackend' backend of type 'test' for all objects.  Otherwise, it's easy to build tests that depend on the details of this default configuration, and then those tests fail if the config is changed.

db/versions/0055.yml
```diff
@@ -0,0 +1,27 @@
+version: 55
+description: add upload db fn to objects service
+methods:
+  upload_object:
```
djmitche, line 4: This doesn't actually *upload* anything.  Maybe `create_object` is a better name?

services/object/src/api.js
```diff
@@ -10,4 +11,25 @@ let builder = new APIBuilder({
...
+  category: 'Backend',
+  scopes: 'object:upload:<name>/<projectId>',
+  title: 'Upload backend data',
+  description: [
+    'Upload backend data.',
+  ].join('\n'),
+}, async function(req, res) {
+  let { name, projectId } = req.params;
+  const backend = this.backends.forUpload({ name, projectId });
+  let expires = fromNow('1 year');
```
djmitche, line 30: The API should probably allow this to be specified.  Perhaps this should be included in the request body?

services/object/src/api.js
```diff
@@ -10,4 +11,25 @@ let builder = new APIBuilder({
   context: ['cfg', 'db', 'backends'],
 });
 
+builder.declare({
+  method: 'post',
+  route: '/upload/:name/:projectId',
```
djmitche, line 16: This will preclude object names containing `/`, which I think we will want.

Maybe move the `projectId` into the request body?

services/object/src/api.js
```diff
@@ -10,4 +11,25 @@ let builder = new APIBuilder({
...
+  name: 'uploadObject',
+  input: undefined,
+  output: undefined,
```
djmitche, line 19: I think we'll want to return something that will tell the caller how to actually upload the data.  Since there are no actual backends yet, make something up, using https://docs.taskcluster.net/docs/reference/platform/queue/api#createArtifact for inspiration. :)

services/object/src/api.js
```diff
@@ -10,4 +11,25 @@ let builder = new APIBuilder({
...
+  scopes: 'object:upload:<name>/<projectId>',
+  title: 'Upload backend data',
+  description: [
+    'Upload backend data.',
+  ].join('\n'),
+}, async function(req, res) {
+  let { name, projectId } = req.params;
+  const backend = this.backends.forUpload({ name, projectId });
+  let expires = fromNow('1 year');
+  await this.db.fns.upload_object(name, { backendId: backend.backendId, name, projectId }, expires);
```
djmitche, line 31: I suspect that we'll want to be able to query by backendId and by projectId in order to calculate costs.  Let's make those top-level column on the table, rather than JSON properties.  And, the name is already a column, so there's no need to include it in `data`.

services/object/src/api.js
```diff
@@ -10,4 +11,25 @@ let builder = new APIBuilder({
...
 
+builder.declare({
+  method: 'post',
+  route: '/upload/:name/:projectId',
+  name: 'uploadObject',
+  input: undefined,
+  output: undefined,
+  stability: 'experimental',
+  category: 'Backend',
+  scopes: 'object:upload:<name>/<projectId>',
```
djmitche, line 22: This scope pattern would cause issues when we want to allow `*` patterns in the name, as the `*` would swallow up the projectId.

services/object/src/api.js
```diff
@@ -10,4 +11,25 @@ let builder = new APIBuilder({
...
 });
 
+builder.declare({
+  method: 'post',
+  route: '/upload/:name/:projectId',
+  name: 'uploadObject',
+  input: undefined,
+  output: undefined,
+  stability: 'experimental',
+  category: 'Backend',
```
djmitche, line 21:
```suggestion
  category: 'Upload',
```
rationale: we'll probably have 2-3 methods for uploading objects, and 2-3 methods for downloading.  A user of the API will probably only be concerned with one or the other of those.  So those will make nice categories.

djmitche (changes requested):
A few extra things to think about.  These don't necessarily need to be solved in this version since it's a temporary API, but are worth thinking about now as we "feel out" the design of the upload. Imagine we're writing an S3 backend, since queue does something pretty similar.

 * How do we know when an upload is complete (that is, when the data is in the cloud)?
 * How is the backend implementation involved in the upload process?


I think the MVP (finished later this month) should answer most of these questions.  Questions about content-type, content-encoding, hash validation, and so on definitely do not need to be solved in the MVP.  And, it's fine to bump some of these questions out into other issues that we can solve later in the sprint, if you'd prefer.

db/test/versions/0056_test.js
```diff
@@ -0,0 +1,6 @@
+const testing = require('taskcluster-lib-testing');
+
+suite(testing.suiteName(), function() {
+// this version only adds a method,
+// so there is nothing to test on upgrade
```
djmitche, line 5: It does add two columns :)

But, I think we will end up squashing all these DB revisions before merging to main anyway, so this is fine.

services/object/src/api.js
```diff
@@ -10,4 +10,27 @@ let builder = new APIBuilder({
...
 });
 
+builder.declare({
+  method: 'post',
+  route: '/upload/:name',
+  name: 'uploadObject',
+  input: 'upload-object-request.yml',
+  stability: 'experimental',
+  category: 'Upload',
+  scopes: 'object:upload:<name>',
```
djmitche, line 20: I think we will want to be able to limit uploads by projectId, so this should probably be
```suggestion
  scopes: 'object:upload:<projectId>:<name>',
```

djmitche (approved):
Looking good!

services/object/src/api.js
helfi92: Removing `projectId` from the route but adding it here throws:

```
  services/object/test/api_test.js
    ✓ ping
    1) should be able to upload
    2) "after each" hook for "should be able to upload"


  1 passing (2s)
  2 failing

  1) services/object/test/api_test.js
       should be able to upload:
     Internal Server Error, incidentId 38f16edb-f25e-4ce7-be7d-6e0a1c95be9d. Error (not shown in production):
```
Error: Deferred auth was never checked!
    at ServerResponse.res.reply (/Users/hassanali/Documents/Mozilla/projects/taskcluster/libraries/api/src/middleware/schema.js:72:15)
    at Proxy.<anonymous> (/Users/hassanali/Documents/Mozilla/projects/taskcluster/services/object/src/api.js:33:14)
```

djmitche: Yeah, we'd need to call `req.authenticate` in that case since projectId is in the request body.

PR #3875, opened 2020-11-09
Add get-task scope to notify service

### LINKED ISSUE #3874: Notify handler doesn't have enough scopes
**Describe the bug**
On a Taskcluster instance without an anonymous role, the notify handler service prints errors about not having sufficient scopes:

````
"Error: Client ID static/taskcluster/notify does not have sufficient scopes and is missing the following scopes:

```
queue:get-task:ZthZ3zgeTs6TDPyfWiV2Qw
```

This request requires the client to satisfy the following scope expression:

```
queue:get-task:ZthZ3zgeTs6TDPyfWiV2Qw
```

---

* method:     task
* errorCode:  InsufficientScopes
* statusCode: 403
* time:       2020-11-09T10:42:53.303Z
    at /app/clients/client/src/client.js:403:21
    at async Handler.onMessage (/app/services/notify/src/handler.js:96:16)
    at async PulseConsumer.handleMessage (/app/libraries/monitor/src/monitor.js:148:9)
    at async PulseConsumer._handleMessage (/app/libraries/pulse/src/consumer.js:320:5)
    at async /app/libraries/pulse/src/consumer.js:202:13"
````

**To Reproduce**

1. Create a new task, with a notify route, e.g.: `notify.slack-channel.CHANNEL.on-transition`.
2. Notice no notifications are sent.
3. Check the logs for notify-handler

**Expected behavior**
Notifications should be sent.

**Taskcluster version**
v38.0.4

**Additional context**
It looks like we just need to add `queue:get-task:*` to notify, I'll put in a PR.

ricky26:
Github Bug/Issue: Fixes #3874

This adds `queue:get-task:*` to the notify service, which lets the handler run without `get-task` being public. I've tested that this is the only required scope by adding just this to anonymous on our testing TC.

### DIFF
--- a/changelog/issue-3874.md
+++ b/changelog/issue-3874.md
@@ -1,0 +1,5 @@
+audience: general
+level: patch
+reference: issue 3874
+---
+The notify service now has enough scopes to handle notifications on Taskcluster instances without the anonymous role.
--- a/services/auth/src/static-scopes.json
+++ b/services/auth/src/static-scopes.json
@@ -36,6 +36,7 @@
   {
     "clientId": "static/taskcluster/notify",
     "scopes": [
+      "queue:get-task:*"
     ]
   },
   {
--- a/services/notify/scopes.yml
+++ b/services/notify/scopes.yml
@@ -1,0 +1,1 @@
+- queue:get-task:*


djmitche (approved):
Thanks!

PR #3879, opened 2020-11-09
Error message change of IP, NIC, and VM in azure.js

### LINKED ISSUE #3808: [azure provider] include failed SDK method in worker pool error log
There are errors that just say "Cannot parse the request" in the worker pool error log, with title "VM Creation Error".  That's not too helpful!

That title comes from a `catch` block in `services/worker-manager/src/providers/azure.js` that catches errors from a number of operations on IPs, NICs, and VMs.

It'd be nice if it was refactored so that there was some indication which of those failed.

SushantAA:
Github Bug/Issue: Fixes #3808 
I have separated the try-catch. Now IP, NIC, and Vm have their separate try-catch. the description of the error message is not complete.

### DIFF
--- a/changelog/issue-3808.md
+++ b/changelog/issue-3808.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3808
+---
--- a/services/worker-manager/src/providers/azure.js
+++ b/services/worker-manager/src/providers/azure.js
@@ -579,43 +579,32 @@ class AzureProvider extends Provider {
    * This function is expected to be called several times per worker as
    * resources are created.
    */
+
   async provisionResources({ worker, monitor }) {
+    
+    let ipErrorFlag = false;
+    let nicErrorFlag = false;
+
+    try {
       // IP
       let ipConfig = {
         location: worker.providerData.location,
         publicIPAllocationMethod: 'Dynamic',
       };
+      worker = await this.provisionResource({
+        worker,
+        client: this.networkClient.publicIPAddresses,
+        resourceType: 'ip',
+        resourceConfig: ipConfig,
+        modifyFn: () => {},
+        monitor,
+      });
 
-      try{
-        worker = await this.provisionResource({
-          worker,
-          client: this.networkClient.publicIPAddresses,
-          resourceType: 'ip',
-          resourceConfig: ipConfig,
-          modifyFn: () => {},
-          monitor,
-        });
-      } catch (err) {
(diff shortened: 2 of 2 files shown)


djmitche (changes requested):
This is looking good.  The test failures were not your fault -- we had to roll something back and it caused those client tests to fail.  Sorry about that!

There are three things to fix here:
 1. The 'description' is now a fixed string, where it used to contain `err.message`.  So while this now gives the context (IP, VM, or NIC), it loses the actual error!  I think we can include both.  I'd suggest adjusting the title ("VM Creation Error", "NIC Creation Error", etc.) and leaving the description equal to `err.message`.
 1. Previously, when an error occurred, the function would report the error and then exit.  With this change, it reports the error and then tries to create the next resource type.  For example, if provisionResources fails for the IP, it will still try to create the NIC.  Let's fix that up :)
 1. This duplicates the error-handling code three times.  How could that be avoided?

SushantAA:
I reduced duplicates by using two bool variables and changed the title

djmitche (commented):
Oh, I like this approach! A single exception handler and tracking of what state things were in when the exception was thrown.  I hadn't thought of that!

Could this be simplified a little more by putting the error title in a single variable, and just changing that variable's content before each call to `provisionResource`?  That would remove the need for the if/else-if/else block to translate the booleans into a string, by just starting with that string.

djmitche:
Ah, also once you get `yarn` stuff running:
```
yarn changelog --silent
yarn lint --fix
```
will fix up the changelog and lint failures.

SushantAA:
I have done changes. I am not able to install go till now so I haven’t done yarn generate and other yarn stuff . I think It will take time to figure it out. If u have time can you do yarn generate and other yarn stuff  pls ?

djmitche (approved):
Nice work -- I think the iterations on this issue ended up at a pretty elegant fix!

I'll run the `yarn` stuff, push it here, and merge once the tests are green.

djmitche:
What kind of thing would you like to work on next?

djmitche:
https://dustin.taskcluster-dev.net/provisioners is running this patch and looks 👌🏽

SushantAA:
I am comfortable in Javascript and Python, want to solve bigger bugs. Anything other than that (like - go, shell) I need to learn basic first.

djmitche:
How about https://bugzilla.mozilla.org/show_bug.cgi?id=1564512?  I will move that to an issue shortly.

Issue #3716, opened 2020-10-20
[object service] build a download API with future-expandable "approaches" and a temporary approach


djmitche:
Let's build a way of downloading files that is future-expandable.  Basically, the client sends the approaches it knows how to use, and the server responds with the best of those approaches that it also supports, or an error.

This is for the "full" download API, so the response is a data structure with enough information for the caller to reliably get the data, rather than a redirect or anything like that.

Implement an initial, temporary approach that gets the job done, but which we don't plan to keep compatibility with for very long.

helfi92:
I had a few questions:

> Basically, the client sends the approaches it knows how to use

Do you have examples of approaches?

> the server responds with the best of those approaches that it also supports

How is "best" determined?

> the response is a data structure with enough information for the caller to reliably get the data

Do you have an example of "enough information"? Would that just be a `url` to send the request to download?

Thanks!

djmitche:
Those are good questions so I don't have precise answers to them.  But I can try to at least fill out the ideas :)

> Do you have examples of approaches?

Approaches might be:
 * A single URL to which a GET request should be made
 * A URL that supports Range requests
 * A bittorrent-style download seed so workers can share data with each other (crazy!)

The general idea is that we want to be future-compatible with interesting new ideas for how to perform a download.  We would probably always have a "preferred" approach, maybe one or two deprecated approaches, and maybe some interesting optional approaches like bittorrent.  So something like scriptworker would probably stick to the preferred approach, while the fanciest workers would support both the preferred approach and the bittorrent approach.

From the negative perspective, the general idea is to avoid locking ourselves into an approach that doesn't support some kind of functionality we would like.

> How is "best" determined?

To be determined.  I would guess that the services define an order of preferences, and just take the most-preferred of the options that both the API client and the object's backend support.  For example, bittorrent > url-with-ranges > single-url.  But not all backends support bittorrent.  So a download request where the client supports [bittorrent, url-with-ranges, single-url] and the backend supports [url-with-ranges, single-url] would choose url-with-ranges.

> Do you have an example of "enough information"? Would that just be a url to send the request to download?

It would depend on the approach.  It would include a URL or a bittorrent seed or whatever information is necessary to get the data, plus metadata like hashes, size, and compression.

PR #3886, opened 2020-11-10
Add full download API

helfi92:
Github Bug/Issue: Fixes https://github.com/taskcluster/taskcluster/issues/3716.

Pair programmed with @petemoore.


services/object/schemas/v1/download-object-request.yml
```diff
@@ -0,0 +1,17 @@
+$schema: "/schemas/common/metaschema.json#"
+title:              "Download Object Request"
+description: |
+  Download object request.
+type:               object
+properties:
+  acceptProtocols:
+    type: array
+    title: "Supported protocols"
```
djmitche, line 9: Let's get some additional documentation in here about what "protocol" means in this context (I think it's what I was calling "approach", right?  I agree that "approach" isn't a great word for it).

Rather than write several paragraphs in the schema, it might make sense to put this in a document under `ui/docs/reference` (e.g., "Download Protocols") and just refer to that document here.

services/object/schemas/v1/get-object-http.yml
```diff
@@ -0,0 +1,22 @@
+$schema: "/schemas/common/metaschema.json#"
+title: "Download Object Request"
+description: |
+  Download object request.
```
djmitche, line 4: This is a response, right?

services/object/src/backends/test.js
```diff
@@ -4,6 +4,14 @@ const { Backend } = require('./base');
  * The test backend type is only available when running tests.
  */
 class TestBackend extends Backend {
+  getDetails(name) {
```
djmitche, line 7: This method name seems pretty vague.  And, it should be defined in the `Backend` class with an explanation of its meaning.  Also, it looks like the test implementation here has a different set of arguments (just name) to the way it's called in `api.js`.

djmitche (dismissed):
I think the code here is in pretty good shape -- it just needs some docs and comments to give structure to what's implicit in the code.

services/object/schemas/v1/download-object-request.yml
```diff
@@ -0,0 +1,17 @@
+$schema: "/schemas/common/metaschema.json#"
+title:              "Download Object Request"
+description: |
+  Download object request. See [Transport Protocols](https://docs.taskcluster.net/docs/reference/platform/object/transport-protocols) for more detail.
+type:               object
+properties:
+  acceptProtocols:
+    type: array
+    title: "Supported protocols"
+    uniqueItems: true
```
imbstack, line 10: while it's true this should be unique, do we want to reject the request if it isn't? Maybe better to accept whatever and uniq them in the service? I'm happy either way.

services/object/src/api.js
```diff
@@ -33,4 +33,33 @@ builder.declare({
   return res.reply({});
 });
 
+builder.declare({
+  method: 'put',
+  route: '/download/:name',
```
imbstack, line 38: Maybe what is making me uncomfortable about object-service is that this is sorta pretending to be a download endpoint for the object when really it is just fetching metadata about how to download it. Even the title and description more accurately say what's going on here. What if we just called this `queryData`, `fetchData`, `objectMetadata`, etc? Thoughts?

imbstack (commented):
making sense so far

ui/docs/reference/platform/object/transport-protocols.mdx
```diff
@@ -0,0 +1,29 @@
...
+The Object Service does not directly upload and download objects, but acts as
+an intermediary on behalf of other data storage service providers. To upload an
+object, clients must make a request to the Object Service describing the object
+to upload (`object.uploadObject`), and the Object Service responds with details
+for the client to perform the upload directly.
+
+Similarly, to download an object (`object.downloadObject`), a request must be
+made to the Object Service describing the object, and the Object Server
+responds with information for the client to download the object directly
+(typically a URL).
```
jwhitlock, line 19: This description doesn't seem to match what is built. I would expect something like:

To download an object (`object.downloadObject`), a request must be made to the Object Service with the object name and an ordered list of acceptable protocols (see below), and the Object Server responds with the server-selected protocol, along with the details (typically a URL) for the client to download the object directly.

services/object/src/api.js
jwhitlock: `:name` implies that the callers are picking the name of the download - is that true? I'd expect something like `:id` for a server-generated ID or a UUID.

I'm also uncomfortable, but I think that's because I don't have a clear idea of what the object service does in response to these requests. Is it just picking the "best" option from requested protocols? In that case, why would the best be any other than the first choice? Doesn't the client have to do some work to pick the request order? If the object server is just picking a protocol, then it seems we could reverse the logic, and the endpoint could return the menu of protocols, and the client could pick the one it likes.

But, I expect there is some work done to prepare a download via protocol. Some possible work:

* An authorized client could be asking for a public URL of a previously private download. The object server copies the file to a public bucket, or updates the metadata
* A client could ask for the gzipped version of a file. The object server uploads a gzipped version to a bucket and returns the URL
* A client asks for a URL to a private resource with baked-in credentials. The object server generates the URL with credentials in the query string
* A client asks for compressed content, with a list of compression formats. The server picks one that is ready to go, rather than re-compresses for the first requested format.

If this is the kind of work the object server does, then it makes more sense to me that the server picks. It could still be accomplished with a menu of protocol options, but it goes from two to three requests:

1. The client requests the metadata, maybe with a caching header for the last time the client asked. The server responds with the menu of protocol options, with data like "I can gzip this, but I haven't yet"
2. The client requests the selected protocol, with any protocol-specific options. The server does some work, and responds with the connection details, such as a URL
3. The client requests the download

But, that's probably a different object service. I'm happy to see where this one goes, and what use cases it supports.

jwhitlock (commented):
Thanks @helfi92 and @petemoore! This is interesting work, and close to merging. It is making me ask a bunch of questions, so it is a successful prototype. I think a working prototype is more important than getting all the questions answered now.

I think @djmitche's documentation suggestion was solid, and it should sound plausible before we continue. The description of how to request download details didn't seem to line up with what I was seeing in the API. I'd appreciate another round of edits. 

Other comments, on the schema and code, should be taken as discussion questions or comments, and not requested changes.

services/object/src/api.js
djmitche: Who decides the name is a good question, that we haven't discussed (at least not recently).  I've been assuming that the client would choose the name, imagining names like `task/<taskId>/<runId>/<artifactName>`.  That allows pattern-matching on names in a useful in the `backendMap` configuration, for example putting `.*\.log$` in a reduced-redundancy bucket.  And it means that simple download URLs would "make sense" to the user, containing the taskId, runId, and artifact name that they expected.  It also helps with analysis and debugging.  And, it also leaves a bit of space for other (very speculative!) names like `worker-cache/<workerPoolId>/<cacheName>` or, for supporting mirroring, `replica/<replicaId>/<objectName>`.

Regarding the process for negotiation, for downloads I see this negotiation as a way to "decouple" download clients and servers, allowing us to introduce new functionality on the server with high confidence that the change won't break old clients because the old code (server-side and client-side) is maintained unchanged.  In that sense, "best" might just be "most recently implemented".  This can also support "weird" approaches -- the example we've been using is [bittorrent](https://docs.aws.amazon.com/AmazonS3/latest/dev/S3Torrent.html) -- that aren't supported by all backends.  

So hypothetically, imagine there's an `old-way` and `new-way`, as well as weird `bittorrent` and `blockchain` (it can do anything!) protocols.  The client makes a request for object `foo` with protocols `old-way`, `new-way`, `newer-way`, and `bittorrent`.  The object `foo` has an Azure backend, which doesn't support `bittorrent`, and the server doesn't know about `newer-way`, so the remaining possibilities are `old-way` and `new-way`.  The server selects `new-way` as it's the best of those two.  It calls the backend requesting protocol `new-way` and things proceed from there.

I think the key difference between the situations we've described is that you're seeing a lot more details involved in the negotiation (compression formats, cloud region, authorization, URL expiration time).  In my conception, those are parameters to the approach that are only consulted after the server chooses the (named) approach.

services/object/schemas/v1/download-object-request.yml
helfi92: Sounds good. We can rename to download methods.

services/object/schemas/v1/download-object-request.yml
helfi92: We discussed and prefer to enforce uniqueness via the schema rather than add the complexity in the application code and possibly help implementers catch bugs earlier.

ui/docs/reference/platform/object/transport-protocols.mdx
helfi92: I think this description is good but is very specific to the current implementation choice. Perhaps this would be good addition to the schema docs. The purpose of this page is to give a very high level view but the specifics of the data exchange is defined at a schema reference level. Another advantage of including only our high level summary here is that as the API evolves this high level description will still apply.

services/object/src/api.js
petemoore: > `:name` implies that the callers are picking the name of the download - is that true? I'd expect something like `:id` for a server-generated ID or a UUID.

Yes, in the current design, the caller chooses the name.

> I'm also uncomfortable, but I think that's because I don't have a clear idea of what the object service does in response to these requests. Is it just picking the "best" option from requested protocols? In that case, why would the best be any other than the first choice? Doesn't the client have to do some work to pick the request order? If the object server is just picking a protocol, then it seems we could reverse the logic, and the endpoint could return the menu of protocols, and the client could pick the one it likes.
> 
> But, I expect there is some work done to prepare a download via protocol. Some possible work:
> 
>     * An authorized client could be asking for a public URL of a previously private download. The object server copies the file to a public bucket, or updates the metadata
> 
>     * A client could ask for the gzipped version of a file. The object server uploads a gzipped version to a bucket and returns the URL
> 
>     * A client asks for a URL to a private resource with baked-in credentials. The object server generates the URL with credentials in the query string
> 
>     * A client asks for compressed content, with a list of compression formats. The server picks one that is ready to go, rather than re-compresses for the first requested format.
> 
> 
> If this is the kind of work the object server does, then it makes more sense to me that the server picks. It could still be accomplished with a menu of protocol options, but it goes from two to three requests:
> 
>     1. The client requests the metadata, maybe with a caching header for the last time the client asked. The server responds with the menu of protocol options, with data like "I can gzip this, but I haven't yet"
> 
>     2. The client requests the selected protocol, with any protocol-specific options. The server does some work, and responds with the connection details, such as a URL
> 
>     3. The client requests the download
> 
> 
> But, that's probably a different object service. I'm happy to see where this one goes, and what use cases it supports.

Regarding all of these ideas, I think they would be good enhancements for a future version of the object service.

Regarding compression negotiation and signed URLs, we could see it being added in future, as additional fields in the request body of this API method. Although compression could sometimes be negotiated in the download request directly, this may not always be possible, if the backing provider doesn't support it (e.g. we haven't found a way for S3 artifacts to be served supporting multiple content encodings) - so indeed the Object Service could handle this by creating additional S3 objects on-demand in this API call, if a new content encoding is required. Similarly, we could add an expiration property for signed URLs if the backing provider supports it.

services/object/schemas/v1/download-object-request.yml
```diff
@@ -0,0 +1,17 @@
+$schema: "/schemas/common/metaschema.json#"
+title:              "Download Object Request"
+description: |
+  Download object request. See [Transport Protocols](https://docs.taskcluster.net/docs/reference/platform/object/transport-protocols) for more detail.
+type:               object
+properties:
+  acceptProtocols:
+    type: array
```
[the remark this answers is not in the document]
petemoore: It is a good question whether the download method should also carry parameters. Currently indeed only a string is assumed, with the name of the download method to use, rather than an object with parameters.

So long as `acceptDownloadMethods` is an `enum`, in future json objects could be added, so that parameterised download methods could be supported, like in your example, but for the first revision, this probably isn't needed, and I think since this is controlled at a schema level, it would be possible to enhance this in future if a new download method takes parameters.

ui/docs/reference/platform/object/upload-download-methods.mdx
```diff
@@ -0,0 +1,29 @@
...
+Similarly, to download an object (`object.downloadObject`), a request must be
+made to the Object Service describing the object, and the Object Server responds
+with information for the client to download the object directly (typically a
+URL).
+
+## Download Methods
+
+Currently the Object Service supports the single download method `HTTP:GET`. In
+order to support additional download methods in future, download requests from
+clients should include an ordered list of preferred download methods to use, in
```
djmitche, line 25: If the server is selecting the preferred method, then this order doesn't matter.

ui/docs/reference/platform/object/upload-download-methods.mdx
```diff
@@ -0,0 +1,29 @@
...
+the `acceptDownloadMethods` property of the request body. When additional
+download methods are available, the Object Service will select a preferred
+download method from the union of its supported download methods and the
```
djmitche, line 28:
```suggestion
download method from the intersection of its supported download methods and the
```

services/object/src/api.js
```diff
@@ -33,4 +33,33 @@ builder.declare({
...
+  name: 'downloadObject',
+  input: 'download-object-request.yml',
+  output: 'download-object-response.yml',
+  stability: 'experimental',
+  category: 'Download',
+  scopes: 'object:download:<name>',
+  title: 'Download object data',
+  description: [
+    'Download object data.',
```
djmitche, line 47: Let's add a link here to the documentation.

djmitche (approved):

PR #3891, opened 2020-11-10
Allow third-party clients to have assume:anonymous

### LINKED ISSUE #3884: Users' clients get disabled if they have `assume:anonymous`
I created a client with `taskcluster signin` without specifying any scopes, so it just made a copy of my scopes.  It was disabled overnight, and I bet that was because the client had `assume:anonymous` and the web-server crontask didn't take that into account.

djmitche:
Github Bug/Issue: Fixes #3884

### DIFF
--- a/changelog/issue-3884.md
+++ b/changelog/issue-3884.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 3884
+---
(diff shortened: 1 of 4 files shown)


petemoore (approved):
lgtm, thanks!

helfi92 (approved):

PR #3893, opened 2020-11-11
Dynamically load Azure signing certs for attested document

### LINKED ISSUE #3669: azure-provider should download intermediate certs from microsoft
From https://bugzilla.mozilla.org/show_bug.cgi?id=1607922
---
This bug concerns verifying worker identity based on the attested data signature that we use for registering workers. The signature is a PKCS#7 message, which is a bundle that has message, signature, and an embedded certificate. The embedded certificate is signed by an intermediate certificate, which is in turn signed by a trusted root certificate.

The initial way we're handling intermediate certs is to embed them alongside the code and loading them on startup. These certs expire in ~4 years, so we should instead download intermediate certs from Microsoft based on the authorityInfoAccess certificate extension in the embedded certificate.

RFCs 3280, 4325, 5280 have some helpful information on authorityInfoAccess.

jwhitlock:
This PR changes the validation of the Azure worker identity proof in the `worker-manager` service. The identify proof is the output of the [attested data API](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#attested-data), which includes details about the worker and is signed by the Azure platform. The attested data is a [PKCS#7](https://en.wikipedia.org/wiki/PKCS_7) message, signed by an included signing certificate. The signing certificate is issued by an intermediate certificate, which is in turn issued by a root certificate (or root CA).

Previously, the `worker-manager` checked that the message signer was issued by one of four published intermediate certificates (Microsoft IT TLS CA 1, 2, 4, and 5) issued by a single root CA (Baltimore CyberTrust Root CA). This root CA doesn't comply with one of the CA/Browser Forum Baseline requirements, so Azure is planning to expand to five more root CAs, and revoke the current intermediate certificates on February 15, 2021. See [Azure TLS certificate changes](https://docs.microsoft.com/en-us/azure/security/fundamentals/tls-certificate-changes) for details. They have not published a list of new known intermediate certificates, and may stop this practice.

Several changes are needed to prepare for these upstream changes:
* The trusted certificate store is populated from Node.js's bundled certificates (see [tls.rootCertificates](https://nodejs.org/api/tls.html#tls_tls_rootcertificates)). Node v12.19.0, the version we use in development, does not include two of the planned root CAs, so these are checked in and loaded as well. We considered only loading the 6 root CAs, but we'd like to avoid maintaining this list, and are trusting the policies of these trusted root CAs.
* The 4 "Microsoft IT TLS CA" continue to be added to the trusted certificate store. We now verify that they are issued by a root CA. As of this month (November 2020), they were used in the signing chain in the attested data for a newly launched Azure VM.
* If the attested data messages signing certificate is issued by an unknown intermediate certificate, then we download it, with a 5 second timeout. The certificate URL is stored in the [AuthorityAccessInfo extension](https://tools.ietf.org/html/rfc5280#section-4.2.2.1) (AIA) of the signing certificate. This is downloaded synchronously, and checked against the known trusted certificates. If this succeeds, the new certificate is added to the trusted certificate store, and a new monitor message ``registration-new-intermediate-cert`` is emitted with the details. If anything goes wrong, worker registration fails, with a generic error sent to the user and a monitor message ``registration-error-warning`` with the details.
* We now check that the signing cert is named `metadata.azure.com` or a subdomain. This is true for public cloud azure but not for "sovereign cloud" like Azure Government. See the [attested data API](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#attested-data) for details.

Some existing tests needed tuning for the new algorithm, and the new code paths are tested. There are some related documentation changes and other minor changes.

With all these changes, there should be no impact on `worker-manager` until Azure ships the certificate changes. If the new certificate uses RSA, then there will be a delay of up to 5 seconds (less than a second locally) as the certificate is downloaded for the first time. If the new certificate uses ECC, then registration will fail with a monitor message.

Future work is tracked in these issues:

* Issue #3922 - Check intermedia certificates for revocation. We know that the "Microsoft IT TLS CA" certs will be revoked in the future. This will strengthen the chain-of-trust process.
* Issue #3923 - Add support for ECC / ECDSA certificates. This will be needed to support the root certificate "Microsoft ECC Root Certificate Authority 2017". [node-forge](https://github.com/digitalbazaar/forge) does not include ECC, so it is unable to load the Microsoft ECC certificate, or 21 of the 138 certificates bundled with Node.js.
* Issue #3924 - Remove bundled root CAs after upgrading node. These are included in node v15, and may be backported to a node v12 release.
* Issue #3925 - Remove or repurpose legacy Microsoft intermediate certificates when revoked. These should be removed from the trusted certificate store, but may still be useful for testing.

Github Bug/Issue: Fixes #3669


djmitche (commented):
This implementation looks fine so far -- no particular comments there.

looking at https://github.com/taskcluster/taskcluster/blob/32720727a859e1d03f913d2f96cb9a9c01066572/services/worker-manager/src/providers/azure.js#L288-L292 has me hoping I'm missing something :slightly_smiling_face:

Imagine Mallory gets a signing certificate similar to MS's signing certificate, from some recognized CA.  They create a cert/key and sign the cert with the signing certificate.  Then they create an attested document containing a malicious vmID, a signature made with the key, and the cert.  And hand it to us.

1. ✔️ The embedded document was signed with the private key corresponding to the embedded public key
2. ✔️ The embedded public key has a proper certificate chain back to a trusted CA
3. ✔️ The embedded message contains the vmId that matches the worker making the request

So basically we've verified that someone signed the message and that the PKI infrastructure can tell us who that someone is -- but we don't actually check that it was Microsoft.

jwhitlock:
I agree, just verifying the PKI chain seems insufficient. Reading [Sample 2: Validating that the VM is running in Azure](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#sample-2-validating-that-the-vm-is-running-in-azure), it seems like the check for ``.metadata.azure.com`` would also be needed (and might be sufficient).

djmitche:
It's frustrating that the sample just prints stuff.  I wonder how many people have copy/pasted that and said "great, now we verify it!"

jwhitlock:
As I said in the issue, the thumbprint algorithm appears to be correct. Node v12 doesn't have the two MS certs, they were added in node v15. node-forge still can't load the ECC certificate, though.

djmitche:
It sounds sort of boring, but maybe the right option is just to set up a way for deployers to configure the allowed certificates, and continue to ship the current set of certificates.  Then if MS changes the certs out from under us, the fix is just to change that configuration, not necessarily wait for and deploy a new version of Taskcluster.  We could even write a short script that syncs from that page to the repo, just to make it easy to stay up-to-date.

djmitche:
Printing `crt.subject.attributes` in `registerWorker` and running the tests shows
```
[
  {
    type: '2.5.4.3',
    value: 'metadata.azure.com',
    valueTagClass: 19,
    name: 'commonName',
    shortName: 'CN'
  }
]
```

so, that's where we find the `metadata.azure.com` referenced in the Azure docs.

Going back to Mallory's exploit above, I don't think checking that value makes a difference.  Mallory can just create a certificate with that CN and sign it with their signing certificate.  Maybe getting a signing certificate[*] that's chained from a root CA involves a rigorous promise to behave responsibly, and it's OK to assume Mallory can't get such a signing certificate, just like we assume that nothing chained from the root CAs would sign a TLS certificate without some kind of validation?

[*] I'm using "signing certificate" here to mean the intermediate certificate that signs the leaf certificate that's included in the attested document.

jwhitlock:
I don't feel like I can answer your questions, which sound like security or operations questions. I think I'd need to dig into PCKS#7 signing, to see if your attack is possible. I also suspect that the current implementation is susceptible in other ways, like replaying with a different PCKS#7 message. I think I'd have to look at the AWS implementation of the similar code, to see what is acceptable over there.

I assumed a lot of this research went into writing #3669 and previous bugs. Should I stop work on this fix and switch to research mode?

djmitche:
I think the most open question there is actually a policy question, around who gets CA certificates and what limits are in place on them.

djmitche:
Following up: we investigated that question.  Long story short, we can depend on recognized roots and the intermediates they sign to not sign a malicious certificate for `metadata.azure.com`.  So just checking for that CN is sufficient.  This is exactly the same model as TLS for HTTPS: web client checks the hostname against the cert it receives from the web server and checks that the cert chain is valid (even if that involves intermediates).

jwhitlock:
The intermediate certificate download feature was a lot of code and a lot of tests. I've updated the PR description with a summary. I'm pretty happy with the existing test coverage, and most of the new code is covered as well. A couple of the tests download an intermediate cert, so that is tested as well.

My preference would be to continue loading the 4 Microsoft IT certificates, even though they could be downloaded on first reference. I get nervous whenever the network is involved. I could be persuaded to remove them for this PR. In either case, there are a few ``TODO`` lines to fix up.

I've got the `metadata.azure.com` check left, as well as a changelog entry. I hope to have it ready for review Monday morning.

jwhitlock:
I've finished up this PR, it is ready for review. I've added @ajvb, who reviewed the original Azure code, and may be able to quickly review these changes.

This feels like a shippable chunk of work, that will remove a lot of the risk of the Azure TLS changes. I've identified future work, some clean-up (remove some bundled certs after a Node update) and some that may be critical (add ECC support).

services/worker-manager/test/provider_azure_test.js
```diff
@@ -1087,6 +1180,154 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
+            __dirname, '../src/providers/azure-ca-certs/microsoft_it_tls_ca_4.pem');
+          const intermediateCertPem = fs.readFileSync(intermediateCertPath);
+          const intermediateCert = forge.pki.certificateFromPem(intermediateCertPem);
+          const deletedCert = provider.caStore.removeCertificate(intermediateCert);
+          assert(deletedCert);
+
+          // Download is not a binary certificate
+          provider.downloadBinaryPromise = (url) => {
+            return new Promise((resolve, reject) => {
+              resolve('<html><body><h1>Apache2 Default Page</h1></body></html>');
```
djmitche, line 1302: nice :)

djmitche (approved):
This looks great -- you've done a _very_ thorough job of it, including handling a bunch of annoying compatibility issues.  The tests are quite comprehensive, too.

I've left some line comments here but you can feel free to apply or ignore them at your option.

services/worker-manager/src/providers/azure.js
```diff
@@ -77,6 +166,70 @@ class AzureProvider extends Provider {
...
+        let chunks = [];
+        res.on('data', chunk => {
+          chunks.push(Buffer.from(chunk));
+        });
+        res.on('end', () => {
+          const body = Buffer.concat(chunks);
+          resolve(body.toString('binary'));
+        });
+      });
+      request.on('error', (error) => reject(error));
```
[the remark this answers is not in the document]
jwhitlock: Thanks, I can use all the syntax advice I can get!

services/worker-manager/src/providers/azure.js
```diff
@@ -329,6 +501,84 @@ class AzureProvider extends Provider {
...
+
+    // download the intermediate certificate if needed
+    let issuer = this.caStore.getIssuer(crt);
+    if (issuer === null) {
+      const authorityAccessInfo = getAuthorityAccessInfo(crt);
+      let method, location;
+      for (let i = 0; i < authorityAccessInfo.length; i++) {
+        method = authorityAccessInfo[i].method;
+        location = authorityAccessInfo[i].location;
+        if (method === 'CA Issuer' && location.startsWith('http:')) {
```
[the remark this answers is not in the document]
jwhitlock: I was surprised too, but it makes sense - if you a bootstrapping a certificate chain, you don't want to get into a situation where you need a different certificate to make the SSL connection to get a certificate...

[RFC 5280](https://tools.ietf.org/html/rfc5280#section-4.2.2.1) doesn't say HTTPS is forbidden, but does say 

> Conforming applications that support HTTP or FTP for accessing certificates MUST be able to accept individual DER encoded certificates and SHOULD be able to accept "certs-only" CMS messages.

and
 
> When the id-ad-caIssuers accessMethod is used, at least one instance SHOULD specify an accessLocation that is an HTTP [RFC2616] or LDAP [RFC4516] URI.

There are plenty of cases in the spec (FTP, LDAP, certs-only message instead of DER) that would be conforming but that we wouldn't parse. The fall-though logic will at least log the Authority Access Info so we can see what to add (or just download manually). Who knows, `https:` may show up some day...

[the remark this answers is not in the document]
djmitche: Let's hope we don't have to support LDAP :)

services/worker-manager/src/providers/azure.js
```diff
@@ -77,6 +166,70 @@ class AzureProvider extends Provider {
...
+    const issuer = this.caStore.getIssuer(cert);
+    if (issuer === null) {
+      throw Error(`Issuer "${dnToString(cert.issuer)}"` +
+                  ` for "${dnToString(cert.subject)}" is not a known Root CA`);
+    }
+    this.caStore.addCertificate(cert);
+  }
+
+  // Return Promise that will download a binary file
+  async downloadBinaryPromise(url) {
```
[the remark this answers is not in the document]
jwhitlock: I had a lot of trouble getting ``http.get`` working, and I found the documentation to be pretty bad. I wasn't aware of ``superagent``, I'm inclined to switch.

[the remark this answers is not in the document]
djmitche: Let's do it.  Events and streams are _not_ JS's strong points.

services/worker-manager/src/monitor.js
```diff
@@ -171,3 +171,21 @@ MonitorManager.register({
     error: 'Error message from cloud that triggered this',
   },
 });
+
+MonitorManager.register({
+  name: 'registrationNewIntermediateCertificate',
+  title: 'Registration of New Intermediate Certificate',
+  type: 'registration-new-intermediate-certificate',
+  version: 1,
+  level: 'warning',
```
imbstack, line 180: Is this a warning because we want to eventually take some action from it?

services/worker-manager/test/provider_azure_test.js
```diff
@@ -165,13 +179,10 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
       ]);
 
       // Calculate the thumbprint of a certificate
-      // From https://github.com/digitalbazaar/forge/issues/596
+      // Microsoft's thumbprint is OpenSSL's fingerprint in lowercase, no colons
```
imbstack, line 182: they always find a way to make it harder on us

services/worker-manager/src/providers/azure.js
```diff
@@ -500,6 +501,18 @@ class AzureProvider extends Provider {
       throw error();
     }
 
+    // verify the subject of the signing certificate
+    const signerCommonName = crt.subject.getField({ name: 'commonName' });
+    if (!(signerCommonName &&
+          (signerCommonName.value === 'metadata.azure.com' ||
+           signerCommonName.value.endsWith('.metadata.azure.com')))) {
+      this.monitor.log.registrationErrorWarning({
```
imbstack, line 509: The registered report here is nice but I think this might be easier later if we use `monitor.reportError`. That way all of this ends up in the sentry report. Either that or log like this _and_ attach `crt.subject` etc to the error?

imbstack (approved):
Exciting! This looks really good to me.

services/worker-manager/src/monitor.js
jwhitlock: Thanks, I wanted some feedback on this.

One way to interpret it is "WARNING: jwhitlock's code just got called for the first time in production!". Someone may want to look at the data, download the certificate, and make sure it was correct.

It is also possible that Azure will publish a new list of expected intermediate certificates, and we should just pre-load those, and save the network request in the future. In this case, the Warning would be "a new cert was published, go look at the documentation and see if we need a PR".

In both cases, "info" might be appropriate now, or in a few months.

services/worker-manager/src/providers/azure.js
jwhitlock: I'm not familiar enough with ``monitor`` to understand your suggestion. Can you suggest a code change?

I tried to follow the conventions I could see in this code:

* Log the error details with ``registrationErrorWarning``:
  - Use a constant string for the ``message``
  - Add details in ``error``
* Throw the generic ``error()``, initialized near the start of ``registerWorker``, to avoid leaking information to an attacker

jwhitlock: I found the `library/monitor` docs, I understand your comment now.

Maybe the pattern of "log a ``registerionErrorWarning``, throw a generic ``error``" should be moved to a function, which could also send the info to sentry with ``monitor.reportError``. That would change the existing code as well, but would remove a lot of repetition.

All of this code is either a trap for hackers, or a obstacle course for maintainers. Or maybe both! Depending on which way you see it, a log message and/or a Sentry report makes sense.

jwhitlock:
After some out-of-band discussion, we're going to merge this now. @ajvb, feel free to review when you can, and we'll address any feedback in a future PR.

PR #3898, opened 2020-11-11
move docker-worker files around to match other services

djmitche:
Credit to @imbstack for the idea.  This will help us feel at least a little more "at home" in docker-worker.

services/* and libraries/* have source code in a `src/` directory, and
the main entry point is `main.js`, so let's make docker-worker match
this.

This also adjusts the startup script created in `release.sh` to use
`main.js`.

Aside from changes to import paths, this contains no source-code
changes.


djmitche:
OK, *now* the tests pass

jwhitlock (approved):
Thanks @djmitche, looks like a straightforward file move. I'm trusting CI to pick up any missing path modifications.

djmitche:
Me too.  I finally grepped for `\.\./\.\.` and found the few stragglers :)

PR #3902, opened 2020-11-12
Issue 3901 - fix forbidden error when using actions with public artifact URL signing

### LINKED ISSUE #3901: actions.json cannot be fetched if public S3 artifacts are signed
**Describe the bug**
When viewing a task group or task where a `public/actions.json`, with signing enabled for public S3 artifacts, an orange banner saying 'Forbidden' is displayed.

**To Reproduce**

1. Enable public S3 artifact signing.
2. Run a decision task which uploads `public/actions.json`.
3. Navigate to the task view for the decision task.

**Expected behavior**
The actions should be fetched without error.

**Taskcluster version**
v38.0.4

**Additional context**
It looks like this might be related to the signature appended to the S3 URL in the case that public artifact signing is turned on. The signature contains some unsafe URL characters which end up being encoded.

ricky26:
Github Bug/Issue: Fixes #3901

This issue was caused by `node-fetch` passing headers to further redirects, including the eventual S3 signed URL. In particular Taskcluster's `fetch()` sets `Content-Type` by default, which is a header which must be included in the signature.

### DIFF
--- a/changelog/issue-3901.md
+++ b/changelog/issue-3901.md
@@ -1,0 +1,5 @@
+audience: general
+level: patch
+reference: issue 3901
+---
+Fixed a bug where signing public S3 artifacts would result in Forbidden errors on the task and task group views.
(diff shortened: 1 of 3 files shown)


ricky26:
Actually, I'm wondering whether the better fix here is to change `fetch()` to not set `Content-Type` for `GET` requests... since that's invalid anyway IIUC.

djmitche (commented):
That does seem like a better fix, and it seems this is the only place that `utils/fetch.js` is used anyway, so it should be low-impact.

djmitche (approved):

PR #3904, opened 2020-11-12
Expand scopes in between pulling them from cert and checking authorizedScopes

### LINKED ISSUE #3903: Using authorizedScopes with temporary credentials doesn't use expanded scopes
**Describe the bug**
When using temporary credentials with authorized scopes (for example, running actions), only the scopes in the certificate are considered, instead of the expansion of all those scopes.

For example, when trying to run an action, you can receive an error like:

````
Supplied credentials do not satisfy authorizedScopes; credentials have scopes:

```
assume:github-team:org/my-team
assume:github-org-admin:org
assume:login-identity:me
```

authorizedScopes are:

```
secrets:get:my-secret
```
````

Viewing my profile, clearly shows `secrets:*`.

**To Reproduce**

1. Run a decision task which produces an `actions.json` for running another task with some required scopes not part of your login identity, e.g. secrets:get:something
2. Run the task.
3. Try to run the action after the task.

**Expected behavior**
Authorized scopes should take into account all the expanded scopes of temporary credentials.

**Taskcluster version**
v38.0.4

**Additional context**
It looks like is is fallout from #3615, where the two `expandScopes` calls were removed from `limitClientWithExt`. It might still need to do the `expandScopes` after loading temporary credentials.

imbstack:
This is pretty much a return to how things
worked before rfc165 here. We now expand no matter what before checking
`authorizedScopes`.

Github Bug/Issue: Fixes #3903

### DIFF
--- a/changelog/issue-3903.md
+++ b/changelog/issue-3903.md
@@ -1,0 +1,6 @@
+audience: general
+level: patch
+reference: issue 3903
+---
+Scopes are now expanded in between using a certificate's scopes and checking `authorizedScopes`
+as well.
--- a/services/auth/src/signaturevalidator.js
+++ b/services/auth/src/signaturevalidator.js
@@ -30,7 +30,7 @@ const parseExt = function(ext) {
  * modified (otherwise it returns the original).
  */
 const limitClientWithExt = function(credentialName, issuingClientId, accessToken, scopes,
-  expires, ext) {
+  expires, ext, expandScopes) {
   let issuingScopes = scopes;
   let res = { scopes, expires, accessToken };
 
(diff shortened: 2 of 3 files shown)


djmitche (approved):
This looks fine.

#3903 specifically addresses temporary credentials *and* authorizedScopes.  The scopes passed to `limitClientWithExt` are already expanded, so in the [not temporary credentials] and authorizedScopes state, this expansion will be redundant.

But I think that's fine, and in case I'm wrong, it doesn't really hurt to be sure :)

helfi92 (approved):
LGTM, thank you!

PR #3913, opened 2020-11-13
docker-worker: Un-gzip images with gzip content-encoding

### LINKED ISSUE #3899: docker-worker fails to load image from gzip’ed artifact
https://community-tc.services.mozilla.com/tasks/KEHqmyN_SZCqh4rkKILD1g/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FKEHqmyN_SZCqh4rkKILD1g%2Fruns%2F0%2Fartifacts%2Fpublic%252Flogs%252Flive.log#L33

```
[taskcluster 2020-11-11 19:10:01.594Z] Downloading artifact "public/image.tar.lz4" from task ID: f5coRt3YTxuwrhaCY_LzEQ.
[…]
[taskcluster:error] Pulling docker image has failed.
[taskcluster:error] Error: Error loading docker image. Could not decompress image file.Exit Code: 44 Errors: Error 44 : Unrecognized header : file cannot be decoded
```

This image is built by another task, and rebuild on demand when the `Dockerfile` changes or when the previous artifact expires. Here it was recently rebuilt: https://community-tc.services.mozilla.com/tasks/f5coRt3YTxuwrhaCY_LzEQ#artifacts

Because of https://github.com/taskcluster/taskcluster/pull/2809 this artifact is served gzipped:

```
$ curl -sLI https://community-tc.services.mozilla.com/api/queue/v1/task/f5coRt3YTxuwrhaCY_LzEQ/runs/0/artifacts/public%2Fimage.tar.lz4 \
  | grep encoding        
content-encoding: gzip
```

djmitche:
Github Bug/Issue: Fixes #3899.

### DIFF
--- a/workers/docker-worker/src/lib/util/artifact_download.js
+++ b/workers/docker-worker/src/lib/util/artifact_download.js
@@ -38,13 +38,22 @@ module.exports = async function(queue, stream, taskId, artifactPath, destination
     try {
       let expectedSize = 0;
       let receivedSize;
-      let req = request.get(artifactUrl);
+      let req = request.get({
+        method: 'GET',
+        url: artifactUrl,
+        gzip: true,
+      });
       req.on('response', (res) => {
+        // measure the expected HTTP response body size and actual, to compare later
         expectedSize = parseInt(res.headers['content-length']);
         receivedSize = 0;
+        res.on('data', chunk => {
+          // note that this event is emitted with compressed data
+          receivedSize += chunk.length;
+        });
       });
       req.on('data', (chunk) => {
-        receivedSize += chunk.length;
+        // note that this event is emitted with uncompressed data
         hash.update(chunk);
       });
 
@@ -68,13 +77,15 @@ module.exports = async function(queue, stream, taskId, artifactPath, destination
         throw error;
       }
 
+      // can only check the expected size if there was no content-encoding; otherwise,
+      // we assume that request checked it on the compressed data.
       if (receivedSize !== expectedSize) {
-        throw new Error(`Expected size is '${expectedSize}' but received '${receivedSize}'`);
+        throw new Error(`Expected response size is '${expectedSize}' but received '${receivedSize}'`);
       }
 
       stream.write(fmtLog('Downloaded artifact successfully.'));
       stream.write(fmtLog(
-        `Downloaded ${(expectedSize / 1024 / 1024).toFixed(3)} mb`,
+        `Downloaded ${(receivedSize / 1024 / 1024).toFixed(3)} mb over the wire`,
       ));
       return `sha256:${hash.digest('hex')}`;
     } catch(e) {
--- a/changelog/issue-3899.md
+++ b/changelog/issue-3899.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 3899
+---
+Docker-worker now decompresses downloaded images when they have a compressed content-encoding, as artifacts produced by docker-worker now have.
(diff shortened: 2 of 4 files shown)


djmitche:
Mostly looking for tests to pass here, but assuming they do please review :)

djmitche:
TFW you add debug prints and the test stops failing.

djmitche:
Oh, right, I *did* fix something ;)

djmitche:
OK, tests are now passing -- please review :)

workers/docker-worker/src/lib/util/artifact_download.js
```diff
@@ -38,13 +38,22 @@ module.exports = async function(queue, stream, taskId, artifactPath, destination
...
         expectedSize = parseInt(res.headers['content-length']);
         receivedSize = 0;
+        res.on('data', chunk => {
+          // note that this event is emitted with compressed data
+          receivedSize += chunk.length;
+        });
       });
       req.on('data', (chunk) => {
-        receivedSize += chunk.length;
+        // note that this event is emitted with uncompressed data
```
jwhitlock, line 56: Does this portion also run for compressed content, but after uncompression? Otherwise the hash will be empty (maybe `e3b0c44298fc1c1`)

workers/docker-worker/src/lib/util/artifact_download.js
```diff
@@ -38,13 +38,22 @@ module.exports = async function(queue, stream, taskId, artifactPath, destination
     try {
       let expectedSize = 0;
       let receivedSize;
-      let req = request.get(artifactUrl);
+      let req = request.get({
+        method: 'GET',
+        url: artifactUrl,
+        gzip: true,
```
jwhitlock, line 44: I wasn't able to find the `gzip` option in the `superAgent` docs or (latest) source code. Is it necessary?

jwhitlock (approved):
Thanks @djmitche! The `superAgent` docs don't appear to address this case, it must have taken some digging.

I tried for a few hours to get tests running locally, with the hopes of seeing what the hash is of the image, and how the test exercises the gzipped path. However, it takes a while to get Virtualbox and vagrant setup, and the local setup instructions appear to be missing some steps.

So, I haven't verified the changes locally, but it looks like it works, and CI is happy. 🚢 SHIP IT

workers/docker-worker/src/lib/util/artifact_download.js
djmitche: Yeah, my understanding is that both `req` and `res` emit `data` events in any case, and that in the content-encoding case one of those is pre- and one is post-decoding.  The tests check an uncompressed artifact as well, verifying this.

workers/docker-worker/src/lib/util/artifact_download.js
djmitche: So this is pretty crazy, but docker-worker is full of crazy.  In this case we're using [`request`](https://classic.yarnpkg.com/en/package/request), not superagent.  (Really I suspect this was just written this way long before `request` was deprecated).  Here are the docs, such as they are -- https://github.com/request/request#examples

jwhitlock: Oops, I've been staring at a lot of `const request = require('superagent');`, I missed this was different. I guess a lot of code was using `request`, and that's why it isn't `const superagent = require('super agent')`

djmitche: Believe it or not, superagent itself recommends
```
const request = require('superagent');
```
I have no idea why..

PR #3918, opened 2020-11-13
Give tc-gh handler queue client some scopes

### LINKED ISSUE #3867: Parts of Github integration currently won't work in deployments without an `anonymous` role
**Describe the bug**
taskcluster-github queries taskcluster-queue a lot for various bits of information needed to update statuses in github checks/statuses. This is done with a client with no scopes available. This was fine prior to rfc165 but now we need to give this client certain scopes in its `static/taskcluster/github` client for locked down deployments with no `anonymous` role.

This is fairly easy to do but it does raise some questions about what this would be allowing. The most obvious issue is with custom check annotations. taskcluster-github loads artifacts for this and as it stands can only access public artifacts. If it instead is using its entire set of scopes, it could read almost any artifact in the deployment!

There should be some compromise we can reach here.

imbstack:
I believe this should solve this issue in the most direct way possible.

What we're doing here is giving the queue client in tc-gh's handler creds so that it
can use its own scopes and then limiting it to only a task's scopes when it is
acting on behalf of a task in a way that the task can alter what it does.

This should not break anything in community/firefoxci because those have `anonymous`
roles and so it will work without this fix even.

Github Bug/Issue: Fixes #3867

### DIFF
--- a/ui/docs/reference/integrations/github/checks.mdx
+++ b/ui/docs/reference/integrations/github/checks.mdx
@@ -6,14 +6,21 @@ At the moment, the feature is in beta. If you would like to try it out:
 * Add `reporting: checks-v1` line at the root level of your `.taskcluster.yml`
 * You will need `queue:route:checks` scope for your `repo:github.com/org/repo:<action>` role.
 * If you are using a decision task, make sure you add `route.checks` to the `routes` of your tasks
+* If you have a locked-down deployment without allowing public artifacts fetching in your `anonymous` role, you must add
+  `queue:get-artifact:public/github/customCheckRunText.md` and `queue:get-artifact:public/github/customCheckRunAnnotations.json`
+  to the scopes of your task (`task.scopes`) to avoid an error comment being added to your
+  commits. Note that this will change if you choose a custom artifact name (see below)
 
 ## Custom Text Output in Checks
 
- When a task completes, the Taskcluster-GitHub service can optionally include task-specific information in the check run
- displayed in the GitHub checks tab.  To do so, it looks for an artifact on the task named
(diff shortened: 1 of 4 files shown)


imbstack:
Actually... this won't work. This will cause every run without this scope added to make a `403` comment here. Back to the drawing board.

imbstack:
Ok, just doc'ed around the issues that remain. This should be ready for review now

ui/docs/reference/integrations/github/checks.mdx
```diff
@@ -6,14 +6,20 @@ At the moment, the feature is in beta. If you would like to try it out:
...
- When a task completes, the Taskcluster-GitHub service can optionally include task-specific information in the check run
- displayed in the GitHub checks tab.  To do so, it looks for an artifact on the task named
- `public/github/customCheckRunText.md` and uses the artifact content as the check run text.
- The artifact name can be customized by setting `task.extra.github.customCheckRun.textArtifactName` in the task definition.
- For example:
+When a task completes, the Taskcluster-GitHub service can optionally include task-specific information in the check run
+displayed in the GitHub checks tab.  To do so, it looks for an artifact on the task named
+`public/github/customCheckRunText.md` and uses the artifact content as the check run text.
+The artifact name can be customized by setting `task.extra.github.customCheckRun.textArtifactName` in the task definition.
+Whichever name you choose must be included in the task's scopes (`queue:get-artifact:<textArtifactName>`) as mentioned above.
```
djmitche, line 20: I think this could be rephrased a little to say this is required only if the scope is not in `anonymous` and to be clearer that this means `task.scopes`.

djmitche (approved):
Adjust (or not) and ship :)

PR #3938, opened 2020-11-16
Implement method negotiation in the downloadObject API endpoint

djmitche:
..and fix a few other minor inconsistencies in the API endpoint.

This was some preparatory work I did for #3931, and I thought I'd put it up for review while I work on the rest of that issue.

### DIFF
--- a/clients/client-go/tcobject/tcobject.go
+++ b/clients/client-go/tcobject/tcobject.go
@@ -115,7 +115,10 @@ func (object *Object) UploadObject(name string, payload *UploadObjectRequest) er
 
 // Stability: *** EXPERIMENTAL ***
 //
-// Download object data.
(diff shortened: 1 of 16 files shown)


djmitche:
I'll rebase this since the `yarn generate` stuff conflicts, but feel free to review anyway :)

djmitche:
I'll take anyone's review :)

jwhitlock (approved):
@djmitche Thanks, the changes look good. Only the metadata test fails, and I agree that a changelog entry is not needed. 

The biggest change appears to be ``protocol`` to ``method``. I've seen (and participated) in discussions about that name, ``method`` is a decent choice and looks good when used in code.

djmitche:
I think this just changed one or two leftover uses of "protocol", leaving the majority at "method".  Consistent, at least :)

PR #3942, opened 2020-11-17
Add a simple download method

djmitche:
This also
 * refactors the schemas for downloadObject to use `definitions/` and put request and response in the same file
 * changes the request format to include room for parameters to each requested method, prevent listing the same method twice, and avoid using an array so that it's clear order is not significiant
 * splits the docs up a bit, separating upload, download, and simple download

The first commit here is #3938, so only review the second commit plz.


ui/docs/reference/platform/object/download-methods.mdx
```diff
@@ -0,0 +1,72 @@
+---
+order: 3
+title: Download Methods
+---
+# Download Methods
+
+## Introduction
+
+The Object Service does not directly handle object content, but acts as an
+intermediary on behalf of other data storage service providers. That makes
```
imbstack, line 10: could it be called a "directory service to other data storage providers" or something like that? Intermediaries still could be handling the bits in my reading of this.

imbstack (approved):
seems reasonable. just some bikeshedding about what we write in docs

djmitche:
Rebased onto main, and I had to fix a bug in tc-lib-references.  @imbstack if you want to look at that commit, cool, otherwise I'll merge tomorrow morning.

imbstack (commented):
I am always 30% surprised that tc-lib-references works at all so this makes sense

PR #3943, opened 2020-11-17
Avoid spamming comments from tc-gh

### LINKED ISSUE #3637: tc-github intermittently spews out 100+ comments on some commits.
**Describe the bug**
Sometimes, when someone pushes to a tc-enabled repo, taskcluster-github repeatedly comments with the error:

```
Failed to get your artifact.
```

**To Reproduce**

This issue was first reported to me by @badboy, 7 days ago. It happened on https://github.com/mozilla-mobile/android-components/commit/d7e7482c115d477595b1430d492bb711f5c66db2, which was a regular commit on a non-fork repo.

Another instance of this bug happened yesterday, with @alopezz: https://github.com/mozilla-mobile/android-components/commit/da6225aa1558768c3b39042351b29a45d02ccba1. This commit was on a PR.

I don't have more information on how to reproduce it. I guess tc-gh logs will help 

**Expected behavior**
Either a single error or getting tc-gh schedule the decision task.

**Taskcluster version**
 v37.3.0 for yesterday's instance.

**Screenshots**
N/A

N/A

**Additional context**
None.

imbstack:
This probably isn't my dream fix for all of this but I think we need to get a quick fix in and then start building a bigger update

Github Bug/Issue: Fixes #3637

### DIFF
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -1,3 +1,6 @@
+const _ = require('lodash');
+const stringify = require('fast-json-stable-stringify');
+const crypto = require('crypto');
 const taskcluster = require('taskcluster-client');
 const libUrls = require('taskcluster-lib-urls');
 const yaml = require('js-yaml');
@@ -47,6 +50,8 @@ class Handlers {
 
     this.handlerComplete = null;
     this.handlerRejected = null;
+
+    this.commentHashCache = [];
 
     this.jobPq = null;
     this.resultStatusPq = null;
@@ -207,7 +212,7 @@ class Handlers {
             'Taskcluster-GitHub attempted to create a task for this event with the following scopes:',
             '',
             '```',
-            JSON.stringify(scopes, null, 2),
+            stringify(scopes, null, 2),
             '```',
             '',
             'The expansion of these scopes is not sufficient to create the task, leading to the following:',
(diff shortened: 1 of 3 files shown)


services/github/src/handlers.js
```diff
@@ -220,12 +225,32 @@ class Handlers {
...
 
+  commentKey(idents) {
+    return crypto
+      .createHash('md5')
+      .update(stringify(idents))
+      .digest('hex');
+  }
+
+  isDuplicateComment(...idents) {
+    return _.indexOf(this.commentHashCache, this.commentKey(idents)) !== -1;
```
jwhitlock, line 236: My brain is in optimization mode. It probably doesn't matter. And yet, I need to comment so I can move on...

It seems inefficient to iterate through up to 1,000 items in the (hopefully default) case of adding a new comment. A data structure like a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) would be a better match for this part. But, it doesn't allow removing the oldest item.

I guess what we want is an ordered set, so we can quickly test for inclusion, and also remove the last item. I've seen this implemented as a list, to remember insertion order, paired with a set, for quick lookups. To rotate the cache, you pop the end of the list, and remove the related element from the set.

This should work as a proof of concept, and could be extended to an ordered set later.

services/github/src/handlers.js
```diff
@@ -256,6 +282,7 @@ class Handlers {
       commit_sha: sha,
       body,
     });
+    this.markCommentSent(organization, repository, sha, error, pullNumber);
```
jwhitlock, line 285: I wondered if ``pullNumber`` should be sent in this call, since we know it is falsey, but I think we need to do it to match the call to ``isDuplicateComment``, so 👍

jwhitlock (approved):
Thanks @imbstack! This looks like the smallest possible thing that could work, and gets the code onto the stable JSON stringier.

services/github/src/handlers.js
djmitche: We use an LRU cache library a few other places.

imbstack: Yeah agreed. This just stole the approach from how we do it in tc-notify which could also be improved with the library I expect. Just going to land this as-is now though and put off a smarter approach for later.

PR #3944, opened 2020-11-18
Stop using two json-stable-stringify packages

imbstack:
Just noticed this while I was working on something else. I can't see any reason that both need to exist


djmitche:
https://bugzilla.mozilla.org/show_bug.cgi?id=1630958#c4 highlights what *was* the reason.  But, tc-lib-entities is gone now, so it's probably OK (CI will tell!)

djmitche (approved):
👍🏽

imbstack:
converting to draft until we get a way to get pretty-printing of some of this

petemoore:
@imbstack shall we just close this?

PR #3950, opened 2020-11-20
Implement a basic AWS object storage backend

### LINKED ISSUE #3719: [object service] Build an AWS backend

### LINKED ISSUE #3932: Build testing support for verifying the details of a download method for a storage backend

djmitche:
Github Bug/Issue: Fixes #3719, Fixes #3932.

This also implements a testing framework that can help ensure that all
backends behave the same way (which isn't so hard yet, since they do so
little!) -- that's #3932.

And it fills out the temporary upload mechanism so that it, y'know, actually uploads something :)

### DIFF
--- a/services/object/schemas/v1/upload-object-request.yml
+++ b/services/object/schemas/v1/upload-object-request.yml
@@ -1,7 +1,7 @@
 $schema: "/schemas/common/metaschema.json#"
 title:              "Upload Object Request"
 description: |
-  Representation of the object entry to insert.
+  Representation of the object entry to insert.  This is a temporary API.
 type:               object
 properties:
   projectId:
@@ -15,7 +15,13 @@ properties:
       Date at which this entry expires from the object table.
     type:           string
     format:         date-time
+  data:
+    title: "Data"
+    description: |
+      The data to upload, base64-encoded
+    type: string
 additionalProperties: false
 required:
   - expires
   - projectId
+  - data
--- a/services/object/src/backends/aws.js
+++ b/services/object/src/backends/aws.js
@@ -22,6 +22,14 @@ class AwsBackend extends Backend {
     this.s3 = new aws.S3({ region: this.region, ...credentials });
   }
 
+  async temporaryUpload(object, data) {
+    await this.s3.putObject({
+      Bucket: this.config.bucket,
+      Key: object.name,
+      Body: data,
+    }).promise();
+  }
+
   async availableDownloadMethods(object) {
     return ['simple'];
   }
--- a/services/object/src/backends/aws.js
+++ b/services/object/src/backends/aws.js
@@ -63,11 +63,14 @@ class AwsBackend extends Backend {
     // When an object is being expired, we first delete the object from S3.  Even if lifecycle
     // or other S3 policies would accomplish the same thing, this serves as a backstop to
     // prevent removing the database record while the object itself is still on S3, thereby
-    // leaking storage.
+    // leaking storage.  Note that s3.deleteObject is idempotent: this will not fail if the
+    // object does not exist.
     await this.s3.deleteObject({
       Bucket: this.config.bucket,
       Key: object.name,
     }).promise();
+
+    return true;
   }
 }
 
--- a/services/object/src/backends/aws.js
+++ b/services/object/src/backends/aws.js
@@ -42,9 +42,10 @@ class AwsBackend extends Backend {
           url = await this.s3.getSignedUrlPromise('getObject', {
             Bucket: this.config.bucket,
             Key: object.name,
-            // 30 seconds is copied from the queue; the idea is that the download
+            // 30 minutes is copied from the queue; the idea is that the download
             // begins almost immediately.  It might also make sense to use the
             // expiration time of the object here.
+            // https://github.com/taskcluster/taskcluster/issues/3946
             Expires: 30 * 60,
           });
         } else {
--- a/services/object/src/backends/aws.js
+++ b/services/object/src/backends/aws.js
@@ -1,0 +1,75 @@
+const { Backend } = require('./base');
+const assert = require('assert');
+const aws = require('aws-sdk');
+
+class AwsBackend extends Backend {
+  constructor(options) {
+    super(options);
+
+    for (let prop of ['accessKeyId', 'secretAccessKey', 'bucket', 'signGetUrls']) {
+      assert(prop in options.config, `backend ${options.backendId} is missing ${prop}`);
+    }
+
+    this.config = options.config;
+  }
+
+  async setup() {
+    const credentials = {
+      accessKeyId: this.config.accessKeyId,
+      secretAccessKey: this.config.secretAccessKey,
+    };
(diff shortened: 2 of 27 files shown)


services/object/src/backends/aws.js
```diff
@@ -0,0 +1,86 @@
...
+  async availableDownloadMethods(object) {
+    return ['simple'];
+  }
+
+  async downloadObject(object, method, params) {
+    switch (method){
+      case 'simple': {
+        let url;
+        if (this.config.signGetUrls) {
+          url = await this.s3.getSignedUrlPromise('getObject', {
```
jwhitlock, line 42: A signed URL stretches my definition of ``simple``, but at least you can ``GET`` it.

services/object/src/backends/aws.js
```diff
@@ -0,0 +1,86 @@
...
+            Bucket: this.config.bucket,
+            Key: object.name,
+            // 30 seconds is copied from the queue; the idea is that the download
```
jwhitlock, line 45: is ``30 * 60`` 30 seconds or minutes?

services/object/src/backends/aws.js
jwhitlock, line 42: Signed URLs stretch my concept of ``simple``, but it is retrievable with a ``GET``. So stretch but not break.

services/object/src/backends/aws.js
jwhitlock, line 45: Is ``30 * 60``thirty seconds or minutes?

services/object/src/backends/aws.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  async expireObject(object) {
+    // When an object is being expired, we first delete the object from S3.  Even if lifecycle
+    // or other S3 policies would accomplish the same thing, this serves as a backstop to
+    // prevent removing the database record while the object itself is still on S3, thereby
+    // leaking storage.  Note that s3.deleteObject is idempotent: this will not fail if the
+    // object does not exist.
+    await this.s3.deleteObject({
```
jwhitlock, line 68: Random digression: If you turn on [Object Versioning](https://docs.aws.amazon.com/AmazonS3/latest/dev/ObjectVersioning.html), which I think is a checkbox in manual S3 setup and feels like a good idea, then deletion just adds a "deleted" layer. You can still access the data by adding a previous version ID, and you still get to pay for storage.

Maybe we'll have a use case for versioned buckets someday.

services/object/src/backends/base.js
```diff
@@ -21,6 +19,13 @@ class Backend {
   async setup() {
   }
 
+  /**
+   * Temporary upload method that includes the uploded data in the reuqest body
```
jwhitlock, line 23:
```suggestion
   * Temporary upload method that includes the uploded data in the request body
```

services/object/test/api_test.js
```diff
@@ -14,9 +16,14 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
     await helper.apiClient.ping();
   });
 
-  suite('uplodaObject method', function() {
+  suite('uploadObject method', function() {
```
jwhitlock, line 19: ❤️

ui/docs/reference/platform/object/aws-backend.mdx
```diff
@@ -0,0 +1,68 @@
...
+                "arn:aws:s3:::<BUCKET_NAME>/*"
+            ]
+        }
+    ]
+}
+```
+
+The use of `signGetUrls` is described above.
+If this configuration value is `false`, then the bucket should be configured to allow public read access.
+AWS has many layers of "yes, I'm really quite sure" that must be clicked through to accopmlish this.
```
jwhitlock, line 68:
```suggestion
AWS has many layers of "yes, I'm really quite sure" that must be clicked through to accomplish this.
```

jwhitlock (approved):
Thanks @djmitche! With a working AWS backend, the object storage service design is becoming clearer.

services/object/src/backends/aws.js
djmitche: Yeah, simple _for the caller_

services/object/schemas/v1/upload-object-request.yml
```diff
@@ -15,7 +15,13 @@ properties:
       Date at which this entry expires from the object table.
     type:           string
     format:         date-time
+  data:
+    title: "Data"
+    description: |
+      The data to upload, base64-encoded
```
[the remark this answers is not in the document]
djmitche: Yeah, that's the plan.  That is outside of the realm of the MVP though (thankfully).

services/object/src/backends/aws.js
djmitche: oh good point!

services/object/src/backends/aws.js
djmitche: Yeah, that can also be used for allowing a few days' time to undo deletions.  I think in this case it might not be useful.  But if you want to read more about S3, there's some interesting newish stuff about automatically migration objects to other storage classes that might be useful..

djmitche:
(I also added the necessary secret, and fixed the meta test)

services/object/src/backends/aws.js
```diff
@@ -0,0 +1,75 @@
...
+    this.config = options.config;
+  }
+
+  async setup() {
+    const credentials = {
+      accessKeyId: this.config.accessKeyId,
+      secretAccessKey: this.config.secretAccessKey,
+    };
+    this.region = await getBucketRegion({ bucket: this.config.bucket, credentials });
+    this.s3 = new aws.S3({ region: this.region, ...credentials });
```
imbstack, line 22: Just a question that this brings to mind (not blocking landing this) how does this interact with being able to roll aws creds?

ui/docs/reference/platform/object/aws-backend.mdx
```diff
@@ -65,4 +64,4 @@ The provided AWS credentials should have the following policy (with `<BUCKET_NAM
 
 The use of `signGetUrls` is described above.
 If this configuration value is `false`, then the bucket should be configured to allow public read access.
-AWS has many layers of "yes, I'm really quite sure" that must be clicked through to accopmlish this.
+AWS has many layers of "yes, I'm really quite sure" that must be clicked through to accomplish this.
```
imbstack, line 67: occasionally I spend a few minutes thinking about a spellcheck lint on our comments... one day

imbstack (approved):
makes sense

services/object/src/backends/aws.js
djmitche: I think just configuring new creds and restarting the service would do the trick.  But it's a good point regarding expiration of links, if we use years-long expirations..

ui/docs/reference/platform/object/aws-backend.mdx
djmitche: That would kill me :(

PR #3955, opened 2020-11-20
Update vnc-test image to use a newer alpine

### LINKED ISSUE #3954: docker-worker tests failing because taskcluster/vnc-test image is missing
I assume docker hub deleted this based on their new policies.  I thought those policies would only delete images that hadn't been used, but apparently that's not the case.

djmitche:
Docker-hub deleted the older version, and installs on alpine-3.2 no
longer work.

Github Bug/Issue: Fixes #3954

### DIFF
--- a/changelog/issue-3954.md
+++ b/changelog/issue-3954.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3954
+---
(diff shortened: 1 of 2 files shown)


jwhitlock (approved):
Thanks @djmitche!

PR #3990, opened 2020-11-30
escape '.' in regular expressions in tests, to make security scanners…

djmitche:
...stop complaining.

There are 7 "CodeQL" alerts for this. Easy enough to make those go away.

### DIFF
--- a/changelog/f5uwdd7GSkiLQ39h-uBKcQ.md
+++ b/changelog/f5uwdd7GSkiLQ39h-uBKcQ.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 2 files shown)


jwhitlock (changes requested):
Thanks @djmitche, but that didn't seem to make the security scanner any happier. I'm not familiar with this code, but maybe use a regex literal like `/http://tc-tests\.example\.com/` instead?

djmitche:
I think I needed `\\` (or to use JS `/regexp syntax/` maybe).

jwhitlock (approved):
Thanks @djmitche, the scanner is happy! I, however, am sad that I can't think of a slash-themed comment. 😞

PR #3992, opened 2020-12-01
add peer dependency on @octokit/core

djmitche:
I spotted this unmet peer dependency.  Something else was pulling in the package, so nothing failed -- but we should be explicit.


petemoore (approved):
Nice spot!

Issue #3721, opened 2020-10-21
Routes containing slashes do not get indexed, with no visible error


djmitche:
Thanks to @ricky26 for spotting this.  

https://community-tc.services.mozilla.com/tasks/c4FFtHALTq6MqZqZregE1Q
```
{
  "provisionerId": "built-in",
  "workerType": "succeed",
  "schedulerId": "taskcluster-ui",
  "taskGroupId": "c4FFtHALTq6MqZqZregE1Q",
  "dependencies": [],
  "requires": "all-completed",
  "routes": [
    "index.project.taskcluster.foo/bar/slash"
  ],
  "priority": "lowest",
  "retries": 0,
  "created": "2020-10-21T15:52:53.138Z",
  "deadline": "2020-10-22T03:52:53.138Z",
  "expires": "2020-10-28T15:52:53.138Z",
  "scopes": [],
  "payload": {},
  "metadata": {
    "name": "Testy",
    "owner": "unkown@localhost.local",
    "source": "http://foobar.com",
    "description": "jonas damn you <-- :heart:"
  },
  "tags": {},
  "extra": {}
}
```
this task succeeded, but no index entry appeared.  I don't see any logging (aside from db-pool-counts) from index at this time.  Modifying the task to change `/` to `.` in the route succeeds.

Nothing in the [AMQP Spec](https://www.rabbitmq.com/resources/specs/amqp-xml-doc0-9-1.pdf) disallows `/`, so I think this is an issue in the index, and I'm not sure how best to make the error obvious to users.

djmitche:
The proximal cause appears to be
https://github.com/taskcluster/taskcluster/blob/b3a64063e941cbd55c3c650b21cbf429dcde840a/services/index/src/helpers.js#L7
that format is used to filter routes:
https://github.com/taskcluster/taskcluster/blob/b3a64063e941cbd55c3c650b21cbf429dcde840a/services/index/src/handlers.js#L76-L82
and also appears in the API, restricting the `index.insertTask` method:
https://github.com/taskcluster/taskcluster/blob/b3a64063e941cbd55c3c650b21cbf429dcde840a/services/index/src/api.js#L25-L28

I'm not sure if `/` was disallowed there by accident, or on purpose.  The reasons I can think of to intentionally disallow it are to avoid confusion in a URL and to avoid scopes shadowing one another.  Those would happen where a namespace is embedded in a string that is looking for a `/` after the namespace.

Scopes don't appear to be an issue:
```
[...]

PR #3997, opened 2020-12-01
Bug 1680064 - Allow findTask to search for indexes that contain /

JohanLorenzo:
Bugzilla Bug: [1680064](https://bugzilla.mozilla.org/show_bug.cgi?id=1680064)

The new regex has been tested on https://regex101.com/ against `mobile.v2.fenix.branch.releases/v84.0.0.revision.707dd46284878c4081e8f1e194f61e189f88af0f.taskgraph.decision`. This index is not visible on https://firefox-ci-tc.services.mozilla.com/tasks/index/mobile.v2.fenix.branch, although it was accepted when the task was submitted https://firefox-ci-tc.services.mozilla.com/tasks/UUiXNj-pS56IqghasEbkRA

### DIFF
--- a/changelog/bug-1680064.md
+++ b/changelog/bug-1680064.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: bug 1680064
+---
+Allow findTask to search for indexes that contain /
--- a/services/index/src/helpers.js
+++ b/services/index/src/helpers.js
@@ -4,7 +4,7 @@ const { paginateResults } = require('taskcluster-lib-api');
 const { UNIQUE_VIOLATION } = require('taskcluster-lib-postgres');
 
 /** Regular expression for valid namespaces */
-exports.namespaceFormat = /^([a-zA-Z0-9_!~*'()%-]+\.)*[a-zA-Z0-9_!~*'()%-]+$/;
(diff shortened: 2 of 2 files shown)


JohanLorenzo:
> although it was accepted when the task was submitted 

Hmmm, maybe it was just silently ignored, per this test: https://github.com/taskcluster/taskcluster/blob/f552ded612a9763fc91a7e020837bbc37e0fdc55/services/index/test/index_test.js#L28 🤔

djmitche:
We did some research on this topic in #3721 and, unfortunately, determined that `/` is omitted from that regexp for good reason.

It'd be great if the queue refused such things in `task.routes`, but `/` is valid in a pulse route in general -- just not for index namespaces.  I posted [an idea](https://github.com/taskcluster/taskcluster/issues/3721#issuecomment-718887739) of how to accomplish that, but its effect would only be to give an error earlier, not to allow `/`.  I filed #3998 for that issue, since it's come up twice now.

JohanLorenzo:
Thanks for the super quick feedback and for the context! I just found [bug 1150304](https://bugzilla.mozilla.org/show_bug.cgi?id=1150304) which first introduced that restriction. It seems to be a more complex issue than I originally thought. 

I think the easiest fix on our end would be to stop using `/` in Git branch names. This way, we don't have to makes such a heavy change 🙂

PR #4000, opened 2020-12-01
Support uploadIds for uploads

### LINKED ISSUE #3939: [object service] Add an uploadId to the upload API
@petemoore suggested adding an id for an upload, provided by the caller.  This allows the usual API idempotency: a second upload call with the same uploadId would return the same response (allowing retries for HTTP failures), while a second upload call with a different uploadId would fail with a 409.

This will also be useful if we add additional upload-related API methods, such as a progress/completion method.  That can use the same uploadId, preventing another caller (who does not know the uploadId) from updating that information.

Note that this is an ID for the upload process, not for the object itself.  It's not visible from the API, and is irrelevant once the object is uploaded.

djmitche:
- requests witht he same uploadId are idempotent
 - objects are not considered ready for download until uploadId is NULL
 - objects "stuck" uploading are expired sooner than their expiration date

Github Bug/Issue: Fixes #3939


djmitche:
There appears to be some issue with a race between resetting the tables and running the tests.. a forgotten await, no doubt :(

petemoore (approved):
Looks good, many thanks!

djmitche:
(rebased)

PR #4004, opened 2020-12-01
fix handling of named parameters to db functions

### LINKED ISSUE #3999: db.fns named parameters only work if given in order
`libraries/postgres/src/Database.js` uses [`yesql`](https://classic.yarnpkg.com/en/package/yesql) to convert named parameters to unnamed parameters before calling postgres with them.  It provides the names of the parameters to `yesql` in the order they were given in the argument to the DB function.

So, given a DB function `foo(a_in string, b_in string, c_in string)`, this will work:

```js
db.fns.foo({
  a_in: 'aardvark',
  b_in: 'bullfrog',
  c_in: 'caterpillar',
});
```
but this will not:
```js
db.fns.foo({
  c_in: 'caterpillar',
  b_in: 'bullfrog',
  a_in: 'aardvark',
});
```

Postgres [supports named arguments directly](https://www.postgresql.org/docs/11/sql-syntax-calling-funcs.html) so we could just use that functionality rather than yesql.

djmitche:
Github Bug/Issue: Fixes #3999

### DIFF
--- a/libraries/postgres/src/Database.js
+++ b/libraries/postgres/src/Database.js
@@ -7,7 +7,6 @@ const Keyring = require('./Keyring');
 const assert = require('assert').strict;
 const { READ, WRITE, DUPLICATE_OBJECT, UNDEFINED_TABLE } = require('./constants');
 const { MonitorManager } = require('taskcluster-lib-monitor');
-const named = require('yesql').pg;
 const { parse: parseConnectionString } = require('pg-connection-string');
 const { runMigration, runOnlineMigration, runDowngrade, runOnlineDowngrade } = require('./migration');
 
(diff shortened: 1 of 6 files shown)


libraries/postgres/test/database_test.js
```diff
@@ -94,14 +94,14 @@ helper.dbSuite(path.basename(__filename), function() {
             return 'got ' || x;
           end`,
         },
-        add_from_numbers_named: {
+        sub_from_numbers_named: {
```
imbstack, line 97: why change this to subtraction?

libraries/postgres/src/Database.js
```diff
@@ -109,9 +108,11 @@ class Database {
           try {
             let res;
             if (hasNamedArguments) {
-              const placeholders = Object.keys(args[0]).map(arg => `:${arg}`).join(',');
-              const { text, values } = named(`select * from "${method.name}"(${placeholders})`, { useNullForMissing: true })(args[0]);
-              res = await client.query(text, values);
+              const obj = args[0];
+              const keys = Object.keys(obj);
+              const values = keys.map(k => obj[k]); // order must match keys!
+              const placeholders = keys.map((key, i) => `${key} => $${i + 1}`).join(',');
```
imbstack, line 114: could you inject something by naming a key in a certain way here?

imbstack (approved):
I think this makes sense. I wonder what the point of `yesql` even is if it requires order to be maintained?

libraries/postgres/test/database_test.js
djmitche: Because addition is commutative so the test would otherwise still pass if the order were reversed (!)

libraries/postgres/src/Database.js
djmitche: Yes, good point, although I'd expect those not to come from user input.  But just in case, I can add a somewhat more restrictive regex where we check for the `_in` suffix.

libraries/postgres/test/database_test.js
imbstack: ah! makes sense

libraries/postgres/src/Database.js
imbstack: ok great. I figured it would be fine for now but who knows what future-us will end up doing with this

imbstack (approved):

PR #4014, opened 2020-12-02
Backend setup() fail in object service

### LINKED ISSUE #3961: Allow backend setup() to fail and retry
The object service calls `setup` for every backend when it starts up.  Some of those `setup` functions make API calls to cloud providers.  If cloud provider A is down when the process starts, that will mean the setup fails and the object service is totally down, even if cloud provider B is still working fine.

Let's find a way to
 * handle failures of a single backend's `setup`
 * retry those failures periodically
 * make all other calls to the backend fail until `setup` is complete

This should ensure that a single-cloud outage doesn't take out the entire object service.

SushantAA:
Github Bug/Issue: Fixes #3961 
try-catch add in index.js and aws.js
error message in index.js setup() fail

I am starting this PR for the code logic check and haven't done yarn generate.

### DIFF
--- a/services/object/src/backends/aws.js
+++ b/services/object/src/backends/aws.js
@@ -18,8 +18,19 @@ class AwsBackend extends Backend {
       accessKeyId: this.config.accessKeyId,
       secretAccessKey: this.config.secretAccessKey,
     };
+
+    try{
     this.region = await getBucketRegion({ bucket: this.config.bucket, credentials });
+    }catch(err){
+      return;
+    }
+    
+    try{
     this.s3 = new aws.S3({ region: this.region, ...credentials });
+    }catch(err){
+      return;
+    }
+    
   }
 
   async temporaryUpload(object, data) {
--- a/services/object/src/backends/aws.js
+++ b/services/object/src/backends/aws.js
@@ -19,15 +19,24 @@ class AwsBackend extends Backend {
       secretAccessKey: this.config.secretAccessKey,
     };
 
+    let oldRegion = this.region;
+
     try{
-    this.region = await getBucketRegion({ bucket: this.config.bucket, credentials });
+      this.region = await getBucketRegion({ bucket: this.config.bucket, credentials });
     }catch(err){
+      this.region = oldRegion;
+      this.monitor.reportError(err.message);
       return;
     }
+
+    let oldS3 = this.s3;
     
     try{
-    this.s3 = new aws.S3({ region: this.region, ...credentials });
+      this.s3 = new aws.S3({ region: this.region, ...credentials });
     }catch(err){
+      this.region = oldRegion;
+      this.s3 = oldS3;
+      this.monitor.reportError(err.message);
       return;
     }
     
--- a/services/object/src/backends/index.js
+++ b/services/object/src/backends/index.js
@@ -24,6 +24,11 @@ class Backends {
     if (!cfg.backends) {
       throw new Error('No backends configured');
     }
+
+    let loopOver = false;
+
+    while(!loopOver){
+
     for (const [backendId, config] of Object.entries(cfg.backends)) {
       const Backend = BACKEND_TYPES[config.backendType];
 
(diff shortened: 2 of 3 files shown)


services/object/src/backends/aws.js
```diff
@@ -18,8 +18,19 @@ class AwsBackend extends Backend {
...
+    try{
     this.region = await getBucketRegion({ bucket: this.config.bucket, credentials });
+    }catch(err){
+      return;
+    }
+    
+    try{
     this.s3 = new aws.S3({ region: this.region, ...credentials });
+    }catch(err){
+      return;
```
djmitche, line 31: The effect of this change is just to ignore erorrs!  We definitely don't want to do that.  This issue is about reporting errors and allowing other providers to continue to operate, with just the failed provider not working.

services/object/src/backends/index.js
```diff
@@ -39,8 +40,16 @@ class Backends {
         config,
       });
       this._backends.set(backendId, backend);
-      await backend.setup();
+      
+      try{
+        await backend.setup();
+      }catch(err){
+        this.monitor.reportError(new Error(err.message));
```
djmitche, line 47:
```suggestion
        this.monitor.reportError(err.message);
```

This is the right idea, but with the changes in `aws.js` would not end up reporting any errors (because they'd be ignored in that file).

djmitche (changes requested):
The other requirement here is to find a way to avoid calling other methods on the provider when its setup() hasn't succeeded.  For example, we need to prevent calls from `provision(..)` on the aws provider if `setup` didn't succeed, as otherwise `this.region` and `this.s3` might not be set correctly.

services/object/src/backends/aws.js
SushantAA: I am a bit confused. so I have to  add an error message or remove try-catch

djmitche: Hm, I got a notification of a comment here, but now I don't see it.. did it get deleted?  Deleting comments in a conversation is really confusing!

SushantAA: ok,
sorry for that 
I have made some changes

djmitche (changes requested):
I see you've added `oldRegion` and `oldS3`, but neither `this.region` nor `this.s3` are set before setup completes, so those will just be `undefined`.  I'm not sure what that accomplishes..

Maybe some test-driven development will help, by writing a new test in backends_test.js. Configure TestBackend so that its setup method will fail with an error (we usually use the message "uhoh" so it's obvious that it's not a "real" error) when called in this test, and then call `backends.setup()`.  That call should return successfully, but an error should be logged.  And, a subsequent call to `backends.get` that returns that backend should fail with an error indicating that the backend's setup failed.

Once you've got that test in place, then it will be a bit easier to tell what are the right changes to make to the code.

SushantAA:
This question is silly but I have no about the changes which I make are working or not. because I am confused about how to see it or get an idea about it.

SushantAA:
I have moved a line below which was basically adding the backend (id, backend) to the map( _backends ). Every time other functions are called then it was getting added to the map (_backends
) and didn’t check anything. Other function uses the map ( _backends ) to get backend and backend-Id. I think it should work, I don’t know how to check if it’s working or not.

services/object/src/backends/index.js
```diff
@@ -38,9 +35,17 @@ class Backends {
...
       });
+      
+      try{
+        await backend.setup();
+      }catch(err){
+        this.monitor.reportError(err.message);
+        continue;
+      }
+
       this._backends.set(backendId, backend);
```
djmitche, line 46: This looks good -- the effect is that the backend isn't in `_backends` if it didn't set up correctly.  That might leave some objects inaccessible, but that's the best we can do in the circumstances!

djmitche (commented):
Based on our chat, I'm going to close this for now.  We can re-open or open a new PR when you come back to this.  Thanks for working on it!!

PR #4029, opened 2020-12-03
Update highlight.js

helfi92:
Relates to https://github.com/taskcluster/taskcluster/issues/3630.


jwhitlock (approved):
Thanks @helfi92! The `yarn.lock` diff looks as awkward as ever, but seems to get the correct versions.

PR #4030, opened 2020-12-03
Stop shipping packed versions of tc-client-web

### LINKED ISSUE #4006: Use `core-js` to enable polyfills
https://neutrinojs.org/packages/react/
> Important! If you need polyfills in your code, consider including core-js in your package.json. This is will configure @babel/preset-env to automatically include polyfills based on usage. *

I think this might be why we're having trouble with hawk's requirements (mozilla/hawk#278).

djmitche:
Github Bug/Issue: Fixes #4006.

This drops Neutrino (it wasn't adding much, and was potentially holding us back), and just uses webpack to run the karma tests.  Like most libraries, it's up to the user of the library to webpack it if they're using it in a browser.  So, no more 

The two babel plugins appear to be unnecessary -- tests pass without, anyway.

I tested this with the as-yet-unreleased hawk@9.0 and it works just fine, polyjuice potion and all.


petemoore (commented):
Seems ok to me but I'll defer to @helfi92 as he may have more context than me.

helfi92 (approved):

PR #4038, opened 2020-12-04
Add "cdn" middleware to the object service.

### LINKED ISSUE #3951: try building a CDN "wrapper" backend
The idea here would be that the CDN backend uses a CDN to support simple downloads, but defers everything else to another backend.

Something like
```yaml
backends:
  storage:
    backendType: aws
    bucket: all-my-objects
  cdn:
    backendType: cdn
    cdnType: cloudfront  # I suspect CDNs from different clouds are similar enough that
    cdnId: E023985019    # it is OK to just handle them all in the same backend type
    nextBackendId: storage  # defers to the 'storage' backend for everything but simple downloads
backendMap:
  # send public artifact downloads to the CDN
  - backendId: cdn
    when:
      name: {regexp: "^public/"}
  # all other artifacts just use the storage backend directly
  - backendId: storage
    when: all
```

I think this idea of "wrapping" backends could be really powerful.  This is a pretty limited way to experiment with it.

djmitche:
This is *super* basic!  I can imagine we might eventually want to be able to "filter" more precisely.  Like, maybe we want to use a CDN for only some projects, or different object names on different projects.  Maybe we should find a way to generalize the backendMap functionality into a kind of filtering expression that could be used here, too.

I can also imagine that some CDNs might need something more complex than just "append the object name here".  Maybe we'll want some kind of regexp / replace functionality.

In any case, I think this is good for now but I'm happy to be told otherwise.

Github Bug/Issue: Fixes #3951

### DIFF
--- a/generated/docs-search.json
+++ b/generated/docs-search.json
@@ -399,6 +399,13 @@
     "title": "Object Service"
   },
   {
+    "element": "h3",
+    "id": "cdn",
+    "path": "/manual/deploying/object-service",
+    "subtitle": "CDN",
+    "title": "Object Service"
+  },
+  {
     "element": "h2",
     "id": "middleware-types",
     "path": "/manual/deploying/object-service",
--- a/services/object/src/middleware/cdn.js
+++ b/services/object/src/middleware/cdn.js
@@ -1,0 +1,27 @@
+const assert = require('assert');
+const { Middleware } = require('./base.js');
+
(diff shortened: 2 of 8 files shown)


djmitche:
(since it's not yet in use I think we can skip changelogs for object-service stuff, ymmv)

imbstack (approved):
very neat!

PR #4039, opened 2020-12-04
Add a GCS backend based on the AWS backend

### LINKED ISSUE #3718: [object service] build a GCS backend

petemoore:
Github Bug/Issue: Fixes #3718 

I've successfully configured a GCS bucket and uploaded and downloaded objects to it using the AWS client.

I spent some time with @helfi92 trying to get tests to work but haven't yet been successful, so some guidance here would be helpful.

Just as when using a real AWS S3 backend with the object service, an access key ID and a secret access key are both needed. However, additionally, the endpoint URI is required as configuration, and needs to be set to `https://storage.googleapis.com`. More information is available at https://cloud.google.com/storage/docs/interoperability.

Under the `pmoore-dev` project of our gcloud account, I was able to create a bucket `pmoore-potatoes`, create an HMAC and service account, and configure my local s3 client to use them, and then was able to upload and download objects...

```
pmoore@pmoore-laptop:~/git/taskcluster issue3718 $ aws s3 --endpoint-url https://storage.googleapis.com cp ~/Desktop/2880px-Miles_Gordon_Technology_Logo.svg.png s3://pmoore-potatoes
pmoore@pmoore-laptop:~/git/taskcluster issue3718 $ aws s3 --endpoint-url https://storage.googleapis.com ls pmoore-potatoes
2020-12-04 15:54:08      88854 2880px-Miles_Gordon_Technology_Logo.svg.png
2020-12-01 16:17:58     177348 WhatsApp Image 2020-11-27 at 11.49.22.jpeg
pmoore@pmoore-laptop:~/git/taskcluster issue3718 $ 
```

The changes to the code are in fact slight: we simply needed to add the capability to support a different S3 endpoint, in order to have the requests sent to GCS rather than S3.

The problems we are having running the tests seem to be related to the helper methods that create the private and public buckets, and I suspect there are some API calls somewhere in the test harness that also need to be updated to use the endpoint url that we've added to the service code.

For now, we've left the backend name as aws, rather than creating a new one called gcs or google, since the code is identical. Our assumption is that a running taskcluster service would use only one backend at a time, as a configuration setting, rather than uploading some objects to s3 and some to gcs. This is also because the object service API does not provide a means for a request to specify a specific backend, so our assumption is that all object service requests will be serviced by the same backend, which would be a deployment configuration setting.

Please let us know what additional testing we should do, if any, or if this is ok to ship as it is.

Many thanks!

### DIFF
--- a/changelog/issue-3718.md
+++ b/changelog/issue-3718.md
@@ -1,0 +1,6 @@
+audience: deployers
+level: silent
+reference: issue 3718
+---
+Provides support for using Google Cloud Storage as a back end for the Object Service. Under the hood this uses the S3 protocol via the interoperability API.
+See more at https://cloud.google.com/storage/docs/interoperability.
--- a/services/object/src/backends/aws.js
+++ b/services/object/src/backends/aws.js
@@ -72,17 +72,17 @@ class AwsBackend extends Backend {
     // or other S3 policies would accomplish the same thing, this serves as a backstop to
     // prevent removing the database record while the object itself is still on S3, thereby
     // leaking storage.  Note that s3.deleteObject is idempotent in AWS, but not in Google,
-    // and since we don't care about objects that couldn't be deleted, we swallow any errors.
+    // and since we don't care about objects that couldn't be deleted, we swallow NoSuchKey errors.
     try {
       await this.s3.deleteObject({
         Bucket: this.config.bucket,
         Key: object.name,
       }).promise();
     } catch (error) {
-      // This can happen if object expired between the time the object metadata was retrieved
-      // and the delete was executed, or if there was a network blip that caused a retry to
-      // occur.  In no case do we really care if an object couldn't be deleted, as this
-      // expiration process is only a backstop to e.g. a bucket level expiration policy.
+      // Ignore NoSuchKey errors
+      if (error.code !== "NoSuchKey") {
+        throw error;
+      }
     }
     return true;
   }
--- a/services/object/src/backends/aws.js
+++ b/services/object/src/backends/aws.js
@@ -18,8 +18,15 @@ class AwsBackend extends Backend {
       accessKeyId: this.config.accessKeyId,
       secretAccessKey: this.config.secretAccessKey,
     };
-    this.region = await getBucketRegion({ bucket: this.config.bucket, credentials });
-    this.s3 = new aws.S3({ region: this.region, ...credentials });
+    // only include endpoint if included, since included for gcp backend but not needed for aws backend
+    let options = { ...credentials };
+    if ('endpoint' in this.config) {
+      options.endpoint = new aws.Endpoint(this.config.endpoint);
+    } else {
+      this.region = await getBucketRegion({ bucket: this.config.bucket, endpoint: this.config.endpoint, credentials });
+      options.region = this.region;
+    }
+    this.s3 = new aws.S3(options);
   }
 
   async temporaryUpload(object, data) {
@@ -64,19 +71,29 @@ class AwsBackend extends Backend {
     // When an object is being expired, we first delete the object from S3.  Even if lifecycle
     // or other S3 policies would accomplish the same thing, this serves as a backstop to
     // prevent removing the database record while the object itself is still on S3, thereby
-    // leaking storage.  Note that s3.deleteObject is idempotent: this will not fail if the
-    // object does not exist.
-    await this.s3.deleteObject({
-      Bucket: this.config.bucket,
-      Key: object.name,
-    }).promise();
-
+    // leaking storage.  Note that s3.deleteObject is idempotent in AWS, but not in Google,
+    // and since we don't care about objects that couldn't be deleted, we swallow any errors.
+    try {
+      await this.s3.deleteObject({
+        Bucket: this.config.bucket,
+        Key: object.name,
+      }).promise();
+    } catch (error) {
+      // This can happen if object expired between the time the object metadata was retrieved
+      // and the delete was executed, or if there was a network blip that caused a retry to
+      // occur.  In no case do we really care if an object couldn't be deleted, as this
+      // expiration process is only a backstop to e.g. a bucket level expiration policy.
+    }
     return true;
   }
 }
 
-const getBucketRegion = async ({ bucket, ...credentials }) => {
-  const s3 = new aws.S3(credentials);
+const getBucketRegion = async ({ bucket, endpoint, ...credentials }) => {
+  let options = credentials;
+  if (endpoint) {
+    options.endpoint = new aws.Endpoint(endpoint);
+  }
+  const s3 = new aws.S3(options);
   const { LocationConstraint } = await s3.getBucketLocation({
     Bucket: bucket,
   }).promise();
--- a/generated/docs-search.json
+++ b/generated/docs-search.json
@@ -372,9 +372,9 @@
   },
   {
     "element": "h3",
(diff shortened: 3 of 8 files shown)


djmitche:
> For now, we've left the backend name as aws, rather than creating a new one called gcs or google, since the code is identical. Our assumption is that a running taskcluster service would use only one backend at a time, as a configuration setting, rather than uploading some objects to s3 and some to gcs. This is also because the object service API does not provide a means for a request to specify a specific backend, so our assumption is that all object service requests will be serviced by the same backend, which would be a deployment configuration setting.

This doesn't sound quite right -- can you check in with the existing docs and code and confirm?  We definitely support multiple backends :)

petemoore:
> > For now, we've left the backend name as aws, rather than creating a new one called gcs or google, since the code is identical. Our assumption is that a running taskcluster service would use only one backend at a time, as a configuration setting, rather than uploading some objects to s3 and some to gcs. This is also because the object service API does not provide a means for a request to specify a specific backend, so our assumption is that all object service requests will be serviced by the same backend, which would be a deployment configuration setting.
> 
> This doesn't sound quite right -- can you check in with the existing docs and code and confirm? We definitely support multiple backends :)

I couldn't see anything in https://github.com/taskcluster/taskcluster/blob/main/services/object/schemas/v1/upload-object-request.yml - am I looking in the wrong place?

djmitche:
Yeah, configuration for backends is a deployment configuration -- https://github.com/taskcluster/taskcluster/blob/main/ui/docs/manual/deploying/object-service.mdx

djmitche (changes requested):
That's pretty simple :D

This will need some detail added to the deployment configuration:
 * how to set up the buckets (including any necessary policy)
 * how to acquire the credentials
 * how to configure the backend

I think it makes sense to configure a GCS bucket with `{backendType: "aws", ..}` as it's pretty honest about what's happening with the compatibility layer.  If we eventually need to implement dedicated functionality, we can add a `gcs` backendType without any compatibility worries.

As for testing, the comment mentions helper methods for creating buckets.  Creating buckets as part of testing has gotten us in trouble before, as it's often not an immediate operation (Azure) and there are low limits on the number of buckets allowed (S3).  For the AWS backend, I used our existing buckets, with comments indicating how they were set up, and with credentials in [the secret](https://community-tc.services.mozilla.com/secrets/project%2Ftaskcluster%2Ftesting%2Ftaskcluster-object).  The key is to use unique object names so that parallel test runs don't interfere with one another.  I suspect a similar approach could work for GCS.  Also of note, I only use one (public) bucket for AWS, having verified that S3 checks URL signatures even for public buckets.  Assuming GCS has the same behavior, that might reduce the setup burden.  I think if you create the bucket in the `taskcluster-dev` project then we can all have access to it without any further setup.

changelog/issue-3718.md
```diff
@@ -0,0 +1,5 @@
+audience: deployers
+level: major
```
djmitche, line 2: I think this would be `silent` per [this doc](https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/changelog.md#levels).  It would be `minor` for "New feature users want to know about and take advantage of" except that the object service is disabled in deployments so users can't take advantage of it (yet).

jwhitlock (commented):
Thanks @petemoore and @helfi92, that's an impressively small list of changes for GCS support!

I think this would work until we need to support something like multi-part uploads, when we'd need to switch to GCS-specific functions. But, we can implement ``GCSBackend`` when we get to it.

I agree with @djmitche, some documentation is needed to complete this change.

petemoore:
> > For now, we've left the backend name as aws, rather than creating a new one called gcs or google, since the code is identical. Our assumption is that a running taskcluster service would use only one backend at a time, as a configuration setting, rather than uploading some objects to s3 and some to gcs. This is also because the object service API does not provide a means for a request to specify a specific backend, so our assumption is that all object service requests will be serviced by the same backend, which would be a deployment configuration setting.
> 
> This doesn't sound quite right -- can you check in with the existing docs and code and confirm? We definitely support multiple backends :)

Ah indeed, thanks for pointing that out. I decided on reflection to create an AwsBackend type, and a GoogleBackend type, to match our naming convensions elsewhere. The implementation is currently the same, and by virtue of the two new classes extending an underlying S3 class, which contains the commonality (which is currently everything). Maybe this will make it easier to add features to the google backend without affecting aws backend in future. So now both should be supported.

petemoore:
> That's pretty simple :D
> 
> This will need some detail added to the deployment configuration:
> 
>     * how to set up the buckets (including any necessary policy)
> 
>     * how to acquire the credentials
> 
>     * how to configure the backend
> 
> 

I've added a link to the Google docs for setting up a bucket. Regarding policy, I'm currently researching the exact permissions that are required.

> I think it makes sense to configure a GCS bucket with `{backendType: "aws", ..}` as it's pretty honest about what's happening with the compatibility layer. If we eventually need to implement dedicated functionality, we can add a `gcs` backendType without any compatibility worries.

In the end I created a second type (GoogleBackend) to make it clear that GCS is the destination and not S3. Hope that is also ok.

> 
> As for testing, the comment mentions helper methods for creating buckets. Creating buckets as part of testing has gotten us in trouble before, as it's often not an immediate operation (Azure) and there are low limits on the number of buckets allowed (S3). For the AWS backend, I used our existing buckets, with comments indicating how they were set up, and with credentials in [the secret](https://community-tc.services.mozilla.com/secrets/project%2Ftaskcluster%2Ftesting%2Ftaskcluster-object). The key is to use unique object names so that parallel test runs don't interfere with one another. I suspect a similar approach could work for GCS. Also of note, I only use one (public) bucket for AWS, having verified that S3 checks URL signatures even for public buckets. Assuming GCS has the same behavior, that might reduce the setup burden. I think if you create the bucket in the `taskcluster-dev` project then we can all have access to it without any further setup.

Thanks for this! I've now created a `taskcluster-test` bucket under the `taskcluster-dev` project with basic set up (a blanket 2 hour expiration for all objects, global bucket permissions with file-specific permissions disabled). Am testing.

changelog/issue-3718.md
petemoore: Fixed, thanks!

services/object/src/backends/aws.js
```diff
@@ -64,19 +71,29 @@ class AwsBackend extends Backend {
...
-    }).promise();
-
+    // leaking storage.  Note that s3.deleteObject is idempotent in AWS, but not in Google,
+    // and since we don't care about objects that couldn't be deleted, we swallow any errors.
+    try {
+      await this.s3.deleteObject({
+        Bucket: this.config.bucket,
+        Key: object.name,
+      }).promise();
+    } catch (error) {
```
djmitche, line 81: We probably do want to check that this is `NoSuchKey`, so that other errors do bubble through properly.

services/object/src/backends/aws.js
```diff
@@ -64,19 +71,29 @@ class AwsBackend extends Backend {
...
+      // This can happen if object expired between the time the object metadata was retrieved
+      // and the delete was executed, or if there was a network blip that caused a retry to
```
djmitche, line 83: I don't understand this comment -- when do we retrieve object metadata?  I think the explanation in the comment above is probably sufficient..

djmitche (commented):
I'll fix up the notes below and push a commit.

djmitche (approved):
(approved assuming that's green)

PR #4042, opened 2020-12-04
Fix schema for the simple download method

djmitche:
The schema included a `details` object, which is not necessary for this
method and was not what the backends returned.


jwhitlock (approved):
Thanks @djmitche!

PR #4060, opened 2020-12-08
Fix GitHub service producing errors without the anonymous role

### LINKED ISSUE #4059: GitHub service fails to fetch checks metadata without anonymous role
**Describe the bug**
When a GitHub PR is created with `checks-v1` an error is posted: 
```
Uh oh! Looks like an error! Details
Failed to get your artifact.
```
(The task in question has `queue:get-artifact:public/github/*`.)

The error reported to the log is:
```
"Error: cannot GET /api/queue/v1/task/NHdnST8MRVqo6lrUn5Njcw/runs/0/artifacts/public%2Fgithub%2FcustomCheckRunAnnotations.json (403)
    at Response.toError (/app/node_modules/superagent/lib/node/response.js:98:13)
    at ResponseBase._setStatusProperties (/app/node_modules/superagent/lib/response-base.js:119:48)
    at new Response (/app/node_modules/superagent/lib/node/response.js:44:8)
    at Request._emitResponse (/app/node_modules/superagent/lib/node/index.js:930:18)
    at /app/node_modules/superagent/lib/node/index.js:1127:42
    at IncomingMessage.<anonymous> (/app/node_modules/superagent/lib/node/parsers/json.js:22:7)
    at IncomingMessage.emit (events.js:326:22)
    at IncomingMessage.EventEmitter.emit (domain.js:483:12)
    at endReadableNT (_stream_readable.js:1241:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)"
```

The GitHub service now uses a non-anonymous client for fetching things like the checks body but calls `.buildUrl` instead of `.buildSignedUrl` which looks like it results in 403s when fetching the metadata artefacts.

Additionally, some of the extra information is missing from the errors (I think because the result of a `.concat()` isn't being used).

**To Reproduce**

1. Remove all scopes from anonymous.
2. Start a Task via GitHub with `reporting: checks-v1`
3. Notice that Taskcluster posts an error onto your commit.

**Expected behavior**
GitHub service should use its own credentials (masked against task scopes) to fetch the GitHub-specific metadata.

**Taskcluster version**
v39.0.0

This looks like a straight-forward fix, I'll get a PR in shortly.

ricky26:
Github Bug/Issue: Fixes #4059

This changes `requestArtifact` to use `buildSignedUrl` so that it works properly without the anonymous role and also fixes some of the error body being omitted.

### DIFF
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -458,24 +458,24 @@ async function requestArtifact(artifactName, { taskId, runId, debug, instGithub,
     const limitedQueueClient = this.queueClient.use({
       authorizedScopes: scopes,
     });
-    const url = limitedQueueClient.buildUrl(limitedQueueClient.getArtifact, taskId, runId, artifactName);
+    const url = limitedQueueClient.buildSignedUrl(limitedQueueClient.getArtifact, taskId, runId, artifactName);
     const res = await utils.throttleRequest({ url, method: 'GET' });
 
     if (res.status >= 400 && res.status !== 404) {
       let errorMessage = "Failed to get your artifact.\n";
       switch (res.status) {
         case 403:
-          errorMessage.concat("Make sure your artifact is public. See the documentation on the artifact naming.");
+          errorMessage = errorMessage.concat("Make sure your artifact is public. See the documentation on the artifact naming.");
           break;
         case 404:
-          errorMessage.concat("Make sure the artifact exists, and there are no typos in its name.");
+          errorMessage = errorMessage.concat("Make sure the artifact exists, and there are no typos in its name.");
           break;
         case 424:
-          errorMessage.concat("Make sure the artifact exists on the worker or other location.");
+          errorMessage = errorMessage.concat("Make sure the artifact exists on the worker or other location.");
           break;
         default:
           if (res.response && res.response.error & res.response.error.message) {
-            errorMessage.concat(res.response.error.message);
+            errorMessage = errorMessage.concat(res.response.error.message);
           }
           break;
       }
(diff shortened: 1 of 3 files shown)


services/github/src/handlers.js
```diff
@@ -458,24 +458,24 @@ async function requestArtifact(artifactName, { taskId, runId, debug, instGithub,
...
-    const url = limitedQueueClient.buildUrl(limitedQueueClient.getArtifact, taskId, runId, artifactName);
+    const url = limitedQueueClient.buildSignedUrl(limitedQueueClient.getArtifact, taskId, runId, artifactName);
     const res = await utils.throttleRequest({ url, method: 'GET' });
 
     if (res.status >= 400 && res.status !== 404) {
       let errorMessage = "Failed to get your artifact.\n";
       switch (res.status) {
         case 403:
-          errorMessage.concat("Make sure your artifact is public. See the documentation on the artifact naming.");
+          errorMessage = errorMessage.concat("Make sure your artifact is public. See the documentation on the artifact naming.");
```
djmitche, line 468: Is this message still accurate?

djmitche (approved):

services/github/src/handlers.js
ricky26: Ah, interesting point. I guess we could reword that to something like 'Make sure your artifact is accessible.'?

'Make sure you have assigned appropriate scopes to your task and repo roles.'?

I'm really not sold on either of those being amazingly easy to understand. :sweat_smile: 

Is it worth adding the artifact path to the first message? (That was something which wasn't obvious when I was trying to track the cause down here...)

djmitche: "Make sure your task has the scope `..`" might be the most explicit, and also would imply the artifact name.  I'll merge this fix, if you want to adjust the message in a followup?

Issue #4054, opened 2020-12-07
'Expiration time' claim ('exp') must be a numeric value representing the future time at which the assertion expires


djmitche:
https://sentry.prod.mozaws.net/operations/taskcluster-community/issues/10211271/

This is coming from GitHub.  We're getting about 15 / minute.

djmitche:
This happened for 2 hours and then stopped, and hasn't happened in 16 hours.  It's possible this was a GitHub issue, but it also started just 2.75 hours after deploying 39.0.0 to community (deployment was at 20:15 on Dec 7).

Let's see if this recurs.

PR #4062, opened 2020-12-08
Downgrade sift

### LINKED ISSUE #4061: Task group actions are not displayed
**Describe the bug**
The 'more' button which usually appears in the bottom right when there are task group actions is no longer appearing.

**To Reproduce**

1. Start a decision task which produces actions with `kind: 'task', context: []`.
2. Wait for the task to complete.
3. Notice that the button to access the action does not appear.

**Expected behavior**
The actions should still be displayed.

**Taskcluster version**
v39.0.0

**Additional context**
It looks like this is caused by upgrading the `sift` library to v13. Negating all of the incoming queries (by wrapping with `{$not: x}` seems to work, alarmingly). Rolling back to v8 resolves the issue.

When viewing the decision task itself, the action is returned from `/graphql` but is not displayed.

djmitche:
See https://github.com/taskcluster/taskcluster/issues/4054

Github Bug/Issue: Fixes #4061

### DIFF
--- a/changelog/issue-4061.md
+++ b/changelog/issue-4061.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 4061
+---
+This version fixes an issue with the "actions" button not appearing for task groups.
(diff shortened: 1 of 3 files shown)


ricky26 (approved):
I tested with `^8.0.0`, but I believe `8.x` is identical (and it looks to have resolved to the same version). Thanks for this! :+1:

imbstack (approved):

PR #4075, opened 2020-12-09
enable auto-merge for top-level packages

### LINKED ISSUE #4011: Configure renovate to auto-merge non-UI non-major PRs
We don't have tests for the UI, so auto-merging PRs there is dangerous.  We can decide separately to just disable renovate entirely for the UI or continue to manually test them.

We have previously decided to use the "forking renovate" for TC because there was a security issue in the renovate app that would have allowed arbitrary code injection, and at the time we were auto-deploying from Git.  That issue was quickly fixed, giving us a lot of confidence in the security posture of Renovate at the time; the tool has since been acquired by a company with more resources and investment, increasing that confidence; and we no longer auto-deploy.  Also, I believe the branch rules on the repo would prevent renovate from pushing directly to `main`, requiring any malicious content to arrive in a PR where we would have a chance of spotting it.  All in all, I feel that puts renovate's risk level well below that of the JS ecosystem in general -- if I were going to try to compromise the TC source, it would be by compromising some dependency of TC and releasing it, rather than by compromising renovate itself.  And that's a risk we have (unhappily!) accepted forever.

djmitche:
I assume this will need further tweaking.  I'm not sure automerging will work with our current GitHub settings (in particular, I think renovate can't merge with pending reviews, so we may need to solve that).  And I'm not sure this is the right way to express this in renovate config anyway.  Hopefully the addition of `_comment` doesn't cause errors and helps understanding the intent.

Github Bug/Issue: Fixes #4011

### DIFF
--- a/package.json
+++ b/package.json
@@ -216,6 +216,7 @@
     "semanticCommits": "disabled",
     "packageRules": [
       {
+        "_comment": "packages which get very frequent updates are limited to once per week",
         "packagePatterns": [
           "^googleapis$",
           "^aws-sdk$",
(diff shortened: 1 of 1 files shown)


imbstack (approved):
Let's try it out!

PR #4091, opened 2020-12-10
Issue 4034: paginate artifacts better

### LINKED ISSUE #4034: Some tasks are unavailable while their artifacts are.
The task XATXkY3lQeymtCnN21v8JA doesn't exist for the task inspector (https://firefox-ci-tc.services.mozilla.com/tasks/XATXkY3lQeymtCnN21v8JA), but has available artifacts, such as https://firefoxci.taskcluster-artifacts.net/XATXkY3lQeymtCnN21v8JA/1/public/logs/live_backing.log

djmitche:
See individual commit comments for details.

Github Bug/Issue: Fixes #4034


db/versions/0057.yml
```diff
@@ -0,0 +1,47 @@
...
+          queue_artifacts.run_id,
+          queue_artifacts.name,
+          queue_artifacts.storage_type,
+          queue_artifacts.content_type,
+          queue_artifacts.details,
+          queue_artifacts.present,
+          queue_artifacts.expires
+        from queue_artifacts
+        where
+          (queue_artifacts.task_id = task_id_in or task_id_in is null) and
```
jwhitlock, line 31: This line took me a few times. I think it means "all artifacts if `task_id_in` is null, or just the task artifacts if `task_id_in` is set"

db/versions/0057.yml
```diff
@@ -0,0 +1,47 @@
...
+          (queue_artifacts.run_id = run_id_in or run_id_in is null) and
+          (queue_artifacts.expires < expires_in or expires_in is null) and
+          (after_task_id_in is null or
+            (queue_artifacts.task_id > after_task_id_in or
+              (queue_artifacts.task_id = after_task_id_in and
+                (queue_artifacts.run_id > after_run_id_in or
+                  (queue_artifacts.run_id = after_run_id_in and
+                    queue_artifacts.name > after_name_in
```
jwhitlock, line 39: and this monster is "for the next page, pick up where we left off, which may be in a really weird spot"

db/fns.md
```diff
@@ -1906,6 +1909,7 @@ All parameters must be supplied.
 * `create_queue_worker_type(provisioner_id_in text, worker_type_in text, expires_in timestamptz, last_date_active_in timestamptz, description_in text, stability_in text)` (compatibility guaranteed until v40.0.0)
 * `expire_queue_provisioners(expires_in timestamptz)` (compatibility guaranteed until v40.0.0)
 * `expire_queue_worker_types(expires_in timestamptz)` (compatibility guaranteed until v40.0.0)
+* `get_queue_artifacts(task_id_in text, run_id_in integer, expires_in timestamptz, page_size_in integer, page_offset_in integer)` (compatibility guaranteed until v41.0.0)
```
jwhitlock, line 1912: I didn't see that there's a deprecated function mechanism, I really like it.

changelog/issue-4034.md
```diff
@@ -0,0 +1,5 @@
+audience: deployers
+level: patch
+reference: issue 4034
+---
+The queue's artifact expiration crontask now uses a much more efficient query and should be able to keep up with the load.
```
jwhitlock, line 5: Do we usually point out new / deprecated functions, or omit it as implementation details?

libraries/postgres/README.md
```diff
@@ -314,9 +314,40 @@ An online migration may also be interrupted and replaced with an online downgrad
...
+    size: 1000, // optional, defaults to 1000
+  })) {
+    // ..do something with `row`
+  }
+}
+```
+
+### Offset / Limit-Based
+
+Many database functions which return many rows have `page_size int, page_offset int` as their last two arguments, and return a page of the given size at the given offset.
```
jwhitlock, line 347: Should we mention the deprecation in the docs?

libraries/postgres/src/util.js
```diff
@@ -81,26 +81,48 @@ exports.annotateError = (query, err) => {
...
+      }
+
       const next = async () => {
         if (rows.length > 0) {
           return { value: rows.shift(), done: false };
         }
         if (done) {
           // If this iterator has already "finished", just return that status
-          // without calling out to the DB again (and potentialyl returning more
+          // without calling out to the DB again (and potentially returning more
```
jwhitlock, line 120: ❤️

jwhitlock (approved):
Thanks @djmitche, I'm excited to see if this solves the pagination issues. The tests are excellent, covering the new and old pagination, including some cases around bad next page params.

db/versions/0057.yml
djmitche: Indeed, and I wish there was an easier way to represent this.  I looked for tuple comparison support (`(queue_artifatcts.task_id, queue_artifacts.run_id, queue_artifacts.name) > (after_task_id_in, after_run_id_in, after_name_in)`) but didn't find anything.

db/versions/0057.yml
djmitche: Correct.  I can add some comments

changelog/issue-4034.md
djmitche: Since it's internal to the services, it doesn't need mention in the changelog.

libraries/postgres/README.md
djmitche: I'm not sure.. I think the existing pagination is actually fine in a lot of cases, and as part of "stabilization" I don't want to kick off a long-term migration to index-based pagination.  That's why I wrote "preferred" :)

djmitche:
Also, I just realized that this fixes #3278 as well.  Even there, I wrote "preferred" so I don't think there's a need to keep an issue open to track migrating other pagination implementations to index-based.

PR #4094, opened 2020-12-10
Generate the generic-worker README in 'yarn generate'

### LINKED ISSUE #2981: Auto-generate generic-worker README.md as part of yarn generate
See [comment from #2854](https://github.com/taskcluster/taskcluster/pull/2854/files#r432045951) - the [`refresh_readme.sh`](https://github.com/taskcluster/taskcluster/blob/9d9885692504e26e1bb5b0f47cb655b8edf4cf47/workers/generic-worker/refresh_readme.sh) script should run as part of `yarn generate`.

djmitche:
This removes some specifics from the README:
 - the default for tasksDir (which differs per platform)
 - the enine used to generate the help
 - the version (which otherwise would need to be updated on release)

Github Bug/Issue: Fixes #2981


workers/generic-worker/usage.go
```diff
@@ -203,7 +201,7 @@ and reports back results to the queue.
           taskclusterProxyPort              Port number for taskcluster-proxy HTTP requests.
                                             [default: 80]
           tasksDir                          The location where task directories should be
-                                            created on the worker. [default: ` + fmt.Sprintf("%q", defaultTasksDir()) + `]
+                                            created on the worker. [default varies by platform]
```
jwhitlock, line 204: I guess this was necessary to generate the help text? I'm sad to lose the information (although, to be fair, I've never used it). Maybe some thing that matches and replaces `created on the worker. [default:.*`, similar to the first line replacement?

jwhitlock (approved):
Thanks @djmitche! Automation for the win! 🤖

workers/generic-worker/usage.go
djmitche: My thinking was that the information in the README was incorrect (and I have in the past wondered why it uses `/Users` on linux.. turns out it doesn't).  But it's a good point that the `--help` output can be platform-specific.  Let me see what I can do..

djmitche: (I only noticed this because running `refresh_readme.sh` locally changed `/Users` to `/home`)

jwhitlock (approved):
Thanks for keeping the default output!

Issue #3044, opened 2020-06-12
[UI] Enhance the empty state for the hooks view


helfi92:
**To Reproduce**

1. Go to http://localhost:5080/hooks

**Expected behavior**
If there are no hooks to display, it should say something other than "No items for search term".

**Actual behavior**
"No items for search term "

<img width="855" alt="Screen Shot 2020-06-12 at 1 35 04 PM" src="https://user-images.githubusercontent.com/3766511/84530829-89d75180-acb1-11ea-9c9f-eb177ec931a9.png">

**Taskcluster version**
v30.1.1

**Additional context**
This only happens when there are no hooks to display.

Manish-Giri:
@helfi92 May I take this up?

helfi92:
Hey! Go for it :)

codershona:
Hello Is this issue still open, can I work on this issue?

djmitche:
Sure, it's all yours!

codershona:
Thank you

djmitche:
So the issue here is that when there are no hooks at all (rather than a search that didn't return anything), the text is "No items for search term".  Which is confusing, because there is no search term.

So in the case where there are no hooks *and* no search term, let's display "No hooks are defined", instead.

rayyanQ:
I am a new contributor and would like to work on this issue, if it is still open.

djmitche:
Sure thing!  #4104 should have some extra context on what the desired state is here.

rayyanQ:
Thanks!

cerkiewny:
I think this issue is slightly outdated, I will include fix for that in my [PR](https://github.com/taskcluster/taskcluster/pull/4919) as it is changing a lot about the hook tree anyway

jwhitlock:
Fixed in PR #4921 and #4948.

PR #4104, opened 2020-12-14
enhance the empty state by using typography theme

codershona:
#  Enhance the empty state for the hooks view #3044

**Explanation of my method:**
I had tried to customize the theme by adding a component and to expect the behaviour I had to try added an updated hook method as when we try to get updated hooks pf contained in empty search.

### DIFF
--- a/ui/src/views/Hooks/ListHooks/index.jsx
+++ b/ui/src/views/Hooks/ListHooks/index.jsx
@@ -37,23 +37,29 @@ export default class ListHooks extends Component {
 
   handleHookSearchSubmit = hookSearch => {
     const query = parse(window.location.search.slice(1));
+    const { location, history } = this.props;
 
-    this.props.history.push({
-      search: stringify({
+    if (query.search !== hookSearch) {
+      const newQuery = {
         ...query,
         search: hookSearch,
-      }),
-    });
+
+      };
+      history.push({
+        search: stringify(newQuery, {addQueryPrefix: true}),
+      });
+    }
   };
 
   render() {
     const {
       classes,
       description,
+      location,
       data: { loading, error, hookGroups },
     } = this.props;
     const query = qs.parse(this.props.location.search.slice(1));
-    const hookSearch = query.search;
+    const hookSearch = query.search ? query.search : '';
     const tree = hookGroups
       ? hookGroups.map(group => ({
           value: group.hookGroupId,
@@ -85,11 +91,11 @@ export default class ListHooks extends Component {
             key={hookSearch}
             defaultExpanded={Boolean(hookSearch)}
             listItemProps={{ color: classes.listItemProps }}
-            searchTerm={hookSearch || null}
+            searchTerm={hookSearch}
             softSearch
             tree={tree}
             onEmptySearch={
-              <Typography variant="h3" component="h3" color="primary">
+              <Typography variant="h3" color="secondary">
                 No hooks are defined {hookSearch}
               </Typography>
             }
--- a/ui/src/views/Hooks/ListHooks/index.jsx
+++ b/ui/src/views/Hooks/ListHooks/index.jsx
@@ -94,9 +94,9 @@ export default class ListHooks extends Component {
             tree={tree}
             onEmptySearch={
               <Typography
-                variant="subtitle1"
+                variant="h3"
                 placeholder="Updated Hook contains"
-                color="inherit"
+                color="primary"
                 onClick={this.handleHookUpdateSearch}>
                 No items for search term {hookSearch}
               </Typography>
--- a/ui/src/views/Hooks/ListHooks/index.jsx
+++ b/ui/src/views/Hooks/ListHooks/index.jsx
@@ -35,10 +35,6 @@ export default class ListHooks extends Component {
     this.props.history.push('/hooks/create');
   };
 
-  handleHookUpdateSearch = () => {
-    this.props.history.push('/hooks/update');
-  };
-
   handleHookSearchSubmit = hookSearch => {
     const query = parse(window.location.search.slice(1));
 
@@ -93,12 +89,8 @@ export default class ListHooks extends Component {
             softSearch
             tree={tree}
             onEmptySearch={
-              <Typography
-                variant="h3"
-                placeholder="Updated Hook contains"
-                color="primary"
-                onClick={this.handleHookUpdateSearch}>
-                No items for search term {hookSearch}
+              <Typography variant="h3" component="h3" color="primary">
+                No hooks are defined {hookSearch}
               </Typography>
             }
             Link={Link}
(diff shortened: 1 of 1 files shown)


ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -35,6 +35,10 @@ export default class ListHooks extends Component {
     this.props.history.push('/hooks/create');
   };
 
+  handleHookUpdateSearch = () => {
+    this.props.history.push('/hooks/update');
```
djmitche, line 39: What is this route for?

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -89,7 +93,11 @@ export default class ListHooks extends Component {
             softSearch
             tree={tree}
             onEmptySearch={
-              <Typography variant="subtitle1">
+              <Typography
+                variant="subtitle1"
+                placeholder="Updated Hook contains"
```
djmitche, line 98: What does this mean?  I believe when this is displayed, there are no hooks, so nothing has been "updated"..

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -89,7 +93,11 @@ export default class ListHooks extends Component {
...
+                color="inherit"
```
djmitche, line 99: Why is this required?

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -89,7 +93,11 @@ export default class ListHooks extends Component {
...
+                onClick={this.handleHookUpdateSearch}>
                 No items for search term {hookSearch}
```
djmitche, line 101: This text looks different from what #3044 shows. A bit of exploring the git history shows this issue already got fixed, in #2779.  So, maybe there's nothing to change here?

codershona: So where I am a bit that where could I change as I thought it was about to enhance the empty state by using the color component in the typography theme.

djmitche: Nope, the expected behavior doesn't say anything about color:
>  If there are no hooks to display, it should say something other than "No items for search term".

djmitche: Oh, I've misunderstood the issue, too!

I'll update the issue itself with a better explanation.

codershona: ok.

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -89,8 +89,8 @@ export default class ListHooks extends Component {
             softSearch
             tree={tree}
             onEmptySearch={
-              <Typography variant="subtitle1">
-                No items for search term {hookSearch}
+              <Typography variant="h3" component="h3" color="primary">
+                No hooks are defined {hookSearch}
```
djmitche, line 93: This is looking better, but now when there *is* a search term, for example `proj-taskcluster`, it's going to say something like
> No hooks are defined proj-taskcluster

which doesn't make a lot of sense.  How could it be made to say 

> No hooks are defined

when there's no hooks defined and

> No hooks found matching "proj-taskcluster"

When there's a search term but no matching results?

djmitche:
We're all back to work now so I'll have a look at this soon.  Sorry for the delay!

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -37,23 +37,29 @@ export default class ListHooks extends Component {
...
+    if (query.search !== hookSearch) {
+      const newQuery = {
         ...query,
         search: hookSearch,
-      }),
-    });
+
+      };
+      history.push({
+        search: stringify(newQuery, {addQueryPrefix: true}),
```
djmitche, line 49: I'm curious why this was required?  This causes `stringify` to add a `?` to the beginning of the query, but [history already does that](https://github.com/ReactTraining/history/blob/3f69f9e07b0a739419704cffc3b3563133281548/modules/LocationUtils.js#L18-L23).

It looks like history is documented as expecting the `?` to be present so this is not a bad change to make, but doesn't address the issue this PR is solving.

ui/src/views/Hooks/ListHooks/index.jsx
djmitche: It looks like this still only ever says "No hooks are defined" -- I don't see the other strings anywhere in the PR.

djmitche (commented):
Good progress here :)

If you run `yarn lint --fix, it will fix the stylistic issues and then the checks on the PR will make a test version of the site, which we can use to check out your changes.  As it is, because of the lint issues, those checks are failing.

ui/src/views/Hooks/ListHooks/index.jsx
codershona: Hello, @djmitche  I am a bit confused will the string will look like this, 

`No hooks are defined ${hooksearch}`

ui/src/views/Hooks/ListHooks/index.jsx
codershona: Here if i try to use this **logic** will it would be correct to use it,

```
handleHookSearchSubmit = hookSearch => {
    const query = parse(window.location.search.slice(1));
    const {
        location,
        history
    } = this.props;

    if (query.search !== hookSearch) {
        const newQuery = {
            ...query,
            search: hookSearch
        }
        else {
            location.search = '?' + location.search;
        }

    }
}
```

djmitche: I think it would be correct to omit the change entirely in this PR, and focus on fixing the issue in #3044.

ui/src/views/Hooks/ListHooks/index.jsx
djmitche: That would have the same effect as what's already written.

It might be good to make a diagram or notes of what the different states are, and what the component should look like in each of those states.  Then think of a way to put the component in each of those states in your browser, so you can verify whether your changes have achieved the desired result.

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -91,12 +85,12 @@ export default class ListHooks extends Component {
...
-            searchTerm={hookSearch}
+            searchTerm={hookSearch || null}
             softSearch
             tree={tree}
             onEmptySearch={
-              <Typography variant="h3" color="secondary">
-                No hooks are defined {hookSearch}
+              <Typography variant="subtitle1">
+                 {`No hooks are defined ${hookSearch}`}
               </Typography>
```
codershona, line 94: I had added a string logic here but I am still confused about the no display hooks logics.

djmitche:
Let's take a bit of a step back here.  What is the issue looking for?

* When _____ then the component should ______
* When _____ then the component should ______

Once you answer that, we can take the next step.

djmitche:
I'm going to close this PR -- we can continue this conversation on #3304.

PR #4106, opened 2020-12-14
remove redundant hawk requirement in devDependencies

djmitche:
Spotted this after a warning from `yarn`.


jwhitlock (approved):
Nice catch @djmitche!

The check "meta-generate" failed, not sure if it is an intermittent failure or not.

PR #4107, opened 2020-12-14
Remove github octokit throttling until we understand why it is doing what it is doing

imbstack:
Quickest fix we can think of to get this out of the way


djmitche (approved):
Looks good.  Let's get a bug filed for the throttling plugin, too?  1 request per second is 💯% bogus, especially without documenting it.

djmitche:
(needs a yarn generate)

PR #4109, opened 2020-12-15
Use `base64.encodebytes` in Python 3.

jschwartzentruber:
This replaces `base64.encodestring` which was [deprecated in 3.1 and removed in 3.9](https://bugs.python.org/issue39351). This is the only change I had to make to run `tc-admin` with Python 3.9.

### DIFF
--- a/changelog/FuHmLyHZS--UbHLntDvDRw.md
+++ b/changelog/FuHmLyHZS--UbHLntDvDRw.md
@@ -1,0 +1,4 @@
+audience: developers
+level: silent
+---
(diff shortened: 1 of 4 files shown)


djmitche (changes requested):
Thanks!!

It looks like this will need a [changelog](https://github.com/taskcluster/taskcluster/blob/master/dev-docs/best-practices/changelog.md).  Also, please add a py3.9 test suite in `taskcluster/ci/client/kind.yml`.

djmitche (approved):
Looks good!

djmitche:
(the changelog error is something we'll need to fix)

PR #4111, opened 2020-12-15
Handle artifacts better in the UI, without huge URLs that time out

### LINKED ISSUE #3878: log viewer URLs are huge and temporary
Observe:

https://stage.taskcluster.nonprod.cloudops.mozgcp.net/tasks/IN-ZKBSzRxevRR_9xolC9Q/runs/0/logs/https%3A%2F%2Fstage.taskcluster.nonprod.cloudops.mozgcp.net%2Fapi%2Fqueue%2Fv1%2Ftask%2FIN-ZKBSzRxevRR_9xolC9Q%2Fruns%2F0%2Fartifacts%2Fpublic%252Flogs%252Flive.log%3Fbewit%3DbW96aWxsYS1hdXRoMC9hZHxNb3ppbGxhLUxEQVB8ZG1pdGNoZWxsXDE2MDQ5NDQyNjdceHJZUFBJVzg4UnFLYVgzbkd3Y0F6dEsyNEM2ZkZjblkycXNadTdpRVhLQT1cZXlKalpYSjBhV1pwWTJGMFpTSTZleUoyWlhKemFXOXVJam94TENKelkyOXdaWE1pT2xzaVlYTnpkVzFsT20xdmVtbHNiR0V0WjNKdmRYQTZSR1YyVTJWeWRtbGpaWE5YYVd0cElpd2lZWE56ZFcxbE9tMXZlbWxzYkdFdFozSnZkWEE2UjNKaGNHaHpRV1J0YVc0aUxDSmhjM04xYldVNmJXOTZhV3hzWVMxbmNtOTFjRHBKYm5SeVlXNWxkRmRwYTJraUxDSmhjM04xYldVNmJXOTZhV3hzWVMxbmNtOTFjRHBTWld4RmJtZFhhV3RwSWl3aVlYTnpkVzFsT20xdmVtbHNiR0V0WjNKdmRYQTZVM1JoZEhORVlYTm9ZbTloY21RaUxDSmhjM04xYldVNmJXOTZhV3hzWVMxbmNtOTFjRHBUZVhOQlpHMXBibGRwYTJraUxDSmhjM04xYldVNmJXOTZhV3hzWVMxbmNtOTFjRHBoWTNScGRtVmZjMk50WDJ4bGRtVnNYekVpTENKaGMzTjFiV1U2Ylc5NmFXeHNZUzFuY205MWNEcGhZM1JwZG1WZmMyTnRYMnhsZG1Wc1h6SWlMQ0poYzNOMWJXVTZiVzk2YVd4c1lTMW5jbTkxY0RwaFkzUnBkbVZmYzJOdFgyeGxkbVZzWHpNaUxDSmhjM04xYldVNmJXOTZhV3hzWVMxbmNtOTFjRHBoYkd4ZmJHUmhjRjkxYzJWeWN5SXNJbUZ6YzNWdFpUcHRiM3BwYkd4aExXZHliM1Z3T21Gc2JGOXpZMjFmYkdWMlpXeGZNU0lzSW1GemMzVnRaVHB0YjNwcGJHeGhMV2R5YjNWd09tRnNiRjl6WTIxZmJHVjJaV3hmTWlJc0ltRnpjM1Z0WlRwdGIzcHBiR3hoTFdkeWIzVndPbUZzYkY5elkyMWZiR1YyWld4Zk15SXNJbUZ6YzNWdFpUcHRiM3BwYkd4aExXZHliM1Z3T21GMFpXRnRJaXdpWVhOemRXMWxPbTF2ZW1sc2JHRXRaM0p2ZFhBNllYZHpYek15TURRMk5ESXdOVE00Tmw5eVpXRmtYMjl1YkhraUxDSmhjM04xYldVNmJXOTZhV3hzWVMxbmNtOTFjRHBpZFdsc1pIUmxZVzBpTENKaGMzTjFiV1U2Ylc5NmFXeHNZUzFuY205MWNEcGpiRzkxWkdobFlXeDBhQzF6ZEdGdVpHRnlaQ0lzSW1GemMzVnRaVHB0YjNwcGJHeGhMV2R5YjNWd09tVm5aVzVqYVdGZmRYTWlMQ0poYzNOMWJXVTZiVzk2YVd4c1lTMW5jbTkxY0RwcGJuWmxiblJ2Y25raUxDSmhjM04xYldVNmJXOTZhV3hzWVMxbmNtOTFjRHBwYm5abGJuUnZjbmxmWW5WcGJHUWlMQ0poYzNOMWJXVTZiVzk2YVd4c1lTMW5jbTkxY0RwdGNIUjJjRzRpTENKaGMzTjFiV1U2Ylc5NmFXeHNZUzFuY205MWNEcHVZV2RwYjNOaFpHMXBiaUlzSW1GemMzVnRaVHB0YjNwcGJHeGhMV2R5YjNWd09tOW1abWxqWlY5eVpXMXZkR1VpTENKaGMzTjFiV1U2Ylc5NmFXeHNZUzFuY205MWNEcHdhRzl1WldKdmIydGZZV05qWlhOeklpd2lZWE56ZFcxbE9tMXZlbWxzYkdFdFozSnZkWEE2YzNCbFkyOXdjeUlzSW1GemMzVnRaVHB0YjNwcGJHeGhMV2R5YjNWd09uUmhZbXhsWVhWZmRYTmxjbk1pTENKaGMzTjFiV1U2Ylc5NmFXeHNZUzFuY205MWNEcDBaV0Z0WDIxdlkyOGlMQ0poYzNOMWJXVTZiVzk2YVd4c1lTMW5jbTkxY0RwMFpXRnRYM1JoYzJ0amJIVnpkR1Z5SWl3aVlYTnpkVzFsT20xdmVtbHNiR0V0WjNKdmRYQTZkRzl2YkhSdmIyeGxaR2wwYjNJaUxDSmhjM04xYldVNmJXOTZhV3hzWVMxbmNtOTFjRHAyY0c1ZllXUnRhVzRpTENKaGMzTjFiV1U2Ylc5NmFXeHNZUzFuY205MWNEcDJjRzVmWTI5eWNDSXNJbUZ6YzNWdFpUcHRiM3BwYkd4aExXZHliM1Z3T25ad2JsOWtaV1poZFd4MElpd2lZWE56ZFcxbE9tMXZlbWxzYkdFdFozSnZkWEE2ZG5CdVgyZHBkR2xmZEdOZmMyVmpjbVYwYzE5eWR5SXNJbUZ6YzNWdFpUcHRiM3BwYkd4aExXZHliM1Z3T25ad2JsOXVZV2RwYjNOZmQyVmlJaXdpWVhOemRXMWxPbTF2ZW1sc2JHRXRaM0p2ZFhBNmRuQnVYM0psYkdWdVp5SXNJbUZ6YzNWdFpUcHRiM3BwYkd4cFlXNXpMV2R5YjNWd09taGxjbTlyZFMxdFpXMWlaWEp6SWl3aVlYTnpkVzFsT214dloybHVMV2xrWlc1MGFYUjVPbTF2ZW1sc2JHRXRZWFYwYURBdllXUjhUVzk2YVd4c1lTMU1SRUZRZkdSdGFYUmphR1ZzYkNKZExDSnpkR0Z5ZENJNk1UWXdORGswTWpFd016TTJOQ3dpWlhod2FYSjVJam94TmpBME9UUXpPVEF6TXpZMExDSnpaV1ZrSWpvaVVsZGxMWFp6TkVkU2RHMXJjRkZJWVUxZk1VRmtaMFZPYWxOamVWOWpVVUpwWW5WV2FYcDVPVTV4YmtFaUxDSnphV2R1WVhSMWNtVWlPaUp0TDFKSFlrMDVPRmxwVjNWTksxUjNPVWRSZVZWbVUxTk9XbFZ6UjFSbU5XRjJhVk5FYzBVMllsaEJQU0lzSW1semMzVmxjaUk2SW5OMFlYUnBZeTkwWVhOclkyeDFjM1JsY2k5M1pXSXRjMlZ5ZG1WeUluMTk

The problem is that we're passing URLs via react-router, and those URLs include a bewit.  The bewit is only valid for a short length of time, so even if someone did paste this link into a bug or issue or whatever, it would be useless soon after.

From first principles, the information required for a log is
 * taskId
 * runId
 * artifact name

So, let's change things around such that only those three parameters appear in the URL.

That will require a change to how the logs are fetched.  That can either be via a new GraphQL interface, or by generating the signed URL within the UI code using taskcluster-client-web.

djmitche:
Github Bug/Issue: Fixes #3878

See the comments inline on the commits.


djmitche:
So, including tc-client-web from ui is not working very well.  Something calls path.resolve, so that despite the symlinks webpack can't find the client's dependencies because it's looking in the parent directories of `./client/client-web` for `node_modules` directories, which don't exist.  I bet webpack-v5 fixes this :(

djmitche:
I think this will need to wait for #4127, then a release of 39.1.3 or later.  Unless anyone has any better ideas of how to require a library elsewhere in the monorepo using webpack.

djmitche:
> I think this will need to wait for #4127, then a release of 39.1.3 or later. Unless anyone has any better ideas of how to require a library elsewhere in the monorepo using webpack.

I'd love to hear better ideas :)

ui/src/views/Tasks/routes.js
```diff
@@ -59,12 +59,12 @@ export default path => [
   },
   {
     component: TaskLog,
-    path: `${path}/:taskId/runs/:runId/logs/live/:logUrl`,
+    path: `${path}/:taskId/runs/:runId/logs/live/:name+`,
```
jwhitlock, line 62: What does the trailing `+` in `/:name+` do?

ui/package.json
```diff
@@ -89,6 +89,7 @@
     "shebang-loader": "^0.0.1",
     "snake-case": "^3.0.3",
     "subscriptions-transport-ws": "^0.9.16",
+    "taskcluster-client-web": "^39.2.0",
```
jwhitlock, line 92: This seems OK to me. The client feels like a utility library, why not install it rather than sneak the code in via the monorepo.

ui/src/components/TaskRunsCard/index.jsx
```diff
@@ -247,8 +262,50 @@ export default class TaskRunsCard extends Component {
...
+    const {
+      icon: Icon,
+      isLogFile,
+      url,
+      handleArtifactClick,
+    } = this.getArtifactInfo(artifact);
+
+    return (
+      <TableRow
+        key={name}
```
jwhitlock, line 277: If I'm reading this correctly, the `key` value is a lot shorter. Did you intend to change this key from `run-artifact-${run.taskId}-${run.runId}-${artifact.name}` to `${artifact.name}`. Also I don't know what `key` does.

jwhitlock (approved):
Thanks @djmitche, the code looks better after the refactor, and the cached signed URLs are interesting. I didn't try this code (and I'm not familiar with the issue it is solving), but just tried  to read and understand the changes.

ui/src/components/TaskRunsCard/index.jsx
djmitche: It's a React thing -- as long as each element in this array has a different key, everything's fine.  So it didn't need to be so long.

ui/package.json
djmitche: Cool, thanks.  It feels weird to me that we'll almost always be pulling in an older version of this library (so TC 40.0.0 will use tc-client-web 39.x.y).  But aside from being weird it doesn't really matter.

ui/src/views/Tasks/routes.js
djmitche: It allows it to contain `/` characters, so the URLs can be nice-looking and avoid excessive urlencoding.  In double-checking where this is encoded, I see that it correctly uses encodeURI (which does not escape `/`), but that it also leaves in the long comment about react-router, which is no longer relevant.  I'll remove that comment and push.

djmitche: Oh, ignore that, it's addressed in another commit.

ui/src/components/TaskRunsCard/index.jsx
djmitche: https://reactjs.org/docs/lists-and-keys.html#keys

ui/package.json
ricky26: I know I'm late but have you considered NPM's relative syntax?
```json
"taskcluster-client-web": "file:../clients/client-web",
```

That way you get the same behaviour as the versioned import with the one downside that it doesn't always rely on the stable version of the library.

djmitche: I tried that, and tc-client-web does not find its requirements unless `yarn install` has been run in that directory.  I think that's due to Node's (and webpack's) search pattern for `node_modules` -- it never looks in `ui/node_modules`.  I missed this in my first push to this PR because I had a `clients/client-web/node_modules` laying around from development in that directory, but it did not exist in CI.

I tried `link:../clients/client-web` as well, which in theory fixes this -- it makes a link from `ui/node_modules/taskcluster-client-web` -> `../../clients/client-web`.  The `yarn install` process correctly installs taskcluster-client-web's deps in `ui/node_modules`, and iirc from my experimentation Node correctly finds them.  But, webpack does not -- it seems that webpack is using `path.resolve` or similar to translate that symlink into a direct path, and then seeing the same issue as with `file:..`.

It's possible that using a newer version of Webpack would fix this.  It's also possible I am missing some setting or other config that would fix it.  I tried various permutations of aliases with no success.

ricky26: Ah yeah, I forgot that it doesn't recursively install, I've run into that before. What a shame. :neutral_face:

PR #4112, opened 2020-12-15
Update object service enpoint names to be more descriptive

### LINKED ISSUE #3940: Choose new name for object service "full download"
As it stands we're going to roughly have a `download` method and a `simpleDownload` method which correspond to an endpoint that fetches a bunch of data about an object and one that simple does a download.

Let's change these names to `?????` and `download` to correspond to what they do.

One note @djmitche brings up is that the `????` actually might take some action on behalf of the user (copying an object around in the cloud I assume) so just a simple `fetchDetails` doesn't cover everything.

This bug is for bikeshedding about this to our heart's content.

imbstack:
Silent changelog just because this is an "unreleased" service. I can add a real one if that's not a good assumption.

Github Bug/Issue: Fixes #3940

### DIFF
--- a/changelog/issue-3940.md
+++ b/changelog/issue-3940.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3940
+---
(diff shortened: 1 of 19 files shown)


djmitche (approved):

djmitche:
Typo: "enpoints" in the commit message.

Typos are my "thing".. back off! ;)

PR #4116, opened 2020-12-15
add site-specific details for notify

### LINKED ISSUE #3739: Use SiteSpecific in docs to document notify, github plugins
Let's include in the docs the bot names for the notify service in the docs, when they are available.

This would mean adding params to `ui/src/components/SiteSpecific/index.jsx` and documenting them, then including that info in the deployed configs.

djmitche:
This should help users find the bots.

Note that the github bot is already adequately documented with SiteSpecific -- there's no need for users to know its github username.

Github Bug/Issue: Fixes #3739.

### DIFF
--- a/changelog/issue-3739.md
+++ b/changelog/issue-3739.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3739
+---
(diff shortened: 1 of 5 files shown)


djmitche:
Red x's are just the mac generic-worker instance -- it's so far behind that its tasks are reaching their deadline.  So, unrelated to this PR.

imbstack (approved):
ah sweet

PR #4121, opened 2020-12-16
Drop support for VNC in interactive tasks.

### LINKED ISSUE #3542: Remove support for v2 shells/displays from UI
Per #3541, we only actually use `v=1` shells and displays.  Removing this support from the UI will allow us to remove a dependency on `ws-shell` and its sub-dependencies, which will at least lighten the bundle and give us fewer packages to keep updated.

I don't think there's any sense in keeping v2 -- nobody who still works here knows how it works, and anyway it would only be supported in docker-worker which is deprecated, so there's zero chance that we'll ever implement the worker side.

djmitche:
Github Bug/Issue: Fixes #3542, which has the context.


djmitche:
I verified the UI still works for shell connections.

imbstack (approved):
red, red, red

djmitche:
I need to build a docker-worker image from this and test that out..

djmitche:
My dev env isn't set up with the right DNS and public IPs and whatnot, but this generates a `shell.html` artifact as expected.

djmitche:
OK, confirmed with a test worker pool in community.  It works!

PR #4122, opened 2020-12-16
use taskQueueId internally to the queue (functions and code for tasks)

### LINKED ISSUE #3578: use taskQueueId internally to the queue

alopezz:
Github Bug/Issue: Fixes #3578 

With #3987, the db will already store `task_queue_id`s rather than `provisioner_id` and `worker_type` for tasks. This PR implements the follow up of making the code also do this internally (while keeping external interfaces unchanged).


db/versions/0061.yml
```diff
@@ -0,0 +1,395 @@
+version: 61
+description: Deprecate functions that use provisioner_id/worker_type and add equivalents using task_queue_id instead, for tasks table
+methods:
+  create_task:
```
djmitche, line 4: From what I can see, this is the same function from version 0060, but with `deprecated: true`.  As a shortcut, you can omit everything except `deprecated` for such changes.

djmitche (changes requested):
Aside from the below minor detail, this looks great!

djmitche (approved):
Looks good once it's green!  I pushed a `yarn generate`.

alopezz:
Yay! #3578 is finally really closed!

PR #4127, opened 2020-12-16
support signing URLs synchronously when not using credentialAgent

djmitche:
This is broken out from #4111.  I think this will need to land first and be released so that the `ui/` directory can require `taskcluster-client-web` from the package registry, rather than using the local copy.


jwhitlock (approved):
Thanks @djmitche, looks like a clean refactor!

PR #4129, opened 2020-12-17
tc-proxy should not follow redirects

### LINKED ISSUE #4056: Taskcluster-proxy should not attach authorization header to requests with a bewit
Now that we support requiring authentication for all API endpoints, it's much more common to work with a signed URL.  As in #4022, though, if tc-proxy gets a request for a signed URL (containing `bewit=..` in the query), then it adds an authorization header anyway.  Auth doesn't like that.  A simple fix will be to pass signed URLs through unchanged.

djmitche:
Fixes #4056.

### DIFF
--- a/changelog/issue-4056.md
+++ b/changelog/issue-4056.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 4056
+---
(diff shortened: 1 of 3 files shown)


djmitche:
How on earth did this drop the coverage %?

imbstack:
Seems it just discovered web-server for some reason? https://app.codecov.io/gh/taskcluster/taskcluster/compare/4129/tree/services

imbstack (approved):

PR #4130, opened 2020-12-17
Do not follow redirects in clients.

### LINKED ISSUE #4057: Clients should not follow redirects
In #4022, the Python client is following a redirect from the `index.findArtifactFromTask` API method.  The clients should be generating the correct URL to start with, and should not be following redirects.  This API method is a unique case of a "blob" response, so as mentioned in #4022 callers should not be calling the API method directly anyway.

djmitche:
Github Bug/Issue: Fixes #4057

### DIFF
--- a/changelog/issue-4057.md
+++ b/changelog/issue-4057.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 4057
+---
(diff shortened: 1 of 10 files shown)


jwhitlock (approved):
Thanks @djmitche! This is a nice cross-comparison for the different languages as well. My favorite API for "don't allow redirects" is still Python's requests.

djmitche:
I think Requests spends a good deal of time thinking about ergonomics, and it shows.

PR #4131, opened 2020-12-17
Improve cancelTask idempotency

### LINKED ISSUE #3466: Check for conflict when canceling a task
https://sentry.prod.mozaws.net/operations/taskcluster-firefox-ci/issues/9502677/?query=is%3Aunresolved
```
Error: There should exist a run after cancelTask!
  File "/app/services/queue/src/api.js", line 845, col 15, in Proxy.<anonymous>
    let err = new Error('There should exist a run after cancelTask!');
  ?, in runMicrotasks
  File "internal/process/task_queues.js", line 97, col 5, in processTicksAndRejections
```

I suspect this occurs when two cancelTask calls are made for the same task at about the same time, and the second of the `db.fns.cancel_task` calls fails.

Step one would be creating a test for this condition.

Step two would be to fix it.  I believe `updateStatusWith` returns a boolean that signals the DB call failed.  Maybe just bail out in that case?

djmitche:
Github Bug/Issue: Fixes #3466

### DIFF
--- a/changelog/issue-3466.md
+++ b/changelog/issue-3466.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3466
+---
--- a/services/queue/src/api.js
+++ b/services/queue/src/api.js
@@ -843,8 +843,12 @@ builder.declare({
   }
 
   // Modify the task
-  task.updateStatusWith(
-    await this.db.fns.cancel_task(taskId, 'canceled'));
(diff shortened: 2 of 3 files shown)


djmitche:
I suspect lots of other API methods have similar issues, but I figured I should make a targeted fix here.

imbstack (approved):

PR #4136, opened 2020-12-18
Remove superseding support

### LINKED ISSUE #3773: Remove superseding support
Rationale:
* Firefox-CI's usage of superseding was dropped in https://bugzilla.mozilla.org/show_bug.cgi?id=1620610.  I don't know of any other uses.
* The two worker implementations implement superseding in vastly different ways
* There are security issues around having workers (not the task) call out to an arbitrary URL in the task definition
* The design is unlike anything else in Taskcluster

djmitche:
This begins as a draft because the removal from docker-worker is incomplete.  I'd like to get tests passing with this version, then go deeper in the removal.

Github Bug/Issue: Fixes #3773


services/queue/schemas/v1/task-status.yml
```diff
@@ -88,6 +88,7 @@ properties:
             - failed
             - deadline-exceeded
             - canceled
+            # deprecated, but still allowed for old tasks
```
imbstack, line 91: same as above

djmitche: good points, I'll do that and merge..

PR #4138, opened 2020-12-18
Fix max reregistrationTimeout

### LINKED ISSUE #3888: worker-manager is trying to create credentials over 31 days
https://sentry.prod.mozaws.net/operations/taskcluster-community/issues/9983723/

Its call to `createTemporaryCredentials` is returning this error.  Maybe this is just from `registerWorker` being called with too long a timeout?  In any case, this condition should be checked for and handled by returning an error message to the caller.

djmitche:
...with the correct value for "30 days"

Github Bug/Issue: Fixes #3888

### DIFF
--- a/changelog/issue-3888.md
+++ b/changelog/issue-3888.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3888
+---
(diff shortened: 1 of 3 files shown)


jwhitlock (approved):
I approve this math

PR #4140, opened 2020-12-18
safer use of Object.keys

### LINKED ISSUE #2875: [docker-worker] Add defaults for Object.keys
https://github.com/taskcluster/taskcluster/pull/2874#pullrequestreview-417029786

djmitche:
Github Bug/Issue: Fixes #2875

### DIFF
--- a/changelog/issue-2875.md
+++ b/changelog/issue-2875.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 2875
+---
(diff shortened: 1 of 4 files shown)


jwhitlock (approved):
Looks good, thanks @djmitche!

Also, [here's an article](https://www.figma.com/blog/inside-figma-a-case-study-on-strict-null-checks/) about how Figma turned on Typescript's strict null checks in a large codebase.

Issue #3665, opened 2020-10-16
Fix EADDRINUSE intermittent on auth service


djmitche:
```
  25) services/auth/test/testauth_test.js
       testAuthGet (real)
         "before all" hook for "valid creds":
     Error: listen EADDRINUSE: address already in use :::60552
      at Server.setupListenHandle [as _listen2] (net.js:1313:16)
      at listenInCluster (net.js:1361:12)
      at Server.listen (net.js:1447:7)
      at /taskcluster/libraries/app/src/index.js:60:12
      at new Promise (<anonymous>)
      at Function.createServer (/taskcluster/libraries/app/src/index.js:45:10)
      at app (/taskcluster/libraries/app/src/index.js:170:14)
      at setup (src/main.js:157:37)
      at /taskcluster/libraries/loader/src/index.js:109:33
      at new Promise (<anonymous>)
      at /taskcluster/libraries/loader/src/index.js:107:18
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
      at async Object.sticky [as load] (/taskcluster/libraries/testing/src/stickyloader.js:23:19)
      at async Context.<anonymous> (test/helper.js:239:17)
```

I think this is because it's using a hard-coded port value, rather than requesting a port 0 and letting the kernel pick an unused port.

I have a vague memory that this was easy to do for all services except auth, which is why auth is still using a hard-coded port.  Maybe that's wrong, or no longer the case.

djmitche:
> I think this is because it's using a hard-coded port value, rather than requesting a port 0 and letting the kernel pick an unused port.

Looking around, all of the services still have hard-coded paths.  I must have changed that in a daydream :)

But it's only auth that fails this way.

djmitche:
^^ hopefully that debugging info will help next time this happens.  It seems to happen a few times a week in CI.

PR #4144, opened 2020-12-21
Add debugging for EADDRINUSE errors

djmitche:
Github Bug/Issue: Refs #3665.


imbstack (approved):

djmitche:
From the earlier logging, it looks like either something else is using the port, which would be weird since the docker container isolates ports; or for some reason the port is not being released in suiteTeardown.  So hopefully a bit of extra logging of that port release will help.

PR #4157, opened 2020-12-22
UI: Handle dependencies that have expired

### LINKED ISSUE #3983: web-server ISEs fetching dependencies that do not exist
Currently https://community-tc.services.mozilla.com/tasks/FaCUR6i9Q0SrZ4vl2_IgHg fails with "t is null", with a more helpful message in the console:

```
Error from web-server: 
{…}
​
code: "INTERNAL_SERVER_ERROR"
​
exception: {…}
​​
body: Object { code: "ResourceNotFound", message: "`DBjXmNj-R6yzekOOSpaXXg` does not correspond to a task that exists.\nAre you sure this task has already been submitted?\n\n---\n\n* method:     task\n* errorCode:  ResourceNotFound\n* statusCode: 404\n* time:       2020-11-30T18:29:11.199Z", requestInfo: {…} }
​​
code: "ResourceNotFound"
​​
statusCode: 404
```

That `DB..` task is a dependency of `FaCUR..`, and indeed does not exist (it has expired).

That situation should be handled more appropriately.

djmitche:
This introduces a new DependentTask type that may have a null status and
metadata, and adjusts the UI to handle that case.

Github Bug/Issue: Fixes #3983

### DIFF
--- a/ui/src/components/TaskDetailsCard/index.jsx
+++ b/ui/src/components/TaskDetailsCard/index.jsx
@@ -109,7 +109,8 @@ export default class TaskDetailsCard extends Component {
      */
     dependentTasks: arrayOf(
       shape({
-        taskId: string,
+        taskId: string.isRequired,
+        // note that status and metadata may be missing
         status: shape({
           state: string,
         }),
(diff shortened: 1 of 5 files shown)


ui/src/components/TaskDetailsCard/index.jsx
```diff
@@ -319,16 +320,20 @@ export default class TaskDetailsCard extends Component {
...
+                        // note that the task might not exist, if it has
+                        // expired
                         <Link key={task.taskId} to={`/tasks/${task.taskId}`}>
                           <ListItem
                             button
                             className={classes.listItemButton}
                             title="View Task">
-                            <StatusLabel state={task.status.state} />
+                            <StatusLabel
+                              state={task.status?.state || 'EXPIRED'}
```
imbstack, line 331: is this our first `.?`?

imbstack (approved):

ui/src/components/TaskDetailsCard/index.jsx
djmitche: No, it's like the 7th -- owlish used it in a few other places (which is how I learned about it)

PR #4158, opened 2020-12-22
Include a card with links to reference docs

### LINKED ISSUE #2398: Make it easier to find API docs from the reference documentation
I opened https://docs.taskcluster.net/docs/reference/core/index to find the documentation for the index APIs, and I was disappointed. The page does a great job explaining concepts behind the index, but doesn't actually contain the API doc. Nor did I see an obvious link to it - I ignored the "next page" and "previous page" buttons at the bottom, one of which would have helped me.

It would be nice if there was a more obvious pointer to the API docs in the body of the page.

**Taskcluster version**
88017ca6eef228b8ec042e8f05f028c814d135a2

djmitche:
This should make the API specs, etc. a bit more discoverable.

Github Bug/Issue: Fixes #2398

### DIFF
--- a/changelog/issue-2398.md
+++ b/changelog/issue-2398.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 2398
+---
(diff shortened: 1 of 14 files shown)


jwhitlock (approved):
Thanks @djmitche, the code looks good to me. I did not attempt to generate and view the docs.

djmitche:
Sorry, I should have grabbed a screenshot.

PR #4162, opened 2020-12-23
client-shell: Display a warning if input is expected on stdin

### LINKED ISSUE #3758: [client-shell] remind user that an API method expects a payload
It's easy to type something like
```
$ taskcluster api workerManager createWorker somePool someGroup someWorker
```
and then let it sit there, thinking it's hung.  It's not, it's just waiting for the API method payload on stdin.

In cases where stdin is a tty, let's help the user by printing "Waiting for API method payload on stdin" to stderr.

djmitche:
Github Bug/Issue: Fixes #3758

### DIFF
--- a/changelog/issue-3758.md
+++ b/changelog/issue-3758.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 3758
+---
+The `taskcluster` command will now disiplay a warning after a short delay if it is expecting a request payload on stdin.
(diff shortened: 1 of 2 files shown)


jwhitlock (approved):
Thanks @djmitche, I think this will solve the issue. I tried the code, the message appeared at 1 second, and then I hit Ctrl-C to cancel.

I imagine the intended use of reading from stdin is to pipe a payload JSON. I imagine the failures cases are:

* You pipe from stdin, but it takes more than a second to complete. You see a message which you can ignore.
* You are super fast and start typing. You ignore the message which might screw up your input.

There may be some ways to deal with this, but we can see first if anyone complains.

PR #4163, opened 2020-12-23
tc-proxy: avoid using ServeMux, as it mangles URLs

### LINKED ISSUE #2721: [tc-proxy] Bug when rewriting url with %2F%2F
https://github.com/taskcluster/taskcluster-proxy/issues/44

djmitche:
Github Bug/Issue: Fixes #2721

### DIFF
--- a/tools/taskcluster-proxy/routes.go
+++ b/tools/taskcluster-proxy/routes.go
@@ -79,6 +79,22 @@ func (routes *Routes) setHeaders(res http.ResponseWriter) {
 	}
 }
 
+// Implement the Handler interface, dispatching requests to the various
+// *Handler methods.  Note that we cannot use ServeMux for this as it mangles
+// URLs and sends redircts.
+func (routes *Routes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	url := r.URL.String()
+	if strings.HasPrefix(url, "/bewit") {
+		routes.BewitHandler(w, r)
+	} else if strings.HasPrefix(url, "/credentials") {
+		routes.CredentialsHandler(w, r)
+	} else if strings.HasPrefix(url, "/api") {
+		routes.APIHandler(w, r)
+	} else {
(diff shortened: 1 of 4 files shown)


tools/taskcluster-proxy/routes.go
```diff
@@ -151,6 +151,7 @@ func (routes *Routes) RootHandler(res http.ResponseWriter, req *http.Request) {
 	defer routes.lock.RUnlock()
 
 	targetPath, err := routes.services.ConvertPath(req.URL)
+	log.Printf("targetPath %s", targetPath)
```
imbstack, line 170: do we want to leave this in here or is it debug cruft? /me happy either way

tools/taskcluster-proxy/routes.go
```diff
@@ -79,6 +79,22 @@ func (routes *Routes) setHeaders(res http.ResponseWriter) {
 	}
 }
 
+// Implement the Handler interface, dispatching requests to the various
+// *Handler methods.  Note that we cannot use ServeMux for this as it mangles
+// URLs and sends redircts.
+func (routes *Routes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	url := r.URL.String()
+	if strings.HasPrefix(url, "/bewit") {
```
imbstack, line 87: will this handle (and do we want to handle) a situation where this is `"//bewit"`, etc?

imbstack (approved):

tools/taskcluster-proxy/routes.go
djmitche: It should be removed -- thanks!

tools/taskcluster-proxy/routes.go
djmitche: I don't think we want to handle `http://taskcluster//bewit`, unless I'm misunderstanding the question.

PR #4165, opened 2020-12-23
Fix parsing of authorizedScopes in client-shell config

### LINKED ISSUE #4072: Trying to use authorizedScopes with Taskcluster CLI produces errors
**Describe the bug**
In order to use `authorizedScopes` with the Taskcluster CLI, it has to be added to the config. However, when added, it produces errors.

For example, in `~/.config/taskcluster.yml`:
```yaml
config:
  authorizedScopes:
  - queue:get-artifact:hams/*
```

Results in:
```
failed to load configuration file, error: invalid value '["queue:get-artifact:hams/*"]' for config option 'config.authorizedScopes', error: Must be a list of strings
```

**To Reproduce**

1. Fill in your taskcluster CLI config file as above.
2. Run `taskcluster config`.
3. See error

**Expected behavior**
`authorizedScopes` should work in the Taskcluster CLI.

**Taskcluster version**
v39.0.0

**Additional context**
This appears to be due to the validation / casting done in the `authorizedScopes` implementation in `client-shell`. It might need to transformed rather than just checked (from `[]interface{}` to `[]string`).

CC @djmitche

djmitche:
Github Bug/Issue: Fixes #4072

### DIFF
--- a/changelog/issue-4072.md
+++ b/changelog/issue-4072.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4072
+---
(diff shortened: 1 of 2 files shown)


imbstack (approved):
:confounded:

PR #4171, opened 2020-12-24
Fix reregistrationTimeout

### LINKED ISSUE #4125: Worker pool is busy despite high maxCapacity
**Describe the bug**

I have a periodic hook that fires 16 tasks every 6 hours, with a deadline of 6 hours (to run continuously). Including the decision, I expect the maximum tasks required to be 17, but setting the `maxCapacity` to 17 results in tasks remaining pending until they hit their deadline and raise an exception. I have tried doubling the `maxCapacity` to 33, and I still see pending tasks.

**To Reproduce**

The hook in question is here:
https://community-tc.services.mozilla.com/hooks/project-fuzzing/linux-pool8

... and the worker pool has the same name:
https://community-tc.services.mozilla.com/worker-manager/proj-fuzzing%2Flinux-pool8

**Expected behavior**

Worker pool with `maxCapacity` should be able to run that many tasks simultaneously.

**Taskcluster version**

39.1.1-6-g65facdf39

**Screenshots**

image
(although it can't be shown, there are no other tasks in the running state in that pool, observed in prior task groups and in the provisioner worker view).

 - docker-worker (Linux)
 
**Additional context**

This could very well be a misconfiguration in the pool that I'm not seeing.

imbstack:
This should make this field interpreted correctly. I hope to never multiply or divide a time
by 1000 again. This whole bit of w-m could use a re-work at some point.
Github Bug/Issue: Fixes #4125

### DIFF
--- a/changelog/issue-4125-1.md
+++ b/changelog/issue-4125-1.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 4125
+---
+Workerpools now correctly understand the `reregistrationTimeout` option.
--- a/services/worker-manager/src/api.js
+++ b/services/worker-manager/src/api.js
@@ -773,8 +773,10 @@ builder.declare({
   }
 
   // defaults to 96 hours if reregistrationTimeout is not defined
+  // make sure to turn milliseconds into seconds here since we store it in millieconds
+  // after the first interpretation.
   const { terminateAfter } = Provider.interpretLifecycle({ lifecycle: {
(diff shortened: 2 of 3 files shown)


djmitche (approved):
This seems like the fix to me.  This definitely needs a rework, though.  Perhaps that should come in the form of a change from "integer number of seconds" to "a valid fromNow expression" in the schema.  The it's clear that the only two kinds of reregistrationTimeout allowed are strings and integers, and it will be much harder to mix them up.  Let's do that sooner than later.

Issue #4050, opened 2020-12-07
The live log from generic-worker is not viewable after the task ends without assume:anonymous


ricky26:
**Describe the bug**
After a generic-worker task completes on a Taskcluster instance with an empty anonymous role, the live log link no longer works, instead returning 403.

**To Reproduce**

1. Remove all scopes from the anonymous role.
2. Run a task with generic-worker.
3. Wait for the task to complete.
4. Try to view the live log.
5. See error

**Expected behavior**
The live log link should redirect to `public/live_backing.log` after task completion.

**Taskcluster version**
v39.0.0

**Platform (please complete the following information):**

 - OS: [e.g. windows]
 - Browser if applicable [e.g. firefox, safari]
 - Version if applicable [e.g. 57]

**Additional context**
The UI redirects to:
- https://taskcluster.my.org/api/queue/v1/task/Kko3Y4iMSump_cB7SLdDgA/runs/0/artifacts/public%2Flogs%2Flive.log?bewit=BEWIT
- https://taskcluster.my.org/api/queue/v1/task/Kko3Y4iMSump_cB7SLdDgA/runs/0/artifacts/public%2Flogs%2Flive_backing.log

The second redirect has no credentials (no headers are passed to redirects and there is no bewit), so unless `queue:get-artifact:public/live_backing.log` is fulfilled, the redirect fails.

Example error payload from navigating to the artefact path in the browser:
````
{
  "code": "InsufficientScopes",
  "message": "This request requires Taskcluster credentials that satisfy the following scope expression:\n\n```\nqueue:get-artifact:public/logs/live_backing.log\n```\n\n---\n\n* method:     getArtifact\n* errorCode:  InsufficientScopes\n* statusCode: 403\n* time:       2020-12-07T14:09:32.923Z",
  "requestInfo": {
    "method": "getArtifact",
    "params": {
      "0": "public/logs/live_backing.log",
      "taskId": "Kko3Y4iMSump_cB7SLdDgA",
      "runId": "0",
      "name": "public/logs/live_backing.log"
    },
    "payload": {},
    "time": "2020-12-07T14:09:32.923Z"
  }
}
````

I'm not sure what the best way to solve this is, since the redirect artefact is designed to be flexible. Possible blunt solutions could be along the lines of:
- Allow claimed task credentials to be able to sign URLs that expire when the task does, then store the signed URL in the redirect.
- If the next URL is an artefact URL in the same taskcluster instance, sign the target before redirecting. (ick)

[...]

Issue #4215, opened 2021-01-05
API methods with deferred auth can leak info via 404s


djmitche:
Some methods that use deferred auth (calling `req.authorize`) can leak a little bit of information before they check that.  For example, `queue.createArtifact` will return a 404 if the given task or run does not exist, even if the caller has no scopes.

The information leaked in this case is information that was public before we added anonymous scopes support -- so basically absence of things in the DB.  Especially for tasks, it's kind of impractical to use this -- taskIds are random, so if an attacker doesn't have a taskId then there's no use in probing for one.

PR #4216, opened 2021-01-06
Add support for a 'link' artifact type

djmitche:
This implements the link artifact, but doesn't update the workers to create it.  Next PR.

Github Bug/Issue: Refs #4050.


services/queue/schemas/v1/post-artifact-request.yml
```diff
@@ -38,13 +38,11 @@ oneOf:
...
-      This allows you to reference artifacts that aren't managed by the queue.
-      The queue will still authenticate the request, so depending on the level
-      of secrecy required, secret URLs **might** work. Note, this is mainly
-      useful for public artifacts, for example temporary files directly
-      stored on the worker host and only available there for a specific
-      amount of time.
+      Request the queue to redirect fetches for this artifact to a URL.  An
+      existing artifact can be replaced with a RedirectArtifact as long as the
+      task is still executing.  When a RedirectArtifact is fetched, the URL is
+      returend verbatim as a Location header in a 303 (See Other) response.
```
jwhitlock, line 44:
```suggestion
      returned verbatim as a Location header in a 303 (See Other) response.
```

services/queue/schemas/v1/post-artifact-response.yml
```diff
@@ -55,6 +55,20 @@ oneOf:
     additionalProperties: false
     required:
       - storageType
+  - title:          "Link Artifact Response"
+    description: |
+      Response for an artifact with storageType lin.
```
jwhitlock, line 60:
```suggestion
      Response for an artifact with `storageType` `link`.
```

services/queue/src/artifacts.js
```diff
@@ -216,51 +226,48 @@ builder.declare({
...
-        'Existing artifact information: {{originalArtifact}}', {
-          originalArtifact: {
-            storageType: artifact.storageType,
-            contentType: artifact.contentType,
-            expires: artifact.expires,
-          },
-        });
-    }
+    let ok = true;
+    ok = ok && original.storageType === storageType;
```
jwhitlock, line 233: Interesting way to break up a long conditional, very readable. I miss `&=` and `|=`.

services/queue/src/artifacts.js
```diff
@@ -449,29 +484,16 @@ builder.declare({
...
 }, async function(req, res) {
   let taskId = req.params.taskId;
   let runId = parseInt(req.params.runId, 10);
   let name = req.params.name;
+  let names = [name];
 
-  return replyWithArtifact.call(this, taskId, runId, name, req, res);
+  await req.authorize({ names });
+
+  return replyWithArtifact.call(this, taskId, runId, name, req, res, names);
```
jwhitlock, line 496: It seems like misplaced responsibilityfor the caller to remember to initialize `names = [name]`, and call `req.authorize({ names })`. If this was Python, I'd make the function signature `names = None`, and do this work in ``replyWithArtifact``. I'm not sure if JavaScript has a similar pattern. 

But, it also appears to be called from just two places in the same file, so maybe OK?

jwhitlock (approved):
Thanks @djmitche. This new type looks like it will fix the issue and has some cool uses. The tests appear to be comprehensive as well.

services/queue/src/artifacts.js
djmitche: I agree, this is a little ugly.  I had actually used default arguments in an earlier iteration, but switched to being explicit because the same value is used in the `req.authorize` above.  And the `req.authorize` needs to happen early to avoid a new instance of #4215.  Anyway, this is all internal to the file, so I agree with "maybe OK" :)

PR #4217, opened 2021-01-06
Allow taskQueueId in task definitions

### LINKED ISSUE #3580: allow taskQueueId in task definitions, return it along with provisionerId/workerType

alopezz:
Github Bug/Issue: Fixes #3580

I'm not completely sure of all the places that taskQueueId should appear alongside the old identifiers.

I'm guessing that other than what I've done so far I should add it to task statuses as well?

### DIFF
--- a/changelog/issue-3580.md
+++ b/changelog/issue-3580.md
@@ -1,0 +1,7 @@
+audience: users
+level: minor
+reference: issue 3580
+---
+The queue service API responses will now include the taskQueueId, which will match provisionerId/workerType,
+which are also returned. Also, it is now possible to create tasks supplying a taskQueueId instead of the
+separate provisionerId and workerType identifiers.
--- a/clients/client-go/tcqueue/types.go
+++ b/clients/client-go/tcqueue/types.go
@@ -168,6 +168,11 @@ type (
 		// Syntax:     ^[a-zA-Z0-9-_]{1,38}$
 		ProvisionerID string `json:"provisionerId"`
 
+		// Unique identifier for a task queue
+		//
+		// Syntax:     ^[a-zA-Z0-9-_]{1,38}/[a-z]([-a-z0-9]{0,36}[a-z0-9])?$
+		TaskQueueID string `json:"taskQueueId"`
+
 		// Unique identifier for a worker-type within a specific provisioner
 		//
 		// Syntax:     ^[a-z]([-a-z0-9]{0,36}[a-z0-9])?$
--- a/clients/client-go/tcqueue/types.go
+++ b/clients/client-go/tcqueue/types.go
@@ -163,7 +163,9 @@ type (
 		PendingTasks int64 `json:"pendingTasks"`
 
 		// Unique identifier for a provisioner, that can supply specified
-		// `workerType`
+		// `workerType`. Deprecation is planned for this property as it
+		// will be replaced, together with `workerType`, by the new
+		// identifier `taskQueueId`.
 		//
 		// Syntax:     ^[a-zA-Z0-9-_]{1,38}$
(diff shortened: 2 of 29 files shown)


djmitche:
> I'm guessing that other than what I've done so far I should add it to task statuses as well?

Yes, that sounds correct.  We want to delete workerPool and providerId properties eventually, so we'll need taskQueueId everywhere those appear.

djmitche (commented):
Looks great!  Toss in a changelog, and this should be good to go!  I think that can be a minor changelog -- the Go client's structs change, but only to add a field, which shouldn't cause any issues.

What do you think about adding a sentence about deprecation to the providerId / workerType properties in the schema?

alopezz:
> Looks great! Toss in a changelog, and this should be good to go! I think that can be a minor changelog -- the Go client's structs change, but only to add a field, which shouldn't cause any issues.

Actually I'm not finished yet as there are still places where I haven't added the identifier. I'll soon commit the changes involving the status and some other case, but there is one case that I'm not sure how to handle, which is for task creation (`create-task-request.yml`). I think it's the only place where it's the client who would have to add the new field, and I'm not sure what's expected here.

> What do you think about adding a sentence about deprecation to the providerId / workerType properties in the schema?

Will do.

djmitche:
Haha, well, I shouldn't have assumed ;)

For the client-provided value, I think allow
* provisionerId and workerType; or
* taskQueueId; or
* all three, as long as they match

I know that's a bit of work, but I think it's maximally compatible -- the last point covers the case where someone calls `queue.task`, modifies the result, and calls `queue.createTask` with it.

alopezz:
Having a look at the things that are failing, I suspect that the go client code is not being generated correctly for the changes I made on the Task Definition Request schema, very probably due to the use of `anyOf` that tries to encode the two ways to satisfy the requirement.

My knowledge of Go is limited, but this doesn't look good:

https://github.com/taskcluster/taskcluster/blob/ba3a2ee3eac57e60278641838668f4395cda9b98/clients/client-go/tcqueue/types.go#L821-L829
https://github.com/taskcluster/taskcluster/blob/ba3a2ee3eac57e60278641838668f4395cda9b98/clients/client-go/tcqueue/types.go#L1009-L1010

djmitche:
Yeah, that's always been a bit janky.  And even outside of Go, json-schema's ability to enforce complex schema requirements like this with intelligible error messages is limited at best.  It might be best to just add all three fields as "regular" and optional fields, and do the "either or matching both" check in JS.  That will also limit the number of non-compatible changes to the Go data structures to one (when `WorkerType` and `ProvisionerId` are removed).

alopezz:
I don't know why those tests are failing...

petemoore:
Should the `anyOf` be `oneOf` ?

petemoore:
> I don't know why those tests are failing...

The only failure I see is the missing changelog:

```
$ node infrastructure/tooling/src/main.js changelog:check --pr 4217
Changelog OK
ERROR: Pull Request 4217 does not modify any files in 'changelog/'
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

djmitche:
The missing secrets are a result of some credentials rotation we did recently. A rebase will fix it, or just ignore `service-auth` on this PR.  Sorry!

alopezz:
Everything's green now after rebasing.

djmitche (approved):
It looks good to me.  @petemoore do you see any issues with this?

djmitche:
Ah, I think @petemoore is away at the moment.  Let's get this landed before it develops conflicts ;)

petemoore:
Sorry I saw the mention too late! All looks fine to me. Thanks!

PR #4218, opened 2021-01-06
Upgrade to go 1.15.6

petemoore:
It's been a while since we've looked at workers, and in the meantime go has moved on a couple of versions, so this PR upgrades generic-worker to use the latest go version. Major Go releases are scheduled every 6 months, so this picks up a year of go standard library and language fixes and improvements.


djmitche (approved):
This looks cool!  For node, `yarn generate` treats the version in `package.json` as authoritative and updates all of the other places, ensuring nothing gets left behind.  There's a little bit of that for the go version in `infrastructure/tooling/src/generate/generators/go-version.js`, but it looks like you've updated a bunch more spots than are mentioned in that file.  It also appears that file references `.taskcluster.yml` when in fact that file no longer needs a go version.  So updating all of that would be a good followup.

imbstack (approved):

infrastructure/docker-images/README.md
```diff
@@ -36,4 +36,4 @@ This is pretty much exactly the standard node image but with rabbitmq-server ins
 ## Node-Go
 
 This is a base Node image with the necessary bits of the Golang image copied in.
-This takes advantage of the fact that both imgaes are based on Debian "stretch".
+This takes advantage of the fact that both imgaes are based on Debian "buster".
```
jwhitlock, line 39:
```suggestion
This takes advantage of the fact that both images are based on Debian "buster".
```

workers/generic-worker/gw-decision-task/tasks.yml
```diff
@@ -106,7 +106,7 @@ WorkerPools:
   proj-taskcluster/gw-ci-windows10-arm:
     Platform: 'Windows 10 (arm)'
     OS: 'windows'
-    # There is no arm release for go 1.13.7 on windows, but 386 release works
+    # There is no arm release for go 1.15.6 on windows, but 386 release works
```
jwhitlock, line 109: Confirmed that this is still the case on https://blog.golang.org/ports. There is [an open issue](https://github.com/golang/go/issues/36439) to track support for ARM64 Windows 10 systems.

jwhitlock (approved):
Thanks, the changes look good.

workers/generic-worker/gw-decision-task/tasks.yml
petemoore: I think we may even run Windows/arm32 (see https://github.com/golang/go/issues/26148). I can check, I guess.

PR #4227, opened 2021-01-07
remove minification from yarn generate, and run it via 'yarn minify'

### LINKED ISSUE #4226: Remove minification from yarn generate
The `yarn minify` is a good idea, but means that renovate PRs almost always fail and require some followup.  That's preventing automerge.

What if we make `yarn minify` a separate admin command, that we can apply manually in its own PR from time to time (and if we don't get around to it, oh well too bad)?  Then Renovate PRs will always succeed.

@imbstack as the one with the most context on yarn deps, your thoughts?

djmitche:
Github Bug/Issue: Fixes #4226

### DIFF
--- a/changelog/issue-4226.md
+++ b/changelog/issue-4226.md
@@ -1,0 +1,5 @@
+audience: developers
+level: patch
+reference: issue 4226
+---
+The `yarn generate` command no longer performs minifcation of `yarn.lock` files, so that automatic dependency upgrade PRs will succeed.  Run `yarn minify` to do this manually.
--- a/changelog/issue-4226.md
+++ b/changelog/issue-4226.md
@@ -2,4 +2,4 @@ audience: developers
 level: patch
 reference: issue 4226
 ---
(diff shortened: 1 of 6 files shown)


changelog/issue-4226.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
+level: patch
+reference: issue 4226
+---
+The `yarn generate` command no longer performs minifcation of `yarn.lock` files, so that automatic dependency upgrade PRs will succeed.  Run `yarn minify` to do this manually.
```
jwhitlock, line 5:
```suggestion
The `yarn generate` command no longer performs minification of `yarn.lock` files, so that automatic dependency upgrade PRs will succeed.  Run `yarn minify` to do this manually.
```

or maybe reworded

```suggestion
The `yarn generate` command no longer combines redundant lines in `yarn.lock` files, so that automatic dependency upgrade PRs will succeed.  Run `yarn minify` to do this manually.
```

jwhitlock (approved):

PR #4233, opened 2021-01-08
Return an API response from artifact methods

### LINKED ISSUE #4058: "Simple" download functions should also return the URL in the response body
As noted in #4022, there are some API methods that redirect to a binary blob instead of returning a JSON body.  This is a little sketchy!

Let's try *also* returning a response body containing `{"url": "..."}` so that these methods are useful when called directly, too (after #4057 anyway).

djmitche:
Github Bug/Issue: Fixes #4058


libraries/api/src/middleware/schema.js
```diff
@@ -90,8 +90,8 @@ const validateSchemas = ({ validator, absoluteSchemas, rootUrl, serviceName, ent
       if (!json) {
         return res.status(204).send();
```
imbstack, line 91: I know this isn't the normal case but if I specify an undefined `json` and pass a `responseCode` I feel like maybe this should reply with the response code I provide?

imbstack (approved):

libraries/api/src/middleware/schema.js
djmitche: Fair enough - I've added that.  I'll merge if green

djmitche:
(Intermittent #3665)

djmitche:
Oh, I'm going to revert this until we release 40.0.1.  That big green button is so appealing!

PR #4234, opened 2021-01-08
Fix React error boundaries to not crash browsers

### LINKED ISSUE #4110: Error boundary causes infinite loops
..and these loops crash the browser, as they don't stop when Firefox's "do you want to stop this script" prompt appears.

STR (I think):

* Write a render method that raises an exception

The error appears in the console but moves too quickly and crashes the browser too quickly to scroll back.  I think it shows the original exception with traceback, followed by "React will try this in your error boundary, AppComponent".

I've seen this during development -- and I've also see exceptions in a render method *not* cause this!  But @lissyx has seen similar things in the production UI (at least, hanging the browser).

Maybe we should just remove the error boundary?

djmitche:
Github Bug/Issue: Fixes #4110

### DIFF
--- a/changelog/issue-4110.md
+++ b/changelog/issue-4110.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4110
+---
--- a/ui/package.json
+++ b/ui/package.json
@@ -76,6 +76,7 @@
     "react-apollo": "^3.0.0",
     "react-copy-to-clipboard": "^5.0.1",
     "react-dom": "^16.8.6",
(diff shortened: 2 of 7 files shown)


imbstack (approved):
I won't pretend to 100% get what's going on here but anything that claims to make the UI stop doing that is appreciated

djmitche:
https://reactjs.org/docs/error-boundaries.html is the docs, and the `react-error-boundary` lib basically implements that for us.

I did test the main two of these (App and Dashboard) by injecting fake errors, and they worked and did not crash :)

PR #4235, opened 2021-01-10
fix: web-server should check session when fetching

### LINKED ISSUE #3868: web-server should check session when fetching
https://github.com/taskcluster/taskcluster/blob/12f59699fa1efeea1136decb98b1eca49bcd19bd/services/web-server/src/login/PostgresSessionStore.js#L48-L58

This fetches the session by hashedSessionId, but does not verify that the sessionId actually matches the row.  Given a hash collision (unlikely), this could be used to steal sessions given access to the DB.  So it's not a huge deal.  But we should fix it.

rbrishabh:
Fixes #3868 
 
 Hi @djmitche, sorry for the delay in getting back with this really short fix! 
 
 How does this look?
 I am afraid I will trouble you a little bit more!
 
 I've made the change but I'm not 100% sure how to check if it's working! Could you please help me with that?
 
 Thank you!

### DIFF
--- a/services/web-server/src/login/PostgresSessionStore.js
+++ b/services/web-server/src/login/PostgresSessionStore.js
@@ -52,7 +52,7 @@ module.exports = function ({ session, db, options = {} }) {
 
         if (!row) {
           return callback();
-        } else if (db.decrypt({ value: row.data.encrypted_session_id }).toString('utf8') !== sessionId){
+        } else if (db.decrypt({ value: row.encrypted_session_id }).toString('utf8') !== sessionId){
           return callback();
         }
 
--- a/services/web-server/src/login/PostgresSessionStore.js
+++ b/services/web-server/src/login/PostgresSessionStore.js
@@ -52,6 +52,8 @@ module.exports = function ({ session, db, options = {} }) {
 
         if (!row) {
           return callback();
+        } else if (db.decrypt({ value: row.data.encrypted_session_id }).toString('utf8') !== sessionId){
+          return callback();
         }
 
         return callback(null, row.data);
--- a/services/web-server/test/session_store_test.js
+++ b/services/web-server/test/session_store_test.js
@@ -4,6 +4,7 @@ const session = require('express-session');
 const { promisify } = require('util');
 const helper = require('./helper');
 const PostgresSessionStore = require('../src/login/PostgresSessionStore');
+const hash = require('../src/utils/hash');
 
 helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
   helper.withDb(mock, skipping);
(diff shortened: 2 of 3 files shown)


djmitche (approved):
Looks perfect!  I think this is sufficiently low-level that we can skip a changelog.

As for testing, I don't see an easy way to do so other than testing manually in a deployment, which I will do now.  I'll change the hash in the DB and in my browser, and see that the session s no longer valid.

djmitche (changes requested):
Ah, it does look like `session_store_test.js` needs to be updated.

Within that file, you could test it as follows:

* Create a new session with `sessionid1`
* Use a DB query to update the `hashed_session_id` column to the hash `sessionid2`
* Try `session.get('sessionid2')`

I think that might involve calling `await helper.db._withDbClient(async client => { .. })` which is a little awkward since that `_` suggests you shouldn't be calling it -- but it's OK since this is in tests.

rbrishabh:
@djmitche thanks for your review!
how does this look now?
Thanks!

djmitche:
I'll take a look next week, but right off I notice that the test suite failed in CI -- perhaps you can take another look before then?

A changelog can be added with `yarn changelog --issue 3868 --silent`.

services/web-server/test/session_store_test.js
```diff
@@ -67,6 +68,28 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
+      const currentHash = hash(currentSession);
+
+      const newSession = 'sessionid2';
+      const newHash = hash(newSession);
+
+      //creates a new session
+      await store.set(currentSession, { cookie: {} });
+
+      // updates the hash of that session
+      await helper.db._withDbClient(async client => {
```
djmitche, line 84: and here, I think you want `_withClient` :)

services/web-server/src/login/PostgresSessionStore.js
```diff
@@ -52,6 +52,8 @@ module.exports = function ({ session, db, options = {} }) {
 
         if (!row) {
           return callback();
+        } else if (db.decrypt({ value: row.data.encrypted_session_id }).toString('utf8') !== sessionId){
```
djmitche, line 55: I think you want just `row.encrypted_session_id` here -- `data` is the session data itself.  That should fix the errors about `kid` in the tests.

djmitche (changes requested):

rbrishabh:
thanks @djmitche! I'll get to those changes soon!
part of the reason why I couldn't notice those errors was because I'm yet to run taskcluster on my system! The changes I've made so far are just by looking at the schema!
It's in my plans to provision taskcluster and get back into full force very soon!
In the meantime, thanks for your patience and for your quick reviews!

djmitche:
Sure thing.  If you want a bit more sync help, you can check out #taskcluster at https://chat.mozilla.org/.

djmitche:
See #4280 for the better way to access `_withClient`.

services/web-server/test/session_store_test.js
djmitche: (also, it needs a first argument of 'read' or, in this case 'write' -- that's another thing that #4280 makes easier)

djmitche:
@rbrishabh how is it going?

rbrishabh:
hi @djmitche! Had a busy week with work.
Getting to this during weekend!
Sorry for the delay!

rbrishabh:
Hi @djmitche!
Once again, I really apologize for the delay in making these tiny changes!!
I had a superbusy week!

Also, I think this looks good now!
What do you think?

djmitche (approved):
```
Error: Direct uses of DB found in services/: services/web-server/test/session_store_test.js:  helper.withDbClient = fn => helper.db._withClient('write', fn); (while executing task Services are not using node-pg)
```
wow, we even have a test for that -- I had forgotten!

I'll land my changes in #4279 right after this, and doing so should fix that failing test.

PR #4241, opened 2021-01-11
fix yaml.safe{Load,Dump} in ui

### LINKED ISSUE #4239: Task create page only shows an error
**Describe the bug**
When trying to create a task, the error `Function yaml.safeDump is removed in js-yaml 4. Use yaml.dump instead, which is now safe by default.` is seen. It looks like this was caused by an upgrade to `js-yaml`.

**To Reproduce**

1. Go to the task creation page
2. See error

**Expected behavior**
The task creation page should display the task editor.

**Taskcluster version**
v40.0.0

djmitche:
🤦🏽 

I think Webpack didn't because safeLoad and safeDump are still *deifned*, they are just functions that throw an error.  Typescript would hopefully have caught this?

Github Bug/Issue: Fixes #4239

### DIFF
--- a/changelog/issue-4239.md
+++ b/changelog/issue-4239.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 4239
+---
+This version fixes a bug in the user-interface causing messages about `yaml.safeDump` having been removed.  The developers regret te error.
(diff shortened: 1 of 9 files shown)


djmitche:
(I did test this fix manually)

PR #4246, opened 2021-01-12
long migrations should be flagged as major

djmitche:
For example, https://github.com/taskcluster/taskcluster/releases/tag/v40.0.0 should have had

> [MAJOR] #3578
> The queue service now uses taskQueueId internally instead of the pair provisionerId/workerType for tasks.
> This involves an online database migration that scans each row in the `queue_tasks` table and thus might take considerable time for a production deployment.  A well-equipped database can scan about 600 rows per second.


djmitche:
(this is just docs so doesn't need a changelog)

imbstack (approved):

PR #4251, opened 2021-01-12
Fix updating from reference to link artifact

### LINKED ISSUE #4247: Replacing a reference artifact with a link artifact causes 403 errors
**Describe the bug**
When a reference artifact is replaced with a link artifact `storage_type` is not updated in the database, resulting in a "reference" artifact where `url` is unset.

**To Reproduce**

1. Create a reference artifact, which redirects to an external URL.
2. Create a link artifact with the same name.
3. Click on the artifact link on the Task page.
4. Notice that you are redirected to `undefined`.

**Expected behavior**
Replacing an artifact should update the storageType.

**Taskcluster version**
v40.0.3

djmitche:
I added some additional bits to the tests in hopes of encountering additional bugs due to failure of imagination, but no luck.

Github Bug/Issue: Fixes #4247


imbstack (approved):
seems to make sense

PR #4255, opened 2021-01-13
Return responses from artifact-related API methods

### LINKED ISSUE #4058: "Simple" download functions should also return the URL in the response body

djmitche:
* artifact-related API methods now return a JSON response body (instead of the default text response from express)
* clients read this response body and return it as a normal JSON response

The upshot is that it's no longer necessary to generate signed URLs and fetch them with a library following redirects in order to download an artifact.  Instead:

```js
const { url } = await queue.getArtifact(..);
const res = await request.get(url);
```

It doesn't seem that the browser-side `fetch` API allows JS to see the body of a redirect response.  Setting `Request.redirect` to `"manual"` results only in a Response object with `redirected` and `url` properties, but nothing else.  So this behavior is impossible to support in the browser.

Github Bug/Issue: Fixes #4058


services/queue/src/artifacts.js
```diff
@@ -439,18 +446,12 @@ builder.declare({
...
-    'compliant client, but has extra features to ensure the download is valid.',
-    'It is strongly recommend that consumers use either taskcluster-lib-artifact (JS),',
-    'taskcluster-lib-artifact-go (Go) or the CLI written in Go to interact with',
-    'artifacts.',
+    '**Response**: the HTTP response to this method is a 303 redirect to the',
+    'URL from which the artifact can be downloaded.  The body of that response',
+    'contains the data described in the output schema, contianing the same URL.',
+    'Callers are encouraged to use whichever method of gathering the URL is',
+    'most convenient.  Standard HTTP clients will follow the redirect, while',
+    'API  client libraries will return the JSON body.',
```
jwhitlock, line 454:
```suggestion
    'API client libraries will return the JSON body.',
```

jwhitlock (approved):
The code and tests look good, thanks @djmitche! 

I'm confused by the need to repeat the URL in the body of the response. It seems like the web client is going to use the Location header, and the other clients could as well. I'm guessing this is a workaround for third-party code that treats empty response bodies as errors, but I could be missing some context.

jwhitlock:
If I'd just looked up, I would have seen the answer. Putting the URL in the body makes this target code possible:

```javascript
const { url } = await queue.getArtifact(..);
const res = await request.get(url);
```

A little repetition to improve the client code is a good tradeoff.

clients/client-shell/apis/provider.go
```diff
@@ -228,6 +228,16 @@ func execute(
 	}
 
 	res, err := req.Send()
+	if err != nil {
+		// Got considers redirects an error, but we would like to read the body as if it
```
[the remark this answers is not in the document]
djmitche: It started in JS with [sindresorhus](https://github.com/sindresorhus/got), but has been ported to go by [TC folks](https://github.com/taskcluster/go-got).  It's not totally clear that was a good idea.  Between this issue, and the similar issue with httpbackoff (which is also a separate library owned by TC), it might make sense to reconsider our choices.  In another PR :)

services/queue/src/artifacts.js
djmitche: This will require `yarn generate` so I'll apply it locally and push again.

PR #4267, opened 2021-01-15
Bug 1562993 - generic-worker: only log first error encountered in task log

petemoore:
Bugzilla Bug: [1562993](https://bugzilla.mozilla.org/show_bug.cgi?id=1562993)

### DIFF
--- a/changelog/bug-1562993.md
+++ b/changelog/bug-1562993.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: bug 1562993
+---
+generic-worker now only reports the first error it encounters when a task fails or hits an exception.
--- a/go.mod
+++ b/go.mod
@@ -49,9 +49,9 @@ require (
 	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
 	github.com/xeipuuv/gojsonschema v1.2.0
 	golang.org/x/crypto v0.0.0-20201016220609-9e8e0b390897
-	golang.org/x/net v0.0.0-20201016165138-7b1cca2348c0
+	golang.org/x/net v0.0.0-20201021035429-f5854403a974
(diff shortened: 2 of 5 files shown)


djmitche (approved):
I guess the reason to have an error accumulator like this is that it's common for many errors to occur at the end of an unsuccessful task, and this provides a simple way to deal with them (`err.add(..)`) without worrying about which came first.

It looks like the only thing that consults any error but the first is WorkerShutdown.  Everything else (`resolve`, `occurred`) ignores all but the first.

Maybe it would make sense to refactor this a little to only keep the first error, with a boolean to track whether a graceful shutdown had occurred (maybe a global?).

I'm also curious why the `go.mod` changes -- was there something in older versions of those libraries that was broken?

petemoore:
> I guess the reason to have an error accumulator like this is that it's common for many errors to occur at the end of an unsuccessful task, and this provides a simple way to deal with them (`err.add(..)`) without worrying about which came first.

That's exactly right. 

> It looks like the only thing that consults any error but the first is WorkerShutdown. Everything else (`resolve`, `occurred`) ignores all but the first.
> 
> Maybe it would make sense to refactor this a little to only keep the first error, with a boolean to track whether a graceful shutdown had occurred (maybe a global?).

I agree we could have a follow-up PR to simplify the handling, although I'm not 100% sure this PR will stick, so I thought as a first pass better to make the smaller change, check all is as wanted, and then to refactor once we're sure this is how we want things. But for sure, if this sticks, this would be a nice cleanup to have.

> I'm also curious why the `go.mod` changes -- was there something in older versions of those libraries that was broken?

Running `build.sh` updated `go.mod` and `go.sum`, pulling in newer versions. I'm not entirely sure why, but I did see the updates are for newer versions, so I was happy to accept the changes. The go module version pinning mechanics are still a little mysterious to me, but to be honest I haven't done a deep dive to better understand what prompted the upgrade.

PR #4271, opened 2021-01-15
Add projectId to the queue API

### LINKED ISSUE #4262: Add `projectId` to tasks
Per [RFC#163](https://github.com/taskcluster/taskcluster-rfcs/blob/main/rfcs/0163-project-id.md) tasks should grow a `projectId` property, defaulting to `none`.

While this will require a DB migration, it is just adding a nullable column, so it should be quick.

@alopezz  is planning to add `taskQueueId` in #4217, so this is a good time to land a similar change.

djmitche:
Github Bug/Issue: Fixes #4262


djmitche:
Huh, I somehow clobbered db version 62 in a rebase here.  I'll fix that up.

imbstack (approved):
this was less code than I thought it would be for some reason

PR #4280, opened 2021-01-19
Add `helper.withDbClient` in tc-lib-testing

### LINKED ISSUE #4279: Allow tests to get direct DB access
This works and is fine: (from #4235)
```
      await helper.db._withDbClient(async client => {
        await client.query('update sessions set hashed_session_id = $1 where hashed_session_id = $2', [newHash, currentHash]);
      }); 
```

but that `_` in `_withDbClient` suggests we shouldn't be calling it.  Let's fix that up in such a way that *tests* can use a DB client, but it's still difficult for non-test code to do so.

djmitche:
Github Bug/Issue: Fixes #4279

### DIFF
--- a/changelog/issue-4279.md
+++ b/changelog/issue-4279.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4279
+---
--- a/libraries/testing/README.md
+++ b/libraries/testing/README.md
@@ -305,6 +305,7 @@ The resulting database is injected into the taskcluster-lib-loader as `db` and a
 
 It is up to the test suite implementation to reset the contents of the database between tests.
 Ideally this is done via `helper.db.fns` methods.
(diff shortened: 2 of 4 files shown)


jwhitlock (approved):
Thanks @djmitche, this makes using a writable DB client look more official.

I think skipping the changelog is OK for a dev-only change. I wouldn't mind changing a test in this PR, like the one suggested in #4279, to use the new facility.

djmitche:
> I think skipping the changelog is OK for a dev-only change. I wouldn't mind changing a test in this PR, like the one suggested in #4279, to use the new facility.

We can wait to land this and rebase this one on top of it.

imbstack (approved):

djmitche:
Rebased on top of #4235, with the temporary `helper.withDbClient` in that PR removed, meta-tests verified to pass, and a silent changelog.

PR #4283, opened 2021-01-20
Role deletion from Roles Table

### LINKED ISSUE #2393: [UI] Add a delete action in roles, clients, secrets, hooks views

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Taskcluster version**
v24.3.0

**Additional context**

mrsVelena:
Github Bug/Issue: Fixes #2393

Added a delete action to each role in the list of roles

.


ui/src/components/RolesTable/index.jsx
```diff
@@ -28,10 +53,12 @@ export default class RolesTable extends Component {
     onPageChange: func.isRequired,
     /** A search term to refine the list of roles */
     searchTerm: string,
+    onDialogActionOpen: func,
```
djmitche, line 56: This prop is not checked for null-ness before being used, so it should have `.isRequired` like `onPageChange` does.  That also means you can omit it from `defaultProps`, below.

djmitche (changes requested):
This looks like it's got the necessary pieces in place.  But, it has some linting failures that are preventing netlify from deploying it.  If you can get that fixed up, I can take another look.

The `service-auth` failure is because your work is based on an older branch (from a week or so ago).  It's fine to just leave that failing, as you haven't modified that service, but a rebase onto the latest commit on `main` will fix it.

petemoore:
It looks like there may be some test failures in https://community-tc.services.mozilla.com/tasks/LHu8K21YQN6LmHKMD46ZQA/runs/0/logs/public/logs/live.log but I'm not too sure what is wrong...

djmitche (approved):
This is fantastic -- nice work!

One minor note that doesn't matter here, but will for the next PR: you should usually make a "topic branch" to work on a PR like this, so that your `main` branch can be updated to new versions of the shared `main` without losing your changes.  The [GitHub Flow](https://guides.github.com/introduction/flow/) doc has more information on that.  Once this is merged, you'll want to do something like

```
git checkout main
git fetch https://github.com/taskcluster/taskcluster main
git reset --hard FETCH_HEAD
```
and that will reset your `main` branch to the same commit as the taskcluster repo's `main` branch. Then you can create a topic branch for your next project:
```
git checkout -b my-next-project
# (or whatever you want to call the branch)
```

PR #4289, opened 2021-01-21
Update online migrations to use `state_in` instead of `state`

### LINKED ISSUE #4250: 0059 migration is running at 50 items/sec
That's really slow!  Why?

image

djmitche:
This is the input variable containing the correct place to begin
searching for the next row in the batch.  `state` is always empty at
this point, causing scans to begin at the beginning of the table.

Github Bug/Issue: Fixes #4250 (again)


jwhitlock (approved):
once more, with feeling!

PR #4291, opened 2021-01-21
Track whether a provider is set up correctly

### LINKED ISSUE #4276: [worker-manager] allow worker-manager to start up even if a provider's setup method fails
Similar to #3961 but for worker-manager.

The idea is that if a provider's `setup` fails, an error should be reported and the provider should be flagged as not-ready-to-go and the setup call retried after a delay (say, every minute).  All other operations on the provider should then be silently skpiped.

This would mean that if, for example, AWS is down, then the AWS providers will just cycle waiting for it to come back up, while all of the other providers will operate normally.

It would avoid things like https://bugzilla.mozilla.org/show_bug.cgi?id=1687053

djmitche:
This allows the failure of one provider, or a provider backend such as a
cloud provider, from causing the entire process to fail a startup.

Github Bug/Issue: Fixes #4276

### DIFF
--- a/changelog/issue-4276.md
+++ b/changelog/issue-4276.md
@@ -1,0 +1,5 @@
+audience: deployers
+level: patch
+reference: issue 4276
+---
+The worker-manager service will now start up even if one of its providers is down or misconfigured.  Worker pools using that provider will not be provisioned, but other pools will continue to operate normally.
--- a/clients/client-go/tcworkermanager/tcworkermanager.go
+++ b/clients/client-go/tcworkermanager/tcworkermanager.go
@@ -9,6 +9,10 @@
 // /references/worker-manager/v1/api.json
 // This service manages workers, including provisioning for dynamic worker pools.
 //
+// Methods interacting with a provider may return a 503 response if that provider has
+// not been able to start up, such as if the service to which it interfaces has an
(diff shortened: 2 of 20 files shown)


djmitche:
This doesn't have tests yet, but I'd like a 30% review to see if this seems like a good approach.

The only thing I *really* don't like, aside from a certainty that I've made some silly error here, is the handling of messages to the provisioner service.  The comment explains my concerns.  Is there another way?

imbstack (commented):
I think this definitely makes sense so far. If some of the awkwardness is around the pulse listeners, we didn't really end up using any of that and it can probably all be removed?

djmitche:
@imbstack: By "any of that" do you mean the `createResources` and `updateResources` methods?
https://github.com/taskcluster/taskcluster/blob/b46c554eedeae7d38bea141b32eb7cb328a0f89e/services/worker-manager/src/provisioner.js#L90-L104

djmitche:
Testing the failure conditions is tricky, since a failed provider is retried forever and puts error messages into the error log.  I've skipped testing the `if setupFailed` in `api.js`, `provisioner.js`, and `worker-scanner.js` here.  Maybe that's a bad idea -- please feel free to tell me so.

imbstack (approved):
looks great and a side bonus of plenty of red diff

PR #4292, opened 2021-01-22
Sentry no session tracker

jwhitlock:
This extends Dependabot PR #4282 to [explicitly disable](https://docs.sentry.io/platforms/javascript/configuration/releases/#release-health) the new [session tracking feature](https://docs.sentry.io/product/releases/health/). If desired, I can instead apply this commit to the original PR, now or when the expected minor release 6.0.2 is applied.

### DIFF
--- a/changelog/Q1047xsrS2KoRLzwSE5bkw.md
+++ b/changelog/Q1047xsrS2KoRLzwSE5bkw.md
@@ -1,0 +1,7 @@
+audience: admins
+level: patch
+---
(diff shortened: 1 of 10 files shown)


djmitche (approved):
This looks good to me.  It seems like the 6.0.2 fix doesn't matter in this context, so maybe there's no reason to wait?

If this is merged I think renovate will automatically close its PR, so I say go for it.

imbstack (approved):

PR #4293, opened 2021-01-22
Doc that the artifact.contentType is informational

djmitche:
This value is used by the UI to select an icon, and that purpose is just
fine.  But it is redundant to the content-type of the actual data in the
artifact, and it's important to be clear about which value is
authoritative.

I had previously marked this value as deprecated, but that's not quite right.


petemoore (approved):
Useful clarification, thanks!

PR #4301, opened 2021-01-25
rust client: document where  should appear instead of panicing

### LINKED ISSUE #4269: Thread task.projectId through UI
* add projectId to web-server's Task entity
* add projectId to the default task-creator task, with a site-specific value
* show projectId in task inspector

### LINKED ISSUE #4296: Be more lenient about trailing slashes in client-rust's `path_prefix`
This currently panics if there are trailing slashes, and @imbstack pointed out that it'd be just as easy to simply strip those.

djmitche:
Github Bug/Issue: Fixes #4296

### DIFF
--- a/changelog/issue-4296.md
+++ b/changelog/issue-4296.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4296
+---
--- a/clients/client-rust/src/client.rs
+++ b/clients/client-rust/src/client.rs
@@ -70,10 +70,9 @@ impl ClientBuilder {
     /// Set the path_prefix; this will be included between the root URL and the path given to
     /// `request`, `make_url`, and `make_signed_url`.  This is typically used when building a
     /// client that will address a single service, such as `api/queue/v1/`.  The path prefix
-    /// must end with a `/` character.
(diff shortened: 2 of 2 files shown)


jwhitlock (approved):
Thanks @djmitche, I agree that we should get away from panics. I do wonder if we should validate the ``path_prefix`` somewhere. Maybe a ``PathPrefix`` class that wraps the string? Maybe in ``Client::new``? It looks like the internal code is the most frequent caller, so maybe it is OK. I'll defer to @imbstack, who raise the issue the first time.

Also, next time you touch the code, please replace ``Unkown`` with ``Unknown``

djmitche:
I skipped validation because it's internal code, yes.  I wanted to avoid adding `Result` to the return type for the functions that check this, but now that I look they all already return `Result`.  So I'll just add an error message.

djmitche:
(Except `.path_prefix`, which is part of the builder pattern.  I think the assert is OK there)

jwhitlock (approved):
Thanks, I'm more comfortable with this. Which means little from a beginner Rust developer...

djmitche:
I think your intuition is good :)

PR #4308, opened 2021-01-26
Allow project-id related scopes in task manipulation

### LINKED ISSUE #4270: Require projectId scopes to manipulate tasks
`queue.cancelTask`, rerun, etc. should all accept `queue:<verb>-task:project-id:<projectId>` to permit task manipulation.

djmitche:
Github Bug/Issue: Fixes #4270

Note that projectId is not visible in the UI, except in the "view full task definition", but that's tracked as an issue.


djmitche:
Looks like the `service-web-server` failure was an unrelated intermittent yarn failure.

imbstack (approved):
wow this is so exciting. literally a problem as old as tc... fixed!

PR #4318, opened 2021-01-27
Add taskQueueId and projectId to the UI and docs

### LINKED ISSUE #4269: Thread task.projectId through UI

djmitche:
Github Bug/Issue: Fixes #4269

This updates docs, too, and also makes some drive-by docs fixes for old/awkward stuff.  Hopefully most of those diffs are self-explanatory, such as removing the v0 `.taskcluster.yml` from the notify docs.

### DIFF
--- a/changelog/issue-4269.md
+++ b/changelog/issue-4269.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 4269
+---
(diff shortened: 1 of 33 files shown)


imbstack (approved):
seems reasonable to me although good chunks of this are "sure looks like some react/graphql happening there, r+"

PR #4322, opened 2021-01-27
Added Delete button for individual Client records

### LINKED ISSUE #2393: [UI] Add a delete action in roles, clients, secrets, hooks views

mrsVelena:
Github Bug/Issue: Fixes #2393 for Clients page


changelog/issue-2393.md
```diff
@@ -2,4 +2,6 @@ audience: developers
 level: minor
```
djmitche, line 2: I missed this on the earlier PR, but let's change this to "patch".  The guidelines for these levels are [here](https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/changelog.md), In this case, "A patch change is one that fixes a bug or makes a minor change which users can do their work without." -- users can do their work without this feature, just a bit more slowly.

djmitche (changes requested):
One minor request: let's move the delete icon to the rightmost end of the list item, so instead of

image

it looks like

image

djmitche:
Ah!  I didn't see the changes you pushed :(  There's a little circular-arrow button next to the reviews in the GitHub UI that will re-request a review, and hopefully get a response a little sooner.  Sorry!  I'll have a look shortly.

djmitche (changes requested):
This looks great, but seems to have one lingering bug. When I click the delete icon, it displays a dialog that says "This will delete the `github/28673|djmitche/cli-ZN_L9m` client" (which is the clientId I was trying to delete).  When I click the "Delete Client" button, the dialog changes to

image

Any idea what's wrong?

mrsVelena:
I'm sorry for the late answer (I was on vacation and didn't have access to my computer). I actually cannot reach this step, because I don't have account with permissions to delete anything. But I assumed, there could be this problem, because I don't close Dialog onComplete after deleting. I tried to fix this, please check if this helped.

djmitche:
That fix seems perfect!  I noticed, though, that the client is still in the list after the deletion occurs.  Is that something you could fix?

Here's how to create clients:
 * On https://community-tc.services.mozilla.com/profile (after signing in) look for a scope like `auth:create-client:github/<something>`; mine is `auth:create-client:github/28673|djmitche/*`.  That gives you permission to create clients with that prefix (In my case, I can create clients like `github/28673|djmitche/foo` and `github/28673|djmitche/bar`).
 * Create a new client for yourself with a name you have permission for; mine might be `github/28673|djmitche/issue-2393`.  Give it scopes to create and delete clients -- for me that would be `auth:create-client:github/28673|djmitche/*` and auth:delete-client:github/28673|djmitche/*`.  Set the expiration to a few months from now.
 * On your development server, sign in with this new clientId and the accessToken it gave you (leave the certificate blank).  Now you can use your development server to create and delete clients all you want!

Sorry that's so complicated -- signing in is always difficult in development environments, it seems.

mrsVelena:
I fixed list update after client deletion. But I think the same bug would be on Roles page, too. What would be a better way to fix it? Through creating new PR?

djmitche:
Good thinking.  Yes, this PR has been around for a while and is (hopefully) ready to merge, so a new one is a good idea.

djmitche (approved):
And, indeed, this is great -- thanks for your patience and persistence!

djmitche:
(I merged main in here and fixed merge conflicts -- I'll merge back to main when the tests pass)

PR #4334, opened 2021-01-29
support negotiating upload methods

### LINKED ISSUE #4272: Support negotiation of upload methods similarly to download methods
One thing that's a little trickier here than downloads is that we may need to do multiple rounds of negotiation.  There was [some discussion about this way back when](https://github.com/taskcluster/taskcluster-rfcs/pull/120#discussion_r183140955).  Initial implementations might not need this, but it's something to think about in the initial design.

All upload methods will use the existing uploadId approach to ensure idempotency.  That might also ease the multiple rounds.

As a first pass, let's have a "data-inline" method that includes the artifact data inline, with a pretty strict size limit like 8k.  That will be simple to implement, and is also potentially useful in some corner cases and for testing.

djmitche:
Github Bug/Issue: Fixes #4272


services/object/src/api.js
```diff
@@ -68,20 +83,27 @@ builder.declare({
...
 
-  // ..upload the object..
-  await backend.temporaryUpload(object, data);
+  // default uploadMethod to an empty object, meaning no matching methods
+  let uploadMethod = await backend.createUpload(object, proposedUploadMethods);
 
-  // ..and mark its completion
-  await this.db.fns.object_upload_complete({ name_in: name, upload_id_in: uploadId });
+  // XXX temporary - until we have a finishUpload endpoint, and an upload method that does
+  // not finish immediately, finish uploads automatically when they succeed
```
petemoore, line 96: Is this ok to leave in?

petemoore (commented):
It seems reasonable, let's go through it together on zoom.

services/object/src/api.js
djmitche: It's addressed by #4347

PR #4337, opened 2021-02-01
Fix error reporting from worker-runner

### LINKED ISSUE #4336: `workerGroup` / `workerId` are required in wm.reportWorkerError, but not sent by worker-runner
I don't recall if these should not have been required, or if worker-manager should have been providing them.

djmitche:
This has likely never worked!

Github Bug/Issue: Fixes #4336


imbstack (approved):
ah interesting

PR #4341, opened 2021-02-01
[generic-worker] Do not fail if tasks-resolved-count.txt is empty

### LINKED ISSUE #4340: generic-worker should treat a zero-byte tasks-resolved-count.txt as missing
If this file contains zero bytes, the ReadTaskResolvedFile panics with a confusing message.  I suspect this occurs when the disk fills up?

djmitche:
Github Bug/Issue: Fixes #4340

### DIFF
--- a/changelog/issue-4340.md
+++ b/changelog/issue-4340.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4340
+---
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -288,11 +288,14 @@ func setupExposer() (err error) {
 func ReadTasksResolvedFile() uint {
 	b, err := ioutil.ReadFile("tasks-resolved-count.txt")
 	if err != nil {
+		log.Printf("could not open tasks-resolved-count.txt: %s (ignored)", err)
 		return 0
 	}
 	i, err := strconv.Atoi(string(b))
 	if err != nil {
-		panic(err)
+		// treat an invalid (usually empty) file as nonexistent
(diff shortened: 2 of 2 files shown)


petemoore (commented):
Do we know what causes this?
Is it worth logging the error?
Should we special case that len(b) == 0 and panic in other cases?
Perhaps the output could suggest deleting the file, and include the full path to the file in the log message? It could be a cause for concern if this file exists but has invalid content. I wonder how it happened...

djmitche:
We've seen this happen before when disks fill up, so I expect that's the issue.

Logging is a good idea -- I can add that.

What is the consequence of not having this data?

djmitche:
(I added the logging)

petemoore (approved):
I think this is fine. The consequence of not being able to read the file is very small: if a worker administrator has set up a worker to run a limited number of tasks (e.g. after 10 tasks, don't run anymore) then the counter would be reset. However, this is quite a contrived setup, and probably this has never been done.

The `numberOfTasksToRun` config setting was added when we had the simple engine, so that a machine administrator could guarantee that some event occurred after no more than X tasks had run (e.g. a puppet run), by having generic-worker exit after X tasks. Then when multiuser engine was created, the running count was persisted to disk in order to be functionally similar despite the reboots.

In reality, if there are reboots, administrators would probably run their external task (such as a puppet run) on machine bootup, before generic-worker is started, so it isn't really useful.

When multiuser no longer needs to reboot between tasks for non-UI tasks, and this becomes more important again, the task count will be in memory already, so won't need to persist that state to disk, so then the problem also goes away.

So in other words, this is all good. Another option is that we just don't support the feature when there are reboots, and don't write the file to disk, but we can do that later if there are further problems after this one is solved.

PR #4342, opened 2021-02-01
fix: allPurgeRequests returns a continuationToken when there's nothing left

### LINKED ISSUE #3789: allPurgeRequests returns a continuationToken when there's nothing more`
```
}(ve) rubin ~/tmp $ curl https://dustin.taskcluster-dev.net/api/purge-cache/v1/purge-cache/list
{
  "continuationToken": "",
  "requests": []
}
```

When there's no more data, the `continuationToken` property should be omitted, not just an empty string.

Depending on implementation, that could cause a caller to loop endlessly calling this endpoint.

Other endpoints (at least `auth.listClients`) get this right, so what's the difference for this method?

rbrishabh:
fixes #3789

@djmitche do you think this needs a changelog?
Also, does it look ok?

### DIFF
--- a/changelog/issue-3789.md
+++ b/changelog/issue-3789.md
@@ -1,4 +1,5 @@
-audience: general
-level: silent
+audience: developers
+level: patch
 reference: issue 3789
 ---
+Fixed an issue where when there's no more data, the continuationToken property was not being omitted, but being returned as just an empty string. Depending on implementation, that could cause a caller to loop endlessly calling the purge cache endpoint.
--- a/changelog/issue-3789.md
+++ b/changelog/issue-3789.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 3789
+---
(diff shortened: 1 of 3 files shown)


djmitche:
This can do with a `--patch` changelog, yes.  I'll have a look at the patch soon!

djmitche (changes requested):
This looks like the right fix, but to be sure, we should check this in the tests.

https://github.com/taskcluster/taskcluster/blob/7a56dae6320dfdc1d7fe9625cb826a2993fb02d3/services/purge-cache/test/purgecache_test.js#L30-L33
looks like a good place to start, with an extra assertion that there's no continuationToken property.

rbrishabh:
Alright! I'll get to it
Thanks for the prompt review!

rbrishabh:
Hi @djmitche, I think this is ready to go in!
Feel free to review whenever :)

djmitche (approved):
Nicely done!

PR #4347, opened 2021-02-01
Add an object.finishUpload method

### LINKED ISSUE #4273: Add object.uploadComplete
This API method will allow uploaders to signal when an upload is complete.  For an upload method that requires additional steps beyond the `uploadObject` call, this allows the object service to prevent access to an in-progress upload.

We already define an object with `upload_id != null` to be in-progress, and even have an API method to complete an upload, so this is just an API method.

djmitche:
Github Bug/Issue: Fixes #4273

Still WIP pending landing of #4334

### DIFF
--- a/changelog/issue-4273.md
+++ b/changelog/issue-4273.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4273
+---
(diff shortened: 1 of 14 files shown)


djmitche:
Rebased, but still pending #4334.

imbstack (approved):
makes sense to me

PR #4362, opened 2021-02-04
guard against integer overflow

djmitche:
This should silence a CodeQL warning that appeared out of nowhere today.

### DIFF
--- a/tools/worker-runner/cfg/workerconfig.go
+++ b/tools/worker-runner/cfg/workerconfig.go
@@ -102,6 +102,10 @@ func merge(v1, v2 interface{}) interface{} {
 	arr1, arr1ok := v1.([]interface{})
 	arr2, arr2ok := v2.([]interface{})
 	if arr1ok && arr2ok {
+		// check for overflow
+		if len(arr1) > 2*1024*1024*1024 || len(arr2) > 2*1024*1024*1024 {
+			panic("arrays too large (>2**30 items)")
+		}
 		res := make([]interface{}, 0, len(arr1)+len(arr2))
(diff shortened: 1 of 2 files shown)


djmitche:
Worth noting this was not a security issue -- the things being merged are all controlled by the deployer

tools/worker-runner/cfg/workerconfig.go
```diff
@@ -102,6 +102,10 @@ func merge(v1, v2 interface{}) interface{} {
 	arr1, arr1ok := v1.([]interface{})
 	arr2, arr2ok := v2.([]interface{})
 	if arr1ok && arr2ok {
+		// check for overflw
```
jwhitlock, line 105: nit
```suggestion
		// check for overflow
```

tools/worker-runner/cfg/workerconfig.go
```diff
@@ -102,6 +102,10 @@ func merge(v1, v2 interface{}) interface{} {
...
+		if len(arr1) > 2*1024*1024*1024 || len(arr2) > 2*1024*1024*1024 {
```
jwhitlock, line 106: Glad we still support merging 1 billion item lists, I'm sure the system will have no problems at that scale

jwhitlock (approved):
Thanks @djmitche!

tools/worker-runner/cfg/workerconfig.go
djmitche: I feel like at this point I should claim that I leave these here for you intentionally

PR #4381, opened 2021-02-09
Fix call to purgeCache.purgeCache in web-server to use workerPoolId

djmitche:
Let's land this before releasing 41.


djmitche:
I want to cut v41 before we land too many more renovate changes, etc. -- so I merged this.

PR #4396, opened 2021-02-11
Add a putUrl upload method, supported in AWS and Google

### LINKED ISSUE #4274: Add a put-url upload method
This should be similar to the current `s3` object type in the queue, but without S3-specific bits.  Basically, caller specifies some metadata and gets back a URL to which they can make a PUT request.

Things to worry about:
 * Some way to indicate to the caller whether they can or cannot have a non-identity content-encoding (GCS can [automatically decode a gzip encoding](https://cloud.google.com/storage/docs/transcoding) while AWS cannot)
 * If possible, allow the caller to specify a hash (GCS and AWS both seem to use MD5?) that the cloud provider will verify
 * The method should be idempotent, so for example if the upload fails and must be retried, a call to `uploadObject` with the same parameters should result in a newly signed URL for the same object

It's OK to make sub-optimal decisions here (such as "Content-Encoding must be identity").  Looking at the upload docs for GCS and AWS shows that they are not at all compatible, so we will likely have `s3-multipart` and `gcs-resumable` upload methods that allow the full expression of each cloud's functionality.

djmitche:
This is the bicycle of upload methods: not fast, but it will always
work.

Github Bug/Issue: Fixes #4274

### DIFF
--- a/changelog/issue-4274.md
+++ b/changelog/issue-4274.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4274
+---
--- a/clients/client-go/tcobject/types.go
+++ b/clients/client-go/tcobject/types.go
@@ -127,6 +127,37 @@ type (
 		// the payload grows. In general, this is useful for testing and for metadata objects such as
 		// separate cryptographic signatures.
 		DataInline DataInlineUploadRequest `json:"dataInline,omitempty"`
+
+		// Request a URL to which a PUT request can be made.
+		Puturl PutURLUploadRequest `json:"puturl,omitempty"`
+	}
+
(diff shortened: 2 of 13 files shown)


djmitche:
The uploads to the Google endpoint aren't working.. investigating.

djmitche:
(It was just tests interfering with each other)

jwhitlock (approved):
Thanks @djmitche! It is interesting that the Google backend's AWS interface covers this method as well, there was less code there then I was expecting. There is a good set of tests and docs.

djmitche:
Yeah, I expect we'll soon run up against the edges of that compatibility layer, but not yet! 😃

PR #4414, opened 2021-02-16
Create json schema for generic-worker.config file

petemoore:
It is a long-standing issue that we don't have a json schema for the generic-worker config file, but instead validate the config file by primitive means. This cleanup will remove the duplication of the properties in the `usage*.go` files and the `gwconfig/*`, and will also provide better lexical validation of properties based on the rich validation capabilities of json schema.

This will also make it easier to auto-deploy worker documentation on release to the docs site, which can be rendered consistently with the other schemas that we publish for worker task payloads and service API request/response payloads.


petemoore:

PR #4418, opened 2021-02-16
Fix triggerHook response schema to include taskQueueId

### LINKED ISSUE #4417: hooks.triggerHook response does not allow taskQueueId
```
Error: Output schema validation error: 
Schema Validation Failed!
Rejecting Schema: https://community-tc.services.mozilla.com/schemas/hooks/v1/trigger-hook-response.json#
Errors:
  * data.status should NOT have additional properties: "taskQueueId"
  * data....
  File "/app/libraries/api/src/middleware/schema.js", line 79, in ServerResponse.res.reply
    const err = new Error('Output schema validation error: ' + error);
  File "/app/services/hooks/src/api.js", line 605, in Proxy.triggerHookCommon
    return res.reply(resp);
  ?, in runMicrotasks
```

djmitche:
Github Bug/Issue: Fixes #4417

### DIFF
--- a/changelog/issue-4417.md
+++ b/changelog/issue-4417.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 4417
+---
+The `hooks.triggerHook` method no longer fails with a 500 error, and now correctly includes the `taskQueueId` property.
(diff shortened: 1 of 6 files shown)


imbstack (approved):
just need to generate

djmitche:
that's twice I've forgotten `yarn generate` today :(

PR #4419, opened 2021-02-16
Allow specifying a content-type for link artifacts

### LINKED ISSUE #4411: Link artifact logs do not use the log viewer
**Describe the bug**
When using link artifacts for live logs, the redirect works but the log is shown as plain text in a new tab.

**To Reproduce**

1. Create a new task which takes its own task ID.
2. In the task, create a log artifact (content-type 'text/plain' & .log extension).
3. Create a link artifact to the log artifact.
3. In the task view, notice that clicking on the log artifact displays the log viewer.
4. Notice that clicking the link artifact does *not* show the log viewer.

**Expected behavior**
Link artifacts which link to log artifacts should use the log viewer.

**Taskcluster version**
v41.0.0

**Additional context**
It looks like this is because `listArtifacts` returns an empty `contentType` for link artifacts. There are two fixes I can think of:
a) Make it so the DB function does a JOIN for link artifacts to retrieve the content type.
b) Include the artifact name in the DB function and have the taskcluster UI piece together the information.

djmitche:
Github Bug/Issue: Fixes #4411

### DIFF
--- a/changelog/issue-4411.md
+++ b/changelog/issue-4411.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 4411
+---
+The `queue.createArtifact` method now allows specifying a contentType for "link" artifacts, which is necessary to indicate to the UI that a link can be viewed as a logfile.
--- a/clients/client-go/tcqueue/types.go
+++ b/clients/client-go/tcqueue/types.go
@@ -253,6 +253,15 @@ type (
 		// Name of the artifact to which to link.
 		Artifact string `json:"artifact"`
 
+		// Expected content-type of the artifact.  This is informational only:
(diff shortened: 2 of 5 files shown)


imbstack (approved):

djmitche:
@ricky26 can you confirm that this looks like it will fix the issue?  If so, I'll make a 41.0.1 release including this.

ricky26:
Yeah, this looks like a pretty low-complexity way of making everything work. :slightly_smiling_face: Thanks for the quick turn around! (A point release would be much appreciated.)

djmitche:
v41.0.1 = https://github.com/taskcluster/taskcluster/commit/39c33ece400d484852be6775cbd10ecee083bf2e

PR #4427, opened 2021-02-17
Document how to generate timestamps in Rust

### LINKED ISSUE #4405: Add something similar to fromNow to rust client
The python and JS libraries have a `fromNow(..)` function that takes a string and returns a timestamp suitable for use in API payloads.

Rust already has some nice date-time arithmetic, so a `fromNow` that parses strings is probably not a good idea.  But we should have a supported way to generate datestamps in API payload.

djmitche:
Github Bug/Issue: Fixes #4405


petemoore (approved):
Looks good to me, thanks for the doc update!

PR #4428, opened 2021-02-17
Allow projectId in hooks.triggerHook

### LINKED ISSUE #4417: hooks.triggerHook response does not allow taskQueueId

djmitche:
A few other options turned out to be difficult or impossible:
 * `$ref` between services' schemas aren't allowed (and changing this would be hard)
 * Just copying `task-status.yml` from `services/queue` to `services/hooks` does not work since it uses `$ref` to pull in lots of other queue schemas
 * Switching the `triggerHook` response body to just `{type: "object", additionalProperties: true, description: "This is the response from queue.createTask.."}` would change the Golang API incompatibly, requiring a major version bump and potentially doing more harm than a bugfix should.
 * Setting `additionalProperties: true` but keeping all of the defined properties also changes the Golang API incompatibly (replacing the whole thing with `json.RawMessage`.

Github Bug/Issue: Fixes #4417

### DIFF
--- a/changelog/issue-4417.md
+++ b/changelog/issue-4417.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 4417
+---
+In a followup to a bug partially fixed in v41.0.1, the `hooks.triggerHook` function no longer crashes due to the `projectId` property from `queue.createTask`.
--- a/clients/client-go/tchooks/types.go
+++ b/clients/client-go/tchooks/types.go
@@ -290,6 +290,18 @@ type (
 		// must have an expiration that is no later than this.
 		Expires tcclient.Time `json:"expires"`
 
+		// The name for the "project" with which this task is associated.  This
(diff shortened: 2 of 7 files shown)


djmitche:
I confirmed that I can trigger a hook in my dev environment with this patch included.

imbstack (approved):
seems like the best fix given the conditions

PR #4440, opened 2021-02-18
Check upload return values

djmitche:
I discovered that the AWS backend's `putUrl` impl was returning data in the wrong format.  The fix is one-line, but this also adds some tests to avoid this in the future.

### DIFF
--- a/libraries/validate/src/index.js
+++ b/libraries/validate/src/index.js
@@ -118,10 +118,12 @@ class SchemaSet {
     return (obj, id) => {
       id = id.replace(/#$/, '');
       id = id.replace(/\.ya?ml$/, '.json');
-      if (!_.endsWith(id, '.json')) {
-        id += '.json';
+      if (!/#/.test(id)) {
+        if (!_.endsWith(id, '.json')) {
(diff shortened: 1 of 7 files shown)


libraries/validate/src/index.js
```diff
@@ -118,10 +118,12 @@ class SchemaSet {
     return (obj, id) => {
       id = id.replace(/#$/, '');
       id = id.replace(/\.ya?ml$/, '.json');
-      if (!_.endsWith(id, '.json')) {
-        id += '.json';
+      if (!/#/.test(id)) {
+        if (!_.endsWith(id, '.json')) {
+          id += '.json';
+        }
+        id += '#';
```
imbstack, line 125: I'm surprised we've had this wrong for so long. This was added like 4.5 years ago!

imbstack (approved):

libraries/validate/src/index.js
djmitche: I think this is really just adding a feature we've never needed before -- all of the other validation we do is by "file", not sub-schema.

PR #4442, opened 2021-02-18
Drop all fields but taskId from trigger-hook method responses

### LINKED ISSUE #4437: hooks.triggerHook should just return a taskId
Currently this method returns whatever it gets back from `queue.createTask`, which as we've seen in #4417 that can change as we amend the task definition, and can cause validation errors if we don't update that schema properly.

This is the only place where one service returns another service's output unchanged.  And, it's not really necessary -- in fact, all the caller really needs to know is the taskId, with which they could call `queue.status` to get the remaining data if required.

So, let's just return the taskId.  For compatibility, we could return that at the same JSON path where it's always been returned (`status.taskId`), so callers looking for the taskId do not need to change.  So the response would be something like

So
https://github.com/taskcluster/taskcluster/blob/422fb15eabb23d4d6e2efdfc21a118b484a9a454/services/hooks/schemas/v1/trigger-hook-response.yml#L8-L14
would change to
```yaml
anyOf:
 - properties:
     taskId:
       title:          "Task Identifier"
       description: |
         TaskId of the task created by triggering the hook.
       type:           string
       pattern:        {$const: slugid-pattern}
     status:
       type: object
       properties:
         taskId:
           description: |
             Another copy of the taskId, at the location where it was published in
             Taskcluster versions before v42.  Prefer to use the top-level property,
             as this property will be removed in future versions.
           type:           string
           pattern:        {$const: slugid-pattern}
       additionalProperties: false
       required: []
     additionalProperties: false
     required: [taskId]
 - type: object
   description: Empty response indicating no task was created
   properties: {}
   additionalProperties: false
   required: []
```

djmitche:
Github Bug/Issue: Fixes #4437.


djmitche:
(we can wait a bit before landing this, since this will put us onto v42)

imbstack (approved):
makes sense!

PR #4464, opened 2021-02-24
Fix the putUrl response body, where headers must be strings

djmitche:
The missing `await` meant that the schema verification wasn't actually verifying.

I have now checked that it actually fails when the value does not match the schema.  I had done so in the last PR as well, but before factoring out the `helper.assertSatisfiesSchema` function.


imbstack (approved):
ah yes, the missing `await`. scourge of js projects everywhere

PR #4466, opened 2021-02-24
[client-js] Add upload/download functionality

### LINKED ISSUE #4423: Add support for uploading and downloading to the JS client

djmitche:
This works against a "real" deployment (mine) but won't work against the current deployment, so left as a draft.

This API is good for now, but I can imagine cases where we will need more sophisticated support, such as for parallel uploads and downloads.  The `streamFactory` means that we can only read or write sequentially.  We can address that when we support such a thing!

Github Bug/Issue: Fixes #4423

### DIFF
--- a/changelog/issue-4423.md
+++ b/changelog/issue-4423.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4423
+---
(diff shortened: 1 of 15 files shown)


djmitche:
We should probably have a way for the caller to get the content-type, as well.

imbstack (approved):
looks great. the refactors are nice to have

Issue #4422, opened 2021-02-16
Add support for uploading and downloading to the Rust client


djmitche:

djmitche:
I'm estimating a bunch of points on this because I'm doing it first so will be breaking some new ground..

djmitche:
Work in progress here -- https://github.com/taskcluster/taskcluster/compare/main...djmitche:issue4422

The tricky bit has been testing.  I'd like to do the bulk of the testing as unit tests, rather than against a running service, with a few integration style tests against the running service.  So I needed a way to fake the object service.  I tried using `httptest` to fake it at the HTTP level, but this was going to require quite a bit of arcane `httptest::matcher` magic.  So I shifted to an approach similar to that used in generic-worker, defining an "interface" that the service client satisfies.  In Go, just implementing the right functions automatically implements an interface, but Rust requires a trait and an explicit `impl` block (which it will optimize away).  I had to do a little bit of extra gymnastics to keep this trait private.

The result is that I can inject various fakes at test time, to validate the all of the branches of the `upload` method.

djmitche:
I'm going to set this aside for the moment.  Much of the last few days has been finding (minor) bugs in the object service, so that has worked out well.  However, finding a good async Rust API for reliable uploads and downloads is not trivial, and needs some more thought (lots of TODOs in the branch on the topic!).

PR #4477, opened 2021-02-26
Re-export chrono for user convenience

djmitche:
Just about any user of the TC client in Rust will need to generate
timestamps, so let's save the effort of adding a `chrono` dependency
(and any possible version conflicts) and re-export it directly.

Pulled from #4422.


imbstack (approved):
makes sense to me

PR #4493, opened 2021-03-03
[client-go] Add upload/download functionality

### LINKED ISSUE #4425: [client-go] Add support for uploading and downloading objects to/from the object service

petemoore:
Issue: Fixes #4425


clients/client-go/README.md
```diff
@@ -205,6 +205,67 @@ All timestamp arithmetic should be performed with the built-in `time` package.
 
 To generate SlugIDs, such as for TaskIDs, use [github.com/taskcluster/slugid-go](https://github.com/taskcluster/slugid-go).
 
+### Uploading and Downloading Objects
+
+The Object service provides an API for reliable uploads and downloads of large
+objects (files). This library provides convenience methods to implement the
+client portion of those APIs, providing well-tested, resilient upload and
+download functionality. These methods will negotiate the appropriate method
+with the object service and perform the required steps to transfer the data.
```
djmitche, line 214: Is it worth highlighting that these are not REST API methods?

clients/client-go/creds.go
```diff
@@ -65,6 +66,8 @@ type Client struct {
 	HTTPClient ReducedHTTPClient
 	// Context that aborts all requests with this client
 	Context context.Context
+	// HTTPBackoffClient holds the HTTP backoff settings for retries
+	HTTPBackoffClient *httpbackoff.Client
```
djmitche, line 70: This is new functionality for general clients, right?  Should it be highlighted in the README?

djmitche (commented):
I'll look some more, but .. were `upload_test.go` and `download_test.go` not `git add`ed?

clients/client-go/tcobject/download.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	var attempts int
+	resp, attempts, err = client.Retry(retryFunc)
+	if err != nil {
+		err = HTTPRetryError{
+			Attempts: attempts,
+			Err:      err,
+		}
+	}
+	defer resp.Body.Close()
```
djmitche, line 130: Should this be before the error conditional?  Why defer as the last item in a function?

clients/client-go/tcobject/upload.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/taskcluster/slugid-go/slugid"
+	tcclient "github.com/taskcluster/taskcluster/v42/clients/client-go"
+)
+
+const (
+	DataInlineMaxSize = 8192
+)
+
+// UploadFromBuf is a convenience method to publish an Object to the Object
+// Server with content buf and given projectID, name, contentType and expires.
```
djmitche, line 25:
```suggestion
// Service with content buf and given projectID, name, contentType and expires.
```

clients/client-go/tcobject/upload.go
```diff
@@ -0,0 +1,160 @@
...
+	defer file.Close()
+	fileInfo, err := file.Stat()
+	if err != nil {
+		return err
+	}
+	contentLength := fileInfo.Size()
+	return object.UploadFromReadSeeker(projectID, name, contentType, contentLength, expires, file)
+}
+
+// UploadFromReadSeeker publishes an Object to the Object Server, with given
```
djmitche, line 49:
```suggestion
// UploadFromReadSeeker publishes an Object to the Object Service, with given
```

clients/client-go/tcobject/upload.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+		b64 := base64.StdEncoding.EncodeToString(content)
+		proposedUploadMethods.DataInline = DataInlineUploadRequest{
+			ContentType: contentType,
+			ObjectData:  b64,
+		}
+	}
+
+	proposedUploadMethods.PutURL = PutURLUploadRequest{
+		// TODO: find out why this is float64 when it should be int64
```
djmitche, line 71: I think this is because it's a "number" in JSON.  Changing the json-schema type to "integer" will probably fix it.

clients/client-go/tcobject/upload.go
```diff
@@ -0,0 +1,160 @@
...
+			UploadID:              uploadID,
+			ProposedUploadMethods: proposedUploadMethods,
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	// Ensure FinishUpload is called even if upload fails by placing in a defer
+	// block.
```
djmitche, line 90: FinishUpload shouldn't be called if the upload isn't finished :)

clients/client-go/tcobject/upload.go
```diff
@@ -0,0 +1,160 @@
...
+		for headerName, headerValue := range uploadMethod.PutURL.Headers {
+			httpRequest.Header.Set(headerName, headerValue)
+		}
+		putResp, tempError = httpClient.Do(httpRequest)
+		if tempError != nil {
+			return
+		}
+		// bug 1394557: s3 incorrectly returns HTTP 400 for connection
+		// inactivity, so make this a temp error rather than a permanent error
+		// if it occurs, in case this is a put request to s3
```
djmitche, line 143: Can you add an issue to do the same in the other clients?

djmitche (changes requested):
I found a few minutes to look at `upload.go` and `download.go`.

It looks like the tests for these are under `workers/generic-worker/`.  I'd prefer to see tests in the package that they test.  It looks like those tests could be moved to `clients/client-go` without too much trouble.

clients/client-go/tcobject/download.go
petemoore: If there is an error, the response body will not be opened, so we should check for an error before closing the body.

Having a defer in there is just a best practice to help future-proof the code. Let's say something was added that read from the response body, it would need to return early if there was an error. Having the close in a defer just ensures that even if an error causes an early return, resources are still released. The small danger with not putting a defer block is that a future author may simply add something in before the `resp.Body.Close()` line thinking everything is ok, but if there is a possible return from the function introduced, the response body wouldn't get closed. This way, it is much harder to introduce such a resource leakage accidentally in a future modification of this routine.

clients/client-go/tcobject/upload.go
petemoore: Created #4623

clients/client-go/tcobject/upload.go
petemoore: Created #4493

clients/client-go/tcobject/download.go
```diff
@@ -0,0 +1,132 @@
...
+// the downloaded object.
+func (object *Object) DownloadToBuf(name string) (buf []byte, contentType string, contentLength int64, err error) {
+	writeSeeker := &writerseeker.WriterSeeker{}
+	contentType, contentLength, err = object.DownloadToWriteSeeker(name, writeSeeker)
+	if err != nil {
+		return
+	}
+	reader := writeSeeker.BytesReader()
+	buf = make([]byte, reader.Len())
+	_, err = writeSeeker.BytesReader().Read(buf)
```
djmitche, line 35:
```suggestion
	_, err = reader.Read(buf)
```

clients/client-go/tcobject/upload.go
```diff
@@ -0,0 +1,156 @@
...
+		// data is already uploaded -- nothing to do
+		return nil
+	case uploadResp.UploadMethod.PutURL.URL != "":
+		return putURLUpload(object.HTTPBackoffClient, uploadResp.UploadMethod, readSeeker)
+	}
+	return errors.New("Could not negotiate an upload method")
+}
+
+func putURLUpload(httpBackoffClient *httpbackoff.Client, uploadMethod SelectedUploadMethodOrNone, readSeeker io.ReadSeeker) error {
+	// perform http PUT to upload to S3...
```
djmitche, line 112:
```suggestion
	// perform http PUT to upload to the given URL
```

clients/client-go/tcobject/upload_download_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/cenkalti/backoff/v3"
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/assert"
+	"github.com/taskcluster/httpbackoff/v3"
+	"github.com/taskcluster/taskcluster/v42/clients/client-go/tcobject"
+	"github.com/taskcluster/taskcluster/v42/workers/generic-worker/mocktc"
```
djmitche, line 16: Let's move this package to `github.com/taskcluster/taskcluster/internal/mocktc` since it's now shared between this package and generic-worker.

clients/client-go/tcobject/upload_download_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (th *testHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	th.mu.Lock()
+	defer th.mu.Unlock()
+	switch th.attempts {
+	case 0:
+		w.WriteHeader(500)
+		_, err := w.Write([]byte("uhoh"))
+		if err != nil {
+			th.t.Fatal(err)
```
djmitche, line 121: This can't be called from goroutines other than the main test goroutine.  Maybe use a channel?

https://golang.org/pkg/testing/#T
> A test ends when its Test function returns or calls any of the methods FailNow, Fatal, Fatalf, SkipNow, Skip, or Skipf. Those methods, as well as the Parallel method, must be called only from the goroutine running the Test function.

djmitche (changes requested):
Looking good!

The `t.Fatalf` issue needs fixed.  If you'd prefer to move `mocktc` in a followup that's OK, too.

clients/client-go/codegenerator/model/model.go
```diff
@@ -256,10 +256,10 @@ func (apiDefs APIDefinitions) GenerateCode(goOutputDir string) {
 
 	for i := range apiDefs {
 		if strings.Contains(apiDefs[i].PackageName, "events") {
-			amqpApiLinks += "\n" + "* http://godoc.org/github.com/taskcluster/taskcluster/clients/client-go/" + apiDefs[i].PackageName + "\n"
+			amqpApiLinks += "\n" + "* http://godoc.org/github.com/taskcluster/taskcluster/v42/clients/client-go/" + apiDefs[i].PackageName + "\n"
```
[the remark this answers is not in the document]
petemoore: Created #4628.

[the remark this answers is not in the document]
petemoore: And #4630.

clients/client-go/tcobject/upload_download_test.go
petemoore: I agree. I tried this just now, and it created a lot of noise for this PR - is it ok for me to do that in a follow-up PR?

clients/client-go/tcobject/upload_download_test.go
petemoore: I decided to remove the response body entirely, as it isn't needed for the test, and then it saves the more complex error handling.

clients/client-go/tcobject/upload_download_test.go
djmitche: Yeah, sounds good as long as it's not P5 :)

djmitche (approved):
Looks good!

clients/client-go/tcobject/upload_download_test.go
djmitche: Created #4696

PR #4495, opened 2021-03-03
Fix 400 errors when creating / retriggering tasks

### LINKED ISSUE #4470: Task Creator fails when given taskQueueId
STR:
 * Go to task creator
 * enter
```yaml
retries: 5
created: '2021-02-25T22:26:27.058Z'
deadline: '2021-02-26T22:26:27.058Z'
expires: '2022-02-26T22:26:27.058Z'
taskQueueId: built-in/succeed
schedulerId: taskcluster-ui
tags: {}
payload: {}
extra: {}
metadata:
  name: task
  description: testing
  owner: taskcluster-internal@mozilla.com
  source: https://github.com/taskcluster/generic-worker
```
 * update the datestamps
 * create

You will get a 400 error with some Graphql complaints:

> Field \"provisionerId\" of required type \"String!\" was not provided."

Fix: the GraphQL should be treating this task definition as a single blob of JSON and just handing it off to the `queue.createTask` method, without interpretation.

djmitche:
Github Bug/Issue: Fixes #4470

### DIFF
--- a/services/web-server/src/graphql/Tasks.graphql
+++ b/services/web-server/src/graphql/Tasks.graphql
@@ -179,102 +179,6 @@ type Task {
   rawDefinition: JSON
 }
 
-input TaskInput {
-  # Unique identifier for a provisioner, that can supply a specified `workerType`.
-  provisionerId: String!
-
-  # Unique identifier for a worker-type within a specific provisioner.
-  workerType: String!
-
-  # Identifier for the scheduler that _defined_ this task.
-  # This can be an identifier for a user or a service like `"task-graph-scheduler"`.
-  # Along with the `taskGroupId`, this is used to form the permission scope
-  # `queue:assume:scheduler-id:<schedulerId>/<taskGroupId>`
-  # This scope is necessary to _schedule_ a defined task, or _rerun_ a task.
-  schedulerId: String
-
-  # Identifier for a group of tasks scheduled together with this task,
-  # by the scheduler identified by `schedulerId`. For tasks scheduled by the
-  # task-graph scheduler, this is the `taskGraphId`.
-  # Defaults to `taskId` if property isn't specified.
-  taskGroupId: String
-
-  # List of dependent tasks. These must either be _completed_ or _resolved_
-  # before this task is scheduled. See `requires` for semantics.
-  dependencies: [String]
-
-  # The task's relation to its dependencies. This property specifies the
-  # semantics of the `task.dependencies` property. If `ALL_COMPLETED` is given
-  # the task will be scheduled when all dependencies are resolved _completed_
-  # (successful resolution). If `ALL_RESOLVED` is given the task will be
-  # scheduled when all dependencies have been resolved, regardless of what
-  # their resolution is.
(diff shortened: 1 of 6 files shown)


services/web-server/src/graphql/Tasks.graphql
```diff
@@ -179,102 +179,6 @@ type Task {
...
-
-  # If present, these options are passed to the taskcluster-client's `use' method
-  # https://github.com/taskcluster/taskcluster/tree/main/clients/client#calling-api-end-points
-  # before calling queue.createTask.  This is typically used to specify
-  # `options.authorizedScopes`.  Note that the supplied options are not
-  # considered part of the task definition, and this value is only seen in calls
-  # to the createTask mutation.
-  options: JSON
-}
-
```
petemoore, line 277: Happy to see this duplication removed!

Are there any side effects of allowing an arbitrary blob (such as possible abuses)? I haven't really understood what role GraphQL plays when our client libraries are generated based on our schemas and references. Is GraphQL used just by the tools UI for the web client?

ui/src/utils/gqlToApi.js
```diff
@@ -0,0 +1,23 @@
+import { omit } from 'ramda';
+import cloneDeep from 'lodash.clonedeep';
+import removeKeys from './removeKeys';
+import { TASK_ADDED_FIELDS } from './constants';
+
+/**
+ * Given a task in the format embedded in graphql queries, convert it
+ * into a "normal" task payload.
+ */
```
petemoore, line 9: Is the graphql task format documented somewhere? If so it might be nice to cross reference it here.

petemoore (approved):
I'll let Brian do a deeper review as I haven't really understood what role GraphQL plays when our client libraries are generated based on our schemas and references. Is GraphQL used just by the tools UI for the web client?

For a future PR:
  * Do we have other redundant graphql models that could also be derived from our API schemas? If so:
    * could a similar approach be used to remove the duplication? Or
    * could we generate the graphql models?

The above questions are based on my (mis?)understanding that this issue arose due to a mismatch between our graphql model and our API schemas which were kept in sync by hand, and that this PR solved the issue for the task definition model by deriving it from the API schema rather than maintaining an equivalent graphql definition. If that isn't right, let me know!

ui/src/utils/gqlToApi.js
djmitche: The graphql schema itself "documents" it -- but it's not a public API, so we don't publish that in any way.

djmitche:
There's been a good bit of discussion, and in general I think we agree that graphql is not a good fit.  So, yes, there are other areas where graphql's idea of what the API looks like and the actual API differ, and those could probably be addressed piecemeal.  And, if we wanted to go all-in on graphql, we could probably do some sort of automatic generation.  But I think we want to go all-out on graphql instead (meaning, not use it at all).  But that's a big project and we have nobody to do it, so for the moment we are minimally patching what we've got to keep it operating.

imbstack (approved):

PR #4503, opened 2021-03-03
read client-library secrets into env vars for all client-library tasks

djmitche:
This adds credentials from a secret for all client-library tasks.  This will be necessary for the upload/download functionality.

### DIFF
--- a/test/client-library-secrets.py
+++ b/test/client-library-secrets.py
@@ -1,0 +1,31 @@
+#! /usr/bin/python
+
+# Python is the common scripting language between the docker images used to
+# test the various Taskcluster client libraries.  This fetches the
+# `testing/client-libraries` secret and writes out shell-compatible export
+# statements for each value it contains.
+
+try:
+    import urllib.request
+    PY27 = False
+except ImportError:
+    import urllib
+    PY27 = True
+import os
+import json
+
+proxy_url = os.environ['TASKCLUSTER_PROXY_URL'].rstrip('/')
+
+secret_url = proxy_url + '/api/secrets/v1/secret/project/taskcluster/testing/client-libraries'
+if PY27:
+    response = urllib.urlopen(secret_url)
+    if response.getcode() != 200:
+        raise RuntimeError('non-200 response from ' + secret_url)
+    secret = json.load(response)
+else:
+    with urllib.request.urlopen(secret_url) as response:
+        if response.status != 200:
+            raise RuntimeError('non-200 response from ' + secret_url)
+        secret = json.load(response)
+for k, v in secret['secret'].items():
+    print('export %s="%s"' % (k, v))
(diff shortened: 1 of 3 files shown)


djmitche:
https://github.com/mozilla/community-tc-config/pull/396 lands the client and secret this reads -- it's already applied.

test/client-library-secrets.py
```diff
@@ -0,0 +1,31 @@
...
+# test the various Taskcluster client libraries.  This fetches the
+# `testing/client-libraries` secret and writes out shell-compatible export
+# statements for each value it contains.
+
+try:
+    import urllib.request
+    PY27 = False
+except ImportError:
+    import urllib
+    PY27 = True
```
petemoore, line 13: What is the reason to support python 2.7?

test/client-library-secrets.py
```diff
@@ -0,0 +1,31 @@
...
+    if response.getcode() != 200:
+        raise RuntimeError('non-200 response from ' + secret_url)
+    secret = json.load(response)
+else:
+    with urllib.request.urlopen(secret_url) as response:
+        if response.status != 200:
+            raise RuntimeError('non-200 response from ' + secret_url)
+        secret = json.load(response)
+for k, v in secret['secret'].items():
+    print('export %s="%s"' % (k, v))
```
petemoore, line 31: I suspect the environment variable names and values here are known to be limited to a specific set of characters that do not need to be escaped, in which case this is ok. But it may still be worth a comment to explain why the key/value are not escaped on line 31. It would serve as a useful reminder in case other env vars get added in future that can potentially contain characters that require escaping. There is even an argument for escaping them, even if not currently needed, just for good hygiene.

test/client-library-secrets.py
```diff
@@ -0,0 +1,31 @@
+#! /usr/bin/python
+
+# Python is the common scripting language between the docker images used to
+# test the various Taskcluster client libraries.  This fetches the
+# `testing/client-libraries` secret and writes out shell-compatible export
+# statements for each value it contains.
```
petemoore, line 6: Let's highlight that this secret contains _taskcluster credentials_ (TASKCLUSTER_ACCESS_TOKEN / TASKCLUSTER_CLIENT_ID / TASKCLUSTER_ROOT_URL ?) required for the integration tests which connect to a real taskcluster deployment rather than a mock/fake.

petemoore (approved):

test/client-library-secrets.py
djmitche: The TC client still supports it, and also that's the version of Python available in the node images.  I wanted to write one script to do this secret-fetching that would work in all of the test docker images, and py27 was the common factor.

test/client-library-secrets.py
djmitche: That's a good point -- I'll add a comment.  We do the same already in `test/shared-secrets.js` -- this is in some ways a Python port of that script.

PR #4520, opened 2021-03-09
Fix errors creating tasks in the UI with priority or requires

### LINKED ISSUE #4519: createTask calls from the frontend mangle requires and priority
**Describe the bug**
Trying to create a task with `priority` or `requires` results in an error:
```
Schema Validation Failed! Rejecting Schema: https://taskcluster.DOMAIN/schemas/queue/v1/create-task-request.json# Errors:

data.requires should be equal to one of the allowed values
data.priority should be equal to one of the allowed values
```

The sent request clearly contains, for example: `priority: LOWEST` or `require: ALL_COMPLETED`. It looks like this is because of a transform which used to be required for the GraphQL task schema, which is no longer needed.

**To Reproduce**

1. Go to the create task page and fill in `priority` or `require`.
2. Click create

**Expected behavior**
Task creation with priority or requires should succeed.

**Taskcluster version**
v42.1.0

**Additional context**
@djmitche it looks like this is fallout from the GraphQL schema change.

Happy to put a PR in for this one, it seems pretty straight forward! (It looks like `formatTaskMutation` just needs to be removed.)

ricky26:
Github Bug/Issue: Fixes #4519


ricky26:
@imbstack this fix is for bugs _introduced_ in #4495. The removal of the schema for tasks did fix retriggering but a couple of places weren't updated to use the API task format instead of the old GraphQL one.

djmitche (approved):
Thanks!  I did miss those cases in my testing.

PR #4543, opened 2021-03-15
Update the codeql workflow to the latest version

### LINKED ISSUE #4523: Fix broken CodeQL scanning
CodeQL appears to have broken itself:
```
Warning: 1 issue was detected with this workflow: git checkout HEAD^2 is no longer necessary. Please remove this step as Code Scanning recommends analyzing the merge commit for best results.
Error: The repo on which this action is running is not opted-in to CodeQL code scanning.
```
if this is an easy fix, great!  Otherwise -- this has never identified anything useful, so let's just turn it off.

djmitche:
I found this by clicking the "set up CodeQL" option and copy-pasting.

Github Bug/Issue: Fixes #4523

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,36 +1,51 @@
-name: "Code scanning - action"
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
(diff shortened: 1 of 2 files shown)


jwhitlock (approved):
I like the separate per-language test steps. Maybe CodeQL will be more reliable with this change? 🤞

djmitche:
Here's hoping!

PR #4557, opened 2021-03-16
Add new queue artifact methods

### LINKED ISSUE #4547: Add `queue.artifact` that returns artifact metadata

djmitche:
Github Bug/Issue: Fixes #4547 (see that issue for background)


services/queue/src/artifacts.js
```diff
@@ -33,47 +121,6 @@ builder.declare({
...
-    'properties the request will succeed as the operation is idempotent.',
-    'This is useful if you need to refresh a signed URL while uploading.',
-    'Do not abuse this to overwrite artifacts created by another entity!',
-    'Such as worker-host overwriting artifact created by worker-code.',
-    '',
-    '**Immutability Special Cases**:',
-    '',
-    '* A `reference` artifact can replace an existing `reference` artifact`.',
-    '* A `link` artifact can replace an existing `reference` artifact`.',
-    '* Any artifact\'s `expires` can be extended.',
```
petemoore, line 76: It this removed here because it is moved to ui/docs/reference/platform/queue/artifacts.mdx ?

I liked having it in the reference section, because then it is generated e.g. in the go client. With it no longer there, users of the go client won't discover this summary, I wonder if it might be better to keep it here?

djmitche: Yes, because the artifact types are relevant to more than just createArtifact.  For example, artifact users should know that reference artifacts cannot be cached.

It's still in the reference section, just not tied to one specific endpoint.  Go (and rust and Python) client users couldn't understand Taskcluster as a whole just by reading the go docs, anyway -- they also need this kind of more-than-an-API-method context.

petemoore: I think it would be best to put it in the APIBuilder description of the Queue service. This way, it will still be included in the package docs of the queue clients (e.g. [here](https://pkg.go.dev/github.com/taskcluster/taskcluster/v42@v42.1.1/clients/client-go/tcqueue#pkg-overview), but won't be bound to specific API endpoints. My concern with moving service docs outside of the schemas and references is that information about the services doesn't have a single home, and it isn't clear where it should go. I think if we keep it in the service description, the clients should get it, the docs site, and anything that crawls the APIs etc. It is also much closer to the code, and having docs directly with the code seems like a good practice.

Is there an advantage of having it in ui/docs/reference/platform/queue/artifacts.mdx instead of services/queue/src/api.js ?

djmitche: Fair enough.  The advantage is that it shows up in the table of contents, but i can add some links.

jwhitlock (approved):
Thanks @djmitche, the code and tests look good to me. I looked at some of the doc strings on the generated code, and these look good as well, with some needed grammar updates.

PR #4561, opened 2021-03-17
Bug 1564512 - service/github: Do not ignore if no .taskcluster.yml in default branch

### LINKED ISSUE #3927: a .taskcluster.yml in a PR without one in the master branch is skipped
copied from https://bugzilla.mozilla.org/show_bug.cgi?id=1564512
---
https://github.com/taskcluster/taskcluster/blob/8bf75678626567ec2fdae3889dff2f6d86235958/services/github/src/handlers.js#L603-L605

This means experimenting with `.taskcluster.yml` in a PR silently does nothing.

Could we assume the most restrictive policy (collaborators), instead?

mraethel:
If the the repo has no `taskcluster.yml`, create a pseudo object for `defaultBranchYml` containing the necessary entries needed by the following collaborator checks. The Policy is set to `collaborators_quiet` since the issue comment created by `collaborators` does not reflect this case.

Bugzilla Bug: [1564512](https://bugzilla.mozilla.org/show_bug.cgi?id=1564512)

Github Bug/Issue: Fixes #3927

### DIFF
--- a/changelog/issue-4561.md
+++ b/changelog/issue-4561.md
@@ -1,0 +1,5 @@
+audience: deployers
+level: patch
+reference: issue 4561
+---
+The GitHub service now allows collaborators to test out a `.taskcluster.yml` in a PR, when there is no such file in the default branch initialized yet.
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -766,7 +766,9 @@ async function jobHandler(message) {
 
     if (!defaultBranchYml) {
       debug(`${organization}/${repository} has no '.taskcluster.yml' at ${defaultBranch}.`);
-      return;
+      
+      // If the repository does not contain a '.taskcluster.yml' file, collaborators should be able to test before initializing.
+      defaultBranchYml = {version: 1, policy: {pullRequests: 'collaborators_quiet'}};
     }
 
     if (this.getRepoPolicy(defaultBranchYml).startsWith('collaborators')) {
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -766,9 +766,10 @@ async function jobHandler(message) {
 
     if (!defaultBranchYml) {
       debug(`${organization}/${repository} has no '.taskcluster.yml' at ${defaultBranch}.`);
-      
-      // If the repository does not contain a '.taskcluster.yml' file, collaborators should be able to test before initializing.
-      defaultBranchYml = {version: 1, policy: {pullRequests: 'collaborators_quiet'}};
+
+      // If the repository does not contain a '.taskcluster.yml' file, collaborators should be able to test before
+      // initializing.
+      defaultBranchYml = { version: 1, policy: { pullRequests: 'collaborators_quiet' } };
     }
 
     if (this.getRepoPolicy(defaultBranchYml).startsWith('collaborators')) {
--- a/services/github/test/handler_test.js
+++ b/services/github/test/handler_test.js
@@ -617,7 +617,7 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
       });
       await simulateJobMessage({ user: 'imbstack', eventType: 'release' });
 
-      assert(github.inst(5828).repos.createCommitComment.callCount === 1);
+      assert(github.inst(5828).repos.createCommitComment.callCount === 0);
     });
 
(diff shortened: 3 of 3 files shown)


djmitche (commented):
Nice -- the change looks good!  Can you add
 * a test for this
 * a changelog entry (`yarn changelog`)
 * a commit message describing briefly what the commit does (and update the PR title to match)

mraethel:
Thanks for your quick feedback, I hope this satisfies the requirements now. I am not sure if the assure() statement is correct. It could be that I mixed up `repos.createCommitStatus.callCount` and `issues.createComment.callCount`, in which case I probably need to write a second body for the already existing issue comment at [here](https://github.com/taskcluster/taskcluster/blob/8a6a0e9415139c3294bfd1ebb6764b18f094a761/services/github/src/handlers.js#L810-L825), differentiating between the `.taskcluster.yml` existing and not existing. My local tests seem to pass though.

djmitche:
I'll have a look.  The `lint-nodejs` failure shows some linting errors, in particular tabs vs. spaces.  Usually `yarn lint --fix` will fix those up, and a few code-editor tweaks will help use spaces instead of tabs.  If the patch otherwise looks good, then I'll run `yarn lint --fix` and merge.  I'll have a deeper look soon.

djmitche (changes requested):
Thanks for the updates I requested last time!  Please also fix up the lint issues as I mentioned.

I think you're right about the test -- it would pass without the changes you made in `handlers.js`.  That's a good heuristic to judge the effectiveness of a test -- does it fail if the fix is not in place?

I can see two options to address the test:
 1. make the user a collaborator (`setRepoCollaborator`), and check that a task is actually created; or
 2. do not make the user a collaborator, but check that the `checkCollaborator` fake was called, which is only done if the "collaborators" policy is used.

I prefer the first one of those, since it more realistically tests the user's expectations (that a task is created).

Especially on mature software like Taskcluster, often writing the tests is the hard part :)

mraethel:
Ahh, I thougt imbstack was defined as collaborator globally. Now the over tests make sense as well...
Linting should now be fixed as well.

djmitche:
@mraethel how's this going?

mraethel:
My local test failed.
I guess the issue is that I don't quite understand what `github.inst(5828).repos.createCommitComment.callCount` does, I assumed it is called when the taskcluster tasks are being executed, but I'm prbably wrong about that.
Meaning if the previous code I wrote is correct, then the only thing left would be to find the right `assure` statement checking for the success of the github service or at least the [jobHandler](https://github.com/taskcluster/taskcluster/blob/main/services/github/src/handlers.js#L662) function.

djmitche:
`createCommitComment` is a means of calling [this API endpoint](https://docs.github.com/en/rest/reference/repos#create-a-commit-comment) which adds a comment to a GH commit.  There's also `createComment` which adds a comment to a GH issue or pull request.  Depending on whether it can figure out the relevant PR, taskcluster-github will use one or the other of those functions.

Thinking broadly about what the test is testing -- we want a collaborator's PR to have tasks created, even if there's no .taskcluster.yml in the default branch.  So breaking that in to "given" and "then":

Given:
 * a PR is opened; and
 * it contains a .taskcluster.yml
 * the author is a collaborator; and
 * there is no .taskcluster.yml
Then
 * Tasks should be created from the `.taskcluster.yml` in the PR.

So that "then" should correspond to the assertions/`assume` at the end of the test definition.

mraethel:
I think the latest commit should wrap this up then.

mraethel:
At least the mechanism for starting the tasks, the creation of tasks from the PR still needs to implemented.

djmitche:
That already happens -- if there *is* a `.taskclsuter.yml` on the default branch, the tasks for the PR are created from the version of that file in the PR.

djmitche (approved):
Looks good, and I checked that changing `goodBuddy` to `badBuddy` makes it fail, so it's a good test :)

djmitche:
Nice work on that!!

djmitche:
Oh, by the way, [GitHub Flow](https://guides.github.com/introduction/flow/) suggests using a branch other than `main` or `master` for making pull requests.  I often name mine after the issue I'm working on (so `issue3927` in this case).

mraethel:
Thanks! I will do it on the next one.

PR #4562, opened 2021-03-17
Switch to graphql-scalars

### LINKED ISSUE #4453: Stop using graphql-iso-date
This appears unmaintained, and says

> NOTICE: The scalars defined in this repository have moved to the GraphQL-scalars repository where they will be maintained.

djmitche:
This passes tests, and worked for me locally viewing a task (which contains DateTime values).  I don't think we actually use Date and Time, but in any case if DateTime works with this change, those do too.

Github Bug/Issue: Fixes #4453


djmitche:
Hm, GitHub seems to be timing out on the clone operation lately.  The change here only affects service-web-server, so the others are all GitHub failures.

yarn.lock
```diff
@@ -6299,47 +6301,52 @@ tar-stream@^2.0.0:
...
-    lodash "^4.17.4"
-    slugid "^2.0.0"
-    taskcluster-lib-urls "^13.0.0"
+  version "0.0.0"
+  uid ""
 
 "taskcluster-db@link:db":
-  version "42.1.1"
+  version "0.0.0"
+  uid ""
```
jwhitlock, line 6309: This change looks suspicious

jwhitlock (approved):
Thanks @djmitche, looks like a straightforward switch to the maintained library.

The new ``0.0.0`` versions of ``taskcluster-*`` libraries look strange, however.

yarn.lock
djmitche: It's a longstanding yarn bug that adds commit noise but doesn't actually hurt anything.

PR #4563, opened 2021-03-17
Fail with 400 when an object's backend is not found

### LINKED ISSUE #4420: Fail operations if backend is not found
```
  const backend = this.backends.forUpload({ name, projectId }); 
```

If this doesn't find anything, we should return a suitable error message.

djmitche:
Github Bug/Issue: Fixes #4420

### DIFF
--- a/changelog/issue-4420.md
+++ b/changelog/issue-4420.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4420
+---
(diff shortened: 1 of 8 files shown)


djmitche:
Ugh, this is a simple fix but devilishly hard to fix since it requires changing the backends config

imbstack (approved):
seems reasonable

djmitche:
Amusingly, this would have helped with debugging the config change today :)

PR #4564, opened 2021-03-17
yarn minify

djmitche:


djmitche:
`client-nodejs` looks like a yarn intermittent.  I'll close and re-open to try to trigger that again.

jwhitlock (approved):
🚢 SHIP IT! Quick, before the file gets updated again!

PR #4565, opened 2021-03-17
Remove use of deprecated typeface-roboto

### LINKED ISSUE #3993: Replace typeface-robot with fontsource
Per deprecation notice in https://github.com/KyleAMathews/typefaces

djmitche:
This also removes the use of the FontStager and unmaintained
react-fout-stager (not a typo!).  This tool had the effect of rendering
the UI before the fonts were loaded, which was ugly (that's what the U
stands for!).  Let's not do that!

Github Bug/Issue: Fixes #3993


imbstack (approved):
hooray! I don't remember what exactly but fout-stager ruined a few days for me a while ago

PR #4568, opened 2021-03-18
Upgrade to Ajv-7.x

### LINKED ISSUE #4133: Upgrade Ajv to 7.x
Continued from #4114 which also has a few commits and some notes.

djmitche:
Github Bug/Issue: Fixes #4133


petemoore (approved):
It looks ok to me, but I'd prefer a node expert to look through it, so this is more of a rubber stamp with my eyes closed.

djmitche:
@imbstack was involved with the Ajv work, so a second review from him would be great

imbstack (approved):
thanks very much. easy to review split into commits and it all makes sense

how did we live without strict mode for so long? way more mistakes than I thought we would end up having

djmitche:
Yeah, it's pretty scary!  We tried valiantly with the metaschema (or as I like to type it, meatschema), but it missed a lot of stuff.

PR #4569, opened 2021-03-18
balance load.{save,restore} in obj-service tests

djmitche:
This fixes a bug I introduced yesterday that caused failures in the condition where there are no AWS/GCP credentials available.


jwhitlock (approved):
👍

PR #4578, opened 2021-03-19
Make the task claim time configurable

### LINKED ISSUE #3981: Make reclaim time configurable per Taskcluster instance
At the moment, reclaim time is fixed as 20 minutes, if it was exposed as config, smaller instances could use a much smaller reclaim time to improve the responsiveness of cancellations (and thus save some compute resource).

djmitche:
..and document it

Github Bug/Issue: Fixes #3981


services/queue/src/workclaimer.js
```diff
@@ -129,7 +129,7 @@ class WorkClaimer extends events.EventEmitter {
    *   db:            // Database object
    *   queueService:  // queueService from queueservice.js
    *   monitor:       // monitor object from taskcluster-lib-monitor
-   *   claimTimeout:  // Time for a claim to timeout in ms
+   *   claimTimeout:  // Time for a claim to timeout in seconds
```
jwhitlock, line 132: nice catch!

ui/docs/manual/deploying/queue-config.mdx
```diff
@@ -0,0 +1,93 @@
...
+Until everything uses the Object Service, the queue service must be configured to support S3 artifacts.
+That involves the following Helm values:
+
+* `queue.artifact_region`
+* `queue.public_artifact_bucket`
+* `queue.private_artifact_bucket`
+* `queue.sign_public_artifact_urls` (optional, default false)
+* `queue.public_artifact_bucket_cdn` (optional)
+* `queue.aws_access_key_id`
+* `queue.aws_secret_access_key`
```
jwhitlock, line 20: Do we suggest defaults or typical values for Helm values?

jwhitlock (approved):
Thanks @djmitche, looks like a straightforward movement to a configured value, plus some good documentation additions and corrections.

ui/docs/manual/deploying/queue-config.mdx
djmitche: We don't really document Helm values in any consistent way right now.  None of the above would have defaults, aside from the one that is documented to do so.  And below `queue.task_claim_timeout`'s default is documented with some context on what appropriate values may be.

Or maybe I misunderstood the question?

jwhitlock: No, I think you understood. If I was setting these values, I'd be curious if ``queue.public_artifact_bucket`` was ``my-bucket-name`` or ``https://my-bucket-name.s3.us-east-2.amazonaws.com``. Just having the list is a huge improvement, and we can see if it makes sense to the users later.

djmitche: Good point -- I think at some point we will regularize all of this, at which time that might be a good improvement.

PR #4598, opened 2021-03-22
Bug 1419577 - fix handling of special characters in artifact names

djmitche:
Bugzilla Bug: [1419577](https://bugzilla.mozilla.org/show_bug.cgi?id=1419577)


services/object/src/api.js
```diff
@@ -11,6 +11,9 @@ const DOWNLOAD_METHODS = [
   'HTTP:GET',
 ];
 
+const NAME_PATTERN = /^[\x20-\x7e]+$/;
```
jwhitlock, line 14: This made me miss my C days. I had to load up an ASCII table to check it out. It is space, letters, numbers, and all the expected punctuation characters from a US keyboard.

services/object/src/api.js
```diff
@@ -42,6 +49,7 @@ builder.declare({
     'multiple times if necessary, either to propose new upload methods or to',
     'renew credentials for an already-agreed upload.',
     '',
+    'The `name` parameter can contain any printable ASCII character (0x20 - 0x7e).',
```
jwhitlock, line 52: 👍 thanks for the doc update

jwhitlock (approved):
Thanks @djmitche, this is a good change with a good set of tests.

I skimmed [bug 1419577](https://bugzilla.mozilla.org/show_bug.cgi?id=1419577), is there a reason it needs to stay open?

services/object/src/api.js
djmitche: It's the same pattern we use for scopes, so i didn't think to comment it.  I'll add a comment and land.

djmitche:
I'll close the bug once this lands.

glandium:
Is this deployed on community-tc? I'm getting 403s on artifacts with a + in their name. e.g. https://community-tc.services.mozilla.com/api/queue/v1/task/aRUxZsQfQcSMmbDR6Qt7nQ/artifacts/public/mercurial-0+hg32399d0813e0-cp27-cp27mu-linux_x86_64.whl

djmitche:
No, sorry -- this is in the object service, which isn't used for artifacts yet.

djmitche:
(tracked by https://github.com/taskcluster/scrum/issues?q=is%3Aissue+is%3Aopen+label%3Ainitiative%3Aobject-service)

PR #4601, opened 2021-03-22
Drop support for legacy assume:worker-id:.. queue scopes

### LINKED ISSUE #4586: Deprecate `assume:worker-id:..` scopes
This scope pattern is already deprecated in almost every API endpoint *except* `queue.createArtifact`.

```
  scopes: { AnyOf: [
    'queue:create-artifact:<taskId>/<runId>',
    { AllOf: [
      'queue:create-artifact:<name>',
      'assume:worker-id:<workerGroup>/<workerId>',                                                                                
    ] },
  ] },
```

I suspect it's unused there, as well -- that all usage by workers uses the first alternative in the AnyOf.  If so, let's consider it to have been completely deprecated and remove support.  Otherwise, let's fix this usage and declare a future date after which support will be removed.

* https://github.com/taskcluster/taskcluster-rfcs/pull/71
* https://bugzilla.mozilla.org/show_bug.cgi?id=1374978

djmitche:
Github Bug/Issue: Fixes #4586


imbstack (approved):

imbstack:
Although tests are failing

djmitche:
I'll have a look-see, thanks!

djmitche:
I hadn't put the right creds in to run `artifact_test.js` locally, so I had forgotten to update the scopes it uses.  I force-pushed with only changes to that file, so I'll merge if green.

djmitche:
The failures here are all due to a yarn outage, it looks like.

PR #4613, opened 2021-03-23
Update documentation for queue.rerunTask

### LINKED ISSUE #4587: Update rerunTask docs regarding completed, unscheduled tasks
https://bugzilla.mozilla.org/show_bug.cgi?id=1458896

djmitche:
..and add some tests to verify the new docs are accurate.

Github Bug/Issue: Fixes #4587

### DIFF
--- a/services/queue/test/reruntask_test.js
+++ b/services/queue/test/reruntask_test.js
@@ -35,15 +35,22 @@ helper.secrets.mockSuite(testing.suiteName(), ['aws'], function(mock, skipping) 
       purpose: 'taskcluster-testing',
     },
   };
+  let taskId;
 
-  test('create, claim, complete and rerun (is idempotent)', async () => {
-    const taskId = slugid.v4();
+  setup(() => {
+    taskId = slugid.v4();
+  });
 
+  const createTask = async taskDef => {
     debug('### Creating task');
     await helper.queue.createTask(taskId, taskDef);
     helper.assertPulseMessage('task-defined');
-    helper.assertPulseMessage('task-pending', m => m.payload.runId === 0);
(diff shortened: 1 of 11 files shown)


djmitche:
Red x's are
```
error An unexpected error occurred: "https://registry.yarnpkg.com/@slack/logger/-/logger-2.0.0.tgz: Request failed \"500 Internal Server Error\"".
```
we should probably figure out a way to use a yarn cache in our CI..

services/queue/test/reruntask_test.js
```diff
@@ -35,32 +35,64 @@ helper.secrets.mockSuite(testing.suiteName(), ['aws'], function(mock, skipping)
...
     });
     helper.assertPulseMessage('task-running');
+  };
+
+  test('rerun a scheduled task', async () => {
+    // create a self-dependent task, so it is in state 'scheduled'
+    await createTask({ ...taskDef, dependencies: [taskId] });
+
+    debug('### Requesting task rerun');
+    helper.scopes('queue:rerun-task:my-scheduler-extended-extended/dSlITZ4yQgmvxxAi4A8fHQ/*');
```
petemoore, line 68: How was this task originally created?
Do we need to account for it in any development docs?

petemoore (approved):

services/queue/test/reruntask_test.js
djmitche: It's created on line 65 :)

djmitche: Ah, the `dslITZ..` slugId is the taskGroupId, which is hard-coded a bit further up.  None of this stuff runs against the production queue -- it's running against the queue under test.

PR #4615, opened 2021-03-23
Do not consider quarantined workers in calcuation of current/existing worker pool capacity

### LINKED ISSUE #3794: Quarantined workers still contribute to capacity estimation
**Describe the bug**
If a worker on a cloud provisioner with a high capacity is quarantined it can take a long time for new tasks to get scheduled because the quarantined worker still contributes to available capacity.

**To Reproduce**

1. Create a new worker pool using a gcp provider, with `idleTimeoutSecs` at least 600.
2. Create a task so that it spins up a new worker.
3. Quarantine the worker.
4. Create a new task.
3. Notice that the task remains pending (and no worker is created) for at least 10 minutes.

**Expected behavior**
Quarantined workers should not contribute to existing capacity and new workers should be spun up to handle the currently pending tasks.

**Taskcluster version**
v37.3.0

It looks like this could just be adjusted in worker-manager's provider, similarly to how requested workers don't contribute to total requested capacity.

I'm happy to attempt a PR for this if there are no terrifying gotchas I'm not aware of!

djmitche:
Github Bug/Issue: Fixes #3794


imbstack (approved):
thanks, makes sense. this functionality seems to have infinite things to take into account that we haven't yet

djmitche:
BTW this is done in a way that will be forward-compatible when we get to the point of having better state sharing between the queue and worker-manager.

PR #4620, opened 2021-03-24
Separate generated code comments from generated package docs with bla…

petemoore:
The go client godocs contained spurious code comments which were not intended to be part of the package docs.

This PR introduces a blank line between the generated code comments and generated package comments, so that the code comments are not included in the package docs.


petemoore:
Note, it looks like other generated comments may need some attention too, perhaps when we imported the clients into the monorepo some of the formatting of links etc was broken. We can fix that up in subsequent PRs.

djmitche (approved):

PR #4621, opened 2021-03-24
Objects are immutable

djmitche:
I figure we should say this somewhere :)


jwhitlock (approved):
Thanks @djmitche, that may save some user frustration!

PR #4622, opened 2021-03-24
Use integer instead of number for object service PutURLUploadRequest.…

petemoore:
Originated from https://github.com/taskcluster/taskcluster/pull/4493#discussion_r599006194


djmitche (approved):

djmitche:
`service-web-server` is a 500 from yarn, and there's no need for a changelog on this PR.

PR #4628, opened 2021-03-25
Move http://godoc.org links to https://pkg.go.dev and include version in path

petemoore:
Originated from https://github.com/taskcluster/taskcluster/pull/4493#discussion_r600833582

### DIFF
--- a/changelog/PMBzq5oKTVOV6usMDCeJ_g.md
+++ b/changelog/PMBzq5oKTVOV6usMDCeJ_g.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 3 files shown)


petemoore:
Note, the `v42` will get updated automatically on release by [this line](https://github.com/taskcluster/taskcluster/blob/9a123cb0432a5ee738cee97ee2d8fedbc98f89ee/infrastructure/tooling/src/release/tasks.js#L200).

djmitche (approved):
Nice!  I think there's a bugzilla bug around somewhere to switch to pkg.go.dev too.  We'll find it in triage eventually :)

PR #4632, opened 2021-03-25
Require content-encoding: identity for putUrl uploads.

### LINKED ISSUE #4610: Set Content-Encoding: identity on S3 objects
When uploading an object to an S3(-like) backend, require that it be uploaded with `Content-Encoding: identity` so that uploaders cannot surreptitiously set another encoding.  S3 does not support content negotiation, so all artifacts must be in the `identity` encoding.  GCP can do better.

djmitche:
There's room for some nuance here -- if this header *isn't* set, then a backend might support other content-encodings.  But, which?  For example, GCS only supports `gzip`.  Instead of trying to address that here, let's leave the `putUrl` method simple, and always require `identity`.  The more advanced, cloud-specific upload methods can be more specific.

Github Bug/Issue: Fixes #4610

### DIFF
--- a/changelog/issue-4610.md
+++ b/changelog/issue-4610.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4610
+---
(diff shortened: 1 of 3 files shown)


petemoore (approved):
Good idea, I like this.

PR #4636, opened 2021-03-25
Disable testing the simple engine on macos

djmitche:
The single mac worker we have is by far the slowest worker, and its
slowness causes delays to merging PRs, especially Renovate PRs.  This
will cut its load by 50% and hopefully allow it to keep up.

### DIFF
--- a/changelog/CxWEWKD_Rx6GXxX-vqDtWQ.md
+++ b/changelog/CxWEWKD_Rx6GXxX-vqDtWQ.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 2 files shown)


djmitche:
Even on this PR, it's been over an hour and still waiting on the Mac task.

petemoore (approved):
A longer-term solution would be to introduce a second mac to the worker pool, but I guess this is ok for now.
One concern I have is that firefox ci uses the simple engine on macOS (afaik) - perhaps we should think about EOL'ing it after checking with Release Engineering that they are not blocked on migration. Also we should check with Well Played Games and also our community customers, if they use Simple Engine on macOS. For community customers we can probably audit the worker pools ourselves to find out.

petemoore:
Another way to solve this problem would be to only run generic-worker tasks if something in workers/generic-worker has changed......... We could possibly rerun the generic-worker build task if the go client has been updated, but we probably don't need to run the tests (which is the main time consumer). Rerunning all generic-worker tests on all platforms/engines every time _anything_ in taskcluster changes, on a single worker, is really the root problem I think.

That would massively cut the load.

djmitche:
That worker used to keep up just fine, so maybe it just needs a reimage?

Those are good longer-term fixes, but this will solve the shorter-term issues at least.

Issue #4631, opened 2021-03-25
Make a new deployment outside of GCP, fix resulting issues


djmitche:
I'm working on this with @masterWayZ.

Issues encountered:
 * ~~update `development-process.md` (#4638)~~
   * ~~move to another file, since this is not required for development~~
   * ~~azure account is not required~~
   * ~~postgres version 11.x~~
 * Is Helm 2 still supported? (#4758)
 * AWS is not required initially (after #4549 lands)
 * CloudAMQP's free tier is not enough (doesn't allow managing users, perms) (#4758)
 * support for non-GKE clusters (#3295)
 * bugs in `dev:init` (#4757)
 * TC requires RabbitMQ use TLS (#4648) 
 * `yarn build --push` tries to push to the official repo (#4649)

djmitche:
@MasterWayZ how's this going?  Are there issues to file based on your notes?  Maybe even PRs? :)

djmitche:
I think everything here is broken out into other issues or PRs -- @MasterWayZ is there anything else you would like to address here, or should we close this issue?

MasterWayZ:
@djmitche I think we have everything, unless you also would want a doc on setting up the stateless DNS server? Other than that I can't think of anything else.

djmitche:
I'm hesitant to document that at all!  It's really not a sustainable solution, or one we would be comfortable fixing bugs in.

PR #4638, opened 2021-03-25
Reorganize docs on dev deployments

djmitche:
@MasterWayZ please have a look?  There's a TODO in here that I need some help with.

Github Bug/Issue: Refs #4631.


MasterWayZ:
Will post results here.

One thing I noticed in the commit, is that at the top it says ` *Helm installed (either 2 or 3 should Just Work)` on line 14, indicating either Helm 2 or 3, but then further down on line 94, it says `* Helm error `Error: stat taskcluster: no such file or directory` - make sure you have helm3 installed.` indicating that you need helm 3.

djmitche:
Good point!  @imbstack should we just drop Helm 2 support?

PR #4640, opened 2021-03-26
Update postgres.sql to work for new deployments

MasterWayZ:
In an effort with Dustin, I'm working on deploying Taskcluster and we try to fix all issues we encounter (or document) along the way. Some of this is tracked in issue 4631.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -7,7 +7,7 @@ env:
   mocha: true
   es6: true
 parserOptions:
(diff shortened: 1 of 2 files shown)


MasterWayZ:
The changes work but it seems to upset lint-nodejs.

djmitche:
(a missing changelog is fine for this cange)

djmitche (approved):

PR #4647, opened 2021-03-26
Log rather noisily when pulse blocks startup

### LINKED ISSUE #4646: Log at the warning level when unable to connect to pulse on startup
Apps that publish pulse messages block startup on pulse being initialized (so they can declare exchanges, etc.).

I think that's OK -- the app shouldn't do anything until pulse is working, anyway.  But, there's no logging about this.

Let's log INFO when that connection begins, INFO when it's complete, and WARNING after, say, 30s of startup not completing.

djmitche:
We came across this in #4631, and I think that's not the first time I've been left puzzling about this.  It's really the only reason that a process would *hang* on startup and not crash.  I think this bit of logging should at least make that hang visible.

Github Bug/Issue: Fixes #4646

### DIFF
--- a/libraries/pulse/src/publisher.js
+++ b/libraries/pulse/src/publisher.js
@@ -195,7 +195,19 @@ class PulsePublisher {
     this._setChannel(null);
     this.stopHandlingConnections = this.client.onConnected(this._handleConnection);
 
-    await this._assertExchanges();
+    // these tasks can stall waiting for a rabbitmq connection, so we log when
+    // they begin and end, and if they take too long
+
+    this.client.monitor.info('Asserting pulse exchanges exist');
+    const msg = 'Still attempting to assert pulse exchanges; is RabbitMQ configuration correct?';
+    const interval = setInterval(() => this.client.monitor.warning(msg), 30000);
+    try {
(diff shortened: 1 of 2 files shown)


libraries/pulse/src/publisher.js
```diff
@@ -195,7 +195,19 @@ class PulsePublisher {
     this._setChannel(null);
     this.stopHandlingConnections = this.client.onConnected(this._handleConnection);
 
-    await this._assertExchanges();
+    // these tasks can stall waiting for a rabbitmq connection, so we log when
+    // they begin and end, and if they take too long
+
+    this.client.monitor.info('Asserting pulse exchanges exist');
+    const msg = 'Still attempting to assert pulse exchanges; is RabbitMQ configuration correct?';
+    const interval = setTimeout(() => this.client.monitor.warning(msg), 30000);
```
imbstack, line 203: What about `setInterval` for this?

imbstack (approved):

libraries/pulse/src/publisher.js
djmitche: Whoops, that's what I meant (that's why I used `interval`)!

PR #4658, opened 2021-03-26
Use task credentials in built-in-workers

### LINKED ISSUE #4655: built-in-workers must use claim scopes to resolve tasks
Previously, this service had `queue:resolve-task`:
https://github.com/taskcluster/taskcluster/blob/bd0db705c6c939dd68c0958c3d5dc63ec15f2bb2/services/built-in-workers/scopes.yml#L4
which let it resolve any task.  But, after #4586 that scope is no longer used.

Like any other worker, this worker should be using the temporary scopes it gets back from the queue to resolve tasks.

This is _broken in main_ and must be fixed before the next release

djmitche:
Github Bug/Issue: Fixes #4655

### DIFF
--- a/changelog/issue-4655.md
+++ b/changelog/issue-4655.md
@@ -1,0 +1,5 @@
+audience: deployers
+level: patch
+reference: issue 4655
+---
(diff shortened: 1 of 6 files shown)


imbstack (approved):
huh

PR #4689, opened 2021-03-31
Roles list update after deleting a role

### LINKED ISSUE #2393: [UI] Add a delete action in roles, clients, secrets, hooks views

mrsVelena:
Github Bug/Issue: Fixes #2393 for Roles page

### DIFF
--- a/changelog/issue-2393.md
+++ b/changelog/issue-2393.md
@@ -1,0 +1,5 @@
+audience: developers
+level: patch
+reference: issue 2393
+---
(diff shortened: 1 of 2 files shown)


djmitche (approved):
This looks great -- thanks for remembering this after the issue was _again_ auto-closed!

PR #4693, opened 2021-04-01
Allow generic-worker livelog ports to be configured

### LINKED ISSUE #4691: Running two generic-worker instances on the same machine causes tasks to fail
**Describe the bug**
When running multiple simple workers on the same machine at the same time tasks fail with an ominous 'exited with status 1', in the worker logs, there is a message from livelog:

`listen tcp :60099: bind: address already in use`

It looks like these port numbers are hard-coded.

**To Reproduce**

1. Setup a generic-worker instance on a machine.
2. Copy the setup and give it a unique worker ID.
3. Start both instances.
4. Start a task which should pick either worker.
5. Start another task.

**Expected behavior**
Concurrent tasks with multiple workers should succeed.

**Taskcluster version**
v42.1.1

**Screenshots**

 - OS: macOS

**Additional context**

This looks simple, if there are no objections to just adding this to the worker config, I'm happy to PR it.

ricky26:
Github Bug/Issue: Fixes #4691 

Not sure if it's appropriate to use a unit test for this. I've updated the existing ones to use the config field for setting the ports they use, so in some means it is tested.

If this isn't appropriate, give me a shout. :smile:

### DIFF
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -207,6 +207,10 @@ func loadConfig(configFile *gwconfig.File) error {
 			DownloadsDir:                   "downloads",
 			IdleTimeoutSecs:                0,
 			LiveLogExecutable:              "livelog",
+			// The ports on which the livelog process listens locally.  These ports are not exposed
+			// outside of the host.  However, in CI they must differ from those of the generic-worker
+			// instance running the test suite.
+			LiveLogPortBase:                60098,
 			NumberOfTasksToRun:             0,
 			ProvisionerID:                  "test-provisioner",
 			RequiredDiskSpaceMegabytes:     10240,
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -207,9 +207,6 @@ func loadConfig(configFile *gwconfig.File) error {
 			DownloadsDir:                   "downloads",
 			IdleTimeoutSecs:                0,
 			LiveLogExecutable:              "livelog",
-			// The ports on which the livelog process listens locally.  These ports are not exposed
-			// outside of the host.  However, in CI they must differ from those of the generic-worker
-			// instance running the test suite.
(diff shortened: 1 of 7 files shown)


workers/generic-worker/main.go
```diff
@@ -207,6 +207,10 @@ func loadConfig(configFile *gwconfig.File) error {
 			DownloadsDir:                   "downloads",
 			IdleTimeoutSecs:                0,
 			LiveLogExecutable:              "livelog",
+			// The ports on which the livelog process listens locally.  These ports are not exposed
+			// outside of the host.  However, in CI they must differ from those of the generic-worker
+			// instance running the test suite.
```
djmitche, line 212: The "in CI" in this comment means "in the g-w tests", so I think this comment would be better where the value is set to 30538 in `helper_test.go`.

workers/generic-worker/usage.go
```diff
@@ -149,6 +149,9 @@ and reports back results to the queue.
           livelogExecutable                 Filepath of LiveLog executable to use; see
                                             https://github.com/taskcluster/livelog
                                             [default: "livelog"]
+          livelogPortBase                   Set the base port number for livelog. Livelog requires two
+                                            ports.
```
djmitche, line 153: perhaps mention that it will use the given port and `livelogPortBase + 1`?

djmitche (changes requested):
Looks good!  It'd be nice if these ports were determined dynamically by the kernel, but that's a much bigger project.

djmitche:
Looks like this has some lint / go-mod-tidy issues.  And needs a generate.

ricky26:
Right! Sorry about that, I didn't realise that my IDE _wasn't_ using gofmt. I guess it must have slightly different rules. I've run generate too.

djmitche (approved):
Looks good -- I'll merge when it's green

PR #4699, opened 2021-04-01
Add support for content-hashes on object upload

### LINKED ISSUE #4608: Add support for content hashes
When uploading an object, the caller should be able to supply content hashes using one or more algorithms.

The object service should store these values securely.  We talked about disallowing updates to the table containing them, but that would need more thought (an attacker could just delete and insert).

The object service should then make these values available on download.

Client libraries should offer convenience functionality to generate hashes, and pass that on upload.  Client libraries should verify hashes on download.

djmitche:
Github Bug/Issue: Fixes #4608 

The clients don't yet send the hashes in the `finishUpload` call, because the current deployed version of TC doesn't yet support that.  Once this PR is deployed, we can update the TODO items there.


db/test/fns/object_test.js
```diff
@@ -337,4 +337,104 @@ suite(testing.suiteName(), function() {
...
     assert.deepEqual(res, []);
   });
+
+  helper.dbTest('delete_object with hashes', async function(db) {
+    const expires = fromNow('1 day');
+    const uploadId = taskcluster.slugid();
+    await insertData([
+      { name: 'object-1', backend_id: 'be', project_id: 'prj', data: {}, expires, upload_id: uploadId },
+    ]);
+    await db.fns.add_object_hashes('object-1', JSON.stringify({ sha4: '1234', sha6: 'abcdef' }));
```
jwhitlock, line 347: I appreciate the use of made-up algorithms in the tests. I was curious if there was a whitelist of algorithms, and this confirms it is whatever the client wants to send, at least at this level.

jwhitlock (approved):
Thanks @djmitche, this looks like a clean addition of the multi-hash data, with client implementation and tests.

changelog/issue-4608-1.md
```diff
@@ -0,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4608
```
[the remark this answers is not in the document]
djmitche: Hah, good point.  I've gotten myself in the habit of `yarn changelog` before pushing so I don't waste CI resources, but apparently now I'm over-changelogging!

db/test/fns/object_test.js
djmitche: At the DB level there's no limit.  The API level does limit it, mostly to avoid the possibility of the number of hashes for an object growing without bound, or abuse of the field to store other data.

PR #4701, opened 2021-04-01
move mocktc into internal/

### LINKED ISSUE #4696: Move mocktc into `internal/`
Per https://github.com/taskcluster/taskcluster/pull/4493#discussion_r601591448

djmitche:
Github Bug/Issue: Fixes #4696


djmitche:
service-auth failure appears to be an azure outage

petemoore (approved):
This looks good! Too bad I didn't see it before we made the release, as I suspect that is what caused the merge conflicts. Once those are resolved, this looks fine to me. Thanks for moving it!

Note, the mock services were intentionally implemented over an HTTP interface, in order that the same mock services could be used for testing javascript code too, without us needing to implement mocks in both languages. One disadvantage of making this an internal package is that it makes it harder to do that, but since that isn't on our immediate roadmap, I don't see it as a blocker. We can always move it somewhere public if we decide to reuse the mocks for any javascript test suites.

The review approval is obviously subject to resolving the merge conflicts. :-)

djmitche:
Thanks!  Yeah, those /v42/ -> /v43/ changes are awful for merge conflicts :(

I suspect that if we did use this for JS, it'd be by building and running a binary, so this might work as-is!

petemoore:
> I suspect that if we did use this for JS, it'd be by building and running a binary, so this might work as-is!

That's true - good point!

djmitche:
worker-manager failure is #4744

PR #4706, opened 2021-04-02
Add an 'object' artifact storageType

### LINKED ISSUE #4548: Add an `object` artifact type to the queue

djmitche:
Github Bug/Issue: Fixes #4548


djmitche:
I'd appreciate a 10% review of this.  This is based on #4699 so I could get the DB versions right, so you can ignore most of the commits except the last three.  The bit I'd like feedback on is, does the process for uploading artifacts make sense?  Did I miss any critical bits?

djmitche:
OK, well, ready for 100% review now :)

services/queue/src/artifacts.js
```diff
@@ -323,6 +354,34 @@ builder.declare({
   }
 
   switch (artifact.storageType) {
+    case 'object': {
+      // use a generated clientId with useful information in it
```
imbstack, line 358: oh interesting

imbstack (approved):
seems reasonable to me but I would block on pmoore signing off

services/queue/src/artifacts.js
djmitche: we do it for task creds, too :)

petemoore: yes, nice pattern

services/queue/src/artifacts.js
```diff
@@ -358,6 +417,102 @@ builder.declare({
...
+      status: task.status(),
+      artifact: artifactUtils.serialize(artifact),
+      workerGroup: run.workerGroup,
+      workerId: run.workerId,
+      runId,
+    }, task.routes);
+  };
+
+  // if this is of a type that is already finished, that's not allowed
+  if (artifact.storageType !== 'object') {
```
petemoore, line 472: The API description lists the artifact types that do not support the finish method, rather than the one that does, but the check here is against the one that does ('object'). Perhaps the API description should instead just say that this is for object artifacts only, to match the code here. I also see that the description is helpful to understand _why_ reference, link and error do not have an associated finish call, so I'm also ok to leave the description as it is. I'll let you decide!

services/queue/src/artifacts.js
```diff
@@ -358,6 +417,102 @@ builder.declare({
...
+    if (err.statusCode !== 404) {
+      throw err;
+    }
+    return res.reportError('InputError', 'Object storing artifact data does not exist', {});
+  }
+  await this.db.fns.queue_artifact_present(taskId, runId, name);
+  await publishCreatedMessage();
+
+  // clear out the details, since the uploadId is no longer relevant
+  await this.db.fns.update_queue_artifact_2({
```
petemoore, line 504: `update_queue_artifact_2` doesn't feel like an ideal function name - is that new in this PR, or something that already existed?

services/queue/src/artifacts.js
```diff
@@ -379,6 +534,36 @@ let replyWithArtifactDownload = async function({ taskId, runId, name, req, res,
   res.set('x-taskcluster-artifact-storage-type', storageType);
 
   // Handle S3 artifacts
```
petemoore, line 536: This comment ("Handle S3 artifacts") should be further down in the code next to the s3 handling. It could probably be removed, as it is pretty obvious from the code.

services/queue/schemas/v1/object-credentials.yml
```diff
@@ -0,0 +1,29 @@
...
+  accessToken:
+    type:       string
+    minLength:  1
+    description: |
+      The `accessToken` for the temporary credentials.
+  certificate:
+    type:       string
+    minLength:  1
+    description: |
+      The `certificate` for the temporary credentials.
```
petemoore, line 24: Do we have a schema already for temp credentials that we could reuse here? It feels like this could be a shared definition.

petemoore (approved):

services/queue/src/artifacts.js
djmitche: Good point -- I switched this around late in the game (initially I allowed finishing any artifact, and just made it a no-op for most artifact types, but later decided that was too much flexibility).  I'll make a constant defining what does and does not require finishing.

services/queue/src/artifacts.js
djmitche: It's been around for a while.  We started out using `_2`, .. when we modified DB function signatures, but have lately been using slightly more descriptive suffixes like `_tqid` or `_with_foo`.

services/queue/schemas/v1/object-credentials.yml
djmitche: We do - `task-credentials.yml`.  However, it contains comment specific to task credentials (around expiration and usage), and I wanted to add comments specific to object credentials, so I split them up.

djmitche:
worker-manager failure is #4744

PR #4718, opened 2021-04-06
Always deprovision STOPPING azure workers

### LINKED ISSUE #4715: [azure-provider] STOPPING workers aren't further deprovisioned
When a worker goes into state STOPPING, as via a `removeWorker` call (either a REST API call, or because the worker is in `PowerState/stopped`, for example), no further deprovisioning is done.

The `checkWorker` call fetches the vm's state and, if the vm still exists calls `removeWorker`, which just marks it as STOPPING.

The only way `deprovisionResources` gets called is for the vm to not exist (404).

I think this was introduced in https://github.com/taskcluster/taskcluster/commit/66afe24d0a05e2a7e1fe8e6531cad5b855ff4e01 -- before that time, `removeWorker` also deprovisioned, so would begin by deleting the vm immediately.

I think the right fix is to call `depreovisionResources` whenever an instance is in the STOPPING state, regardless of whether the vm exists or not.

This is causing issues for the work on azure instance support in firefox-ci, so we may need to backport the fix.

djmitche:
Github Bug/Issue: Fixes #4715 

Reviewers: please review the first commit individually, as that's the one I'll backport to the v39.1.1 branch running in firefox-ci.  The refactor passes tests, but is higher-risk and I won't backport it.

### DIFF
--- a/changelog/issue-4715.md
+++ b/changelog/issue-4715.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 4715
+---
+The worker-manager service now deprovisions workers when `removeWorker` is called and when the workers terminate themselves.  Previously it would wait forever for such workers to be deleted, without attempting that deletion.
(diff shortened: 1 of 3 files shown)


imbstack (approved):
makes sense

Issue #4698, opened 2021-04-01
[client-py] call async upload/download from sync upload/download


djmitche:
Instead of implementing all of the upload/download functionality twice, let's call the async version from the sync version.  That will provide users with a simple sync API, but with only one implementation to maintain.

The "reader" and "writer" involve callbacks, so that will be a little tricky, but we can defer the sync `read` and `write` methods to threads.

Thanks to @escapewindow for the suggestion!

PR #4721, opened 2021-04-06
Fix function names in client-py's readme

alopezz:
I found this, which looks like a mistake, while having a quick look at the python client docs (as I was having a look at maybe working on #4698).


djmitche (approved):
Ah, yes, copy-pasta error.  Thanks!

PR #4724, opened 2021-04-07
Support tagging S3 objects

### LINKED ISSUE #3948: [aws backend] Support tagging S3 objects with projectId and other tags in deployment config
Let's tag objects for later analysis.  At least with the projectId, and let's let deployers put other tags in the deployment config that the backend will also add.

djmitche:
This automatically tags objects with the ProjectId, and allows deployers to add other custom tags on a per-backend basis.

Github Bug/Issue: Fixes #3948

### DIFF
--- a/services/object/src/backends/aws.js
+++ b/services/object/src/backends/aws.js
@@ -3,6 +3,7 @@ const assert = require('assert');
 const aws = require('aws-sdk');
 const { reportError } = require('taskcluster-lib-api');
 const taskcluster = require('taskcluster-client');
+const qs = require('qs');
 
 const PUT_URL_EXPIRES_SECONDS = 45 * 60;
 
@@ -31,6 +32,9 @@ class AwsBackend extends Backend {
       options.region = this.region;
     }
     this.s3 = new aws.S3(options);
+
+    // determine whether we are talking to a genuine AWS S3, or an emulation of it
+    this.isAws = !this.config.endpoint;
   }
 
   async createUpload(object, proposedUploadMethods) {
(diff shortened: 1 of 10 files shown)


services/object/src/backends/aws.js
```diff
@@ -31,6 +32,18 @@ class AwsBackend extends Backend {
       options.region = this.region;
     }
     this.s3 = new aws.S3(options);
+
+    // determine whether we are talking to a genuine AWS S3, or an emulation of it
+    this.isAws = !this.config.endpoint;
```
imbstack, line 37: The hit NBC drama "This is Aws"

services/object/src/backends/aws.js
```diff
@@ -136,6 +173,34 @@ class AwsBackend extends Backend {
...
+      ProjectId: object.project_id,
+    };
+  }
+
+  /**
+   * Construct the "Tagging" argument containing the tags for this object, as included
+   * in the `x-amx-tagging` header.
+   */
+  objectTaggingHeader(object) {
+    return qs.stringify(this._objectTags(object));
```
imbstack, line 192: I'm not deeply bothered by adding a new dependency to our pile of dependencies but superagent or got can't do something like this?

imbstack (approved):
I think this makes sense

djmitche:
I'll look at why the tests fail :(

djmitche:
just bad luck:
```
error An unexpected error occurred: "https://registry.yarnpkg.com/@azure/core-auth/-/core-auth-1.3.0.tgz: Request failed \"500 Internal Server Error\"".
```

PR #4729, opened 2021-04-07
remove special-case in go for retrying 400s on upload

### LINKED ISSUE #4623: All clients should retry HTTP 400's when uploading via object service convenience methods
Originating from: https://github.com/taskcluster/taskcluster/pull/4493#discussion_r599007535

djmitche:
I can't reproduce this error, so for the moment we'll make all upload operations behave the same way and not retry on 400's.

Github Bug/Issue: Fixes #4623

### DIFF
--- a/changelog/issue-4623.md
+++ b/changelog/issue-4623.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4623
+---
--- a/clients/client-go/tcobject/upload.go
+++ b/clients/client-go/tcobject/upload.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/base64"
 	"errors"
-	"fmt"
 	"io"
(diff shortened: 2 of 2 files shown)


petemoore (approved):
Since this only affects the object service uploads, I'm ok with this, although not convinced yet it is an improvement. It could be, for example, that this issue is only present on specific platforms like Windows 10, or macOS etc - in which case some manual testing might not have uncovered it. But since this is a new service, we can try without the fix, and see if it gets reported. Note, I wouldn't want to remove the fix from generic-worker queue uploads without more thorough testing, since this issue was a real pain for the firefox CI pipeline in the past.

djmitche:
That definitely makes sense.  If we can reproduce this, I'd like to recognize it more specifically than just "400" -- maybe there's a regex on the response body that we can use, or the HTTP status, or something like that.

PR #4730, opened 2021-04-07
Call the async versions of upload and download from their sync versions in python client

### LINKED ISSUE #4698: [client-py] call async upload/download from sync upload/download

alopezz:
Github Bug/Issue: Fixes #4698 

I've started by doing the upload part; I'm hoping to get feedback of whether this looks right, as doing the download part is almost the same.

This would probably benefit from more thorough tests to be confident that it does what it has to.

### DIFF
--- a/clients/client-py/taskcluster/aio/upload.py
+++ b/clients/client-py/taskcluster/aio/upload.py
@@ -71,7 +71,7 @@ async def upload(*, projectId, name, contentType, contentLength, expires,
 
     async def hashingReaderFactory():
         nonlocal hashingReader
-        hashingReader = HashingReader(await ensureCoro(readerFactory)())
+        hashingReader = HashingReader(await readerFactory())
         return hashingReader
 
     async with aiohttp.ClientSession() as session:
@@ -157,15 +157,8 @@ class HashingReader:
         self.bytes = 0
 
     async def read(self, max_size):
-        chunk = await ensureCoro(self.inner.read)(max_size)
-        try:
-            self.update(chunk)
-        except TypeError:
-            # Best effort: since ensureCoro fails to identify certain cases as
-            # coroutines (e.g. with aiofiles), when we fail to get a bytes-like object
-            # on the first try, we retry without the ensureCoro wrapping.
-            chunk = await self.inner.read(max_size)
-            self.update(chunk)
+        chunk = await self.inner.read(max_size)
+        self.update(chunk)
         return chunk
 
     def update(self, chunk):
--- a/clients/client-py/taskcluster/aio/upload.py
+++ b/clients/client-py/taskcluster/aio/upload.py
@@ -21,11 +21,12 @@ if six.PY2:
     raise ImportError("upload is only supported in Python 3")
 
 import base64
+import hashlib
 
 import aiohttp
 
 import taskcluster
-from taskcluster.upload import HashingReader as SyncHashingReader
+from .asyncutils import ensureCoro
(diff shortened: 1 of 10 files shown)


clients/client-py/taskcluster/upload.py
```diff
@@ -63,105 +51,13 @@ def upload(*, projectId, name, contentType, contentLength, expires,
...
-    def hashes(self, contentLength):
-        """Return the hsahes in a format suitable for finishUpload, first checking that all the bytes
-        in the content were hashed."""
-        if contentLength != self.bytes:
-            raise RuntimeError(f"hashed {self.bytes} bytes but content length is {contentLength}")
-        return {
-            "sha256": self.sha256.hexdigest(),
-            "sha512": self.sha512.hexdigest(),
-        }
+    return asyncio.run(aio_upload.upload(
```
djmitche, line 54: this could probably use `**kwargs` too

djmitche (commented):
This is fantastic!  Let's see if the tests confirm that before I click "approve" :)

What additional tests did you have in mind?  This doesn't remove the existing sync tests, so it should be getting a pretty thorough workout.

alopezz:
> What additional tests did you have in mind? 

I may have to take that back, as I hadn't looked closely at the `test_upload_download_real` tests, so I thought that only `upload_from_buf` was being tested.

And there *is* a failing test among those, I'll see what I can do about it.

alopezz:
So it turned out that `aiofiles` [methods](https://github.com/Tinche/aiofiles/blob/26decd33a86f1fa93599a1fad117cd8dead28c5a/src/aiofiles/threadpool/utils.py#L40-L46) are not really *coroutine functions* in a way that `asyncio.iscoroutinefunction()` recognizes, so I had to work around that.

Also, maintaining compatibility with py36 means no [`asyncio.run`](https://docs.python.org/3/library/asyncio-task.html#asyncio.run), so that had to be implemented differently.

With that out of the way, I'll do the corresponding work for downloads.

djmitche:
Sometimes the ducks are just too ducky for duck typing!

clients/client-py/taskcluster/aio/upload.py
```diff
@@ -146,10 +147,38 @@ async def reader_gen():
...
+        chunk = await ensureCoro(self.inner.read)(max_size)
+        try:
+            self.update(chunk)
+        except TypeError:
+            # Best effort: since ensureCoro fails to identify certain cases as
+            # coroutines (e.g. with aiofiles), when we fail to get a bytes-like object
+            # on the first try, we retry without the ensureCoro wrapping.
+            chunk = await self.inner.read(max_size)
+            self.update(chunk)
         return chunk
```
djmitche, line 169: This concerns me, since it's calling `read(..)` twice, and that function has side-effects (incrementing the current position).

We actually know the cases where passed functions are sync, so maybe we could do this wrapping explicitly at that point, in the sync versions -- at least for the readerFactory and writerFactories.

alopezz:
The failing test looks like an unrelated intermittent failure, I think.

djmitche:
Yeah, that's #4744 :(

djmitche (approved):
Awesome!  That saves a _bunch_ of duplicated code!

PR #4733, opened 2021-04-08
Support amqp (without TLS) for communicating with pulse

### LINKED ISSUE #4648: Make it so that you can specify rabbitmq to use TLS or not
Some people, especially in development deployments, may want to do a simple RabbitMQ installation with no SSL, this issue is there so we can make this possible.

djmitche:
Github Bug/Issue: Fixes #4648

I'll test this out in my dev deployment.


djmitche:
This worked for me locally.  Note (as it confused me for a bit!) that amqp / port 5672 works on the hip-macaw cluster but not on the orange-antelope cluster!

petemoore (approved):
Looks about right to me. Thanks!

Does it also look OK to you @MasterWayZ?

djmitche:
I'll merge, but I will make another PR if the answer is "no" :)

MasterWayZ:
It does look good, though the TLS/non-TLS port can be confusing. (It did for me).

PR #4734, opened 2021-04-08
Node.js upgrade 14.16.0 -> 14.16.1

petemoore:
I've published `docker.io/taskcluster/ci-image:node14.16.1-pg11-go1.15.6`.

Do I also need to run `build-browser-test.sh` and `build-rabbit-test.sh`?

They are based on debian stretch, but ci-image is based on the more recent debian buster. Should I update browser-test and rabbit-test to debian buster too?


djmitche:
Yeah, you'll need to build all of the images

djmitche (approved):
Woo!

PR #4740, opened 2021-04-08
Update reverse proxy example to avoid weird authentication errors caused by default apache config

### LINKED ISSUE #4739: Current apache reverse proxy example will cause weird authentication errors
Caused by how apache handles encoded slashes and other characters by default.

MasterWayZ:
Github Bug/Issue: Fixes #4739


djmitche (approved):
Nice, thanks!

PR #4743, opened 2021-04-09
Go upgrade 1.15.6 -> 1.16.3

petemoore:
The go upgrade is a prerequisite for providing darwin/arm64 builds of `generic-worker` and `start-worker` (worker runner).


petemoore:
Note, [`infrastructure/tooling/src/generate/generators/go-version.js`](https://github.com/taskcluster/taskcluster/blob/go1.16.3/infrastructure/tooling/src/generate/generators/go-version.js) and [`infrastructure/docker-images/push-all.sh`](https://github.com/taskcluster/taskcluster/blob/go1.16.3/infrastructure/docker-images/push-all.sh) did all the magic.

petemoore:
The error I see is very strange, and suggests that some go installations have got mangled on the worker.

In [a task failure above](https://community-tc.services.mozilla.com/tasks/EdwnfPGER8W9OksoJeA1IQ/runs/0/logs/public/logs/live.log#L11229) we see the line:

```
level=warning msg="[runner] Can't run linter unused: buildssa: analysis skipped: errors in package: [/task_161797554348158/taskcluster/workers/generic-worker/fileutil/file.go:4:2: could not import archive/zip (/task_161797554348158/go1.16.3/go/src/archive/zip/reader.go:639:2: IsDir redeclared)]"
```

If we look at the go source tree, we find [line 639 of `reader.go`](https://github.com/golang/go/blob/go1.16.3/src/archive/zip/reader.go#L639) does not contain a definition for `IsDir` but instead other content. There is a definition of `IsDir` later on in that file, on [line 655](https://github.com/golang/go/blob/go1.16.3/src/archive/zip/reader.go#L655) so I suspect there is an older version of that file on the worker, presumably from an earlier go release.

I'm not quite sure if this is just a dirty cache, and even if it is, I'm not sure how it happened, since the cache names should include the go version, which has changed.

Even stranger, the go1.15.6 version of this file was only 609 lines long, so if it is a cached version of that file from a previous build, it isn't from go version 1.15.6, which is what we've been using for a very long time.

Still investigating. 



https://github.com/golang/go/blob/go1.16.3/src/archive/zip/reader.go#L639

petemoore:
The service-worker-manager failure is unrelated and due to #4744.

It looks like upgrading golangci-lint fixed the previous problem. I suspect it was a mismatch between the standard library calls compiled into the golangci-lint binary and the standard library definition from go 1.16.3. In any case it is nice to be using a more recent `golangci-lint`.

changelog/K1TJedwpRByrvgTHS-9MXg.md
```diff
@@ -0,0 +1,4 @@
+audience: worker-deployers
+level: minor
+---
+Go major version upgrade for generic-worker and worker-runner (go 1.15.6 -> go 1.16.3). Prerequisite step for providing native darwin/arm64 binaries for both (native Apple Silicon builds).
```
djmitche, line 4: Why ["minor"](https://github.com/taskcluster/taskcluster/blob/master/dev-docs/best-practices/changelog.md)?

djmitche (approved):

PR #4749, opened 2021-04-09
Serve HTML documents as attachments

djmitche:
Bugzilla Bug: [1442024](https://bugzilla.mozilla.org/show_bug.cgi?id=1442024)

### DIFF
--- a/changelog/bug-1442024.md
+++ b/changelog/bug-1442024.md
@@ -1,0 +1,5 @@
+audience: deployers
+level: patch
+reference: bug 1442024
+---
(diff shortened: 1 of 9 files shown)


djmitche:
@ajvb the change to ui/docs/reference/platform/object/simple-downloads.mdx is probably all you need to look at here, as the rest just implements it.

I think the big question I have is: is just `text/html` enough?  Do we need to consider image types? stylesheets? other things?

petemoore (approved):
This looks like a welcome security improvement, but I'll let @ajvb give a more thorough interpretation. Nice work!

ajvb (approved):
From a security perspective, this lgtm.

djmitche:
Thanks all!  I'll fix the lint and merge.

PR #4759, opened 2021-04-12
Add `taskcluster download` commands

### LINKED ISSUE #4576: Add a `taskcluster download` command to client-shell
Once the go client supports uploads and downloads (#4425), let's use that functionality to implement a `taskcluster download` command in the shell client.  For the moment at least, uploads need not be supported -- downloading will be by far the more common operation.

djmitche:
Github Bug/Issue: Fixes #4576

I chose to extract the response as JSON because (a) it was pretty easy and (b) the [autogenerated types](https://pkg.go.dev/github.com/taskcluster/taskcluster/v43/clients/client-go/tcqueue#GetArtifactContentResponse) are numbered (GetArtifactContentResponse1, GetArtifactContentResponse2, etc.) and that seems like something we should address -- otherwise adding a storageType could renumber those values (maybe it already has?) causing a major version bump.

If you (@petemoore) think we can do something about that, let's open a new issue.


clients/client-shell/cmds/download/summary.go
```diff
@@ -0,0 +1,17 @@
+package download
+
+import (
+	"fmt"
+	"io"
+	"time"
+)
+
+func summarize(out io.Writer, contentLength int64, contentType string, duration time.Duration) {
+	// only calculate the rate if it's well-defined and a useful value
```
petemoore, line 10: ++

clients/client-shell/cmds/download/artifact.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	switch storageType {
+	case "reference", "s3":
+		// `reference` and `s3` both have a URL from which we should download, so handle them the same
+		start := time.Now()
+
+		var urlContent struct {
+			URL string `json:"url"`
+		}
```
petemoore, line 72: nice dynamic typing

clients/client-shell/cmds/download/artifact.go
```diff
@@ -0,0 +1,151 @@
...
+		Use:   "artifact <taskId> [<runId>] <name> <filename>",
+		Short: "Download an artifact attached to the given task; if runId is omitted, the latest run is used",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			taskID := args[0]
+			var (
+				runID    int64
+				name     string
+				filename string
+				err      error
+			)
```
petemoore, line 127: Nice syntax

internal/geturl/geturl.go
```diff
@@ -0,0 +1,67 @@
...
+	"net/http"
+
+	"github.com/cenkalti/backoff/v3"
+	"github.com/taskcluster/httpbackoff/v3"
+)
+
+type HTTPRetryError struct {
+	Attempts int
+	Err      error
+}
```
petemoore, line 14: User code that imports the taskcluster client will often benefit from being able to inspect a non-nil error that is returned from a client method. If calling code receives an error type that comes from an internal package, will it still be able to access the internal fields of the error? Perhaps HTTPRetryError needs to be a public type for calling code to be able to access the underlying error?

petemoore (approved):

petemoore:
> Github Bug/Issue: Fixes #4576
> 
> I chose to extract the response as JSON because (a) it was pretty easy and (b) the [autogenerated types](https://pkg.go.dev/github.com/taskcluster/taskcluster/v43/clients/client-go/tcqueue#GetArtifactContentResponse) are numbered (GetArtifactContentResponse1, GetArtifactContentResponse2, etc.) and that seems like something we should address -- otherwise adding a storageType could renumber those values (maybe it already has?) causing a major version bump.
> 
> If you (@petemoore) think we can do something about that, let's open a new issue.

I've created #4761 for this - thanks!

internal/geturl/geturl.go
djmitche: It's re-exported in the download package.

djmitche:
I didn't actually mean to mark this for review yet, but thank you!  I need to add tests, and I want to factor out the entire artifact-download system (queue.getContent and the switch on storageType) so that it can be shared with generic-worker.

internal/geturl/geturl.go
djmitche: *in the tcobject package

internal/mocktc/object.go
```diff
@@ -84,6 +94,19 @@ func (object *Object) StartDownload(name string, payload *tcobject.DownloadObjec
...
+func (object *Object) FakeObject(name string) {
+	object.objects[name] = Obj{
+		uploadRequest:  &tcobject.CreateUploadRequest{},
+		uploadFinished: true,
+		onMockS3:       true,
+	}
+}
+
+/////////////////////////////////////////////////
+
```
petemoore, line 109: Although not heavily tested and not yet documented, the `GW_TESTS_USE_EXTERNAL_TASKCLUSTER` env var was intended to be used as a means to be able to select whether tests should run against mocks (for speed) or against a real deployment, for integration testing purposes. If tests create objects outside of the API interface, they would not work if run against a real deployment. Since an object can be created via the API, would it be possible for the tests to use that instead, so that the tests could still be run against a real deployment? My thinking is that we could run tests against a real deployment when making a release, to give us further certainty that things function end-to-end (e.g. this may have helped protect against the taskQueueId failures that we introduced recently, that didn't get picked up by the mocks when the task definition structure changed).

If you like, you could keep this convenience method, but in the body wrap the object creation API call rather than updating the underlying map directly, to preserve the capability.

petemoore (changes requested):

internal/mocktc/object.go
djmitche: The problem with testing against a "real thing" is that it depends on the configuration of that real thing, and requires quite a bit of complexity to establish the conditions for the test -- in this case, creating and claiming a task, creating an artifact on that task, and uploading that artifact.  The result is that the condition-establishment becomes more brittle than the test itself.  In this case, it would depend on configuration of the object service's backends, for example, as well as taskQueues.  It also eliminates the possibility of lots of conditions.  In the example of upload or download support, for example, there's no way to test for upload failures.  And finally, it ties development into the deployment loop: a test won't succeed until the change being tested is running on the TC deployment where CI runs.  That ends up adding weeks to development cycles, as we saw with the client upload/download functionality.  And, it's worth noting, that's not even completely tested, as writing those tests is hard.

In short, I don't think that tests which run against both an upstream Taskcluster deployment and a mock are practical.  Given more engineering bandwidth, maybe we could do some interesting things (perhaps spin up a deployment with minikube for every PR?) but those ideas are immensely far outside our capacity right now.

I appreciate the introduction of a mock, and the use of interfaces to ensure at least some amount of fidelity to the real thing.  But even that is difficult -- I spent a few hours getting the mocks/tests in this PR to work.

petemoore (approved):

PR #4806, opened 2021-04-21
[azure-provider] Generate names for dataDisks

djmitche:
..to allow their deletion during deprovisioning

I'm going to test this out in my dev environment.

Bugzilla Bug: [1631824](https://bugzilla.mozilla.org/show_bug.cgi?id=1631824)

### DIFF
--- a/changelog/bug-1631824.md
+++ b/changelog/bug-1631824.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: bug 1631824
+---
+The Azure provider of the worker-manager service now assigns unique names to all data disks attached to a VM, allowing those disks to be removed when the worker is removed.
--- a/generated/references.json
+++ b/generated/references.json
@@ -1119,6 +1119,21 @@
               "storageProfile": {
                 "additionalProperties": true,
                 "properties": {
+                  "dataDisks": {
+                    "items": {
+                      "additionalProperties": true,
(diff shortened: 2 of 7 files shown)


djmitche:
The failures above are unrelated.. but I can't figure out how to set up the Azure provider for my dev env, so I've not finished that yet.

djmitche:
Welp, I finally got the provider working, and it didn't delete any of the disks (including the OS disk)

djmitche:
The `disks` array appears to be empty in providerData.

djmitche:
Ah, I see why: once the instance exists, `checkWorker` assumes everything's OK and doesn't call `provisionResources` again, so the `modifyFn` is never called, which is what fills in the disk information.

There's a pretty bad race condition here, anyway, between finishing the provisioning process (gathering the information about the worker) and the worker registering and going to the RUNNING state.

djmitche:
This has been awful, which is a bit better than I expected.  I'll work on addressing that race condition tomorrow.  Ideally:
 * can fix the race
 * can just name all disks in advance so that we don't need to capture them back after the VM is created

djmitche:
Well, I tried both of those "ideally" items.  [narrator voice]Let's see what happens[/narrator voice]

djmitche:
OMG it's green

djmitche:
So, I think this is another bug, but if you delete an instance when it's still in the REQUESTED state, then `provisionResources` just keeps patiently waiting for it to show up.  I'll slap another coat of paint over that crack and hope it doesn't get any bigger.

djmitche:
OK, this version works in dev, both for when a worker is removed with `removeWorker` and for when the vm is deleted via the portal.

imbstack (approved):
I think this makes sense. Thanks for dealing with this

PR #4825, opened 2021-04-27
Add github mocks using nock

### LINKED ISSUE #3985: Mock github in tc-github via nock
Mocking the Octokit interface is hard to keep in sync with Octokit, and is getting harder as they move more functionality to plugins.  It also means we're not actually testing our use of those plugins.

Let's switch to mocking GitHub at the HTTP level.  That API is better (?) defined and documented, and of necessity more stable.

In particular, that will let us use the Octokit pagination support, which tbh we should be using for all `*.list*` calls.

sladyn98:
Github Bug/Issue: Fixes #3985

### DIFF
--- a/services/github/test/github-auth.js
+++ b/services/github/test/github-auth.js
@@ -2,6 +2,7 @@ const Debug = require('debug');
 const sinon = require('sinon');
 const _ = require('lodash');
 const assert = require('assert');
+const nock = require('nock');
 
 class FakeGithub {
   constructor(installation_id) {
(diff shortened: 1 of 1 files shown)


services/github/test/github-auth.js
```diff
@@ -238,16 +239,22 @@ class FakeGithub {
...
 
   hasNextPage() {
     return false;
   }
 }
 
+nock('https://api.github.com')
+  .get('${owner}/${repo}@${number}')
+  .reply(200, {
+    body: 'Task failed here'
```
sladyn98, line 255: @djmitche I am not too sure if this is the API URL is to be intercepted :> The tests pass anyway

sladyn98: yarn workspace taskcluster-secrets test

djmitche:
It looks like the decision task is failing because this is based on an old version of `main`.  A rebase should fix that up, and I don't think it will generate any conflicts.

services/github/test/github-auth.js
djmitche: If it's meant to replace the `getComments` API method, that's [`/repos/{owner}/{repo}/issues/{issue_number}/comments`](https://docs.github.com/en/rest/reference/issues#list-issue-comments).  That link also shows the format of the response body.  I suspect that a good habit will be to include a link to the GitHub docs for each nock'd endpoint.  And, the response should have the right structure, but can omit fields that we do not use.

I suspect you'll want to use [regular expressions with nock](https://classic.yarnpkg.com/en/package/nock#specifying-path) for the path components, and then parse the values [out of `this.req.path`](https://classic.yarnpkg.com/en/package/nock#access-original-request-and-headers).  That seems like something that could be made easier with a utility function, but if it were me I'd write out the first one fully, and then figure out how to build a utility function.

jwhitlock:
@sladyn98 do you want to continue working on this PR?

jwhitlock:
Closing due to inactivity, feel free to re-open when ready for the next round.

PR #4838, opened 2021-04-30
Automerge in all package.json's

djmitche:
This change causes renovate to match any path containing package.json
and enable automerge.  We had previously wanted to avoid this, since
`ui/` and `workers/docker-worker` are poorly tested, but in practice
we're merging those anyway.  And in fact, the build process for `ui/`
ends up catching most issues with things like Babel, so tihs has done
quite well.

This should further reduce the team's required toill.

### DIFF
--- a/package.json
+++ b/package.json
@@ -261,7 +261,7 @@
         ],
         "matchCurrentVersion": "!/^0/",
         "paths": [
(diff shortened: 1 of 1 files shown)


petemoore (approved):
Love it!

PR #4863, opened 2021-05-07
Serialize errors for logging more succinctly

### LINKED ISSUE #4807: Errors from Azure API include the whole request, don't parse as JSON
I spent a bunch of time looking for errors from the Azure API, and eventually found them by removing all filters in StackDriver.  They were huge blobs of JSON text that StackDriver wouldn't parse as JSON (because they were too long).  This appears to be mostly because they include the entire request object in the error body.

It'd be great to fix this in some systematic way -- I can imagine lots of other errors are also too huge to get registered.

djmitche:
This avoids issues with errors that have helpful details like an HTTP
request attached.  In practice, most error logging contains only
top-level strings, so this should not be a problem.

Github Bug/Issue: Fixes #4807

### DIFF
--- a/changelog/issue-4807.md
+++ b/changelog/issue-4807.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4807
+---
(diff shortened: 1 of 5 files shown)


imbstack (approved):
makes sense to me!

PR #4877, opened 2021-05-11
[client-rust] support downloading artifacts

### LINKED ISSUE #4764: Support downloading artifacts in all clients
All clients now have support for downloading *objects*, but let's add support for downloading artifacts, as well.  Currently, that means handling `s3` / `reference` artifacts (which are pretty much the same) as well as handling `object` artifacts by deferring to the existing object-download functionality.

This will make issues like #4760 moot, as we'll have a one-liner to download a named artifact (to a buffer, to a file, or to a writer)

djmitche:
This layers artifact support on top of the existing object support in this client, using a very similar approach.

Github Bug/Issue: Fixes #4764


clients/client-rust/download/src/test_helpers.rs
```diff
@@ -0,0 +1,139 @@
...
+use anyhow::Error;
+use async_trait::async_trait;
+use httptest::{matchers::*, responders::*, Expectation};
+use serde_json::Value;
+use std::fmt;
+use std::sync::{Arc, Mutex};
+
+const PLAINTEXT_BODY: &[u8] = b"hello, world";
+
+const GZIPPED_BODY: &[u8] = &[
```
imbstack, line 12: I always get disappointed when there isn't some fun string hidden in things like this!

imbstack (approved):
looks good to me

clients/client-rust/download/src/test_helpers.rs
djmitche: Sorry to disappoint!  I tried to compress it at runtime, but httptest's `.body(..)` wants a static string, so I just plunked it in a const.

PR #4883, opened 2021-05-13
Pass servername to amqplib.connect

### LINKED ISSUE #4882: Pass servername as a socket option to amqplib
Due to https://github.com/squaremo/amqp.node/pull/567, amqplib doesn't work with SNI.

djmitche:
Github Bug/Issue: Fixes #4882


imbstack (approved):
I think this makes sense

PR #4893, opened 2021-05-19
Rust worker

djmitche:
I don't expect this to go anywhere, but I'm making a draft PR just to keep a pointer to it.


djmitche:
This has accidentally gone somewhere.  See `workers/rust/README.md` for a TODO list.  Done so far:
 * testing framework for worker implementations
 * support for uploading artifacts
 * support for a task log
 * support for claiming multiple tasks in parallel
 * basic container support (`image` and `command`)

At this point, this is probably enough to write something like `scriptworker` (/cc @escapewindow) or other utility workers.

PR #4911, opened 2021-06-01
[Bug 1712924] Encode postmessage content to avoid issues

imbstack:
Bugzilla Bug: [1712924](https://bugzilla.mozilla.org/show_bug.cgi?id=1712924)

### DIFF
--- a/services/web-server/src/utils/login.js
+++ b/services/web-server/src/utils/login.js
@@ -1,5 +1,15 @@
+const assert = require('assert');
+const _ = require('lodash');
+
 // A function that returns a middleware that renders callback.ejs
 module.exports = (publicUrl) => async (request, response) => {
+  assert(request.user, 'Must have a user');
+  assert(request.user.profile);
+  assert(request.user.identityProviderId);
+  assert(request.user.providerExpires);
+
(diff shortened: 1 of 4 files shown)


services/web-server/src/utils/login.js
```diff
@@ -1,5 +1,17 @@
...
+
 // A function that returns a middleware that renders callback.ejs
 module.exports = (publicUrl) => async (request, response) => {
+  assert(request.user, 'Must have a user');
+  assert(request.user.profile);
+  assert(request.user.identityProviderId);
+  assert(request.user.providerExpires);
+
+  if (!_.isString(request.user.profile)) {
+    request.user.profile = new Buffer.from(JSON.stringify(request.user.profile)).toString('base64');
```
jwhitlock, line 12: I'm not familiar w/ JS. Can this be assigned to ``request.user.encodedProfile`` instead of overwriting ``user.profile``? The same name makes it look like `callback.ejs` has the same bug as before.

jwhitlock (approved):
Thanks @imbstack!

services/web-server/src/utils/login.js
ncdt138: Thank you to all my participating friends who helped me.

Aaryan9898:
Nice patch.

PR #4921, opened 2021-06-14
issue2749 - change migrate to material-ui/treeview

### LINKED ISSUE #2749: Remove material-ui-treeview
We started using https://github.com/helfi92/material-ui-treeview/ in https://community-tc.services.mozilla.com/hooks because material-ui did not have a tree view component available at the time, but now they do. Let's move away from what we were using and use material-ui's [TreeView](https://material-ui.com/api/tree-view/#treeview-api) component. The [demo](https://material-ui.com/components/tree-view/#rich-object) shows that they support rendering the same tree structure that we have so switching to it shouldn't be too demanding from our end.

cerkiewny:
Github Bug/Issue: Fixes #2749

### DIFF
--- a/ui/src/views/Hooks/ListHooks/index.jsx
+++ b/ui/src/views/Hooks/ListHooks/index.jsx
@@ -17,6 +17,36 @@ import ErrorPanel from '../../../components/ErrorPanel';
 import Link from '../../../utils/Link';
 import hooksQuery from './hooks.graphql';
 
+const TreeItems = ({ groups }) => {
+  return groups
+    ? groups.map(group => {
+        const nodes = group.hooks.map(hook => (
+          <TreeItem
+            key={hook.hookId}
+            nodeId={hook.hookId}
+            label={
+              <Link
+                to={`/hooks/${group.hookGroupId}/${encodeURIComponent(
+                  hook.hookId
+                )}`}>
+                {hook.hookId}
+              </Link>
+            }
+          />
+        ));
+
+        return (
+          <TreeItem
+            key={group.hookGroupId}
+            nodeId={group.hookGroupId}
+            label={group.hookGroupId}>
+            {nodes}
+          </TreeItem>
+        );
+      })
+    : [];
+};
+
 @graphql(hooksQuery, {
   options: {
     fetchPolicy: 'network-only',
@@ -94,33 +124,7 @@ export default class ListHooks extends Component {
       data: { loading, error, hookGroups },
     } = this.props;
     const { search } = parse(window.location.search.slice(1));
-    const tree = hookGroups
-      ? hookGroups.map(group => {
-          const nodes = group.hooks.map(hook => (
-            <TreeItem
-              key={hook.hookId}
-              nodeId={hook.hookId}
-              label={
-                <Link
-                  to={`/hooks/${group.hookGroupId}/${encodeURIComponent(
-                    hook.hookId
-                  )}`}>
-                  {hook.hookId}
-                </Link>
-              }
-            />
-          ));
-
-          return (
-            <TreeItem
-              key={group.hookGroupId}
-              nodeId={group.hookGroupId}
-              label={group.hookGroupId}>
-              {nodes}
-            </TreeItem>
-          );
-        })
-      : [];
+    const tree = <TreeItems groups={hookGroups} />;
     const { expanded, selected } = this.state;
 
     return (
--- a/ui/src/views/Hooks/ListHooks/index.jsx
+++ b/ui/src/views/Hooks/ListHooks/index.jsx
@@ -7,6 +7,7 @@ import TreeView from '@material-ui/lab/TreeView';
 import TreeItem from '@material-ui/lab/TreeItem';
 import PlusIcon from 'mdi-react/PlusIcon';
 import { parse, stringify } from 'qs';
+import Typography from '@material-ui/core/Typography';
 import Spinner from '../../../components/Spinner';
 import Dashboard from '../../../components/Dashboard';
 import HelpView from '../../../components/HelpView';
@@ -82,6 +83,10 @@ export default class ListHooks extends Component {
     }
   }
 
+  toggleState(stateName) {
+    return (event, nodeIds) => this.setState({ [stateName]: nodeIds });
+  }
+
   render() {
     const {
       classes,
@@ -116,8 +121,6 @@ export default class ListHooks extends Component {
           );
         })
       : [];
-    const handleState = stateName => (event, nodeIds) =>
-      this.setState({ [stateName]: nodeIds });
     const { expanded, selected } = this.state;
 
     return (
@@ -133,16 +136,19 @@ export default class ListHooks extends Component {
         }>
         {!hookGroups && loading && <Spinner loading />}
         <ErrorPanel fixed error={error} />
-        {hookGroups && (
-          <TreeView
-            {...{ expanded, selected }}
-            onNodeToggle={handleState('expanded')}
-            onNodeSelect={handleState('selected')}
-            defaultCollapseIcon={<ArrowDropDownIcon />}
-            defaultExpandIcon={<ArrowRightIcon />}>
-            {tree}
-          </TreeView>
-        )}
+        {!loading &&
+          (tree.length ? (
+            <TreeView
+              {...{ expanded, selected }}
+              onNodeToggle={this.toggleState('expanded')}
+              onNodeSelect={this.toggleState('selected')}
+              defaultCollapseIcon={<ArrowDropDownIcon />}
+              defaultExpandIcon={<ArrowRightIcon />}>
+              {tree}
+            </TreeView>
+          ) : (
+            <Typography variant="subtitle1">No hooks are defined</Typography>
+          ))}
         <Button
           spanProps={{ className: classes.actionButton }}
           tooltipProps={{ title: 'Create Hook' }}
--- a/ui/src/views/Hooks/ListHooks/index.jsx
+++ b/ui/src/views/Hooks/ListHooks/index.jsx
@@ -1,10 +1,12 @@
 import React, { Component } from 'react';
 import { graphql } from 'react-apollo';
 import { withStyles } from '@material-ui/core/styles';
-import Typography from '@material-ui/core/Typography';
-import MuiTreeView from 'material-ui-treeview';
+import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
+import ArrowRightIcon from '@material-ui/icons/ArrowRight';
+import TreeView from '@material-ui/lab/TreeView';
+import TreeItem from '@material-ui/lab/TreeItem';
 import PlusIcon from 'mdi-react/PlusIcon';
-import qs, { parse, stringify } from 'qs';
+import { parse, stringify } from 'qs';
 import Spinner from '../../../components/Spinner';
 import Dashboard from '../../../components/Dashboard';
 import HelpView from '../../../components/HelpView';
@@ -29,6 +31,11 @@ import hooksQuery from './hooks.graphql';
   },
 }))
 export default class ListHooks extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { expanded: [], selected: [] };
+  }
+
   handleCreateHook = () => {
     this.props.history.push('/hooks/create');
   };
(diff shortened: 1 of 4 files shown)


sarah-clements:
Hi! I'll take a look at this tomorrow. :)

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -46,25 +48,80 @@ export default class ListHooks extends Component {
...
+      : [];
+
+    this.setState({ expanded: expandedHookGroups, selected: [] });
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { expanded: [], selected: [] };
+    this.updateExpanded();
+  }
```
sarah-clements, line 74: The constructor should be moved to the top of the class (function).

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -46,25 +48,80 @@ export default class ListHooks extends Component {
...
+          )
+          .map(group => group.hookGroupId)
+      : [];
+
+    this.setState({ expanded: expandedHookGroups, selected: [] });
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { expanded: [], selected: [] };
```
sarah-clements, line 72: I would write this as `this.state = this.updateExpanded()` but then you'd return an object in `updateExpanded` rather than setting state. The alternative is to calls `this.updateExpanded()` in `componentDidMount`.

ui/src/views/Hooks/ListHooks/index.jsx
cerkiewny: Moved all the way up

ui/src/views/Hooks/ListHooks/index.jsx
cerkiewny: moved to component did mount

ui/src/views/Hooks/ListHooks/index.jsx
sarah-clements: Great :) I'm still reviewing this pr - I'll let you know when I'm done so you can batch the changes together.

cerkiewny: Oj, sorry I thought you made all the comments in batch "review" feature, my bad.

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -46,25 +48,80 @@ export default class ListHooks extends Component {
     });
   };
 
+  updateExpanded() {
+    this.query = qs.parse(this.props.location.search.slice(1));
```
sarah-clements, line 57: I don't think this needs to be a method on the class. You can just extract the query string dynamically like so ` const { somethingMeaningful } = qs.parse(this.props.location.search)` and then you don't need line 56 (just use the variable you destructured in place of hookSearch).

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -80,21 +137,15 @@ export default class ListHooks extends Component {
...
-              <Typography variant="subtitle1">
-                No items for search term {hookSearch}
-              </Typography>
-            }
-            Link={Link}
-          />
+          <TreeView
+            expanded={this.state.expanded}
+            selected={this.state.selected}
+            onNodeToggle={handleToggle}
```
sarah-clements, line 146: You can use destructuring assignment for these two state values in render: `const { expanded, selected } = this.state`

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -46,25 +48,80 @@ export default class ListHooks extends Component {
...
+          );
+        })
       : [];
+    const handleToggle = (event, nodeIds) => {
+      this.setState({ expanded: nodeIds });
+    };
+
+    const handleSelect = (event, nodeIds) => {
+      this.setState({ selected: nodeIds });
+    };
```
sarah-clements, line 127: These two functions (handleToggle and handleSelect) should be with the other class methods rather than defined in render. Alternatively, you can define them inline on lines 141 and 142 if they're only used once: `expanded={(event, nodeIds) => this.setState({ selected: nodeIds })`.

sarah-clements (commented):
This looks good :) I haven't tested it manually though. @cerkiewny have you manually tested this locally?

ui/src/views/Hooks/ListHooks/index.jsx
sarah-clements: No worries :)

cerkiewny:
> This looks good :) I haven't tested it manually though. @cerkiewny have you manually tested this locally?

Yes I did, but could use a 2nd eye on the angle of the design, the functionality itself was working.

ui/src/views/Hooks/ListHooks/index.jsx
sarah-clements: Another option, reduce the duplicate code by combining these two functions into one: `handleUpdates = (stateValue, nodeIds) => this.setState({ [stateValue]: nodeIds })` and then `onNodeSelect={(nodeIds) => this.handleUpdates('selected', nodeIds)}`. You basically use a string to point to the correct key on the state object.

sarah-clements:
> > This looks good :) I haven't tested it manually though. @cerkiewny have you manually tested this locally?
> 
> Yes I did, but could use a 2nd pair of eyes to look at the design aspect, the functionality itself was working.

Ok, I have this up and running locally. One problem I've hit is that if I just go to http://localhost:5080 and then go to the menu icon on the left, select 'hook', I see an error: this.query is undefined. You can look in the console for some hints as to the location of the error.

There's also some tests that have failed. Any with a red X you should click into the details link -> view in taskcluster -> click on the live log link on the right side of the taskcluster UI, and that will give you more details. If the messages are a bit vague then you can ask about them on here and me or petemoore can help you figure them out.

sarah-clements:
@cerkiewny Please @ mention me when you've fixed the bug and made all changes and I'll take another look.

ui/src/views/Hooks/ListHooks/index.jsx
cerkiewny: This change introduced the query problem there is time between constructor and component did mount in which it is undefined, I can either move updateExpanded back to constructor or add this.query="" in constructor to give it initial value, what do you think?

ui/src/views/Hooks/ListHooks/index.jsx
cerkiewny: did something similar, let me know if it is good enough

ui/src/views/Hooks/ListHooks/index.jsx
cerkiewny: done

ui/src/views/Hooks/ListHooks/index.jsx
cerkiewny: actually after removing the query from class it is no longer relevant

ui/src/views/Hooks/ListHooks/index.jsx
cerkiewny: done

cerkiewny:
> @cerkiewny Please @ mention me when you've fixed the bug and made all changes and I'll take another look.

@sarah-clements I think it's ready for next round

ncdt138:
Thanks you guys gonna particular the future of humanity

ncdt138:
Thanks ,you guys gonna particular the future of humanity

ncdt138:
Thanks ,you guys gonna particular the future of humanity

cerkiewny:
Just realized I accidental removed the empty hook text, so adding it back and changing it to resolve #3044

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -118,13 +118,9 @@ export default class ListHooks extends Component {
...
-    const handleToggle = (event, nodeIds) => {
-      this.setState({ expanded: nodeIds });
-    };
-
-    const handleSelect = (event, nodeIds) => {
-      this.setState({ selected: nodeIds });
-    };
+    const handleState = stateName => (event, nodeIds) =>
+      this.setState({ [stateName]: nodeIds });
+    const { expanded, selected } = this.state;
```
sarah-clements, line 124: Its not convention to have a function like this in render unless there's some reason it needs to be here. If should be moved into the body of the class.

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -46,25 +54,74 @@ export default class ListHooks extends Component {
...
-          value: group.hookGroupId,
-          nodes: group.hooks.map(hook => ({
-            value: hook.hookId,
-            href: `/hooks/${group.hookGroupId}/${encodeURIComponent(
-              hook.hookId
-            )}`,
-          })),
-        }))
+      ? hookGroups.map(group => {
+          const nodes = group.hooks.map(hook => (
```
sarah-clements, line 97: I would move this whole block into a separate component and file so when you pass the hook into this new component as a prop, its only going to re-render everything in that component if the prop has changed. Right now, if a user selects or expands (state change), this is going to be rendered even if the hookGroups hasn't changed.

sarah-clements:
These changes look good, just a few things I missed from last review :) I'll be off next week, so I won't be able to review changes until the following week.

ui/src/views/Hooks/ListHooks/index.jsx
cerkiewny: can we have that as a follow up ticket? I would be adding a generic treeview component instead, this one is limited to depth 2 so it might be confusing, alternatively I could just move it as a functional component on top of the same file so we achieve short term rendering performance, what do you think?

ui/src/views/Hooks/ListHooks/index.jsx
cerkiewny: can we have that as a follow up ticket? I would be adding a generic treeview component instead, this one is limited to depth 2 so it might be confusing, alternatively I could just move it as a functional component on top of the same file so we achieve short term rendering performance, what do you think?

petemoore:
Weird, the failure was:

```
[3/5] Fetching packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/@octokit/plugin-retry/-/plugin-retry-3.0.7.tgz: Request failed \"502 Bad Gateway\"".
```

I'm guessing this was just an intermittent?

petemoore:
@cerkiewny are you still working on this, or is it ready for review again? Thanks!

cerkiewny:
It's ready for another round, I was just waiting for Sarah to get back from her days off.

ui/src/views/Hooks/ListHooks/index.jsx
sarah-clements: Moving it to the top of the file is probably fine.

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -135,16 +136,19 @@ export default class ListHooks extends Component {
...
+            <TreeView
+              {...{ expanded, selected }}
+              onNodeToggle={handleState('expanded')}
+              onNodeSelect={handleState('selected')}
+              defaultCollapseIcon={<ArrowDropDownIcon />}
+              defaultExpandIcon={<ArrowRightIcon />}>
+              {tree}
+            </TreeView>
+          ) : (
+            <Typography variant="subtitle1">No hooks are defined</Typography>
```
sarah-clements, line 152: I think this situation, there's nothing the user can do to fix it so this may not be that helpful of an error message. :) I'd show a generic error message (you can see if there are already some in helper files) like 'Something went wrong - please file a bug if the problem persists'.

cerkiewny: Hmm, It is something that existed as behavior a prior my changes, I just accidentally ate it in one of commits, so this change is more restoration to status quo... also there is a bug to change the text to "No hooks are defined" if hooks view is empty.

sarah-clements: Ok, I'd definitely switch to fetching directly from upstream (taskcluster repo) and rebasing so you keep branch clean, rather than merging. We can leave this for now, but fwiw, its definitely ok to change existing code if its not clear or not useful anymore :)

sarah-clements: I think you're going to need to rebase and drop any of the merge commits that don't belong here so you can get all tests to pass.

sarah-clements:
> > This looks good :) I haven't tested it manually though. @cerkiewny have you manually tested this locally?
> 
> Yes I did, but could use a 2nd pair of eyes to look at the design aspect, the functionality itself was working.

Btw, I did test this out and it looks good to me.

ui/src/views/Hooks/ListHooks/index.jsx
sarah-clements: @cerkiewny Hi, just following up with this in case you missed my reply :)

cerkiewny: I haven't missed it just had little time for TC last week and spent it trying to set up my k8s for it, I will probably follow it up today/tomorrow.

cerkiewny: so, should I remove this? (It is btw in case of empty task cluster hooks not in case of errors) I fixed all the other things I think.

sarah-clements: Well, wouldn't an empty taskcluster hook be an application error? It's not based on something the user did "wrong" so that message that is shown to the user doesn't seem very helpful in my opinion. If I saw that when trying to access menu items, I'd think that part of the app is broken. 

But you can just leave it for now so we can get this landed. It seems the service-github task is failing. @petemoore do you know if this is an intermittent failure?

sarah-clements (approved):
Good work :)

ui/src/views/Hooks/ListHooks/index.jsx
cerkiewny: I thought it is meant for clean installation of taskcluster, if you dev init/ and start the backend, if there are no hooks defined yet, basically the message would show.

jwhitlock (approved):
Thanks for the changes @cerkiewny, and thanks for the review @sarah-clements!

The `services-github` test is failing due to a network issue, and I'd expect it to pass on re-run. I ran the test locally (`yarn workspace taskcluster-github test`) and it passed.

PR #4922, opened 2021-06-15
remove react-hot-loader and replace it with neutrino preset

cerkiewny:


petemoore:
Another one for you @sarah-clements if you have time. :-)

sarah-clements:
hrm, I'm not that familiar with neutrino but I'm not seeing where we're actually adding the dev-server to some sort of config file (besides package.json). @cerkiewny Can you please add a description to this pr that explains in detail the change (if it just works magically under the code, that'd be useful for the reviewer to know) and confirms that you have tested this and the hot loading "just works".

cerkiewny:
@sarah-clements as far as I understand it neutrino is something that is able to store the configs as a module (I don't know it too well either), adding dev-server in package.json installs a config that allows dev server to use hot reloading, hence rendering the decorator + hot reload module not needed. I am not 100% sure why exactly the hot reloading works as from docs I should create devServer with hot: true option set, but I am expecting that the server itself is also created as portion of neutrino config hence it works out of the box.

ui/yarn.lock
```diff
@@ -1650,7 +1650,7 @@
   dependencies:
     copy-webpack-plugin "^5.1.2"
 
-"@neutrinojs/dev-server@9.5.0":
+"@neutrinojs/dev-server@9.5.0", "@neutrinojs/dev-server@^9.5.0":
```
sarah-clements, line 1653: The lock file isn't capturing all of the package updates. If I run `yarn` locally after checking out this branch, I see a bunch of changes that need to be captured in this pr.

sarah-clements:
> @sarah-clements as far as I understand it neutrino is something that is able to store the configs as a module (I don't know it too well either), adding dev-server in package.json installs a config that allows dev server to use hot reloading, hence rendering the decorator + hot reload module not needed. I am not 100% sure why exactly the hot reloading works as from docs I should create devServer with hot: true option set, but I am expecting that the server itself is also created as portion of neutrino config hence it works out of the box.

Confirmed it works as expected!

ui/yarn.lock
cerkiewny: Updated yarn lock.

jwhitlock (approved):
Thank you for the PR @cerkiewny, and thanks @sarah-clements for the expert review.

I've run this locally as well, ready to merge.

PR #4933, opened 2021-07-19
[Bug 1714200] add generic-worker-simple darwin arm64 build

### LINKED ISSUE #4926: generic-worker-simple-darwin-arm64 is missing
generic-worker-simple-darwin-arm64 is missing. Can this be added for future releases?
image

_Originally posted by @davehouse in https://github.com/taskcluster/taskcluster/issues/4791#issuecomment-875907818_

davehouse:
Bugzilla Bug: [1714200](https://bugzilla.mozilla.org/show_bug.cgi?id=1714200)

Github Bug/Issue: Fixes #4926

### DIFF
--- a/changelog/issue4926.md
+++ b/changelog/issue4926.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 4926
+---
+Adds release binary for generic-worker-simple on MacOS arm64.
--- a/ui/docs/reference/workers/generic-worker/support-tiers.mdx
+++ b/ui/docs/reference/workers/generic-worker/support-tiers.mdx
@@ -24,6 +24,7 @@ The Tier-1 platforms are:
 * generic-worker-multiuser-windows-386
 * generic-worker-multiuser-windows-amd64
 * generic-worker-simple-darwin-amd64
+* generic-worker-simple-darwin-arm64
 * generic-worker-simple-linux-amd64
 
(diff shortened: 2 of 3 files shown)


petemoore (approved):
Hi Dave, thanks for this.

I think this is ok, but I should add that the simple build was not really intended for production use, and we are hoping to stop supporting it at some point. Unfortunately, I can't log in to Bugzilla at the moment to see bug 1714200 for more context, but I'm guessing there is an issue with using the multiuser build, so I'm happy for this to go through, with the following request:

Could you also add it as a tier 2 platform to ui/docs/reference/workers/generic-worker/support-tiers.mdx ?

When I'm back in bugzilla, I'll have a look at the bug to see if I can help unblock any issues you might be having with multiuser, and then ideally the workers would migrate to multiuser and we can retire the simple engine.

Many thanks Dave!

davehouse:
Thanks Pete! There is no rush on this. So I'm good with waiting until you're back/ready.

We are running the single user generic-worker on the Catalina and BigSur macos macminis (x86 and M1). We ran into some issues with the multiuser worker and since we don't need the security of the user separation, we switched to single. We could look at switching to multiuser again (which we run on the mac builders), but singleuser is working well for the mac unit and perf tests.

davehouse:
thanks! I'll test it (and upgrade) at the next release

PR #4941, opened 2021-07-30
First cuts at telescope

imbstack:
- Fix monitoring hints table
- First cut at autogenerating telescope config

Unsure of how much of this is useful!


jwhitlock (commented):
Thanks @imbstack!

I think we're going to work on a set of service checks first, along with a manually created SVG, and then come back around to Taskcluster documentation. I think some of these changes will be useful, especially for bootstrapping monitoring of new TC clusters.

There are some changes we'd want regardless of what we do, such as a typo fix, and displaying the table. It may make sense to pull those into a new PR, so we don't forget to do it.

petemoore:
@jwhitlock I suspect we should close this in favour of the work we plan for observability in H1 2022, but we could keep it open if it serves as a useful reference. What are your thoughts? I think I'll remove assigned reviewers for now though.

jwhitlock:
Yes, let's close this. There may be useful bits, but I don't want this to constrain the next round of monitoring work.

PR #4945, opened 2021-08-05
[client-py] remove asyncclient retry debug log

### LINKED ISSUE #4944: `TypeError` in aio/asyncclient debug logging
**Describe the bug**
In the client-py `aio/asyncclient.py`, we log a debug message about which attempt we're trying [here](https://github.com/taskcluster/taskcluster/blob/797867c78082986e0814639b505c5b00eb43d735/clients/client-py/taskcluster/aio/asyncclient.py#L144). However, the variable name we're using, `retry`, points at a [function, not an int](https://github.com/taskcluster/taskcluster/blob/797867c78082986e0814639b505c5b00eb43d735/clients/client-py/taskcluster/aio/asyncclient.py#L16).

**To Reproduce**

1. Set the python `logging` level to debug
2. Make an API call through the python async client
3. See `TypeError` exception

**Expected behavior**
No exception thrown.

**Taskcluster version**
44.1.0

**Screenshots**
n/a

 - OS: MacOS 11.4

**Additional context**
PR incoming

escapewindow:
This appears to be a copy of this line [1] from the synchronous python
`client.py`. However, in that file, `retry` is an integer tracking which
attempt number we're on [2].

In the `asyncclient`, `retry` is a function [3], and does not format
into a %d. We likely didn't catch this because python logging doesn't
format % strings format % strings unless the log level includes the log
message to output. Because this is in a `log.debug` call and we probably
don't run with debug logging enabled, we haven't caught this issue.

We can rename `retry`, or import it under a different name, or track the
attempt under a different variable name and log that. Removing the
logging line seemed like the most expedient fix.

[1] https://github.com/taskcluster/taskcluster/blob/797867c78082986e0814639b505c5b00eb43d735/clients/client-py/taskcluster/client.py#L489
[2] https://github.com/taskcluster/taskcluster/blob/797867c78082986e0814639b505c5b00eb43d735/clients/client-py/taskcluster/client.py#L458
[3] https://github.com/taskcluster/taskcluster/blob/797867c78082986e0814639b505c5b00eb43d735/clients/client-py/taskcluster/aio/asyncclient.py#L16

Github Bug/Issue: Fixes #4944

### DIFF
--- a/changelog/issue-4944.md
+++ b/changelog/issue-4944.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 4944
+---
--- a/clients/client-py/taskcluster/aio/asyncclient.py
+++ b/clients/client-py/taskcluster/aio/asyncclient.py
@@ -141,7 +141,6 @@ class AsyncBaseClient(BaseClient):
                 # outside this loop.
                 headers['Content-Type'] = 'application/json'
 
-            log.debug('Making attempt %d', retry)
             try:
                 response = await asyncutils.makeSingleHttpRequest(
                     method, url, payload, headers, session=self.session


jwhitlock (approved):
Thanks @escapewindow! None of our linting tools can find a function formatted as an integer 😢 

If you think the debug message is useful, I think it could be added to the log message in ``retry``: 

https://github.com/taskcluster/taskcluster/blob/b239801344df21f733e3361991a072f533e23cc4/clients/client-py/taskcluster/aio/retry.py#L37-L39

Otherwise, we can merge as-is.

jwhitlock:
Merging, we can add the retry count later as needed.

PR #4953, opened 2021-08-10
tyop :p

petemoore:
Trivial typo...

### DIFF
--- a/changelog/DTGgHfyiTjeO3wYQkknuYQ.md
+++ b/changelog/DTGgHfyiTjeO3wYQkknuYQ.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
--- a/clients/client-py/README.md
+++ b/clients/client-py/README.md
@@ -376,7 +376,7 @@ A "writer" is an object with a `write(data)` method which writes the given data,
 A "writer factory" is a callable (again either async or sync) which returns a fresh writer, ready to write the first byte of the object.
 When uploads are retried, the writer factory may be called more than once.
 
-### Artifact Downlodas
+### Artifact Downloads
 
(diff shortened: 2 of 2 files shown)


jwhitlock (approved):
Thanks @petemoore! Feel free to fix this generate markdown weirdness if desired.

jwhitlock:
Oops, I detailed the weirdness but did not submit my comment.

README.md is marked as a generated file:

https://github.com/taskcluster/taskcluster/blob/main/.github/CODEOWNERS#L8-L12

which means it is collapsed by default in code reviews. However, it doesn't appear to be generated. There is this bit in the ``Makefile``:

https://github.com/taskcluster/taskcluster/blob/1296cadbe80dac4f4fdbf3732c0ddd0decf9f3cd/clients/client-py/Makefile#L22-L26

But [grip](https://github.com/joeyespo/grip) generates HTML from markdown, so ``README.html`` should be marked as the generated file. And, since GitHub natively renders Markdown since forever, I don't think we use this anymore.

petemoore:
Very interesting! Thanks John!

So I did a bit of digging, and it looks to me like that Makefile is no longer called. I installed some required dependencies, to see if it still works, and discovered that it runs a bunch of tests, of which most pass, but some fail, although that may be because I need some valid credentials with certain scopes. It also refers to a file that doesn't exist (`devDep.py`).

Rather than throw away the Makefile entirely at this point, I'll file a follow-up issue to investigate this stuff. If there are a bunch of tests that pass when appropriate credentials are used, probably they should be called as part of the CI. I'd be interested to know at what point we stopped running the tests, and why, or if they were just never integrated into the CI but just run by the module owner of the time.

So I'll land this, as is, and leave the rest for the follow up...

Issue #4614, opened 2021-03-23
Drop Python 2.7 support in Python client


jwhitlock:
The Python client supports Python 2.7 as well as recent Python 3 versions:

https://github.com/taskcluster/taskcluster/blob/cf14539f4cd5ba92e293eb456291e1897c2eb41c/clients/client-py/setup.py#L96-L99

We have a similar test matrix that includes later Python 3 versions:

https://github.com/taskcluster/taskcluster/blob/cf14539f4cd5ba92e293eb456291e1897c2eb41c/clients/client-py/tox.ini#L2-L7

Many projects have dropped Python 2.7 support after it reached end-of-line on Jan 1, 2020. Taskcluster should drop it as well, or plan to in a future release.

The documentation implies that the synchronous code is for Python 2 and the async code is for Python 3. The sync code should continue to work in Python 3, and I think deciding about future support of synchronous code should be separate from dropping Python 2.7.

djmitche:
I agree that we should keep the sync support even in Python 3.

I'm not sure if the `taskgraph` code still uses Python 2.  If so, we may want to favor keeping compatibility until it becomes problematic to do so.

jwhitlock:
Taskgraph on Python 3 may be tracked in [bug 1473144](https://bugzilla.mozilla.org/show_bug.cgi?id=1473144), but there's no activity for 2 years.

ahal:
The in-tree instance of taskgraph is already converted, but there's an out-of-tree one used for mobile projects (like Fenix) that is still Python 2. I filed [bug 1700716](https://bugzilla.mozilla.org/show_bug.cgi?id=1700716) to track that.

djmitche:
It looks like this is in the icebox for the moment.  If it becomes difficult to continue this maintenance, or if the out-of-tree taskgraph is updated, then we can put it on the schedule.

jwhitlock:
On May 25th [bug 1700716](https://bugzilla.mozilla.org/show_bug.cgi?id=1700716) was marked ``	RESOLVED FIXED``, so the out-of-tree taskgraph is now on Python 3 as well. I think we are unblocked for dropping Python 2.7 support in the Python client.

ahal:
Sorry not quite. Out of tree taskgraph now supports both but we still need to migrate consumers over. I think this will be unlocked when out of tree taskgraph drops py2 support.

[...]

PR #4958, opened 2021-08-11
Fix the example value for `registered_clients` in the dev config

Eijebong:
It's not expecting a map but a list. If you copied the previous value,
the web-server service wouldn't start and give this error:
`cfg.login.registeredClients.map is not a function`


petemoore (changes requested):
Many thanks for this! Unfortunately, the CI failed, and it looks like this is because the change gets overwritten when `yarn generate` is called. I think the change will need to made in infrastructure/tooling/src/generate/generators/k8s.js or some associated file that it uses.

Eijebong:
Woops, this completely slipped out of my mind. CI should be happier now.

djmitche:
(that client-rust failure is an intermittent)

petemoore:
@Eijebong I only see the changelog file now, but no code changes - can you take a look? Many thanks.

djmitche:
I pushed an update at https://github.com/djmitche/taskcluster/pull/new/fix-dev-config-example that should enable this.  It's worth looking for other config values that have the wrong type and adding them to the list.

petemoore:
> I pushed an update at https://github.com/djmitche/taskcluster/pull/new/fix-dev-config-example that should enable this. It's worth looking for other config values that have the wrong type and adding them to the list.

Hi @djmitche, are you going to create that as a separate PR, or should @Eijebong include that change in this PR? Thanks!

djmitche:
Please pull it into this PR.  I'm not a maintainer anymore so I can't push to this PR.

djmitche:
@petemoore do you want to have another look?  Are there other fields that should be included in this list?

petemoore (approved):
Looks great, many thanks!

djmitche:
I'm just a lowly contributor now, so you'll have to click for me!
image

petemoore:
@djmitche Is the rust client failure the same intermittent? And any idea about the python client failure?

djmitche:
The Python failure was temporary (but #5015  and #4614!), and yes, it's the same rust intermittent.

PR #4967, opened 2021-08-18
Add note for 44.2.1

jwhitlock:
The release for 44.2.0 failed when uploading the monoimage. Adding a note for the changelog.


jwhitlock:
Filed #4968 to look into the `docker-worker-chunk-1` failure, or at least fix one bit of it.

I'm going to go ahead and merge. In the past, the patch version was bumped with no changelog entry (v37.5.1, v34.0.1), so worst case is this one does have an entry.

Issue #4947, opened 2021-08-06
Deprecated usage of highlight.js 10.7.0


jwhitlock:
**Describe the bug**

When viewing a task in the UI, this message is displayed in the console:

```
Deprecated as of 10.7.0. highlight(lang, code, ...args) has been deprecated.
Deprecated as of 10.7.0. Please use highlight(code, options) instead.
https://github.com/highlightjs/highlight.js/issues/2277
```
The message is generated from ``Tasks.TaskGroup.cd2cd260.js:11:71365``.

**To Reproduce**

1. Open the developer tools, Console
2. Go to https://community-tc.services.mozilla.com/tasks
3. Click to view a task

**Expected behavior**

No deprecation message.

**Taskcluster version**

* Present in 44.1.0 on Community-TC
* Present in 43.2.0 on Firefox-ci

**Additional context**

The linked issue https://github.com/highlightjs/highlight.js/issues/2277 has a working example of changing the call to `highlight`.

Tanmay000009:
Hey, I am new around here. Can I contribute?

jwhitlock:
Thanks @Tanmay000009! A good place to start is getting a development environment setup. See https://github.com/taskcluster/taskcluster/blob/main/dev-docs/development-process.md. You'll need the first two sections, "Setup" and "Hacking on the UI". You then should be able to reproduce the issue in your development environment and start working on the fix.

Tanmay000009:
@jwhitlock thank you so much! The documentation is really helpful and easy to understand. I would like to work on the issue and try to contribute, if it's possible, please allot the issue to me.

Tanmay000009:
@jwhitlock  I am getting errors:
`yarn start`

yarn run v1.22.10
$ cross-env NODE_ENV=development node src/main.js devServer
AssertionError [ERR_ASSERTION]: readDbUrl is required
    at Function.setup (/media/tanmay000009/Web Dev/taskcluster/libraries/postgres/src/Database.js:60:5)
    at Object.exports.setup (/media/tanmay000009/Web Dev/taskcluster/db/src/setup.js:6:25)
    at setup (/media/tanmay000009/Web Dev/taskcluster/services/web-server/src/main.js:186:50)
    at /media/tanmay000009/Web Dev/taskcluster/libraries/loader/src/index.js:109:33
    at new Promise (<anonymous>)
    at /media/tanmay000009/Web Dev/taskcluster/libraries/loader/src/index.js:107:18
[...]

Issue #2637, opened 2020-04-10
AssertionError [ERR_ASSERTION]: readDbUrl is required


aryamanpuri:
**Describe the bug**
Unable to start the graphql server

**To Reproduce**

Follow the dev docs to produce local web server using https://github.com/taskcluster/taskcluster/blob/master/dev-docs/development-process.md#hacking-on-the-ui

**Error Log**
```
AssertionError [ERR_ASSERTION]: readDbUrl is required
    at Function.setup (/home/aryaman/taskcluster/libraries/postgres/src/Database.js:54:5)
    at Object.exports.setup (/home/aryaman/taskcluster/db/src/setup.js:6:25)
    at setup (/home/aryaman/taskcluster/services/web-server/src/main.js:188:48)
    at /home/aryaman/taskcluster/libraries/loader/src/index.js:109:33
    at new Promise (<anonymous>)
    at /home/aryaman/taskcluster/libraries/loader/src/index.js:107:18
```

**Additional context**
I don't if there are any server code changes that are giving this error.

helfi92:
We're in the midst of a migration today. Once the migration is done, I think you would be able to have it point at the community deployment db url as a default. For now, I think you can just `git checkout 012c24e4073df8a8e67c41836ec9977324ba5571` (commit right before the db change) and then try again. It should work.

djmitche:
Credentials for that URL won't be available.  I think we'll have to run postgres in docker from here on out, even for running the web-server process.

aryamanpuri:
It is not working. Giving the same error. Hold on until the migration is done.
>  For now, I think you can just `git checkout 012c24e4073df8a8e67c41836ec9977324ba5571` (commit right before the db change) and then try again. It should work.

djmitche:
Do the docs in #2638 help?

rohanharikr:
@aryamanpuri if you are still facing this issue, try these commands
```export READ_DB_URL=postgresql://postgres@localhost/postgres

export WRITE_DB_URL=postgresql://postgres@localhost/postgres
```
thanks @helfi92 !

PR #4977, opened 2021-08-30
Update development-process.md

Tanmay000009:
Github Bug/Issue: Fixes https://github.com/taskcluster/taskcluster/issues/4947#issuecomment-908614772

Added two snippets under: https://github.com/taskcluster/taskcluster/blob/main/dev-docs/development-process.md#hacking-on-the-ui under 
- Set both READ_DB_URL and WRITE_DB_URL to the URL for your postgres server, as described above.

Reference: https://github.com/taskcluster/taskcluster/issues/2637#issuecomment-617997973 

Also added a note about setting all environment variables in the same shell session.

### DIFF
--- a/dev-docs/development-process.md
+++ b/dev-docs/development-process.md
@@ -92,8 +92,6 @@ To run the Taskcluster UI:
     For example:
     ```sh
     export READ_DB_URL=$TEST_DB_URL
-    ```
-    ```sh
     export WRITE_DB_URL=$TEST_DB_URL
     ```
     
--- a/dev-docs/development-process.md
+++ b/dev-docs/development-process.md
@@ -89,6 +89,14 @@ To run the Taskcluster UI:
 * In a shell window for taskcluster-web-server:
   * Make sure Postgres is set up as described above.
   * Set both `READ_DB_URL` and `WRITE_DB_URL` to the URL for your postgres server, as described above.
+    For example:
+    ```sh
+    export READ_DB_URL=$TEST_DB_URL
+    ```
+    ```sh
+    export WRITE_DB_URL=$TEST_DB_URL
+    ```
+    
   * Set `TASKCLUSTER_ROOT_URL` to point to a Taskcluster deployment you wish to represent in the UI.
     For example:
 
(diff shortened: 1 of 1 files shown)


dev-docs/development-process.md
```diff
@@ -89,6 +89,14 @@ To run the Taskcluster UI:
 * In a shell window for taskcluster-web-server:
   * Make sure Postgres is set up as described above.
   * Set both `READ_DB_URL` and `WRITE_DB_URL` to the URL for your postgres server, as described above.
+    For example:
+    ```sh
+    export READ_DB_URL=$TEST_DB_URL
+    ```
+    ```sh
+    export WRITE_DB_URL=$TEST_DB_URL
+    ```
```
jwhitlock, line 98: Let's combine the two into a single block:

```suggestion
    ```sh
    export READ_DB_URL=$TEST_DB_URL
    export WRITE_DB_URL=$TEST_DB_URL
    ```
```

dev-docs/development-process.md
```diff
@@ -106,6 +114,8 @@ To run the Taskcluster UI:
     You will only need to do this when the dependencies change.
   * Run `yarn start` to start the development server.
     It will output a localhost URL on which you can see the site.
+    
+    *Note* : Make sure you have set all the above mentioned environment variables in the shell session.
```
jwhitlock, line 118: This note should be in the previous section. The commands run in `services/web-server` need the environment variables, while the ones run in `ui` do not.

There's already a note in that block, you may want to format both as sentences, like 

> "If you get an error like `readDbUrl is required`, ensure you've set all the environment variables in the shell session."

or both notes like this new style:

> *Note 1*: The pulse note
> *Note 2*: The environment note

jwhitlock (commented):
Thanks @Tanmay000009! I think the sample command improves the docs. The note about setting environment variables is in the wrong section, can be moved up.

dev-docs/development-process.md
Tanmay000009: Sure got it!

Tanmay000009:
I have got the changes, and will make a clean PR with changelogs aswell!

PR #4981, opened 2021-09-01
Fix "People of Taskcluster page"

petemoore:
Assuming I'm not entirely incompetent, this should add @leplatrem as staff and change @imbstack and @djmitche to staff alumni...

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -46,7 +46,7 @@
       "profile": "http://code.v.igoro.us/",
       "contributions": [
         "code",
-        "staff"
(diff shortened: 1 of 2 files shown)


leplatrem (approved):

jwhitlock (approved):
Thanks @petemoore! I think the bot updates the README on merge to main, let's see...

petemoore:
Looks like it worked! https://docs.taskcluster.net/docs/people

Although now I see from reading that page I could have used the CLI or bot to generate that change, never mind!

PR #4984, opened 2021-09-01
Fix spelling of spelling.js

jwhitlock:
Remove a bit of comic irony from the code base.

The `meta` plugins are [loaded by path](https://github.com/taskcluster/taskcluster/blob/main/infrastructure/tooling/src/meta/checks/index.js), so there should be no string change in `index.js`.


leplatrem (approved):
:rofl:

Issue #4999, opened 2021-10-07
worker-manager: Azure workers register when state != REQUESTED


jwhitlock:
**Describe the bug**

About 100 times a day, the worker-manager for Firefox-CI logs the error "Worker was already running.". This is from the `register()` method for the Azure provider:

https://github.com/taskcluster/taskcluster/blob/685a0470d6ddb845552f33c245a18503c990c634/services/worker-manager/src/providers/azure/index.js#L620-L623

**To Reproduce**

In the GCP log explorer for the Firefox-CI project, filter with:

```
labels.k8s-pod/app_kubernetes_io/component="taskcluster-worker-manager-web"
jsonPayload.Type="registration-error-warning"
```

**Expected behavior**

Workers should only register in the ``REQUESTED`` state. They could re-register, but that is a different endpoint.

The log message assumes the worker is in the ``RUNNING`` state, but does not provide enough information to determine the actual state.

**Taskcluster version**

Firefox CI is running 43.2.0. This is probably present in the current main branch and community-tc, but there are no Azure worker pools in community-tc.

lotas:
Debugging one of those issues, found that this error happens right after worker was removed (or marked for removal):
<img width="1643" alt="image" src="https://user-images.githubusercontent.com/83861/158397017-65de1dfc-e94f-4fb8-afa5-4ece7c57df86.png">

lotas:
after following logs for couple of such VMs, they all follow similar pattern: (timing approximate)

* `t+0` - `worker-requested` - provisioner is creating new instance
* `t+22min` - `"vm instance view not found"`  - worker-scanner is doing `checkWorker()` and `queryInstance()`
* right after: `"provisioning resource` - `azure.checkWorker()` + `azure.provisionResource()`
* `t+90min` - same as above - `"vm instance view not found"`
* `t+120min` - same as above -`"vm instance view not found"`
* `t+121min` - `"fetched instance view"` - instance is in the `running` state, but  `"worker-removed"` says  that `"terminateAfter time exceeded"`
and the last one is coming either from `DELETE /workers/../../..` call or from `worker-scanner's` `azure.checkWorker()`

So as I understand it instances are being created, after ~2 hours are being terminated because of `terminateAfter` flag on the instance says it should be terminated regardless of the state

https://github.com/taskcluster/taskcluster/blob/main/services/worker-manager/src/providers/azure/index.js#L1042-L1047

**Questions:**
1. is this `terminateAfter` time set properly?
[...]

PR #5000, opened 2021-10-07
Expand Azure register warning context

jwhitlock:
Add ``workerPoolId``, ``providerId``, and ``workerId`` to ``registration-error-warning`` context, logged from the Azure provider's ``register()`` function in worker-manager.

Add ``workerState`` to the warning context when the ``state`` is not ``REQUESTED``, to help diagnose registration with an unexpected state.

Github Bug/Issue: #4999 (helps diagnose, doesn't yet fix)

### DIFF
--- a/services/worker-manager/src/providers/azure/index.js
+++ b/services/worker-manager/src/providers/azure/index.js
@@ -466,7 +466,13 @@ class AzureProvider extends Provider {
       let asn1 = forge.asn1.fromDer(forge.util.createBuffer(decodedMessage));
       message = forge.pkcs7.messageFromAsn1(asn1);
     } catch (err) {
-      this.monitor.log.registrationErrorWarning({ message: 'Error extracting PKCS#7 message', error: err.toString() });
+      this.monitor.log.registrationErrorWarning({
+        message: 'Error extracting PKCS#7 message',
+        error: err.toString(),
+        workerPoolId: workerPool.workerPoolId,
+        providerId: this.providerId,
+        workerId: worker.workerId,
+      });
       throw error();
     }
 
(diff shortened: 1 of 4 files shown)


petemoore (approved):
Looks great, thanks John!

jwhitlock:
Every time I think about merging this, I look at the repeated code and think there must be a way to get those variables automatically into the logging context.

I'm still holding off until 1) I can research that, or 2) we cut a new TC version

services/worker-manager/src/providers/azure/index.js
```diff
@@ -466,7 +466,13 @@ class AzureProvider extends Provider {
...
       message = forge.pkcs7.messageFromAsn1(asn1);
     } catch (err) {
-      this.monitor.log.registrationErrorWarning({ message: 'Error extracting PKCS#7 message', error: err.toString() });
+      this.monitor.log.registrationErrorWarning({
+        message: 'Error extracting PKCS#7 message',
+        error: err.toString(),
+        workerPoolId: workerPool.workerPoolId,
+        providerId: this.providerId,
+        workerId: worker.workerId,
+      });
```
djmitche, line 475: Inside this method (so, inside the scope wehre workerPool, this, and worker are defined), you could define a closure to do this:
```js
   let logRegError = (err, message) => {
      this.monitor.log.registrationErrorWarning({
        message, err,
        workerPoolId: workerPool.workerPoolId,
        providerId: this.providerId,
        workerId: worker.workerId,
      });
```

Another option is to make a child logger that contains some of those values, but then you're passing a reference to that logger everywhere.

TBH I wouldn't worry about the verbosity here.

jwhitlock: *option 3 - Add a comment and Dustin will prototype the solution* 😁  

Thanks! I'm going to try the closure, see what it looks like.

jwhitlock: I ran out of time to make this look nice. Maybe when someone re-writes the Azure provisioner they can fix this...

PR #5001, opened 2021-10-08
Changed the widget for non logged in user in viewclient to signin dialog

### LINKED ISSUE #2679: [UI] Prompt for logging in immediately when using `taskcluster signin`.
**Describe the bug**
When signing in via `taskcluster sign`, if you are not signed via the web (or the signin has expired), it gives you a page that says "sign in to create client". You need to then click in the top right to sign in. 

**Expected behavior**

It would be nice if instead, it automatically prompted you to sign in, rather that requiring the extra click.

cerkiewny:
Github Bug/Issue: Fixes #2679

### DIFF
--- a/changelog/issue-2679.md
+++ b/changelog/issue-2679.md
@@ -1,0 +1,5 @@
+audience: users
+level: minor
+reference: issue 2679
+---
+Now after `taskcluster signin` the sign in dialog is shown instead of message if user is not logged in.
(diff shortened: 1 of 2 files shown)


cerkiewny:
@jwhitlock usually it's clean up function used to clearing the "dialog open" state, in this case I am always rendering it as open because the higher up component manages the logic. Login in should clear the dialog and display the correct page.

PR #5005, opened 2021-10-25
Fix broken link

petemoore:
Spotted this broken link while browsing a task log.

Some [history](https://bugzilla.mozilla.org/show_bug.cgi?id=1549312) on the problem of broken links.


petemoore:
I've created #5006 for the broken CI - seems to be unrelated to the change.

PR #5013, opened 2021-12-14
Bug 1635730 - clean up /private/var/folders on macOS multiuser

petemoore:
Bugzilla Bug: [1635730](https://bugzilla.mozilla.org/show_bug.cgi?id=1635730)

Note, this should be tested on a real macOS worker. It may be that the root user won't have permission to do this cleanup, so the code won't fail if the delete fails, it will simply report the failure.

### DIFF
--- a/changelog/bug-1635730.md
+++ b/changelog/bug-1635730.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: bug 1635730
+---
+generic-worker multiuser engine running on macOS will now attempt to cleanup /private/var/folders when deleting a task OS user account.
--- a/workers/generic-worker/runtime/runtime_darwin.go
+++ b/workers/generic-worker/runtime/runtime_darwin.go
@@ -44,6 +44,10 @@ func (user *OSUser) CreateNew(okIfExists bool) (err error) {
 }
 
 func DeleteUser(username string) (err error) {
+	err = host.Run("/usr/bin/find", "/private/var/folders", "-user", username, "-delete")
+	if err != nil {
+		log.Printf("WARNING: Error when trying to delete files under /private/var/folders belonging to %v: %v", username, err)
+	}
 	err = host.Run("/bin/bash", "-c", `/usr/bin/sudo dscl . -delete '/Users/`+username+`'`)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


petemoore:
CC @djmitche @davehouse

djmitche (approved):
This is worth a try.  But, I'm worried that it will take hours even on a lightly-loaded Mac.  I think the thing that was slow when i was running similar commands was the error handling for the un-deletable files.  So this may end up in a spot where whatever phase of the worker this deletion occurs in (startup?) takes longer and longer.

davehouse:
It looks like this will run before claiming a taskf or the previous user and after completing a task: https://github.com/taskcluster/taskcluster/blob/2f46d0118f89b49b293f9bf1b59a202acbd86b52/workers/generic-worker/main.go#L440
Maybe if workers take a long time on this cleanup after tasks, we could put a timeout or something on it?

petemoore:
> This is worth a try. But, I'm worried that it will take hours even on a lightly-loaded Mac. I think the thing that was slow when i was running similar commands was the error handling for the un-deletable files. So this may end up in a spot where whatever phase of the worker this deletion occurs in (startup?) takes longer and longer.

I think the reason the cleanup [took hours in the past ](https://bugzilla.mozilla.org/show_bug.cgi?id=1635730#c7) is because there was a significant backlog of files from previous task runs. With this change, the cleanup should occur after every task, so there shouldn't be a build up of files, if all goes well. In other words, I believe there is a good chance that the cleanup time will not be significant, and not grow over time, but indeed we won't know for sure until we try it.

petemoore:
> It looks like this will run before claiming a taskf or the previous user and after completing a task:
> 
> https://github.com/taskcluster/taskcluster/blob/2f46d0118f89b49b293f9bf1b59a202acbd86b52/workers/generic-worker/main.go#L440
> 
> Maybe if workers take a long time on this cleanup after tasks, we could put a timeout or something on it?

I agree that could be an option, but it may also mask the problem. In a way I would prefer to roll it out as is so we can monitor how long it takes, and whether it increases over time (and also what affect it has on log sizes), and then if we notice it is growing over time, we should investigate how to fix it, rather than just timeout, and still have a growing filesystem over time, and have a time delay too. Does that work for you too?

davehouse:
Thanks!

Issue #5016, opened 2021-12-15
Rust client CI fails


djmitche:
The Rust worker's CI is failing frequently, e.g., in https://github.com/taskcluster/taskcluster/pull/4958

djmitche:
@petemoore if you want to assign this to me, I can take a look.

petemoore:
Thanks Dustin!

djmitche:
This is intermittent, and I'm having a hard time reproducing it locally.  I may resort to adding some debug output and waiting for it to fail in CI.

djmitche:
Locally, I can't even reproduce this running in a tight cycle or running the whole test suite (in case there's something "leaking" between test cases).

djmitche:
(I tried running the docker container locally, too, to no avail)

djmitche:
No failures on main since this landed :(

djmitche:
I tried running a few of these on the tutorial worker -- https://community-tc.services.mozilla.com/tasks/groups/GluSZhtQR7uCHqz4bCcPgg

djmitche:
```
running 5 tests
test factory::test::cursor_reader_twice ... ok(B
test test::small_data_inline_upload ... ok(B
test test::large_data_inline_upload ... ok(B
test factory::test::file_reader_twice ... FAILED(B
test test::put_url ... ok(B
failures:
---- factory::test::file_reader_twice stdout ----
thread 'factory::test::file_reader_twice' panicked at 'assertion failed: `(left == right)`
  left: `[]`,
 right: `[72, 69, 76, 76, 79, 47, 87, 79, 82, 76, 68]`', upload/src/factory.rs:87:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
That's not the error I was trying to reproduce :(

djmitche:
@petemoore is this still failing frequently?

djmitche:
I'm assuming this is more-or-less fixed..

PR #5018, opened 2021-12-17
add debug info to client::tests::test_simple_request_with_perm_creds

djmitche:
Github Bug/Issue: #5016

This just adds debug info.  Assuming I can make this fail on the PR, I shouldn't need to merge this.

### DIFF
--- a/changelog/issue-5106.md
+++ b/changelog/issue-5106.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 5106
+---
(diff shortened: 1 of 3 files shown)


petemoore:
@djmitche do you want to try pushing some empty commits to trigger the intermittent failure, or do you want me to rubber stamp it and merge it (and potentially revert it later if it is too noisy)?

djmitche:
close/reopen should trigger new runs.  I'll try to do a few of those.  If it still doesn't reproduce, then the `println!`s I added are probably adjusting some timing such that the error doesn't occur -- which suggests this is a race condition :(

djmitche:
Well, this _is_ frustrating, isn't it.

djmitche:
(btw I don't have perms to rerun or retrigger or edit-and-create)

djmitche:
@petemoore let's merge this to main.  It's definitely not noisy.  And if it doesn't fail anymore, 🤷🏽 ?

petemoore (approved):

PR #5020, opened 2021-12-21
Simpler markdown yaml json

petemoore:
Bugzilla Bug: [1735159](https://bugzilla.mozilla.org/show_bug.cgi?id=1735159)

Simplifies rendering code for markdown / json / yaml content.

### DIFF
--- a/ui/src/components/Code/index.jsx
+++ b/ui/src/components/Code/index.jsx
@@ -40,12 +40,10 @@ export default class Code extends Component {
       ignoreIllegals: true,
     }).value;
 
-    /* eslint-disable react/no-danger */
     return (
       <pre className={classNames(`language-${language}`, className)} {...props}>
-        {code && <code dangerouslySetInnerHTML={{ __html: code }} />}
+        {code && <code>{code}</code>}
       </pre>
     );
-    /* eslint-enable react/no-danger */
   }
 }
--- a/ui/src/components/Markdown/index.jsx
+++ b/ui/src/components/Markdown/index.jsx
@@ -200,16 +200,10 @@ export default class Markdown extends Component {
   render() {
     const { classes, children, className, ...props } = this.props;
 
-    /* eslint-disable react/no-danger */
     return (
-      <span
-        className={classNames(classes.root, className)}
-        dangerouslySetInnerHTML={{
-          __html: markdown.render(children),
-        }}
-        {...props}
-      />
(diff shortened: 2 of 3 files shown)


petemoore:
I suspect this may break the rendering, since probably the markdown and json/yaml rendering code generates native html, but just confirming my suspicions here...

djmitche:
It looks like this is mixing two PRs?

ui/src/components/Code/index.jsx
```diff
@@ -40,12 +40,10 @@ export default class Code extends Component {
       ignoreIllegals: true,
     }).value;
 
-    /* eslint-disable react/no-danger */
     return (
       <pre className={classNames(`language-${language}`, className)} {...props}>
-        {code && <code dangerouslySetInnerHTML={{ __html: code }} />}
+        {code && <code>{code}</code>}
```
djmitche, line 45: I believe this will render literal HTML, which is not what we want.  I believe the output of `highlight` is escaped and considered safe.

ui/src/components/Markdown/index.jsx
```diff
@@ -199,17 +199,8 @@ export default class Markdown extends Component {
...
-    /* eslint-disable react/no-danger */
     return (
-      <span
-        className={classNames(classes.root, className)}
-        dangerouslySetInnerHTML={{
-          __html: markdown.render(children),
-        }}
-        {...props}
-      />
+      <span className={classNames(classes.root, className)} {...props}>{markdown.render(children)}</span>
```
djmitche, line 203: This will cause all Markdown to be rendered as onscreen HTML.  I don't think that's what you want :)

I believe that the output of `markdown.render` is considered safe.

djmitche (commented):
You can verify both of my "I believe" :)

ui/src/components/Code/index.jsx
petemoore: Oh sorry, you don't have access to the bug, but this was a quick test to see if this would cause double-escaping (for the reason you cited)! I made it a draft PR because I didn't expect it to be needed, but if it rendered correctly, it would have implied that escaping wasn't being done underneath, and would have been straightforward to mark it as ready for review.

ui/src/components/Markdown/index.jsx
petemoore: Same as above!

PR #5021, opened 2021-12-21
[client-py] for `upload_artifact` let `contet` be `bytes` or `str`.

### LINKED ISSUE #5022: upload_artifact from python client should be also able to upload bytes besides str
As we've discussed on matrix taskcluster channel `upload_artifact` should also be able to upload bytes stream instead of str.

abpostelnicu:
As discussed on the matrix `taskcluster` channel we should also be able to send bytes.
Closes #5022

### DIFF
--- a/changelog/issue-5021.md
+++ b/changelog/issue-5021.md
@@ -1,0 +1,5 @@
+audience: developers
+level: minor
+reference: issue 5021
+---
+for `upload_artifact` from `client-py` let `contet` be `bytes` or `str`.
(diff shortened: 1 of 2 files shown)


petemoore (approved):
Looks great, many thanks!

djmitche:
I do want to emphasize
https://github.com/taskcluster/taskcluster/blob/23e90e1b3a0e0c0e5148928ddb7ef62064a79384/clients/client-py/taskcluster/helper.py#L149-L152

That's not reason not to land this change, but this method will likely be removed soon (not least because it does not support non-S3 objects), so please plan to move away.

abpostelnicu:
Yes, we shall do that. I want to use this as a temporary fix before we schedule the transition to the new API.

PR #5026, opened 2021-12-22
Temp fix - use curl `--insecure` with docker image node:14.17.5

petemoore:
Release 44.3.0 failed due to an issue with the docker image node:14.17.5.

This is a workaround for that issue, until it has been resolved upstream.

See the [failed release task](https://community-tc.services.mozilla.com/tasks/bVGir4inRTiglqpaIN0lPA/runs/0/logs/public/logs/live.log#L904) (around line 904).

### DIFF
--- a/changelog/bRGHs0kbQ5ijt2Sec0Mr_A.md
+++ b/changelog/bRGHs0kbQ5ijt2Sec0Mr_A.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
--- a/workers/docker-worker/release.sh
+++ b/workers/docker-worker/release.sh
@@ -34,7 +34,24 @@ mkdir $DW_ROOT/node
 curl https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.xz | tar -C $DW_ROOT/node --strip-components=1 -xJf -
 
 # Install Yarn (later to be removed)
-curl -L https://yarnpkg.com/latest.tar.gz | tar --transform 's|yarn-[^/]*/|yarn/|' -C $DW_ROOT -zvxf -
+# TODO: remove --insecure from command below once either the following command
+# runs successfully (implies issue fixed upstream in node:14.17.5 docker image):
+#
+#   docker run -ti --rm node:14.17.5 curl -LI https://yarnpkg.com/latest.tar.gz
+#
+# or once we have upgraded to node 14.18.2 or higher (where this issue is no
+# longer present).
+#
+# The above docker command currently fails with:
+#   "curl: (60) SSL certificate problem: certificate has expired"
+# which is why we added `--insecure` to the command. The issue is with the
+# certificate bundle in the node:14.17.5 docker image.
+#
+# Note, this is also reported in https://github.com/yarnpkg/yarn/issues/8700
+# but it is _not_ a problem with yarnpkg/yarn - instead it is a problem with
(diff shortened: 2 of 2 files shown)


djmitche (approved):
Can you link to the issue in the comments? 

I noticed the issue linked in Matrix was for `classic.yarnpkg.com`, where this is `yarnpkg.com`.

Is this fixed with a newer node?  Maybe this fix is "until we update node" (like to v16)

petemoore:
> Can you link to the issue in the comments?

I could link to [the issue](https://github.com/yarnpkg/yarn/issues/8700) in the comments, but it isn't something yarn can fix (where it was reported) but unfortunately an issue with the certificate bundle on docker image node:14.17.5. In other words, the status of that issue might not accurately reflect when the upstream fix is ready.

However, I checked and node:14.18.2 is working:

```
$ docker run -ti --rm node:14.18.2 curl -LI https://yarnpkg.com/latest.tar.gz
```

This above invocation runs without problems, so therefore I think instead I'll update the comments to say we should remove `--insecure` when we have upgraded to node 14.18.2 or higher (not sure if we can/should upgrade to node 16/17?).

> I noticed the issue linked in Matrix was for `classic.yarnpkg.com`, where this is `yarnpkg.com`.

It turns out to be the same issue. The output is:

```
1234567891011121314151617181920212223242526272829303132333435363738394041pmoore@Petes-iMac:~/git/taskcluster main $ docker run -ti --rm node:14.17.5 curl -LI https://yarnpkg.com/latest.tar.gz
Unable to find image 'node:14.17.5' locally
14.17.5: Pulling from library/node
eb18d230e067: Pull complete 
83600c1b4583: Pull complete 
4ae15c65bfa0: Pull complete 
c19c058edda5: Pull complete 
05cdaa0fd103: Pull complete 
8461dcff50c4: Pull complete 
84be4117f79d: Pull complete 
4ccb803887fd: Pull complete 
ab52680a5469: Pull complete 
Digest: sha256:c1fa7759eeff3f33ba08ff600ffaca4558954722a4345653ed1a0d87dffed9aa
Status: Downloaded newer image for node:14.17.5
HTTP/2 301 
date: Wed, 22 Dec 2021 12:50:11 GMT
content-type: text/plain; charset=utf-8
content-length: 56
location: https://classic.yarnpkg.com/latest.tar.gz
cache-control: public, max-age=14400
x-nf-request-id: 01FQH329BA0HJT59V1QKKAEH0X
cf-cache-status: MISS
expires: Wed, 22 Dec 2021 16:50:11 GMT
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
server: cloudflare
cf-ray: 6c1979754cb1d721-FRA

curl: (60) SSL certificate problem: certificate has expired
More details here: https://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
pmoore@Petes-iMac:~/git/taskcluster main $ 
```

Looking closely you'll see:

```
HTTP/2 301 
....
location: https://classic.yarnpkg.com/latest.tar.gz
```

PR #5031, opened 2022-01-03
Drop support for Python-2.7 in the client.

### LINKED ISSUE #4614: Drop Python 2.7 support in Python client

djmitche:
Github Bug/Issue: Fixes #4614 

Always fun to delete some code :)

### DIFF
--- a/clients/client-py/README.md
+++ b/clients/client-py/README.md
@@ -6,8 +6,7 @@
 **A Taskcluster client library for Python.**
 
 This library is a complete interface to Taskcluster in Python.  It provides
-both synchronous and asynchronous interfaces for all Taskcluster API methods,
-in both Python-2 and Python-3 variants.
+both synchronous and asynchronous interfaces for all Taskcluster API methods.
 
 ## Usage
 
@@ -22,9 +21,8 @@ required, and it's unusual to configure any other options aside from
 `credentials`.
 
 For each service, there are sync and async variants.  The classes under
-`taskcluster` (e.g., `taskcluster.Queue`) are Python-2 compatible and operate
-synchronously.  The classes under `taskcluster.aio` (e.g.,
-`taskcluster.aio.Queue`) require Python >= 3.6.
+`taskcluster` (e.g., `taskcluster.Queue`) operate synchronously.  The classes
+under `taskcluster.aio` (e.g., `taskcluster.aio.Queue`) are asynchronous.
 
 #### Authentication Options
 
@@ -335,7 +333,6 @@ This library provides convenience methods to implement the client portion of tho
 These methods will negotiate the appropriate method with the object service and perform the required steps to transfer the data.
 
 All methods are available in both sync and async versions, with identical APIs except for the `async`/`await` keywords.
-These methods are not available for Python-2.7.
 
 In either case, you will need to provide a configured `Object` instance with appropriate credentials for the operation.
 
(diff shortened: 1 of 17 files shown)


clients/client-py/README.md
```diff
@@ -22,9 +21,8 @@ required, and it's unusual to configure any other options aside from
 `credentials`.
 
 For each service, there are sync and async variants.  The classes under
-`taskcluster` (e.g., `taskcluster.Queue`) are Python-2 compatible and operate
-synchronously.  The classes under `taskcluster.aio` (e.g.,
-`taskcluster.aio.Queue`) require Python >= 3.6.
```
grahamalama, line 27: > require Python >= 3.6
Is this still true?

clients/client-py/setup.py
```diff
@@ -32,11 +31,10 @@
     'mohawk>=0.3.4',
     'slugid>=2',
     'taskcluster-urls>=12.1.0',
-    'six>=1.10.0',
 
     # python-3 only
```
grahamalama, line 35: Probably don't need this heading anymore, right?

clients/client-py/setup.py
```diff
@@ -96,8 +90,10 @@ def run_tests(self):
         },
         cmdclass={'test': Tox},
         zip_safe=False,
-        classifiers=['Programming Language :: Python :: 2.7',
-                     'Programming Language :: Python :: 3.5',
-                     'Programming Language :: Python :: 3.6',
-                     'Programming Language :: Python :: 3.7'],
+        classifiers=[
```
grahamalama, line 93: Two questions here:
- is this client Python 3.10 compatible? (or is that out of scope to account for in this PR?)
- I noticed in other places that we removed references to Python 3.6 (which is also at EOL). Should we remove the classifier here?

clients/client-py/tox.ini
```diff
@@ -1,6 +1,5 @@
 [tox]
 envlist =
-    py27
     py36
```
grahamalama, line 3: Similar question to above about Python 3.6 & 3.10

grahamalama (commented):
Love cutting lines of code. Just a few questions left inline.

clients/client-py/setup.py
djmitche: I think that's out of scope.  We typically haven't dropped support for old versions until they become a nuisance (as 2.7 did recently), so I'd suggest keeping 3.6 for now.  Adding 3.10 is hopefully just a matter of adding a test task for it.

The removals of `; python_version>="3.6"` were removals of a filter to only install those deps on Python-3, not to drop Python-3.6 support.

djmitche:
Hm, the "re-request review" button isn't working for me, but plz re-review :)

grahamalama (approved):

petemoore (approved):

Issue #5041, opened 2022-01-08
Add a config option to be able to use a docker image from a private repository in helm charts


Eijebong:
I have a patch that works when you specify a sercret but needs to be tested without one.

<details><summary>Untested patch</summary>
<p>

```diff
diff --git a/infrastructure/k8s/templates/taskcluster-auth-cron-purgeExpiredClients.yaml b/infrastructure/k8s/templates/taskcluster-auth-cron-purgeExpiredClients.yaml
index 26f47781c..0296d93ed 100644
--- a/infrastructure/k8s/templates/taskcluster-auth-cron-purgeExpiredClients.yaml
+++ b/infrastructure/k8s/templates/taskcluster-auth-cron-purgeExpiredClients.yaml
@@ -19,6 +19,10 @@ spec:
         metadata:
           labels: *ref_0
         spec:
+        {{- with.Values.imagePullSecrets }}
+          imagePullSecrets:
+            {{- toYaml . | nindent 12 }}
+        {{- end }}
           restartPolicy: OnFailure
           containers:
             - name: taskcluster-auth-purgeexpiredclients
diff --git a/infrastructure/k8s/templates/taskcluster-auth-deployment-web.yaml b/infrastructure/k8s/templates/taskcluster-auth-deployment-web.yaml
index a0d88a16c..044dba9a0 100644
--- a/infrastructure/k8s/templates/taskcluster-auth-deployment-web.yaml
+++ b/infrastructure/k8s/templates/taskcluster-auth-deployment-web.yaml
@@ -17,6 +17,10 @@ spec:
         checksum/secret: '{{ include (print $.Template.BasePath "/taskcluster-auth-secret.yaml") . | sha256sum }}'
       labels: *ref_0
     spec:
+    {{- with.Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
       serviceAccountName: taskcluster-auth
       containers:
         - name: taskcluster-auth-web
diff --git a/infrastructure/k8s/templates/taskcluster-built-in-workers-deployment-server.yaml b/infrastructure/k8s/templates/taskcluster-built-in-workers-deployment-server.yaml
index 302fadac2..55bbb1e55 100644
--- a/infrastructure/k8s/templates/taskcluster-built-in-workers-deployment-server.yaml
+++ b/infrastructure/k8s/templates/taskcluster-built-in-workers-deployment-server.yaml
[...]

PR #5060, opened 2022-01-17
Add `imagePullSecrets` to all helm charts that pull a docker containers

Eijebong:
This allows using a private docker repository with the helm chart.

### DIFF
--- a/infrastructure/k8s/values.schema.json
+++ b/infrastructure/k8s/values.schema.json
@@ -7,6 +7,10 @@
       "description": "The name of this deployment of Taskcluster.",
       "type": "string"
     },
+    "imagePullSecrets": {
+      "description": "Secrets to get the docker image",
+      "type": "array"
+    },
     "auth": {
       "additionalProperties": false,
       "properties": {
--- a/changelog/Pn46CTBsQ-qH-pKCgUC-mg.md
+++ b/changelog/Pn46CTBsQ-qH-pKCgUC-mg.md
@@ -1,0 +1,4 @@
+audience: deployers
+level: minor
+---
+Added `imagePullSecrets` to the helm chart configuration to allow private docker repositories for deployment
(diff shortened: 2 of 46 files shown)


infrastructure/k8s/values.schema.json
```diff
@@ -7,6 +7,10 @@
       "description": "The name of this deployment of Taskcluster.",
       "type": "string"
     },
+    "imagePullSecrets": {
+      "description": "Secrets to get the docker image",
```
petemoore, line 11: It isn't clear to me what the format of each secret is, and how they are used to return the docker image, and which  docker image the description refers to (although I guess it is the one and only image that is shared across all the service containers). Would it be ok to add a couple of words to explain how the secrets are used to retrieve the docker image, and just clarify which docker image with something like "the taskcluster services docker image"? Thanks @Eijebong !

petemoore (commented):
Thanks for creating this @Eijebong.

Do we have any other deployment docs where it might make sense to introduce/explain this new property?

petemoore:
Fixes #5041.

infrastructure/k8s/values.schema.json
lotas: This is from https://kubernetes.io/docs/concepts/containers/images/#using-a-private-registry

lotas: Format there is indeed an array, but elements should be  a dict with `name` key present. Would be good to specify it, indeed.

```yaml
imagePullSecrets:
   - name: secret name
```

How often do we need more than one secret name? 

I guess, one way to avoid using schema is to use single entry like:

```yml
      imagePullSecrets:
        - name: {{ .Values.imagePullSecrets }}
```   
Wdyt?

lotas:
@Eijebong I'll close this in favor of #5642  
Those yamls needs to be autogenerated with some hacks to overcome json-e strictness

Issue #5079, opened 2022-01-24
Fix code scanning alert - Missing rate limiting


matt-boris:
Tracking issue for:
- [ ] https://github.com/taskcluster/taskcluster/security/code-scanning/94

lotas:
I don't think it makes sense to implement this feature, I commented in both PR and on the code-scanning alert. Let's close it

PR #5080, opened 2022-01-24
For #5079: Add rate limiting to web server auth requests

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/5079


petemoore:
@djmitche @Eijebong @MasterWayZ does this look ok to you?

Eijebong:
If it's just the auth endpoint the default limit seems high, if it's global to the application then it seems low.
Might be a good idea to document the scope of the limit and that setting it to 0 will disable it too.

jwhitlock (dismissed):
Thanks @matt-boris for looking into this. This seems fine to me, but is also a bit weird.

If I go to Firefox-CI, I get two options:

Screen Shot 2022-01-25 at 9 17 34 AM

If I chose "Sign in with Auth0", then I go through Mozilla's SSO workflow in a new tab. The tab auto-closes, so I can't see if it calls `/login/oauth/authorize`. But maybe this is the one getting rate limited. The `POST` is done in the user's browser, so it would be associated with the user's IP (used by [express-rate-limit](https://github.com/nfriedly/express-rate-limit)).

If you chose "Sign in with credentials", you enter a Client ID and Access Token. These are turned into an authentication headers, and you do not `POST` to `/login/oauth/authorize`. The token is checked when used.

If you sign in on community-tc, the option is "Sign In with Github". In a similar way, a tab is opened, and I suspect it calls `POST` to `/login/oauth/authorize` before closing. Maybe this could be confirmed in the server logs.

So, the potential negative is that valid users will log in with SSO or GitHub, and get rate limited on the final step of the OAuth workflow. If the users are using a Proxy, this is more likely to happen. I think our corporate VPN or other VPNs will look like a proxy, so employees are more likely to see this issue. There are settings to [trust one or more levels of proxy](https://github.com/nfriedly/express-rate-limit#troubleshooting-proxy-issues) to make this less likely. Still, 200 in one minute is a lot!

This mitigation will rate limit requests pretending to be the final steps of an OAuth workflow. Or valid final steps of an OAuth workflow. It does not rate limit:

* Guessing the access tokens for static client IDs
* Guessing OAuth credentials

In other applications, I would add a report of 429 rate limits seen in a day, to see if this worked at all (or negatively impacted actual customs). Maybe we add that as a TODO item for the monitoring project.

petemoore (dismissed):
Looks great, many thanks @matt-boris!

changelog/issue-5079.md
```diff
@@ -0,0 +1,7 @@
+audience: deployers
+level: patch
+reference: issue 5079
+---
+The new `app.authRateLimitMaxRequests` configuration value allows setting the maximum number of web server requests per minute to help mitigate DoS attacks. Default is 5. Set to 0 to disable rate limiting.
+
+Another configuration value, `app.numberOfProxies` was added to set the number of proxies between the user and the server to prevent the rate limiter from acting like a global one. Default is 0. A value of 0 means that the first untrusted address would be req.socket.remoteAddress, i.e. there is no reverse proxy.
```
Eijebong, line 7: Just to clarify, that means hitting the docker container itself directly right? Might be worth changing the default to 1 in helm charts then since you'll always have an ingress in front AFAIK?

jwhitlock: Interesting idea. If the ingress acts as a proxy, then we would see all traffic from one IP, and be limiting ourselves (default case) to 5 logins a minute. This may be a reasonable default - I have no idea what the statistics are for manual login on either deployment.

I'll see if I can figure out rates and the ingress proxy question from logs.

matt-boris: I'd really love to be able to pinpoint a max value to use here for number of proxies.

On the [express-rate-limit docs](https://www.npmjs.com/package/express-rate-limit#troubleshooting-proxy-issues) they mention a way to get the number of proxies. Wasn't sure if this would be possible to try or not.

infrastructure/tooling/src/dev/taskcluster.js
```diff
@@ -51,6 +51,8 @@ module.exports = async ({ userConfig, answer, configTmpl }) => {
   // Simple defaults that can be updated by the user later
   setDefault('worker_manager.providers', {});
   setDefault('web_server.public_url', rootUrl);
+  setDefault('web_server.auth_rate_limit_max_requests', 5);
```
lotas, line 54: this feels like a very low limit, but I don't know what clients are using this to authorize, so might be good enough

changelog/issue-5079.md
jwhitlock: Sorry, I wasn't able to check the logs to see if I could figure out the proxy levels.

services/web-server/src/servers/createApp.js
```diff
@@ -134,8 +135,19 @@ module.exports = async ({ cfg, strategies, auth, monitor, db }) => {
...
+    legacyHeaders: false, // Disable the `X-RateLimit-*` headers
+  });
+
+  // set number of proxies between the user and the server
+  // to ensure rate limiter doesn't behave like a global one
+  app.set('trust proxy', cfg.app.numberOfProxies);
+
   // 1. Render a dialog asking the user to grant access
-  app.get('/login/oauth/authorize', cors(corsOptions), authorization);
+  app.get('/login/oauth/authorize', authLimiter, cors(corsOptions), authorization);
```
lotas, line 150: so this is a `GET` call which is used to start the authentication process. I don't think it makes sense to rate-limit it at all. 
This is not actively used and is probably invoked by 3rd party (like Treeherder) when they try to authorize using TC as provider

lotas:
I will close this one, commented in the ticket

PR #5084, opened 2022-01-25
Update to Go 1.17.6,  Node.js 16.13.2

### LINKED ISSUE #5070: CI taskgraph failing due to Python 2.7 running Python 3 code
**Describe the bug**

In https://github.com/taskcluster/taskcluster/pull/5069#issuecomment-1018725852, the decision task is running Python 2.7 against our Python 3-only code.

**Expected behavior**

Python 3 should be used to run the decision task

**Taskcluster version**

v44.2.2 (although Python 2.7 wasn't dropped until [v44.4.0](https://github.com/taskcluster/taskcluster/blob/main/CHANGELOG.md#developers)

**Additional context**
We're using 2-year old versions of the taskgraph code and the taskgraph/decision Docker image. An update might fix.

jwhitlock:
* Update Go 1.16.7 to 1.17.6. This makes `yarn generate` work on my M1 MacBook (arm64)
* Update Node.js from 14.17.15 to 16.13.2, the [LTS version](https://nodejs.org/en/about/releases/) until 2022-10-18.
* Add notes for how to update Go and Node.js

I've built and pushed the new Docker images based on these updates.

Github Bug/Issue: Fixes #5070


dev-docs/npn-and-go-upgrades.md
```diff
@@ -0,0 +1,45 @@
...
+development environments before deploying new versions.
+
+# Updating node
+
+Dependabot will open PRs for Node.js, but these will often fail since
+additional steps are needed.
+
+To update the node version:
+
+* Install the new node version, such as `nvm install 16.13.2`
```
matt-boris, line 14: `nvm install --lts` will also install the latest LTS, wonder if that's worth mentioning here? Up to you.

matt-boris (commented):
Looking good! Just some small failing tasks. Left a couple small comments.

I believe we can now get rid of the `--insecure` flag passed to `curl` command in the docker-worker release script https://github.com/taskcluster/taskcluster/blob/cde6847a3ec187d8dde77ae41df747ea9cd25a6b/workers/docker-worker/release.sh#L37-L54

dev-docs/npn-and-go-upgrades.md
```diff
@@ -0,0 +1,45 @@
...
+
+To update the go version:
+
+* Install the new go version, such as `gvm install go1.17.6`
+* Update `.go-version`
+* Run `yarn generate`
+* Create a changelog file and commit
+* Update the package hashes in `workers/generic-worker/gw-decision-task/tasks.yml`
+  with the hashes on the [go downloads page](https://go.dev/dl/)
+* Build and push new images (see below)
```
matt-boris, line 31: Could you please add to this section the extra steps you took such as https://github.com/taskcluster/taskcluster/pull/5084/commits/970d91d07d2e2490290ae58de00d6e728996cc5e and https://github.com/taskcluster/taskcluster/pull/5084/commits/559a9fea690a5844233592862aa03ab708bdbdc0?

jwhitlock: Yes. I'm looking for a taskcluster-approved way to run them...

jwhitlock: I can not find an easy way to run all the formatting stuff, so I added the commands from the build step.

jwhitlock:
Progress report at the end of my day:

go 1.17.6 had significant changes with `go fmt`. `yarn lint:go` checks the formatting, but fails if there are differences. I tried to extract the formatting commands from the build step.

Node.js 15.0 changed the way unhandled Promise rejects are handled. From the changelog:

> **Throw On Unhandled Rejections** - (issue reference)
>
> As of Node.js 15, the default mode for ``unhandledRejection`` is changed to ``throw`` (from ``warn``). In ``throw`` mode, if an ``unhandledRejection`` hook is not set, the ``unhandledRejection`` is raised as an uncaught exception. Users that have an ``unhandledRejection`` hook should see no change in behavior, and it’s still possible to switch modes using the ``--unhandled-rejections=mode`` process flag.

This broke a monitor library test. My reading of the test code is that it is reporting what happens in the default case, so I changed the test to current behavior, rather than try to replicate the old behavior by calling with ``--unhandled-rejections=warn`` or adding a ``unhandledRejection`` hook in the tested mode (the monitor library has settings to add this hook).

It appears that ``docker-worker-chunk-5`` is failing due to ``fs-ext`` not building. This may be because ``node-gyp`` finds Python 3.4.3 and needs >=3.6.0. I tried updating `workers/docker-worker/Dockerfile`, but that didn't seem to help and I'll probably undo that change. There is so much broken in the `docker-worker`, I find it hard to tell what are "known errors" and what are breaking errors.

djmitche:
That change to the monitor library sounds like the right choice.

> There is so much broken in the docker-worker, I find it hard to tell what are "known errors" and what are breaking errors.

I hear that!

tools/worker-runner/exit/shutdown_unix.go
```diff
@@ -1,3 +1,4 @@
+//go:build linux || darwin || freebsd
```
petemoore, line 1: Do we need both forms of the build constraint (or can we delete line 2)?

Note, this comment applies to everywhere there is the syntax change in this PR.

petemoore: Looking at [the new build constraints draft design](https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md) I think we can get away with just the new format, and delete the old lines. I'm not sure if the draft design was replaced with a final version, but I presume there should be no reason to keep both forms.

petemoore: Ah, reading the Compatibility section at the bottom of the page, I see it is intended that both forms are preserved, to enable backward compatibility (presumably so people can continue compiling with an older go version, or importing code as a library). Since we require the newer go version, and this isn't library code that others import, I think it should be safe to remove the lines with the older syntax.

jwhitlock:
I'm going to split this up in some new PRs. I think the order is:

* Go 1.17.6 update with new developer notes - fixes #5070 for us M1 / arm64 users
* Update the CI Docker images - At least [worker-ci](https://github.com/taskcluster/taskcluster/blob/3eb6ffa544879bc64c537b9f9058a59f38b14648/infrastructure/docker-images/worker-ci/Dockerfile#L1) is on ``ubuntu:14.04``. I think an upgrade to 20.04 is a good idea in general, and will be supported until 2025. It looks like the newer versions of `node-gyp` require Python 3.6, and this may get us closer.
* Update to Node.js 16.13.2. This should be smoother once the first two are done.

This may be more of a week-long project than a one-day project, but I think it is possible. Feel free to merge any npm package updates, they will not block the effort until we start on the Node.js update.

jwhitlock:
The next batch of work has been in other PRs. PR #5088 updated to Go 1.17.6, which also included:
* Adding the developer upgrade notes
* Removing the older build constraint format.

It has been nice to run `yarn generate` on my new laptop, and seems to run smoothly in CI.

I'm making progress on #5095, which includes:
* Updating to Node 16.13.2
* Removing the temp fix for Node 14.17.5
* Updating a test for node 1.15.0 behavior
* Updating `worker-ci` to Ubuntu 20.04. This image is tagged with the Node version, so it makes sense to me to update it in the same PR. All the `docker-worker-chunk-X` tests pass, except for one, which reproduces in the https://github.com/taskcluster/docker-exec-websocket-client library.

I think that's everything from code review on this PR, so I'm going to close this one.

PR #5090, opened 2022-01-27
Issue 5081: Lock OS thread on Windows when deleting a profile

### LINKED ISSUE #5081: generic-worker stuck deleting a user profile
**Describe the bug**
A generic-worker hung, trying to delete a user profile.

Last line of log:

```
122022/01/24 18:17:59Z Making system call DeleteProfileW with args: [C0004B6000 0 0]
```

A reboot was issued, and then logs continued:

```
2022/01/25 09:58:22Z Making system call GetProfilesDirectoryW with args: [C000139380 C00037A248]
```

Each syscall is logged before being issued, and then the results of the syscall are logged. Since the log message that the syscall will be issued is shown, and the log message of the result did not appear, we can see the syscall hung.

Nothing in the UI to suggest there was an issue (such as a dialogue window reporting an error).

The code that didn't complete is https://github.com/taskcluster/taskcluster/blob/v44.2.2/workers/generic-worker/win32/win32_windows.go#L808-L812

Under the hood this calls https://github.com/golang/go/blob/go1.16.7/src/syscall/dll_windows.go#L335-L340

**Expected behavior**
We should probably have a timeout. And if cleanup of old users fails, that probably shouldn't be a critical failure and we should purge old users on a best-effort basis.

**Taskcluster version**
44.2.2

 - OS: Windows (unknown version)

**Additional context**

* Reported by @Eijebong for taskcluster deployment https://divvun-tc.thetc.se
* @petemoore has full logs, if needed

petemoore:
Github Bug/Issue: Fixes #5081 (if we are lucky)

We don't have a good way to test this fix for this intermittent issue. Creating a draft PR for now so that at least we'll generate binaries, and @Eijebong can test the updated binary on his system to see if the issue disappears. This is at least useful for determining when this fix isn't appropriate, in case the issue occurs again, then this fix wasn't helpful.

The [DeleteProfileW](https://docs.microsoft.com/en-us/windows/win32/api/userenv/nf-userenv-deleteprofilew) system call is an expensive one, and may take considerable time to complete.

My suspicion is, that if in the intervening time of the DeleteProfileW call running, the calling goroutine is parked, when it is awoken, it may be assigned to a different underlying system thread. Since the system call is passed a pointer to a memory address of the original thread, perhaps the kernel ends up writing data to the data/bss section of the original thread, corrupting the memory. The new thread is not aware that the operation is completed, and hangs forever waiting to be updated.


Eijebong:
For the record, running this binary and got the deadlock again

petemoore:
Some more things we can try here:

* check if the undocumented `//go:uintptrescapes` syntax has changed ([context](https://github.com/golang/go/issues/23045#issuecomment-350275695)) that we use [here](https://github.com/taskcluster/taskcluster/blob/a401198287c6eb46f9d2b695fd0db02254bcf5d3/workers/generic-worker/win32/wrap_dll_windows.go#L30)
* check if we are missing [`runtime.KeepAlive(x interface{})`](https://pkg.go.dev/runtime#KeepAlive) calls
* try not calling [(*windows.LazyProc).Call](https://pkg.go.dev/golang.org/x/sys/windows#LazyProc.Call) method but calling [syscall.Syscall](https://pkg.go.dev/syscall#Syscall) directly
* when setting up a minimal test to reproduce the issue, have a go routine pass a byte down a channel to another goroutine every few milliseconds to try to force thread switch during profile deletion
* double check the new hang @Eijebong experienced happens at the same syscall ([userenv.DeleteProfileW](https://docs.microsoft.com/en-us/windows/win32/api/userenv/nf-userenv-deleteprofilew)) (since we make quite a lot of different calls into win32)
* make the syscalls from a C/C++/rust program, and execute that program in its own process, rather than via [FFI](https://en.wikipedia.org/wiki/Foreign_function_interface)
* review the latest [golang.org/x/sys package](https://pkg.go.dev/golang.org/x/sys/windows) to see how the standard library calls into win32, to see if there are differences, or if maybe there is already a DeleteProfile syscall already defined

Potentially related issues:
* https://github.com/golang/go/issues/34684
* https://github.com/golang/go/issues/34810
* https://github.com/golang/go/issues/42680

Eijebong:
> double check the new hang @Eijebong experienced happens at the same syscall (userenv.DeleteProfileW) (since we make quite a lot of different calls into win32)

Yup, it was. Logs were stopped at the same place and the stacktrace from delve was exactly the same

petemoore:
> Some more things we can try here:
> 
>     * check if the undocumented `//go:uintptrescapes` syntax has changed ([context](https://github.com/golang/go/issues/23045#issuecomment-350275695)) that we use [here](https://github.com/taskcluster/taskcluster/blob/a401198287c6eb46f9d2b695fd0db02254bcf5d3/workers/generic-worker/win32/wrap_dll_windows.go#L30)

wasn't that - syntax is still same in go 17

>     * check if we are missing [`runtime.KeepAlive(x interface{})`](https://pkg.go.dev/runtime#KeepAlive) calls

I've added some in df607f4df947f0db1410dda2fec08991bf25d73b above - let's see if that helps

>     * try not calling [(*windows.LazyProc).Call](https://pkg.go.dev/golang.org/x/sys/windows#LazyProc.Call) method but calling [syscall.Syscall](https://pkg.go.dev/syscall#Syscall) directly

Trying that in 3556c2a948f827184c8342fd5f07600b701740c1 above - let's see if it helps

>     * when setting up a minimal test to reproduce the issue, have a go routine pass a byte down a channel to another goroutine every few milliseconds to try to force thread switch during profile deletion

Tried that in https://github.com/petemoore/gohang/commit/9231aae46265b421676298da79cb5519857fff9e but wasn't able to reproduce the issue.

>     * double check the new hang @Eijebong experienced happens at the same syscall ([userenv.DeleteProfileW](https://docs.microsoft.com/en-us/windows/win32/api/userenv/nf-userenv-deleteprofilew)) (since we make quite a lot of different calls into win32)

confirmed above by @Eijebong 

>     * make the syscalls from a C/C++/rust program, and execute that program in its own process, rather than via [FFI](https://en.wikipedia.org/wiki/Foreign_function_interface)
> 
>     * review the latest [golang.org/x/sys package](https://pkg.go.dev/golang.org/x/sys/windows) to see how the standard library calls into win32, to see if there are differences, or if maybe there is already a DeleteProfile syscall already defined
> 
> 
> Potentially related issues:
> 
>     * [proposal: unsafe: clarify unsafe.Pointer rules for package syscall golang/go#34684](https://github.com/golang/go/issues/34684)
> 
>     * [runtime: manual instrumentation of KeepAlive is fraught golang/go#34810](https://github.com/golang/go/issues/34810)
> 
>     * [x/sys/windows: clarify safety of Proc.Call, LazyProc.Call golang/go#42680](https://github.com/golang/go/issues/42680)

petemoore:
Closing this issue in favour of #5100 which offers an alternative solution.

PR #5137, opened 2022-02-02
Additional Dependabot Config

matt-boris:
* Change Dependabot schedule from `daily` to `monthly`
* Automerge semver-patch version upgrades via GitHub Actions Workflow (thanks @grahamalama!)

### DIFF
--- a/.github/workflows/dependabot-automerge.yml
+++ b/.github/workflows/dependabot-automerge.yml
@@ -1,0 +1,23 @@
+name: Dependabot auto-merge
+on: pull_request
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1.1.1
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Enable auto-merge for Dependabot semver-patch PRs
+        if: ${{steps.metadata.outputs.update-type == 'version-update:semver-patch'}}
+        run: gh pr merge --auto --merge "$PR_URL"
+        env:
+          PR_URL: ${{github.event.pull_request.html_url}}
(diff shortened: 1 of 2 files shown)


.github/workflows/dependabot-automerge.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1.1.1
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
```
petemoore, line 17: what privileges does this token have?

matt-boris: This is what's suggested by GitHub - https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/automating-dependabot-with-github-actions#enable-auto-merge-on-a-pull-request

matt-boris: @petemoore I've made all checks required now. Refresh this PR to see the required badge next to all the tasks that run.

matt-boris: I _believe_ this should respect our setting ` Require branches to be up to date before merging` we have enabled.

petemoore (approved):

petemoore:
I see there are just a couple of checks that might not be required - but I'm not sure if they are important. They look like netlify checks - do you know if they would be affected by dependabot updates and catch potential issues? If so, we should probably have a follow up to add them, if not, no change needed.

petemoore:
* Header rules - taskcluster-web
* Mixed content - taskcluster-web 
* Pages changed - taskcluster-web
* Redirect rules - taskcluster-web

matt-boris:
@petemoore I've gone ahead and added the netlify ones as well. I was initially under the impression that since _some_ PRs end these with status `Deploy canceled`, that it would fail the required check. After enabling, I see that's not the case! Similar reason to the `Dependabot auto-merge / dependabot` check and it behaves the same as the netlify ones too.

PR #5095, opened 2022-01-28
Update Node.js from 14.17.15 to 16.13.2 LTS

jwhitlock:
Update Node.js from 14.17.15 to 16.13.2, the [LTS version](https://nodejs.org/en/about/releases/) until 2022-10-18.

This was originally part of PR #5084, where it failed on the ``docker-worker-chunk-X`` steps. These (mostly) pass after updating the `worker-ci` image from Ubuntu 14.04 to 20.04, which also includes:

* Updating from Docker 18.06.3 to 20.10.12
* Implementing a new method for registering the Docker package repo
* Switch from GZip to the XZ-compressed Node.js package

PR #5154, opened 2022-02-08
Update to docker-exec-websocket-* 3.0.0

jwhitlock:
Update `docker-exec-websocket-client` and `docker-exec-websocket-server`. These should work under Node v14, and are required for Node v16. They fix an issue when the `docker exec` command is reading and writing data
to the container.

This is part of PR #5095. The tests are failing under Node 16.13.2, and I'm curious if they also fail under the current Node 14.17.5.

### DIFF
--- a/changelog/bdp4Ifc0SGe81snujuhMrQ.md
+++ b/changelog/bdp4Ifc0SGe81snujuhMrQ.md
@@ -1,0 +1,11 @@
+audience: general
+level: minor
+---
+This release updates the `docker-worker-websocket-client` and
+`docker-worker-websocket-server` libraries, used by `docker-worker` to execute
+commands inside a running container. These updates fix a bug when reading and
+writing data to the process in the container, which may have been broken since
+2015, and be a part of why VNC was broken
+(see [issue 3542](https://github.com/taskcluster/taskcluster/issues/3542#issuecomment-746934147)).
+This change required for Node v16, and may affect tasks that use this library like the
+[interactive feature](https://docs.taskcluster.net/docs/reference/workers/docker-worker/features#feature-interactive).
--- a/ui/package.json
+++ b/ui/package.json
@@ -42,7 +42,7 @@
     "deep-sort-object": "^1.0.2",
     "deepmerge": "^4.0.0",
     "dexie": "^3.0.0",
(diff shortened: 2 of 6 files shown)


jwhitlock:
The test was:

https://github.com/taskcluster/taskcluster/blob/949fcf0a14ce280ec01cbebc6984310693a5bd2f/workers/docker-worker/test/integration/interactive_test.js#L261

[.compare()](https://nodejs.org/api/buffer.html#bufcomparetarget-targetstart-targetend-sourcestart-sourceend) is similar to ``compare`` from the C library, and returns 0 if equal, -1 if the left side should be sorted first, and 1 if the right side should be sorted first. I believe the intent is to check that `buf` (a megabyte of random data) is equal to `data` (concatenated buffer of the output of piping that random data to `cat`), so this test was confirming that the feature was _not_ working.

The code was changed to use `.compare()` in 2015, to add something call "noVNC" support: https://github.com/taskcluster/docker-worker/pull/185. Previously, it checked the input was the same as the output, byte by byte, so it is very possible this change broke the feature. This might be related to VNC not working, which was dropped in https://github.com/taskcluster/taskcluster/pull/4121.

I've switched to use [.equals()](https://nodejs.org/api/buffer.html#bufequalsotherbuffer), which does what you'd expect.

jwhitlock:
I've been staring at that test code for at least two weeks, and didn't notice `compare` until I added some debug lines and saw everything else was equal. Node in 2015 was a different beast, `.compare` and `.equals()` weren't added until v0.11.13.

I looked for other uses of `.compare` in the JS code, and just found this one, which correctly asserts the return is zero:

https://github.com/taskcluster/taskcluster/blob/949fcf0a14ce280ec01cbebc6984310693a5bd2f/workers/docker-worker/test/integration/docker_save_test.js#L87-L88

Still, `.equals` would be less awkward...

matt-boris (approved):

djmitche:
> so this test was confirming that the feature was not working

Aren't surprises fun?

lotas (commented):
Nice!

djmitche:
I doubt it's worth the effort to get something that complex working for docker-worker :)

PR #5159, opened 2022-02-09
Run yarn upgrade

matt-boris:
It appears some package upgrades were only changed in the `package.json` file and not also updated in the `yarn.lock` file, which is what's used by dependabot for scanning. I ran `yarn upgrade` on all of our `yarn.lock` files aside from the one in the `/ui` directory (will handle in another PR due to breaking changes).

This should help us clean up some open dependabot alerts.


petemoore (approved):
WCPGW?!

PR #5162, opened 2022-02-10
Auto-merge semver-minor depandabot updates (previously only semver-pa…

petemoore:
Let's auto-merge dependabot minor updates too. Previously we only auto-merged patch updates.


lotas (approved):
lgtm

PR #5194, opened 2022-02-16
Fix for #5193: incosistent UI deps

### LINKED ISSUE #5193: CI needs to run/test `yarn build`
**Describe the bug**

When releasing v44.6.0, we found a couple (at least) bugs in the build of the docker image due to us not running this task at PR creation time.

**To Reproduce**

1. Run `yarn build`

**Expected behavior**

Image can be built successfully. We need to be doing this on each PR so we know when things are breaking before we go to release.

**Taskcluster version**

v44.6.0

lotas:
Inconsistent npm dependencies

Github Bug/Issue: Fixes #5193

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,7 +62,7 @@ RUN rm -rf ui/node_modules ui/src
 # build the final image
 
 FROM node:16.14.0-alpine as image
-RUN apk --no-cache --update add nginx bash
+RUN apk --no-cache add --update  nginx bash
 COPY --from=build /base/app /app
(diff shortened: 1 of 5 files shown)


matt-boris (approved):
Awesome fix, thanks Yarik! Two birds with one stone here 👏🏻 🎉

lotas:
> Awesome fix, thanks Yarik! Two birds with one stone here 👏🏻 🎉

Yeah, that's very good that we discovered this issue, and could find a good fix, instead of reverting 👍

PR #5195, opened 2022-02-17
View Task page improvements

### LINKED ISSUE #4983: ViewTask UI doesn't handle similar named actions well
I recently added a new action called `rerun-pr`, which is very similar in name to the existing `rerun` action. Depending on the situation, one or the other will be available as an action. The UI handles this fine for the case where `rerun` is enabled but `rerun-pr` is not, for example: https://firefox-ci-tc.services.mozilla.com/tasks/A0MP43PSTSCtPkChsME-XQ

But it is not handled well for the reverse. In these cases, there's two issues:
1) The `rerun-pr` action is shown, but with the wrong icon (a hammer instead of the "restart" icon)
2) The `rerun` action is _still_ shown, even though it's not present in `actions.json`

I poked at this a bit. The icon part is easily fixable by tweaking https://github.com/taskcluster/taskcluster/blob/5020df14256764a50d3b4d3f4e621c3ffda65f4a/ui/src/views/Tasks/ViewTask/index.jsx#L761

I'm pretty sure that https://github.com/taskcluster/taskcluster/blob/5020df14256764a50d3b4d3f4e621c3ffda65f4a/ui/src/views/Tasks/ViewTask/index.jsx#L941 is the cause of the second issue, although I wasn't able figure out why that block is being added despite `rerun` not being a key in `actionData`.

lotas:
- Improved snapshot testing with react-testing-library (instead of Enzyme).
- Fixes UI menu for the task actions: correct icon for tasks named `^(rerun|retrigger)`. Fixes #4983
- Improves ViewTask page: show all details without "Show more/show less".
- Shows artifacts by default, if there's less than 10 of them.
- Moves Runs navigation on top of the card to take fixed position when switching between runs

Working demo: https://yarik.taskcluster-dev.net/tasks/Bs64AD3QSbm1Pm0MOgaJZQ

image


ui/src/components/SpeedDial/index.jsx
```diff
@@ -19,12 +19,25 @@ const styles = withStyles(theme => ({
...
-    setOpen(false);
+  function handleClose(evt) {
+    if (evt.type === 'click') {
+      setOpen(false);
+    } else {
+      if (timeout) {
+        clearTimeout(timeout);
+      }
+
+      timeout = setTimeout(() => setOpen(false), 2000);
```
lotas, line 32: This is some little tweak to allow popup menu stay for some time without closing immediately

ui/src/components/AuthConsent/__snapshots__/index.test.jsx.snap
```diff
@@ -1,54 +1,334 @@
...
-      }
-    }
-    onExpirationChange={[MockFunction]}
-    onInputChange={[MockFunction]}
-    onScopesChange={[MockFunction]}
-    registeredClientId="registeredClientId"
-    transactionID="transactionID"
-  />
-</Router>
+<DocumentFragment>
```
lotas, line 4: Turned out enzyme snapshots / serializing was useless. Switched to react-testing-library to have the actual html tree here

ui/src/components/TaskRunsCard/index.jsx
```diff
@@ -337,10 +329,34 @@ export default class TaskRunsCard extends Component {
         <div>
           <CardContent
             classes={{
-              root: classNames(classes.cardContent, {
-                [classes.collapsedCard]: !showMore && run,
-              }),
+              root: classNames(classes.cardContent),
             }}>
+            <MobileStepper
```
lotas, line 334: Moved stepper to the top of the card to improve navigation. Details are shown in full, so in order to move to next/previous run, navigation is displayed in the fixed position on top

matt-boris (approved):

Issue #3362, opened 2020-08-06
Support badges when using the checks API


djmitche:
Currently, badges only consult the table maintained for tasks using the status API.  Let's support badges with the checks API, too.

lotas:
Implemented in #5209

PR #5209, opened 2022-02-19
Add support for checks in the github badge API

### LINKED ISSUE #5139: Github badges don't work when using `reporting: checks-v1`
The current API for it is using the status API. Would be nice to support both

https://github.com/taskcluster/taskcluster/blob/7e7aa2d7b9a026cee78a3c970244f91d638c45a1/services/github/src/api.js#L394

Eijebong:
When taskcluster uses the `reporting: checks-v1` it doesn't create a
global github status anymore but a check per task. The taskcluster
github badge API used to only check for said status which would thus not
be there and return "No data yet".

This commit adds support for the github service to request the checks
associated with the reference we're requesting a badge for. I chose to
combine those results with a status in case it's also present. I don't
think this can happen in real situations but it doesn't cost anything.

Since checks are per task instead of task group, here are the rules for
the badge state:

- If any check is a failure, the state is marked as a failure.
- If no checks have a failure but there's at least one that pending,
  state is marked as pending
- Otherwise all checks are green, mark the state as success.
- If no checks are present, this keeps the same behavior as before.

A quick note on implementation: We are using the app ID to find which
check is coming from the taskcluster bot because that's what github
provides in its checks API. That differs from the status filtering which
is using the user id of the github app.

Fixes #5139


djmitche (approved):
lgtm!

lotas:
great!
The ui test is failing because dates were not fixed in tests, will fix this on Monday in a separate branch

Eijebong:
Fixes #3362 too

lotas (approved):
:+1:

PR #5210, opened 2022-02-20
Make the whole row clickable in the secrets table

### LINKED ISSUE #5027: Clicking on a secret row only works if you click on the text part of it
Same as #4962 but on the secrets page

Eijebong:
Fixes #5027


lotas (approved):
Awesome! Thanks

PR #5211, opened 2022-02-20
Fix UI tests by mocking the date in the TaskDetailsCard snapshot test

Eijebong:
There's an `X days ago` in there that would break all the time
otherwise.


lotas:
Oh! Thanks @Eijebong

I missed your PR and added new also #5212 :) With only difference that I set this date mock for all tests.
I think it's fine to merge both

lotas (approved):

Issue #5179, opened 2022-02-11
The roles page takes a long time to render when there are a few hundred roles


Eijebong:
UI side, the graphql API responds way faster than the page takes to render. Example here: https://divvun-tc.thetc.se/auth/roles

lotas:
I've discovered that material UI table is causing those slow renders. Also a problem for filtering that I've introduced in #5218

lotas:
I've spent some time debugging the issue. Problem there that Roles page (and most of the others) are using `1000` as default page size to render all elements. 
Rendering of each single row takes time, because of dozens of components wrapped around it.

I see two approaches how we can solve this:

1. Reduce markup to the bare minimum (For example list of `280` roles renders around `4700` dom elements. This can be simplified to have twice as fewer components). This is still slowish, but feels faster than it was.
2. Change the way how `ConnectionDataTable` component renders such tables. Currently it renders all rows at once. We can apply `react-virtualized`/`react-window` List wrappers, to render only what's inside the viewport. Rendering of such pages will be instant

Second approach will require slightly more refactoring, as it will touch most of the pages where lists/tables are used. 
Another downside would be that you cannot use native browser Find in Page functionality (as not all rows are rendered). Considering we have filtering on the tables themselves, I think we can sacrifice this functionality in order to make pages load faster and be more responsive.

lotas:
@Eijebong can you please remind how big was your Roles table? Community has around 280 roles so it is somewhat bearable. Iirc you had around 600? 
I think the easiest fix is to set page size to ~100, or just implement custom rendering

Eijebong:
Well I'm hitting pagination limits so more than 1000

lotas:
Fix was merge. It will now should render only rows visible on the page. Layout was simplified a bit also (removed link icon, as the whole row is clickable anyway)

Eijebong:
Will test as soon as a release contains the patch and let you know ;)

Please remind me in two weeks that I haven't tested it because there's a 99.9% change I will forget though.

lotas:
Woo-hoo! Great, now we need to go through the rest of the tables with grids and possibly do the same.. 
Do you have any other pages with hundreds of rows?

PR #5218, opened 2022-02-23
Add Copy URL  button for artifacts, rows filtering in tables

### LINKED ISSUE #5181: Copying artifact links in task UI should be simple
**Describe the bug**
It's difficult to copy an artifact link (rather than downloading / opening it) from the task UI.

**To Reproduce**

1. Open a task in the UI, e.g: https://firefox-ci-tc.services.mozilla.com/tasks/HkcTlRWqSwGV9x1Oy9lOfQ#artifacts
2. Try to copy the artifact link
3. The only way to do so is to right click and use the browser's context menu (e.g "Copy Link Address" in Firefox).
4. This URL includes the `bewit` parameter which means if you send it to someone else, they won't be able to open it
5. So you first need to paste the URL into somewhere editable and manually remove the query parameter

**Expected behavior**

Copying the artifact URL should be easy.

**Additional Context**

In Releng we often just want to link people to the artifact. E.g, linking to the signed build in the task above.

lotas:
Resolves #5181

- [x] Adds "Copy" button to copy artifact url without `bewit` parameter
- [x] Adds "Filter rows" functionality to "long tables"

image

image

There's a performance issue with the tables itself, which makes filtering slower. Will be addressed in #5179

### DIFF
--- a/ui/.neutrinorc.js
+++ b/ui/.neutrinorc.js
@@ -61,9 +61,11 @@ module.exports = {
         proxy: {
           '/login': {
             target: 'http://localhost:3050',
+            changeOrigin: true,
           },
           '/graphql': {
             target: 'http://localhost:3050',
+            changeOrigin: true,
           },
           '/subscription': {
             ws: true,
--- a/changelog/issue-5181.md
+++ b/changelog/issue-5181.md
@@ -1,0 +1,8 @@
+audience: users
+level: patch
+reference: issue 5181
+---
+
+Added "Copy URL" to the artifacts table.
+
+Added filter row functionality for big tables.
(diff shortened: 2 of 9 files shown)


ui/.neutrinorc.js
```diff
@@ -60,15 +60,17 @@ module.exports = {
...
           '/graphql': {
-            target: 'http://localhost:3050',
+            target: 'https://yarik.taskcluster-dev.net',
+            changeOrigin: true,
           },
           '/subscription': {
             ws: true,
             changeOrigin: true,
-            target: 'ws://localhost:3050',
+            target: 'wss://yarik.taskcluster-dev.net',
```
lotas, line 73: oops..

matt-boris (approved):
Nice! 🎉

ahal:
Thank you!

PR #5219, opened 2022-02-24
lbheartbeat and version endpoints for all services

matt-boris:


libraries/api/src/index.js
```diff
@@ -32,6 +36,53 @@ const ping = {
...
+  title: 'Taskcluster Version',
+  category: 'Taskcluster Version',
+  description: [
+    'Respond with the JSON version object.',
+    'https://github.com/mozilla-services/Dockerflow/blob/main/docs/version_object.md',
+  ].join('\n'),
+  handler: function(_req, res) {
+    try {
+      const taskclusterVersionFile = path.resolve(REPO_ROOT, 'version.json');
+      const taskclusterVersion = fs.readFileSync(taskclusterVersionFile).toString().trim();
```
lotas, line 78: this should be done outside of the handler. `version.json` will not be changed during whole deployment, so there is no need to do a blocking read inside the handler, and pause event loop for other requests.

lotas: or at least, if this is dynamic (which is probably not) we can do an async call here to avoid blocking event loop

matt-boris: Good call, thanks @lotas

lotas (approved):

lotas (approved):

lotas:
tested on dev cluster one of the services, it returns `__version__` and `__lbheartbeat__`

PR #5220, opened 2022-02-24
pre-commit

matt-boris:
Will add more to the `.pre-commit-config.yaml` file to check changelog existence, lint, etc. Here are some built-in hooks we could use as well https://github.com/pre-commit/pre-commit-hooks.

@lotas any experience here? I know you tossed out the idea! 🙂

PR is likely easier to look at commit by commit since tons of files changed due to this addition.


lotas (approved):
👏

Eijebong (commented):
I like the idea. I have one objection though about the trailing-whitespace one. While I hate them and removing them is :ok_hand:, I think CI should check for them too if we're going this way because it's very easy to not setup pre-commit hooks and have people commit trailing whitespaces and then have the next PR modify completely unrelated lines because of that (or have 300 "fix whitespace" commits which is also bad).

lotas:
> I think CI should check for them too

thanks @Eijebong, this is a very important point. Pre-commit hooks should not contradict CI checks. 

> or have 300 "fix whitespace" commits which is also bad

I think this should be a one-off operation to delete all trailing spaces and add new lines. 

Idea behind pre-commit is that it should only check the stuff that you are committing now, and is also an optional step. If you don't run it locally, nothing bad will happen. CI will still do that work.

Eijebong:
Yup :+1: I think biting the bullet once to get those things in is fine, I would just not want it not to be enforced by CI and have random "fix whitespace" commits all the time.

lotas (approved):

ahal:
> I like the idea. I have one objection though about the trailing-whitespace one. While I hate them and removing them is 👌, I think CI should check for them too if we're going this way because it's very easy to not setup pre-commit hooks and have people commit trailing whitespaces and then have the next PR modify completely unrelated lines because of that (or have 300 "fix whitespace" commits which is also bad).

I think it's good practice to stand up a task that runs the same pre-commit checks in CI, see:
https://pre-commit.com/#usage-in-continuous-integration

We can't enforce that all contributors set up the hooks, so we want to avoid breakage slipping in and then preventing someone else who has the hooks from committing due to something completely unrelated.

matt-boris:
We're going to be adding the https://pre-commit.ci/ github app to this repo to handle the pre-commit CI checks.

PR #5232, opened 2022-02-28
Parse application id as an int before comparing while fetching checks

Eijebong:
Turns out the helm configuration wants a string for
`github.application_id` and the checks API returns an int. This wasn't
caught because the config file doesn't enforce the type of
`credentials.appId` and I did my testing with an int, not a string


services/github/src/api.js
```diff
@@ -172,7 +172,7 @@ async function findTCChecks(github, owner, repo, branch, configuration) {
     }
     throw e;
   }
-  return checks.filter(checkObject => checkObject.app.id === configuration.github.credentials.appId);
+  return checks.filter(checkObject => checkObject.app.id === parseInt(configuration.github.credentials.appId));
```
lotas, line 175: Maybe we can add radix to the parseInt function here `parseInt(x, 10)`

This is mostly not an issue in the latest ecmascript, but just to be safe

lotas (approved):

services/github/src/api.js
Eijebong: Oh yeah, I wonder if it would take it as base 8 if the ID started with a 0, I'll fix that

lotas (approved):

Issue #5246, opened 2022-03-01
Hide pagination in the artifacts list unless there's more than 1000


Eijebong:
Right now the pagination arrows are present even if there's only one artifact, it looks a bit silly IMO. I think 1000 is the page limit

![92625813-7239-445a-b03a-31629dcc1337](https://user-images.githubusercontent.com/3650385/156143609-b31e25d5-79df-487e-9f66-c6a3e38e427f.png)

lotas:
Thanks for QA, fix coming ;)

Eijebong:
This was fixed by #5248

PR #5248, opened 2022-03-01
fix(ui): Pagination and filters shown conditionally

### LINKED ISSUE #5247: Hide the table filter when there's only one row
This looks very silly :)
92625813-7239-445a-b03a-31629dcc1337

lotas:
Pagination not shown when table doesn't have next and previous pages.
Filter is shown when there are more than 10 items in a table.

Fixes #5247 #5246

### DIFF
--- a/changelog/issue-5247.md
+++ b/changelog/issue-5247.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 5247
+---
+
+Pagination and filters shown conditionally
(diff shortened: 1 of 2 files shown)


changelog/issue-5247.md
```diff
@@ -0,0 +1,4 @@
+audience: developers
+level: silent
```
matt-boris, line 2: I feel like this should be for `users` and at level `patch`.

matt-boris (approved):
Nice, quick fix! Just one small request. I'd even be fine with you skipping running the whole CI for that change since it's just the changelog.

changelog/issue-5247.md
matt-boris: Make sure you include a little snippet in the file too for what's changed.

matt-boris (approved):

PR #5260, opened 2022-03-01
Add pre-commit badge

matt-boris:
Also, skip the `taskcluster_yml` hook on ci because it appears outgoing traffic is blocked and it cannot reach the schema to do validations.


lotas (approved):

matt-boris:
@ahal do you have experience with setting up this pre-commit ci? Looks like outgoing traffic is blocked so we're just disabling the `.taskcluster.yml` validator for now.

PR #5261, opened 2022-03-01
Fix changelog generation to not include whitespace

matt-boris:
https://github.com/taskcluster/taskcluster/commit/09391656a8e448071121b9226318903c22807315 fails pre-commit ci checks because of whitespace created in the changelog. This removes that from happening again.

### DIFF
--- a/infrastructure/tooling/src/changelog/index.js
+++ b/infrastructure/tooling/src/changelog/index.js
@@ -161,9 +161,9 @@ class ChangeLog {
     }
 
     const levelLabels = {
-      'major': '[MAJOR] ',
-      'minor': '[minor] ',
-      'patch': '[patch] ',
+      'major': '[MAJOR]',
+      'minor': '[minor]',
+      'patch': '[patch]',
     };
 
     const silent = this.snippets.filter(sn => sn.level === 'silent' && sn.reference);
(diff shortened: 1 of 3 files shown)


infrastructure/tooling/src/changelog/index.js
```diff
@@ -194,7 +194,7 @@ class ChangeLog {
         const snippets = categorizedSnippets[audience]
           .map(({ level, reference, body }) => (
             '▶ ' + levelLabels[level] +
-            (reference ? reference : '') + '\n' +
+            (reference ? ' ' + reference : '') + '\n' +
```
lotas, line 197: I think leading space was ok? to pad string to the '>' above?

infrastructure/tooling/src/changelog/index.js
```diff
@@ -161,9 +161,9 @@ class ChangeLog {
     }
 
     const levelLabels = {
-      'major': '[MAJOR] ',
-      'minor': '[minor] ',
-      'patch': '[patch] ',
+      'major': '[MAJOR]',
```
lotas, line 164: perfect! Thanks

lotas (approved):

infrastructure/tooling/src/changelog/index.js
matt-boris: This adds the space after the level in the case where there is a reference to an issue/bug

matt-boris: You can see it prints `levelLabels[level]` first. This extra space is for after that.

Issue #5242, opened 2022-02-28
Dockerflow endpoints for auth-web


matt-boris:
https://github.com/taskcluster/taskcluster/blob/main/infrastructure/k8s/templates/taskcluster-auth-deployment-web.yaml

Issue #5241, opened 2022-02-28
Dockerflow endpoints for object-web


matt-boris:
https://github.com/taskcluster/taskcluster/blob/main/infrastructure/k8s/templates/taskcluster-object-deployment-web.yaml

Issue #5240, opened 2022-02-28
Dockerflow endpoints for queue-web


matt-boris:
https://github.com/taskcluster/taskcluster/blob/main/infrastructure/k8s/templates/taskcluster-queue-deployment-web.yaml

Issue #5239, opened 2022-02-28
Dockerflow endpoints for index-web


matt-boris:
https://github.com/taskcluster/taskcluster/blob/main/infrastructure/k8s/templates/taskcluster-index-deployment-web.yaml

PR #5264, opened 2022-03-02
feat(api): add `/__heartbeat__` to each service

matt-boris:
Addresses #5234, #5236, #5237, #5238, #5239, #5240, #5241, and #5242.

> Added initial `/__heartbeat__` endpoint to all service APIs. Simply returning a 200 empty JSON object for now - implementation to follow in individual PRs per service.


services/worker-manager/src/api.js
```diff
@@ -838,3 +838,22 @@ builder.declare({
 
   return res.reply({ expires: expires.toJSON(), credentials, secret: newSecret });
 });
+
+builder.declare({
+  method: 'get',
+  route: '/__heartbeat__',
```
lotas, line 844: I think it would be good to add either unit-tests for those calls, or include those in smoke test. 
WDYT?

matt-boris: Yep, I like that. Good call, thanks!

lotas (approved):

lotas (approved):

PR #5276, opened 2022-03-09
Renew worker creds on time, even after hibernation

### LINKED ISSUE #5269: worker-runner credential renewal does not work if the system goes to sleep
After having a windows multi-user worker running idle for a while, credentials expired and were not refreshed by start-worker meaning it was stuck and not accepting tasks anymore. start-worker should refresh those credentials and restart the worker I guess.

djmitche:
Github Bug/Issue: Fixes #5269

See the issue -- I'm not sure this is worth fixing?

### DIFF
--- a/tools/worker-runner/util/nonmonotonic_test.go
+++ b/tools/worker-runner/util/nonmonotonic_test.go
@@ -16,7 +16,8 @@ func TestSleepUntilWallClockTime(t *testing.T) {
 
 func TestSleepUntilWallClockTimeCancel(t *testing.T) {
 	until := time.Now().Add(time.Second * 10)
-	ctx, _ := context.WithTimeout(context.Background(), 10*time.Millisecond)
+	ctx, cx := context.WithTimeout(context.Background(), 10*time.Millisecond)
+	defer cx()
 	require.False(t, SleepUntilWallClock(until, ctx))
 	require.Less(t, time.Now().UnixMilli(), until.UnixMilli())
 }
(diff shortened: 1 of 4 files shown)


tools/worker-runner/util/nonmonotonic_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+func TestSleepUntilWallClockTime(t *testing.T) {
+	until := time.Now().Add(time.Millisecond * 10)
+	require.True(t, SleepUntilWallClock(until, context.Background()))
+	require.GreaterOrEqual(t, time.Now().UnixMilli(), until.UnixMilli())
+}
+
+func TestSleepUntilWallClockTimeCancel(t *testing.T) {
+	until := time.Now().Add(time.Second * 10)
+	ctx, _ := context.WithTimeout(context.Background(), 10*time.Millisecond)
```
petemoore, line 19:
```
Running golangci-lint..
tools/worker-runner/util/nonmonotonic_test.go:19:7: lostcancel: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (govet)
	ctx, _ := context.WithTimeout(context.Background(), 10*time.Millisecond)
	     ^
```

petemoore (changes requested):

tools/worker-runner/util/nonmonotonic_test.go
djmitche: I really wish govet didn't run on tests, but ok :)

matt-boris (approved):

PR #5280, opened 2022-03-09
fix(ui): JS error blocked real error

### LINKED ISSUE #5277: `can't access property "length", r is undefined` on hooks page when you don't have hooks scopes
Other pages tell you which scopes you're missing. The hooks page tells you `can't access property "length", r is undefined` instead.

lotas:
"Cannot access property length of undefined" prevented real message from
being shown for unauthenticated users.

Fixes #5277

image

### DIFF
--- a/changelog/issue-5277.md
+++ b/changelog/issue-5277.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 5277
+---
+
+Fix "can't access property length of undefined" that prevented showing missing permissions error.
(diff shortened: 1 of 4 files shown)


matt-boris (approved):

djmitche (approved):
OMG TESTS!

lotas:
> OMG TESTS!

:) not very useful though. Once I'll find an easy way to mock whole auth/graphql stuff, I would improve those tests to cover different scenarios. For now this is just a dummy integration test to see if page can be rendered

PR #5286, opened 2022-03-10
Fix usage of `temporary.writeFile` in docker worker

Eijebong:
That function called was not an `async` one so awaiting had no effect.
If anyone used that function and was unlucky, they would end up with an
empty tmp file instead of one containing the string.

### DIFF
--- a/workers/docker-worker/src/upload_to_s3.js
+++ b/workers/docker-worker/src/upload_to_s3.js
@@ -50,7 +50,7 @@ module.exports = async function uploadToS3 (
     // write the source out to a temporary file so that it can be
     // re-read into the request repeatedly
     if (typeof source === 'string') {
-      await tmp.writeFile(source);
+      tmp.writeFileSync(source);
       hash.update(source);
     } else {
       let stream = fs.createWriteStream(tmp.path);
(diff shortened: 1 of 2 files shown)


workers/docker-worker/src/upload_to_s3.js
```diff
@@ -50,7 +50,7 @@ module.exports = async function uploadToS3 (
     // write the source out to a temporary file so that it can be
     // re-read into the request repeatedly
     if (typeof source === 'string') {
-      await tmp.writeFile(source);
```
lotas, line 53: hmm.. this is generally a bad idea to use sync functions as it blocks whole eventloop. But I see that this `temporary` library does not handle async writes properly.

I guess for now we can go with this solution, but we should aim to rewrite this using native `fs` functions

lotas (approved):

workers/docker-worker/src/upload_to_s3.js
matt-boris: Just made https://github.com/taskcluster/taskcluster/issues/5287. Thanks for the notes @lotas!

lotas: Thank you!

lotas (approved):

PR #5290, opened 2022-03-11
Bootstrap Taskgraph from pypi in decision task

ahal:
Releng has tweaked the recommended way of bootstrapping Taskgraph from a
clone + checkout pinned revision, to Pypi and a requirements.txt.

This provides a few benefits:
1. Since Taskgraph uses SemVer, updating will provide a hint if there
   will be backwards incompatible changes.
2. Using version numbers is nicer than revisions (e.g, Taskgraph's
   CHANGELOG will be useful).
3. Provides a mechanism to pin other non-Taskgraph dependencies.

The reason I'm submitting this now is I'd like to migrate Taskgraph to a
new repository. By moving consumers over to Pypi first, the migration
can happen seamlessly without breaking anything.

Jira: RELENG-818


ahal:
I recommend reviewing this with the "Hide whitespace" option checked.

petemoore:
@ahal Apologies, I broke Windows worker type. Should be fixed now. Just testing in https://community-tc.services.mozilla.com/tasks/PJtSKGOzSxGfXavQbOHvBA

If that passes, maybe you can just repush the commit, with a different timestamp etc, so we get a new commit sha? I think the rerun/retrigger functionality is broken (bad actions.json in decision task).

lotas (approved):

ahal:
> @ahal Apologies, I broke Windows worker type. Should be fixed now. Just testing in https://community-tc.services.mozilla.com/tasks/PJtSKGOzSxGfXavQbOHvBA
> 
> If that passes, maybe you can just repush the commit, with a different timestamp etc, so we get a new commit sha? I think the rerun/retrigger functionality is broken (bad actions.json in decision task).

Re-pushed, but looks like it invalidated review.. sorry.

matt-boris (approved):
Looks great, thanks @ahal. You mention,
>Releng has tweaked the recommended way of bootstrapping Taskgraph

I was curious if there are any docs that you could include in the PR that points to this new recommendation (if there's some formal doc) so it's easy to link back to if need be?

escapewindow:
> Looks great, thanks @ahal. You mention,
> 
> > Releng has tweaked the recommended way of bootstrapping Taskgraph
> 
> I was curious if there are any docs that you could include in the PR that points to this new recommendation (if there's some formal doc) so it's easy to link back to if need be?

https://github.com/mozilla-releng/releng-rfcs/blob/main/rfcs/0037-decision-python-deps.md probably.

matt-boris:
Thanks @escapewindow!

ahal:
There's also a link in a comment of the requirements.in :)

matt-boris:
> There's also a link in a comment of the requirements.in :)

I completely missed that! Thanks again @ahal!

PR #5292, opened 2022-03-11
fix(api): fix dockerflow endpoints for web-server

matt-boris:
Still a WIP. Likely need to edit nginx config with a reverse proxy.


services/web-server/src/main.js
```diff
@@ -123,14 +124,15 @@ const load = loader(
     httpServer: {
       requires: ['app', 'schema', 'context', 'monitor'],
       setup: async ({ app, schema, context, monitor }) => {
+        const httpServer = createServer(app);
         const server = new ApolloServer({
           schema,
           context,
           formatError,
           tracing: true,
+          plugins: [ApolloServerPluginDrainHttpServer({ httpServer })],
```
matt-boris, line 133: Says this is _highly recommended_ https://github.com/apollographql/apollo-server/blob/apollo-server-core%403.6.4/docs/source/api/plugin/drain-http-server.mdx.

matt-boris, line 133: Says this is _highly recommended_ https://github.com/apollographql/apollo-server/blob/apollo-server-core%403.6.4/docs/source/api/plugin/drain-http-server.mdx.

lotas (approved):

PR #5293, opened 2022-03-15
Worker-manager exit when timeouts lead to stuck process

### LINKED ISSUE #5003: worker-manager: Timeout leads to stuck process
In firefox-ci, a timeout was detected and logged:

https://github.com/taskcluster/taskcluster/blob/beeca281a2f2b0158b3f34613c5a58e5abf64c3c/services/worker-manager/src/provisioner.js#L36-L39

Usually, this results in worker-manager coming back (the mechanism may be that the process causes the kubernetes pod to terminate and be recreated), and worked twice in the last 7 days. However, after one timeout on October 8th, ``provisioningLoopAlive`` was stuck on ``true``, leading to ``provision`` exiting early each iteration:

https://github.com/taskcluster/taskcluster/blob/beeca281a2f2b0158b3f34613c5a58e5abf64c3c/services/worker-manager/src/provisioner.js#L66-L69

**To Reproduce**

I do not know how to reproduce this.

**Expected behavior**

The ``worker-manager`` process is killed and recreated.

**Taskcluster version**

This happened in firefox-ci, 43.2.0

**Additional context**

The node documentation suggests throwing an uncaught error rather than using ``process.exit``, but mostly so that any async output is correctly emitted:

https://nodejs.org/api/process.html#process_process_exit_code

---- 

Also see: https://github.com/taskcluster/taskcluster-retrospectives/blob/master/retrospectives/2021-10-08-worker-manager-underprovisioning.md

lotas:
Force provisioner loop to exit in case of timeouts

Github Bug/Issue: Fixes #5003

### DIFF
--- a/services/worker-manager/src/provisioner.js
+++ b/services/worker-manager/src/provisioner.js
@@ -1,6 +1,7 @@
 const Iterate = require('taskcluster-lib-iterate');
 const { paginatedIterator } = require('taskcluster-lib-postgres');
 const { WorkerPool, Worker } = require('./data');
+const { ApiError } = require('./providers/provider');
 
 /**
  * Run all provisioning logic
@@ -64,8 +65,10 @@ class Provisioner {
    */
   async provision() {
     if (this.provisioningLoopAlive) {
-      this.monitor.notice('loop-interference', {});
-      return;
+      this.monitor.alert('loop-interference', {});
+      // should be treated as terminal error
+      // to let the pod to restart and avoid getting stuck in a loop
+      throw new ApiError('provision loop interference');
     }
     try {
       this.provisioningLoopAlive = true;
--- a/services/worker-manager/test/fakes/aws.js
+++ b/services/worker-manager/test/fakes/aws.js
@@ -47,7 +47,7 @@ const ec2Method = (context, method) => {
  * A fake region-specific EC2 client.  This has the SDK methods (all with a `.promise()`
  * trailer).  It also has properties:
  *
- *  - runInstancesCalls: an array of the launch config passsed to each runInstances call
+ *  - runInstancesCalls: an array of the launch config passed to each runInstances call
  *  - terminatedInstances: instance IDs for which terminateInstances has been called
  *  - instanceStatuses: values returned from describeInstanceStatus, in the form {instanceId: state}
  */
--- a/services/worker-manager/test/fakes/azure.js
+++ b/services/worker-manager/test/fakes/azure.js
@@ -36,7 +36,7 @@ class FakeAzure extends FakeCloud {
   }
 
   _reset() {
-    // managers indexed by resoruceType
+    // managers indexed by resourceType
     this._managers = {
       vm: new VMResourceManager(this, 'vm', 'azure-vm.yml'),
       // (no schema for disks as provider does not create them directly)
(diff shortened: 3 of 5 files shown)


services/worker-manager/test/fakes/aws.js
```diff
@@ -47,7 +47,7 @@ const ec2Method = (context, method) => {
  * A fake region-specific EC2 client.  This has the SDK methods (all with a `.promise()`
  * trailer).  It also has properties:
  *
- *  - runInstancesCalls: an array of the launch config passsed to each runInstances call
+ *  - runInstancesCalls: an array of the launch config passed to each runInstances call
```
matt-boris, line 50: 🙂

services/worker-manager/test/fakes/azure.js
```diff
@@ -36,7 +36,7 @@ class FakeAzure extends FakeCloud {
   }
 
   _reset() {
-    // managers indexed by resoruceType
+    // managers indexed by resourceType
```
matt-boris, line 39: Love the clean up!

services/worker-manager/test/provisioner_test.js
```diff
@@ -358,6 +358,39 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
+      }).create(helper.db);
+      const provisioner = await helper.load('provisioner');
+      try {
+        await Promise.all([
+          provisioner.provision(),
+          provisioner.provision(),
+        ]);
+      } catch (err) {
+        if (err.message !== 'provision loop interference') {
+          throw err;
```
matt-boris, line 383: I've always used something like `assert.fail(err)` instead of throwing the `err`.

As long as throwing works the same and fails the test properly, I'm fine with what you have here.

services/worker-manager/src/provisioner.js
```diff
@@ -64,8 +65,10 @@ class Provisioner {
    */
   async provision() {
     if (this.provisioningLoopAlive) {
-      this.monitor.notice('loop-interference', {});
-      return;
+      this.monitor.warning('loop-interference', {});
+      // should be treated as terminal error
+      // to let the pod to restart and avoid getting stuck in a loop
+      throw new ApiError('provision loop interference');
```
matt-boris, line 71: I haven't looked yet, but I wonder if there are string constants for errors that we could use here (or you could make a new one and use it here if that _is_ the practice in the repo). If constants for error messages aren't used, I'd call this good! Just wanna be consistent.

matt-boris (dismissed):
Nice! Just left a couple small comments.

services/worker-manager/src/provisioner.js
lotas: I like the idea, but I haven't seen much constants either. In other places it was also like this. 
It is different for monitor though, where you need to register proper format.

services/worker-manager/test/provisioner_test.js
lotas: ah yeah, it should be something like `assert.throws()` can be changed. Also used same test suite for guidance :)

matt-boris: `assert.throws()` is asserting that some error will be thrown in the call. I don't think that's what you want here since you're just wanting the test itself to fail if the error isn't `provision loop interference`, so I think you need `assert.fail(err)` here.

lotas: I think I've figured it finally out. I am testing if the error did happen. So here `assert.rejects` is working exactly as it should. Please check updated code

matt-boris (dismissed):
LGTM!

matt-boris (approved):

Issue #5217, opened 2022-02-22
reduce number of clicks to get to the log


escapewindow:
From the [community meeting notes](https://docs.google.com/document/d/1dCQScxwH04VInIntxesYVQ8te1IvLTZXWYrHNgi1Irc/edit):

[ahal] especially from Github
- [x] Link from checks UI
- [ ] Tail of the log
  - Idea: load last N lines through web-server
  - Idea: Show last N lines of logs on failed tasks in the task definition page

Personally, I almost never want the log UI, and almost always want the raw log. Having to click through to the log UI and then to the raw log is extra clicks I'd prefer to skip.

escapewindow:
For the Github UI portion, IIRC we're able to add some text to each of the tasks in the checks view? Adding a log tail to that view may be a solution before we even get to the TC UI.

ahal:
Yes, I looked into the checks API awhile back for this very purpose. We can add arbitrary [markdown formatted text](https://docs.github.com/en/rest/reference/checks#output-object). So a log tail + direct link to the log would be very helpful.

lotas:
todo: check if there is a 64kb limit for a check run description

matt-boris:
Introduced last 250 lines of `live.log` in GH checks UI in [v44.20.0](https://github.com/taskcluster/taskcluster/releases/tag/v44.20.0).

PR #5298, opened 2022-03-16
feat(github): get to logs quicker through GitHub checks UI

matt-boris:
Github checks UI link for logs https://github.com/taskcluster/taskcluster/issues/5217

### DIFF
--- a/services/github/src/handlers.js
+++ b/services/github/src/handlers.js
@@ -7,7 +7,7 @@ const yaml = require('js-yaml');
 const assert = require('assert');
 const { consume } = require('taskcluster-lib-pulse');
 const { UNIQUE_VIOLATION } = require('taskcluster-lib-postgres');
-const { CONCLUSIONS, CHECKRUN_TEXT, CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME, CUSTOM_CHECKRUN_ANNOTATIONS_ARTIFACT_NAME } = require('./constants');
+const { CONCLUSIONS, CHECKLOGS_TEXT, CHECKRUN_TEXT, CUSTOM_CHECKRUN_TEXT_ARTIFACT_NAME, CUSTOM_CHECKRUN_ANNOTATIONS_ARTIFACT_NAME } = require('./constants');
 const utils = require('./utils');
 
 /**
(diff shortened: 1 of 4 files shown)


matt-boris:
Failing tasks were successful on rerun:
[`library-validate` rerun](https://community-tc.services.mozilla.com/tasks/B-GPQonQScaGkTGH4BSZMA)
[`client-rust` rerun](https://community-tc.services.mozilla.com/tasks/QRCUFrquRgqDNnZ0eIQLjw)

services/github/src/handlers.js
```diff
@@ -620,7 +622,7 @@ async function statusHandler(message) {
         output: {
           title: `${this.context.cfg.app.statusContext} (${event_type.split('.')[0]})`,
           summary: `${taskDefinition.metadata.description}`,
-          text: `[${CHECKRUN_TEXT}](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n${customCheckRunText || ''}`,
+          text: `[${CHECKRUN_TEXT}](${taskUI(this.context.cfg.taskcluster.rootUrl, taskGroupId, taskId)})\n[${CHECKLOGS_TEXT}](${taskLogUI(this.context.cfg.taskcluster.rootUrl, taskId)})\n${customCheckRunText || ''}`,
```
djmitche, line 625: There's a runId in scope in this function, too, that should be included in that taskLogUI function (where the `/0/` appears).

matt-boris: Ah perfect, thank you! I'll get that added tomorrow.

matt-boris: I also actually couldn't think of an instance where the run id isn't 0 in this context.

I figured every new push to a PR would start an entirely new task graph.

Is it in the case where once we fix the rerunning of tasks, on rerun, an individual task's run id is incremented?

lotas: I wonder if we can already tail the log in here as well? 
Or at this point file does not exist yet?

lotas (dismissed):

services/github/src/handlers.js
djmitche: A rerun can occur either due to a transient TC error (`worker-expired` for example) or due to a `queue.rerunTask` API call, so it's quite possible, even if unusual.

matt-boris: @djmitche would you consider it safe to keep it hardcoded or 0 (or even remove the additional link to the logs since they wouldn't exist) in the `taskDefinedHandler`?

https://github.com/taskcluster/taskcluster/blob/7888f56b64c86be4b36efb6b8d2ca4c21143c2d4/services/github/src/handlers.js#L945-L953

djmitche: Hm, good question.  When the task is initially defined, run 0 may not even exist (if the task has unresolved dependencies).  And it won't have a log until it starts.  So that link would 404 for a while, but would begin working before the other handler gets it.  So, maybe it makes sense to include the link with run 0 in that case.

lotas (approved):

matt-boris:
Issues stemming from https://www.githubstatus.com/incidents/sksd097hm0y5.

lotas (dismissed):

PR #5335, opened 2022-04-01
Document exchanges for worker-manager

djmitche:
Tested manually

image

### DIFF
--- a/changelog/N3fQy9ZuTx2jk0GOAJBiMA.md
+++ b/changelog/N3fQy9ZuTx2jk0GOAJBiMA.md
@@ -1,0 +1,4 @@
+audience: general
+level: patch
+---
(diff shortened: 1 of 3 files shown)


matt-boris (approved):
LGTM, thank you! Just one note is your commit comes up as `Unverified`. Wonder what GitHub doesn't like about your signature..

Screen Shot 2022-04-01 at 11 35 26 AM

Eijebong:
@matt-boris from https://github.com/djmitche 

> NOTE: if you see "Unverified" next to my commits, that's because they're signed with my [SSH key](https://github.com/djmitche.keys), and GitHub does not yet support SSH signatures.

petemoore (approved):
Awesome, thanks!!

djmitche:
(I have a pending support case open with GitHub for that..)

Issue #4942, opened 2021-08-04
Provide worker counts and capacity by state for worker pools


jwhitlock:
A worker manager pool provides "Current Capacity", which is the sum of the related worker capacities when the state is not `stopped`. This is calculated in the database function [get_worker_pool_with_capacity](https://github.com/taskcluster/taskcluster/blob/b239801344df21f733e3361991a072f533e23cc4/db/versions/0013.yml#L13), so it should be reasonably efficient. A worker can handle a single task at a time (``capacity=1``) or multiple tasks.

`stopped` is one of [four possible worker states](https://github.com/taskcluster/taskcluster/blob/66941613e4242d69dd2aff3fb560359eb633d59c/services/worker-manager/src/data.js#L481-L486):

* ``requested`` - A new worker is being provisioned, is not yet running
* ``running`` - A worker is ready to run tasks, and may be running tasks
* ``stopping`` - A worker is being shutdown and related resources released
* ``stopped`` - A worker has been shutdown

Operations is often interested in counts of workers by state. @davehouse uses the Taskcluster APIs to fetch all related workers for statically-created pools, counts by state, and feeds the data to Graphana to visualize worker pools over time. M. Cornmesser is debugging issues with an Azure provisioner in [bug 1723789](https://bugzilla.mozilla.org/show_bug.cgi?id=1723789#c4), and a breakdown by worker state may help diagnose delays backlogs of pending tasks and other provisioning issues. Getting these statistics requires [iterating over the workers by pool](https://github.com/jwhitlock/tc-scripts/blob/main/worker_pool_stats.py), which transfers a great deal of data over many requests to get a few aggregated numbers.

This suggests new data returned by DB function `get_worker_pool_with_capacity` and related functions, or their replacements:

* Worker count in `requested` state
* Worker count in `running` state
* Worker count in `stopping` state

These related stats could be included for completeness:

* Worker count in `stopped` state
* Total worker count
* Worker capacity in `requested` state
* Worker capacity in `running` state
* Worker capacity in `stopping` state
* Worker capacity in `stopped` state

The new data could be present in the [worker manager API calls](https://firefox-ci-tc.services.mozilla.com/docs/reference/core/worker-manager/api):

* ``workerPool`` - Get details of a worker pool
* ``listWorkerPools`` - Paginated list of all worker pools

[...]

PR #5340, opened 2022-04-04
feat(db): provide worker counts and capacities by state for worker pools

matt-boris:
> Addresses #4942. Add `get_worker_pool_with_capacity_and_counts_by_state`, `get_worker_pools_with_capacity_and_counts_by_state`, and `update_worker_pool_with_capacity_and_counts_by_state` functions to get worker counts and capacity by state for worker pools.
Deprecates `get_worker_pool_with_capacity`, `get_worker_pools_with_capacity`, and `update_worker_pool_with_capacity`.


db/versions/0070.yml
```diff
@@ -0,0 +1,248 @@
+version: 70
```
matt-boris, line 1: May be helpful to compare to https://github.com/taskcluster/taskcluster/blob/main/db/versions/0013.yml.

db/versions/0070.yml
```diff
@@ -0,0 +1,248 @@
...
+          worker_pools.previous_provider_ids,
+          worker_pools.description,
+          worker_pools.config,
+          worker_pools.created,
+          worker_pools.last_modified,
+          worker_pools.owner,
+          worker_pools.email_on_error,
+          worker_pools.provider_data,
+          coalesce((
+            select sum(workers.capacity) from workers where
```
lotas, line 32: Ok, so in this query, according to `Explain` there will be 1 main query + 9 sub-queries (although very fast ones, because they will be filtering by primary key)
image


This can be rewritten to just have one join and calculation inside using `CASE WHEN`

```sql
        select
          worker_pools.worker_pool_id,
          worker_pools.provider_id,
          worker_pools.previous_provider_ids,
          worker_pools.description,
          worker_pools.config,
          worker_pools.created,
          worker_pools.last_modified,
          worker_pools.owner,
          worker_pools.email_on_error,
          worker_pools.provider_data,
          coalesce(sum(CASE WHEN workers.state != 'stopped' THEN workers.capacity ELSE 0 END))::integer as non_stopped_capacity,          
          coalesce(sum(CASE WHEN workers.state = 'running' THEN workers.capacity ELSE 0 END))::integer as running_capacity,
          coalesce(sum(CASE WHEN workers.state = 'requested' THEN workers.capacity ELSE 0 END))::integer as requested_capacity,
          coalesce(sum(CASE WHEN workers.state = 'stopping' THEN workers.capacity ELSE 0 END))::integer as stopping_capacity,
          coalesce(sum(CASE WHEN workers.state = 'stopped' THEN workers.capacity ELSE 0 END))::integer as requested_capacity,
          coalesce(count(CASE WHEN workers.state = 'requested' THEN 1 ELSE 0 END))::integer as requested_workers,
          coalesce(count(CASE WHEN workers.state = 'running' THEN 1 ELSE 0 END))::integer as running_workers,
          coalesce(count(CASE WHEN workers.state = 'stopping' THEN 1 ELSE 0 END))::integer as stopping_workers,
          coalesce(count(CASE WHEN workers.state = 'stopped' THEN 1 ELSE 0 END))::integer as stopped_workers
        from worker_pools
        left join workers ON workers.worker_pool_id = worker_pools.worker_pool_id
        group by worker_pools.worker_pool_id
        order by worker_pools.worker_pool_id
```

This results in two sequential scans
image

And it feels like doing second one is ~30-50% faster on community tc database (probably because query needs to scan all rows anyways)

lotas: So it looks like single join is performing better and I think the other query should also be rewritten :+1:

matt-boris: Updated with joins! https://github.com/taskcluster/taskcluster/pull/5340/files#diff-44414ecd761296d11428d72a4701ed1db030571c35858b90e4ac65f48d1ff7a2

Thanks for the help @lotas 👍🏻

lotas: Do you also want to update `update_worker_pool_with_capacity_and_counts_by_state`, it still appears with sub-queries in the diff?

matt-boris: I started to do so, but got `aggregate functions are not allowed in RETURNING`. I can continue to dig to see if I can figure it out, but also figured at the same time, this function will be way less used. What do you think?

db/test/fns/worker_manager_test.js
```diff
@@ -43,12 +43,26 @@ suite(testing.suiteName(), function() {
...
+      assert(wp.stopping_capacity !== undefined);
+      assert(wp.stopped_capacity !== undefined);
+      delete wp.requested_count;
+      delete wp.running_count;
+      delete wp.stopping_count;
+      delete wp.stopped_count;
+      delete wp.requested_capacity;
+      delete wp.running_capacity;
+      delete wp.stopping_capacity;
+      delete wp.stopped_capacity;
```
petemoore, line 63: I see the original test only checked if the capacity was `!== undefined` rather than something more restrictive - is there value in the test being that the value is a positive integer, for example? I'm not entirely sure how the test framework works, and whether that adds any value or not (perhaps the db schema definition is sufficient to ensure that it is an integer, if it isn't defined), but on initial glance the check looks like it might be a little permissive.

services/worker-manager/schemas/v1/worker-pool-full.yml
```diff
@@ -62,6 +62,46 @@ properties:
     type: integer
     minimum: 0
     description: Total capacity available across all workers for this worker pool that are currently not "stopped"
+  requestedCount:
+    title: Requested Worker Count
+    type: integer
+    minimum: 0
+    description: Total woker count in "requested" state for this worker pool
```
petemoore, line 69: s/woker/worker/

Are the various worker states defined somewhere that we can link to? I think this could be useful context for the docs, or a few words about e.g. what the requested state represents. I don't know if we already have such a document, but ideally we'd have a reference doc with a finite state diagram showing how the workers transition through the states, together with a description of what each state means, and then we'd reference that in these descriptions, so the reader knows what the different states mean. If we don't already have something like that, I think just having a few words about what the state means would also have value, as I realise writing a new doc is a bit of scope creep, and we could do that in a different (follow-up) PR.

petemoore (dismissed):
This looks pretty good to me, and it looks like you've got your head around the db migration process etc. My SQL isn't good enough to know if the queries can't be simplified/optimised a bit.

@djmitche Do you have any thoughts on this?

db/versions/0070.yml
```diff
@@ -0,0 +1,180 @@
...
+          worker_pools.created,
+          worker_pools.last_modified,
+          worker_pools.owner,
+          worker_pools.email_on_error,
+          existing.provider_id as previous_provider_id,
+          coalesce((
+            select sum(workers.capacity) from workers where
+              workers.worker_pool_id = worker_pool_id_in and
+              workers.state != 'stopped'),
+            0)::integer,
```
djmitche, line 139: Hm, this worries me a little with respect to performance -- is it performing 8 additional index scans of the workers table?

djmitche (commented):
This looks nicely implemented.  I saw some comments earlier about explain/analyze of the get_ queries, and that looked OK.

I'd be very nervous about doing joins between this table and the tasks table, though! (mentioned since it came up in the issue)

db/versions/0070.yml
lotas: Afaik there are some difficulties to add a join and grouping in the update return statement.
We can also solve this on the api level, this function is only being used in `POST /worker-pool/:id` method, so we can drop that sub-query here and trigger previously defined new function to get those stats. 

I'm not sure how Postgres internally will handle transaction if we'd introduce join in that update query. Would it need to lock whole `workers` table?

lotas:
@djmitche I'm just curious about this approach in general, as I haven't seen it yet. It looks like business logic is implemented fully inside stored procedures. So clients are becoming very "dumb" and only care about stored procedures api, not table internals. 
Can you please confirm that this was the main intent for doing it this way, or briefly describe if there's more to it? :)

lotas (dismissed):

db/versions/0070.yml
djmitche: That is a good question -- adding load with more scans is bad, but locking a table is _very_ bad!  I don't think it would lock the table, but it's good to be sure!

djmitche:
Yes, that's precisely the intent.  It appears more clearly in e.g., queue methods where we take a task from one state to another.  It's less interesting for the CRUD methods.

My impression is that a lot of "enterprise" software is built this way, but it does seem to be unusual in OSS.

I think the design decision that might be more regrettable is that REST API functions tend to return the same schema, even for update operations.  When that schema contains information that  is derived, and that derivation is expensive, that becomes a problem.

lotas (approved):

petemoore (approved):

PR #5351, opened 2022-04-12
fix(graphql): update schema for new counts/capacities per state

matt-boris:
GraphQL was missing schema for the new columns returned by the API.


lotas (approved):
nice!

PR #5354, opened 2022-04-13
Tweak description of task dependents

petemoore:
Minor wording change for description of dependent tasks. Now I've submitted it, I'm not sure it is any better, so it isn't preferred, I'm ok with closing and not merging. My feeling was that the term "blocks" might make it sound like this task prevents the other tasks from being scheduled.


matt-boris (approved):
Definitely like this wording better, thanks!

Issue #4366, opened 2021-02-05
Track last date active for queue workers, too


djmitche:
We track last-seen dates for task queues, but not for workers -- just a first-claim and an expiration.

That leads to confusion such as https://bugzilla.mozilla.org/show_bug.cgi?id=1690950 where there's no useful data to indicate that some of those workers are no longer doing anything.

Let's add a `last_date_active` field to workers as well, and update that on claimTask, claimWork and reclaimTask.

This should have the same kind of client-side rate-limiting that the `task_queues.last_date_active` does, but with a shorter timeframe (perhaps 30 minutes?), so that any `last_date_active` less than 30 minutes ago should be considered "current".

Displaying this in the UI is a separate project.

PR #5365, opened 2022-04-19
feat(db): add `last_date_active` to `queue_workers`

matt-boris:
Address #4366.

> Add `last_date_active` column to `queue_workers` table. Add `queue_worker_seen_with_last_date_active`, `quarantine_queue_worker_with_last_date_active`, `get_queue_worker_tqid_with_last_date_active`, and `get_queue_workers_tqid_with_last_date_active` functions for this new column.
>
> Deprecates `quarantine_queue_worker`, `get_queue_worker_tqid`, `get_queue_workers_tqid`, and `queue_worker_seen`.


services/queue/schemas/v1/taskqueue-response.yml
```diff
@@ -29,8 +29,8 @@ properties:
   lastDateActive:
     title:        "Task queue Last Date Active"
     description: |
-      Date of the last time this task queue was seen active. `lastDateActive` is updated every 6 hours
-      but may be off by up-to 6 hours. Nonetheless, `lastDateActive` is a good indicator
+      Date of the last time this task queue was seen active. `lastDateActive` is updated every half hour
+      but may be off by up-to half an hour. Nonetheless, `lastDateActive` is a good indicator
```
petemoore, line 33: This is a problem with the original comment, rather than your changes, but maybe we can fix it at the same time. I'm not sure what the returned date is, but I suspect it is one of the following:

* Most recent time a worker legally called queue.claimWork for this task queue
* Most recent time a worker legally called queue.claimWork or queue.reclaimTask or queue.resolveTask or queue.createArtifact / .....
* Most recent time a worker legally called queue.XXXXXXX or a worker legally called workerManager.registerWorker / workerManager.XXXXX

In other words, I think the description makes it hard to guess which date it really returns, which means when there is a problem, it is harder to see where the problem is. With a bit more clarification it should help aid troubleshooting when there are problems.

services/queue/src/workerinfo.js
```diff
@@ -10,7 +10,7 @@ class WorkerInfo {
 
     // update `expires` values in postgres at this frequency; larger values give less accurate
     // expires times, but reduce database traffic.
-    this.updateFrequency = '6 hours';
+    this.updateFrequency = '30 min';
```
djmitche, line 13: It might be a good idea to make this a deployment config option, so it can be changed without making a new release or rolling back.

djmitche (commented):
This looks good!  I agree with Pete that being explicit about what "seen" means would be good.

db/versions/0072.yml
```diff
@@ -0,0 +1,176 @@
...
+  get_queue_workers_tqid:
+    deprecated: true
+  queue_worker_seen:
+    description: |-
+      Recognize that a worker has been seen by the queue, creating it if necessary.  This is called
+      when workers claim or re-claim work.  The expiration time is not allowed to move backward.
+
+      This function always writes to the DB, so calls should be suitably rate-limited at the
+      client side.
+    deprecated: true
```
lotas, line 25: as I understand it, if we deprecate a function, it should not be changed, so all other props are not needed here

matt-boris: So since we guarantee all DB functions must work for 2 major TC versions, this needed to be updated. Check out the points under the first checkbox [here](https://github.com/taskcluster/taskcluster/tree/main/db#checklist).

lotas: I was thinking more of [this](https://github.com/taskcluster/taskcluster/tree/main/db#stored-functions). Basically signature changes (new return column), so name should also be changed as I understand this

matt-boris: Ah ok I see. So not having a re-implementation of this DB func fails tests and would fail if anyone uses it because the new column has a not-null constraint.

So for example, if I changed this to just `deprecated: true`, the output of `yarn workspace taskcluster-db test` is:
```bash
  463 passing (1m)
  1 failing

  1) db/test/fns/queue_test.js
       queue_workers
         queue_worker_seen (deprecated):
     error: null value in column "last_date_active" violates not-null constraint
DETAIL: Failing row contains (wg, wi, [], 2012-04-21 13:21:16.684619+00, 2022-04-22 13:21:16.684+00, 2022-04-21 13:21:16.684619+00, prov/wt, null).
WHERE: SQL statement "insert
    into queue_workers (task_queue_id, worker_group, worker_id, quarantine_until, expires, first_claim, recent_tasks)
    values (
      task_queue_id_in,
      worker_group_in,
      worker_id_in,
      now() - interval '10 years',
      expires_in,
      now(),
      jsonb_build_array()
    )
    on conflict (task_queue_id, worker_group, worker_id) do update
    set
      expires = greatest(coalesce(expires_in, queue_workers.expires), queue_workers.expires)
    where
      queue_workers.task_queue_id = task_queue_id_in and
      queue_workers.worker_group = worker_group_in and
      queue_workers.worker_id = worker_id_in"
PL/pgSQL function queue_worker_seen(text,text,text,timestamp with time zone) line 2 at SQL statement
CODE: 23502
```

lotas: Oh, that probably means that upgrade script will also fail, since all of the existing records do not have default value, and field itself is marked as `not null`

db/versions/0072.yml
```diff
@@ -0,0 +1,176 @@
+version: 72
+description: add last_date_active to queue_workers
+migrationScript: |
+  begin
+    alter table queue_workers add column last_date_active timestamp with time zone not null;
```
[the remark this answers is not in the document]
matt-boris: Nice catch, thanks! So, now the question is - what do we default this column as? Or just make it nullable?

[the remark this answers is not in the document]
lotas: I think it's best to make it default null. 
Otherwise I'm not even sure which date to use there, we'd have to create a temporary table and fill it in with some dates from other tables? (`workers.created`)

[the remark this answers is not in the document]
matt-boris: Yeah I'm on board with just making it null. @petemoore @djmitche you guys have thoughts?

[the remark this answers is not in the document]
djmitche: The downside of NULL is that, after the migration, you won't see that value again.  So all the conditionals to handle a null there will get very little exercise.  But, you're right that setting it to anything else makes the migration substantially more complex.

So, null makes sense, but maybe call it out as a possibility in the sql function definitions and in the API schema.

services/queue/schemas/v1/worker-response.yml
```diff
@@ -55,6 +55,14 @@ properties:
       Date of the first time this worker claimed a task.
     type:         string
     format:       date-time
+  lastDateActive:
+    title:        "Worker Last Date Active"
+    description: |
+      Date of the last time this worker was seen active. `lastDateActive` is updated every half hour
+      but may be off by up-to half an hour. Nonetheless, `lastDateActive` is a good indicator
+      of when the worker was last seen active.
```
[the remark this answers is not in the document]
matt-boris: Whatcha think @petemoore? https://github.com/taskcluster/taskcluster/pull/5365/commits/7d95ce9f06bf54b9d764e7d42108da39d6e2252d

lotas (approved):
nice!

PR #5366, opened 2022-04-19
fixed bug5313 in the UI

### LINKED ISSUE #5313: Save filters in worker-manager UI
**Describe the bug**
The worker-manager UI loses its filter when you click "View Errors" / "View Workers" then go back again. This is a minor annoyance :)

**To Reproduce**

1. Go to 'https://firefox-ci-tc.services.mozilla.com/worker-manager'
2. Apply a filter (e.g `gcp`)
3. Click "View Errors" for one of the provisioniers
4. Go back
5. The GCP filter is lost and now you need to type it out again so you can inspect the next one. 

**Expected behavior**
Would be nice if the filter were saved somehow (url parameter?) so I didn't have to re-type everything. My work around is to just always open the View Workers / View Errors pages in a new tab (but sometimes I forget).

Come to think of it, encoding the filter in the url would also make sharing links to specific workers better.

**Taskcluster version**
44.8.1

abdelquodr:
Fixed the worker-manager filter value loss. saving the filter value so that you don't always need to retype(repeat yourself) every time.

@ahal, @lotas, and @djmitche kindly check and give feedback/suggestions.

thanks

Fixes  #5313

### DIFF
--- a/ui/.neutrinorc.js
+++ b/ui/.neutrinorc.js
@@ -197,5 +197,3 @@ module.exports = {
     }),
   ],
 };
-
-
--- a/ui/.neutrinorc.js
+++ b/ui/.neutrinorc.js
@@ -60,20 +60,20 @@ module.exports = {
         },
         proxy: {
           '/login': {
-            target: 'https://community-tc.services.mozilla.com/',
+            target: 'http://localhost:3050',
             changeOrigin: true,
           },
           '/graphql': {
-            target: 'https://community-tc.services.mozilla.com/',
+            target: 'http://localhost:3050',
             changeOrigin: true,
           },
           '/subscription': {
             ws: true,
             changeOrigin: true,
-            target: 'https://community-tc.services.mozilla.com/',
+            target: 'ws://localhost:3050',
           },
           '/api/web-server': {
-            target: 'https://community-tc.services.mozilla.com/',
+            target: 'http://localhost:3050',
             changeOrigin: true,
           },
         },
--- a/ui/.neutrinorc.js
+++ b/ui/.neutrinorc.js
@@ -197,3 +197,5 @@ module.exports = {
     }),
   ],
 };
+
+
--- a/ui/.neutrinorc.js
+++ b/ui/.neutrinorc.js
@@ -60,20 +60,20 @@ module.exports = {
         },
         proxy: {
           '/login': {
-            target: 'http://localhost:3050',
+            target: 'https://community-tc.services.mozilla.com/',
             changeOrigin: true,
           },
           '/graphql': {
-            target: 'http://localhost:3050',
+            target: 'https://community-tc.services.mozilla.com/',
             changeOrigin: true,
           },
           '/subscription': {
             ws: true,
             changeOrigin: true,
-            target: 'ws://localhost:3050',
+            target: 'https://community-tc.services.mozilla.com/',
           },
           '/api/web-server': {
-            target: 'http://localhost:3050',
+            target: 'https://community-tc.services.mozilla.com/',
             changeOrigin: true,
           },
         },
--- a/ui/src/components/ConnectionDataTable/index.jsx
+++ b/ui/src/components/ConnectionDataTable/index.jsx
@@ -152,6 +152,12 @@ export default class ConnectionDataTable extends Component {
 
   pages = new Map();
 
+  componentDidMount() {
+    this.setState({
+      filterValue: this.props.history.location.state || this.state.filterValue,
+    });
+  }
+
   componentDidUpdate(prevProps) {
     if (
       !this.props.connection.pageInfo.previousCursor &&
@@ -209,6 +215,9 @@ export default class ConnectionDataTable extends Component {
   };
 
   handleFilterValueChange = e => {
+    const { history } = this.props;
+
+    history.replace(this.state, e.target.value);
     this.setState({ filterValue: e.target.value });
   };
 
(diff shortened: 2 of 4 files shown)


ui/.neutrinorc.js
```diff
@@ -60,20 +60,20 @@ module.exports = {
...
             ws: true,
             changeOrigin: true,
-            target: 'ws://localhost:3050',
+            target: 'https://community-tc.services.mozilla.com/',
           },
           '/api/web-server': {
-            target: 'http://localhost:3050',
+            target: 'https://community-tc.services.mozilla.com/',
             changeOrigin: true,
           },
```
djmitche, line 78: I suspect these changes were made during development?  You'll want to be careful not to check such changes in.  But, these changes shouldn't be needed if you follow the [Hacking the UI](https://github.com/taskcluster/taskcluster/blob/main/dev-docs/development-process.md#hacking-on-the-ui) steps.

djmitche, line 78: I suspect these changes were made during development?  You'll want to be careful not to check such changes in..

djmitche (changes requested):
It's best to describe the changes in a PR directly, rather than referring to a bug number for context.  Then someone looking at the PR, or at the commits in Git, has the context necessary to know what they are seeing.  So, please update the PR description and commit description to summarize what the commit does.  Also, you'll need to add a changelog -- see the advice in https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/changelog.md for how to write that.

I don't actually have comments on the React changes you've made -- I'm pretty rusty on that part of the codebase!

ui/src/components/WMWorkerPoolsTable/index.jsx
```diff
@@ -305,6 +305,7 @@ export default class WorkerManagerWorkerPoolsTable extends Component {
             String(workerPool.providerId).includes(filterValue) ||
             String(workerPool.owner).includes(filterValue)
           }
+          {...this.props}
```
djmitche, line 308: It looks like this component just needs the `history` prop -- perhaps better to only pass that single property along?

abdelquodr: ok, I'll change it.
Thanks

ui/.neutrinorc.js
```diff
@@ -60,20 +60,20 @@ module.exports = {
...
           '/subscription': {
             ws: true,
             changeOrigin: true,
-            target: 'ws://localhost:3050',
+            target: 'https://community-tc.services.mozilla.com/',
           },
           '/api/web-server': {
-            target: 'http://localhost:3050',
+            target: 'https://community-tc.services.mozilla.com/',
```
lotas, line 76: As @djmitche pointed out, it's better not to change those lines here. I only suggested adding those to skip running other services locally. 

We would definitely address it in a separate issue. I think we can support `TASKCLUSTER_ROOT_URL` variable here, but it will go in a new issue.

abdelquodr: Okay. I understannd @lotas 
Thanks

ui/src/components/ConnectionDataTable/index.jsx
```diff
@@ -209,6 +215,9 @@ export default class ConnectionDataTable extends Component {
   };
 
   handleFilterValueChange = e => {
+    const { history } = this.props;
```
lotas, line 218: I think we can add `history` to the `propTypes`, as we are using it here. Not sure if we have linting rules covering this.

lotas:
Thanks @abdelquodr, looks good. 
If you could also change back urls in the proxy config, and change PR title to match the name of the issue ticket that would be awesome. Thank!

ui/src/components/ConnectionDataTable/index.jsx
abdelquodr: Okay

ui/.neutrinorc.js
lotas: Added new ticket, linking here for the records https://github.com/taskcluster/taskcluster/issues/5373

abdelquodr:
@lotas please recheck the PR for merge@lotas please recheck the PR for merge

lotas:
> @lotas please recheck the PR for merge

Checking it locally
When I open one of the provisioners page like `http://localhost:5080/provisioners/proj-fuzzing`  I get this error. This means that some components do not have router passed to them properly. Would you be able to debug this?

image

Thanks!

lotas:
One more request, as it turned out, the original page from #5313 is `http://localhost:5080/worker-manager` which doesn't use this `ConnectionDataTable` component, and requires similar implementation

Thanks!

abdelquodr:
I’m sorry, my system develop a fault. I’m glad for the opportunity and will
like to continue working on the issue. Thanks for reaching out to me

Best regards

petemoore:
Hi @abdelquodr 
I hope all is well with you. Just checking in, to see if you are still actively working on this. Many thanks!

abdelquodr:
Hello Pete Moore,

thanks for reaching out. I just fixed my system back, and I am active to
continue contributing to the source code.

Thank you

lotas:
@abdelquodr please let us know if you would still like to finish this ticket. For now I will close it, if you would like to continue, just reopen or start a new PR. Thanks!

abdelquodr:
Yes, I’ll like to finish up the issue. Thanks

PR #5369, opened 2022-04-19
fix(github): use checks API to fulfill latest endpoint

### LINKED ISSUE #5364: Github `latest` endpoint always returns a 404
**Describe the bug**
According to the docs, the `latest` endpoint in the Github service is supposed to redirect to a status page:
https://docs.taskcluster.net/docs/reference/integrations/github/api#latest

This also means that the official docs for adding badges provide broken links:
https://docs.taskcluster.net/docs/reference/integrations/github#adding-status-badges-to-your-projects-readme

**To Reproduce**

Follow below links:
https://community-tc.services.mozilla.com/api/github/v1/repository/taskcluster/taskcluster/main/latest
https://firefox-ci-tc.services.mozilla.com/api/github/v1/repository/mozilla-mobile/fenix/main/latest

They hit a 404:
```
{
  "code": "ResourceNotFound",
  "message": "Status not found\n\n---\n\n* method:     latest\n* errorCode:  ResourceNotFound\n* statusCode: 404\n* time:       2022-04-18T18:05:01.189Z",
  "requestInfo": {
    "method": "latest",
    "params": {
      "owner": "mozilla-mobile",
      "repo": "fenix",
      "branch": "main"
    },
    "payload": {},
    "time": "2022-04-18T18:05:01.189Z"
  }
}
```

**Taskcluster version**
44.11.1

ahal:
The `/repository/<owner>/<repo>/<branch>/latest` endpoint relies on the
Github `statuses` API to retrieve the link to the task group of a
commit. But for some reason most repositories I've tested, seem to
return an empty list of statuses for this API, rending the latest
endpoint useless. Practically speaking, this means the badge links for
most projects are busted.

This PR tweaks the endpoint to use the Github `checks` API to build the URL
instead. Unfortunately this API does not contain the task group id
anywhere, so linking to the task group is not possible. Instead, this
links to the run id of the check in question.

Fixes #5364

### DIFF
--- a/changelog/issue-5364.md
+++ b/changelog/issue-5364.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 5364
+---
+The `github/v1/repository/<owner>/<repo>/<branch>/latest` endpoint now supports projects using `checks-v2` reporting.
--- a/services/github/src/api.js
+++ b/services/github/src/api.js
@@ -532,13 +532,26 @@ builder.declare({
 
   // Get task group ID
   if (instGithub) {
-    let status = await findTCStatus(instGithub, owner, repo, branch, this.cfg);
-
-    if (!status) {
-      return res.reportError('ResourceNotFound', 'Status not found', {});
+    // First inspect the status API. This will be set if consumer repos are
+    // using status reporting (the default).
+    const status = await findTCStatus(instGithub, owner, repo, branch, this.cfg);
+    if (status) {
+      return res.redirect(status.target_url);
     }
 
-    return res.redirect(status.target_url);
+    // Next inspect the checks API. This will be set if consumer repos have
+    // opted into the checks-v2 reporting.
+    const checks = await findTCChecks(instGithub, owner, repo, branch, this.cfg);
+
+    if (checks.length > 0) {
+      // If there are multiple checks we can't redirect to all of them, so just
+      // pick the first one in the array.
+      let check = checks[0];
+      return res.redirect(check.repository.html_url + "/runs/" + check.id);
+    }
+
+    // Otherwise there is no status available for the given branch.
+    return res.reportError('ResourceNotFound', 'Status not found', {});
   }
 });
 
--- a/services/github/test/api_test.js
+++ b/services/github/test/api_test.js
@@ -119,7 +119,7 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
       repo: 'checksRepo',
       ref: 'success',
       info: [
-        { name: "check1", conclusion: 'success', app: { id: 66666 } },
+        { id: 123, name: "check1", conclusion: 'success', app: { id: 66666 }, repository: { html_url: "https://github.com/abc123/checksRepo" } },
         { name: "check2", conclusion: 'success', app: { id: 66666 } },
         { name: "check3", conclusion: 'failure', app: { id: 12345 } },
       ],
@@ -467,7 +467,7 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
       err => err.code === 'InsufficientScopes');
   });
 
-  test('link for clickable badges', async function() {
+  test('link for clickable badges with status', async function() {
     let res;
 
     await testing.fakeauth.withAnonymousScopes(['github:latest-status:abc123:*'], async () => {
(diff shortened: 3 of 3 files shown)


ahal:
I mostly wrote this to familiarize myself with the Github service a bit. There's a good chance we'll want to WONTFIX this PR as it sort of circumvents the issue by redirecting to the Github checks UI rather than to the Taskcluster task group.

We might want to instead continue linking to the task group and figure out why many repos are returning an empty list when calling the `statuses` API.

ahal:
Another option here is to first try the `statuses` API and then link to the task group if it returns something, and then fallback to the `check` API if it doesn't.

petemoore:
Thanks for this Andrew!

I'm not too familiar with the specifics here, but from initial inspection, I _think_ the issue here is that when you create a `.taskcluster.yml` file you can essentially choose whether the integration is via the github status api or the github checks api. The choice of which api to use will affect the metadata that is available via github. The taskcluster endpoint here github metadata to return an appopriate page.

I'm not entirely sure why we provide both status and checks api integrations, perhaps one was available before the other, and we never deprecated the former approach, so essentially we support two different integrations.

My proposal would be that we have a separate endpoint for checks api metadata and one for status api. Since I think the status api was there first, we should probably leave it is for backward compatibility, e.g.

* /repository/<owner>/<repo>/<branch>/latest -> uses github status api
* /repository/<owner>/<repo>/<branch>/latest/statusapi -> same as above
* /repository/<owner>/<repo>/<branch>/latest/checksapi -> uses checks api

I don't know if it is possible to have a checks api and status api implementation in a .taskcluster.yml, but for example if it was, then you could get different metadata for each.

I suspect the better long term solution is for us to choose which api (checks or status) is better, and go with that. Or maybe we should have entirely separate /checks and /status api calls. At the moment it all looks a bit mixed together, so maybe a lot of code is trying to handle both possibilities, and it could get a bit tangled up. Having explicit separate paths may be a way to clean that up a bit, and we can always mark e.g. the status api calls as deprecated, so we can eventually move off of them in a future release, and then just drop the endpoints, and everything should continue to work.

@djmitche @matt-boris @lotas @ahal I may have misunderstood what is going on here entirely with my superficial scan of the code/docs, so curious on your thoughts.

Eijebong:
How about checking through the status API and if empty through the checks API?

Because as is, pretty sure this will break for people using status reporting instead of the checks one.

petemoore:
If we do decide against having a separate endpoint for the checks api version of this api call, is the information about whether the .taskcluster.yml uses checks api or status api internally available, so we don't have to try both, but only need to try the appropriate one?

> How about checking through the status API and if empty through the checks API?
>
> Because as is, pretty sure this will break for people using status reporting instead of the checks one.

Is it possible that a repo uses checks api for taskcluster, but has some other app integration (e.g. code coverage tool) that uses status api checks, that would cause the status api check not to be empty?

I feel like the code not _knowing_ which one to use, and then trying both could be a footgun for bugs later on. Either it should know which api to call, or if it doesn't, we should provide different endpoints, since the user should know which one it needs to call (I'm assuming here this endpoint is called by users, rather than used internally by our services - and if our own services use the api - that they should know whether to use the checks variant, or the status variant).

Eijebong:
> Is it possible that a repo uses checks api for taskcluster, but has some other app integration (e.g. code coverage tool) that uses status api checks, that would cause the status api check not to be empty?

No because both `findTCstatus` and `findTCchecks` filter the returned objects to get the taskcluster ones only.

https://github.com/taskcluster/taskcluster/blob/f4450ce48a583491761821b24857bf1154ced5f7/services/github/src/api.js#L157
https://github.com/taskcluster/taskcluster/blob/f4450ce48a583491761821b24857bf1154ced5f7/services/github/src/api.js#L175

Also having only one endpoint would match what I did for badges in #5209

ahal:
Assuming no one wants to tackle said refactor, I think our options here are:
1. First try status API, then fallback to checks API.
2. Create a second endpoint for checks API.

To my untrained mind, option 1 seems nicer to the users. But I guess whether we are planning to deprecate the `status` reporting or not should weigh in pretty heavily here.

djmitche:
> I'm not entirely sure why we provide both status and checks api integrations, perhaps one was available before the other, and we never deprecated the former approach, so essentially we support two different integrations.

Yes, GitHub replaced Status with Checks several years ago.  There was a general plan to switch and drop status support, but we never really "finished" checks (this being one of the remaining issues) or made a deprecation plan for the status API support.

ahal:
As per discussion in the Taskcluster community meeting, I'm going to implement the fallback method for now (so first query `status` endpoint, then if empty query the `checks` endpoint). That's the simplest solution and is a clear improvement over a 404.

Down the road we might still decide to split this into two separate endpoints, or more likely just drop support for `status` reporting. I'll work on this next week when I get a chance.

lotas (approved):
Looks good, thanks!

Issue #3060, opened 2020-06-15
Mix information from queue and worker-manager in UI and worker-manager


djmitche:

djmitche:
Let's have a little brainstorm here.

At the top level, we currently have "Worker Manager" and "Workers" menu items.  The "Workers" is currently structured by "provisioner" and then "workerType" and then a list of workers (notably, workerGroup is not a level, and that's probably just as well).

Instead, we really have "task queues" and "worker pools".  These are sort-of joined by name, but it's possible to have one without the other. For example, the `built-in/succeed` task queue exists on the queue, but not in worker-manager.  And worker-manager may have pools defined which have no tasks and no workers, and thus no task-queue exists.  So, the view I'm picturing has a row for each task-queue/worker-pool:

* taskQueueId = workerPoolId (linking to workers view)
* queue
  * pending tasks
  * stability
  * description
* worker-manager
  * providerId
  * current capacity
  * owner
  * link to errors
  * link to definition edit/delete

Some of that information might make more sense in a sidebar, or even in a full-page view.  Especially description, and we have long talked about adding a schema for each task queue as well.

For rows that are missing one or the other, those fields would be blank.

Information about workers also comes from both the queue and worker-manager, but in a fashion that's less visible to users.  Again we can have workers known to the queue but not worker-manager (such as `built-in/*`) and workers known to worker-manager but not the queue (such as workers that have been created but not yet claimed a task, or have otherwise failed).  Also, both services keep track of workers for some time after they have stopped doing work.  We should give a unified view of workers, but make it clear what different pieces of information mean, as that information is often crucial to debugging.

So, a similar tabular view for workers makes sense, with the following information:
 * workerGroup / workerId
 * queue
   * last seen by queue (#4366)
   * most recent task / state
   * quarantine state + button
 * worker-manager
   * state
   * capacity
   * providerId
   * terminate button

Similarly, some of this information, and perhaps more (first claim, created, list of tasks, etc.) might make sense in a sidebar.

Setting aside the UI, it's not clear how best to get all of this information from the DB, since it is spread over two services.  The data we want is basically an OUTER JOIN of the queue and worker-manager tables.  Our DB access rules allow a service to have *read* access to another service's tables, so we would have to select one service or the other to run that OUTER JOIN.  It probably makes sense to do so in worker-manager, in keeping with our standard that platform services don't know about core services.

[...]

PR #5370, opened 2022-04-20
feat(ui): show last active in queue workers view

matt-boris:
Addresses UI part of #4366.

note: as more and more columns are added to this table, I feel like the UX is worse and worse, as horizontal scrolling is then needed. We may need to address this as more columns will be added from https://github.com/taskcluster/taskcluster/issues/3060.


lotas (approved):
LGTM

Issue #3163, opened 2020-07-06
Measure and improve performance of the worker query in provisioning loop


djmitche:
The worker-manager provisioner process [seems to be running slowly](https://bugzilla.mozilla.org/show_bug.cgi?id=1650909).  Our guess is that this is due to

https://github.com/taskcluster/taskcluster/blob/master/services/worker-manager/src/provisioner.js#L145-L151

First of all, this is pulling out a bunch of rows and immediately discarding most of them (stopped).  Second, it's just getting aggregate counts, which could probably be done much more quickly in the DB.  That's a little tricky since it's also revealing the providerId's still running in each pool.

Third, we don't know if this is the issue!

So:
 * Rewrite this query to be more efficient -- minimally, let's add a method that will return all running workers, and then continue with the JS `seen()` logic.  Maximally, let's do the entire analysis in the DB.
 * Add timing logs for the various parts of the provisioning loop: maybe one for the overall loop, containing durations for the components, plus adding a "duration" property to the `this.monitor.log.workerPoolProvisioned` call for each worker pool.

helfi92:
Any idea why this would run faster on azure? We're still doing the same amount of querying.

djmitche:
Good question.  I don't think we're sure, just a hunch.  But, we had some issues before with scanning this table where its width seemed to cause the issue -- a query selecting `provider_data` was *much* slower than one that did not.  We never did figure out why that was the case, but the guess was that it's because that column is [toast](https://www.postgresql.org/docs/11/storage-toast.html) column and thus accessing it was quite a bit slower.

PR #5374, opened 2022-04-22
Worker manager extra stats

lotas:
* include total number of workers seen for a provider
* add duration for a single worker pool provisioning loop

Related to #3163

### DIFF
--- a/changelog/issue-3163.md
+++ b/changelog/issue-3163.md
@@ -1,0 +1,6 @@
+audience: worker-deployers
+level: patch
+reference: issue 3163
+---
+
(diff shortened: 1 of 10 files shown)


changelog/issue-3163.md
```diff
@@ -0,0 +1,6 @@
+audience: general
+level: silent
```
matt-boris, line 2: Don't think silent level writes out to the changelog (may be wrong), probably better as patch I think.

matt-boris (dismissed):
Nice! Looks great 🎉

matt-boris (approved):

PR #5382, opened 2022-04-26
feat(worker-manager): Immediate worker provisioning for Azure

lotas:
Further improve worker manager Azure provisioning. 

Till now, Azure was the only provider that was not actually doing any cloud api calls in the `provisioner` loop, and only created worker record in `REQUESTED` state, that was later picked by `workerScannerAzure`.
This sometimes can take 10-20 minutes, and, considering overall provisioning slowness, it often ends with worker being terminated, as they fail to start within given time (`terminateAfter`) (Leading to errors seen in #4999)

Related to #4999

### DIFF
--- a/services/worker-manager/test/provider_azure_test.js
+++ b/services/worker-manager/test/provider_azure_test.js
@@ -50,6 +50,35 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
     monitor = await helper.load('monitor');
   });
 
+  const assertProvisioningState = async (expectations) => {
+    // re-fetch the worker, since it should have been updated
+    const workers = await helper.getWorkers();
+    assert.equal(workers.length, 1);
+    const worker = workers[0];
+
+    for (let resourceType of ['ip', 'vm', 'nic']) {
+      const name = worker.providerData[resourceType].name;
+      switch (expectations[resourceType]) {
+        case 'none':
+          assert(!worker.providerData[resourceType].operation);
+          assert(!worker.providerData[resourceType].id);
+          break;
+        case 'inprogress':
+          assert.equal(worker.providerData[resourceType].operation, `op/${resourceType}/rgrp/${name}`);
+          assert(!worker.providerData[resourceType].id);
+          break;
+        case 'allocated':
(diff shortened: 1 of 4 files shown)


services/worker-manager/test/provider_azure_test.js
```diff
@@ -50,6 +50,35 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
     monitor = await helper.load('monitor');
   });
 
+  const assertProvisioningState = async (expectations) => {
```
lotas, line 53: this is moved out of one suit, because it is needed in two different

services/worker-manager/test/provider_azure_test.js
```diff
@@ -634,11 +628,46 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
       await assertProvisioningState({ ip: 'allocated', nic: 'allocated', vm: 'none' });
       assert(provider.removeWorker.called);
     });
+  });
 
-    test('successful provisioning of VM without public ip', async function() {
-      await assertProvisioningState({ ip: 'none', nic: 'none', vm: 'none' });
+  suite('provisionResources without public IP', function() {
```
lotas, line 633: extracted this test in separate case, as its success depends on initial `setup()`, worker pool needs to have proper `workerConfig` to set `skipPublicIP` properly

services/worker-manager/test/provider_azure_test.js
```diff
@@ -657,13 +686,22 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
+
     setup('create un-provisioned worker', async function() {
       const workerPool = await makeWorkerPool();
       const workerInfo = {
         existingCapacity: 0,
         requestedCapacity: 0,
       };
+
+      // prevent the worker from being immediately provisioned
+      sandbox.stub(provider, 'checkWorker').returns('ok');
```
lotas, line 699: this just ensures that internal state for fake calls are not interfered by calling `checkWorker` inside `provision()` call

matt-boris (approved):
Awesome! Can't wait to see even more improvements from this 👍🏻 🎉

Issue #5348, opened 2022-04-11
Copy artifact URL button is clipping the wrong link


ahal:
**Describe the bug**
When clicking the "Copy" button next to an artifact, the text copied to the clipboard looks similar to:
https://firefox-ci-tc.services.mozilla.comhttps//firefox-ci-tc.services.mozilla.com/api/queue/v1/task/M0BdslNvR1KAQ7F-u2or0Q/runs/1/artifacts/public%2Fbuild%2FMozillaVPN.pkg

Note that the root URL is listed twice, and the second one is missing the colon after `https`. I'm pretty sure this used to work.

**To Reproduce**

1. Go to a task
2. Click the "Copy" button next to an artifact
3. Paste it somewhere and inspect

**Taskcluster version**
44.8.4

lotas:
Interesting. Some of the artifacts are fine on that very task. Looks like text files are being handled differently, and the bug was in this url concat:

https://github.com/taskcluster/taskcluster/blob/c569cf38ada0bc4f31dc3fed7e4d681bf9adbfbc/ui/src/components/TaskRunsCard/index.jsx#L289-L291

Shubhamchinda:
@ahal I'd like to check this.

lotas:
@Shubhamchinda not sure what is the status of your fix, but I had it to fix in my branch in order to fix something else, sorry :)

78a017fa1204296adff2ca0e8b73b18b98869fd1

lotas:
Fix merged to main

Issue #5373, opened 2022-04-21
UI: Improve proxy url configuration


lotas:
At the moment to run UI you need to also run additional services (postgres, service-web).

It is also possible to run local UI against public Taskcluster.

Configuration is defined here: https://github.com/taskcluster/taskcluster/blob/8e8bc011239c1910c9007f78ae233c78489d08d9/ui/.neutrinorc.js#L61-L78

TODO: 
* add a way to use `TASKCLUSTER_ROOT_URL` to define root url for the  proxy configuration, defaulting to the `localhost:3500`

Issue #5362, opened 2022-04-14
Put the log files at the top of the list in task artifacts


Eijebong:
Currently artifacts are sorted alphabetically, however the chances of you wanting anything but logs there is pretty slim (as far as my workflow goes, maybe my workflow is too specific).

I feel like it would help if the log artifacts would always show up first in the artifacts list instead of me having to look for it.in the list. And yeah, there's a filter, but that's extra work :)

lotas:
@Eijebong I just realized that it was already fixed in the past :)

![image](https://user-images.githubusercontent.com/83861/165490356-63995706-e2ed-402e-b5b9-4d513428988d.png)

Link to the logs is a huge button on top. Do you see the same on your instance? 

I will still place them on top in the list

Eijebong:
![39dd72b7-e654-4326-8038-37d88d65791a](https://user-images.githubusercontent.com/3650385/165491147-a2098fc7-3c23-427e-97d9-f4c39f385f07.png)

It's not there if you don't have livelog enabled

lotas:
Fix merged

PR #5384, opened 2022-04-27
UI artifacts list improvements

lotas:
* fix copy URL functionality for some artifacts (#5348)
* allow to run local UI against remote Taskcluster instance (#5373)
* put log files on top of the list (#5362)

Closes bugs: #5348 #5373 #5362

image

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -19,6 +19,15 @@ http://localhost:3050, but otherwise need not be publicly accessible. The
 development server for this repo will proxy requests as necessary to
 http://localhost:3050.
 
+### Using existing Taskcluster server
+
+Instead of running web-server locally you can use existing Taskcluster installation.
+To point proxy endpoints, you need to set `TASKCLUSTER_ROOT_URL` environment variable:
+
+```bash
+TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com/ yarn start
+```
+
 ## Development
 
 To get started local development, just:
--- a/ui/src/components/TaskRunsCard/index.jsx
+++ b/ui/src/components/TaskRunsCard/index.jsx
@@ -37,6 +37,7 @@ import { withAuth } from '../../utils/Auth';
 import { getArtifactUrl } from '../../utils/getArtifactUrl';
 import splitTaskQueueId from '../../utils/splitTaskQueueId';
 import Link from '../../utils/Link';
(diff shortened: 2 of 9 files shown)


ui/src/components/TaskRunsCard/sort.js
```diff
@@ -0,0 +1,23 @@
...
+  return artifacts
+    .map(a => ({ ...a, priority: getPriority(a) }))
+    .sort((a, b) => {
+      if (a.priority === b.priority) {
+        return a.node?.name?.localeCompare(b.node?.name);
+      }
+
+      return a.priority - b.priority;
+    });
+};
```
matt-boris, line 23: Love this!

ui/README.md
```diff
@@ -19,6 +19,15 @@ http://localhost:3050, but otherwise need not be publicly accessible. The
 development server for this repo will proxy requests as necessary to
 http://localhost:3050.
 
+### Using existing Taskcluster server
```
matt-boris, line 22: 🎉

matt-boris (approved):
Awesome! LGTM

ui/src/components/TaskRunsCard/sort.js
lotas: thanks! Spend some time trying to figure out how to do it with `.sort()` only, but didn't find a good solution. Had to attach those virtual priorities

ui/src/components/TaskRunsCard/index.jsx
```diff
@@ -252,13 +253,7 @@ export default class TaskRunsCard extends Component {
 
     return {
       ...artifacts,
-      edges: [...artifacts.edges].sort((a, b) => {
-        if (a.node.isPublic === b.node.isPublic) {
```
lotas, line 256: btw, search code base didn't revel any place where this value was set. There were only some checks on the queue api side, but it was never exposed, so I wonder if it ever worked at all

djmitche: I recall this property (`isPublic`) was calculated in a weird way in a weird place, and I think I removed that.  Maybe I missed this spot.

lotas: Thanks! I decided that calculating it in the UI is enough for now, so didn't change API returned object.

PR #5399, opened 2022-04-29
fix(release): prepare `docker-worker-chunk` for `release-publish` dependency

matt-boris:
> `docker-worker-chunk-{1,2,3,4,5}` tasks would non-deterministically fail during releases due to them depending on the `release-publish` task passing successfully before they run. This will be fixed by setting a dependency on `release-publish` in the `docker-worker-chunk-{1,2,3,4,5}` tasks. For now, on releases, two task groups are created, making it impossible to define a dependency on `release-publish` in the `docker-worker-chunk-{1,2,3,4,5}` tasks. Once we combine the two task groups to one, on release, this dependency can be set to fix the issue.

### DIFF
--- a/taskcluster/ci/docker-worker/kind.yml
+++ b/taskcluster/ci/docker-worker/kind.yml
@@ -4,8 +4,12 @@ loader: taskgraph.loader.transform:loader
 transforms:
   - src.transforms:taskcluster_images
   - src.transforms:add_task_env
+  - src.transforms:docker_worker_chunk
   - taskgraph.transforms.job:transforms
   - taskgraph.transforms.task:transforms
+
+kind-dependencies:
+  - release
 
 job-defaults:
   worker-type: dw-ci
@@ -32,23 +36,7 @@ job-defaults:
       DOCKER_TESTS: '1'
 
 jobs:
-  chunk-1:
-    description: 'docker-worker test chunk #1'
+  chunk:
+    description: 'docker-worker test chunk'
     run:
-      command: ./test/docker-worker-test --this-chunk 1 --total-chunks 5
-  chunk-2:
-    description: 'docker-worker test chunk #2'
-    run:
-      command: ./test/docker-worker-test --this-chunk 2 --total-chunks 5
-  chunk-3:
-    description: 'docker-worker test chunk #3'
-    run:
-      command: ./test/docker-worker-test --this-chunk 3 --total-chunks 5
-  chunk-4:
-    description: 'docker-worker test chunk #4'
-    run:
-      command: ./test/docker-worker-test --this-chunk 4 --total-chunks 5
-  chunk-5:
-    description: 'docker-worker test chunk #5'
-    run:
-      command: ./test/docker-worker-test --this-chunk 5 --total-chunks 5
+      command: ./test/docker-worker-test
--- a/taskcluster/ci/docker-worker/kind.yml
+++ b/taskcluster/ci/docker-worker/kind.yml
@@ -38,5 +38,6 @@ job-defaults:
 jobs:
   chunk:
     description: 'docker-worker test chunk'
+    chunks: 5
     run:
       command: ./test/docker-worker-test
--- a/taskcluster/src/transforms/__init__.py
+++ b/taskcluster/src/transforms/__init__.py
@@ -1,3 +1,4 @@
+import copy
 import os
 import json
 
@@ -105,3 +106,20 @@ def parameterize_mounts(config, jobs):
                             rust_version=rust_version,
                             node_version=node_version)
         yield job
+
+
+@transforms.add
+def docker_worker_chunk(config, jobs):
+    total_chunks = 5
+    for job in jobs:
+        for chunk in range(1, total_chunks + 1):
+            c_task = copy.deepcopy(job)
+            c_task["name"] += f"-{chunk}"
+            c_task["description"] += f" #{chunk}"
+            c_task["run"]["command"] += f" --this-chunk {chunk} --total-chunks {total_chunks}"
+
+            # include release-publish dependency in the case of a release
+            if config.params["tasks_for"] == "github-push":
+                c_task["dependencies"] = dict(release="release-publish")
+
+            yield c_task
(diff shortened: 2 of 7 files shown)


lotas:
As I understood, @ahal mentioned that it could be simply done with new transformer, something like this: 
https://github.com/taskcluster/taskcluster/blob/eb5ed5c6ad3783bc70571ca1077a770808802579/taskcluster/src/job.py#L39-L61

Where we check if this is `release == release-publish` and adding this dependency.

ahal:
> As I understood, @ahal mentioned that it could be simply done with new transformer, something like this:
> 
> https://github.com/taskcluster/taskcluster/blob/eb5ed5c6ad3783bc70571ca1077a770808802579/taskcluster/src/job.py#L39-L61
> 
> Where we check if this is `release == release-publish` and adding this dependency.

Yes, not only can you add this in a `transform`, but you can split the `chunks` via transform as well. I'll add an example a bit later.

lotas (approved):
looks good to me

matt-boris:
Need to test the release dependency still. Tried locally but got this issue https://github.com/taskcluster/taskgraph/issues/30. Talking with @ahal on how best to do that.

matt-boris:
Am now second guessing that this will even work because there are two task groups that are created on a release - like this one https://github.com/taskcluster/taskcluster/runs/6230348821 - generates both https://community-tc.services.mozilla.com/tasks/groups/KMSME5jMQparn-7-8hyOdA and https://community-tc.services.mozilla.com/tasks/groups/Bcj7kXiKQY-vBkRf-vxDNA.

And I now see that the `release-publish` task isn't even in the same group as the `docker-worker-chunk` tests. So further work may be needed to combine these two task groups into one on release.

taskcluster/src/transforms/__init__.py
```diff
@@ -105,3 +106,20 @@ def parameterize_mounts(config, jobs):
...
+    total_chunks = 5
+    for job in jobs:
+        for chunk in range(1, total_chunks + 1):
+            c_task = copy.deepcopy(job)
+            c_task["name"] += f"-{chunk}"
+            c_task["description"] += f" #{chunk}"
+            c_task["run"]["command"] += f" --this-chunk {chunk} --total-chunks {total_chunks}"
+
+            # include release-publish dependency in the case of a release
+            if config.params["tasks_for"] == "github-push":
```
lotas, line 122: I guess this is not enough to determine if this is a release or not.

https://github.com/taskcluster/taskcluster/blob/6185123416f71712fc63eb69a3505bbfac4862a3/taskcluster/src/target_tasks.py#L7-L11

Or this will also be controlled by the logic above `only-on` and doesn't need to be added here. Hm.. @ahal? :)

lotas:
> Am now second guessing that this will even work because there are two task groups that are created on a release - like this one https://github.com/taskcluster/taskcluster/runs/6230348821 - generates both https://community-tc.services.mozilla.com/tasks/groups/KMSME5jMQparn-7-8hyOdA and https://community-tc.services.mozilla.com/tasks/groups/Bcj7kXiKQY-vBkRf-vxDNA.
> 
> And I now see that the `release-publish` task isn't even in the same group as the `docker-worker-chunk` tests. So further work may be needed to combine these two task groups into one on release.

Hmm.. I have two ideas:
1. I can guess that's because all tasks were filtered out on release before (see `target_tasks.py`) which would drop all non `only-on: release` tasks, and since there is a dependency on it, taskgraph decides to include all
2. this is because of the insufficient checks if this is release or not, which results in extra set of tasks: https://github.com/taskcluster/taskcluster/pull/5399#pullrequestreview-960105660

petemoore:
> Am now second guessing that this will even work because there are two task groups that are created on a release - like this one https://github.com/taskcluster/taskcluster/runs/6230348821 - generates both https://community-tc.services.mozilla.com/tasks/groups/KMSME5jMQparn-7-8hyOdA and https://community-tc.services.mozilla.com/tasks/groups/Bcj7kXiKQY-vBkRf-vxDNA.
> 
> And I now see that the `release-publish` task isn't even in the same group as the `docker-worker-chunk` tests. So further work may be needed to combine these two task groups into one on release.

Would it be possible to combine them also when it isn't a release? I like the idea of a single task group regardless, so it is easier not to overlook which tasks have run and the dependencies between them.

I wonder if some of the logic in the json-e of .taskcluster.yml could be moved into the taskgraph logic, so that the json-e is pretty straightforward, just generates a decision task, and the decision task does all the rest?

taskcluster/src/transforms/__init__.py
ahal: Yeah, you might want to create a new parameter called `release` or something that does this early on in the generation, and then that way you can re-use it throughout.

Alternatively you could fake the `tasks_for` value in your `.taskcluster.yml` to be `github-release` when a tag push is detected... but that feels hacky to me.

ahal (changes requested):
Matt and I were chatting about this in PM and the problem here is that there are actually two separate graphs being created. One generated via the main Github push event, and another for the tag push event.

Unfortunately the docker-worker tasks are part of the former, and the release task is part of the latter. So there's no way to make them depend on one another. Solving this will likely involve re-working the `.taskcluster.yml` a bit and putting some thought into which graphs are generated when, and what is part of them.

Requesting changes for now.

lotas (approved):

taskcluster/ci/docker-worker/kind.yml
```diff
@@ -4,9 +4,13 @@ loader: taskgraph.loader.transform:loader
 transforms:
   - src.transforms:taskcluster_images
   - src.transforms:add_task_env
+  - src.transforms:docker_worker_chunk
   - taskgraph.transforms.job:transforms
   - taskgraph.transforms.task:transforms
 
+kind-dependencies:
+  - release
```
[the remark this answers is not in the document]
ahal: Yeah that's fine, it just means the `release` tasks will be generated before the `docker-worker` ones.

ahal (approved):
Nice, looks good!

PR #5410, opened 2022-05-02
Fix log autoscroll behaviour

### LINKED ISSUE #5336: Sharing a log link with a line anchor doesn't work anymore
I'm guessing that https://github.com/taskcluster/taskcluster/pull/5283 broke it since you end up at the bottom of the log even when there's a line anchor in the URL.

To reproduce:

- Find a long log from a task
- Click on a line number in the middle of the log. Get teleported back up (that's also the wrong behavior)
- Copy the URL, paste it into a new tab, get teleported at the bottom

lotas:
Avoid scrolling to bottom when link contain line references

Fixes #5336

### DIFF
--- a/changelog/issue-5336.md
+++ b/changelog/issue-5336.md
@@ -1,0 +1,6 @@
+audience: worker-deployers
+level: patch
+reference: issue 5336
+---
+
+Fix scroll to bottom link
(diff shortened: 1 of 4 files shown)


matt-boris (approved):
Thanks for fixing this Yarik!! 🎉

PR #5413, opened 2022-05-03
docker-worker: Disallow unknown properties in task payload.

### LINKED ISSUE #5412: Docker worker payload allows unknown properties
**Describe the bug**
It is possible to submit a task to docker-worker containing properties that are misspelled or unknown to docker-worker, and it ignores them. It should disallow undefined properties in its payload, to avoid problems and confusion. For example, there could be features which are specific to another worker, such as generic-worker, that are included in a docker-worker payload, and docker-worker would not object, but execute the task without enabling the feature. Or for example, properties could be misspelled. 

**To Reproduce**

Submit a task to docker-worker containing additional undefined properties.

**Expected behavior**
It is much easier if there is strict validation of the payload, so only payloads that docker-worker understands are accepted.

**Taskcluster version**
44.12.3 (but it seems the problem exists ever since docker-worker was created)

petemoore:
Github Bug/Issue: Fixes #5412


lotas (approved):
great!

petemoore:
Note this was originally attempted in https://github.com/taskcluster/docker-worker/pull/409 but later rolled back.

PR #5422, opened 2022-05-04
fix(queue): add null check to `lastDateActive` in queue workers `serialize()` func

matt-boris:
> Add null check to `lastDateActive` in queue workers serialize() func.

Got this sentry issue when releasing to FxCI: https://sentry.prod.mozaws.net/operations/taskcluster-firefox-ci/issues/19477419/


lotas:
thanks!

Issue #2776, opened 2020-05-06
Show worker tasks from most-recent to least-recent


petemoore:
**Describe the bug**

The Worker historic tasks view has a default task ordering that lists the tasks from oldest to newest, but would be much nicer to reverse the order and list from newest to oldest.

See e.g. https://community-tc.services.mozilla.com/provisioners/proj-taskcluster/worker-types/gw-ci-macos/workers/proj-taskcluster/63-135-170-250

helfi92:
If you click on the "Started" column, it will re-order the list. The link will also be deep linked so it can be shared for example:

https://community-tc.services.mozilla.com/provisioners/proj-taskcluster/worker-types/gw-ci-macos/workers/proj-taskcluster/63-135-170-250?sortBy=started&sortDirection=desc

Does that help?

PR #5432, opened 2022-05-05
feat(ui): improvements to provisioners ViewWorker

matt-boris:
Addresses https://github.com/taskcluster/taskcluster/issues/2776.

> Show worker tasks from most recent to least recent. Also, link the `taskId` field to the task page.

### DIFF
--- a/ui/src/components/WorkerTable/index.jsx
+++ b/ui/src/components/WorkerTable/index.jsx
@@ -95,12 +95,22 @@ export default class WorkerTable extends Component {
     });
   };
 
+  componentDidMount() {
+    let query = parse(this.props.location.search.slice(1));
+
+    query = query.sortBy ? query : { sortBy: 'started', sortDirection: 'desc' };
+
+    this.props.history.replace({
+      search: stringify(query, { addQueryPrefix: true }),
+    });
+  }
+
   render() {
     const { classes, worker } = this.props;
     const query = parse(this.props.location.search.slice(1));
     const { sortBy, sortDirection } = query.sortBy
       ? query
-      : { sortBy: null, sortDirection: null };
+      : { sortBy: 'started', sortDirection: 'desc' };
     const iconSize = 16;
     const items = this.getTableData({ sortBy, sortDirection, worker });
     const headers = [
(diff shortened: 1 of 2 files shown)


ui/src/components/WorkerTable/index.jsx
```diff
@@ -95,12 +95,22 @@ export default class WorkerTable extends Component {
...
+    return query;
+  };
+
   render() {
     const { classes, worker } = this.props;
     const query = parse(this.props.location.search.slice(1));
     const { sortBy, sortDirection } = query.sortBy
       ? query
-      : { sortBy: null, sortDirection: null };
+      : this.setDefaultSortBy();
```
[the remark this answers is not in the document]
matt-boris: Ah perfect, yeah thank you!

ui/src/components/WorkerTable/index.jsx
```diff
@@ -95,6 +95,14 @@ export default class WorkerTable extends Component {
     });
   };
 
+  componentDidMount() {
+    const query = { sortBy: 'started', sortDirection: 'desc' };
+
+    this.props.history.replace({
```
lotas, line 101: ah, wait, no :) this will always push new query. I think you forgot to add a check, if those values are not present in query, so you can set them. 

Ugh, maybe it's even better not to put those directly in url, since those are defaults, but just replace the default on line 111 - `{ sortBy: null, sortDirection: null }` => `{ sortBy: 'started', sortDirection: 'desc' }`
and if user changes sorting, it would only then change URL. 

Sorry :)

matt-boris: So I had initially set the defaults like you mentioned here, but what that does is makes the user need to click on the `Started` column header two times to reverse the sort direction. First time they click, it'll add the query to the URL, then it'll switch on the second click. That's why I initially did it in the `render()` func. I guess I could just add the check in `componentDidMount()`??

lotas: ah.. yeah.. then this, or you can also change the link that leads to this page, so user would end up here already with those values in query string

petemoore: I'm glad you guys understand this stuff! 🤣

lotas (approved):
nice! :) Thanks

matt-boris:
I think something similar should be done with the workers view as well https://firefox-ci-tc.services.mozilla.com/provisioners/gecko-t/worker-types/win10-64-2004.

Looks to be sorted by the worker ID by default - maybe we should make this one by last active? I'll ask in #firefox-ci-proj

PR #5436, opened 2022-05-09
feat(#4999): Worker manager cloud call timeout

lotas:
Add timeout to all cloud api calls to avoid stuck processes.

Problem described in https://github.com/taskcluster/taskcluster/issues/4999#issuecomment-1118659890 

Github Issue:  #4999

### DIFF
--- a/services/worker-manager/src/main.js
+++ b/services/worker-manager/src/main.js
@@ -223,9 +223,7 @@ let load = loader({
         ownName,
         providers,
         monitor: monitor.childMonitor('worker-scanner'),
-        iterateConf: cfg.app.workerScannerIterateConfig || {
-          maxIterationTime: 9000000,
-        },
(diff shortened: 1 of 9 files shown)


services/worker-manager/src/main.js
```diff
@@ -223,9 +223,7 @@ let load = loader({
         ownName,
         providers,
         monitor: monitor.childMonitor('worker-scanner'),
-        iterateConf: cfg.app.workerScannerIterateConfig || {
-          maxIterationTime: 9000000,
-        },
+        iterateConf: cfg.app.workerScannerIterateConfig || {},
```
lotas, line 226: this is not needed anymore, process times out for a different reason

matt-boris (dismissed):
Nice! All LGTM!

matt-boris (approved):

PR #5443, opened 2022-05-12
Issues #5363, #5396, Bugs 1595217: Invalidate cache entries for files/dirs that have disappeared from host

### LINKED ISSUE #5363: Cache missing or corrupted problems repeat
**Describe the bug**
https://bugzilla.mozilla.org/show_bug.cgi?id=1595217 If a file is missing on the filesystem but still in the file-cache.json or directory-cache.json, then generic-worker attempts to copy it and panics when the filesystem move fails.

https://bugzilla.mozilla.org/show_bug.cgi?id=1539681 If a cached file/dir is corrupted, generic-worker does not run a checksum to catch this. Instead when generic-worker runs a task using the corrupted cache file(s), it moves the bad file(s) into place and allows the task to fail. Because of this, we see repeated failures on the affected worker for tasks using the corrupt cache file(s). This appears as an intermittent failure to sheriffs.

**To Reproduce**

Missing cached file
1. Remove a cached file from the filesystem.
2. Run a task that uses the cached file.
3. Notice that generic-worker fails when trying to move the missing cache into place.
4. Loop to step 2

Corrupt cached file
1. Modify a cached file or directory on the filesystem.
2. Run a task that uses the cached file.
3. Notice that generic-worker fails when running the task.
4. Loop to step 2

**Expected behavior**
Could generic-worker check if cached files exist before trying to use them and remove them from the cache record if they're missing?
Could generic-worker as part of this check run a checksum for the on-filesystem cached file(s) to verify it is not corrupted?

generic-worker does remove cached file(s) if there is a failure moving it back into the caches directory (back out from within the task_XX dir), but only if it fails to move it. This could also be a place to verify the checksum of the cache to make sure the finished task did not alter the cache contents?

**Taskcluster version**
v38
The failures have increased over the last 3-4 years. Perhaps this has increased with the increased use of the cache

**Screenshots**

 - OS: Linux, MacOS (I assume WIndows g-w also could have this happen if not already happening)

**Additional context**

### LINKED ISSUE #5396: generic-worker failure: [mounts] Not able to rename dir
**Describe the bug**
MacOS builds in mozillavpn recently started failing with this generic-worker error:
https://firefox-ci-tc.services.mozilla.com/tasks/CUSlv4wpQlCPUoyGbmWcBw/runs/0/logs/public/logs/live.log

Afaict, these tasks have not changed recently. Heitor and I have been standing up signing tasks for these builds, but the signing tasks use a different pool, so not sure how that would be related.

**Taskcluster version**
44.11.2

petemoore:
Bugzilla Bugs:
* [1595217](https://bugzilla.mozilla.org/show_bug.cgi?id=1595217)

Github Bug/Issue:
* Fixes #5363
* Fixes #5396

### DIFF
--- a/workers/generic-worker/cache_test.go
+++ b/workers/generic-worker/cache_test.go
@@ -1,0 +1,13 @@
+package main
+
+import "testing"
+
+func TestIssue5363(t *testing.T) {
+	cacheMap := &CacheMap{}
+	// This cache file has two entries with locations that do not exist on the filesystem, and three that do
+	cacheMap.LoadFromFile("testdata/testcaches.json", "fakedir")
+	// Make sure two entries were removed
+	if len(*cacheMap) != 3 {
+		t.Errorf("Was expecting 3 cache entries in testdata/testcaches.json but found %v", len(*cacheMap))
+	}
+}
--- a/workers/generic-worker/testdata/testcaches.json
+++ b/workers/generic-worker/testdata/testcaches.json
@@ -1,0 +1,17 @@
+{
+  "bad1": {
+    "location": "fakelocation"
+  },
+  "good1": {
+    "location": "."
+  },
+  "bad2": {
+    "location": "fakelocation"
+  },
+  "good2": {
+    "location": "testdata"
+  },
+  "good3": {
+    "location": "main.go"
+  }
+}


workers/generic-worker/mounts.go
```diff
@@ -144,7 +144,18 @@ func (cm *CacheMap) LoadFromFile(stateFile string, cacheDir string) {
...
-		(*cm)[i].Owner = *cm
+		// Github Issues: #5363, #5396
+		// Bugzilla Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1595217
+		// Loop through cache entries, and remove any that refer to files/dirs that do not exist.
+		// Log a warning in worker log, and don't raise an exception.
+		if _, err = os.Stat((*cm)[i].Location); err != nil {
+			log.Printf("WARNING: Cache %#v missing on worker - corrupt internal state, ignoring!", (*cm)[i])
+			// note, this should be safe, despite looking scary:
+			// https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-map-within-a-range-loop
+			delete(*cm, i)
```
matt-boris, line 155: Interesting to see why this is safe too! https://stackoverflow.com/a/23231539

matt-boris (dismissed):
Code LGTM, but is there a good way to test this?

PR #5447, opened 2022-05-16
build(deps): upgrade to go1.18.2

matt-boris:
> Go update from 1.18.1 to 1.18.2. Also upgrade `golangci-lint` from 1.45.2 to 1.46.1.


matt-boris:
@djmitche do you think removing the additional `ne.Temporary()` calls would be fine?

```
WARN [linters context] structcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649. 
tools/websocktunnel/cmd/wst-client/main.go:102:16: SA1019: ne.Temporary has been deprecated since Go 1.18 because it shouldn't be used: Temporary errors are not well-defined. Most "temporary" errors are timeouts, and the few exceptions are surprising. Do not use this method. (staticcheck)
				if !ok || !ne.Temporary() {
				           ^
workers/generic-worker/expose/portforward.go:17:40: SA1019: ne.Temporary has been deprecated since Go 1.18 because it shouldn't be used: Temporary errors are not well-defined. Most "temporary" errors are timeouts, and the few exceptions are surprising. Do not use this method. (staticcheck)
			if ne, ok := err.(net.Error); ok && ne.Temporary() {
			                                    ^
```

lotas (approved):

petemoore (approved):
I'm also fine with this, although I'm not too sure of what the consequences are for removing the `Temporary` error handling, so am happy to take the lead from @djmitche on that one.

djmitche (dismissed):
I think that code is meant to mimic http.Server#Serve, which [still uses err.Temporary()](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/net/http/server.go;l=3038-3072).

I think that errors are rare for Accept, especially timeout errors, so this is probably fine.

Issue #5440, opened 2022-05-11
Add functionality to remove cloud worker from UI 


lotas:
According to sheriff's [how-to notes](https://wiki.mozilla.org/Sheriffing/How_To/Intermittent_bugs#Machine-specific_failures) it was possible to terminate cloud worker from the UI, but it is not anymore.

TODO

- Add remove worker functionality to remove single cloud VM. This should call worker-manager's `removeWorker` endpoint.
- Add batch action to remove all workers in the pool

lotas:
@Archaeopteryx I was going through my notes and noticed this one. I believe this is still an issue for you? 
How often do you go through this manual worker removal? 
How often do you remove all of the workers in a single pool? (I mean if it makes sense to add this to UI, or this is something to be better done manually?)

Thanks!

matt-boris:
I'd imagine that a `terminate` button could live on each row in the table here https://firefox-ci-tc.services.mozilla.com/provisioners/gecko-t/worker-types/win10-64-2004-gpu but am questioning the best location for a `terminate all` button.

Maybe one at the top of that same page? Or maybe in each row of this table https://firefox-ci-tc.services.mozilla.com/worker-manager? Or even the worker pool detail view https://firefox-ci-tc.services.mozilla.com/worker-manager/comm-1%2Fb-linux.

What do you guys think? @Archaeopteryx @lotas

Archaeopteryx:
> How often do you go through this manual worker removal?
Usually 0-3x per week

> How often do you remove all of the workers in a single pool?
Once per quarter or less. This might be more interesting for Pete or RelOps when new AMIs get deployed. But doing it from the command line should be fine. At the moment code sheriffs [retrieve the list of machines known to the pool](https://wiki.mozilla.org/Sheriffing/How_To/Intermittent_bugs#Machine-specific_failures) (even ones which got released) and try to terminate each.

Code sheriffs' workflow is to enter the TC UI on the page for a task, navigate to the machine for the worker instance which lists the tasks it executed and their results. Here they can see if the machine needs to be terminated. Terminating a single machine from the list view would be an uncommon workflow for code sheriffs but might be the fastest way for RelEng. They could search the list for the ID and then initiate the termination.

matt-boris:
@Archaeopteryx thanks for these notes.

Just to clarify what you're referring to, I'm going to provide a link to a page.

> navigate to the machine for the worker instance which lists the tasks it executed and their results

[...]

PR #5448, opened 2022-05-16
feat: mix queue and worker info to provide additional data/functionality in Worker views

matt-boris:
Addresses https://github.com/taskcluster/taskcluster/issues/3060 and https://github.com/taskcluster/taskcluster/issues/5440.


services/queue/src/api.js
```diff
@@ -1949,7 +1949,7 @@ builder.declare({
   },
   name: 'listWorkers',
   scopes: 'queue:list-workers:<provisionerId>/<workerType>',
-  stability: APIBuilder.stability.experimental,
+  stability: APIBuilder.stability.deprecated,
```
lotas, line 1952: Hmm, why do you want to deprecate `listWorkers` and `getWorker`? Afaik they are quite popular and are used outside if TC UI (cli tools).

lotas: Actually yes, it was marked as experimental anyway, but we should keep it for at least 1 major release more I think

services/worker-manager/schemas/v1/worker-full.yml
```diff
@@ -80,7 +80,7 @@ properties:
     type: integer
     minimum: 1
 
-additionalProperties: false
+additionalProperties: true
```
lotas, line 83: Do you remember how it was used before and why is it needed to allow extra props now?

services/worker-manager/src/api.js
```diff
@@ -5,6 +5,8 @@ const { ApiError, Provider } = require('./providers/provider');
 const { UNIQUE_VIOLATION } = require('taskcluster-lib-postgres');
 const { WorkerPool, WorkerPoolError, Worker } = require('./data');
 const { createCredentials } = require('./util');
+const { joinTaskQueueId, addSplitFields } = require('../../queue/src/utils');
```
lotas, line 8: I think services should not be aware of other service's internals. Can we move that to a common lib instead?

services/worker-manager/src/api.js
```diff
@@ -5,6 +5,8 @@ const { ApiError, Provider } = require('./providers/provider');
...
+const { TaskQueue } = require('../../queue/src/data');
```
lotas, line 9: This is where it gets interesting :) 
As soon as we start to introduce such dependencies, we will have one double-micro-service :D 
If we really need this class, we can copy it like you did with the schema. Having copy-pasted class would be better than having highly coupled services imho.

services/worker-manager/src/api.js
```diff
@@ -839,6 +841,137 @@ builder.declare({
...
+    'option. By default this end-point will list up to 1000 workers in a single',
+    'page. You may limit this with the query-string parameter `limit`.',
+  ].join('\n'),
+}, async function(req, res) {
+  const quarantined = req.query.quarantined || null;
+  const requested = req.query.requested || null;
+  const running = req.query.running || null;
+  const stopping = req.query.stopping || null;
+  const stopped = req.query.stopped || null;
+  const workerState = requested || running || stopping || stopped || null;
```
lotas, line 880: Why don't we just use enum-like query parameter? It seems like they are mutually exclusive anyway and only will be filtered by one of the values. So instead of having 5 boolean params, why not have 1 `workerState` of type string/enum? Validation is already done above with regexes.

services/worker-manager/src/api.js
lotas: Actually those are not needed at all, worker manager has own utils:

https://github.com/taskcluster/taskcluster/blob/5a25a717299e9eae61d7ed0935ceb028b8319a26/services/worker-manager/src/util.js#L12-L25

services/worker-manager/src/api.js
```diff
@@ -839,6 +841,137 @@ builder.declare({
...
+  const { provisionerId, workerType } = req.params;
+  const now = new Date();
+  const taskQueueId = joinTaskQueueId(provisionerId, workerType);
```
lotas, line 883: So once you will switch to worker's manager implementation it would  be named `workerPoolId`. Because `taskQueueId` is terminology from the Queue service ;)

Check [docs](https://firefox-ci-tc.services.mozilla.com/docs/reference/platform/queue/worker-hierarchy#worker-pools-and-task-queues)

services/worker-manager/src/api.js
```diff
@@ -839,6 +841,137 @@ builder.declare({
...
+  stability: APIBuilder.stability.experimental,
+  output: 'worker-response.yml',
+  title: 'Get a worker-type',
+  category: 'Worker Metadata',
+  description: [
+    'Get a worker from a worker-type.',
+  ].join('\n'),
+}, async function(req, res) {
+  const { provisionerId, workerType, workerGroup, workerId } = req.params;
+  const taskQueueId = joinTaskQueueId(provisionerId, workerType);
```
lotas, line 944: Let's also use `workerPoolId` here

services/worker-manager/src/data.js
```diff
@@ -178,7 +179,7 @@ class WorkerPoolError {
   // or one elements.  This matches the semantics of get_worker_pool_error.
   static fromDbRows(rows) {
     if (rows.length === 1) {
-      return WorkerPool.fromDb(rows[0]);
+      return WorkerPoolError.fromDb(rows[0]);
```
lotas, line 182: ouch! nice catch! :)

ui/src/components/WorkerDetailsCard/index.jsx
```diff
@@ -1,52 +1,189 @@
...
  * Render information in a card layout about a worker.
  */
+@withAuth
 export default class WorkerDetailsCard extends Component {
   static propTypes = {
     /** A GraphQL worker response. */
     worker: worker.isRequired,
   };
 
+  state = {
```
lotas, line 29: iirc, nested objects inside state are not so common, and usually state is a plain object.
Since you don't have anything else there besides `dialogState`, maybe it can all go to the top level?

ui/src/components/WorkerDetailsCard/index.jsx
```diff
@@ -1,52 +1,189 @@
...
+        error: null,
+      },
+    });
+
+    try {
+      if (typeof user?.credentials === 'undefined') {
+        throw new Error('User credentials not found. Please log in.');
+      }
+
+      const wm = new WorkerManager({
```
lotas, line 72: would be a good idea to move this logic to util function, like https://github.com/taskcluster/taskcluster/blob/3f2a65aa0945bcb3851eb0b4f5c27bfc19e695be/ui/src/utils/getArtifactUrl.js#L1-L12

ui/src/components/WorkerDetailsCard/index.jsx
```diff
@@ -1,52 +1,189 @@
...
+              secondary={
+                quarantineUntil
+                  ? format(parseISO(quarantineUntil), 'yyyy/MM/dd')
+                  : 'n/a'
+              }
+            />
+          </ListItem>
+          {providerId !== NULL_PROVIDER ? (
+            <ListItem>
+              <Button
```
lotas, line 157: one suggestion here: if you switch to `ui/src/components/Button`  you will be able to use `requiresAuth` prop, which will disable button if user is not authenticated, and you will not have to do the check on line 68-70

ui/src/components/WorkerDetailsCard/index.jsx
```diff
@@ -1,52 +1,189 @@
...
+                Terminate Worker
+              </Button>
+            </ListItem>
+          ) : (
+            <ListItem>
+              <Label mini status="warning">
+                Scheduled for termination
+              </Label>
+            </ListItem>
+          )}
```
lotas, line 174: this came out of nowhere :) 
if provider is not null provider -> display "Delete" button, else -> worker is terminated? Are we jumping into conclusions about `worker.state` here?  :)

lotas: Speaking of states.. I don't think we show here worker state at all, should we add this?

ui/src/components/WorkersTable/index.jsx
```diff
@@ -88,6 +111,75 @@ export default class WorkersTable extends Component {
...
+
+      const wm = new WorkerManager({
+        rootUrl: window.env.TASKCLUSTER_ROOT_URL,
+        credentials: user.credentials,
+        authorizedScopes: [
+          `worker-manager:remove-worker:${taskQueueId}/${workerGroup}/${workerId}`,
+        ],
+      });
+
+      await wm.removeWorker(taskQueueId, workerGroup, workerId);
```
lotas, line 152: I've seen this code above, let's move it to utils function :)

ui/src/components/WorkersTable/index.jsx
```diff
@@ -152,112 +249,158 @@ export default class WorkersTable extends Component {
...
+                    onClick={this.handleDialogActionOpen(
+                      taskQueueId,
+                      workerGroup,
+                      workerId
+                    )}>
+                    <DeleteIcon size={iconSize} />
+                  </IconButton>
+                ) : (
+                  <Label mini status="warning" className={classes.button}>
+                    Scheduled for termination
```
lotas, line 371: why is it scheduled for termination?

services/worker-manager/src/api.js
lotas: Ok, after checking the component's code below, maybe extra params are fine, we just need to fix line 880, as I'm pretty sure it can also be any of `true|false|null` values so will not be helpful to filter states

services/worker-manager/schemas/v1/worker-full.yml
matt-boris: This was needed in the two new wm APIs, `listWorkers` and `getWorker` because they both return a _little_ more data (from the `queue_workers` table) because of that join happening. I decided to just use the same schema that was there and enabled additional props. I could separate this out into a new schema for just those two new APIs that have the queue data.

services/worker-manager/src/api.js
matt-boris: Perfect, thanks for pointing this out. And the point where services should not be aware of other services' internals. 👍🏻

services/worker-manager/src/api.js
matt-boris: I'm gonna combine to just one `workerState`. I initially had it written that way and changed my mind for some reason.

services/worker-manager/src/api.js
```diff
@@ -839,6 +841,137 @@ builder.declare({
...
+      };
+      if (worker.recentTasks.length > 0) {
+        entry.latestTask = worker.recentTasks[worker.recentTasks.length - 1];
+      }
+      if (worker.quarantineUntil.getTime() > now.getTime()) {
+        entry.quarantineUntil = worker.quarantineUntil.toJSON();
+      }
+      return entry;
+    }),
+  };
```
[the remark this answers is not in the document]
lotas: we can also do it separately, but would be really good to do this before the release, as I think this is going to hit performance, as some pools are really big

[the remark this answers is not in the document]
matt-boris: Yeah, try filtering by quarantined today on FxCI.. pretty slow... I'll make an issue for it.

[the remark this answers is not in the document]
matt-boris: Just created https://github.com/taskcluster/taskcluster/issues/5456.

ui/src/components/WorkerDetailsCard/index.jsx
matt-boris: This is one of the things I'd love to work alongside you with to make sure we get it right! :)

ui/src/components/WorkersTable/index.jsx
matt-boris: Same https://github.com/taskcluster/taskcluster/pull/5448#discussion_r881864891

services/worker-manager/src/api.js
djmitche: Thanks for spotting this!  100% agree that services shouldn't be referencing each other (and IIRC that's somewhere in the dev docs).

lotas: Indeed, that's in the best practices doc. https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/microservices.md#independence

matt-boris:
Reran [`lint-nodejs`](https://community-tc.services.mozilla.com/tasks/AQNKsOy_ST-tMyhagQyLag) task with success.

ui/src/views/Provisioners/ViewWorker/index.jsx
```diff
@@ -151,6 +157,49 @@ export default class ViewWorker extends Component {
...
+  handleTerminateDeleteClick = async () => {
+    const { workerPoolId, workerGroup, workerId } = this.state;
+    const { user } = this.props;
+
+    this.setState({
+      terminateDialogError: null,
+    });
+
+    try {
+      removeWorker({ workerPoolId, workerGroup, workerId, user });
```
lotas, line 181: this should very likely be `await removeWorker` unless you want to to run in "background", in which case try-catch is useless

ui/src/views/Provisioners/ViewWorkers/index.jsx
```diff
@@ -36,6 +43,10 @@ import workersQuery from './workers.graphql';
         parse(location.search.slice(1)).filterBy === 'quarantined'
           ? true
           : null,
+      workerState:
+        parse(location.search.slice(1)).filterBy in STATES
```
lotas, line 47: to be honest, I haven't seen this python approach being popular in js to check if key exists. As `in` operator also checks props down the prototypes inherited, like:

```
const a = {"one": 1}
"toString" in a // returns true here, because a.toString() is defined in object's prototype
```

It would be safer to use `.hasOwnProperty()` or just simplify it to do a check against allowed values like `Object.values(STATES).includes(location.search..)` 

wdyt?

ui/src/views/Provisioners/ViewWorker/index.jsx
matt-boris: Great catch, thank you!

PR #5449, opened 2022-05-17
Deprecate old Azure endpionts: Tables/Containers

lotas:


services/auth/src/azure.js
```diff
@@ -18,7 +19,7 @@ builder.declare({
   name: 'azureAccounts',
   input: undefined,
   output: 'azure-account-list-response.yml',
-  stability: 'stable',
+  stability: APIBuilder.stability.stable,
```
petemoore, line 22: Should this be deprecated?

services/auth/test/azure_test.js
```diff
@@ -15,304 +13,9 @@ helper.secrets.mockSuite(testing.suiteName(), ['azure'], function(mock, skipping
...
-  suiteSetup('get azure test account name', async function() {
-    const cfg = await helper.load('cfg');
-    testaccount = _.keys(cfg.azureAccounts)[0];
-  });
-
   test('azureAccounts', function() {
     return helper.apiClient.azureAccounts().then(function(result) {
       assert.deepEqual(result.accounts, _.keys(helper.cfg.azureAccounts));
     });
   });
```
petemoore, line 20: Can this go too? Can we remove the whole file?

petemoore (approved):
Looks good, just a couple of questions!

changelog/V-VAtsIKQuGv2Q-_LVDz1w.md
```diff
@@ -0,0 +1,6 @@
+audience: general
+level: minor
```
petemoore, line 2: Maybe patch, since no new functionality is added?

lotas: sure, yes

matt-boris (approved):

services/auth/src/azure.js
lotas: discussed offline, yes, switching to deprecated. This endpoint just returned existing configurations, which could be done with secrets service

Issue #5437, opened 2022-05-09
Taskcluster CI does not report to Github the status of a task re-run


jonalmeida:
**Describe the bug**

The Github UI in a pull request does not update when a failed task has been re-run. So it looks like the task is still failed until you click into it and go to the tc UI or until a success is posted.

**To Reproduce**

1. Have a Github project with Taskcluster CI running on pull requests.
2. Have one of the CI tasks fail.
3. Observe status of the task on Github UI.
4. Re-run the task from taskcluster CLI (or any other preferred way).
3. Observe status of the task on Github UI.

**Expected behaviour**

- At 3, the status should show a failure.
- At 5, the status should show running.

**Actual behaviour**
- At 3, the status shows a failure.
- At 5, the status shows a failure.

**Taskcluster version**

44.13.5

**Screenshots**

This task was restarted and is currently running, but the UI still shows the task as a failure:
<img width="924" alt="Screen Shot 2022-05-09 at 3 13 07 PM" src="https://user-images.githubusercontent.com/1370580/167481793-39d807b9-b3e9-4149-a540-31f0d46609fd.png">

**Platform (please complete the following information):**

 - OS: Mac OS
 - Browser if applicable: firefox
 - Version if applicable: 102.0a1 (2022-05-06) (64-bit)

**Additional context**

lotas:
Hi @jonalmeida  thanks for reporting this. 
Can you please tell me how you do "Re-run the task from taskcluster CLI" this step?

jonalmeida:
@lotas I used the CLI tool for my device. For example: [`taskcluster-darwin-amd64`](https://github.com/taskcluster/taskcluster/releases/download/v44.13.6/taskcluster-darwin-amd64)

Then:
1. `TASKCLUSTER_ROOT_URL=https://firefox-ci-tc.services.mozilla.com/; taskcluster signin`
2. Authenticated.
3. `taskcluster task rerun [taskId]`

lotas:
Indeed. Let's keep both until resolved.

jonalmeida:
I'm quite alright with closing this as a dupe since Aki has commented on that bug on behalf of MoCo mobile. :)

lotas:
alright :) Let's then close this one and work on the fix in #5085. 
Thanks!

Issue #5085, opened 2022-01-26
When rerunning a failed action, the github check doesn't go back from failed to running


Eijebong:
Run an action from the github integration that fails, rerun it from the taskcluster interface, the github check stays failed instead of going back to running.

escapewindow:
This is also causing the MoCo mobile team issues since they often have a failing task they need to rerun to merge a PR.

jonalmeida:
I see some merged changes linked to this bug, but I'm still seeing facing this issue in mozilla-mobile/android-components with tasks that are re-run, and then put in the the pending state on TC, but not reflected in the Github UI.

lotas:
hey @jonalmeida thanks for letting us know. This needs a bit more testing, I got stuck with missing roles and discussions who should configure those. This is also part of our priorities for this quarter. Hopefully it will be working soon

lotas:
@jonalmeida quick update from my side, I was just fixing something related in github integration and also tested 're-run' functionality: #5617 

It worked for me, but I had to make few tweaks:
- before it didn't send an error message properly if TC was setup incorrectly (missing scopes). Now taskcluster bot should send a comment with a detailed error message.
- you need to make sure that taskcluster is configured properly: it should have a role which allows tasks to be rerun. This can either be through a generic `repo:github.com/org/repo:*` role, or dedicated role `repo:github.com/org/repo:rerun`, that will allow `queue:rerun-task:*` and `queue:rerun-task-in-project:*`

By looking at [current scopes](https://firefox-ci-tc.services.mozilla.com/auth/scopes/expansions?scopes%5B0%5D=assume%3Arepo%3Agithub.com%2Fmozilla-mobile%2Fandroid-components%3A%2A) I think it is missing the `queue:rerun-task-in-project:*` scope

Upd: checking service logs, I see this:

```
Client ID static/taskcluster/github does not have sufficient scopes and is missing the following scopes:

{
  "AnyOf": [
    "queue:rerun-task:mobile-level-1/KRO4xszhSze7HeQF7wB6Rg/M4rbApUATVC7XIGwpT6Q9g",
    "queue:rerun-task-in-project:none",
    {
      "AllOf": [
        "queue:rerun-task",
        "assume:scheduler-id:mobile-level-1/KRO4xszhSze7HeQF7wB6Rg"
      ]
    }
  ]
}
```

[...]

PR #5458, opened 2022-05-30
WIP: Github handlers

### LINKED ISSUE #41: Rerun task doesn't update status on PR
Happened on https://github.com/mozilla-releng/build-puppet/pull/82. 

# STR

1. Open PR and let tests run.
2. Have one task failing
3. When the task graph is done, rerun failed task.
4. Let failed task pass.
5. See taskgraph being green, but PR still showing red.

lotas:
This PR is WIP

Aims to resolve #41 #5085 #5437


services/github/schemas/v1/github-rerun-message.yml
```diff
@@ -0,0 +1,59 @@
...
+      The GitHub webhook deliveryId. Extracted from the header 'X-GitHub-Delivery'
+    pattern: {$const: github-guid-pattern}
+  details:
+    type:         object
+    description: |
+      Metadata describing the push (for version 0)
+  body:
+    type: object
+    description: |
+      The raw body of github event (for version 1)
```
matt-boris, line 38: Are there properties in the `body` that we should be defining in this schema?

matt-boris (commented):
Looks great! Just left a small question about the schema

services/github/schemas/v1/github-rerun-message.yml
lotas: I don't think it is needed. At least all other github schemas there don't define it.

changelog/issue-5085.md
```diff
@@ -0,0 +1,6 @@
+audience: users
+level: minor
+reference: issue 5085
+---
+
+Listen to github's check_run events that requests task to be rerun.
```
djmitche, line 6: You might want to phrase this more from a user's perspective?

db/test/versions/0076_test.js
```diff
@@ -0,0 +1,4 @@
+const testing = require('taskcluster-lib-testing');
+
+suite(testing.suiteName(), function() {
```
djmitche, line 3: no tests?

services/github/src/api.js
```diff
@@ -89,6 +89,21 @@ function getReleaseDetails(eventData) {
...
+    'event.head.user.login': eventData.sender.login,
+    'event.head.user.id': eventData.sender.id,
+    'event.check.name': eventData.check_run.name,
+    'event.check.run.id': String(eventData.check_run.id),
+    'event.check.run.url': eventData.check_run.url,
+    'event.check.suite.url': eventData.check_run.check_suite.url,
+    'event.check.suite.id': String(eventData.check_run.check_suite.id),
+    'event.head.repo.name': eventData.repository.name,
+  };
+}
```
djmitche, line 105: Are all of these used? I think this pattern was because these keys are exposed to the user in the v0 taskcluster.yml.  Since (a) v0 isn't supported anymore and (b) this message isn't going to result in a JSON-e rendering of taskcluster.yml, you'll only need the "details" that the handler requires, and it may be simpler to just send `eventData` directly rather than select and rename fields like this.

services/github/src/handlers/index.js
```diff
@@ -0,0 +1,376 @@
...
+    assert(!this.resultStatusPq, 'Cannot setup twice!');
+    assert(!this.initialTaskStatusPq, 'Cannot setup twice!');
+    assert(!this.deprecatedResultStatusPq, 'Cannot setup twice!');
+    assert(!this.deprecatedInitialStatusPq, 'Cannot setup twice!');
+    assert(!this.rerunPq, 'Cannot setup twice!');
+
+    // This is a powerful Queue client without scopes to use throughout the handlers for things
+    // where taskcluster-github is acting of its own accord
+    // Where it is acting on behalf of a task, use this.queueClient.use({authorizedScopes: scopes}).blahblah
+    // (see this.createTasks for example)
```
djmitche, line 86: I think this is an important point -- `this.queueClient.rerunTask()` will use tc-github's scopes, which means it can rerun just about any task.  Other handlers have things like
https://github.com/taskcluster/taskcluster/blob/3eddaee81983d2f631952904298bbbf637afc272/services/github/src/handlers.js#L460-L462
in this case, `scopes` would probably be defined as something like `["assume:repo:github.com/<org>/<repo>:rerun"]`.

matt-boris (approved):

djmitche (commented):
I appreciate the refactor into multiple files!  There is certainly room for further improvements to the codebase in this service, but this is a nice start.

I think the scopes question is an important one.  The GitHub service automatically has `assume:repo:github.com/*`, so whatever scopes `queue:rerun-task:...` that implies will apply to re-running a task as this PR is currently written -- which is probably both too broad (one project can rerun another project's tasks) and too narrow (if no role has the necessary `queue:rerun-task:..` scope, the rerun will fail).

The situation is both a little bit risky, because _lots_ of people can click that rerun button in GitHub, and a little bit safe, because it's difficult to forge a GitHub webhook and even more difficult to forge a pulse message, so it's reasonable to assume that the suite/run ID's are something the originating user had access to.

services/github/src/handlers/index.js
lotas: Thanks! Will apply this scope

matt-boris (approved):

lotas:
Scopes are not correctly applied at the moment:

```
Error: Client ID static/taskcluster/github does not have sufficient scopes and is missing the following scopes:

```
{
  "AnyOf": [
    "queue:rerun-task:taskcluster-level-1/Xq9wYVyoRD24ZQJCDJgNJw/UV58TFnxTwWrXp_9kAqmQg",
    "queue:rerun-task-in-project:none",
    {
      "AllOf": [
        "queue:rerun-task",
        "assume:scheduler-id:taskcluster-level-1/Xq9wYVyoRD24ZQJCDJgNJw"
      ]
    }
  ]
}
```

Issue #5459, opened 2022-05-31
Add retries to `execCommand` method


matt-boris:
`release-publish` tasks intermittently fail due to the [`dockerPush`](https://github.com/taskcluster/taskcluster/blob/66941613e4242d69dd2aff3fb560359eb633d59c/infrastructure/tooling/src/utils/docker.js#L246-L291) command failing, usually during the build websocktunnel docker image. This is fixed by rerunning the task over a few times until it succeeds. An extra param passed to `execCommand` with number of retires may be a good solution for the short term.

lotas:
It could also be some quirk of streams/pipes, maybe there's another bug that fails.
But in general retries sounds like a good idea

PR #5460, opened 2022-05-31
feat(release): add exponential backoff retries to `dockerPush` function used in `release-publish` task

matt-boris:
Addresses https://github.com/taskcluster/taskcluster/issues/5459.

> Add exponential backoff retries to the `dockerPush` function to help alleviate intermittent failures in the `release-publish` task.


lotas (approved):
nice! should work

Issue #5361, opened 2022-04-13
Link Worker State Buttons to a filtered view of workers in that state


matt-boris:
On the [Worker Pool details view page](https://firefox-ci-tc.services.mozilla.com/worker-manager/gecko-t%2Fwin10-64-2004), we should make the buttons at the top link to a filtered list of workers in that state, probably the [workers view](https://firefox-ci-tc.services.mozilla.com/provisioners/gecko-t/worker-types/win10-64-2004).

PR #5465, opened 2022-06-01
feat(ui): link to filtered workers view

matt-boris:
Addresses #5361.

> Link Worker State Buttons on Worker Pool Details View to a filtered view of workers in that state.


lotas (approved):
:+1:

PR #5475, opened 2022-06-01
fix(schema): don't require additional wm fields

matt-boris:
Fixes this error:

```
"Output schema validation error: 
Schema Validation Failed!
Rejecting Schema: https://stage.taskcluster.nonprod.cloudops.mozgcp.net/schemas/worker-manager/v1/list-workers-response.json#
Errors:
  * data/workers/0 should have required property 'state'
  * data/workers/0 should have required property 'capacity'
  * data/workers/0 should have required property 'providerId'"
```

> Fix output schema validation error when calling `workerManager.listWorkers()` and `workerManager.getWorker()` methods by not requiring additional worker manager fields.

### DIFF
--- a/changelog/WEXGTNjNQYaDFU7YuJjiDQ.md
+++ b/changelog/WEXGTNjNQYaDFU7YuJjiDQ.md
@@ -1,0 +1,4 @@
+audience: users
+level: patch
+---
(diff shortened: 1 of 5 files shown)


lotas (dismissed):
👍

PR #5490, opened 2022-06-02
fix(api): handle null checks and extract wm data

matt-boris:
Addresses this error:
image

Also, I made sure to extract all possible worker data in the `workerManager.listWorkers()` API.

> Handle some null checks with optional chaining. Also, ensure all data is extracted out during `workerManager.listWorkers()` calls.


services/worker-manager/src/api.js
```diff
@@ -899,14 +899,18 @@ builder.declare({
...
+        state: worker.state,
+        capacity: worker.capacity,
+        providerId: worker.providerId,
       };
       if (worker.recentTasks.length > 0) {
         entry.latestTask = worker.recentTasks[worker.recentTasks.length - 1];
       }
       if (worker.quarantineUntil.getTime() > now.getTime()) {
-        entry.quarantineUntil = worker.quarantineUntil.toJSON();
+        entry.quarantineUntil = worker.quarantineUntil?.toJSON();
```
lotas, line 913: this one is already inside the check above, if it will be null/undefined it will fail before anyways - `worker.quarantineUntil.getTime()`

lotas (approved):

lotas (approved):

PR #5496, opened 2022-06-07
build(deps): upgrade node to 16.15.1

matt-boris:
> Upgrade node to the latest LTS release, v16.15.1. Also upgrade `golangci-lint` to 1.46.2.


matt-boris:
Getting this error in `lint-golang`:

```
Running golangci-lint..
test/go-lint.sh: 12: test/go-lint.sh: golangci-lint: not found
```

matt-boris:
After changing up the shebangs from `#! /bin/bash` to have to space, `#!/bin/bash` and moving some scripts to be called with `bash` instead of `sh`, this failure has gone away!

PR #5497, opened 2022-06-08
build(deps): upgrade to `go1.18.3`

matt-boris:
> Upgrade to `go1.18.3` from `go1.18.2`.


matt-boris:
Rebasing branch off of https://github.com/taskcluster/taskcluster/pull/5496 to get the `lint-golang` task fix. So https://github.com/taskcluster/taskcluster/pull/5496 should be merged before this one.

db/test-setup.sh
```diff
@@ -1,4 +1,4 @@
-#! /bin/bash
+#!/bin/bash
```
lotas, line 1: ouch! nice one

matt-boris: First three commits on this PR are from https://github.com/taskcluster/taskcluster/pull/5496 btw :)

lotas: hmm, actually space is allowed :D https://en.wikipedia.org/wiki/Shebang_(Unix)#History  who knew

lotas: > First three commits on this PR are from #5496 btw :)

Approving both

lotas (approved):
Thanks!

PR #5499, opened 2022-06-09
fix(schema, graphql): use explicit values when a worker is not known by worker manager, add better descriptions, update graphql queries

matt-boris:
Addresses PR comments from @petemoore on #5491.

> Don't allow additional properties in `worker-response.yml` schema. Updated descriptions in `worker-response.yml` and `list-workers-response.yml` schemas to explain some default values that may occur in the case where the queue knows about the worker, but worker manager does not. Also, updated the GraphQL queries to extract additional needed data.


lotas (dismissed):
:+1:

petemoore (approved):
Looks perfect!

PR #5505, opened 2022-06-10
fix(ui): disable `Terminate Worker` button for static and standalone workers

matt-boris:
> This patch makes it so a user cannot click the `Terminate Worker` button on a Static or Standalone worker. This issue was first brought up in v44.15.0


ui/src/components/WorkersTable/index.jsx
```diff
@@ -328,15 +328,22 @@ export default class WorkersTable extends Component {
                 {providerId !== NULL_PROVIDER && state !== 'stopping' && (
                   <Button
                     requiresAuth
-                    disabled={['stopping', 'stopped'].includes(state)}
+                    disabled={
+                      ['stopping', 'stopped'].includes(state) ||
```
lotas, line 332: `stopping` will not reach this point, as it is being filtered out 4 lines above this

ui/src/views/Provisioners/ViewWorker/index.jsx
```diff
@@ -288,6 +279,26 @@ export default class ViewWorker extends Component {
...
+                    onClick={() =>
+                      this.handleTerminateDialogActionOpen(
+                        worker.workerId,
+                        worker.workerGroup,
+                        worker.workerPoolId
+                      )
+                    }
+                    FabProps={{
+                      disabled:
+                        ['stopping', 'stopped'].includes(worker.state) ||
```
lotas, line 297: I know it's minor, but maybe this can be defined in one place (`WorkerTable` uses same logic), like new tiny util function

ui/src/views/Provisioners/ViewWorker/index.jsx
```diff
@@ -288,6 +279,26 @@ export default class ViewWorker extends Component {
                     disabled: actionLoading,
                   }}
                 />
+                {worker.providerId !== NULL_PROVIDER && (
+                  <SpeedDialAction
+                    tooltipOpen
+                    requiresAuth
+                    icon={<DeleteIcon />}
+                    tooltipTitle="Terminate Worker"
```
lotas, line 287: should this tooltip also be dynamic as in WorkerTable?

ui/src/components/WorkersTable/index.jsx
matt-boris: Good catch, thank you! I've decided that for here, I'll just not even show the terminate button to the users if the state is `stopping` or `stopped`.

ui/src/views/Provisioners/ViewWorker/index.jsx
matt-boris: Screen Shot 2022-06-13 at 9 44 00 AM

I decided not to, since the tooltip in this component is essentially the title used describing the button in the speed dial menu. There's no real "tooltip" on hover over these menu options. If that were the case, I'd put the same logic. I felt that just disabling it was sufficient in this case.

lotas (approved):

lotas (approved):
Nice, thanks!

PR #5507, opened 2022-06-13
Log debug information for incoming github webhooks

### LINKED ISSUE #5085: When rerunning a failed action, the github check doesn't go back from failed to running

### LINKED ISSUE #5506: Log additional information for Github webhook events
Currently when trying to troubleshoot intermittent failures for the Github integration, minimal information is logged about the webhook events received (only the installation id and the event id - these identifiers may not be easily accessible by those doing the troubleshooting).

Including additional information from the webhook request would allow less people to start to troubleshoot intermittent failures - organization, repository and action should all be common fields that would give us a bit more immediate context when debugging.

lotas:
Closes #5506

### DIFF
--- a/changelog/issue-5506.md
+++ b/changelog/issue-5506.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 5506
+---
+
+Log debug information for incoming Github webhooks.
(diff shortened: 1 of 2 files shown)


petemoore (approved):
Thanks Yarik, looks great!

PR #5508, opened 2022-06-13
Add missing rerun scope for github handler.

### LINKED ISSUE #5503: Github rerun action missing scopes
Scopes are not correctly applied at the moment:

```
Error: Client ID static/taskcluster/github does not have sufficient scopes and is missing the following scopes:

{
  "AnyOf": [
    "queue:rerun-task:taskcluster-level-1/Xq9wYVyoRD24ZQJCDJgNJw/UV58TFnxTwWrXp_9kAqmQg",
    "queue:rerun-task-in-project:none",
    {
      "AllOf": [
        "queue:rerun-task",
        "assume:scheduler-id:taskcluster-level-1/Xq9wYVyoRD24ZQJCDJgNJw"
      ]
    }
  ]
}
```

_Originally posted by @lotas in https://github.com/taskcluster/taskcluster/pull/5458#issuecomment-1152531070_

lotas:
Fixes #5503

To be able to use `re-run` events, github project must be set up properly.
Github handler will assume `repo:github.com:/$org/$repo:pull-request` role when calling `queue.rerunTask()`
This means this role needs to be defined, and it should include necessary scopes to make that call:  

https://github.com/taskcluster/taskcluster/blob/cc0c17e2601075ad328ae5d2af2ce2edcf71d266/services/queue/src/api.js#L761-L767

### DIFF
--- a/changelog/issue-5503.md
+++ b/changelog/issue-5503.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 5503
+---
+
+Add missing task-rerun scope to github handler.
--- a/services/github/src/handlers/index.js
+++ b/services/github/src/handlers/index.js
@@ -196,7 +196,7 @@ class Handlers {
         bindings: rerunBindings,
         queueName: this.rerunQueueName,
       },
-      this.monitor.timedHandler('tasklistener', callHandler('rerun', rerunHandler).bind(this)),
+      this.monitor.timedHandler('rerunlistener', callHandler('rerun', rerunHandler).bind(this)),
     );
 
   }
--- a/services/github/src/handlers/rerun.js
+++ b/services/github/src/handlers/rerun.js
@@ -2,6 +2,9 @@ const { makeDebug } = require('./utils');
 
 /**
  * Github events that request a task rerun.
+ *
+ * To be able to rerun a task, client will assume `repo:github.com/<owner>/<repo>:rerun` role.
+ * This role needs to be defined and should include scopes necessary to call `queue.rerunTask`
  **/
 async function rerunHandler(message) {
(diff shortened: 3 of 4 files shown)


petemoore (approved):
That looks reasonable to me, and I think should be sufficient, although I'm happy for @djmitche to take a look too.

Am I right in thinking to rerun a github task you still need `repo:<repo>:rerun` scope?

lotas:
Checking API code: https://github.com/taskcluster/taskcluster/blob/main/services/queue/src/api.js#L761-L768 

`queue:rerun-task` is actually a "Legacy" scope based on the comment in code. So probably I should not be using that hmm

djmitche:
This is a case where TC-Github is doing something "on behalf of" the repository, so it should probably be using a role that the user has to define (although [this doc](https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/scopes.md) does not specifically address this case).  That's what assuming the `repo:...:rerun` role would accomplish -- it's the one place where "assume" really means assume, as TC-GitHub is assuming the role of the repo for purposes of rerunning.

I suspect that there are cases where projects do not want their tasks to be re-run.  Fuzzing comes to mind as an example.  But any project that's paying for its resources might also want to control this.  That said, _maybe_ relying on GitHub's permissions to click the "rerun" button is sufficient, and cases like fuzzing are sufficiently unusual to ignore?  I'm not sure -- but it's the question to ask, I think.

lotas:
@petemoore @djmitche thanks for the help, I fell like I finally figured it out :) 

I am setting this to assume `repo:github.com/org/repo:pull-request` role, but have a feeling that this will require extra work (although one-time work) to add those permissions.

matt-boris (approved):

PR #5528, opened 2022-06-23
Remove azure library

petemoore:
Curious if we can remove these libraries.


petemoore:
Doesn't look like we're ready to get rid of them quite yet

lotas:
@petemoore this is probably fine to remove, can you reabase to see what tests would fail? I think none of them should

matt-boris (approved):
It's a shame you needed to add three more lines to the monorepo 😉

lotas (approved):
zero tests failed? amazing!

djmitche:
You may want to remove https://docs.taskcluster.net/docs/reference/platform/auth/api#azure-credentials as well (at least azureContainerSAS)

PR #5530, opened 2022-06-24
fix(release): update docker-compose.yml file on `yarn release`

### LINKED ISSUE #5529: `meta-generate` task fails on newly released TC version
**Describe the bug**
When a new version is released, the `meta-generate` task fails (like this, for example: https://community-tc.services.mozilla.com/tasks/MJpcbwBPQV-vWlUPrZuFKA) due to the taskcluster docker image not being updated in the `docker-compose.yml` file during a `yarn release`.

**To Reproduce**

1. `yarn release` to cut a new release.
2. View `meta-generate` task failure on the commit that's pushed on release.

**Expected behavior**
Task should pass. The docker image version should likely be updated in the file during `yarn release`.

**Taskcluster version**
v44.16.4

matt-boris:
Fixes #5529.

> This patch makes it so that the `docker-compose.yml` file is updated with the new taskcluster docker image version on a `yarn release`. Previously, the version wasn't updated, so the `meta-generate` task would fail on releases. This issue first appeared in v44.16.4.

### DIFF
--- a/infrastructure/tooling/src/release/tasks.js
+++ b/infrastructure/tooling/src/release/tasks.js
@@ -215,12 +215,18 @@ module.exports = ({ tasks, cmdOptions, credentials }) => {
         changed.push(file);
       }
 
-      utils.status({ message: 'Dockerfile' });
+      utils.status({ message: 'Update Dockerfile' });
       await modifyRepoFile('Dockerfile',
         contents => contents.replace(
           /\\"version\\":\s*\\"([0-9]{1,2}\.[0-9]{1,2}\.[0-9]{1,2})/gm,
           `\\"version\\": \\"${requirements['release-version']}`));
       changed.push('Dockerfile');
+
+      const dockerCompose = 'docker-compose.yml';
+      utils.status({ message: `Update ${dockerCompose}` });
+      await modifyRepoFile(dockerCompose, contents =>
+        contents.replace(/taskcluster\/taskcluster:v([0-9]{1,2}\.[0-9]{1,2}\.[0-9]{1,2})/g, releaseImage));
+      changed.push(dockerCompose);
 
       return { 'version-updated': changed };
     },
--- a/changelog/issue-5529.md
+++ b/changelog/issue-5529.md
@@ -1,0 +1,5 @@
+audience: general
+level: patch
+reference: issue 5529
+---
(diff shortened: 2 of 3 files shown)


infrastructure/tooling/src/release/tasks.js
```diff
@@ -215,13 +215,19 @@ module.exports = ({ tasks, cmdOptions, credentials }) => {
...
       await modifyRepoFile('Dockerfile',
         contents => contents.replace(
           /\\"version\\":\s*\\"([0-9]{1,2}\.[0-9]{1,2}\.[0-9]{1,2})/gm,
           `\\"version\\": \\"${requirements['release-version']}`));
       changed.push('Dockerfile');
 
+      const dockerCompose = 'docker-compose.yml';
+      utils.status({ message: `Update ${dockerCompose}` });
+      await modifyRepoFile(dockerCompose, contents =>
+        contents.replace(/taskcluster\/taskcluster:v([0-9]{1,2}\.[0-9]{1,2}\.[0-9]{1,2})/g, releaseImage));
```
petemoore, line 228: Elsewhere in this file the version number matching isn't quite so strict (e.g. `v[0-9.]*`), which would result in inconsistent behaviour if for some reason an obscure release is made such as `v53.2.6.1` (although unlikely we'd ever do such a thing and our automation probably will never support it).

I think this regexp also only allows 1 or 2 character version numbers, so e.g. v100.1.1 wouldn't be allowed (which is a number we could realistically hit).

My instinct is to be as least restrictive as reasonably possible, to allow for a more generous range of possible version numbers, so that we don't have to encode release version numbering in too many places. So I would be happy to go with `v[0-9.]*` here too, for example. However, maybe this is all rather hypothetical because we are very likely to stick with semantic versioning, and if we change, we would change the code here anyway.

Perhaps we can have a global constant that defines the regexp for the version number, and use that everywhere where we find/replace version number, to be consistent.

If you disagree, we can also go with your regexp, but let's allow numbers to three digits long, and then also use the same RE in the other lines that do a version number find/replace.

Thanks Boris!

matt-boris: Less restrictive now with https://github.com/taskcluster/taskcluster/pull/5530/commits/e3898cc51067f8495e449cb0d5a3b5bc550951c9! Thanks for the call out. I've tested these cases too with 3 digit major/minor/patches on [regex101.com](regex101.com) and all looks 👍🏻

petemoore (approved):

Issue #5456, opened 2022-05-25
Filter workers by state and quarantined should happen in a separate db func


matt-boris:
@lotas commented in this PR https://github.com/taskcluster/taskcluster/pull/5448#discussion_r881444469 that adds additional filtering by worker state using the same pattern that was used in the queue service.

Try toggling the `Filter By` selection to notice the slowness: https://firefox-ci-tc.services.mozilla.com/provisioners/gecko-t/worker-types/win10-64-2004?sortBy=Last%20Active&sortDirection=desc

PR #5531, opened 2022-06-24
feat(db,api): filter workers in db, not in browser

matt-boris:
Addresses #5456.

> This change adds more DB functions to allow for filtering in the DB based on worker state and quarantined status for the workers page. Previously, filtering would only happen on the initial page loaded from the DB if results were paginated. This should also speed up the workers page rendering when a filter is applied.


services/worker-manager/src/api.js
```diff
@@ -868,10 +868,7 @@ builder.declare({
     'page. You may limit this with the query-string parameter `limit`.',
   ].join('\n'),
 }, async function(req, res) {
-  const quarantined = req.query.quarantined || null;
```
lotas, line 871: those are listed in the api definition above on line 848-849, if they are no longer used, should be dropped from declaration.
Are they still used?

lotas: ah, full query is passed I see, nevermind

services/worker-manager/src/data.js
```diff
@@ -337,13 +337,32 @@ class Worker {
     const fetchResults = async (query) => {
       const { continuationToken, rows } = await paginateResults({
         query,
-        fetch: (size, offset) =>
-          db.fns.get_queue_workers_with_wm_join(
+        fetch: (size, offset) => {
+          if (Object.keys(query).includes('workerState')) {
```
lotas, line 341: we should probably put this into endpoint description that you can only use one at a time to filter value: either state or qurantined

lotas (approved):

services/worker-manager/src/data.js
matt-boris: Ah true, good call. Thanks. I was thinking of only _our_ use case which is directly through the UI where only one filter is selectable in the first place. Will update.

matt-boris: It's actually already in the docs :) https://firefox-ci-tc.services.mozilla.com/docs/reference/core/worker-manager/api#listWorkers

PR #5535, opened 2022-06-28
build(deps): upgrade deps and fix vulns snyk found

matt-boris:
> This patch upgrades dependencies to their latest minor/patch versions. Doing so fixes a handful of vulns found within Snyk and docker scans.

See vulns by running `docker scan taskcluster/taskcluster:v44.16.4`.


services/web-server/src/main.js
```diff
@@ -132,6 +132,7 @@ const load = loader(
           tracing: true,
           plugins: [ApolloServerPluginDrainHttpServer({ httpServer })],
           csrfPrevention: true,
+          cache: 'bounded',
```
matt-boris, line 135: https://github.com/apollographql/apollo-server/blob/main/CHANGELOG.md#v390

petemoore (approved):
Thanks Boris!

services/web-server/src/main.js
matt-boris: https://github.com/taskcluster/taskcluster/pull/5535/files#diff-34ea755b1bde6647b339b37ff4f3b5f6152613dc4a7a269dbea38dc337d9bda2R135-R136

petemoore (approved):
Awesome, thanks Boris!

PR #5542, opened 2022-06-30
Remove unused config value

lotas:
`auditLog` was added at some point to log custom audit messages to AWS
It was used in Monitor, but was refactored at https://github.com/taskcluster/taskcluster/commit/4aa762e4ca3d067a5cbf98e32d1e51822178cf4f and never used in the service after

Monitor lib functionality was refactored and auditLogs were removed at #152


lotas:
This seems to be a class that used to submit logs to AWS but was lost during migration to monorepo, or removed because something https://github.com/taskcluster/taskcluster-lib-monitor/blob/master/src/auditlogs.js

matt-boris (approved):

Issue #5557, opened 2022-07-10
Upgrade to Debian 10 docker images


matt-boris:
Debian 9 ("stretch") reached EOL on 2022-06-30. We should upgrade to Debian 10 ("buster") which has LTS support until June 2024.

https://lists.debian.org/debian-lts-announce/2022/07/msg00002.html

- [ ] Upgrade and test in `infrastructure/docker-images/build-rabbit-test.sh`
- [ ] Upgrade and test in `infrastructure/docker-images/build-browser-test.sh`

PR #5558, opened 2022-07-10
build(deps): upgrade node to 16.16.0

matt-boris:
> Upgrade node to the latest LTS release, v16.16.0.
This is a security release. More info can be found [here](https://nodejs.org/en/blog/vulnerability/july-2022-security-releases/#update-07-july-2022-security-releases-available).

Also in this PR, https://github.com/taskcluster/taskcluster/issues/5557 is addressed by upgrading to the current LTS Debian version.

> This patch upgrades to Debian 10 docker images, as Debian 9 hit EOL.


petemoore (approved):
Great stuff, thanks Boris!

PR #5561, opened 2022-07-11
Issue 5560 - multiarch docker builds

### LINKED ISSUE #5560: Create multi-arch docker images
It is useful to have multi-arch docker images to e.g. improve performance of macOS/ARM dev deployments.

petemoore:
Github Bug/Issue: Fixes #5560

### DIFF
--- a/infrastructure/docker-images/build-browser-test.sh
+++ b/infrastructure/docker-images/build-browser-test.sh
@@ -15,5 +15,4 @@ FROM node:${node_version}-buster
 RUN apt-get update && apt-get install -y firefox-esr xvfb
 EOF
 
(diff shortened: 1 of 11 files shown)


lotas:
TIL: there's a [docker manifest](https://docs.docker.com/desktop/multi-arch/) that creates a merged image info (so that docker client could select the correct architecture)
And it looks like `buildx` command is already doing this, so there is no need to build and tag individual images for each platform (see [this article](https://www.howtogeek.com/devops/what-is-a-docker-image-manifest/))

lotas (approved):
This looks good, should we ship it yet @petemoore ? :)

lotas (approved):

Issue #5554, opened 2022-07-06
Split docker-compose into prod and dev versions


matt-boris:
We can split `docker-compose.yml` in two: 

* one for prod-like deployments: `docker-compose up -d` and you can use it
* dev-like deployment where source code will be mounted as volumes (for dev/debug purposes)

For dev deployment we need to add ui container that will run webpack (yarn start) in live mode, instead of serving statically built artifacts

PR #5564, opened 2022-07-12
feat: split out dev/prod docker compose files

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/5554.

> This patch splits the docker compose file into separate dev and prod configuration files. For prod-like deployments, where you want to use the latest `taskcluster/taskcluster` docker image, use the command `docker compose -f docker-compose.yml -f docker-compose.prod.yml up -d`. For development deployments, where local source code mounts as volumes for testing/debugging purposes, use the command `docker compose -f docker-compose.yml -f docker-compose.dev.yml up -d`.
> 
> This change also switches `docker-compose` (v1) references over to `docker compose` (v2). See [here](https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command) for more details.


docker-compose.override.yml
```diff
@@ -0,0 +1,77 @@
+services:
```
lotas, line 1: I would rather make it part of base `docker-compose.yml` that is suited for local dev. So you can still do some extra local overrides without needing to commit changes to this file

Eijebong:
So I tried this and was not able to get the UI container to work properly. Whatever I did I was always getting an `Invalid Host header` when visiting `127.0.0.1:8080`. Not sure if I missed something. Running the UI locally with a `TASKCLUSTER_ROOT_URL` pointing to a deployed instance and running `yarn start` did work though. 

While writing this, I remembered that we do this: https://github.com/taskcluster/taskcluster/blob/76da16819c1e2869952d32ea4ec543b5e80349ff/docker/nginx.conf#L21 which might be why it got angry?

matt-boris:
@Eijebong This work's still a WIP because of that issue. Thanks for pointing out that nginx config though!

docker-compose.yml
```diff
@@ -97,9 +97,6 @@ services:
       GRAPHQL_SUBSCRIPTION_ENDPOINT: http://localhost:8080/graphql
       GRAPHQL_ENDPOINT: http://localhost:8080/graphql
       UI_LOGIN_STRATEGY_NAMES: local
-      DEBUG: '*'
-    command:
-      - ui/web
     ports:
```
lotas, line 100: Do you remember why this command was removed? Now when you simply run `docker compose up` it will fail to start `ui` container because it lacks entrypoint command

infrastructure/tooling/src/generate/generators/docker-compose.js
```diff
@@ -284,7 +286,7 @@ exports.tasks.push({
             MINIO_ROOT_PASSWORD: 'miniopassword',
           },
         }),
-        ui: serviceDefinition('ui', { command: ['ui/web'] }),
+        ui: serviceDefinition('ui'),
```
lotas, line 289: Do you remember why this command was removed? Now when you simply run docker compose up it will fail to start ui container because it lacks entrypoint command

lotas (approved):

PR #5565, opened 2022-07-13
build(deps): upgrade to go1.18.4

matt-boris:
> Go update from 1.18.3 to 1.18.4. Also updates the git version for generic worker decision tasks from `git2.24.0.2` to `git2.37.1`.


workers/generic-worker/gw-decision-task/tasks.yml
```diff
@@ -119,7 +119,7 @@ WorkerPools:
     # The next comment is edited by infrastructure/tooling/src/generate/generators/go-version.js
     #   DO NOT CHANGE HERE!
 
-    # There is no arm release for go 1.18.3 on windows, but 386 release works
+    # There is no arm release for go 1.18.4 on windows, but 386 release works
```
petemoore, line 122: I wonder if we should switch to arm64/windows instead. I believe the Lenovo/ARM laptops we run generic-worker on run the 32 bit ARM instruction set (either ARMv7 or below, or if ARMv8 or higher, running aarch32 rather than aarch64).

@davehouse do you know?

I suspect both aarch32 and ARMv7 or lower are a thing of the past that we might need to support in the future.

petemoore (approved):
Thanks Boris! You're so quick, I didn't even realise go1.18.4 was already out!

matt-boris:
> I didn't even realise go1.18.4 was already out!

@petemoore my editor (VS Code) tells me when an update's available 😉

PR #5567, opened 2022-07-15
Change the access token input to be a password field

Eijebong:
It doesn't much sense to have what is essentially a password be
displayed in plain text.
This also makes it so a password manager would recognize the form as a
login form and is able to automatically fill it.


matt-boris (approved):
Thank you! 👍🏻

PR #5568, opened 2022-07-15
Fix docker compose sometimes not starting up the ingress container

Eijebong:
The issue here is that `proxy_pass http://something` will always try to
resolves `something` as soon as nginx starts. By using an intermediate
variable, we force nginx to resolve it later at runtime [1].
Because the `github-web` container exits without any extra config, it
would becomes unresolvable DNS wise and prevent nginx from starting.
Same with other containers if they took too long to start up.

1: https://forum.nginx.org/read.php?2,215830,215832#msg-215832

### DIFF
--- a/changelog/TY9RHwsbTDC4afWvLixzoQ.md
+++ b/changelog/TY9RHwsbTDC4afWvLixzoQ.md
@@ -1,0 +1,4 @@
+audience: developers
+level: patch
+---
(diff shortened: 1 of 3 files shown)


lotas (approved):

Eijebong:
Yeah the issue with putting every service in there is mostly the github one since it requires extra configuration.

lotas:
thanks for the fix! Merging it

PR #5569, opened 2022-07-18
Create a new 'public_restricted' pullRequestPolicy

ahal:
This PR implements [RFC 175](https://github.com/taskcluster/taskcluster-rfcs/blob/main/rfcs/0175-restricted-pull-requests.md)

### DIFF
--- a/services/github/test/handler_test.js
+++ b/services/github/test/handler_test.js
@@ -252,9 +252,22 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
         handlers.handlerComplete = resolve;
         handlers.handlerRejected = reject;
       });
+      const [eventBase, eventAction] = eventType.split('.');
 
-      let body = {
-        pull_request: {
+      let body = {};
+      let details = {
+        'event.type': eventType,
+        'event.base.repo.branch': 'tc-gh-tests',
+        'event.head.repo.branch': 'tc-gh-tests',
+        'event.head.user.login': user,
+        'event.head.repo.url': 'https://github.com/TaskclusterRobot/hooks-testing.git',
+        'event.head.sha': head || '03e9577bc1ec60f2ff0929d5f1554de36b8f48cf',
+        'event.head.ref': 'refs/heads/tc-gh-tests',
+        'event.base.sha': base || '2bad4edf90e7d4fb4643456a4df333da348bbed4',
+        'event.head.user.id': 190790,
+      };
+      if (eventBase === 'pull_request') {
+        body['pull_request'] = {
           user: {
             login: opener || user || 'octocat',
           },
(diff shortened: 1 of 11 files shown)


ahal:
Thought creating a "draft PR" would not request review... This is not ready for review (but feel free to take a look and comment if you like)

ahal:
Note: Ticking `Hide Whitespace` makes the review a lot easier :)

ahal:
Also if someone could merge https://github.com/taskcluster/taskcluster-rfcs/pull/175 that would be appreciated :)

ahal:
Note for reviewers, the diff is easier to read with the "Hide whitespace" option checked.

ahal:
@petemoore @matt-boris @lotas review ping :)

services/github/test/handler_test.js
```diff
@@ -407,6 +407,83 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
+      let args = handlers.createTasks.firstCall.args[0];
+      let taskGroupId = args.tasks[0].task.taskGroupId;
+      let [build] = await helper.db.fns.get_github_build(taskGroupId);
+      assert.equal(build.organization, 'TaskclusterRobot');
+      assert.equal(build.repository, 'hooks-testing');
+      assert.equal(build.sha, '03e9577bc1ec60f2ff0929d5f1554de36b8f48cf');
+      assert.equal(build.state, 'pending');
+    });
+
+    test('valid pull_request (user is not a collaborator, policy is public_restricted) creates a taskGroup', async function() {
```
lotas, line 458: It looks like both tests are identical with only change being `tcyaml['policy']` value?
And if I understand correctly, for the `public_restricted`, `createTasks()` would be called with different role, is it possible to add this check also? Or scopes do not go through the `createTasks` call directly?

lotas (dismissed):
Solid job! :) Thanks @ahal

services/github/test/handler_test.js
ahal: Yeah good call, I'll add assertions for the scopes!

And yes the tests are basically the same. I tend to prefer duplicating tests so that you can tell at a glance what's broken and what's not, but if you like I could try refactoring them to reduce duplication.

ahal: Sounds good, turns out the `simulateJobMessage` helper wasn't setting the `payload.tasks_for` property, and once I add that some other tests start failing. Trying to dig into why and whether those failures are legit bugs or not.

ahal: Tracked it down, turns out setting the `tasks_for` just unlocked a code path that needed an additional mocked property in the event body.

lotas (approved):
:+1:

ahal:
For posterity, I also tested this by hooking up a live Github app to my dev instance and opening PRs on a real repo (both as collaborator and as public). Everything seemed to work as expected, though I can't claim to have manually tested all edge cases :).

lotas:
Thanks! Let's merge it then 🎉

PR #5571, opened 2022-07-20
chore(k8s): Spread cron times to avoid CPU spikes and increased db loads

lotas:
Fixes bug [1633440](https://bugzilla.mozilla.org/show_bug.cgi?id=1633440)

### DIFF
--- a/changelog/bug-1633440.md
+++ b/changelog/bug-1633440.md
@@ -1,0 +1,6 @@
+audience: deployers
+level: patch
+reference: bug 1633440
+---
+
+Spread cron task times that started at 00:00 to minimize CPU spikes and DB loads.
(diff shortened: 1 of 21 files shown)


matt-boris (dismissed):
👍🏻 ty!

matt-boris (dismissed):

djmitche (approved):
The frequency of these operations is (vaugely) important, but the precise time isn't -- even for credental expiration.  The queries for those check the `expires` column, so practically speaking expiration is accurate to the second.  Purging the expired data is just a way of keeping the table sizes reasonable.

lotas:
Forgot to document the decision process here,
So the ones that deal with auth and credentials are still running at 00:00, the rest are spread a bit.
I've checked last runs and their durations and most of them are quick jobs, running 10s - 100s at most

PR #5578, opened 2022-07-25
feat(cd): use cloudbuild for CD to a dev env

### LINKED ISSUE #5553: Deploy built image after cloudbuild to alpha env
Add extra step in  `cloudbuild.yaml` to automatically deploy new image to the GKE cluster.
We can create new namespace `dev` and link it to `dev.alpha.taskcluster-dev.net` domain where we can deploy our images that were built from `main` branch

matt-boris:
Fixes #5553.

> This change adds continuous deployment support to the `cloudbuild.yaml` file so that each change to `main` results in a new deployment to [`https://dev.alpha.taskcluster-dev.net/`](https://dev.alpha.taskcluster-dev.net/).

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -40,7 +40,7 @@ options:
   dynamicSubstitutions: true
 substitutions:
   _NODE_VERSION: 16.16.0
-  _VERSION: '{"version":"${BRANCH_NAME}_dev","commit":"${SHORT_SHA}","source":"https://github.com/taskcluster/taskcluster","build":"${BUILD_ID}"}'
+  _VERSION: '{"version":"${BRANCH_NAME}_${SHORT_SHA}","commit":"${SHORT_SHA}","source":"https://github.com/taskcluster/taskcluster","build":"${BUILD_ID}"}'
   _IMAGE_NAME: gcr.io/${PROJECT_ID}/${PROJECT_ID}/${BRANCH_NAME}
   _IMAGE_VERSION: latest
 availableSecrets:
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -3,16 +3,49 @@ steps:
     args:
       - build
       - '--build-arg'
-      - 'DOCKER_FLOW_VERSION=${_VERSION}'
+      - DOCKER_FLOW_VERSION=${_VERSION}
       - '-t'
-      - '${_IMAGE_NAME}'
+      - ${_IMAGE_NAME}:${_IMAGE_VERSION}
       - .
-timeout: 1500s
+    id: Build
+  - name: gcr.io/cloud-builders/docker
+    args:
+      - push
+      - ${_IMAGE_NAME}:${_IMAGE_VERSION}
+    id: Push
+  - name: bash
+    args:
+      - '-c'
+      - echo "$$DEV_CONFIG" > /workspace/dev-config.yml && echo "$$INGRESS" > /workspace/infrastructure/k8s/templates/ingress.yaml
+    id: Get secrets
+    secretEnv:
+      - DEV_CONFIG
+      - INGRESS
+  - name: node:${_NODE_VERSION}
+    args:
+      - '-c'
+      - curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash && curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" && install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl && echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add - && apt-get update -y && apt-get install google-cloud-cli google-cloud-sdk-gke-gcloud-auth-plugin -y && gcloud container clusters get-credentials ${PROJECT_ID} --region us-east1 --project ${PROJECT_ID} && yarn && yarn dev:db:upgrade && yarn dev:apply
+    id: Deploy
+    entrypoint: bash
+  - name: gcr.io/cloud-builders/gcloud
+    args:
+      - '-c'
+      - gcloud container images list-tags ${_IMAGE_NAME} --filter='-tags:*' --format='get(digest)' --limit=unlimited | xargs -I {sha} gcloud container images delete "${_IMAGE_NAME}@{sha}"
+    id: Delete old images
+    entrypoint: bash
+timeout: 1800s
 images:
-  - '${_IMAGE_NAME}'
+  - ${_IMAGE_NAME}:${_IMAGE_VERSION}
 options:
   dynamicSubstitutions: true
 substitutions:
-  _IMAGE_NAME: 'gcr.io/taskcluster-dev/${PROJECT_ID}/${BRANCH_NAME}:latest'
-  _VERSION: >-
-    {"version":"${BRANCH_NAME}_dev","commit":"${SHORT_SHA}","source":"https://github.com/taskcluster/taskcluster","build":"${BUILD_ID}"}
+  _NODE_VERSION: 16.16.0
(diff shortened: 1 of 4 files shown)


cloudbuild.yaml
```diff
@@ -3,16 +3,49 @@ steps:
...
 options:
   dynamicSubstitutions: true
 substitutions:
-  _IMAGE_NAME: 'gcr.io/taskcluster-dev/${PROJECT_ID}/${BRANCH_NAME}:latest'
-  _VERSION: >-
-    {"version":"${BRANCH_NAME}_dev","commit":"${SHORT_SHA}","source":"https://github.com/taskcluster/taskcluster","build":"${BUILD_ID}"}
+  _NODE_VERSION: 16.16.0
+  _VERSION: '{"version":"${BRANCH_NAME}_dev","commit":"${SHORT_SHA}","source":"https://github.com/taskcluster/taskcluster","build":"${BUILD_ID}"}'
+  _IMAGE_NAME: gcr.io/${PROJECT_ID}/${PROJECT_ID}/${BRANCH_NAME}
+  _IMAGE_VERSION: latest
```
petemoore, line 45: Could this introduce a race condition where, e.g. two commits land on main in a short time, and the older one ends up getting deployed because the cloud build for the older one pushes the docker images to docker hub after the newer cloud build, and then the last images to get deployed are the older ones?
Or will previous jobs be cancelled, if a new commit lands?
Are there other races that could happen that I haven't thought of?
One option I can imagine is that we are explicit with version and refer to e.g. the git commit sha in the docker image version. This might not solve the races, but at least we can inspect the cluster to see which version of the images are deployed in the containers. However, we'd also then need to clean up the docker registry and not keep images around for too long, so it becomes more complicated.

So long story short, I think this could work for most of the time, but I'm concerned we could intermittently end up with the wrong version of something deployed, and perhaps a more complicated handling of version number may mitigate the issue, albeit introducing more complexity.

petemoore (commented):
Not sure how important/relevant my observation is, so keen to hear your assessment of the possible race conditions...

cloudbuild.yaml
lotas: this can indeed produce such scenario, but yeah, it will not happen that often. Plus, we can always trigger this build manually to rebuild the latest or just redeploy latest.

matt-boris: I can edited our [Cloud Build API Concurrent Builds quota](https://console.cloud.google.com/iam-admin/quotas?service=cloudbuild.googleapis.com&project=taskcluster-dev) from 10 to 1, so that additional jobs will queue instead of running in parallel.

I think it's a pretty solid solution to this potential problem! :) Good callout Pete

https://stackoverflow.com/a/63120998

cloudbuild.yaml
```diff
@@ -3,16 +3,49 @@ steps:
...
+    args:
+      - push
+      - ${_IMAGE_NAME}:${_IMAGE_VERSION}
+    id: Push
+  - name: bash
+    args:
+      - '-c'
+      - echo "$$DEV_CONFIG" > /workspace/dev-config.yml && echo "$$INGRESS" > /workspace/infrastructure/k8s/templates/ingress.yaml
+    id: Get secrets
+    secretEnv:
```
lotas, line 21: Can you please leave somewhere in Readme a not about how to properly set up this env at least manually? In case we would need to reproduce it or someone else would like to run their own cloud build? 
Thanks

cloudbuild.yaml
petemoore: I guess we can leave it as is, and change it if it becomes a problem. My only hesitation is that we might not always know exactly which version got deployed, e.g. the version number here is `main_dev` rather than e.g. a commit hash.

Screenshot 2022-07-26 at 15 14 51

On the plus side, we tend to get a lot of updates due to renovate/dependabot/Synk etc, so even if we do "lose a commit" from time to time, probably it will be resolved relatively quickly as the dependency update PRs tend to land quite frequently.

Is there any way to get that `main_dev` reference to change to an actual commit sha though?

petemoore: > I can edit our [Cloud Build API Concurrent Builds quota](https://console.cloud.google.com/iam-admin/quotas?service=cloudbuild.googleapis.com&project=taskcluster-dev) from 10 to 1, so that additional jobs will queue instead of running in parallel.
> 
> I think it's a pretty solid solution to this potential problem! :) Good callout Pete
> 
> https://stackoverflow.com/a/63120998

Ah nice! Do you know if this can be done in combination with auto-cancelling previous jobs that are already running, just in case e.g. several commits land in sequence, and then we need to wait for the full deployment cycle to run for each of them.

cloudbuild.yaml
```diff
@@ -3,16 +3,49 @@ steps:
...
+  - name: node:${_NODE_VERSION}
+    args:
+      - '-c'
+      - curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash && curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" && install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl && echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add - && apt-get update -y && apt-get install google-cloud-cli google-cloud-sdk-gke-gcloud-auth-plugin -y && gcloud container clusters get-credentials ${PROJECT_ID} --region us-east1 --project ${PROJECT_ID} && yarn && yarn dev:db:upgrade && yarn dev:apply
+    id: Deploy
+    entrypoint: bash
+  - name: gcr.io/cloud-builders/gcloud
+    args:
+      - '-c'
+      - gcloud container images list-tags ${_IMAGE_NAME} --filter='-tags:*' --format='get(digest)' --limit=unlimited | xargs -I {sha} gcloud container images delete "${_IMAGE_NAME}@{sha}"
```
lotas, line 33: hmm. how does this work? Is it excluding currently built one? Or registry removes some tags from older images, so only those are being listed?

cloudbuild.yaml
matt-boris: > Is there any way to get that `main_dev` reference to change to an actual commit sha though?

I could definitely change the `dev` suffix to be a short sha.

> Do you know if this can be done in combination with auto-cancelling previous jobs that are already running, just in case e.g. several commits land in sequence, and then we need to wait for the full deployment cycle to run for each of them?

This I'm not too sure of. A polling step at the beginning that uses the `gcloud` cli could probably be used for this, but I feel like it's a non-issue for now. I think we should wait and see how this works for us. I think general case, we should have no issues.

cloudbuild.yaml
matt-boris: Registry only ever has one `latest` tag and this is filtering on images with _no_ tags (so, not `latest`). It'll get that list and delete all of them.

cloudbuild.yaml
matt-boris: https://github.com/taskcluster/taskcluster/pull/5578/commits/fd47dc069d7ae6ec3d27de7853289fa04480c215#diff-5b9888d738c268a2fca37f35bd3ddf988ebcac318b536ea2be2b45a3ea676f32R356-R361

cloudbuild.yaml
matt-boris: @petemoore changed the version to be `main_<short SHA>` in https://github.com/taskcluster/taskcluster/pull/5578/commits/fd47dc069d7ae6ec3d27de7853289fa04480c215#diff-0b4a5fc26445836e6672c426692297004cfaba21997343b63e6f58b946bd2a5cR43

lotas (approved):
nice! Thanks

petemoore (commented):
Is it also possible change the name from `dev` to e.g. "Continuous Deployment" here:
Screenshot 2022-07-26 at 15 12 58
Also, is it possible to change the description to something like "This environment is deployed by Google Cloud Build every time a there is a push to github.com/taskcluster/taskcluster main branch" ? I think the description is allowed to be markup, so that link could even be a real hyperlink, I believe.

Perhaps the change to do this isn't part of this PR but is part of the configuration of google cloud build?

Do we need a document that explains how to change config such as this, or how to access the google cloud build configuration, or is it reasonably self explanatory?

PR #5581, opened 2022-07-26
fix(db): fix quarantine filter for static workers

### LINKED ISSUE #5555: The quarantine "filter by" doesn't seem to work for static workers
I have a macos worker that is quarantined but the `filter by` filter for it doesn't seem to be working
quarantine1
-> 
quarantine2

matt-boris:
Fixes #5555.

> This patch fixes an issue with filtering workers based on quarantined status. The issue only occurs with static workers that are quarantined. When the filter was active, those static, quarantined workers would not be displayed in the list. This issue was first brought up in v44.17.0.


lotas (approved):
:+1: looks good
but probably needs some testing on the environment with some workers

matt-boris:
> but probably needs some testing on the environment with some workers

Exactly what this is waiting for @lotas

PR #5582, opened 2022-07-27
feat(cd): run `yarn smoketest` after deploy to dev

matt-boris:
> This patch makes it so that a `yarn smoketest` on our dev environment is run after a successful deploy.


lotas (approved):
nice!

PR #5584, opened 2022-07-28
feat(dev): Autoreload services in docker compose dev mode

lotas:
Add ability to automatically restart services when source code changes.

`nodemon` is used

* This will not work until new version is released and build, as `nodemon` is expected to be inside main dependencies


infrastructure/tooling/src/generate/generators/docker-compose.js
```diff
@@ -445,11 +453,16 @@ exports.tasks.push({
...
+          volumes: [
+            ...devVolumes,
+            `./services/${name}:/app/services/${name}`, // service should only care about own code
+          ],
           environment: {
             LEVEL: 'info',
             DEBUG: 'taskcluster:*',
           },
+          command: '',
+          entrypoint: `/bin/sh -c "${procs[proc].command.replace(/^node /, `npx nodemon ${nodemonFlags(name)} `)}"`,
```
lotas, line 465: This is arguable.. We can also change the base image and always install `nodemon` with dependencies, but it will increase slightly image and will add +1 dependency to the tree.
Otoh `npx nodemon` works perfectly by fetching, caching and running it, only adding small delay at first start.

Instead of `/entrypoint.sh` we use command directly and just replace `node` with `nodemon`

infrastructure/tooling/src/generate/generators/docker-compose.js
```diff
@@ -445,11 +453,16 @@ exports.tasks.push({
         }
 
         dockerCompose.services[`${name}-${proc}`] = {
-          volumes: devVolumes,
+          volumes: [
+            ...devVolumes,
+            `./services/${name}:/app/services/${name}`, // service should only care about own code
```
lotas, line 458: this is to prevent all services from restarting at the same time when you change something for one service only

lotas:
Changes are applied quite fast, but it comes with a short down time during restart, couple of seconds. For most of the services it's okay, but for some like auth/queue, which are central to other services, it might lead to some temp error messages in in console. It goes back to normal within 5-10s

lotas:
Looks like this dumb approach to live-reload all services at once comes with extra cost.. which are resources. In `nodemon` mode each service consumes x2-x3 times more memory and takes much longer to start, which make some healthckecks fail on machine with lower resources.
It is also unlikely that someone will want to live-develop all services at once. I will experiment with compose profiles to start dev containers only for the services that are being developed on.

docker-compose.dev.yml
```diff
@@ -1,13 +1,11 @@
 x-autogenerated: This file is autogenerated
-version: '3'
```
lotas, line 2: dropping version, as it was deprecated already [docs](https://docs.docker.com/compose/compose-file/#compose-file)

dev-docs/development-process.md
```diff
@@ -202,20 +202,21 @@ It's easy: `yarn build` in the root directory.
 
 ## Running everything locally using `docker compose`
 
+> Note: you'll need recent version of `docker` with [`docker-compose-plugin`](https://docs.docker.com/compose/install/compose-plugin/) |(or alternatively [docker-compose](https://github.com/docker/compose/releases))
+>
+> For non-linux systems it is advised to run docker with at least 4 CPU cores and at least 6 Gb of memory.
```
matt-boris, line 207: nit
```suggestion
> For non-linux systems it is advised to run docker with at least 4 CPU cores and at least 6 GB of memory.
```

dev-docs/development-process.md
```diff
@@ -242,6 +243,37 @@ To avoid having HAWK authentication issues and be able to see worker logs, modif
...
+> Note: starting all services in developer mode might be slower and will require extra resources. As alternative you can only start some of the services in development mode:
+
+```sh
+# first run everything
+docker compose up -d
+
+# then start some of the services you want to develop
+docker compose -f docker-compose.yml -f docker-compose.dev.yml up -d queue-web auth-web
+
+# now queue-web and auth-web are started with nodemon and will be live-reloaded when source file will change
```
matt-boris, line 272: Nice, I like this 👍🏻

infrastructure/tooling/src/generate/generators/docker-compose.js
```diff
@@ -427,12 +424,22 @@ exports.tasks.push({
             './generated:/app/generated',
             '.all-contributorsrc:/app/.all-contributorsrc',
             './ui:/app/ui',
-            'ui_node_modules:/app/ui/node_modules/', // prevent previous mount to overwrite modules inside container
```
matt-boris, line 430: No longer needed?

matt-boris (commented):
Nice! Just left a couple comments/questions/suggestions 👍🏻

dev-docs/development-process.md
lotas: Thanks! :)

matt-boris (approved):

package.json
```diff
@@ -5,7 +5,7 @@
   "engine-strict": true,
   "engines": {
     "node": "16.16.0",
-    "yarn": "^1.22.19"
+    "yarn": "^1.22.10"
```
matt-boris, line 8: This looks to have been accidentally downgraded

lotas: ouch, .. was reverting nodemon, thanks for spotting this!

matt-boris (approved):
Nice! 🚢

PR #5603, opened 2022-08-03
Introduce compose profiles

### LINKED ISSUE #5602: Docker compose: include all background jobs under different profile(s): https://docs.docker.com/compose/profiles/
TODO: include all background jobs from procs to the same `docker-compose.yml` with different profiles (either single `background` profile for everything, or independent `${serviceName}-jobs` individual to allow fine-grained control of what to start.

lotas:
Compose profiles allow to run background jobs that are not started by default.

Usage documentation: [development-process.md: using compose profiles](https://github.com/taskcluster/taskcluster/blob/66c44af97c507fc5c1a6b2d70a5dcd472e056b0b/dev-docs/development-process.md#using-compose-profiles)

Refactored docker-compose.yml to load environment variables for services from dedicated env files.
This was done because there are multiple services that share same set of environment variables (web+background+cron)

Fixes #5602


dev-docs/development-process.md
```diff
@@ -275,6 +275,47 @@ yarn dev:start queue-web auth-web
...
+# stop all
+yarn stop
+
+# start only one specific background job by name
+yarn start notify-background-handler
+```
+
+> Note: although `docker compose` command allows passing profile names through arguments like `docker compose --profile cron up -d`, those services will not be stopped if you run `docker compose down` afterwards. To stop them you would need to use same arguments `docker compose --profile cron down`.
+>
+> This way using exported environment variable makes compose behaviour more intuitive.
```
lotas, line 317: this is probably arguable, because you need to keep in mind that this variable controls what will be started and stopped.. :shrug:

infrastructure/tooling/src/generate/generators/docker-compose.js
```diff
@@ -58,7 +59,7 @@ const defaultValues = {
   NODE_ENV: 'development',
 
   DEBUG: '',
-  LEVEL: 'warning',
+  LEVEL: 'info',
```
lotas, line 62: Switching again to a slightly more verbose mode 🙈 Having some logs is better than no logs, as it feels like nothing is happening

infrastructure/tooling/src/generate/generators/docker-compose.js
```diff
@@ -143,13 +154,18 @@ const uiConfig = [
...
-  title: `Generate docker-compose.yml`,
+  title: `Generate docker-compose files`,
   requires: [
     ...SERVICES.map(name => `configs-${name}`),
     ...SERVICES.map(name => `procslist-${name}`),
     'target-nginx.conf',
   ],
-  provides: ['target-docker-compose.yml'],
+  provides: [
+    'target-docker-compose.yml',
```
lotas, line 164: combined three targets into one, as it includes now more logic around which services should be included. Easier to manage in single iteration

matt-boris: Nice! Solid cleanup too

PR #5606, opened 2022-08-04
Use devel image for docker-compose.dev.yml

lotas:
* switched to `-devel` image in `.dev.yml` compose
* added `tc-admin` script to provision local instance of taskcluster (default roles/pool definition). Also serves as example 
* `yarn dev:*` commands conditionally include `docker-compose.override.yml` if it exists. Turned out, if you run compoes with multiple configs, it will not autoload the override file.


matt-boris (dismissed):
Nice 👍🏻 🚢

matt-boris (approved):

PR #5612, opened 2022-08-08
feat(github): Allow to skip CI for some commits.

### LINKED ISSUE #5311: Add a way to skip CI with github integration when commits contain `[ci skip]`
Most CI systems have that, it'd be nice to have for the github integration. I could implement that in my decision task of course but that would still require workers to run that and would change the CI status.

lotas:
For a single commit push event with a message that contains `"[skip ci]"` or `"[ci skip]"` no tasks would be created.

Fixes #5311


lotas:
This is currently implemented for the "push" events with a single commit in it only.

I'm not sure if we should also support pull requests with titles that includes "[ci skip]" keyword.

petemoore:
I like it.
It occurs to me it might(?) already be possible to implement this in the .taskcluster.yml via json-e, if there is a way to check if the commit message contains the allowed skip terms via json-e, I guess it could return an empty list of tasks. But I've no idea how that would look in json-e and I think your solution is much easier for users.

One thing I'd like to call out is that in general, interpretation of commit messages is likely to be project-specific, so I think taskgraph is a better place to interpret directives in commit messages for anything else we add, but for the one and only case of skipping tasks, it indeed needs to be either json-e or the taskcluster github service that does this, since taskgraph requires that a task is already running.

petemoore (approved):
Love it, thanks for adding the tests too!

PR #5613, opened 2022-08-10
Add support for ingress nginx and certbot to ingress.yml

### LINKED ISSUE #4913: Current ingress YAML does not work with nginx ingress
Currently, `infrastructure/k8s/templates/ingress.yaml` uses `/*` in all its paths. However, nginx ingress, or at least the nginx ingress addon for minikube does not support this. The fix is to remove all the `*` and then it will work instead of 404.

lotas:
This introduces few changes:

* `ingress.yml` can be configured through `dev-config.yml` option `ingressType: nginx` to generate proper routes (`/` instead of `/*`)
* `ingress.yml` can include cert-manager annotations to provision certificates automatically
* Improved `dev-deployment.yml` to include information on how to use those options

Fixes #4913


entrypoint
```diff
@@ -87,6 +87,7 @@ script/dev:db:downgrade) shift; exec yarn run dev:db:downgrade "${@}";;
 script/dev:apply) shift; exec yarn run dev:apply "${@}";;
 script/dev:delete) shift; exec yarn run dev:delete "${@}";;
 script/dev:verify) shift; exec yarn run dev:verify "${@}";;
+script/dev:templates) shift; exec yarn run dev:templates "${@}";;
```
lotas, line 90: this is to dump k8s templates with substituted variables into a file for debug purposes

infrastructure/tooling/src/dev/helm.js
```diff
@@ -111,14 +115,23 @@ const actions = [
     provides: ['target-verify'],
     run: async (requirements, utils) => ({
       'target-verify': await execCommand({
-        command: ['kubectl', 'apply', '--dry-run', '-f', '-'],
+        command: ['kubectl', 'apply', '--dry-run=client', '-f', '-'],
```
lotas, line 118: this is the preferred way now, old was deprecated

infrastructure/tooling/templates/k8s/ingress.yaml
```diff
@@ -6,14 +6,24 @@ metadata:
   annotations:
     'kubernetes.io/ingress.global-static-ip-name': '{{ .Values.ingressStaticIpName }}'
     'ingress.gcp.kubernetes.io/pre-shared-cert': '{{ .Values.ingressCertName }}'
+    'cert-manager.io/cluster-issuer': '{{ .Values.certManagerClusterIssuerName | default "" }}'
```
lotas, line 9: those extra annotations will do no harm when left blank to any existing deployment.

Plus mozilla is overriding those for moz-specific deployments

petemoore (approved):
Looks great, many thanks Yarik!

@Eijebong Do you have anything to add?

Eijebong (approved):
Got nothing to add, this looks good to me.

PR #5615, opened 2022-08-11
Remove hardcoded ingress, use latest ingress nginx support.

lotas:
Cloudbuild improvements:

* using #5613 ingress nginx support to replace hardcoded ingress.yml
* Adding intermediate deployment image for faster deployments - will be built first time and then reused in later builds.
* Switching to higher CPU machines for faster builds (suggested by google cloud)

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,5 +1,11 @@
 steps:
   - name: gcr.io/cloud-builders/docker
+    id: Prefetch deploy image
+    entrypoint: bash
+    args:
+      - '-c'
+      - docker pull ${_DEPLOY_IMAGE_NAME}:latest || exit 0
+  - name: gcr.io/cloud-builders/docker
     args:
       - build
       - '--build-arg'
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -36,16 +58,17 @@ steps:
 timeout: 1800s
 images:
   - ${_IMAGE_NAME}:${_IMAGE_VERSION}
+  - ${_DEPLOY_IMAGE_NAME}:${_IMAGE_VERSION}
```
lotas, line 61: images defined here will be automatically pushed to registry, so no need to additionally use `docker push`

matt-boris (approved):
Nice, thanks! Only small thought I have is to maybe pull out the inline Dockerfile into its own file. Up to you though, as this works perfectly fine.

lotas:
> Nice, thanks! Only small thought I have is to maybe pull out the inline Dockerfile into its own file. Up to you though, as this works perfectly fine.

I thought about it yes, but I decided that this Dockerfile doesn't belong outside of this file, not used anywhere else, so not worth it

PR #5624, opened 2022-08-16
Remove duplicate ingress paths

lotas:
Google load balancer uses Url Maps, and having wildcard in path is mandatory to match all sub-paths.
Default k8s ingress and nginx-ingress do not use wildcard in paths, so it should be removed.

Tested on dev

### DIFF
--- a/services/web-server/procs.yml
+++ b/services/web-server/procs.yml
@@ -3,7 +3,6 @@ web:
   command: node services/web-server/src/main.js server
   readinessPath: .well-known/apollo/server-health
   paths:
-    - '/login'
     - '/login/*'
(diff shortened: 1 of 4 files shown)


services/web-server/procs.yml
```diff
@@ -3,7 +3,6 @@ web:
   command: node services/web-server/src/main.js server
   readinessPath: .well-known/apollo/server-health
   paths:
-    - '/login'
     - '/login/*'
```
lotas, line 6: this should be safe, as I couldn't find any references for the strict match, all routes are only `/login/something`

matt-boris (approved):

PR #5625, opened 2022-08-17
UI: reload task on updates, show recent run on rerun

### LINKED ISSUE #5046: When rerunning an action, it would be nice if the UI would automatically show the newly created run
Right now it creates the run but doesn't switch to it, requiring me to click on the right arrow. Would be nice if it did that on its own

lotas:
* UI: Task page: subscribe to task real-time updates to display latest status and run information
* UI: Task page: automatically show latest run after triggering "Rerun" 
* Docker compose: only run `standalone` generic worker on start, `static` can be run with `COMPOSE_PROFILES=workers`
* UI: Pulse messages: autocomplete known exchanges + submit on enter

image

This should help quickly subscribe to known exchanges without thinking  long for the correct format of the exchange name. It only fetches exchange names of the taskcluster services

Github Bug/Issue: Fixes #5046

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -189,6 +189,8 @@ services:
       TASKCLUSTER_ROOT_URL: http://taskcluster
       TASKCLUSTER_CLIENT_ID: static/generic-worker-compose-client
       TASKCLUSTER_ACCESS_TOKEN: j2Z6zW2QSLehailBXlosdw9e2Ti8R_Qh2M4buAEQfsMA
+    profiles:
+      - workers
     depends_on:
       rabbitmq:
         condition: service_healthy
--- a/ui/src/views/Tasks/ViewTask/index.jsx
+++ b/ui/src/views/Tasks/ViewTask/index.jsx
@@ -54,6 +54,7 @@ import { nice } from '../../../utils/slugid';
 import Link from '../../../utils/Link';
 import submitTaskAction from '../submitTaskAction';
 import taskQuery from './task.graphql';
+import taskSubscription from './taskSubscription.graphql';
 import scheduleTaskQuery from './scheduleTask.graphql';
 import rerunTaskQuery from './rerunTask.graphql';
 import cancelTaskQuery from './cancelTask.graphql';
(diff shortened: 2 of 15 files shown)


docker-compose.yml
```diff
@@ -189,6 +189,8 @@ services:
       TASKCLUSTER_ROOT_URL: http://taskcluster
       TASKCLUSTER_CLIENT_ID: static/generic-worker-compose-client
       TASKCLUSTER_ACCESS_TOKEN: j2Z6zW2QSLehailBXlosdw9e2Ti8R_Qh2M4buAEQfsMA
+    profiles:
```
lotas, line 192: I've added profile for static worker, so it will not launch with rest of the services. 
We have one generic worker (standalone). One is enough

lotas:
wanted to go wild on the pulse page, but will keep it for later :) 
We can also show exchange descriptions, or have a dictionary of all available ones

ui/src/views/Tasks/ViewTask/index.jsx
```diff
@@ -100,7 +101,7 @@ const getCachesFromTask = task =>
 @graphql(taskQuery, {
   options: props => ({
     fetchPolicy: 'network-only',
-    pollInterval: TASK_POLL_INTERVAL,
+    pollInterval: TASK_POLL_INTERVAL * 10,
```
matt-boris, line 104: `TASK_POLL_INTERVAL` was updated to be 60,000ms (60s). Multiplying by 10 here brings this polling interval up to 10min. Before, this interval was 30s.

My guess is this `* 10` can be removed. And why was the `TASK_POLL_INTERVAL` increased from 30s to 60s?

matt-boris (commented):
Thanks for splitting your work out into separate commits too, made for easier reviewing! 🎉 Looks good! Just left a quick comment/question.

ui/src/views/Tasks/ViewTask/index.jsx
lotas: yes! Well spotted.. Oh, `10` was a debug one.. will remove.
In fact, we don't really need it much, since there is a subscribe method listening for updates. So I thought that once a minute would be more than enough, and then forgot about that multiplier. Will drop it. Thanks

matt-boris (approved):

PR #5628, opened 2022-08-18
Supress output of dev:apply as it outputs raw secrets, cleanup old pods

lotas:


matt-boris (approved):
LGTM, thanks! This may allow us to forward the build logs to GitHub too, which is a nice perk!

PR #5636, opened 2022-08-23
Issue 5634: build livelog docker image FROM scratch

### LINKED ISSUE #5634: Build go docker images "from scratch"
**Describe the bug**
We used to build docker images using `FROM scratch` so that the docker image contained just the self-contained statically linked standalone go binary that it was for, plus any additional files it required, such as a certs file, or an empty `/tmp` directory.

It looks like this changed for livelog in #3866. In that case we probably are just missing the empty /tmp directory and/or the certs file.

Let's check all the docker images we create that are shipped as docker images, and see if we can return to the leaner containers containing only their actual runtime dependencies.

petemoore:
Github Bug/Issue: Fixes #5634

I'm not really sure why the Dockerfile is generated on the fly, rather than being a static file in the repo.

I've also spotted that `livelog` is built _outside_ of the docker container, and then copied in. I'm not sure why this is. This won't work for multiarch builds, so I think it would make more sense to build the go app inside the docker container in the first stage, and copy it into the second stage.

I can't see how to reference the go version in the Dockerfile so I can have something like:

```
FROM golang:1.18.5
```

in the first stage, where the version number comes from the `.go-version` file. There is [go-version.js](https://github.com/taskcluster/taskcluster/tree/from-scratch-again/infrastructure/tooling/src/generate/generators/go-version.js) but I'm not sure that currently sets any variable that can be referenced from the context of [livelog.js](https://github.com/taskcluster/taskcluster/tree/from-scratch-again/infrastructure/tooling/src/build/tasks/livelog.js) but I'm looking into it. Therefore keeping this as draft for now, so I can fix that too.

### DIFF
--- a/changelog/issue-5634.md
+++ b/changelog/issue-5634.md
@@ -1,0 +1,7 @@
+audience: worker-deployers
+level: patch
+reference: issue 5634
+---
+The livelog docker image used by docker-worker now is not based on busybox, but
+contains only the livelog binary, /etc/ssl/certs/ca-certificates.crt and an
+empty /tmp directory. This effectively reverses the change from #3866.
--- a/infrastructure/tooling/src/build/tasks/livelog.js
+++ b/infrastructure/tooling/src/build/tasks/livelog.js
@@ -72,9 +72,15 @@ module.exports = ({ tasks, cmdOptions, credentials, baseDir, logsDir }) => {
       // this simple Dockerfile just packages the binary into a Docker image
       const dockerfile = path.join(contextDir, 'Dockerfile');
       fs.writeFileSync(dockerfile, [
-        'FROM busybox', // We use busybox rather than scratch to have `/tmp` and such things
+        'FROM ubuntu:latest AS certs',
+        'apt-get update',
+        'apt-get upgrade -y',
+        'apt-get install -y ca-certificates',
+        'FROM scratch',
         'EXPOSE 60023',
         'EXPOSE 60022',
+        'RUN mkdir /tmp'
+        'COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt',
         'COPY version.json /app/version.json',
         'COPY livelog /livelog',
         'ENTRYPOINT ["/livelog"]',
--- a/changelog/ejVAaJPFTT2p5sOMcqC-7A.md
+++ b/changelog/ejVAaJPFTT2p5sOMcqC-7A.md
@@ -1,0 +1,12 @@
+audience: general
+level: patch
+---
+The azure library has been removed that was used by the following (removed) endpoints:
+
+  * `auth.azureCredentials`
+  * `auth.azureTables`
+  * `auth.azureTablesSAS`
+  * `auth.azureContainers`
+  * `auth.azureContainersSAS`
+
+ which were removed in release .


infrastructure/tooling/src/build/tasks/livelog.js
```diff
@@ -72,9 +72,15 @@ module.exports = ({ tasks, cmdOptions, credentials, baseDir, logsDir }) => {
       // this simple Dockerfile just packages the binary into a Docker image
       const dockerfile = path.join(contextDir, 'Dockerfile');
       fs.writeFileSync(dockerfile, [
-        'FROM busybox', // We use busybox rather than scratch to have `/tmp` and such things
+        'FROM ubuntu:latest AS certs',
```
lotas, line 75: no latest please :)

infrastructure/tooling/src/build/tasks/livelog.js
```diff
@@ -72,9 +72,15 @@ module.exports = ({ tasks, cmdOptions, credentials, baseDir, logsDir }) => {
...
+        'apt-get update',
```
lotas, line 76: `RUN` missing?

infrastructure/tooling/src/build/tasks/livelog.js
petemoore: I think this is an exceptional case - the fetch is for root certs only, and with the certificate chains, I think we _always_ want the latest version. Agreed that normally we would want to pin, but the risk of pinning certs here is that they will become out-of-date, and we'll only notice one day in production when suddenly certs are missing or invalid.

infrastructure/tooling/src/build/tasks/livelog.js
petemoore: Nice spot!

infrastructure/tooling/src/build/tasks/livelog.js
lotas: ah, sure, we agreed that latest is fine for this :)

petemoore:
Note, this successfully build the docker image here: https://community-tc.services.mozilla.com/tasks/IWuo9Pt_TFGNeHpfyw13aA/runs/0/logs/debug-logs/livelog-docker-build.log

However, it was not published, so I'm not really sure how to test it.

I'm also not sure why we generate a Dockerfile on the fly.

I could try to modify that task to persist the docker image that is created, and then try to test that.

lotas (approved):

infrastructure/tooling/src/build/tasks/livelog.js
```diff
@@ -72,9 +72,16 @@ module.exports = ({ tasks, cmdOptions, credentials, baseDir, logsDir }) => {
       // this simple Dockerfile just packages the binary into a Docker image
       const dockerfile = path.join(contextDir, 'Dockerfile');
       fs.writeFileSync(dockerfile, [
-        'FROM busybox', // We use busybox rather than scratch to have `/tmp` and such things
+        'FROM ubuntu:latest AS certs',
+        'RUN apt-get update',
+        'RUN apt-get upgrade -y',
+        'RUN apt-get install -y ca-certificates',
+        'RUN mkdir /empty',
```
matt-boris, line 79: These _could be_ reduced to one `RUN` command to reduce the layers, but not too big of a deal here I don't think. Up to you!

changelog/issue-5634.md
```diff
@@ -0,0 +1,7 @@
+audience: worker-deployers
+level: patch
+reference: issue 5634
+---
+The livelog docker image used by docker-worker now is not based on busybox, but
+contains only the livelog binary, /etc/ssl/certs/ca-certificates.crt and an
+empty /tmp directory. This effectively reverses the change from #3866.
```
matt-boris, line 7:
```suggestion
empty /tmp directory. This effectively reverses the change from [#3866](https://github.com/taskcluster/taskcluster/pull/3866).
```

matt-boris (commented):
Looks good! Just left a couple small comments/suggestions 👍🏻

infrastructure/tooling/src/build/tasks/livelog.js
petemoore: we don't keep the image, so i don't think it makes a difference :-)

we throw it away, and create a new image from scratch :-)

changelog/issue-5634.md
petemoore: sorry, i saw this too late!

PR #5640, opened 2022-08-25
cloudbuild: Use commit SHA instead of latest to trigger redeployment

lotas:
Problem with using ":latest" is that deployment configuration doesn't change, so the pods are not being restarted

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -12,12 +12,15 @@ steps:
       - DOCKER_FLOW_VERSION=${_VERSION}
       - '-t'
       - ${_IMAGE_NAME}:${_IMAGE_VERSION}
+      - '-t'
+      - ${_IMAGE_NAME}:latest
       - .
     id: Build
   - name: gcr.io/cloud-builders/docker
     args:
       - push
-      - ${_IMAGE_NAME}:${_IMAGE_VERSION}
+      - '-a'
+      - ${_IMAGE_NAME}
     id: Push
   - name: gcr.io/cloud-builders/docker
     id: Build deploy image using cache from previous build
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -12,20 +12,23 @@ steps:
       - DOCKER_FLOW_VERSION=${_VERSION}
       - '-t'
       - ${_IMAGE_NAME}:${_IMAGE_VERSION}
+      - '-t'
+      - ${_IMAGE_NAME}:latest
```
lotas, line 16: adding `:latest` tag to make sure any dev can use `gcr.io/taskcluster-dev/taskcluster-dev/main:latest` for their needs

cloudbuild.yaml
```diff
@@ -12,20 +12,23 @@ steps:
...
       - .
     id: Build
   - name: gcr.io/cloud-builders/docker
     args:
       - push
-      - ${_IMAGE_NAME}:${_IMAGE_VERSION}
+      - '-a'
+      - ${_IMAGE_NAME}
```
lotas, line 23: pushing both tags

cloudbuild.yaml
```diff
@@ -12,20 +12,23 @@ steps:
...
     id: Push
   - name: gcr.io/cloud-builders/docker
     id: Build deploy image using cache from previous build
     entrypoint: bash
     args:
       - '-c'
       - |
-        docker build -t ${_DEPLOY_IMAGE_NAME}:${_IMAGE_VERSION} --cache-from ${_DEPLOY_IMAGE_NAME}:latest -<<EOF
+        docker build -t ${_DEPLOY_IMAGE_NAME}:latest --cache-from ${_DEPLOY_IMAGE_NAME}:latest -<<EOF
```
lotas, line 31: deploy helper image is always the same, so single `:latest` tag for it is enough

cloudbuild.yaml
```diff
@@ -78,8 +82,7 @@ steps:
       - TASKCLUSTER_ACCESS_TOKEN
 timeout: 1800s
 images:
-  - ${_IMAGE_NAME}:${_IMAGE_VERSION}
-  - ${_DEPLOY_IMAGE_NAME}:${_IMAGE_VERSION}
+  - ${_DEPLOY_IMAGE_NAME}:latest
```
lotas, line 85: we only allow gcloud build to push this deploy image automatically, as we are pushing main monoimage manually already

matt-boris (approved):
LGTM, thanks!

PR #5644, opened 2022-08-29
Fix hover bug

### LINKED ISSUE #5295: When hovering over a task in a group task, the background color should change for the whole row
Right now it only changes for half the row and on a large screen, with lots of jobs I sometimes struggle to associate the status with the name.
90262c84-f322-429d-8572-ed7f74f9ed83

ahmedDevs:
Fix incomplete background colour change when hovering over a row at the group task page.

Issue was causing the background colour to change only for half the row when hovering over a row at the group task page, which was causing confusion  for users with wider screens as to what row is being hovered. 

Github Bug/Issue: Fixes #5295

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -876,6 +876,15 @@
       ]
     },
     {
+      "login": "ahmedDevs",
+      "name": "Ahmed A.",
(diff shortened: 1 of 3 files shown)


lotas (approved):
Looks great! Thanks

PR #5647, opened 2022-08-31
docker buildx release fix

lotas:
[ci skip]


lotas:
I think we can abandon this idea for now. This only increases build complexity and doesn't have any current business need to have arm images.
We can get back to it once we have a real use for this

PR #5654, opened 2022-09-02
Fix github checks updates

### LINKED ISSUE #5653: Github integration fails to change status updates
Introduced in the latest release 44.19.1, github falls into some sort of race condition or uses incorrect check run to update status.
Example: https://github.com/mozilla-releng/build-mar/runs/8135486860 
Although task was resolved https://firefox-ci-tc.services.mozilla.com/tasks/X0wOnzqKTEmMEBYPqDpNbg 

Related to #5085

lotas:
Since 44.19.1 release github service started tracking additional task state changes, and this resulted in a race condition between "taskDefined" and "status" handlers where both of them would create new check run at the same time. Wrong check run would later get all status updates, while Github UI will be showing a different check run which didn't receive all the updates.

Fixes #5653 

Description of the issue:
task status handler was doing roughly this:
1. get `checkRun` from db
2. do several calls to queue to fetch several artifacts (this may take couple of seconds!)
3. create new `checkRun` if it didn't exist yet

And in the same time `taskDefined` would be creating initial check run when task was initially defined.

So depending on luck and network latency there might be several concurrent routines calling check run create and resulting in multiple check runs being created for the same task.

Fix here is to update status handler to group calls together and swap steps 1. and 2.

### DIFF
--- a/changelog/issue-5653.md
+++ b/changelog/issue-5653.md
@@ -1,0 +1,12 @@
+audience: general
+level: patch
+reference: issue 5653
+---
+Fix a bug with github status checks not being updated.
+
+In 44.19.1 release github service started tracking additional task
+state changes, and this resulted in a race condition between "taskDefined"
+and "status" handlers where both of them would create new check run at
+the same time. Wrong check run would later get all status updates, while
+Github UI will be showing a different check run which didn't receive all
+the updates.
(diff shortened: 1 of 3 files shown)


matt-boris (approved):

jgraham:
Thanks for fixing this! Do we know how soon the fix will be in prod (context: this is causing ongoing issues for web-platform-tests on the community tc instance).

matt-boris:
@jgraham we've released v44.20.2 to Community CI and that appears to have fixed the issue, sorry about that!

PR #5662, opened 2022-09-06
feat(github): Trim shell control sequences from logs sent to github

lotas:
Reduces the noise in then output


matt-boris (approved):
Nice! Thank you for cleaning this up

lotas:
> Nice! Thank you for cleaning this up

we discussed using ansi2html last week, but I've discovered that github will be stripping most of the html tags anyways.. so instead of increasing payload size, it's better to keep the minimum, so txt will be fine

PR #5665, opened 2022-09-07
Refactor status handler, remove taskDefined handler

lotas:
This resolves an issue where two different handlers were trying to check/create/update check_run at the same time, which led to several check_runs being created and status reports were being sent to the wrong one.

Now single status handler manages all state transitions including task definition, running, and all of the completion states.

Besides that, if multiple events try to update same `taskId`, it will be ensured that they are executed strictly in order to not overwrite future events or create multiple check runs.

This also removes `/rerequested` functionality for re-run tasks, as this feature is broken on github side, status remains forever completed no matter what. Instead, new check run is being created if task was rerun.


petemoore:
This looks good, but I'm wondering how best we can test the various types of race conditions that caused us pains before, aside from testing it in production. For example, could we deploy it to staging before releasing it, and then mirror production traffic to staging, to see how it handles things?

I'm guessing it is going to be pretty tricky to mock production like conditions without building a big framework to emulate heavy load etc.

lotas:
> This looks good, but I'm wondering how best we can test the various types of race conditions that caused us pains before, aside from testing it in production. For example, could we deploy it to staging before releasing it, and then mirror production traffic to staging, to see how it handles things?
> 
> I'm guessing it is going to be pretty tricky to mock production like conditions without building a big framework to emulate heavy load etc.

I will be testing it on dev.alpha right now, with the single [connected repo](https://github.com/taskcluster/tc-dev-integration-test) to confirm that refactoring still works.
Plus all unit tests still gives us a good signal that the expected functionality and outputs are still the same.

Goal of this refactoring were to avoid concurrent handlers trying to create/update check runs for a single task.
This was possible in the past (although extremely rare) because of two different handlers - `taskDefined` and `status`. And as we saw in [this](https://github.com/web-platform-tests/wpt/runs/8202440583) task, it happened because both of those handlers were creating check run just `0.005s` apart :) 
image

Now, both events will be sent to the same queue, and as a result only one event will be handled at a time. `PulseConsumer` is waiting for handler to finish before receiving new message from that queue. As long as we don't scale `github-worker` to multiple instances it should be fine ;) And if we do, we'd have to come up with some distributed state machines .. huh :)

services/github/src/handlers/status.js
```diff
@@ -115,69 +244,59 @@ async function statusHandler(message) {
...
+
+    if (liveLogText) {
+      // doesn't make sense to include link to live log if it was not fetched
+      githubCheck.output.addText(`[${CHECKLOGS_TEXT}](${taskLogUI(this.context.cfg.taskcluster.rootUrl, runId, taskId)})`);
+      githubCheck.output.addText(`\n\`\`\`bash\n${tailLog(liveLogText)}\n\`\`\``);
+    }
+
+    if (customCheckRunText) {
+      githubCheck.output.addText(customCheckRunText);
+    }
```
matt-boris, line 280: Will you please flip these two blocks? I got a note from @AaronMT about the ordering here and discussed with him that it makes sense to put the user's custom check run text above the log. Their team uses the custom check run text feature to generate info about the artifact and the build, etc. but now it's pushed all the way to the bottom of this checks page like here https://github.com/mozilla-mobile/fenix/pull/26793/checks?check_run_id=8230762755.

matt-boris (commented):
Lookin good!

services/github/src/handlers/status.js
lotas: oh! didn't know someone was using it :) 
Sure!

lotas: I also thought that hiding logs link when it's not there is a good thing, but then maybe we would keep it for consistency, and later improve UI to show any log (livelog <> live_backing.log issue)

matt-boris: > oh! didn't know someone was using it :)

I didn't know either! Their use case makes total sense though.



> but then maybe we would keep it for consistency, and later improve UI to show any log

👍🏻

changelog/ZuGEp6tWRWW4ED2sVXTkuQ.md
```diff
@@ -0,0 +1,9 @@
+audience: developers
+level: minor
+---
+
+Refactored github status checks handler to do handle task status transitions in single place.
+
+Previous implementaition relied on two handlers: taskDefined and statusChanged.
```
matt-boris, line 7:
```suggestion
Previous implementation relied on two handlers: taskDefined and statusChanged.
```

services/github/src/github-auth.js
```diff
@@ -4,6 +4,31 @@ const { retry } = require("@octokit/plugin-retry");
 
 const PluggedOctokit = Octokit.plugin(retry);
 
+const tokenCache = new Map();
+/**
+ * Github app access token have an expiration date of 1h
+ * https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#authenticating-as-an-installation
+ *
+ * As we run multiple handlers for the same repository, we can save a lot of resources by using cached token
+ * On average it takes 0.5-1s to get one.
```
matt-boris, line 13: Nice! 🎉

services/github/src/handlers/index.js
```diff
@@ -128,12 +125,8 @@ class Handlers {
...
     ];
 
-    // Listen for taskDefined event to create initial status on github
-    const taskBindings = [
-      queueEvents.taskDefined(`route.${this.context.cfg.app.checkTaskRoute}`),
-    ];
-
-    const callHandler = (name, handler) => message => {
+    // handler returned must be async, as timedHandler will not be able to time it correctly
+    const callHandler = (name, handler) => message =>
```
matt-boris, line 129: 🦅 👁️

matt-boris (commented):
Just a couple small suggestions! But overall - 💪🏻 💪🏻

changelog/ZuGEp6tWRWW4ED2sVXTkuQ.md
lotas: Thanks! :)

services/github/src/handlers/status.js
```diff
@@ -1,30 +1,221 @@
...
+  }
+
+  addText(text, appendText = '\n') {
+    this.text += `${text}${appendText}`;
+  }
+
+  /**
+   * Github has a limit of 64Kb for the whole payload
+   */
+  getRemainingMaxSize() {
```
lotas, line 62: this will ensure that if someone includes custom annotations, we will consider its size too

services/github/src/handlers/status.js
```diff
@@ -1,30 +1,201 @@
...
 async function statusHandler(message) {
-  let { taskGroupId, state, runs, taskId } = message.payload.status;
-  let { runId } = message.payload;
-  let { reasonResolved } = runs[runId];
+  const { taskGroupId, state, runs, taskId } = message.payload.status;
+  const { runId } = message.payload;
+  const { reasonResolved } = runs[runId] || {};
+  const taskDefined = state === undefined;
+
+  await qLock.acquire(taskId);
```
lotas, line 182: this ensures we only run single status update for a given task. Consequent handlers will pause here until the previous is done.

services/github/src/queue-lock.js
```diff
@@ -0,0 +1,45 @@
...
+      resolver();
+    } else {
+      this.queue[name].push(resolver);
+    }
+
+    return promise;
+  }
+
+  release(name) {
+    const nextResolver = this.queue[name].shift();
```
lotas, line 28: This is the only part that can cause issues, if client code doesn't call `.release()` all pending callers will be stuck. 
At the moment status handler ensures that it always calls it before the end.

For the future we might want to implement some fail-safe watchdog timers to force-release lock after some timeout

services/github/src/queue-lock.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * Implements locked queue to allow one routine running at a time
+ */
+class QueuedLock {
```
lotas, line 4: is there a better name?

services/github/src/handlers/index.js
lotas: this caused function to return early, and all `monitor.timedHandler()` were returning over-optimistic durations :)

petemoore:
I'm not feeling too confident about being a good reviewer for this change. @djmitche Do you have any thoughts on this?

services/github/config.yml
```diff
@@ -5,7 +5,6 @@ defaults:
     deprecatedResultStatusQueue: 'stat-result'
     deprecatedInitialStatusQueue: 'stat-init'
     resultStatusQueue: 'ch-result'
-    initialStatusQueue: 'ch-init'
```
lotas, line 8: I guess we'll also need to drop this queue manually

services/github/src/github-auth.js
djmitche: Nice catch!  The cache size will be on the order of the number of installations, so in the 10's at most, so it should be pretty small.

services/github/src/handlers/status.js
```diff
@@ -1,30 +1,201 @@
...
 
-  let debug = makeDebug(this.monitor, { taskGroupId, taskId });
-  debug(`Handling state change for task ${taskId} in group ${taskGroupId}, reason=${reasonResolved || state}`);
+  let debug = makeDebug(this.monitor, { taskGroupId, taskId, id: `id-${counter}` });
+  counter += 1;
```
djmitche, line 185: Could this be an instance variable instead?

services/github/src/queue-lock.js
djmitche: I don't know if a timer is the right fix here.  If something is stuck and taking forever (like the GitHub API), probably waiting for it to fail is the right approach.  The risk is that a `return` gets added to the function without a corresponding `release`.  One way to avoid that is to have the `acquire` take an async callback which is called with the lock held, and unconditionally release the lock when the callback returns.

services/github/src/queue-lock.js
djmitche: I wonder if something from https://github.com/sindresorhus/promise-fun would serve this purpose, or at least could be used for the map values?

djmitche (dismissed):
This looks pretty solid -- my notes aren't anything serious.

How clear is it to someone deploying TC that they can't run more than one handler process?  I suspect that previously this wasn't necessary but wasn't expressly forbidden.  Now it must be forbidden.

services/github/src/queue-lock.js
lotas: I wasn't able to find anything that matched my use-case

services/github/src/queue-lock.js
lotas: Yeah, I was also thinking about this callback approach, but didn't want to make code more complicated on the handler side

services/github/src/handlers/status.js
lotas: thanks! Realized that it is even better to move it to `makeDebug()` so it would be applied to all handlers using this utility function, and not only for status handler :+1:

PR #5668, opened 2022-09-08
build(deps): upgrade to go1.19.3

matt-boris:
> Go upgrade from 1.18.5 to 1.19.3. Also upgrades golangci-lint version to 1.50.1 for go1.19 support.


lotas (dismissed):
looks okayish :) 

👍

matt-boris:
@lotas I should've changed this back to draft, as Pete and I were looking into why https://github.com/taskcluster/taskcluster/pull/5668/commits/1581b5e2d861da23189c48a8f7a711df2a1b03e0 was necessary. Looking as if there's a go bug, but finalizing our findings here soon 👍🏻

petemoore (changes requested):
I think it would be worthwhile adding a code comment here that explains that c.session.Accept() blocks until c.session.Close() is closed, which happens in c.Close(), but c.Close() also locks c.m, so we have to release it while c.session.Accept() runs, in order to avoid the deadlock. We could mention that this deadlock was first discovered when https://go-review.googlesource.com/c/go/+/409537 started enforcing that listener go routines complete before active connections are cleaned up after a server close / shutdown, and that prior to this, the deadlocks would still have been happening, but went unnoticed, and that there would have been deadlocked zombie go routines.

petemoore (approved):
Love it, thanks so much Boris!

matt-boris:
Thanks for the help @petemoore and @djmitche!!

PR #5670, opened 2022-09-11
Add completions for other shells

### LINKED ISSUE #4534: client-shell: Support completions for other shells
It looks like cobra supports command completions for `bash`, `zsh`, `fish` and `PowerShell` out of the box:
https://github.com/spf13/cobra/blob/master/shell_completions.md

But `taskcluster-cli completions` only supports `bash`. Would be awesome if we could hook up cobra's default completion command here instead.

samihda:
In addition to bash completions, add completions for fish, powershell, and zsh.

The [default Cobra completion command](https://pkg.go.dev/github.com/spf13/cobra#CompletionOptions) is still in place though. It can probably be removed in a separate (breaking?) update if necessary.

Github Bug/Issue: Fixes #4534

### DIFF
--- a/changelog/issue-4534.md
+++ b/changelog/issue-4534.md
@@ -2,3 +2,4 @@ audience: general
 level: minor
 reference: issue 4534
 ---
+Add completions for other shells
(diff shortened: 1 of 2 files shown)


lotas (dismissed):
Thanks for adding this!

lotas (approved):

PR #5693, opened 2022-10-03
build(deps): dep upgrades w/PMAC

matt-boris:
> Upgrade many deps with the following command:
`pmac add 5692 5691 5690 5689 5688 5687`

This also addresses some of the recent high vulnerabilities found [here](https://github.com/taskcluster/taskcluster/security/dependabot).


petemoore (approved):
Yay, `pmac`! Thanks Boris!

PR #5710, opened 2022-10-14
docs: update Github deployment docs for new Github interface

ahal:
I noticed that the strings mentioned in the deployment docs don't quite match up with what exists in Github when creating a new app. It's not too hard to figure out which options are *meant* to be set, but figured it would be better if the docs mentioned the exact permission strings visible in the Github interface.


lotas (approved):
Great additions, thanks Andrew!

PR #5714, opened 2022-10-17
Generic worker docker image

lotas:
Build was using incorrect Dockerfile for the generic worker.

Manually rebuilt and pushed latest `taskcluster/generic-worker:v44.22`.

The only reason it was working locally is because `build` directive was building it each time it started. Now it is moved to `dev` config.


petemoore (dismissed):
Thanks Yarik!

matt-boris (dismissed):
Ty!

matt-boris (approved):

PR #5727, opened 2022-10-19
Issue 5726: github service fetches backing logs instead of live logs

### LINKED ISSUE #5726: "certificate has expired" sentry alerts for github service
**Describe the bug**
Sentry issue: https://sentry.io/organizations/mozilla/issues/3555315507/?project=6459385

Exception "certificate has expired" being raised by github service when trying to fetch live log of a task.

**Expected behavior**

**Taskcluster version**
Sentry reports this started in taskcluster release 44.20.0.

**Screenshots**
image

**Additional context**
An abbreviated log was added to the github checks output that contained the last lines of the log in release 44.20.0. This retrieved the task log from task artifact `public/logs/live.log`. While the task is running, this artifact will redirect to a live http feed on the worker, which in the case of docker-worker, will be a url with a hostname provided by the stateless dns service. Docker worker configures these hostnames to be valid for the [max run time of the task](https://github.com/taskcluster/taskcluster/blob/ec95016e24808759f3df3db5b6173fc6371a76e7/workers/docker-worker/src/task.js#L790). Under normal operation, the task will complete within the max run time, and replace the live log artifact with a redirect artifact to the backing log. However, since there is no leniency in the expiration of the DNS record, presumably if the livelog url is queried immediately after the expiration of the dns record, but before docker worker has had the chance to upload the backing log and publish the redirect artifact for the livelog, the request will result in an expired certificate. This could also happen if docker-worker hangs for some reason, and does not publish the backing log.

In all cases it is probably desirable to only fetch the log once the task has completed, which is perhaps already the case. In this case, fetching the backing log is desirable, since it should contain the entire log, and only exist if the task has completed. This also avoids that the github service makes active connections to running workers, but instead retrieves artifacts directly from artifact storage (currently, aws s3). Live logs are really only meant to serve users directly, rather than backend services.

petemoore:
Fixes #5726.


matt-boris:
Yep, these ones :) https://github.com/taskcluster/taskcluster/pull/5727/checks?check_run_id=8984373419

matt-boris (approved):

services/github/src/handlers/utils.js
```diff
@@ -5,7 +5,7 @@ const taskUI = (rootUrl, taskGroupId, taskId) =>
 const taskGroupUI = (rootUrl, taskGroupId) =>
   libUrls.ui(rootUrl, `${rootUrl === 'https://taskcluster.net' ? '' : '/tasks'}/groups/${taskGroupId}`);
 const taskLogUI = (rootUrl, runId, taskId) =>
-  libUrls.ui(rootUrl, `/tasks/${taskId}/runs/${runId}/logs/public/logs/live.log`);
+  libUrls.ui(rootUrl, `/tasks/${taskId}/runs/${runId}/logs/public/logs/live_backing.log`);
```
lotas, line 8: https://github.com/taskcluster/taskcluster/commit/ce02f749c0e773d95ad005190f6177699a73aa31#diff-25dcdc2c35f64d1a1447614a17d90d675ac53a6e2188631790bbe4b052637ff4R51


here's the actual

matt-boris (approved):

PR #5738, opened 2022-10-27
Add auto-release to github status handlers

lotas:
Logs show that few github handlers might error or hang without releasing own lock. `maxLockTimeMs` for QueueLock will ensure that handlers do not run beyond given time, and forces their lock to be released. This will allow next waiting handlers to start work.

### DIFF
--- a/changelog/issue-5728.md
+++ b/changelog/issue-5728.md
@@ -1,0 +1,6 @@
+audience: general
+level: patch
+reference: issue 5728
+---
+
+Adds auto release lock functionality to queued locks to prevent some handlers to run forever and keep the queue locked.
--- a/changelog/issue-5728.md
+++ b/changelog/issue-5728.md
@@ -3,4 +3,4 @@ level: patch
 reference: issue 5728
 ---
 
-Adds auto release lock functionality to queued locks to prevent some handlers to run forever and keep the queue locked.
+Adds auto release lock functionality to queued locks to prevent some GitHub handlers to run forever and keep the queue locked.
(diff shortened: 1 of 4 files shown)


services/github/src/handlers/status.js
```diff
@@ -17,7 +17,9 @@ const { taskUI, makeDebug, taskLogUI, GithubCheck } = require('./utils');
  * This can lead to situations where newer event is being overwritten by an older one,
  * because it took him longer to reach update calls.
  */
-const qLock = new QueueLock();
+const qLock = new QueueLock({
+  maxLockTimeMs: 5 * 60 * 1000, // sometimes queue and github calls get delayed
```
matt-boris, line 21: Just wanted to hear how you got to this 5 min max lock time here? It just seems sorta long, but if the calls can take up to that amount of time, then by all means we can keep it like this. I thought more of a fail-fast timing here may be better.

changelog/issue-5728.md
```diff
@@ -0,0 +1,6 @@
+audience: general
+level: patch
+reference: issue 5728
+---
+
+Adds auto release lock functionality to queued locks to prevent some handlers to run forever and keep the queue locked.
```
matt-boris, line 6:
```suggestion
Adds auto release lock functionality to queued locks to prevent some GitHub handlers to run forever and keep the queue locked.
```
Just thought this could be a little bit more specific.

matt-boris (commented):
Left one main question and a small suggestion to the changelog! Nice though!

changelog/issue-5728.md
lotas: Thanks!

Issue #5739, opened 2022-10-27
Grant dw-ci tasks the docker-worker capability disableSeccomp


petemoore:
The `disableSeccomp` capability is required by the dw-ci task that runs the docker-worker tests, in order that the docker-worker tests themselves have the capability.

_Originally posted by @jschwartzentruber in https://github.com/taskcluster/taskcluster/issues/5695#issuecomment-1271591913_

matt-boris:
Completed via https://github.com/taskcluster/taskcluster/pull/5744

PR #5695, opened 2022-10-04
 Add a capability disableSeccomp to disable seccomp system call filter.

jschwartzentruber:
This is required to run `rr` inside a container, as described here: [Docker (rr wiki)](https://github.com/mozilla/rr/wiki/Docker)

This was previously opened as https://github.com/taskcluster/docker-worker/pull/532, but stalled during testing.

PR #5740, opened 2022-10-27
dw-ci tasks to have `disableSeccomp: true` set in payload.capabilities

petemoore:
Fixes: #5739

Note: still in DRAFT because we will also need a new release of taskgraph including https://github.com/taskcluster/taskgraph/pull/140.

Also note, I'm not entirely sure this will solve the remaining issue in #5695 (failing test) but I believe it is necessary as part of the fix.

@jschwartzentruber FYI.


petemoore:
> Note: still in DRAFT because we will also need a new release of taskgraph including https://github.com/taskcluster/taskgraph/pull/140.

(this is why the Decision Task is failing)

matt-boris:
Updated in https://github.com/taskcluster/taskcluster/pull/5744

PR #5745, opened 2022-10-31
Improves props handling for search query for TaskGroup page

### LINKED ISSUE #5737: Save task group filters as a URL parameter
This is essentially a clone of https://github.com/taskcluster/taskcluster/issues/5313 except for the task group filter. Sometimes graphs can have hundreds (or thousands) of tasks and you need to filter down to the set you care about. Only if you navigate away (e.g by clicking a task) and then go back, the filter is lost and you need to type it out again.

If the filter were added as a URL param, the browsers back/forward cache will ensure the filter is not lost. It would also make it possible to share that view as a link with others.

lotas:
Fixes #5737

### DIFF
--- a/changelog/issue-5737.md
+++ b/changelog/issue-5737.md
@@ -1,0 +1,6 @@
+audience: general
+level: patch
+reference: issue 5737
+---
+
+Fix React props handling for TaskGroup page in production mode.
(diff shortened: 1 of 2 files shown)


lotas:
> Seems reasonable, although I'm no React expert! Maybe best to get a review from @matt-boris too...

Thanks Pete. I'll merge to test on our dev. Last time it also worked locally.

PR #5753, opened 2022-11-01
build(deps): Dependabot minor/patch upgrades via PMAC

matt-boris:
> 
> `pmac add 5746 5747 5749 5750 5751`


lotas (approved):
looks ok :+1:

PR #5758, opened 2022-11-04
build(taskgraph): Dependabot will manage taskgraph dependencies

matt-boris:


lotas:
Great! Thanks.

I wonder if we should do something for the decision task? 
https://github.com/taskcluster/taskcluster/blob/main/.taskcluster.yml#L168

Maybe use latest always, but it wouldn't pin sha256, so probably not the best idea

lotas (approved):

Issue #5774, opened 2022-11-14
`build.sh` modifies `go.sum/mod` in a fresh checkout


djmitche:
**Describe the bug**
The `build.sh` script for `generic-worker`, run in a fresh checkout, modifies `go.mod` and `go.sum`

**To Reproduce**

```
⸩ cd workers/generic-worker/
✺ funk ~/p/taskcluster/workers/generic-worker ⸨main⸩
⸩ go version
go version go1.18.5 linux/amd64
✺ funk ~/p/taskcluster/workers/generic-worker ⸨main⸩
⸩ ./build.sh
.... .... ....
Build successful!
On branch main
Your branch is ahead of 'origin/main' by 2699 commits.
  (use "git push" to publish your local commits)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   ../../go.mod
        modified:   ../../go.sum

no changes added to commit (use "git add" and/or "git commit -a")
✺ funk ~/p/taskcluster/workers/generic-worker ⸨main*⸩
⸩ 
```

**Expected behavior**
`go.mod` and `go.sum` should be the way Go "wants them to be", so `./build.sh` should not produce any changes.

**Taskcluster version**
`main`

**Screenshots**

**Platform (please complete the following information):**

 - OS: [e.g. windows]
 - Browser if applicable [e.g. firefox, safari]
 - Version if applicable [e.g. 57]

**Additional context**
Here's the `go.mod` diff.  `go.sum` isn't as interesting..
```patch
diff --git a/go.mod b/go.mod
index a0b25f258..9cfdc29ec 100644
--- a/go.mod
+++ b/go.mod
@@ -37,19 +37,19 @@ require (
        github.com/taskcluster/httpbackoff/v3 v3.1.0
        github.com/taskcluster/pulse-go v1.0.0
        github.com/taskcluster/shell v0.0.0-20191115171910-c688067f12d3
        github.com/taskcluster/slugid-go v1.1.0
        github.com/taskcluster/taskcluster-lib-urls v13.0.1+incompatible
        github.com/tent/hawk-go v0.0.0-20161026210932-d341ea318957
        github.com/xeipuuv/gojsonschema v1.2.0
        golang.org/x/crypto v0.1.0
-       golang.org/x/net v0.1.0
-       golang.org/x/sys v0.1.0
-       golang.org/x/tools v0.2.0
[...]

PR #5775, opened 2022-11-14
update instructions for generic-worker dev

djmitche:
This reflects what I had to do to get `./build.sh -t` passing.

It looks like some of these `go get` lines are in `build.sh`, but not the `go install` lines.  I recall that in a recent version (go1.18?) `go get` stopped installing binaries.  Maybe it makes sense to have `build.sh` run `go install` for _all_ of these, instead of this solution?


workers/generic-worker/README.md
```diff
@@ -281,8 +281,19 @@ See the [![GoDoc](https://pkg.go.dev/github.com/taskcluster/taskcluster/v42/work
 Set up to build Taskcluster in general.
 See [development process](../../dev-docs/development-process.md).
 
-* Run `go get github.com/taskcluster/livelog`
- Run `go get github.com/taskcluster/taskcluster/v30/tools/taskcluster-proxy`
+* Run
+
+  ```shell
+  go get github.com/taskcluster/livelog`
+  go install github.com/taskcluster/livelog`
```
matt-boris, line 288:
```suggestion
  go get github.com/taskcluster/livelog
  go install github.com/taskcluster/livelog
```

matt-boris (changes requested):
Can confirm these steps work in getting `./build.sh -t` to run successfully.

Just requested changes for a couple stray ` + a missing changelog, thanks!

workers/generic-worker/README.md
```diff
@@ -281,8 +281,19 @@ See the [![GoDoc](https://pkg.go.dev/github.com/taskcluster/taskcluster/v42/work
...
+  go get github.com/taskcluster/taskcluster/v30/tools/taskcluster-proxy
+  go install github.com/taskcluster/taskcluster/v30/tools/taskcluster-proxy
```
petemoore, line 290: v30?

petemoore (changes requested):
Thanks for updating the docs.

I think you can avoid needing the `go get` and the `go install` by installing a specific version, e.g.

```
go install github.com/taskcluster/taskcluster/v44/tools/taskcluster-proxy@v44.9.2
```

If we want this to stay up-to-date, we should probably punch in the version numbers at release time via regexp. What do you think?

petemoore:
> Maybe it makes sense to have `build.sh` run `go install` for _all_ of these, instead of this solution?

That's a fair point - I'm also happy for us to do it that way too, and the instructions be clear that you just run `build.sh`.

workers/generic-worker/README.md
djmitche: That was just from copy-paste :)

djmitche:
It appears that this fixes #5774!

petemoore (dismissed):

petemoore:
@djmitche Can you add a silent changelog?

djmitche:
> @djmitche Can you add a silent changelog?

Done

lotas (approved):

petemoore (approved):
Many thanks!

lotas:
Somehow I merged it locally after doing rebase, and it didn't automatically closed this PR, sorry.

Here's the merge commit https://github.com/taskcluster/taskcluster/commit/ed6c810036c94393ecd38e49e578d88042b23d09  (all commits preserved in history also)

PR #5777, opened 2022-11-15
[generic-worker] support uploading object artifacts

### LINKED ISSUE #4605: [generic-worker] Upload artifacts as objects
It's OK to drop the existing s3 upload support, as this will be a major release.

djmitche:
This adds support for uploading artifacts containing data as type 'object' instead of type 's3', still defaulting to 's3'.

The PR begins with some refactoring.  The two interesting bits are adding the new object-service support (which is quite simple, since it relies on existing client support for object-service uploads) and using the configuration value to decide whether to create an S3Artifact or ObjectArtifact.

Github Bug/Issue: Fixes #4605


djmitche:
I'll need to figure out how to test this in the community instance.  There's really not enough code here to usefully test by mocking out the queue/object services.

petemoore:
@djmitche This looks good so far, but I think the test failures are genuine. Do you want to put it in draft status, until it is ready for review? Thanks!

petemoore:
By the way, the go test failures are best seen by searching for 'FAIL:' in the logs. The failures don't always appear at the end of the log. We could probably change it so that tests exit as soon as one fails...

djmitche:
Looking better now.  I think those tests don't run locally because I don't have credentials locally.

petemoore (approved):

petemoore:
This looks great, many thanks Dustin!

djmitche:
I merged from main again.  Please merge if successful?

lotas:
Done

PR #5787, opened 2022-11-20
add a getUrl download method

### LINKED ISSUE #4624: Add a `getUrl` download method

djmitche:
This adds a `geturl` download method to supplement the existing `simple` download method.

 * [x] documentation
 * [x] service support
 * [x] backend support

Github Bug/Issue: See #4624.


djmitche:
OK, let's land just the server side of this.  The client tests -- for the JS client, at least -- expect the local server to support the required method, so they won't be able to pass until the server side is deployed.

djmitche:
(I'll make a followon PR for the clients, once that's ready)

services/object/test/backends/google_test.js
```diff
@@ -61,20 +65,29 @@ helper.secrets.mockSuite(testing.suiteName(), ['google'], function(mock, skippin
...
-  const makeObject = async ({ name, data }) => {
-    const projectId = 'test-proj';
+  const projectId = 'test-proj';
+
+  const makeObject = async ({ name, data, hashes, gzipped }) => {
     const expires = taskcluster.fromNow('1 hour');
     const uploadId = taskcluster.slugid();
 
-    await helper.db.fns.create_object_for_upload(name, projectId, 'google', uploadId, expires, {}, expires);
+    await helper.db.fns.create_object_for_upload(name, projectId, 'aws', uploadId, expires, {}, expires);
```
lotas, line 74: Should it really be `'aws'` here in a `google` test?

lotas (dismissed):
Nice! Thanks

services/object/test/backends/google_test.js
djmitche: It should not! :)

djmitche: Oh, sorry -- it should.  The "google" support is implemented through GCS's S3 emulation.

petemoore (approved):
I like it!

I'm happy you have another PR open for the client implementations, so that indeed we can state that the new method is supported everywhere. We need to make sure we land this and the other PR(s) in the same release, so that docs are correct.

We discussed a little over matrix some other ideas, which we decided against. Making a note of our discussion here for our future selves:

1) Possibly stating which release the new method became available in, both for clients and Object Service. Decided against since we don't have a guaranteed way of knowing which release this will make it into, depending on whether another release is made before / after all the PRs are approved and merged.

2) We've decided not to reserve any properties for future expansion (such as having the ability to specify location, so that a region-specific url could be served, in the case of content that is mirrored between various regions/clouds). We decided not to do this, since we were not sure what that might look like, and therefore if additional properties come along, those would be done in a new method (not `getUrl`). We decided setting `additionalProperties: true` would be too loose and could lead to problems with misspellings going unnoticed).

3) We talked about the name `getUrl` and that the `simple` method also returns a url, and that the difference with this method is that it allows you to provide expected hashes and have a content encoding. However, we couldn't agree on a good name that represented these differences with the `simple` method, and since you are getting rid of `simple` we thought `getUrl` was ok and simple.

4) We noted that if you do get rid of `simple` we will need to make sure that where Object Service is already being used (e.g. in external deployments) that the services should be updated before clients, in order that negotiation doesn't fail if client removes `simple` support, but Object Service doesn't yet support `getUrl` method.

changelog/issue-4624.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: patch
```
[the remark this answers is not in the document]
petemoore: Actually, let's do that in the other PR where we remove simple method, but here should be `minor` since we are adding a new feature...

lotas (approved):

ui/docs/reference/platform/object/download-methods.mdx
```diff
@@ -55,20 +55,48 @@ For example, a request that can use either the `simple` or `superFancy` methods
...
 The known methods are defined below.
+See the API definitions for the object service for details on the request and response payloads.
 
 ### `simple` Download Method
 
-The `simple` method implements [simple downloads](/simple-downloads).
+The `simple` method is always supported, as it is used to implement [simple downloads](/simple-downloads).
 The response contains a URL against which a `GET` request can be made to retrieve the object's data.
+This method should only be used by generic clients that are not aware of the particulars of the object service, such as `curl` or web browsers.
+Taskcluster-aware clients should use the `getUrl` method, which is always available.
```
petemoore, line 65: @djmitche Can we remove this line, until this is implemented, because I think you wanted to release the server parts before implementing the client parts?

changelog/issue-4624.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: minor
+reference: issue 4624
+---
+The object service now supports an additional download method, `getUrl`, which handles gziped content and requires that hashes be validated.
```
petemoore, line 5: Can you add an additional comment here, that the client parts are not yet implemented, and will be provided in a future release?

petemoore (changes requested):

ui/docs/reference/platform/object/download-methods.mdx
djmitche: I think this line is valid with only this PR landed, though -- callers should prefer the `getUrl` method.  That's just not easy (yet) with the provided libraries.

changelog/issue-4624.md
djmitche: Yeah, that makes sense.

ui/docs/reference/platform/object/download-methods.mdx
petemoore: The part I find confusing is "which is always available", as this seems to suggest (to me) that the clients support the `getUrl` method in their native download methods (otherwise I'm not sure what it means).

Perhaps just replace "which is always available" with "where possible"? Or we could add a note to say that clients that wish to download objects may call `object.startDownload` with `{ acceptDownloadMethods: { getUrl: true }}` and then use the returned url to retrieve the object, and that in future, there will be a native helper function/method that will provide this functionality (and maybe cite the method/function name if it is standard across clients?).

Does my point of confusion make sense?

djmitche: It's always available _on the server_, regardless of the object service backend.  The client libraries are just a convenience provided to make it easier to interact with the object service (they are _very_ convenient, especially for the object service), so it's implicit that they would call `startDownload` with `getUrl: true` if they wanted to use this method.

Anyway, this is a tiny, transient detail -- the client libraries will support it soon.  It's hard to see how this would cause anyone any confusion in the interim, since nothing is using the object service.

petemoore (approved):
Thanks Dustin!

matt-boris (approved):

PR #5796, opened 2022-11-24
Call finishArtifact for object artifacts

### LINKED ISSUE #5795: generic-worker must call finishArtifact for object artifacts
All other artifact types do not require this, so I missed adding it for type=object

djmitche:
Object artifacts must be "finished" before they appear on the queue, so this does so.

Github Bug/Issue: Fixes #5795


internal/mocktc/queue.go
```diff
@@ -167,6 +167,14 @@ func (queue *Queue) CreateArtifact(taskId, runId, name string, payload *tcqueue.
 	return &par, nil
 }
 
+func (queue *Queue) FinishArtifact(taskId, runId, name string, payload *tcqueue.FinishArtifactRequest) error {
+	queue.mu.Lock()
+	defer queue.mu.Unlock()
+
+	// do nothing (for now)
+	return nil
+}
```
petemoore, line 176: Should we also update `CreateArtifact` method above, and add handling of `"object"`storage type? At the moment it will cause a test failure - so I guess we don't have any tests for object storage type - should we add at least one?

workers/generic-worker/artifacts/artifacts.go
```diff
@@ -54,3 +58,12 @@ type (
 func (base *BaseArtifact) Base() *BaseArtifact {
 	return base
 }
+
+// FinishArtifact implements TaskArtifacg#FinishArtifact.
```
petemoore, line 62:
```suggestion
// FinishArtifact implements TaskArtifact#FinishArtifact.
```

petemoore (approved):
Thanks Dustin!

internal/mocktc/queue.go
djmitche: It [is tested](https://github.com/taskcluster/taskcluster/blob/51ac97dcac93f8a219a464c0753a8d914899a449/workers/generic-worker/artifacts_test.go#L119-L147) like the other upload methods.  But it looks like that only tests creating the TaskArtifact objects, without actually executing the upload.

Ah, I found tests that actually create objects in the broken-on-docker file!

petemoore: Oh indeed, that isn't named nicely (broken-on-docker). I think we'll probably get rid of the build constraints entirely next year.

djmitche:
Huh, I'm not sure why
```
        [taskcluster 2022-11-25T22:11:26.455Z] Executing command 1: cp '/task_166941419113895/taskcluster/workers/generic-worker/testdata/SampleArtifacts/_/X.txt' 'SampleArtifacts/_/X.txt'
        cp: cannot stat '/task_166941419113895/taskcluster/workers/generic-worker/testdata/SampleArtifacts/_/X.txt': No such file or directory
```
I think that part is unchanged?

djmitche:
...and it works for me locally:
```
2022/11/25 22:47:44Z Executing command 0: ["mkdir" "-p" "SampleArtifacts/_"]
2022/11/25 22:47:44Z Executing command 1: ["cp" "/home/dustin/p/taskcluster/workers/generic-worker/testdata/SampleArtifacts/_/X.txt" "SampleArtifacts/_/X.txt"]
```
any idea what's up?

Eijebong:
While I don't know what's happening, I was able to reproduce locally:

```
[taskcluster 2022-11-26T15:34:38.785Z] Executing command 1: cp '/home/eijebong/code/taskcluster/workers/generic-worker/testdata/SampleArtifacts/_/X.txt' 'SampleArtifacts/_/X.txt'
cp: cannot stat '/home/eijebong/code/taskcluster/workers/generic-worker/testdata/SampleArtifacts/_/X.txt': No such file or directory
```

The file definitely exists:

```
[eijebong@plutonium] ~/code/taskcluster/workers/generic-worker (pr-5796) >>> cat /home/eijebong/code/taskcluster/workers/generic-worker/testdata/SampleArtifacts/_/X.txt
test artifact
```

This is on an up to date archlinux machine.
`go version go1.19.3 linux/amd64`

djmitche:
I wonder if it has to do with the order the tests run in, somehow??

Eijebong:
Could be. I ran tests with `./build.sh -t`

djmitche:
Ah -- I usually don't run that since it takes so long (6.5 minutes)!

djmitche:
Oh, I bet these are broken on docker.  I don't know how to tell if a test is running with the docker tag, in the build.sh output.

djmitche:
Nope, that wasn't it.  And I tried using an artifact with a different name in hopes that some other test was deleting X.txt at an inopportune moment.  The file exists after the tests run.  And its modification time hasn't changed.  Is there some `chroot` thing going on here?

djmitche:
Oh, I think I forgot to `git add` the new file, lol.

petemoore (approved):
Looks great, many thanks!

PR #5801, opened 2022-11-28
fix(github): point livelog URL to streamed livelog

matt-boris:
> Updated livelog link in GitHub checks UI to points to a streaming livelog. Previous link would only render the logs once the task was complete.


petemoore (approved):
This looks fine, just one thing to note though:

The ui urls and the service api urls are diverging a little bit, but they were already inconsistent. We don't have a formal interface around ui urls, which is something we may wish to address in future, so that people and systems can safely link to ui elements.

The queue.getArtifact URL path is:
`<ROOT_URL>/api/queue/v1/task/<TASK_ID>/runs/<RUN_ID>/artifacts/<ARTIFACT_PATH>`

The Task Inspector URL path for the live log artifact is now:
`<ROOT_URL>/tasks/<TASK_ID>/runs/<RUN_ID>/logs/live/<ARTIFACT_NAME>`

As you see, the path elements had already diverged (e.g. `/task` -> `/tasks`).

PR #5802, opened 2022-11-28
build(deps): upgrade crates and rust to 1.65.0

matt-boris:
> Upgrades some rust crates and bumps rust version from 1.60.0 to 1.65.0.


petemoore (approved):
Looks sane, thanks Boris!

Issue #5804, opened 2022-11-29
https://firefox-ci-tc.services.mozilla.com/hooks/project-relman/code-review-production errors out with 504


marco-c:
Loading https://firefox-ci-tc.services.mozilla.com/hooks/project-relman/code-review-production or https://firefox-ci-tc.services.mozilla.com/hooks/project-relman/code-review-testing doesn't work. After a while, the following error is displayed:
> Network Error (504): ServerParseError: JSON.parse: unexpected character at line 1 column 1 of the JSON data

lotas:
So there's a network time out. 
And I believe that's because it tries to fetch _all_ last fires without pagination.
For the `project-relman/code-review-production` it already has `203461` records, which will probably take long to serialize to respond.

We should introduce proper pagination and maybe split calls to load page without that initially.

lotas:
related to #4870

lotas:
Discovered that it's not only the UI that tries to fetch everything, hooks api is also loading all of the fires to determine the most recent one

Issue #5573, opened 2022-07-20
UI: Hooks view fetches all executions of a hook


lotas:
Hooks that run frequently may produce lots of tasks (like [this hook](https://community-tc.services.mozilla.com/hooks/project-mozci/check-backfills-production)).

UI page is doing a query to fetch all recent runs, which loads all tasks. Some responses reach megabytes in size and take some time to fetch from db.

TODO: 
- [ ] Add server-side pagination to load less
- [ ] Add an option for UI to display more than 5 runs 
- [x] Change DB queries to fetch last hooks sorted by creation time
- [ ] Change hooks API to fetch latest by date and not iterate all records

lotas:
will close this one as #5804 #5805 solved the main issue

Issue #4870, opened 2021-05-10
hooks.listLastFires returning old records


petemoore:
$ taskcluster api hooks listLastFires hg-push try | jq '.lastFires[] | .taskId' | wc -l
98434

This results in problems in the UI...

<img width="1072" alt="Screenshot 2021-05-10 at 16 51 04" src="https://user-images.githubusercontent.com/190790/117678744-f56ebb00-b1af-11eb-87ee-665f6dd68f8d.png">

jwhitlock:
This issue was brought up in https://bugzilla.mozilla.org/show_bug.cgi?id=1722140, because viewing the hook https://firefox-ci-tc.services.mozilla.com/hooks/hg-push/try failed when debugging an issue with try builds.

Retrieving the JSON from the API takes about 30 seconds, and we're up to 125,162 items.

```
curl https://firefox-ci-tc.services.mozilla.com/api/hooks/v1/hooks/hg-push/try/last-fires > try.json
cat try.json | jq '.lastFires | length'
```

Is the issue that old tasks are not being cleaned up, and we need a cronjob to periodically clean them? Or do we need to paginate last fires, similar to the [queue list task group API](https://firefox-ci-tc.services.mozilla.com/docs/reference/platform/queue/api#listTaskGroup)?

PR #5805, opened 2022-11-29
Hook last fires pagination

lotas:
This PR changes:
- `get_last_fires` uses sorting by creation date to avoid doing full scan to find latest
- `hooks_last_fires` table has new index to allow faster ORDER BY + LIMIT queries for time column
- changes graphql that adds pagination to `lastHookFires` to avoid sending everything to the client

Closes: #5804 #5573 #4870

### DIFF
--- a/db/test/helper.js
+++ b/db/test/helper.js
@@ -253,7 +253,7 @@ const queryIndexInfo = async (client, table, index, column) =>
       AND a.attname = '${column}'`);
 
 /**
- * Assert that the table contains given index on specified column
+ * Assert that the table contains given index on specific column
  */
 exports.assertIndexOnColumn = async (table, index, column) => {
   await exports.withDbClient(async client => {
@@ -263,7 +263,7 @@ exports.assertIndexOnColumn = async (table, index, column) => {
 };
 
 /**
- * Assert that the table doesn't contains given index on a column
+ * Assert that the table doesn't contain given index on a column
  */
 exports.assertNoIndexOnColumn = async (table, index, column) => {
   await exports.withDbClient(async client => {
--- a/services/hooks/src/api.js
+++ b/services/hooks/src/api.js
@@ -667,7 +667,7 @@ builder.declare({
     'This endpoint will return information about the the last few times this hook has been',
     'fired, including whether the hook was fired successfully or not',
     '',
-    'By default this endpoint will return up to 1000 last fires in one request.',
+    'By default this endpoint will return up to 1000 most recent fires in one request.',
   ].join('\n'),
 }, async function(req, res) {
   const { hookGroupId, hookId } = req.params;
--- a/changelog/issue-5804.md
+++ b/changelog/issue-5804.md
@@ -1,0 +1,9 @@
+audience: users
+level: minor
+reference: issue 5804
+---
+
+Adds pagination to the hooks last fires api call.
+
+This prevents loading all last fires for the hooks that have thousands of records, which results in 500 errors.
+Changes the behaviour of the existing `get_last_fires` function by using a different sort column - creation time instead of task_id.
--- a/clients/client-go/tchooks/tchooks.go
+++ b/clients/client-go/tchooks/tchooks.go
@@ -348,7 +348,7 @@ func (hooks *Hooks) TriggerHookWithToken(hookGroupId, hookId, token string, payl
 // This endpoint will return information about the the last few times this hook has been
 // fired, including whether the hook was fired successfully or not
 //
-// By default this endpoint will return up to 1000 last fires in one request.
+// By default this endpoint will return up to 1000 most recent fires in one request.
 //
 // Required scopes:
 //
--- a/clients/client-py/taskcluster/generated/aio/hooks.py
+++ b/clients/client-py/taskcluster/generated/aio/hooks.py
@@ -206,7 +206,7 @@ class Hooks(AsyncBaseClient):
         This endpoint will return information about the the last few times this hook has been
         fired, including whether the hook was fired successfully or not
 
-        By default this endpoint will return up to 1000 last fires in one request.
+        By default this endpoint will return up to 1000 most recent fires in one request.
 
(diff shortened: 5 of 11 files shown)


db/versions/0079.yml
```diff
@@ -0,0 +1,41 @@
+version: 79
+description: Speed up hooks fetching last fires by date
+migrationScript: |-
+  begin
+    create index hooks_last_fires_time on hooks_last_fires (hook_group_id,hook_id,task_create_time);
```
lotas, line 5: This was tested on the production dataset locally, which has over 2.000.000 records in that table.
Index creation is fast, and overhead is few hundred MB in storage

db/versions/0079.yml
```diff
@@ -0,0 +1,41 @@
...
+          hooks_last_fires.task_id,
+          hooks_last_fires.task_create_time,
+          hooks_last_fires.result,
+          hooks_last_fires.error,
+          public.gen_random_uuid()
+        from hooks_last_fires
+        where
+          hooks_last_fires.hook_group_id = hook_group_id_in and
+          hooks_last_fires.hook_id = hook_id_in
+        order by hook_group_id, hook_id, task_create_time DESC
```
lotas, line 38: this is the only part changed from the original, instead if `task_id` - `task_create_time DESC` is used.

ui/src/views/Hooks/ViewHook/index.jsx
```diff
@@ -20,6 +21,9 @@ import exchangesList from '../../../utils/exchangesList';
     variables: {
       hookGroupId: params.hookGroupId,
       hookId: decodeURIComponent(params.hookId),
+      hookConnection: {
+        limit: VIEW_CLIENTS_PAGE_SIZE,
```
lotas, line 25: UI doesn't have a full pagination built in yet, so it will only display first 1000 records.
My reasoning is that we might not need to go beyond that.

db/test/helper.js
```diff
@@ -232,6 +232,46 @@ exports.assertNoTableColumn = async (table, column) => {
...
+ */
+exports.assertIndexOnColumn = async (table, index, column) => {
+  await exports.withDbClient(async client => {
+    const res = await queryIndexInfo(client, table, index, column);
+    assert.deepEqual(res.rows, [{ table, index, column }]);
+  });
+};
+
+/**
+ * Assert that the table doesn't contains given index on a column
```
petemoore, line 266:
```suggestion
 * Assert that the table doesn't contain given index on a column
```

petemoore (commented):
I can rubber stamp this, but I think I'd prefer someone else to review the pagination changes that understands the code better than me. Looks like a big piece of work, many thanks for this Yarik!

db/test/helper.js
```diff
@@ -232,6 +232,46 @@ exports.assertNoTableColumn = async (table, column) => {
...
+      AND a.attnum = ANY (ix.indkey)
+      AND t.relkind = 'r'
+      AND t.relname = '${table}'
+      AND i.relname = '${index}'
+      AND a.attname = '${column}'`);
+
+/**
+ * Assert that the table contains given index on specified column
+ */
+exports.assertIndexOnColumn = async (table, index, column) => {
```
djmitche, line 258: Nice!

db/versions/0079.yml
djmitche: I'd say this is _technically_ a breaking change, except that the old docs said it was ordered by worker_id which is obviously bogus :)

I don't think it will cause any issues to have old JS code call this updated function, nor to have new JS code call the old function.  So it should be fine to replace the function in-place.

That said, it might be good to mention this in the changelog ("..changes the behavior of the existing get_last_fires function..")

services/hooks/src/api.js
```diff
@@ -664,32 +662,38 @@ builder.declare({
   title: 'Get information about recent hook fires',
   stability: 'stable',
   category: 'Hook Status',
+  query: paginateResults.query,
   description: [
     'This endpoint will return information about the the last few times this hook has been',
     'fired, including whether the hook was fired successfully or not',
+    '',
+    'By default this endpoint will return up to 1000 last fires in one request.',
```
djmitche, line 670: Let's include that these are the most recent.

Before this change, the call would return _all_ last fires.  Now, the same call (without pagination query) will return _some_ last fires.  But since those are the most recent, and TC periodically flushes the less recent last-fires, I think that's a legitimate change.

ui/src/views/Hooks/ViewHook/index.jsx
djmitche: I think that's fine..

djmitche (approved):
I was never too clear on the GraphQL stuff and have forgotten whatever I might have known, so I'll take your word for that.  But the rest looks good!

services/hooks/src/api.js
lotas: Thanks! 
Regarding the flushing of last fires I'm not so sure this is happening, as there are few hooks with thousands of those last fires. 
upd: yes, the `expire_last_fires` only deletes those older than 1 year, so apparently it can still get a lot of tasks:

image

db/test/helper.js
lotas: Thanks Pete, will add

lotas:
thanks for the review @djmitche

services/hooks/src/api.js
djmitche: Ah, right -- IIRC we thought about deleting if there are more than N for a hook, but that's a more complicated query.  That might be worth considering anyway?  In another PR of course :)

lotas: ah, that's actually another possible solution for this problem.. and an easier one. :)
I'll create a ticket to add such kind of expiration by number as well

lotas: https://github.com/taskcluster/taskcluster/issues/5839  added here

petemoore (approved):
Thanks Yarik!

PR #5715, opened 2022-10-17
[WIP] Docker worker documentation

lotas:
Minor docker worker documentation fixes

[skip ci]

PR #5809, opened 2022-11-30
Migrate from gliderlabs/alpine:<various> to alpine:latest docker images

petemoore:
CI started failing due to a seemingly no longer maintained base docker image (gliderlabs/alpine).

This PR switches our CI tests to the latest vanilla alpine docker image.

### DIFF
--- a/workers/docker-worker/test/images/dind-test/Dockerfile
+++ b/workers/docker-worker/test/images/dind-test/Dockerfile
@@ -1,5 +1,5 @@
-FROM        gliderlabs/alpine:3.2
-MAINTAINER  Jonas Finnemann Jensen <jopsen@gmail.com>
+FROM        alpine:latest
+MAINTAINER  Taskcluster <taskcluster-accounts+docker-maintainer@mozilla.com>
 
 # Install dependencies
 RUN         apk-install docker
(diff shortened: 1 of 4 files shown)


workers/docker-worker/test/images/dind-test/Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM        gliderlabs/alpine:3.2
-MAINTAINER  Jonas Finnemann Jensen <jopsen@gmail.com>
+FROM        alpine:latest
+MAINTAINER  Taskcluster <taskcluster-accounts+docker-maintainer@mozilla.com>
 
 # Install dependencies
 RUN         apk-install docker
```
lotas, line 5: this will not run in latest alpine 
check this https://github.com/taskcluster/taskcluster/pull/5715/files 

Also maybe let's pin the version?

lotas, line 5:
```suggestion
RUN         apk add docker
```

lotas (approved):
Thanks!

workers/docker-worker/test/images/dind-test/Dockerfile
petemoore: Discussed via slack, we decided this is ok, as many other tests use `:latest`, and this way they are at least consistent.

PR #5824, opened 2022-12-01
build(deps): Dependabot minor/patch upgrades via PMAC (Dec. 1)

matt-boris:
> 
> `pmac add 5822 5821 5819 5818 5817 5816 5815 5814 5813 5790 5789`


lotas (approved):
LGTM! Thanks!

PR #5825, opened 2022-12-01
Add ".deb" to the exclusion list for artifact compression

jcristau:
Bugzilla Bug: [1803745](https://bugzilla.mozilla.org/show_bug.cgi?id=1803745)

### DIFF
--- a/changelog/bug-1803745.md
+++ b/changelog/bug-1803745.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: bug 1803745
+---
+Docker-worker and generic-worker now skip gzipping artifacts with a `.deb` extension.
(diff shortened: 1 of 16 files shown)


jcristau:
> This looks great, is it ok to also update the schemas?
> 
> workers/generic-worker/schemas/docker_posix.yml workers/generic-worker/schemas/multiuser_posix.yml workers/generic-worker/schemas/multiuser_windows.yml workers/generic-worker/schemas/simple_posix.yml

Thanks for pointing that out.  Do I also need to update the `generated_*` files in generic-worker?  If so, how?

petemoore:
> Thanks for pointing that out. Do I also need to update the `generated_*` files in generic-worker? If so, how?

Thanks Julien! If you run `./build.sh` in the workers/generic-worker subdirectory, that should do it.

I've made an issue (#5829) to hopefully help with this in future.

petemoore (dismissed):
Perfect, thank you Julien!

Issue #5781, opened 2022-11-16
Generic worker test failures difficult to find


petemoore:
When generic-worker tests fail in the CI, the failure text is often not at the end of the log, but at the point the test occurred, which can make it harder for people not familiar with this, to find the failures. It might (?) be nicer if we specify `-failfast` when running generic-worker tests so that tasks stop as soon as they hit a failure.

Note, an alternative would be to log a message after the failure saying something like "Please be aware you should search for `FAIL:` in the output above to find the source of the test failure" but that might be making things too complicated.

An advantage of running all tests is that you don't just tackle failures one-by-one, but get all the failure information in one shot. But maybe this is outweighed by the benefits of saving resources in CI and having the failure at the bottom of the log.

Another option to consider at this time is removing "verbose" output in the logs (`go test -v`). Passing tests should produce no output. Sometimes the verbose output is useful to compare succeeding and failing tasks, but probably this is more the exception than the rule.

djmitche:
Part of the issue is that "FAIL" appears frequently in the normal test output, so it's a bit difficult to find.  I find this a little more difficult when running tests locally, than in CI -- my scrollback buffer is so big and its search facilities are not great.

petemoore:
Yeah, I tend to suffix a `:`, i.e. search for `FAIL:` rather than `FAIL` but indeed, it isn't ideal. We could also update `build.sh` to not generate the verbose output too...

petemoore:
> We could also update `build.sh` to not generate the verbose output too...

I've added cc505b1249f694155e3b7effa01b0a3ac3ae8cb9 in #5782 to do that now...

escapewindow:
One thing I've done is:
- run the tests through a wrapper
- have the wrapper parse the output for substrings like `FAIL:` while also outputting to the console or logfile
- at the end of the run, output a summary of the run, e.g. all `FAIL:` lines detected with a count of the failures

petemoore:
Looks like #5782 was not enough to resolve this issue. See e.g. https://community-tc.services.mozilla.com/tasks/M8Q49_00Qy-7Ro0fTUdv3A/runs/0/logs/live/public/logs/live.log

It seems that `go test` and `go test .` [operate differently](https://stackoverflow.com/questions/31663229/how-can-stdout-be-captured-or-suppressed-for-golang-testing) in this regard, and `go test` needs to be replaced with `go test .` to suppress the output.

petemoore:
Fixed by #5782 and #5827.

PR #5827, opened 2022-12-02
Issue 5781 part 2: specify package(s) name when calling go test to suppress stdout/stderr in passing tests

petemoore:
I hadn't spotted that my changes in PR #5782 were not enough to implement issue #5781. This is a follow-on that should hopefully put the issue to bed.

This time I'll also check the output of the affected CI tasks to make sure that the intended changes are effective before merging! (whoops)

### DIFF
--- a/changelog/JpzzjXzuTkqyZ_ekVL8EjA.md
+++ b/changelog/JpzzjXzuTkqyZ_ekVL8EjA.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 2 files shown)


petemoore:
Note, the key change is changing `go test` to `go test .` as highlighted by [this](https://stackoverflow.com/a/57930037) post.

At the same time I swapped the order of the two `go test` commands under Windows, to be consistent with Linux/macOS tasks.

petemoore:
Also note, one drawback is that we now don't see which tests are actually running. This isn't easy to rectify due to https://github.com/golang/go/issues/21926#issuecomment-338789080.

petemoore:
> This time I'll also check the output of the affected CI tasks to make sure that the intended changes are effective before merging! (whoops)

Looking much better! https://community-tc.services.mozilla.com/tasks/H_47VWUpQASm83Ugu2LCxw/runs/0/logs/live/public/logs/live.log

Only ~500 lines now, rather than ~10,000.

lotas (approved):

lotas (approved):

PR #5835, opened 2022-12-02
build(deps): upgrade taskgraph to v4.0.0

matt-boris:
> Upgrade `taskcluster-taskgraph` version to v4.0.0 as well as upgrade other python packages.


lotas:
Thanks!

PR #5837, opened 2022-12-03
Be more precise about object hashes

### LINKED ISSUE #5791: Require hashes on objects
Let's set some kind of enforced minimum of hashes for an object upload to be accepted, which downloaders can then expect to find.

For the moment, we define `sha256` and `sha512`, but don't say anything about when they must be defined.  Over the years, we'll probably want to add hash algorithms, and eventually older hash algorithms will be deprecated (as md5 and sha1 have been), so we should have a policy that allows some kind of evolution.  Perhaps:

 * must include at least one of [algos still considered secure] (e.g., sha256, sha512)
 * should include [most secure algo] (e.g., sha512)
 * can also include [deprecated algos] (e.g., md5, sha1)

Downloaders should be expected to verify all recognized algorithms for which there are hashes, and to verify that at least one of the first set are present.

This gives enough time for some evolution: assuming sha1024 is the next best algorithm, uploaders can start adding it but also including sha256 and sha512.  During that time, downloaders will validate sha256 and sha512, and new implementations will also validate sha1024.  Then by the time sha256 is deprecated, uploaders and downloaders will already be using sha512 and sha1024, and can drop sha256.

I think this would be a docs-only fix, for now.

Thoughts?

djmitche:
This adds some guidance in the form of comments, and also sets `additionalProperties: false`, for reasons outlined in those comments.

Github Bug/Issue: Fixes #5791


lotas:
For some reason linter and go client tests fail:
```
clients/client-go/tcobject/hashingreadseeker_test.go:23:94: hashes.Sha256 undefined (type json.RawMessage has no field or method Sha256) (typecheck)
	require.Equal(t, "98b1ae45059b004178a8eee0c1f6179dcea139c0fd8a69ee47a6f02d97af1f17", hashes.Sha256)
	                                                                                            ^
clients/client-go/tcobject/hashingreadseeker_test.go:24:158: hashes.Sha512 undefined (type json.RawMessage has no field or method Sha512) (typecheck)
	require.Equal(t, "e0ea5ae6e392bb46d27eebabf5e7eb817d242505a960079cd9871559eaa94c613aff4034b709ea3cbd7747b304e7da5564083df50ea51f389cddcb942d2a4a09", hashes.Sha512)
	                                                                                                                                                            ^
clients/client-go/tcobject/hashingreadseeker_test.go:46:94: hashes.Sha256 undefined (type json.RawMessage has no field or method Sha256) (typecheck)
	require.Equal(t, "98b1ae45059b004178a8eee0c1f6179dcea139c0fd8a69ee47a6f02d97af1f17", hashes.Sha256)
	                                                                                            ^
clients/client-go/tcobject/hashingreadseeker_test.go:47:158: hashes.Sha512 undefined (type json.RawMessage has no field or method Sha512) (typecheck)
	require.Equal(t, "e0ea5ae6e392bb46d27eebabf5e7eb817d242505a960079cd9871559eaa94c613aff4034b709ea3cbd7747b304e7da5564083df50ea51f389cddcb942d2a4a09", hashes.Sha512)
```

djmitche:
(not a commit error, but hopefully fixed now)

clients/client-go/tcobject/upload.go
```diff
@@ -49,6 +49,7 @@ func (object *Object) UploadFromFile(projectID string, name string, contentType
 // object content read from readSeeker. The value of contentLength is not
 // validated prior to upload.
 func (object *Object) UploadFromReadSeeker(projectID string, name string, contentType string, contentLength int64, expires time.Time, uploadID string, readSeeker io.ReadSeeker) (err error) {
+	return nil
```
djmitche, line 52: Whoops, this was added while debugging!

petemoore (approved):
Thanks Dustin!

PR #5846, opened 2022-12-06
Add a backend to object-service in dev environments

djmitche:
This was a step on the way to testing the object service's support for getUrl.


infrastructure/tooling/src/generate/generators/docker-compose.js
```diff
@@ -101,6 +101,10 @@ const defaultValues = {
   // Notify
   EMAIL_SOURCE_ADDRESS: 'root@local',
 
+  // Objectj
```
petemoore, line 104:
```suggestion
  // Object
```

petemoore (approved):

PR #5847, opened 2022-12-06
Include geturl schema in startDownload

djmitche:
One more missed thing in adding this new support.  But, at least now I can catch these locally with a dev environment.

### DIFF
--- a/clients/client-go/tcobject/types.go
+++ b/clients/client-go/tcobject/types.go
@@ -94,6 +94,7 @@ type (
 	//
 	// One of:
 	//   * SimpleDownloadResponse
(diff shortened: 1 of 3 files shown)


petemoore (approved):
Thanks Dustin! I think we're ok without a changelog here, as we have changelogs in the other PRs that cover everything.

djmitche:
> Thanks Dustin! I think we're ok without a changelog here, as we have changelogs in the other PRs that cover everything.

Sorry, I tend to forget changelogs when I cherry-pick a commit out of another branch like this.  I agree -- this is a fix to something unreleased that is already covered by a changelog.

PR #5848, opened 2022-12-06
feat(ui): Fix empty error message for hook last fire

lotas:
Fixes in this PR:

* Extend ErrorPanel component to display errors that look like json serialized API response error.  In hooks last fires it was showing empty red panel.
* Task group stats adds safe min width for the graph, so group with few tasks wouldn't have really wide bars on it. Looks better

### DIFF
--- a/ui/src/components/TaskGroupStats/index.jsx
+++ b/ui/src/components/TaskGroupStats/index.jsx
@@ -149,7 +149,7 @@ export default class TaskGroupStats extends Component {
       searchTerm,
       graphAll ? +Infinity : maxTasksInGraph
     );
-    const width = (sampledTasks.length + 1) * barWidth;
+    const width = (Math.max(25, sampledTasks.length) + 1) * barWidth;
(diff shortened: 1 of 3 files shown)


matt-boris (dismissed):
👍🏻

ui/src/components/TaskGroupStats/index.jsx
```diff
@@ -149,7 +149,7 @@ export default class TaskGroupStats extends Component {
       searchTerm,
       graphAll ? +Infinity : maxTasksInGraph
     );
-    const width = (sampledTasks.length + 1) * barWidth;
+    const width = (Math.max(25, sampledTasks.length) + 1) * barWidth;
```
lotas, line 152: As whole graph is being stretched 100% we add here safe min `25` bars for it. So groups with just few bars, would not be stretched all the way, but would take those 25 placeholders instead

matt-boris (approved):

lotas:
I'll merge it despite few tests failing (I believe it is temp issue) to test it on dev

PR #5854, opened 2022-12-09
Update incorrect link in changelog

lotas:


matt-boris (approved):

changelog/F-pEwPQpS3CaBH1GspLyqA.md
```diff
@@ -0,0 +1,5 @@
+audience: general
+level: patch
+---
+
+Update incorrect link to a ticket in changelog.
```
petemoore, line 5: Maybe we can be more explicit here, and say that release 46.0.0 incorrectly listed issue 2791 in the release notes, when it should have listed issue 5791?

I think this is useful because when you read the release notes, you might not have the git history available.

matt-boris (dismissed):

PR #5856, opened 2022-12-09
Fix python client tests

lotas:
Python tests were failing [py38](https://github.com/taskcluster/taskcluster/pull/5854/checks?check_run_id=9995796033)

According to [documentation](https://tox.wiki/en/4.0.3/config.html) external tools needs to be whitelisted


matt-boris (approved):
Nice, thanks for the quick fix!

PR #5858, opened 2022-12-10
test(generic-worker): simplify setup cleanup with t.Cleanup

Juneezee:
This PR simplifies the cleanup of  `setup(t)` calls with `t.Cleanup` in `generic-worker` tests. The main reason for using `t.Cleanup` is to simplify the resource creation and cleanup logic during a test in a test helper function, for example:

```go
func createTestResource(t *testing.T) *resource {
	resource, cleanup := createResource()

	t.Cleanup(func() {
		cleanup()
	})

	return resource
}

func TestFoo(t *testing.T) {
	resource := createTestResource(t)
	// do things with resource...
}
```
In this way, we can focus on the test logic without cluttering the code with repeated cleanup calls.


petemoore (approved):
This looks great! Many thanks @Juneezee for this simplification!

Not sure what the macOS failure was, but running it again now...

PR #5859, opened 2022-12-11
Support getUrl downloads in clients

### LINKED ISSUE #4624: Add a `getUrl` download method

djmitche:
In fact, the clients no longer use simple downloads, meaning that all downloads performed via a client will verify hashes on the downloaded content.

Github Bug/Issue: Fixes #4624

### DIFF
--- a/clients/client-rust/download/src/hashing.rs
+++ b/clients/client-rust/download/src/hashing.rs
@@ -1,0 +1,119 @@
+use super::factory::AsyncWriterFactory;
+use anyhow::Result;
+use sha2::{Digest, Sha256, Sha512};
+use std::collections::HashMap;
+use std::pin::Pin;
+use std::sync::{Arc, Mutex};
+use std::task::{Context, Poll};
+use tokio::io::AsyncWrite;
+
+/// A hasher reader factory wraps an AsyncWriterFactory in such a way that it will capture hashes of
+/// the content, making those hashes available once at least one complete read-through of the data
+/// is complete.
+pub(super) struct HasherAsyncWriterFactory<'f, AWF: AsyncWriterFactory> {
+    /// the underlying writer factory
+    writer_factory: &'f mut AWF,
+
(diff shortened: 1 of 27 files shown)


djmitche:
It looks like this needs to run against a deployment with #5838 and #5847 before it will pass CI.

matt-boris:
> It looks like this needs to run against a deployment with #5838 and #5847 before it will pass CI.

I'll release to Community TC right now 👍🏻

matt-boris:
@djmitche Community TC is now upgraded to v46.0.0

djmitche:
_reruns_

clients/client-rust/download/src/hashing.rs
```diff
@@ -0,0 +1,119 @@
+use super::factory::AsyncWriterFactory;
+use anyhow::Result;
+use sha2::{Digest, Sha256, Sha512};
+use std::collections::HashMap;
+use std::pin::Pin;
+use std::sync::{Arc, Mutex};
+use std::task::{Context, Poll};
+use tokio::io::AsyncWrite;
+
+/// A hasher rader factory wraps an AsyncWriterFactory in such a way that it will capture hashes of
```
lotas, line 10:
```suggestion
/// A hasher writer factory wraps an AsyncWriterFactory in such a way that it will capture hashes of
```

lotas (approved):
wow, that's a lot of work! Thanks Dustin.
I have no idea what rust part is doing :D but rest seems to be just fine 👍

matt-boris (approved):
Wow.. thanks a lot for all this Dustin! Similar to Yarik, the rust looks pretty foreign, but it looked _sound_. 👏🏻 🎉

djmitche:
Hopefully no UB in the Rust :)

It's pretty similar to what the other languages are doing -- adds a stream implementation that can calculate hashes.  Then it calculates that hash.  And, it also looks at the `expires` property of the `getUrl` response and, if necessary, re-invokes startDownload.

PR #5862, opened 2022-12-13
[wip] Worker implementation metrics

### LINKED ISSUE #5798: Add dashboard with worker distribution status
We would want to track how many tasks are being run by which worker type.
This will give us a quantitative and visual clues when we will migrate off to generic worker

UPD: Instead of guessing worker type inside services, and then exposing this information through logs, we'd go the other way. We would use existing utils  [mozilla-history](https://github.com/taskcluster/mozilla-history/) and [community-history](https://github.com/taskcluster/community-history) to collect and expose this information

lotas:
This aims to provide extra information for the docker worker deprecation effort

This PR adds few extra metrics:

- `workerImplementation` for the `worker-pool-provisioned` and `worker-requested` events
- `workerImplementation` for the `claim-work` event

Those would be later exported as counters into grafana to visualise our progress of migrating away from docker worker to generic worker

Fixes #5798


db/access.yml
```diff
@@ -41,6 +41,7 @@ queue:
     queue_workers: write
     azure_queue_messages: write
     tasks: write
+    worker_pools: read
```
lotas, line 44: messy, but we only need to read some info about the pool if it exists

db/fns.md
```diff
@@ -2687,6 +2689,26 @@ Otherwise, page_size rows are returned at offset page_offset.
 
 Get an existing worker. The returned table will have one or (if no such worker is defined) zero rows.
 
+### get_worker_pool
```
lotas, line 2692: bringing back long deprecated friend, to give info only about the pool itself without extra stats

services/worker-manager/src/data.js
```diff
@@ -154,6 +159,22 @@ class WorkerPool {
...
   }
+
+  // guess what type of worker is being used
+  // we assume that provider only contains launch configs for the workers of the same type
+  guessWorkerType() {
+    if (!Array.isArray(this.config?.launchConfigs)) {
+      return 'unknown';
+    }
+    if (this.config.launchConfigs.some(cfg =>
+      // some worker pools with docker worker might include genericWorker config by mistake with 1-2 entries
```
lotas, line 170: this assumption was correct at the time of implementation, but then owners were notified and those worker pool configs should be fixed, so  maybe we don't need an extra check for the number of keys in that config?

services/worker-manager/test/worker_test.js
```diff
@@ -48,4 +48,75 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
       assert.equal(worker.capacity, 3);
     });
   });
+
+  suite('worker pool', () => {
```
lotas, line 52: yes, this is a copy-paste because two services have independent implementations

petemoore (commented):
I don't think the services should guess the worker implementation, perhaps workers can provide metadata about themselves, when calling workerManager.registerWorker, or when calling queue.claimWork. The worker implementations should depend on the service APIs, but I think if services start having knowledge of worker implementations, we'll introduce a cyclic dependency. Another option is that metrics are collected per workerpool, and something external splits the data into generic-worker vs docker-worker etc. But ideally if we just provide data when registering, we can include not just the implementation, but also the version, the platform, etc, so you can break down metrics by more fields than just worker implementation.

services/queue/src/data.js
```diff
@@ -537,3 +537,60 @@ class Worker {
...
+  }
+
+  // Get a worker pool from the DB, or undefined if it does not exist.
+  static async get(db, workerPoolId) {
+    return WorkerPool.fromDbRows(await db.fns.get_worker_pool(workerPoolId));
+  }
+
+  // guess what type of worker is being used
+  // we assume that provider only contains launch configs for the workers of the same type
+  guessWorkerType() {
```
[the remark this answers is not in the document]
lotas: ah, good point, thanks Pete, will change naming

lotas:
I am going to abandon this idea and instead we'll use the approach @petemoore used before:
https://github.com/taskcluster/community-history/
https://github.com/taskcluster/mozilla-history/

PR #5863, opened 2022-12-13
chore(ci): use updated worker pools

matt-boris:
> This patch switches running CI tasks on generic-worker-windows2012r2 worker pool to the new, windows 2022 worker pool.

> Use updated gw-ci-macos-10-14 worker pool.


petemoore (approved):
Love it!

Issue #5844, opened 2022-12-05
Investigate high number of failed registerWorker calls with 404/400 errors


lotas:
There is a big number of failed `registerWorker` calls.

Investigate if this contributes to overall cost overhead.

```
resource.labels.cluster_name="taskcluster-firefoxcitc-v1"
jsonPayload.Type="monitor.apiMethod"
jsonPayload.Fields.name="registerWorker"
-jsonPayload.Fields.statusCode="200"
```

djmitche:
Are those related to the issues discussed in the meeting on Monday?  If workers take a long time to start up, or have been deleted via script before they start up, then their registration may fail.

lotas:
ha! One interesting observation - lots of calls from same IP. Could it be some incorrectly configured `worker-runner` or some very stubborn worker.

<img width="1367" alt="image" src="https://user-images.githubusercontent.com/83861/205875698-8d35009a-6b05-46c5-9672-a0c13860d173.png">

which is macstadium in this case

lotas:
@davehouse can you please check who is trying to register worker form macstadium devices?

lotas:
> @lotas could you share the full request payload for an example request or two? (or email/slack them to me?) Macstadium could only tell me that the requests were from our firewall outside ip. If the request payload has the workerId that is trying to register, that will be simpler to translate to a machine.

@davehouse I would be able to show something like this only after next release. Unfortunately we don't log this information yet. 
But we are going to do more for the https://github.com/taskcluster/taskcluster/issues/2982 so I can include this information to be logged as well, and then will let you know

lotas:
@davehouse  so this is what appears in logs:
I hope it gives you more clues 
```
{"providerId":"static","workerGroup":"mslv","workerId":"macmini-m1-1nor","workerIdentityProof":{"staticSecret":"**"},"workerPoolId":"releng-hardware/gecko-t-osx-1100-m1-1nor"}
```

lotas:
thanks @davehouse, I can confirm that calls have stopped now :) 
We will try to implement something smarter in the future to allow us notice such events in time.
I will close this for now

PR #5865, opened 2022-12-14
feat(w-m): Log extra information on failed worker registration

lotas:
Required to debug some unknown workers that are trying to register unsuccessfully.  #5844


petemoore (approved):
Looks good!

PR #5866, opened 2022-12-14
Go lint updates

petemoore:
Version numbers were hidden inside `workers/generic-worker/build.sh` script so would never get auto-updated. These versions are only for CI lint checks so don't affect release bits of generic-worker, only the lint checks. This keeps them on latest, which is probably a safer bet them pinning them to a version which quickly could outdated without us noticing.

Additionally run the linter against all generic-worker engines, not just multiuser engine. Also reinstated `goimports -w` which was temporarily suppressed.

### DIFF
--- a/changelog/QXcPePhQRKm2s1stITCfWQ.md
+++ b/changelog/QXcPePhQRKm2s1stITCfWQ.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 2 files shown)


petemoore:
Notice, some of the generic-worker specific checks are probably duplicated with some of the repo-wide golangci-lint checks, so we may be able to remove some/all of them in a future PR, or think about a way of invoking all tests that apply to a given subdirectory, etc.

matt-boris (approved):
👍🏻 thanks!

PR #5870, opened 2022-12-14
Better github service error reporting

### LINKED ISSUE #5851: `/api/github/v1/repository/.../latest` returns a 500 when the provided  branch doesn't exist
Example: `https://{taskcluster_root}/api/github/v1/repository/divvun/kbdi/foobar/latest`

Full error:

```
{"EnvVersion":"2.0","Fields":{"code":422,"incidentId":"2441a83a-0255-46c5-9018-f243daff0bb6","message":"No commit found for SHA: main","method":"latest","name":"HttpError","stack":"HttpError: No commit found for SHA: main\n    at /app/node_modules/@octokit/request/dist-node/index.js:86:21\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at async Job.doExecute (/app/node_modules/bottleneck/light.js:405:18)","status":422,"v":1},"Hostname":"taskcluster-github-web-566f99f899-vlkzc","Logger":"taskcluster.github.api","Pid":1,"Severity":3,"Timestamp":1670501952432000000,"Type":"monitor.error","message":"HttpError: No commit found for SHA: main\n    at /app/node_modules/@octokit/request/dist-node/index.js:86:21\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at async Job.doExecute (/app/node_modules/bottleneck/light.js:405:18)","requestId":"5c375a20-c6cf-456a-a8fd-b2c93f266b3a","serviceContext":{"service":"github","version":"45.0.0"},"severity":"ERROR","traceId":"79a7e597-a91c-4cc2-8468-667e734d3798"}
```

It should probably return a 404 with a message telling the user the branch doesn't exist. Note that it already does that when the provided repo doesn't exist.

lotas:
Adds extra handling for the 422 error and also uses standard error handling.

Throwing errors inside api handlers seems to be not catched by middlewares and results in timeouts

Fixes #5851

### DIFF
--- a/services/github/src/api.js
+++ b/services/github/src/api.js
@@ -175,7 +175,7 @@ async function findTCChecks(github, owner, repo, branch, configuration) {
   try {
     checks = (await github.checks.listForRef({ owner, repo, ref: branch })).data.check_runs;
   } catch (e) {
-    if (e.code === 404) {
+    if (e.code === 404 || e.code === 422) {
       return [];
     }
     throw e;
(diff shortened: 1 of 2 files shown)


services/github/src/api.js
```diff
@@ -474,64 +474,64 @@ builder.declare({
...
 
-  if (instGithub) {
-    try {
-      let state;
-
-      let status = await findTCStatus(instGithub, owner, repo, branch, this.cfg);
-      if (status) {
-        state = status.state;
-      }
+  if (!instGithub) {
```
lotas, line 477: just moving the `else` block up, and rest as is

services/github/src/api.js
```diff
@@ -474,64 +474,64 @@ builder.declare({
...
+      // otherwise, it's a commit without a TC status, which probably means a new repo
+      fileConfig.headers['X-Taskcluster-Status'] = 'newrepo';
+      return res.sendFile('newrepo.svg', fileConfig);
+    }
+  } catch (e) {
+    if (e.code < 500) {
+      fileConfig.headers['X-Taskcluster-Status'] = 'error';
+      return res.sendFile('error.svg', fileConfig);
+    }
+    res.reportError('InternalError', e.message, {});
```
lotas, line 534: I wonder if we should send the original exception message back to client, or only log it

services/github/src/api.js
```diff
@@ -554,25 +554,26 @@ builder.declare({
...
+  } catch (e) {
+    if (e.code > 400 && e.code < 500) {
       return res.reply({ installed: false });
-    } catch (e) {
-      if (e.code > 400 && e.code < 500) {
-        return res.reply({ installed: false });
-      }
-      throw e;
     }
+    res.reportError('InternalError', e.message, {});
```
lotas, line 575: same here, not sure if we should send the error back

matt-boris (approved):

services/github/src/api.js
Eijebong: It would be nice if a client could figure out what went wrong without having to acess logs. Would be nice if it could report when something wasn't found.

lotas: alright, makes sense. I hope there will be nothing sensitive in those exceptions. But should be safe, as those would be most likely coming from github itself

PR #5871, opened 2022-12-14
Update dependabot.yml to include rust client and e2e ui tests

matt-boris:
I also ran a `cargo update` to avoid all the dependabot PRs to come after this lands.


lotas (approved):
:+1: thanks!

PR #5885, opened 2022-12-15
Filter logged debug information for registerWorker calls

lotas:
Remove some undesired information from logs

### DIFF
--- a/services/worker-manager/src/util.js
+++ b/services/worker-manager/src/util.js
@@ -50,3 +50,17 @@ const createCredentials = (worker, expires, cfg) => {
   });
 };
 exports.createCredentials = createCredentials;
+
+const PAYLOAD_SENSITIVE_KEYS = ['workerIdentityProof'];
+// remove sensitive keys from the request payload that would be safe for logging
+const sanitizeRegisterWorkerPayload = (obj = {}) => {
+  return Object.keys(obj).reduce((res, key) => {
+    if (PAYLOAD_SENSITIVE_KEYS.includes(key)) {
+      res[key] = '*';
+    } else {
+      res[key] = obj[key];
+    }
+    return res;
+  }, {});
+};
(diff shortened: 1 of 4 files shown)


services/worker-manager/src/util.js
```diff
@@ -50,3 +50,21 @@ const createCredentials = (worker, expires, cfg) => {
   });
 };
 exports.createCredentials = createCredentials;
+
+const PAYLOAD_SENSITIVE_KEYS = ['staticSecret', 'secret'];
```
petemoore, line 54: we could probably add `accessToken` and `password` to that list, just to be safe, what do you think?

petemoore (approved):
Looks great!

services/worker-manager/src/util.js
lotas: I checked request payload schema in documentation and it doesn't list those, but we can include those, sure

services/worker-manager/src/util.js
```diff
@@ -50,3 +50,21 @@ const createCredentials = (worker, expires, cfg) => {
   });
 };
 exports.createCredentials = createCredentials;
+
+const PAYLOAD_SENSITIVE_KEYS = ['staticSecret', 'secret', 'accessToken', 'password'];
```
djmitche, line 54: It might be confusing to see this later -- can you add a comment that these were added just-in-case, even though they are not allowed in the schema?  If the worker is sending an access token to worker-manager, we have worse design issues than just logging that information!

petemoore:
Maybe I've added more confusion here. I thought this might be a general purpose function that would be applied generally across logging, as a guard rail against logging any of the most common secret fields used in the platform. Indeed if this is only effective for a particular call, I'm fine with leaving it as it was. Agreed it is more confusing if it filters keys that never exist.

matt-boris (approved):

PR #5889, opened 2022-12-16
feat(cli): build taskcluster cli with goreleaser

matt-boris:
This fixes https://github.com/taskcluster/homebrew-tap/issues/2.

> This patch makes it so the taskcluster shell client (cli) is built with `goreleaser`.
>
> `goreleaser` also will automatically keep our [homebrew-tap](https://github.com/taskcluster/homebrew-tap/blob/main/Formula/taskcluster.rb) formula up-to-date during the release process.
>
> GitHub releases will now also contain zipped Windows executables of this cli supporting both amd64 and arm64. arm64 binaries for linux have been added as well.
>
> The darwin and linux binaries are now tarballs.

Followed https://goreleaser.com/customization/homebrew/ as well as many of their other docs.

This tool is incredibly powerful. We can use it to publish to chocolatey and snap, for example. I think all of our go builds could be migrated over to this. It handles archiving and also can create a checksum file for each file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,6 @@ __pycache__
 # docker-compose artifacts
 /docker/buckets
 docker-compose.override.yml
+
+dist/
+.DS_Store
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,89 @@
+builds:
+  - main: ./clients/client-shell
+    id: taskcluster
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - darwin
+      - linux
+      - windows
+    goarch:
+      - amd64
+      - arm64
+
+archives:
+  - id: taskcluster
+
+    # Archive name template.
+    name_template: "taskcluster-{{ .Os }}-{{ .Arch }}"
+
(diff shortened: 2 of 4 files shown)


.goreleaser.yaml
```diff
@@ -0,0 +1,89 @@
...
+    id: taskcluster
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - darwin
+      - linux
+      - windows
+    goarch:
+      - amd64
+      - arm64
```
matt-boris, line 12: I can easily add more arches here if you guys think, like 386.

lotas (approved):
Sweet! Great stuff, thanks Matt

lotas (approved):
nice :+1:

.gitignore
```diff
@@ -79,3 +79,6 @@ __pycache__
 # docker-compose artifacts
 /docker/buckets
 docker-compose.override.yml
+
+dist/
+.DS_Store
```
lotas, line 84: usually advised to put those into your global gitignore, and only keep what actually belongs to the project in the repo's .gitignore, but this is just an opionion :)

lotas (approved):

PR #5891, opened 2022-12-16
Remove duplicate workers from queue view.

### LINKED ISSUE #5890: get_queue_workers_with_wm_join returns non-unique rows, which fails respons validation
worker manager has few errors in logs:

```
Output schema validation error: 
Schema Validation Failed!
Rejecting Schema: https://community-tc.services.mozilla.com/schemas/worker-manager/v1/list-workers-response.json#
Errors:
  * data/workers should NOT have duplicate items (items ## 0 and 1 are identical)
```

On community-ci `/provisioners/proj-taskcluster/worker-types/gw-ci-macos-10-14/workers?limit=1000`

this happens because `get_queue_workers_with_wm_join` returns non-unique items after join

lotas:
This change ensures that only unique workers are being returned.

Problem: fetching workers for the queue view returned non-unique workers, i.e. same static worker that was assigned to two different worker_pool_id.

Fixes #5890

However, I'm not sure if this should be allowed in the first place.
On community we happen to have same static worker which is assigned to two different worker pools.
Maybe this is a valid scenario where worker can serve both queues?


lotas:
by looking at fxci db I can see there are multiple macmini workers assigned to different queues, so I guess this is a valid scenario

db/versions/0080.yml
```diff
@@ -0,0 +1,72 @@
...
+          queue_workers.expires,
+          queue_workers.first_claim,
+          queue_workers.recent_tasks,
+          queue_workers.last_date_active,
+          workers.state,
+          workers.capacity,
+          workers.provider_id,
+          public.gen_random_uuid()
+        from queue_workers
+        full outer join workers on workers.worker_id = queue_workers.worker_id and workers.worker_pool_id = queue_workers.task_queue_id
```
lotas, line 30: this is the part that changes - extra condition for `worker_pool_id = task_queue_id`

db/versions/0080.yml
```diff
@@ -0,0 +1,72 @@
...
```
lotas, line 65: same here, only change is `worker_pool_id = task_queue_id`

petemoore (approved):
Thanks Yarik!

I'd prefer Boris to look at it too if possible, but from my side it looks ok. 👍

matt-boris (approved):
LGTM thanks Yarik!

PR #5892, opened 2022-12-16
build(deps): Dependabot minor/patch upgrades via PMAC (Dec. 1)

matt-boris:
>
> `pmac add 5887 5881 5880 5879 5878 5876 5874 5873 5872`


lotas (approved):
Cool!

PR #5905, opened 2022-12-19
fix(workflow): use TC bot gh token to automerge

matt-boris:
This makes it so the steps in the workflow use the taskcluster-bot PAT to approve and merge these PRs, instead of the github-actions user that's used by default.


lotas (approved):
🤞

PR #5912, opened 2022-12-19
fix(changelog): don't require changelog on dep upgrades

matt-boris:
This will fix it so that `meta-changelog-pr` tasks in `client-rust` and `taskgraph` dependency upgrades won't fail like they are in https://github.com/taskcluster/taskcluster/pull/5904 and https://github.com/taskcluster/taskcluster/pull/5908.


lotas (approved):
cool :+1:

PR #5913, opened 2022-12-20
UI static asset caching and 404 response

lotas:
* Adds client cache header for static assets to avoid downloading same big files over and over.
* Adds 404 status responses for all `/api/*` and `/graphql` routes since they are not part of the UI service, but due to the UI being default backend in k8s it sometimes accepts calls for non-existing `/api/*` endpoints but  returns 200 as SPA serves everything from `index.html` 
* local docker-compose serves both public and private buckets


petemoore (approved):
Very fancy, I like it!

Is it possible to add a test?

matt-boris (approved):
Nice 👍🏻 Thanks!

lotas:
So I was investigating impact locally with running xpcshell-tests for firefox and was running a web server on expected port. After that I stopped the server and run the tests again, and nothing bad happened.
I think firefox service handles remote settings errors quite good already, and it doesn't depend on the output of the mocked server response, so this change shouldn't affect fxci pipelines. But we should probably still test it on staging

lotas:
> Is it possible to add a test?

this is only a deployment modification. So most likely we can do something in smoketests

PR #5916, opened 2022-12-21
build(deps): major version bumps using `pmac` tool

matt-boris:
>Upgrade some major version dependency bumps using `pmac` tool:
>
>`pmac add 5748 5811 5902 5903`

If you look into each of the 4 PRs I've combined (#5748 #5811 #5902 #5903), you'll see my comment as to why it should be fine to merge (typically just dropped support for old versions of Node.js).


lotas (approved):
awesome! Thanks :)

PR #5920, opened 2022-12-22
Add preinstall hook to avoid having inconsistent installs.

lotas:
This happens because taskcluster-client is linking clients/client package which is not private. And "link:" dependencies work on the "private":true packages.
Yarn is not supposed to install all such dependencies of the linked packages, and in some conditions fails with ENOENT error.


package.json
```diff
@@ -204,7 +202,8 @@
     "smoketest": "node infrastructure/tooling/src/main.js smoketest",
     "test:meta": "node infrastructure/tooling/src/main.js test:meta",
     "db:upgrade": "node db/src/main.js upgrade",
-    "db:downgrade": "node db/src/main.js downgrade"
+    "db:downgrade": "node db/src/main.js downgrade",
+    "preinstall": "cd clients/client && yarn --froze-lockfile"
```
matt-boris, line 206:
```suggestion
    "preinstall": "cd clients/client && yarn --frozen-lockfile"
```

matt-boris (changes requested):
Tiny fix! 😉

petemoore (approved):
Awesome work Yarik! ❤️

PR #5924, opened 2022-12-23
Lint all the generic-worker engines, not just multiuser

petemoore:
Simple update to lint tests to make sure all generic-worker engines are linted, not just multiuser engine.


lotas (approved):
Cool 👍

PR #5926, opened 2022-12-23
Improve monoimage by removing unnecessary files and fixing RUN commands

lotas:
* tidy-up `.dockerignore` file to add few more entries that don't belong in resulting monoimage
* install dependencies for `clients/client` to match the expected behavior from other workflows
* fixes issue with `rm -rf` that used bash expansion which doesn't work since RUN doesn't use bash

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,21 +6,31 @@ FROM node:16.19.0 as build
 RUN mkdir -p /base/cache
 ENV YARN_CACHE_FOLDER=/base/cache
 
+# prepare top level dependencies
 RUN mkdir -p /base/yarn
 COPY /yarn.lock /.yarnrc /package.json /base/yarn/
 COPY /.yarn /base/yarn/.yarn/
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -5,22 +5,33 @@ FROM node:16.19.0 as build
 
 RUN mkdir -p /base/cache
 ENV YARN_CACHE_FOLDER=/base/cache
+ENV NODE_ENV=production
```
lotas, line 8: oops, removing this from build stage, this prevents `ui` to install webpack and all of its dev-dependencies.
Other parts are already using `--production` so we can safely leave this env var in the last stage only

lotas: ok, turns out having NODE_ENV=production breaks graphql endpoints.. no idea why yet. Will remove that also

matt-boris (approved):
Nice, LGTM! Thanks 👍🏻

PR #5937, opened 2023-01-03
Prevent actions.json from being uploaded to artifacts.

### LINKED ISSUE #4950: Add ability to re-run PR checks for intermittant network issues
On a test run for PR #4948, the db-library test failed:

https://community-tc.services.mozilla.com/tasks/EfEaT4CZSAGk_wLAeneZUw

```
info There appears to be trouble with your network connection. Retrying...
error An unexpected error occurred: "https://registry.yarnpkg.com/googleapis/-/googleapis-73.0.0.tgz: unexpected end of file".
```

This appears to be an intermittent network failure. Downloading the file worked locally, and this test passed on other recent PRs.

Taskcluster has several features that would appear to address retrying tasks, but they do not work.

There is a "Re-run" link next to the failed task in GitHub. Clicking this displays the message "You have successfully requested db-library be re-run". It does not appear to run a test in Taskcluster.

github rerun

When viewing the failed task in Taskcluster, the action menu has a "Rerun" action and a "Retrigger (disabled)" action:
tc task actions

This displays a re-run page with a message: 

> Rerun a task.
> This only works on failed or exception tasks in the original taskgraph, and is CoT friendly.
> This action triggers hook project-taskcluster/in-tree-action-1-generic/f62bbc6185.

with a Action ``{}`` and a Schema:

```
properties: {}
type: object
```

Clicking "RERUN" with these defaults results in an error:

> Action is misconfigured: decision task's scopes do not satisfy in-tree:hook-action:project-taskcluster/in-tree-action-1-generic/f62bbc6185

tc task rerun failure

Ideally:

* Tasks that are known to fail due to intermittent network issues would retry, perhaps as part of the original task definition
* The "Re-run" button in GitHub would work
* The "Rerun" action in Taskcluster would work

lotas:
This file is auto-generated by taskgraph but is not configured and is not used in taskcluster repository. It relies on the hooks that should be created by third-party tools on every change of .taskcluster.yml.

Fixes #4950

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -198,7 +198,8 @@ tasks:
                         --head-repository='${repoUrl}'
                         --head-ref='${head_ref}'
                         --head-rev='${head_sha}'
-                        --head-tag='${head_tag}'
+                        --head-tag='${head_tag}' &&
+                        rm artifacts/actions.json # taskcluster is not using actions.json
                   artifacts:
                       'public':
                           type: 'directory'
--- a/changelog/issue-4950.md
+++ b/changelog/issue-4950.md
@@ -1,0 +1,6 @@
+audience: developers
+level: patch
+reference: issue 4950
+---
+
+Remove auto-generated `actions.json` which is not properly configured and is not used in this repo.


lotas:
it is possible to use use UI actions directly now, i.e in this [task](https://community-tc.services.mozilla.com/tasks/bPYEuJUnR7WCeKwwDmZhjw). Rerun and cancel work

petemoore (approved):

PR #5940, opened 2023-01-04
 Assume different role for pre-release event.

### LINKED ISSUE #5518: [github] Assume a separate role for Github pre-releases
With Github releases, you can specify that a release is a "pre-release". The intent here is that QA can then validate the release prior to it being ready to publish. Once validated, the release can be updated to a regular release.

For release automation, it might be desirable to run parts ofthe pipeline at the "pre-release" phase (e.g signing tasks), but other parts at the "release" phase (e.g actually deploying). Currently the Taskcluster Github service simply assumes a single `<org>/<repo>:release` role for all `release` events. So it's not possible to block deploy tasks via scopes on pre-releases. This means a project could inadvertently ship a "pre-release" if they don't have their tasks crafted carefully.

To resolve this, I propose the Github service assumes a separate role (e.g `<org>/<repo>:pre-release` if a "pre-release" is detected. This change would be backwards incompatible and possibly break existing pipelines without correlated `ci-config` changes. Though in practice I'd be surprised if it impacted anyone.

lotas:
This adds new variable 'action' for release payload that includes type of the release: [github docs](https://docs.github.com/developers/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=released#release)

In case of `prereleased` action `<>:pre-release` role would be assumed instead of `<>:release`.

Fixes #5518

This is possibly a breaking change if some repositories relied on different workflows?

### DIFF
--- a/services/github/src/api.js
+++ b/services/github/src/api.js
@@ -1,7 +1,7 @@
 const crypto = require('crypto');
 const { APIBuilder, paginateResults } = require('taskcluster-lib-api');
 const _ = require('lodash');
-const { EVENT_TYPES, CHECK_RUN_ACTIONS, PUBLISHERS } = require('./constants');
+const { EVENT_TYPES, CHECK_RUN_ACTIONS, PUBLISHERS, GITHUB_TASKS_FOR } = require('./constants');
 const { shouldSkipCommit, shouldSkipPullRequest } = require('./utils');
 
 // Strips/replaces undesirable characters which GitHub allows in
@@ -293,7 +293,7 @@ builder.declare({
         msg.details = getPullRequestDetails(body);
         msg.installationId = installationId;
         publisherKey = PUBLISHERS.PULL_REQUEST;
-        msg.tasks_for = 'github-pull-request';
+        msg.tasks_for = GITHUB_TASKS_FOR.PULL_REQUEST;
         msg.branch = body.pull_request.head.ref;
         break;
 
@@ -309,7 +309,7 @@ builder.declare({
         msg.details = getPushDetails(body);
         msg.installationId = installationId;
         publisherKey = PUBLISHERS.PUSH;
-        msg.tasks_for = 'github-push';
+        msg.tasks_for = GITHUB_TASKS_FOR.PUSH;
         msg.branch = body.ref.split('/').slice(2).join('/');
         break;
 
@@ -321,7 +321,7 @@ builder.declare({
         msg.details = getReleaseDetails(body);
         msg.installationId = installationId;
         publisherKey = PUBLISHERS.RELEASE;
-        msg.tasks_for = 'github-release';
+        msg.tasks_for = GITHUB_TASKS_FOR.RELEASE;
         msg.branch = body.release.target_commitish;
         break;
 
@@ -354,7 +354,7 @@ builder.declare({
         publisherKey = PUBLISHERS.RERUN;
         msg.organization = sanitizeGitHubField(body.repository.owner.login);
         msg.details = getRerunDetails(body);
-        msg.tasks_for = 'github-push';
+        msg.tasks_for = GITHUB_TASKS_FOR.PUSH;
         msg.installationId = installationId;
         msg.checkRunId = body.check_run.id;
         msg.checkSuiteId = body.check_run.check_suite.id;
--- a/services/github/src/constants.js
+++ b/services/github/src/constants.js
@@ -18,6 +18,17 @@ const GITHUB_TASKS_FOR = {
   PULL_REQUEST_UNTRUSTED: 'github-pull-request-untrusted',
 };
 
+// https://docs.github.com/developers/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=released#release
+const GITHUB_RELEASE_ACTION = {
+  CREATED: 'created',
+  DELETED: 'deleted',
+  EDITED: 'edited',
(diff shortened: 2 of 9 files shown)


lotas:
@ahal @JohanLorenzo  please check if this works for you, and let us know how much of a breaking change it would be.
Thanks!

services/github/src/api.js
```diff
@@ -321,7 +321,8 @@ builder.declare({
         msg.details = getReleaseDetails(body);
         msg.installationId = installationId;
         publisherKey = PUBLISHERS.RELEASE;
-        msg.tasks_for = 'github-release';
+        msg.tasks_for = GITHUB_TASKS_FOR.RELEASE;
+        msg.action = body.action;
```
ahal, line 325: So just to clarify, this means that there will be an `action` variable passed into the json-e template, right? If so this looks perfect!

services/github/src/tc-yaml.js
```diff
@@ -199,9 +200,11 @@ class VersionOne extends TcYaml {
           prefix + payload.details['event.base.repo.branch'],
         ];
       }
-    } else if (payload.tasks_for === 'github-release') {
+    } else if (payload.tasks_for === GITHUB_TASKS_FOR.RELEASE) {
+      // Assume separate role if this is a pre-release action
+      const role = payload.action === GITHUB_RELEASE_ACTION.PRERELEASED ? 'pre-release' : 'release';
```
ahal, line 205: Does Taskcluster process all the release events? Or just `preleased` and `released`. I'm wondering if we're scheduling a whole bunch of unnecessary graphs has Github releases are created..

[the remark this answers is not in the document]
ahal: Then again, technically this change is backwards incompatible anyway. To your earlier question, I don't expect this to cause much (if any) disruption. But it's definitely possible it breaks someone.

services/github/src/api.js
lotas: Oh! that's a good question. Let me add some tests to make sure it reaches the template. Yes, the goal was to have it there so you could use it.

services/github/src/tc-yaml.js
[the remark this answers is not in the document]
lotas: Indeed. I think we should not change the `release` one.

ahal: Follow-up question, if we do process all event types, maybe we should assume a separate role for each of them? I think you're trying to make this change as backwards compatible as possible. But since even this is backwards incompatible, I think I'd vote do it properly even it means a higher chance of something breaking.

If we do end up assuming a separate role for each action, maybe the format should be something like:
```
assume:repo:github.com/${ payload.organization }/${ payload.repository }:release:${action}
```

That way in the configuration we could grant scopes to `release:*` if we didn't want to bother doing each action individually. But then this raises another point that `pull-requests` also have an action, so maybe we should make the same change there :p.

I do think this change would be a great improvement (both for pull-requests and releases). I guess it boils down to how much time you want to spend on it. I don't think there's anything blocked on this feature atm (I initially filed the issue for VPN, but we have since worked around it).

services/github/src/constants.js
```diff
@@ -11,6 +11,24 @@ const TASK_STATE_TO_CHECK_RUN_STATE = {
...
 
+const GITHUB_TASKS_FOR = {
+  PUSH: 'github-push',
+  RELEASE: 'github-release',
+  PULL_REQUEST: 'github-pull-request',
+  PULL_REQUEST_UNTRUSTED: 'github-pull-request-untrusted',
+};
+
+// https://docs.github.com/developers/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=released#release
+const GITHUB_RELEASE_ACTION = {
```
ahal, line 22: Nice, definitely can be a follow-up, but if we're adding this here it would be good to also add `GITHUB_PULL_REQUEST_ACTION`.

services/github/src/api.js
lotas: upd: the original payload was available inside the templates already as `event` variable. So maybe I don't need to add a separate key here. 
In `.taskcluster.yml` it was possible to access it under `${event.action}` for `tasks_for == release`

services/github/src/tc-yaml.js
lotas: hmm.. that is a valid point. The webhook handler didn't filter out those events by action: https://github.com/taskcluster/taskcluster/blob/main/services/github/src/api.js#L319-L326 

Which means that all release events go through (or as many as github sends) and are published into `release` exchange. Later those are handled by the [job handler](https://github.com/taskcluster/taskcluster/blob/main/services/github/src/handlers/job.js)

So I would assume that all will end up creating new tasks. If we didn't notice any unwanted builds in the past it might be that github was not sending too many of those? I think you mentioned you saw 3 events for each publish event. Unless I'm missing something, I think it should create 3 graphs in such case.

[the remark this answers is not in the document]
lotas: alright, since this is a breaking change, we make this role match the action of the event to be consistent: `assume:repo:github.com/<owner>/<repo>:release:<action>`

services/github/src/constants.js
lotas: there are too many of those :) maybe we don't need it atm

matt-boris (approved):
👍🏻 LGTM

Issue #3924, opened 2020-11-16
worker-manager: Remove bundled root CAs after upgrading node


jwhitlock:
The worker-manager uses the [tls.rootCertificates](https://nodejs.org/docs/latest-v12.x/api/tls.html#tls_tls_rootcertificates) bundled with node. Node v12.19.0 does not include two certificates that Azure plans to use [by February 15, 2020](https://docs.microsoft.com/en-us/azure/security/fundamentals/tls-certificate-changes):

* Microsoft RSA Root Certificate Authority 2017
* Microsoft ECC Root Certificate Authority 2017

These two certs are included in node v15 (https://github.com/nodejs/node/commit/44a66adbaa9d82cf70b6569701359b9a871acd31), which synchronized with [NSS 3.56](https://hg.mozilla.org/projects/nss/raw-file/NSS_3_56_RTM/lib/ckfw/builtins/certdata.txt), released on 2020-08-21. There are two other certs added, and 4 removed.

We're bundling these two certificates, and loading the RSA certificate. The ECC certificate can't be loaded (issue #3923). When we update to node v16, or if this change is back-ported to v12, then we can remove the two certificates and the side-loading code.

PR #5942, opened 2023-01-04
Azure certs updated

### LINKED ISSUE #5941: Azure tests failing 2023/01
This task failed on azure tests: https://community-tc.services.mozilla.com/tasks/QbWDqY3nRtCfuygOitaSUA/runs/0/logs/live/public/logs/live.log

Updating certificate locally produces some other errors:

```
         fetch-in-register
           certificate download timeout:

      AssertionError [ERR_ASSERTION]: "TimeoutError: Timeout awaiting 'request' for 1ms; location=http://www.microsoft.com/pkiops/certs/Microsoft%20Azure%20TLS%20Issuing%20CA%2005%20-%20xsign.crt" == "TimeoutError: Timeout awaiting 'request' for 1ms; location=http://www.microsoft.com/pki/mscorp/Microsoft%20RSA%20TLS%20CA%2001.crt"
      + expected - actual

      -TimeoutError: Timeout awaiting 'request' for 1ms; location=http://www.microsoft.com/pkiops/certs/Microsoft%20Azure%20TLS%20Issuing%20CA%2005%20-%20xsign.crt
      +TimeoutError: Timeout awaiting 'request' for 1ms; location=http://www.microsoft.com/pki/mscorp/Microsoft%20RSA%20TLS%20CA%2001.crt
      
```

lotas:
Github Bug/Issue: Fixes #5941


petemoore (approved):
Thanks for sorting this out @lotas !

Should we create an issue for automated cert refreshing for future, so it doesn't happen again? Or is there a way the service can retrieve the azure cert at runtime, rather than it being checked into our repo?

lotas:
> Thanks for sorting this out @lotas !
> 
> Should we create an issue for automated cert refreshing for future, so it doesn't happen again? Or is there a way the service can retrieve the azure cert at runtime, rather than it being checked into our repo?

So the problem here is that I need to start manually new instance, then ssh into it, and do two curl calls to fetch  metadata and certificate.. This is annoying but is mostly needed once a year. Not sure if it is worth automation 

I'm also checking #3924 .. looks like we can also cleanup those root certs that were included in node:tls finally. But that will go with a separate branch later.

PR #5951, opened 2023-01-11
Exclude .git from dockerignore

lotas:
Some of the docker images rely on the presence of .git folder during release

### DIFF
--- a/infrastructure/tooling/src/build/tasks/client-shell.js
+++ b/infrastructure/tooling/src/build/tasks/client-shell.js
@@ -27,7 +27,7 @@ module.exports = ({ tasks, cmdOptions, credentials, baseDir, logsDir }) => {
       if (cmdOptions.staging || !cmdOptions.push) {
         // --snapshot will generate an unversioned snapshot release,
         // skipping all validations and without publishing any artifacts
-        goreleaserCmd = goreleaserCmd.push('--snapshot');
+        goreleaserCmd.push('--snapshot');
       }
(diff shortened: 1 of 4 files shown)


petemoore (dismissed):
Thanks for this Yarik!

infrastructure/tooling/src/build/tasks/client-shell.js
```diff
@@ -27,7 +27,7 @@ module.exports = ({ tasks, cmdOptions, credentials, baseDir, logsDir }) => {
       if (cmdOptions.staging || !cmdOptions.push) {
         // --snapshot will generate an unversioned snapshot release,
         // skipping all validations and without publishing any artifacts
-        goreleaserCmd = goreleaserCmd.push('--snapshot');
+        goreleaserCmd.push('--snapshot');
```
lotas, line 30: this was assigning result of `Array.push()` (that returns int) to the same variable, which changes it from array to int. 
and later fails on the command execution step

lotas:
stage release worked fine this time https://community-tc.services.mozilla.com/tasks/UGii7WokQGi1hYcjA2WcIw/runs/0/logs/live/public/logs/live.log

infrastructure/tooling/src/build/tasks/client-shell.js
matt-boris: Ah thanks for catching! 🙏🏻

PR #5954, opened 2023-01-12
feat(tools): automatically update sha256 in go generator

### LINKED ISSUE #5266: `yarn generate` should update the sha256 for Go Upgrades
Instead of having to do [these steps](https://github.com/taskcluster/taskcluster/blob/main/dev-docs/node-and-go-upgrades.md?plain=1#L37-L38), we should make `yarn generate` fetch these for us automatically.

matt-boris:
Fixes #5266.

> This patch fetches `https://go.dev/dl/?mode=json` in order to automatically update the sha256 values of each of the go binaries used in the `workers/generic-worker/gw-decision-task/tasks.yml` file.

Note: the `js-yaml` package does not support keeping comments while dumping the yaml https://github.com/nodeca/js-yaml/issues/689.


lotas (approved):
cool! Seems like less manual labor on next upgrade.

matt-boris:
> Seems like less manual labor on next upgrade.

Yes, absolutely! Previously had to go to https://go.dev/dl/ to manually copy/paste each hash.

lotas:
I remember @petemoore mentioned few times that we should eventually get rid of this separate decision task as it was created before we had taskgraph or so
would be good to have single task generating rules, not sure if there was a ticket for that yet

matt-boris:
I'm going to wait on @petemoore's look over before merging this because all of the comments being removed. I want to see if he thinks my top comment block is sufficient.

petemoore (approved):
Love it!

PR #6013, opened 2023-01-26
Ensure docker worker env var values are strings

### LINKED ISSUE #6014: docker worker should enforce that env var values are strings
Currently docker worker payloads allow an env var to be something other than a string, e.g. a bool or a var.

So a task can be subitted, with the following payload

```json
"payload": {
  "env": {
    "DEBUG": true
  }
}
```

clearly, the value should be a string, e.g.

```json
"payload": {
  "env": {
    "DEBUG": "true"
  }
}
```

Note, if someone was using python, where true is represented as `True`, then they might be surprised to see that when it reaches the task definition it became `true` instead, and I have no idea if docker worker will crash, or convert it to some string representation of `true` etc. But there is no reason not to restrict the env var value to being a string, since env vars are meant to be strings (or []byte).

petemoore:
Spotted this issue when looking at docker-worker to generic-worker conversion mechanics. I have a nasty feeling I saw this once before, and there was tasks in the wild that were setting env vars to bools or ints unintentionally - I think this is reasonable, and will probably never reach production fxci since we are not intending on deploying updated docker-workers in fxci from now on. If tasks do break though, they should be very easy to fix.

Fixes #6014.

### DIFF
--- a/changelog/issue-6014.md
+++ b/changelog/issue-6014.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 6014
+---
+Bug fix: docker worker no longer accepts non-strings for env var values in task payloads.
--- a/generated/references.json
+++ b/generated/references.json
@@ -8823,6 +8823,9 @@
           "type": "array"
         },
         "env": {
+          "additionalProperties": {
+            "type": "string"
+          },
           "description": "Example: ```\n{\n  \"PATH\": '/borked/path'\n  \"ENV_NAME\": \"VALUE\"\n}\n```",
           "title": "Environment variable mappings.",
           "type": "object"
--- a/workers/docker-worker/schemas/v1/payload.yml
+++ b/workers/docker-worker/schemas/v1/payload.yml
@@ -148,6 +148,8 @@ properties:
       }
       ```
     type: object
+    additionalProperties:
+      type: string
   maxRunTime:
     type: number
     title: Maximum run time in seconds


petemoore:
How do I still always forget the changelog?! 🤣

petemoore:
So that task did pass payload validation, and even created env vars with javascript string conversions of the underlying types, but I don't think this is good. It isn't well defined behaviour and if you pass something that isn't a string, it is very likely a bug. So I prefer to be strict and make sure an env var is a string, and if you want to convert an object to a string, you do that yourself, and put that in the payload explicitly, rather than let docker-worker guess exactly which utf8 character sequence you wanted. So to be clear, with this patch, the above task would no longer pass payload validation, and would be resolved by docker-worker as `exception/malformed-payload`, which I believe to be the correct handling.

changelog/issue-6014.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: patch
```
lotas, line 2: this could also be a breaking change, if some of the existing tasks would start failing?

lotas: or maybe not.. I wonder if we track somewhere payload validation errors

lotas (approved):

changelog/issue-6014.md
petemoore: I thought about this too, but decided since we have officially deprecated Docker Worker, and anything that is submitting such payloads are already buggy and should probably be fixed, it is reasonable to argue that it is a bugfix and that docker worker wasn't working correctly, since in any case the conversion mechanism of non-strings to strings wasn't documented or referenced, and as undocumented behaviour, under no contract of behaviour, is free to change.

lotas: Spotted one from wpt project: https://community-tc.services.mozilla.com/tasks/HsL1bxj8SAWUtzqp_vUXeA

lotas: And one more https://community-tc.services.mozilla.com/tasks/PUB9BKR1RSCJXnbZDiEN8w Project Orion

PR #6018, opened 2023-01-30
build(deps): upgrade node to 18.13.0

matt-boris:
>This upgrades Node.js to the latest LTS version available, v18.13.0. Previous version was v16.19.0 and it was scheduled to hit EoL later this year.
>
>View the release schedule [here](https://github.com/nodejs/release#release-schedule).


ui/package.json
```diff
@@ -4,13 +4,13 @@
...
+    "node": "18.13.0",
     "yarn": "^1.22.19"
   },
   "scripts": {
-    "build": "webpack --mode production",
-    "start": "cross-env GENERATE_ENV_JS=1 webpack-dev-server --mode development",
-    "start:docker": "cross-env GENERATE_ENV_JS=1 webpack-dev-server --host 0.0.0.0 --disable-host-check --hot --mode development",
+    "build": "export NODE_OPTIONS=--openssl-legacy-provider && webpack --mode production",
+    "start": "export NODE_OPTIONS=--openssl-legacy-provider && cross-env GENERATE_ENV_JS=1 webpack-dev-server --mode development",
+    "start:docker": "export NODE_OPTIONS=--openssl-legacy-provider && cross-env GENERATE_ENV_JS=1 webpack-dev-server --host 0.0.0.0 --disable-host-check --hot --mode development",
```
matt-boris, line 13: Since Node.js v17, it now [supports OpenSSL v3.0](https://nodejs.org/en/blog/release/v17.0.0/#openssl-3-0), which brings tightened restrictions on the allowed algorithms and key sizes. Because we can't upgrade webpack to the latest version (Neutrino doesn't support), we need to use the legacy provider to avoid the ERR_OSSL_EVP_UNSUPPORTED error.

package.json
```diff
@@ -179,7 +179,7 @@
     "release": "node infrastructure/tooling/src/main.js release",
     "staging-release": "node infrastructure/tooling/src/main.js staging-release",
     "release:publish": "node infrastructure/tooling/src/main.js release:publish",
-    "generate": "node infrastructure/tooling/src/main.js generate",
+    "generate": "export NODE_OPTIONS=--dns-result-order=ipv4first && node infrastructure/tooling/src/main.js generate",
```
matt-boris, line 182: Since Node.js v17, the DNS result order is returned how the resolver sent the list [PR](https://github.com/nodejs/node/pull/39987). This breaks some of our testing/infra scripts because IPv6 localhosts are now returned instead of IPv4.

libraries/api/package.json
```diff
@@ -9,7 +9,7 @@
     "coverage": "c8 yarn test",
     "coverage:report": "c8 yarn test && c8 report --temp-directory ./coverage/tmp --reporter json --report-dir ../../artifacts",
     "lint": "eslint src/*.js test/*.js",
-    "test": "mocha test/*_test.js"
+    "test": "export NODE_OPTIONS=--dns-result-order=ipv4first && mocha test/*_test.js"
```
matt-boris, line 12: Since Node.js v17, the DNS result order is returned how the resolver sent the list [PR](https://github.com/nodejs/node/pull/39987). This breaks some of our testing/infra scripts because IPv6 localhosts are now returned instead of IPv4.

clients/client-web/package.json
```diff
@@ -10,7 +10,7 @@
   ],
   "scripts": {
     "lint": "eslint src/*.js test/*.js",
-    "test": "karma start --single-run"
+    "test": "export NODE_OPTIONS=--openssl-legacy-provider && karma start --single-run"
```
matt-boris, line 13: Since Node.js v17, it now [supports OpenSSL v3.0](https://nodejs.org/en/blog/release/v17.0.0/#openssl-3-0), which brings tightened restrictions on the allowed algorithms and key sizes. Because we can't upgrade webpack to the latest version (Neutrino doesn't support), we need to use the legacy provider to avoid the ERR_OSSL_EVP_UNSUPPORTED error.

package.json
treysis: Instead of interfering with DNS order returned by the operating system, it's probably better to make use of Happy Eyeballs*, which was finally implemented in Node v18:
```suggestion
    "generate": "export NODE_OPTIONS=--enable-network-family-autoselection && node infrastructure/tooling/src/main.js generate",
```
*Happy Eyeballs is used by every web browser. It also solves issues in round-robin cases where one returned IP address isn't accessible.

libraries/api/package.json
treysis:
```suggestion
    "test": "export NODE_OPTIONS=--enable-network-family-autoselection && mocha test/*_test.js"
```
see previous comment

package.json
matt-boris: Awesome, thanks for this note @treysis!

matt-boris: @treysis it seems like this command line option is not available yet in v18:

```bash
/usr/local/bin/node: --enable-network-family-autoselection is not allowed in NODE_OPTIONS
```

matt-boris: I'll likely just revert to what we had and will come back to this if/when that CLI option is backported to v18.

@treysis do you happen to know if that's on the roadmap? I looked through some open issues/PRs and wasn't able to tell for sure.

treysis: Oops. No idea how I overlooked that. I was sure the cli option made it into 18.13. Apparently it's just the general support for Happy Eyeballs that made it into 18.13. It requires to set the option `autoSelectFamily=true` for each `net`-call separately.

Yes, it's a bit confusing. Seems to depend on https://github.com/nodejs/node/pull/45994 to make it into 18.x. It's worked on. So fingers crossed!

PR #6043, opened 2023-02-01
build(deps): Dependabot minor/patch upgrades via PMAC (Feb. 1)

matt-boris:
>Minor and patch version bumps via Dependabot using `pmac`:
>
>`pmac add 6040 6039 6038 6036 6035 6034 6032 6030 6029 6028 6027 6026 6025 6024 6023`


petemoore (approved):

PR #6044, opened 2023-02-01
Fix for 'unsupported docker v2s2 media type: application/vnd.oci.image.manifest.v1+json'

petemoore:
This is plaguing our CI at the moment, and all CI runs are currently failing due to this issue. The issue is that generic-worker docker engine was running tasks under `ubuntu:latest` using podman, but since ubuntu:latest was updated 21 hours ago, we are getting the error:

```
unsupported docker v2s2 media type: application/vnd.oci.image.manifest.v1+json
```

The "fix" here is to freeze the ubuntu version to the previous working version. We intend to drop support for the docker engine at some point, or replace it internally with the new d2g translation layer, therefore freezing this version, for something that isn't used in any production system, should be fine.

### DIFF
--- a/changelog/VvP2TgMtQ5Sjk-HEI4gjjg.md
+++ b/changelog/VvP2TgMtQ5Sjk-HEI4gjjg.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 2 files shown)


matt-boris (approved):
Awesome, thanks! 🎉

PR #6049, opened 2023-02-02
Docker worker exit code statuses now integer

petemoore:
Fixed the exit code statuses so they are now integer too.


matt-boris (approved):
Nice, thanks!

PR #6054, opened 2023-02-07
Cancel task group endpoint

### LINKED ISSUE #3652: Be able to cancel an entire Task Group
It would be nice to be able to do the operations one can do on a Task (e.g. Cancel/Retrigger/Rerun etc.) at the Task Group level. My primary use-case for this is to decrease CI load when I end up making multiple pushes in quick succession and only care about the latest running to completion.

lotas:
This introduces new method for queue service `cancelTaskGroup` which will:

* cancel all tasks that are scheduled, running or pending (similarly to what `taskcluster group cancel taskGroupId` is doing)
* shell client drops cancel task group command, api should be used instead
* new db functions: `get_task_group_size`, `cancel_task_group`
* UI action to cancel task group

Fixes #3652


docker/postgres/init.sql
```diff
@@ -11,3 +11,6 @@ CREATE USER taskcluster_queue PASSWORD 'queue_password';
 CREATE USER taskcluster_secrets PASSWORD 'secrets_password';
 CREATE USER taskcluster_web_server PASSWORD 'web_server_password';
 CREATE USER taskcluster_worker_manager PASSWORD 'worker_manager_password';
+
+CREATE DATABASE "taskcluster-test";
+GRANT ALL ON DATABASE "taskcluster-test" TO taskcluster;
```
lotas, line 16: This is an accidental addition to this PR - create separate database in local environment for tests and generation.

Problem is if you develop with single database locally and then run `yarn generate` - it would wipe out whole database. 

`export TEST_DB_URL=postgresql://postgres@localhost:5432/taskcluster-test` for `yarn generate` avoids this

services/queue/schemas/v1/cancel-task-group-response.yml
```diff
@@ -0,0 +1,40 @@
...
+  Response from a `cancelTaskGroup` request.
+type:               object
+properties:
+  taskGroupId:
+    title:          "Task-Group Identifier"
+    description: |
+      Identifier for the task-group being listed.
+    type:           string
+    pattern:        {$const: slugid-pattern}
+  totalCount:
```
lotas, line 13: I'm not sure if this is worth for inclusion as this number might change between calls, if new tasks are being added. But at the same time it could be an indicator that subsequent calls do something, as the number of tasks in the group grows

services/queue/src/api.js
```diff
@@ -300,6 +300,132 @@ builder.declare({
...
 });
 
+/** Cancel all tasks in a group */
+builder.declare({
+  method: 'post',
+  route: '/task-group/:taskGroupId/cancel',
+  name: 'cancelTaskGroup',
+  scopes: {
+    AnyOf: [
+      'queue:cancel-task-group:<schedulerId>/<taskGroupId>',
```
lotas, line 310: tricky part here is that we call this endpoint with `taskGroupId` parameter, and in order to figure out what `schedulerId` is, we have to fetch first task in the group to construct proper scopes. And since `schedulerId/taskGroupId` is unique, this should be fine

services/queue/src/api.js
```diff
@@ -300,6 +300,132 @@ builder.declare({
...
+    'This method will cancel all tasks with a given `taskGroupId` that are not resolved yet.',
+    'That means all tasks in either `unscheduled`, `pending` or `running` states.',
+    'Behaviour is similar to the `cancelTask` method.',
+    '',
+    '**Remark** this operation does not guarantee that new tasks created for this `taskGroupId`',
+    'would automatically be rejected. For example when running task keeps creating new tasks.',
+  ].join('\n'),
+}, async function(req, res) {
+  const taskGroupId = req.params.taskGroupId;
+  // allow query.reasonResolved ? cancel | superseded
```
lotas, line 329: do we need this? pass extra resolution status to the `reasonResolved`? 
Default could stay `cancelled` but for the upcoming github auto-cancellation it could become `superseded`

services/queue/src/api.js
```diff
@@ -300,6 +300,132 @@ builder.declare({
...
+  }
+
+  await req.authorize({
+    taskGroupId,
+    schedulerId: someTaskInGroup.schedulerId,
+    projectId: someTaskInGroup.projectId,
+  });
+
+  // naïve approach
+  const allTasks = await this.db.fns.get_tasks_by_task_group_projid(taskGroupId, null, null);
```
lotas, line 359: This is a simple approach to fetch all tasks for the group and then cancel those that are not yet resolved and have deadline in the future. 

Each task that was cancelled emits corresponding messages to the queue

petemoore: Although more efficient to operate in the database directly rather than to call queue.cancelTask multiple times, I'm worried operating on the db directly could introduce the risk that the scope handling of queue.cancelTask and queue.cancelTaskGroup could deviate over time. So this concern isn't so much with the current risk, more about future proofing that the two mechanisms don't deviate in terms of scope requirements.

I'm curious how much performance difference there is between operating directly on the database, and calling queue.cancelTask API instead. Perhaps the overhead isn't so great, and could ensure that the two methods don't diverge?

services/queue/src/api.js
```diff
@@ -300,6 +300,132 @@ builder.declare({
...
+
+  const response = {
+    totalCount: allTasks.length,
+    taskIds: [],
+    cancelledCount: 0,
+    taskGroupId,
+  };
+
+  const allowedStates = ['unscheduled', 'pending', 'running'];
```
petemoore, line 368: I'm curious what the resolved states are, and if a blacklist of resolved states is better future protection than a whitelist of unresolved states. But this is probably fine. If a new state got added, I'm not sure if it is better to include it in cancellations by default or not.

services/queue/src/api.js
```diff
@@ -300,6 +300,132 @@ builder.declare({
...
+
+  for (let task of allTasks) {
+    task = Task.fromDb(task);
+    // skip resolved and deadline exceeded tasks
+    if (!allowedStates.includes(task.state()) || task.deadline.getTime() < new Date().getTime()) {
```
petemoore, line 373: I don't think you need to check deadline here. If the deadline has passed, but the queue hasn't yet resolved it as deadline-exceeded, I think we would still want to resolve it as cancelled, because it could still be running.

services/queue/src/api.js
```diff
@@ -300,6 +300,132 @@ builder.declare({
...
+      continue;
+    }
+
+    // copy-paste from cancelTask method
```
petemoore, line 377: can you move this into a shared function, to avoid the copy/paste?

services/queue/src/api.js
```diff
@@ -300,6 +300,132 @@ builder.declare({
...
+      'No tasks found for task-group with taskGroupId: `{{taskGroupId}}`', {
+        taskGroupId,
+      },
+    );
+  }
+
+  await req.authorize({
+    taskGroupId,
+    schedulerId: someTaskInGroup.schedulerId,
+    projectId: someTaskInGroup.projectId,
```
petemoore, line 355: Is it a requirement that all tasks in the same taskgroup have the same projectId? I'm pretty confident the same schedulerId is required, but I didn't know `projectId` had to be the same.

petemoore (commented):
Thanks for this Yarik, looks like a great start. I have a few opinions/questions/requests...

1) It would be good to document what happens if there is an exception cancelling a task midway through. I'm not sure if that can happen, since I guess it happens in a single DB transaction, so the whole transaction would fail if there was an issue updating one task - is that correct? Is there anything we need to document here about possible partial success cases?
2) I'm not 100% sure we should operate directly on the database. Perhaps there are significant performance improvements with direct DB access, but I'm worried that the two APIs to cancel tasks may diverge in their implementation or introduce subtle differences, especially in regards to how they handle scopes.
3) I tend to think this API call would be better if it _did_ prevent future tasks from being added to the task group. Since a decision task can run for up to 17 minutes after being cancelled, it seems reasonable to me that we should be able to prevent the Queue from allowing tasks to be created with a given taskGroupId, so that that decision task would be ineffective. Without this, it seems quite easy/possible/likely that cancelTaskGroup will often be ineffective and cause frustration since a graph could be cancelled, and then 15 minutes later spawn a large number of tasks.

Although it doesn't currently exist, maybe we should expose it as an API call directly (e.g. `queue.FreezeTaskGroup`), which `queue.cancelTaskGroup` could call under the hood. But e.g. the decision task could also call it in taskgraph, after it finishes adding all tasks, as some offer of safety that the task graph isn't later modified.

Maybe we can discuss this with the broader community in the Taskcluster Community Meeting later today.

docker/postgres/init.sql
petemoore: Does this get run on production deployments too?

services/queue/schemas/v1/cancel-task-group-response.yml
petemoore: Maybe `taskGroupSize`? Naming is so hard. I think it isn't obvious what "total" refers to...

services/queue/schemas/v1/cancel-task-group-response.yml
```diff
@@ -0,0 +1,40 @@
...
+    type:           integer
+    minimum:        0
+    title:          "Total number of tasks"
+    description: |
+      Total number of tasks that existed at the moment of the call.
```
petemoore, line 18: "Current count of tasks in the task group."?

services/queue/schemas/v1/cancel-task-group-response.yml
```diff
@@ -0,0 +1,40 @@
...
+  cancelledCount:
+    type:           integer
+    minimum:        0
+    title:          "Total number of cancelled tasks"
+    description: |
+      Total number of tasks that were cancelled with this call.
+      It includes all non-resolved tasks with deadline in the future.
```
petemoore, line 25: Let's omit "deadline in the future" in the description and in the code. Deadline handling is the responsibility of a different component, and I think it is cleanest to uncouple it from here.

services/queue/schemas/v1/cancel-task-group-response.yml
```diff
@@ -0,0 +1,40 @@
...
+  taskIds:
+    type:           array
+    title:          "List of task IDs"
+    description: |
+      List of IDs of tasks that were cancelled with this call
```
petemoore, line 30:
```suggestion
      List of `taskIds` cancelled by this call.
```

docker/postgres/init.sql
petemoore: Do we need to update any of the developer docs to explain how multiple local databases should be used, and any local config that should be applied for a working system?

services/queue/schemas/v1/cancel-task-group-response.yml
lotas: suggestion accepted! Thanks!

services/queue/src/api.js
lotas: true! Thanks

services/queue/src/api.js
lotas: Yes, i'm rewriting this

services/queue/src/api.js
lotas: I don't think there is a strict rule, code doesn't enforce all tasks to have same projectId in the group. But according to docs this is the mechanism to set up the permission system. In the call above `taskGroupId`, `schedulerId` and `projectId` would be injected in scope templates:
```
 'queue:cancel-task-group:<schedulerId>/<taskGroupId>',
 'queue:cancel-task-in-project:<projectId>',
```
to see if current user has permissions. 

This mimics the behavior of the `cancelTask` method which uses similar scopes:
```
    'queue:cancel-task:<schedulerId>/<taskGroupId>/<taskId>',
    'queue:cancel-task-in-project:<projectId>',
```

services/queue/schemas/v1/cancel-task-group-response.yml
lotas: applied, thanks

services/queue/src/api.js
petemoore: So the issue here is that the code takes an arbitrary task from the task group to see what projectId it has. If scopes are sufficient for _that_ task, then the transaction is approved for _all_ tasks in that task group, even if the client that calls this endpoint doesn't have the scopes to delete all the tasks. This should be changed to return the unique projectIds, and require `queue:cancel-task-in-project:<project>` for _all_ projectIds that exist in the set of tasks that it would like to cancel. This is one of the reasons I am wary of implementing this directly in the database, and not gating it on the `cancelTask` implementation (which would ensure consistent scope treatment).

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -27,6 +27,7 @@ import HelpView from '../../../components/HelpView';
 import TaskGroupProgress from '../../../components/TaskGroupProgress';
 import TaskGroupTable from '../../../components/TaskGroupTable';
 import TaskActionForm from '../../../components/TaskActionForm';
+import Snackbar from '../../../components/Snackbar';
```
petemoore, line 30: Thanks. Now I'm feeling hungry.

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -499,6 +547,20 @@ export default class TaskGroup extends Component {
...
   };
 
+  handleSnackbarOpen = ({ message, variant = 'success', open }) => {
+    this.setState({ snackbar: { message, variant, open } });
+  };
+
+  handleSnackbarClose = (event, reason) => {
+    if (reason === 'clickaway') {
+      return;
+    }
```
petemoore, line 557: Don't close the snack bar!!! 😆

petemoore (commented):
Awesome work Yarik!

ui/src/views/Tasks/TaskGroup/index.jsx
lotas: that was not a goal here. Focus :D

changelog/issue-3652.md
```diff
@@ -0,0 +1,10 @@
+audience: users
+level: major
+reference: issue 3652
+---
+
+It is now possible to cancel sealed task group by using `cancelTaskGroup` method.
```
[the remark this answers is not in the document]
lotas: thanks

changelog/issue-3652.md
```diff
@@ -0,0 +1,10 @@
...
+This will cancel all scheduled/pending/running tasks within given group.
+
+`taskcluster group cancel` was removed from shell client,
+as it is available under `taskcluster api queue cancelTaskGroup` now.
```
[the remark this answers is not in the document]
lotas: Good point. 
Maybe worth then keeping the old version because it was doing something different. But probably add a warning that this method is deprecated to give anyone who's using it heads up. 👍

services/queue/src/api.js
```diff
@@ -309,6 +309,117 @@ builder.declare({
...
+    taskGroupSize: taskGroupSize[0].get_task_group_size,
+    taskIds: [],
+    cancelledCount: 0,
+    taskGroupId,
+  };
+
+  for (let task of allTasks) {
+    task = Task.fromDb(task);
+    await cancelSingleTask(task, this);
+    response.cancelledCount++;
```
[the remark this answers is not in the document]
lotas: no, it should only count the ones that were actually cancelled with this call.

services/queue/src/api.js
```diff
@@ -309,6 +309,117 @@ builder.declare({
...
+  const response = {
+    taskGroupSize: taskGroupSize[0].get_task_group_size,
+    taskIds: [],
+    cancelledCount: 0,
+    taskGroupId,
+  };
+
+  for (let task of allTasks) {
+    task = Task.fromDb(task);
+    await cancelSingleTask(task, this);
```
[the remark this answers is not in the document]
lotas: Canceling is done in the DB directly, this is only to notify pulse subscribers and tell queue service that this task was resolved with exception, so it can update tracked dependencies. 

I don't think it should be a big hit on performance, but I didn't test it on tasks groups with thousands tasks.
The call itself might take a bit longer, but for sure faster than all individual task cancel api calls. I can try to put some stress test to see

services/queue/src/api.js
```diff
@@ -309,6 +309,117 @@ builder.declare({
...
+  output: 'cancel-task-group-response.yml',
+  title: 'Cancel Task Group',
+  description: [
+    'This method will cancel all unresolved tasks (`unscheduled`, `pending` or `running` states)',
+    'with the given `taskGroupId`. Behaviour is similar to the `cancelTask` method.',
+    '',
+    'It is only possible to cancel a task group if it has been sealed using `sealTaskGroup`.',
+    'If the task group is not sealed, this method will return a 409 response.',
+    '',
+    'Every task that was canceled with will trigger a `task-exception` message.',
```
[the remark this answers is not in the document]
lotas: Hmm.. I'll check, I think  I saw it was mentioned somewhere as well in docs, but I might be wrong.

services/queue/src/api.js
```diff
@@ -309,6 +309,117 @@ builder.declare({
...
+
+  return status;
+};
+
+/** Cancel all tasks in a group */
+builder.declare({
+  method: 'post',
+  route: '/task-group/:taskGroupId/cancel',
+  name: 'cancelTaskGroup',
+  scopes: 'queue:cancel-task-group:<schedulerId>/<taskGroupId>',
```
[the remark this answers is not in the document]
lotas: After some few other discussions we decided to require all of the scopes:

```
queue:seal-task-group:scheduler-id:<schedulerId>
queue:seal-task-group:task-group-id:<taskGropuId>
queue:seal-task-group:project-id:<projectId>
```

This way if someone wants to have flexibility, role with few scopes using `:*` can be used while some scope will be strict.

clients/client-shell/cmds/group/actors.go
```diff
@@ -18,16 +16,6 @@ import (
 var listFormat string
 
 func init() {
-	cancelCmd := &cobra.Command{
-		Use:   "cancel <taskGroupId>",
-		Short: "Cancel a whole group by taskGroupId.",
-		RunE:  executeHelperE(runCancel),
-	}
```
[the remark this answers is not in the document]
lotas: yes, agree. Restoring it

db/fns.md
```diff
@@ -1389,6 +1393,30 @@ reason.  If the task is unscheduled, a run with that status is
...
+  * `  expires timestamptz`
+  * `  retries_left integer`
+  * `  routes jsonb`
+  * `  runs jsonb`
+  * `  taken_until timestamptz `
+* *Last defined on version*: 82
+
+This cancels all non-resolved tasks for the given task group
+by calling existing cancel_task() procedure. It will return
+only the tasks that were cancelled with this call.
```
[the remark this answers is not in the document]
lotas: I'm not sure, this is just a function that does one thing. Group sealing logic is handled on the API level, so probably not relevant.

db/test/fns/queue_test.js
```diff
@@ -644,6 +644,115 @@ suite(testing.suiteName(), function() {
...
+        await create(db, { taskId: slugid.v4(), taskGroupId });
+        const res2 = await db.fns.get_task_group_size(taskGroupId);
+        assert.equal(res2.length, 1);
+        assert.equal(res2[0].get_task_group_size, 2);
+      });
+    });
+
+    suite('cancel_task_group', function() {
+      helper.dbTest('no such task group', async function(db) {
+        const res = await db.fns.cancel_task_group(taskId, 'because');
```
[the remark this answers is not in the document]
lotas: This is fine. Those values are hardcoded in the calling api handler, and are not user-defined values. 
Same behavior as the `cancelTask` method

services/queue/src/api.js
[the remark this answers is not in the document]
lotas: True, I'll drop this line. 
Question is though: seal task group emits new event `seal-task-group` but this one doesn't.. do you think we need a `cancel-task-group` event published to exchange in addition to the individual tasks?

services/queue/src/api.js
```diff
@@ -309,6 +309,117 @@ builder.declare({
   return res.reply(result);
 });
 
+const cancelSingleTask = async (task, ctx) => {
+  // Get the last run, there should always be one
+  let run = _.last(task.runs);
+  if (!run) {
+    let err = new Error('There should exist a run after cancelTask!');
```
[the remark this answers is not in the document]
lotas: oh! and I just noticed that there was a bug in previous version as well. Down below the lines there was early return or check later if run is defined. 
```
if (!run) {
 report error
} 
.. if (run.state == 'xx') .. 
```

if error ever happened, then the whole call was probably a `500`, gonna add a check

services/queue/src/api.js
```diff
@@ -309,6 +309,117 @@ builder.declare({
...
+  if (!taskGroup) {
+    return res.reportError('ResourceNotFound',
+      'No task-group with taskGroupId: `{{taskGroupId}}`', {
+        taskGroupId,
+      },
+    );
+  }
+  if (!taskGroup.sealed) {
+    return res.reportError('RequestConflict',
+      'Cannot cancel unsealed task group: `{{taskGroupId}}`', {
```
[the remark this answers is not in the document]
lotas: yay!

services/queue/src/api.js
```diff
@@ -309,6 +309,117 @@ builder.declare({
...
+        taskGroupId,
+      },
+    );
+  }
+
+  await req.authorize({
+    taskGroupId,
+    schedulerId: taskGroup.schedulerId,
```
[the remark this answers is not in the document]
lotas: no, let's keep it :)
`'queue:cancel-task-group:<schedulerId>/<taskGroupId>'`

from the discussion above

ui/src/views/Tasks/TaskGroup/index.jsx
lotas: sorry, it's all user's fault

services/queue/src/api.js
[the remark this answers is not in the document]
lotas: @djmitche here is a brainstorming of how ideal scope should look like :)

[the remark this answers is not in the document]
djmitche: Per https://docs.taskcluster.net/docs/manual/tasks/groups, schedulerId _can_ be used to control who can manipulate tasks, but projectId is preferred (in fact, projectId was invented because schedulerId was doing too many things and impractical for this purpose).

That said, I think it makes sense to make this method parallel to [queue.cancelTask](https://docs.taskcluster.net/docs/reference/platform/queue/api#cancelTask):

any of
* queue:cancel-task:<schedulerId>/<taskGroupId>/<taskId>; or
* queue:cancel-task-in-project:<projectId>;
* all of
  * queue:cancel-task; and
  * assume:scheduler-id:<schedulerId>/<taskGroupId>

The third of those ("all of..") is deprecated, so no need to duplicate that. You've said you'll introduce project-id support later, which will correspond to the second. So, I think it makes sense to mirror the first option here, which is option 1 above.

This allows a few things:
 * `queue:$verb-task-group:*` -> cancel/seal anything you want
 * `queue:$verb-task-group:tc-github/*` -> cancel/seal tasks created by tc-github
 * `queue:$verb-task-group:$schedulerId/$taskGroupId` - permission to cancel this specific task group. This might be granted in temporary credentials, for example.

clients/client-shell/cmds/group/actors.go
```diff
@@ -21,7 +21,12 @@ func init() {
 	cancelCmd := &cobra.Command{
 		Use:   "cancel <taskGroupId>",
 		Short: "Cancel a whole group by taskGroupId.",
-		RunE:  executeHelperE(runCancel),
+		Long: "This method fetches all tasks in the given task group, and cancels\n" +
+			"all tasks with respect to worker-type filter if given.\n\n" +
+			"Note that there is a more efficient way to cancel a task group with API:\n\n" +
+			"taskcluster api queue sealTaskGroup <taskGroupId> # seal task group \n" +
```
matt-boris, line 27: I feel like we should mention how this step is _required_ for being able to cancel the task group. Maybe something like this?
```suggestion
			"taskcluster api queue sealTaskGroup <taskGroupId> # seal task group is required before calling cancelTaskGroup\n" +
```

services/queue/src/api.js
```diff
@@ -1014,36 +1129,7 @@ builder.declare({
...
-    );
-
-    // Publish message about the exception
-    const runId = task.runs.length - 1;
-    await this.publisher.taskException(_.defaults({
-      status,
-      runId,
-    }, _.pick(run, 'workerGroup', 'workerId')), task.routes);
-    this.monitor.log.taskException({ taskId, runId });
-  }
```
matt-boris, line 1046: Nice pulling out into a func since it's shared now 👍🏻

services/queue/test/canceltaskgroup_test.js
```diff
@@ -0,0 +1,173 @@
...
+    routes: [],
+    retries: 5,
+    created: taskcluster.fromNowJSON(),
+    deadline: taskcluster.fromNowJSON('1 day'),
+    scopes: [],
+    payload: {},
+    metadata: {
+      name,
+      description: 'Task created during unit tests',
+      owner: 'jonsafj@mozilla.com',
```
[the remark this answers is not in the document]
matt-boris: Just gonna leave it to keep the history there? 😄

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -448,8 +476,28 @@ export default class TaskGroup extends Component {
...
+        data: { cancelTaskGroup },
+      } = await apolloClient.mutate({
+        mutation: cancelTaskGroupQuery,
+        variables: {
+          taskGroupId: taskGroup.taskGroup.taskGroupId,
+        },
+      });
+
+      this.handleSnackbarOpen({
+        message: `Tasks cancelled: ${cancelTaskGroup.cancelledCount} out of ${cancelTaskGroup.taskGroupSize}.`,
```
matt-boris, line 499: I wonder if the case where `cancelTaskGroup.cancelledCount < cancelTaskGroup.taskGroupSize`, additional info could be added into the message letting the user know they can try calling again to cancel remaining tasks.

matt-boris (commented):
Looking great Yarik! Nice work! 🎉

Left a few small comments/suggestions!

petemoore (approved):
Thanks Yarik!

clients/client-shell/cmds/group/actors.go
lotas: Thanks! WIll add

db/fns.md
```diff
@@ -1662,6 +1690,17 @@ where the page of results should begin, and must all be specified if any
...
+### get_task_group_size
+
+* *Mode*: read
+* *Arguments*:
+  * `task_group_id_in text`
+* *Returns*: `integer`
+* *Last defined on version*: 82
+
+Return number of tasks that exist in the current task group
+at the moment of the call.
```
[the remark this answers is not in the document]
lotas: Thanks! will remove now

db/versions/0082.yml
```diff
@@ -0,0 +1,74 @@
...
+                tasks.taken_until
+              FROM tasks
+              WHERE tasks.task_id = task_record.task_id;
+          END IF;
+        END LOOP;
+      end
+  get_task_group_size:
+    description: |-
+      Return number of tasks that exist in the current task group
+      at the moment of the call.
```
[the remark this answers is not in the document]
lotas: yes! thanks, Pete also pointed out in the generated code, gonna drop it 👍

services/queue/test/canceltaskgroup_test.js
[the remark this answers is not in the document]
lotas: alright, gonna rename

ui/docs/manual/tasks/groups.mdx
```diff
@@ -41,3 +41,16 @@ This can help to save resources.
...
 Once a task group is sealed, it will no longer accept new tasks and will throw
 an error if new task is being added to it.
+
+
+## Cancelling Task Groups
+
+To cancel all tasks within a single task group, utilize the queue.cancelTaskGroup method.
+This method cancels all tasks that are yet to be resolved,
+while tasks that have already expired will remain unchanged.
+It is important to note that this operation is only possible when the task group has been sealed.
```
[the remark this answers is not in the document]
lotas: Thanks Matt, will add

ui/src/views/Tasks/TaskGroup/index.jsx
lotas: So if this happens, it probably means that some tasks were already resolved, or task group was already cancelled. As `cancelTaskGroup` would only cancel non-resolved ones. 
And we already show this in a message before user clicks the action button 

https://github.com/taskcluster/taskcluster/blob/67139383d1db65eb27f2eaeeaf2f100d69fd8e90/ui/src/views/Tasks/TaskGroup/index.jsx#L83-L87

services/queue/src/api.js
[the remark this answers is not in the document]
lotas: Thanks Dustin!

I wonder if we should already add project specific scopes now or do in the follow up PR

matt-boris (dismissed):
Looks great Yarik! Thanks 🎉

matt-boris (approved):

petemoore (approved):
Thanks Yarik!

PR #6060, opened 2023-02-14
Seal Task Group

### LINKED ISSUE #6059: Implement sealTaskGroup method
During implementation of #3652  we had an idea that it would make sense to have functionality to "seal" task group, that would make it impossible to add new tasks for this group.

This should might be used together with `cancelTaskGroup` to save resources and be sure that new tasks wouldn't be added from some decision tasks that are still running.

Discussion points:
* should this be reversible? `sealTaskGroup` + `unsealTaskGroup`?
* does it need any new scope like `queue:seal-task-group:<taskGroupId>` ?

lotas:
Seal task group:

* [x] adds new method `queue.sealTaskGroup(taskGroupId)`
* [x] changes `queue.createTask()` to throw `409` error when task group is sealed
* [x] adds new db functions `get_task_group2`, `get_task_group_project_ids`, `is_task_group_sealed`, `seal_task_group`
* [x] updates documentation with new section "Task Group"
* [x] fixes small bug with `docker-compose.dev.yml` - adds mapping to `generated/` folder so database could parse latest schema and functions
* [x] Display information in the UI if task group is sealed
* [x] Add "seal task group" action in the UI for the task group
* [x] `taskGroupResolved`  event includes `expires` + `sealed` fields
* [x] `listTaskGroup` includes task group meta fields: `schedulerId`, `expires`, `sealed`

image

Github Bug/Issue: Fixes #6059


ui/docs/manual/tasks/groups.mdx
```diff
@@ -0,0 +1,26 @@
...
+filename: tasks/groups.mdx
+title: Task Groups
+order: 50
+---
+
+# Task Groups
+
+Task Groups are a way to group tasks together based on their `taskGroupId`.
+When multiple tasks are being created at once, they often have the same `taskGroupId`.
+This is useful in situations such as when tasks are created for a Github
```
matt-boris, line 11:
```suggestion
When multiple tasks are created simultaneously, they often have the same `taskGroupId`.
This is useful in situations where tasks are created for a Github
```

matt-boris (commented):
Looks great! Nice!

Are you planning on tackling those last two checkboxes in your PR description in this PR? If not, I'll go ahead and approve. 👍🏻

ui/docs/manual/tasks/groups.mdx
lotas: thanks!

matt-boris (dismissed):
Awesome!

clients/client-shell/cmds/group/actors.go
```diff
@@ -297,3 +307,48 @@ func filterListTask(status tcqueue.TaskStatusStructure, flags *pflag.FlagSet) bo
...
+
+	fmt.Fprintf(out, `Task Group sealed:
+taskGroupId: %s
+schedulerId: %s
+expires:     %s
+sealed:      %s
+`, updated.TaskGroupID, updated.SchedulerID, updated.Expires, updated.Sealed)
+
+	return nil
+}
```
petemoore, line 354: Same as above - I don't think this is needed either, since it is covered by `taskcluster api queue sealTaskGroup <taskGroupId>` already.

clients/client-shell/cmds/group/actors.go
```diff
@@ -36,6 +36,16 @@ func init() {
...
+	sealCmd := &cobra.Command{
+		Use:   "seal <taskGroupId>",
+		Short: "Seal the task group to disallow addition of new tasks",
+		Long:  "This operation is irreversible and calling it multiple times will only seal it once.",
+		RunE:  executeHelperE(runSeal),
+	}
+	sealCmd.Flags().BoolP("force", "f", false, "Skip sealing confirmation.")
+
+	Command.AddCommand(sealCmd)
+
```
petemoore, line 48: The new API method will already be available via `taskcluster api queue sealTaskGroup <taskGroupId>`, so I don't think this is needed.

clients/client-shell/cmds/group/actors_test.go
```diff
@@ -105,6 +106,17 @@ func listTaskGroupHandler(w http.ResponseWriter, _ *http.Request) {
...
+func sealHandler(w http.ResponseWriter, _ *http.Request) {
+	status := `{
+		"schedulerId": "test-scheduler",
+		"taskGroupId": "e4WPAAeSdaSdKxeWzDCBA",
+		"expires": "2023-01-30T15:49:31.389Z",
+		"sealed": "2023-01-30T16:49:31.389Z"
+	}`
+	_, _ = io.WriteString(w, status)
+}
+
```
petemoore, line 119: Same here...

clients/client-shell/cmds/group/actors_test.go
```diff
@@ -125,6 +137,18 @@ func (suite *FakeServerSuite) TestRunCancel() {
...
+	buf, cmd := setUpCommand()
+	cmd.Flags().Bool("force", true, "")
+
+	// run the command
+	args := []string{fakeGroupID}
+	assert.NoError(suite.T(), runSeal(&tcclient.Credentials{}, args, cmd.OutOrStdout(), cmd.Flags()))
+
+	suite.Equal("Task Group sealed:\ntaskGroupId: e4WPAAeSdaSdKxeWzDCBA\nschedulerId: test-scheduler\nexpires:     2023-01-30T15:49:31.389Z\nsealed:      2023-01-30T16:49:31.389Z\n", buf.String())
+}
+
```
petemoore, line 151: Same here...

db/fns.md
```diff
@@ -93,13 +93,15 @@
    * [`get_dependent_tasks`](#get_dependent_tasks)
    * [`get_queue_artifact`](#get_queue_artifact)
    * [`get_queue_artifacts_paginated`](#get_queue_artifacts_paginated)
-   * [`get_task_group`](#get_task_group)
+   * [`get_task_group_project_ids`](#get_task_group_project_ids)
+   * [`get_task_group2`](#get_task_group2)
```
petemoore, line 97: Is there a way to avoid having the 2 in the name?

db/fns.md
```diff
@@ -93,13 +93,15 @@
...
    * [`get_task_projid`](#get_task_projid)
    * [`get_task_queue`](#get_task_queue)
    * [`get_task_queues`](#get_task_queues)
    * [`get_tasks_by_task_group_projid`](#get_tasks_by_task_group_projid)
    * [`is_task_blocked`](#is_task_blocked)
    * [`is_task_group_active`](#is_task_group_active)
+   * [`is_task_group_sealed`](#is_task_group_sealed)
```
petemoore, line 104: What is the difference between is_task_group_active and is_task_group_sealed? I wonder if is_task_group_active was the old way to guess if a task group was still in progress, and perhaps now we can use the new method and delete the old one entirely?

db/fns.md
```diff
@@ -1658,18 +1664,30 @@ where the page of results should begin, and must all be specified if any
 are specified.  Typically these values would be drawn from the last item
 in the previous page.
 
-### get_task_group
+### get_task_group_project_ids
```
petemoore, line 1667: should this be `get_sealed_task_group_project_ids` ? i feel like it would be safer to only allow it for sealed task groups.

db/fns.md
```diff
@@ -1658,18 +1664,30 @@ where the page of results should begin, and must all be specified if any
...
+
+* *Mode*: read
+* *Arguments*:
+  * `task_group_id_in text`
+* *Returns*: `table`
+  * `project_id text`
+* *Last defined on version*: 81
+
+Get unique projectIds of the tasks in the task group
```
petemoore, line 1676: In this case we should update description too

db/fns.md
```diff
@@ -2046,6 +2088,10 @@ client side.
 Update a queue artifact, including its storageType.
 Returns the up-to-date artifact row that have the same task id, run id, and name.
 
+### deprecated methods
+
+* `get_task_group(task_group_id_in text)` (compatibility guaranteed until v49.0.0)
```
petemoore, line 2093: I don't think the database schema is a defined API into our system - the only approved route to collect data is via API calls, so I think we are free to change the database function inputs and outputs. I guess this is why we now have get_task_group2 - but I don't think it is needed.

db/test/fns/queue_test.js
```diff
@@ -125,14 +125,39 @@ suite(testing.suiteName(), function() {
         await db.fns.ensure_task_group('0cM7dCL2Rpaz0wdnDG4LLg', 'sched', expires);
         await db.fns.ensure_task_group('jcy-h6_7SFuRuKLPByiFTg', 'sched', expires);
 
-        const tgs = await db.fns.get_task_group('jcy-h6_7SFuRuKLPByiFTg');
+        const tgs = await db.fns.get_task_group2('jcy-h6_7SFuRuKLPByiFTg');
```
petemoore, line 128: i don't think this needs to change (see above - we should be free to change the inputs/outputs of our database functions as they are not exposed as a data interface directly.

db/test/fns/queue_test.js
```diff
@@ -125,14 +125,39 @@ suite(testing.suiteName(), function() {
...
+        assert.equal(false, isSealed[0].is_task_group_sealed);
+
+        const [tg2] = await db.fns.seal_task_group(taskGroupId);
+        assert.equal(tg2.task_group_id, taskGroupId);
+        assert.notEqual(tg2.sealed, null);
+
+        const isSealed2 = await db.fns.is_task_group_sealed(taskGroupId);
+        assert.equal(true, isSealed2[0].is_task_group_sealed);
+
+        // multiple calls should not change sealed timestamp
```
petemoore, line 154: ❤️

db/versions/0081.yml
```diff
@@ -0,0 +1,113 @@
...
+        sealed timestamptz
+      )
+    body: |-
+      declare
+        updated_row task_groups%ROWTYPE;
+      begin
+        UPDATE task_groups
+        SET sealed = NOW()
+        WHERE task_groups.task_group_id = task_group_id_in
+        AND task_groups.sealed IS NULL
```
petemoore, line 89: nice

docker-compose.dev.yml
```diff
@@ -36,6 +36,7 @@ services:
       DEBUG: '*'
     volumes:
       - ./db:/app/db
+      - ./generated:/app/generated
```
petemoore, line 39: is this related to this PR?

infrastructure/tooling/src/generate/generators/docker-compose.js
```diff
@@ -240,6 +240,7 @@ exports.tasks.push({
       },
       volumes: [
         './db:/app/db', // in case of new migrations
+        './generated:/app/generated', // db schema
```
petemoore, line 243: should this be a different PR?

infrastructure/tooling/src/generate/generators/go-version.js
```diff
@@ -77,7 +77,7 @@ exports.tasks = [{
       ).replace(
         /go[0-9]+\.[0-9]+\.[0-9]+/g,
         `${goVersion}`));
-    const goDownloadsJson = await (await fetch('https://go.dev/dl/?mode=json')).json();
+    const goDownloadsJson = await (await fetch('https://go.dev/dl/?mode=json&include=all')).json();
```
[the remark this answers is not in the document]
petemoore: i think this has landed in a different PR, and shouldn't be part of this one.

services/queue/schemas/v1/task-group-response.yml
```diff
@@ -0,0 +1,31 @@
...
+properties:
+  taskGroupId:
+    title:          "Task-Group Identifier"
+    description: |
+      Identifier for the task-group being sealed.
+    type:           string
+    pattern:        {$const: slugid-pattern}
+  sealed:
+    type:           string
+    format:         date-time
```
petemoore, line 15: does this allow a null/empty value?

services/queue/schemas/v1/task-group-sealed.yml
```diff
@@ -0,0 +1,25 @@
+$schema: "/schemas/common/metaschema.json#"
+title:              "Task Group Sealed Message"
+description: |
+  Message written once a task group has been sealed.
```
petemoore, line 4: Is there a reason not to return the full task group response (e.g. including the sealed time?) (i.e. reuse the other json schema and not have a different one)

services/queue/src/api.js
```diff
@@ -550,6 +658,10 @@ builder.declare({
     '**Scopes**: Note that the scopes required to complete this API call depend',
     'on the content of the `scopes`, `routes`, `schedulerId`, `priority`,',
     '`provisionerId`, and `workerType` properties of the task definition.',
+    '',
+    'If the task group was sealed, this end-point will return `409` reporting',
+    '`RequestConflict` to indicate that it is no longer possible to add new tasks',
+    'for this `taskGroupId`.',
```
petemoore, line 664: Nice!

services/queue/src/exchanges.js
```diff
@@ -358,3 +358,19 @@ exchanges.declare({
...
+  description: [
+    'A message is published on task-group-sealed whenever task group is sealed.',
+    'This task group will no longer allow creation of new tasks.',
+  ].join('\n'),
+  routingKey: buildTaskGroupRoutingKey(),
+  schema: 'task-group-sealed.json#',
+  messageBuilder: commonMessageBuilder,
+  routingKeyBuilder: taskGroupRoutingKeyBuilder,
+  CCBuilder: commonCCBuilder,
+});
```
petemoore, line 376: nice!!!

ui/docs/manual/tasks/groups.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+Sometimes, it may be necessary to block the entire task group from accepting new tasks.
+For example, if multiple push events occur on the same branch in Github
+and create separate task groups, we can stop previous runs and seal the entire
+task group to prevent currently running decision tasks from adding new ones.
+This can help to save resources.
+
+To seal a task group, you can use the `queue.sealTaskGroup` method.
+Once a task group is sealed, it will no longer accept new tasks and will throw
+an error if new task is being added to it.
```
petemoore, line 26: nice!!!

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -672,14 +749,36 @@ export default class TaskGroup extends Component {
...
+                  <CopyToClipboardListItem
+                    tooltipTitle={tgInfo.sealed}
+                    textToCopy={tgInfo.sealed}
+                    primary="Task Group Sealed"
+                    secondary={<DateDistance from={tgInfo.sealed} />}
+                  />
+                )}
+              </Grid>
+            </Grid>
+          </React.Fragment>
```
petemoore, line 781: This is awesome!!!

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -745,7 +844,8 @@ export default class TaskGroup extends Component {
                 tooltipOpen
                 key={action.title}
                 FabProps={{
-                  disabled: actionLoading,
+                  disabled:
+                    actionLoading || this.groupActionDisabled(action.name),
```
petemoore, line 848: What does this do?

petemoore (commented):
Hey Yarik, this is incredible work. Amazing how you've put this together so quickly and thoroughly. Love it!

clients/client-shell/cmds/group/actors.go
lotas: true! Thanks

db/fns.md
lotas: this is mostly same function but with changed return signature. I've seen few other functions were using same approach. 
Technically we can change to something like `get_task_group_full` but I think it's fine.

db/fns.md
lotas: This is something different. `is_task_group_active` checks all tasks for the given `taskGroupId` and returns true if there are non-resolved tasks

db/fns.md
lotas: Problem here is that we only need this to check permissions :) To know if someone should be able to perform this action. 

We could as well drop those project scopes and ignore this for sealing

db/fns.md
lotas: this is the description. You mean name of the function?

db/fns.md
lotas: This is a convention. If one function is deprecated, we create new one with new name, and mark previous one as deprecated. 
We probably don't delete them after either (unless db:migrate command takes care of deprecated 2+ versions ago functions).

This is true that db is not a public api, but we have to keep it because of the running processes.  When we make a new deployment in kubernetes, there would be different versions of services running at the same time. Some background tasks might take hours to complete. So if migration removes those "unused" ones, it will quite likely break those processes in the middle of some work.
We can't change existing functions for the same reason.

db/test/fns/queue_test.js
lotas: Unfortunately we can't. I've written in the comment above. 
See: [db docs](https://github.com/taskcluster/taskcluster/blob/main/db/README.md#stored-functions) and [db best practices](https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/db.md#redefining-db-functions)

docker-compose.dev.yml
lotas: this is a small fix for the local dev env. Without it dev containers didn't pick up new db functions

infrastructure/tooling/src/generate/generators/docker-compose.js
lotas: I don't know. I can extract it if you feel like it should be better to have a separate PR for this.

infrastructure/tooling/src/generate/generators/go-version.js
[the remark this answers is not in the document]
lotas: Was that merged already? I'll rebase and remove it

services/queue/schemas/v1/task-group-response.yml
lotas: I had struggled a bit with default null value for this, trying out different schemas here. Sometimes api validation would fail, sometimes generated GO code would complain about this. 
And I noticed that other schemas using same approach for `expired` timestamp - string type, but not required. 
So I did it same way - strict type, but field is optional. so schema allows `sealed: null` in the response. And clients didn't complain :)

services/queue/schemas/v1/task-group-sealed.yml
lotas: Haven't thought about that! 
On one hand sealed timestamp cannot be changed, so one can query db/api to fetch the date if they are interested.
On the other hand it doesn't hurt to include more and to avoid clients do extra calls
:+1:

ui/src/views/Tasks/TaskGroup/index.jsx
lotas: So there is a popup menu like on the task view page. And it will have few options. For example if task group is already sealed, corresponding option would be disabled, because it doesn't make sense to seal it twice.

services/queue/src/api.js
```diff
@@ -300,6 +300,100 @@ builder.declare({
...
+        taskGroupId,
+      });
+  }
+
+  // fetch project ids to construct scopes: `queue:seal-task-group:<taskGroupId>`
+  let projectIds = await taskGroup.getProjectIds(this.db);
+
+  await req.authorize({
+    taskGroupId,
+    projectIds,
```
petemoore, line 335: How do the projectIds relate to the scopes here? Doesn't the caller just need "queue:list-task-group:<taskGroupId>" ?

petemoore: Have a look at [req.authorize](https://github.com/taskcluster/taskcluster/blob/2e992676afd1035d044349b7f50926e97675e499/libraries/api/src/middleware/auth.js#L42-L89) - I don't think you need to call `req.authorize` at all, but I am not deeply familiar with this middleware.

services/queue/src/api.js
```diff
@@ -300,6 +300,100 @@ builder.declare({
...
+  });
+
+  const updated = TaskGroup.fromDbRows(await this.db.fns.seal_task_group(taskGroupId));
```
petemoore, line 388: There is a race condition here. If more projects are added to the task group between fetching the projectIds and updating the database, then there will be a privilege escalation.

db/fns.md
petemoore: It is a tough one. Maybe we should also have queue.sealTaskGroupForProject as a separate endpoint, that only does it on behalf of a single project that is specified by the caller. The problem with that though is then it doesn't make sense for a task group to be sealed or not, because it then becomes a property of taskgroup/project combination. I'm starting to think the idea of multiple projects per taskgroup might be too compilcated, and perhaps one taskgroup, one project, is going to make life a lot simpler and also still be pretty usable.

db/fns.md
petemoore: I think I meant that _if_ we change it so you can only list projectIds of sealed task groups, we should update this description to say _sealed_ task group. But, I think based on your other feedback, I'm tending to think that having multiple projects per taskgroup is getting a bit too complicated because of the potential race conditions, and if a task group has multiple projects, needing multiple seals also is messy (one per project). So that would break things too because everything expects zero or one seal per task group.

db/fns.md
petemoore: Ah, of course! Thanks Yarik!

services/queue/src/api.js
lotas: 🤦 thanks pete.. I copy-pasted too much.. this is a different endpoint, it shouldn't be here.
👍

infrastructure/tooling/src/generate/generators/docker-compose.js
petemoore: It is fine. I just wasn't sure what it was doing, and wasn't sure if it was related.

services/queue/schemas/v1/task-group-response.yml
petemoore: Ah, it is included in the response, even if it isn't set? Can we change that, so the response doesn't include it, if it is null?

lotas: hm.. it's already optional, not required. Do you see a problem with this? 
I was checking other schemas how similar fields behave, and found similar approach.

We can exclude this field from response if it's null, sure

services/queue/src/api.js
lotas: yes, we'll have to ignore the project part for now. Somehow multiple projects beg for multiple task ids, and we should really force having it so

infrastructure/tooling/src/generate/generators/docker-compose.js
lotas: yeah, it was required to make tings work locally. Let's keep it then, thanks!

db/fns.md
lotas: yes, we have to discuss projectId idea further and see how we can improve it, or maybe get rid of it, since all the tasks existing in both deployments never set it! All tasks have "none" for projectId

clients/client-shell/cmds/group/actors.go
lotas: absolutely! thanks

db/fns.md
```diff
@@ -1794,6 +1799,16 @@ Return true if the task has remaining un-satisfied dependencies.
 
 temp, removed in next commit
```
djmitche, line 1800: huh, this sounds like me.  whoops?

db/test/fns/queue_test.js
```diff
@@ -89,7 +89,7 @@ suite(testing.suiteName(), function() {
       });
     });
 
-    suite('ensure_task_group/get_task_group', function() {
```
djmitche, line 92: It's probably best to actually duplicate this suite.  That way the deprecated `get_task_group` is verified to still be functional, as well as the new function.

services/queue/schemas/v1/task-group-response.yml
```diff
@@ -0,0 +1,31 @@
...
+    title:          "Sealing timestamp"
+    description: |
+      Date and time when task group was sealed.
+  expires:
+    type:           string
+    format:         date-time
+    title:          "Task group expiry date"
+    description: |
+      Date and time when task group expires.
```
djmitche, line 24: It will be worth noting here that this date is after the last expiration of any task in the task group, and for an unsealed taskgroup could change to a later date.

services/queue/schemas/v1/task-group-response.yml
```diff
@@ -0,0 +1,31 @@
+$schema: "/schemas/common/metaschema.json#"
+title:              "Task-Group Response"
+description: |
+  Response containing information about a task group.
+type:               object
+properties:
```
djmitche, line 6: Would it make sense to include schedulerId (and soon projectId) here as well, since they're already easily accessible?

ui/docs/manual/tasks/groups.mdx
```diff
@@ -0,0 +1,26 @@
...
+title: Task Groups
+order: 50
+---
+
+# Task Groups
+
+Task Groups are a way to group tasks together based on their `taskGroupId`.
+When multiple tasks are created simultaneously, they often have the same `taskGroupId`.
+This is useful in situations where tasks are created for a Github
+push event with a [decision task](/docs/manual/design/conventions/decision-task).
```
djmitche, line 12: This has a good bit of overlap with https://docs.taskcluster.net/docs/manual/tasks/taskgroupid-schedulerid.  I like the idea of describing a taskgroup as a "thing" ("entity"? "resource"?) here.  In the existing docs, it's really just an id, and now that it can be sealed and there's a queue.getTaskGroup, that isn't really appropriate anymore.

djmitche (approved):
Nice work!!

db/test/fns/queue_test.js
lotas: sure! thanks

ui/docs/manual/tasks/groups.mdx
lotas: Yeah, I was not sure if adding sealing to that doc would make sense, so created a new page.
I will move information from that one and drop that (or maybe add redirect if it was linked somewhere)

db/fns.md
lotas: thanks for spotting and bringing that back!

services/queue/schemas/v1/task-group-response.yml
lotas: yeah! It is already included in the bottom. We'll add projectId later

services/queue/schemas/v1/task-group-sealed.yml
```diff
@@ -0,0 +1,38 @@
...
+    type:           string
+    format:         date-time
+    title:          "Task group expiry date"
+    description: |
+      Date and time when task group expires.
+required:
+ - taskGroupId
+ - schedulerId
+ - expires
+additionalProperties: false
```
petemoore, line 38: I don't think this whole file is needed, because task-group-response and task-group-sealed should be the same schema. Can you merge them into one, so we just need to maintain this schema in one place?

services/queue/src/api.js
petemoore: For onlookers - with "ignore the project part", Yarik means we have decided to change the required scopes to just `queue:seal-task-group:<taskGroupId>` i.e. no scopes based on `projectId`. It doesn't mean we are ignoring a privilege escalation path! The privilege escalation path has been removed by requiring stricter scopes. :-)

services/queue/schemas/v1/task-group-sealed.yml
lotas: Ah. this is actually a message schema that is being published to the pulse queue

lotas: Rest of the code is doing the same - separating queue schemas and api

matt-boris (approved):
👏🏻 🎉

services/queue/schemas/v1/task-group-sealed.yml
petemoore: There is no reason to duplicate pulse schemas (or any other). See for example:

* https://github.com/taskcluster/taskcluster/blob/main/services/hooks/src/exchanges.js#L58-L107

Here we have three different hooks exchanges all using the same schema: `"pulse-hook-changed-message.json#"`.

If you've seen this duplicated elsewhere, can you create an issue for it? We should avoid the code duplication as much as possible, but it has a tendency to slip in. Thanks Yarik!

petemoore (changes requested):

services/queue/schemas/v1/task-group-sealed.yml
petemoore: Yarik has pointed out to me that one is an http api schema, and the other is a schema for a pulse message, and indeed, we tend not to share these, so let's leave them as separate after all, apologies for the noise.
I think the only thing we might want to change is rename:

* task-group-sealed.yml -> task-group-sealed-message.yml
* task-group-resolved.yml -> task-group-resolved-message.yml

just to be consistent with the naming scheme we are using elsewhere.

We can probably review elsewhere (in a different issue/PR) if we want to start sharing schemas between http api input/outputs and pulse messages, or if it makes sense to keep them separate.

Thanks Yarik!

services/queue/src/exchanges.js
```diff
@@ -353,7 +353,7 @@ exchanges.declare({
     'be submitted against an already resolved task group.',
   ].join('\n'),
   routingKey: buildTaskGroupRoutingKey(),
-  schema: 'task-group-resolved.json#',
+  schema: 'task-group-changed-message.json#',
```
lotas, line 356: @djmitche do you remember if there were any practices for pulse message schema changes? 

I see hard-coded `version: 1` field, so I don't think anyone was ever using different versions of the same message. 

What happens here is that we let several events share one schema, and that schema would include extra fields, which were not defined in the previous one.

I'm marking this as a major release because of this to at least warn whoever was using clients for the previous version

djmitche: Much like `/v1/` in the API paths, I think `version: 1` was added without much thought to how it might be used in the future.

In general, though, adding to APIs isn't an incompatible change, so no special compatibility work is required.  We've added fields to the github service's messages a few times, iirc.

Marking it as a major change probably still makes sense, though, for exactly the reason you stated.

lotas: Right, thanks! 

I checked rabbitMQ and the only client listening to that exchange is actually a github service. 
And it seems like payload validation is on the publisher side, not the consumer.

petemoore (dismissed):

PR #6062, opened 2023-02-15
Fix: collapse rendering of rust client generated code in PRs

petemoore:
I spotted in #6060 that generated rust code was being rendered - this is a fix for that.


lotas (approved):
Nice! Thanks

djmitche (approved):

PR #6066, opened 2023-02-15
Move taskcluster CI decision task to generic-worker

petemoore:
Seeing if I can switch our own docker-worker CI decision task to generic-worker, using d2g tool...

Let's see how many attempts this takes, at the same time I've tried to reformat the output a little bit to make it easier to read, and due to all the encodings inside encodings, this may take a few attempts to get right...

PR #6068, opened 2023-02-15
bump taskgraph decision image

matt-boris:
Suggested keeping this change separate from Pete's work https://github.com/taskcluster/taskcluster/pull/6066#discussion_r1107477154


petemoore (approved):
Thanks @matt-boris !

PR #6069, opened 2023-02-16
Bugzilla 1753249 task id clickable field

jakeNesom:
Bugzilla Bug: [1753249](https://bugzilla.mozilla.org/show_bug.cgi?id=1753249)

- [x] Makes TaskID a click-to-copy field on the Task Details Page
- [x]  Regenerates Jest snapshot
- [x] Updates the .gitignore for the UI project per [Yarn](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored) package manager's recommendation

Screen Shot 2023-02-15 at 10 49 05 PM

### DIFF
--- a/ui/.gitignore
+++ b/ui/.gitignore
@@ -13,3 +13,13 @@ src/static/env.js
 coverage
 test/e2e/cypress/screenshots/
 test/e2e/cypress/videos/
+
+
+# yarn package manager
+.yarn/*
+!.yarn/cache
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
+!.yarn/versions
--- a/ui/.gitignore
+++ b/ui/.gitignore
@@ -13,13 +13,3 @@ src/static/env.js
 coverage
 test/e2e/cypress/screenshots/
 test/e2e/cypress/videos/
-
-
-# yarn package manager
-.yarn/*
-!.yarn/cache
-!.yarn/patches
-!.yarn/plugins
-!.yarn/releases
-!.yarn/sdks
-!.yarn/versions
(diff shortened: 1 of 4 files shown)


lotas:
Thank you for submitting this patch!

ui/.gitignore
```diff
@@ -13,3 +13,13 @@ src/static/env.js
 coverage
 test/e2e/cypress/screenshots/
 test/e2e/cypress/videos/
+
+
+# yarn package manager
+.yarn/*
```
lotas, line 19: I'm wondering if we really need this change. Can you please tell how you get there? 
We've been using both `ui` and top-level services with `yarn` but we never had any files created in `.yarn` folder other than [.yarn/releases/yarn-1.22.19.cjs](https://github.com/taskcluster/taskcluster/blob/main/.yarn/releases/yarn-1.22.19.cjs) that we have on the top level.

Maybe you have some custom yarn configuration in your local dev environment? 
Thanks!

lotas (approved):
Looks good! Thanks!

lotas:
`meta-generate` test fails for known reasons, I checked it locally, it passes with the patch that was not released yet

ui/.gitignore
jakeNesom: Yeah you are right - I don't think its necessary. 

For some reason running 'yarn' was installing the .yarn/cache directory in the /ui project root but i think i fixed it.   My first time using yarn & nvm.

lotas: Cool! Let's then revert those changes for this file.
Thank you

jakeNesom: Cool - i made another commit to revert the changes to the .gitignore  Hope that's ok

lotas (approved):

ui/.gitignore
lotas: Awesome! Thanks again!

PR #6070, opened 2023-02-16
Replace `github.com/ghodss/yaml` with `sigs.k8s.io/yaml`

Juneezee:
The package `github.com/ghodss/yaml` is no longer actively maintained. See discussion in https://github.com/ghodss/yaml/issues/75 and https://github.com/ghodss/yaml/issues/80. [`sigs.k8s.io/yaml`](https://github.com/kubernetes-sigs/yaml) is a permanent fork of `github.com/ghodss/yaml`.

The notable change is that `github.com/ghodss/yaml` uses `gopkg.in/yaml.v2 v2.2.2`, but `sigs.k8s.io/yaml` uses `gopkg.in/yaml.v2 v2.4.0`. Changes can be seen here [v2.2.2...v2.4.0](https://github.com/go-yaml/yaml/compare/v2.2.2...v2.4.0), mostly bug fixes.

### DIFF
--- a/changelog/YHDKOgBXSZ-t6gQX9OvHbg.md
+++ b/changelog/YHDKOgBXSZ-t6gQX9OvHbg.md
@@ -1,0 +1,5 @@
+audience: developers
+level: patch
+---
+
+Replace `github.com/ghodss/yaml` with `sigs.k8s.io/yaml`
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,6 @@ require (
 	github.com/elastic/go-sysinfo v1.9.0
 	github.com/fatih/camelcase v1.0.0
 	github.com/getsentry/raven-go v0.2.0
-	github.com/ghodss/yaml v1.0.0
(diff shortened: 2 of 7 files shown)


petemoore:
Thanks for this @Juneezee ! Indeed, it looks like the original repo is no longer actively maintained. I see there are a few different forks, but this seems like a reasonable choice.

petemoore (approved):
❤️

changelog/YHDKOgBXSZ-t6gQX9OvHbg.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
+level: silent
```
matt-boris, line 2:
```suggestion
level: patch
```

matt-boris (changes requested):
Would you please change this to a patch level? I'll approve after this! Thanks for replacing!

matt-boris (approved):

PR #6071, opened 2023-02-16
Simplify decision task: set env vars only in container, not in host process

petemoore:
It occurred to me it that for environment variables which are only used by programs in the (podman) container and not by commands on the host, that it is an unnecessarily complication and overhead to set these env vars in the host process too. In other words, this:

```yaml
env:
  ENV_VAR: "<VALUE>"
  ....
command:
  ....
  podman -e "ENV_VAR=$${ENV_VAR}"
  ....
```

simplifies to:

```yaml
command:
  ....
  podman -e "ENV_VAR=<VALUE>"
  ....
```

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -153,19 +153,9 @@ tasks:
                   env:
                       # run-task uses these to check out the source; the inputs
                       # to `mach taskgraph decision` are all on the command line.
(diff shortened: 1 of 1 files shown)


matt-boris (approved):
Nice, thanks for this cleanup! 🎉

PR #6074, opened 2023-02-17
build(deps): upgrade node to 18.14.1

matt-boris:
>This upgrades Node.js to the latest LTS version available, v18.14.1. This is a security release.

More info can be found [here](https://nodejs.org/en/blog/vulnerability/february-2023-security-releases/).


petemoore (approved):
Thanks Boris!

PR #6076, opened 2023-02-19
Document db fn is_task_group_active

djmitche:
It looks like this got here through a revert in #3243.


petemoore (approved):
Thanks Dustin! ❤️

PR #6077, opened 2023-02-20
Grant Access to WinStation and Desktop for Application Container SIDs on Windows workers.

### LINKED ISSUE #6058: Grant Access to WinStation and Desktop for Application Container SIDs on Windows workers.
[Bug 1815711 - Grant Access to WinStation and Desktop for Application Container SIDs on Windows workers.](https://bugzilla.mozilla.org/show_bug.cgi?id=1815711)

At the moment we grant access to Everyone (S-1-1-0) on Window Station WinSta0 and Desktop Default for the Windows workers.
I think that this is done [here](https://github.com/taskcluster/taskcluster/blob/a431591bb38a3739bebc787d535a89f066f370bd/workers/generic-worker/process/multiuser_windows.go#L36-L39).

In order to test Low Privileged Application Containers with taskcluster we need to also add access to ALL RESTRICTED APPLICATION PACKAGES (S-1-15-2-2).
It probably makes sense to add ALL APPLICATION PACKAGES (S-1-15-2-1) as well for normal Application Containers.
The equivalent Window Station and Desktop for a normal logon session already have these permissions.

bobowen:
Bug 1815711 - Grant Access to WinStation and Desktop for Application Container SIDs on Windows workers.

Github Bug/Issue: Fixes #6058

In order to test Low Privileged Application Containers with taskcluster this adds access to ALL RESTRICTED APPLICATION PACKAGES (S-1-15-2-2). It also add ALL APPLICATION PACKAGES (S-1-15-2-1) for normal Application Containers.

The equivalent Window Station and Desktop for a normal logon session already have these permissions.


petemoore:
Many thanks for this @bobowen! Shout if you need any help while you are working on it. Looking pretty good so far. :-)

bobowen:
> Many thanks for this @bobowen! Shout if you need any help while you are working on it. Looking pretty good so far. :-)

Hi @petemoore, I've struggled trying to get a dev environment set up for this.
Matt Boris has already been helping me with that, but I hit issues both with WSL and a separate Ubuntu VM.

Realistically I don't know that I'll be making lots of contributions to taskcluster, so would it be possible for someone else to take on this patch.
What I'm really hoping exists is a "test try" environment where the change can be included and I can push my Firefox LPAC changes to make sure that this change does in fact fix the problem.

Thanks for your help.

petemoore:
@bobowen No worries, I think I can take it from here! We are on a work week this week, so it might be next week until I get a chance to work on it - is that also ok with you? Then we can push out a new release for you to play around with, to check it does what you want.

petemoore:
> Matt Boris has already been helping me with that

That's great to hear. You are in very good hands! :-)

bobowen:
> @bobowen No worries, I think I can take it from here! We are on a work week this week, so it might be next week until I get a chance to work on it - is that also ok with you? Then we can push out a new release for you to play around with, to check it does what you want.

Yes next week is absolutely fine, thanks for taking it on.
I'm still in the process of landing the change behind a pref anyway.
I won't need these changes until I'm going to turn it on by default.
It would be good to prove the fix though, it took a lot of remote printf debugging on try to work out how to reproduce locally and then kernel debugging to finally get to (what I very much hope is) the root cause.

petemoore:
@bobowen I'm really sorry this got pushed back a bit with other activities going on - I will be looking at this on Monday again. Sorry for the delay!

petemoore:
```
2023/04/18 12:15:26Z Making system call SetUserObjectSecurity with args: [1A8 C0002B3460 C000235300]
2023/04/18 12:15:26Z   Result: 0 0 Access is denied.
2023/04/18 12:15:26Z Error granting S-1-15-2-1 full control of interactive windows station and desktop:
2023/04/18 12:15:26Z SetUserObjectSecurity: Access is denied.
```

bobowen:
Looks like that is what it is doing (if I'm interpreting the code correctly):
https://github.com/taskcluster/taskcluster/blob/66941613e4242d69dd2aff3fb560359eb633d59c/workers/generic-worker/win32/win32acl_windows.go#L87-L109

It's creating a new ACL adding the ACE(s) and then setting that on the object.
I think it should be getting the existing ACL and then using something like SetEntriesInAcl to create a new ACL with the new access granted.

petemoore:
Separately I've created #6188 to remove the sharing of the current console session Windows Station and Desktop with Everyone (`S-1-1-0`) in the case that the access token is not refreshed.

The only time access to the active Windows Station and Desktop needs to be granted is when the currently available access token for the interactive desktop session does not already have access, which can happen if a permission change is required that itself requires a new logon. In this case, we call [advapi32 LogonUserW](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-logonuserw) to generate a new access token to incorporate additional permissions which were not applied when the previous logon took place. Currently this is only when additional `osGroups` are specified in the task payload, that the user should be added to.

petemoore:
Closing here as this was fixed by #6188

PR #6103, opened 2023-03-06
Worker manager considers stoppingCapacity in estimator

### LINKED ISSUE #6067: Worker manager multiple errors for the same worker pool
there was an incident on 2023-02-15 when few worker pool configs included `tags` that were injected in the azure api calls and were failing schema validation:

`parameters.tags.deploymentId with value "12345678" must be of type string.`

Problem was that client library in ci-admin was converting that string into a number because it contained only digits.

Ideas:
1. ~worker manager can add extra schema validation for tags, to ensure that only strings are passed~
2. when something like this happens, worker-manager should be smart and not to try to  spin new workers with bad payload, which resulted in almost 50k error messages in just few hours. 
Probably some kind of "quarantine" or backoff timeouts for worker pool is necessary to prevent starting new instances in big amounts ..
3. for the scanner loop limit number of workers per pool per scan, to avoid single pool "stealing" whole loop

lotas:
Stopping workers require multiple calls for Azure to deprovision. When Workers that fail to start immediately due to the configuration or cloud availability error, go into stopping mode. Next provisioner loop doesn't count those against total capacity and keeps spawning them. This leads to the situations where stopping capacity grows significantly over short period of time and blocks scanning loop.

Fixes #6067

### DIFF
--- a/changelog/issue-6067.md
+++ b/changelog/issue-6067.md
@@ -1,0 +1,6 @@
+audience: deployers
+level: patch
+reference: issue 6067
+---
+
+Worker-manager now considers `stoppingCapacity` when estimating the required number of workers to start, preventing failed to start workers from growing beyond `maxCapacity` and slowing down the scanner loop.
--- a/generated/references.json
+++ b/generated/references.json
@@ -13321,13 +13321,14 @@
             "pendingTasks": "The number of tasks the queue reports are pending for this worker pool",
             "requestedCapacity": "Amount of capacity that this estimator thinks we should add",
             "scalingRatio": "The ratio of workers to spawn per pending task for this worker pool.",
+            "stoppingCapacity": "Amount of capacity being stopped",
             "workerPoolId": "The worker pool name"
           },
           "level": "any",
           "name": "simpleEstimate",
           "title": "Simple Estimate Provided",
           "type": "simple-estimate",
-          "version": 3
+          "version": 4
         },
         {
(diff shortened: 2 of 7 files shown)


matt-boris (approved):
Nice, looks great! 🎉

petemoore (approved):
It looks good to me! You might want to run it past @djmitche too, or someone else that has touched Worker Manager code. I appreciate that you've written some tests though. Do we have a way to roll this out to an environment and put some load on it which intentionally creates a bunch of stopping instances (e.g. intentionally use machine images)? It would be nice if we had a way to exercise the new logic before it goes to fxci, but I don't know how much work that would be.

lotas:
I am trying to adjust [taskcluster-provisioning-simulator](https://github.com/taskcluster/taskcluster-provisioning-simulator/) to add this `stopping` state into simulations to see if it makes any difference

lotas:
Made few quick updates: https://github.com/taskcluster/taskcluster-provisioning-simulator/commit/897259074beb14fbcb7fb1a55575f12c9eb98cc7 

Ran with firefox simulator, got this. Pretty much the same as the simulator without the stopping capacity, but I'm semi-confident this is a valid testing scenario for this

image

djmitche:
> You might want to run it past @djmitche too

Thanks for thinking of me! It's good to see that the provisioning simulator was useful here!

Issue #5993, opened 2023-01-17
Support custom log artifact paths in generic-worker


petemoore:

lotas:
@petemoore can you provide any description why this might be a good idea? :) 
iiuc `/public/logs/live.log` can be replaced with any other path, right? This might break the existing github service and the ui, as they rely on that path to be present. Would have to be updated as well.

PR #6107, opened 2023-03-13
feat(worker): add `task.payload.logs` and `task.payload.features.{liveLog,backingLog}` fields to generic worker

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/5993

Work to follow will happen in [taskcluster/d2g](https://github.com/taskcluster/d2g) which will address https://github.com/taskcluster/taskcluster/issues/5978.

>Adds the `liveLog` and `backingLog` feature flags to the generic worker payload so they can be disabled for a task. These are enabled by default.
Adds the `logs` property to the generic worker payload allowing customization of the live and backing log artifact names.

Note: PR is easier to review commit by commit, as they're chunked up into which part of the monorepo I edited: generic worker, github service, and ui.

### DIFF
--- a/services/github/src/handlers/utils.js
+++ b/services/github/src/handlers/utils.js
@@ -1,3 +1,4 @@
+const path = require('path');
 const libUrls = require('taskcluster-lib-urls');
 const { CHECK_RUN_STATES } = require('../constants');
 
@@ -5,9 +6,8 @@ const taskUI = (rootUrl, taskGroupId, taskId) =>
   libUrls.ui(rootUrl, rootUrl === 'https://taskcluster.net' ? `/groups/${taskGroupId}/tasks/${taskId}/details` : `/tasks/${taskId}`);
 const taskGroupUI = (rootUrl, taskGroupId) =>
   libUrls.ui(rootUrl, `${rootUrl === 'https://taskcluster.net' ? '' : '/tasks'}/groups/${taskGroupId}`);
-const taskLogUI = (rootUrl, runId, taskId) =>
-  libUrls.ui(rootUrl, `/tasks/${taskId}/runs/${runId}/logs/live/public/logs/live.log`);
-
+const taskLogUI = (rootUrl, runId, taskId, liveLogName = 'public/logs/live.log') =>
+  libUrls.ui(rootUrl, path.join(`/tasks/${taskId}/runs/${runId}/logs/live/`, liveLogName));
 let debugCounter = 0;
 
 /**
--- a/services/github/test/handler_test.js
+++ b/services/github/test/handler_test.js
@@ -21,6 +21,7 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
 
   const URL_PREFIX = 'https://tc-tests.example.com/tasks/groups/';
   const CUSTOM_CHECKRUN_TASKID = 'apple';
+  const CUSTOM_LIVELOG_NAME_TASKID = 'banana';
   const CUSTOM_CHECKRUN_TEXT = 'Hi there! This is your custom text';
   const LIVE_LOG_TEXT = 'Hi there! This is your live log';
   const CUSTOM_CHECKRUN_ANNOTATIONS = JSON.stringify([
@@ -114,6 +115,19 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
                   customCheckRun: {
                     textArtifactName: 'public/text.md',
                   },
+                },
+              },
+            });
+
+          case CUSTOM_LIVELOG_NAME_TASKID:
+            return Promise.resolve({
+              metadata: {
+                name: 'Task with custom live log path',
+                description: 'Task Description',
+              },
+              payload: {
+                logs: {
+                  live: 'apple/banana.log',
                 },
               },
             });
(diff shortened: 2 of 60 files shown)


lotas:
Does this create an extra log artifact, or replaces the existing one? 

I'm not sure why this change is needed, tickets don't have any description :) but it might break few things:
`/public/logs/live.log` path is already hardcoded in github service (to fetch last lines), documentation and in the UI. 
To keep those working correctly we should probably update those as well.

matt-boris:
@lotas thanks for the comments!

> Does this create an extra log artifact, or replaces the existing one?

This replaces the default livelog location of `/public/logs/live.log` with the user-defined path.

> I'm not sure why this change is needed, tickets don't have any description

Sorry about the ticket being pretty bare! Basically, we need to bring generic worker up to parity with what docker worker has, in terms of features/capabilities. This is already a feature of docker worker, being able to define a custom livelog path.

> but it might break few things: `/public/logs/live.log` path is already hardcoded in github service (to fetch last lines), documentation and in the UI.

Yes, good call, thanks! I'll update those paths in this PR as well. Tailing the log was updated to fetch the `/public/logs/live_backing.log` file (which is _always_ written out), so I'll update the few instances where we're linking to `/public/logs/live.log` to `/public/logs/live_backing.log`.

petemoore:
I think there may still be some references to live_backing.log:

```
workers/generic-worker/main.go:	logName   = "public/logs/live_backing.log"
workers/generic-worker/main.go:	logPath   = filepath.Join("generic-worker", "live_backing.log")
workers/generic-worker/main.go:			// public/logs/live_backing.log inadvertently.
```

workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -269,6 +269,10 @@ properties:
           title: Exit codes
           type: integer
           minimum: 1
+  log:
+    title: Livelog artifact name
```
petemoore, line 273: I've just realised, the description is probably needed after all, to explain that this setting affects both the name of the live log, _and_ the name of the backing log. It should explain how the backing log name is derived from the live log artifact name.

matt-boris:
Ready for review now @lotas!

@petemoore and I worked a bunch together on this one, so we'd appreciate a different set of 🦅 👀! 🙏🏻

services/github/src/handlers/utils.js
```diff
@@ -1,13 +1,13 @@
...
 const { CHECK_RUN_STATES } = require('../constants');
 
 const taskUI = (rootUrl, taskGroupId, taskId) =>
   libUrls.ui(rootUrl, rootUrl === 'https://taskcluster.net' ? `/groups/${taskGroupId}/tasks/${taskId}/details` : `/tasks/${taskId}`);
 const taskGroupUI = (rootUrl, taskGroupId) =>
   libUrls.ui(rootUrl, `${rootUrl === 'https://taskcluster.net' ? '' : '/tasks'}/groups/${taskGroupId}`);
-const taskLogUI = (rootUrl, runId, taskId) =>
-  libUrls.ui(rootUrl, `/tasks/${taskId}/runs/${runId}/logs/live/public/logs/live.log`);
-
+const taskLogUI = (rootUrl, runId, taskId, liveLogName = 'public/logs/live.log') =>
```
lotas, line 9: Would default value produce correct path?

I'm checking one of the livelogs in a task:
https://community-tc.services.mozilla.com/tasks/M6s3bmwiThKwqvf55o6Blg/runs/0/logs/public/logs/live.log

so the path here is `/tasks/M6s3bmwiThKwqvf55o6Blg/runs/0/logs/public/logs/live.log`
in the next line it would be joined and would produce something like:
`/tasks/M6s3bmwiThKwqvf55o6Blg/runs/0/logs/live/public/logs/live.log`
which will have extra `/live/` inside, no?

services/github/test/handler_test.js
```diff
@@ -118,6 +119,19 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
...
 
+          case CUSTOM_LIVELOG_NAME_TASKID:
+            return Promise.resolve({
+              metadata: {
+                name: 'Task with custom live log path',
+                description: 'Task Description',
+              },
+              payload: {
+                logs: {
+                  live: 'apple/banana.log',
```
lotas, line 130: 🍎  🍌

services/github/src/handlers/utils.js
matt-boris: The extra `/live/` within the path tells it to stream the logs. You can try putting the `/live/` within the path of the link you commented and see that it'll still load it up. What's different is when there's a running task. You need the `/live/` part to follow the logs.

https://github.com/taskcluster/taskcluster/blob/14ec3e80bfa85518ef6500a1727f0b678ebe5c80/ui/src/views/Tasks/routes.js#L64-L71

test/go-lint.sh
```diff
@@ -8,7 +8,7 @@ if [ -n "${unformatted_files}" ]; then
     exit 1;
 fi
 
-echo "Running golangci-lint.."
-golangci-lint run --build-tags multiuser
-golangci-lint run --build-tags simple
-golangci-lint run --build-tags docker
+for engine in multiuser simple docker; do
```
lotas, line 11: carbon-neutral refactoring 😂  4 lines in, 4 lines out 👍 
nice

ui/docs/manual/using/artifacts.mdx
```diff
@@ -9,8 +9,10 @@ import Warning from 'taskcluster-ui/views/Documentation/components/Warning';
 
 ## Creating Artifacts
 
-Most workers automatically produce a conventional `public/logs/live.log`
-artifact containing the main log of the artifact.
+Workers conventionally create a `public/logs/live.log`
+artifact containing the main task log. This default path can be overridden
+in Generic Worker task payloads using the `logs.live` property and the
+`log` property for Docker Worker task payloads.
```
lotas, line 15: stupid question, if you define `logs.live` would it replace whole path, or just the last part of it? 
So if `logs.live: 'non-public/banana.log'` then artifact path would be `non-public/banana.log`?

matt-boris: Your example is correct :)

workers/generic-worker/chain_of_trust.go
```diff
@@ -63,6 +62,7 @@ type ChainOfTrustData struct {
 type ChainOfTrustTaskFeature struct {
 	task           *TaskRun
 	ed25519PrivKey ed25519.PrivateKey
+	disabled       bool
```
lotas, line 65: is it something that was present in docker worker? 
I can't find any other comments explaining this addition

matt-boris: We discovered a small bug where if a task feature (chain of trust in this case) fails to `Start()` but still publishes sig/cert files, so we still need to explicitly get the feature to `Stop()`, which is how this new member is being used.

lotas (dismissed):
Tremendous!!! 
Thanks!

workers/generic-worker/schemas/multiuser_windows.yml
```diff
@@ -231,6 +231,24 @@ properties:
           `generic-worker:run-as-administrator:<provisionerId>/<workerType>`.
 
           Since: generic-worker 10.11.0
+      liveLog:
```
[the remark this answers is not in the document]
petemoore: True. It would be nice to clean this up. There may be some option with using `$ref` json schema references, you are right.

services/github/src/handlers/utils.js
lotas: ah! then all good, thanks!

workers/generic-worker/chain_of_trust.go
lotas: all good, I was just missing the context for this change. Maybe adding a comment would be enough.

lotas (approved):

PR #6110, opened 2023-03-15
Remove colons from paths in worker-manager API methods

### LINKED ISSUE #6109: Error when calling Get a Worker API Endpoint
**Describe the bug**
`listWorkersForWorkerPool` method one is too greedy and gets also the requests coming for worker call

**To Reproduce**

Open powershell, populate the clientId and accessToken, and run the below.

```Powershell
Invoke-RestMethod -Uri "https://firefox-ci-tc.services.mozilla.com/api/worker-manager/v1/workers/gecko-t/win11-64-2009/canadacentral/vm-bfarycpest6jyozcgdfjgazzkhve9t86bzj" -Headers @{
    clientId    = $ENV:TASKCLUSTER_CLIENT_ID
    accessToken = $ENV:TASKCLUSTER_ACCESS_TOKEN
}
```

**Expected behavior**

**Taskcluster version**

**Screenshots**

 - OS: Windows

**Additional context**
Existing bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1560645

djmitche:
Github Bug/Issue: Fixes #6109

I tested manually that the old paths still work. Doing so in unit tests is hard because the withColon methods require scopes, but the client libraries don't create a function to call them.


lotas:
Awesome! Thank you Dustin for the quick fix. 🎉 
I tested it locally, and it works. The only thing worth keeping in mind that `/` needs to be escaped in order to work properly

services/worker-manager/src/api.js
```diff
@@ -38,6 +38,23 @@ let builder = new APIBuilder({
...
+// un-published method at that path, for backward compatibility. See #6109.
+const declareWithTrailingColon = (options, handler) => {
+  // declare with the colon included
+  const colonOptions = {
+    ...options,
+    route: options.route.replace("/workers/:workerPoolId/", "/workers/:workerPoolId:/"),
+    name: options.name + 'WithColon',
+    noPublish: true,
+  };
+  builder.declare(colonOptions, handler);
```
lotas, line 52: nice!

services/worker-manager/src/api.js
```diff
@@ -420,9 +437,9 @@ builder.declare({
...
 });
 
-builder.declare({
+declareWithTrailingColon({
   method: 'get',
-  route: '/workers/:workerPoolId:/:workerGroup/:workerId',
+  route: '/workers/:workerPoolId/:workerGroup/:workerId',
   name: 'worker',
   scopes: 'worker-manager:get-worker:<workerPoolId>/<workerGroup>/<workerId>',
   title: 'Get a Worker',
```
lotas, line 445: I'm not sure if it is documented anywhere else, but maybe we should add also in description that `workerPoolId` should be escaped/urlencoded before by the clients? Or is it too obvious? :)

lotas (approved):

services/worker-manager/src/api.js
petemoore: I think we've documented that all parameters need to be urlencoded, and the various clients take care of this.

lotas: great then!

petemoore: Well, I say that, but now I can't find it here: https://community-tc.services.mozilla.com/docs/manual/using/api

petemoore (approved):
Thanks for doing this Dustin!

services/worker-manager/src/api.js
djmitche: That's definitely where it should be documented.  You're right that in the bug @jwmoss identified (https://bugzilla.mozilla.org/show_bug.cgi?id=1560645) we added the flexibility to allow an unencoded _final_ component, but that's not allowed on non-final components.

djmitche:
changelog added

petemoore (approved):
Thanks Dustin!

PR #6113, opened 2023-03-20
jsonschema2go defaults

petemoore:
In #6111 we worked around the issue that jsonschema default values are not supported in jsonschema2go by changing `bool` to values to `*bool`. The underlying type change enabled the type system to represent absence of a value from a json object as a nil pointer. However, this didn't overcome the underlying issue that the default values specified in a schema would be automatically respected when unmarshaling the data into the go type, and put the burder on the client to duplicate this logic.

This PR applies a different approach to solve the same problem. Instead of switching the underlying type from `bool` to `*bool`, it encodes default values as struct tags directly into the generated go types, and the caller may (should) use [https://github.com/mcuadros/go-defaults](go-defaults) to pre-initialise the default go values, before unmarshaling the json data into them.

Callers are already required to validate json using e.g. gojsonschema before unmarshaling, and therefore at some point it may make sense for us to additionally generate helper functions in jsonschema2go to do everything (validate json payload, apply schema defaults, and unmarshal data).


petemoore:
Note, the first commit is a simple revert of #6111 and the second commit is the fix to jsonschema2go plus the generated code changes. So probably easiest to review the commits one by one...

lotas (dismissed):

matt-boris (commented):
Nice, this looks so much better than what we had! Thanks for cleaning this up Pete and for keeping the PR in separate commits - much easier to review 🙂

Left a tiny suggestion to remove a newline.

tools/jsonschema2go/jsonschema.go
```diff
@@ -1078,8 +1078,19 @@ func (s *Properties) AsStruct(disableNested bool, extraPackages StringSet, rawMe
...
 			}
+			defaultStructTag := ""
+			if def := s.Properties[j].Default; def != nil {
+				switch (*def).(type) {
+				// for now, let's keep this simple, and limit support to types
+				// that currently have a default; we can expand on this if and
+				// when more types are needed
+				case string, bool:
+					defaultStructTag = fmt.Sprintf(` default:"%v"`, *def)
+
```
matt-boris, line 1089:
```suggestion
```

PR #6114, opened 2023-03-20
Dashboard displays stopping capacity, values autoreloaded.

lotas:
Few updates to the dashboard:

* `stoppingCapacity` is being displayed, since worker manager uses it during provisioning
* Values are being grouped by provisioning and general stats
* worker manager stats are being refreshed every 30s
* visual animations when values change

image


matt-boris (approved):
Nice! 🎉

PR #6123, opened 2023-03-22
Combine artifacts.S3Artifact and artifacts.ObjectArtifact struct members

petemoore:
Types `artifacts.S3Artifact` and `artifacts.ObjectArtifact` both contain the following two struct members:

* `Path` - the file path, relative to the task directory, of the file whose content holds the artifact content
* `RawContentFile` - the absolute file path of the file whose content holds the artifact content

We don't need both, and since some artifacts (e.g. live log, backing log, rdp info file, etc) should not even exist inside the task directory (since the worker creates them, not the task) we should free ourselves from the constraint that there needs to be a relative path at all. In fact, the relative path is only used for logging that artifacts are being uploaded, and we already include both the relative path and the absolute path in those messages. Therefore, we can drop the relative path. However, the struct member name `Path` is preferable to `RawContentFile`, so this PR essentially removes the `Path` members from go types `artifacts.S3Artifact` and `artifacts.ObjectArtifact`, and then renames `RawContentFile` to `Path`. Logging is updated to only include the absolute path.


lotas (approved):
:+1:

`+56 - 69 = -13` 👏

matt-boris (approved):
Nice, thanks Pete!

lotas (approved):

PR #6129, opened 2023-03-27
Statuses API set back status to pending when task was restarted

lotas:
Fixes Bugzilla Bug: [1590886](https://bugzilla.mozilla.org/show_bug.cgi?id=1590886)

### DIFF
--- a/services/github/src/handlers/deprecatedStatus.js
+++ b/services/github/src/handlers/deprecatedStatus.js
@@ -49,6 +49,10 @@ async function deprecatedStatusHandler(message) {
 
   if (message.exchange.endsWith('task-exception') || message.exchange.endsWith('task-failed')) {
     state = 'failure';
+  } else if (message.exchange.endsWith('task-running') && build.state !== 'pending') {
+    // if build is not pending, it means it was already resolved as success or failure
+    // seeing a running task means it was retried, so we should set the status back to pending
+    state = 'pending';
   }
 
   await this.context.db.fns.set_github_build_state(taskGroupId, state);
--- a/services/github/src/handlers/deprecatedStatus.js
+++ b/services/github/src/handlers/deprecatedStatus.js
@@ -24,9 +24,11 @@ async function deprecatedStatusHandler(message) {
     installationId: build.installation_id,
   });
 
+  const { exchangeNames } = this;
+
   let state = 'success';
 
-  if (message.exchange.endsWith('task-group-resolved')) {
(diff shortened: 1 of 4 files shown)


services/github/src/handlers/deprecatedStatus.js
```diff
@@ -49,6 +49,10 @@ async function deprecatedStatusHandler(message) {
 
   if (message.exchange.endsWith('task-exception') || message.exchange.endsWith('task-failed')) {
     state = 'failure';
+  } else if (message.exchange.endsWith('task-running') && build.state !== 'pending') {
```
petemoore, line 52: It looks like this should work, but it feels a little fragile to have these if blocks handling all the different exchanges. Perhaps it would be cleaner to separate out the handlers for the various exchanges? If we do keep it this way, perhaps a single if/else if block to encompass them all would be better? For example, at the moment, line 29 and line 50 aren't part of the same if/else if block, but 50 and 52 are, so I think it would be best either to have multiple independent blocks (29, 50, 52), or one if/else if block (or is there something like a switch statement in node.js?) containing all options, including a default for anything missed. The reason to have two in a combined block, but one not, isn't clear. It therefore isn't immediately clear if all cases are caught, or if there are changes to which exchanges are published to, how it would affect the logic. Also matching by the textual ending of the exchange name also feels fragile, it just feels a bit back-to-front. In other words all the handlers are merged into one, and then we are working backwards to attribute each logic block to its original source, rather than having individual handlers from the start. Does that make sense? In any case, it looks like it will work, it just doesn't feel highly maintainable. Also checking the ending of the exchange name doesn't feel safe if another exchange is created with a different name but has the same ending, or if the exchange name is changed, etc.

petemoore (approved):

services/github/src/handlers/deprecatedStatus.js
lotas: Thanks Pete!
Yeah, I agree. I guess those handlers were build around github functionality and rabbitMQ queues. 
This handler is responsible for updating single build status. If we split those, we would have another problem. Three handlers that are doing the same but only differ in `pending|success|failure` state that they are passing to the github api.

Let me try to refactor it a bit inside this handler, and if you think we still need to change it, we can create a ticket for this.

lotas: @petemoore combined if-else in single block and also made names a bit more deterministic by using real exchange names for comparisons

petemoore (approved):
Thanks Yarik! ❤️

Issue #6130, opened 2023-03-27
Generic worker panic when `workerPoolID` doesn't have a slash


matt-boris:
**Describe the bug**
This error occurs when the worker runner config file specifies a `workerPoolID` without a slash in it.
```
panic: runtime error: index out of range [1] with length 1
```
It occurs on this line https://github.com/taskcluster/taskcluster/blob/8a04c2354df5d9348b4744ba2da880546d64cc98/tools/worker-runner/worker/genericworker/genericworker.go#L82

We should just do a sanity check in the code that the array `splitWorkerPoolID` has two elements, otherwise, report an error about the problem.

PR #6131, opened 2023-03-27
fix(generic-worker): ensure workerPoolID contains slash

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6130


lotas (approved):
:+1:

PR #6132, opened 2023-03-27
feat(github): Skip updating statuses when task is running

lotas:
Patching https://github.com/taskcluster/taskcluster/pull/6129 that was released in 48.2.0

Since new event was added to that handler, it was invoked too often and resulted in unnecessary status api updates.
When task starts running  it is only worth updating status if whole build was already fail/success. 
Initial `pending` state is handled by a different handler, so safe to skip it here


matt-boris (approved):
👍🏻 ty!

PR #6133, opened 2023-03-28
docker-worker: set a timeout when downloading artifact

jcristau:
If the connection to download the docker image hangs for whatever reason, we could wait indefinitely, and because the maxRunTime timer isn't set up by that point, we'd only stop when hitting the task's deadline and failing to reclaim it.
Instead, time out the connection if it stays idle for 60 seconds, so we can properly retry the download.

Bugzilla Bug: [1824937](https://bugzilla.mozilla.org/show_bug.cgi?id=1824937)

### DIFF
--- a/changelog/bug-1824937-0.md
+++ b/changelog/bug-1824937-0.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: bug 1824937
+---
+docker-worker no longer waits indefinitely when downloading an image artifact, and will retry if the connection is idle for 60s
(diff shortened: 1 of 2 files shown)


lotas:
thanks @jcristau for adding this. 
Are you planning to add anything? This PR is marked as draft still and doesn't include a [changelog](https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/changelog.md) file yet.

Regarding 60s timeout, as I understand from the [docs](https://github.com/request/request) it is  only for connection and waiting for response, so doesn't include the actual downloading time. Which is more than enough I believe and is a good default value 👍 
Thanks!

jcristau:
I'll add a changelog and move this out of draft.

lotas (approved):
:+1: Looks good, thanks!

petemoore (approved):
Looks good to me, thanks @jcristau!

PR #6134, opened 2023-03-28
docker-worker: add a timer for task setup, defaulting to 10 minutes

jcristau:
Limit the amount of time we can spend setting up the docker container for the task.
    
Bugzilla Bug: [1824937](https://bugzilla.mozilla.org/show_bug.cgi?id=1824937)

### DIFF
--- a/changelog/bug-1824937.md
+++ b/changelog/bug-1824937.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: bug 1824937
+---
+Limit the amount of time docker-worker can spend setting up the container for the task, to 10 minutes by default
--- a/workers/docker-worker/config.yml
+++ b/workers/docker-worker/config.yml
@@ -113,6 +113,9 @@ defaults:
     # Tasks should be removed from the queue if they have been dequeued a lot.
     # Possible signs that the task is bad
     dequeueCount: 15
+    # Maximum amount of time to wait for the task setup (e.g. downloading
+    # docker images), in seconds
+    maxSetupTime: 600
 
   taskQueue:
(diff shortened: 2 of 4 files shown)


petemoore:
Thanks for looking into this Julien! Do you think it would be possible to add a unit test? Ideally it would be great if the unit test could use a shorter timeout (e.g. 10ms) so that the unit test runs quickly. I realise this is still in draft, but wanted to give early feedback... Many thanks!

jcristau:
Thanks for the feedback Pete!  The lack of a test is the reason this is still in draft actually :)

jcristau:
The approach here is all wrong, so I'm going to close this.

lotas:
Hey @jcristau do you mind sharing more details? Did you find some other issue or have a better fix on your mind?
Thanks!

jcristau:
My patch didn't actually kill/interrupt the blocking operations, so it didn't fix anything (even aside from the fact that it caused docker-worker to call abortRun twice, and crash on the second call).

I was basing my change on the maxRunTime handling, but that's a lot different because what it needs to interrupt is in a separate process which it can just kill, whereas the blocking operation here may be things like an in-process http download or an out-of-process docker pull operation.
It might be easier to make changes similar to #6133 to other blocking operations, e.g. adding a timeout to the dockerode instance that we use for pulls.

As an aside, I've noticed e.g. https://github.com/taskcluster/taskcluster/blob/2c07f52a5dc41aec21513d605400df26b07bc51b/workers/generic-worker/mounts.go#L817 doesn't seem to set a timeout, either (if I'm reading it correctly), so there may be similar issues lurking in g-w as well.

PR #6135, opened 2023-03-28
Quarantine worker with comment

### LINKED ISSUE #4343: quarantine metadata / comment
In [bug 1689822](https://bugzilla.mozilla.org/show_bug.cgi?id=1689822), we had machines quarantined that shouldn't have been. We don't know who quarantined a machine, when, or why, just if it's quarantined, and the time it will come out of quarantine.

Can we:
- add metadata to the quarantine UI: who quarantined and when? (If we don't store this info, then we may need to update the backend as well)
- add metadata to the backend+UI: a comment and/or why section. I imagine being able to query for this information through the api would be good for both the UI and for people looking to generate reports.

This isn't actively blocking any priority task, but resolving this issue would help us maintain hardware or other long-lived pools.

lotas:
* `queue.quarantineWorker` accepts `quarantineInfo` string which is stored along with the `clientId` to keep the history of changes made to the worker
* `worker-manger.getWorker` returns additional field `quarantineDetails` which will be a list of all quarantine requests
* UI allows to provide a comment when quarantining worker and also displays the list of recent changes

image

Fixes #4343

Note: only single `getWorker` call returns this field `quarantineDetails`. Assumption here was that if this information will be needed, it will be on a per-worker basis, and not for all workers at once. Also this reduces amount of transmitted data.


changelog/issue-4343.md
```diff
@@ -0,0 +1,10 @@
+audience: admins
+level: minor
+reference: issue 4343
+---
+
+Workers can be quarantined with optional comment. `queue.quarantineWorker` accepts `quarantineInfo` string.
```
matt-boris, line 6:
```suggestion
Workers can be quarantined with an optional comment. `queue.quarantineWorker` accepts a `quarantineInfo` string.
```

services/worker-manager/schemas/v1/worker-response.yml
```diff
@@ -49,6 +49,14 @@ properties:
       somewhere in the past).
     type:         "string"
     format:       date-time
+  quarantineDetails:
+    title:       "Worker Quarantine Details"
+    description: |
+      This is a list of changes to the worker's quarantine status. Each entry is a an object
```
matt-boris, line 55:
```suggestion
      This is a list of changes to the worker's quarantine status. Each entry is an object
```

services/worker-manager/src/data.js
```diff
@@ -320,7 +321,7 @@ class Worker {
   // Get a queue worker from the DB, or undefined if it does not exist.
   static async getQueueWorker(db, workerPoolId, workerGroup, workerId, expires) {
     return Worker.fromDbRows(
-      await db.fns.get_queue_worker_with_wm_join(
+      await db.fns.get_queue_worker_with_wm_join2(
```
matt-boris, line 324:
```suggestion
      await db.fns.get_queue_worker_with_wm_join_2(
```

matt-boris (changes requested):
Nice Yarik! Mostly small things here, but one definitely needs updated.

In `services/worker-manager/src/data.js`, `get_queue_worker_with_wm_join2` db func is used instead of `get_queue_worker_with_wm_join_2`

services/worker-manager/test/api_test.js
lotas, line 1: it shows too many changes because lint complained about the formatting, had to apply `--fix`

services/worker-manager/src/api.js
```diff
@@ -918,7 +918,6 @@ builder.declare({
         capacity: worker.capacity || 0,
         providerId: worker.providerId || 'none',
         quarantineUntil: worker.quarantineUntil?.toJSON(),
-        quarantineDetails: worker.quarantineDetails,
```
matt-boris: Why was this removed?

lotas: yeah, that’s where the test failed that I added. So this is a listWorkers endpoint, and I decided not to include this field there
only single getWorker returns it
it was added by accident there initially

matt-boris (approved):
Awesome! 🎉

matt-boris (dismissed):

petemoore (approved):

services/queue/src/api.js
```diff
@@ -2251,14 +2251,22 @@ builder.declare({
   ].join('\n'),
 }, async function(req, res) {
   const { provisionerId, workerType, workerGroup, workerId } = req.params;
-  const { quarantineUntil } = req.body;
+  const { quarantineUntil, quarantineInfo } = req.body;
   const taskQueueId = joinTaskQueueId(provisionerId, workerType);
 
-  const [result] = await this.db.fns.quarantine_queue_worker_with_last_date_active({
+  const quarantineDetails = {
+    clientId: await req.clientId() || 'unknown-client',
```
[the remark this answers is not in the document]
lotas: Oh! great catch.
It looks like middleware would give same string if it cannot determine clientId, so this check is not needed

https://github.com/taskcluster/taskcluster/blob/main/libraries/api/src/middleware/auth.js#L205-L212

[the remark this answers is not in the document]
lotas: And I'm not completely sure where `clientId` is coming from. Most likely from the hawk headers. And most likely it would be set if that endpoint is accessible to anonymous scope

services/queue/schemas/v1/worker-response.yml
```diff
@@ -49,6 +49,47 @@ properties:
       somewhere in the past).
     type:         "string"
     format:       date-time
+  quarantineDetails:
+    title:       "Worker Quarantine Details"
+    description: |
+      This is a list of changes to the worker's quarantine status. Each entry is an object
+      containing information about the time, clientId and reason for the change.
+    type:        array
+    uniqueItems: false
```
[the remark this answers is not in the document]
lotas: Thanks!

lotas:
i merged it locally because github statuses were not updating checks

PR #572, opened 2019-04-14
Add docs for collaborators_quiet policy

iFlameing:
Bugzilla Bug: [1446768](https://bugzilla.mozilla.org/show_bug.cgi?id=1446768)

PR #327, opened 2019-02-28
Bug[1446768] Adding new policy in .taskcluster.yml

iFlameing:
Bugzilla Bug: [1446768](https://bugzilla.mozilla.org/show_bug.cgi?id=1446768)

This pr belongs to the 1st part of this bug.

@owlishDeveloper I added options in policy. How I can check the policy if it is collaborators or not?

PR #6137, opened 2023-03-29
Github: add 'collaborators_quiet' to the schema.

lotas:
It looks like this was forgotten during initial implementations in #327 and #572 of the <https://bugzilla.mozilla.org/show_bug.cgi?id=1446768>

### DIFF
--- a/services/github/test/handler_test.js
+++ b/services/github/test/handler_test.js
@@ -747,6 +747,19 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
 
       assert(handlers.createTasks.calledWith({ scopes: sinon.match.array, tasks: sinon.match.array }));
     });
+
+    test('using collaborators_quiet policy should not create comment', async function () {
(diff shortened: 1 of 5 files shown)


matt-boris (dismissed):
Nice, thanks!

services/github/test/handler_test.js
```diff
@@ -747,6 +747,19 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
 
       assert(handlers.createTasks.calledWith({ scopes: sinon.match.array, tasks: sinon.match.array }));
     });
+
+    test('using collaborators_quiet policy should not create comment', async function () {
```
petemoore, line 751: I thought a comment should be created, but only once per PR?

lotas: As I understand this policy and implementation is that it should not send comment if `_quiet` part of it is used.

lotas:
merged it locally because github actions were slow :)

PR #6144, opened 2023-03-30
Force stop instances that are running but not visible to the queue

### LINKED ISSUE #6142: Worker Manager not killing rogue/zombie instances
**Describe the bug**
Worker Manager should kill rogue/zombie instances, when they are not effective.

**To Reproduce**

Run Worker Manager, then have a worker that crashes with out of memory, or something similar.

**Expected behavior**

Worker Manager notices that the worker is not routinely calling `queue.claimWork` or `queue.reclaimTask`, and therefore terminates the instance it is running on.

**Taskcluster version**
All versions

**Screenshots**

image

**Additional context**

Worker Manager should kill all instances that:

a) were requested more than e.g. 30 minutes ago, and are not yet known by the Queue, and
b) workers that are known by the Queue, but haven’t called `queue.claimWork` nor `queue.reclaimTask` in the last hour

lotas:
Worker manager force stops instances that are running but are not claiming any tasks (not visible to the queue).
This is a safety mechanism to prevent workers from running indefinitely when generic-worker (or docker-worker) fails to start properly or died for some reason (like OOM) and can no longer shutdown itself.

Fixes #6142

[tests missing]


db/versions/0084.yml
```diff
@@ -0,0 +1,52 @@
...
+          (workers.state = 'running') and
+          (providers_filter_cond is null or providers_filter_value is null or
+            case
+              when providers_filter_cond = '='
+                then workers.provider_id = ANY(string_to_array(providers_filter_value, ','))
+              when providers_filter_cond = '<>'
+                then workers.provider_id <> ALL(string_to_array(providers_filter_value, ','))
+            end
+          )
+          and workers.last_modified < now() - interval '2 hours'
```
petemoore, line 47: Could the `interval `'2 hours'` be a parameter of the function?

lotas: nevermind, i think I will re-do this PR.. I initially thought that worker scanner doesn't see those instances at all, but that was incorrect. So we need to augment existing query with something that would tell that queue doesn't know about this worker.

PR #6166, opened 2023-04-02
refactor!: drop support for python 3.6

matt-boris:
Python 3.6 is no longer a [supported version](https://devguide.python.org/versions/#supported-versions). This also adds python 3.10 and 3.11 support.

>Remove python 3.6 support as it's past its end-of-life date.
>
>Add python 3.10 and python 3.11 support.


lotas (approved):
nice, thanks @matt-boris

PR #6172, opened 2023-04-12
Delete artifacts improvements.

lotas:
Using dedicated db function `get_expired_artifacts_for_deletion` that only uses expires column and performs very good even without an index as there is just one sequential scan. 

Batch size increased to 1000 and `deleteObjects` is used to delete multiple s3 files at once.

Removing multiple db entries with a single call of `delete_queue_artifacts`

Bugzilla Bug: [1826417](https://bugzilla.mozilla.org/show_bug.cgi?id=1826417)


lotas:
I was running tests on a local table which had 10% of the prod size - 45.000.000 records and queries were  fast.
Same query (selecting expired) works very fast on prod also, so this should give us a boost to catch up with artifacts removal

services/queue/src/utils.js
```diff
@@ -42,86 +41,108 @@ const artifactUtils = {
   async expire({ db, publicBucket, privateBucket, ignoreError, monitor, expires }) {
     let count = 0;
 
-    // Get all expired artifacts using a handler function, operating on one
-    // page of results at a time.  This does not use paginatedIterator because
-    // that would result in a much slower one-by-one expiration, instead of
-    // the parallel handling of each batch.
-    const getExpiredArtifacts = async (expires, handler) => {
-      let after_task_id_in = null;
```
lotas, line 50: we no longer need to mark the "last read" position, since new `get_expired_artifacts_for_deletion` always returns _any_ _first_ records that db gives us that satisfies `expires < ts` filter. 
Those returned rows can even be random, but we don't care since we know we are going to remove them afterwards anyway and they will not appear in any query again

services/queue/src/utils.js
```diff
@@ -42,86 +41,108 @@ const artifactUtils = {
...
-          after_run_id_in = lastRow.run_id;
-          after_name_in = lastRow.name;
-        }
+    // Fetch all expired artifacts and batch delete the S3 ones
+    // then remove the entity from the database
+    // repeat until there are no more expired artifacts
+    while (true) {
+      const rows = await db.fns.get_expired_artifacts_for_deletion({
+        expires_in: expires,
+        page_size_in: 1000,
```
lotas, line 50: on prod db table selecting 1000 records is performed under 200ms on average, so this should speed things up

services/queue/src/utils.js
```diff
@@ -42,86 +41,108 @@ const artifactUtils = {
...
-          let err = new Error('Expiring artifact with bucket which isn\'t ' +
-            'configured for use. Please investigate!');
-          err.bucket = artifact.details.bucket;
-          err.taskId = artifact.taskId;
-          err.runId = artifact.runId;
-          monitor.reportError(err);
-          return;
+      const entries = rows.map(exports.artifactUtils.fromDb);
+
+      const s3public = [];
```
lotas, line 58: since we are doing batch removals, we group by bucket

services/queue/src/utils.js
```diff
@@ -42,86 +41,108 @@ const artifactUtils = {
...
       }
 
-      // note that 'object' artifacts are deleted at expiration by the object
-      // service, so nothing needs done here.
+      const errors = [];
+      const deleteObjects = async (bucket, entries) => {
+        if (entries.length) {
+          try {
+            const response = await bucket.deleteObjects(entries.map(entry => entry.details.prefix));
+            if (response.Errors && response.Errors.length) {
```
lotas, line 84: https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html#API_DeleteObjects_ResponseSyntax

lotas:
tested everything locally with communitytc database. Created few artifacts in minio s3 to make sure they are removed. All works fine

db/test/fns/queue_artifacts_test.js
```diff
@@ -456,6 +456,67 @@ suite(testing.suiteName(), function() {
...
+    }
+
+    let rows = await db.fns.get_expired_artifacts_for_deletion({
+      expires_in: today,
+      page_size_in: 1000,
+    });
+    assert.deepEqual(rows.map(({ run_id }) => run_id), [0, 2, 4, 6, 8]);
+  });
+
+  helper.dbTest('delete_queue_artifacts', async function(db) {
```
matt-boris, line 484: Since the `get_expired_artifacts_for_deletion` test is the exact same as the first lines of this `delete_queue_artifacts` test, I think it's safe to just keep this second test and wipe out the first.

```suggestion
  helper.dbTest('get_expired_artifacts_for_deletion and delete_queue_artifacts', async function(db) {
```

db/versions/0084.yml
```diff
@@ -0,0 +1,54 @@
...
+          queue_artifacts.name,
+          queue_artifacts.storage_type,
+          queue_artifacts.content_type,
+          queue_artifacts.details,
+          queue_artifacts.present,
+          queue_artifacts.expires
+        from queue_artifacts
+        where
+          queue_artifacts.expires < expires_in
+        limit get_page_limit(page_size_in);
```
matt-boris, line 31: What's the need for this `get_page_limit()` call? Does `limit page_size_in` not work? Just curious.

services/queue/src/bucket.js
```diff
@@ -135,6 +135,7 @@ Bucket.prototype.deleteObject = function(prefix) {
 /** Delete a list of objects */
 Bucket.prototype.deleteObjects = function(prefixes) {
   assert(prefixes instanceof Array, 'prefixes must be an array');
+  assert(prefixes.length <= 1000, 'not more than 1000 prefixes can be deleted');
```
matt-boris, line 138: Is this an S3 limit?

services/queue/src/utils.js
matt-boris, line 84: Shouldn't this be `Error`? Or did I misread something in those docs? 😄

```suggestion
            if (response.Error && response.Error.length) {
```

matt-boris (commented):
Looking good! I left a few small suggestions/questions/comments

db/versions/0084.yml
matt-boris: Ah, nvm I see it defined in 0009.yml - makes sense now 👍🏻

db/test/fns/queue_artifacts_test.js
lotas: Yeah, I think you're right.. I thought that maybe at some point they will diverge, so having individual tests is better, but until this happens we can bundle them :+1:

services/queue/src/bucket.js
lotas: yes! I should probably add a link to the doc in the comments

lotas, line 138:
```suggestion
  // This is a S3 API limit: https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html
  assert(prefixes.length <= 1000, 'not more than 1000 prefixes can be deleted');
```

services/queue/src/utils.js
lotas: huh! Controversial one! :) I think the documentation lists response as XML, but when it is parsed by client library, it is being turned into `response.Deleted` and `response.Errors` which are both arrays.. let me try to find some backup for this claim 😬

lotas: Cannot link the exact section, but here js library explains the response object for this call: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#deleteObjects-property 

```
Errors — (Array<map>) 
```

lotas: we should switch to typescript 😬

services/queue/test/helper.js
```diff
@@ -63,7 +63,11 @@ exports.withS3 = (mock, skipping) => {
       exports.load.cfg('aws.accessKeyId', undefined);
       exports.load.cfg('aws.secretAccessKey', undefined);
 
-      const mock = new mockAwsS3.S3({});
+      const mock = new mockAwsS3.S3({
+        params: {
```
lotas, line 67: this is needed for mockAwsS3 to properly construct temp folder path.
Without it, some calls are working fine (the one which contain `Bucket` key, and the ones that don't - fail)

matt-boris (approved):
Nice! 🎉

matt-boris (approved):
*Take two...* Nice! 🎉 😄

matt-boris (approved):

PR #6175, opened 2023-04-14
docker-compose several improvements

lotas:
* `s3_init_buckets:` doesn't remove and create buckets if they exist.
* added multi-line yaml for docker-compose files for readability 
* `nginx.conf` fixed bucket urls to support POST operations on a bucket. Trailing slash was not matched on `"POST /public-bucket?delete"` action

First one was very annoying to debug anything on cron job level, as those services depend on `s3_init_buckets` which removed all contents before service starts, thus making it hard to tell if deletion through API really works or was it this startup script :)


matt-boris (approved):
👍🏻

PR #6181, opened 2023-04-17
Fix profile name handling in UI

### LINKED ISSUE #5773: taskcluster pages fail to load: TypeError: can't access property "displayName", n.profile is undefined
**Describe the bug**
When I opened https://firefox-ci-tc.services.mozilla.com/worker-manager today, the error message `TypeError: can't access property "displayName", n.profile is undefined` got shown in red. The behavior was reproducible with visits to https://firefox-ci-tc.services.mozilla.com/

The issue was resolved after I deleted cookies, local storage etc.

The account used for login is not a `@mozilla.com` and if I remember correct no first name and last name are bound to it - https://firefox-ci-tc.services.mozilla.com/profile says "Signed In As Archaeopteryx Archaeopteryx".

lotas:
In some rear scenarios #5773 `user.profile` or `user` objects became `undefined` or `null`. 

Although the root cause of this issue remains unclear and cannot be reproduced, enhancing the UI's resilience can prevent pages from crashing when such instances occur

Fixes #5773

### DIFF
--- a/changelog/issue-5773.md
+++ b/changelog/issue-5773.md
@@ -1,0 +1,6 @@
+audience: users
+level: minor
+reference: issue 5773
+---
+
+Fix UI bug when user profile was missing and causing whole page to crash.
--- a/changelog/issue-5773.md
+++ b/changelog/issue-5773.md
@@ -1,5 +1,5 @@
 audience: users
-level: minor
+level: patch
 reference: issue 5773
 ---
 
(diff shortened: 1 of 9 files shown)


changelog/issue-5773.md
```diff
@@ -0,0 +1,6 @@
+audience: users
+level: minor
```
matt-boris, line 2: Think this is minor? Seems a little more like a patch bump to me 🤷🏻‍♂️

matt-boris (approved):
LGTM, maybe patch bump though and not minor?

changelog/issue-5773.md
lotas: yes! will update

lotas, line 2:
```suggestion
level: patch
```

matt-boris (approved):
LGTM

PR #6182, opened 2023-04-18
Set status back to pending as soon as task restarts

lotas:
Before it was only listening to `taskRunning` and it took some time to switch github state after rerun, as it takes time to get scheduled and start running

Bugzilla Bug: [1590886](https://bugzilla.mozilla.org/show_bug.cgi?id=1590886)


matt-boris (approved):
Nice 👍🏻

matt-boris (approved):

PR #6148, opened 2023-03-31
[WIP] feat(worker): add `task.payload.onExitStatus.purgeCaches` feature to generic-worker

matt-boris:
Fixes #6147.

>Adds `task.payload.onExitStatus.purgeCaches` feature to generic worker to bring to parity with an existing docker worker feature.
>
>`purgeCaches` is an array of exit status code integers that the user wants all caches associated with the task to be purged.

PR #6183, opened 2023-04-18
fix(generic-worker): only purge task caches

matt-boris:
https://github.com/taskcluster/taskcluster/pull/6148 incorrectly would purge _all_ caches on the worker. This patch makes it so only caches related to the task run get purged. I ensured that the new test that I added fails on the `main` branch.

>Generic worker feature `task.Payload.OnExitStatus.PurgeCaches` now only purges caches related to the task, instead of all caches on the worker.


lotas (approved):

petemoore (approved):
Love it, thanks Boris!

PR #6189, opened 2023-04-26
Fix quarantine worker payload schema

### LINKED ISSUE #6185: web ui makes it seem like new quarantine 'reason' field is optional, but it's not and fails silently
**Describe the bug**
When quarantining a host (see https://firefox-ci-tc.services.mozilla.com/provisioners/proj-autophone/worker-types/gecko-t-bitbar-gw-unit-p5/workers/bitbar/pixel5-05?sortBy=started&sortDirection=desc as an example) the 'reason' field appears optional. If it's omitted there's no indication of error, but the quarantine request doesn't complete.

**To Reproduce**

1. Go to https://firefox-ci-tc.services.mozilla.com/provisioners/proj-autophone/worker-types/gecko-t-bitbar-gw-unit-p5/workers/bitbar/pixel5-05
2. Click on Update Quarantine
3. Click Update (leave options alone).
4. Host will not be quarantined.
  
**Expected behavior**
Host should be quarantined or an error should appear (that no reason was provided).

**Taskcluster version**
49.1.1

lotas:
Reason field was marked as optional, but validation required min length 1

Fixes #6185

### DIFF
--- a/changelog/issue-6185.md
+++ b/changelog/issue-6185.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 6185
+---
+
+Fixed quarantine worker 'reason' field schema to be optional.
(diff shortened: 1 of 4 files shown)


matt-boris (approved):
Nice, LGTM 👍🏻

PR #6204, opened 2023-05-02
Extra check before worker is terminated on terminateAfter time exceeded.

### LINKED ISSUE #6186: Investigate "terminateAfter time exceeded" errors
While investigating `claim-expired` exceptions, I've noticed some pattern in azure workers - they are being created, register themselves, start claiming tasks, but then get shutdown with `terminateAfter time exceeded` reason.

There are hundreds of those daily and about 2k in a week. 
Usually termination and registration are 5-10min apart. 

My theory is that `worker-scanner` is operating on a "stale" data. It fetches batches of 1000 workers and then iterates them. 
What I think happens is that between `worker-scanner` has fetched this worker in batch, and by the time it reached it to check on state, worker has already started up, registered and started claiming work. Registration call should have updated `provider_data->>'terminateAfter'` to be +96hours (based on pool config). 
But `worker-scanner` doesn't refetch this record, so when it sees instance is running but `terminateAfter` has previous value, it forces this worker to be shut down, despite it doing some work already.

lotas:
This is an attempt to reduce the amount of workers that are being stopped even after they were registered and started claiming work. In some conditions, especially for Azure workers, scan loops can take 10-20 minutes. And since all records are being fetched in memory once, it may happen that worker record was updated. We simply reload the record before finally removing worker.

Fixes #6186 

I'm only 99% confident this is the main cause why normally working workers are being removed, because `update_worker_2` function is actually checking `etag` of the record when doing an update, so it's not clear how it can overwrite something that was already updated. But maybe not all updates are being handled with try/catch or are not being logged. 
From the investigation into 2k of killed workers, 85% of those were already registered and working for 2-8 minutes before being killed by worker-scanner.

### DIFF
--- a/services/worker-manager/config.yml
+++ b/services/worker-manager/config.yml
@@ -1,6 +1,5 @@
 defaults:
   app:
-    errorsExpirationDelay: '-4 days' # Anything older than a few days goes away
     provisionerIterateConfig: !env:json:optional PROVISIONER_ITERATE_CONFIG
     workerScannerIterateConfig: !env:json:optional WORKER_SCANNER_ITERATE_CONFIG
   monitoring:
@@ -39,8 +38,6 @@ test:
     trustProxy: false
   taskcluster:
     rootUrl: "https://tc.example.com"
-  app:
-    errorsExpirationDelay: '-1 hour'
   postgres:
     dbCryptoKeys:
       - id: 'worker-manager'
--- a/services/worker-manager/src/worker-scanner.js
+++ b/services/worker-manager/src/worker-scanner.js
@@ -56,7 +56,7 @@ class WorkerScanner {
     const fetch =
       async (size, offset) => await this.db.fns.get_non_stopped_workers_quntil_providers(
         null, null, null, this.providersFilter.cond, this.providersFilter.value, size, offset);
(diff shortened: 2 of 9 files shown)


services/worker-manager/config.yml
```diff
@@ -1,6 +1,5 @@
 defaults:
   app:
-    errorsExpirationDelay: '-4 days' # Anything older than a few days goes away
```
lotas, line 3: this wasn't used anywhere in the code

services/worker-manager/src/worker-scanner.js
```diff
@@ -56,7 +56,7 @@ class WorkerScanner {
     const fetch =
       async (size, offset) => await this.db.fns.get_non_stopped_workers_quntil_providers(
         null, null, null, this.providersFilter.cond, this.providersFilter.value, size, offset);
-    for await (let row of paginatedIterator({ fetch })) {
+    for await (let row of paginatedIterator({ fetch, size: 500 })) {
```
lotas, line 59: reducing size of batches a bit to also lower the risk of records getting outdated by the time they are checked

matt-boris (approved):
👍🏻

PR #6206, opened 2023-05-04
build(deps): node/go upgrades to 18.16.0/1.19.9

matt-boris:
>Upgrade Node.js version to 18.16.0 and go version to 1.19.9.


lotas (approved):
awesome! Thanks

PR #6210, opened 2023-05-08
Add options for jsonschema2go cli for nested structs and defaults

petemoore:
This adds some optional arguments to the jsonschema2go cli for:

* outputting nested structs
* outputting default struct tags

This is needed by d2g. In that project, the docker worker payload schema has some features which are `true` by default (such as `payload.features.bulkLog`). In order for an absent value to be interpreted as `true` rather than `false`, the default will need to be specified in the generated go type as a default struct tag.

### DIFF
--- a/tools/jsonschema2go/jsonschema2go/cli.go
+++ b/tools/jsonschema2go/jsonschema2go/cli.go
@@ -58,11 +58,13 @@ The go type names will be "normalised" from the json subschema Title element.
     cat urls.txt | jsonschema2go -o main
 
   Usage:
-    jsonschema2go -o GO-PACKAGE-NAME
+    jsonschema2go [-d] [-n] -o GO-PACKAGE-NAME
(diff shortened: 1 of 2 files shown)


matt-boris (commented):
Changes LGTM! Toss in a changelog and we can merge 👍🏻

tools/jsonschema2go/jsonschema2go/cli.go
```diff
@@ -75,7 +77,8 @@ func main() {
 		Package:              arguments["-o"].(string),
 		ExportTypes:          true,
 		URLs:                 parseStandardIn(),
-		DisableNestedStructs: true,
+		EnableDefaults:       arguments["-d"].(bool),
```
lotas, line 80: Is this being used anywhere? Or is it passed somewhere downstream that accepts this property?

matt-boris (approved):

PR #6211, opened 2023-05-08
Move workermanager.reregisterWorker into category 'Worker Interface'

petemoore:
This moved workermanager.reregisterWorker into the category "Worker Interface" to match workermanager.registerWorker. This affects where it appears in the docs site, and under which heading. See e.g. https://community-tc.services.mozilla.com/docs/reference/core/worker-manager/api#reregisterWorker.


matt-boris (approved):
Nice, ty

PR #6214, opened 2023-05-09
Issue 6213: remove unused docker engine from Generic Worker

### LINKED ISSUE #6213: Remove Generic Worker Docker Engine
The Generic Worker Docker Engine was an experimental engine that was never used in production. It was an intended starting point for adding support for docker-worker style payloads. However, a new approach to running Docker Worker payloads in the multiuser engine was agreed, and is under [active development](https://github.com/orgs/taskcluster/projects/14), and this will provide the same functionality that the docker engine was intended to provide. Therefore the old, incomplete, and unused docker engine should be removed.

petemoore:
Fixes #6213


matt-boris (approved):
Awesome, thanks Pete! Kill it! 💣

PR #6221, opened 2023-05-10
Cancel older task groups automatically

### LINKED ISSUE #5621: Auto-cancellation of task groups
Create a configuration option to auto-cancel the previously running task group on a new commit to a PR. This will free up resources, time, and $$$.

lotas:
This is the functionality built on top of #6060 and #6054  

`.taskcluster.yml` has additional property `autoCancelPreviousChecks` (`true` by default) that will allow github handler to cancel previous (currently running) task groups.

This will happen in two cases:
1. `pull_request` event. We consider all running builds as obsolete, and will cancel them as soon as the new task graph is created
2. `push` event for the same `SHA` (non default branch)

Note: 
Currently not implemented: Cancelling builds for the same branch if no pull request exists. Do we want this? If there are some specific branch workflows (like release/deploy branches that are non default ones), then probably each push has some kind of attached context, so stopping running builds might not be wise.

Implements https://github.com/taskcluster/taskcluster-rfcs/blob/main/rfcs/0180-Github-cancel-previous-tasks.md 
Fixes #5621


db/fns.md
```diff
@@ -470,6 +476,22 @@ Get github check run id and check suite id
...
 
+### get_github_checks_by_task_group_id
+
+* *Mode*: read
+* *Arguments*:
+  * `page_size_in integer`
+  * `page_offset_in integer`
+  * `task_group_id_in text`
+* *Returns*: `table`
+  * `   task_group_id text`
```
matt-boris, line 487: Screenshot 2023-05-11 at 7 43 44 AM

Not sure it's truly an issue with whitespace or if it's how github's presenting the file, but this line looks more indented for some reason. Maybe there's some extra whitespace in db/versions/0085.yml?

services/github/src/handlers/job.js
```diff
@@ -28,7 +28,12 @@ async function jobHandler(message) {
   let repository = message.payload.repository;
   let sha = message.payload.details['event.head.sha'];
   debug = debug.refine({ owner: organization, repo: repository, sha });
-  let pullNumber = message.payload.details['event.pullNumber'];
+  let pullNumber = message.payload.details['event.pullNumber'] || message.payload.body.number;
+
+  debug(`Retrieving  ${organization}/${repository}@${sha}...`);
```
matt-boris, line 33: Do you still want this debug?

```suggestion
  debug(`Retrieving ${organization}/${repository}@${sha}...`);
```

matt-boris: Ah I see it was a copy/paste

services/github/src/handlers/job.js
```diff
@@ -237,6 +238,16 @@ async function jobHandler(message) {
     return await this.createExceptionComment({ debug, instGithub, organization, repository, sha, error: e });
   }
 
+  // only cancel previous tasks after we have successfully created new ones
+  // Cancel existing builds for non-default branches
+  if (graphConfig.autoCancelPreviousChecks !== false) {
```
matt-boris, line 243: Curious as to why you're explicitly checking vs the `false` value here? Does this not work?

```js
if (graphConfig.autoCancelPreviousChecks) {
```

matt-boris (commented):
Nice! Looking great! This will be awesome to have 👍🏻

Just left a few comments/questions!

services/github/src/handlers/job.js
lotas: eagle-eye :) thanks!

services/github/src/handlers/job.js
lotas: yeah, so turns out it comes straight empty. Schema only validates but doesn't add default values to the missing props. 
So here we expect tc.yaml owner to explicitly say `false`, anything else is enabling this `true`, `undefined`

PR #6227, opened 2023-05-12
Allow github client to seal and cancel task groups

lotas:
This is a followup on #5621 to give extra scopes for github client, as it currently cannot cancel own task groups. 

Github client will get following scopes:
- `queue:seal-task-group:taskcluster-github/*`
- `queue:cancel-task-group:taskcluster-github/*`

When some repository is using different `schedulerId` then corresponding `repo:github.com/` role should be updated to have adjusted roles (to make github client capable of stopping previous builds)

### DIFF
--- a/changelog/issue-5621.md
+++ b/changelog/issue-5621.md
@@ -1,0 +1,7 @@
+audience: admins
+level: patch
+reference: issue 5621
+---
+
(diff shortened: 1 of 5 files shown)


petemoore (approved):
Love it, thanks Yarik!

PR #6231, opened 2023-05-15
Remove unneeded build constraint

petemoore:
Without docker engine, this build constraint is no longer necessary.


matt-boris (approved):
:shipit:

matt-boris (approved):

PR #6233, opened 2023-05-15
Auto-cancellation improvements

lotas:
Several improvements to auto-cancel running builds functionality:

- [x] report errors directly to github comments
- [x] queue client uses limited scopes to make sure other repos cannot overwrite anything
- [ ] implement `github.cancelBuilds({ oraganization, reository, sha?, pullNumber? })` to cancel existing running builds


db/versions/0086.yml
```diff
@@ -0,0 +1,37 @@
...
+          github_builds.event_id,
+          github_builds.pull_request_number,
+          public.gen_random_uuid()
+        from github_builds
+        where
+          github_builds.organization = organization_in and
+          github_builds.repository = repository_in and
+          (sha_in is null or github_builds.sha = sha_in) and
+          (pull_request_number_in is null or github_builds.pull_request_number = pull_request_number_in) and
+          github_builds.state in ('pending')
```
lotas, line 33: this is the only difference with `get_github_builds_pr`, but I wanted to have a dedicated function to hold this logic. Debatable :)

services/github/schemas/v1/build-list.yml
```diff
@@ -35,7 +35,7 @@ properties:
           description: Github revision associated with the build.
         state:
           type: string
-          enum: ['pending', 'success', 'error', 'failure']
+          enum: ['pending', 'success', 'error', 'failure', 'cancelled']
```
lotas, line 38: auto-cancelled builds get this status now `cancelled`.
This is only `github_builds` internal representation and state

services/github/src/api.js
```diff
@@ -449,6 +456,96 @@ builder.declare({
   });
 });
 
+builder.declare({
+  method: 'post',
+  route: '/builds/:owner/:repo/cancel',
+  name: 'cancelBuilds',
+  scopes: 'github:cancel-builds:<owner>:<repo>',
```
lotas, line 463: in order to use parameters in `scopes`, we have to include parameters with the same name to the `route`

This is probably less ideal, as it provides now a different url signature (unlike `.builds()`) but also requires both owner and repo (unlike `builds()`)

services/github/src/handlers/index.js
```diff
@@ -79,15 +81,6 @@ class Handlers {
     assert(!this.deprecatedInitialStatusPq, 'Cannot setup twice!');
     assert(!this.rerunPq, 'Cannot setup twice!');
 
-    // This is a powerful Queue client without scopes to use throughout the handlers for things
```
lotas, line 82: this is moved to the `main.js` so we can inject different fake mocks during testing

petemoore (dismissed):
It looks ok to me, but you may want to get another reviewer that is more familiar with the service code. Thanks @lotas!

matt-boris (approved):
Thanks Yarik!

matt-boris (approved):

PR #6240, opened 2023-05-17
chore: remove docker engine reference

matt-boris:
This is no longer a valid schema to validate against in the create task view on the UI.


lotas (approved):
nice! Thank you

PR #6242, opened 2023-05-17
feat(github): Fix incorrect query for auto-cancellation of builds

lotas:


services/github/src/handlers/index.js
```diff
@@ -287,7 +287,7 @@ class Handlers {
         null,
         organization,
         repository,
-        sha ? null : sha, // we only want to filter either by sha or pull number, not both
+        pullNumber ? null : sha, // we only want to filter either by sha or pull number, not both
```
lotas, line 290: this slipped in after refactoring in previous PR .. from two different calls to one, with this broken logic

matt-boris (approved):

PR #6244, opened 2023-05-19
Clean up

petemoore:
Software Jenga / Kerplunk.


services/auth/test/clients.json
lotas, line 1: I wonder if it will break this test: 
https://github.com/taskcluster/taskcluster/blob/main/services/auth/test/scoperesolver_test.js#L525

services/auth/test/roles.json
lotas, line 1: same with roles https://github.com/taskcluster/taskcluster/blob/main/services/auth/test/scoperesolver_test.js#L488-L525

services/auth/test/clients.json
petemoore: Good catch — you were absolutely right that these tests depend on these fixture files. Investigating your comment led us to discover that the tests were actually silently not running at all, due to a subtle mocha bug: the `performance` suite callback was declared `async`, but mocha doesn't `await` suite callbacks. The `await helper.loadJson()` at line 490 caused everything after it — the six "real roles" tests and the "resolve all clients" test — to simply never be registered. That's why CI passed even with the files deleted.

We've fixed the async bug (replaced with synchronous `readFileSync`), kept the fixture files, and confirmed all 7 tests now run and pass. These fixtures serve as a valuable golden test for the scope resolver — they ensure that any future refactors to the resolution logic produce identical results for the same inputs, even though the data itself is from the old `auth.taskcluster.net` deployment. That kind of regression safety net is worth preserving.

services/auth/test/roles.json
petemoore: Same story here — the roles fixture feeds the `testRealRoles` assertions (lines 499–524) which verify that scope expansion produces the correct output for realistic role hierarchies. Thanks for flagging this; the files are restored and the tests are back in action.

matt-boris (approved):
Looks fine to me, ty 👍🏻

PR #6246, opened 2023-05-19
Issue 6245: generic-worker: support mounting indexed content

### LINKED ISSUE #6245: generic-worker: support mounting indexed artifacts via index namespace + artifact name
Generic Worker supports [four source types](https://github.com/taskcluster/taskcluster/blob/0c0486db2dd0e9f501b0926b81d4e06e2979b2c1/workers/generic-worker/schemas/multiuser_posix.yml#L463-L552) for mounting file system content into a task directory:
* Artifact Content
* URL Content
* Raw Content
* Base 64 Content

It would be useful to have a fifth type, _Indexed Content_ which would be like Artifact Content, except the task ID would be replaced with an Index namespace, for looking up the task ID.

petemoore:
Github Bug/Issue: Fixes #6245

### DIFF
--- a/changelog/issue-6245.md
+++ b/changelog/issue-6245.md
@@ -1,0 +1,14 @@
+audience: users
+level: minor
+reference: issue 6245
+---
+Generic Worker payload now supports declaratively mounting indexed artifacts into the task directory. For example:
+
+```yml
+payload:
+  mounts:
+    content:
+      namespace: my.index.namespace
+      artifact: public/image.jpg
+    file: pics/image.jpg
+```
--- a/workers/generic-worker/mounts.go
+++ b/workers/generic-worker/mounts.go
@@ -838,7 +838,7 @@ func (ic *IndexedContent) Download(taskMount *TaskMount) (file string, sha256 st
 	}
 	ac := &ArtifactContent{
 		Artifact: ic.Artifact,
-		Sha256:   "",
+		SHA256:   "",
 		TaskID:   itr.TaskID,
 	}
 	return ac.Download(taskMount)
--- a/workers/generic-worker/mounts.go
+++ b/workers/generic-worker/mounts.go
@@ -16,6 +16,7 @@ import (
 	"github.com/taskcluster/httpbackoff/v3"
 	"github.com/taskcluster/slugid-go/slugid"
 	tcclient "github.com/taskcluster/taskcluster/v50/clients/client-go"
+	"github.com/taskcluster/taskcluster/v50/internal/mocktc/tc"
 	"github.com/taskcluster/taskcluster/v50/internal/scopes"
 	"github.com/taskcluster/taskcluster/v50/workers/generic-worker/fileutil"
 )
@@ -77,13 +78,13 @@ func (cache *Cache) Rating() float64 {
 	return float64(cache.Hits)
 }
 
-func (cache *Cache) Evict(task *TaskRun) error {
-	if task != nil {
-		task.Infof("[mounts] Removing cache %v from cache table", cache.Key)
+func (cache *Cache) Evict(taskMount *TaskMount) error {
+	if taskMount.task != nil {
+		taskMount.Infof("Removing cache %v from cache table", cache.Key)
 	}
 	delete(cache.Owner, cache.Key)
-	if task != nil {
-		task.Infof("[mounts] Deleting cache %v file(s) at %v", cache.Key, cache.Location)
+	if taskMount.task != nil {
+		taskMount.Infof("Deleting cache %v file(s) at %v", cache.Key, cache.Location)
 	}
 	// delete the cache on the file system
 	return os.RemoveAll(cache.Location)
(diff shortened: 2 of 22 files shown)


workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -487,6 +487,31 @@ definitions:
...
+        artifact:
+          type: string
+          maxLength: 1024
+        sha256:
+          type: string
+          title: SHA 256
+          description: |-
+            The required SHA 256 of the content body.
+
+            Since: generic-worker 10.8.0
```
matt-boris, line 509:
```suggestion
```

Should we remove this? Seems like maybe a copy/pasta error?

workers/generic-worker/schemas/multiuser_windows.yml
```diff
@@ -518,6 +518,31 @@ definitions:
...
+        artifact:
+          type: string
+          maxLength: 1024
+        sha256:
+          type: string
+          title: SHA 256
+          description: |-
+            The required SHA 256 of the content body.
+
+            Since: generic-worker 10.8.0
```
matt-boris, line 540:
```suggestion
```

Same as above.

workers/generic-worker/schemas/simple_posix.yml
```diff
@@ -465,6 +465,31 @@ definitions:
...
+        artifact:
+          type: string
+          maxLength: 1024
+        sha256:
+          type: string
+          title: SHA 256
+          description: |-
+            The required SHA 256 of the content body.
+
+            Since: generic-worker 10.8.0
```
matt-boris, line 487:
```suggestion
```

Same as above.

workers/generic-worker/mounts.go
```diff
@@ -756,50 +810,79 @@ func UnmarshalInto(c json.RawMessage, fsContent FSContent) (FSContent, error) {
...
+	itr, err := taskMount.index.FindTask(ic.Namespace)
+	if err != nil {
+		return "", "", err
+	}
+	ac := &ArtifactContent{
+		Artifact: ic.Artifact,
+		Sha256:   "",
+		TaskID:   itr.TaskID,
+	}
+	return ac.Download(taskMount)
```
matt-boris, line 844: _Love_ how we can just take advantage of the `ArtifactContent` methods here!

workers/generic-worker/mounts.go
```diff
@@ -756,50 +810,79 @@ func UnmarshalInto(c json.RawMessage, fsContent FSContent) (FSContent, error) {
...
 }
 
+func (ic *IndexedContent) Download(taskMount *TaskMount) (file string, sha256 string, err error) {
+	itr, err := taskMount.index.FindTask(ic.Namespace)
+	if err != nil {
+		return "", "", err
+	}
+	ac := &ArtifactContent{
+		Artifact: ic.Artifact,
+		Sha256:   "",
```
matt-boris, line 841: Should we calculate this and set it?

workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -487,6 +487,31 @@ definitions:
...
+          type: string
+          maxLength: 255
+        artifact:
+          type: string
+          maxLength: 1024
+        sha256:
+          type: string
+          title: SHA 256
+          description: |-
+            The required SHA 256 of the content body.
```
matt-boris, line 507: The code suggests this may not be required (it's setting to `""` by default). Also, the schema doesn't list it as required. Probably needs to be reworded here and in the other schemas :)

matt-boris (commented):
Looking nice! Left a few comments/suggestions :)

workers/generic-worker/mounts.go
petemoore: The mounts feature calculates the value, and compares it to what is passed in here, so if we did calculate it, the two values would match, but there would be a little bit more overhead. If we don't set it, the value isn't checked. Note, I could have just not included it in the struct literal, because the zero value would be the empty string, but I just included it here to be a bit more explicit, so that it is clear that it exists, but isn't used.

petemoore:
I'll see if I can add some tests on Monday, so will leave this open. Thanks for the review so far!

petemoore:
@matt-boris By the way, just a heads up that the biggest code change is to workers/generic-worker/mounts.go - but github is collapsing (hiding) those changes in the "Files changed" tab.

matt-boris (dismissed):
Perfect! Thanks Pete

changelog/issue-6245.md
```diff
@@ -0,0 +1,14 @@
+audience: users
+level: minor
+reference: issue 6245
+---
+Generic Worker payload now supports declaratively mounting indexed artifacts into the task directory. For example:
```
lotas, line 5: is it worth adding that workers would need an extra scope to be able to "talk to index artifacts"?

lotas (approved):
👍

PR #6250, opened 2023-05-20
Update the rabbitmq image

Eijebong:
I've had issues with the previous not starting and eating all of my RAM, not sure why but updating fixes it 🤷.


petemoore (approved):
Thanks for the fix!

PR #6251, opened 2023-05-20
Mount the changelog in the UI container in dev mode

Eijebong:
The changelog is necessary to be able to compile
`ui/src/views/Documentation/Changelog/index.jsx` and needs to be at the root of the app.


petemoore (commented):
Awesome, thanks @Eijebong!

petemoore (approved):

lotas:
oops, thanks @Eijebong

PR #6255, opened 2023-05-23
Get rid of node-fetch that was only used in tests

lotas:
Minus one dependency.

While trying to figure out how to migrate to ESM, I realized we don't need `node-fetch` at all.
This doesn't solve the migration problem later on, but avoids us being blocked by security updates to ESM-only packages ;) 

it is still used by some of the dependencies, but this is not our concern


services/web-server/src/loaders/tasks.js
```diff
@@ -53,7 +53,7 @@ module.exports = ({ queue, index }, isAuthed, rootUrl, monitor, strategies, req,
             'public/actions.json',
           );
 
-          const raw = await fetch(url);
+          const raw = await got(url).json();
```
lotas, line 56: `got` supports [retries](https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md#retry) by default, so no need for custom fetch implementation

changelog/AQ5cFhMXRpOKM2RwvYecmQ.md
```diff
@@ -0,0 +1,3 @@
+audience: developers
+level: silent
```
matt-boris, line 2:
```suggestion
level: patch
```

Wdyt? Plus a small changelog about the remove/replace too?

matt-boris (approved):

PR #6267, opened 2023-05-26
Quarantine worker on removeWorker

### LINKED ISSUE #6247: claimWork calls succeed when worker is shutting down already
There is a race condition within workers that are polling for new tasks and shutting themselves down at the same time. 
From the logs we can see that there are many instances of workers being removed `workerManager.removeWorker` and then within few seconds `queue.claimWork` successfully gives task to such workers. This results in `claim-expired` errors

This issue can be partially solved with newer versions of generic-worker and its improved interaction with worker-runer (or working without it). But it will take long time to replace all existing pools. Also upgrading docker-worker images might not be possible.
So it makes sense to "optimize" queue service to reject such calls.

At the moment queue service doesn't know anything worker states, as it is not its main concern. 

Possible fixes could be:
- `auth` service could reject worker's temporary credentials (blacklist approach), so queue service would simply fail with `403`. But this will result in race conditions between worker-manager and auth service - adding workerId to blacklist
- `queue` service might "talk" to worker-manager to query worker state before returning task.

Problem, however, is that `claimWork` itself is not instant, it can "sleep" for 20s if there are no tasks immediately available, and worker shutdown event can happen during this period.

lotas:
This patch should theoretically lower the number of `claim-expired` exceptions when `queue` service serves tasks to the workers that were requested to be removed. 

Queue service already checks quarantine field to avoid giving tasks to such workers, so this approach doesn't make any changes to the queue itself, only uses existing functionality

It will not solve all race conditions, as `claimWork` is waiting for up to 20s on each call, so worker could be already making this call, when instance would be removed by worker-runner.

Fixes #6247

### DIFF
--- a/changelog/issue-6247.md
+++ b/changelog/issue-6247.md
@@ -1,0 +1,8 @@
+audience: admins
+level: minor
+reference: issue 6247
+---
+
+Worker manager now also quarantines worker on `removeWorker` call. This is used to prevent some race conditions when worker is still polling for new work and is removed/shutdown at the same time.
+
+This also requires configuration changes so that `static/taskcluster/worker-manager` client would be able to quarantine workers: adding `queue:quarantine-worker:<..>/*` scope to the `worker-type:*` role should be enough.
--- a/changelog/issue-6247.md
+++ b/changelog/issue-6247.md
@@ -4,5 +4,3 @@ reference: issue 6247
 ---
 
 Worker manager now also quarantines worker on `removeWorker` call. This is used to prevent some race conditions when worker is still polling for new work and is removed/shutdown at the same time.
-
-This also requires configuration changes so that `static/taskcluster/worker-manager` client would be able to quarantine workers: adding `queue:quarantine-worker:<..>/*` scope to the `worker-type:*` role should be enough.
--- a/services/auth/src/static-scopes.json
+++ b/services/auth/src/static-scopes.json
@@ -93,6 +93,7 @@
       "notify:email:*",
       "queue:claim-work:*",
       "queue:pending-count:*",
+      "queue:quarantine-worker:*",
       "queue:worker-id:*",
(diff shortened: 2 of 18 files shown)


petemoore (approved):
I think we can do this, but roll it back if it doesn't really help.

I think what we decided is:
* Worker Runner sometimes thinks a worker isn't active when it is
* Worker Runner calls Worker Manager and tells it to remove the worker
* Worker Runner triggers a shutdown of the machine
* Possibly something causes the worker to restart, so it calls claimWork again, and is given a task
* Worker dies before task is resolved, so is eventually resolved exception/claim-expired by some Queue process

We decided this patch may be useful for reducing the cases of this happening, but the ideal fix would be to fix worker runner from falsely thinking the worker is running, or perhaps docker-worker is really hanging, but then when worker manager shuts down the worker, docker-worker is starting up again for some reason. However, since we are discontinuing docker-worker, and possibly worker-runner too, this is a reasonable fix to try, to see if it improves the situation.

If after this, things are no better, we will roll this back.
If this fixes the issue only for docker-worker, and the needle doesn't change much on generic-worker, we will also roll this back once we have removed docker-worker from taskcluster.
If this fixes things also with worker-runner/generic-worker, we can try removing worker runner to see if things improve. If there is no worker-runner, and there is no docker-worker, and this fix continues to improve the situation, we need to revisit the design to understand what is going wrong. In short, this fix shouldn't really be needed, it is masking a problem which is occurring elsewhere, and we are only making this fix this way for now, since we are planning on getting rid of docker-worker and possible worker-runner too.

@lotas Have I accurately captured our discussion, and is this all ok with you?

lotas:
@petemoore yes, this is correct.

Plus there is situation when worker manager shuts down workers on its own, which might be doing something (terminateAfter) but those should be quite rare

PR #6271, opened 2023-05-31
generic-worker: add tests for mounting indexed artifacts

petemoore:
This is a follow-up from https://github.com/taskcluster/taskcluster/pull/6246#issuecomment-1554976149


petemoore:
Note, this is just an initial indexed artifact test. It could be expanded in future to cover more test cases:
1) that when the index is updated (new task ID) for an artifact that has already been cached, and then a task references it, that the new artifact is downloaded
2) that when the index is not updated for an artifact that has already been cached, or is updated with e.g. a higher rank but the task ID remains the same, and a new task references the indexed artifact, that the cached artifact is used

matt-boris (commented):
I like it Pete, thanks! Only tiny nit from me is variable naming. Using `payload` is a _bit_ confusing since we generally refer to payloads as the task payload. So it hurts readability just a tiny bit. Instead, maybe we could use `itr` for `InsertTaskRequest`? Wdyt?

matt-boris (approved):

workers/generic-worker/mounts_test.go
```diff
@@ -322,8 +325,9 @@ func TestValidSHA256(t *testing.T) {
 	setup(t)
 	taskID := CreateArtifactFromFile(t, "unknown_issuer_app_1.zip", "public/build/unknown_issuer_app_1.zip")
 
-	// whether permission is granted to task user depends if running under windows or not
-	// and is independent of whether running as current user or not
+	// Whether permission is granted to task user depends if running multiuser
+	// engine or simple engine but is independent of whether running as current
```
matt-boris:
```suggestion
	// engine or insecure engine but is independent of whether running as current
```

workers/generic-worker/mounts_test.go
```diff
@@ -392,8 +396,9 @@ func TestFileMountNoSHA256(t *testing.T) {
...
```
matt-boris:
```suggestion
	// engine or insecure engine but is independent of whether running as current
```

workers/generic-worker/mounts_test.go
```diff
@@ -550,8 +555,9 @@ func TestWritableDirectoryCacheNoSHA256(t *testing.T) {
...
```
matt-boris:
```suggestion
	// engine or insecure engine but is independent of whether running as current
```

workers/generic-worker/mounts_test.go
```diff
@@ -907,8 +913,9 @@ func TestCacheMoved(t *testing.T) {
...
```
matt-boris:
```suggestion
	// engine or insecure engine but is independent of whether running as current
```

workers/generic-worker/mounts_test.go
```diff
@@ -1006,8 +1013,9 @@ func TestMountFileAndDirSameLocation(t *testing.T) {
...
```
matt-boris:
```suggestion
	// engine or insecure engine but is independent of whether running as current
```

PR #6274, opened 2023-05-31
fix GW_TESTS_USE_EXTERNAL_TASKCLUSTER: wip

### LINKED ISSUE #6272: GW_TESTS_USE_EXTERNAL_TASKCLUSTER broken
In https://github.com/taskcluster/taskcluster/blob/63b4aefdacff0ed776203880c4946a3bd9d54847/workers/generic-worker/helper_test.go#L463 we provide the possibility of running generic-worker tests against a real taskcluster deployment rather than against the mocktc version of taskcluster written in go.

This is a nice feature, and we should probably run the tests against the real service before making a release. We stopped using real services in the CI because it was pretty slow, and running against mocks was very fast. However, there is always a risk that a service behaviour changes, and we don't notice, since we are running against the mocked version.

It looks like this is broken at the moment, if you set GW_TESTS_USE_EXTERNAL_TASKCLUSTER to e.g. "true" and then run tests, you will get failures. We need to find out what is going wrong and fix it.

This is also useful for situations when new mocks need to be written, and we want to be sure the mocks behave like the real services (which is something I am doing right now in #6271).

petemoore:
Fixes #6272


matt-boris (approved):
👍🏻 Nice!

matt-boris (approved):

PR #6275, opened 2023-05-31
Revert "Quarantine worker on removeWorker"

lotas:
Reverts taskcluster/taskcluster#6267

it is a temporary (experimental) fix in attempt to reduce `claim-expired` exceptions. 
It was released with `51.1.0`.
If it doesn't give us the desire effect and amount of `claim-expired` wouldn't go down, we should revert this pull request https://github.com/taskcluster/taskcluster/pull/6267 

Upd:
Evidence suggests that there was no significant impact on the `claim-expired` exceptions number, so no need to keep that longer in codebase

### DIFF
--- a/changelog/issue-6247.md
+++ b/changelog/issue-6247.md
@@ -1,0 +1,6 @@
+audience: admins
+level: minor
+reference: issue 6247
+---
+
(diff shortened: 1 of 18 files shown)


matt-boris (approved):
👍🏻

PR #6302, opened 2023-06-01
Don't run generic-worker decision task and worker runner tests as root user

petemoore:
We have two distinct types of worker pools in proj-taskcluster:

* proj-taskcluster/gw-ci-\<platform\> (privileged - tasks run as root/LocalSystem)
* proj-taskcluster/gw-\<platform\> (regular - tasks run as regular task users)

The first type is intended for generic-worker CI, which when testing generic worker, needs to be able to create OS users on-the-fly.

The second type is for regular tasks that don't require the additional privileges. The generic-worker decision task and worker runner tests don't require these privileges, so in this PR I am moving them to the regular (less privileged) worker pools.

### DIFF
--- a/changelog/SH9GLDhmS86nZh2mY3JzTw.md
+++ b/changelog/SH9GLDhmS86nZh2mY3JzTw.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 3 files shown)


lotas (approved):
cool 👍

PR #6303, opened 2023-06-02
Include empty .go file for windows in interactive package

petemoore:
`workers/generic-worker/build.sh -a` was failing with: `package github.com/taskcluster/taskcluster/v51/workers/generic-worker/interactive: build constraints exclude all Go files in /Users/pmoore/git/taskcluster/workers/generic-worker/interactive`.

This fixes that issue.

Not sure why this didn't break releases, will investigate.


petemoore:
> Not sure why this didn't break releases, will investigate.

Looks like releases run `build.sh -p` rather than `build.sh -a` and that skips the building of the test packages.

lotas (approved):

PR #6307, opened 2023-06-02
Fix: fail task if decision task fails

petemoore:
I happened to notice that a decision task was resolved as successful, even though it failed (unfortunately I've lost the reference to the task/github check) and it was because the exit code was from the `rm artifacts/actions.json` command rather than the `~/.local/bin/taskgraph decision` command.

The fix is to make sure `bash` runs with the `-e` option.


lotas (approved):
nice

PR #6309, opened 2023-06-03
Fix an intermittent test failure in client-rust

### LINKED ISSUE #6280: Intermittent rust test failure
**Describe the bug**
I had a rust test failure today that I've not seen before. My guess is that it is intermittent, and unrelated to my changes.

Issue was here: https://github.com/taskcluster/taskcluster/runs/13918398445

Looking at the code, it appears to write DATA to a file ('HELLO/WORLD') and then reads it from a file, checking that the content matches what was written to the file. It attempts to read the data twice. The first read was the one that failed in this case, so wasn't related to data being read twice.

My _guess_ is that perhaps:

```rs
file.write_all(DATA).await?;
```

assumes that the statement blocks until the data is flushed to the file and the file handle is closed. But perhaps code continues before the OS has flushed the data, so that the subsequent lines that read the data from the file don't see the new data that has been written to the file. This would explain why the left side is empty (no bytes) and the right side is the full data (the data that was meant to be written to the file) and also why the issue would be intermittent, because probably normally the OS immediately flushes the data:

```rs
assert_eq!(&copy_from_factory(&mut factory).await?, DATA);
```

@djmitche What do you think?

djmitche:
This is due to a tokio bug where `try_clone` does not wait for pending operations to finish.

I wasn't able to reproduce after making this fix.

Github Bug/Issue: Fixes #6280

### DIFF
--- a/changelog/issue-6280.md
+++ b/changelog/issue-6280.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 6280
+---
--- a/clients/client-rust/upload/src/factory.rs
+++ b/clients/client-rust/upload/src/factory.rs
@@ -81,7 +81,12 @@ mod test {
     #[tokio::test]
     async fn file_reader_twice() -> Result<()> {
         let mut file: File = tempfile()?.into();
+
         file.write_all(DATA).await?;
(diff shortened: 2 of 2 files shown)


lotas (approved):
awesome! Thanks

PR #6263, opened 2023-05-24
Added .taskcluster.yml validation tool

lotas:
Experimental for now.
Intention is to allow devs to test their `.taskcluster.yml` before committing it, or detect possible issues with event configuration.

This might later be merged together with "Github Quickstart" but for now is  a standalone view

image

image

PR #6312, opened 2023-06-05
feat(github): Adds renderTaskclusterYaml endpoint

lotas:
This is a followup for #6265 and #6263 that introduces api endpoint with the same functionality.

This even goes further and re-uses actual task rendering routines and produces both tasks and scopes. 

UI allows to render graph for all combinations of pull request and release with actions

todo:
- [x] introduce `github.renderTaskclusterYml` endpoint
- [x] update UI debugger to use that endpoint instead of rendering in-browser

image

image

endpoint accepts:
- `body` as taskcluster.yml raw contents
- `organization` - github org
- `repository` - github repo
- `fakeEvent.type` to simulate one of the github events
- `fakeEvent.action` to check for different PR actions
- `fakeEvent.overrides` - custom object to be passed to jsone directly

It will use fake github payloads to simulate those events: https://github.com/taskcluster/taskcluster/blob/f2d95c81d3a34c1988e8b49980ef1fb8710100e5/services/github/src/fake-payloads.js


clients/client-go/tcgithub/types.go
```diff
@@ -110,6 +112,98 @@ type (
...
+		Repository string `json:"repository,omitempty"`
+	}
+
+	// Rendered .taskcluster.yml output.
+	RenderTaskclusterYmlOutput struct {
+
+		// Scopes that will be used for the tasks
+		//
+		// Array items:
+		Scopes []string `json:"scopes"`
```
petemoore, line 198: isn't this also in the tasks returned?

clients/client-go/tcgithub/types.go
```diff
@@ -110,6 +112,98 @@ type (
...
+
+		// Rendered task
```
petemoore, line 200: this says "task" but it looks like an ordered list of tasks

clients/client-go/tcgithub/types.go
```diff
@@ -110,6 +112,98 @@ type (
...
+		//
+		// Array items:
+		// Additional properties allowed
+		Tasks []json.RawMessage `json:"tasks"`
```
petemoore, line 204: isn't this guaranteed to be a task structure rather than an array of arbitrary objects?

clients/client-go/tcgithub/tcgithub.go
```diff
@@ -338,6 +338,19 @@ func (github *Github) CreateComment(owner, repo, number string, payload *CreateC
...
 }
 
+// Stability: *** EXPERIMENTAL ***
+//
+// This endpoint allows to render the .taskcluster.yml file for a given event or payload.
+// This is useful to preview the result of the .taskcluster.yml file before pushing it to
+// the repository.
+//
+// See #renderTaskclusterYaml
+func (github *Github) RenderTaskclusterYaml(payload *RenderTaskclusterYmlInput) (*RenderTaskclusterYmlOutput, error) {
```
petemoore, line 348: Maybe we should add YAML to [commonInitialisms](https://github.com/taskcluster/taskcluster/blob/f81b376f030cf2f7fc5a0898529614a286b5492d/tools/jsonschema2go/text/text.go#L44)

clients/client-go/tcgithub/types.go
```diff
@@ -110,6 +112,98 @@ type (
 	// Syntax:     ^[a-zA-Z0-9]{8}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{12}$
 	GithubGUID string
 
+	// Render .taskcluster.yml for given event.
+	RenderTaskclusterYmlInput struct {
```
petemoore, line 116: Let's consistently use yaml or yml if possible. Further up we have RenderTaskclusterYaml. If we go with yml instead of yaml (I don't mind which) then let's add that as a common initialism instead. But maybe I prefer YAML to YML.

services/github/schemas/v1/taskcluster-render-yaml-input.yml
```diff
@@ -0,0 +1,76 @@
...
+type:         object
+properties:
+  body:
+    title: Taskcluster.yml contents
+    description: The contents of the .taskcluster.yml
+    type: string
+  organization:
+    type: string
+    description: The organization that owns the repository.
+    pattern: "^[a-zA-Z0-9_-]+$"
```
petemoore, line 14: did you get this regexp from GitHub docs?

services/github/schemas/v1/taskcluster-render-yaml-input.yml
```diff
@@ -0,0 +1,76 @@
...
+    default: TaskclusterRobot
```
petemoore, line 15: why is there a default here? is this what the github service users as a default if no org is specified? when does GitHub not have an org in an event message?

services/github/schemas/v1/taskcluster-render-yaml-input.yml
```diff
@@ -0,0 +1,76 @@
...
+  repository:
+    type: string
+    description: The repository name.
+    pattern: "^[a-zA-Z0-9_-]+$"
+    default: hooks-testing
```
petemoore, line 20: same question here

services/github/schemas/v1/taskcluster-render-yaml-input.yml
```diff
@@ -0,0 +1,76 @@
...
+      - unpublished
+      - created
+      - edited
+      - deleted
+      - prereleased
+      - released
+  fakeEventData:
+    type: object
+    description: |
+      Additional data to be added to the event object.
```
petemoore, line 71: could you provide more context here? when would you want to do this, what type of information can be provided? in a real situation if there was additional data here, where would it have come from? are there e.g. github docs that describe what additional data is allowed?

services/github/schemas/v1/taskcluster-render-yaml-input.yml
```diff
@@ -0,0 +1,76 @@
...
+      - labeled
+      - ready_for_review
+      - review_requested
+      - review_request_removed
+      - unassigned
+      - unlabeled
+  fakeReleaseAction:
+    type: string
+    description: Name of the release action
+    default: published
```
petemoore, line 59: curious about the choice of defaults overall. I'm not sure it doesn't complicate things without asking users to make an explicit choice.

services/github/schemas/v1/taskcluster-render-yaml-output.yml
```diff
@@ -0,0 +1,22 @@
+$schema: "/schemas/common/metaschema.json#"
+title: "Render .taskcluster.yml output"
+description: Rendered .taskcluster.yml output.
+type: object
+properties:
+  tasks:
+    type: array
+    description: Rendered task
```
petemoore, line 8: Rendered tasks?

services/github/schemas/v1/taskcluster-render-yaml-output.yml
```diff
@@ -0,0 +1,22 @@
...
+    uniqueItems: false
+    items:
+      type: object
```
petemoore, line 11: why not a task definition?

services/github/schemas/v1/taskcluster-render-yaml-output.yml
```diff
@@ -0,0 +1,22 @@
...
+      additionalProperties: true
+  scopes:
```
petemoore, line 13: is this not already in all of the rendered tasks?

clients/client-go/tcgithub/types.go
lotas: Let me see if I can reuse those schemas somehow, yesterday go generator was throwing too many errors at me when I tried to

clients/client-go/tcgithub/tcgithub.go
lotas: oh, so it is supposed to be uppercase always?

services/github/schemas/v1/taskcluster-render-yaml-input.yml
lotas: No, this is actually a soft-check .. couldn't find any "official regexps". 
There are some on [stackoverflow](https://stackoverflow.com/questions/59081778/rules-for-special-characters-in-github-repository-name/59082561#comment132563138_59082561) though.

services/github/schemas/v1/taskcluster-render-yaml-input.yml
lotas: good question. This endpoint was intended to be used for debug purposes, so you could send your .taskcuster.yml as string and one of the events to be faked, that would produce task definitions using mocked data. 
We can remove default value from schema, but we'll still use something similar in code when fake event would be mocked

services/github/schemas/v1/taskcluster-render-yaml-input.yml
lotas: 👍 will add. 
If someone wants to send his test payload, instead of doing mocked one, this could be a way to send full event payload, that will be later passed to jsone context and rendered into tasks

services/github/schemas/v1/taskcluster-render-yaml-input.yml
lotas: good question! My assumption was that it should be easy to test generic events like

```
http POST http://taskcluster/api/github/v1/taskcluster-yml body=@.taskcluster.yml fakeEventType=github-push
http POST http://taskcluster/api/github/v1/taskcluster-yml body=@.taskcluster.yml fakeEventType=github-pull-request
```
and if you need to be more specific you could 
```
http POST http://taskcluster/api/github/v1/taskcluster-yml body=@.taskcluster.yml fakeEventType=github-pull-request fakePullRequestAction=synchronize
```

But I see your point. We can also require all of it and don't assume any defaults at all 🤷

services/github/schemas/v1/taskcluster-render-yaml-output.yml
lotas: go codegen didn't like me initially, will see if I can make him love me again

services/github/schemas/v1/taskcluster-render-yaml-output.yml
lotas: those are different scopes! but it is a very good observation. 
I should add a description that those scopes are used for `queueClient` when github service is creating those tasks in the first place. But tasks, themselves, have other scopes inside their definition body

clients/client-go/tcgithub/types.go
lotas: I spent some time experimenting and thinking about it, and decided not to mix queue-service knowledge into this endpoint. Same goes with validation. 
I think we should keep it simple

clients/client-go/tcgithub/tcgithub.go
lotas: replaced namings

services/github/schemas/v1/taskcluster-render-yaml-input.yml
lotas: restructured payload, and added descriptions, it should be more clear now

services/github/schemas/v1/taskcluster-render-yaml-output.yml
lotas: I decided against mixing queue entities into github service to be consistent with design and existing functionality. Github currently doesn't validate produced tasks that it later sends to the queue service to create such tasks. So I think it's better to keep it this way, and if someone wants to check the generated payload, he is able to do it once he gets rendered tasks out.
Btw, this is now implemented in the UI section. UI uses github service to render yaml into tasks, and then validates those tasks against `create-task-request.json` schema, and presents them in the UI

services/github/schemas/v1/taskcluster-render-yaml-output.yml
lotas: improved documentation

clients/client-go/tcgithub/types.go
```diff
@@ -110,6 +112,136 @@ type (
...
+		Organization string `json:"organization,omitempty"`
+
+		// Syntax:     ^[-a-zA-Z0-9_.]{1,100}$
+		Repository string `json:"repository,omitempty"`
+	}
+
+	// Rendered .taskcluster.yml output.
+	RenderTaskclusterYmlOutput struct {
+
+		// Scopes that will be used by the github client to create tasks.
```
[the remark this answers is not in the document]
lotas: they actually do! Depending on type of event, branch, and tag version, they would produce different set of scopes. 

The reason I decided to include those is because this is somewhat hidden to the user and not obvious that different there can be different scopes.

clients/client-go/tcgithub/types.go
```diff
@@ -110,6 +112,136 @@ type (
 	// Syntax:     ^[a-zA-Z0-9]{8}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{12}$
 	GithubGUID string
 
+	// .taskcluster.yml supports `github-pull-request` and `github-pull-request-untrusted` events.
+	// The difference is that `github-pull-request-untrusted` will use different set of scopes.
```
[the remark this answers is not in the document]
lotas: yes, this is coming from this RFC: https://github.com/taskcluster/taskcluster-rfcs/blob/main/rfcs/0175-restricted-pull-requests.md 
I can leave a link there

services/github/schemas/v1/render-taskcluster-yml-output.yml
```diff
@@ -0,0 +1,24 @@
...
+    description: Rendered tasks
+    uniqueItems: false
+    items:
+      type: object
+      additionalProperties: true
+  scopes:
+    type: array
+    description: |
+      Scopes that will be used by the github client to create tasks.
+      Those are different that the scopes inside the tasks itself.
```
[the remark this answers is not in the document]
lotas: it is influenced indeed. 
I think it is useful, as it makes github service more transparent by exposing both parts of the task creation process.

clients/client-go/tcgithub/types.go
```diff
@@ -110,6 +112,136 @@ type (
...
+		// This can be used to set some specific properties of the event or override the existing ones.
+		// For example:
+		//   "ref": "refs/heads/main"
+		//   "before": "000"
+		//   "after": "111"
+		//
+		// Additional properties allowed
+		Overrides json.RawMessage `json:"overrides,omitempty"`
+
+		// Possible values:
```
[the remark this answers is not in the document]
lotas: I've updated the descriptions with extra information

changelog/HzTmgpCWRVO5g5-UW1-OUA.md
```diff
@@ -0,0 +1,7 @@
+audience: users
+level: patch
+---
+
+Introduced `github.renderTaskclusterYml` endpoint to render provided `.taskcluster.yml` file for various events.
+This might be used for debug purposes or to validate the taskcluster.yml file
```
petemoore, line 6:
```suggestion
This might be used for debug purposes or to validate the .taskcluster.yml file
```

petemoore (approved):
Looks good to me, many thanks Yarik!

services/github/schemas/v1/render-taskcluster-yml-input.yml
```diff
@@ -0,0 +1,128 @@
...
+      This can be used to set some specific properties of the event or override the existing ones.
+      For example:
+        "ref": "refs/heads/main"
+        "before": "000"
+        "after": "111"
+      To make sure which properties are available for each event type,
+      please refer to the github [documentation](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads)
+    additionalProperties: true
+
+type:         object
```
petemoore, line 24: lot of space here, is that intentional?

ui/src/views/TcYamlDebug/index.jsx
```diff
@@ -35,16 +35,65 @@ const prefetchSchema = async () => {
...
+  'dequeued',
+  'edited',
+  'enqueued',
+  'labeled',
+  'ready_for_review',
+  'review_requested',
+  'review_request_removed',
+  'unassigned',
+  'unlabeled',
+];
```
petemoore, line 83: does this belong in the ui / frontend code? it looks closely tied to github service logic

petemoore (dismissed):

ui/src/views/TcYamlDebug/index.jsx
lotas: I guess this applies to the whole of UI - it is closely tied to everything already heh :) It also knows about github service and queue service.
Also I didn't see that there are "dictionary endpoints" that return some constants/lists. Do you think we should get one?

PR #6313, opened 2023-06-06
Enable CORS on api endpoints

lotas:
This allows to run UI locally on different port and domain and still let it talk to API services running under docker-compose.


petemoore (approved):
I guess this only affects local deployments, so this seems fine to me, thanks!

matt-boris (approved):
Thanks Yarik!

PR #6331, opened 2023-06-14
feat(generic-worker): add loopbackVideo support

### LINKED ISSUE #5994: Support loopback video in generic-worker

matt-boris:
Fixes #5994.

>Generic Worker: Adds `task.payload.feature.loopbackVideo` for loopback video device support on Linux.
>
>The `v4l2loopback` kernel module must be installed on the host system for this feature to work, although it does not _need_ to be loaded. Generic Worker loads the module with `modprobe` and generates the virtual video device with a `v4l2loopback` command. Under the multiuser engine, it also manages file ownership of the device with `chown` to ensure that only tasks with suitable scopes have read/write access to the virtual device.
>
>For tasks that enable the feature, the virtual video device location will be provided to the task commands via the environment variable `TASKCLUSTER_VIDEO_DEVICE`. The value of the environment variable depends on deployment configuration, and therefore tasks should not assume a fixed value. Its value will however take the form `/dev/video<DEVICE_NUMBER>` where `<DEVICE_NUMBER>` is an integer between 0 and 255. The Generic Worker config setting `loopbackVideoDeviceNumber` may be used to change the device number. Future releases of Generic Worker may provide the capability of having more than one virtual video device; currently only one virtual video device is supported.


changelog/issue-5994.md
```diff
@@ -0,0 +1,9 @@
+audience: users
+level: minor
+reference: issue 5994
+---
+Adds `task.payload.feature.loopbackVideo` for loopback video device support.
```
petemoore, line 5:
```suggestion
Generic Worker: Adds `task.payload.feature.loopbackVideo` for loopback video device support on Linux multiuser engine.
```

workers/generic-worker/loopback_video.go
```diff
@@ -0,0 +1,85 @@
+//go:build darwin || linux
+
+// TODO: are these right? ^^
+// I was having issues with them only being set to linux
```
petemoore, line 4: I think you can rename it to loopback_video_linux.go to just be for Linux.

workers/generic-worker/loopback_video.go
```diff
@@ -0,0 +1,85 @@
...
+		{"generic-worker:capability:device:loopbackVideo", fmt.Sprintf("generic-worker:capability:device:loopbackVideo:%s", lvt.task.Definition.TaskQueueID)},
+	}
+}
+
+func (lvt *LoopbackVideoTask) ReservedArtifacts() []string {
+	return []string{}
+}
+
+func (lvt *LoopbackVideoTask) Start() *CommandExecutionError {
+	err := host.Run("/bin/bash", "-c", "/usr/bin/sudo modprobe v4l2loopback video_nr=0")
```
petemoore, line 62: You don't need to invoke a shell here, as the arguments are static and e.g. there is no parsing of environment variables. `sudo` can also be avoided as generic-worker is already running as root in multiuser engine. So `host.Run("/usr/sbin/modprobe", "v4l2loopback", "video_nr=0")` would be more efficient.

workers/generic-worker/loopback_video.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return executionError(internalError, errored, fmt.Errorf("Could not load the v4l2loopback kernel module: %v", err))
+	}
+
+	err = host.Run("/bin/bash", "-c", "/usr/bin/sudo chmod 660 /dev/video0")
```
petemoore, line 67: Same here

workers/generic-worker/loopback_video.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return executionError(internalError, errored, fmt.Errorf("Could not chmod 660 the /dev/video0 device: %v", err))
+	}
+
+	err = makeFileOrDirReadWritableForUser(false, "/dev/video0", taskContext.User)
```
petemoore, line 72: Not sure if taskContext is initialised at this point. Hopefully! I just can't remember. My own fault - having taskContext as a global variable was a pretty ugly design decision of mine.

workers/generic-worker/loopback_video.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return executionError(internalError, errored, fmt.Errorf("Could make the /dev/video0 device readwritable for task user: %v", err))
+	}
+
+	return nil
+}
+
+func (lvt *LoopbackVideoTask) Stop(err *ExecutionErrors) {
+	removeModuleErr := host.Run("/bin/bash", "-c", "/usr/bin/sudo modprobe -r v4l2loopback")
```
petemoore, line 81: Can simplify here too.

workers/generic-worker/multiuser_posix.go
```diff
@@ -25,6 +25,7 @@ func (task *TaskRun) formatCommand(index int) string {
 func platformFeatures() []Feature {
 	return []Feature{
 		&InteractiveFeature{},
+		&LoopbackVideoFeature{},
```
petemoore, line 28: We probably need to split platform features up into linux and darwin, since now they diverge. Does interactive work on mac? If not, we should remove it. Could be fun to try it out - maybe it is close enough to linux that it might work?

Alternative: we could see what is involved in getting a virtual video device running on macOS. Maybe there is something similar to v4l2loopback for mac? If so, would be nice to support both platforms.

workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -6,8 +6,8 @@ description: |-
   Taskcluster Task definition.
 type: object
 required:
-- command
-- maxRunTime
+  - command
+  - maxRunTime
```
petemoore, line 10: I suspect this doc was already formatted, but perhaps using a different indentation strategy (non-indented list items).

We can switch, but we should probably decide on a standard so that we don't switch back and forth. Perhaps we can do that in a different PR to reduce the noise here, and add a linting setting at the same time to keep the same formatting standard?

workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -247,6 +247,15 @@ properties:
           is one.
 
           Since: generic-worker 49.2.0
+      loopbackVideo:
+        type: boolean
+        title: Loopback Video device
+        description: |-
+          Video loopback device created using v4l2loopback.
+          A video device will be available to the task user
+          at `/dev/video0`.
```
petemoore, line 256: This is ok for a first pass, but i think eventually we'll need to make it configurable on the worker. Eventually we will want to support capacity > 1 on generic-worker workers, and then they will need to use unique paths.

workers/generic-worker/loopback_video.go
matt-boris: Looks like it's set beforehand :) https://community-tc.services.mozilla.com/tasks/CmLsFsBDTzeCVgaDX4EasA/runs/0/logs/public/logs/live.log#L191

workers/generic-worker/schemas/multiuser_posix.yml
matt-boris: I can switch it to be a worker configuration that defaults to `/dev/video0` for now. We'll have to keep this in mind though as we move forward with the capacity > 1 on generic worker workers.

workers/generic-worker/multiuser_posix.go
matt-boris: I've decided to _not_ split it up and instead just return a malformed payload in the case where a user tries to run this on a a mac.

workers/generic-worker/loopback_video_posix.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return executionError(internalError, errored, fmt.Errorf("Could not chmod 660 the %s device: %v", devicePath, err))
+	}
+
+	err = makeFileOrDirReadWritableForUser(false, devicePath, taskContext.User)
+	if err != nil {
+		return executionError(internalError, errored, fmt.Errorf("Could make the %s device readwritable for task user: %v", devicePath, err))
+	}
+
+	err = os.Setenv("TASKCLUSTER_VIDEO_DEVICE", devicePath)
```
petemoore, line 75:
```suggestion
	err = lvt.task.setVariable("TASKCLUSTER_VIDEO_DEVICE", devicePath)
```

Your version would set the env var in the current worker process, but you need to set the env var in the task process(es) that are created later for each task command.

When you add a test, it would be great for the test to echo the env var, and then to check the log that it output it (so we can be sure this works).

workers/generic-worker/loopback_video_posix.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return executionError(internalError, errored, fmt.Errorf("Could not set TASKCLUSTER_VIDEO_DEVICE environment variable: %v", err))
+	}
+
+	lvt.task.Infof("Loopback video device is available at %s", devicePath)
```
petemoore, line 80: nice

workers/generic-worker/loopback_video_posix.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return nil
+}
+
+func (lvt *LoopbackVideoTask) Stop(err *ExecutionErrors) {
+	removeModuleErr := host.Run("/usr/sbin/modprobe", "-r", "v4l2loopback")
```
petemoore, line 86: I think we only need to ensure that the kernel module is loaded in order for the tasks to run, but we don't need to ensure it is unloaded when we are not using it, since generic-worker doesn't know what other processes might need it (e.g. if there are several generic-worker processes running, other ones may need it for their tasks).

workers/generic-worker/loopback_video_posix.go
```diff
@@ -0,0 +1,90 @@
...
+	if removeModuleErr != nil {
+		err.add(executionError(internalError, errored, fmt.Errorf("Could not remove the v4l2loopback kernel module: %v", removeModuleErr)))
+	}
```
petemoore, line 89: I think you should chown it its original state here (`root:root`?) I can't remember what the default setting is - but we should check.

workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -247,6 +247,21 @@ properties:
...
+          A video device will be available to the task user
+          at `/dev/video0` by default (device number can be
+          configured using the `loopbackVideoDeviceNumber`
+          field within the worker config) and an environment
+          variable `TASKCLUSTER_VIDEO_DEVICE` will be created
+          for the task user to use.
+          Ex. `loopbackVideoDeviceNumber: 1` will create
+          `/dev/video1` for the task user.
+
+          Since: generic-worker 53.1.0
```
petemoore, line 264: This needs to say that this is only for linux, and be explicit that any task submitted for macOS or freebsd with this feature enabled will resolve as `exception/malformed-payload`. In fact, we should have an explicit test for that for both freebsd and macOS, since that is easy to test.

workers/generic-worker/schemas/simple_posix.yml
```diff
@@ -225,6 +225,21 @@ properties:
...
+          A video device will be available to the task user
+          at `/dev/video0` by default (device number can be
+          configured using the `loopbackVideoDeviceNumber`
+          field within the worker config) and an environment
+          variable `TASKCLUSTER_VIDEO_DEVICE` will be created
+          for the task user to use.
+          Ex. `loopbackVideoDeviceNumber: 1` will create
+          `/dev/video1` for the task user.
+
+          Since: generic-worker 53.1.0
```
petemoore, line 242: Here too

workers/generic-worker/loopback_video_posix.go
```diff
@@ -0,0 +1,90 @@
...
+	}}
+}
+
+func (lvt *LoopbackVideoTask) ReservedArtifacts() []string {
+	return []string{}
+}
+
+func (lvt *LoopbackVideoTask) Start() *CommandExecutionError {
+	if runtime.GOOS == "darwin" {
+		return MalformedPayloadError(errors.New("Loopback video is not supported on darwin"))
```
petemoore, line 56: Thanks for adding this.

In general I think in taskcluster we refer to the platform as macOS, so I think "Loopback video is not supported on macOS" would be more consistent with our other messages (although I haven't double checked all of them). e.g. we provide setup instructions for macOS.

This feature also won't work with freebsd I guess (I think v4l2loopback is a linux only thing). If we added support, the implementation would be different for linux/freebsd/macOS so I think I would prefer it if you rename this file to loopback_video_linux.go and create a stub for the other platforms that just return a MalformedPayloadError and have nothing else. Then, if we later add support, it will go in its own file, and if we don't, it will be clear to anyone looking at the code that it is linux only (they won't have to _notice_ that there is an if statement on line 55 - because when they go to open the code, they will see immediately there is a linux version, a freebsd and a macOS version).

petemoore (commented):
Getting closer!

workers/generic-worker/loopback_video_posix.go
matt-boris: Perfect, thanks Pete!

workers/generic-worker/loopback_video_posix.go
matt-boris: It was `root:video`, would `root:root` be sufficient in our case though?

lotas (approved):
lgtm

matt-boris:
Example task with it enabled where you can see the `/dev/video0` device and also see the environment variable being written out as well https://community-tc.services.mozilla.com/tasks/W5mxcix3Q9Oas_dw6DddSA/runs/0/logs/public/logs/live.log

Example task with it disabled where the `/dev/video0` device is not available https://community-tc.services.mozilla.com/tasks/dnWv1ZrST3CO1CfYGNNFrw/runs/0/logs/public/logs/live.log

changelog/issue-5994.md
```diff
@@ -0,0 +1,9 @@
+audience: users
+level: minor
+reference: issue 5994
+---
+Generic Worker: Adds `task.payload.feature.loopbackVideo` for loopback video device support on Linux.
+
+The `v4l2loopback` kernel module is required on the host system for this feature to work, as generic worker will load this module and create the video device for the worker to use.
+
+The resulting video device will be located at `/dev/video0` by default (device number can be configured using the `loopbackVideoDeviceNumber` field within the worker config) and an environment variable `TASKCLUSTER_VIDEO_DEVICE` will be created for the task user to use. Ex. `loopbackVideoDeviceNumber: 1` will create `/dev/video1` for the task user.
```
petemoore, line 9:
```suggestion
The `v4l2loopback` kernel module must be installed on the host system for this feature to work, although it does not _need_ to be loaded. Generic Worker loads the module with `modprobe` and generates the virtual video device with a `v4l2loopback` command. Under the multiuser engine, it also manages file ownership of the device with `chown` to ensure that only tasks with suitable scopes have read/write access to the virtual device.

For tasks that enable the feature, the virtual video device location will be provided to the task commands via the environment variable `TASKCLUSTER_VIDEO_DEVICE`. The value of the environment variable depends on deployment configuration, and therefore tasks should not assume a fixed value. Its value will however take the form `/dev/task<DEVICE_NUMBER>` where `<DEVICE_NUMBER>` is an integer between 0 and 255. The Generic Worker config setting `loopbackVideoDeviceNumber ` may be used to change the device number. Future releases of Generic Worker may provide the capability of having more than one virtual video device; currently only one virtual video device is supported.
```

workers/generic-worker/loopback_video_darwin_test.go
```diff
@@ -0,0 +1,26 @@
...
+	payload := GenericWorkerPayload{
+		Command:    helloGoodbye(),
+		MaxRunTime: 30,
+		Features: FeatureFlags{
+			LoopbackVideo: true,
+		},
+	}
+	defaults.SetDefaults(&payload)
+	td := testTask(t)
+	td.Scopes = append(td.Scopes, "generic-worker:loopback-video")
```
petemoore, line 23: could add a comment that is malformed-payload because the feature isn't implemented on macOS?

workers/generic-worker/loopback_video_freebsd_test.go
```diff
@@ -0,0 +1,26 @@
...
+		MaxRunTime: 30,
+		Features: FeatureFlags{
+			LoopbackVideo: true,
+		},
+	}
+	defaults.SetDefaults(&payload)
+	td := testTask(t)
+	td.Scopes = append(td.Scopes, "generic-worker:loopback-video:freebsd/loopback-video")
+
+	_ = submitAndAssert(t, td, payload, "exception", "malformed-payload")
```
petemoore, line 25: here too an explanation that the feature isn't implemented on freebsd

workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -247,6 +247,26 @@ properties:
...
+        description: |-
+          Video loopback device created using v4l2loopback.
+          A video device will be available to the task user
+          at `/dev/video0` by default (device number can be
+          configured using the `loopbackVideoDeviceNumber`
+          field within the worker config) and an environment
+          variable `TASKCLUSTER_VIDEO_DEVICE` will be created
+          for the task user to use.
+          Ex. `loopbackVideoDeviceNumber: 1` will create
+          `/dev/video1` for the task user.
```
petemoore, line 262:
```suggestion
          A video device will be available for the task. Its
          location will be passed to the task via environment
          variable `TASKCLUSTER_VIDEO_DEVICE`. The
          location will be `/dev/video<N>` where `<N>` is
          an integer between 0 and 255. The value of `<N>`
          is not static, and therefore either the environment
          variable should be used, or `/dev` should be
          scanned in order to determine the correct location.
          Tasks should not assume a constant value.
```

workers/generic-worker/loopback_video_linux.go
```diff
@@ -0,0 +1,46 @@
...
+
+package main
+
+import (
+	"fmt"
+
+	"github.com/taskcluster/taskcluster/v53/workers/generic-worker/host"
+)
+
+func setupVideoDevice(task *TaskRun) *CommandExecutionError {
```
petemoore, line 11: I feel like this should be:

```go
func (lvt *LoopbackVideoTask) setupVideoDevice() *CommandExecutionError {
```

workers/generic-worker/loopback_video_linux.go
```diff
@@ -0,0 +1,46 @@
...
+	err := host.Run("/usr/sbin/modprobe", "v4l2loopback", fmt.Sprintf("video_nr=%d", config.LoopbackVideoDeviceNumber))
+	if err != nil {
+		return executionError(internalError, errored, fmt.Errorf("Could not load the v4l2loopback kernel module: %v", err))
+	}
+
+	devicePath := fmt.Sprintf("/dev/video%d", config.LoopbackVideoDeviceNumber)
```
petemoore, line 17: I feel like devicePath should be a struct member of LoopbackVideoTask

workers/generic-worker/loopback_video_linux.go
```diff
@@ -0,0 +1,46 @@
...
+		return executionError(internalError, errored, fmt.Errorf("Could not set TASKCLUSTER_VIDEO_DEVICE environment variable: %v", err))
+	}
+
+	task.Infof("Loopback video device is available at %s", devicePath)
+
+	return nil
+}
+
+func resetVideoDevice() *CommandExecutionError {
+	devicePath := fmt.Sprintf("/dev/video%d", config.LoopbackVideoDeviceNumber)
```
petemoore, line 39: If this was a struct member of LoopbackVideoTask you would not have to calculate it here again...

petemoore (commented):
Thanks Boris!

workers/generic-worker/loopback_video_linux_test.go
```diff
@@ -0,0 +1,61 @@
...
+	_ = submitAndAssert(t, td, payload, "completed", "completed")
+
+	logText := LogText(t)
+	if !strings.Contains(logText, "/dev/video0") {
+		t.Fatalf("Expected log to contain /dev/video0, but it didn't\n%s", logText)
+	}
+	if !strings.Contains(logText, "crw-rw----") {
+		t.Fatalf("Expected log to contain crw-rw----, but it didn't\n%s", logText)
+	}
+}
```
petemoore, line 37: This looks like a copy of the test below, but with a fixed value for the location. I don't think this test tests anything that the test below doesn't already test? Maybe we can delete it.

workers/generic-worker/loopback_video_linux_test.go
```diff
@@ -0,0 +1,61 @@
...
+	defaults.SetDefaults(&payload)
+	td := testTask(t)
+	td.Scopes = append(td.Scopes, "generic-worker:loopback-video")
+
+	_ = submitAndAssert(t, td, payload, "completed", "completed")
+
+	logText := LogText(t)
+	if !strings.Contains(logText, "/dev/video0") {
+		t.Fatalf("Expected log to contain /dev/video0, but it didn't\n%s", logText)
+	}
```
petemoore, line 60: I think this should be a regexp, i.e. `/dev/video[0-9]+` (or however you do digits in go regexp)

workers/generic-worker/loopback_video_linux.go
matt-boris: Good idea, thanks.

workers/generic-worker/loopback_video_linux_test.go
matt-boris: This test should always be device 0, as it's hardcoded in the generic worker test config.

workers/generic-worker/loopback_video_linux_test.go
matt-boris: So what I was trying to do by separating between different test cases was differentiate between the `strings.Contains(logText, "/dev/video0")` calls between the `ls -l /dev/video0` and `echo $TASKCLUSTER_VIDEO_DEVICE`. If one of those commands logged out `/dev/video0`, then the test would pass, and I wouldn't have successfully singled out which part actually was passing, the video device existing, or the environment variable existing.

petemoore: Ah that makes sense. Thanks. I think you can do this if when you echo it you have text, like `echo "Device: ${TASKCLUSTER_VIDEO_DEVICE}"` and then check for text `Device: /dev/video0`. I think `ls -l` will fail if the device does not exist, so then it should not be completed/completed, and I think test will fail already.
The test setup might set the device number to zero, but that lives in another file etc, that could get out of sync. Would it be worth adding a comment here that we can hardcode 0 because it is hardcoded in the setup method? Or maybe read it directly from the config here, so if it is changed in the setup, it is changed here too? (something like `fmt.Sprintf("/dev/video%v", config.LoopbackVideoDevice)`). Just want to make sure we don't have the zero hardcoded in multiple places, and if one changes, the other breaks. But e.g. a comment to say it has to match test setup is also ok, I think I just prefer reading the variable so the information isn't duplicated.

workers/generic-worker/loopback_video_linux_test.go
petemoore: Same here, maybe we can read it from config value, so that it works even if test setup is changed to use a different device number. For example, I suspect we can't always assume /dev/video0 is free, if someone is running the tests on a linux development machine that has a webcam, maybe /dev/video0 is already assigned to their real webcam. So I feel like at some point, we should probably change the test setup to use the first available free device, but since none of us are using linux dev machines we probably haven't hit that issue yet. In general, we can't assume /dev/video0 is available.

petemoore (approved):
Thanks Boris!

PR #6333, opened 2023-06-16
Stop instances that are not doing anything useful anymore

### LINKED ISSUE #6142: Worker Manager not killing rogue/zombie instances

lotas:
This can happen when docker/generic worker process dies and stops calling `queue.claimWork`/`queue.reclaimTask`
Such workers would either be not known to the queue, or would have their last_date_active way in the past.
In such cases killing instance is the only option to free up resources.

This PR introduces `queueInactivityTimeout` for the worker `lifecycle` parameter (default being `2h`). 

It works in following scenarios: 

* if worker is registered, but didn't call `queue.claimWork` it would be terminated after `queueInactivityTimeout` (`firstClaim` is not set)
* if worker claimed work, but didn't call `queue.reclaimTask`, or died and stopped calling `claimWork`, it would be terminated after `2 hours` (`lastDateActive` is older than `queueInactivityTimeout`)
   

Quarantined workers are not affected, since they keep calling `claimWork` but are being ignored there.

Fixes #6142


services/worker-manager/src/providers/provider.js
```diff
@@ -84,6 +84,46 @@ class Provider {
...
+   * If worker fails to call queue.claimWork, `queue_worker.first_claim` would be set to null.
+   * If worker does not call queue.reclaimTask or stops calling queue.claimWork,
+   * `queue_worker.last_date_active` would not be updated.
+   *
+   * Workers that are registered, but don't have `first_claim`
+   * or `last_date_active` is older than inactivity timeout are considered to be zombies.
+   */
+  static isZombie({ worker, activityTimeout, createdTimeout }) {
+    activityTimeout = activityTimeout || 1000 * 60 * 60 * 2; // last active within 2 hours
+    createdTimeout = createdTimeout || 1000 * 60 * 60 * 1; // created at least 1 hour ago
```
djmitche, line 103: These should probably be values specified in the `lifecycle` property of the worker pool, rather than hard-coded. As mentioned in the meeting today, they can probably be much smaller for AWS and Google.

djmitche (commented):
I mentioned in the meeting that changing the existing `lifetime` parameters to something a lot smaller than 96h would have a similar effect -- it will terminate workers that have not (re)registered in that timeframe.

This PR would terminate workers that were successfully re-registering, but were not claiming work. Typically worker-runner handles the worker registration, while of course the worker handles the claiming. The cases where I have seen a worker re-registering but not claiming are when docker-worker is confused into thinking that it has zero capacity (such as when it starts without enough disk space). So I think this is a valid situation for worker-manager to monitor for.

services/worker-manager/src/providers/provider.js
lotas: Thanks Dustin,  yes, I moved them into the `lifecycle` and left a single `claimTimeout` parameter

lotas:
> I mentioned in the meeting that changing the existing lifetime parameters to something a lot smaller than 96h would have a similar effect -- it will terminate workers that have not (re)registered in that timeframe.

Do you think we should lower that default here now as well? Do you remember why `96h` was chosen initially?

I think it still makes sense to have `claimTimeout` as it "reacts faster" to the situations where queue interaction is broken for either reason.

services/worker-manager/src/providers/aws.js
```diff
@@ -359,6 +360,10 @@ class AwsProvider extends Provider {
           await this.removeWorker({ worker, reason: 'terminateAfter time exceeded' });
         }
       }
+      const { isZombie, reason } = Provider.isZombie({ worker });
```
lotas, line 363: naming is hard .. but  I think `zombie` is a good analogy, same as zombie-processes

djmitche:
> > I mentioned in the meeting that changing the existing lifetime parameters to something a lot smaller than 96h would have a similar effect -- it will terminate workers that have not (re)registered in that timeframe.
> 
> Do you think we should lower that default here now as well? Do you remember why `96h` was chosen initially?
> 
> I think it still makes sense to have `claimTimeout` as it "reacts faster" to the situations where queue interaction is broken for either reason.

The 96h was originally hard-coded in the workers. The `aws-provisioner-v1` service would call TerminateInstances after 96 hours as a backup, so workers terminated themselves before that.

We set the lifecycle defaults to match that so that enabling them didn't change behavior, with the idea that someone would followup and reduce those timeouts :)

db/fns.md
```diff
@@ -2667,15 +2667,19 @@ Returns a count of rows that have been deleted.
...
 
 Get non-stopped workers filtered by the optional arguments,
 ordered by `worker_pool_id`, `worker_group`, and  `worker_id`.
 If the pagination arguments are both NULL, all rows are returned.
 Otherwise, page_size rows are returned at offset `page_offset`.
 The `quaratine_until` contains NULL or a date in the past if the
 worker is not quarantined, otherwise the date until which it is
-quaratined. `providers_filter_cond` and `providers_filter_value` used to
+quaratined. `first_claim` and `last_date_active` contains information
+known to the queue service about the worker.
```
petemoore, line 2681: How is last date active defined? Last time any API request was made to the queue, or only a specific subset?

services/worker-manager/schemas/v1/worker-lifecycle.yml
```diff
@@ -30,5 +30,18 @@ properties:
       will be terminated.  This value also dictates the lifetime of the
       temporary credentials granted to the worker, meaning that it must be
       less than 30 days. The default value is 4 days.
+  claimTimeout:
```
petemoore, line 33: Claim timeout sounds like something different - it sounds like how long the claim is, until reclaimTask needs to be called. Normally that is 20 minutes.

Perhaps notClaimingTimeout / betweenClaimsTimeout / offlineTimeout / awolTimeout / queueSilenceTimeout ?

services/worker-manager/src/providers/aws.js
petemoore: I think it is ok too, but we should probably document somewhere what zombie means.

services/worker-manager/src/providers/provider.js
```diff
@@ -84,6 +84,46 @@ class Provider {
...
+
+    let reason = null;
+    let isZombie = false;
+
+    if (!worker.firstClaim && worker.created < lastActiveAfter) {
+      isZombie = true;
+      reason = `worker never claimed work, created=${worker.created}`;
+    }
+
+    if (!worker.lastDateActive && worker.firstClaim < lastActiveAfter) {
```
petemoore, line 114: what sets lastDateActive? i'm keen to know what active means in this context.

ui/docs/reference/core/worker-manager/worker-interaction.mdx
```diff
@@ -58,6 +58,19 @@ The expiration time of the new credentials comes from the worker pool configurat
 which gets set when calling [`workerManager.registerWorker`](/docs/reference/core/worker-manager/api#registerWorker).
 The default behavior, when `reregistrationTimeout` is omitted, is to extend the credentials by 96 hours.
 
+## Claim Work Timeout
+
+To make sure that spawned workers are not stuck in a "zombie" state, the worker-manager will check with queue service
```
petemoore, line 63: This would be a great place to explain what a zombie state is.

petemoore (commented):
Thanks Yarik!

I like that you used the term "non-stopped-workers" to mean all workers not in stopped state, since other terms like "active" would be more vague. So +1 for that! I think claimTimeout though is a bit confusing, as when workers claim a task, they get a claimTimeout already, and it means something different (how long until if they haven't reclaimed the task will be resolved as exception/claim-expired). Other than that, some clarity over lastActiveDate is useful, although maybe that already exists elsewhere in documentation. Thanks!

db/fns.md
matt-boris: This looks like he's just using what I had added previously https://github.com/taskcluster/taskcluster/pull/5365/files#diff-3dcdf963e193e0ce3968b21e9a477bf3cb166b7480385482d217b5d9773f3116R60-R66

>Updated each time a worker calls `queue.claimWork`, `queue.reclaimTask`, and `queue.declareWorker` for this task queue.

lotas: Yes, this is handled by queue, whenever `claimWork` or `reclaimTask` calls happens, queue marks workers this way, to know that they were doing something at this period of time

services/worker-manager/schemas/v1/worker-lifecycle.yml
lotas: thanks for the feedback! I'm open to changes here to get more clarity.

`queueInactivityTimeout` ?

services/worker-manager/src/providers/provider.js
lotas: this is something queue service keeps track of for each worker. Whenever worker calls queue methods, those values are being updated. 

Same way we know that worker never called queue service - those values would be missing.

ui/docs/reference/core/worker-manager/worker-interaction.mdx
lotas: great! as soon as we come up with final parameter name for this one :)

services/worker-manager/schemas/v1/worker-lifecycle.yml
```diff
@@ -30,5 +30,21 @@ properties:
...
+    type: integer
+    minimum: 600 # 10 minutes
+    default: 7200 # 2 hours
+    description: |
+      In order to prevent workers from being stuck without doing any work,
+      `claimTimeout` controls how long a worker can do something other than
+      working on a task. If worker process dies, or it stops calling `claimWork`
+      or `reclaimTask` it should be considered dead and terminated.
+
+      Minimum allowed value is 10 minutes, to give worker chance to start claiming tasks.
```
matt-boris, line 44: This worries me a _little_ bit. Did you get this 10 min from anywhere or just _hoping_ that this'll be the case? I prefer to go a little more conservative here. Wdyt?

ui/docs/reference/core/worker-manager/worker-interaction.mdx
```diff
@@ -58,6 +58,19 @@ The expiration time of the new credentials comes from the worker pool configurat
...
+to see if there was any recent activity from a given worker.
+
+Unfortunately, it is not always possible to handle critical errors gracefully on a worker itself. For example,
+if worker process is being terminated with out of memory error, it is not possible to catch it and report it to the worker-manager.
```
matt-boris, line 67:
```suggestion
if worker process is being terminated with an out of memory error, it is not possible to catch it and report it to the worker-manager.
```

ui/docs/reference/core/worker-manager/worker-interaction.mdx
```diff
@@ -58,6 +58,19 @@ The expiration time of the new credentials comes from the worker pool configurat
...
+Also, when image is not configured properly, worker might be able to start and register, but would fail to call `queue.claimWork`.
```
matt-boris, line 68:
```suggestion
Also, when an image is not configured properly, the worker might be able to start and register, but would fail to call `queue.claimWork`.
```

matt-boris (commented):
Nice, looking good! 🎉

services/worker-manager/schemas/v1/worker-lifecycle.yml
```diff
@@ -30,5 +30,21 @@ properties:
       will be terminated.  This value also dictates the lifetime of the
       temporary credentials granted to the worker, meaning that it must be
       less than 30 days. The default value is 4 days.
+  queueInactivityTimeout:
+    title: Claim Timeout
```
matt-boris, line 34:
```suggestion
    title: Queue Inactivity Timeout
```

services/worker-manager/test/provider_test.js
```diff
@@ -25,46 +26,117 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
-  test('greater reregistrationTimeout', async function() {
-    assert.deepEqual({
-      terminateAfter: 10100,
-      reregistrationTimeout: 100000,
-    }, Provider.interpretLifecycle({ lifecycle: {
-      registrationTimeout: 10,
-      reregistrationTimeout: 100,
-    } }));
+  suite('isZombie', function() {
+    test('default claim timeout', function() {
```
matt-boris, line 88:
```suggestion
    test('default queue inactivity timeout', function() {
```

matt-boris (commented):
Some small, additional cleanup.

services/worker-manager/schemas/v1/worker-lifecycle.yml
lotas: renamed it

ui/docs/reference/core/worker-manager/worker-interaction.mdx
lotas: done

services/worker-manager/schemas/v1/worker-lifecycle.yml
lotas: yeah, this is totally debatable. I just wanted something to prevent misconfiguration lead to workers being killed before they even got a chance to do something 
I can imagine it takes time to start up and prepare. Also, this would include reboots. On windows it probably takes longer.

@matt-boris do you suggest we increase this number or decrease? 

@petemoore from your past experience, do you know what should be average max time between tasks? GC, reboot, etc

matt-boris: > @matt-boris do you suggest we increase this number or decrease?

I would think 30 is a safer bet. This is quite a bit more conservative, which might be overkill, but it's hard to tell for sure.

lotas: I've been running some [queries](https://sql.telemetry.mozilla.org/queries/92851/source) on the data we have, and noticed that reboots might take 5-15min on average, with some outliers with up to 60+ minutes
Maybe those pools are special, or maybe just the data is not very representative. 

Data source is an aggregation already - total number of seconds spent on reboots and number of instances.. So I'm drawing my conclusions from reboot/instances to see what those numbers are 

Now I'm also wondering if 2h is a good default 🤔

lotas: ok, if I only query for the pools with 20+ instances, then .75%% gives all reboots within 2h, so I think it should be safe

services/worker-manager/schemas/v1/worker-lifecycle.yml
lotas: 🤦 thank you Matt

matt-boris (dismissed):
LGTM thanks Yarik!! 🚀

services/worker-manager/schemas/v1/worker-lifecycle.yml
djmitche: I think there's a difference between a minimum value in the platform and the configured value. If for example we had a provisioner that ran QEMU VMs on demand, 10 minutes might be _far_ too long.  It's probably short for a default value, but not for a _minimum_ value.

It's worth wondering if a host that takes 2+ hours to reboot is going to ever manage to run a build or test task in a reasonable amount of time. But, that decision should be up to the deployers (and thus not encoded in the schema) IMHO.

lotas: Agree, I'll put something smaller.

PR #6336, opened 2023-06-22
feat(tools): migrate `d2g` into monorepo

matt-boris:
Fixes https://github.com/taskcluster/d2g/issues/11.


petemoore (approved):
Thanks so much Boris!

Can you also archive https://github.com/taskcluster/d2g and delete all the files except the README saying that it has moved here (with a link to the new home)?

PR #6341, opened 2023-06-23
chore: additional cleanup from PR #6339

### LINKED ISSUE #6340: Higher priority decision tasks, or run in dedicated pool etc
Decision tasks block other tasks, yet they are scheduled with same priority as regular tasks. This means that sometimes entire task graphs are blocked by competing tasks of the same priority on the same worker pool, e.g. a build or test task could be delaying a decision task for a different push. I noticed this because we have a Generic Worker decision task which first has to be created by the mono repo decision task, and so it can take a long time before e.g. a Generic Worker build/test task is even generated, let alone run. We should solve that by moving the Generic Worker decision task into the main decision task, but it still highlights an issue. Note, this doesn't have to be solved with using a different priority, another option is having dedicated worker pool for decision tasks, which scales quickly, and has generous max capacity. Another option is fewer worker pools, and adjusted priority. This is probably more of a discussion point right now than a hard requirement. We have a few knobs we can tweak, would be nice to think about the situation holistically, and work out a nice solution that balances concerns and optimises end-to-end times as much as possible within given cost constraints etc.

matt-boris:
Pardon the branch name.. I was initially thinking I'd be able to easily edit the task priority of the gw decision task. I'll tackle this in a different PR.

### DIFF
--- a/changelog/aRbeyY13Rfm9GQkNDh2fzw.md
+++ b/changelog/aRbeyY13Rfm9GQkNDh2fzw.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
--- a/taskcluster/src/transforms/__init__.py
+++ b/taskcluster/src/transforms/__init__.py
@@ -1,4 +1,3 @@
-import copy
 import os
 import json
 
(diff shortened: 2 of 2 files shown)


lotas (approved):
👍

PR #6342, opened 2023-06-23
chore: use generic worker for CI tasks

matt-boris:
48 tasks used to run on docker worker. These have all been migrated to run on generic worker.

Because of these changes, we'll be able to remove all of our docker-worker worker pools in Community-TC with https://github.com/mozilla/community-tc-config/pull/604.

Next up will be migrating the generic-worker specific decision task in with the main decision task.


infrastructure/docker-images/build-browser-test.sh
```diff
@@ -11,7 +11,7 @@ fi
 tmpdir=$(mktemp -d)
 trap "cd /; rm -rf ${tmpdir}" EXIT
 cat > ${tmpdir}/Dockerfile <<EOF
-FROM node:${node_version}-buster
+FROM node:${node_version}-bookworm
```
lotas, line 14: I wonder if we could do something similar here to the `{node_version}` to avoid doing `s/old/new/` in so many places? 
just a thought :)

lotas (approved):
this is awesome! ❤️

PR #6344, opened 2023-06-27
fix(generic-worker): set defaults before validating payload

### LINKED ISSUE #6237: Malformed Payload exception without task log
**Describe the bug**
Generic Worker is meant to provide a log when there is a malformed payload exception, to explain what was invalid in the payload. I've never noticed it _not_ doing this before, but today it happened in https://community-tc.services.mozilla.com/tasks/Hn755kwxSC6def6V3o9KIw#artifacts.

There should be a log file with the payload exception. Not sure how/when this broke. We should certainly have a unit test for it.

matt-boris:
Fixes #6237.

>This patch fixes the case where a generic worker won't upload its log on a malformed payload error. This has been broken since v48.2.0 from PR [#6107](https://github.com/taskcluster/taskcluster/pull/6107).

### DIFF
--- a/changelog/issue-6237.md
+++ b/changelog/issue-6237.md
@@ -1,0 +1,5 @@
+audience: general
+level: patch
+reference: issue 6237
+---
+Fix the case where a generic worker won't upload its log on a malformed payload error. This has been broken since v48.2.0 from PR [#6107](https://github.com/taskcluster/taskcluster/pull/6107).
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -584,6 +584,11 @@ func ClaimWork() *TaskRun {
 
 func (task *TaskRun) validatePayload() *CommandExecutionError {
 	jsonPayload := task.Definition.Payload
+	defaults.SetDefaults(&task.Payload)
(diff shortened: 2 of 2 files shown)


changelog/issue-6237.md
```diff
@@ -0,0 +1,5 @@
+audience: general
+level: patch
+reference: issue 6237
+---
+This patch fixes the case where a generic worker won't upload its log on a malformed payload error. This has been broken since v48.2.0 from PR [#6107](https://github.com/taskcluster/taskcluster/pull/6107).
```
lotas, line 5:
```suggestion
Fix the case where a generic worker won't upload its log on a malformed payload error. This has been broken since v48.2.0 from PR [#6107](https://github.com/taskcluster/taskcluster/pull/6107).
```

probably better to phrase it from the changelog reader point of view?

lotas (dismissed):
👏

changelog/issue-6237.md
lotas: Fix/Fixed not sure :)

lotas (approved):
🥇

Issue #2915, opened 2020-05-28
merge generic-worker decision task logic into taskgraph decision task logic


imbstack:
Right now taskgraph triggers a task that runs g-w decision task which then spawns tasks. Let's just cut out the middleman

petemoore:
Fixed in #6348 !

PR #6348, opened 2023-06-28
refactor: remove generic worker decision task

### LINKED ISSUE #6340: Higher priority decision tasks, or run in dedicated pool etc

matt-boris:
Fixes #6340.

This should also help our "First to last" stat on the task group page because all tasks are now kicked off by the decision task. Before, we would need to wait on an additional, generic worker decision task to run in order to create the build/test generic worker tasks (which all got created with low priority). Now we're a bit more asynchronous.

Stats before (from https://community-tc.services.mozilla.com/tasks/groups/CBSf0IvRSvKMuSyxCQH0Lg):
Screenshot 2023-06-28 at 7 37 24 PM

Stats after (from https://community-tc.services.mozilla.com/tasks/groups/H7zH2iIwTyGS0NPVoq369w):
Screenshot 2023-06-28 at 7 37 50 PM


lotas:
awesome! Here's also a related ticket that should be closed after: https://github.com/taskcluster/taskcluster/issues/2915

lotas (approved):
This is wonderful!

PR #6351, opened 2023-06-29
feat(cli): integrate `d2g` into taskcluster cli

matt-boris:
>This change adds the `d2g` subcommand to the `taskcluster` cli.
>
>It can be used to translate a Docker Worker payload to a Generic Worker payload.
>Both the input and output are JSON. You can either pass the input as a file or pipe it in to the command.
>
>View help with:
>
>```shell
>taskcluster d2g -h
>```
>
>Example usages:
>
>```shell
>taskcluster d2g -f /path/to/input.json
>```
>
>_OR_
>
>```shell
>taskcluster d2g --file /path/to/input.json
>```
>
>_OR_
>
>```shell
>cat /path/to/input.json | taskcluster d2g
>```
>
>_OR_
>
>```shell
>echo '{"image": "ubuntu", "command": ["bash", "-c", "echo hello world"], "maxRunTime": 300}' | taskcluster d2g
>```

### DIFF
--- a/clients/client-shell/cmds/d2g/d2g.go
+++ b/clients/client-shell/cmds/d2g/d2g.go
@@ -1,34 +1,60 @@
-package main
+package d2g
 
 import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log"
 	"os"
 	"strings"
 
 	"github.com/mcuadros/go-defaults"
+	"github.com/taskcluster/taskcluster/v54/clients/client-shell/cmds/root"
 	"github.com/taskcluster/taskcluster/v54/tools/d2g"
 	"github.com/taskcluster/taskcluster/v54/tools/d2g/dockerworker"
 	"github.com/taskcluster/taskcluster/v54/tools/d2g/genericworker"
 	"github.com/xeipuuv/gojsonschema"
+
+	"github.com/spf13/cobra"
 )
 
-func main() {
-	log.SetFlags(0)
-	log.SetPrefix("d2g: ")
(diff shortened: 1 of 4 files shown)


clients/client-shell/cmds/d2g/d2g.go
```diff
@@ -0,0 +1,124 @@
...
+	// Validate the JSON output against the schema
+	err = validateJSON(formattedActualGWPayload, genericworker.JSONSchema())
+	if err != nil {
+		return fmt.Errorf("output validation failed: %v", err)
+	}
+
+	fmt.Fprintln(cmd.OutOrStdout(), string(formattedActualGWPayload))
+
+	// Copy output to clipboard
+	err = clipboard.WriteAll(string(formattedActualGWPayload))
```
lotas: I'm wondering if this is something that taskcluster command is doing elsewhere, or was a requirement? 
It feels a bit weird, especially since you have tools like `pbcopy` or `xclip` you can easily pipe the output to clipboard

Maybe it would be better to just let it write to stdout? Also probably it is unexpected from a user perspective if you want to convert the payload, but have something in your clipboard it would be overwritten without user realizing this.
wdyt?

lotas (approved):
awesome! Thank you

clients/client-shell/cmds/d2g/d2g.go
matt-boris: Removed! Thanks for the suggestion :)

PR #6355, opened 2023-06-30
feat(generic-worker): integrate `d2g` to interpret docker worker payloads

### LINKED ISSUE #5967: Add capability to generic-worker to be able to directly execute docker-worker tasks via internal transformation of task payload to native generic-worker format using go library

matt-boris:
Fixes #5967.

>This change integrates the `d2g` tool into Generic Worker so that it can accept a valid, Docker Worker payload.

### DIFF
--- a/tools/d2g/genericworker/generated_types.go
+++ b/tools/d2g/genericworker/generated_types.go
@@ -4,6 +4,7 @@ package genericworker
 
 import (
 	"encoding/json"
+	"errors"
 
 	tcclient "github.com/taskcluster/taskcluster/v54/clients/client-go"
 )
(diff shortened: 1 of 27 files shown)


workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -1,369 +1,604 @@
...
+
+              Note, setting `contentEncoding` on a directory artifact will apply the same content
+              encoding to all the files contained in the directory.
+
+              Since: generic-worker 16.2.0
+        required:
+        - type
+        - path
+    features:
+      title: Feature flags
```
petemoore, line 192:
```suggestion
      title: Docker Worker feature flags
```

This will avoid the go type getting named `FeatureFlags1`.

tools/d2g/genericworker/generated_types.go
```diff
@@ -224,6 +353,53 @@ type (
 		TaskclusterProxy bool `json:"taskclusterProxy,omitempty"`
 	}
 
+	// Used to enable additional functionality.
+	FeatureFlags1 struct {
```
petemoore, line 357: This should get fixed when regenerating...

ui/docs/reference/workers/generic-worker/README.mdx
```diff
@@ -17,7 +17,7 @@ Depending on the operating system of the worker, and the generic-worker engine t
 Please see the following pages for guidance:
 
 <!-- BEGIN PAYLOAD LINKS -->
- * [Generic worker payload - multiuser, posix](/docs/reference/workers/generic-worker/multiuser-posix-payload)
+ * [Payload - multiuser, posix](/docs/reference/workers/generic-worker/multiuser-posix-payload)
```
petemoore, line 20: Is there any way we can fix this?

workers/generic-worker/d2g_test.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	payload := dockerworker.DockerWorkerPayload{
+		Command:    []string{"/bin/bash", "-c", "echo hello world"},
+		Image:      json.RawMessage(imageBytes),
+		MaxRunTime: 10,
+	}
+	defaults.SetDefaults(&payload)
+	td := testTask(t)
+
+	switch fmt.Sprintf("%s:%s", runtime.GOOS, engine) {
```
petemoore, line 31: Nice!

workers/generic-worker/d2g_test.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		t.Fatalf("Error marshaling JSON: %v", err)
+	}
+	payload := dockerworker.DockerWorkerPayload{
+		Command:    []string{"/bin/bash", "-c", "echo hello world"},
+		Image:      json.RawMessage(imageBytes),
+		MaxRunTime: 10,
+	}
+	defaults.SetDefaults(&payload)
+	td := testTask(t)
```
petemoore, line 29: where does the payload get set in the test? I see payload is set on line 23, but I don't see it used anywhere.

workers/generic-worker/d2g_test.go
```diff
@@ -0,0 +1,57 @@
...
+	imageBytes, err := json.Marshal(image)
+	if err != nil {
+		t.Fatalf("Error marshaling JSON: %v", err)
+	}
+	payload := dockerworker.DockerWorkerPayload{
+		Command: []string{"/bin/bash", "-c", "echo hello world"},
+		Image:   json.RawMessage(imageBytes),
+	}
+	defaults.SetDefaults(&payload)
+	td := testTask(t)
```
petemoore, line 54: Same here.

workers/generic-worker/helper_test.go
```diff
@@ -58,14 +59,14 @@ func testWorkerType() string {
 	return "test-" + strings.ToLower(strings.Replace(slugid.Nice(), "_", "", -1)) + "-a"
 }
 
-func scheduleTask(t *testing.T, td *tcqueue.TaskDefinitionRequest, payload GenericWorkerPayload) (taskID string) {
+func scheduleTask[P GenericWorkerPayload | dockerworker.DockerWorkerPayload](t *testing.T, td *tcqueue.TaskDefinitionRequest, payload P) (taskID string) {
```
petemoore, line 62: Could we use `DockerWorkerPayload` here instead of `dockerworker.DockerWorkerPayload`?

workers/generic-worker/helper_test.go
```diff
@@ -58,14 +59,14 @@ func testWorkerType() string {
...
 	t.Helper()
 	taskID = slugid.Nice()
 	scheduleNamedTask(t, td, payload, taskID)
 	return
 }
 
-func scheduleNamedTask(t *testing.T, td *tcqueue.TaskDefinitionRequest, payload GenericWorkerPayload, taskID string) {
+func scheduleNamedTask[P GenericWorkerPayload | dockerworker.DockerWorkerPayload](t *testing.T, td *tcqueue.TaskDefinitionRequest, payload P, taskID string) {
```
petemoore, line 69: Same here

workers/generic-worker/helper_test.go
```diff
@@ -199,7 +200,7 @@ func LogText(t *testing.T) string {
 	return string(bytes)
 }
 
-func submitAndAssert(t *testing.T, td *tcqueue.TaskDefinitionRequest, payload GenericWorkerPayload, state, reason string) (taskID string) {
+func submitAndAssert[P GenericWorkerPayload | dockerworker.DockerWorkerPayload](t *testing.T, td *tcqueue.TaskDefinitionRequest, payload P, state, reason string) (taskID string) {
```
petemoore, line 203: And here

workers/generic-worker/main.go
```diff
@@ -584,52 +583,25 @@ func ClaimWork() *TaskRun {
...
+	if validateErr != nil {
+		return validateErr
 	}
-	schemaLoader := gojsonschema.NewStringLoader(JSONSchema())
-	docLoader := gojsonschema.NewStringLoader(string(jsonPayload))
-	result, err := gojsonschema.Validate(schemaLoader, docLoader)
+	payload := map[string]interface{}{}
+	err := json.Unmarshal(jsonPayload, &payload)
 	if err != nil {
 		return MalformedPayloadError(err)
```
petemoore, line 593: I think this would need to be an internal error - maybe you even need to panic if there is an error here. At this point, you have ensured that it complies to the payload schema, but then it cannot be unmarshaled into a map[string]interface{} - which would be a very serious error, and not because of a malformed payload.

workers/generic-worker/main.go
```diff
@@ -584,52 +583,25 @@ func ClaimWork() *TaskRun {
...
-			task.Errorf("- %s", desc)
+	if _, exists := payload["image"]; exists {
+		err := task.convertDockerWorkerPayload()
+		if err != nil {
+			return err
+		}
+	} else {
+		err := json.Unmarshal(jsonPayload, &task.Payload)
+		if err != nil {
+			return MalformedPayloadError(err)
```
petemoore, line 603: again, this should probably panic here if there is an error. we have already seen that it passes json schema validation, so not being able to unmarshal is pretty serious.

petemoore (commented):
Very nice, thanks Boris!

workers/generic-worker/helper_test.go
matt-boris: `DockerWorkerPayload` is undefined for the other schemas/engines though. It would only be defined for multiuser posix.

workers/generic-worker/d2g_test.go
matt-boris: In the `submitAndAssert()` helper :)

workers/generic-worker/d2g_test.go
matt-boris: Same as https://github.com/taskcluster/taskcluster/pull/6355#discussion_r1264102645.

workers/generic-worker/schemas/multiuser_posix.yml
matt-boris: Nice catch! Thank you 🎉

ui/docs/reference/workers/generic-worker/README.mdx
matt-boris: I think the only _good way_ to fix this would be to fix the generator to just use "Generic worker payload" as the name as all of them as opposed to taking the name from the schema..

If you change the name of the other two schemas to also be `Payload`, then `GenericWorkerPayload` is not defined anymore and you run into a massive rabbit hole and/or chicken/egg problem while updating all the references. 🙃

workers/generic-worker/d2g_test.go
petemoore: lol

petemoore (approved):
Nice job Boris!

PR #6370, opened 2023-07-04
UI: Refactor service schema loading and validation

lotas:
Previously same schema was used on two different pages/components and there were errors if both pages would be loaded, saying that schema was already loading.
This refactoring adds single `loadServiceSchema()` function that loads resource once and adds it to `ajv` once to avoid those issues.

### DIFF
--- a/ui/src/views/TcYamlDebug/index.jsx
+++ b/ui/src/views/TcYamlDebug/index.jsx
@@ -12,7 +12,6 @@ import TextField from '../../components/TextField';
 import CodeEditor from '../../components/CodeEditor';
 import Dashboard from '../../components/Dashboard';
 import Button from '../../components/Button';
-import urls from '../../utils/urls';
 import { siteSpecificVariable } from '../../utils/siteSpecific';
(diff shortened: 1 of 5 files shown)


ui/src/views/TcYamlDebug/index.jsx
```diff
@@ -12,40 +12,33 @@ import TextField from '../../components/TextField';
...
-  );
-  ajv.addSchemaOnce(
-    await (await fetch(urls.schema('queue', 'v1/task.json'))).json()
-  );
-  ajv.addSchemaOnce(
-    await (
-      await fetch(urls.schema('queue', 'v1/create-task-request.json'))
-    ).json()
-  );
+  await ajv.loadServiceSchema('common', 'metaschema.json');
```
matt-boris, line 22: Why don't you put this in the `Promise.all()` block as well? Because it's not from a service?

matt-boris (approved):
Changes LGTM. Just had one small question. Code works either way.

PR #6379, opened 2023-07-06
Deduplicate worker payload schemas

petemoore:
We currently have docker-worker and generic-worker schemas checked in in two places in the monorepo. This PR is a simplification so that we only have one copy, and use the same generated code everywhere, rather than generating identical code in two places.


lotas (dismissed):
👍

matt-boris (dismissed):
👍🏻

matt-boris (dismissed):

matt-boris (approved):

Issue #6380, opened 2023-07-06
Merge tools/d2g/.taskcluster.yml into CI


petemoore:
It may be that parts of it are already taken care of by existing tasks, but we should make sure we haven't lost any of the steps that are currently in tools/d2g/.taskcluster.yml and remove it from the repo once it is fully integrated.

PR #6382, opened 2023-07-07
chore: extra cleanup after d2g migration

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6380.


taskcluster/ci/go/kind.yml
```diff
@@ -2,30 +2,71 @@
...
+        git --version
+        go version
+        go env
+        if [ ! -d taskcluster/.git ]; then
+          rm -rf taskcluster
+          git clone {head_repository} taskcluster
+        fi
+        cd taskcluster
+        git fetch {head_repository} "+${{GITHUB_SHA}}:refs/heads/X${{TASK_ID}}"
+        git checkout -f "X${{TASK_ID}}"
```
lotas, line 31: just curious, do you remember how that `X{{TASK_ID}}` refs is created? Do we create a branch for each task, or previous `get fetch` is doing something I'm not aware of 🤔

tools/jsonschema2go/jsonschema.go
```diff
@@ -661,7 +661,7 @@ func (subSchema *JsonSubSchema) link(job *Job) (err error) {
 	if ref := subSchema.Ref; ref != nil && *ref != "" {
 		subSchema.RefSubSchema = job.result.SchemaSet.all[subSchema.RefSchemaURL]
 		if subSchema.RefSubSchema == nil {
-			return fmt.Errorf("Subschema %v not loaded when updating %v", subSchema.RefSchemaURL, subSchema.SourceURL)
+			return fmt.Errorf("subschema %v not loaded when updating %v", subSchema.RefSchemaURL, subSchema.SourceURL)
```
lotas, line 664: u no like uppercase? :)

lotas: let's hope no-one is doing case-sensitive checks for those errors 😬

lotas (approved):
nice!

taskcluster/ci/go/kind.yml
matt-boris: I'm not sure.. this is how other tasks do it as well so I just went along with the magic 😂

I think a new branch is created for each task though, yes.

tools/jsonschema2go/jsonschema.go
matt-boris: Haha, it fails the staticcheck :)

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

lotas: good to know, thanks

taskcluster/ci/go/kind.yml
lotas: aaah.. that's the syntax to checkout specific SHA, yet create a branch (local ref) with this name.. neat

PR #6388, opened 2023-07-11
Align `go mod tidy` usage with current go semantics

petemoore:
As per `go help install`:

> If the arguments have version suffixes (like @latest or @v1.0.0), "go install"
> builds packages in module-aware mode, ignoring the go.mod file in the current
> directory or any parent directory, if there is one. This is useful for
> installing executables without affecting the dependencies of the main module.

By adding `@latest` to go install commands, we should ensure that installing dev tools (such as goimports) will not modify `go.mod` and `go.sum` files.

By continuing to check that running `go mod tidy` does not alter the contents of `go.mod` or `go.sum`, we also ensure that these files are consistent with the go code.

### DIFF
--- a/changelog/Poh6dB8NQ5S_-IAOgyYoew.md
+++ b/changelog/Poh6dB8NQ5S_-IAOgyYoew.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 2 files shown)


lotas (approved):
👍

matt-boris (approved):
👍🏻 Thanks Pete!

PR #6385, opened 2023-07-10
chore: add retry to podman run transforms

matt-boris:
Here's an example of this happening: https://community-tc.services.mozilla.com/tasks/RmJe8MoKR-2lxCfM9BhgGQ/runs/0/logs/public/logs/live.log#L65.

PR #6390, opened 2023-07-11
fix: update d2g to retry on exit 125

matt-boris:
Similar to how it was done in https://github.com/taskcluster/taskcluster/pull/6385, I've included it in `d2g` now as @petemoore suggested [here](https://github.com/taskcluster/taskcluster/pull/6385#issuecomment-1630356729).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -43,6 +43,7 @@ require (
 	github.com/tent/hawk-go v0.0.0-20161026210932-d341ea318957
 	github.com/xeipuuv/gojsonschema v1.2.0
 	golang.org/x/crypto v0.11.0
+	golang.org/x/exp v0.0.0-20230711023510-fffb14384f22
 	golang.org/x/net v0.12.0
 	golang.org/x/sys v0.10.0
 	golang.org/x/tools v0.9.2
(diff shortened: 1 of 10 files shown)


tools/d2g/d2g.go
```diff
@@ -295,6 +296,14 @@ func setMaxRunTime(dwPayload *dockerworker.DockerWorkerPayload, gwPayload *gener
 
 func setOnExitStatus(dwPayload *dockerworker.DockerWorkerPayload, gwPayload *genericworker.GenericWorkerPayload) {
 	gwPayload.OnExitStatus.Retry = dwPayload.OnExitStatus.Retry
+
+	// An error sometimes occurs while pulling the docker image:
+	// Error: reading blob sha256:<SHA>: Get "<URL>": remote error: tls: handshake failure
+	// And this exits 125, so we'd like to retry.
+	if !slices.Contains(gwPayload.OnExitStatus.Retry, 125) {
```
petemoore, line 303: The implementation in slices is pretty trivial: https://cs.opensource.google/go/x/exp/+/fffb1438:slices/slices.go;l=106-113

As per the go proverb [_a little copying is better than a little dependency_](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s) I think I would prefer it if we just looped through the slice here and didn't import the full slices package.

petemoore (approved):
Thanks for doing this!

go.mod
```diff
@@ -72,7 +72,7 @@ require (
 	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
-	golang.org/x/mod v0.10.0 // indirect
+	golang.org/x/mod v0.11.0 // indirect
```
petemoore, line 75: I guess this happened as a consequence of adding and then removing `slices` dependency, but I don't think it does any harm.

petemoore (approved):
Thanks Boris!

PR #6391, opened 2023-07-11
Refactor Hooks page do display extra info.

lotas:
Last fire and task state are shown, as well as schedule/binding details

image
image


matt-boris (approved):
Nice 😎

PR #6399, opened 2023-07-13
Fix freebsd

petemoore:
In the process of implementing osGroups for posix, I realised that we should either drop support for FreeBSD, or start testing it, like we test the other platforms.

I started by manually setting up a FreeBSD worker in AWS, and am iterating through fixing bugs in this PR. Eventually I hope to add a FreeBSD image set to community-tc-config (mozilla/community-tc-config#612).

### DIFF
--- a/changelog/Hj8UAm3LSKClWjTXC0ojiQ.md
+++ b/changelog/Hj8UAm3LSKClWjTXC0ojiQ.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
--- a/workers/generic-worker/build.sh
+++ b/workers/generic-worker/build.sh
@@ -1,4 +1,5 @@
-#!/bin/bash -e
+#!/usr/bin/env bash
+set -e
 
 cd "$(dirname "${0}")"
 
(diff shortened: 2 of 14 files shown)


petemoore:
Example task: https://community-tc.services.mozilla.com/tasks/SpvmiqIZQny42a9LxnyygQ/

matt-boris (approved):
Changes LGTM! I'm curious, why'd you have to replace `/bin/bash` with `/usr/bin/env bash` everywhere?

petemoore:
> Changes LGTM! I'm curious, why'd you have to replace `/bin/bash` with `/usr/bin/env bash` everywhere?

On FreeBSD it gets installed to /usr/local/bin/bash.
There is a similar issue on macOS, if you brew install bash - in that case /bin/bash will be bash 3, and the homebrew bash will be bash 5. By using `/usr/bin/env bash` you get whichever bash is in your PATH, which is kinda nice so you aren't stuck with a bash 3 on macOS, and also means you don't need to create a symlink on FreeBSD.

PR #6406, opened 2023-07-14
Bulk artifact removal is controlled by config value

### LINKED ISSUE #6405: Artifact expire job is failing on GCS
Artifact storage on community and fxci were switched from S3 to GCS.

```
exception InvalidArgument: Invalid argument.
POST https://storage.googleapis.com/communitytc-private-artifacts?delete [400]
```

```
InvalidArgument: Invalid argument.
  File "/app/node_modules/aws-sdk/lib/services/s3.js", line 711, col 35, in Request.extractError
    resp.error = AWS.util.error(new Error(), {
  File "/app/node_modules/aws-sdk/lib/sequential_executor.js", line 106, col 20, in Request.callListeners
    listener.apply(self, args);
  File "/app/node_modules/aws-sdk/lib/sequential_executor.js", line 78, col 10, in Request.emit
    this.callListeners(listeners, eventArgs, doneCallback);
  File "/app/node_modules/aws-sdk/lib/request.js", line 686, col 14, in Request.emit
    origEmit.call(this, eventName, args, function (err) {
  File "/app/node_modules/aws-sdk/lib/request.js", line 22, col 10, in Request.transition
    self.emit(self._asm.currentState, function(err) {
  File "/app/node_modules/aws-sdk/lib/state_machine.js", line 14, col 12, in AcceptorStateMachine.runTo
    state.fn.call(bindObject || self, inputError, function(err) {
  File "/app/node_modules/aws-sdk/lib/state_machine.js", line 26, col 10, in <anonymous>
    self.runTo(finalState, done, bindObject, err);
  File "/app/node_modules/aws-sdk/lib/request.js", line 38, col 9, in Request.<anonymous>
    done(self.response.error);
  File "/app/node_modules/aws-sdk/lib/request.js", line 688, col 12, in Request.<anonymous>
    done.call(this, err);
  File "/app/node_modules/aws-sdk/lib/sequential_executor.js", line 116, col 18, in Request.callListeners
    doneCallback.call(self, error);
  File "/app/node_modules/aws-sdk/lib/sequential_executor.js", line 78, col 10, in Request.emit
    this.callListeners(listeners, eventArgs, doneCallback);
  File "/app/node_modules/aws-sdk/lib/request.js", line 686, col 14, in Request.emit
    origEmit.call(this, eventName, args, function (err) {
  File "/app/node_modules/aws-sdk/lib/request.js", line 22, col 10, in Request.transition
    self.emit(self._asm.currentState, function(err) {
  File "/app/node_modules/aws-sdk/lib/state_machine.js", line 14, col 12, in AcceptorStateMachine.runTo
    state.fn.call(bindObject || self, inputError, function(err) {
  File "/app/node_modules/aws-sdk/lib/state_machine.js", line 26, col 10, in <anonymous>
    self.runTo(finalState, done, bindObject, err);
  File "/app/node_modules/aws-sdk/lib/request.js", line 38, col 9, in Request.<anonymous>
    done(self.response.error);
```

lotas:
Falling back to single artifact deletion for expire artifact job to support storage services without bulk delete support.

By default this is disabled.
If deployment supports bulk removal, it can set `AWS_USE_BULK_DELETE` to `true` to use bulk deletion instead.

Fixes #6405


services/queue/src/utils.js
```diff
@@ -37,17 +37,22 @@ const artifactUtils = {
...
    * deleting the underlying artifact.
+   *
+   * Not all S3-compatible storage providers support bulk delete, so we
+   * need to handle that case.
    */
-  async expire({ db, publicBucket, privateBucket, ignoreError, monitor, expires }) {
+  async expire({ db, publicBucket, privateBucket, ignoreError, monitor,
+    expires, useBulkDelete = false }) {
     let count = 0;
+    const maxPageSize = useBulkDelete ? 1000 : 100;
```
lotas, line 47: maybe this requires some explanation .. 
bulk delete operation can support up to `1000` records, where in single removal we don't want to fire too many parallel requests, so we use `100` which was used before bulk feature was added.

petemoore (approved):
Looks reasonable to me, thanks for this Yarik! Maybe Boris can take a look too before we land this. 👍

services/queue/src/utils.js
matt-boris: What do you think about making both of these values configurable as well? Maybe single deletes happen faster on GCS and we want to bump up that `100` value without having to make a code change?

matt-boris (commented):
Changes LGTM, just added one quick question.

matt-boris (dismissed):
Good stuff! LGTM 🎉

services/queue/src/utils.js
petemoore: > maybe this requires some explanation ..

Can you add this explanation to the code as a comment too?

lotas: yes! just added, thanks

petemoore (approved):
Thanks Yarik!

PR #6407, opened 2023-07-14
Livelog working in local dev env

### LINKED ISSUE #6395: Failed to fetch live.log and backing_live.log
I followed the document to run taskcluster locally, but can not open live.log and live_backing.log.
image

**Steps to reproduce:**
1. git clone https://github.com/taskcluster/taskcluster
2. cd taskcluster
3. yarn start
4. add “127.0.0.1 taskcluster” to /etc/hosts 
5. sign in (**Client Id:** static/taskcluster/root, **Access Token:** j2Z6zW2QSLehailBXlosdw9e2Ti8R_Qh2M4buAEQfsMA)
6. create a task and save
image
7. click live.log or live_backing.log

**Expected behavior:**
click live.log or live_backing.log and show the log in ui.

**Actual behavior:**
image

**Environment information:**
- Operating System: Windows 10
- Browser: Chrome

lotas:
By default it will remain `0` and so port would be allocated dynamically.

But for local development environment where generic-worker runs inside the container, and `livelog` binary is executed within same container, assigning ports dynamically would make it hard to expose it using static `docker-compose.yml` file. 

This way we provide predefined port that will be used to expose livelog, which then could be accessed on the host and in the UI.

generic worker config may look like this:
```json
  "publicIP": "127.0.0.1",
  "livelogPortBase": 60000,
  "livelogExposePort": 59999
```

This means it would run `livelog` with `60000`, `60001` ports, and read port would be exposed on `59999` port, which will be used for the reference artifact, and would make it accessible in the UI.

Fixes #6395

### DIFF
--- a/changelog/issue-6395-1.md
+++ b/changelog/issue-6395-1.md
@@ -3,4 +3,4 @@ level: patch
 reference: issue 6395
 ---
 
-Local development environment supports livelogging of the running tasks in the UI.
+Local development environment now supports live log.
--- a/changelog/issue-6395-1.md
+++ b/changelog/issue-6395-1.md
@@ -1,0 +1,6 @@
+audience: developers
+level: patch
+reference: issue 6395
+---
+
+Local development environment supports livelogging of the running tasks in the UI.
--- a/workers/generic-worker/expose/local.go
+++ b/workers/generic-worker/expose/local.go
@@ -8,13 +8,14 @@ import (
 
 type localExposer struct {
 	publicIP net.IP
+	port     uint16
 }
 
 // Create a local exposer implementation.  Local exposers are useful for local
 // testing, and simply exposes the URL as given, or (for ExposeTCPPort) proxies
 // a websocket on localhost to the port.
-func NewLocal(publicIP net.IP) (Exposer, error) {
-	return &localExposer{publicIP}, nil
+func NewLocal(publicIP net.IP, port uint16) (Exposer, error) {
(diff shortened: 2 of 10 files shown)


changelog/issue-6395-1.md
```diff
@@ -0,0 +1,6 @@
+audience: developers
+level: patch
+reference: issue 6395
+---
+
+Local development environment supports livelogging of the running tasks in the UI.
```
matt-boris, line 6:
```suggestion
Local development environment now supports live log.
```

workers/generic-worker/expose/local.go
```diff
@@ -8,13 +8,14 @@ import (
 
 type localExposer struct {
 	publicIP net.IP
+	port		 uint16
```
matt-boris, line 11:
```suggestion
	port     uint16
```

matt-boris (commented):
Changes look great, thanks Yarik! Just left a couple small change requests :)

matt-boris (dismissed):

petemoore (changes requested):
Looks good, just usage.go needs updating with some text about the config setting. The docs update should say that 0 means random port, and should highlight that the default is 0 (usually with `[default: 0]` at the end of the text - see the other properties for an example).

Thanks Yarik!

matt-boris (approved):

petemoore (approved):

Issue #6400, opened 2023-07-13
Deadlocks in interactive feature


petemoore:
A deadlock occurs intermittently in the Generic Worker CI.

Example: https://github.com/taskcluster/taskcluster/runs/14960216397

petemoore:
Introduced in #6249

petemoore:
Not sure why this didn't get closed automatically, but it was fixed in #6446.

PR #6408, opened 2023-07-14
Don't allow writing on the websocket twice at the same time

Eijebong:
Because the websocket library isn't thread safe and an error can occur and try to write the error to the websocket at the same time as the process outputs something, we have to lock the websocket while we're writing to it

### DIFF
--- a/changelog/S2l2EKSATh2TGiafzqHXxw.md
+++ b/changelog/S2l2EKSATh2TGiafzqHXxw.md
@@ -1,0 +1,3 @@
+audience: users
+level: silent
+---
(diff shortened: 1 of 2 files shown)


petemoore:
Thanks @Eijebong! Is this a complete fix for #6400?

Eijebong:
Nope, that only fixes the race condition, not the deadlock

changelog/S2l2EKSATh2TGiafzqHXxw.md
```diff
@@ -0,0 +1,3 @@
+audience: users
+level: silent
```
matt-boris, line 2: Mind updating this to `patch` and providing a short description?

matt-boris (commented):
Nice! Thanks for this Bastien! Just want a small update to the changelog is all 👍🏻

And lol at your branch name 😂

petemoore (dismissed):
Agree with everything Boris says. Thanks Bastien! ❤️

matt-boris (approved):

PR #6413, opened 2023-07-16
chore: clean up gw generator and add generate command for d2g

matt-boris:


petemoore (approved):
Thanks so much Boris!

PR #6431, opened 2023-07-19
fix(changelog): correctly filter on git log author

matt-boris:
Though, the previous arguments technically worked on a local machine, it seems like the regex filtering doesn't work when running the `child_process.execFile()` function (what is used to call `git log ...`).


lotas (approved):
awesome! Thanks for fixing this

PR #6434, opened 2023-07-20
chore(generic-worker): extra d2g log

matt-boris:
>Adds a task log letting the user know their Docker Worker payload is being converted to a Generic Worker payload using d2g.

### DIFF
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -595,6 +595,7 @@ func (task *TaskRun) validatePayload() *CommandExecutionError {
 		panic(err)
 	}
 	if _, exists := payload["image"]; exists {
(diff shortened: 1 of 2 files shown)


workers/generic-worker/main.go
```diff
@@ -595,6 +595,7 @@ func (task *TaskRun) validatePayload() *CommandExecutionError {
 		panic(err)
 	}
 	if _, exists := payload["image"]; exists {
+		task.Info("Docker Worker payload detected. Converting to a Generic Worker payload using d2g.")
```
lotas, line 598: I guess we don't want to dump resulting converted payload, as it would be visible in live log anyways?

lotas (approved):
👍 👍

PR #6443, opened 2023-07-31
Python lint fixes

### LINKED ISSUE #6442: Python client: flake8 E721 failures
In #6441 there were some python lint failures. I have _no_ idea why they showed up there and nowhere else. Maybe E721 checks did not previously run by default, and now do? It seems to be some state outside of our repo that triggered this change, which is concerning.

petemoore:
Fixes #6442

### DIFF
--- a/changelog/issue-6442.md
+++ b/changelog/issue-6442.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 6442
+---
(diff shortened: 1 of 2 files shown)


petemoore:
Oh interesting, I guess we don't pin our flake8 dependency. Thanks dustin!

petemoore:
Thanks Bastien!

PR #6447, opened 2023-08-01
Issue 6414: fix intermittent TestExponentialBackoffFailure

### LINKED ISSUE #6414: Intermittent test failue on `go-tools` CI task
The `TestExponentialBackoffFailure` test intermittently fails now for some reason. This is _likely_ do to the fact that the task now runs within podman on generic worker vs on docker worker. I believe the test has some racey timing logic that should be looked into. The test will normally succeed on a rerun or two.

https://github.com/taskcluster/taskcluster/blob/c3dc8d618fd9cc17cfd5972c55c8afb8cc37f18b/tools/websocktunnel/client/client_test.go#L84C1-L120

Here's an example failure https://community-tc.services.mozilla.com/tasks/VHnhysmsQEeuY1uwRmQodQ/runs/0/logs/public/logs/live.log#L502-521.

petemoore:
Fixes #6414
Introduced in 057a6db3781cd5057831b3fb7033a55528f45825.

### DIFF
--- a/tools/websocktunnel/client/client_test.go
+++ b/tools/websocktunnel/client/client_test.go
@@ -114,7 +114,13 @@ func TestExponentialBackoffFailure(t *testing.T) {
 		t.Fatalf("should not run for more than %d milliseconds", maxTime)
 	}
 
-	if count > 10 || count < 4 {
+	// Note, sometimes 11 requests occur, presumably because the attempts occur
+	// when clock is 0ms, 200ms, 400ms, 600ms, 800ms, 1000ms, 1200ms, 1400ms,
+	// 1600ms, 1800ms, 2000ms, or there is a timeout go routine that isn't
+	// guaranteed to be immediately scheduled after 2 seconds. We may even
+	// want to bump the limit to 12, no need to be super strict here. Let's see
+	// how it is with 11 for now.
+	if count > 11 || count < 4 {
 		t.Fatalf("wrong number of retries: %d", count)
 	}
 }
--- a/changelog/issue-6414.md
+++ b/changelog/issue-6414.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 6414
+---


tools/websocktunnel/client/client_test.go
```diff
@@ -114,7 +114,13 @@ func TestExponentialBackoffFailure(t *testing.T) {
...
 	}
 
-	if count > 10 || count < 4 {
+	// Note, sometimes 11 requests occur, presumably because the attempts occur
+	// when clock is 0ms, 200ms, 400ms, 600ms, 800ms, 1000ms, 1200ms, 1400ms,
+	// 1600ms, 1800ms, 2000ms, or there is a timeout go routine that isn't
+	// guaranteed to be immediately scheduled after 2 seconds. We may even
+	// want to bump the limit to 12, no need to be super strict here. Let's see
+	// how it is with 11 for now.
+	if count > 11 || count < 4 {
```
Eijebong, line 123: Wouldn't the condition above also be an issue? Sounds like this couldn't handle 11 retries anyway?

```
maxTime := retry.MaxElapsedTime + 200*time.Millisecond
if end.Sub(start) > maxTime {
```

petemoore: Good point. I haven't seen that failure case (line 114), but indeed there is a chance we will get intermittent failures there too if the machine is running really slowly. Until now I think we've only had failures with count being 11 (which I think is retries+1, i.e. retries = 10 when count = 11) so maybe let's try it like this, and adjust line 113 too if we start seeing failures there too?

Eijebong (approved):
🤞

tools/websocktunnel/client/client_test.go
petemoore: > Yep, makes sense, I though count was only the number of retries, not the number of tries.

I think the counter on line 89 is bumped every time a request is made, including the first request, before any retries occur.

Issue #2942, opened 2020-05-29
Allow longer deadlines


djmitche:
The task deadline limit is based on Azure's limit for queue entry lifetime (7 days).  We're not using Azure anymore, though.

https://bugzilla.mozilla.org/show_bug.cgi?id=1604175

We can actually lift this restriction now, but let's wait for #2940 so as not to cause headaches.

gabrielBusta:
#6269

petemoore:
I think this issue is about task deadline, not maxRunTime.

gabrielBusta:
Thanks for clarifying @petemoore I got them mixed up. This is also an issue for the translations training pipeline. If we run tasks with a high `maxRunTime` by the time its dependencies run they will already have exceeded the deadline. It sounds like this depends on which queue implementation we are using now (is it Pulse?)

PR #6465, opened 2023-08-04
feat(g-w): Allow the maximum on the maxRunTime property to be configured

### LINKED ISSUE #6269: Allow a configurable maxRunTime for generic worker
As I understand it, generic worker currently hardcodes a maximum run time of 1 day (86400 seconds). This has been a pretty sane and safe default for a long time, but there are certain use cases where longer run times are useful and necessary. The specific example I'm working with now is training language models for Firefox Translations, but I know that @rail has unrelated use cases as well (and I'm sure there's others out there, eg: possibly fuzzing).

Allowing the max run time to be overridden in the generic worker config would be one way to accomplish this, and reduce the possibility of huge max run times for tasks in pools where it is absolutely not needed.

gabrielBusta:
Some tinkering around #6269 to unblock https://github.com/mozilla/firefox-translations-training/issues/153

### DIFF
--- a/changelog/issue-6269.md
+++ b/changelog/issue-6269.md
@@ -1,0 +1,5 @@
+audience: developers
+level: minor
+reference: issue 6269
+---
+The maximum value for `maxRunTime` on `generic-worker` payloads can now be configured using a `maxTaskRunTime` property on the `generic-worker` config.
--- a/changelog/issue-6269.md
+++ b/changelog/issue-6269.md
@@ -2,4 +2,4 @@ audience: developers
 level: minor
 reference: issue 6269
 ---
-The maximum value for `maxRunTime` on `generic-worker` payloads can now be configured using a `maxTaskRunTime` property on the `generic-worker` config.
+Generic Worker now provides configuration property `maxTaskRunTime` as an upper bound for task payload property `maxRunTime`. Tasks with `maxRunTime` exceeding this value will be resolved as `exception/malformed-payload`.
--- a/tools/d2g/genericworker/generated_types.go
+++ b/tools/d2g/genericworker/generated_types.go
@@ -484,11 +484,11 @@ type (
 		Logs Logs `json:"logs,omitempty"`
 
 		// Maximum time the task container can run in seconds.
+		// The maximum value for `maxRunTime` is set by a `maxTaskRunTime` config property specific to each worker-pool.
 		//
 		// Since: generic-worker 0.0.1
 		//
 		// Mininum:    1
-		// Maximum:    86400
 		MaxRunTime int64 `json:"maxRunTime"`
 
 		// Directories and/or files to be mounted.
@@ -1110,8 +1110,7 @@ func JSONSchema() string {
           "type": "object"
         },
         "maxRunTime": {
-          "description": "Maximum time the task container can run in seconds.\n\nSince: generic-worker 0.0.1",
-          "maximum": 86400,
+          "description": "Maximum time the task container can run in seconds.\nThe maximum value for ` + "`" + `maxRunTime` + "`" + ` is set by a ` + "`" + `maxTaskRunTime` + "`" + ` config property specific to each worker-pool.\n\nSince: generic-worker 0.0.1",
           "minimum": 1,
           "multipleOf": 1,
           "title": "Maximum run time in seconds",
--- a/workers/generic-worker/gwconfig/config.go
+++ b/workers/generic-worker/gwconfig/config.go
@@ -45,6 +45,7 @@ type (
 		LiveLogPortBase                uint16                 `json:"livelogPortBase"`
 		LiveLogExposePort              uint16                 `json:"livelogExposePort"`
 		LoopbackVideoDeviceNumber      uint8                  `json:"loopbackVideoDeviceNumber"`
+		MaxTaskRunTime                 uint32                 `json:"maxTaskRunTime"`
 		NumberOfTasksToRun             uint                   `json:"numberOfTasksToRun"`
 		PrivateIP                      net.IP                 `json:"privateIP"`
 		ProvisionerID                  string                 `json:"provisionerId"`
--- a/workers/generic-worker/helper_test.go
+++ b/workers/generic-worker/helper_test.go
@@ -372,7 +372,7 @@ func GWTest(t *testing.T) *Test {
 			// These ports are not exposed outside of the host. However, in CI they must differ from those of the
 			// generic-worker instance running the test suite.
 			LiveLogPortBase:    30583,
-			MaxTaskRunTime:     180,
+			MaxTaskRunTime:     300,
 			NumberOfTasksToRun: 1,
 			PrivateIP:          net.ParseIP("87.65.43.21"),
 			ProvisionerID:      "test-provisioner",
--- a/workers/generic-worker/helper_test.go
+++ b/workers/generic-worker/helper_test.go
@@ -372,6 +372,7 @@ func GWTest(t *testing.T) *Test {
 			// These ports are not exposed outside of the host. However, in CI they must differ from those of the
 			// generic-worker instance running the test suite.
 			LiveLogPortBase:    30583,
+			MaxTaskRunTime:     86400,
 			NumberOfTasksToRun: 1,
 			PrivateIP:          net.ParseIP("87.65.43.21"),
 			ProvisionerID:      "test-provisioner",
--- a/workers/generic-worker/helper_test.go
+++ b/workers/generic-worker/helper_test.go
@@ -372,7 +372,7 @@ func GWTest(t *testing.T) *Test {
 			// These ports are not exposed outside of the host. However, in CI they must differ from those of the
 			// generic-worker instance running the test suite.
 			LiveLogPortBase:    30583,
-			MaxTaskRunTime:     60,
+			MaxTaskRunTime:     180,
 			NumberOfTasksToRun: 1,
 			PrivateIP:          net.ParseIP("87.65.43.21"),
 			ProvisionerID:      "test-provisioner",
--- a/workers/generic-worker/helper_test.go
+++ b/workers/generic-worker/helper_test.go
@@ -372,7 +372,7 @@ func GWTest(t *testing.T) *Test {
 			// These ports are not exposed outside of the host. However, in CI they must differ from those of the
 			// generic-worker instance running the test suite.
 			LiveLogPortBase:    30583,
-			MaxTaskRunTime:     86400,
+			MaxTaskRunTime:     60,
 			NumberOfTasksToRun: 1,
 			PrivateIP:          net.ParseIP("87.65.43.21"),
 			ProvisionerID:      "test-provisioner",
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -210,6 +210,7 @@ func loadConfig(configFile *gwconfig.File) error {
 			LiveLogExecutable:              "livelog",
 			LiveLogPortBase:                60098,
 			LoopbackVideoDeviceNumber:      0,
+			MaxTaskRunTime:                 86400,
 			NumberOfTasksToRun:             0,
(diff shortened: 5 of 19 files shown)


workers/generic-worker/helper_test.go
```diff
@@ -372,6 +372,7 @@ func GWTest(t *testing.T) *Test {
 			// These ports are not exposed outside of the host. However, in CI they must differ from those of the
 			// generic-worker instance running the test suite.
 			LiveLogPortBase:    30583,
+			MaximumMaxRunTime:  86400,
```
gabrielBusta, line 375: This is a bit weird. When I run the test I get a 0 for this value (that's why I set it again here) but when I do `generic-worker run` locally I see it defaulting to 86400

workers/generic-worker/gwconfig/config.go
```diff
@@ -45,6 +45,7 @@ type (
 		LiveLogPortBase                uint16                 `json:"livelogPortBase"`
 		LiveLogExposePort              uint16                 `json:"livelogExposePort"`
 		LoopbackVideoDeviceNumber      uint8                  `json:"loopbackVideoDeviceNumber"`
+		MaximumMaxRunTime              uint32                 `json:"maximumMaxRunTime"`
```
gabrielBusta, line 48: We need a `uint32` to store the value of a month's worth of seconds

workers/generic-worker/main.go
```diff
@@ -945,6 +946,11 @@ func (task *TaskRun) Run() (err *ExecutionErrors) {
 	task.logHeader()
 
 	err.add(task.validatePayload())
+
+	if task.Payload.MaxRunTime > int64(config.MaximumMaxRunTime) {
```
gabrielBusta, line 950: Note: move this into `validatePayload` because it's validating a payload value?

tools/d2g/genericworker/generated_types.go
```diff
@@ -278,7 +278,6 @@ type (
 		// Maximum time the task container can run in seconds.
 		//
 		// Mininum:    1
-		// Maximum:    86400
```
petemoore, line 281: Would be nice to add a note to the description that the maximum value is variable, but worker-pool specific.

petemoore (commented):
Looking good so far!

changelog/issue-6269.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
+level: minor
+reference: issue 6269
+---
+The maximum value for `maxRunTime` on `generic-worker` payloads can now be configured using a `maxTaskRunTime` property on the `generic-worker` config.
```
petemoore, line 5:
```suggestion
Generic Worker now provides configuration property `maxTaskRunTime` as an upper bound for task payload property `maxRunTime`. Tasks with `maxRunTime` exceeding this value will be resolved as `exception/malformed-payload`.
```

workers/generic-worker/helper_test.go
```diff
@@ -372,6 +372,7 @@ func GWTest(t *testing.T) *Test {
 			// These ports are not exposed outside of the host. However, in CI they must differ from those of the
 			// generic-worker instance running the test suite.
 			LiveLogPortBase:    30583,
+			MaxTaskRunTime:     86400,
```
petemoore, line 375: For tests, I don't think we need such a high value. It doesn't hurt, but perhaps a value like 60 might be better, so that if someone writes a test that needs to take longer than 60s to run, they would at least see they have escaped the bounds of existing tests.

workers/generic-worker/main.go
```diff
@@ -210,6 +210,7 @@ func loadConfig(configFile *gwconfig.File) error {
 			LiveLogExecutable:              "livelog",
 			LiveLogPortBase:                60098,
 			LoopbackVideoDeviceNumber:      0,
+			MaxTaskRunTime:                 86400,
```
petemoore, line 213: Might be nice to add a comment that 86400s is 24 hours.

workers/generic-worker/payload_test.go
```diff
@@ -167,8 +168,32 @@ func TestArtifactExpiresBeforeDeadline(t *testing.T) {
...
+      "type": "file",
+      "path": "public/some/artifact",
+      "expires": "` + tcclient.Time(now.Add(time.Minute*20)).String() + `"
+    }
+  ]
+}`)
+	task.Definition.Deadline = tcclient.Time(now.Add(time.Minute * 10))
+	task.Definition.Expires = tcclient.Time(now.Add(time.Minute * 20))
+	ensureMalformedPayload(t, task)
+}
```
petemoore, line 192: it would be very easy for this task to resolve as malformed payload because it contains a raw string which is quite complicated which is meant to be json. If there was a missing quotation mark etc, it would resolve as malformed payload. There are other tests that don't use raw strings, and create a very simple task. Would it be possible to carbon cult one of those instead? Another thing that would be useful here is checking that the error message contained the expected text about the maxRunTime being too high. There are other tests that check the log messages - can you use one of those as a basis? This way, we can be sure the malformed payload really is due to the maxRunTime and not one of the other causes of a malformed payload exception, such as invalid json, or if the syntax of artifacts changes, etc.

petemoore (commented):
Thanks @gabrielBusta !

petemoore:
> Some tinkering around #6269 #2942 to unblock [mozilla/firefox-translations-training#153](https://github.com/mozilla/firefox-translations-training/issues/153)

Note, #2942 is about task deadline, rather than maxRunTime.

workers/generic-worker/payload_test.go
gabrielBusta: Thanks for the feedback, that cleaned up nicely.

gabrielBusta: Thanks for the feedback. I updated the test to declare the payload using `GenericWorkerPayload`struct instead of a raw string and refactored it so that it checks that the test fails due to a malformed payload exception caused by an out of bounds `MaxTaskRunTime`.

workers/generic-worker/helper_test.go
gabrielBusta: I made it 180s because there are several tests that set that as the `MaxRunTime`.

gabrielBusta: Scratch that, I made it 300s because there is a test that sets that as the `MaxRunTime`.

gabrielBusta:
I want to ability to set the `MaxTaskRunTime` on the workers because it's different depending on where they are running. But, now I am wondering, is there a limit we should set on this configuration parameter taking into account limits on the queue's task deadline?

gabrielBusta:
I want to ability to set the MaxTaskRunTime on the workers because it's different depending on where they are running (and their purpose.) But, now I am wondering, is there a limit we should set on this configuration parameter taking into account limits on the queue's task deadline?

petemoore:
> I want to ability to set the MaxTaskRunTime on the workers because it's different depending on where they are running (and their purpose.) But, now I am wondering, is there a limit we should set on this configuration parameter taking into account limits on the queue's task deadline?

Oh I see what you mean! Looking at queue.createTask, indeed the deadline requirements state "Note, deadline cannot be more than 5 days into the future".
I believe this was historically due to a limitation of the Azure Queues we were using to manage the task queues before we migrated to Postgres. Now that we have a postgres backend, we can probably lift this restriction. Although some of these restrictions did also serve as protections for the platform in general, so we should be careful about what the impact of lifting such a restriction could be. For example, regardless of how many tasks you submit today, you can be sure that they won't impact worker throughput in 5 days' time, since those tasks will have exceeded their deadline.

As this PR stands now, it would not be effective, due to the Queue restriction you have highlighted, so I don't think it makes sense to land it as is. I have a strong feeling it would be better to invest more effort into snapshotting training task state so that training sessions can be paused and continued. I think the impact on taskcluster of allowing long lived tasks may be introducing more problems and complexity than it solves, and since the stated goal is to be able to dice up the long lived tasks into shorter ones, it would be better to focus on that, rather than introduce features which will be redundant once the preferred solution has been implemented. Perhaps we can meet in Montreal to discuss this next week?

CC @bhearsum @gabrielBusta @eu9ene

petemoore:
> ... and since the stated goal is to be able to dice up the long lived tasks into shorter ones, it would be better to focus on that, rather than introduce features which will be redundant once the preferred solution has been implemented.

https://github.com/mozilla/firefox-translations-training/issues/154 is the issue which tracks implementing Marian checkpoints

bhearsum:
I agree that pausing and resuming training is something we should support.

However, even if and when we have that, we _still_ want to be able to configure our max run times and deadlines in the Firefox CI cluster. For translations, in addition to the spot instances that we already use, we will be hooking up hardware machines that we own as workers for Translations. It would also not be shocking if we put other hardware in a datacenter, or look at using reserved instances in GCP. None of these are cases where we need to worry about termination.

But more generally - as a deployer of Taskcluster, I should be in control over max run times and max deadlines. As I understand it, there are no technical reasons why these can't be higher at this point, which makes it a policy decision that should be up to the owner of the cluster IMO.

I agree that Taskcluster should have sensible deadlines (and I think the current ones are very sensible), but IMO Taskcluster, being a generic task execution platform, should not be placing artificial limits on how long tasks can run for, or be scheduled for.

eu9ene:
> the stated goal is to be able to dice up the long lived tasks into shorter ones, it would be better to focus on that, rather than introduce features which will be redundant once the preferred solution has been implemented

@petemoore The goal is to support long-running tasks on TC. It is an absolute blocker for the training pipeline. Some regular tasks (not training) can also take longer than a day (for example cleaning, especially after we integrate OpusCleaner all-in-one tool). Training is the most critical one. Having many small tasks and not being able to look at the full training log complicates observability and make an investigation of issues a nightmare. Also, I believe when Marian restarts from a checkpoint it might not start training from exactly the same place because of the dataset reshuffling. Continuation support is needed just to not lose the trained checkpoint if something goes wrong which is useful for debugging and testing new configs and is not a regular work path. 

ML tasks often operate on large datasets and are not always parallelizable to smaller chunks which makes them long-running. If we position TC as a general-purpose task execution framework and not just a CI tool, there shouldn't be artificial limitations for task time, the timeouts should be configurable, ideally per task.

lotas:
Queue service is checking `deadline` for a task here: https://github.com/taskcluster/taskcluster/blob/7e0ca9b54b5e87641e436172747319c8901be587/services/queue/src/api.js#L631-L640 

Currently it is limited to 5 days. 

Historically this was a limitation of Azure tables, as Pete mentioned before, but since migration to postgres, it can be increased. 
I'm not sure what is the best way here, as increasing deadline for all tasks might result in taskgraphs taking much longer to resolve. 

Not all tasks might require long execution time, so maybe it would be good to place this under a separate scope? To make sure that regular tasks/users would not be able to make tasks run forever.
And I also believe that ability to pause/continue is a different issue, and Taskcluster should allow users to run longer tasks.

bhearsum:
> Queue service is checking `deadline` for a task here:
> 
> https://github.com/taskcluster/taskcluster/blob/7e0ca9b54b5e87641e436172747319c8901be587/services/queue/src/api.js#L631-L640
> 
> Currently it is limited to 5 days.
> 
> Historically this was a limitation of Azure tables, as Pete mentioned before, but since migration to postgres, it can be increased. I'm not sure what is the best way here, as increasing deadline for all tasks might result in taskgraphs taking much longer to resolve.

To be clear: on the deadlines, the request is to allow tasks to configure a longer deadline than 5 days. Most likely, we'd only make use of this for Translations tasks at the moment. (Firefox tasks already set their deadlines at 24h - so those aren't even making use of the 5 days allowed at the moment. We'd most likely keep the default in Taskgraph at 24h as well.)

petemoore:
OK, you've all convinced me that it would be wrong to introduce artificial limitations. I agree with your arguments. Thank you for taking the time to write up your thoughts.

petemoore (approved):
Looks great, many thanks @gabrielBusta!

PR #6477, opened 2023-08-15
Parameterized role doc fixes

petemoore:
Some doc fixes.

Note, we should probably auto generate this doc based on the `project-admin:*` role, since they easily get out of sync with each other. However, it is specific to community-tc, and is a convention rather than a requirement, so that might not be so straightforward. I also noticed that role has some non-project scopes in there, which doesn't seem right, but I will look at that in a separate PR (presumably configured in community-tc-config repo, so a PR for that repo instead).


matt-boris (approved):
Thanks Pete

lotas (approved):

PR #6479, opened 2023-08-16
Don't force the interactive process to exit

Eijebong:
I added that when fixing test failures in
dca2377bc8a27cdc6e4ee928a95058b7f53ca528 and
i6026a3fa263175550cda877a6d7b9123301becc7 in the hope that it would make tests fail faster instead of deadlocking should something go wrong with my patch.

Turns out it added a new kind of intermittent test failure because the interactive process could exit before we finish, signal it, close the server-side websocket before the test is done reading from it.


matt-boris (approved):
Thanks Bastien 👍🏻

Issue #6481, opened 2023-08-17
Can't set `metadata.source` to a Github ssh URL


ahal:
**Describe the bug**
I'm trying to create a Decision task on a private repository, so SSH urls are needed to clone the repo. So I have:
```
          repoUrl:
              $switch:
                  'tasks_for == "github-push"': '${event.repository.ssh_url}'
                  'tasks_for[:19] == "github-pull-request"': '${event.pull_request.head.repo.ssh_url}'
                  'tasks_for in ["cron", "action"]': '${repository.url}'
```

in the `.taskcluster.yml`. Later on we have:
```
              metadata:
                      - source: "${repoUrl}/raw/${head_sha}/.taskcluster.yml"
```

But this results in an error:
![image](https://github.com/taskcluster/taskcluster/assets/273734/027ce482-d3f7-4597-912a-88239a953c02)

I assume the url format is something like `git@github.com:mozilla...`

**Expected behavior**

Github SSH urls are valid for `metadata.source`.

ahal:
This is easily worked around by using the http URL for `metadata.source` and the SSH url for cloning. So I'm just filing this for future improvement.

PR #6482, opened 2023-08-17
Allow for git ssh source URLs

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6481.

>Allow git SSH urls in `metadata.source`.

### DIFF
--- a/services/queue/schemas/constants.yml
+++ b/services/queue/schemas/constants.yml
@@ -113,7 +113,7 @@ scope:
     pattern: "^[\x20-\x7e]*$"
 
 # Pattern for source URLs; this rules out insecure URLs like data: or javascript:
-source-pattern: "^(https?|ssh)://"
+source-pattern: "^(https?:\/\/|ssh:\/\/|git@)"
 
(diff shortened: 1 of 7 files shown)


services/queue/schemas/constants.yml
```diff
@@ -113,7 +113,7 @@ scope:
     pattern: "^[\x20-\x7e]*$"
 
 # Pattern for source URLs; this rules out insecure URLs like data: or javascript:
-source-pattern: "^(https?|ssh)://"
+source-pattern: "^(https?:\/\/|ssh:\/\/|git@)"
```
lotas, line 116: I wonder if there are any tests that we can extend to check this as well?

lotas (approved):
lgtm

services/queue/schemas/constants.yml
matt-boris: Sorry about that! I had automerge enabled. I've added a test here https://github.com/taskcluster/taskcluster/pull/6483.

Issue #6476, opened 2023-08-14
Don't cache generic worker indexed artifacts


petemoore:
We introduced indexed artifacts in [Generic Worker 51.1.0](https://github.com/taskcluster/taskcluster/blob/main/CHANGELOG.md#v5110).

Unfortunately, it uses the same caching mechanism as the other artifact types, which means, if it the artifact has already been downloaded, Generic Worker won't query the Index to see if there is a new version available.

Generic Worker should always check the Index to see if there is a new version available. If there isn't, it is fine to use its cached copy, but if there is, it should update its cache.

Note, d2g is currently not relying on this feature, and instead generates the code to query the Index directly, so Docker Worker indexed artifacts are working correctly, only in Generic Worker they are not.

Note, once this is fixed in Generic Worker, d2g could use the native Generic Worker feature rather and benefit from cached downloads of artifacts.

PR #6486, opened 2023-08-18
fix(generic-worker): check for new Indexed Artifact before accessing cached artifact

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6476.

>Generic Worker now checks the Index to see if there is a new version of an Indexed Artifact available. If there isn't, it is fine to use its cached copy, but if there is, it updates its cache.


workers/generic-worker/mounts.go
```diff
@@ -855,12 +858,13 @@ func (ic *IndexedContent) String() string {
...
-	return "artifact:" + ac.TaskID + ":" + ac.Artifact
+func (ac *ArtifactContent) UniqueKey(taskMount *TaskMount) (string, error) {
+	return "artifact:" + ac.TaskID + ":" + ac.Artifact, nil
 }
 
-func (ic *IndexedContent) UniqueKey() string {
-	return "indexed:" + ic.Namespace + ":" + ic.Artifact
+func (ic *IndexedContent) UniqueKey(taskMount *TaskMount) (string, error) {
+	itr, err := taskMount.index.FindTask(ic.Namespace)
+	return "indexed:" + ic.Namespace + ":" + ic.Artifact + ":" + itr.TaskID, err
```
petemoore, line 867:
```suggestion
	return "artifact:" + itr.TaskID + ":" + ic.Artifact, err
```

This way, we won't have two copies of the same artifact in the disk cache, if they represent the same content, but one is an indexed artifact, and the other explicitly refers to the task id. Also if there are two separate indexes (e.g. a nightly index, and a latest index) that happen to point to exactly the same taskId/artifact combination, this also won't be duplicated in the disk cache.

petemoore (approved):
Hey Boris, this looks awesome. I just spotted one minor optimisation, but with that, I think it is good to go! Thanks for taking care of this! ❤️

PR #6489, opened 2023-08-29
workerManager.getWorker always returns worker 

### LINKED ISSUE #6117: Worker manager shows incorrect number of running instances
There was an incident where generic-worker process was paniced and exited, while keeping instance running for several days.

Worker pool showed number of instances including this one, but when going into "list workers" view, it was not displayed there.

Further investigation showed that worker record existed in `workers` table but not in `queue_workers` which probably caused this not showing up issue.

> I've got 2 clogged worker pools.. https://community-tc.services.mozilla.com/worker-manager/proj-fuzzing%2Fwindows-pool55 and https://community-tc.services.mozilla.com/worker-manager/proj-fuzzing%2Fwindows-pool85 both have idle workers that are "running" but not claiming tasks.
> pool55 also says 2 running but I can only see 1

> `i-0d08846ed25267639` appears in logs within few hours after it was created, on 18th..
> then last message from websocktunnel `session removed` and that's it

Worker died because of OOM:
```
2023/03/18 16:20:14Z Error: Post "https://community-tc.services.mozilla.com/api/queue/v1/task/CUFljGCBS6mMV7Z8YxtRtA/runs/0/reclaim": dial tcp 34.102.144.36:443: bind: An operation on a socket could not be performed because the system lacked sufficient buffer space or because a queue was full.

2023/03/18 16:21:04Z Received task status change: Reclaimed
panic: Could not PUT to http://localhost:80/credentials: Put "http://localhost:80/credentials": dial tcp [::1]:80: bind: An operation on a socket could not be performed because the system lacked sufficient buffer space or because a queue was full.

goroutine 55 [running]:
main.(*TaskclusterProxyTask).Start.func1({0x93f172, 0x9})
	/task_167821116098671/taskcluster/workers/generic-worker/taskcluster_proxy.go:103 +0x56b
main.(*TaskStatusManager).updateStatus(0xc0006fe7e0, {0x93f172, 0x9}, 0xc000fb5e90, {0xc0003be1a0?, 0x2, 0x2})
	/task_167821116098671/taskcluster/workers/generic-worker/taskstatus.go:304 +0x60d
main.(*TaskStatusManager).reclaim(0xc0006fe7e0)
	/task_167821116098671/taskcluster/workers/generic-worker/taskstatus.go:157 +0x8a
main.NewTaskStatusManager.func1()
	/task_167821116098671/taskcluster/workers/generic-worker/taskstatus.go:367 +0x44f
created by main.NewTaskStatusManager
	/task_167821116098671/taskcluster/workers/generic-worker/taskstatus.go:340 +0x1d8

```

lotas:
For some historical reasons `queue.getWorker` was returning `404` for worker that were quarantined or expired.
After it was deprecated and `workerManager.getWorker` method was introduced, I no longer think we should return `404` when record still exists.
This leads to situations where UI shows one number of workers, but when you try to open it, you get "Worker not found" error

Fixes #6117


matt-boris (approved):
Nice, thanks! LGTM

petemoore:
Would the response show if the worker was quarantined?

I ran a test for a non-quarantined worker and got back these fields:

```
$ taskcluster api workerManager worker proj-fuzzing/grizzly-reduce-worker us-east1 9188080371018223247
{
  "workerPoolId": "proj-fuzzing/grizzly-reduce-worker",
  "workerGroup": "us-east1",
  "workerId": "9188080371018223247",
  "providerId": "community-tc-workers-google",
  "created": "2023-08-26T21:36:34.148Z",
  "expires": "2023-09-06T21:38:15.068Z",
  "state": "running",
  "capacity": 1,
  "lastModified": "2023-08-26T21:37:25.887Z",
  "lastChecked": "2023-08-30T10:50:56.671Z"
}
```

If we change the API so that quarantined workers are now listed, that could break integrations that expect only active workers to be listed (e.g. maybe RelOps has something) - so we should make this a breaking change if we do go with this. At the moment this is marked as `patch` level, but I think it should be `major`.

I also see `removeWorkerManagerData: true` in the code (admittedly, this hasn't changed) - I'm curious what data that removes and why, but that isn't related to this change.

lotas:
Thanks for the check @petemoore 
I'm wondering if it makes sense in your example not to return it? If you quarantine some worker, it is probably because you want it back later, so you want to query and see the details for this worker. 

I'm really curious to know why someone would not expect to get the worker that is still running but just quarantined?

lotas:
Actually, it is more of an `expired` check than quarantined. 
Workers are being removed from queue view every hour, and from worker-manager view once a day. 
So the only inconsistency I see here is when worker is scheduled for removal (expires before the job runs) and we still show it.

PR #6491, opened 2023-08-30
feat(generic-worker): log out d2g translated payload

matt-boris:
>Generic Worker: If a Docker Worker payload is received, the resulting, d2g-translated Generic Worker payload will be logged out to the user.

### DIFF
--- a/changelog/U4SFSYxkTWW283bMcwZdbg.md
+++ b/changelog/U4SFSYxkTWW283bMcwZdbg.md
@@ -1,0 +1,4 @@
+audience: users
+level: patch
+---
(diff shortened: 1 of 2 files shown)


petemoore (commented):
I like this a lot, but I feel like we should output the full task defintion if possible, or at least include the scopes. The thing I like about a full task definition is that it includes everything to reproduce the task, so you have the worker pool etc, and you can copy/paste the full thing into the task creator, rather than needing to manually work out the missing parts from the original task. What do you think? It will be a bit of work to output it nicely, because the payload is stored in a different place to the rest of the task definition, but I think you should be able to copy the task definition, and then replace the payload part with json.RawMessage of the converted payload, and then call json.MarshalIndent against the full thing...

PR #6509, opened 2023-09-01
Remove renovate and group dependabot PRs

matt-boris:


petemoore (approved):
Will this still work with dependabot changelog updates, or will we now get one update for all the changes? Maybe also not a bad thing.

matt-boris:
> Will this still work with dependabot changelog updates, or will we now get one update for all the changes? Maybe also not a bad thing.

Good question.. I vote, we see how it works out and move from there 🙂

petemoore:
> > Will this still work with dependabot changelog updates, or will we now get one update for all the changes? Maybe also not a bad thing.
> 
> Good question.. I vote, we see how it works out and move from there 🙂

I think we'll find out if we make a staging release, won't we?

PR #6522, opened 2023-09-05
Fix code-scanning #318: possible overflow

petemoore:
Fixes: https://github.com/taskcluster/taskcluster/security/code-scanning/318

Note, this updated implementation is marginally less efficient, but removes the warning, which is kind of bogus, since practically I don't think it would be possible to hit the limits with a real payload, as the payload would have to be so big, I suspect some other limit would prevent such a task being created anyway.

### DIFF
--- a/changelog/A3EkBRJlRkSxwkS-Ke50BA.md
+++ b/changelog/A3EkBRJlRkSxwkS-Ke50BA.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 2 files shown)


lotas (approved):
I don't think it is less efficient, just more lines :)

Thanks!

petemoore:
> I don't think it is less efficient, just more lines :)
> 
> Thanks!

The inefficiency is just in allocation, the array may grow beyond its original allocation, and need reallocating, whereas with an explicit size upfront, there should just be one allocation, but I think it really doesn't matter.

PR #6523, opened 2023-09-05
no-op: simplify string replacement expression in gw-codegen

petemoore:
This is just a trivial simplification, I don't know why I used the more complicated form originally, quite bizarre.


lotas (approved):
👍 not sure what it does besides removing indenting.. no tests broken though, so must be good :)

matt-boris (approved):
👍🏻

PR #6525, opened 2023-09-05
Rename dependabot groups

matt-boris:
I felt like `build(deps-dev): bump the deps-dev group in /clients/client with 1 update` sounded a little strange for a PR title, so dropped the `-dev` part so it'll read, `build(deps-dev): bump the deps group in /clients/client with 1 update` in future PRs.


lotas (approved):
nice! Thanks!

PR #6531, opened 2023-09-07
Add support for sha256 webhook signatures

### LINKED ISSUE #6528: Upgrade github webhook to use x-hub-signature-256
sha256 is recommended by [github](https://docs.github.com/en/webhooks/using-webhooks/securing-your-webhooks#validating-payloads-from-github) so we should upgrade webhook handler to use this hash

lotas:
Errors are being reported to Sentry additionally now

Fixes #6528


petemoore (commented):
This looks reasonable to me - and thanks for moving the helper functions into the utils.js file! My node isn't good enough to do a thorough review, so I'll leave that to others. Thanks for specifically testing the monitor logging for the HTTP 403 case to ensure we are reporting to sentry when the webhook signature is invalid. Also thanks for covering sha1 when the taskcluster deployment is oeprating against a Github Enterprise account that doesn't have sha256 validation - nice thinking. I guess Mozilla taskcluster deployments won't have this issue, but nice to think about other deployments. 👍

matt-boris (approved):
Nice Yarik! Thanks 👍🏻

PR #6532, opened 2023-09-07
Replace slow db query with the one that uses index

lotas:
`get_github_check_by_task_id` was only called with `task_id` which couldn't be looked up efficiently using index.
`get_github_check_by_task_group_and_task_id` uses both `task_group_id` and `task_id` which uses primary index for lookup.

According to GCP console, `get_github_check_by_task_id` was called about 50k times a day, with average response time being ~`800ms` - `1100ms`, which is pretty bad.

Here's `EXPLAIN` output for both:

**existing version**:

```
Gather  (cost=1000.00..95735.16 rows=1 width=68)
  Workers Planned: 2
  ->  Parallel Seq Scan on github_checks  (cost=0.00..94735.06 rows=1 width=68)
        Filter: (task_id = 'TASK_ID'::text)
```

**updated version**:

```
Index Scan using github_checks_pkey on github_checks  (cost=0.56..8.58 rows=1 width=68)
  Index Cond: ((task_group_id = 'TASK_GROUP_ID'::text) AND (task_id = 'TASK_ID'::text))
```

So in first case postgres will do scan on the full table, while the second one will use index to find row index and then fetch the index.

Tested on the prod db, queries are faster but still in a 100ms-200ms range (which is ~4-10 times faster)


matt-boris (approved):
Nice 👍🏻

lotas:
looks like I "killed" it 😄 
pulse turned into a straight line

image

PR #6533, opened 2023-09-07
Handle GCP preemption notifications

jschwartzentruber:
Add polling for GCP preemption notifications of spot and preemptible VMs. This allows graceful termination of tasks.

Github Bug/Issue: Fixes #6930


petemoore (approved):
This is awesome, thank you so much Jesse! ❤️

PR #6535, opened 2023-09-07
Upgrade CronJobs to use latest batch/v1 api

### LINKED ISSUE #6534: Upgrade k8s cronjob api version
Our cronjobs are using `apiVersion: batch/v1beta1` version which is outdated and blocks k8s cluster from being auto-upgraded to `1.25`

The correct current api version for cron is `batch/v1` according to [docs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/)

lotas:
This is also preventing production clusters from being upgraded to 1.25

Fixes #6534

### DIFF
--- a/changelog/issue-6534.md
+++ b/changelog/issue-6534.md
@@ -1,0 +1,6 @@
+audience: deployers
+level: patch
+reference: issue 6534
+---
+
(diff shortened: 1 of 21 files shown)


matt-boris (approved):
Incredible!! 🎉 😉

Issue #5514, opened 2022-06-14
Support Postgresql > 11


mostlygeek:
PostgreSQL 11 is EOL Nov/2023.   This issue is to track work updating Taskcluster to support PostgreSQL > 11.

matt-boris:
More information here https://www.postgresql.org/support/versioning/



Version | Current minor | Supported | First Release | Final Release
-- | -- | -- | -- | --
14 | 14.3 | Yes | September 30, 2021 | November 12, 2026
13 | 13.7 | Yes | September 24, 2020 | November 13, 2025
12 | 12.11 | Yes | October 3, 2019 | November 14, 2024
11 | 11.16 | Yes | October 18, 2018 | November 9, 2023

PR #6536, opened 2023-09-07
chore(db): allow for pg v15

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/5514.

Note, once this lands, we will upgrade our dev cluster's db to pg15 and ensure things are working. From there, we'll do community tc, then let SREs know about the ending support of v11. At that time, it might be best for them to make a copy of the pg11 db, and perform the upgrade to v15, just to see what the expected downtime might be while running the upgrade on Firefox CI.

>Adds support for postgres version 15.
>
>Note: if you want to migrate your local dev db to pg15, you'll need to either erase the existing db with `docker volume rm taskcluster_db-data` before you migrate, or, if you'd prefer to keep your local dev data, you'll need to manually dump the db contents and then import them into the upgraded db.
>
>Support for postgres v11 will be dropped from Taskcluster on November 9, 2023 (v11 EoL date) and that will be a breaking change.

### DIFF
--- a/changelog/issue-5514.md
+++ b/changelog/issue-5514.md
@@ -1,0 +1,9 @@
+audience: deployers
+level: minor
+reference: issue 5514
+---
+Adds support for postgres version 15.
+
+Note: if you want to migrate your local dev db to pg15, you'll need to either erase the existing db with `docker volume rm taskcluster_db-data` before you migrate, or, if you'd prefer to keep your local dev data, you'll need to manually dump the db contents and then import them into the upgraded db.
+
+Support for postgres v11 will be dropped from Taskcluster on November 9, 2023 (v11 EoL date) and that will be a breaking change.
--- a/db/README.md
+++ b/db/README.md
@@ -270,5 +270,5 @@ This approach is slower, but is appropriate for testing.
 
 ## Development
 
-To test this library, you will need a Postgres database, running the latest release of Postgres 11.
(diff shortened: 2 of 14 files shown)


lotas:
I think we should drop a note somewhere saying that local development will be affected by this change. Existing docker volume would be incompatible with new pg15.
One way is to `docker volume rm taskcluster_db-data` before starting to erase existing database. 

If there's something worth migrating, best option is to manually dump contents before upgrade and then importing after pg15 will be running

changelog/issue-5514.md
```diff
@@ -0,0 +1,7 @@
+audience: deployers
+level: minor
```
lotas, line 2: maybe major since it will break local db?

lotas (approved):
nice 👍 
thanks!

changelog/issue-5514.md
matt-boris: I'm not sure.. I'm tempted to keep it as minor since this PR still allows for v11 to be running. It only breaks the local db if they migrate their db to v15. Wdyt?

lotas: sure, i'm not insisting on major. But adding a note here in changelog would be good 👍

lotas (approved):
awesome, thank you!

matt-boris:
Post-upgrade steps https://cloud.google.com/sql/docs/postgres/upgrade-major-db-version-inplace#complete_the_major_version_upgrade

PR #6538, opened 2023-09-08
Provide a script to remove all unused taskcluster images locally

lotas:
I realised that we do releases very often, and each new release bring new image version. 
Over time they pile up and may take 10gb+ of disk space on local development machine.

This is only for the images that are referenced in `docker-compose.yml`


matt-boris (approved):
👍🏻

PR #6539, opened 2023-09-11
Add missing event from github documentation

lotas:
thanks @Eijebong for pointing out

### DIFF
--- a/changelog/eh3jyK_FSMWR2U7RIs-M8w.md
+++ b/changelog/eh3jyK_FSMWR2U7RIs-M8w.md
@@ -1,0 +1,4 @@
+audience: developers
+level: patch
+---
+Updates Github integration documentation to tell which events to subscribe to.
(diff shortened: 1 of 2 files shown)


changelog/eh3jyK_FSMWR2U7RIs-M8w.md
```diff
@@ -0,0 +1,3 @@
+audience: developers
+level: silent
```
Eijebong, line 2: This probably shouldn't be silent in case people need to modify their configuration if they only read the documentation? Maybe add a note?

petemoore (dismissed):
Looks great! Agree with @Eijebong though that it probably shouldn't be silent. Just a word or two to say that Github repo owners should subscribe to Check Run events if they want to be able to rerun failed checks directly from Github UI. Thanks Yarik!

changelog/eh3jyK_FSMWR2U7RIs-M8w.md
lotas: true! Will add, thanks

matt-boris (approved):
Thanks Yarik!

Issue #6543, opened 2023-09-11
d2g: look into loading environment values by key from the host


matt-boris:
Use the `-e VAR` notation documented [here](https://docs.docker.com/engine/reference/commandline/run/#env).

PR #6544, opened 2023-09-11
fix(d2g): only pass env var name

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6543. Thanks for the suggestion @jschwartzentruber!

>Generic Worker: d2g no longer passes the environment variable values to the `podman run` command. Instead, just the variable name is passed as `-e VAR` which tells podman to take the value from the host. This will tidy up the run command and will help with any escaping issues users may have been having.

### DIFF
--- a/changelog/issue-6543.md
+++ b/changelog/issue-6543.md
@@ -1,0 +1,5 @@
+audience: general
+level: patch
+reference: issue 6543
+---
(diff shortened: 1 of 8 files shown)


lotas (approved):
nice! looks much cleaner now 👍

petemoore (approved):
Nice job, thanks Boris!

PR #6545, opened 2023-09-12
Mount libraries in dev mode for migrations

lotas:
When db layer has changed in main branch but not released yet, it has one issue that `yarn start pg_init_db` or `yarn dev:start pg_init_db` may not work. Compose references pg15, but library only allows pg11, makes db initialization to fail.

Steps to fix it (removing old data!)

        docker volume rm taskcluster_db-data
        yarn dev:start postgres pg_init_db

This way we ensure that migration task has recent library changes and is able to apply all migrations.


matt-boris (approved):
Nice, thanks Yarik!

lotas:
also discovered that in order to have `yarn generate` working properly you need to have `pg_dump` compatible with latest db, so 15+
on mac solved by 

```
brew tap homebrew/core
brew install postgresql@15
```

PR #6548, opened 2023-09-13
Issue 6470: D2G now properly escapes artifact path in podman cp commands

### LINKED ISSUE #6470: d2g: incorrect escaping of arguments
D2g does not correctly escape arguments. For example, with this task payload:

```yaml
payload:
  artifacts:	
    public/build.tar.gz:
      path: /etc/passwd' artifact0; whoami; echo foo > /root/bar; cp 'foo
      expires: '2024-05-28T16:12:56.693817Z'
      type: file
```

you would be able to escape the podman container. Note, d2g does not provide a security boundary, so escaping the podman container is not a security concern (you do not even need to run payloads inside a container if you do not want to). Despite not a security concern, it is of course still a bug and incorrect behaviour.

Caused by:
```go
fmt.Sprintf("podman cp '%s:%s' %s", containerName, dwPayload.Artifacts[gwArtifacts[i].Name].Path, gwArtifacts[i].Path))
```

Obviously we need to properly escape arguments here. We should check all the escaping/encoding of d2g, and maybe even consider running a fuzzer against it.

Credit: @Eijebong reported this

petemoore:
Fixes #6470

### DIFF
--- a/changelog/issue-6470.md
+++ b/changelog/issue-6470.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 6470
+---
+D2G now properly escapes artifact paths in generated `podman cp` commands.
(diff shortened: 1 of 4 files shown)


lotas (dismissed):
👍

matt-boris (approved):

matt-boris (dismissed):

matt-boris (approved):

PR #6549, opened 2023-09-13
feat(d2g): use gw indexed artifact type

matt-boris:
>D2G now takes advantage of Generic Worker Indexed Artifacts, introduced in Generic Worker 51.0.0. D2G translates Indexed Docker Images into Generic Worker mounted Indexed Artifacts. Previously, D2G generated commands to query the taskcluster Index and fetch the docker image.
>With this improvement, docker images are now cached on workers, docker image dependencies between tasks are declarative (and thus inspectable), and generated Generic Worker task payloads are simpler and easier to understand.

### DIFF
--- a/tools/d2g/d2gtest/testdata/testcases/docker_image_tests.yml
+++ b/tools/d2g/d2gtest/testdata/testcases/docker_image_tests.yml
@@ -62,15 +62,16 @@ testSuite:
           - - bash
             - '-cx'
             - >-
-              curl -fsSL -o path "${TASKCLUSTER_PROXY_URL}/index/v1/task/test.namespace/artifacts/test/path"
-
-              IMAGE_NAME=$(podman load -i path | sed -n '1s/.*: //p')
-
               podman run -t --rm
               -e RUN_ID -e TASKCLUSTER_ROOT_URL -e
               TASKCLUSTER_WORKER_LOCATION -e TASK_ID
-              "${IMAGE_NAME}" 'echo "Hello world"'
+              docker-archive:path 'echo "Hello world"'
         maxRunTime: 3600
+        mounts:
+          - content:
+              artifact: test/path
+              namespace: test.namespace
+            file: path
         features:
           taskclusterProxy: true
         onExitStatus:
@@ -92,19 +93,16 @@ testSuite:
           - - bash
             - '-cx'
             - >-
-              curl -fsSL -o path.lz4 "${TASKCLUSTER_PROXY_URL}/index/v1/task/test.namespace/artifacts/test/path.lz4"
-
-              unlz4 'path.lz4'
-
-              rm 'path.lz4'
-
-              IMAGE_NAME=$(podman load -i path | sed -n '1s/.*: //p')
-
               podman run -t --rm
               -e RUN_ID -e TASKCLUSTER_ROOT_URL -e
               TASKCLUSTER_WORKER_LOCATION -e TASK_ID
-              "${IMAGE_NAME}" 'echo "Hello world"'
+              docker-archive:path.lz4 'echo "Hello world"'
         maxRunTime: 3600
+        mounts:
+          - content:
+              artifact: test/path.lz4
+              namespace: test.namespace
+            file: path.lz4
         features:
           taskclusterProxy: true
         onExitStatus:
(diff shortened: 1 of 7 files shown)


tools/d2g/dockerimageartifact.go
```diff
@@ -38,5 +28,5 @@ func (dia *DockerImageArtifact) FileMounts() ([]genericworker.FileMount, error)
 }
 
 func (dia *DockerImageArtifact) String() (string, error) {
-	return `"${IMAGE_NAME}"`, nil
+	return fmt.Sprintf("docker-archive:%s", filepath.Base(dia.Path)), nil
```
petemoore, line 31:
```suggestion
	return "docker-archive:dockerimage"
```

Elsewhere, rather than attempting to preserve file names, which indeed might clash, we have chosen to used fixed names, chosen by d2g. For example artifacts get the names `artifact0`, `artifact1`, .... These names cannot conflict, because they are named by d2g, and the file names in the original tasks have no bearing over the file names created on the host, only in the container. This works since d2g chooses names which do not conflict with each other, but can break if a file can be created on the host where the name is arbitrary and chosen by the task.

tools/d2g/indexeddockerimage.go
```diff
@@ -1,59 +1,31 @@
...
+	}
+	raw, err := json.MarshalIndent(&indexedContent, "", "  ")
+	if err != nil {
+		return nil, fmt.Errorf("cannot marshal indexed content %#v into json: %w", indexedContent, err)
+	}
+	return []genericworker.FileMount{
+		{
+			Content: json.RawMessage(raw),
+			// TODO check if this could conflict with other files(?)
+			File: filepath.Base(idi.Path),
```
petemoore, line 24:
```suggestion
			File: "dockerimage",
```

Note, below I propose supporting lz4 compression in mounted Read Only Directories and mounted Writable Directory Caches so that the mounts feature automatically performs the lz4 decompression. If we follow this approach, here we would need to check if this is a `.lz4` file, and if so, set `Directory` instead of `File`, and set `Format` to `"lz4"`. This would cause the mounts feature to decompress the file. Presumably the lz4 decompression simply strips the file `.lz4` filename suffix, and the compressed file does not have the ability to choose the target file name.

We should probably do a review of the compression formats we support, to see which are pure compression, and which are a combination of archiving and compression. I believe lz4 is pure compression, and no archiving. Pure compression probably makes sense also for file artifacts, whereas archive formats probably only make sense for mounted directories. Currently I think we don't support any compression formats for file artifacts.

tools/d2g/dockerimageartifact.go
matt-boris: Nice, thanks for the great idea Pete!

matt-boris:
Leaving in draft until https://github.com/taskcluster/taskcluster/issues/6553 is complete.

tools/d2g/d2gtest/testdata/testcases/docker_image_tests.yml
```diff
@@ -62,15 +62,16 @@ testSuite:
...
             - >-
-              curl -fsSL -o path "${TASKCLUSTER_PROXY_URL}/index/v1/task/test.namespace/artifacts/test/path"
-
-              IMAGE_NAME=$(podman load -i path | sed -n '1s/.*: //p')
-
               podman run -t --rm
               -e RUN_ID -e TASKCLUSTER_ROOT_URL -e
               TASKCLUSTER_WORKER_LOCATION -e TASK_ID
-              "${IMAGE_NAME}" 'echo "Hello world"'
+              docker-archive:dockerimage 'echo "Hello world"'
```
lotas, line 68: Is this a hardcoded name for all images? 
If worker is configured to run multiple tasks concurrently, wouldn't it cause a collision if two tasks would request two different indexed images, but the registry on the host is the same? 
Or is there some sort of isolation happening for this scenario?

matt-boris: Good eye. Yes, this is hardcoded for all images. Generic worker's capacity is ever 1, though. So it doesn't and can't run tasks concurrently.

petemoore: When (if ever) https://bugzilla.mozilla.org/show_bug.cgi?id=1609102 is fixed, it will be possible to run multiple concurrent tasks under multiuser engine. However, since the `podman run` commands are executed as the task user, I suspect that the cached images and tags will be stored on a per-user basis, rather than a global cache on the host. In other words, I suspect that if OS user A tags a docker image with the name `dockerimage`, and OS user B tags a different docker image with the name `dockerimage`, that they won't conflict. But we should check this... Nice spot, @lotas!

lotas (dismissed):
👍

tools/d2g/indexeddockerimage.go
```diff
@@ -1,59 +1,49 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot marshal indexed content %#v into json: %w", indexedContent, err)
+	}
+	return []genericworker.FileMount{
+		{
+			Content: json.RawMessage(raw),
+			// Instead of trying to preserve the artifact filename, we use a
+			// hardcoded name to prevent filename collisions.
+			// This _may_ cause issues once concurrent tasks are supported
+			// on generic worker (see https://bugzil.la./1609102).
```
petemoore, line 27:
```suggestion
			// on generic worker (see https://bugzil.la/1609102).
```

petemoore (dismissed):
Wonderful!

Issue #2940, opened 2020-05-29
Migrate queue's queue to a more visible form


djmitche:
The queue uses tc-lib-azqueue for its core "queueing" operations.  This supports both queues for tasks for workers, and delay queues for things like deadlines and task claims.

Because it mixes so many things in there, this queue is not easily introspected to find out, say, what tasks are pending for a particular task queue.

Let's find a way to do task queueing that's linked to the tasks table in a way that allows querying (#2939) the tasks in the queue.  The `pendingTasks` API method should be upgraded in the process to take account of cancelled tasks (https://bugzilla.mozilla.org/show_bug.cgi?id=1434851).

Requires #2932.

djmitche:
Note that postgres supports NOTIFY/LISTEN, allowing us to avoid polling (or for safety's sake, polling very infrequently) the DB when nothing is changing.

lotas:
current thought progress documented in this gist https://gist.github.com/lotas/57779afe067e4f8eb586f37fe25016c9

lotas:
Execution of this transition:

- [x] Part 1: prepare `azure_queue_messages` table #6550 
- [ ] Part 2: migrate data from `azure_queue_messages` into separate tables, keep things running #6554 
- [ ] Part 3: add "listen" functionality to avoid polling, drop `azure_queue_messages`

after that #2939 will add new functionality to introspect queues and change task priorities

lotas:
I've extracted that last piece into as separate ticket and will close this one https://github.com/taskcluster/taskcluster/issues/6950

PR #6550, opened 2023-09-14
Prepare azure_queue_messages for upcoming migration.

lotas:
As it stands, azure_queue_messages table doesn't contain name of the task queue or priority of the task. Payload is also a base64 encoded string which makes it harder to extract necessary information.

In following release we would introduce extra tables to split messages functionality into following tables:
1. queue_task_deadlines
2. queue_resolved_tasks
3. queue_pending_tasks
4. queue_claimed_tasks

To simplify live migration of existing records into those tables we would need some information in place, thus adding task queue id in clear form and also payload in jsonb which would allow to extract data

Migration strategy is outlined in https://gist.github.com/lotas/57779afe067e4f8eb586f37fe25016c9

Related to #2940 

---

I tested it locally, tasks are being created, necessary fields appears. Background tasks run without errors - claim resolver, dependency resolver, deadline resolver

Going to work on the second part, to test that those fields are enough.


db/versions/0090.yml
```diff
@@ -0,0 +1,78 @@
+version: 90
+description: prepare queue for upcoming migration and refactoring
+migrationScript: |-
+  begin
+    alter table azure_queue_messages
+      add column task_queue_id text null, -- expose cryptic names
+      add column priority int null, -- don't forget about the priority
+      add column payload jsonb null; -- instead of base64 encoded string use jsonb
```
matt-boris, line 8: Only thing going around in my head here is total size the jsonb is using up. If you don't think it'll be an issue, then np from me.

matt-boris (approved):
Changes LGTM. Just added one comment about using the new jsonb column.

db/versions/0090.yml
lotas: i think should be fine, shouldn't take much more space than base64 encoded json that is used currently.
plus those new fields are not used in fetching, so wouldn't affect existing querying 👍 

but later it would be much easier to extract columns during second phase - https://github.com/taskcluster/taskcluster/pull/6554/files 

As I'm still playing with that part, I might discover that jsonb is not necessary, then I'll drop it here. 
The most problematic part is "cryptic" queue names for taskQueueId

lotas: alright, I think I'm going to drop this afterall, thanks for making me rethink this :) 

I can also do 
```
select 
queue_name, 
convert_from(decode(message_text, 'base64'), 'utf-8')::jsonb->>'taskId',
convert_from(decode(message_text, 'base64'), 'utf-8')::jsonb->>'taskGroupId',
convert_from(decode(message_text, 'base64'), 'utf-8')::jsonb->>'schedulerId'
from azure_queue_messages
limit 100;
```
so no need to keep intermediary column 👍

matt-boris (approved):

PR #6554, opened 2023-09-14
Queue internals refactoring Part 2/3

lotas:
Migrating internal queue structures from single azure-like table into separate tables. 

This migration should be backwards-compatible and allow db downgrade to v91 by storing extra columns, that would be removed in next migration. Existing `azure_*` methods are re-purposed to use new tables to avoid data loss at migration time

**Goals** of this PR:

- get rid of `azqueue` library
- get rid of single `azure_queue_messages`  table that holds four different kinds of information (task deadlines, tasks that were claimed, tasks that were resolved and tasks that are pending)
- build a foundation for new queue features (changing priority of a task, seeing what tasks are currently pending)

This is Part 2 of a 3 step migration plan:

- [x] Part 1: prepare `azure_queue_messages` table #6550 
- [ ] Part 2: migrate data from `azure_queue_messages` into separate tables, keep things running #6554 
- [ ] Part 3: add "listen" functionality to avoid polling, drop `azure_queue_messages`

## Changes

- `taskcluster-lib-azqueue` removed
- `azure_queue_messages` table is split into 4
- existing data is migrated into new tables
- update existing code to use new structure

## Risks

- some messages will be "lost", as migration copies non-expired only `expires > now()` so if cron didn't remove those before, they will be lost. Those are not relevant as they wouldn't be fetched anyway
- some events might be delivered twice, once for the services that are running during migration, and new updated services after data migrated to new tables. Worst case some **tasks could run twice** although highly unlikely as there are mechanisms in place to avoid serving same task to two different workers. 

## Migration details

Tested locally with prod db data, takes less than a second to migrate ~170k records.

1. `azure_queue_messages` table is locked in exclusive mode, which should prevent existing clients writing and reading from it
2. `queue_task_deadlines` table created and all messages for `"deadline-queue"` are copied into this table
3. `queue_resolved_tasks` table created and all messages for `"resolved-queue"` are copied into this table
4. `queue_claimed_tasks` table created and all messages for `"claim-queue"` are copied into this table 
5. indexes are being added to the tables
6. existing data is removed from `azure_queue_messages` to avoid running services from picking up "migrated" messages
7. `azure_queue_messages` and all related functions would be dropped on next release

Part 2 of #2940 

### To-do:

- [x] migrate data into new tables
- [x] define queue pending tasks functions
- [x] define deadline queue functions
- [x] define claimed queue functions
- [x] define resolved queue functions
- [x] update tests if needed
- [x] put indexes where needed
- [x] add tests for new db functions
- [x] document queue internals


db/test/fns/queue_test.js
```diff
@@ -10,74 +10,218 @@ const taskcluster = require('taskcluster-client');
...
+      await db.fns.queue_pending_tasks_put('tq1', 9, 'task1', 0, 'hint1', fromNow('10 seconds'));
+      await db.fns.queue_pending_tasks_put('tq1', 3, 'task2', 0, 'hint1', fromNow('10 seconds'));
+      await db.fns.queue_pending_tasks_put('tq1', 1, 'expiredTask', 0, 'hint1', fromNow('0 seconds'));
+      assert.deepEqual(
+        await db.fns.queue_pending_tasks_count("tq1"),
+        [{ queue_pending_tasks_count: 2 }],
+      );
+    });
+
+    // TODO: This requires support from the taskcluster-lib-postgres library
```
lotas, line 35: I will do this in a follow up PR, to keep this strictly to the data migration and not build too many new features on top

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,166 @@
+begin
+  -- migration is only possible when all tasks were migrated to new column structure
+  -- following query is commented out for documentation purposes, to prevent failed migration on prod deployment
```
lotas, line 3: data is already ready on prod, so should be safe

db/versions/0091.yml
```diff
@@ -0,0 +1,391 @@
...
+  azure_queue_delete:
+    deprecated: true
+  azure_queue_update:
+    deprecated: true
+  azure_queue_put_extra:
+    deprecated: true
+  azure_queue_delete_expired:
+    deprecated: true
+
+  queue_pending_tasks_get:
```
lotas, line 22: there are 4 `queue_*_get` methods that are doing the same that `azure_queue_get` was doing - selecting rows with `FOR UPDATE SKIP LOCKED` to guarantee that rows would only be fetched once.

db/versions/0091.yml
```diff
@@ -0,0 +1,391 @@
...
+            priority = priority_in,
+            run_id = greatest(run_id_in, queue_pending_tasks.run_id),
+            hint_id = hint_id_in
+          WHERE
+            queue_pending_tasks.task_queue_id = task_queue_id_in
+            AND queue_pending_tasks.task_id = task_id_in
+            AND queue_pending_tasks.pop_receipt is null;
+
+        -- notify listeners that there is a new task in the queue
+        EXECUTE 'NOTIFY task_pending, ' || quote_literal(task_queue_id_in) || ';';
```
lotas, line 89: this will be used in followup PR. Listener would be added in `WorkClaimer` to listen to notifications and when `taskQueueId` in the payload will match listened one, claimer would fetch new tasks using existing functions.

I think that not sending full payload is fine, because we wouldn't be able to use that information directly anyway. And there might be zero listeners at the time. This is just to inform that new task is avaliable for the `taskQueueId`

db/versions/0091.yml
```diff
@@ -0,0 +1,391 @@
...
+          task_group_id_in,
+          task_id_in,
+          scheduler_id_in,
+          resolution_in,
+          now(),
+          now()
+        );
+
+        -- notify listeners that task was resolved
+        EXECUTE 'NOTIFY task_resolved, ' || quote_literal(task_id_in) || ';';
```
lotas, line 346: same for resolved. `DependencyResolver` would be able to know there is a new task resolved and fetch the details right after

services/queue/config.yml
```diff
@@ -34,25 +34,15 @@ defaults:
...
-
     # Number of ms before deadline expiration message arrives, past deadline
     deadlineDelay:                60000
 
     # Configuration for deadline-resolver
     deadlineResolver:
       # Number of ms to sleep after polling from an empty queue
       pollingDelay:               5000
+      # Number of records to process at once
+      count:                      32
```
lotas, line 45: this was the default from azqueue times, maybe it would be safe to increase this number.

services/queue/src/hintpoller.js
```diff
@@ -0,0 +1,126 @@
+let assert = require('assert');
```
lotas, line 1: This class is extracted from `workerclaimer.js` file for now.

services/queue/src/queueservice.js
```diff
@@ -47,204 +36,54 @@ const PRIORITIES = [
...
-      // Don't cache negative results
-      this.deadlineQueueReady = null;
-      throw err;
-    });
-    return this.deadlineQueueReady = ready;
   }
 
   /** Enqueue message to become visible when claim has expired */
-  async putClaimMessage(taskId, runId, takenUntil) {
+  async putClaimMessage(taskId, runId, takenUntil, taskQueueId, workerGroup, workerId) {
```
lotas, line 70: this class is only being adjusted to new functions after azqueue library was removed

services/queue/src/workclaimer.js
```diff
@@ -151,21 +38,29 @@ class WorkClaimer extends events.EventEmitter {
     this._hintPollers = {}; // provisionerId/workerType -> HintPoller
   }
 
+  getHintPoller(taskQueueId) {
```
lotas, line 41: this is extracted from `claim()` while loop. Not sure why it was done inside the loop anyway 🤷

services/queue/src/workclaimer.js
```diff
@@ -151,21 +38,29 @@ class WorkClaimer extends events.EventEmitter {
...
+    if (!this._hintPollers[taskQueueId]) {
+      this._hintPollers[taskQueueId] = new HintPoller(taskQueueId, {
+        monitor: this._monitor,
+        pollPendingQueue: async (count) => this._queueService.pollPendingQueue(taskQueueId)(count),
```
lotas, line 45: `HintPoller` has single dependency `monitor` now. 

It might be that we don't even need this class anymore (at least in its current form). It will likely be rewritten in upcoming PRs to use `listen/notify`

services/queue/src/workclaimer.js
```diff
@@ -228,7 +123,7 @@ class WorkClaimer extends events.EventEmitter {
     // put the claim-expiration message into the queue first.  If the
     // subsequent claim_task fails, the claim-expiration message will be
     // ignored when it appears.
-    await this._queueService.putClaimMessage(taskId, runId, takenUntil);
+    await this._queueService.putClaimMessage(taskId, runId, takenUntil, task.taskQueueId, workerGroup, workerId);
```
lotas, line 126: storing extra info in claim table, so we could also answer those questions in future:
- how many tasks are currently running for task queue, worker group, worker

services/queue/test/queueservice_test.js
```diff
@@ -174,11 +172,9 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
 
     // Poll message again
     message = await testing.poll(async () => {
-      for (let i = 0; i < poll.length; i++) {
-        let messages = await poll[i](1);
-        if (messages.length === 1) {
-          return messages[0];
-        }
+      let messages = await poll(1);
```
lotas, line 175: before `queueService.pendingQueues` returned list of functions, one per each priority, to ask iteratively starting from highest to lowest. 
We no longer need this, since single function can return tasks ordered by priority already

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,166 @@
...
+    pop_receipt
+  FROM azure_queue_messages
+  WHERE queue_name = 'resolved-queue'
+  AND expires > now();
+
+  CREATE INDEX queue_resolved_task_idx ON queue_resolved_tasks (task_id);
+  CREATE INDEX queue_resolved_task_vis_idx ON queue_resolved_tasks (visible);
+
+  -- Claimed Tasks
+  -- queries: what tasks are running and waiting to be reclaimed or resolved
```
djmitche, line 79: This is good, and I believe visibility here is based on the reclaim time?

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,166 @@
...
+    task_queue_id text not null,
+    worker_group text not null,
+    worker_id text not null,
+    claimed timestamptz not null,
+    taken_until timestamptz not null,
+    visible timestamptz not null,
+    pop_receipt uuid null
+  );
+
+  -- migrate data to resolved queue
```
djmitche, line 92:
```suggestion
  -- migrate data to claimed queue
```

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,166 @@
...
+  INSERT INTO
+    queue_claimed_tasks (task_id, run_id, task_queue_id, worker_group, worker_id, claimed, taken_until, visible, pop_receipt)
+  SELECT
+    convert_from(decode(message_text, 'base64'), 'utf-8')::jsonb->>'taskId',
+    CAST(convert_from(decode(message_text, 'base64'), 'utf-8')::jsonb->>'runId' AS INTEGER),
+    'unknown/tq',
+    'unknonwn-wg',
+    'unknonwn-wi',
```
djmitche, line 100: There are typos in the last two of these (a typo I make all the time!)

It would probably slow down the migration to do a join with the `tasks` table to fetch these values, and I suppose wouldn't add much value. Maybe add a comment describing the temporary impact of using these dummy values?

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,166 @@
...
+  WHERE queue_name = 'claim-queue'
+  AND expires > now();
+
+  CREATE INDEX queue_claimed_task_idx ON queue_claimed_tasks (task_id, run_id);
+  CREATE INDEX queue_claimed_task_vis_idx ON queue_claimed_tasks (visible);
+  CREATE INDEX queue_claimed_task_queue_idx ON queue_claimed_tasks (task_queue_id, worker_group, worker_id);
+
+
+  -- Pending (scheduled)
+  -- purpose: keep the record of all tasks that were scheduled and claimed
```
djmitche, line 115: I think claimed tasks aren't in this queue? Looking at `workclaimer.js`, the rows from this table (hints) are `.remove()`d when the task is actually claimed.

The comment at the top of that file describes this table quite nicely, actually: a task appears in this table if a tsk is pending, but a non-pending task _may_ also appear in this table.

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,166 @@
...
+  GRANT select, insert, update, delete ON queue_resolved_tasks to $db_user_prefix$_queue;
+  GRANT select, insert, update, delete ON queue_claimed_tasks to $db_user_prefix$_queue;
+
+  -- delete old data
+  -- or delete whole table? will there be workers still running at the time of migration?
+  -- maybe it's better to leave the table as is to let remaining workers finish what was started
+  -- and in worst case new workers might restart few tasks
+  -- other queues should not suffer much, as deadline/resolved/claims are handling errors
+  -- table can be dropped in a followup migration on next release
+  -- DELETE FROM azure_queue_messages;
```
djmitche, line 165: The migration should update the stored functions in the same transaction, so that once the transaction commits, nothing references the old table. So, it should be safe to delete it.

db/versions/0091-downgrade.sql
```diff
@@ -0,0 +1,6 @@
+begin
+  drop table queue_pending_tasks;
+  drop table queue_claimed_tasks;
+  drop table queue_resolved_tasks;
+  drop table queue_task_deadlines;
```
djmitche, line 5: This is a bit of a cheat -- if, for some reason, the migration succeeds but there's some issue that leads you to downgrade an installation, that installation's queues will be out of sync with the task tables. As written, the queues will contain "stale" data from the moment the upgrade transaction committed, while the task tables will contain up-to-date information about tasks. So for example you might have pending tasks for which there's no entry in the pending queue.

I know it seems like a nightmare of SQL that is unlikely to ever be needed, but I think it's a good idea to write a full downgrade process that reconstructs `azure_queue_messages` from these four tables, using an exclusive table lock just like `0091-migration.sql`.

db/versions/0091.yml
```diff
@@ -0,0 +1,391 @@
...
+  azure_queue_count:
+    deprecated: true
+  azure_queue_delete:
+    deprecated: true
+  azure_queue_update:
+    deprecated: true
+  azure_queue_put_extra:
+    deprecated: true
+  azure_queue_delete_expired:
+    deprecated: true
```
djmitche, line 20: This feels like cheating a little, too -- any existing JS code still calling these functions will access the old table, potentially for example inserting a pending task, which new JS code will not see. And, this cheat really won't work if the old table is deleted! I suspect it's worthwhile re-implementing these functions in terms of the new queues. That might be difficult with the pending queues, though, since they are named after the taskQueueId. Maybe a not-quite-as-good solution is to just make all of these functions either return an empty set or fail, so that any old JS just fails and retries until the new JS starts running.

db/versions/0091.yml
djmitche: I'm not sure the taskId will be useful here -- this should just be a trigger for the DependencyResolver to stop sleeping and do its loop, rather than look for a specific taskId (and possibly miss other taskIds that were NOTIFY'd when it wasn't LISTENing). I suppose it doesn't hurt to include the taskId even if it's not used.

services/queue/README.md
```diff
@@ -21,3 +21,91 @@ For S3 we have a dummy bucket called `test-bucket-for-any-garbage` which stores
...
+
+```mermaid
+sequenceDiagram
+  participant client
+  participant queue as Queue API
+  participant deadlineQueue as Deadline Queue
+  participant pendingQueue as Pending Queue
+
+  client->>+queue: createTask()
+  queue-->>deadlineQueue: QueueService.putDeadlineMessage()
```
djmitche, line 48: The `alt` here makes it look like _either_ a deadline message or a pending message is enqueued, but in fact a deadline message is always enqueued. Maybe I'm just misreading the diagram?

services/queue/README.md
```diff
@@ -21,3 +21,91 @@ For S3 we have a dummy bucket called `test-bucket-for-any-garbage` which stores
...
+  alt is scheduled:
+    queue-->>pendingQueue: if task can be scheduled<br> QueueService.putPendingMessage()
+  end
+
+  queue->>-client: task status response
+```
+
+### Worker claiming work
+
+Workers would be calling `queue.claimWork()` to get tasks from specific `taskQueueId`.
```
djmitche, line 59: It might be useful to include reclaimTask in this diagram, as it "bumps" the visibility in the claim queue.

services/queue/config.yml
djmitche: It's probably not worthwhile -- the deadline resolver gets this many messages and then handles them with `Promise.all`, so it's basically defining the concurrency of a single resolver instance. That concurrency is probably already limited by the maximum number of DB connections. I suspect that if the deadline resolver is falling behind, `parallelism` is already sufficient to adjust this concurrency.

Anyway, adding this as a config.yml value doesn't hurt, but I wouldn't suggest making it configurable with an env var.

djmitche (commented):
I've looked at everything except `services/queue/{src,test}`. I'll find more time to look soon, but I thought this much would be useful to you now.

db/versions/0091-downgrade.sql
lotas: Great point! 

I think it's easy to put deadline/claim/resolve back, but for pending you need to create queue name which is bit more tricky, but I agree, data is lost on downgrade.
Will add 👍

db/versions/0091-migration.sql
lotas: oh, indeed!
Then I will live it not commented, to give fail corrupting pending queues for deployments that will try to do run few upgrades at once. 
I'll also live a comment in changelog to say that this depends on successful migration from the previous release 👍

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,166 @@
...
+  -- ) THEN
+  --   RAISE EXCEPTION 'Not possible to migrate, some records are still in old format';
+  -- END IF;
+
+  -- migration of data
+  -- prevent reads and writes from table, get exclusive access
+  LOCK TABLE azure_queue_messages;
+
+  -- Task Deadlines
+  -- purpose: know when particular task expires, so dependencies/task groups can be resolved
```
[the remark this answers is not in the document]
lotas: thanks! I will update the comment. It was coming from the pre-refactoring draft queries I made to test it and forgot to update 👍

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,166 @@
...
+  CREATE TABLE queue_task_deadlines (
+    task_id text not null,
+    task_group_id text not null,
+    scheduler_id text not null,
+    created timestamptz not null,
+    deadline timestamptz not null,
+    visible timestamptz not null,
+    pop_receipt uuid null
```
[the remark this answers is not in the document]
lotas: Good call! Name migrated from azqueue indeed.

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,166 @@
...
+    pop_receipt
+  FROM azure_queue_messages
+  WHERE queue_name = 'deadline-queue'
+  AND expires > now();
+
+  CREATE INDEX queue_task_deadline_idx ON queue_task_deadlines (task_id);
+  CREATE INDEX queue_task_deadline_vis_idx ON queue_task_deadlines (visible);
+
+  -- Resolved Tasks
+  -- queries: what tasks are not scheduled for given task_group_id/scheduler_id
```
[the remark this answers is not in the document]
lotas: huh, silly copy-paste probably :) Thanks! Updating desription

db/versions/0091-migration.sql
lotas: Yes, this is the same value, until the records is fetched for processing. 
We could probably still use `taken_until` to act as `visible` column, and update it, to temporarily hide it during processing

db/versions/0091-migration.sql
lotas: I should install spell checker probably 🤦 

I wanted to add those fields for future uses like query all tasks for a given worker or task queue.. but this also can be added later also. 
I can also just drop those columns for now, since they are not used  yet

db/versions/0091-migration.sql
lotas: uh, sorry.. I didn't update comments from the initial draft before sending for review! Thanks!

db/versions/0091-migration.sql
lotas: So with this migration we are only marking old `azure_*` methods as deprecated. And we didn't drop functions ad deprecation time before, as far as I know. 

But since we also remove so much, maybe we should also drop existing methods?

db/versions/0091.yml
lotas: Yes, agree. As dependency resolver takes all tasks and only needs to know that there are new records available.

services/queue/README.md
lotas: uhm.. right.. I was not sure how to better depict this. Will update to show that that deadline message is always added 👍

services/queue/config.yml
lotas: Indeed! Maybe I should put it as a constant in code instead, and document the reasoning behind this number

db/versions/0091-migration.sql
djmitche: I figured that is what the fields were for. If they will be populated as new tasks are added, then it makes sense to include them now -- then by the time there's functionality to _read_the columns, they will have valid data in them.

db/versions/0091-migration.sql
djmitche: In this context, "deprecated" means "still exists and works as documented, but new code should not use it". As written, though, these functions will _not_ work as documented, since they address the old `azure_queue_messages` table.

Typically that's meant re-implementing the deprecated functions in terms of the new tables or columns. For example, when we changed from provisionerId/workerType to workerPoolId, the functions that referred to `provisioner_id` and `worker_type` were re-implemented to read the `worker_pool_id` column from the DB, split on `/`, and return the two parts as `provisioner_id` and `worker_type`. These deprecated functions only run during the time when _old_ code is running against a migrated DB.

Generally after a migration completes, the code is upgraded "soon" after, so the functions are only used for the duration of "soon". Maybe it's acceptable to have those deprecated functions just crash, to avoid the effort of re-implementing them in terms of the new tables. Maybe it makes sense for _some_ functions to just crash (e.g., those used by deadline resolver, which can crash-loop for a few minutes without any user-visible impact), while re-implementing others (e.g., putting messages in the pending queues, so that `queue.createTask` calls don't fail).

db/versions/0091.yml
djmitche: Oops, my last comment ("In this context, "deprecated" means ...") should probably have been here.

db/versions/0091-migration.sql
lotas: Thanks Dustin! 
Totally makes sense. Those functions would be useless during migration and would lead to data loss likely.
I will try to repurpose those for new tables if it would be possible.

services/queue/README.md
lotas: Oh! I missed to also check this flow :) Thanks again!

So the `reclaimTask` calls adds new record in claim queue. 
Which means that long running tasks could have multiple records in the claim queue table. 

We can make `(taskId, runId)` unique, and on conflict just update the `takenUntil`. Or delete existing record before.
There is no point in letting ClaimResolver process "dead" messages

lotas: ah, checked prod db, there are multiple records for the same task, so adding unique key at that time wouldn't help.

lotas: oh, interesting, some workers reclaim tasks more often, which results in some tasks having `4` claim records for the same `(taskId, runId)` 

Probably worth optimizing it by keeping only the latest one

lotas: ok, updated migration and function with unique taskId+runId

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,183 @@
...
+    inserted,
+    expires,
+    visible,
+    pop_receipt
+  FROM azure_queue_messages
+  WHERE queue_name NOT IN ('claim-queue', 'deadline-queue', 'resolved-queue')
+  AND task_queue_id IS NOT NULL
+  AND expires > now();
+
+  CREATE UNIQUE INDEX queue_pending_task_idx ON queue_pending_tasks (task_id);
```
lotas, line 167: @djmitche, @petemoore as I was reading older comments and docs I noticed that before it was allowed to have multiple  `(taskId, runId)` messages in the queue. 
With this change I'm enforcing "there can be only one run scheduled for any task".

Do you think there are any reasons or conditions under which several runs for the same task could be pending? 
I believe this was already enforced in `schedule_task` and `rerun_task` methods.

So I think it is fine that we always keep single record here for any task to let workers claim only the latest run

djmitche: Historically, queues have been "over-estimations", so for example a task being in the pending queue meant that the task _might_ actually be pending. But a task that was not in the pending queue was definitely not pending. This is a pretty common distributed-algorithm design, but was necessary with azure queues since it's impossible to delete a message from a queue - or know if a message exists in a queue - until that message is visible and comes to the top of the queue.

So I think a (task_id, run_id) could appear in the queue multiple times if there are some overlaps or failed operations, and might prevent retrying an idempotent operation. I can't think of a case in "normal operation" where that might occur. So preventing this would change the semantics of the queue and _might_ have undesirable consequences.

I think it's a judgement call on your part whether to change the semantics for some, or all, of the queues. And if you change it, whether to do so in this PR or a followup. The cost to changing this is complexity of the PRs and the possibility of bugs. Those bugs would be of the sort where a pending task never runs, or a worker failure does not result in `claim-expired`, or something like that -- the sort of bug that is rare and hard to track down.

If you choose to keep the existing queue semantics, this PR is already a big improvement: values are broken out into columns that can be manipulated with SQL. For example, if you allow that multiple pending messages for the same run of the same task can exist in the pending queue, you can still find all pending tasks for a taskQueueId with something like
```sql
SELECT * from queue_tasks JOIN (
  SELECT unique task_id
  FROM queue_pending_tasks
  WHERE task_queue_id = $taskQueueId
) as maybe_pending
WHERE queue_tasks.task_id = maybe_pending.task_id
AND queue_tasks.status = 'pending'     -- I forgot how this status is stored, but you get the idea
```

djmitche:
I mentioned this in the community meeting, but I hope it's clear here, too -- I have no stake in the decisions you're making here, and you should be the one to decide. Where I've given my opinion, you should feel free to decide otherwise!

lotas:
> I mentioned this in the community meeting, but I hope it's clear here, too -- I have no stake in the decisions you're making here, and you should be the one to decide. Where I've given my opinion, you should feel free to decide otherwise!

Thank you Dustin, your comments really help! And I mostly agree with them :)

db/versions/0091-migration.sql
lotas: This is helpful. I prefer not to change the semantics now as I don't have a complete understanding about the impacts. 

Agree, I'd keep it for now as it was before to reduce the risk of having unwanted side-effects during this migration.

services/queue/src/claimresolver.js
```diff
@@ -99,25 +103,18 @@ class ClaimResolver {
       }
     }));
 
-    // If there were no messages, back of for a bit.  This avoids pounding
-    // Azure repeatedly for empty queues, at the cost of some slight delay
-    // to finding new messages in those queues.
+    // If there were no messages, back of for a bit.
```
matt-boris, line 106: I see it was there beforehand too, but here's a quick fix :)

```suggestion
    // If there were no messages, back off for a bit.
```

services/queue/src/deadlineresolver.js
```diff
@@ -103,25 +108,18 @@ class DeadlineResolver {
       }
     }));
 
-    // If there were no messages, back of for a bit.  This avoids pounding
-    // Azure repeatedly for empty queues, at the cost of some slight delay
-    // to finding new messages in those queues.
+    // If there were no messages, back of for a bit.
```
matt-boris, line 111:
```suggestion
    // If there were no messages, back off for a bit.
```

services/queue/src/utils.js
```diff
@@ -233,3 +233,7 @@ const useOnlyTaskQueueId = (obj) => {
   delete obj.workerType;
 };
 exports.useOnlyTaskQueueId = useOnlyTaskQueueId;
+
+/** Sleep for `delay` ms, returns a promise */
+const sleep = (delay) => new Promise((accept) => setTimeout(accept, delay));
+exports.sleep = sleep;
```
matt-boris, line 239: Nice cleanup 👍🏻

services/queue/src/main.js
```diff
@@ -33,6 +33,10 @@ const DEFAULT_UPDATE_FREQUENCY = '30 minutes';
 // for supported bulk deletion S3 operations
 const MAX_BULK_DELETE_SIZE = 1000;
 
+// maximum number of records to process at once in claim, deadline, and dependency resolvers
+// this is to limit total amount of concurrent updates to DB
+const NUMBER_OF_RECORDS_TO_PROCESS = 32;
```
matt-boris, line 38: I'm wondering if we should make this configurable and default to 32 so that we can have finer control over it without having to make additional code changes. Wdyt?

matt-boris (commented):
Incredible efforts here Yarik, thank you! 💪🏻

Just left a few comments/questions and small/nit suggestions 👍🏻

matt-boris:
I'll drop the `library-azqueue` required CI check now too.

lotas:
> >   I think we can get rid of this column, and instead pass deadline < now() + deadlineDelay offset to avoid using this column 👍

> Are you still wanting to do this?

no, sorry, forgot to comment. `visible` column is needed in combination with `pop_receipt` to ensure that messages "disappear" during processing. Same with other tables. And if we change `deadline` we would lost the ability to check in `DeadlineResolver` if task was indeed expired or not

services/queue/src/main.js
lotas: Haha, this was initially in `config.yml` :) But it's more of a node<>pg library that has own limits on number of concurrent connections. And after Dustin's comments I moved it back to constant. I'll keep it as is, no need to modify on deployment level

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,179 @@
...
+  -- if someone upgrades db from 89 to 91 directly it would fail if there are pending tasks
+  -- this is to prevent data loss
+  IF EXISTS (
+    SELECT message_id FROM azure_queue_messages
+    WHERE queue_name NOT IN ('claim-queue', 'deadline-queue', 'resolved-queue')
+    AND task_queue_id IS NULL
+    LIMIT 1
+  ) THEN
+    RAISE EXCEPTION 'Not possible to migrate, some records are still in old format - task_queue_id is NULL for pending tasks';
+  END IF;
```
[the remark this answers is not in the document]
lotas: cool :)

db/versions/0091-migration.sql
```diff
@@ -0,0 +1,166 @@
...
+
+  -- Task Deadlines
+  -- purpose: know when particular task expires, so dependencies/task groups can be resolved
+  CREATE TABLE queue_task_deadlines (
+    task_id text not null,
+    task_group_id text not null,
+    scheduler_id text not null,
+    created timestamptz not null,
+    deadline timestamptz not null,
+    visible timestamptz not null,
```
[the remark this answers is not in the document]
lotas: I take back my comment, sorry :) 
I realized all methods are using `pop_receipt + visible` fields to "check-out" messages and make sure no one is working on them concurrently. So if I'd do this with `deadline`, `DeadlineResolver` would fail to acknowledge and process this entry, because original deadline defined in task would not match.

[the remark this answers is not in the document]
djmitche: Good spot - that makes total sense!

db/versions/0091.yml
```diff
@@ -0,0 +1,609 @@
+version: 91
+description: Migrate existing azure queue messages into new tables in a reversible way
```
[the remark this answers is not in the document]
djmitche: I was wondering about that -- the design was definitely for migrations to wait, even for the background migrations, to complete before moving on. If that's not the case then it's a bug :)

matt-boris (approved):

Issue #6553, opened 2023-09-14
Support additional compression formats in Genereic Worker FileMount and add `tar.lz4` support for directory caches


matt-boris:
lz4, bz2, gz, xz, and zst for FileMount

PR #5093, opened 2022-01-27
[generic-worker] Add support for mounting tar.xz and tar.zst files.

jschwartzentruber:
Revive #3351. This updates archiver to the latest version now that mholt/archiver/issues/213 is fixed. Add support for generic-worker mounts in `tar.xz` and `tar.zst` formats.

Github Bug/Issue: Fixes #3540

PR #6555, opened 2023-09-14
feat(generic-worker): support compression formats for FileMount

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6553.

>Generic Worker File Mounts now include an optional `format` field to specify the compression format for the content. Generic Worker will decompress the retrieved content using the format specified before writing to disk. To avoid decompression, do not include the format field.
>
>Allowed compression formats are: `bz2`, `gz`, `lz4`, `xz`, and `zst`.
>
>This change additionally adds support for the `tar.lz4` format for Writable Directory Caches and Read Only Directories.


workers/generic-worker/mounts.go
```diff
@@ -764,13 +747,80 @@ func extract(fsContent FSContent, format string, dir string, taskMount *TaskMoun
...
+		if err != nil {
+			// this could be a system error, but it can also be that e.g. the task
+			// specified an invalid path, so resolve as malformed payload rather
+			// than panic
+			taskMount.Errorf("Not able to mount content from %v at path %v", fsContent.String(), file)
+			return err
+		}
+		return nil
+	default:
+		log.Fatalf("Unsupported format %v", format)
```
petemoore, line 804: I think this was originally added [here](https://github.com/taskcluster/taskcluster/pull/5093/files#diff-0c9721edee76ea6b023e15639a4e25f9d51080ccedc80b6d19bdb0e1593f7b16R686) and I hadn't spotted that it was using `log.Fatalf`. But that is bad, because it will cause worker to immediately exit without resolving the task. Instead this should either be a `panic` (that would be recovered from, so task gets resolved as exception and logs are uploaded) or it should cause task to resolve as malformed payload. Note this can only happen if schema becomes out of sync with the switch statement, i.e. there is a format in the schema that is missing from the switch statement. But if that happens, we would like task logs, so we should fix this.

workers/generic-worker/schemas/multiuser_windows.yml
```diff
@@ -399,6 +399,19 @@ definitions:
 
           Since: generic-worker 5.4.0
         "$ref": "#/definitions/content"
+      format:
+        title: Format
+        type: string
+        description: |-
+          Compression format of the preloaded content.
+
+          Since: generic-worker 55.2.0
```
petemoore, line 408:
```suggestion
          Since: generic-worker 55.3.0
```

workers/generic-worker/schemas/simple_posix.yml
```diff
@@ -389,6 +389,19 @@ definitions:
 
           Since: generic-worker 5.4.0
         "$ref": "#/definitions/content"
+      format:
+        title: Format
+        type: string
+        description: |-
+          Compression format of the preloaded content.
+
+          Since: generic-worker 55.2.0
```
petemoore, line 398:
```suggestion
          Since: generic-worker 55.3.0
```

workers/generic-worker/mounts_test.go
```diff
@@ -421,6 +421,63 @@ func TestFileMountNoSHA256(t *testing.T) {
...
+				taskID,
+			},
+			TaskRunResolutionState: "completed",
+			TaskRunReasonResolved:  "completed",
+			PerTaskRunLogExcerpts: [][]string{
+				// Required text from first task with no cached value
+				pass1,
+				// Required text from second task when download is already cached
+				pass2,
+			},
```
petemoore, line 476: Would be nice to set `Payload` member here too, to cat the mounted file, to check it was compressed.

workers/generic-worker/mounts_test.go
```diff
@@ -421,6 +421,63 @@ func TestFileMountNoSHA256(t *testing.T) {
...
+	// No cache on first pass
+	pass1 := append([]string{
+		`Downloading task ` + taskID + ` artifact public/build/compressed-file-mount.txt.gz to .*`,
+		`Downloaded 89 bytes with SHA256 a37856e8cd10250f76dc076bb03d380b16a870dec31f3461223f753124a4b28a from task ` + taskID + ` artifact public/build/compressed-file-mount.txt.gz to .*`,
+		`Content from task ` + taskID + ` artifact public/build/compressed-file-mount.txt.gz .* matches required SHA256 a37856e8cd10250f76dc076bb03d380b16a870dec31f3461223f753124a4b28a`,
+		`Creating directory .* with permissions 0700`,
+		`Decompressing gz file .* to .*` + t.Name(),
+	},
+		granting...,
+	)
```
petemoore, line 441: Would be nice to add `` `testing file mounts with compression!` `` here after `cat`ing the file (see comment below about specifying a custom `Payload` to include the `cat` command). This way, we are really sure it was decompressed and e.g. not just copied.

workers/generic-worker/mounts_test.go
```diff
@@ -421,6 +421,63 @@ func TestFileMountNoSHA256(t *testing.T) {
...
+
+	// On second pass, cache already exists
+	pass2 := append([]string{
+		`Found existing download for artifact:` + taskID + `:public/build/compressed-file-mount.txt.gz .* with correct SHA256 a37856e8cd10250f76dc076bb03d380b16a870dec31f3461223f753124a4b28a`,
+		`Creating directory .* with permissions 0700`,
+		`Decompressing gz file .* to .*` + t.Name(),
+	},
+		granting...,
```
petemoore, line 449: Same here.

petemoore (approved):
Really awesome job! And thanks too for adding the test.

I really like that the SHA256 remains that of the compressed source version, and that it remains compressed in the cache, so that two tasks can use it, one decompresses it, the other doesn't, but they still share the same cache entry.

workers/generic-worker/mounts.go
```diff
@@ -764,13 +747,80 @@ func extract(fsContent FSContent, format string, dir string, taskMount *TaskMoun
...
 		log.Fatalf("Unsupported format %v", format)
 		return fmt.Errorf("Unsupported archive format %v", format)
 	}
 	return unarchiver.Unarchive(cacheFile, dir)
 }
 
+func decompress(fsContent FSContent, format string, file string, taskMount *TaskMount) error {
+	cacheFile, err := ensureCached(fsContent, taskMount)
+	if err != nil {
+		log.Printf("Could not cache content: %v", err)
```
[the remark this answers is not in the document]
matt-boris: Just stole it from the `extract()` function :) I'll leave for consistency.

workers/generic-worker/mounts_test.go
matt-boris: I like this, the only issue is, the `LogTest()` helper checks specifically for the `[mounts]` prefix on this output. I'll add a check elsewhere, not in these `pass` variables.

petemoore (approved):
Thanks Boris! And sorry, I had forgotten that the test cases were looking for lines that are prefixed with `[mounts] `. 😬
But thanks for doing it anyway!

PR #6556, opened 2023-09-15
Check markdown urls return HTTP 200 in CI

petemoore:
Can't hurt.


lotas:
oh this is cool! 
and there are quite a few of those broken links

petemoore:
I've had problems with this github action in another repo, and had to remove it, so let's not bother with this. I don't think it works reliably or is well maintained.

PR #6558, opened 2023-09-18
Allow arbitrary per-task test code in mounts tests

petemoore:
I made a slight change here, because at the moment, the current mechanism only allowed you to specify a single string which had to appear in the log, so wouldn't work if you wanted to check for e.g. two different strings, or perform some other per-task test logic. So I simplified it a bit so you just pass an optional function for extra test code, and that function can do whatever you like.


matt-boris (dismissed):
So much cleaner, thanks Pete!

matt-boris:
Looks like there's a lint failure:

```bash
workers/generic-worker/mounts_test.go:487:25: test helper function should start from t.Helper() (thelper)
			PerTaskExtraTesting: func(t *testing.T) {
```

matt-boris (approved):

PR #6560, opened 2023-09-19
Removes deprecated queue.claimTask endpoint

### LINKED ISSUE #6559: Remove deprecated `queue.claimTask()` method
It seems like this method was deprecated long time ago, doesn't contain any documentation and was not seen to be used in logs in the past month.

It is only being used in tests and `docker-worker` has a reference to it. 

Since it's not being used, it should be removed.

Also, it will simplify refactoring of https://github.com/taskcluster/taskcluster/blob/main/services/queue/src/workclaimer.js for the #2940

lotas:
In order to simplify queue refactoring in #2940 we can get rid of this undocumented endpoint that wasn't used

Fixes #6559


clients/client-shell/cmds/task/actors.go
```diff
@@ -162,46 +162,5 @@ func runRetrigger(credentials *tcclient.Credentials, args []string, out io.Write
...
-		Certificate: c.Credentials.Certificate,
-	})
-	r, err := wq.ReportCompleted(taskID, fmt.Sprint(c.RunID))
-	if err != nil {
-		return fmt.Errorf("could not complete the task %s: %v", taskID, err)
-	}
-
-	fmt.Fprintln(out, getRunStatusString(r.Status.Runs[c.RunID].State, r.Status.Runs[c.RunID].ReasonResolved))
-	return nil
+	return fmt.Errorf("tasks could no longer be completed this way", err)
```
lotas, line 165: probably better to drop whole command

lotas:
ok, gonna close this for now as it might be useful in critical situations when some graph needs to be unblocked and let dependent tasks run.

PR #6566, opened 2023-09-25
Add explicit dependency on 's3' for the 'taskcluster' service

bhearsum:
`taskcluster` doesn't start without `s3`, so AFAICT this is strictly necessary.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -115,6 +115,7 @@ services:
     networks:
       - local
     depends_on:
+      - s3
       - ui
       - web-server-web
     volumes:


lotas:
let me check how to fix this properly. `docker-compose.yml` is autogenerated, so shouldn't be modified directly.
We usually modify this file, which later generates resulting yamls https://github.com/taskcluster/taskcluster/blob/87f00abf143940d1890d3c8933da9ab40dc514ef/infrastructure/tooling/src/generate/generators/docker-compose.js

lotas:
Oh, I know what the issue is! Sorry, let me fix it properly. 
The way `nginx.conf` was generated is it has direct reference to `http://s3..` in one of the upstreams.. and weirdly enough, it tries to resolve it at launch time. But, if you use a variable instead, it would delay resolution until service starts or location is being matched, so doesn't result in failure.. so weird :) 

Here's the proper fix: https://github.com/taskcluster/taskcluster/pull/6568

Issue #6564, opened 2023-09-25
Running single UI docker container locally doesn't work because of CSP


lotas:
If you run `docker compose up ui` it would start UI on `localhost:3022` but it would not be able to load resources because of Content Security Policies:

```
Content-Security-Policy: The page’s settings blocked the loading of a resource at http://localhost:3022/assets/11.cdde2ebb.css (“style-src”).
```

This problem doesn't exist when we run also `taskcluster` ingress (which is a nginx with a separate config that removes CSP  headers, for local development purposes) `docker compose up ui taskcluster`

We can fix it by updating HTML file to specify overrides for cs-policies, or mount slightly modified `nginx.conf` that would do this on nginx level

PR #6568, opened 2023-09-25
Fixes local UI running issues

lotas:
Removing startup dependency on `s3` proxy service running. 
Hiding CSP headers for local compose env

Fixes: #6564

### DIFF
--- a/changelog/issue-6564.md
+++ b/changelog/issue-6564.md
@@ -3,4 +3,4 @@ level: patch
 reference: issue 6564
 ---
 
-Fixes CSP related issue with running single UI container locally `docker compose ui`. If run with shipped nginx.conf, it would not be able to load some resources correctly becaues of the strictre 'Content-Security-Policy' headers.
+Fixes CSP related issue with running single UI container locally `docker compose ui`. If run with shipped nginx.conf, it would not be able to load some resources correctly because of the stricter 'Content-Security-Policy' headers.
--- a/changelog/issue-6564.md
+++ b/changelog/issue-6564.md
@@ -1,0 +1,6 @@
+audience: general
+level: patch
+reference: issue 6564
+---
+
+Fixes CSP related issue with running single UI container locally `docker compose ui`. If run with shipped nginx.conf, it would not be able to load some resources correctly becaues of the strictre 'Content-Security-Policy' headers.
--- a/docker/nginx-ui-local-dev-only.conf
+++ b/docker/nginx-ui-local-dev-only.conf
@@ -1,7 +1,6 @@
 # This file overrides ui/web-ui-nginx-site.conf
 # for the local docker compose usage.
 # It is a simplified version and doesn't include stricter security headers.
-# It is to be used in production!
 
 worker_processes  1;
 error_log  stderr;
--- a/docker/nginx-ui-local-dev-only.conf
+++ b/docker/nginx-ui-local-dev-only.conf
@@ -1,0 +1,49 @@
+# This file overrides ui/web-ui-nginx-site.conf
+# for the local docker compose usage.
+# It is a simplified version and doesn't include stricter security headers.
+# It is to be used in production!
+
+worker_processes  1;
+error_log  stderr;
+events {
+  worker_connections  1024;
+}
+http {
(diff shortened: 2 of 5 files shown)


infrastructure/tooling/src/generate/generators/docker-compose.js
```diff
@@ -340,6 +340,9 @@ exports.tasks.push({
         ui: serviceDefinition('ui', {
           command: 'ui/web',
           _useEnvFile: true,
+          volumes: [
+            './docker/nginx-ui.conf:/app/ui/web-ui-nginx-site.conf',
```
lotas, line 344: this is the least intrusive change imho.

I also tried `sed -i \'/Content-Security-Policy/d\' /app/ui/web-ui-nginx-site.conf;` but then realised if someone would like to use `docker compose` to run TC in production (for some strange reason), at least it wouldn't strip all of the security headers. And since it is an extra config file, it can be overwritten or volume mounts could be removed

changelog/issue-6564.md
```diff
@@ -0,0 +1,6 @@
+audience: general
+level: patch
+reference: issue 6564
+---
+
+Fixes CSP related issue with running single UI container locally `docker compose ui`. If run with shipped nginx.conf, it would not be able to load some resources correctly becaues of the strictre 'Content-Security-Policy' headers.
```
matt-boris, line 6:
```suggestion
Fixes CSP related issue with running single UI container locally `docker compose ui`. If run with shipped nginx.conf, it would not be able to load some resources correctly because of the stricter 'Content-Security-Policy' headers.
```

infrastructure/tooling/src/generate/generators/docker-compose.js
```diff
@@ -340,6 +340,9 @@ exports.tasks.push({
         ui: serviceDefinition('ui', {
           command: 'ui/web',
           _useEnvFile: true,
+          volumes: [
+            './docker/nginx-ui-local-dev-only.conf:/app/ui/web-ui-nginx-site.conf',
```
lotas, line 344: renamed file to make it clear it is not production ready :)

infrastructure/tooling/src/generate/generators/docker-compose.js
```diff
@@ -622,7 +625,8 @@ http {
       proxy_set_header Connection "";
       chunked_transfer_encoding off;
 
-      proxy_pass http://s3:9000;
+      set $pass http://s3:9000;
```
lotas, line 628: this is the missing bit. 
for some reason nginx when encounters `proxy_pass http://s3` like directives would try to resolve those at startup time.
But if you declare a variable, parsing would be delayed (?) or until this location was hit.

Other services were already using this approach

docker/nginx-ui-local-dev-only.conf
```diff
@@ -0,0 +1,49 @@
+# This file overrides ui/web-ui-nginx-site.conf
+# for the local docker compose usage.
+# It is a simplified version and doesn't include stricter security headers.
+# It is to be used in production!
```
Eijebong, line 4: "It is *not* to be used in production" I guess

infrastructure/tooling/src/generate/generators/docker-compose.js
matt-boris: Yes, I remember @Eijebong fixed something like this for us sometime back. Thanks for fixing here Yarik!

docker/nginx-ui-local-dev-only.conf
matt-boris: Good catch :)

lotas: lol.. my bad
Thanks :)

matt-boris (approved):

infrastructure/tooling/src/generate/generators/docker-compose.js
Eijebong: Yup, guess I missed this one or it was added later.

lotas: yeah, i added it later probably, so messed it up. Thanks!

Issue #6569, opened 2023-09-25
Increase max length allowed for `hookId` and `hookGroupId`


matt-boris:
@ahal ran into the following issue:

```bash
taskcluster.exceptions.TaskclusterRestFailure: Invalid URL patterns:
URL parameter 'hookId' given as 'cron-task-mozilla-releng-staging-firefox-ios/tab-performance-test' must match regular expression: '/^[a-zA-Z0-9-_\/]{1,64}$/'
```

It looks like it used to be set to 22 characters, but was increased to 64 in this commit https://github.com/djmitche/taskcluster-hooks/commit/a9e03a25e9ef890db3e54b7631728e1508ac013b.

I'll look into doubling the limit to 128.

matt-boris:
Looks like the regex should've been updated in https://github.com/taskcluster/taskcluster/issues/1986 but was missed. I'll update this to be 1000, as that's what the PR bumped the artificial limit up to :)

PR #6570, opened 2023-09-25
fix(hooks): allow for 1000 characters in `hookGroupId` and `hookId`

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6569.

>This patch updates the regex for the `hookGroupId` and `hookId` params for the hooks API to allow for up to 1000 characters.

cc @ahal


lotas (approved):
👍 

lgtm, 1000 is so generous :)

PR #6588, opened 2023-10-02
Issue 6472: Generic Worker osGroups feature: add SysProcAttr.Credential.Groups

### LINKED ISSUE #6472: d2g kvm feature will need osGroup: kvm
I believe kvm feature of docker worker ensures that the host user is in the `kvm` OS group, in order for the docker container to also have access to kvm. I believe in d2g, tasks that use the kvm feature will also need to add `osGroups: kvm` to the task payload. It isn't entirely clear if `osGroups` feature is working on Linux, so this is something we should explicitly test, e.g. by testing a working payload that uses kvm feature under worker pool:

* `proj-taskcluster/old-docker-worker`

and then testing the same payload under worker pool:

* `proj-taskcluster/gw-ubuntu-22-04`

petemoore:
Fixes #6472


petemoore:
Closing, as it references the wrong issue.

PR #6589, opened 2023-10-02
fix(worker-runner): set the `TASKCLUSTER_WORKER_LOCATION` variable for generic worker

### LINKED ISSUE #6471: d2g does not set all TASKCLUSTER_* env vars that Docker Worker sets
I [created](https://github.com/mozilla/community-tc-config/pull/665) a Docker Worker worker pool and a Generic Worker 54.4.1 worker pool, to compare results of identical tasks, to be sure that the Docker Worker compatibility layer is functioning correctly in Generic Worker.

Early results show some differences in environment. Specifically:

Docker worker was setting environment variables:

* `TASKCLUSTER_WORKER_POOL`
* `TASKCLUSTER_WORKER_GROUP`
* `TASKCLUSTER_PUBLIC_IP`
* `TASKCLUSTER_WORKER_TYPE`
* `TASKCLUSTER_INSTANCE_TYPE`

Generic Worker does not set any of the above environment variables. I have some reservations about setting such environment variables, in general, since they encourage task authors to use the information in tasks, which would very much tie them to the taskcluster environment (e.g. making it harder to run tasks locally). However, maybe we need to provide this information to be backwardly compatible. It might be that nothing uses these env vars, but it is hard to know.

Furthermore, Docker Worker was setting this env var correctly, and Generic Worker is not, it seems:

* `TASKCLUSTER_WORKER_LOCATION`

Generic Worker seems to just set it to the empty string.

matt-boris:
Fixes #6471.

>Worker Runner now sets the `TASKCLUSTER_WORKER_LOCATION` variable for Generic Worker.

### DIFF
--- a/changelog/issue-6471.md
+++ b/changelog/issue-6471.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 6471
+---
+Worker Runner now sets the `TASKCLUSTER_WORKER_LOCATION` variable for Generic Worker.
--- a/tools/d2g/d2gtest/testdata/testcases/artifacts_test.yml
+++ b/tools/d2g/d2gtest/testdata/testcases/artifacts_test.yml
@@ -32,12 +32,22 @@ testSuite:
         command:
           - - bash
             - '-cx'
-            - |-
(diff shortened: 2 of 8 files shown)


petemoore (approved):
Awesome, thanks Boris!

PR #6593, opened 2023-10-02
Issue 6592: Generic Worker requires unique items in osGroups array of task payload

### LINKED ISSUE #6592: Generic Worker: osGroups arrays should contain unique items
Currently it is allowed to specify the same group twice in the array of osGroups in Generic Worker. Only unique items should be allowed, since there is no meaning to being in the same group twice.

petemoore:
Fixes #6592

### DIFF
--- a/changelog/issue-6592.md
+++ b/changelog/issue-6592.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 6592
+---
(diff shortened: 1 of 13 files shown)


matt-boris (dismissed):
Thanks 👍🏻

matt-boris (approved):

PR #6595, opened 2023-10-04
Fix auto cancel github builds

lotas:
Events were not properly checked correctly inside the cancelPreviousTaskGroup routine, assuming the caller would decided. Added check inside to make sure we don't cancel more than we should, especially since
[RFC](https://github.com/taskcluster/taskcluster-rfcs/pull/177) said we would only cancel `push` and `pull_request` events

This patch ensures only `push` and `pull_request` are cancelled


matt-boris (approved):
Thanks for the quick fix!

PR #6597, opened 2023-10-04
chore(ui): switch default payload view to yaml

matt-boris:
>UI: Task Definitions and Payloads now display in YAML by default for readability.


petemoore (approved):
LGTM, but I'll let Yarik respond too, as he understands this stuff better than me! 😆

lotas (approved):
LGTM! Thanks

PR #6603, opened 2023-10-06
Switch to ESM modules 

### LINKED ISSUE #4260: Upgrade to ECMAScript Modules
I believe this is an all-or-nothing change to the entire JS codebase.

Once this is done, we can use
* `quick-lru` (#4257).  Our use of that specific module is not nearly deep enough to drive this change -- at worst, we could find another (still CJS) LRU package.
* `markdown-table` (#4566)
* `p-queue` (#4678)
* `got` (#5388 )
* ~~`escape-string-regex` (#4783)~~
* `@sentry/browser` (#5547)

lotas:
## Motivation

In node.js ESM (`"type": "module"`) package can import both CommonJS packages, and ESM packages. However, commonjs package cannot import newer ESM packages.

As more and more dependencies switch to ESM-only distribution, making upgrades (especially security upgrades) become hard. 

Taskcluster platform consists of multiple utility (library) packages, services themselves, and clients. 

Library packages: `db`, `api`, `app`, `config`, `iterate`, `loader`, `monitor`, `postgres`, `pulse`, `references`, `testing`, `validate`

Services: `auth`, `built-in-workers`, `github`, `hooks`, `index`, `notify`, `object`, `purge-cache`, `queue`, `secrets`, `web-server`, `worker-manager`

Clients that are published: `taskcluster-client`, `taskcluster-client-web`. 

It is not possible to only convert services packages to ESM, as they depend on library packages, which also consume 3rd-party dependencies.

Clients, however, do not depend on those, so could  possibly remain in CommonJS format to avoid shipping both ESM/CJS builds.

`taskcluster-client` is a published standalone package, but uses unlisted dependencies for testing: `taskcluster-lib-*`, `taskcluster-testing`. Those are not listed in package dependencies, as those packages are not installed, but are internal. 
This was refactored, and implicit dependencies were removed, and tests that depends on those were moved into separate `client/client-test` folder

## importing

ESM requires that files would be imported with extensions. 

```js
// This worked fine in cjs
const { util } = require('./util')

// this wouldn't work anymore
import { util } from './util'

// this will
import { util } from './util.js'
// or if it is a folder
import { util } from './util/index.js'
```

For the same reason (I believe) running `node services/auth/src/main server` will not work, but `node services/auth/src/main.js server` will work

## `__dirname`, `__filename`

Can now be used with the help of `import.meta.url`:

```js
const __filename = new URL('', import.meta.url).pathname;
```

## `module.parent` no longer available

Not possible to do:
```js
if (!module.parent) {
  load.crashOnError(process.argv[2]);
}
```

This will work:
```js
import { fileURLToPath } from "url";
if (process.argv[1] === fileURLToPath(import.meta.url)) {
  load.crashOnError(process.argv[2]);
}
```

## Testing
Since ESM require exports to be statically analyzable, test helper functions are being rewritten:
instead of `exports.xx` and dynamic injections of helper functions, `helper.js` now exposes single `helper = {}` object that is being modified by utility functions like `withDb()`, `withPulse()`

## Exports

Taskcluster internal libraries expose everything as named export and as default to allow mixed usage:

```js
export * from "./upgrade.js";
import { upgrade, downgrade } from './upgrade.js';
export default { upgrade, downgrade };
```

Can be used both ways:
```js
import { upgrade } from 'taskcluster-db';
upgrade()
// or
import db from 'taskcluster-db';
db.upgrade()
```

## Import assertions

[Node.js](https://nodejs.org/api/esm.html#import-assertions) allows to specify import assertions since v17:

```js
// load json module
import jsonSchemaDraft06 from 'ajv/lib/refs/json-schema-draft-06.json' assert { type: 'json' };

// equivalent of 
const jsonSchemaDraft06 = require('ajv/lib/refs/json-schema-draft-06.json');
```

However, it is still experimental and comes with: 
```
(node:40419) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
```
Also [eslint](https://github.com/eslint/eslint/discussions/15305) doesn't implement it yet, waiting for it to be analysed.

Alternatives:
```js
const dataDir = new URL('./data/yml/', import.meta.url).pathname;
const loadJson = filename => JSON.parse(fs.readFileSync(path.join(dataDir, filename), 'utf8'));
// or
const schemaPath = new URL('../../../node_modules/ajv/lib/refs/json-schema-draft-06.json', import.meta.url).pathname;
const jsonSchemaDraft06 = JSON.parse(fs.readFileSync(schemaPath, 'utf-8'));
```

-----

Files were migrated in dumb way `npx cjs-to-es6 path/to/files` and does not result in a working code, there are still many steps to follow:

## TODO

- [x] `db/`
- [x] `infrastructure/tooling`
- [x] `libraries/api`
- [x] `libraries/app`
- [x] `libraries/config`
- [x] `libraries/iterate`
- [x] `libraries/loader`
- [x] `libraries/monitor`
- [x] `libraries/postgres`
- [x] `libraries/pulse`
- [x] `libraries/references`
- [x] `libraries/testing`
- [x] `libraries/validate`
- [x] `services/auth`
- [x] `services/built-in-workers`
- [x] `services/github`
- [x] `services/hooks`
- [x] `services/index`
- [x] `services/notify`
- [x] `services/object`
- [x] `services/purge-cache`
- [x] `services/queue`
- [x] `services/secrets`
- [x] `services/web-server`
- [x] `services/worker-manager`

Fixes #4260


changelog/issue-4260.md
```diff
@@ -0,0 +1,6 @@
+audience: developers
+level: major
```
lotas, line 2: although clients are mostly untouched, and are still CJS, worth making a major

clients/client/test/helper.js
```diff
@@ -23,3 +26,17 @@ exports.withRestoredEnvVars = () => {
     }
   });
 };
+
+const ROOT_DIR = path.resolve(__dirname, '../../..');
+const suiteName = () => {
```
lotas, line 31: this is copied from lib/testing to avoid having to implicitly depend on package from outside.

clients/client/test/retry_test.js
```diff
@@ -1,260 +1,265 @@
...
-    description: 'Place we can call to test something',
-  }, function(req, res) {
-    getInternalErrorCount += 1;
-    res
-      .status(500)
-      .json({
-        message: 'Internal error of sorts',
-      });
-  });
+// suite(testing.suiteName(), function() {
```
lotas, line 17: as this whole test depends on "internal" libraries, it shouldn't be here.
figuring out where to put it

clients/client-test/package.json
```diff
@@ -0,0 +1,35 @@
+{
```
lotas, line 1: this was added to keep `clients/client` clean, and separate tests that used internal libraries from the client itself.

lotas:
closed in favor of https://github.com/taskcluster/taskcluster/pull/6627

PR #6604, opened 2023-10-06
Generic Worker: Use system copy command for better logging

petemoore:
System commands such as copy (Windows) or cp (Linux/macOS/FreeBSD) provide useful output when commands fail. Elsewhere we have been using system commands for such things, let's do this for copying files too.

This is part 1, which separates out the optional gzip compression from the file copy, in preparation for using the system copy command in future.


matt-boris (approved):
Nice, thanks Pete 👍🏻

Issue #6371, opened 2023-07-05
d2g cli should translate full task definitions rather than just the task payload


petemoore:
This is potentially a breaking change, unless we backwardly support translating direct payloads too, although that may not be necessary since I believe we are the only users of d2g right now.

Both payload and scopes of docker-worker tasks need translating to generic-worker format in order to create a valid task, however at the moment, both the `d2g` executable, and the `taskcluster d2g` cli subcommand process just task payloads. The underlying d2g library supports translating both.

Probably the simplest, is to just adapt the code to take a full task definition, and output a full task definition, not altering (nor validating) any properties except for `payload` and `scopes`. The reason _not_ to validate the other fields, is that they are not interpreted by the worker, they are interpreted by the Queue, and are subject to change. The Queue's task definition schema may change over time, yet still continue to work with a released Generic Worker, and therefore there is no reason to validate the properties of a task definition other than the properties that Generic Worker interprets itself (namely, `payload` and `scopes`).

PR #6605, opened 2023-10-06
fix(d2g): add `kvm` OS group for KVM feature

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6472.

>D2G: Add the `kvm` OS group to the Generic Worker payload if the KVM device is enabled in the Docker Worker payload.

Keeping in WIP for now until we can get a working test to prove we need the task user added to these groups.

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -31,19 +31,19 @@ type (
 
 func ConvertTaskDefinition(dwTaskDef json.RawMessage) (json.RawMessage, error) {
 	var gwTaskDef json.RawMessage
-	var parsedTaskDef map[string]interface{}
-	err := json.Unmarshal(dwTaskDef, &parsedTaskDef)
+	var parsedDwTaskDef map[string]interface{}
+	err := json.Unmarshal(dwTaskDef, &parsedDwTaskDef)
 	if err != nil {
 		return nil, fmt.Errorf("cannot parse task definition: %v", err)
 	}
 
-	if _, exists := parsedTaskDef["payload"]; !exists {
+	if _, exists := parsedDwTaskDef["payload"]; !exists {
 		return nil, fmt.Errorf("task definition does not contain a payload")
 	}
 
 	dwPayload := new(dockerworker.DockerWorkerPayload)
 	defaults.SetDefaults(dwPayload)
-	dwPayloadJSON, err := json.Marshal(parsedTaskDef["payload"])
+	dwPayloadJSON, err := json.Marshal(parsedDwTaskDef["payload"])
 	if err != nil {
 		return nil, fmt.Errorf("cannot marshal Docker Worker payload: %v", err)
 	}
@@ -57,12 +57,15 @@ func ConvertTaskDefinition(dwTaskDef json.RawMessage) (json.RawMessage, error) {
 		return nil, fmt.Errorf("cannot convert Docker Worker payload: %v", err)
 	}
 
-	if scopes, exists := parsedTaskDef["scopes"]; exists {
+	if scopes, exists := parsedDwTaskDef["scopes"]; exists {
 		var dwScopes []string
 		for _, scope := range scopes.([]interface{}) {
 			dwScopes = append(dwScopes, scope.(string))
 		}
-		parsedTaskDef["scopes"] = Scopes(dwScopes)
+		parsedDwTaskDef["scopes"], err = Scopes(dwScopes, dwTaskDef)
+		if err != nil {
+			return nil, fmt.Errorf("cannot convert Docker Worker scopes: %v", err)
+		}
 	}
 
 	d2gConvertedPayloadJSON, err := json.Marshal(*gwPayload)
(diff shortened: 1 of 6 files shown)


petemoore (commented):
It looks good, but note you will probably need to update the scopes too, which are required for the osGroups feature.

matt-boris:
> need to update the scopes too

Very true! Thanks for calling this out 👍🏻

matt-boris:
I think this now becomes dependent on https://github.com/taskcluster/taskcluster/issues/6371. Currently, d2g knows nothing of the scopes if it's performing a payload translation. It needs to add the scopes required for the osGroups that are added for the KVM feature.

tools/d2g/d2g.go
```diff
@@ -324,6 +325,14 @@ func setSupersederURL(dwPayload *dockerworker.DockerWorkerPayload, gwPayload *ge
 	gwPayload.SupersederURL = dwPayload.SupersederURL
 }
 
+func setOSGroups(dwPayload *dockerworker.DockerWorkerPayload, gwPayload *genericworker.GenericWorkerPayload) {
+	if dwPayload.Capabilities.Devices.KVM {
+		// task user needs to be in the kvm and libvirt groups for KVM to work
+		// https://help.ubuntu.com/community/KVM/Installation
+		gwPayload.OSGroups = append(gwPayload.OSGroups, "kvm", "libvirt")
```
Eijebong, line 332: You do not need libvirt to use KVM. Libvirt is just an orchestrator allowing you to manage virtual machines from different backends via XML files.

matt-boris: Thanks Bastien 👍🏻 I'll go ahead and remove then. So looking over at https://github.com/taskcluster/community-tc-config/commit/5a7ebdcf319fb6ac8d85932ddfe40ca0ad27eac7, some of these packages may not be needed on the worker? Or would you suggest keeping these installed in case the user may want to use libvirt?

Eijebong: They definitely shouldn't be necessary for getting a VM to run.

I don't know who's using those images, maybe ask them if they need libvirt? I'd personally remove them and add them back if anyone actually needs them and finds it cumbersome to install them during their CI workflow. But I don't know what the stance on that is for community images.

tools/d2g/d2g.go
```diff
@@ -57,22 +57,25 @@ func ConvertTaskDefinition(dwTaskDef json.RawMessage) (json.RawMessage, error) {
...
 	d2gConvertedPayloadJSON, err := json.Marshal(*gwPayload)
 	if err != nil {
 		return nil, fmt.Errorf("cannot marshal Generic Worker payload: %v", err)
 	}
 
-	parsedTaskDef["payload"] = json.RawMessage(d2gConvertedPayloadJSON)
+	parsedDwTaskDef["payload"] = json.RawMessage(d2gConvertedPayloadJSON)
 
-	gwTaskDef, err = json.MarshalIndent(parsedTaskDef, "", "  ")
+	gwTaskDef, err = json.MarshalIndent(parsedDwTaskDef, "", "  ")
```
petemoore, line 78: the left side says it is generic-worker task definition, but right side says it is a docker worker task definition?

tools/d2g/d2g.go
```diff
@@ -84,7 +87,7 @@ func ConvertTaskDefinition(dwTaskDef json.RawMessage) (json.RawMessage, error) {
 // equivalent Generic Worker scopes. These scopes should be used together with
 // a converted Docker Worker task payload (see d2g.Convert function) to run
 // Docker Worker tasks under Generic Worker.
-func Scopes(dwScopes []string) (gwScopes []string) {
+func Scopes(dwScopes []string, dwTaskDef json.RawMessage) (gwScopes []string, err error) {
```
petemoore, line 90: I think we should pass in a DockerWorkerPayload struct here, rather than json.RawMessage. The only reason we do that elsewhere (ConvertTaskDefinition) is because the task definition can evolve independently of the gw/dw payload definitions (e.g. if new properties are added, that the queue cares about, but the worker doesn't). However, a version of d2g only supports specific docker-worker payload format, and can only generate properties in the generic worker payload format that it knows about, so with payloads, it would be better to pass around the go types, for consistency with the Convert method, etc.

tools/d2g/d2g.go
```diff
@@ -102,6 +105,32 @@ func Scopes(dwScopes []string) (gwScopes []string) {
...
+		return nil, fmt.Errorf("cannot marshal Docker Worker payload: %v", err)
+	}
+	err = json.Unmarshal(dwPayloadJSON, &dwPayload)
+	if err != nil {
+		return nil, fmt.Errorf("cannot unmarshal Docker Worker payload: %v", err)
+	}
+
+	if dwPayload.Capabilities.Devices.KVM {
+		gwScopes = append(gwScopes, fmt.Sprintf("generic-worker:os-group:%s/kvm", parsedDwTaskDef["taskQueueId"]))
+	}
```
petemoore, line 132: i think this ends up being the only part you need to keep, once you pass in a native docker worker payload go type

tools/d2g/d2gtest/d2g_test.go
```diff
@@ -33,7 +33,11 @@ func ExampleScopes_mixture() {
 		"docker-worker:capability:device:loopbackVideo:x/y/z",
 		"docker-worker:capability:device:kvm:x/y/z",
 	}
-	gwScopes := d2g.Scopes(dwScopes)
+	gwScopes, err := d2g.Scopes(dwScopes, nil)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
```
petemoore, line 40: I think this doesn't work for full task definition conversions?

tools/d2g/d2g.go
matt-boris: Thanks, just a typo, good catch!

tools/d2g/d2g.go
matt-boris: `taskQueueId` is not in the payload though. That's why I passed in the entire task def.

tools/d2g/d2gtest/d2g_test.go
matt-boris: This is just a simple test to see that the scopes convert. Isn't that ok?

tools/d2g/d2g.go
matt-boris: This is the entire task definition, though. I agree, if I only needed the payload, that I should pass in the go type.

PR #6612, opened 2023-10-11
feat(queue): patch migration 91 to fix invalid columns

lotas:
During Community-TC upgrade it was discovered that one of the patched functions was referencing wrong column: 
`queue_name_old` instead of `queue_name_compat`

Also `azure_queue_put_extra` didn't coalesce null value for `task_queue_id`, which could result in few messages being lost for the claim queue (or `claimWork` call failing with error, and worker would have to call it again)


matt-boris (approved):

djmitche:
Bummer that we missed this in review and in tests :(

lotas:
> Bummer that we missed this in review and in tests :(

too many changes, to few tests :) 

I didn't test what happens to the existing running code that uses live-patched functions, so missed those corner cases

PR #6618, opened 2023-10-13
feat(d2g): translate full task definition

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6371.

>D2G tool now can convert an entire Docker Worker task definition to a Generic Worker task definition.
>
>This is now used under the hood of Generic Worker to unblock PR [#6605](https://github.com/taskcluster/taskcluster/pull/6605).
>
>New `taskcluster d2g -h` output:
>
>```bash
>Converts a docker-worker payload (JSON) to a generic-worker payload (JSON).
>To convert a task definition (JSON), you must use the task definition flag (-t, --task-def).
>
>Usage:
>  taskcluster d2g [flags]
>
>Examples:
>  taskcluster d2g -f /path/to/input/payload.json
>  taskcluster d2g -t -f /path/to/input/task-definition.json
>  cat /path/to/input/payload.json | taskcluster d2g
>  cat /path/to/input/task-definition.json | taskcluster d2g -t
>  echo '{"image": "ubuntu", "command": ["bash", "-c", "echo hello world"], "maxRunTime": 300}' | taskcluster d2g
>
>Flags:
>  -f, --file string   Path to a .json file containing a docker-worker payload or task definition.
>  -h, --help          help for d2g
>  -t, --task-def      Must use if the input is a docker-worker task definition.
>
>Global Flags:
>  -v, --verbose   verbose output
>```


petemoore (changes requested):
Hey Boris, this looks awesome!
One thing I spotted though, is that there is a new dependency on `tcqueue.TaskDefinitionResponse` in the underlying d2g tool. The issue with this is that d2g library is included in e.g. generic-worker binary, so that would effectively burn the task definition schema into the binary at build time. As the schema for task definition evolves over time, these changes would not be picked up, so e.g. new fields would not appear.
I think the correct approach in this case is to take the task definition as an arbitrary blob that does not need to conform to a particular schema, other than it has a `payload` json object, and an optional `scopes` json string array. Those particular properties should be translated, and the rest of the task definition should be left untouched. This way, if new properties appear or are removed over time, or their default value changes etc, those changes would be respected. The only concern d2g should have is with these two properties, and leave the Queue to validate the rest of the task definition. This way, we won't accidentally modify fields outside of `scopes` and `payload` that were defined/modified in the Queue some time after the d2g/generic-worker binary is built.

tools/d2g/d2gtest/schemas/test_suites.yml
```diff
@@ -62,6 +60,12 @@ properties:
               description: |-
                 Detailed information about what the test case tests
               type: string
+            isTaskDef:
+              title: Is Task Definition
+              description: |-
+                Set to true if the test input are task definitions in
+                dockerWorkerTaskDefinition and genericWorkerTaskDefinition.
+              type: boolean
```
petemoore, line 68: I don't think this is really needed, since you can see if a payload or full task definition has been included.

tools/d2g/d2gtest/schemas/test_suites.yml
```diff
@@ -48,8 +48,6 @@ properties:
           required:
             - name
             - description
-            - dockerWorkerTaskPayload
-            - genericWorkerTaskPayload
```
petemoore, line 52: I think it would have been better to use `oneOf` json schema construct here, so that you could guarantee that either a pair of payloads are included, _or_ a pair of task definitions.

tools/d2g/d2gtest/d2g_test.go
```diff
@@ -197,6 +175,88 @@ func validateTestSuite(t *testing.T, schema string, path string) {
...
+		}
+
+		gwRaw, err = json.Marshal(parsedGwTaskDef["payload"])
+		if err != nil {
+			t.Fatalf("Cannot marshal test suite Generic Worker payload %#v: %v", parsedGwTaskDef["payload"], err)
+		}
+	} else {
+		dwRaw = tc.DockerWorkerTaskPayload
+		gwRaw = tc.GenericWorkerTaskPayload
+	}
```
petemoore, line 210: It feels like we should have two different types for a PayloadTestCase and a TaskDefinitionTestCase, which I think sits well with having a `oneOf` too.

petemoore (commented):
What you have works, but I think there may be some room for improvement. I'll let you decide if you think it is worth the effort or not.

tools/d2g/d2gtest/schemas/test_suites.yml
```diff
@@ -74,3 +58,59 @@ properties:
...
+              # than using $ref, since gojsonschema can't interpret yaml
+              # schemas and we prefer not to keep both json and yml versions in
+              # the repo
+              type: object
+            genericWorkerTaskDefinition:
+              # validation of this object is handled in test harness, rather
+              # than using $ref, since gojsonschema can't interpret yaml
+              # schemas and we prefer not to keep both json and yml versions in
+              # the repo
+              type: object
```
petemoore, line 95: It will be easier if you embed all the properties that the d2g test cases include here (such as taskQueueId, payload, scopes, ....). Then you won't need all the unmarshaling in the code. Note, this only pertains to the test case data, so you don't need to worry about task definition properties changing over time, since the input/output pairs are frozen in the tests (so the schema only needs to include the properties that the tests actually contain). Make sure to include "additionalProperties: false" though, so that the generated go code isn't full of json.RawMessage. I would even add an optional (non-required) property "foo", e.g. type object, and create a single test case that includes it with some value, to make sure that it is passed through verbatim to the translated task definition. Only one test needs to test this functionality. Obviously, document "foo" as a fake property for the purposes of testing that unknown properties get passed through.

Incidentally, it occurs to me that when converting tasks, a task definition may include provisionerId/workertype or taskQueueId, but doesn't need to include both. If we translate tasks, I guess we should check if both exist, that they are consistent with each other, and if only one exists, we use that one, and not the other when passing to the Scopes conversion routine. Because if we assume it is present, and then pass through an empty string, we'll get the wrong scope. Just occurred to me. That is also something we can cover with a test case. Normally generic-worker doesn't care, because it gets task definition from Queue service, and that will include both, but if someone is converting tasks definitions that they create themself (e.g. from taskgraph) they might have one of the properties, but not the other, which is valid for taskcluster, but d2g wouldn't understand in its current incarnation.

matt-boris: I've addressed your second paragraph's concern in https://github.com/taskcluster/taskcluster/pull/6618/commits/a18262d3e739c8230558b0fdf0211a96791390ec.

Regarding your first paragraph - there are tradeoffs to doing this. The only field accessed in the d2g tests is the `payload` and that's to validate against the worker's payload schema. So that's two unmarshals.

If the `payload` field is added here, I'll then need to change:
  * marshal the field into JSON to do the validation steps I mentioned above
  * https://github.com/taskcluster/taskcluster/blob/a18262d3e739c8230558b0fdf0211a96791390ec/tools/d2g/d2gtest/d2g_test.go#L151 I'll no longer be able to pass the docker worker task definition in as-is. I'll need to marshal the docker worker task definition into JSON before passing to `d2g.ConvertTaskDefinition()`

Keeping it as-is keeps the test case method for task definitions super straight forward: https://github.com/taskcluster/taskcluster/blob/a18262d3e739c8230558b0fdf0211a96791390ec/tools/d2g/d2gtest/d2g_test.go#L147-L170

Let me know what you think! Thanks Pete

matt-boris: And in https://github.com/taskcluster/taskcluster/pull/6618/commits/77829a2beaa2b93bc7d5379f6cc25783589fdf00, I just added another test to ensure new fields are translated verbatim through to the resulting generic worker task definition.

petemoore (approved):
Love it, thanks!

PR #6619, opened 2023-10-16
push events are no longer cancelled, only pull_request

### LINKED ISSUE #6616: Github auto-cancellation tries to cancel too much
Following commit was attempted to be cancelled, but due to lack of scopes it failed
https://github.com/mozilla-releng/scriptworker-scripts/commit/3a824a199479b63d7f1bd0ef155a4e16eea9b2a6#commitcomment-129821443

This was coming from `push` event.

Upon investigation it was revealed that there are two different push events with the same SHA. Github service didn't keep branch references internally so far, so it couldn't differentiate that two pushes were done on two different branches: `master` and `production`.
And in this case none of those should have been cancelled or even attempted. 

TODO: 
- Github service should additionally store branch reference associated with github build
- when push event tries to auto cancel previous builds it should include branch name to make sure only commits with the same SHA in the same branch are cancelled

lotas:
This is to prevent cancelling builds two builds for the same commit, as they might belong to different branches.

And to keep it simple we only cancel `pull_request` events.

Fixes #6616

### DIFF
--- a/changelog/issue-6616.md
+++ b/changelog/issue-6616.md
@@ -1,0 +1,9 @@
+audience: users
+level: patch
+reference: issue 6616
+---
+
+Github service no longer cancels builds for the same SHA for `push` events.
+Only `pull_request` events would cancel running builds for the same pull request if they exist.
+
+This is to avoid canceling same commit pushed to different branches.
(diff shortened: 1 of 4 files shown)


matt-boris (approved):
Nice, thanks!

PR #6624, opened 2023-10-17
Allow specifying `0` as a value for `onExitStatus.purgeCaches`

Eijebong:
I was trying to abuse the caching system to download a binary that I was developing on another machine to run in a task. Took me by surprise that it didn't allow automatically purging the cache when the command succeeded. As far as I can tell, docker-worker did not have such a limit and I don't see why there should be one anyway.

### DIFF
--- a/changelog/cBmAFvklRfu3hYByRUiRGA.md
+++ b/changelog/cBmAFvklRfu3hYByRUiRGA.md
@@ -1,0 +1,4 @@
+audience: users
+level: patch
+---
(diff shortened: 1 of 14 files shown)


matt-boris (commented):
Thanks Bastien, I agree with you. This artificial requirement didn't make much sense, unlike for the retry option.

It looks like you need to run `yarn generate` now, though.

matt-boris (approved):

PR #6627, opened 2023-10-18
ESmodules

### LINKED ISSUE #4260: Upgrade to ECMAScript Modules

lotas:
Same as https://github.com/taskcluster/taskcluster/pull/6603 
But split into per-folder commits

## Motivation

In node.js ESM (`"type": "module"`) package can import both CommonJS packages, and ESM packages. However, commonjs package cannot import newer ESM packages.

As more and more dependencies switch to ESM-only distribution, making upgrades (especially security upgrades) become hard. 

Taskcluster platform consists of multiple utility (library) packages, services themselves, and clients. 

Library packages: `db`, `api`, `app`, `config`, `iterate`, `loader`, `monitor`, `postgres`, `pulse`, `references`, `testing`, `validate`

Services: `auth`, `built-in-workers`, `github`, `hooks`, `index`, `notify`, `object`, `purge-cache`, `queue`, `secrets`, `web-server`, `worker-manager`

Clients that are published: `taskcluster-client`, `taskcluster-client-web`. 

It is not possible to only convert services packages to ESM, as they depend on library packages, which also consume 3rd-party dependencies.

Clients, however, do not depend on those, so could  possibly remain in CommonJS format to avoid shipping both ESM/CJS builds.

`taskcluster-client` is a published standalone package, but uses unlisted dependencies for testing: `taskcluster-lib-*`, `taskcluster-testing`. Those are not listed in package dependencies, as those packages are not installed, but are internal. 
This was refactored, and implicit dependencies were removed, and tests that depends on those were moved into separate `client/client-test` folder

## importing

ESM requires that files would be imported with extensions. 

```js
// This worked fine in cjs
const { util } = require('./util')

// this wouldn't work anymore
import { util } from './util'

// this will
import { util } from './util.js'
// or if it is a folder
import { util } from './util/index.js'
```

For the same reason (I believe) running `node services/auth/src/main server` will not work, but `node services/auth/src/main.js server` will work

## `__dirname`, `__filename`

Can now be used with the help of `import.meta.url`:

```js
const __filename = new URL('', import.meta.url).pathname;
```

## `module.parent` no longer available

Not possible to do:
```js
if (!module.parent) {
  load.crashOnError(process.argv[2]);
}
```

This will work:
```js
import { fileURLToPath } from "url";
if (process.argv[1] === fileURLToPath(import.meta.url)) {
  load.crashOnError(process.argv[2]);
}
```

## Testing
Since ESM require exports to be statically analyzable, test helper functions are being rewritten:
instead of `exports.xx` and dynamic injections of helper functions, `helper.js` now exposes single `helper = {}` object that is being modified by utility functions like `withDb()`, `withPulse()`

## Exports

Taskcluster internal libraries expose everything as named export and as default to allow mixed usage:

```js
export * from "./upgrade.js";
import { upgrade, downgrade } from './upgrade.js';
export default { upgrade, downgrade };
```

Can be used both ways:
```js
import { upgrade } from 'taskcluster-db';
upgrade()
// or
import db from 'taskcluster-db';
db.upgrade()
```

## Import assertions

[Node.js](https://nodejs.org/api/esm.html#import-assertions) allows to specify import assertions since v17:

```js
// load json module
import jsonSchemaDraft06 from 'ajv/lib/refs/json-schema-draft-06.json' assert { type: 'json' };

// equivalent of 
const jsonSchemaDraft06 = require('ajv/lib/refs/json-schema-draft-06.json');
```

However, it is still experimental and comes with: 
```
(node:40419) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
```
Also [eslint](https://github.com/eslint/eslint/discussions/15305) doesn't implement it yet, waiting for it to be analysed.

Alternatives:
```js
const dataDir = new URL('./data/yml/', import.meta.url).pathname;
const loadJson = filename => JSON.parse(fs.readFileSync(path.join(dataDir, filename), 'utf8'));
// or
const schemaPath = new URL('../../../node_modules/ajv/lib/refs/json-schema-draft-06.json', import.meta.url).pathname;
const jsonSchemaDraft06 = JSON.parse(fs.readFileSync(schemaPath, 'utf-8'));
```

-----

Files were migrated in dumb way `npx cjs-to-es6 path/to/files` and does not result in a working code, there are still many steps to follow:

## TODO

- [x] `db/`
- [x] `infrastructure/tooling`
- [x] `libraries/api`
- [x] `libraries/app`
- [x] `libraries/config`
- [x] `libraries/iterate`
- [x] `libraries/loader`
- [x] `libraries/monitor`
- [x] `libraries/postgres`
- [x] `libraries/pulse`
- [x] `libraries/references`
- [x] `libraries/testing`
- [x] `libraries/validate`
- [x] `services/auth`
- [x] `services/built-in-workers`
- [x] `services/github`
- [x] `services/hooks`
- [x] `services/index`
- [x] `services/notify`
- [x] `services/object`
- [x] `services/purge-cache`
- [x] `services/queue`
- [x] `services/secrets`
- [x] `services/web-server`
- [x] `services/worker-manager`

Fixes #4260


matt-boris:
d6ab4790b8a678ff9c15f9a14d50d3a8be49ff2f LGTM.

Continuing on...

matt-boris:
Have one question in b3fcfd0261d26088a2a58bb1c05408587b4a18df, but overall LGTM.

matt-boris:
86fa244d0e627ff3d28527ac90895a88ed76d1b0 LGTM

lotas:
> Have one question in https://github.com/taskcluster/taskcluster/commit/b3fcfd0261d26088a2a58bb1c05408587b4a18df, but overall LGTM.

have you published it yet? Don't see any 

I owe you many beers for reviewing this already 😬

matt-boris:
e9d8b39053fcb2183e33c18691a81241cbf2100c LGTM 😅

dev-docs/best-practices/testing.md
matt-boris, line 1: Thanks for going around and fixing the *.md files as well! 🎉

package.json
matt-boris, line 1: I wonder if we should try running `npx depcheck` on our `package.json` files to help remove any unused deps? I see you removed a couple, so that got me thinking about [`depcheck`](https://www.npmjs.com/package/depcheck).

taskcluster/ci/client/kind.yml
```diff
@@ -30,6 +30,16 @@ tasks:
         {{ yarn --frozen-lockfile || exit 99; }} &&
         yarn test &&
         yarn package-test
+  nodejs-client-test:
```
matt-boris, line 33: Nice, thanks! We'll need to update branch protection settings too to require this check once this is all merged 👍🏻

taskcluster/ci/client/kind.yml
```diff
@@ -30,6 +30,16 @@ tasks:
...
+    description: nodejs client tests
```
matt-boris, line 34: Maybe use a different description vs the task defined above this?

matt-boris:
9798a50502e9bd28d06d527e2221506abd3a808c looks pretty good, I left a couple comments/questions.

matt-boris (commented):
What an undertaking!! 💪🏻

Left a few comments/questions. Thanks so much for tackling this Yarik! 🎉

clients/client-test/test/clients_test.js
```diff
@@ -0,0 +1,24 @@
+import taskcluster from 'taskcluster-client';
+import testing from 'taskcluster-lib-testing';
+import assert from 'assert';
+
+suite(testing.suiteName(), function () {
+  test('Main clients exposed', function () {
```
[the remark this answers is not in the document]
matt-boris: I had that feeling.. I'm sure a lot of these fixes required careful hand-crafting, so no worries about doing it here as well. Just think the proper solution would be in a generator :)

package.json
lotas: this is already happening, there is a `depcheck.js` ;) 
it was improved to support imports as well

taskcluster/ci/client/kind.yml
lotas: nicely spot! will rename

clients/client-test/test/clients_test.js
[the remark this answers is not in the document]
lotas: Rewrote it to be generated now.

infrastructure/tooling/src/generate/generators/tc-client.js
matt-boris, line 1: Awesome!! 🎉

matt-boris (approved):
Awesome work! 💪🏻

PR #6628, opened 2023-10-19
chore(ci): add deps to help not waste resources

### LINKED ISSUE #6621: CI tasks should depend on linting to succeed before running
This will help to not waste resources.

Here's an example with go lint failing https://community-tc.services.mozilla.com/tasks/groups/JJQw9Kj6RIaN6ykMO6jXnw. Ideally, the tests wouldn't start until `lint-golang` is successful.

This should be done for go, node, and python (if applicable).

matt-boris:
Fixes #6621.

This only applied to the go tasks, as they'd fail if there was a lint error.

### DIFF
--- a/changelog/issue-6621.md
+++ b/changelog/issue-6621.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 6621
+---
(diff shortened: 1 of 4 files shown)


lotas (approved):
Cool! Thanks

PR #6630, opened 2023-10-23
feat(docker): Move image names into .env file

lotas:
Version tags were changed after each release which resulted in all compose files being changed. 
This introduces unnecessary changes in history.

This will keep changing versions in `.env` file and `docker-compose*` files would reference it using `image: ${IMAGE}` format, as `docker` by default reads variables from `.env` file


matt-boris (approved):
This is great, thanks Yarik! I'd even say this fits on the https://github.com/orgs/taskcluster/projects/17 board as far as a better dev experience!

PR #6632, opened 2023-10-23
Hook service display 'unknown' state when task doesn't exist anymore

lotas:
This [hook](https://firefox-ci-tc.services.mozilla.com/hooks/project-releng/cron-task-mozilla-central%2Fship-geckoview) has tasks that expire after 7d, so no longer available. Displaying `"unscheduled"` as its state might not be accurate, as they were scheduled at that time.


matt-boris (approved):
👍🏻

Issue #2923, opened 2020-05-28
add yarn scripts for making writing db migration easier


imbstack:
This would basically automate some of [this checklist](https://github.com/taskcluster/taskcluster/tree/master/db#checklist) and comprise:

`yarn db:migration:create`: using some templates, create the files needed to make a migration, incrementing version numbers automatically, etc

`yarn db:migration:check`: this can be run locally and also should be run in CI for PRs to check that all files mentioned in the checklist exist and are "correct" as much as we can tell in a test

PR #6633, opened 2023-10-23
DB helper scripts to create new migrations

lotas:
To make it easier to create new migrations.

Moved old `renumber-db-verison.js` script into db utils

Introduces helper database utilities to create new version (migration) and renumber existing.

`yarn db:new` creates new db migration and test
`yarn db:renumber old new` renumbers existing version

Fixes: #2923


lotas:
yeah, that CodeQL complaint is about writing files in concurrent system, so doesn't apply much for dev machines

PR #6639, opened 2023-10-26
docker-worker: stringify scope expressions in error message

jcristau:
It turns out seeing `[object Object]` in the task log isn't too helpful.

This is untested, and rather academic as we're not deploying updated docker-workers, but maybe it's trivial enough that it's still worth it?


lotas:
thanks @jcristau 
I hope someone would be able to deploy it still. 

We have stopped running tests for docker-worker already, but I don't think this feature requires tests

lotas (approved):

PR #6652, opened 2023-10-31
feat(worker-manager): Updated expired dev certs

lotas:
Steps are documented in: 
https://github.com/taskcluster/taskcluster/blob/main/services/worker-manager/README.md#testing


matt-boris (approved):
Awesome, thanks for updating Yarik!

Issue #6656, opened 2023-11-01
D2G should shell escape environment variable name


matt-boris:
If an environment variable name has spaces in it, single quotes need to be placed around the name before passing to the podman run command so that it's still properly formed.

PR #6660, opened 2023-11-01
fix(d2g): shell escape env var names

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6656.

>D2G now shell escapes environment variable key names in case they contain spaces or special characters that would previously mess up the `podman run...` command.

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -451,7 +451,7 @@ func createVolumeMountsString(dwPayload *dockerworker.DockerWorkerPayload, wdcs 
 }
 
 func podmanEnvSetting(envVarName string) string {
-	return fmt.Sprintf(" -e %s", envVarName)
+	return fmt.Sprintf(" -e %s", shell.Escape(envVarName))
 }
 
 func imageObject(payloadImage *json.RawMessage) (Image, error) {
--- a/tools/d2g/d2gtest/testdata/testcases/env_escaping_test.yml
+++ b/tools/d2g/d2gtest/testdata/testcases/env_escaping_test.yml
@@ -13,21 +13,21 @@ testSuite:
         image: ubuntu
         maxRunTime: 3600
         env:
-          test123 --help ; whoami ;: testing
+          ' test123 --help  ; whoami ; ': testing
       genericWorkerTaskPayload:
         command:
           - - bash
             - '-cx'
             - >-
               podman run -t --rm
+              -e ' test123 --help  ; whoami ; '
               -e RUN_ID
               -e TASKCLUSTER_ROOT_URL
               -e TASKCLUSTER_WORKER_LOCATION
               -e TASK_ID
-              -e 'test123 --help ; whoami ;'
               ubuntu 'echo "Hello world"'
         env:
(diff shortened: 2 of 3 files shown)


tools/d2g/d2g.go
```diff
@@ -502,7 +502,7 @@ func podmanEnvMappings(dwPayload *dockerworker.DockerWorkerPayload) string {
 
 	envVarNames := []string{}
 	for envVarName := range dwPayload.Env {
-		envVarNames = append(envVarNames, envVarName)
+		envVarNames = append(envVarNames, shell.Escape(envVarName))
```
petemoore, line 505: It looks like the env var name is being stored in its escaped form. I think it would be cleaner to store it in its raw form, but escape it when writing it the command.

tools/d2g/d2gtest/testdata/testcases/env_escaping_test.yml
```diff
@@ -0,0 +1,38 @@
...
+            - >-
+              podman run -t --rm
+              -e 'test123 --help ; whoami ;'
+              -e RUN_ID
+              -e TASKCLUSTER_ROOT_URL
+              -e TASKCLUSTER_WORKER_LOCATION
+              -e TASK_ID
+              ubuntu 'echo "Hello world"'
+        env:
+          test123 --help ; whoami ;: testing
```
petemoore, line 30: Is this correct? I'm surprised there is no string quoting here too. When this is converted to json, what does it look like? My understanding of yaml isn't sufficient to know how this is parsed.

tools/d2g/d2g.go
matt-boris: If `shell.Escape()` is called around this entire function as `shell.Escape(podmanEnvMappings(dwPayload))`, then it'll put single quotes around the ` -e ENV0 -e ENV1` list and podman won't interpret the flags properly at that point. The way I have it, only the `ENV0` or `ENV1` names will be shell escaped.

lotas (approved):
awesome!

tools/d2g/d2gtest/testdata/testcases/env_escaping_test.yml
matt-boris: Yeah, json output of this is:

```json
"env": {
    "test123 --help ; whoami ;": "testing"
},
```

tools/d2g/d2g.go
matt-boris: Ah I see what you mean. I can escape inside of `podmanEnvSetting()`.

tools/d2g/d2gtest/testdata/testcases/env_escaping_test.yml
matt-boris: Looks like it strips trailing spaces

matt-boris: Added in https://github.com/taskcluster/taskcluster/pull/6660/commits/f4ae953781fcc8d286f2748a069183afbcc2fea4.

PR #6646, opened 2023-10-27
chore: bump to yarn v4 (stable)

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6636.

Corepack a really neat new feature in nodejs that helps manage the package managers.

Two key core workflows it simplifies (from https://nodejs.org/api/corepack.html):
  * It eases new contributor onboarding, since they won't have to follow system-specific installation processes anymore just to have the package manager you want them to.
  * It allows you to ensure that everyone in your team will use exactly the package manager version you intend them to, without them having to manually synchronize it each time you need to make an update.

I followed the migration guide here https://yarnpkg.com/migration/guide#migration-steps.

>Upgrades to modern Yarn version 4.
>
>To enable on your machine:
>
>```bash
>corepack enable
>yarn set version berry
>```

PR #6678, opened 2023-11-09
chore: use `yarnPath` config and `stable` version

matt-boris:
Some small cleanup/changes after https://github.com/taskcluster/taskcluster/pull/6646.

These changes make it so locally, yarn will resolve to the version specified in source code. I had it switch back to legacy version a couple times on me. This will fix that.

This is also causing an issue when a new version of yarn is released. The CI was downloading it, then there was a diff in the source vs what's running on the CI due to the yarn version being updated.


matt-boris:
Testing a `staging-release` https://community-tc.services.mozilla.com/tasks/groups/LCU8k6YkRr2SnfFSi3376A.

petemoore (approved):
Awesome, thanks!

PR #6679, opened 2023-11-09
fix(generic-worker): validate the task user can read/execute the generic-worker binary

matt-boris:
As a follow-up to https://github.com/taskcluster/taskcluster/pull/6673, this change will check to ensure that the generic-worker binary is readable and executable by the task user. If not, the worker will exit with exit code 69, internal error.

>Generic Worker: Adds validation that the task user is able to read and execute the generic-worker binary on startup of the worker. If the task user is not able to read and execute the binary, the worker will exit with exit code 69, internal error.

### DIFF
--- a/changelog/cNPiQAqZTEeqDmtRcz-kKQ.md
+++ b/changelog/cNPiQAqZTEeqDmtRcz-kKQ.md
@@ -1,0 +1,4 @@
+audience: general
+level: patch
+---
+Generic Worker: Adds validation that the task user is able to read and execute the generic-worker binary on startup of the worker. If the task user is not able to read and execute the binary, the worker will exit with exit code 69, internal error.
--- a/ui/docs/reference/workers/generic-worker/installing.mdx
+++ b/ui/docs/reference/workers/generic-worker/installing.mdx
@@ -11,7 +11,9 @@ worker-runner, and contains some outdated references to the
 "legacy" deployment of Taskcluster at https://taskcluster.net.
 A critical instruction to follow is to ensure that the generic-worker
 binary is readable and executable by the task user. Make sure it's not only the
-root user who can read and execute the binary.
+root user who can read and execute the binary. If the generic-worker binary
+is not readable and executable by the task user, the worker will exit
+with exit code 69, internal error.
 
 They document _possible_ (and simple) ways to install and run generic-worker on
 various platforms. Real life production deployments may be integrated quite
--- a/workers/generic-worker/artifacts.go
+++ b/workers/generic-worker/artifacts.go
@@ -12,7 +12,6 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
-	"testing"
 	"time"
 
 	"github.com/taskcluster/httpbackoff/v3"
@@ -360,23 +359,7 @@ func (task *TaskRun) uploadArtifact(artifact artifacts.TaskArtifact) *CommandExe
 }
 
 func copyToTempFileAsTaskUser(filePath string) (tempFilePath string, err error) {
-	// We want to run generic-worker, which is os.Args[0] if we are running generic-worker, but if
-	// we are running tests, os.Args[0] will be the test executable, so then we use relative path to
-	// installed binary. This hack will go if we can impersonate the logged on user.
-	var exe string
-	if testing.Testing() {
-		exe = os.Getenv("GOPATH")
-		switch runtime.GOOS {
-		case "windows":
-			exe += `\bin\generic-worker.exe`
-		default:
-			exe += "/bin/generic-worker"
-		}
-	} else {
-		exe = os.Args[0]
-	}
-
-	cmd, err := gwCopyToTempFile(exe, filePath)
+	cmd, err := gwCopyToTempFile(filePath)
 	if err != nil {
 		return "", fmt.Errorf("Failed to create new command to copy file %s to temporary location as task user: %v", filePath, err)
 	}
--- a/workers/generic-worker/artifacts_multiuser.go
+++ b/workers/generic-worker/artifacts_multiuser.go
@@ -2,8 +2,11 @@
 
 package main
 
-import "github.com/taskcluster/taskcluster/v58/workers/generic-worker/process"
+import (
+	"github.com/taskcluster/taskcluster/v58/workers/generic-worker/process"
+	"github.com/taskcluster/taskcluster/v58/workers/generic-worker/runtime"
+)
 
-func gwCopyToTempFile(exe, filePath string) (*process.Command, error) {
-	return process.NewCommandNoOutputStreams([]string{exe, "copy-to-temp-file", "--copy-file", filePath}, taskContext.TaskDir, []string{}, taskContext.pd)
+func gwCopyToTempFile(filePath string) (*process.Command, error) {
+	return process.NewCommandNoOutputStreams([]string{runtime.GenericWorkerBinary(), "copy-to-temp-file", "--copy-file", filePath}, taskContext.TaskDir, []string{}, taskContext.pd)
 }
--- a/workers/generic-worker/fileutil/fileutil_posix.go
+++ b/workers/generic-worker/fileutil/fileutil_posix.go
@@ -4,6 +4,7 @@ package fileutil
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
 	"strconv"
 	"strings"
@@ -56,15 +57,25 @@ func SecureFiles(filepaths ...string) (err error) {
 	return nil
 }
 
-func UnsecureFiles(filepaths ...string) (err error) {
-	for _, path := range filepaths {
-		err = os.Chmod(
-			path,
-			0777,
-		)
-		if err != nil {
-			return
-		}
+func ResetPermissions(path string, permissions fs.FileMode) error {
+	return os.Chmod(
+		path,
+		permissions,
+	)
+}
+
+func GetPermissionsString(path string) (string, error) {
+	fileMode, err := GetPermissions(path)
+	if err != nil {
+		return "", err
 	}
-	return
+	return fileMode.String(), nil
+}
+
+func GetPermissions(path string) (fs.FileMode, error) {
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return 0, err
+	}
+	return fileInfo.Mode().Perm(), nil
 }
--- a/workers/generic-worker/fileutil/fileutil_posix.go
+++ b/workers/generic-worker/fileutil/fileutil_posix.go
@@ -55,3 +55,16 @@ func SecureFiles(filepaths ...string) (err error) {
 	}
 	return nil
 }
+
+func UnsecureFiles(filepaths ...string) (err error) {
+	for _, path := range filepaths {
+		err = os.Chmod(
+			path,
+			0777,
+		)
+		if err != nil {
+			return
+		}
+	}
+	return
+}
--- a/workers/generic-worker/fileutil/fileutil_posix.go
+++ b/workers/generic-worker/fileutil/fileutil_posix.go
@@ -30,12 +30,12 @@ func SecureFiles(filepaths ...string) (err error) {
 	uidString = strings.TrimSpace(uidString)
 	uid, err := strconv.Atoi(uidString)
 	if err != nil {
-		return fmt.Errorf("Could not convert %v to an integer uid: %v", uidString, err)
+		return fmt.Errorf("could not convert %v to an integer uid: %v", uidString, err)
 	}
 	gidString = strings.TrimSpace(gidString)
 	gid, err := strconv.Atoi(gidString)
 	if err != nil {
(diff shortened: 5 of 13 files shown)


changelog/cNPiQAqZTEeqDmtRcz-kKQ.md
```diff
@@ -0,0 +1,4 @@
+audience: general
+level: patch
+---
+Generic Worker: Adds validation that the task user is able to read and execute the generic-worker binary. If the task user is not able to read and execute the binary, the task will fail as `exception/internal-error`.
```
petemoore, line 4: Let's make this a start up check, so that it does not claim a task if it cannot upload artifacts.

ui/docs/reference/workers/generic-worker/installing.mdx
```diff
@@ -11,7 +11,9 @@ worker-runner, and contains some outdated references to the
 "legacy" deployment of Taskcluster at https://taskcluster.net.
 A critical instruction to follow is to ensure that the generic-worker
 binary is readable and executable by the task user. Make sure it's not only the
-root user who can read and execute the binary.
+root user who can read and execute the binary. If you try to submit a task
+and the generic-worker binary is not readable and executable by the task user,
+the task will resolve as `exception/internal-error`.
```
petemoore, line 16: we'll need to adjust this text here too

workers/generic-worker/artifacts.go
```diff
@@ -360,23 +359,7 @@ func (task *TaskRun) uploadArtifact(artifact artifacts.TaskArtifact) *CommandExe
...
-		exe = os.Getenv("GOPATH")
-		switch runtime.GOOS {
-		case "windows":
-			exe += `\bin\generic-worker.exe`
-		default:
-			exe += "/bin/generic-worker"
-		}
-	} else {
-		exe = os.Args[0]
-	}
```
matt-boris, line 377: Pulled this out into a helper to reduce duplicative code and this also makes it so some other functions don't need to pass the `exe` string through anymore.

https://github.com/taskcluster/taskcluster/pull/6679/files#diff-47bb9ac1b0971421de4be385184e1f8ba471bf0cc53654f0ee5e3b033457bb5b

workers/generic-worker/multiuser_test.go
```diff
@@ -56,3 +57,26 @@ func TestWhoAmI(t *testing.T) {
...
+	}
+	defer func() {
+		err := os.Rename(backupExe, exe)
+		if err != nil {
+			t.Fatalf("Could not move generic-worker binary to original location: %v", err)
+		}
+	}()
+
+	execute(t, INTERNAL_ERROR)
+}
```
petemoore, line 82: I don't think the generic-worker binary not existing is a likely scenario, since it would need to be removed between generic-worker starting and the generic-worker initialisation completing. Can we have a Windows helper that runs icacls to remove the Everyone execute permission, and a posix helper that runs os.Chmod to remove the "all" execute permission? This would hopefully ensure that the code that checks it is executable by the task user really is using the task user to check that, rather than if some code slipped in so that it was checking if root could execute it by mistake. This test I think wouldn't notice if generic-worker started running the check as root/LocalSystem instead of the task user.

workers/generic-worker/process/multiuser_windows.go
```diff
@@ -171,16 +171,7 @@ func GrantSIDWinstaAccess(sid string, pd *PlatformData) {
...
-		// we are running tests, os.Args[0] will be the test executable, so then we use relative path to
-		// installed binary. This hack will go if we can use ImpersonateLoggedOnUser / RevertToSelf instead.
-		var exe string
-		if testing.Testing() {
-			exe = os.Getenv("GOPATH") + `\bin\generic-worker.exe`
-		} else {
-			exe = os.Args[0]
-		}
-		cmd, err := NewCommand([]string{exe, "grant-winsta-access", "--sid", sid}, ".", []string{}, pd)
+		cmd, err := NewCommand([]string{runtime.GenericWorkerBinary(), "grant-winsta-access", "--sid", sid}, ".", []string{}, pd)
```
petemoore, line 174: nice!

workers/generic-worker/runtime/runtime.go
```diff
@@ -16,3 +20,23 @@ import (
...
+			exe += `\bin\generic-worker.exe`
+		default:
+			exe += "/bin/generic-worker"
+		}
+	} else {
+		exe = os.Args[0]
+	}
+
+	return exe
+}
```
petemoore, line 42: Thanks for moving this into runtime package! 👍

workers/generic-worker/runtime_multiuser.go
```diff
@@ -0,0 +1,12 @@
+//go:build multiuser
+
+package main
+
+import (
+	"github.com/taskcluster/taskcluster/v57/workers/generic-worker/process"
+	"github.com/taskcluster/taskcluster/v57/workers/generic-worker/runtime"
+)
+
+func gwVersion() (*process.Command, error) {
```
petemoore, line 10: It would be good to add a code comment that we do this as the task user, because really this is just checking that gw binary is executable, rather than just being a helper to see what the generic-worker version is (that we know already from the `version` global variable in the main package, iirc.

ui/docs/reference/workers/generic-worker/installing.mdx
```diff
@@ -11,7 +11,9 @@ worker-runner, and contains some outdated references to the
 "legacy" deployment of Taskcluster at https://taskcluster.net.
 A critical instruction to follow is to ensure that the generic-worker
 binary is readable and executable by the task user. Make sure it's not only the
-root user who can read and execute the binary.
+root user who can read and execute the binary. If you try to submit a task
```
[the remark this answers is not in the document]
matt-boris: Good catch, thank you. Was just too fast at updating what I had previously :)

workers/generic-worker/multiuser_test.go
matt-boris: Yeah, good call thanks Pete.

petemoore: fwiw I think the existing `SecureFiles` function has some prior art for running icacls on Windows and os.Chmod on posix, iirc

matt-boris: Yes, I plan on using that. But there's no reversal helper. I think I may just make a backup copy for the test, secure the actual binary, then test, then after it's complete, remove the secured file and restore the backup.

workers/generic-worker/artifacts_multiuser.go
```diff
@@ -2,8 +2,11 @@
...
-import "github.com/taskcluster/taskcluster/v58/workers/generic-worker/process"
+import (
+	"github.com/taskcluster/taskcluster/v58/workers/generic-worker/process"
+	"github.com/taskcluster/taskcluster/v58/workers/generic-worker/runtime"
+)
 
-func gwCopyToTempFile(exe, filePath string) (*process.Command, error) {
-	return process.NewCommandNoOutputStreams([]string{exe, "copy-to-temp-file", "--copy-file", filePath}, taskContext.TaskDir, []string{}, taskContext.pd)
+func gwCopyToTempFile(filePath string) (*process.Command, error) {
+	return process.NewCommandNoOutputStreams([]string{runtime.GenericWorkerBinary(), "copy-to-temp-file", "--copy-file", filePath}, taskContext.TaskDir, []string{}, taskContext.pd)
```
petemoore, line 11: I notice that the multiuser version uses `NewCommandNoOutputStreams` but the simple version uses `NewCommand`. Is there a reason for the difference?

workers/generic-worker/fileutil/fileutil_posix.go
```diff
@@ -55,3 +56,26 @@ func SecureFiles(filepaths ...string) (err error) {
 	}
 	return nil
 }
+
+func ResetPermissions(path string, permissions fs.FileMode) error {
```
petemoore, line 60: Shouldn't this be SetPermissions? I know that we call it when we want to reset the permissions, but the function body itself takes a path and some permissions, and applies them, so really it is just setting the applied permissions, I would say.

workers/generic-worker/fileutil/fileutil_windows.go
```diff
@@ -16,3 +19,15 @@ func SecureFiles(filepaths ...string) (err error) {
 	}
 	return
 }
+
+func ResetPermissions(path string, permissions fs.FileMode) error {
```
petemoore, line 23: It looks like permissions isn't used here. What does reset mean in this context? How does Windows know what the previous permissions were? Is it removing all permissions associated with the path, or applying some previous value?

workers/generic-worker/fileutil/fileutil_windows.go
```diff
@@ -16,3 +19,15 @@ func SecureFiles(filepaths ...string) (err error) {
...
+	return host.Run("icacls", path, "/reset", "/t")
+}
+
+func GetPermissionsString(path string) (string, error) {
+	return host.CombinedOutput("icacls", path)
+}
+
+func GetPermissions(path string) (fs.FileMode, error) {
+	return 0, fmt.Errorf("GetPermissions not implemented on Windows")
+}
```
petemoore, line 33: If it isn't implemented, why do we include it?

workers/generic-worker/main.go
```diff
@@ -691,6 +696,20 @@ func (task *TaskRun) validateJSON(input []byte, schema string) *CommandExecution
 	return MalformedPayloadError(fmt.Errorf("Validation of payload failed for task %v", task.TaskID))
 }
 
+func validateGenericWorkerBinary() error {
+	cmd, err := gwVersion()
+	if err != nil {
+		return fmt.Errorf("could not determine generic-worker binary version: %v", err)
```
petemoore, line 702: I think the error here means you couldn't generate the command to determine what the generic-worker binary is, doesn't it? I think at this point the command hasn't been executed, so this should be a panic if we can't even create the command, I think?

workers/generic-worker/main.go
```diff
@@ -691,6 +696,20 @@ func (task *TaskRun) validateJSON(input []byte, schema string) *CommandExecution
 	return MalformedPayloadError(fmt.Errorf("Validation of payload failed for task %v", task.TaskID))
 }
 
+func validateGenericWorkerBinary() error {
```
petemoore, line 699: Can you add a doc comment here to explain what the function does? It isn't totally clear just from the function name. Thanks!

workers/generic-worker/runtime_simple.go
```diff
@@ -0,0 +1,14 @@
...
+
+package main
+
+import (
+	"github.com/taskcluster/taskcluster/v58/workers/generic-worker/process"
+	"github.com/taskcluster/taskcluster/v58/workers/generic-worker/runtime"
+)
+
+// This is executed as the task user to ensure that the
+// generic-worker binary is readable/executable.
```
petemoore, line 11: Something like:

gwVersion executes `generic-worker --version` as the task user. This is useful to test that that the task user has permission to run generic-worker subcommands. It does not return the version as a result - the result is already available as a global variable in the main package. A non-nil error is returned if the command could not be executed.

We could call the function something like testExecutingGWSubcommandAsTaskUser to make it clearer that it isn't a function for getting the gw version.

workers/generic-worker/multiuser_test.go
```diff
@@ -56,3 +60,53 @@ func TestWhoAmI(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Could not get permissions (string) of GOPATH: %v", err)
+	}
+	var fileMode fs.FileMode
+	if runtime.GOOS != "windows" {
+		fileMode, err = fileutil.GetPermissions(goPath)
+		if err != nil {
+			t.Fatalf("Could not get permissions of GOPATH: %v", err)
+		}
+	}
```
petemoore, line 90: Could GetPermissions return a String representation of the permissions _and_ a function that resets the permissions back to these values? Then you wouldn't need to special case Windows and have both the GetPermissions and GetPermissionsString versions, you would just have a single function that returned a string and a reset function.

workers/generic-worker/fileutil/fileutil_windows.go
matt-boris: I had to make the function signature the same but have two different implementations between posix and windows. That's why `permissions` isn't used here, where it is on posix.

And `/reset`'s docs on https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/icacls says:

>`[/reset [/t] [/c] [/l] [/q]] 	Replaces ACLs with default inherited ACLs for all matching files.`

workers/generic-worker/fileutil/fileutil_windows.go
matt-boris: Because the multiuser test is calling GetPermissions so wouldn't there be a compilation error on windows if this wasn't defined?

petemoore (approved):
Looks great, many thanks!

PR #6687, opened 2023-11-15
fix(generic-worker): disallow relative paths in file mounts

matt-boris:
Fixes bugzilla bug 1858424.

>Generic Worker: Disallow relative file mounts. All relative file paths will result in the mount happening from within the task user's directory.


matt-boris:
We won't want to disallow relative paths, in case the user wants to mount in a directory outside the task user's directory. Closing for now.

PR #6690, opened 2023-11-16
Bug fix: Generic Worker: Process absolute paths in task payload correctly

### LINKED ISSUE #6689: Generic Worker evaluates absolute paths as relative in mounts and artifacts
There is a bug in Generic Worker which occurs when a task payload specifies a `path` to an artifact, or a `file` or `directory` inside a mount. Instead of respecting the provided path, Generic Worker effectively strips any leading slash and then evaluates the path as a relative path from the task directory. This is due to it internally using go's path/filepath package and evaluating the target path as:

```go
filepath.Join(<task directory>, <provided path>)
```

Instead, Generic Worker should first check if the provided path is absolute, and only join it to the task directory if it is not.

petemoore:
Github Bug/Issue: Fixes #6689


petemoore:
Converting back to draft, I realise I should probably add a test or two...

Issue #6682, opened 2023-11-14
large increase in claim expired tasks on Firefox CI cluster around the start of October, 2023


bhearsum:
As shown in https://sql.telemetry.mozilla.org/queries/95924/source?p_date=2023-03-01#236915.

The first small bump appears to be October 5th, followed by a much larger bump on October 10th. It drops a bit after that (but not back to normal levels), and then hits a new peak on November 8th.

bhearsum:
https://bugzilla.mozilla.org/show_bug.cgi?id=1859204 was filed about this as well.

bhearsum:
One thing I wondered was whether this was spot termination related. Here's an example of where this seems to have happened on a bitbar machine, which might disprove that?
```
provisionerId: proj-autophone
workerType: gecko-t-bitbar-gw-perf-a51
taskQueueId: proj-autophone/gecko-t-bitbar-gw-perf-a51
schedulerId: gecko-level-3
projectId: none
taskGroupId: f8jCP1KrSwKLNBdCfvLRVg
dependencies:
  - F9DsHj7jS0C2V0CzLyAFbA
  - FGME-1ZhR6ypEZHtJYFIgw
  - GizFup4qRx29HhWmDGXdHA
  - JvnukrsZTou0hotu606M0A
  - VzxY2QkvSXajQxzK1cJiAQ
  - cWGy4YtxQkGNtz0X1N4ATQ
  - dSJjH8_kRUSk1xLmr3W1dA
requires: all-completed
routes:
  - tc-treeherder.v2.autoland.9591bcc532c4599dba688d3555c6a0135139c39a
priority: low
retries: 5
created: '2023-10-09T16:54:53.829Z'
deadline: '2023-10-10T16:54:53.829Z'
expires: '2024-10-08T16:54:53.829Z'
scopes:
  - project:releng:services/tooltool/api/download/public
  - project:releng:services/tooltool/api/download/internal
  - secrets:get:project/perftest/gecko/level-3/perftest-login
payload:
  env:
    LANG: en_US.UTF-8
    PYTHON: python3
    NEED_XVFB: 'false'
    MOZ_FETCHES: >-
      [{"artifact": "public/build/browsertime.tar.zst", "extract": true, "task":
      "cWGy4YtxQkGNtz0X1N4ATQ"}, {"artifact": "public/build/fix-stacks.tar.zst",
      "extract": true, "task": "dSJjH8_kRUSk1xLmr3W1dA"}, {"artifact":
[...]

PR #6692, opened 2023-11-17
Increase `server.keepAliveTimeout`

lotas:
Tweaking `server.keepAliveTimeout` to fix downstream errors in reverse proxy and load balancer.

Default node's http `server.keepAliveTimeout` is `5s` which might be an issue when working behind a reverse proxy which has bigger timeouts.

To reduce number of `502` errors, application's keep alive timeout should be larger than the one of the reverse proxy,
and that in turn, should be larger than the Load Balancer's one.

This introduces new configuration option for server: `KEEP_ALIVE_TIMEOUT_SECONDS` that can be used in deployment-specific scenarios. 

Related to #6682

Load Balancer (GLB) has `10min` keep alive timeout
nginx reverse proxy (iprepd in fxci) has `2min` keep alive timeout
taskcluster services are using default node timeout which is `5s`

So what happens is that each of the downstream servers tries to keep the connection open to the upstream server, but if request comes in right before this time is out, the connection would be closed, and proxy server would not  have a chance to read the response back.

So Ideally GLB 10min -> reverse proxy 10min+X -> app  10min+X+Y to make this chain work without high number of 502s


petemoore (approved):
Got to be worth a try. We can always back out if it doesn't help. Thanks Yarik!

Issue #4998, opened 2021-09-22
client-py does not handle redirects in `getArtifact` endpoint


ahal:
**Describe the bug**
The `artifacts` endpoint [is supposed to return a 303 response](https://docs.taskcluster.net/docs/reference/platform/queue/api#getArtifact). Specifically "Response: the HTTP response to this method is a 303 redirect to the URL from which the artifact can be downloaded". However the Python client appears to [fail on any 303 status](https://github.com/taskcluster/taskcluster/blob/main/clients/client-py/taskcluster/client.py#L514). So rather than following the redirect (or returning the URL pointed at by the redirect as the schema says we are supposed to), we raise an exception.

**To Reproduce**

Steps to reproduce the behaviour:

```python
>>> import taskcluster
>>> queue = taskcluster.Queue({"rootUrl": "https://firefox-ci-tc.services.mozilla.com"})
>>> queue.getArtifact("cCEzo2wFTCmPe65GJyO7ng", 0, "public/taskgraph/diffs/diff_autoland-onpush.txt")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ahal/.pyenv/versions/tc-bug/lib/python3.9/site-packages/taskcluster/generated/queue.py", line 467, in getArtifact
    return self._makeApiCall(self.funcinfo["getArtifact"], *args, **kwargs)
  File "/home/ahal/.pyenv/versions/tc-bug/lib/python3.9/site-packages/taskcluster/client.py", line 270, in _makeApiCall
    response = self._makeHttpRequest(entry['method'], _route, payload)
  File "/home/ahal/.pyenv/versions/tc-bug/lib/python3.9/site-packages/taskcluster/client.py", line 538, in _makeHttpRequest
    raise exceptions.TaskclusterRestFailure(
taskcluster.exceptions.TaskclusterRestFailure: None
```

After adding some debugging, I determined the response code here is a `303` (as the docs mention is possible).

**Expected behavior**

I'm a bit unsure about what is actually expected here. The docs say that the response schema is:
```
{
    "url": <url>
}
```

and that it's up to the consumer to handle downloading it. But I know of a consumer that is using this API (reviewbot) and it seems to both avoid the `303` error as well as return the contents of the artifact (rather than the URL). So I'm not sure what is going on here.

Either we should return the artifact contents (and the docs are wrong), or we should return the URL (and there's a second bug in the implementation here).

[...]

PR #6693, opened 2023-11-17
fix(client-py): Don't raise exception on 300 status code

ahal:
Issue: #4998

### DIFF
--- a/changelog/issue-4998.md
+++ b/changelog/issue-4998.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 4998
+---
+This fixes a bug in the Python client where an exception was raised on 300 HTTP
+status.
--- a/clients/client-py/test/test_client.py
+++ b/clients/client-py/test/test_client.py
@@ -345,7 +345,7 @@ class ObjWithDotJson(object):
         return self.x
 
     def raise_for_status(self):
-        if self.status_code >= 300 or self.status_code < 200:
+        if self.status_code >= 400 or self.status_code < 200:
             raise requests.exceptions.HTTPError()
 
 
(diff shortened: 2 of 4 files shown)


clients/client-py/test/test_client.py
```diff
@@ -442,6 +442,14 @@ def test_success_fifth_try_connection_errors(client, apiPath):
         assert expected == v
 
 
+def test_redirect_status_code(client, apiPath):
```
ahal, line 445: I was going to add test for the async client, but noticed that there are no test fixtures set up there to make it easy, so decided to skip it. But would be nice to add tests for the async client at some point.

petemoore (approved):
This looks ok to me, but also happy for others to look over it with more python familiarity and who might actively be using the python client that might be affected (@matt-boris @lotas @djmitche @jschwartzentruber @marco-c).

changelog/issue-4998.md
```diff
@@ -0,0 +1,6 @@
+audience: users
+level: patch
```
petemoore, line 2: I'm torn between this being patch and being major.

clients/client-py/test/test_client.py
lotas: 👍 so we don't forget https://github.com/taskcluster/taskcluster/issues/6694

lotas (approved):

changelog/issue-4998.md
lotas: I'm not sure how much does this change break for existing integrations.

ahal: The workaround for this bug should keep working after this.

Also in general, it's possible for users to depend on the buggy behaviour of almost any bug, so personally I don't think it should be major.

matt-boris (approved):
Thanks ahal 👍🏻

PR #6699, opened 2023-11-22
build(deps): upgrade to rust 1.74.0 and run `cargo audit fix`

matt-boris:
>Upgrades rust version to 1.74.0 and ran `cargo audit fix` to fix insecure crates.


lotas (approved):
Thanks!!

PR #6705, opened 2023-11-30
chore(db)!: remove pg11 support

matt-boris:
Related to #5514. Now that we've upgraded all of our deployments' pg databases to v15, we can drop v11.

>Removes postgres v11 support.


petemoore (approved):
Thanks so much for seeing this through Boris, really appreciated!

PR #6729, opened 2023-12-06
chore(ci): separate out gw tests

matt-boris:


petemoore (approved):
Thanks Boris!

This is certainly a gain. Ideally it would be nice to have less task payload duplication, but it is certainly an improvement, so happy to approve. I also do know know how to share parts of the payload across different tasks in taskgraph, so maybe it isn't so easy to do, or would add more complexity than it removes(?) so I think this is good. 👍

PR #6734, opened 2023-12-08
ci: upgrade to Taskgraph 7.0.0

ahal:


ahal:
I think the failure is unrelated, I see this task failing on other PRs too.

matt-boris (approved):
Nice, thank you! I've triggered a rerun on that failing task.

matt-boris:
Not sure what's going on with our macstadium workers. Maybe this will resolve by the morning. If not, we'll look into it further. Thanks again for the patch!

PR #6735, opened 2023-12-08
chore: use alpine base images

matt-boris:
>Use alpine base images for generic worker dockerfile and for fetching latest certs for livelog and taskcluster proxy docker images.


workers/generic-worker/build.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
```
petemoore, line 1: Is this script definitely compatible with non-bash shells?
Just curious what the reason was for this line change.

petemoore (approved):

workers/generic-worker/build.sh
matt-boris: Alpine linux doesn't have bash installed by default and builds the generic worker image just fine while running this script. :)

PR #6745, opened 2023-12-21
fix(generic-worker): don't log out private redirect URLs

matt-boris:
>Generic Worker: No longer logs out the redirect URL for Redirect Artifacts whose name begins with `private/`.


changelog/bE5Htd8cQja6Jzk722vy1A.md
```diff
@@ -0,0 +1,4 @@
+audience: general
+level: patch
+---
+Generic Worker: No longer logs out the redirect URL for Redirect Artifacts whose name begins with `private/`.
```
petemoore: let's just change this to say that we don't log the url of interactive artifacts, since they are signed urls... something like that

petemoore (approved):
Perfect! Many thanks Boris!

PR #6760, opened 2024-01-04
Kubernetes config improvements

### LINKED ISSUE #6759: Switch k8s pods to use real healthcheck liveness probe instead of restarting pods daily
Current implementation makes pods restart daily by telling kubernetes it should be killed.
This results in connections being dropped and increased 502 rate. 

Note: 502 are not directly related to this problem and is rather a separate issue that needs to be solved around pod termination.

TODO: 

- Replace `livenessProbe` unconditional termination with periodic healtcheck endpoint call `/ping` to let k8s know if pod is still responsive. 
- Set memory limits to kill pods automatically in case of memory leaks (size tbd)

lotas:
This changes container lifecycle in kubernetes, therefor a major version.

- web containers no longer killed daily and will be running as long as process responds to `livenessProbe`. (There are no indications that containers leak memory or use large amounts of CPU when running longer than few days)
- non-web containers keep the old logic, but will be terminated after 5 days instead (experiment to see if this can be removed or better not)
- tweaks `initialDelaySeconds` for readiness probe, as `3` was too short and caused too many errors to be logged.
- additionally listens to `SIGINT`
- grace termination period increased to `120s` (k8s will not kill stopping pod for so long)
- `lifecycle.preStop` sleep timeout reduced to `10s` because this happens before container's process receives `SIGTERM`, and thus no `server.close()` is being called

Documentation is not yet updated as this is still an experiment to figure out proper combination of timeouts and calls.

How it should work theoretically:
- pod is being created and starts serving requests
- once it is being terminated (by autoscaler or deployment) k8s will run `lifecycle.preStop` hook, it will wait 10s and then main process would receive `SIGTERM`. (This delay is needed, I believe, to give enough time to `kubelet` and `kubelet-proxy` to exclude this pod from network endpoints and stop sending new traffic). 
- `SIGTERM` should be handled by main process, and `server.close()` would be invoked that, in theory, should wait until all open connections will be closed. Additionally, it would tell all open connections (keepalive) to close connection and destroy socket 
- when all connections will be closed, `server.close()` would terminate process and container should be finally killed
- this should all happen within `terminationGracePeriodSeconds`

Fixes #6759

### DIFF
--- a/infrastructure/tooling/templates/k8s/deployment.yaml
+++ b/infrastructure/tooling/templates/k8s/deployment.yaml
@@ -18,7 +18,7 @@ spec:
       # allow server to terminate gracefully while still serving existing connections
       # new connections would be rejected as server will be closed
       # and under normal conditions pod will stop faster than this timeout
-      terminationGracePeriodSeconds: 45
+      terminationGracePeriodSeconds: 120
       containers:
       - name: ${projectName}-${lowercase(procName)}
         image: '{{ .Values.dockerImage }}'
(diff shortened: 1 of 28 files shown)


infrastructure/tooling/templates/k8s/deployment.yaml
```diff
@@ -55,23 +55,29 @@ spec:
...
+            timeoutSeconds: 3
+            initialDelaySeconds: 15
+            periodSeconds: 30
+          else:
+            exec:
+              command:
+              - /bin/sh
+              - -c
+              - exit $(awk 'BEGIN{srand(); print (rand() > 0.3)}')
+            initialDelaySeconds: 432000 # 5 days
```
lotas, line 74: this is to experiment if we need this for background jobs, i.e. they leak memory or run slower after 1 day. Bump from 1d to 5d. 
Kinda to be safe, but maybe not necessary. We'll see

libraries/app/src/index.js
```diff
@@ -1,5 +1,5 @@
-import express from 'express';
 import _ from 'lodash';
+import express from 'express';
```
lotas, line 2: not sure how this line got reordered 🙈

matt-boris (approved):
Worth a shot! Thanks for the detailed PR description 👍🏻

Issue #6762, opened 2024-01-08
Generic Worker: Intermittent issues pulling docker image on `podman run ...` command


matt-boris:
```bash
Trying to pull docker.io/library/node:18.19.0...
Getting image source signatures
Copying blob sha256:1076f3af5a4574b5b93db1b41c8f578eed04571a64f3a4d247f024738e3a39bd
Copying blob sha256:917ee5330e73737d6095a802333d311648959399ff2c067150890162e720f863
Copying blob sha256:bc0734b949dcdcabe5bfdf0c8b9f44491e0fce04cb10c9c6e76282b9f6abdf01
Copying blob sha256:9b5852c9e7e7085c3f6de31e692245df85fa0ad9291c75b6d7ff176dfd7e7b79
Copying blob sha256:b5de22c0f5cd2ea2bb6c0524478db95bff5a294c99419ccd4a9d3ccc9873bad9
Copying blob sha256:b43bd898d5fbe0e1606380820047fd1e8b421722c9e69ac12757474305bd6702
Copying blob sha256:1076f3af5a4574b5b93db1b41c8f578eed04571a64f3a4d247f024738e3a39bd
Copying blob sha256:917ee5330e73737d6095a802333d311648959399ff2c067150890162e720f863
Copying blob sha256:9b5852c9e7e7085c3f6de31e692245df85fa0ad9291c75b6d7ff176dfd7e7b79
Copying blob sha256:b5de22c0f5cd2ea2bb6c0524478db95bff5a294c99419ccd4a9d3ccc9873bad9
Copying blob sha256:bc0734b949dcdcabe5bfdf0c8b9f44491e0fce04cb10c9c6e76282b9f6abdf01
Copying blob sha256:8de389b640c64a7082091e57ca04cb52f7c3c9c31811b339c193828cc92bad4f
Copying blob sha256:b43bd898d5fbe0e1606380820047fd1e8b421722c9e69ac12757474305bd6702
Copying blob sha256:8de389b640c64a7082091e57ca04cb52f7c3c9c31811b339c193828cc92bad4f
Copying blob sha256:88bd787cb576487a259144faa4f28fc286713908345b0547d1ca1a6809c3a96d
Copying blob sha256:88bd787cb576487a259144faa4f28fc286713908345b0547d1ca1a6809c3a96d
Copying config sha256:638c4f90a264b4d166219a0de84b8076051339601976283c4733842e11508584
Writing manifest to image destination
Storing signatures
[...]

PR #6763, opened 2024-01-08
fix(d2g): auto retry on exit code 128

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6762.

>Generic Worker: Tasks internally translated by D2G will add exit code 128 to the retry exit status array for retrying on an intermittent docker image pull issue.


lotas (approved):
lgtm

PR #6764, opened 2024-01-08
fix(release): fix path to kind.yml file for autogeneration on releases

matt-boris:
This path should've been updated in https://github.com/taskcluster/taskcluster/pull/6734.


lotas (approved):
awesome! thanks

PR #6765, opened 2024-01-08
[Bug] d2g subcommand changes done to add support for YAML file.

### LINKED ISSUE #6731: taskcluster d2g subcommand to support yaml
We now point users at the d2g subcommand if they use a docker worker payload. It is likely many users will now look at this utility and start using it, once this lands on firefox-ci.

I suspect several will stumble / give up, when they see that .taskcluster.yml file is yaml, and they need to submit json.

I think it would be reasonable for the d2g subcommand to automatically switch between json/yaml depending on what it receives. If it receives yaml, it outputs yaml back, and vice versa. We could also add a command line option to switch format, but that already might be overkill. Presumably if you have a task definition in one format, and want to change it, you are happy to receive the output in the same format that you sent it, because in all likelihood, you will just replace the input with the output in the original file you got it from.

anirudh-pandey:
## ✅ Tasks Done:-
- -f option with filename functionality done.
- WIP: task def coding done, testing remaining due to incomplete test case(payload option missing)
- WIP: Piped input coding remaining, need to detect file type passed.

<hr>

Github Bug/Issue: Fixes #6731

### DIFF
--- a/clients/client-shell/cmds/d2g/d2g.go
+++ b/clients/client-shell/cmds/d2g/d2g.go
@@ -15,6 +15,7 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 
 	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
 )
 
 func init() {
@@ -37,8 +38,14 @@ To convert a task definition (JSON), you must use the task definition flag (-t, 
 func convert(cmd *cobra.Command, args []string) (err error) {
 	isTaskDef, _ := cmd.Flags().GetBool("task-def")
 	filePath, _ := cmd.Flags().GetString("file")
+	// Default file extension is json(temporary, handled for piped input)
+	fileExtension := "json"
 
-	input, err := userInput(filePath)
+	// Read file extension(temporary, handled for piped input)
+	if len(filePath) >= 4 {
+		fileExtension = filePath[len(filePath)-4:]
+	}
+	input, err := userInput(filePath, fileExtension)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


clients/client-shell/cmds/d2g/d2g.go
```diff
@@ -37,8 +38,14 @@ To convert a task definition (JSON), you must use the task definition flag (-t,
...
 	filePath, _ := cmd.Flags().GetString("file")
+	// Default file extension is json(temporary, handled for piped input)
+	fileExtension := "json"
 
-	input, err := userInput(filePath)
+	// Read file extension(temporary, handled for piped input)
+	if len(filePath) >= 4 {
+		fileExtension = filePath[len(filePath)-4:]
+	}
+	input, err := userInput(filePath, fileExtension)
```
petemoore, line 48: Rather than checking the file extension, it would be better to check the content. There are many variations for the file extension, including, but not limited to no extension, `.yaml`, `.yml`, `.YAML`, `.YML`, `.json`, `.JSON`. What matters really is only whether the content can be successfully interpreted as a particular format, and if it passes the payload validation (i.e. includes required fields, etc). For this, it is only necessary to attempt to parse the content in a particular format, and then decide based on whether the data was successfully parsed, whether the data is of the particular format you are trying.

clients/client-shell/cmds/d2g/d2g.go
```diff
@@ -95,16 +102,23 @@ func convert(cmd *cobra.Command, args []string) (err error) {
 			return fmt.Errorf("failed to unmarshal generic worker task definition: %v", err)
```
petemoore, line 102: This would need to be changed. If you aren't able to unmarshal, it suggests the input format is not json. At this point, knowing it is not json, the next task is to see if it is yaml.

clients/client-shell/cmds/d2g/d2g.go
```diff
@@ -179,3 +207,8 @@ func validateJSON(input []byte, schema string) error {
 
 	return nil
 }
+
+func isYAMLExtension(fileExtension string) bool {
```
petemoore, line 211: As mentioned above, this is not the best way to determine the format of the content. Furthmore, this function would consider a file called `foo.notyaml` to be a yaml file. The standard library provides [filepath.Ext](https://pkg.go.dev/path/filepath#Ext) for this purpose which handles variable length extensions, but we shouldn't need to consider the filename, especially because the input can come from standard in, so we need a means to be able to deduce it without a filename anyway.

petemoore (changes requested):

PR #6773, opened 2024-01-12
feat(generic-worker): add `unarchive` subcommand

matt-boris:
>Generic Worker: adds `unarchive` subcommand to the `generic-worker` binary.


petemoore (approved):

PR #6777, opened 2024-01-15
feat(web): Fixes auth0 token fetching

lotas:
After making pods run for longer than 24h issue with token expiry was uncovered. Auth0 returned token that would expire in 24h but due to the bug in code, check never considered those as expired (off by 1000 times error)

Bug #1874568


services/web-server/src/login/strategies/mozilla-auth0.js
```diff
@@ -28,11 +28,7 @@ export default class MozillaAuth0 {
 
   // Get a personAPI instance, by requesting an API token as needed.
   // See https://github.com/mozilla-iam/cis/blob/f90ba5033785fd4fb14faf9f066e17356babb5aa/docs/PersonAPI.md#do-you-have-code-examples
-  async getPersonApi() {
-    if (this._personApi && new Date().getTime() / 1000 < this._personApiExp - 10) {
```
lotas, line 32: the real issue was here.
`_personApiExp` was already in ms.. so there was no need for an extra `/1000`

petemoore (approved):
Awesome! Looks perfect.

PR #6787, opened 2024-01-22
Generic Worker: exit with exit code 82 if chain of trust key missing

### LINKED ISSUE #6785: Generic Worker: dedicated exit code for missing chain of trust key
@aerickson requested a dedicated exit code if the (chain of trust) signing key is missing. This seems like a good idea.

petemoore:
Github Bug/Issue: Fixes #6785

### DIFF
--- a/changelog/issue-6785.md
+++ b/changelog/issue-6785.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: minor
+reference: issue 6785
+---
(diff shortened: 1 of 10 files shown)


matt-boris (dismissed):

matt-boris (approved):
Thanks Pete! I like separating this out into its own exit code 👍🏻

PR #6790, opened 2024-01-24
Fix Issue 6789

### LINKED ISSUE #6789: Generic Worker passes d2g-translated scopes to Taskcluster Proxy
**Describe the bug**
When running Docker Worker task payloads under Generic Worker, Generic Worker calls d2g to translate the task payload and scopes to its native format. The translated scopes are needed to determine whether the task has the permissions to perform the Generic Worker functions. However, since outbound requests to external taskcluster services via Taskcluster Proxy are limited to the task's scopes, and the temporary client created by the Queue that is passed to Generic Worker as `credentials` in the [`queue.claimWork`](https://docs.taskcluster.net/docs/reference/platform/queue/api#claimWork) and [`queue.reclaimTask`](https://docs.taskcluster.net/docs/reference/platform/queue/api#reclaimTask) responses is not authorized to use the converted scopes, the original task scopes should be passed to the Taskcluster Proxy as Authorized Scopes, rather than the d2g-converted scopes.

**To Reproduce**

Retrigger task [`DtABS8hFTDuEihBQYUGUkA`](https://community-tc.services.mozilla.com/tasks/DtABS8hFTDuEihBQYUGUkA).

**Expected behavior**

Generic Worker should keep a copy of the original task scopes from the task definition it claims, before converting the scopes, and pass these unaltered task scopes to Taskcluster Proxy as Authorized Scopes when it launches Taskcluster Proxy.

**Taskcluster version**
60.2.0

**Additional context**
Discovered while working on [bug 1545497 - Support /dev/kvm in the Android PGO worker](https://bugzilla.mozilla.org/show_bug.cgi?id=1545497).

petemoore:
Fixes #6789

This PR comprises two commits. The first adds a test, which should hopefully cause the CI to fail. The second provides a fix, which should hopefully cause the CI to pass!

### DIFF
--- a/changelog/issue-6789.md
+++ b/changelog/issue-6789.md
@@ -1,0 +1,15 @@
+audience: users
+level: patch
+reference: issue 6789
+---
+Generic Worker no longer modifies task scopes passed to Taskcluster Proxy.
+Previously there was a bug where Taskcluster Proxy would be passed the
+d2g-modified scopes by Generic Worker rather than the original task scopes from
+the task definition of the `queue.claimWork` response body. If the task was not
+also explicitly assigned the required generic-worker scopes, this would result
+in HTTP 401 errors from Taskcluster Proxy calls.
+
+This has now been fixed, so that it is sufficient for tasks with a Docker
+Worker payload to contain only Docker Worker scopes, not have the associated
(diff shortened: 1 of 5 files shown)


lotas (approved):
Thanks! LGTM

Issue #6802, opened 2024-01-26
spot instance terminations in GCP aren't handled correctly


bhearsum:
https://firefox-ci-tc.services.mozilla.com/tasks/GYYVwr5RS-61EM8otXtMNg/runs/0 is a task that I confirmed to be killed by a spot instance preemption. Logs show it being noticed and handled:
> Jan 25 11:17:12Z [translations-1-b-linux-v100-gpu-4-300g-zlyc9cmhtlqiaqxuzsym0q-1](https://my.papertrailapp.com/systems/translations-1-b-linux-v100-gpu-4-300g-zlyc9cmhtlqiaqxuzsym0q-1/events?focus=1687134496819937292&selected=1687134496819937292) [start-worker](https://my.papertrailapp.com/groups/1141234/events?q=program%3Astart-worker&focus=1687134496819937292&selected=1687134496819937292) 2024/01/25 11:17:11 GCP Metadata Service says termination is imminent
Jan 25 11:17:12Z [translations-1-b-linux-v100-gpu-4-300g-zlyc9cmhtlqiaqxuzsym0q-1](https://my.papertrailapp.com/systems/translations-1-b-linux-v100-gpu-4-300g-zlyc9cmhtlqiaqxuzsym0q-1/events?focus=1687134496819937294&selected=1687134496819937294) [start-worker](https://my.papertrailapp.com/groups/1141234/events?q=program%3Astart-worker&focus=1687134496819937294&selected=1687134496819937294) 2024/01/25 11:17:11 Got graceful-termination request with finish-tasks=false
Jan 25 11:17:12Z [translations-1-b-linux-v100-gpu-4-300g-zlyc9cmhtlqiaqxuzsym0q-1](https://my.papertrailapp.com/systems/translations-1-b-linux-v100-gpu-4-300g-zlyc9cmhtlqiaqxuzsym0q-1/events?focus=1687134496819937297&selected=1687134496819937297) [start-worker](https://my.papertrailapp.com/groups/1141234/events?q=program%3Astart-worker&focus=1687134496819937297&selected=1687134496819937297) 2024/01/25 11:17:11 Killing process tree with parent PID 1637... (0xc000308a20)
[...]

PR #6806, opened 2024-01-30
fix(worker-runner): use a hanging GET for GCP spot termination detection

matt-boris:
Will test on a staging pool to see if this helps with #6802.

>Worker Runner no longer polls the metadata service for the Google provider. Instead, we've added `?wait_for_change=true` to the endpoint to perform a hanging GET request that'll return as soon as the metadata has changed and the VM has been preempted.

### DIFF
--- a/tools/worker-runner/provider/google/google.go
+++ b/tools/worker-runner/provider/google/google.go
@@ -111,7 +111,7 @@ func (p *GoogleProvider) SetProtocol(proto *workerproto.Protocol) {
 }
 
 func (p *GoogleProvider) checkTerminationTime() bool {
-	value, err := p.metadataService.queryMetadata(TERMINATION_PATH)
+	value, err := p.metadataService.queryMetadata(TERMINATION_PATH + "?wait_for_change=true")
 	// if the file exists and contains TRUE, it's time to go away
 	if err == nil && value == "TRUE" {
 		log.Println("GCP Metadata Service says termination is imminent")
@@ -130,13 +130,10 @@ func (p *GoogleProvider) checkTerminationTime() bool {
 }
 
 func (p *GoogleProvider) WorkerStarted(state *run.State) error {
-	// start polling for graceful shutdown
-	p.terminationTicker = time.NewTicker(15 * time.Second)
 	p.proto.AddCapability("graceful-termination")
 
 	go func() {
 		for {
-			<-p.terminationTicker.C
 			log.Println("polling for termination-time")
 			p.checkTerminationTime()
 		}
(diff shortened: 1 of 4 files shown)


ahal:
This is fine as a test, but if we do proceed with this option, we should have some kind of retry logic per:

> Caveat: The metadata server may return the 503 response if it's temporarily unavailable (this is very rare, but happens), so certain retry logic is required. This especially true for the long-running second form (with ?wait_for_change=true), as the pending request may return at any time with the code 503. Your code should be ready to handle this and restart the query. curl does not return the HTTP error code directly, but you can use the fact that x=$(curl ....) expression returned an empty string if you scripting it; your criterion for positive detection of preemption is [[ $x == TRUE ]] in this case.

(from https://stackoverflow.com/questions/34005480/how-to-detect-system-acpi-g2-s5-soft-off-event-with-python-on-linux/57862925#57862925)

matt-boris:
> we should have some kind of retry logic

This uses http retry logic under the hood https://github.com/taskcluster/taskcluster/blob/4889be0ef8acccb855d9b162535d36ef3c786781/tools/worker-runner/provider/google/metadata.go#L58

tools/worker-runner/provider/google/google.go
```diff
@@ -130,13 +130,10 @@ func (p *GoogleProvider) checkTerminationTime() bool {
 }
 
 func (p *GoogleProvider) WorkerStarted(state *run.State) error {
-	// start polling for graceful shutdown
-	p.terminationTicker = time.NewTicker(15 * time.Second)
 	p.proto.AddCapability("graceful-termination")
 
 	go func() {
 		for {
-			<-p.terminationTicker.C
```
bhearsum, line 139: Just to be clear: removing this ticker means that we do this in an infinite loop with no wait time? (I think that's fine...just making sure I understand the code.)

matt-boris: So it's a little bit confusing, I agree. The infinite for loop could _probably_ be removed, as the `p.checkTerminationTime()` call will be the one waiting for the HTTP GET response. I left the for loop in here just in case the method returned without an actual response yet from the API. I didn't look much into https://pkg.go.dev/github.com/taskcluster/httpbackoff/v3@v3.1.0#ClientDo but I thought if it had a timeout or something I'd rather just retry the `wait_for_change=true` call.

bhearsum: Absolutely - yeah. I agree that keeping the loop is a good idea.

matt-boris: 👍🏻 yeah, mainly just a safeguard to be sure we're still listening

matt-boris:
The metadata service doesn't respond to the hanging API request before preempting the instance, so this approach won't work.

PR #6821, opened 2024-02-07
feat(ui): Fixes role/clients filtering in the UI

### LINKED ISSUE #6820: scopes UI doesn't filter correctly
Eg:
image

Which is filtering on `project:releng:ship-it:server:production` which is definitely not granted to the `anonymous` role.

lotas:
Fixes roles and clients view, which were not filtered properly:

image
image
image

Fixes #6820

### DIFF
--- a/ui/src/components/ClientScopesTable/index.jsx
+++ b/ui/src/components/ClientScopesTable/index.jsx
@@ -1,33 +1,40 @@
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
 import { arrayOf, func, shape, string } from 'prop-types';
-import {
-  filter,
-  pipe,
-  path,
-  map,
-  toLower,
-  identity,
-  includes,
-  contains,
-  length,
-  sort as rSort,
-} from 'ramda';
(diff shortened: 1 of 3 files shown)


ui/src/components/ClientScopesTable/index.jsx
```diff
@@ -56,60 +63,55 @@ export default class ClientScopesTable extends Component {
   clients = null;
 
   createSortedClientsConnection = memoize(
-    (clientsConnection, selectedScope) => {
-      const extractClients = pipe(
-        filter(
-          pipe(
-            path(['node', 'expandedScopes']),
-            filter(pipe(toLower, includes(selectedScope), length))
```
lotas, line 64: Problem was that those nested filter-pipes were not filtering properly.
I tried to figure out for a while how to debug ramda currying pipelines and decided that modern javascript provides all the tools to achieve same results in a more friendly way. 🤷 

I think we can also get rid of it at some point maybe

matt-boris (approved):
Nice, thanks Yarik 👍🏻

PR #6825, opened 2024-02-08
Go client doc fix

petemoore:
AMQP API links were in HTTP section, HTTP API links were in AMQP section 😬

Introduced in #2666.


lotas (approved):
nicely spotted!

PR #6826, opened 2024-02-08
Go client doc fix - some more fallout from #2666

petemoore:
This line break was removed in #2666 but was intentionally included so that the comment about auto-generation is a code comment, not a package comment. It has no value for users to see this comment in the generated go docs. They simply wish to know how to use the package, and do not care that the docs are auto-generated. Developers care, because they need to know they should not be making manual changes to the file, which is why it is included as a code comment, and should not be part of the go docs.

### DIFF
--- a/clients/client-go/codegenerator/model/model.go
+++ b/clients/client-go/codegenerator/model/model.go
@@ -233,18 +233,31 @@ func (apiDefs APIDefinitions) GenerateCode(goOutputDir string) {
 		FormatSourceAndSave(typesSourceFile, result.SourceCode)
 
 		fmt.Printf("Generating functions and methods for %s\n", job.Package)
-		content := `
-// The following code is AUTO-GENERATED. Please DO NOT edit.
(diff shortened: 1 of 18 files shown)


clients/client-go/codegenerator/model/model.go
```diff
@@ -243,6 +243,7 @@ func (apiDefs APIDefinitions) GenerateCode(goOutputDir string) {
 
 // This package was generated from the schema defined at
 // ` + apiDefs[i].URL + `
+
```
lotas, line 246: how could we make sure this newline wouldn't be removed in future by someone who would notice this extra blank line? 
Maybe worth commenting it in the code that this is intended?

lotas (approved):
👍

matt-boris (dismissed):

matt-boris (approved):
LGTM, but you need to add a changelog :)

matt-boris (approved):

PR #6828, opened 2024-02-08
chore: go cleanup

matt-boris:
>Switched to use `math/rand/v2` ([new in go1.22](https://tip.golang.org/doc/go1.22#math_rand_v2)), removed [deprecated](https://pkg.go.dev/golang.org/x/sys@v0.16.0/windows#OpenCurrentProcessToken) call to `windows.OpenCurrentProcessToken()`, fixed `staticcheck` errors, and added a `staticcheck` GitHub actions workflow for our repo.


petemoore (approved):
Looks great, thanks Boris!

tools/worker-runner/perms/perms_windows.go
```diff
@@ -160,12 +160,7 @@ func verifyPrivateToOwner(filename string) (err error) {
...
 // Get the SID of the current user
 func getCurrentUser() (sid *windows.SID, err error) {
-	token, err := windows.OpenCurrentProcessToken()
-	if err != nil {
-		return
-	}
-	defer token.Close()
-
+	token := windows.GetCurrentProcessToken()
 	tokenuser, err := token.GetTokenUser()
```
petemoore, line 164: Oh - this isn't something new you've added, but won't this `err` shadow the one that gets returned?

matt-boris: Addressed in https://github.com/taskcluster/taskcluster/pull/6829 :) Nice find!

PR #6829, opened 2024-02-09
fix(worker-runner): remove shadowing

matt-boris:
Should've been addressed in https://github.com/taskcluster/taskcluster/pull/6828.


petemoore (approved):
Perfect, thanks Boris!

PR #6830, opened 2024-02-12
Refreshes raw log url before opening it

### LINKED ISSUE #5033: The `Raw Log` link expires very quickly when public URLs signing is enabled
Guessing it's the S3 link that isn't refreshed when one clicks on the link

### LINKED ISSUE #6795: Task log page "Raw Log" button should not use stale signed URLs
**Describe the bug**
When clicking on "Raw Log" you can currently receive a 401 error (even for public artifacts!), if you opened then task log page more than 60s beforehand. It would be better to sign the url with a bewit at the time the user clicks "Raw Log" rather than when the page is rendered.
Note, #6724 is about not signing urls for public artifacts, which is related, but not the same as this issue.

lotas:
To avoid seeing link expired errors we  need to refresh it

Fixes #6795 
Fixes #5033

### DIFF
--- a/changelog/issue-6795.md
+++ b/changelog/issue-6795.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 6795
+---
+
+Fixes "Raw Log" button in UI that can point to an expired artifact.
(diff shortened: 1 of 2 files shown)


matt-boris (approved):
👍🏻

PR #6831, opened 2024-02-12
chore: update browserslist

matt-boris:
Used `npx update-browserslist-db@latest`.

Can read more about why we should do this regularly here https://github.com/browserslist/update-db?tab=readme-ov-file#readme.

Used the workflow example from https://github.com/c2corg/browserslist-update-action.


lotas (approved):
magic! 
thanks

PR #6846, opened 2024-02-21
Issue 6845: D2G support disableSeccomp capability

### LINKED ISSUE #6845: d2g: add support for task.payload.capabilities.disableSeccomp
disableSeccomp is a feature that was removed from Docker Worker, but was still used by bugmon project which pinned to a legacy version of Docker Worker.

In order to ease the migration from Docker Worker to Generic Worker, we should add support for this former Docker Worker feature to d2g, so that bugmon can be migrated to Generic Worker workers. See taskcluster/community-tc-config#721 for details.

petemoore:
Fixes #6845


lotas (approved):
LGTM!

PR #6849, opened 2024-02-22
Don't close the websocket when invalid UTF-8 is output by an interactive session

### LINKED ISSUE #6848: non-utf8 output in interactive tasks causes them to abort
non-utf8 output to the shell in an interactivate task causes the session to be closed. For example, cat'ing /dev/urandom is a quick way to reproduce (but there are also ways this can happen that happen in real life scenarios).

On the worker, we see:
```2024/02/20 20:57:44 Error occured: websocket: close 1006 (abnormal closure): unexpected EOF```

Debugging with Chrome, I found: "Could not decode a text frame as UTF-8."

The first error we in websocktunnel code comes from [copyWsData's call to Gorilla's Conn.NextReader](https://github.com/taskcluster/taskcluster/blob/3d8b26bd128c5cf085d652cb0960899a33712966/tools/websocktunnel/wsproxy/wsproxy.go#L141). After some additional debugging, it looks like that error originates from [advanceFrame](https://github.com/gorilla/websocket/blob/695e9095ce8736ac99c83939ca6b0fe93768f680/conn.go#L819-L821), and based on our ultimate message it seems like that we're hitting [this path which returns an unexpectedEOF](https://github.com/gorilla/websocket/blob/695e9095ce8736ac99c83939ca6b0fe93768f680/conn.go#L370-L372).

It _seems_ unlikely to me that this is a bug in the websocktunnel or the Gorilla library that we're using. My guess is that something is happening in the interactive task UI that's causing some invalid (or possibly empty) data to be sent to the websocktunnel, which is what's making Gorilla unhappy. There's clearly a call to [TextEncode for every keystroke](https://github.com/taskcluster/taskcluster/blob/3d8b26bd128c5cf085d652cb0960899a33712966/ui/src/components/Shell/index.jsx#L113), although I haven't been able to confirm what is or is not happening there.

@Eijebong - I don't suppose you have any insight here? It looks like you touched it last :grimacing: 

(See https://github.com/mozilla/firefox-translations-training/issues/355 for some early investigation of this.)

Eijebong:
There's 2 parts to this. First, the worker was using a `TextMessage` for websockets which would assume the message was valid UTF-8. This ended up being the main issue, it would just close the websocket when it saw invalid UTF-8.

The second part was the DECODER (which wasn't used in version 2 of interactive sessions until now), which was spamming the console with errors when it received invalid UTF-8.

Fixing both of those allows one to see invalid UTF-8 output in their interactive sessions.

I kept backwards compatibility with generic-worker which are still sending `TextMessage` messages by checking that the received data is a string and kept the old behavior in that case.

Fixes #6848

### DIFF
--- a/changelog/issue-6848.md
+++ b/changelog/issue-6848.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 6848
+---
+
+Fix an issue where an interactive session would close up when the shell would output invalid UTF-8.
--- a/ui/src/components/Shell/index.jsx
+++ b/ui/src/components/Shell/index.jsx
@@ -5,7 +5,7 @@ import { hterm, lib } from 'hterm-umdjs';
 import { DockerExecClient } from 'docker-exec-websocket-client';
 import withAlertOnClose from '../../utils/withAlertOnClose';
 
-const DECODER = new TextDecoder('utf-8');
+const DECODER = new TextDecoder('utf-8', { fatal: false });
 const defaultCommand = [
   'sh',
   '-c',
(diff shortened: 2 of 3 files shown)


Eijebong:
image
:)

petemoore (approved):
Fantastic! Thanks Bastien!

PR #6851, opened 2024-02-22
Add a proper `TERM` environment variable to interactive sessions

### LINKED ISSUE #6850: docker worker interactive tasks do not work properly curses applications
From a user report:
> the output to the screen is garbled and input doesn't seem to work correctly or it's just confusing with the garbled text
> I'm seeing two odd things, tmux is now not working either.
> tmux says "open terminal failed: terminal does not support clear"
> And then when I load vim the old output is still on the screen.

image

I gave this a try myself, and it looks to me like they're behaving similarly to when you use  `docker run` without `-t`. It's quite clear in the code that we are [requesting a tty](https://github.com/taskcluster/taskcluster/blob/3d8b26bd128c5cf085d652cb0960899a33712966/ui/src/components/Shell/index.jsx#L56-L81) though, and resizing works (which also requires a tty) - so this is probably a red herring.

@Eijebong - I'm not sure if know much about this one, but I'll flag you anyways on the off chance :)

Eijebong:
The `TERM` environment variable is used to indicate the capabilities of the current terminal emulator. Applications use it to know whether they can use some features or not. An example of that is tmux which looks for the `clear` capability which is missing from the default `TERM` assigned by bash (`dumb`).
Thankfully this was fairly easy as `hterm` is a supported terminal.

Github Bug/Issue: Fixes #6850

### DIFF
--- a/changelog/issue-6850.md
+++ b/changelog/issue-6850.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 6850
+---
+Add a proper TERM environment variable to interative sessions. This helps with some ncurses apps and tmux for example.
--- a/workers/generic-worker/multiuser_posix.go
+++ b/workers/generic-worker/multiuser_posix.go
@@ -66,10 +66,13 @@ func (task *TaskRun) generateInteractiveCommand(ctx context.Context) (*exec.Cmd,
 	var processCmd *process.Command
 	var err error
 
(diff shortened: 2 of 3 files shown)


lotas (approved):
Thanks Bastien!
this is a very nice and simple fix 👍

petemoore (approved):
Nice! Thanks Bastien!

PR #6856, opened 2024-02-22
Fix docker worker shell UI to set row/cols properly

bhearsum:
It looks like this is broken on first load, due to the second `resize` call, as well as on resize (onTerminalResize and DockerExecClient.resize's arg order doesn't match).

There's an initial `this.client.resize.apply` call which is correct...and seems like it doesn't need to exist...but this is working fine so it seems best not to try to remove it.

### DIFF
--- a/ui/src/components/Shell/index.jsx
+++ b/ui/src/components/Shell/index.jsx
@@ -92,10 +92,12 @@ export default class Shell extends Component {
           });
 
           this.client.resize(
-            terminal.screenSize.width,
-            terminal.screenSize.height
+            terminal.screenSize.height,
+            terminal.screenSize.width
           );
-          io.onTerminalResize = (c, r) => this.client.resize(c, r);
+          // onTerminalResize provides width then height;
+          // DockerExecClient.resize needs height then width
+          io.onTerminalResize = (cols, rows) => this.client.resize(rows, cols);
           this.client.stdout.on('data', data =>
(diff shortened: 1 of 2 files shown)


ui/src/components/Shell/index.jsx
```diff
@@ -92,10 +92,12 @@ export default class Shell extends Component {
           });
 
           this.client.resize(
+            terminal.screenSize.height,
             terminal.screenSize.width,
-            terminal.screenSize.height
           );
-          io.onTerminalResize = (c, r) => this.client.resize(c, r);
+          // onTerminalSize provides width then height; DockerExecClient.resize
```
matt-boris, line 98: nit:

```suggestion
          // onTerminalResize provides width then height; DockerExecClient.resize
```

matt-boris (commented):
Thanks Ben 👍🏻

ui/src/components/Shell/index.jsx
```diff
@@ -92,10 +92,12 @@ export default class Shell extends Component {
           });
 
           this.client.resize(
+            terminal.screenSize.height,
```
matt-boris, line 95: And it looks like you need to remove the trailing comma after `terminal.screenSize.width` now.

matt-boris (approved):

matt-boris (approved):

PR #6861, opened 2024-02-29
Fix smoketest for web-server

lotas:
No longer needed to treat it differently from the rest of the services, since apollo server does not expose healthcheck endpoints anymore


petemoore (dismissed):
Nice simplification!

petemoore (approved):

PR #6863, opened 2024-03-01
Fix validation error for hooks page

lotas:
Since validationRules property was misplaced before, it didn't do anything, so this problem wasn't caught. After it complained about hooksGroup query validation complexity. As that query includes lists and queries external entities with additional queries, complexity rises to ~4000.

### DIFF
--- a/services/web-server/src/main.js
+++ b/services/web-server/src/main.js
@@ -150,7 +150,7 @@ const load = loader(
           validationRules: [
             queryLimit(1000),
             depthLimit(10),
-            createComplexityLimitRule(1000),
+            createComplexityLimitRule(4500),
           ],
         });
         await server.start();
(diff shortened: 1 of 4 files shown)


services/web-server/src/main.js
```diff
@@ -150,7 +150,7 @@ const load = loader(
           validationRules: [
             queryLimit(1000),
             depthLimit(10),
-            createComplexityLimitRule(1000),
+            createComplexityLimitRule(4500),
```
petemoore, line 153: lol

petemoore (approved):
Looks reasonable, thanks Yarik!

services/web-server/src/main.js
lotas: 🙈 .. hooks query is quite aggressive.. fetching all groups+hooks, and then fetching last fire details, which combined produces a very complex query.. but relatively fast despite

petemoore (approved):

PR #6865, opened 2024-03-01
Issue 6864: D2G: pass --privileged to podman run when passing --device=/dev/shm

### LINKED ISSUE #6864: d2g: payload.capabilities.devices.hostSharedMemory requires --privileged under podman
**Describe the bug**
d2g translates the Docker Worker device capability `hostSharedMemory` into an additional `podman run` command argument `--device=/dev/shm`. However, in isolation, this appears not to provide access to `/dev/shm` which appears to be rectified when passing the additional argument `--privileged`.

Example failing task:
  * https://community-tc.services.mozilla.com/tasks/a9ERETC4TY2etXKijCxbqg/runs/5/logs/public/logs/live.log

Example successful task, when `--privileged` is added to the `podman run` command:
  * https://community-tc.services.mozilla.com/tasks/BUbILqarQhSVCLUMrwQEQA/runs/0/logs/live/public/logs/live.log

I tried an alternative version, using `--group-add keep-group` instead of `--privileged` here:
  * https://community-tc.services.mozilla.com/tasks/ccf7rsOwRou2LZo4CTVJ-A/runs/0/logs/live/public/logs/live.log
after reading [this article](https://www.redhat.com/sysadmin/files-devices-podman). However, this was unsuccessful, therefore I don't think this is an appropriate fix, and `--privileged` is a more appropriate option.

Please note, the `podman run` command is still being run as a non-privileged user, so there is no increased security concern by passing the `--privileged` flag, since the podman command does not have elevated privileges.

petemoore:
Fixes #6864

### DIFF
--- a/changelog/issue-6864.md
+++ b/changelog/issue-6864.md
@@ -1,0 +1,10 @@
+audience: users
+level: patch
+reference: issue 6864
+---
+D2G now passes `--privileged` flag to the generated `podman run` command when
+Docker Worker payload enables device capability `hostSharedMemory`.  Without
+this option, the podman container could not successfully access the shared
+memory, despite the inclusion of argument `--device=/dev/shm`. With both
+arguments present (`--privileged` and `--device=/dev/shm`), shared memory now
+appears to be available inside the podman container.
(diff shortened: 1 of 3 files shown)


lotas (approved):
👍

PR #6878, opened 2024-03-01
build(deps): upgrade to node v20 (new LTS)

matt-boris:
>Upgrades to Node.js LTS v20.


lotas:
Related: https://github.com/tschaub/mock-fs/issues/380, https://github.com/tschaub/mock-fs/issues/381

Node 20.8 started optimizing sync fs functions.
I noticed that mocks for `references` at least can be fixed by wrapping test into a setTimeout.
Comments also suggest that it can be resolved by using async fs functions

lotas (approved):
awesome! thanks for the updates :)

matt-boris:
Thank _you_ too Yarik! 🎉

PR #6879, opened 2024-03-05
build(deps): bump to go1.22.1

matt-boris:
>Upgrades to go1.22.1 which is a [security release](https://go.dev/doc/devel/release#go1.22.1).


petemoore (approved):
Nice job Boris!

PR #6883, opened 2024-03-06
fix(generic-worker): fix malformed payload log when enabling interactive task

matt-boris:
>Generic Worker: fixed incorrect log line when an Interactive Task fails due to the worker pool config not being set.

### DIFF
--- a/changelog/Ap0wKwVAT-CSQUTGw7gOtw.md
+++ b/changelog/Ap0wKwVAT-CSQUTGw7gOtw.md
@@ -1,4 +1,4 @@
 audience: users
 level: patch
 ---
-Generic Worker: fixed incorrect log line when an Interactive Task fails due to the worker pool config not being set.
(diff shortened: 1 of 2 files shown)


changelog/Ap0wKwVAT-CSQUTGw7gOtw.md
```diff
@@ -0,0 +1,4 @@
+audience: users
+level: patch
+---
+Generic Worker: fixed incorrect log line when an Interactive Task fails due to the worker pool config not being set.
```
petemoore, line 4:
```suggestion
Generic Worker now correctly reports the Worker Pool ID when an interactive task is attempted on a worker pool with the interactive feature disabled. Previously the task log would report the Worker Pool ID in the `exception/malformed-payload` task run as `<workerGroup>/<workerType>`; now it correctly reports it as `<provisionerId>/<workerType>`. The Interactive feature is considered disabled when Generic Worker config setting `enableInteractive` is either absent or explicitly set to `false` in the config.
```

petemoore (changes requested):
Thanks Boris! Nice spot! :-)

PR #6891, opened 2024-03-08
Issue 6890: D2G: pass --privileged to all generated podman run commands

### LINKED ISSUE #6890: D2G: several tasks seem to need --privileged flag passed to podman run
There appear to be an increasing number of tasks that ran under docker-worker without privileged capability that are not running successfully under `podman` without `--privileged` flag. Since for podman, "privileged" doesn't increase privileges of task, it just increases the privileges inside the container that the task user already has outside the container, let's just pass `--privileged` regardless. This should more closely match the privileges that the docker worker tasks have.

Examples:
  * Bugmon monitor tasks (fuzzing)
    * Issue #6888 
    * Task: https://community-tc.services.mozilla.com/tasks/evzR-_XzS0mKjTDanC8G0g
  * Loopback video task
    * No issue
    * https://firefox-ci-tc.services.mozilla.com/tasks/U5dS50DdRlqh7sOztqqjSQ/runs/0

petemoore:
Fixes #6890

### DIFF
--- a/changelog/issue-6890.md
+++ b/changelog/issue-6890.md
@@ -1,0 +1,10 @@
+audience: users
+level: patch
+reference: issue 6890
+---
+D2G now always passes `--privileged` to the generated `podman run` command.
+Without this option, some tasks that ran successfully under Docker Worker,
+including tasks without Docker Worker capabilities, would not run correctly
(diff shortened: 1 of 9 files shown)


matt-boris (approved):
LGTM 👍🏻

Issue #6892, opened 2024-03-08
Worker Manager: support test payloads in registerWorker 


petemoore:
For the purposes of integration testing, it is useful for worker tests to be able to call real taskcluster services in addition to mocks. This way, if the behavior of mocks and real services unexpectedly deviate, we can catch the issues before making a release.

Currently there is no way to successfully call `workerManager.registerWorker` for a non-`static` worker unless Worker Manager created the worker where the API call originates, which is not the case for integration testing. This is intentional in a production scenario, however severely limiting in an integration testing scenario.

In order to address this, it would be useful to reserve a special `provisionerId` (`test-provisioner-id`) that when used will slightly alter the behaviour of `workerManager.registerWorker`. Rather than looking up the worker in the database of Worker-Manager-spawned workers, and only returning successfully if exactly one suitable worker can be found, instead, the worker credentials, secret, and expiry can be given fake (reserved) values, and the worker config can be returned as if the call was successful.

PR #6893, opened 2024-03-08
workerManager.registerWorker: treat test-provisioner-id specially

petemoore:
Fixes: #6892


petemoore:
Just realised, to get the worker config, you need the talk to the provider code, and the provider then needs to know the region, for which it needs to know what worker was created.

I might end up abandoning this... It might not be so easy to special case. For the fake worker pool definitions though, it would be sufficient to pull out the first workerConfig section from the first launchConfig, we don't need to do a region-lookup. So the test would create a fake worker pool definition like [this one](https://community-tc.services.mozilla.com/worker-manager/test-provisioner%2Ftest-kmkji2cutniadgbb6aewg-a) containing a single workerConfig in a single launch config, and registerWorker could return it.

But it only makes sense to implement this if it is a light touchpoint to the code e.g. inside a simple `if` statement to check the worker pool id, and pass back what it has before doing the more complex queries for other information about the worker that was created, that can be stubbed out with fake data. In the end the worker integration tests just care that they can get their config back, and the calls work to retrieve it.

PR #6895, opened 2024-03-12
chore(generic-worker): use `filepath.WalkDir`

matt-boris:
>Generic Worker now utilizes `filepath.WalkDir` instead of `filepath.Walk`.
>
>`filepath.WalkDir` was introduced in go1.16 and is more performant and efficient over `filepath.Walk`.
>
>This _may_ help with race conditions during artifact uploads, where a file was initially seen, but then became unavailable at upload time.

From https://pkg.go.dev/path/filepath#Walk:
>Walk is less efficient than [WalkDir](https://pkg.go.dev/path/filepath#WalkDir), introduced in Go 1.16, which avoids calling os.Lstat on every visited file or directory.

### DIFF
--- a/changelog/XqmVDnNSQeaUj-43ZuK1Sw.md
+++ b/changelog/XqmVDnNSQeaUj-43ZuK1Sw.md
@@ -1,0 +1,8 @@
+audience: general
+level: patch
+---
+Generic Worker now utilizes `filepath.WalkDir` instead of `filepath.Walk`.
+
+`filepath.WalkDir` was introduced in go1.16 and is more performant and efficient over `filepath.Walk`.
+
+This _may_ help with race conditions during artifact uploads, where a file was initially seen, but then became unavailable at upload time.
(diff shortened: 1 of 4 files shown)


djmitche:
We saw this sort of thing with docker-worker, too. If files are appearing or disappearing while artifact uploads are going on, that does seem like user error. Maybe those errors should be caught and included as warnings in the logs?

matt-boris:
Yeah, there seems to be a strange issue with a task the fuzzing team is running where it's trying to upload what appears to be some sort of mutex/lock from the temp directory. I'm not sure the use case, but these changes may help out a bit.

In generic worker, we copy the file to a temporary location and upload from that location to help with issues like these, but seems like there's a race between walking the file tree and beginning to copy that file to a temp location.

In the case where the file isn't readable, the task fails with an appropriate message https://community-tc.services.mozilla.com/tasks/K5_NG8pwRRaZAN9MmXs1ZQ/runs/0/logs/public/logs/live.log#L194.

PR #6899, opened 2024-03-14
chore(generic-worker)!: rename `simple` engine to `insecure` engine

### LINKED ISSUE #6832: Rename simple engine to make it apparent it should not be used in production environments
Simple engine really should not be used in production, since it is inherently insecure. The only real use case is for development setups and unit testing. Update the engine name to reflect this, as it is commonly thought to be suitable for production.

matt-boris:
Fixes #6832.

>The Generic Worker `simple` engine has been renamed to the `insecure` engine.
>
>All future release binaries for this engine will also be renamed (e.g. `generic-worker-simple-darwin-arm64` --> `generic-worker-insecure-darwin-arm64`), so please update any scripts that reference the `simple` engine binary.
>
>This change was made to help make it extremely apparent that it should not be used in production environments and is only recommened for testing and development.


lotas (approved):
Thanks Boris, looks good!

PR #6915, opened 2024-03-19
feat: introduce new API for batch task requests

Alphare:
This is done according to RFC 189¹. In few words, the motivation is to drastically reduce the number of index API calls done by decision tasks, since they sometimes need the task ids for hundreds of tasks.

[1] https://github.com/taskcluster/taskcluster-rfcs/pull/189


Alphare:
I'm submitting this to get some feedback to make sure I'm headed in the right direction. I have a couple of questions inside (in XXX) about auth and validation, I'd be glad to get some pointers there. :)

services/index/src/api.js
```diff
@@ -66,6 +66,37 @@ builder.declare({
...
+  input: 'list-tasks-at-index.yml',
+  stability: APIBuilder.stability.experimental,
+  category: 'Index Service',
+  output: 'list-tasks-response.yml',
+  title: 'Find tasks at indexes',
+  description: [
+    'List the tasks given their labels. XXX description',
+  ].join('\n'),
+}, async function (req, res) {
+  // XXX what about authorization?
```
lotas, line 85: Each method has own `scopes` that is being used by api builder to validate if the caller has necessary scopes and he can make that call.

So far I see `index` services is using several scopes:

`index:find-task:<indexPath>` (`findTask`)
`index:list-tasks:<namespace>` (`listTasks`)

where `<indexPath>` and `<namespace>` are parameters that could be passed to the authz step when they are part of the route (i.e. `/task/:indexPath`)
If not, it is necessary to pass those manually via calling `req.authorize({ indexPath: xx })`

In case there will be multiple indexes passed, we would need to construct `scopes` in a way that all of those indexes would be accounted, for example:

```js
scopes: { AllOf: [{
    for: 'indexPath',
    in: 'indexPaths',
    each: 'index:find-task:<indexPath>'
}], }

// with later call:
req.authorize({
    indexPaths: req.body.indexes //  if it's array
});
```

this way we'll make sure that the caller has permissions for all of the tasks he is trying to find. I think it would be simpler to assume initially he needs to have `allOf` those and not to have to deal with partial access.

services/index/src/api.js
```diff
@@ -66,6 +66,37 @@ builder.declare({
...
+  const indexes = req.body.indexes;
+  // XXX where does the validation happen?
```
lotas, line 87: This is the `input: 'list-tasks-at-index.yml` (maybe add a `-request` to the name?) part of the route definition. All input will be validated automatically and rejected if validation fails

services/index/src/helpers.js
```diff
@@ -169,6 +169,35 @@ export const taskUtils = {
     // Fetch results
     return fetchResults(query ? query.continuationToken : {});
   },
+
+  async findTasksAtIndexes(db, { indexes }, { query } = {}) {
+    assert(_.isArray(indexes), 'indexes must be an Array');
```
lotas, line 174: `input` schema validation would make sure only valid payloads would be passed, but it doesn't hurt making it future-proof also 👍

lotas: sure :)

services/index/test/api_test.js
```diff
@@ -216,7 +216,67 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
         assert(err.statusCode === 404, 'Should have returned 404');
         return;
       }
+      assert(false, "should have caught");
```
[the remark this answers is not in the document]
lotas: let's keep it, thanks!

db/test/fns/index_test.js
```diff
@@ -192,16 +192,76 @@ suite(testing.suiteName(), function() {
...
+        delete data.taskId;
+        expectedTasks.push(data);
+      }
+
+      // Full query
+      let rows = await db.fns.get_tasks_from_indexes(JSON.stringify(expectedIndexes), 1000, 0);
+      assert.equal(rows.length, expectedIndexes.length);
+      assert.deepStrictEqual(rows, expectedTasks);
+
+      // Order of the input doesn't matter
```
lotas, line 249: nice

lotas (approved):
Nice! This looks great, thanks for the contirbution

PR #6917, opened 2024-03-20
Ensure 'taskGroup' is always defined when fetching scopes for submitting an action task

bhearsum:
This is in support of making actions like "Cancel All" work on task groups that have been created by an Action Task (see https://github.com/mozilla/firefox-translations-training/issues/250).

In an ideal world I'd like to understanding how `task.decisionTask` fits into things a bit more, and see if it could be removed...but that's more work than I'm willing to put in at the moment. I think this is a fairly safe fix to take in the meantime.

### DIFF
--- a/ui/src/views/Tasks/submitTaskAction.js
+++ b/ui/src/views/Tasks/submitTaskAction.js
@@ -13,7 +13,19 @@ import ajv from '../../utils/ajv';
 
 export default async ({ task, form, action, apolloClient, taskActions }) => {
   const actions = removeKeys(cloneDeep(taskActions), ['__typename']);
-  const taskGroup = task.taskId === task.taskGroupId ? task : task.decisionTask;
+  // We need to source scopes from somewhere. Typically these come from the
+  // task we're given, which is either a regular decision task or an action
+  // task. Decision tasks are easily identified, because their taskId matches
+  // the taskGroupId that they are in. Action tasks do not have such a
+  // mapping. Instead, we're simply going to try our best by falling back to
+  // `task` if `task.decisionTask` is null or undefined. At the time of
+  // writing, it's not clear whether the `task.decisionTask` case is even
+  // valid anymore, but it's being kept for the time being because verifying
+  // that is not a trivial matter.
+  const taskGroup =
+    task.taskId === task.taskGroupId || !task.decisionTask
+      ? task
+      : task.decisionTask;
   const input = load(form);
   const validate = ajv.compile(action.schema || {});
   const valid = validate(input);
--- a/changelog/N39Wr4udTH-1jL1N6rTggw.md
+++ b/changelog/N39Wr4udTH-1jL1N6rTggw.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+---
+
+Action tasks now work correctly for task groups created by another action task.


ui/src/views/Tasks/submitTaskAction.js
```diff
@@ -13,7 +13,18 @@ import ajv from '../../utils/ajv';
...
   const actions = removeKeys(cloneDeep(taskActions), ['__typename']);
-  const taskGroup = task.taskId === task.taskGroupId ? task : task.decisionTask;
+  // We need to source scopes from somewhere. Typically these come from the task we're given,
+  // which is either a regular decision task or an action task. Decision tasks are easily
+  // identified, because their taskId matches the taskGroupId that they are in. Action tasks
+  // do not have such a mapping. Instead, we're simply going to try our best by falling back
+  // to `task` if `task.decisionTask` is null or undefined.
+  // At the time of writing, it's not clear whether the `task.decisionTask` case is even valid
+  // anymore, but it's being kept for the time being because verifying that is not a trivial
+  // matter.
```
gabrielBusta, line 23: What is `task` when you trigger the action on a task group? The task group's UUID?

lotas (approved):
Good catch!

`task.decisionTask` is graphql thing only used in UI, but definitely possible it can be missing if task was fetched or created not through those graphql calls

Thanks

ui/src/views/Tasks/submitTaskAction.js
lotas: good point. I think it should always be `task` object, checked code. Only used in places where `task` is guaranteed to be defined correctly

bhearsum: It's a full task object. For example, when I use Cancel All on `V1ciHNV6S7eTWB3K1SakZA`, it ends up as:
```
__typename: "Task"
​​
created: "2024-03-20T14:10:57.718Z"
​​
deadline: "2024-03-21T14:10:57.718Z"
​​
dependencies: []
​​
extra: {…}
​​
metadata: {…}
​​​
__typename: "TaskMetadata"
​​​
description: "Initiate part or all of the training pipeline\n\nAction triggered by clientID `mozilla-auth0/ad|Mozilla-LDAP|bhearsum`\n"
​​​
name: "Action: Train"
​​​
owner: "action@noreply.mozilla.org"
​​​
source: "https://github.com/mozilla/firefox-translations-training/raw/dc91cb33531d7d20bcb5dcd4605381f8f2409999/.taskcluster.yml"
​​​
<prototype>: {…}
​​
payload: {…}
​​
priority: "LOWEST"
​​
requires: "ALL_COMPLETED"
​​
retries: 5
​​
routes: (3) […]
​​
schedulerId: "translations-level-1"
​​
scopes: (1) […]
​​​
0: "assume:repo:github.com/mozilla/firefox-translations-training:action:train"
​​​
length: 1
​​​
<prototype>: []
​​
tags: {…}
​​
taskGroupId: "WRffcusETWGVH66P9NI5yQ"
​​
taskId: "V1ciHNV6S7eTWB3K1SakZA"
​​
taskQueueId: "translations-1/decision-gcp"
​​
<prototype>: {…}
```

(Sorry for the bad paste...the Debugger doesn't let me copy the object nicely...)

bhearsum:
> Good catch!
> 
> `task.decisionTask` is graphql thing only used in UI, but definitely possible it can be missing if task was fetched or created not through those graphql calls
> 
> Thanks

I think I'm not clear on the graphql involvement here...in my testing I was loading URLs like http://0.0.0.0:5080/tasks/groups/V1ciHNV6S7eTWB3K1SakZA, then clicking Cancel All. Presumably the tasks and task group info was fetched through graphql?

ui/src/views/Tasks/submitTaskAction.js
gabrielBusta: Oh duh, yeah that makes sense

ui/src/views/Tasks/submitTaskAction.js
```diff
@@ -13,7 +13,18 @@ import ajv from '../../utils/ajv';
 
 export default async ({ task, form, action, apolloClient, taskActions }) => {
   const actions = removeKeys(cloneDeep(taskActions), ['__typename']);
-  const taskGroup = task.taskId === task.taskGroupId ? task : task.decisionTask;
+  // We need to source scopes from somewhere. Typically these come from the task we're given,
```
gabrielBusta, line 16: You might want to make your comment < 80 lines wide to appease the linter :)

gabrielBusta: Looks like it is making the build step fail

bhearsum: I thought `lint:fix` would that...but I guess not!

lotas (approved):
awesome! thanks

lotas (approved):

PR #6919, opened 2024-03-20
Client-node rewritten to ESModules

lotas:
Having CJS only was blocking some of the dependency upgrades which switched to ESM only distributions

`stream-buffers` was removed since it was only used in tests and could be rewritten using Readable/Writable streams from stdlib

### DIFF
--- a/clients/client/src/upload.js
+++ b/clients/client/src/upload.js
@@ -1,7 +1,7 @@
-const got = require('got');
-const { slugid } = require('./utils');
-const retry = require('./retry');
-const { HashStream } = require('./hashstream');
+import got from 'got';
+import { slugid } from './utils.js';
+import retry from './retry.js';
+import { HashStream } from './hashstream.js';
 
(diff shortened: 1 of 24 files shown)


clients/client/src/upload.js
```diff
@@ -62,7 +62,7 @@ const upload = async ({
   const retryCfg = {
     retries: retries === undefined ? 5 : retries,
     delayFactor: delayFactor === undefined ? 100 : delayFactor,
-    randomizationFactor: randomizationFactor === undefined ? randomizationFactor : 0.25,
+    randomizationFactor: randomizationFactor === undefined ? 0.25 : randomizationFactor,
```
lotas, line 65: for some reason it didn't cause any problem before node20 upgrade

matt-boris: 😬

matt-boris: Nice catch!

matt-boris (approved):
Great thanks Yarik 🎉

PR #6924, opened 2024-03-21
Queue batched tasks

Alphare:
This is done according to RFC 189¹. In few words, the motivation is to drastically reduce the number of queue API calls done by decision tasks, since they sometimes need the status for hundreds of tasks.

[1] https://github.com/taskcluster/taskcluster-rfcs/pull/189


Alphare:
@lotas when implementing the batched versions of the database calls, I'm bypassing the queue's LRUCache. Is that fine until we prove otherwise (since I'm assuming this is "just" a cache to speedup single lookups and not something that is actually load-baring)?

services/queue/src/api.js
```diff
@@ -192,6 +192,101 @@ builder.declare({
...
+    }
+    // Handle cases where the task doesn't exist
+    if (!task) {
+      return res.reportError('ResourceNotFound', [
+        '`{{taskId}}` does not correspond to a task that exists.',
+        'Are you sure this task has already been submitted?',
+      ].join('\n'), { taskId });
+    }
+    // Create task definition
+    tasks.push({ taskId, task: await task.definition() });
```
lotas, line 244: btw I think this function is not async anymore https://github.com/taskcluster/taskcluster/blob/main/services/queue/src/data.js#L115 
but I see it is used with await in some other places. We can ignore it here and refactor separatly

lotas:
> @lotas when implementing the batched versions of the database calls, I'm bypassing the queue's LRUCache. Is that fine until we prove otherwise (since I'm assuming this is "just" a cache to speedup single lookups and not something that is actually load-baring)?

Yeah, I think the main benefit there is to fetch all those tasks at once and bypass the cache. makes sense.

I'm not sure though, if whole request should fail if some of the tasks are missing? (maybe already expired, maybe just incorrect id and there is no way to tell why it is missing)
I think it should only return what was found and probably just ignore the missing ones?

Alphare:
> Yeah, I think the main benefit there is to fetch all those tasks at once and bypass the cache. makes sense.

Good!

> I'm not sure though, if whole request should fail if some of the tasks are missing? (maybe already expired, maybe just incorrect id and there is no way to tell why it is missing) I think it should only return what was found and probably just ignore the missing ones?

My latest implementation documents that the responsibility to check for missing tasks is the callers', and I think that's the right decision, since is the fastest way of answering while leaving the possibility open to still check.

Alphare:
This depends on #6915 because of the orders of db migrations.

lotas:
https://github.com/taskcluster/taskcluster/pull/6915 was merged, feel free to rebase @Alphare

Alphare:
Rebased, hopefully this also looks good :)

changelog/BmjAEJCdRZWKU0hhq9sBcQ.md
```diff
@@ -0,0 +1,4 @@
+audience: users
+level: minor
+---
+The queue now exposes `tasks` and `tasks/status` to batch task requests
```
lotas, line 4:
```suggestion
The queue now exposes `tasks` and `statuses` to batch task requests
```

Probably better to reflect real names in the API that generated code produced

services/queue/schemas/v1/tasks-response.yml
```diff
@@ -0,0 +1,32 @@
+$schema: "/schemas/common/metaschema.json#"
+title:              "Task Definition Response"
+description: |
+  Definitions of tasks that can be scheduled
```
lotas, line 4:
```suggestion
  Definitions of tasks
```

Same here, endpoint will return any tasks

services/queue/schemas/v1/tasks-statuses-response.yml
```diff
@@ -0,0 +1,32 @@
+$schema: "/schemas/common/metaschema.json#"
+title:              "Tasks Statuses Response"
+description: |
+  Status of tasks that can be scheduled
```
lotas, line 4:
```suggestion
  Status of multiple tasks
```

services/queue/src/api.js
```diff
@@ -192,6 +192,86 @@ builder.declare({
...
+  input: 'tasks-request.yml',
+  stability: APIBuilder.stability.experimental,
+  category: 'Tasks',
+  idempotent: true,
+  output: 'tasks-response.yml',
+  title: 'Get multiple task definitions',
+  description: [
+    'This end-point will return the task definition for each input task id.',
+    'Notice that the task definitions may have been modified by queue.',
+    'If an optional property is not specified the queue may provide',
```
lotas, line 215:
```suggestion
```

this was probably lost long time ago, as I don't see what optional property is being passed :) let's drop it

services/queue/src/api.js
```diff
@@ -192,6 +192,86 @@ builder.declare({
...
+    'a default value.',
```
lotas, line 216:
```suggestion
```

lotas:
> Rebased, hopefully this also looks good :)

Awesome! Thanks a lot @Alphare 

Looks good, just minor suggestions for the documentation part (If accepted, please rerun `yarn generate`)

Alphare:
Everything should be good now, thanks for catching the optimistic copy-paste ;)

lotas (approved):
Great!

PR #6926, opened 2024-03-22
Replaces single "googleapis" with smaller packages

### LINKED ISSUE #6761: Chore: Replace googleapis package with specific ones
While debugging connections in profiler I noticed that node spends a huge amount of time loading `googleapis` package. 

That package (as of 73.0.0) contains ~218 service that take about 84Mb. And I believe they are all being parsed and loaded on startup. 

Also [package docs](https://github.com/googleapis/google-api-nodejs-client) suggest to load packages individually:

> If you need to reduce startup times, you can alternatively install a submodule as its own dependency. We make an effort to publish submodules that are not in this [list](https://github.com/googleapis/google-cloud-node#google-cloud-nodejs-client-libraries). 

image

(measured with `npx clinic flame -- node src/main.js server`)

lotas:
This will speed up startup times, as whole `googleapis` package is `>100Mb` of js/json which needs a lot of time to parse at startup. 

Replacing single mono-package with smaller ones

Github Bug/Issue: Fixes #6761


lotas:
I'll have to abandon it for now. Auth service was relatively easy to upgrade, but worker-manager requires much bigger refactoring, since separate libraries do not provide same functionality equally good around oauth and iam stuff.

matt-boris (approved):
Nice! Thank you 👍🏻

PR #6932, opened 2024-03-26
Remove maxRunTime limitations for d2g payloads

bhearsum:
We added support for configurable maxRunTime in generic worker in https://github.com/taskcluster/taskcluster/pull/6465. We're now converting our existing docker-worker tasks over to d2g/generic-worker, some of which require a higher maxRunTime than generic-worker's docker-worker schema allows for. I believe this patch will fix that.

I did notice that there's also the same limit in https://github.com/taskcluster/taskcluster/blob/main/workers/docker-worker/schemas/v1/payload.yml, but seeing as docker-worker is deprecated, I don't think it's worthwhile to make any changes there.

(And for completeness: the docker worker maxRunTime was _reduced_ in https://github.com/taskcluster/taskcluster/pull/6019/files, but the version we're running allows for the old `604800` limit, which is how we have tasks configured that high.)

### DIFF
--- a/changelog/C7wI7jeHQHe41TMMOOrVYA.md
+++ b/changelog/C7wI7jeHQHe41TMMOOrVYA.md
@@ -1,0 +1,4 @@
+audience: users
+level: minor
+---
+Remove maxRunTime limitations for docker payloads in generic worker (d2g)
--- a/generated/references.json
+++ b/generated/references.json
@@ -9157,7 +9157,6 @@
             },
             "maxRunTime": {
               "description": "Maximum time the task container can run in seconds.",
-              "maximum": 86400,
               "minimum": 1,
               "multipleOf": 1,
               "title": "Maximum run time in seconds",
@@ -9944,7 +9943,6 @@
         },
         "maxRunTime": {
           "description": "Maximum time the task container can run in seconds.",
-          "maximum": 86400,
           "minimum": 1,
           "multipleOf": 1,
           "title": "Maximum run time in seconds",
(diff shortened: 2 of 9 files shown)


matt-boris:
Hey Ben, if you could share a task with an example of this issue happening, it would help me fully understand what you're seeing. Because I'm currently not seeing how your previous docker worker payload would've been valid if its `maxRunTime` was something over `86400`. My _guess_ is you're running into generic worker's _default_ `maxTaskRunTime` which is `86400`. This is a worker pool config setting. We recently fixed some `bugbug` issues they were running into which I'm _guessing_ is going to be the same as what you're seeing. See https://github.com/taskcluster/community-tc-config/pull/719.

bhearsum:
An example is this task: https://firefox-ci-tc.services.mozilla.com/tasks/OH2dEqpqTKqvG8dJ9UzRAA

Which is using [this worker pool](https://hg.mozilla.org/ci/ci-configuration/file/tip/worker-pools.yml#l591), configured with a `maxRunTime` of `2592000`

An equivalent task on a docker worker is https://firefox-ci-tc.services.mozilla.com/tasks/V52jwhZ3Qnmv67udzYKvnQ, which shows it's using `38.0.5`. I believe this predates #6019 where the docker worker `maxRunTime` was reduced.

matt-boris:
Ah ok perfect. Thanks for these links. You just need to run `yarn generate` and this will fix the CI issues. Feel free to remove the `maxRunTime` in the docker worker payload schema as well as `d2g` currently generates types off of it https://github.com/taskcluster/taskcluster/blob/main/tools/d2g/d2g.go#L1.

matt-boris (approved):
Thanks Ben 👍🏻

PR #6933, opened 2024-03-27
build(deps): upgrade to node v20.12.0

matt-boris:
>Upgrades to Node.js v20.12.0


lotas (approved):
that was fast :) thanks

PR #6955, opened 2024-04-04
build(deps): upgrade to go1.22.2 and node 20.12.1 (SECURITY)

matt-boris:
>Upgrades to go1.22.2 and Node.js 20.12.1 which are both security releases.


lotas (approved):
Thanks!

Issue #6958, opened 2024-04-04
Worker Manager (GCP) should only apply `disks[].initializeParams.labels` to persistent disks


matt-boris:
We'll need to rework this to only apply to persistent disks, we're seeing errors on FxCI:
```json
{
  "config": {
    "zone": "us-central1-a",
    "disks": [
      {
        "boot": true,
        "type": "PERSISTENT",
        "autoDelete": true,
        "initializeParams": {
          "labels": {
            "owner": "release-tc-workers-mozilla-com",
            "created-by": "taskcluster-wm-fxci-level1-gcp",
            "managed-by": "taskcluster",
            "worker-pool-id": "gecko-1-b-linux-gcp"
          },
          "diskSizeGb": 20,
          "sourceImage": "projects/taskcluster-imaging/global/images/docker-firefoxci-gcp-l1-googlecompute-2023-12-04t22-11-57z"
        }
      },
      {
        "type": "SCRATCH",
        "interface": "NVME",
        "autoDelete": true,
        "initializeParams": {
          "labels": {
            "owner": "release-tc-workers-mozilla-com",
            "created-by": "taskcluster-wm-fxci-level1-gcp",
            "managed-by": "taskcluster",
            "worker-pool-id": "gecko-1-b-linux-gcp"
          },
          "diskType": "zones/us-central1-a/diskTypes/local-ssd"
        }
      },
      {
        "type": "SCRATCH",
        "interface": "NVME",
        "autoDelete": true,
        "initializeParams": {
          "labels": {
            "owner": "release-tc-workers-mozilla-com",
            "created-by": "taskcluster-wm-fxci-level1-gcp",
            "managed-by": "taskcluster",
            "worker-pool-id": "gecko-1-b-linux-gcp"
          },
          "diskType": "zones/us-central1-a/diskTypes/local-ssd"
        }
      },
      {
        "type": "SCRATCH",
        "interface": "NVME",
        "autoDelete": true,
        "initializeParams": {
          "labels": {
            "owner": "release-tc-workers-mozilla-com",
            "created-by": "taskcluster-wm-fxci-level1-gcp",
            "managed-by": "taskcluster",
            "worker-pool-id": "gecko-1-b-linux-gcp"
          },
          "diskType": "zones/us-central1-a/diskTypes/local-ssd"
        }
      },
      {
        "type": "SCRATCH",
        "interface": "NVME",
        "autoDelete": true,
        "initializeParams": {
          "labels": {
[...]

PR #6959, opened 2024-04-04
fix(w-m): only apply `initializeParams.labels` to `PERSISTENT` disks on GCP

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6958.

>Worker Manager now only applies GCP disk labels to `PERSISTENT` disk types.
>
>This fixes an issue in v64.2.2 where `initializeParams.labels` was being set on all disk types and caused GCP to error on local SSDs (`SCRATCH` type disks).

### DIFF
--- a/changelog/issue-6958.md
+++ b/changelog/issue-6958.md
@@ -1,0 +1,7 @@
+audience: general
+level: patch
+reference: issue 6958
+---
+Worker Manager now only applies GCP disk labels to `PERSISTENT` disk types.
+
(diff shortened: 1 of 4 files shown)


lotas (approved):
Looks good!

my only concern is if we want to validate launch config schema. If someone provides disks with initializeParams.labels, then it's gonna still fail, but at least this is on the config size.. So let's keep it as is, thanks!

PR #6960, opened 2024-04-05
feat(indexes): Optimize db query to search by multiple indexes

### LINKED ISSUE #6957: `findTasksAtIndex` is really slow, even for queries of a single index
**Describe the bug**

On the production Firefox TC, Querying `findTasksAtIndex` takes 20+ seconds to answer if it does at all before getting timed out, no matter how small the list of indexes provided (aside from the empty list).

**To Reproduce**

```
curl -X POST https://firefox-ci-tc.services.mozilla.com/api/index/v1/tasks/indexes -H 'Content-Type: application/json' -d '{"indexes": ["doesnotexist"]}'
```

**Expected behavior**

I expect this query to take (well) under a second, removing any network latency. I measured the network latency for other calls to be <100ms at the time of testing.

**Taskcluster version**

v64.1.0+

**Screenshots**

The current query and its analysis:

image

A mockup of a better solution:

image

**Additional context**

We have figured out that the culprit is the SQL query doing a sequential read of the database and concatenating the namespace and name of each task before being able to answer, which takes a very long time.

Possible solutions include adding an index for this concatenation or - likely a much better solution - splitting the namespace from the name in the subquery to reduce this all to a hash join.

lotas:
Updated query uses index scan which is faster

`explain analyze` of the new query:

```
Limit  (cost=0.69..872.25 rows=1 width=145) (actual time=0.099..0.874 rows=18 loops=1)
  ->  Nested Loop  (cost=0.69..872.25 rows=1 width=145) (actual time=0.098..0.871 rows=18 loops=1)
        ->  Function Scan on jsonb_array_elements index_pairs  (cost=0.00..1.00 rows=100 width=32) (actual time=0.011..0.014 rows=18 loops=1)
        ->  Index Scan using indexed_tasks_pkey on indexed_tasks  (cost=0.69..8.71 rows=1 width=150) (actual time=0.043..0.043 rows=1 loops=18)
              Index Cond: ((namespace = (index_pairs.value ->> 0)) AND (name = (index_pairs.value ->> 1)))
```

Github Bug/Issue: Fixes #6957


matt-boris (approved):
👍🏻 thanks Yarik

matt-boris (approved):

Issue #6900, opened 2024-03-14
Add support for Azure Spot Instance Eviction Status


jwmoss:
**Describe the bug**
When Release SRE is diagnosing why windows cloud workers exit with exit code `72`, it's confusing whether or not this is due to generic worker exit code indicating `The worker is running on spot infrastructure in AWS EC2 and has been served a spot termination notice, and therefore has shut down` or another reason. In the taskcluster code, it seems that `WORKER_SHUTDOWN` could be caused by two reasons: 

1. An error occured during/afer a [task](https://github.com/taskcluster/taskcluster/blob/main/workers/generic-worker/main.go#L460) (although I could be wrong here, I'm not a golang expert)
2. [Termination requested](https://github.com/taskcluster/taskcluster/blob/main/workers/generic-worker/main.go#L521)

Upon looking at the code for termination requests, it appears [we're checking for the existence of an array of scheduled events and terminating the worker](https://github.com/taskcluster/taskcluster/blob/main/tools/worker-runner/provider/azure/azure.go#L111-L144). 

It would be useful to understand that if `scheduledevents` is sent to the metadata endpoint, if we could determine the `eventType` or glean more information about the cause of the scheduled event. Additionally, having a separate exit code would help track spot instance evictions (and other issues) in the event a specific region/sku is impacted more. More [here](https://learn.microsoft.com/en-us/azure/virtual-machines/windows/scheduled-events#event-properties).

[Link](https://learn.microsoft.com/en-us/azure/architecture/guide/spot/spot-eviction#continuously-monitor-for-eviction) for monitoring for eviction status.

**To Reproduce**

1. Visit papertrail and search for `"exit status 72"`.

**Expected behavior**
What I would expect to see is an exit code per `eventType`, or at least adding support for identifying when the eventType is `Preempt`.

**Taskcluster version**
62.0.0

**Screenshots**

![image](https://github.com/taskcluster/taskcluster/assets/2729151/d88fa7e2-4ebf-4b3d-9875-a07d98bdce24)
![image](https://github.com/taskcluster/taskcluster/assets/2729151/a657806a-a5a1-4b97-8b5b-97a24d3b3565)

PR #6961, opened 2024-04-05
fix(worker-runner azure): don't gracefully terminate on `Freeze` event

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6900.

>Worker Runner on Azure no longer sends a `graceful-termination` message if the scheduled event type is `Freeze`. It will continue to send the message for all other event types: `Reboot`, `Redeploy`, `Preempt`, and `Terminate`.


lotas (approved):
LGTM

PR #6977, opened 2024-04-18
fix(generic-worker): upload artifacts in parallel to speed up graceful shutdowns

### LINKED ISSUE #6972: preempted instance didn't upload all artifacts
After #6802 was fixed, generic worker appears to handle spot terminations a bit better. However, in my [first real world spot termination](https://firefox-ci-tc.services.mozilla.com/tasks/YVMH_VtgTRWvSArOOXFSTA/runs/0) with this patch, only some artifacts were uploaded. Specifically, `public/logs/live.log` and `public/build/model.npz.best-bleu-detok.npz` are still unavailable. The former appears to be trying to use the websocktunnel to grab the log, and the latter simply says the key does not exist. (It's unclear to me if these have the root cause or not.)

I looked through GCP logs and found the following:
* At 13:15:46 worker runner reports `CP Metadata Service says termination is imminent`
* A bunch of things happen (see attachment), and then at 13:16:01 it begins to upload artifacts.
* Only 4 artifacts are attempted to be uploaded (there are 6 on the task, including `live.log`), the last of which is logged at 13:16:09. No further uploads are attempted AFAICT.
* Finally, at 13:17:36 the instance is preempted.

[logs.txt](https://github.com/taskcluster/taskcluster/files/14980002/logs.txt)

matt-boris:
Should help with and/or fix #6972.

>Generic Worker now uploads task payload artifacts in parallel to decrease graceful termination time in the event of a spot termination.
>
>The `insecure` engine no longer performs a file copy command as the task user before the artifact upload process happens to help speed up the process.
>
>Generic Worker (posix only) now tries to put an exclusive file lock on artifacts before upload to prevent the file from being written to by any other process. This is done in lieu of copying the file to a temporary location which was achieving the same thing. If putting the lock on the file fails, Generic Worker will fallback to copying the file.

### DIFF
--- a/changelog/issue-6972.md
+++ b/changelog/issue-6972.md
@@ -1,0 +1,9 @@
+audience: worker-deployers
+level: patch
+reference: issue 6972
+---
+Generic Worker now uploads task payload artifacts in parallel to decrease graceful termination time in the event of a spot termination.
+
+The `insecure` engine no longer performs a file copy command as the task user before the artifact upload process happens to help speed up the process.
+
+Generic Worker (posix only) now tries to put an exclusive file lock on artifacts before upload to prevent the file from being written to by any other process. This is done in lieu of copying the file to a temporary location which was achieving the same thing. If putting the lock on the file fails, Generic Worker will fallback to copying the file.
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/elastic/go-sysinfo v1.13.1
 	github.com/fatih/camelcase v1.0.0
 	github.com/getsentry/raven-go v0.2.0
+	github.com/gofrs/flock v0.8.1
 	github.com/golang-jwt/jwt/v4 v4.5.0
 	github.com/gorilla/mux v1.8.1
 	github.com/gorilla/websocket v1.5.1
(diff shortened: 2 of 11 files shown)


changelog/issue-6972.md
```diff
@@ -0,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 6972
+---
+Generic Worker now uploads task payload artifacts in parallel to decrease graceful termination time in the event of a spot termination.
```
bhearsum, line 5: Is it worth mentioning the change to the insecure engine as well?

bhearsum (commented):
I'll note that this _does_ get rid one of the copies we make in the insecure engine, but we still make a full copy of each artifact [over here](https://github.com/taskcluster/taskcluster/blob/5f78921b3ae8d963e7e2da2b0705a03f4275497d/workers/generic-worker/artifacts/s3.go#L64), which slows down termination.

If I remember correctly, you told me that this copy was to avoid the file being modified while being uploaded. If that's the case, can we move it instead? (This could obviously be handled in a separate PR if it y'all decide to do it...)

changelog/issue-6972.md
matt-boris: Yes I'll add this, thanks 👍🏻

lotas (approved):
Awesome!

PR #6990, opened 2024-04-29
chore: use full go version in go.mod

matt-boris:
As suggested in https://github.com/github/codeql/issues/15647 to fix the below issue:
Screenshot 2024-04-29 at 12 41 11 PM


lotas (approved):
👍

PR #7009, opened 2024-05-14
Issue 6979 - Explicitly set XDG_RUNTIME_DIR in Linux/Multiuser

### LINKED ISSUE #6979: g-w linux: should perform a user login (XDG_RUNTIME_DIR not set)
**Describe the bug**

Debugging Wayland issues, we notice that the XDG_RUNTIME_DIR env var (https://wiki.alpinelinux.org/wiki/Wayland) is not set. It's critical for Wayland functioning. It's normally set by `pam_systemd` when a user login is performed. 

It looks like in https://github.com/taskcluster/taskcluster/blob/5f78921b3ae8d963e7e2da2b0705a03f4275497d/workers/generic-worker/multiuser_posix.go#L136-L139 we execute via os/exec's Command (https://pkg.go.dev/os/exec#Command) which seems to not perform a user login.

**To Reproduce**

Run a task on a Wayland worker in community of firefoxci (gecko-t/t-linux-2204-wayland-exp-a2), the XDG_RUNTIME_DIR will be unset. This causes issues. 

**Expected behavior**
XDG_RUNTIME_DIR is set to a valid path (e.g. from a working laptop running Wayland 22.04: `XDG_RUNTIME_DIR=/run/user/1001`).

**Taskcluster version**
g-w multi: [taskcluster 2024-04-19T05:32:12.111Z]       "version": "64.2.0"

**Screenshots**

**Additional context**

petemoore:
Github Bug/Issue: Fixes #6979

Spent the entire day looking at this to see if there is a comfortable way to generate the env vars that would be set in the gnome desktop environment and pass them to Generic Worker, but after a full day of trying (and with the help of @Eijebong) decided this was the most sensible solution. We had some successes with systemd-run but it all got very complicated, and only set 3 or 4 env vars compared to what would be seen in a full graphical session, so this seemed like the most pragmatic solution.

It isn't ideal having Generic Worker care about desktop settings, but it also isn't currently trivial for a task to launch a desktop session of its own, so this seems like a reasonable compromise.

### DIFF
--- a/changelog/EdA3XVyRQlGVGNjJlRW6yw.md
+++ b/changelog/EdA3XVyRQlGVGNjJlRW6yw.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: minor
+reference: issue 6979
+---
+Generic Worker multiuser engine on Linux now sets environment variable`XDG_RUNTIME_DIR` to `/run/user/<UID>` in task command processes (unless Generic Worker config setting `runTasksAsCurrentUser` is set to `true`).
--- a/workers/generic-worker/multiuser_posix.go
+++ b/workers/generic-worker/multiuser_posix.go
@@ -151,6 +151,9 @@ func (task *TaskRun) EnvVars() []string {
 	}
 	if runtime.GOOS == "linux" {
 		taskEnv["DISPLAY"] = ":0"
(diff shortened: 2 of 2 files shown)


matt-boris (dismissed):
👍🏻

matt-boris (approved):

PR #7015, opened 2024-05-15
feat(ui): Worker Manager shows worker page properly when queue data is missing

lotas:
When only worker-manager data is available, page no longer shows error and gives 404, but displays available info

image

PR #7016, opened 2024-05-15
feat(ui): Display both queue and w-m info when available

lotas:
Small addition to #7015 - always displaying w-m info when available, so views are merged 

image

Workers view includes a button when given provisionerId/workerType has corresponding workerPool associated with it 
image


matt-boris (dismissed):
👍🏻

matt-boris (approved):

PR #7019, opened 2024-05-16
feat(ui): Using consistent UI navigation across worker pages

lotas:
This is a refactoring of navigation components.
Worker manager pages and worker pages use same component

image

image

image

image

image

image


matt-boris (approved):
🎉

PR #7006, opened 2024-05-08
build(deps): upgrade to node 20.13.0 and go1.22.3

matt-boris:
>Upgrades to node v20.13.0 and go1.22.3 (SECURITY release).

PR #7021, opened 2024-05-16
build(deps): upgrade to go1.22.3

matt-boris:
Upgrades to go1.22.3 (SECURITY release). Was supposed to be handled in PR #7006, but was accidentally left out.


petemoore (approved):
Thanks Boris!

lotas (approved):
Node upgrade is missing :-p 

thanks!

PR #7030, opened 2024-05-21
Issue 7029: Generic Worker now resolves tasks as exception/malformed-payload if max run time occurs after deadline

### LINKED ISSUE #7029: Consider not resolving tasks as deadline-expired if they are already running (claimed)
This technically isn't a bug, it is by design, but it feels suboptimal in its current form.

Typically task payload will specify the max run time of a task. This is how long a task is allowed to run for, once it has been claimed. Separately, the task has a deadline, which is the point in time (not a duration) at which if the task has not run, you don't care about the result anymore. After this is the timestamp for the expiry, at which point, the record of the task existing is purged.

This commonly causes confusion, leading to task deadlines being set to creation time + max run time. For example: https://community-tc.services.mozilla.com/tasks/Cg-nlOz1Q1C93WChn4DiZA/definition

```
...
created: '2024-05-21T08:09:07.521Z'
deadline: '2024-05-21T12:09:07.521Z'
expires: '2024-05-28T08:09:07.521Z'
...
payload:
  ...
  maxRunTime: 14400
  ...
```

The max run time of the task is 14400s (= 4 hours) and the deadline was set to task created timestamp + 4 hours.

This can lead to tasks being resolved as deadline-exceeded by the queue, and having no artifacts uploaded by the worker, before the task has had the chance to run to completion.

Although intended, this is often undesirable, and can lead to wasted resources, since the task typically will have no artifacts uploaded, since this typically occurs at the end of the task. Or only the livelog, pointing to a temporary resource which disappears.

I propose we adapt the behaviour of the Queue task resolver, not to resolve tasks as `exception/deadline-exceeded` if there is a currently running task. If a new task run is scheduled, then it may later be resolved as `exception/deadline-exceeded`.

Downside of this approach: it may propagate (or fail to curb) misuse/misunderstanding of the deadline field.

petemoore:
Fixes #7029

### DIFF
--- a/workers/generic-worker/payload_test.go
+++ b/workers/generic-worker/payload_test.go
@@ -146,6 +146,21 @@ func NowMillis(t *testing.T) (now time.Time) {
 	return
 }
 
+// If a task deadline is sooner than maxRunTime
+func TestTaskDeadlineTooClose(t *testing.T) {
+	now := NowMillis(t)
+	task := taskWithPayload(`{
+  "env": {
+    "XPI_NAME": "dist/example_add-on-0.0.1.zip"
+  },
+  "maxRunTime": 3,
+  "command": [` + rawHelloGoodbye() + `]
+}`)
+	task.Definition.Deadline = tcclient.Time(now.Add(time.Second))
+	task.Definition.Expires = tcclient.Time(now.Add(time.Minute * 20))
+	ensureMalformedPayload(t, task)
+}
+
 // If an artifact expires before task deadline we should get a Malformed Payload
 func TestArtifactExpiresBeforeDeadline(t *testing.T) {
 	now := NowMillis(t)
--- a/changelog/issue-7029.md
+++ b/changelog/issue-7029.md
@@ -1,0 +1,5 @@
+audience: users
+level: major
+reference: issue 7029
+---
+Generic Worker now resolves tasks as exception/malformed-payload if the deadline occurs in less time than maxRunTime.
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -651,6 +651,9 @@ func (task *TaskRun) validatePayload() *CommandExecutionError {
 	if task.Payload.MaxRunTime > int64(config.MaxTaskRunTime) {
 		return MalformedPayloadError(fmt.Errorf("task's maxRunTime of %d exceeded allowed maximum of %d", task.Payload.MaxRunTime, config.MaxTaskRunTime))
 	}
+	if time.Time(task.Definition.Deadline).Before(time.Now().Round(0).Add(time.Second * time.Duration(task.Payload.MaxRunTime))) {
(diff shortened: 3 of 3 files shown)


matt-boris (approved):
Nice! I like this approach to solving that issue.

petemoore:
On reflection, I think I might stick with the original suggestion, of changing Queue behaviour rather than worker behaviour. Instead, I think it would be better for the task deadline to be the deadline for workers to _claim_ the task (latest task run), rather than to resolve the task. There are a few reasons why I favour this approach, over resolving the task if maxRunTime is after the deadline:
1. This is independent of task payload, so is generic across worker implementations such as Generic Worker / Script Worker / Docker Worker.
2. Conceptually the deadlines for each phase are typically independent of each other. If a task max run time needs to increase by 10 mins, it isn't ideal to need to increase deadline by 10 mins too. Typically deadline is about how long you are prepared to wait for a worker to claim your task, but if your task now takes longer (or shorter) probably you want to keep the same period of time as a timeout for the task getting claimed.
3. The implementation I did above, which resolves task as exception if there is less than maxRunTime left until deadline may resolve tasks as exception that could have run and would have resolved in time, since maxRunTime is an upper bound, not an expected task completion time.
4. The above implementation resolved tasks as malformed-payload, which would be inappropriate if workers were backlogged or slow to start up. However, resolving as deadline-exceeded instead would also be confusing, since deadline would not have passed at the time the task is resolved, which would be pretty confusing. However, no other resolution states seem appropriate, and also not ideal to introduce new resolution statuses, which complicate any existing logic (which can be custom user code) that already considers resolution status of tasks. This is not a reasonable burden to put on devs that rely on the existing statuses.
5. By tweaking Queue semantics so that deadline is deadline for a task to be claimed, rather than a task to be resolved, the deadline-exceeded status is still representative of the outcome and accurate. Also it keeps the reasonably independent quantities "how much time is reasonable to wait for a task to be claimed" and "after being claimed, how much time is reasonable for task to be resolved" reasonably independent of each other.
6. No new task resolution statuses are required.

When making this change, it will be important to update docs at the same time, to make it clear that deadline-exceeded means task claim deadline, not task resolution deadline.

This should be a major version bump, since it is backwardly incompatible behaviour.

workers/generic-worker/payload_test.go
```diff
@@ -146,6 +146,21 @@ func NowMillis(t *testing.T) (now time.Time) {
...
+func TestTaskDeadlineTooClose(t *testing.T) {
+	now := NowMillis(t)
+	task := taskWithPayload(`{
+  "env": {
+    "XPI_NAME": "dist/example_add-on-0.0.1.zip"
+  },
+  "maxRunTime": 3,
+  "command": [` + rawHelloGoodbye() + `]
+}`)
+	task.Definition.Deadline = tcclient.Time(now.Add(time.Second))
```
petemoore, line 159: This could put deadline in the past by the time the task gets claimed, especially if not run against mocks but a real service, so I think better to set it for e.g. 10 mins in the future.

petemoore (commented):
I don't think we can go with this as is, I think we need some more discussion about it first.

PR #7031, opened 2024-05-22
feat(generic-worker): `insecure` engine accepts docker worker payloads

matt-boris:
>Allows Docker Worker payloads to be used on the `insecure` Generic Worker engine, translated by `d2g`.


lotas (approved):
cool! Thanks

PR #7034, opened 2024-05-31
pg_dump used inside docker if versions mismatch

### LINKED ISSUE #6079: pg_dump dev requirement not documented
To set up a taskcluster development environment, you currently need pg_dump installed on your host. Used here:

* https://github.com/taskcluster/taskcluster/blob/66941613e4242d69dd2aff3fb560359eb633d59c/infrastructure/tooling/src/generate/generators/db-schema-dump.js#L36

We currently don't document that a native postgres client is required in
* https://github.com/taskcluster/taskcluster/blob/fe1d0cbe16ee2f68e500335ae5a140324b5b6369/dev-docs/development-process.md

If this is the only place a native client is used, we might be better off just running some equivalent sql commands and using taskcluster-lib-postgres which already seems to be required in that source code, rather than executing a native binary. Or maybe it isn't trivial to do via sql (although i expect there probably will be some built in postgres function(s) to do it), and we are just better off documenting the requirement.

Not sure if this is the only place a native client is currently required.

### LINKED ISSUE #6567: `yarn generate` doesn't work
Even with a fully fresh postgres container + volume I get this error:
```
yarn run v1.22.19
$ export NODE_OPTIONS=--dns-result-order=ipv4first && node infrastructure/tooling/src/main.js generate
... 79 completed tasks
⣯ DB Schema Dump
  ✔ Reset Test Database
  ⣯ Upgrade Test Database
  ℹ ...updating users
⢿ Generate d2g
⢿ Generate Generic-Worker
⢿ Generate Taskcluster-Client-Go
⢿ Update worker-runner README file
89% finished
node:events:495
      throw er; // Unhandled 'error' event
      ^

Error: spawn pg_dump ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:284:19)
    at onErrorNT (node:internal/child_process:477:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess._handle.onexit (node:internal/child_process:290:12)
    at onErrorNT (node:internal/child_process:477:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawn pg_dump',
  path: 'pg_dump',
  spawnargs: [
    '--schema-only',
    '-h',
    'localhost',
    '-p',
    '5432',
    '-U',
    'postgres',
    '-d',
    'postgres'
  ]
}

Node.js v18.18.0
```

lotas:
This adds `pg_dump` in development process documentation and fixes some of the problems around `yarn generate`.

Generate command expects database to be running so that its schema could be checked with the help of `pg_dump`. 
Sometimes it might result in local `pg_dump` version being different from postgres's server version, in such case `pg_dump` throws an error. 
Sometimes you might not have `pg_dump` installed at all.

Plus local development environment is being built around `docker compose` which always run the required version of `postgres` db.

This fix is trivial and will first run native `pg_dump` and in case of any error will attempt to run it with `docker compose`. If both fail, the generation process will also fail.

Fixes #6567
Fixes #6079

### DIFF
--- a/changelog/issue-6567.md
+++ b/changelog/issue-6567.md
@@ -1,0 +1,6 @@
+audience: developers
+level: patch
+reference: issue 6567
+---
+
+`yarn generate` commands will attempt to run `pg_dump` inside the docker container if local binary is missing or its version is different from the server version.
--- a/dev-docs/development-process.md
+++ b/dev-docs/development-process.md
@@ -34,6 +34,9 @@ The easiest and best way to do this is to use docker, but if you prefer you can 
 *NOTE* the test suites repeatedly drop the `public` schema and re-create it, effectively deleting all data in the database.
 Do not run these tests against a database instance that contains any useful data!
 
+`pg_dump` is being used to detect schema changes. If it is not present on your system,
+`yarn generate` might attempt to run this inside the `postgres` docker container, which is one of the `docker-compose.yml` services.
+
 To start the server using Docker:
 
 ```shell
--- a/infrastructure/tooling/src/generate/generators/db-schema-dump.js
+++ b/infrastructure/tooling/src/generate/generators/db-schema-dump.js
@@ -31,23 +31,38 @@ export const tasks = [{
     // now dump the schema of the DB
     utils.step({ title: 'Dump Test Database' });
     const { host, port, user, password, database } = parseDbURL(process.env.TEST_DB_URL);
-    const pgdump = await execCommand({
-      dir: REPO_ROOT,
-      command: [
-        'pg_dump',
-        '--schema-only',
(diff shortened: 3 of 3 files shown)


petemoore (approved):
This looks good! Were you able to test it?

lotas:
> This looks good! Were you able to test it?

yes, locally when I ran into this issue after another `brew update` upgraded system-wide packages 😬  and versions didn't match. Plus tried out with missing binary, also works.
The only prerequisite is `postgres` container running with docker compose

PR #7048, opened 2024-06-03
feat(ui): Show errors only when no worker info exists

lotas:
Since we might have queue only view, worker-manager only view, both views or none, we only want to show errors, when there were zero records found for both queue and worker-manager


petemoore (approved):
lgtm!

PR #7051, opened 2024-06-04
Remove explit build constraints that are already implied by filename

petemoore:
Looks like some unnecessary build constraints slipped in. Go already applies a platform build constraint if the platform is specified in the filename.


lotas (dismissed):
nice!

PR #7053, opened 2024-06-05
Fix pg_dump usage in yarn generate

lotas:
Previous error handling wasn't sufficient as missing binary would be a fatal error and docker compose wouldn't even run.
New approach detects first if executable exists.
It also prefers to run pg_dump from the same docker container that db is running in. If there are any docker processes with postgres running, generator will execute pg_dump inside that container.

Also instead of only checking `docker compose`, it should also run if postgres is run as a  standalone container, as mentioned in docs.


petemoore (approved):
Awesome!

PR #7054, opened 2024-06-06
Azure CI worker pools no longer have -azure suffix

petemoore:
See https://github.com/taskcluster/community-tc-config/pull/740 for context.


lotas (approved):
lgtm

PR #7056, opened 2024-06-07
feat(k8s): Allow to skip certain resource kinds in helm

### LINKED ISSUE #7035: Add an Option to the Helm Chart to Disable Secret Creation
We are trying to deploy TC using ArgoCD (GitOps approach), and for fetching secrets, we are using the External Secret Operator (ESO). Currently, TC stores secrets as files, which does not work for our setup.

Could you consider adding an extra flag to the Helm chart to disable secret creation? This would allow us to use the secrets defined by ESO instead of the ones defined in the Helm charts.

This enhancement would simplify our deployment process and integrate seamlessly with our existing ESO-managed secrets.

lotas:
This can be useful for deployments that manage resources differently, i.e. having own ingress.yml configuration, or managing secrets/configs with external tools.

Passing `skipResourceTypes:[]` values would tell helm to omit those

Example usage:

```sh
helm template --values values.yaml --values ../../dev-config.yml --set "skipResourceTypes[0]"=secret --set "skipResourceTypes[1]"=ingress 
```

Or by adding those values to a separate yaml config:

```sh
skipResourceTypes:
  - configmap
  - secret
  - ingress
  - serviceaccount
```

Fixes #7035 

> **Note**: `configmap` resources are not yet available, but this option is added for future compatibility


infrastructure/k8s/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: taskcluster
 appVersion: '65.1.0'
 description:  Framework that supports Mozilla's continuous integration and release processes.
-version: 0.1.0
+version: 0.2.0
```
lotas, line 4: why not :)

petemoore (approved):
Thanks Yarik!

I'm not really sure what I'm looking at - is there someone in SRE team that could give a better review? Or maybe one of the anaconda guys?

lotas:
> Thanks Yarik!
> 
> I'm not really sure what I'm looking at - is there someone in SRE team that could give a better review? Or maybe one of the anaconda guys?

I will let our SRE team know for sure. 

Defaults are configured in such way that those templates would not produce any changes if they are to be deployed. 
We would also be able to test after merge to main - our dev deployment will be our canary

lotas:
ok, no policy got it. what about created vs edited
/taskcluster u_good?

PR #7069, opened 2024-06-19
fix(notify): Ensure pulse routes use custom topics from routing key

ahal:
The notify service supports sending notifications out through a pulse exchange. Tasks can make use of this via routes like:

    notify.pulse.some.custom.topic.on-resolved

However, the notify service currently does not use `some.custom.topic` in the routing key it emits. This results in *all* messages using the same routing key, which ends up being `primary._`.

This defeats the purpose of using the notify service's pulse exchange, as consumers need to manually sift through *all* messages sent out over this exchange.

### DIFF
--- a/services/notify/src/exchanges.js
+++ b/services/notify/src/exchanges.js
@@ -28,12 +28,12 @@ const buildCommonRoutingKey = function(options) {
       constant: 'primary',
       required: true,
     }, {
-      name: 'reserved',
-      summary: 'Space reserved for future routing-key entries, you ' +
-                        'should always match this entry with `#`. As ' +
-                        'automatically done by our tooling, if not specified.',
+      name: 'topic',
+      summary: 'Custom topic. This is the <topic> portion of the ' +
+                        '`notify.pulse.<topic>.on-<event>` routes.',
+      maxSize: 100,
       multipleWords: true,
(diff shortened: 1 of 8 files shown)


services/notify/src/exchanges.js
```diff
@@ -28,12 +28,12 @@ const buildCommonRoutingKey = function(options) {
       constant: 'primary',
       required: true,
     }, {
-      name: 'reserved',
-      summary: 'Space reserved for future routing-key entries, you ' +
-                        'should always match this entry with `#`. As ' +
-                        'automatically done by our tooling, if not specified.',
+      name: 'topic',
```
ahal, line 31: I'm not really sure if `topic` is the right word here. It might have a slightly different meaning in RabbitMQ parlance. But I couldn't think of anything better.

ahal:
I'm assuming the generic-worker failure is unrelated to this PR. Please let me know if I'm mistaken!

lotas (approved):
Cool, thanks Andrew

lotas:
> I'm assuming the generic-worker failure is unrelated to this PR. Please let me know if I'm mistaken!
> 
> Edit: Yes this is also failing on `main`

yes, please ignore the failing windows generic-worker build, this is unrelated

PR #7071, opened 2024-06-20
Set TASKCLUSTER_INSTANCE_TYPE env var in Generic Worker tasks

### LINKED ISSUE #7070: generic-worker should set TASKCLUSTER_INSTANCE_TYPE
Found by @glandium in [bug 1903428](https://bugzilla.mozilla.org/show_bug.cgi?id=1903428#c0)

> However, TASKCLUSTER_INSTANCE_TYPE is [only set by docker worker](https://github.com/search?q=repo%3Ataskcluster%2Ftaskcluster%20TASKCLUSTER_INSTANCE_TYPE&type=code), not generic worker.

@petemoore, what would be the complexity to add a line like: 

https://github.com/taskcluster/taskcluster/blob/35b4e2d01200b90ddaaa87669a6ea177577ce499/workers/docker-worker/src/task.js#L364

to generic-worker? 

Having this environment variable will help with cost saving. We will be able to pinpoint what instance types are taking longer to clone hg.mozilla.org, for example.

petemoore:
Fixes #7070


lotas (approved):

djmitche (commented):
This should probably get added to the docs, too, with some indication of what it means for each supported cloud

lotas (approved):

PR #7075, opened 2024-06-24
Add support for yarn run dev:start, yarn run dev:stop

### LINKED ISSUE #6668: `yarn run` doesn't support shell like `npm run`
At the moment `yarn run dev:start` will not work because yarn's shell interpreter [does not support control structures](https://github.com/yarnpkg/berry/tree/master/packages/yarnpkg-shell#non-goals)

`npm run dev:start` still works

This problem can be fixed in several ways:
- generate `docker-compose.override.yml` separately and always assume it exists
- move that logic into separate (shell) script and invoke it instead

thakkarujas5:
This introduces fix for supporting yarn run shell for "dev:start" and "dev:stop" scripts.

Fixes #6668


changelog/issue-6668.md
```diff
@@ -0,0 +1,5 @@
+level: patch
+audience: developers
+reference: issue 6668
+
```
lotas, line 4:
```suggestion
---
```

I think there's a separator missing, so one check fails

thakkarujas5: Resolved: added the separator.

lotas (approved):
Awesome! Thanks again for your contribution

PR #7079, opened 2024-06-25
Worker Runner: Use IMDSv2 in EC2 instead of IMDSv1

### LINKED ISSUE #7076: Worker Runner to support IMDSv2 in EC2
Currently Worker Runner uses IMDSv1 to query metadata, but it looks like support for v1 may be limited, and we have encountered issues with Ubuntu 24.04 where support was not included.

An attempt to provide IMDSv1 support was made in https://github.com/taskcluster/community-tc-config/pull/749 but it appears not to have been successful, and has resurfaced in https://github.com/taskcluster/community-tc-config/issues/766.

Therefore, rather than invest too much effort in supporting v1 further, it might be simplest, more secure, and more future proof, to update Worker Runner to use v2.

Note, we still depend on c5.metal instances running Linux payloads in AWS (e.g. for fuzzing) as we cannot migrate these to GCP or Azure.

petemoore:
Fixes #7076

### DIFF
--- a/changelog/issue-7076.md
+++ b/changelog/issue-7076.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: minor
+reference: issue 7076
+---
+Worker Runner now uses IMDSv2 instead of IMDSv1 in EC2. IMDSv1 is being phased out by Amazon.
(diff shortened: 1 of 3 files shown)


lotas (approved):
awesome!!!

lotas (approved):

petemoore:
Confirmed working in https://community-tc.services.mozilla.com/tasks/UrInAsn8S-mZYvLiSn7ufg

PR #7081, opened 2024-06-25
feat(github): Fixes 404 handling when task group doesn't exist

### LINKED ISSUE #7080: tc-github shouldn't report an error if sealTaskGroup fails with 404
**Describe the bug**

If a github PR fails to create any tasks at some point (missing scopes, buggy .taskcluster.yml, ...), all subsequent pushes will call sealTaskGroup on the never-created group, and report a spurious error to the PR.
See e.g. https://github.com/mozilla-releng/tooltool/pull/1186 and https://github.com/mozilla-releng/tooltool/commit/2a802ebad508f68c23810956fc11b2778dbd77e5#commitcomment-143393335.

**Expected behavior**

tc-github should probably ignore 404s from sealTaskGroup, or not add the group to its db until it actually exists.

**Taskcluster version**

65.1.0

lotas:
Previously incorrect property was checked for an error code, which resulted in same error being sent to github comments

Fixes #7080

### DIFF
--- a/changelog/issue-7080.md
+++ b/changelog/issue-7080.md
@@ -1,0 +1,7 @@
+audience: developers
+level: patch
+reference: issue 7080
+---
+
+Fixes github service issue during cancellation of the previous runs that were not created.
+Response code was not checked properly which resulted in sending same error for each new build.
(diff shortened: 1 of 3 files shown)


petemoore (approved):
Sweet!

PR #7084, opened 2024-06-26
API graceful query handling for pagination

### LINKED ISSUE #7083: Pagination with incorrect payload results in 500 Internal server error
For example:
```
curl -v https://community-tc.services.mozilla.com/api/purge-cache/v1/purge-cache/list\?continuationToken\=this-is-not-right-michael-this-is-not-right                                                                                         
HTTP/2 500

{
  "code": "InternalServerError",
  "message": "Internal Server Error, incidentId d8163ea0-0873-45e5-aeef-39959e275128.\n\n---\n\n* method:     allPurgeRequests\n* errorCode:  InternalServerError\n* statusCode: 500\n* time:       2024-06-26T10:30:57.666Z",
  "requestInfo": {
    "method": "allPurgeRequests",
    "params": {},
    "payload": {},
    "time": "2024-06-26T10:30:57.666Z"
  }
```

Query parameter validation should be `400 Bad request` error

lotas:
Used hashids library throws an error if decoded token contains invalid characters. API in such case would throw `500 InternalServerError` error. However, this should be a client error - `400 InvalidInput`.

Fixes #7083

### DIFF
--- a/changelog/issue-7083.md
+++ b/changelog/issue-7083.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 7083
+---
+
+Fixes query validation in pagination queries that were throwing `500 InternalServerError` instead of `400 InputError`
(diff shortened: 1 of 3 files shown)


petemoore (approved):
Looks great, thanks Yarik!

PR #7090, opened 2024-06-27
feat(queue): resolved tasks are removed from pending queue

### LINKED ISSUE #7089: worker manager thought tasks were pending despite them being cancelled
The worker pool showed:
image

Pending tasks were specifically two runs of the same task (which....seems impossible?):
image

Both runs were `CANCELED`:
image

This happened in a pool where workers were successfully spinning up, but never claiming tasks due to an issue with the image, if it matters.

lotas:
This issue existed for a long time and queue refactoring made it apparent that tasks that were considered pending, were in fact not. 
When task is resolved (cancelled), queue service only removed it form the claimed list, but record in pending queue was left as is. This was not an issue for queue service, because it was also checking real task status at claim time, but worker-manager only saw total numbers. And seeing non-zero pending count resulted in creation of new instances that would start and then do nothing because no tasks were actually pending.

Fixes #7089


petemoore (approved):
Nice job, Yarik!

PR #7092, opened 2024-06-27
D2G: use docker (instead of podman) for privileged tasks

### LINKED ISSUE #7082: kvm broken under d2g
See https://github.com/MozillaSecurity/fuzzing-tc-config/pull/171#issuecomment-2155776248

petemoore:
Fixes #7082

Tested in https://community-tc.services.mozilla.com/tasks/aEp39hd-QYKzs1mFFwvRKQ

### DIFF
--- a/changelog/issue-7082.md
+++ b/changelog/issue-7082.md
@@ -1,0 +1,17 @@
+audience: users
+level: major
+reference: issue 7082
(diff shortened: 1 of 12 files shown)


lotas (approved):
nice 👏

PR #7122, opened 2024-07-03
fix(d2g): disable memory, pid, or ulimits in d2g (fixes #7120)

### LINKED ISSUE #7120: generic-worker/d2g uses default resource limits for running containers
Over in https://github.com/mozilla/firefox-translations-training/issues/630#issuecomment-2197384089 we're seeing issues like being unable to fork or start new threads after switching workers from docker-worker to generic-worker. Neither the workers nor the tasks have changed. After digging, I found that podman has [various options for configuring resource limits](https://docs.podman.io/en/latest/markdown/podman-run.1.html for containers, some of which are set by default. The ones that stood out to me that have a default are:
```
--cpu-period=limit

Set the CPU period for the Completely Fair Scheduler (CFS), which is a duration in microseconds. Once the container’s CPU quota is used up, it will not be scheduled to run until the current period ends. Defaults to 100000 microseconds.

--cpu-shares, -c=shares

CPU shares (relative weight).

By default, all containers get the same proportion of CPU cycles. This proportion can be modified by changing the container’s CPU share weighting relative to the combined weight of all the running containers. Default weight is 1024.

--memory-swap=number[unit]

A limit value equal to memory plus swap. A unit can be b (bytes), k (kibibytes), m (mebibytes), or g (gibibytes).

Must be used with the -m (--memory) flag. The argument value must be larger than that of -m (--memory) By default, it is set to double the value of --memory.


--pids-limit=limit

Tune the container’s pids limit. Set to -1 to have unlimited pids for the container. The default is 2048 on systems that support “pids” cgroup controller.


--shm-size=number[unit]

Size of /dev/shm. A unit can be b (bytes), k (kibibytes), m (mebibytes), or g (gibibytes). If the unit is omitted, the system uses bytes. If the size is omitted, the default is 64m. When size is 0, there is no limit on the amount of memory used for IPC by the container. This option conflicts with --ipc=host.


--shm-size-systemd=number[unit]

Size of systemd-specific tmpfs mounts such as /run, /run/lock, /var/log/journal and /tmp. A unit can be b (bytes), k (kibibytes), m (mebibytes), or g (gibibytes). If the unit is omitted, the system uses bytes. If the size is omitted, the default is 64m. When size is 0, the usage is limited to 50% of the host’s available memory.


--ulimit=option

Ulimit options. Sets the ulimits values inside of the container.

--ulimit with a soft and hard limit in the format =[:]. For example:

$ podman run --ulimit nofile=1024:1024 --rm ubi9 ulimit -n 1024

Set -1 for the soft or hard limit to set the limit to the maximum limit of the current process. In rootful mode this is often unlimited.

Use host to copy the current configuration from the host.

Don’t use nproc with the ulimit flag as Linux uses nproc to set the maximum number of processes available to a user, not to a container.

Use the --pids-limit option to modify the cgroup control to limit the number of processes within a container.
```

It seems plausible that either the pid or ulimits could be causing the acute issues we're seeing, but IMO we should remove the limits anyways. I can see a case for reserving a very small amount of memory, cpu, and iops for generic-worker itself (for example, to allow it to shut itself down cleanly in the case of an OOM, which is currently can't do), but otherwise we should make all of the system resources available to container processes, just as they are to non-container processes.

bhearsum:
These seem to be the key ones in the cases that I've had issues. I also looked at:
- Removing CPU limits - but after a reread of the docs, it looks like that only applies when multiple containers are competing for resources, which is something that rarely happens in practice, and there's no way to "disable" them (you need to set limits and shares for each container, which d2g can't do intelligently)>
- Setting shm size. I didn't change this in the end because there's no way to make it unlimited, and unless we have a reason to believe it needs to be increased, it may as well stay as the default.

Github Bug/Issue: Fixes #7120

### DIFF
--- a/changelog/issue-7120.md
+++ b/changelog/issue-7120.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 7120
+---
+
+Removed memory, pid, and ulimits for d2g payloads.
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -306,6 +306,15 @@ func runCommand(containerName string, dwPayload *dockerworker.DockerWorkerPayloa
 	default:
 		command.WriteString(fmt.Sprintf("timeout %v %v run -t --name %v", dwPayload.MaxRunTime, tool, containerName))
 	}
+	// Do not limit resource usage by the containerName. See
+	// https://docs.podman.io/en/latest/markdown/podman-run.1.html
+	// and https://docs.docker.com/reference/cli/docker/container/run/
+	command.WriteString(" --memory-swap -1 --pids-limit -1")
+	// Only podman supports inheriting host ulimits. `docker` uses docker
+	// daemon settings by default.
+	if tool == "podman" {
+		command.WriteString(" --ulimit host")
+	}
 	// Sometimes --privileged is needed under podman, even where there are no
 	// capabilities defined. Therefore always add it under podman, regardless.
 	// See e.g.  https://github.com/taskcluster/taskcluster/issues/6888
--- a/tools/d2g/d2gtest/testdata/testcases/artifacts_test.yml
+++ b/tools/d2g/d2gtest/testdata/testcases/artifacts_test.yml
@@ -33,7 +33,8 @@ testSuite:
           - - bash
             - '-cx'
             - >-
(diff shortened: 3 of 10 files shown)


gabrielBusta:
The limit does not exist.

petemoore (approved):
Great work, thanks Ben!

PR #7127, opened 2024-07-08
Revert #6891 and #6929

### LINKED ISSUE #7126: d2g: universally passing --privileged (#6891) breaks some tasks
In #6891 `--privileged` was added globally to all generated d2g `podman run` commands. At the time, it was believed this was needed. However, this broke some tasks, for example see https://github.com/web-platform-tests/wpt/pull/46967.

Returning to the original issue that PR #6891 was intended to solve (issue #6890) the reported issue was [no longer reproducible](https://github.com/taskcluster/taskcluster/issues/6890#issuecomment-2210782247).

Since the original issue that required this change is no longer reproducible, and the change is also breaking other tasks, #6891 should be reverted. Furthermore, #6891 also broke other tasks which required a fix in #6929. Therefore #6929 should also be reverted.

petemoore:
Fixes #7126.

### DIFF
--- a/changelog/issue-7126.md
+++ b/changelog/issue-7126.md
@@ -1,0 +1,5 @@
+audience: users
+level: major
+reference: issue 7126
+---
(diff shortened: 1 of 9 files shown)


lotas (approved):
Thanks Pete

PR #7138, opened 2024-07-10
Code simplification: no need to call user.Lookup(username) twice

petemoore:
Spotted that we were calling `user.Lookup(username)` twice unnecessarily...


lotas (approved):
👍

PR #7140, opened 2024-07-10
Set TASK_GROUP_ID env var in Generic Worker tasks

### LINKED ISSUE #7139: set TASK_GROUP_ID env var in Generic Worker tasks
Tasks that create subtasks (typically decision tasks) typically want to give the subtask the same `taskGroupId` that they themselves have, in order that the subtask appears in the same task group.

There is a convention that task group id of a decision task is the same as its task id. However, this is only a convention, and isn't guaranteed. It also limits task groups to a single decision task.

In the case where a task whose `taskGroupId` did not match its `taskId` wanted to create a subtask, it would either need to look up its own task group id (via api calls, e.g. via taskcluster-proxy) in order to pass it to the subtask, or it may mistakenly use its own taskId as the taskGroupId of the subtask, which would end up placing subtasks in a totally different taskGroup to that of the task which created them.

Since workers are already passed the task group ID of the task that they claim, it would make sense to expose this information also to the task. We currently already pass TASK_ID and RUN_ID as environment variables, so it would be reasonable to also pass TASK_GROUP_ID to hold the `taskGroupId` value. This way, tasks which spawn subtasks (not always a canonical "decision-task") can use this to specify the task group id of the tasks they create.

This will in turn enable us to easily update https://github.com/taskcluster/taskgraph/blob/4896be17ec9ef349f61f8acb85699dc9a9c7e555/src/taskgraph/create.py#L45 which currently relies on the task group id and task id of the decision task matching, which leads to unexpected behaviour if the task is retriggered inside the existing task group (the decision task and its subtasks end up in different task groups).

petemoore:
Fixes #7139


lotas (approved):
awesome !

PR #7142, opened 2024-07-10
feat(w-m): avoid sending duplicate email errors

### LINKED ISSUE #7129: worker-manager vs quota errors
When an operation fails due to having run out of quota for a resource, w-m will typically retry it again and again and again, each time sending error email.  Is there a way to be smarter here and e.g. back off for a while?  Or not send email for each and every error?

lotas:
Fixes #7129

### DIFF
--- a/changelog/issue-7129.md
+++ b/changelog/issue-7129.md
@@ -1,0 +1,6 @@
+audience: worker-deployers
+level: patch
+reference: issue 7129
+---
+
+Worker-manager would avoid sending emails with duplicate error messages, as long as error message and information are the same.
--- a/services/worker-manager/src/providers/provider.js
+++ b/services/worker-manager/src/providers/provider.js
@@ -1,4 +1,6 @@
 import assert from 'assert';
+import crypto from 'crypto';
+import _ from 'lodash';
 import libUrls from 'taskcluster-lib-urls';
 import slugid from 'slugid';
 import yaml from 'js-yaml';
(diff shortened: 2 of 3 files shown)


petemoore (commented):
LGTM, although I'm no node expert.

Is there a risk though, that one message is sent, and then is never seen again? Perhaps there is merit in limiting how often the email is sent, rather than there being a problem which is reported once, but is ongoing, but never reported again?

What happens if a Worker Pool has an error, which is resolved, but e.g. 6 months later, the same error starts happening again. Will another email be sent, because the problem has started again, or will the report from 6 months ago be enough to prevent the recurring problem not to be reported again?

I wonder if we should just organise it so that it won't send the same email within a given time period, like a week. But you'll get an email at most once per week, if the issue is ongoing, and hasn't been resolved?

lotas:
> LGTM, although I'm no node expert.
> 
> Is there a risk though, that one message is sent, and then is never seen again? Perhaps there is merit in limiting how often the email is sent, rather than there being a problem which is reported once, but is ongoing, but never reported again?
> 
> What happens if a Worker Pool has an error, which is resolved, but e.g. 6 months later, the same error starts happening again. Will another email be sent, because the problem has started again, or will the report from 6 months ago be enough to prevent the recurring problem not to be reported again?
> 
> I wonder if we should just organise it so that it won't send the same email within a given time period, like a week. But you'll get an email at most once per week, if the issue is ongoing, and hasn't been resolved?

this is done in the the same was as in github/notify services - it is in-memory (last 1000) cache. So the risk that we'd suppress too much is not that high. Usually pods restart from time to time and also errors would be still shown in the UI or exposed in API.

We can also change this hashmap to have some TTL and let them expire after some timeout, instead of relying on pod restarts

petemoore (approved):
Sounds good!

PR #7154, opened 2024-07-14
Change `adduser/deluser` usage to `useradd/userdel`

Eijebong:
`adduser/deluser` are debian specific wrappers around `useradd/userdel` and friends. By changing to `useradd/userdel`, we allow workers to be deployed on non debian derivative distributions. Tested on archlinux.

Flag mapping for useradd:

`--disabled-password`: Disables password prompting and disables the password for the user. `useradd` does that by default.

`--gecos ""`: This sets the GECOS flag for the user to "". Maps to `chfn -f ${username}`

`--debug`: Makes adduser output what commands it runs. Irrelevant here.

`--home ${homedir}`: Creates the home directory at `${homedir}`. `-m` Creates the user home directory and `-d ${homedir}` defines it as being `${homedir}`.

Flag mapping for userdel:

`--remove-home`: This removes the user's home directory when deleting the user. Maps to `-r` which has the same behavior.

### DIFF
--- a/changelog/PUMDkmd0RlmEMQJWef3qVQ.md
+++ b/changelog/PUMDkmd0RlmEMQJWef3qVQ.md
@@ -1,0 +1,8 @@
+audience: worker-deployers
+level: minor
+---
+Change `adduser` usage to `useradd`
+
+`adduser` is a debian specific wrapper around `useradd` and friends. By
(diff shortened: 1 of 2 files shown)


lotas (dismissed):
This is cool! Thanks

petemoore (approved):
Thanks Bastien!

PR #7158, opened 2024-07-16
max task dependencies limit

### LINKED ISSUE #7151: test an increase to the maximum number of dependencies a task can have
In the community meeting today we discussed at length whether or not the current maximum number of dependencies a task can have (100) is still sensible. The quick background is that having no limit caused stability issues in the past (see https://bugzilla.mozilla.org/show_bug.cgi?id=1316593#c1). That comment was written 8 years ago though, and it's not clear that it's all necessarily still true.

We've been able to workaround the limit very trivially in the past, but we have more recent use cases that are causing us to spend time and money copying artifacts through intermediate tasks to work around the limit. If we can remove or raise the limit it would avoid this, and remove complexity from a number of graphs.

In the meeting we agreed that we would do some testing to see how high the limit can go before it causes issues.

lotas:
experiment to increase max task dependencies limit

Fixes #7151


services/queue/src/dependencytracker.js
```diff
@@ -35,38 +35,46 @@ class DependencyTracker {
    * This will return {message, details} if there is an error.
    */
   async trackDependencies(task) {
-    for (let requiredTaskId of task.dependencies) {
-      await this.db.fns.add_task_dependency(task.taskId, requiredTaskId, task.requires, task.expires);
-    }
+    await this.db.fns.add_task_dependencies(
```
lotas, line 38: this used to be the slowest part - inserting records one by one is slow, especially if you have lots of those

ui/src/views/Tasks/ViewTask/index.jsx
```diff
@@ -949,12 +950,16 @@ export default class ViewTask extends Component {
...
             <Grid container spacing={3}>
               <Grid item xs={12} md={6}>
-                <TaskDetailsCard
-                  task={task}
-                  dependentTasks={dependentTasks}
-                  dependents={dependents}
-                  onDependentsPageChange={this.handleDependentsPageChange}
-                />
+                <AuthContext.Consumer>
+                  {auth => (
```
lotas, line 954: injecting Auth here because task dependencies would be fetched from the queue API directly which might require credentials on some deployments (non-public)

ui/src/components/TaskDetailsCard/index.jsx
```diff
@@ -99,45 +105,63 @@ export default class TaskDetailsCard extends Component {
...
+      const definitions = await getTaskDefinitions({
+        taskIds: task.dependencies,
+        user,
+      });
+      const statuses = await getTaskStatuses({
+        taskIds: task.dependencies,
+        user,
+      });
+      // merge back everything into { taskId, status, metadata }
+      const dependentTasks = task.dependencies.map(taskId => {
```
lotas, line 145: since we have batch methods now we can load dependencies after the page load more effectively

services/queue/config.yml
```diff
@@ -137,7 +140,8 @@ test:
       parallelism:                1
     dependencyResolver:
       pollingDelay:               250
-    taskCacheMaxSize: 10
+    taskCacheMaxSize:             10
+    taskMaxDependencies:          100
```
petemoore, line 144: Here we have `taskMaxDependencies: 100` but below in `services/queue/schemas/constants.yml` we have `max-task-dependencies: 10000` and in `services/queue/src/main.js` we also have `DEFAULT_MAX_TASK_DEPENDENCIES = 10000`. What is the reason for there being three constant defaults defined, and why are they different?

services/queue/src/main.js
```diff
@@ -226,6 +229,10 @@ let load = loader({
         workerInfo: ctx.workerInfo,
         LRUcache: new QuickLRU({ maxSize: ctx.cfg.app.taskCacheMaxSize || 10 }),
         objectService: ctx.objectService,
+        taskMaxDependencies: Math.min(
+          ctx.cfg.app.taskMaxDependencies || DEFAULT_MAX_TASK_DEPENDENCIES,
```
petemoore, line 233:
```suggestion
          ctx.cfg.app.taskMaxDependencies ?? DEFAULT_MAX_TASK_DEPENDENCIES,
```
If `ctx.cfg.app.taskMaxDependencies` is intentionally 0, i think `||` would consider that false and `taskMaxDependencies` would end up being `DEFAULT_MAX_TASK_DEPENDENCIES`, rather than 0. Probably not a real use case, as probably `ctx.cfg.app.taskMaxDependencies` should never be 0, but if it was, I think `??` would be the correct operator to have `taskMaxDependencies` be 0 if `ctx.cfg.app.taskMaxDependencies` actually is 0. But I'm no javascript expert, and maybe I'm misunderstanding.

services/worker-manager/schemas/constants.yml
```diff
@@ -118,7 +118,7 @@ scope:
 source-pattern: "^(https?:\/\/|ssh:\/\/|git@)"
 
 # Maximum number of dependencies a single task can have
-max-task-dependencies: 100
+max-task-dependencies: 10000
```
petemoore, line 121: Do you know why this was also defined in worker manager? Shouldn't it be a Queue-only thing? And shouldn't it share the same config value if it is really needed, rather than having a duplicate one that can become out-of-sync?

ui/package.json
```diff
@@ -96,7 +96,8 @@
     "shebang-loader": "^0.0.1",
     "snake-case": "^3.0.3",
     "subscriptions-transport-ws": "^0.11.0",
-    "taskcluster-client-web": "^60.4.2",
+    "taskcluster-client": "68.0.4",
+    "taskcluster-client-web": "68.0.4",
```
petemoore, line 100: any idea why this is needed (why it doesn't just come from source directory), and why it was on such an old version?

ui/src/components/TaskDetailsCard/__snapshots__/index.test.jsx.snap
```diff
@@ -32,7 +32,7 @@ exports[`should render TaskDetailsCard 1`] = `
                 class="MuiTypography-root MuiListItemText-secondary MuiTypography-body2 MuiTypography-colorTextSecondary MuiTypography-displayBlock"
               >
                 <button
-                  class="MuiButtonBase-root MuiButton-root MuiButton-text Label-mini-17 MuiButton-textSizeSmall MuiButton-sizeSmall Label-dense-16 Mui-disabled Label-info-23 Label-disabled-18 Mui-disabled"
+                  class="MuiButtonBase-root MuiButton-root MuiButton-text Label-mini-18 MuiButton-textSizeSmall MuiButton-sizeSmall Label-dense-17 Mui-disabled Label-info-24 Label-disabled-19 Mui-disabled"
```
petemoore, line 35: this looks like a generated file - can we hide it from PR view like the other generated files?

petemoore (approved):
This looks good, but I'll be happy for someone else to review especially the UI/graphql changes as I don't really know what I'm looking at there 😬 .

Thanks Yarik!

Also great job adding batch DB functions for performance. I'm sure that will be very much appreciated by callers!

services/queue/src/dependencytracker.js
matt-boris: Nice! 🎉

services/queue/src/dependencytracker.js
```diff
@@ -128,10 +136,9 @@ class DependencyTracker {
       // First remove task
       await this.db.fns.remove_task(task.taskId);
 
-      // Then remove depdencies, because removing these makes it
+      // Then remove dependencies, because removing these makes it
       // easier to trigger the task. So we remove them after removing the task.
-      await Promise.all(task.dependencies.map(requiredTaskId =>
-        this.db.fns.remove_task_dependency(task.taskId, requiredTaskId)));
+      await this.db.fns.remove_task_dependencies(task.taskId, JSON.stringify(task.dependencies));
```
matt-boris, line 141: Love these batch calls 👍🏻

services/queue/config.yml
lotas: This is a `test:` section of the config and defines variables that will be used in tests only. Also a way to test if overrides are working. 
We could also remove it and let the default take place

ui/src/components/TaskDetailsCard/index.jsx
```diff
@@ -99,45 +105,63 @@ export default class TaskDetailsCard extends Component {
...
+  componentDidMount() {
+    this.fetchDependentTasks();
+  }
+
+  async fetchDependentTasks() {
+    const { task, user } = this.props;
+
+    this.setState({ loading: true });
+
+    if (task.dependencies.length > 0) {
```
matt-boris, line 135: Will `dependencies` _always_ be defined here? I'm just wondering if we need to add optional chaining

services/queue/src/main.js
lotas: Oh! Great eagle-eye! I should probably rewrite it though, as node is "special" 

```
> let a
undefined
> a
undefined
> a || 5
5
> a = 0
0
> a || 5
5
> undefined || 'a'
'a'
> 0 || 'a'
'a'
> 0 || 5
5
```

Looks like `||` works differently for numbers 😬

changelog/issue-7151.md
```diff
@@ -0,0 +1,6 @@
+audience: users
+level: major
```
matt-boris, line 2: Why is this a major bump?

services/queue/src/main.js
lotas: Indeed! Thanks for catching this!

matt-boris (commented):
Looking good! 🎉

services/worker-manager/schemas/constants.yml
lotas: great question! 
so this is coming from this feature https://github.com/taskcluster/taskcluster/pull/5448 which merged some views (queue+w-m)
but, now as I'm checking this values is not really  needed on this side, so I'd drop it. `task.yml` is being defined in schemas only to reference `workerPoolId`, `workerType` and `provisionerId` which we can describe without moving `task.yml`

ui/src/components/TaskDetailsCard/index.jsx
lotas: so `task` is coming from the graphql response, which includes this.. but I guess we can check additionally 👍

changelog/issue-7151.md
lotas: good question.. maybe it isn't?
To me it felt like a breaking change, but it's not really breaking anything (unless it doesn't work on prod haha)

matt-boris: 😄 Definitely seemed more like a minor bump to me, unless I was missing some breaking change, but I don't think there is one.

ui/package.json
matt-boris, line 100: Oh, also, this should probably be:

```suggestion
    "taskcluster-client": "^68.0.4",
    "taskcluster-client-web": "^68.0.4",
```

services/worker-manager/schemas/constants.yml
lotas: I create a separate task for this - https://github.com/taskcluster/taskcluster/issues/7242

ui/package.json
lotas: oh, thanks for spotting! was trying to upgrade client and added the wrong one. Removing.

The other one is not linked directly most likely because we deploy docs to netlify, maybe that complicated linking? I'm not sure to be honest

ui/src/components/TaskDetailsCard/__snapshots__/index.test.jsx.snap
lotas: Didn't know about this feature! Neat!

changelog/issue-7151.md
lotas: convinced :) Minor it is then!

ui/src/components/TaskDetailsCard/index.jsx
```diff
@@ -99,45 +105,63 @@ export default class TaskDetailsCard extends Component {
...
+      const definitions = await getTaskDefinitions({
```
lotas, line 136: this is also sub-optimal as we are doing two separate calls because we have two methods .. one returns definition, the other status .. when we were adding batch queries we didn't think about adding one that would include everything. 
It doesn't increase complexity, just the size of the response.

lotas: will address this in https://github.com/taskcluster/taskcluster/issues/7243

ui/package.json
lotas: thanks! indeed

matt-boris (approved):
LGTM, thanks! 🎉

PR #7169, opened 2024-07-25
fix: look for `chfn` in `/usr/bin` instead of `/usr/sbin`

### LINKED ISSUE #7168: linux runtime references chfn in a distro specific location
https://github.com/taskcluster/taskcluster/pull/7154 added a dependency on this command, which is fine, but on Debian based systems it exists in `/usr/bin/chfn`, not `/usr/sbin/chfn`.

It's not immediately obvious to me what the right way to fix this is, so I'm filing instead of sending a PR :laughing:. I considered replacing it `which chfn`, but that seemed rather hacky.

cc @Eijebong

bhearsum:
The latter doesn't exist on Debian based distributions.

Github Bug/Issue: Fixes #7168


petemoore (approved):

petemoore (approved):

petemoore:
Thanks Ben!

PR #7170, opened 2024-07-25
fix: map uid/gid of user running podman into useful uids/gids in podman containers

bhearsum:
This patch maps uids and gids from the host system into containers that run with Podman, which helps to avoid permission issues within the container on mounted volumes.

### DIFF
--- a/changelog/issue-7128-1.md
+++ b/changelog/issue-7128-1.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 7128
+---
+
+Generic Worker / D2G partial bug fix: support has been improved for running Docker Worker tasks with caches under Generic Worker. Previously, caches from a Docker Worker task running under Generic Worker containing files owned by a user other than root would not be owned by the same (container) user when the cache was mounted in a future task. D2G now consistently maps container uids and gids to host subuids and subgids (when caches are used) in order that cache file ownership, as seen from inside the container, is maintained across task runs. However, this fix does not apply when the privileged capability is enabled in the Docker Worker payload, since privileged tasks are executed under docker rather than podman. This fix only applies when podman is used.
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -312,8 +312,42 @@ func runCommand(containerName string, dwPayload *dockerworker.DockerWorkerPayloa
 	command.WriteString(" --memory-swap -1 --pids-limit -1")
 	// Only podman supports inheriting host ulimits. `docker` uses docker
 	// daemon settings by default.
+	// Also, only podman supports mapping uids
 	if tool == "podman" {
 		command.WriteString(" --ulimit host")
+
+		if len(dwPayload.Cache) > 0 {
+			// We map uids and gids to help avoid issues with permissions within the container
+			// on mounted volumes.
+			// Cribbed from https://stackoverflow.com/questions/70770437/mapping-of-user-ids
+			// and https://github.com/containers/podman/blob/main/troubleshooting.md#solution-36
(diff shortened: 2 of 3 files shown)


bhearsum:
An example of this is: https://firefox-ci-tc.services.mozilla.com/tasks/PG3dhNDWQDeQCfVsZeYmrw/runs/34

bhearsum:
I'm going to mark this as "ready", but I will note that it would be ideal to avoid the hardcode of `64536` for the uid/gid map range.

tools/d2g/d2g.go
```diff
@@ -312,8 +312,40 @@ func runCommand(containerName string, dwPayload *dockerworker.DockerWorkerPayloa
...
+		// lower part of the subuids.
+		command.WriteString(fmt.Sprintf(" --uidmap 0:1:%v", uid_start))
+		// Map the UIDs between $uid+1 and 64536 in the container to the remaining subuids.
+		// Ideally 64536 would be pulled from `podman info` running as a task user,
+		// but we're running as root here, so we can't do that.
+		command.WriteString(fmt.Sprintf(" --uidmap %v:%v:%v", uid_start+1, uid_start+1, mapping_range))
+		// Same thing for GIDs
+		command.WriteString(fmt.Sprintf(" --gidmap %v:0:1", gid_start))
+		command.WriteString(fmt.Sprintf(" --gidmap 0:1:%v", gid_start))
+		command.WriteString(fmt.Sprintf(" --gidmap %v:%v:%v", gid_start+1, gid_start+1, mapping_range))
```
petemoore, line 348: This additional complexity is only required if a Docker Worker cache is used. To keep the payload as simple as possible in the common case, would it be ok to only add the mappings if a cache is used?

changelog/issue-7128-1.md
```diff
@@ -0,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 7128
+---
+
+non-privileged d2g payloads now map uids and gids from the host into the container, which avoids most issues with permissions when mounting volumes into a container
```
petemoore, line 6: See https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/changelog.md#snippet-text for general guidance when writing change logs.

The change log should be comprised of sentences (i.e. starting with a capital letter, ending with a period). Bug fixes should ideally describe under which circumstances the bug occurred. Something like:

> Generic Worker / D2G bug fix: support has been improved for running Docker Worker tasks with caches under Generic Worker. Previously, caches from a Docker Worker task running under Generic Worker containing files owned by a user other than root would not be owned by the same (container) user when the cache was mounted in a future task. D2G now consistently maps container uids and gids to host subuids and subgids (when caches are used) in order that cache file ownership, as seen from inside the container, is maintained across task runs.

My version might be a bit too wordy, but you get the idea. I really wanted to capture the fix pertains to tasks running under Generic Worker, when a Docker Worker task is submitted, and specifically when caches are used, in order to guarantee consistent file ownership as seen from inside the container.

petemoore (dismissed):

petemoore:
> I'm going to mark this as "ready", but I will note that it would be ideal to avoid the hardcode of `64536` for the uid/gid map range.

I think it is sufficient to file a follow-up issue for that. I don't think we need to block on it.

tools/d2g/d2g.go
bhearsum: That seems sensible, yeah.

petemoore (approved):
Looks nice Ben, many thanks!

I'm going to slightly modify the commit, just to remove the noise in the yaml files whose structural content did not change, but the files were slightly formatted. But I'll commit as you.

PR #7175, opened 2024-07-31
fix(queue): Emit pulse message to 'task-exception' regardless of retr…

### LINKED ISSUE #7174: Queue service fails to emit `task-exception` pulse message when run is automatically re-tried
**Describe the bug**

When a run fails with state `exception` such that it is automatically re-tried, the queue service does not emit a pulse event to the `task-exception` exchange. This happens because the call to `this.publisher.taskException` happens in this else clause:
https://github.com/taskcluster/taskcluster/blob/9afed31dcbad17729523fa84de6beaa5b81bb132/services/queue/src/api.js#L1809

I suspect this was implemented this way intentionally, but I would argue that this is actually a bug. Consider my use case. I'm building an ETL that is aiming to track costs. These runs that get automatically retried, incur just as much cost as any other run, so it's important to track them.

Besides, we emit one event per run in all other cases.. so I don't see why whether there is an automatic retry or not should make any difference.

ahal:
…y state

Previously, we were only emitting a message if the run was not being retried. This is a bug as we need to be able to track all runs of a task.

I moved the publish logic above the call to `task-pending` as this matches the order of events (first the old run had an exception, then the new run is pending).

Github Bug/Issue: Fixes #7174

### DIFF
--- a/changelog/issue-7174.md
+++ b/changelog/issue-7174.md
@@ -1,0 +1,5 @@
+audience: users
+level: major
+reference: issue 7174
+---
+Queue service now emits pulse messages to the `exchange/taskcluster-queue/v1/task-exception` exchange when a task has an exception that is automatically retried.
--- a/services/queue/src/api.js
+++ b/services/queue/src/api.js
@@ -1795,10 +1795,19 @@ builder.declare({
     tags: task.tags,
   };
 
-  // If a newRun was created and it is a retry with state pending then we better
-  // publish messages about it. And if we're not retrying the task, because then
-  // the task is resolved as it has no more runs, and we publish a message about
-  // task-exception.
+  // Publish message about taskException
+  await this.publisher.taskException({
+    status,
+    runId,
+    task: taskPulseContents,
+    workerGroup: run.workerGroup,
+    workerId: run.workerId,
+  }, task.routes);
+  this.monitor.log.taskException({ taskId, runId });
+
+  // If a newRun was created and it is a retry with state pending then we
+  // better publish messages about it. If we're not retrying the task, the task
+  // is resolved as it has no more runs.
   let newRun = task.runs[runId + 1];
   if (newRun &&
       task.runs.length - 1 === runId + 1 &&
@@ -1822,16 +1831,6 @@ builder.declare({
       task.schedulerId,
       'exception',
     );
-
-    // Publish message about taskException
-    await this.publisher.taskException({
-      status,
-      runId,
-      task: taskPulseContents,
-      workerGroup: run.workerGroup,
-      workerId: run.workerId,
-    }, task.routes);
-    this.monitor.log.taskException({ taskId, runId });
   }
 
   // Reply to caller
--- a/services/queue/src/claimresolver.js
+++ b/services/queue/src/claimresolver.js
@@ -145,6 +145,15 @@ class ClaimResolver {
 
     let status = task.status();
 
+    // Publish message about task exception
+    await this.publisher.taskException({
+      status: status,
+      runId: runId,
+      workerGroup: run.workerGroup,
+      workerId: run.workerId,
+    }, task.routes);
+    this.monitor.log.taskException({ taskId, runId });
+
     // If a newRun was created and it is a retry with state pending then we
     // better publish messages about it
     let newRun = task.runs[runId + 1];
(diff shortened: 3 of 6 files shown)


ahal:
I think the `service-worker-manager` failure is an intermittent unrelated to this PR?

petemoore:
Thanks Andrew!

Looking at [the comments](https://github.com/taskcluster/taskcluster/pull/7175/files#diff-5658805efc5f832d8836628ad1915970b18576d9c518eedde3569438ae39bc4aL1790-L1792) perhaps the previous behavior was intentional, but I too find it counterintuitive. I suspect the author intended that intermediate states (i.e. temporary exceptions that would be followed by a retry) would not be reported, so that consumers could be reasonably confident if they receive a message of an exception, that the task resolution isn't likely to change. However, it can always change with a rerun, and sometimes it is useful for consumers to track intermediate changes (i.e. track the results of each task run, not just the end state of the task) so I think the simpler 1:1 mapping from task run resolutions to pulse messages makes more sense. But perhaps we should ship this as a major release, as it is a change to previous behavior, and tools that are listening for task exceptions are potentially impacted.

@djmitche @lotas @matt-boris @Eijebong @aerickson @jcristau Curious on your opinions here.

Eijebong:
This change sounds reasonable to me. I don't think automatic retries should be treated differently than manual ones.

However, the linked comment might need to be updated?

ahal:
I agree with your assessment!

> I suspect the author intended that intermediate states (i.e. temporary exceptions that would be followed by a retry) would not be reported, so that consumers could be reasonably confident if they receive a message of an exception, that the task resolution isn't likely to change

Agreed, and I can kind of see where the design decision is coming from. But I do just want to add that doing this makes some big assumptions about what the consumer on the other end of the pulse queue is trying to accomplish :). In my experience, Taskcluster has always been about empowering users, and this design decision is a rare case that goes counter to that.

ahal:
And yeah, I could see it being a major change.. I was debating myself. Want me to update the changelog?

Eijebong:
I was about to comment that it probably warrants a major release since this is technically a breaking change :)

ahal:
Dustin raised a point in the issue that requires some thought before merging this.

petemoore:
> Dustin raised a point in the issue that requires some thought before merging this.

He makes a very valid point, I've responded in the issue too. Let's discuss at MozWeek.

ahal:
@petemoore @lotas sorry for neglecting this for a bit.. I've made it a breaking change and updated a spot in the docs I found. Please let me know if there's anything else I'm missing!

ahal:
I thought about adding a `retried` field to the status schema.. but I think that'll be a bit more work than I'm able to commit atm. I think it would be a good follow-up issue.

ahal:
I think I need to run `yarn generate` but I'm having some troubles. Will debug tomorrow

lotas (approved):
Thanks!

lotas:
@ahal there's something I forgot to move from the meeting notes here:

https://github.com/taskcluster/taskcluster/blob/main/services/queue/src/claimresolver.js#L150-L177

Here is similar logic and worth updating here too

lotas (changes requested):
https://github.com/taskcluster/taskcluster/blob/main/services/queue/src/claimresolver.js#L150-L177

ahal:
@lotas I'm a bit confused about the docs. I ran `yarn generate` successfully, but it didn't result in any changes. Did I edit the wrong file? Do I not need to generate docs in this case?

matt-boris:
@ahal looks like you're fine as far as code/docs generation goes. `meta-generate` task would've failed if you needed to do so.

ahal:
Interesting, I thought doc changes required a generation.

Is that test failure an known intermittent? I don't see any errors that look like they could be caused by this patch and `yarn test` passes for me locally. Could someone re-run it for me?

(Aside: it would also be nice if we could get releng scopes to rerun/cancel tasks on community :))

matt-boris:
> I don't see any errors that look like they could be caused by this patch and `yarn test` passes for me locally.

If it's not intermittent, it's possible your local tests are all passing fine, but once they run on the CI they have `NO_TEST_SKIP: 'true'` set in the env to have it test against real services alongside the mocks.


> (Aside: it would also be nice if we could get releng scopes to rerun/cancel tasks on community :))

Apologies! Had no clue you guys didn't have these scopes! We'll get them to you. I reran the task in the meantime.

ahal:
Looks like it failed again. I don't really understand the log, but it looks like the tests actually all pass, but the task then fails to copy artifacts or something?

lotas:
Thanks @ahal . Failing test might be due to some expected pulse messages not happening, or more messages appear than test expect. I'll check it locally also

matt-boris:
@ahal granting permissions to releng here https://github.com/taskcluster/community-tc-config/pull/842

ahal:
Ah thanks, I missed the reference to `retry_test.js` in the log, I was expecting to see a little red x somewhere

ahal:
This should be ready to merge, assuming it passes review.

matt-boris (approved):
LGTM thanks Andrew

PR #7192, opened 2024-08-02
Poll more frequently for EC2 spot interruption

### LINKED ISSUE #7167: Unexpected graceful termination requests and claim-expired tasks
Take a look at e.g. https://community-tc.services.mozilla.com/tasks/f8pxA7E6R4iHc63Mukj7tQ/runs/3/logs/public/logs/live.log#L223

On the fourth attempt of this task, we see a graceful termination request. On the first 3 runs of this task, we had exception/claim-expired resolutions.

We are hitting [this](https://github.com/taskcluster/taskcluster/blob/1f69e1c2dc0cc6470a50944b8ee97fecdd158e8e/workers/generic-worker/main.go#L1153).

There are only two reasons Generic Worker decides to abort a task iirc, one is if you hit the max run time (that didn't happen in this case) and the other is if it believes the machine is about to be shutdown.

This pool runs in ec2, so the aws spot termination handler is be in use.

We don't see this line in the logs of task run 3: https://github.com/taskcluster/taskcluster/blob/1f69e1c2dc0cc6470a50944b8ee97fecdd158e8e/tools/worker-runner/provider/aws/awsprovider.go#L107

I wonder if somehow the task manages to send Generic Worker a `SIGINT` or something similar, even though it shouldn't be possible.

I think the task finally resolved on the 6th run. According to @choller, this has been happening a lot. @jschwartzentruber might have some ideas. This is Generic Worker, running tasks using KVM in EC2 on `c5.metal` instances.

jschwartzentruber:
Change the polling period for EC2 spot instance interruption notices to 5 seconds, as recommended by [AWS documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-instance-termination-notices.html).

Github Bug/Issue: Fixes #7167

### DIFF
--- a/changelog/issue-7167.md
+++ b/changelog/issue-7167.md
@@ -1,0 +1,5 @@
+audience: deployers
+level: patch
+reference: issue 7167
+---
+Change the polling period for EC2 spot instance interruption notices to 5 seconds, as recommended by AWS documentation.
--- a/tools/worker-runner/provider/aws/awsprovider.go
+++ b/tools/worker-runner/provider/aws/awsprovider.go
@@ -121,7 +121,7 @@ func (p *AWSProvider) checkTerminationTime() bool {
 
 func (p *AWSProvider) WorkerStarted(state *run.State) error {
 	// start polling for graceful shutdown
-	p.terminationTicker = time.NewTicker(30 * time.Second)
+	p.terminationTicker = time.NewTicker(5 * time.Second)
 	p.proto.AddCapability("graceful-termination")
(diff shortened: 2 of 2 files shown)


lotas (approved):
Looks good. The test failures are not related

One was rerun here https://community-tc.services.mozilla.com/tasks/ITTU6wpTRbaY7cp6E3suPw 
and for other is intermittent - https://github.com/taskcluster/taskcluster/issues/7195

PR #7203, opened 2024-08-21
Fixes taskcluster.yml rendering for some payloads

### LINKED ISSUE #7202: Github render taskcluster yaml doesn't escape some values
There are few errors like this:

```
SyntaxError: Expected ',' or '}' after property value in JSON at position 32
    at JSON.parse (<anonymous>)
    at getEventPayload (file:///app/services/github/src/fake-payloads.js:744:22)
    at Proxy.<anonymous> (file:///app/services/github/src/api.js:905:23)
    at file:///app/libraries/api/src/middleware/handle.js:13:28
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) 
```

Which is happening because `github.renderTaskclusterYml` doesn't escape some arguments

lotas:
In some payloads `branch` value was not properly escaped which resulted in invalid json

Fixes #7202

### DIFF
--- a/services/github/src/fake-payloads.js
+++ b/services/github/src/fake-payloads.js
@@ -745,7 +745,7 @@ export const getEventPayload = function getEventPayload(type, action, organizati
     mapping[type]
       .replaceAll('$ORGANIZATION', organization)
       .replaceAll('$REPOSITORY', repository)
-      .replaceAll('$BRANCH', branch)
+      .replaceAll('$BRANCH', encodeURI(branch))
   );
 
   return { ...event, action, ...overrides };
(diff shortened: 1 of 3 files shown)


services/github/src/fake-payloads.js
```diff
@@ -745,7 +745,7 @@ export const getEventPayload = function getEventPayload(type, action, organizati
     mapping[type]
       .replaceAll('$ORGANIZATION', organization)
```
lotas, line 746: `organization` and `branch` were validated with schema, but `branch` was coming from a free-form `overrides` param

matt-boris (approved):

PR #7208, opened 2024-08-26
feat(github): Enhance integration

### LINKED ISSUE #5669: Enhance github integration: Include more useful information in check run output
At the moment, when task is finished, github service is sending `output.text` information for each check run, like [here](https://github.com/taskcluster/taskcluster/pull/7057/checks?check_run_id=26643525105)

At the moment it includes:

```
[link to the task]
[link to the logs]

< contents of the public/github/customCheckRunText.md artifact if it exists >

---
< last 250 lines of live.log >
```

Some ideas what could be included as well, in order to provide as much information as possible:

- total task execution time, when started, when finished, how long it took
- always include beginning of the logs which contains worker details, version and environment info
- link to the task group
- exact status information: failed, canceled, exceptions, number of runs, ..
- attach list of artifacts produced by this build (with links)

thakkarujas5:
This introduces more information in the check run output like task runtime information, status information, artifact list, head of logs.

Fixes #5669


services/github/src/handlers/status.js
```diff
@@ -177,10 +196,13 @@ export async function statusHandler(message) {
         taskDefinition.payload?.logs?.live || taskDefinition.payload?.log,
       ),
     ));
+    output.addText(markdownAnchor(CHECKTASK_TEXT, taskGroupUI(this.context.cfg.taskcluster.rootUrl, taskGroupId)));
+
     if (customCheckRunText) {
       output.addText(customCheckRunText);
     }
     if (liveLogText) {
+      output.addText(markdownLog(headLog(liveLogText, 20, githubCheck.output.getRemainingMaxSize())));
```
lotas, line 205: I think it would be good to add some separator between both lines to make it clear that something got cut off from the middle
Or, maybe, it would be better to have single function that would take this `liveLogText`, split it into lines, then grab first `20` and last `~200` lines and adds `...(X lines hidden)...` 
Would be easier to observe `getRemainingMaxSize()` in single call

Just a suggestion, thanks!

services/github/src/handlers/utils.js
```diff
@@ -193,6 +193,32 @@ export const isCollaborator = async (instGithub, organization, repository, login
   }));
 };
 
+export const getTimeDifference = (timestamp1, timestamp2) => {
+
```
lotas, line 197: probably would be good to check if timestamps and dates that come here are valid. From the code that uses it it might seem as those values might be `undefined` so some check and tests would be helpful

services/github/src/utils.js
```diff
@@ -212,14 +219,20 @@ export const checkGithubSignature = (secret, payload, signature) => {
   return compareSignatures(signature, expected);
 };
 
+export const buildUrl = (taskId, runId, artifactName) => {
+  return `${process.env.TASKCLUSTER_ROOT_URL}/tasks/${taskId}/runs/${runId}/${artifactName}`;
```
lotas, line 223: let's not use `process.env` here and just explicitly pass `rootUrl` as we do in https://github.com/taskcluster/taskcluster/blob/main/services/github/src/handlers/utils.js#L9-L10 

Probably also makes sense to place this function close to the other similar ones?

services/github/src/handlers/status.js
thakkarujas5: Screenshot 2024-08-28 at 12 00 54 AM

In my local dev setup I have logged 10 lines of head, 10 lines of tail and hidden in between, will change these values while pushing.

services/github/src/utils.js
thakkarujas5: Have incorporated the change of not using process.env, didn't get the question of placing function to similar ones, do we mean to keep it in utils inside handlers?

services/github/src/handlers/utils.js
thakkarujas5: For writing the tests, I am first facing issue in running the tests, will get back to you in case I am stuck for too long.

services/github/src/constants.js
```diff
@@ -84,6 +84,7 @@ export const PUBLISHERS = {
 
 export const CHECKLOGS_TEXT = 'View logs in Taskcluster';
 export const CHECKRUN_TEXT = 'View task in Taskcluster';
+export const CHECKTASK_TEXT = 'View task group in Taskcluster';
```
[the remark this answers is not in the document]
thakkarujas5: Have incorporated this change.

services/github/src/handlers/utils.js
lotas: any luck with the tests and checking if passed timestamps are defined?

ujas-tally: I have written the test for this function, will need to add a undefined check for this, slipped out of my mind while updating tests for existing statusHandler function

services/github/src/utils.js
```diff
@@ -173,6 +173,30 @@ export const tailLog = (log, maxLines = 250, maxPayloadLength = 30000) => {
...
+ * Github checks API call is limited to 64kb
+ * @param {string} log
+ * @param {number} headLines
+ * @param {number} tailLines
+ * @param {number} maxPayloadLength
+ * @returns string
+ */
+export const extractLog = (log, headLines = 20, tailLines = 200, maxPayloadLength = 30000) => {
+
+  const logString = ansi2txt(log).substring(0, maxPayloadLength);
```
lotas, line 186: I think this logic here will not work as expected. `maxPayloadLength` is for the github check (max we can attache for github check), but here it would parse the log and take first 30k chars only 
This length should be applied later and probably in some recursive/iterative way. 

We want output to look like:

```
line1
line2
..
line20

... XX lines hidden .. 

lineN-200
lineN-199
..
lineN
```

and the trick is that we don't know how long last lines are so we can be creative and check if last `tailLines` fit into `maxPayloadLength-sum(line1..line20)` and if not, take less lines and check again until we fit.

services/github/src/handlers/utils.js
thakkarujas5: Have written tests for this function, need to add a undefined check, slipped out of my mind while writing tests for existing statusHandler function

services/github/test/handler_test.js
```diff
@@ -48,6 +48,16 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
   let github = null;
   let handlers = null;
 
+  function buildArtifactLinks(limit, taskId){
+
+    const artifactLinks = [];
+
+    for(let i = 0;i < limit;i++) {
+      artifactLinks.push(`[artifact-${i}](${libUrls.testRootUrl()}/tasks/${taskId}/runs/0/artifact-${i})`);
```
lotas, line 56: Is this method used at all? Only see references in tests

lotas: ah sorry... it is for the tests :) nevermind.

services/github/test/handler_utils_test.js
```diff
@@ -106,4 +106,22 @@ helper.secrets.mockSuite(testing.suiteName(), [], function() {
...
+
+    const START_TIMESTAMP = new Date('2024-07-16T18:23:18.118Z');
+    const END_TIMESTAMP_MILLISECONDS = new Date('2024-07-16T18:23:18.128Z');
+    const END_TIMESTAMP_SECONDS = new Date('2024-07-16T18:23:28.118Z');
+    const END_TIMESTAMP_MINUTES = new Date('2024-07-16T18:33:18.118Z');
+    const END_TIMESTAMP_HOURS = new Date('2024-07-17T04:23:18.118Z');
+    const END_TIMESTAMP_DAYS = new Date('2024-07-26T18:23:18.118Z');
+    const END_TIMESTAMP_GENERIC = new Date('2024-07-27T04:33:28.128Z');
+
+    assert.equal(getTimeDifference(START_TIMESTAMP, END_TIMESTAMP_MILLISECONDS), "10 milliseconds");
```
lotas, line 120: this is great! 
would be good to include tests for the missing/invalid dates also since `getTimeDifference` is not validating the input

services/github/src/handlers/status.js
lotas: looks great!

services/github/src/handlers/utils.js
thakkarujas5: Also if the timestamps are undefined, we will ideally want to remove it from the check run output right

lotas: I think it is highly unlikely we won't have any values in there, but if something is missing we can replace value with `n/a`. Would be easier to spot that something is wrong, as if we would hide the line, it wouldn't be visible

services/github/src/utils.js
thakkarujas5: I am thinking of binary search here. The problem I see with recursive or an iterative solution is that we need some fixed step defined for how many lines we would decrement, and worst case lets say tomorrow github increases the size of the check run output from 64kb to 200kb, so as we have room for more number of lines, we wouldn't have to change our code and it would run very efficiently. What are your thoughts?

services/github/test/handler_utils_test.js
thakkarujas5: can we receive invalid dates, or is it just to make the tests robust?

services/github/src/utils.js
lotas: Yeah, I think the most simple solution here is the best. 
This function already accepts `maxPayloadLength` so it shouldn't care if github increases or decreases something, it only knows how many chars to return at most. 

Let's not over-engineer. Maybe taking first 20 lines should be fixed, and then we take last `maxPayloadLength  - length(first20Lines) - somebuffer` and then just take that much data from the end of the log, and split it with newlines. What do you think?

services/github/test/handler_utils_test.js
lotas: Someone might always use this function with some invalid payload. And because we don't have strict types here, we could at least add some defensive mechanisms to return `n/a` if something is not a valid date

services/github/src/utils.js
```diff
@@ -173,6 +173,30 @@ export const tailLog = (log, maxLines = 250, maxPayloadLength = 30000) => {
...
+  const lines = logString.split('\n');
+
+  if(lines.length <= headLines + tailLines) {
+    return lines.join('\n');
+  }
+
+  const headLog = lines.slice(0, headLines).join('\n');
+  const tailLog = lines.slice(-tailLines).join('\n');
```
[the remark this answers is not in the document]
lotas: pls ignore, i missed previous lines that check that

services/github/src/handlers/utils.js
```diff
@@ -193,6 +193,48 @@ export const isCollaborator = async (instGithub, organization, repository, login
...
+  const isValidDate = (date) => {
+    return !isNaN(Date.parse(date));
+  };
+
+  if (timestamp1 === undefined || timestamp2 === undefined) {
+    return null;
+  }
+
```
thakkarujas5, line 205: Added check for invalid dates

services/github/src/handlers/utils.js
thakkarujas5: added n/a condition in check run output for invalid dates.

services/github/src/utils.js
```diff
@@ -173,6 +173,94 @@ export const tailLog = (log, maxLines = 250, maxPayloadLength = 30000) => {
...
+ * @param {string} log
+ * @param {number} headLines
+ * @param {number} tailLines
+ * @param {number} maxPayloadLength
+ * @returns string
+ */
+export const extractLog = (log, headLines = 20, tailLines = 200, maxPayloadLength = 30000) => {
+
+  if(headLines < 0 && tailLines < 0) {return extractLog(log, -1 * tailLines, -1 * headLines, maxPayloadLength);}
+
```
thakkarujas5, line 226: handles cases for negative inputs to this function

services/github/src/handlers/utils.js
```diff
@@ -193,6 +193,48 @@ export const isCollaborator = async (instGithub, organization, repository, login
...
+  if(!isValidDate(timestamp1) || !isValidDate(timestamp2)) {
+    return null;
+  }
+
```
thakkarujas5, line 209: handles cases for invalid dates

services/github/src/utils.js
```diff
@@ -173,6 +173,94 @@ export const tailLog = (log, maxLines = 250, maxPayloadLength = 30000) => {
...
+  const logString = ansi2txt(log);
+  const lines = logString.split('\n');
+
+  if(lines.length <= headLines + tailLines) {
+
+    if (maxPayloadLength <= logString.length) {
+      return extractHeadLinesFromLog(logString, maxPayloadLength);
+    }
+    return logString;
```
thakkarujas5, line 243: within overlapping conditions, added check if the max payload length allows us to log it, ideally this condition should never be hit, but added for safety purpose, will remove if required.

services/github/src/utils.js
thakkarujas5: modified the extractLog function as you suggested, just one thing I had a doubt on, when you say someBuffer, what should be the optimal size for it?

thakkarujas5:
Screenshot 2024-09-09 at 12 49 50 AM

Currently the output looks like a mess as there is no separation between the items. I am thinking to separate it with headings like this

# Task Status
# Task Logs
# Task Links

What are your thoughts on this?

services/github/src/handlers/status.js
```diff
@@ -164,7 +165,28 @@ export async function statusHandler(message) {
...
     });
+
+    const artifactList = await this.queueClient.listArtifacts(taskId, runId, { limit: 50 });
+
     const output = githubCheck.output;
+
+    if(runs.length > 0) {
+      const taskExecutionTime = getTimeDifference(runs[runId]?.started, runs[runId]?.resolved);
+
+      output.addText(`Started: ${runs[runId]?.started !== undefined ? runs[runId]?.started : "n/a"}`);
```
lotas, line 176: nitpick: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing 
```suggestion
      output.addText(`Started: ${runs[runId]?.started ?? "n/a"}`);
```

services/github/src/handlers/utils.js
lotas: thanks!

services/github/src/utils.js
lotas: Do you think it will be useful/used anywhere?

thakkarujas5: I don't think anyone would trigger it knowingly, thought it would be good behaviour if anyone would trigger with negative inputs, it should cycle from back. 

Will remove this check

services/github/src/utils.js
```diff
@@ -173,6 +173,94 @@ export const tailLog = (log, maxLines = 250, maxPayloadLength = 30000) => {
...
+/**
+ *
+ * @param {string} logString
+ * @param {number} maxPayloadLength
+ * @param {number} tailLines
+ * @returns string or null
+ */
+export const extractTailLinesFromLog = (logString, maxPayloadLength, tailLines) => {
+
+  if(tailLines === 0) {return null;}
```
lotas, line 185: I think our eslint config is broken, it should have complained about the spaces and new lines 🙈 
```suggestion
  if (tailLines === 0) {
    return null;
  }
```

thakkarujas5: i had ran yarn lint:fix, didnt show any complaints, will fix this.

services/github/src/utils.js
lotas: Thanks! By `someBuffer` I mean extra formatting that is being used in the end - new lines and "xxx lines hidden" part. Just some random number :) `42` will do

services/github/src/utils.js
thakkarujas5: @lotas should I remove this condition as well then?

lotas:
> Screenshot 2024-09-09 at 12 49 50 AM
> 
> Currently the output looks like a mess as there is no separation between the items. I am thinking to separate it with headings like this
> 
>     1. Task Status
> 
>     2. Task Logs
> 
>     3. Task Links
> 
> 
> What are your thoughts on this?

Looks good!
I like your suggestion and grouping. I would keep though the links as they were in the beginning - when you open a check you don't want to scroll to the bottom to open any of those. 
We can also save some space by placing all on a single line? 
```md
[View task in Taskcluster](..)  | [View task group in Taskcluster](...) | [View full log](...)
```

For artifacts can you try adding a small header and maybe a list? 

```md
#### Artifacts

- [art1](..)
- [art2](..)
```

just suggestions! Thank you for doing this

services/github/src/utils.js
lotas: yeah, I don't know why .. maybe some files are being excluded. let me check

lotas: or maybe I'm wrong and it was never part of the project's eslint.. heh.. nevermind then. Please ignore then :)

lotas: I'll add a separate ticket

services/github/src/utils.js
lotas: Yeah, if lines "fit" into allowed length I think it is not necessary to do those extra manipulations we can just return all

thakkarujas5:
> > Screenshot 2024-09-09 at 12 49 50 AM
> > Currently the output looks like a mess as there is no separation between the items. I am thinking to separate it with headings like this
> > ```
> > 1. Task Status
> > 
> > 2. Task Logs
> > 
> > 3. Task Links
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What are your thoughts on this?
> 
> Looks good! I like your suggestion and grouping. I would keep though the links as they were in the beginning - when you open a check you don't want to scroll to the bottom to open any of those. We can also save some space by placing all on a single line?
> 
> ```
> [View task in Taskcluster](..)  | [View task group in Taskcluster](...) | [View full log](...)
> ```
> 
> For artifacts can you try adding a small header and maybe a list?
> 
> ```
> #### Artifacts
> 
> - [art1](..)
> - [art2](..)
> ```
> 
> just suggestions! Thank you for doing this

Screenshot 2024-09-09 at 8 25 40 PM


Updated as you suggested 👍

services/github/src/utils.js
thakkarujas5: Added the buffer

services/github/src/utils.js
```diff
@@ -173,6 +173,85 @@ export const tailLog = (log, maxLines = 250, maxPayloadLength = 30000) => {
...
+ */
+export const extractTailLinesFromLog = (logString, maxPayloadLength, tailLines) => {
+
+  if (tailLines === 0) {
+    return null;
+  }
+
+  const tailLog = logString.split("\n").slice(-tailLines).join("\n");
+
+  if (logString.length <= maxPayloadLength) {
```
lotas, line 191: should this be above the previous line maybe? If whole log fits, maybe it is redundant to run (somewhat expensive) string split, slice and join?

thakkarujas5: Ahh, yes, correct, my bad. Thanks for pointing this out.

services/github/test/handler_test.js
```diff
@@ -1535,14 +1564,16 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
...
+        resolved: RESOLVED,
       });
 
       assert(github.inst(9988).checks.update.calledOnce, 'checks.update was not called');
       let [args] = github.inst(9988).checks.update.firstCall.args;
       /* eslint-disable comma-dangle */
       assert.strictEqual(
         args.output.text,
-        `[${CHECKRUN_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID})\n[${CHECKLOGS_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID}/runs/0/logs/live/public/logs/live.log)\n${CUSTOM_CHECKRUN_TEXT}\n`
+        `### Task Links\n[${CHECKRUN_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID}) | [${CHECKLOGS_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID}/runs/0/logs/live/public/logs/live.log) | [${CHECK_TASK_GROUP_TEXT}](${libUrls.testRootUrl()}/tasks/groups/${TASKGROUPID})\n### Task Status\nStarted: ${STARTED}\nResolved: ${RESOLVED}\nTask Execution Time: 1 day\nTask Status: completed\nReason Resolved: completed\nRunId: 0\n### Artifacts\n${buildArtifactLinks(50, CUSTOM_CHECKRUN_TASKID)}\n${CUSTOM_CHECKRUN_TEXT}\n`
```
lotas, line 1576:
```suggestion
        `[${CHECKRUN_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID}) | [${CHECKLOGS_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID}/runs/0/logs/live/public/logs/live.log) | [${CHECK_TASK_GROUP_TEXT}](${libUrls.testRootUrl()}/tasks/groups/${TASKGROUPID})\n### Task Status\nStarted: ${STARTED}\nResolved: ${RESOLVED}\nTask Execution Time: 1 day\nTask Status: completed\nReason Resolved: completed\nRunId: 0\n### Artifacts\n${buildArtifactLinks(50, CUSTOM_CHECKRUN_TASKID)}\n${CUSTOM_CHECKRUN_TEXT}\n`
```

services/github/test/handler_test.js
```diff
@@ -1566,14 +1597,16 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
...
+        resolved: RESOLVED,
       });
 
       assert(github.inst(9988).checks.update.calledOnce, 'checks.update was not called');
       let [args] = github.inst(9988).checks.update.firstCall.args;
       /* eslint-disable comma-dangle */
       assert.strictEqual(
         args.output.text,
-        `[${CHECKRUN_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID})\n[${CHECKLOGS_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID}/runs/0/logs/live/public/logs/live.log)\n\n---\n\n\`\`\`bash\n${LIVE_LOG_TEXT}\n\`\`\`\n`
+        `### Task Links\n[${CHECKRUN_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID}) | [${CHECKLOGS_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID}/runs/0/logs/live/public/logs/live.log) | [${CHECK_TASK_GROUP_TEXT}](${libUrls.testRootUrl()}/tasks/groups/${TASKGROUPID})\n### Task Status\nStarted: ${STARTED}\nResolved: ${RESOLVED}\nTask Execution Time: 1 day\nTask Status: completed\nReason Resolved: completed\nRunId: 0\n### Artifacts\n${buildArtifactLinks(50, CUSTOM_CHECKRUN_TASKID)}\n\n---\n\n\`\`\`bash\n${LIVE_LOG_TEXT}\n\`\`\`\n`
```
lotas, line 1609:
```suggestion
        `[${CHECKRUN_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID}) | [${CHECKLOGS_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_CHECKRUN_TASKID}/runs/0/logs/live/public/logs/live.log) | [${CHECK_TASK_GROUP_TEXT}](${libUrls.testRootUrl()}/tasks/groups/${TASKGROUPID})\n### Task Status\nStarted: ${STARTED}\nResolved: ${RESOLVED}\nTask Execution Time: 1 day\nTask Status: completed\nReason Resolved: completed\nRunId: 0\n### Artifacts\n${buildArtifactLinks(50, CUSTOM_CHECKRUN_TASKID)}\n\n---\n\n\`\`\`bash\n${LIVE_LOG_TEXT}\n\`\`\`\n`
```

services/github/test/handler_test.js
```diff
@@ -1597,14 +1630,16 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
...
+        resolved: RESOLVED,
       });
 
       assert(github.inst(9988).checks.update.calledOnce, 'checks.update was not called');
       let [args] = github.inst(9988).checks.update.firstCall.args;
       /* eslint-disable comma-dangle */
       assert.strictEqual(
         args.output.text,
-        `[${CHECKRUN_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_LIVELOG_NAME_TASKID})\n[${CHECKLOGS_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_LIVELOG_NAME_TASKID}/runs/0/logs/live/apple/banana.log)\n\n---\n\n\`\`\`bash\n${LIVE_LOG_TEXT}\n\`\`\`\n`
+        `### Task Links\n[${CHECKRUN_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_LIVELOG_NAME_TASKID}) | [${CHECKLOGS_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_LIVELOG_NAME_TASKID}/runs/0/logs/live/apple/banana.log) | [${CHECK_TASK_GROUP_TEXT}](${libUrls.testRootUrl()}/tasks/groups/${TASKGROUPID})\n### Task Status\nStarted: ${STARTED}\nResolved: ${RESOLVED}\nTask Execution Time: 1 day\nTask Status: completed\nReason Resolved: completed\nRunId: 0\n### Artifacts\n${buildArtifactLinks(50, CUSTOM_LIVELOG_NAME_TASKID)}\n\n---\n\n\`\`\`bash\n${LIVE_LOG_TEXT}\n\`\`\`\n`
```
lotas, line 1642:
```suggestion
        `[${CHECKRUN_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_LIVELOG_NAME_TASKID}) | [${CHECKLOGS_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_LIVELOG_NAME_TASKID}/runs/0/logs/live/apple/banana.log) | [${CHECK_TASK_GROUP_TEXT}](${libUrls.testRootUrl()}/tasks/groups/${TASKGROUPID})\n### Task Status\nStarted: ${STARTED}\nResolved: ${RESOLVED}\nTask Execution Time: 1 day\nTask Status: completed\nReason Resolved: completed\nRunId: 0\n### Artifacts\n${buildArtifactLinks(50, CUSTOM_LIVELOG_NAME_TASKID)}\n\n---\n\n\`\`\`bash\n${LIVE_LOG_TEXT}\n\`\`\`\n`
```

services/github/test/handler_test.js
```diff
@@ -1723,6 +1758,39 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
...
+        started: undefined,
+        resolved: undefined,
+      });
+
+      assert(github.inst(9988).checks.update.calledOnce, 'checks.update was not called');
+      let [args] = github.inst(9988).checks.update.firstCall.args;
+      /* eslint-disable comma-dangle */
+      assert.strictEqual(
+        args.output.text,
+        `### Task Links\n[${CHECKRUN_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_LIVELOG_NAME_TASKID}) | [${CHECKLOGS_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_LIVELOG_NAME_TASKID}/runs/0/logs/live/apple/banana.log) | [${CHECK_TASK_GROUP_TEXT}](${libUrls.testRootUrl()}/tasks/groups/${TASKGROUPID})\n### Task Status\nStarted: n/a\nResolved: n/a\nTask Execution Time: n/a\nTask Status: completed\nReason Resolved: completed\nRunId: 0\n### Artifacts\n${buildArtifactLinks(50, CUSTOM_LIVELOG_NAME_TASKID)}\n\n---\n\n\`\`\`bash\n${LIVE_LOG_TEXT}\n\`\`\`\n`
```
lotas, line 1789:
```suggestion
        `[${CHECKRUN_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_LIVELOG_NAME_TASKID}) | [${CHECKLOGS_TEXT}](${libUrls.testRootUrl()}/tasks/${CUSTOM_LIVELOG_NAME_TASKID}/runs/0/logs/live/apple/banana.log) | [${CHECK_TASK_GROUP_TEXT}](${libUrls.testRootUrl()}/tasks/groups/${TASKGROUPID})\n### Task Status\nStarted: n/a\nResolved: n/a\nTask Execution Time: n/a\nTask Status: completed\nReason Resolved: completed\nRunId: 0\n### Artifacts\n${buildArtifactLinks(50, CUSTOM_LIVELOG_NAME_TASKID)}\n\n---\n\n\`\`\`bash\n${LIVE_LOG_TEXT}\n\`\`\`\n`
```

services/github/src/utils.js
thakkarujas5: Just realized one thing, even if the whole logString fits within the payload, we only want to return the amount of log with the specified tail lines right, so the tailLog has been calculated before this condition.

services/github/src/handlers/status.js
```diff
@@ -176,12 +188,44 @@ export async function statusHandler(message) {
...
+    );
+    const CHECK_TASK_GROUP_TEXT_OUTPUT = markdownAnchor(
+      CHECK_TASK_GROUP_TEXT,
+      taskGroupUI(
+        this.context.cfg.taskcluster.rootUrl,
+        taskGroupId,
+      ),
+    );
+
+    output.addText(`### Task Links`);
```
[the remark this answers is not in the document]
thakkarujas5: Resolved: removed task link from header

services/github/test/handler_test.js
thakkarujas5: Resolved: Removed task links from tests

services/github/test/handler_test.js
thakkarujas5: Resolved: Removed task links from tests

services/github/test/handler_test.js
thakkarujas5: Resolved: Removed task links from tests

services/github/test/handler_test.js
thakkarujas5: Resolved: Removed task links from tests

services/github/test/handler_test.js
lotas: thanks

services/github/src/utils.js
lotas: Nice catch, indeed!

thakkarujas5: Have not modified the code for this.

lotas (approved):
Thank you!

Issue #7012, opened 2024-05-14
could not determine interactive username: number of gnome session users is not exactly one


bhearsum:
Noticed this error showing up in Firefox CI:
```
could not determine interactive username: number of gnome session users is not exactly one - not sure which user is interactively logged on: []string{""}
```

```
{
  "GOOS": "linux",
  "GOARCH": "amd64",
  "engine": "multiuser",
  "rootURL": "https://firefox-ci-tc.services.mozilla.com",
  "workerId": "9144889771845100298",
  "gwVersion": "64.2.0",
  "gwRevision": "9a9e590f97a3cab828ec1ea933610ce71d334c11",
  "workerType": "b-linux-large-gcp-d2g",
  "workerGroup": "us-central1-c",
  "deploymentId": "",
  "instanceType": "projects/887720501152/machineTypes/n2-highmem-32",
  "provisionerId": "translations-1",
  "cleanUpTaskDirs": "true"
}
```

It's not super frequent, but also not a one off. Seen in pools such as https://firefox-ci-tc.services.mozilla.com/worker-manager/translations-1%2Fb-linux-large-gcp-d2g/errors.

PR #7214, opened 2024-08-27
fix(generic-worker): add retries to interactive username logic

matt-boris:
Potentially fixes/helps https://github.com/taskcluster/taskcluster/issues/7012 and https://github.com/taskcluster/community-tc-config/issues/784.

>Generic Worker: Add retries around the logic that gets the current user who is logged in the gnome3 desktop session, in case race conditions are causing the following issue: `could not determine interactive username: number of gnome session users is not exactly one - not sure which user is interactively logged on: []string{""}`.

### DIFF
--- a/changelog/issue-7012.md
+++ b/changelog/issue-7012.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 7012
+---
+Generic Worker: Add retries around the logic that gets the current user who is logged in the gnome3 desktop session, in case race conditions are causing the following issue: `could not determine interactive username: number of gnome session users is not exactly one - not sure which user is interactively logged on: []string{""}`.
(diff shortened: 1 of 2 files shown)


petemoore:
Thanks for this Boris!

I think this might not be ideal, since we already have retries outside of the InteractiveUsername() method where we call it: https://github.com/taskcluster/taskcluster/blob/847b9f3b388f8a6401def4509977d00506151b2a/workers/generic-worker/runtime/runtime_other.go#L43-L51

The only other place we call InteractiveUsername() is on line 63 here:
https://github.com/taskcluster/taskcluster/blob/847b9f3b388f8a6401def4509977d00506151b2a/workers/generic-worker/multiuser.go#L59-L66
but as you see, line 59 has already called it indirectly. Therefore I think the best fix is just to avoid the additional call, and return it in `WaitForLoginCompletion`. I've made a fix for this in #7215.

PR #7215, opened 2024-08-28
Issue 7012: Generic Worker: cache interactive username

### LINKED ISSUE #7012: could not determine interactive username: number of gnome session users is not exactly one

petemoore:
Fixes #7012 (hopefully).

This should hopefully reduce intermittent failures.

In `multiuser.go` we had the following code:

https://github.com/taskcluster/taskcluster/blob/847b9f3b388f8a6401def4509977d00506151b2a/workers/generic-worker/multiuser.go#L59-L66

Line 59 was calling `gwruntime.WaitForLoginCompletion(5 * time.Minute)` which in turn was calling `gwruntime.InteractiveUsername()` indirectly with retries. In issue #7012 this was eventually passing, otherwise we would have seen this error:

https://github.com/taskcluster/taskcluster/blob/847b9f3b388f8a6401def4509977d00506151b2a/workers/generic-worker/runtime/runtime_other.go#L53

Therefore the error from #7012 appears to be from line 63 which happens after the interactive username had already been established successfully inside of the call on line 59. Therefore, for some reason, on Linux, that Generic Worker's method for determining the Gnome Desktop interactive username can intermittently fail, even if a previous call works.

Since the interactive user does not change during the lifetime of the Generic Worker process (only on reboot) it is safe to cache the value. This should reduce the intermittent failures such as the one in #7012. The solution is to surface the interactive username that is determined inside `gwruntime.WaitForLoginCompletion` by adding it to the returned values, in order that it can be used by the caller, without needing to make an additional call to determine its value.


matt-boris (approved):
Nice, thanks Pete 👍🏻

PR #7219, opened 2024-08-30
Generic Worker waits for expected task user to login rather than any user

### LINKED ISSUE #7218: Generic Worker Linux Multiuser: interactive username gdm does not match task user task_XXXXXXXXXXXXXXX
We are still hitting this issue even after #7215.

It happened after e.g. 7 successful tasks on https://community-tc.services.mozilla.com/provisioners/proj-mozci/worker-types/compute-smaller/workers/us-east4-c/974984873804309607?sortBy=started&sortDirection=desc.

Here is the sentry report for that worker: https://mozilla.sentry.io/issues/3330911982/events/af6ae647bcbc4e3e803dbadf6b1112ca/.

The code that determines the logged in gdm user is here:

https://github.com/taskcluster/taskcluster/blob/542c891ebddb13c9c0415193dc477947de23d8d8/workers/generic-worker/gdm3/gdm3_nonfreebsd.go#L12-L25

For some reason this is returning `gdm` as the user rather than the task user that was meant to log in. Currently, we call a function to determine which user is actually logged in to the desktop session, and compare it to the expected task user. If the two don't match, we panic. However, another option would be to pass the expected user to the function that waits for the login, so that we keep running that nasty bash excerpt above to test if our user is logged in, until it returns the required task user, or times out. That way, if for a short period the wrong user is logged in, it won't matter. If this doesn't resolve the issue, we should revert, since we will be introducing 5 minute penalty for every time the wrong user is logged in, which could be expensive. However, hopefully it is just a race condition that will be solved.

petemoore:
Fixes #7218

### DIFF
--- a/changelog/issue-7218.md
+++ b/changelog/issue-7218.md
@@ -1,0 +1,19 @@
+audience: worker-deployers
+level: patch
+reference: issue 7218
+---
+Generic Worker Multiuser engine on Linux, macOS and FreeBSD now waits for the
+required task user to be logged in to the console session, rather than waiting
+for any user to be logged in, and then checking whether it is the anticipated
+user. This subtle change in behaviour means that temporarily a different user
+may be (or appear to be) logged into the console session without causing
+Generic Worker to panic. It is hoped that this will reduce intermittent issues
+where a different user appears to be logged in (such as gdm user on Linux)
+since it is suspected that this might just be a fleeting login that passes due
+to some race condition in the start up of the Gnome Desktop.
+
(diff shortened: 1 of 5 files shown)


lotas (approved):
let's see 😎

matt-boris (approved):
👍🏻 🤞🏻 

Something looks up with the CI. I wonder if it's because you initially pushed to main the force pushed it back off. Could you push a new commit? Or even merge your trivial build constraint fix into this?

matt-boris (approved):

PR #7236, opened 2024-09-04
chore(ci): use taskgraph to build/manage docker images

matt-boris:
This feature makes it so the CI docker images are only built on changes to the `Dockerfile`. Taskgraph will automatically assign task dependencies on the build docker image tasks so that tasks that rely on this image won't run until the new image is built!

Followed recommendations from https://taskcluster-taskgraph.readthedocs.io/en/latest/howto/docker.html#using-in-project-images as well as examples from https://github.com/taskcluster/taskgraph.


taskcluster/config.yml
```diff
@@ -3,6 +3,10 @@ trust-domain: taskcluster
 # this is lower priority than the decision task
 task-priority: high
 
+treeherder:
+  group-names:
+    'I': 'Docker Image Builds'
```
petemoore, line 8: is this needed if we don't have a treeherder integration?

taskcluster/kinds/client/kind.yml
```diff
@@ -21,7 +19,10 @@ task-defaults:
 tasks:
   nodejs:
     description: nodejs client tests
-    docker-image: node:{node_version}
+    worker:
+      docker-image: node:20.16.0
```
petemoore, line 23: do we need a way to bump this version number when we update our node version?

taskcluster/kinds/db/kind.yml
```diff
@@ -26,6 +27,9 @@ tasks:
         cd db && yarn test
   upgrade:
     description: yarn db:upgrade tests
+    worker:
+      env:
+        COREPACK_ENABLE_DOWNLOAD_PROMPT: "0"
```
petemoore, line 32: i'm curious what the COREPACK_ENABLE_DOWNLOAD_PROMPT does?

taskcluster/kinds/docker-image/kind.yml
```diff
@@ -0,0 +1,14 @@
+loader: taskgraph.loader.transform:loader
+
+transforms:
+  - taskgraph.transforms.docker_image:transforms
+  - taskgraph.transforms.cached_tasks:transforms
+  - taskgraph.transforms.task:transforms
+
+tasks:
+  browser-test:
+    symbol: I(browser-test)
```
petemoore, line 10: is this treeherder config? do we need that?

petemoore (commented):
This will be so awesome when it is finished! Mostly just questions for now. Thanks Boris!

taskcluster/config.yml
matt-boris: Yes it is. I want to talk to RelEng about this alongside some other changes I'd like to suggest after implementing these changes.

taskcluster/kinds/client/kind.yml
matt-boris: Yep! I'm going to work on the generators after the main hunk of work is complete so that these will get updated with `yarn generate`.

taskcluster/kinds/db/kind.yml
matt-boris: Messed around with some env var settings to have `corepack` not ask me if I want to enable it and install anything on the system. I think I've worked around having to use this anyway...

taskcluster/kinds/docker-image/kind.yml
matt-boris: Again, yeah this may tie into treeherder (with the symbol?) but this is what's required on Taskgraph's side to get the images built. https://taskcluster-taskgraph.readthedocs.io/en/latest/howto/docker.html#adding-image-builder-tasks

lotas (approved):
great job!

PR #7247, opened 2024-09-10
Headless Tasks

### LINKED ISSUE #4595: [generic-worker] Support headless
See
* https://bugzilla.mozilla.org/show_bug.cgi?id=1588799
* https://bugzilla.mozilla.org/show_bug.cgi?id=1609102

petemoore:
Fixes #4595


workers/generic-worker/process/multiuser_posix.go
```diff
@@ -11,23 +11,23 @@ import (
...
 		return &PlatformData{}, nil
 	}
-	return TaskUserPlatformData()
+	return TaskUserPlatformData(user, headlessTasks)
 }
 
-func TaskUserPlatformData() (pd *PlatformData, err error) {
-	username, err := runtime.InteractiveUsername()
+func TaskUserPlatformData(u *gwruntime.OSUser, headlessTasks bool) (pd *PlatformData, err error) {
+	username, err := gwruntime.InteractiveUsername()
```
Eijebong, line 30: This is still depending on gdm as it ends up trying to get the username with 

`gnomeSessionUserList, err := host.CombinedOutput("/usr/bin/env", "bash", "-c", "PROCPS_USERLEN=20 /usr/bin/w | /bin/grep gnome-[s]ession | /usr/bin/cut -f1 -d' '")`

What I did on my attempt to solve this issue was to copy the code from `gdm3/gdm3_freebsd.go` and ended up with something like that:

```go
func TaskUserPlatformData(config *gwconfig.Config) (pd *PlatformData, err error) {
    var user string
    if !config.DisableGui {
        user, err = runtime.InteractiveUsername()
        if err != nil {
            return nil, fmt.Errorf("could not determine interactive username: %v", err)
        }
    } else {
        // TODO: Dedupe this block, it's also in gdm3/freebsd
        credsFile, err := os.Open("current-task-user.json")
        if err != nil {
            return nil, err
        }
        defer func() {
            credsFile.Close()
        }()
        decoder := json.NewDecoder(credsFile)
        decoder.DisallowUnknownFields()
        decodedUser := map[string]string{}
        err = decoder.Decode(&decodedUser)
        if err != nil {
            return nil, err
        }
        user = decodedUser["name"]
    }
```

petemoore: Ah whoops, thanks for that! I think there is a simple fix:

```
diff --git a/workers/generic-worker/process/multiuser_posix.go b/workers/generic-worker/process/multiuser_posix.go
index a95d67f9b..054f30a4d 100644
--- a/workers/generic-worker/process/multiuser_posix.go
+++ b/workers/generic-worker/process/multiuser_posix.go
@@ -27,12 +27,7 @@ func NewPlatformData(currentUser bool, headlessTasks bool, user *gwruntime.OSUse
 }
 
 func TaskUserPlatformData(u *gwruntime.OSUser, headlessTasks bool) (pd *PlatformData, err error) {
-       username, err := gwruntime.InteractiveUsername()
-       if err != nil {
-               return nil, fmt.Errorf("could not determine interactive username: %w", err)
-       }
-
-       usr, err := user.Lookup(username)
+       usr, err := user.Lookup(u.Name)
        if err != nil {
                return nil, fmt.Errorf("failed to lookup user: %w", err)
        }
```

changelog/issue-4595.md
```diff
@@ -0,0 +1,16 @@
...
+reference: issue 4595
+---
+Generic Worker can now be run in headless mode, meaning tasks do not have a
+dedicated graphical user environment. To do this, the Generic Worker config
+setting `headlessTasks` should be set to true. This can only be enabled or
+disabled at the Worker level, tasks cannot choose if they run in a headless
+environment or not, it depends on the worker settings (i.e. the Worker Pool
+configuration).
+
+There are no reboots in headless mode, and multiple worker instances can
```
lotas, line 12: 🎉  awesome

workers/generic-worker/multiuser.go
```diff
@@ -192,7 +207,7 @@ func deleteExistingOSUsers() (err error) {
 	}
 	allErrors := []string{}
 	for _, username := range userAccounts {
-		if strings.HasPrefix(username, "task_") && username != taskContext.User.Name && username != gwruntime.AutoLogonUser() {
+		if strings.HasPrefix(username, "task_") && username != taskContext.User.Name && username != nextTaskUser {
```
lotas, line 210: probably a silly question - how does it know which task is next? if you have capacity>1, wouldn't there be multiple next users? 
Or this case is not handled yet (or handled in a different place)?

lotas (dismissed):
👏

workers/generic-worker/multiuser.go
petemoore: That isn't a silly question, that is a good one. Under Generic Worker, you need one generic-worker process per unit capacity, so if you want to run 5 tasks in parallel, you would have 5 generic-worker processes running. The `nextTaskUser` is created ahead of time, so at this point in the code, the value has already been initialised.

lotas: ah, got it!

workers/generic-worker/process/multiuser_windows.go
```diff
@@ -172,11 +180,11 @@ func GrantSIDWinstaAccess(sid string, pd *PlatformData) {
 		log.Printf("SID %v NOT found in %#v - granting access...", sid, sidsThatCanControlDesktopAndWindowsStation)
 
 		cmd, err := NewCommand([]string{gwruntime.GenericWorkerBinary(), "grant-winsta-access", "--sid", sid}, ".", []string{}, pd)
-		cmd.DirectOutput(os.Stdout)
-		log.Printf("About to run command: %#v", *(cmd.Cmd))
 		if err != nil {
 			panic(err)
 		}
+		cmd.DirectOutput(os.Stdout)
+		log.Printf("About to run command: %#v", *(cmd.Cmd))
```
matt-boris, line 187: Nice, thanks for moving these down. Helps with readability with that error handling 👍🏻

workers/generic-worker/multiuser_windows_test.go
```diff
@@ -105,24 +106,39 @@ func TestDesktopResizeAndMovePointer(t *testing.T) {
...
-		Command:    commands,
-		MaxRunTime: 90,
-		// Don't assume python 2 is in the default system PATH, but rather
-		// require that python 2 is in the PATH of the test process.
-		Env: map[string]string{
-			"PATH": os.Getenv("PATH"),
-		},
+
+	// We run the same test under both headless and non-headless mode, but
+	// expect different results. So pull it out into its own function...
```
matt-boris, line 111: Nifty 🎉

workers/generic-worker/helper_multiuser_test.go
```diff
@@ -34,10 +36,24 @@ func expectChainOfTrustKeyNotSecureMessage(t *testing.T, td *tcqueue.TaskDefinit
 	expectedArtifacts.Validate(t, taskID, 0)
 }
 
-func newPlatformData(conf *gwconfig.Config) (pd *process.PlatformData) {
-	pd, err := process.NewPlatformData(conf.RunTasksAsCurrentUser)
+func engineTestSetup(t *testing.T, testConfig *gwconfig.Config) {
+	t.Helper()
+	// this should be fun
```
matt-boris, line 41: Wdyt?

```suggestion
```

changelog/issue-4595.md
```diff
@@ -0,0 +1,16 @@
+audience: users
+level: minor
+reference: issue 4595
+---
+Generic Worker can now be run in headless mode, meaning tasks do not have a
+dedicated graphical user environment. To do this, the Generic Worker config
+setting `headlessTasks` should be set to true. This can only be enabled or
```
matt-boris, line 7: There's part of me that's thinking this config option should just be `headless`. But I'm not fully convinced one way or the other, just wanted to toss it out there and hear why you thought to use `headlessTasks`?

matt-boris (commented):
Awesome stuff here Pete, thank you!

petemoore:
I'm afraid I discovered I still need to tweak the handling of current task user / next task user. Since with native GUI logins you need to reboot once before you've even claimed a task, just removing the reboots means a slightly different code path is followed. So I'll need to fix this.

Currently we have something like this...

## Non headless environment (current implementation)

* reboot = true
* if next-task-user.json file exists
  * copy next-task-user.json to current-task-user.json
  * reboot = false
* so long as there isn't precisely one remaining task to run (`numberOfTasksToRun` == 1) create next task user account on the system, the next-task-user.json file, configure auto login on the system

However, for headless environment, the above implementation cannot easily be used, even with `if config.HeadlessTasks` { ... }` wrappers around parts of it, or if it is reused, it gets very messy very quickly.

Instead I'll look at having a headless version which is much simpler:

## Headless environment

* create current task user (no reason to write `.json` files to disk here, as nothing needs to be persisted across reboots)

Note, also no "next task user" is needed, as that is something to protect against tasks that reboot the system or crash etc, so that after an unplanned reboot, we still have a fresh task user prepared. If the headless version suffers an unanticipated reboot during task execution, it already creates a task user immediately before using it, so no need for any such preparation or persistence of state.

changelog/issue-4595.md
petemoore: My thinking was the config settings relates to the execution of the tasks, rather than the execution of the worker (e.g. other config settings manage how the worker behaves, rather than how the tasks behave).

lotas (approved):
LGTM!

Issue #4987, opened 2021-09-02
Combine worker-manager provisioner and worker scanner in to a single processes


jwhitlock:
The ``worker-manager`` has two processes that could be combined:

* ``worker-manager-provisioner`` - Looks at the pending tasks in the ``queue``, and spins up workers (and maybe spins down)
*  ``worker-manager-workerscanner`` - Checks if provisioned workers are still running. Workers are often configured to shut themselves down if they have no tasks to claim.

According to @imbstack, this design was informed by a previous AWS-only provisioner, which was too slow and would get itself stuck. The new provisioner is designed to launch many workers in parallel, as quickly as possible, and not get stuck.  However, this seems to lead to over-provisioning.

His suggestion was to combine the two into a single process that runs the three steps repeatedly:

* Scan the active workers to update state
* Check the queues
* Start an appropriate amount of workers

This would ensure that scanning and provisioning workers remain in sync.

lotas:
From FxCI logs I can see that:

* provisioner is currently quite fast, seconds to minutes (worst cases)
* scanner is extremely slow on FxCI - 10min to 30min and above

If we combine those now, it would mean that provisioning part of it might get delayed, and new task would be waiting for the scanner part for too long.  

Plus there is a huge disproportion in providers, for example at this very moment fxci runs:

| provider | running tasks | % |
| :---: | :---: | :---: |
| aws | 4250 | ~84% |
| azure | 696 | ~13% |
| gcp | 108 | ~2% |

Those values will probably change, but still, some providers will run longer. 

My suggestions:

**Split processes** 
Run separate scanner/provisioner for each cloud (providers would be isolated and would not block each other)

**Optimize scanner** 
Minimize scanning time. 
For example call `list` endpoints to enumerate all running instances and avoid doing multiple calls
    - GCP [instances/list](https://cloud.google.com/compute/docs/reference/rest/v1/instances/list)
    - Azure [list-all](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/list-all#virtualmachine)
    - AWS [DescribeInstances](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html)

 And then to find instances that were missing in api response but are considered as running according to database, and additionally query those, or set to terminated

lotas:
Couple of interesting observations came out when looking at one of the slow loops:

* we didn't hit rate limiting in cloud APIs in recent weeks, which means that we are not being slowed down by cloud provider
[...]

PR #7258, opened 2024-09-13
Allow provision of public IP for g-w in Azure

### LINKED ISSUE #7257: Worker manager should provide an option to get public IP with Azure + g-w
The Azure provider currently skips public IP allocation if the worker type is generic-worker, since it doesn't need it. Public IP can still be helpful for debugging, eg. getting RDP access.

There should be an option in the workerConfig to override the default behaviour and request a public IP in a pool even if the worker is g-w.

lotas:
in #4987 it was decided to skip public IP provisioning for generic-worker since it wasn't used
However, having public IP might still be useful for debug purposes. Worker pool configuration allows `workerManager.publicIP=true` to override that logic

Fixes #7257

### DIFF
--- a/services/worker-manager/schemas/v1/config-azure.yml
+++ b/services/worker-manager/schemas/v1/config-azure.yml
@@ -46,6 +46,22 @@ properties:
           type: integer
           minimum: 1
           description: The number of tasks a single worker of this type can run at any given time.
+        workerManager:
+          title: Worker Manager Config
+          type: object
+          description: |-
+            Worker Manager's own configuration settings per launch configuration.
+            The other properties of the launch configuration section are passed directly through to
+            Azure APIs, whereas this section is interpreted directly by Worker Manager.
+          properties:
+            publicIp:
+              title: Public IP
+              type: boolean
+              description: |
+                If `true`, Worker Manager will request a public IPv4 address for the worker.
+                This is set to `false` by default. It is typically only needed for Windows
+                workers that should allow incoming RDP connections from the public internet.
+          additionalProperties: false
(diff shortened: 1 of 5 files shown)


services/worker-manager/schemas/v1/config-azure.yml
```diff
@@ -46,6 +46,18 @@ properties:
...
+          title: Worker Manager Config
+          type: object
+          description: Configuration specific to the worker-manager and Azure
+          properties:
+            publicIp:
+              title: Public IP
+              type: boolean
+              description: |
+                If true, the worker-manager will request a public IP address for the worker.
+                This is set to `false` by default as it is not needed for most workers.
```
petemoore, line 59:
```suggestion:
                If `true`, Worker Manager will request a public IPv4 address for the worker.
                This is set to `false` by default. It is typically only needed for Windows
                workers that allow incoming RDP connections from the public internet.
```

services/worker-manager/src/providers/azure/index.js
```diff
@@ -911,7 +914,7 @@ export class AzureProvider extends Provider {
 
     let titleString = "";
 
-    // #4987: generic workers do not need Public IP,
+    // #4987: generic workers do not need Public IP unless explicitly requested #7257
```
petemoore, line 917:
```suggestion
    // #4987: workers do not need Public IP unless explicitly requested #7257
```

petemoore (approved):
Thanks Yarik!

petemoore (approved):
Looks awesome, thanks!

PR #7259, opened 2024-09-13
Update rust dependencies

Eijebong:
This is a "major" change because it requires a major version bump for
the rust client as some dependencies (`reqwest`) is exposed through the
public API.


lotas (approved):
Great! Many thanks

PR #7261, opened 2024-09-13
feat(github): allow fetching artifacts from queue

lotas:
For the #5669 we forgot to include necessary scopes

After this feature was merged https://github.com/taskcluster/taskcluster/pull/7208 we noticed we forgot to include one more scope 
https://github.com/taskcluster/tc-dev-integration-test/commit/643820fe0576909515c72f6c40d4990d36837812#comments

### DIFF
--- a/changelog/issue-5669-1.md
+++ b/changelog/issue-5669-1.md
@@ -1,0 +1,4 @@
+audience: developers
+level: silent
+reference: issue 5669
+---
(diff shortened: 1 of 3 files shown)


matt-boris (dismissed):

matt-boris (approved):

lotas:
actually, I'm not 100% convinced it should be static permission, or role for the repository.. so we could configure it on a per-project level?

PR #7262, opened 2024-09-13
fix(github) Fix artifact fetch due to permission issue

### LINKED ISSUE #5669: Enhance github integration: Include more useful information in check run output

thakkarujas5:
Fixes #5669 

Fix artifact url generation
Fix artifact fetch due to missing permission issues
Hide headers in checkrun for absent values

### DIFF
--- a/services/github/src/handlers/utils.js
+++ b/services/github/src/handlers/utils.js
@@ -235,6 +235,10 @@ export const buildUrl = (rootUrl, taskId, runId, artifactName) => {
   return `${rootUrl}/tasks/${taskId}/runs/${runId}/${artifactName}`;
 };
 
+export const buildLogUrl = (rootUrl, taskId, runId, artifactName) => {
+  return `${rootUrl}/tasks/${taskId}/runs/${runId}/logs/${artifactName}`;
+};
+
 export default {
   taskUI,
   taskGroupUI,
@@ -245,4 +249,5 @@ export default {
   isCollaborator,
   getTimeDifference,
   buildUrl,
+  buildLogUrl,
 };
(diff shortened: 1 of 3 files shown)


services/github/src/handlers/utils.js
```diff
@@ -232,7 +232,7 @@ export const getTimeDifference = (timestamp1, timestamp2) => {
 };
 
 export const buildUrl = (rootUrl, taskId, runId, artifactName) => {
-  return `${rootUrl}/tasks/${taskId}/runs/${runId}/${artifactName}`;
+  return `${rootUrl}/tasks/${taskId}/runs/${runId}/logs/${artifactName}`;
```
lotas, line 235: `/logs` was probably a special case of artifacts, sorry. We'd likely need to check the name also: https://github.com/taskcluster/taskcluster/blob/main/ui/src/views/Tasks/routes.js#L63-L71

And maybe this only works for the logs.. 
I think we can only link those two logs (log, live log) directly, and all other artifact links should just point to the task page (so url ends with runId) 
Thanks!

lotas (approved):
Thanks!

PR #7265, opened 2024-09-19
feat(d2g): add `containerEngine` to toggle between `docker`/`podman`

matt-boris:
>D2G: Adds `capabilities.containerEngine` to the Docker Worker payload schema strictly to use as a `docker`/`podman` toggle for the d2g-translated payload.

>Generic Worker: Adds `containerEngine` worker config option to select between `docker` and `podman` to be used during D2G payload translations.
>
>Default is `docker` and this value will be overridden by `task.payload.capabilities.containerEngine`, if specified.

>Generic Worker: Adds `enableD2G` worker config option to internally process Docker Worker payloads using D2G. Defaults to `false` and will return a `malformed-payload` if a Docker Worker payload is detected and this config isn't set to `true`.

### DIFF
--- a/changelog/DIlRBhW6S1aQ6Rl63gLxog.md
+++ b/changelog/DIlRBhW6S1aQ6Rl63gLxog.md
@@ -1,4 +1,4 @@
 audience: users
-level: major
+level: minor
 ---
-D2G: Adds `capabilities.containerEngine` to the Docker Worker payload schema strictly to use as a `docker`/`podman` toggle for the d2g-translated payload. Defaults to `docker`.
+D2G: Adds `capabilities.containerEngine` to the Docker Worker payload schema strictly to use as a `docker`/`podman` toggle for the d2g-translated payload.
--- a/changelog/DIlRBhW6S1aQ6Rl63gLxog.md
+++ b/changelog/DIlRBhW6S1aQ6Rl63gLxog.md
@@ -1,4 +1,4 @@
 audience: users
 level: major
 ---
-D2G: Adds `capabilities.containerEngine` to the Docker Worker payload schema strictly to use as a `docker`/`podman` toggle for the d2g-translated payload. Defaults to `podman`.
+D2G: Adds `capabilities.containerEngine` to the Docker Worker payload schema strictly to use as a `docker`/`podman` toggle for the d2g-translated payload. Defaults to `docker`.
--- a/changelog/DIlRBhW6S1aQ6Rl63gLxog.md
+++ b/changelog/DIlRBhW6S1aQ6Rl63gLxog.md
@@ -1,0 +1,4 @@
+audience: users
+level: major
+---
+D2G: Adds `capabilities.containerEngine` to the Docker Worker payload schema strictly to use as a `docker`/`podman` toggle for the d2g-translated payload. Defaults to `podman`.
--- a/clients/client-shell/cmds/d2g/d2g.go
+++ b/clients/client-shell/cmds/d2g/d2g.go
@@ -29,6 +29,7 @@ To convert a task definition (JSON), you must use the task definition flag (-t, 
   cat /path/to/input/task-definition.json | taskcluster d2g -t
   echo '{"image": "ubuntu", "command": ["bash", "-c", "echo hello world"], "maxRunTime": 300}' | taskcluster d2g`,
 	}
+	cmd.Flags().StringP("engine", "e", "docker", "The container engine to use between docker and podman.")
 	cmd.Flags().StringP("file", "f", "", "Path to a .json file containing a docker-worker payload or task definition.")
 	cmd.Flags().BoolP("task-def", "t", false, "Must use if the input is a docker-worker task definition.")
 	root.Command.AddCommand(cmd)
--- a/clients/client-shell/cmds/d2g/d2g.go
+++ b/clients/client-shell/cmds/d2g/d2g.go
@@ -23,7 +23,7 @@ func init() {
 		Short: `Converts a docker-worker payload (JSON) to a generic-worker payload (JSON).
 To convert a task definition (JSON), you must use the task definition flag (-t, --task-def).`,
 		RunE: convert,
-		Example: `  taskcluster d2g -f /path/to/input/payload.json
+		Example: `  taskcluster d2g -f -e podman /path/to/input/payload.json
   taskcluster d2g -t -f /path/to/input/task-definition.json
   cat /path/to/input/payload.json | taskcluster d2g
   cat /path/to/input/task-definition.json | taskcluster d2g -t
@@ -38,6 +38,7 @@ To convert a task definition (JSON), you must use the task definition flag (-t, 
 func convert(cmd *cobra.Command, args []string) (err error) {
 	isTaskDef, _ := cmd.Flags().GetBool("task-def")
 	filePath, _ := cmd.Flags().GetString("file")
+	engine, _ := cmd.Flags().GetString("engine")
 
 	input, err := userInput(filePath)
 	if err != nil {
@@ -84,7 +85,7 @@ func convert(cmd *cobra.Command, args []string) (err error) {
 		}
 
 		// Convert dwTaskDef to gwTaskDef
-		gwTaskDefJSON, err := d2g.ConvertTaskDefinition(dwTaskDefJSON)
+		gwTaskDefJSON, err := d2g.ConvertTaskDefinition(dwTaskDefJSON, engine)
 		if err != nil {
 			return fmt.Errorf("failed to convert docker worker task definition to a generic worker task definition: %v", err)
 		}
@@ -108,7 +109,7 @@ func convert(cmd *cobra.Command, args []string) (err error) {
 		fmt.Fprintln(cmd.OutOrStdout(), string(gwTaskDefJSON))
 	} else {
 		// Convert dwPayload to gwPayload
-		gwPayload, err := d2g.Convert(dwPayload)
+		gwPayload, err := d2g.Convert(dwPayload, engine)
 		if err != nil {
 			return fmt.Errorf("conversion error: %v", err)
 		}
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -98,7 +98,8 @@ func ConvertTaskDefinition(dwTaskDef json.RawMessage) (json.RawMessage, error) {
 // a converted Docker Worker task payload (see d2g.Convert function) to run
 // Docker Worker tasks under Generic Worker.
 func Scopes(dwScopes []string, dwPayload *dockerworker.DockerWorkerPayload, taskQueueID string) (gwScopes []string) {
-	gwScopes = []string{}
+	// scopes to use docker, by default, should just come "for free"
+	gwScopes = []string{"generic-worker:os-group:" + taskQueueID + "/docker"}
 	for _, s := range dwScopes {
 		switch true {
 		case s == "docker-worker:capability:device:kvm":
@@ -106,11 +107,6 @@ func Scopes(dwScopes []string, dwPayload *dockerworker.DockerWorkerPayload, task
 				gwScopes,
 				"generic-worker:os-group:"+taskQueueID+"/kvm",
 				"generic-worker:os-group:"+taskQueueID+"/libvirt",
-			)
-		case s == "docker-worker:capability:privileged":
-			gwScopes = append(
-				gwScopes,
-				"generic-worker:os-group:"+taskQueueID+"/docker",
 			)
 		case s == "docker-worker:capability:device:loopbackVideo":
 			gwScopes = append(gwScopes, "generic-worker:loopback-video:*")
@@ -136,7 +132,7 @@ func Scopes(dwScopes []string, dwPayload *dockerworker.DockerWorkerPayload, task
 
 // Convert transforms a Docker Worker task payload into an equivalent Generic
 // Worker Multiuser POSIX task payload. The resulting Generic Worker payload is
-// a BASH script which uses Podman to contain the Docker Worker payload. Since
+// a BASH script which uses Podman (by default) to contain the Docker Worker payload. Since
 // scopes fall outside of the payload in a task definition, scopes need to be
 // converted separately (see d2g.Scopes function).
 func Convert(dwPayload *dockerworker.DockerWorkerPayload) (gwPayload *genericworker.GenericWorkerPayload, err error) {
@@ -239,10 +235,7 @@ func artifacts(dwPayload *dockerworker.DockerWorkerPayload) []genericworker.Arti
 }
 
 func command(dwPayload *dockerworker.DockerWorkerPayload, dwImage Image, gwArtifacts []genericworker.Artifact, gwWritableDirectoryCaches []genericworker.WritableDirectoryCache) ([][]string, error) {
-	tool := "podman"
-	if dwPayload.Capabilities.Privileged {
-		tool = "docker"
-	}
+	tool := dwPayload.Capabilities.ContainerEngine
 	containerName := ""
 	if len(gwArtifacts) > 0 {
 		containerName = "taskcontainer"
@@ -467,7 +460,7 @@ func setOSGroups(dwPayload *dockerworker.DockerWorkerPayload, gwPayload *generic
 		// https://help.ubuntu.com/community/KVM/Installation
 		gwPayload.OSGroups = append(gwPayload.OSGroups, "kvm", "libvirt")
 	}
-	if dwPayload.Capabilities.Privileged {
+	if dwPayload.Capabilities.ContainerEngine == "docker" {
 		gwPayload.OSGroups = append(gwPayload.OSGroups, "docker")
 	}
 }
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -30,7 +30,7 @@ type (
 	}
 )
 
-func ConvertTaskDefinition(dwTaskDef json.RawMessage) (json.RawMessage, error) {
+func ConvertTaskDefinition(dwTaskDef json.RawMessage, containerEngine string) (json.RawMessage, error) {
 	var gwTaskDef json.RawMessage
 	var parsedTaskDef map[string]interface{}
 	err := json.Unmarshal(dwTaskDef, &parsedTaskDef)
@@ -53,7 +53,7 @@ func ConvertTaskDefinition(dwTaskDef json.RawMessage) (json.RawMessage, error) {
 		return nil, fmt.Errorf("cannot unmarshal Docker Worker payload: %v", err)
 	}
 
-	gwPayload, err := Convert(dwPayload)
+	gwPayload, err := Convert(dwPayload, containerEngine)
 	if err != nil {
 		return nil, fmt.Errorf("cannot convert Docker Worker payload: %v", err)
 	}
@@ -75,7 +75,7 @@ func ConvertTaskDefinition(dwTaskDef json.RawMessage) (json.RawMessage, error) {
 		if taskQueueID == "" {
 			return nil, fmt.Errorf("taskQueueId ('provisionerId/workerType') is required")
 		}
-		parsedTaskDef["scopes"] = Scopes(dwScopes, dwPayload, taskQueueID)
+		parsedTaskDef["scopes"] = Scopes(dwScopes, dwPayload, taskQueueID, containerEngine)
 	}
 
 	d2gConvertedPayloadJSON, err := json.Marshal(*gwPayload)
@@ -97,9 +97,10 @@ func ConvertTaskDefinition(dwTaskDef json.RawMessage) (json.RawMessage, error) {
 // equivalent Generic Worker scopes. These scopes should be used together with
 // a converted Docker Worker task payload (see d2g.Convert function) to run
 // Docker Worker tasks under Generic Worker.
-func Scopes(dwScopes []string, dwPayload *dockerworker.DockerWorkerPayload, taskQueueID string) (gwScopes []string) {
+func Scopes(dwScopes []string, dwPayload *dockerworker.DockerWorkerPayload, taskQueueID, containerEngine string) (gwScopes []string) {
+	tool := getContainerEngine(dwPayload, containerEngine)
 	gwScopes = []string{}
-	if dwPayload.Capabilities.ContainerEngine == "docker" {
+	if tool == "docker" {
 		// scopes to use docker, by default, should just come "for free"
 		gwScopes = append(gwScopes, "generic-worker:os-group:"+taskQueueID+"/docker")
 	}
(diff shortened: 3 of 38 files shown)


workers/docker-worker/schemas/v1/payload.yml
```diff
@@ -123,6 +123,17 @@ properties:
...
+        title: Container engine
+        description: >-
+          The container engine to use, between `docker` and `podman`.
+          This property is only used during Generic Worker's internal
+          payload translation (d2g). Default is `podman`.
+        type: string
+        enum:
+          - docker
+          - podman
+        default: podman
```
lotas, line 136: Tough choice probably as there are already so many pools on community-tc converted to d2g, but having major fxci upgrade ahead maybe it's worth making `docker` default? 
Although it might require changes on the taskgraph side as well, so doesn't really matter?

matt-boris: Maybe we can tie @taskcluster/releng in here and get their opinion.

matt-boris: Sounds like having a worker config setting is preferred. I'll keep this payload change as well for quick toggling in individual tasks.

tools/d2g/d2gtest/testdata/testcases/container_engine_test.yml
```diff
@@ -0,0 +1,94 @@
...
+              -e TASK_GROUP_ID
+              -e TASK_ID
+              ubuntu 'echo "Hello world"'
+        maxRunTime: 3600
+        onExitStatus:
+          retry:
+            - 125
+            - 128
+
+    - name: Default container engine test with privileged set
```
petemoore, line 33: in addition, one more test where privileged is set but containerEngine isn't set? or do we have that elsewhere? maybe not bad to have it here too, then we have covered all possibilities here

petemoore (approved):

workers/docker-worker/schemas/v1/payload.yml
bhearsum: It sounds like maybe you've moved past this already, but fwiw, I don't think either default would be a big deal. Any migrations to g-w/d2g that we do will involve a bunch of testing on Try or in PRs, and we can fairly easily set this to whatever when needed.

petemoore:
Before we add config options here, let's discuss with RelEng again. My gut feeling is that this isn't the best way to solve the problem, and could lead to more unexpected behaviour and errors due to worker pools being configured differently, than problems it solves. This is compounded by the fact that worker config is typically located in many places and isn't centrally managed. Some worker pools may be manually configured, others managed in puppet, other managed in nested puppet modules and other config managed by tools like tc-admin / ci-config etc, or even in source code in fuzzing repositories like orion.

My feeling is that d2g should emulate what Docker Worker did. Docker Worker ran _everything_ under docker. So I think the better approach here might be for d2g to use docker exclusively, and if someone wants to experiment seeing if their task that runs fine under docker might also run fine under podman, they can update their task to use the Generic Worker native payload that d2g generates for them, and then change it to use podman instead of docker...

But my opinion isn't the only important one, so let's discuss this with others before we decide for sure.

jcristau:
From a fxci perspective, I think having a way to select between podman and docker is valuable:
- longer-term, podman seems like a better/safer choice, with the absence of a privileged daemon
- short/medium term, if we can avoid tying the migration away from docker-worker to the change of tool, we're removing an arguably unnecessary road block
- to Pete's point, all our docker-worker pools (and thus all those where we'll eventually want to be using d2g) are centrally managed AFAIK
- "just use d2g locally and submit your task using a native generic-worker payload" is not something we necessarily want to support, and in any case feels like more trouble (users don't directly call createTask in our environment, everything goes through a vcs push).

As to the mechanism, I have a slight preference for being able to configure this per pool, so that e.g. we can eventually enforce that all builds are on podman, while tests could use either depending on the permissions they need and how far we are in the migration.  It also makes it clearer what needs to be installed on the worker image, and if/when we're ready to get rid of docker.  But we can probably make it work either way.

On the client / taskgraph level, I'm not sure it makes much difference, it'd probably be either setting `containerEngine:docker` in the payload or setting `worker-type: t-whatever-docker`?  As a bonus, the latter would be easy to experiment with using `mach try --worker-override`, without making any code changes.

tools/d2g/d2g.go
```diff
@@ -238,9 +238,11 @@ func artifacts(dwPayload *dockerworker.DockerWorkerPayload) []genericworker.Arti
...
 
-func command(dwPayload *dockerworker.DockerWorkerPayload, dwImage Image, gwArtifacts []genericworker.Artifact, gwWritableDirectoryCaches []genericworker.WritableDirectoryCache) ([][]string, error) {
-	tool := "podman"
-	if dwPayload.Capabilities.Privileged {
+func command(dwPayload *dockerworker.DockerWorkerPayload, dwImage Image, containerEngine string, gwArtifacts []genericworker.Artifact, gwWritableDirectoryCaches []genericworker.WritableDirectoryCache) ([][]string, error) {
+	tool := containerEngine
+	if payloadDefinedContainerEngine(dwPayload) {
+		tool = dwPayload.Capabilities.ContainerEngine
+	} else if dwPayload.Capabilities.Privileged {
 		tool = "docker"
```
petemoore, line 246: I think we should either:
* have a separate worker config for default privileged container engine and default unprivileged container engine to use, or
* we shouldn't switch based on whether the payload is privileged at all (e.g. maybe it is desirable to run `podman --privileged` and using docker always isn't what we want)

For this reason, my instinct says it is better to lose the worker pool setting, keep the task payload setting, and make docker the default, not podman.

petemoore:
> As to the mechanism, I have a slight preference for being able to configure this per pool, so that e.g. we can eventually enforce that all builds are on podman, while tests could use either depending on the permissions they need and how far we are in the migration. It also makes it clearer what needs to be installed on the worker image, and if/when we're ready to get rid of docker. But we can probably make it work either way.
> 
> On the client / taskgraph level, I'm not sure it makes much difference, it'd probably be either setting `containerEngine:docker` in the payload or setting `worker-type: t-whatever-docker`? As a bonus, the latter would be easy to experiment with using `mach try --worker-override`, without making any code changes.

I would much prefer using the task payload here, for the following reasons:
* anything which is explicit in the task is much easier to reason about. it can be confusing for the same task payload to run under docker on one worker pool yet run under podman on a different worker pool
* the fxci config repo is also not part of the firefox monorepo, so having the choice of engine ride the trains and by versioned with the code removes the problems of worker pools migrating from one container to another, and then tasks that were passing beginning to fail when run later on
* in my experience, adding config options to workers has a high cost. it is an additional setting that can be misconfigured, even if it lives in fxci config repo. currently the docker worker workers are managed by fxci, but it isn't unimaginable that at some point we have hardware workers that are not managed by fxci config repo
* being able to switch between podman and docker is something that people greening up tests or devs might wish to do. if it is a worker pool setting, it requires a much more burdensome workflow, requiring code changes to be approved by releng. having it as a task property makes it easy for anyone to try making changes in a try push, without new worker pools needing to be provisioned etc
* there is a benefit of decoupling moving from docker-worker to generic-worker, and switching from docker to podman. I know this well because this is I think a big mistake that I made. I tried to combine the two into a single maneuver, and this combined a lot of complexity from two projects that really should have been independent. by making this a payload-only feature, it means we can migrate everything across using docker initially, and then update taskgraph to start moving jobs over, potentially one-by-one, to podman
* there may be often multiple different tasks that run on the same worker pool. by keeping the choice about docker vs podman to the specific task that is being run, it means we can easily migrate some tasks that run on a worker pool, without migrating all of them at once. this allows a more controlled migration
* having a global worker pool setting might mean that things are working, a new task starts breaking, and the entire set of jobs have to be migrated back to the former container engine (docker). by controlling this on a per-task basis, we never force unrelated tasks to be migrated at the same time, or if new tasks need to run on a worker pool that e.g. don't run under podman, we don't force that all other tasks that have already been adapted to work under podman need to go back to running docker again
* if we really want to force that all jobs need to use podman rather than docker at some point, we can achieve this by either adding an option to disable one engine or the other, or disable d2g entirely, or just not install docker/podman on the host etc. i would prefer to do this with an explicit disable rather than forcing that all tasks just use one engine, because e.g. maybe one day there are 5 different potentialy engines that might be usable, and we might want to allow e.g. 3 of them, rather than have one setting saying which one and only one can be used.

petemoore:
> I have a few thoughts:
> 
>     * Running tasks in containers in a common enough (arguably the default) way to run tasks. Because of this, generic-worker should continue to support it natively in the long term in some way. (This is to say that I don't think that d2g or something like it should ever, eg: be fully removed and stuffed into taskgraph.)

I can imagine you could be right, but I can also imagine the opposite might be true. However, the beauty is I think we don't need to make this decision yet, and we can just decide at any point in the future if we want to drop it, when we have more experience. So long as there is an advantage to keeping it, for sure we can. We would only drop it if we agreed collectively that it was for the better.
> 
>     * I have no opinion (or even the expertise to have much of an opinion) on docker vs. podman. I always thought of d2g as something that was supposed to be entirely transparent, such that task implementors like RelEng (but more importantly, others that have even less docker and taskcluster expertise) don't need to make any change other than the worker type to migrate to generic-worker

I agree. For this reason, it is now my opinion that by default generic-worker should run docker and not podman, because that is exactly what docker-worker did. I am happy to have the ability to explicitly choose to use podman instead (e.g. as part of the task payload) but ideally, we should separate the two different concerns:
* Migrating from docker-worker to generic-worker
* Migrating from docker to podman (for which ever tasks we decide to do this for)
I believe by keeping docker as the default, and not trying to migrate to podman at this early stage, we can be reasonably transparent with the migration, and work on going from docker to podman in the future when this migration is over.

>     * I think it's fine to make the container engine selectable in the payload. However, I'm not really sure how well this plays with the fact that whether or not it will actually work depends on whether or not the worker has `podman` installed. (Come to think of it, it seems like we ought to require all workers to have `podman` installed already, seeing as it's often used?)

The overhead of installing docker and podman isn't very high, I think it is fine to require both. For example, we require other things too, such as the installation of certain system commands, that generic-worker uses to create user accounts, etc. We could probably do a better job of documenting dependencies, this could be a good follow up for this point. Thanks for raising. Note, we could also add the option to disable an engine, if we don't want to install e.g. docker or podman on a worker pool, so that the task will resolve as exception/malformed-payload and log an error that the given engine isn't supported on that worker pool.

> 
>     * More generally, it seems less than ideal that we have a need for supporting both. I suspect there's good reasons we do, but it seems like it would be preferable if all uses could be support with one or the other?

I think this is a secondary migration effort. Docker is the de facto container technology, but comes with some security limitations. Podman is less mainstream, but gaining traction, and offers some favourable security characteristics. So I think the push to support both is so that tasks that can run under podman do so, to improve overall security. Probably the more that can be moved from docker to podman the better, but it may be that not everything can be moved, and doing this during the d2g migration might not be the best time to do this.

changelog/DIlRBhW6S1aQ6Rl63gLxog.md
```diff
@@ -0,0 +1,4 @@
+audience: users
+level: minor
+---
+D2G: Adds `capabilities.containerEngine` to the Docker Worker payload schema strictly to use as a `docker`/`podman` toggle for the d2g-translated payload. Defaults to `podman`.
```
petemoore, line 4: I think this should say that it defaults to `docker`.

tools/d2g/d2g.go
```diff
@@ -98,7 +98,8 @@ func ConvertTaskDefinition(dwTaskDef json.RawMessage) (json.RawMessage, error) {
 // a converted Docker Worker task payload (see d2g.Convert function) to run
 // Docker Worker tasks under Generic Worker.
 func Scopes(dwScopes []string, dwPayload *dockerworker.DockerWorkerPayload, taskQueueID string) (gwScopes []string) {
-	gwScopes = []string{}
+	// scopes to use docker, by default, should just come "for free"
+	gwScopes = []string{"generic-worker:os-group:" + taskQueueID + "/docker"}
```
petemoore, line 102: It occurs to me we could look at dwPayload to see if the engine is set to docker or podman, and only add it if docker is used. Normally we shouldn't add scopes based on what the payload is, but I think in this case it might help restrict the scopes a bit.

tools/d2g/d2g.go
```diff
@@ -107,11 +108,6 @@ func Scopes(dwScopes []string, dwPayload *dockerworker.DockerWorkerPayload, task
 				"generic-worker:os-group:"+taskQueueID+"/kvm",
 				"generic-worker:os-group:"+taskQueueID+"/libvirt",
 			)
-		case s == "docker-worker:capability:privileged":
-			gwScopes = append(
-				gwScopes,
-				"generic-worker:os-group:"+taskQueueID+"/docker",
-			)
```
petemoore, line 114: Note, tasks will now get scope `generic-worker:capability:privileged` added, which does no harm, but doesn't help a lot. That said, there are other scopes it will also not translate, so that might be fine for now, and helps show what the original task had.

tools/d2g/d2g.go
```diff
@@ -136,7 +132,7 @@ func Scopes(dwScopes []string, dwPayload *dockerworker.DockerWorkerPayload, task
 
 // Convert transforms a Docker Worker task payload into an equivalent Generic
 // Worker Multiuser POSIX task payload. The resulting Generic Worker payload is
-// a BASH script which uses Podman to contain the Docker Worker payload. Since
+// a BASH script which uses Podman (by default) to contain the Docker Worker payload. Since
```
petemoore, line 135: docker by default

tools/d2g/d2gtest/testdata/testcases/container_engine_test.yml
```diff
@@ -0,0 +1,122 @@
...
+        command:
+          - echo "Hello world"
+        image: ubuntu
+        maxRunTime: 3600
+      genericWorkerTaskPayload:
+        command:
+          - - bash
+            - '-cx'
+            - >-
+              podman run -t --rm --memory-swap -1 --pids-limit -1 --ulimit host
```
petemoore: docker not podman

tools/d2g/d2gtest/testdata/testcases/container_engine_test.yml
```diff
@@ -0,0 +1,122 @@
...
```
petemoore: docker

petemoore (commented):
A couple of docker/podman the wrong way around, but the rest looks good!

petemoore (approved):
Love it, thanks Boris!

petemoore:
This broken test has caught a genuine issue. It has highlighted that insecure engine does not support specifying osGroups. It allows you to specify an empty list, but that is all.

I think the fix here is to allow a non-empty list, but if there are entries, each entry needs to be a group that the user is a member of (so e.g. generic-worker insecure engine could scan its own groups, and then make sure that every group listed in the array of osGroups, that it is a member of that group, otherwise fail with malformed payload. This way, it ensures that if you are using simple engine, and e.g. the user running generic-worker is not a member of the docker group, it will fail with malformed payload and tell you that you have asked to be in docker group, but that the user running generic-worker insecure engine isn't a member of that group.

petemoore:
Also looks like there are `syscall.Getgroups()` and `user.LookupGroupId` if you want to do it in go natively and not call out to `id` command...

changelog/DIlRBhW6S1aQ6Rl63gLxog.md
```diff
@@ -0,0 +1,4 @@
+audience: users
+level: minor
```
petemoore, line 2: I think this needs to be major, because D2G now uses docker by default for non-privileged tasks, whereas before it used podman.

workers/generic-worker/os_groups_insecure.go
```diff
@@ -13,11 +15,57 @@ type OSGroups struct {
...
 	}
+
+	userGroups, err := currentUserGroups()
+	if err != nil {
+		return executionError(internalError, errored, fmt.Errorf("failed to get user groups: %v", err))
+	}
+
+	notInGroups := checkUserGroupsInList(groupNames, userGroups)
+	if len(notInGroups) > 0 {
+		return MalformedPayloadError(fmt.Errorf("current user is a member of groups: %v, but is not a member of: %v", userGroups, notInGroups))
```
petemoore, line 30:
```suggestion
		return MalformedPayloadError(fmt.Errorf("Task definition contains unsupported osGroups: %v. Allowed values (on this worker pool): %v", notInGroups, userGroups))
```
Does the list include the primary group (e.g. task_XXXXXXXXX)? If so, is there a way to filter that one out, just because that will be variable and is unpredictable, so you can't actually use it. No worries if it isn't trivial.

petemoore (dismissed):
This is so awesome, I love it!!

workers/generic-worker/os_groups_insecure.go
matt-boris: Yes this includes the primary group too. I can filter this out no problem. Thanks for flagging.

workers/generic-worker/os_groups_insecure.go
```diff
@@ -13,11 +15,61 @@ type OSGroups struct {
 }
 
 func (osGroups *OSGroups) Start() *CommandExecutionError {
-	if len(osGroups.Task.Payload.OSGroups) > 0 {
-		return MalformedPayloadError(fmt.Errorf("osGroups feature is not supported on platform %v - please modify task definition and try again", runtime.GOOS))
+	groupNames := osGroups.Task.Payload.OSGroups
+	if len(groupNames) == 0 {
+		return nil
 	}
```
petemoore, line 21: nice optimisation! ❤️

clients/client-shell/cmds/d2g/d2g.go
```diff
@@ -29,6 +29,7 @@ To convert a task definition (JSON), you must use the task definition flag (-t,
   cat /path/to/input/task-definition.json | taskcluster d2g -t
   echo '{"image": "ubuntu", "command": ["bash", "-c", "echo hello world"], "maxRunTime": 300}' | taskcluster d2g`,
 	}
+	cmd.Flags().StringP("engine", "e", "docker", "The container engine to use between docker and podman.")
```
petemoore, line 32: i think there was a readme about this somewhere - do we need to update to include information about the -e option?

tools/d2g/d2g.go
```diff
@@ -97,8 +97,16 @@ func ConvertTaskDefinition(dwTaskDef json.RawMessage) (json.RawMessage, error) {
 // equivalent Generic Worker scopes. These scopes should be used together with
 // a converted Docker Worker task payload (see d2g.Convert function) to run
 // Docker Worker tasks under Generic Worker.
-func Scopes(dwScopes []string, dwPayload *dockerworker.DockerWorkerPayload, taskQueueID string) (gwScopes []string) {
+func Scopes(dwScopes []string, dwPayload *dockerworker.DockerWorkerPayload, taskQueueID, containerEngine string) (gwScopes []string) {
+	if dwPayload.Capabilities.ContainerEngine == "" {
+		dwPayload.Capabilities.ContainerEngine = containerEngine
```
petemoore: I feel like we shouldn't change the docker-worker payload here, because it is passed by reference. So someone fetches the scopes, and then the payload has docker worker payload has changed underneath them, which is a bit unexpected. Instead i think we should use a local variable here to work out which is the appropriate engine to use, based on the payload and the container engine passed in, and use that to decide if we add the os-group scope.

tools/d2g/d2g.go
```diff
@@ -410,6 +408,12 @@ func setFeatures(dwPayload *dockerworker.DockerWorkerPayload, gwPayload *generic
 	}
 }
 
+func setContainerEngine(dwPayload *dockerworker.DockerWorkerPayload, containerEngine string) {
+	if dwPayload.Capabilities.ContainerEngine == "" {
+		dwPayload.Capabilities.ContainerEngine = containerEngine
```
petemoore: same here

workers/generic-worker/os_groups_insecure.go
```diff
@@ -13,11 +15,61 @@ type OSGroups struct {
...
+	groupIDs, err := currentUser.GroupIds()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get current user's group ids: %v", err)
+	}
+
+	primaryGID := currentUser.Gid
+	userGroups := make(map[string]bool)
+	for _, gid := range groupIDs {
+		if gid == primaryGID {
+			continue
```
petemoore, line 54: I don't think this is quite what we want. On some platforms, the group may be fixed (e.g. "staff" on macOS) but on other platforms (e.g. Linux) it may be variable, and match the task user name. I think only in the case it is variable, i.e. matches the user name, we would want to filter it out. So I think here we just need to filter out the case when it matches the username exactly. If it is the primary group, but doesn't match the username, we should keep it.

workers/generic-worker/os_groups_insecure_test.go
```diff
@@ -22,20 +21,15 @@ func TestEmptyOSGroups(t *testing.T) {
...
 }
 
-func TestNonEmptyOSGroups(t *testing.T) {
+func TestNonEmptyOSGroupsUserIsNotIn(t *testing.T) {
 	setup(t)
 	payload := GenericWorkerPayload{
 		Command:    listGroups(),
 		MaxRunTime: 30,
-		OSGroups:   []string{"abc"},
+		OSGroups:   []string{"docker"},
```
petemoore, line 29: it is quite feasible that the user will be in the docker group on some worker pool, I think a more abstract fake group like "abc" is indeed better here.

workers/generic-worker/os_groups_insecure_test.go
```diff
@@ -22,20 +21,15 @@ func TestEmptyOSGroups(t *testing.T) {
...
 	}
 	defaults.SetDefaults(&payload)
 	td := testTask(t)
 
 	_ = submitAndAssert(t, td, payload, "exception", "malformed-payload")
-
-	logtext := LogText(t)
-	if !strings.Contains(logtext, "- osGroups: Array must have at most 0 items") {
```
petemoore, line 38: could we not keep a check that we at least get a snippet of a particular error message we were expecting (e.g. that we don't get completely the wrong message?)

workers/generic-worker/schemas/insecure_posix.yml
```diff
@@ -460,6 +459,16 @@ oneOf:
             still includes feature in order for d2g to work with it.
           type: boolean
           default: false
+        containerEngine:
+          title: Container engine
+          description: >-
+            The container engine to use, between `docker` and `podman`.
+            This property is only used during Generic Worker's internal
+            payload translation (d2g).
```
petemoore, line 467: I think it might be nicer to say "The container engine to use when executing tasks with Docker Worker formatted task payloads." I don't think you need to list the options in the description, because they are already listed in the enum. I like having d2g in brackets. I think "Generic Worker's internal payload translation" is a little bit too opaque, which is why I think I favor being explicit that it is when Generic Worker executes tasks that were designed for Docker Worker.

workers/generic-worker/main.go
```diff
@@ -623,6 +626,12 @@ func (task *TaskRun) validatePayload() *CommandExecutionError {
 		panic(err)
 	}
 	if _, exists := payload["image"]; exists {
+		if !config.EnableD2G {
+			return MalformedPayloadError(errors.New(`docker worker payload detected, but D2G is not enabled on this worker pool.
+If you need D2G to translate your Docker Worker payload so Generic Worker can process it, please do one of two things:
+	1. Contact the owner of the worker pool and ask for D2G to be enabled.
```
petemoore, line 632: It would be really neat here to provide a link to the Worker Pool definition in the UI, since we have the root URL and can create it. This way, they can see who the owner is directly if they follow the link.

petemoore (commented):
Awesome!

petemoore (approved):
Thanks so much!

petemoore (approved):
BEAUTIFUL

Issue #7146, opened 2024-07-12
RFC191: Add launch config expire cron


lotas:
During worker pool configuration updates, previous launch configurations would be marked as archived and kept in the database for a certain amount of time. Usually they should be kept for as long as the workers, that were created with this configuration. Once such workers are expired and removed from db, we can remove the launch configuration as well.

Issue #7125, opened 2024-07-08
launch configuration pulse events


lotas:
Continuation of https://github.com/taskcluster/taskcluster/issues/7085 

When launch configs implemented:

- [ ] launch configuration created `launch-configuration-created`
- [ ] launch configuration deactivated `launch-configuration-deactivated`
- [ ] launch configuration paused event `launch-configuration-paused`
- [ ] launch configuration resumed event `launch-configuration-resumed`

PR #7266, opened 2024-09-20
Worker manager launch configurations

### LINKED ISSUE #7086: RFC191: Introduce launch configuration entity and expose new API endpoints
[RFC191](https://github.com/taskcluster/taskcluster-rfcs/blob/main/rfcs/0191-worker-manager-launch-configs.md)

Add Launch Configuration entity to worker-manager service.

Every launch config should have a unique `launchConfigId` (within a given worker pool). 
If configuration changes, so should `launchConfigId`.
If not specified, worker-manager would automatically generate a unique id.

Existing worker manager API endpoints would make it transparent to the end users, and existing endpoints might be used as they were:
`createWorkerPool`, `updateWorkerPool`, etc

.tbc.

lotas:
Restructuring worker pool configuration to work with launch configs as a separate entity

This aims keep backwards compatibility for all of the existing worker manager API endpoints

## Schema Changes

1. launch configs must be unique (aws/google/azure schema uniqueItems: true) - see below `launchConfigId`

2. launch configs include `workerManager` top level object to configure w-m behaviour
  * `launchConfigId` - has to be unique per pool and unique per configuration
     (if two configs are identical they should have same `launchConfigId` and as such be the same launch config)
     if not specified, then worker manager would generate a unique Id for every launch config
  * `capacityPerInstance` - hint for provisioner to know how many instances to create
  * `initialWeight` - float value `0 .. 1` to tell the provisioner which launch configs are more/less likely to be used
    if two configs with equal `initialWeight` given (`>0`), then they have equal probability of being used for provisioning
    if config has `initialWeight: 0` it will not be used at all. This could be a way to pause certain configs or regions
    This value is later being adjusted dynamically during provision cycle
  * `maxCapacity` - hard limit for the number of non-stopped instances for the current launch config
    This will not allow over-provisioning and w-m will only create as much capacity as given.

`workerManager` should be the only non-cloud related property in the config, and previous `capacityPerInstance` should be migrated at some point. 

## Launch Config ID

Worker-manager uses `launchConfigId` to track worker creation
and attribute any worker or worker-pool errors to their respective launch configurations.

Internally worker-manager would keep history of all launch configs to know which ones were created, updated or archived.
All launch configurations are stored in database as long as there are workers created by it.

Most of the API endpoints that return workers or errors would include `launchConfigId` in its output.

## Provisioning

Provisioner **goals** are:

- distribute load evenly between all launch configs
- observe limits per config (maxCapacity)
- avoid using configs that seen errors in the _last hour_

To understand which launch configurations to choose at provision time, worker-manage would use following data:

- `initialWeight` of the launch config (`0` would be skipped)
- `maxCapacity` to know how much more to create for a given config
- immediate worker pool stats:
  - total number of non-stopped workers
  - total number of errors associated with the given launch config in the _last hour_

Rules for weight adjustment at provision time:

1. If `maxCapacity` is set and `existingCapacity`:
  `adjustedWeight = weight - existingCapacity/maxCapacity`
  This way the weight would be lowered if existing capacity is reaching max allowed capacity,
  and would become zero if it is at max, thus removing it from launch config pool.

2. If worker pool has errors and launch config has errors, weight would be adjusted:
  `adjustedWeight = weight - errorsCount / totalErrors`

  Following scenarios are possible:

  a. If there is one particular launch config causing all the errors, it would be excluded from the selection.

  b. If there are few launch configs causing errors, their selection probility would be lowered, but not zero,
      so worker-manager could keep trying to create instance, hoping that errors would disappear.

  c. If all launch configs have error, their weights would be adjusted accordingly,
     and the ones with less errors would have higher relative weight to the others

Worker-manager will be using `1 hour` sliding window for those errors to let configs auto-recover from errors,
without stopping completely the provisioning.

> [!NOTE]
> This doesn't address the case when there is a broken image that fails to start and pick up new tasks.
> In order to tackle this problem, provisioner could additionally check how many workers did register successfully
and talked to the queue service proportionally to the total number of created workers for this launch config.

--

Rendered "Worker Manager launch configuration" documentation section: https://github.com/taskcluster/taskcluster/blob/ec8ca7f7730b22c10fcd77c1f6dc5991f1e54ee1/ui/docs/reference/core/worker-manager/launch-configurations.mdx 

--

Fixes #7086, #7146, #7125


lotas:
questions during review:
1. if w-p maxCapacity is set to X and l-c maxCapacity is set to Y (Y>X) do we show errors?

petemoore (approved):
Looks great, given that it is backwardly compatible, I would say it is good enough to merge and try out.

services/worker-manager/schemas/v1/worker-pool-launch-configurations-response.ymlx
matt-boris, line 1: Is this supposed to be of file type `.ymlx`?

ui/docs/reference/core/worker-manager/launch-configurations.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+## Configuration
+
+Each launch configuration _may_ include a `workerManager` object that controls how instances are provisioned:
+
+```yaml
+workerManager:
+  launchConfigId: 'unique-config-id'      # Optional: Unique identifier for this configuration
+  capacityPerInstance: 1                  # Number of workers per instance
+  initialWeight: 0.5                      # Float between 0 and 1 (default: 1)
```
matt-boris, line 22:
```suggestion
  initialWeight: 0.5                      # Float between 0 and 1 (default: 0.5)
```

matt-boris (commented):
Looks great Yarik! Awesome work. I've left a few suggestions/comments.

ui/docs/reference/core/worker-manager/launch-configurations.mdx
matt-boris: Apologies! I see default _should be_ 1, but in the schema, you mention how the default is `0.5`, knew I got that value from somewhere 😅

db/tables.yml
```diff
@@ -137,6 +138,16 @@ worker_pools:
...
   previous_provider_ids: jsonb not null
+
+worker_pool_launch_configs:
+  launch_config_id: text not null
+  worker_pool_id: text not null
+  is_archived: boolean not null
+  configuration: jsonb not null
+  created: timestamp with time zone not null
+  last_modified: timestamp with time zone not null
+  # should stats be included in this table or it will be updated too frequently?
```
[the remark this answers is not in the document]
lotas: awesome, thanks, dropping

db/versions/0105-migration.sql
```diff
@@ -0,0 +1,96 @@
...
+  CREATE OR REPLACE FUNCTION _migrate_worker_pools_to_launch_configs()
+  RETURNS void AS $$
+  DECLARE
+      wp RECORD;
+      config JSONB;
+      launch_config_id TEXT;
+  BEGIN
+    truncate table worker_pool_launch_configs;
+
+    -- FOR wp IN SELECT * FROM worker_pools WHERE worker_pool_id LIKE 'proj-taskcluster/%' LOOP
```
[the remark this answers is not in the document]
lotas: well spotted! Thanks

ui/docs/reference/core/worker-manager/launch-configurations.mdx
lotas: brilliant, thanks for spotting this! indeed, constants.yml should be saying that `1` is the default, correcting it

services/worker-manager/schemas/constants.yml
```diff
@@ -204,3 +204,69 @@ quarantine-details-item:
...
+    Must be between 1 and 38 characters long and contain only alphanumeric
+    characters, dashes, and underscores.
+lc-max-capacity:
+  title: Max Capacity
+  type: integer
+  minimum: 0
+  description: |-
+    The maximum capacity of workers of this type (launch configuration)
+    that can be running at any given time.
+    This cannot exceed worker pool's `maxCapacity`.
```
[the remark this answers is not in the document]
lotas: updated description slightly 👍

matt-boris (dismissed):
🎉 🚀 🚢 :shipit:

PR #7268, opened 2024-09-22
refactor: move go test code into correct .go file

petemoore:
Just a trivial cleanup, moving code around into proper place.


matt-boris (approved):
👍🏻

PR #7270, opened 2024-09-23
feat(ui): Fix missing status colors in task view

lotas:
Regression from #7158 
Since we fetch dependencies differently, their status was not upper-cased which didn't let the component to pick right color


petemoore (approved):
lgtm

PR #7271, opened 2024-09-23
D2G: pass --init to docker run / podman run command

### LINKED ISSUE #7255: Command ABORTED on d2g worker
**Describe the bug**
As seen intermittently by our fuzzing instances, some instances fail with a `Command ABORTED` error.  See this [task](https://community-tc.services.mozilla.com/tasks/BXWx-CnNRhiaAP7kReXNVw/runs/0/logs/public/logs/live.log) for further details.

These tasks are running on generic worker with the payload converted by d2g.

petemoore:
Fixes #7255

### DIFF
--- a/tools/d2g/d2gtest/testdata/testcases/artifacts_test.yml
+++ b/tools/d2g/d2gtest/testdata/testcases/artifacts_test.yml
@@ -33,7 +33,7 @@ testSuite:
           - - bash
             - '-cx'
             - >-
-              timeout 3600 podman run -t --name taskcontainer
+              timeout 3600 podman run --init -t --name taskcontainer
               --memory-swap -1 --pids-limit -1 --ulimit host
               -e RUN_ID
               -e TASKCLUSTER_INSTANCE_TYPE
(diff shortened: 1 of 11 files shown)


lotas:
Thanks for adding a quick fix!
Do you think it i will be possible to add an actual test for the `timeout` functionality? Since we didn't catch it before? 
Maybe we can test the actual time it takes to kill the container and compare if it is within some delta?

tools/d2g/d2gtest/testdata/testcases/artifacts_test.yml
```diff
@@ -33,7 +33,7 @@ testSuite:
           - - bash
             - '-cx'
             - >-
-              timeout 3600 podman run -t --name taskcontainer
+              timeout 3600 podman run --init -t --name taskcontainer
```
lotas, line 36: is there a test anywhere that is verifying the timeout part?

lotas (approved):

petemoore:
> Thanks for adding a quick fix! Do you think it i will be possible to add an actual test for the `timeout` functionality? Since we didn't catch it before? Maybe we can test the actual time it takes to kill the container and compare if it is within some delta?

Good call. I've created #7273 for this. Thanks!

matt-boris (commented):
This is great! 🎉

PR #7275, opened 2024-09-23
Docker Compose: Use Generic Worker Multiuser Engine

petemoore:
Note, this depends on #7247.

### DIFF
--- a/generic-worker.Dockerfile
+++ b/generic-worker.Dockerfile
@@ -25,8 +25,8 @@ RUN go build -o /taskcluster
 
 WORKDIR /app/workers/generic-worker
 RUN ./build.sh && \
-  mv generic-worker-multiuser-* /generic-worker-multiuser && \
-  mv generic-worker-insecure-* /generic-worker
+  mv generic-worker-multiuser-* /generic-worker && \
+  mv generic-worker-insecure-* /generic-worker-insecure
 
 FROM ubuntu:jammy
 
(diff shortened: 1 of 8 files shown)


petemoore:
Putting in draft while I test locally. For example, I see that `sudo` is needed in docker image...

generic-worker.Dockerfile
```diff
@@ -53,7 +53,7 @@ else \
     echo \{\"version\": \"69.0.1\", \"commit\": \"local\", \"source\": \"https://github.com/taskcluster/taskcluster\", \"build\": \"NONE\"\} > /version.json; \
 fi
 
-VOLUME /etc/generic-worker/config.json
```
petemoore, line 56: I'm not sure how this worked before, aiui VOLUME should be a directory?

lotas: good catch!

workers/generic-worker/multiuser_posix.go
```diff
@@ -36,12 +36,12 @@ func platformFeatures() []Feature {
 
 func deleteDir(path string) error {
 	log.Print("Removing directory '" + path + "'...")
-	err := host.Run("/usr/bin/sudo", "/bin/chmod", "-R", "u+w", path)
+	err := host.Run("/bin/chmod", "-R", "u+w", path)
```
lotas, line 39: will it cause any issues for the non docker-compose workers? Or will this always be run with the correct permissions?

lotas (approved):
nice!

Issue #7235, opened 2024-09-04
WMIC Deprecation impacting Generic Worker on Windows 11 24H2


jwmoss:
**Describe the bug**
Release SRE are building images to prepare for the next feature release of Windows 11, which is 24H2, slated to be released in October, 2024. Microsoft will deprecate `wmic` (article [here](https://techcommunity.microsoft.com/t5/windows-it-pro-blog/wmi-command-line-wmic-utility-deprecation-next-steps/ba-p/4039242)) in 24H2. When running generic worker, it seems that it relies on `wmic` for creating a new windows user.

**To Reproduce**
Install generic worker on Windows 11 24H2 and run a task. The failure will be similar to `exec: "wmic": executable file not found in %PATH%`.

**Expected behavior**
There should be no errors about `wmic`.

**Taskcluster version**
68.0.2

**Screenshots**
![image](https://github.com/user-attachments/assets/f78b9c09-bb18-4585-a56f-08a60c3890c1)

**Additional context**

petemoore:
`wmic` is currently used at two places in:

https://github.com/taskcluster/taskcluster/blob/773e652d3a6578d78390e19fe653d8425c44e3f7/workers/generic-worker/runtime/runtime_windows.go

ErichDonGubler:
Seems like the solution is to migrate to using the PowerShell commandlets for WMI/CIM? https://learn.microsoft.com/en-us/powershell/scripting/learn/ps101/07-working-with-wmi?view=powershell-7.4

petemoore:
Thanks Erich! Another option might be wrapping the [win32 api](https://learn.microsoft.com/en-us/windows/win32/api/lmaccess/nf-lmaccess-netuseradd) like we do with other win32 apis [here](https://github.com/taskcluster/taskcluster/blob/773e652d3a6578d78390e19fe653d8425c44e3f7/workers/generic-worker/win32/win32_windows.go).

petemoore:
Looking closer, I see we have

```
"net", "user", user.Name, user.Password, "/add", "/expires:never", "/passwordchg:no", "/y"
```

and then later

```
"wmic", "useraccount", "where", "name='" + user.Name + "'", "set", "passwordexpires=false"
```

It isn't clear to me right now why we need to set `passwordexpires=false` in the wmic command when we already specified `/expires:never` in the net user command. But perhaps that relates to user account expiry, rather than password expiry. In any case, it might be able to rewrite the wmic command using just net user?

[...]

PR #7282, opened 2024-09-25
fix(generic-worker)!: remove `wmic` and `net` calls on windows

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/7235.

>Generic Worker (windows): Removes calls to `wmic` (being [deprecated](https://techcommunity.microsoft.com/t5/windows-it-pro-blog/wmi-command-line-wmic-utility-deprecation-next-steps/ba-p/4039242)) and `net` in favor of a more modern approach using PowerShell cmdlets.
>
>The `powershell` executable is required to be in the path.

### DIFF
--- a/workers/generic-worker/runtime/runtime_windows.go
+++ b/workers/generic-worker/runtime/runtime_windows.go
@@ -21,8 +21,8 @@ type OSUser struct {
 func (user *OSUser) CreateNew(okIfExists bool) error {
 	log.Print("Creating Windows user " + user.Name + "...")
 	userExisted, err := host.RunIgnoreError(
-		"The account already exists",
-		"net", "user", user.Name, user.Password, "/add", "/expires:never", "/passwordchg:no", "/y",
+		"User "+user.Name+" already exists",
+		"powershell", "-Command", "New-LocalUser -Name '"+user.Name+"' -Password (ConvertTo-SecureString '"+user.Password+"' -AsPlainText -Force) -AccountNeverExpires -PasswordNeverExpires -UserMayNotChangePassword",
 	)
 	if err != nil {
 		return err
@@ -33,8 +33,7 @@ func (user *OSUser) CreateNew(okIfExists bool) error {
 	log.Print("Created new OS user!")
 	err = host.RunBatch(
 		userExisted,
-		[]string{"wmic", "useraccount", "where", "name='" + user.Name + "'", "set", "passwordexpires=false"},
-		[]string{"net", "localgroup", "Remote Desktop Users", "/add", user.Name},
+		[]string{"powershell", "-Command", "Add-LocalGroupMember -Group 'Remote Desktop Users' -Member '" + user.Name + "'"},
 	)
 	// if user existed, the above commands can fail
 	// if it didn't, they can't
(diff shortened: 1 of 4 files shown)


matt-boris:
Successfully tested on the win2022-staging pool https://community-tc.services.mozilla.com/provisioners/proj-taskcluster/worker-types/gw-windows-2022-staging?sortBy=Last%20Active&sortDirection=desc

Example successful tasks:
* https://community-tc.services.mozilla.com/tasks/IrJDx32mR52B0GFKIEnjMg (with a successful rerun, too)
* https://community-tc.services.mozilla.com/tasks/ecI_CF5GQ4uwDwfRLTjNZA

lotas (approved):
nice

workers/generic-worker/runtime/runtime_windows.go
```diff
@@ -83,12 +85,12 @@ func DeleteUser(username string) (err error) {
...
 	if err != nil {
 		return
 	}
-	for _, line := range strings.Split(out, "\r\n") {
-		trimmedLine := strings.Trim(line, "\r\n ")
+	for _, line := range strings.Split(out, "\n") {
+		trimmedLine := strings.TrimSpace(line)
 		usernames = append(usernames, trimmedLine)
 	}
 	return
```
petemoore, line 96: Were you able to test this locally? I'm sure it will be fine, but it occurs to me that if for some reason it was broken and didn't list historical task user accounts, disks might slowly fill up over time...
Another way to check this would be to run several tasks on a worker, and then run a task to list the users or `dir` the parent of the home directory, just to make sure it really is cleaning up old users.

petemoore (approved):
Awesome, love it! Thanks so much for doing this so quickly!

jwmoss:
> Successfully tested on the win2022-staging pool https://community-tc.services.mozilla.com/provisioners/proj-taskcluster/worker-types/gw-windows-2022-staging?sortBy=Last%20Active&sortDirection=desc
> 
> Example successful tasks:
> 
>     * https://community-tc.services.mozilla.com/tasks/IrJDx32mR52B0GFKIEnjMg (with a successful rerun, too)
> 
>     * https://community-tc.services.mozilla.com/tasks/ecI_CF5GQ4uwDwfRLTjNZA

This looks great! I think my only concern is that it was tested on win2022 server, which still has `wmic.exe` included, although I did a quick search on `wmic` in the taskcluster code and didn't see any other references, so I'm sure this is working as expected!

workers/generic-worker/runtime/runtime_windows.go
matt-boris: Yeah, the output of `wmic` vs the powershell `Get-LocalUser` are quite different, with the powershell cmdlet outputting with just newline separators. It's a good idea though to try running multiple tasks against one worker - I'll do that now. Honestly, the `strings.TrimSpace()` isn't needed at all, too.

matt-boris:
>This looks great! I think my only concern is that it was tested on win2022 server, which still has wmic.exe included, although I did a quick search on wmic in the taskcluster code and didn't see any other references, so I'm sure this is working as expected!

Yeah, good thing to point out here. Yes, this is definitely only tested on win2022 server for right now, but I can assure you, we no longer use `wmic` :)

I'm going to make one small change to the code and rerun some tests then this should be good to go 👍🏻

workers/generic-worker/runtime/runtime_windows.go
matt-boris: Yeah, this is working as expected. https://community-tc.services.mozilla.com/provisioners/proj-taskcluster/worker-types/gw-windows-2022-staging/workers/westus/vm-cspkoykswqp4s8jzuiruqcxk4eh0pq1g1rd?sortBy=started&sortDirection=desc

PR #7283, opened 2024-09-26
Worker Manager  consider idle workers during provisioning

### LINKED ISSUE #7052: Worker-manager provisioner could use total claimed to know how many tasks are running
Form this discussion:

https://github.com/taskcluster/taskcluster-rfcs/pull/191#pullrequestreview-2098453039

https://github.com/taskcluster/taskcluster-rfcs/pull/191#issuecomment-2149691567

It becomes apparent that decision to spawn workers is being made only on the number of workers that are in RUNNING state.
However, worker in RUNNING state isn't guaranteed to be working on a task (either between the tasks, preparing to run a task, or simply failure to start)

If we make provisioner a bit smarter, and take `totalClaimed` into the equation, then it should prevent scenarios where we keep spawning workers for a single pending task. 

We could check `totalClaimed` + `totalPending` and use this instead of number of workers to decide if new one should be launched

lotas:
What happens when workers start but don't do any work (temp or permanent image/config issue) is that worker manager keeps launching new instances in response to `pendingTasks>0`

If you only have 1 pending task and worker pool configuration that doesn't result in "good worker", existing capacity will keep increasing with each provision loop. 

Since we know how many tasks are currently being claimed for any worker/task queue, we are able to determine how many of currently running workers/capacity are **idle**:

```
    const idleCapacity = existingCapacity - claimedTasks
```

With this we can adjust `pendingTasks` count  assuming that some idle workers are going to pick those tasks soon:

```
    const adjustedPendingTasks = Math.max(0, pendingTasks - idleCapacity);
```

----

After some playing with https://github.com/taskcluster/taskcluster-provisioning-simulator/
tests show that using `claimedTasks` can reduce total number of workers created, while only slightly increasing mean wait time
~1% less workers needed with ~1% increase of wait time

| **Metric**                     | **No Claimed Task Logic**           | **With Idle Capacity Logic**          |
|-------------------------------|-------------------------------------|---------------------------------------|
| **Number of Events**          | 319,287                             | 318,759                               |
| **Simulation Duration**       | 5 days (432,000,000 ms)             | 5 days (432,000,000 ms)               |
| **Requested Workers**         | 0 workers requested                 | 0 workers requested                   |
| **Running Workers**           | 0 workers running                   | 0 workers running                     |
| **Shutdown Workers**          | 25,650 workers shutdown             | 25,474 workers shutdown               |
| **Overprovisioned Workers**   | 329 workers                         | 262 workers                           |
| **Overprovisioned Worker Time** | 23 days (2,030,067,814 ms)        | 22 days (1,867,490,814 ms)            |
| **Minimum Wait Time**         | none                                | none                                  |
| **Maximum Wait Time**         | 17 minutes (1,048,073 ms)           | 17 minutes (1,048,073 ms)             |
| **Mean Wait Time**            | less than a minute (19,210.01 ms)   | less than a minute (19,417.52 ms)     |
| **Median Wait Time**          | none                                | none                                  |
| **Minimum Tasks per Worker**  | 0 tasks                             | 0 tasks                               |
| **Maximum Tasks per Worker**  | 28 tasks                            | 29 tasks                              |
| **Mean Tasks per Worker**     | 3.149 tasks                         | 3.171 tasks                           |
| **Median Tasks per Worker**   | 2 tasks                             | 2 tasks                               |

Fixes #7052


services/worker-manager/src/estimator.js
```diff
@@ -4,9 +4,25 @@ export class Estimator {
...
+    workerPoolId,
+    minCapacity,
+    maxCapacity,
+    scalingRatio = 1.0,
+    maxIdlePercent = 0.2, // maximum allowed percentage of idle workers
+    workerInfo: { existingCapacity, stoppingCapacity = 0, requestedCapacity = 0 },
+  }) {
     const { pendingTasks } = await this.queue.pendingTasks(workerPoolId);
-    const { existingCapacity, stoppingCapacity = 0, requestedCapacity = 0 } = workerInfo;
+    const { claimedTasks } = await this.queue.claimedTasks(workerPoolId);
```
lotas, line 16: yet to be implemented, possible both values above in a single api call

changelog/issue-7052.md
```diff
@@ -0,0 +1,7 @@
+audience: worker-deployers
+level: minor
+reference: issue 7052
+---
+
+Worker-manager now uses number of claimed tasks during estimation process to avoid having too much idling workers.
```
matt-boris, line 6:
```suggestion
Worker-manager now uses number of claimed tasks during estimation process to avoid having too many idling workers.
```

services/queue/src/api.js
```diff
@@ -1995,6 +1998,47 @@ builder.declare({
...
+    '',
+    'As task states may change rapidly, this number may not represent the exact',
+    'number of pending and claimed tasks, but a very good approximation.',
+  ].join('\n'),
+}, async function(req, res) {
+  const taskQueueId = req.params.taskQueueId;
+  const { provisionerId, workerType } = splitTaskQueueId(taskQueueId);
+
+  // Get number of pending message
+  let [pendingCount, claimedCount] = await Promise.all([
```
matt-boris, line 2027: Prefer `const` here I think

```suggestion
  const [pendingCount, claimedCount] = await Promise.all([
```

services/queue/src/api.js
```diff
@@ -1406,6 +1406,7 @@ builder.declare({
     );
   }
 
+  await this.queueService.removePendingMessage(taskId, runId);
```
matt-boris, line 1409: Should this have always been here and was noticed in implementing this PR? Seems like a bug fix

lotas: this is only because this method is long being deprecated, and I believe only tests are using this method to claim specific task by taskId. 
Plus tests never tested the exact number of pending tasks because of the extra complexity of the data storage in the past (base64 encoded json strings with different messages in a single azure table)

matt-boris: Cool thanks for the info 👍🏻

matt-boris (approved):
👍🏻

Issue #7222, opened 2024-08-30
Worker Runner: kernel deprecation warning


matt-boris:
```
kernel: [   28.254982] start-worker (530): /proc/530/oom_adj is deprecated, please use /proc/530/oom_score_adj instead.
```

PR #7284, opened 2024-09-26
fix(worker-runner): replace deprecated oom adjuster

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/7222.

>Worker Runner: Replaces deprecated `/proc/<pid>/oom_adj` with `/proc/<pid>/oom_score_adj`.


lotas (approved):
cool 👍

PR #7290, opened 2024-09-30
feat(auth): Fix error handling in continuation token

lotas:
`hashids.decode` throw an error when input contains unexpected characters. 
This wraps the handling and responds 400 to the client instead of 500


matt-boris (approved):
Nice, ty

PR #7291, opened 2024-10-01
feat(w-m): Handling edge-case in estimation

### LINKED ISSUE #7052: Worker-manager provisioner could use total claimed to know how many tasks are running

lotas:
provisioner and estimator are operating with partially stale information

Fixes #7052

### DIFF
--- a/changelog/issue-7052.md
+++ b/changelog/issue-7052.md
@@ -1,0 +1,6 @@
+audience: worker-deployers
+level: patch
+reference: issue 7052
+---
+
+Fixes edge-case in estimation that was introduced in #7283 where claimed count might be greater than the existing capacity.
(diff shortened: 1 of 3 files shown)


petemoore (approved):
Perfect, many thanks!

PR #7306, opened 2024-10-02
Generic Worker task log headers and sentry reports include key worker config settings

### LINKED ISSUE #7305: Log important worker config settings in task header
Some worker config settings fundamentally affect how tasks operated. Therefore it is useful to log these config settings in the task header, together with other information about the worker environment.

In particular, for multiuser enginer, this would be `runTasksAsCurrentUser` and `headlessTasks` config settings.

petemoore:
Fixes #7305

### DIFF
--- a/changelog/issue-7305.md
+++ b/changelog/issue-7305.md
@@ -1,0 +1,11 @@
+audience: users
+level: patch
+reference: issue 7305
+---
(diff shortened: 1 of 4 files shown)


matt-boris (approved):
LGTM thanks Pete. I think we've looked into this before, but as long as maps are passed in by reference automatically, then this is good 👍🏻

PR #7308, opened 2024-10-02
fix(d2g): use `--device` instead of volume mounts

### LINKED ISSUE #7307: d2g: `-v /dev/video0:/dev/video0` does not work in docker
**Describe the bug**

d2g passes the video device as a docker volume, which doesn't seem to work (the device exists in the container, but is not usable).  This was changed from `--device` for the sake of old versions of podman, AIUI.

**To Reproduce**

```
$ docker run --rm -v /dev/video0:/dev/video0 ubuntu:18.04 cat /dev/video0
cat: /dev/video0: Operation not permitted
```

**Expected behavior**

```
$ docker run --rm --device /dev/video0 ubuntu:18.04 cat /dev/video0
cat: /dev/video0: Invalid argument
```

**Taskcluster version**
72.0.0

 - OS: linux

matt-boris:
Fixes #7307.

>Generic Worker (D2G): Pass devices through to the `docker run ...` command using `--device` instead of a volume mount.

### DIFF
--- a/changelog/issue-7307.md
+++ b/changelog/issue-7307.md
@@ -1,0 +1,5 @@
+audience: general
+level: patch
+reference: issue 7307
+---
+Generic Worker (D2G): Pass devices through to the `docker run ...` command using `--device` instead of a volume mount.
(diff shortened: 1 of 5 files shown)


petemoore (approved):
Looks great!

For context, for anyone reading this, it reverts https://github.com/taskcluster/taskcluster/pull/7061/commits/9c0ad40c4b5be437b80d29d55f3959be30a96dac which was added due to issues with podman 3. Since now podman 4 is stable/standard, and the podman 3 fix broke things for docker anyway, it makes sense to pull it out and no longer support podman 3. Thanks Boris!

Issue #7286, opened 2024-09-27
Better document run state transitions


djmitche:
..and also update the task state transitions to support rerun.

djmitche:
Here: https://docs.taskcluster.net/docs/reference/platform/queue/task-life-cycle

PR #7310, opened 2024-10-03
Include function definitions in documentation

djmitche:
In working on #7286, I had a hard time finding all of the DB functions in one place. This adds them hidden behind `<details>` elements.

[Results](https://github.com/djmitche/taskcluster/blob/dumpfns/db/fns.md)


petemoore (approved):
I like this a lot!!!

PR #7311, opened 2024-10-03
notify: add rootUrl to the slack/matrix/email message context

jcristau:


petemoore (approved):
This looks great, many thanks Julien. I particularly appreciate that you updated tests and documentation at the same time. I'd like Yarik/Boris to have a chance to look at the changes too before we merge, as I'm not super familiar with the code here, just in case they see issues that I might miss. Thanks!

lotas (approved):
Great, thank you @jcristau

PR #7312, opened 2024-10-03
fix(tests): don't skip client tests due to nock issue

### LINKED ISSUE #6923: Enable several skipped tests in client library because of nock and node 20.10 issues
During upgrade to node 20 it was discovered that `nock` has some issues with mocking endpoints, this is why three tests were disabled: https://github.com/taskcluster/taskcluster/pull/6878/files#diff-ed1d5ac9fc5bdf958ee722903df5e16bc53f25bcd1764c8b466545ec5b8390faR290 

After this issue gets fixed, we can upgrade and check if the issue is gone https://github.com/nock/nock/issues/2595 

We tested all minor versions of `20` release and discovered that issue first appeared in `20.10`

matt-boris:
Fixes #6923.

Opened because of this comment https://github.com/nock/nock/issues/2595#issuecomment-2391538222.

### DIFF
--- a/changelog/issue-6923.md
+++ b/changelog/issue-6923.md
@@ -1,0 +1,4 @@
+audience: general
+level: silent
+reference: issue 6923
+---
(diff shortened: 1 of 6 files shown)


petemoore (approved):
Awesome!

Issue #7254, opened 2024-09-12
d2g: download images to a shared location/cache


jcristau:
In docker-worker, because images are downloaded to docker's system-wide store, any image only needs to be downloaded once per worker.
It would be good if d2g would similarly arrange for the image storage to be shared across tasks on the same worker.

jcristau:
As an example:

```
[taskcluster 2024-10-02T17:09:07.997Z] [mounts] Downloading task O53-n_L6RAqkbMzIoYcAwQ artifact public/image.tar.zst to downloads/BzfuVEV3Q3-D8gQXzoCGzg
[taskcluster 2024-10-02T17:09:29.280Z] [mounts] Downloaded 1444548990 bytes with SHA256 cdd9f389cbaec67ceae714c72fc6237ad1e1a687ac40168d9d14b82fc4b05d12 from task O53-n_L6RAqkbMzIoYcAwQ artifact public/image.tar.zst to downloads/BzfuVEV3Q3-D8gQXzoCGzg
[...]
[taskcluster 2024-10-02T17:09:29.289Z] [mounts] Decompressing zst file downloads/BzfuVEV3Q3-D8gQXzoCGzg to '/home/task_172788885745726/dockerimage'
[taskcluster 2024-10-02T17:09:52.121Z] [mounts] Granting task_172788885745726 full control of file '/home/task_172788885745726/dockerimage'
[...]
[taskcluster 2024-10-02T17:09:52.122Z] exit "${exit_code}"'
++ docker load --input dockerimage
++ sed -n '0,/^Loaded image: /s/^Loaded image: //p'
+ IMAGE_ID=ubuntu1804-test:latest
++ pwd
++ pwd
+ timeout 5400 docker run --init -t --name taskcontainer [...]
 * Starting system message bus dbus       
[setup 2024-10-02T17:11:37.332Z] run-task started in /builds/worker
```

We spend 20+ seconds downloading the image, another ~20 seconds decompressing it, and then somewhere around 90 or 100 seconds in `docker load`.

jcristau:
For comparison, AIUI d-w saves the image in docker under a unique name (based on task id + artifact name), and remembers which images it knows about (with https://github.com/taskcluster/taskcluster/blob/main/workers/docker-worker/src/docker/image_manager.js).

matt-boris:
A simple POC I have has this payload:
```yaml
command:
  - - /bin/bash
    - '-cx'
    - |-
[...]

PR #7316, opened 2024-10-08
fix(d2g): don't decompress certain image extensions before docker load

matt-boris:
>D2G: No longer specify file mount format on image if compressed with gzip, bzip2, xz, or zstd when using docker. Generic Worker will now no longer decompress these files before running `docker load`. Docs [here](https://docs.docker.com/reference/cli/docker/image/load/).

A small optimization I found while looking into https://github.com/taskcluster/taskcluster/issues/7254.


petemoore (approved):
Looks good, thanks Boris!

lotas (approved):

PR #7323, opened 2024-10-10
fix(w-m): transition aws/gcp workers to stopping on `removeWorker` call

### LINKED ISSUE #7322: AWS/GCP workers never go into stopping state while being terminated
Stopping state was only used for Azure, but this can cause a bit of confusion seeing running workers, even if they're currently being terminated. I think they should transition from running to stopping once `removeWorker` is called, then worker scanner will properly move to stopped once it sees they're fully terminated.

matt-boris:
Fixes #7322.

>Worker Manager: AWS and GCP workers now transition into stopping state on call to `removeWorker`.

### DIFF
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -383,10 +383,10 @@ export class AwsProvider extends Provider {
   }
 
   async removeWorker({ worker, reason }) {
-    await worker.update(this.db, worker => {
-      if ([Worker.states.REQUESTED, Worker.states.RUNNING].includes(worker.state)) {
-        worker.lastModified = new Date();
-        worker.state = Worker.states.STOPPING;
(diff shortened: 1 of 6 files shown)


services/worker-manager/src/providers/aws.js
```diff
@@ -383,6 +383,12 @@ export class AwsProvider extends Provider {
   }
 
   async removeWorker({ worker, reason }) {
+    await worker.update(this.db, worker => {
```
lotas, line 386: I'd probably rename this inner variable to avoid shadowing and guessing which value is being updated

matt-boris: Good call, thanks. It was a copy/pasta, so I updated the original as well 👍🏻

lotas (approved):
Thanks!

PR #7324, opened 2024-10-11
generic-worker, d2g: let containers talk to the tc proxy without sharing the host network namespace

### LINKED ISSUE #6797: d2g: --net=host causes problems for android pgo kvm jobs
**Describe the bug**
runtask is failing at https://searchfox.org/mozilla-central/rev/961556b584938a694d7df6e7bf25d70253c37060/taskcluster/scripts/run-task#797 under Generic Worker.

See broken [treeherder job](https://treeherder.mozilla.org/logviewer?job_id=444475711&repo=try&lineNumber=246). Broken [taskcluster task](https://firefox-ci-tc.services.mozilla.com/tasks/T7ds8_QTTFaeVRCXN3msNw/runs/0/logs/public/logs/live.log#L246).

Under docker worker, this works. See working docker worker [treeherder job](https://treeherder.mozilla.org/jobs?repo=try&revision=7fd3a844e044bb8af7cdacd16ebfe4598e079e64&selectedTaskRun=dUtz5PcyTBWqwTwG59XgJg.0). Working [taskcluster task](https://firefox-ci-tc.services.mozilla.com/tasks/dUtz5PcyTBWqwTwG59XgJg) or maybe [this one](https://treeherder.mozilla.org/jobs?repo=try&revision=fe50bb952ccf4ae6db913684aa00a5bd49ee02a0&selectedTaskRun=XXdjmREoSUS9cHpxpUTnLQ.0) (need to double check).

@jrmuizel says:

> the Xvfb problem looks like it’s caused by the abstract unix domain socket namespace being shared with container because of -net=host. See https://stackoverflow.com/questions/74733639/cannot-bind-x-unix-socket-within-docker-container

The reason we chose `--net=host` rather than e.g. `--net=private` was to keep things simple. Tasks _should_ use `$TASKCLUSTER_PROXY_URL` for accessing the taskcluster proxy (which we could change inside the podman container) _but_ they may also be explicitly using e.g. explicit references to expected proxy location with commands like `curl -fsSL http://localhost:80/api/queue/v1/task/...`, and we [would like these to keep working](https://linuxreviews.org/WE_DO_NOT_BREAK_USERSPACE) too under Generic Worker. So we ideally need to find a way to have taskcluster proxy urls work regardless of whether they are hardcoded with `http://localhost` or whether they use `$TASKCLUSTER_PROXY_URL` but at the same time, we need the kvm stuff to work too that is broken as described above.

### LINKED ISSUE #7320: d2g: don't use `--net=host`
d2g uses `--net=host` if a task wants access to the taskcluster proxy.  That gives the task privileged access to the host, which it shouldn't have.

jcristau:
Github Bug/Issue: Fixes #7320, fixes #6797

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -358,7 +358,7 @@ func runCommand(containerName string, dwPayload *dockerworker.DockerWorkerPayloa
 	}
 	command.WriteString(createVolumeMountsString(dwPayload, wdcs))
 	if dwPayload.Features.TaskclusterProxy {
-		command.WriteString(" --add-host=taskcluster:127.0.0.1 --net=host")
+		command.WriteString(" --add-host=taskcluster:host-gateway -e TASKCLUSTER_PROXY_URL=${TASKCLUSTER_PROXY_URL/localhost/taskcluster}")
 	}
 	command.WriteString(envMappings(dwPayload))
 	dockerImageString, err := dwImage.String(tool)
@@ -566,10 +566,6 @@ func envMappings(dwPayload *dockerworker.DockerWorkerPayload) string {
 		"TASKCLUSTER_WORKER_LOCATION",
 		"TASK_GROUP_ID", // note, docker-worker didn't set this, but decision tasks may in future choose to use it if it is set
 		"TASK_ID",
-	}
-
-	if dwPayload.Features.TaskclusterProxy {
-		additionalEnvVars = append(additionalEnvVars, "TASKCLUSTER_PROXY_URL")
 	}
 
 	envVarNames := []string{}
--- a/changelog/issue-7320.md
+++ b/changelog/issue-7320.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: major
+reference: issue 7320
+---
+D2G: containers no longer use the host's network namespace
--- a/tools/d2g/d2gtest/testdata/testcases/decision_task_tests.yml
+++ b/tools/d2g/d2gtest/testdata/testcases/decision_task_tests.yml
@@ -84,7 +84,8 @@ testSuite:
               timeout -s KILL 3600 docker run -t --name taskcontainer
               --memory-swap -1 --pids-limit -1 --privileged
               -v "$(pwd)/cache0:/builds/worker/checkouts"
-              --add-host=taskcluster:127.0.0.1 --net=host
+              --add-host=taskcluster:host-gateway
+              -e TASKCLUSTER_PROXY_URL=${TASKCLUSTER_PROXY_URL/localhost/taskcluster}
               -e GECKO_BASE_REPOSITORY
               -e GECKO_BASE_REV
               -e GECKO_HEAD_REF
@@ -96,7 +97,6 @@ testSuite:
               -e RUN_ID
               -e TASKCLUSTER_CACHES
               -e TASKCLUSTER_INSTANCE_TYPE
-              -e TASKCLUSTER_PROXY_URL
               -e TASKCLUSTER_ROOT_URL
               -e TASKCLUSTER_WORKER_LOCATION
               -e TASK_GROUP_ID
--- a/tools/d2g/d2gtest/testdata/testcases/fuzzing_tests.yml
+++ b/tools/d2g/d2gtest/testdata/testcases/fuzzing_tests.yml
@@ -38,14 +38,26 @@ testSuite:
         command:
         - - bash
           - "-cx"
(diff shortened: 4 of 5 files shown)


tools/d2g/d2g.go
```diff
@@ -568,10 +568,6 @@ func envMappings(dwPayload *dockerworker.DockerWorkerPayload) string {
 		"TASK_ID",
 	}
 
-	if dwPayload.Features.TaskclusterProxy {
-		additionalEnvVars = append(additionalEnvVars, "TASKCLUSTER_PROXY_URL")
-	}
-
```
matt-boris, line 574: I think I would prefer having any and all env vars be added directly in this helper `envMappings()` function. So this could change to something like:
```go
if dwPayload.Features.TaskclusterProxy {
    additionalEnvVars = append(additionalEnvVars, "TASKCLUSTER_PROXY_URL=${TASKCLUSTER_PROXY_URL}/localhost/taskcluster")
}
```

tools/d2g/d2g.go
```diff
@@ -357,10 +357,10 @@ func runCommand(containerName string, dwPayload *dockerworker.DockerWorkerPayloa
 		command.WriteString(" --security-opt=seccomp=unconfined")
 	}
 	command.WriteString(createVolumeMountsString(dwPayload, wdcs))
+	command.WriteString(envMappings(dwPayload))
```
matt-boris, line 360: Any reason for the different ordering?

tools/d2g/d2g.go
```diff
@@ -357,10 +357,10 @@ func runCommand(containerName string, dwPayload *dockerworker.DockerWorkerPayloa
...
 	if dwPayload.Features.TaskclusterProxy {
-		command.WriteString(" --add-host=taskcluster:127.0.0.1 --net=host")
+		command.WriteString(" --add-host=taskcluster:host-gateway -e TASKCLUSTER_PROXY_URL=${TASKCLUSTER_PROXY_URL/localhost/taskcluster}")
```
matt-boris, line 362: I'm not familiar with adding this `host-gateway`. Do you have any tasks you made manually that uses this that the container has access to the tc proxy?

matt-boris (commented):
Thanks for helping look into this!

tools/d2g/d2g.go
jcristau: That doesn't work, because additionalEnvVars is then passed through `shell.Escape` (via `envSetting`).

jcristau: Also, the `/localhost/taskcluster` does need to be within the braces.

tools/d2g/d2g.go
jcristau: No particular reason.

tools/d2g/d2g.go
jcristau: That would need a g-w with the first patch here, for the proxy to listen on the docker interface.

Per https://docs.docker.com/reference/cli/docker/container/run/#add-host "The `--add-host` flag supports a special `host-gateway` value that resolves to the internal IP address of the host. This is useful when you want containers to connect to services running on the host machine."

tools/d2g/d2g.go
matt-boris: Ah right, thanks, I had forgotten about `shell.Escape`.

tools/d2g/d2g.go
jcristau: Reordered.

matt-boris: It'll make it so you don't have to reorder so many of the test cases, keep the diff to what's truly changing here 👍🏻

tools/d2g/d2g.go
matt-boris: Thanks 👍🏻 Are you planning on doing the generic worker/tc proxy patch too? We can assist however you'd like/need.

matt-boris (commented):
Changes LGTM, let's just build a staging pool off of this commit and do some testing before merging! 👍🏻

matt-boris:
Building our ubuntu staging image off of https://github.com/taskcluster/taskcluster/tree/d2g-proxy.

matt-boris:
Successful tests with docker https://community-tc.services.mozilla.com/tasks/SZGlDbShTn6DCtsXzTngHA/runs/0
and podman https://community-tc.services.mozilla.com/tasks/BbZ9bsxUTr2QF6gwZT3M1A/runs/0

matt-boris (approved):
LGTM, will you please rebase with `main` and we'll get this merged?

PR #7328, opened 2024-10-15
fix(d2g): one-to-one mapping for `--privileged` flag

### LINKED ISSUE #7327: D2G: Map `--privileged` to only the `dwPayload.Capabilities.Privileged` bool
Users shouldn't be able to get `--privileged` by using dind or host shared memory https://github.com/taskcluster/taskcluster/blob/215b304c87ed397534d9f9f11aefa5dce8d61a09/tools/d2g/d2g.go#L351

matt-boris:
Fixes #7327.

>D2G: Don't provide `--privileged` flag for dind and host shared memory use. Only using now as a one-to-one mapping to Docker Worker's privileged payload flag.

### DIFF
--- a/changelog/issue-7327.md
+++ b/changelog/issue-7327.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 7327
+---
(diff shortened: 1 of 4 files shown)


lotas (approved):
awesome!

PR #7332, opened 2024-10-18
Return an error when the rust client fails to upload an artifact

Eijebong:
Without this, the rust client would silently fail when uploading an artifact if the PUT request would return a 4/5xx.


lotas (approved):
Many thanks Bastien!

PR #7341, opened 2024-10-25
fix(d2g): use unique task container names

### LINKED ISSUE #7340: D2G: Aborted tasks sometimes don't remove the running container
In the case where a user manually cancels a task during execution of the `docker run -t --name taskcontainer ...` command, the final `docker rm taskcontainer` is never successfully ran. This results in the following error on subsequent runs on the same worker:

```bash
docker: Error response from daemon: Conflict. The container name "/taskcontainer" is already in use by container "cb3c084614528b9bc8ad4cf9fbecafcce283ddcd437aa51767fc389f9d683037". You have to remove (or rename) that container to be able to reuse that name.
```

We should probably force remove the container before the run begins.

matt-boris:
Fixes #7340.

>D2G: Use unique task container names to avoid container naming conflicts.

cc @jschwartzentruber

### DIFF
--- a/changelog/issue-7340.md
+++ b/changelog/issue-7340.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 7340
+---
+D2G: Use unique task container names to avoid container naming conflicts.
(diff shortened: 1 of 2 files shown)


matt-boris:
I'm going to rework this PR to instead make the name unique.

petemoore (approved):
Love it, thanks!

PR #7342, opened 2024-10-25
Matt boris/revert decision task to use d2 g

matt-boris:
This is to have our decision task use the latest changes we make to d2g. Pete had initially made these changes before we had landed d2g into the monorepo and incorporated it in with generic worker. This was done in an effort to migrate away from our docker worker pools. I think now that we have d2g integrated in and we’re pretty consistently changing things about it, I believe we should switch it back.


petemoore (approved):
wfm, thanks!

PR #7343, opened 2024-10-28
feat(d2g)!: added scope/payload validation

matt-boris:
>D2G: Renamed methods `Convert()` --> `ConvertPayload()` and `Scopes()` --> `ConvertScopes()`.
>
>D2G: `ConvertScopes()` checks that the provided docker worker payload is valid with the supplied scopes. Generic Worker will now resolve a docker worker task as `exception/malformed-payload` if any required docker worker scopes are missing for its payload.

### DIFF
--- a/changelog/fady1OgaQ1ulcxtyDR4fuQ.md
+++ b/changelog/fady1OgaQ1ulcxtyDR4fuQ.md
@@ -1,0 +1,6 @@
+audience: worker-deployers
+level: major
+---
+D2G: Renamed methods `Convert()` --> `ConvertPayload()` and `Scopes()` --> `ConvertScopes()`.
+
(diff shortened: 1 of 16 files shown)


petemoore (commented):
Thanks Boris!

petemoore (commented):
I've just realised this isn't going to work. Rather than replacing docker-worker scopes with generic-worker scopes, we are going to need to keep all the docker-worker scopes but _add_ the generic-worker scopes too. The reason for this is that any task that creates other tasks (i.e. typically decision tasks) need to themselves posses (satisfy) scope X if they create a task that has scope X. So for example, a decision task that creates docker-worker tasks (with docker-worker scopes) would need to posses the docker-worker scopes itself too. If we translate them into generic-worker scopes and discard the original docker-worker scopes, the decision task will not be able to create any tasks that have docker-worker scopes.

petemoore (approved):
Awesome!

PR #7372, opened 2024-11-07
fix(d2g): add scope check validations for loopback audio/video

matt-boris:
Builds on top of https://github.com/taskcluster/taskcluster/pull/7343.

>D2G: `ConvertScopes()` additionally checks scopes for `loopbackAudio` and `loopbackVideo`.

### DIFF
--- a/changelog/Nf0wHHyBR52vH0-l7oaJ0Q.md
+++ b/changelog/Nf0wHHyBR52vH0-l7oaJ0Q.md
@@ -1,0 +1,4 @@
+audience: general
+level: patch
+---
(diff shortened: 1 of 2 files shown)


lotas (approved):
lgtm!

petemoore (commented):
Rather than check for scopes satisfied in each block, we could collect together all the required scopes for all of the features, and then at the very end, one time, check that scopes are satisfied? Reviewing on phone so hard to see for sure, but I think that could remove a lot of the code duplication, and give a single error message with all required scopes at once, couldn't it?

PR #7373, opened 2024-11-07
fix(d2g): perform all scope validations once

matt-boris:
Cleanup from https://github.com/taskcluster/taskcluster/pull/7343.

>D2G: `ConvertScopes()` checks all scopes at once so users would see all missing scopes in one run.


lotas (approved):
nice! looks even better

PR #7377, opened 2024-11-12
Add authentication to websockets

### LINKED ISSUE #3823: web-server provides access to pulse messages in private deployments
As part of building a private deployment, we will want to prevent the web-server process from sharing pulse messages with the UI.

We use pulse messages in two ways:
 * pulse inspector (arbitrary messages)
 * to subscribe to task-group or task status updates

Let's tackle those separately.

thakkarujas5:
Fixes #3823

### DIFF
--- a/changelog/issue-3823.md
+++ b/changelog/issue-3823.md
@@ -1,0 +1,5 @@
+audience: developers
+level: patch
+reference: issue 3823
+---
+Add authentication to websockets at the time of subscribing to pulse messages
--- a/changelog/issue-3823.md
+++ b/changelog/issue-3823.md
@@ -1,5 +1,5 @@
-audience: developers
-level: patch
+audience: users
+level: major
 reference: issue 3823
 ---
 Add authentication to websockets at the time of subscribing to pulse messages
--- a/services/web-server/src/servers/createSubscriptionServer.js
+++ b/services/web-server/src/servers/createSubscriptionServer.js
@@ -2,17 +2,67 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
 import { execute, subscribe } from 'graphql';
 import credentials from './credentials.js';
 import formatError from './formatError.js';
+import scopeUtils from 'taskcluster-lib-scopes';
+import taskcluster from 'taskcluster-client';
+import { decryptToken } from './decryptToken.js';
+import { ErrorReply } from 'taskcluster-lib-api';
 
-export default ({ server, schema, context, path }) =>
+// TODO: Check for expiration of access token when the websocket connection is active
+export default ({ cfg, server, schema, context, path }) => {
+
+  let disconnectTimeout;
   SubscriptionServer.create(
     {
       schema,
       execute,
       subscribe,
-      onConnect(params, socket) {
-        return new Promise(resolve => {
-          credentials()(socket.upgradeReq, {}, resolve);
+      async onConnect(params, socket) {
+
+        disconnectTimeout = setTimeout(() => {
+          if (socket && socket.readyState !== socket.CLOSING && socket.readyState !== socket.CLOSED) {
+            socket.close();
+          }
+        }, cfg.server.socketAliveTimeoutMilliSeconds);
+
+        return new Promise((resolve, reject) => {
+          credentials()(socket.upgradeReq, {}, async () => {
+            try {
+
+              const credentials = params?.Authorization ? decryptToken(params.Authorization) : null;
+
+              const authClient = new taskcluster.Auth({
+                rootUrl: cfg.taskcluster.rootUrl,//process.env['TASKCLUSTER_ROOT_URL'],
+                credentials: credentials,
+              });
+
+              const scopes = await authClient.currentScopes();
+              const satisfyingScopes = scopeUtils.scopesSatisfying(scopes.scopes, { AllOf: ['web:read-pulse'] });
+
+              if (!satisfyingScopes) {
+
+                const message = ([
+                  `This request requires Taskcluster credentials that satisfy the following scope: `,
+                  '',
+                  '```',
+                  'web:read-pulse',
+                  '```',
+                ]).join('\n');
+
+                return reject(new ErrorReply({
+                  code: 'InsufficientScopes',
+                  message: message,
+                }));
+              }
+
+              resolve();
+            } catch (err) {
+              reject(err);
+            }
+          });
         });
+      },
+      onDisconnect(socket) {
+        clearTimeout(disconnectTimeout);
       },
       async onOperation(message, connection) {
         // formatResponse should be replaced when
@@ -34,3 +84,4 @@ export default ({ server, schema, context, path }) =>
       path,
     },
   );
+};
--- a/services/web-server/src/servers/createSubscriptionServer.js
+++ b/services/web-server/src/servers/createSubscriptionServer.js
@@ -41,7 +41,9 @@ export default ({ cfg, server, schema, context, path }) => {
               if (!satisfyingScopes) {
 
                 const message = ([
-                  `This request requires Taskcluster credentials that satisfy the following scope: `,
+                  `Error: InsufficientScopes`,
+                  '',
+                  `Client ID ${credentials?.clientId ?? 'anonymous'} does not have sufficient scopes and is missing the following scopes:`,
                   '',
                   '```',
                   'web:read-pulse',
@@ -51,6 +53,9 @@ export default ({ cfg, server, schema, context, path }) => {
                 return reject(new ErrorReply({
                   code: 'InsufficientScopes',
                   message: message,
+                  details: {
+                    required: ['web:read-pulse'],
+                  }
                 }));
               }
 
--- a/services/web-server/src/servers/createSubscriptionServer.js
+++ b/services/web-server/src/servers/createSubscriptionServer.js
@@ -6,7 +6,6 @@ import scopeUtils from 'taskcluster-lib-scopes';
 import { decryptToken } from './decryptToken.js';
 import { ErrorReply } from 'taskcluster-lib-api';
 
-// TODO: Check for expiration of access token when the websocket connection is active
 export default ({ cfg, server, schema, context, path, authFactory }) => {
 
   const timeoutMap = new WeakMap();
@@ -59,7 +58,6 @@ export default ({ cfg, server, schema, context, path, authFactory }) => {
 
               resolve();
             } catch (err) {
-              console.log(err);
               reject(err);
             }
           });
--- a/services/web-server/src/servers/createSubscriptionServer.js
+++ b/services/web-server/src/servers/createSubscriptionServer.js
@@ -3,14 +3,14 @@ import { execute, subscribe } from 'graphql';
 import credentials from './credentials.js';
 import formatError from './formatError.js';
 import scopeUtils from 'taskcluster-lib-scopes';
-import taskcluster from 'taskcluster-client';
 import { decryptToken } from './decryptToken.js';
 import { ErrorReply } from 'taskcluster-lib-api';
 
 // TODO: Check for expiration of access token when the websocket connection is active
-export default ({ cfg, server, schema, context, path }) => {
+export default ({ cfg, server, schema, context, path, authFactory }) => {
 
-  let disconnectTimeout;
+  const timeoutMap = new WeakMap();
+
   SubscriptionServer.create(
     {
       schema,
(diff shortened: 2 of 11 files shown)


services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -2,16 +2,48 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
...
+          credentials()(socket.upgradeReq, {}, async () => {
+            try {
+
+              const token = params.Authorization;
+
+              if (!token) {
+                reject();
+              }
+
+              const credentials= JSON.parse(Buffer.from(token.split(' ')[1], 'base64').toString());
```
lotas, line 26: Check `credentials()` middleware - I think it is already doing some of this - extracts authorization from request and assigns it to `request.credentials`
And then it should likely be reachable at `socket.upgradeReq.credentials`

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -2,16 +2,48 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
...
+      async onConnect(params, socket) {
+
+        return new Promise((resolve, reject) => {
+          credentials()(socket.upgradeReq, {}, async () => {
+            try {
+
+              const token = params.Authorization;
+
+              if (!token) {
+                reject();
```
lotas, line 23: I don't think it would be necessary here, as we should rely on scopes. And for `anonymous` role on public deployments users should be able to subscribe still.

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -2,16 +2,48 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
...
+              const authClient = new taskcluster.Auth({
+                rootUrl: process.env['TASKCLUSTER_ROOT_URL'],
+                credentials: credentials
+              });
+
+              const scopes = await authClient.currentScopes();
+              const expandedScopes = await authClient.expandScopes(scopes);
+
+              if (!scopeUtils.satisfiesExpression(scopes.scopes, { AllOf: ['web:read-pulse:*'] })) {
+                await authClient.disableClient(token.clientId);
```
lotas, line 37: That would be too harsh :) If you can't access some parts of application it doesn't mean you should be banned immediately ;)

services/web-server/src/servers/createSubscriptionServer.js
thakkarujas5: If client id's are associated with scopes and roles, then how is anonymous role determined without a client id?

services/web-server/src/servers/createSubscriptionServer.js
thakkarujas5: Removed this.

services/web-server/src/servers/createSubscriptionServer.js
thakkarujas5: Written the code for decrypting base64 in a separate function.

services/web-server/src/servers/createSubscriptionServer.js
lotas: good question. To my knowledge UI and clients do not send this header if the request is "public" (requires no authentication). And it is then up to the auth service to decide if credentials were included, and if not assume it is an anonymous role. 
So I think it can happen that this code will not see any tokens at all for public deployments, so we should not reject here

lotas: https://github.com/taskcluster/taskcluster/blob/ded37fc2628f6c9ac291a6969d5ddec117615df5/services/auth/src/signaturevalidator.js#L159

Okay, maybe it's not as trivial, and maybe we'll have to fetch user scopes first and see if those include anonymous or not, and if not manually expand those with anonymous before we compare

ui/src/App/index.jsx
```diff
@@ -89,6 +89,13 @@ export default class App extends Component {
     options: {
       reconnect: true,
       lazy: true,
+      connectionParams: async () => {
+        const user = await this.authController.getUser();
+
+        return {
+          Authorization: `Bearer ${btoa(JSON.stringify(user.credentials))}`,
```
lotas, line 96: can you please check if this is empty or contains something when user is not logged in?

services/web-server/src/servers/createSubscriptionServer.js
lotas: ah, actually, if you just load (without auth) https://community-tc.services.mozilla.com/api/auth/v1/scopes/current  you would see that it includes anonymous already, so calling `auth.currentScopes()` should be enough

services/web-server/src/servers/createSubscriptionServer.js
lotas: great, thanks!

ui/src/App/index.jsx
thakkarujas5: Resolved: Added a null check for this.

services/web-server/src/servers/createSubscriptionServer.js
thakkarujas5: Have added a check, if the token is not available; the authClient initialization will take credentials parameter as null and on using this to call currentScopes will give us the scopes of anonymous role.

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -2,16 +2,55 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
 import { execute, subscribe } from 'graphql';
 import credentials from './credentials.js';
 import formatError from './formatError.js';
+import scopeUtils from 'taskcluster-lib-scopes';
+import taskcluster from 'taskcluster-client';
+import { decryptToken } from './decryptToken.js';
+import { ErrorReply } from 'taskcluster-lib-api';
 
+// TODO: Check for expiration of access token when the websocket connection is active
```
lotas, line 10: I wonder if client does reconnect automatically in case the connection drops?
This we we could simply put a timeout and force to reauthenticate every few hours or so, would be a cheap hacky way to avoid making this code more complex

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -2,16 +2,55 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
...
+
+              const credentials = params?.Authorization ? decryptToken(params.Authorization) : null;
+
+              const authClient = new taskcluster.Auth({
+                rootUrl: process.env['TASKCLUSTER_ROOT_URL'],
+                credentials: credentials,
+              });
+
+              const scopes = await authClient.currentScopes();
+              const satisfyingScopes = scopeUtils.scopesSatisfying(scopes.scopes, { AllOf: ['web:read-pulse:*'] });
```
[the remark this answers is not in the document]
thakkarujas5: Done

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -2,16 +2,55 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
...
+
+                const message = ([
+                  `This request requires Taskcluster credentials that satisfy the following scope: `,
+                  '',
+                  '```',
+                  'web:read-pulse:*',
+                  '```',
+                ]).join('\n');
+
+                reject(new ErrorReply({
```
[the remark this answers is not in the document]
thakkarujas5: Done

thakkarujas5:
> this looks great, thanks @thakkarujas5
> 
> Can you please check if this also works for the task/task group subscription page as well? When you load a task or task group, page subscribes for updates. I think it should use same functions, so should be covered already

Working as expected.

services/web-server/src/servers/createSubscriptionServer.js
thakkarujas5: For this I added a timeout and closed the connection. The connection was being retried but was failing to connect afterward. Still looking into this.

thakkarujas5: what should be the ideal number of hours for the timeout to configure?

thakkarujas5:
Thanks for the review @lotas.

The web server check is failing, what could be the issue here?

services/web-server/src/servers/createSubscriptionServer.js
thakkarujas5: Added the timeout in config.yml 👍

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -2,16 +2,59 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
...
       schema,
       execute,
       subscribe,
-      onConnect(params, socket) {
-        return new Promise(resolve => {
-          credentials()(socket.upgradeReq, {}, resolve);
+      async onConnect(params, socket) {
+
+        setTimeout(() => {
+          socket.close();
```
lotas, line 20: does it  need a check if socket is closed already? 
Maybe we need to handle this to avoid throwing unnecessary errors. 

Also, maybe it would be good alternatively to listen to socket disconnect event and clear this timeout? 

Just a thought

services/web-server/src/servers/createSubscriptionServer.js
lotas: Thanks!

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -2,18 +2,68 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
...
     {
       schema,
       execute,
       subscribe,
-      onConnect(params, socket) {
-        return new Promise(resolve => {
-          credentials()(socket.upgradeReq, {}, resolve);
+      async onConnect(params, socket) {
+
+        disconnectTimeout = setTimeout(() => {
```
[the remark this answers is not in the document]
thakkarujas5: added weakmap 👍

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -2,18 +2,75 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
...
+        }, cfg.server.socketAliveTimeoutMilliSeconds);
+        timeoutMap.set(socket, disconnectTimeout);
+
+        return new Promise((resolve, reject) => {
+          credentials()(socket.upgradeReq, {}, async () => {
+            try {
+
+              const credentials = params?.Authorization ? decryptToken(params.Authorization) : null;
+
+              const authClient = authFactory(credentials);
```
lotas, line 34: I think you need to pass object, since factory expects it in this format, or just change to single argument
```suggestion
              const authClient = authFactory({ credentials });
```

thakkarujas5: Resolved 👍

services/web-server/test/graphql/artifacts_graphql_test.js
```diff
@@ -9,6 +9,19 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
   helper.withServer(mock, skipping);
   helper.resetTables(mock, skipping);
 
+  suiteSetup(function() {
+
+    helper.load.cfg('taskcluster.rootUrl', 'http://taskcluster');
```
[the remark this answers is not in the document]
thakkarujas5: Resolved 👍

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -2,18 +2,75 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
 import { execute, subscribe } from 'graphql';
 import credentials from './credentials.js';
 import formatError from './formatError.js';
+import scopeUtils from 'taskcluster-lib-scopes';
+import { decryptToken } from './decryptToken.js';
+import { ErrorReply } from 'taskcluster-lib-api';
+
+// TODO: Check for expiration of access token when the websocket connection is active
```
lotas, line 9:
```suggestion
```

thakkarujas5: removed

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -2,18 +2,75 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
...
+                  message: message,
+                  details: {
+                    required: ['web:read-pulse'],
+                  },
+                }));
+              }
+
+              resolve();
+            } catch (err) {
+              console.log(err);
```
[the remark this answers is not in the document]
thakkarujas5: removed

changelog/issue-3823.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
+level: patch
```
lotas, line 2: I wonder if we should put `major` here, since anyone who runs public deployment would have to update the anonymous role to include this new scope. Otherwise it will break some functionality if role is not present in anonymous

@petemoore do you agree?

changelog/issue-3823.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
```
lotas, line 1:
```suggestion
audience: users
```

thakkarujas5: Added

djmitche (commented):
I don't remember enough of how web-server works to give this a real "review" but nothing I see in the diff seems wrong to me.

lotas (approved):

PR #7412, opened 2024-12-07
fix(queue): Add expiration check for artifacts

### LINKED ISSUE #4086: Return 404 from queue.getArtifact when an artifact has expired
Artifacts have an expiration time, and a periodic task deletes them after that time.  As a result, it's possible to find an artifact in the database with an expiration time in the past.  The queue service's `getArtifact` method should return a 404 in this case.

To get started on this, set up your development environment and then try to reproduce this situation in a test in `services/queue/test/artifact_test.js`.  Once you've done that -- created a failing test -- look at the implementation of `getArtifact` in services/queue/src/artifacts.js` and make a minimal change to make the test pass.

thakkarujas5:
Fixes #4086

### DIFF
--- a/changelog/issue-4086.md
+++ b/changelog/issue-4086.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 4086
+---
+`queue.getArtifact()` checks if artifact is expired and returns `ResourceExpired - 410` in such cases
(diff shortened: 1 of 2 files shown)


lotas (approved):
This is great, thank you!

changelog/issue-4086.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 4086
+---
+Add check for expiration of artifact in getArtifact API
```
lotas, line 5:
```suggestion
`queue.getArtifact()` checks artifact for expiration and returns `ResourceExpired - 410` in such cases
```

thakkarujas5: fixed 👍

PR #7418, opened 2024-12-13
feat(d2g): disable native payloads

matt-boris:
>Generic Worker: adds `disableNativePayloads` (default: `false`) worker config option (`linux` only) to require all task payloads to be Docker Worker payloads. If this option is set to `true`, the task log will no longer contain the translated task definition and the warning about using Docker Worker payloads.
>
>Tasks submitted with native payloads will be resolved as `exception/malformed-payload`.
>
>Generic Worker: adds `d2gConfig.logTranslation` (default: `true`) worker config to control whether the D2G-translated task definition is logged to the task logs.

### DIFF
--- a/workers/generic-worker/payload_linux.go
+++ b/workers/generic-worker/payload_linux.go
@@ -84,18 +84,20 @@ func (task *TaskRun) convertDockerWorkerPayload() *CommandExecutionError {
 		return executionError(internalError, errored, fmt.Errorf("could not convert task definition from JSON to YAML: %v", err))
 	}
 
-	task.Warn("This task was designed to run under Docker Worker. Docker Worker is no longer maintained.")
-	task.Warn("In order to execute this task, it is being converted to a Generic Worker task, using the D2G")
-	task.Warn("utility (Docker Worker 2 Generic Worker):")
-	task.Warn("    https://github.com/taskcluster/taskcluster/tree/main/clients/client-shell#translating-docker-worker-task-definitionpayload-to-generic-worker-task-definitionpayload")
-	task.Warn("")
-	task.Warn("We recommend that you convert all your Docker Worker tasks to Generic Worker tasks, to ensure")
-	task.Warn("continued support. For this task, see the converted payload below. If you have many tasks that")
-	task.Warn("require conversion, consider using the d2g tool (above) directly. It simply takes a Docker")
-	task.Warn("Worker task payload as input, and outputs a Generic Worker task payload. It can also convert")
-	task.Warn("Docker Worker scopes to equivalent Generic Worker scopes.")
-	task.Warn("")
-	task.Warn("Converted task definition (conversion performed by d2g):\n---\n" + text.Indent(string(d2gConvertedTaskDefinitionYAML), "  "))
+	if !config.PublicPlatformConfig.NativePayloadsDisabled() {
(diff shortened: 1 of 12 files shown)


workers/generic-worker/payload_linux.go
```diff
@@ -84,18 +84,20 @@ func (task *TaskRun) convertDockerWorkerPayload() *CommandExecutionError {
...
+		task.Warn("utility (Docker Worker 2 Generic Worker):")
+		task.Warn("    https://github.com/taskcluster/taskcluster/tree/main/clients/client-shell#translating-docker-worker-task-definitionpayload-to-generic-worker-task-definitionpayload")
+		task.Warn("")
+		task.Warn("We recommend that you convert all your Docker Worker tasks to Generic Worker tasks, to ensure")
+		task.Warn("continued support. For this task, see the converted payload below. If you have many tasks that")
+		task.Warn("require conversion, consider using the d2g tool (above) directly. It simply takes a Docker")
+		task.Warn("Worker task payload as input, and outputs a Generic Worker task payload. It can also convert")
+		task.Warn("Docker Worker scopes to equivalent Generic Worker scopes.")
+		task.Warn("")
+		task.Warn("Converted task definition (conversion performed by d2g):\n---\n" + text.Indent(string(d2gConvertedTaskDefinitionYAML), "  "))
```
lotas, line 99: I can imagine this might still be useful to show what is running underneath? Or was it some sort of a security concern not to show the original command?

lotas (approved):
👏

workers/generic-worker/payload_linux.go
matt-boris: Yeah I debated what to do here. I think for sure, if disable native payloads is set to true, they definitely shouldn't see the warning messages about docker worker payloads being entered. I could add an additional config toggle to show/hide the d2g-converted task definition. Fuzzing have asked for this, I see it to be useful here now. Thanks for calling out.

matt-boris: Addressed in https://github.com/taskcluster/taskcluster/pull/7418/commits/aa44f1ee2abae8dda2e67ef598bfbad9373b627a

workers/generic-worker/payload_test.go
```diff
@@ -148,6 +148,7 @@ func NowMillis(t *testing.T) (now time.Time) {
 
 // If an artifact expires before task deadline we should get a Malformed Payload
 func TestArtifactExpiresBeforeDeadline(t *testing.T) {
+	setup(t)
```
petemoore, line 151: Do you remember why this was needed, just out of interest?

petemoore (approved):
Looks great!

workers/generic-worker/payload_test.go
matt-boris: Yes! Good question. When I pulled these variables out into a different block https://github.com/taskcluster/taskcluster/pull/7418/files#diff-4d9f4299686b08adfe1957964bfc775633d85db82da2a5d71733bb3f41bcdefaR631-R632 I was getting nil pointer exceptions because the config was never set. I'm not sure why `setup(t)` was not called for these tests, but they are now, and things still look good 👍🏻

petemoore: I think because these were more like unit tests rather than integration tests, i.e. they didn't require a full system setup. But maybe now they do, for some reason.

PR #7419, opened 2024-12-14
feat(db): Add audit history

### LINKED ISSUE #5438: enhancement: add history of when a given secret was changed to the secrets UI
Use case: Releng was asked if we knew when a given secret was last changed. We didn't, but Cloudops was able to retrieve this information from the logs.

This could be as simple as "this secret was last changed on `datestring`" (though we may have to store this information in the db if it's not there already).

Bonus points for being able to view a history of the last `n` changes to a given secret, or being able to blame the `clientId` responsible.

thakkarujas5:
Fixes #5438


lotas:
Few brainstorming ideas: https://github.com/taskcluster/taskcluster/issues/5438#issuecomment-2567474498 :)

thakkarujas5:
@lotas Added audit in the following API's inside auth service:

1) Create Role
2) Update Role
3) Delete Role
4) Create Client
5) Update Client
6) Delete Client
7) Enable Client
8) Disable Client
9) Reset Access Token

db/versions/0105.yml
```diff
@@ -0,0 +1,30 @@
...
+          entity_type,
+          client_id,
+          action_type,
+          created
+        ) VALUES (
+          entity_id_in,
+          entity_type_in,
+          client_id_in,
+          action_type_in,
+          created_in
```
lotas, line 28: I guess we might also just use `now()` to avoid passing extra timestamp, unless we want to make sure that this timestamp is identical to the one being saved with the entity itself?

db/versions/0105-migration.sql
```diff
@@ -0,0 +1,15 @@
...
+
+  CREATE TABLE audit_history (
+    entity_id text not null,
+    entity_type text not null,
+    client_id text not null,
+    action_type text not null,
+    created timestamptz not null
+  );
+
+  ALTER TABLE audit_history add primary key (entity_id, entity_type, created);
```
lotas, line 11: I would consider adding here or separately an index on client_id either now or at some later point. 
We don't know how big this table would be as for now, so maybe not an issue at this point

services/auth/src/api.js
```diff
@@ -252,6 +252,9 @@ builder.declare({
   let clientId = req.params.clientId;
   let input = req.body;
   let scopes = input.scopes || [];
+  const created = new Date();
+  const action_type = 'created';
```
lotas, line 256: May I suggest adding some constant enum-like class to have fixed action types and allow easier refactoring later?
`const action_type = AUDIT_ENTRY_TYPE.CREATED` or something like this

db/versions/0105-migration.sql
thakkarujas5: One question, generally when we are searching we would be fetching history against the entity_id and entity_type, shouldn't we index these instead of client_id?

db/versions/0105.yml
thakkarujas5: added default timestamp

services/auth/src/api.js
thakkarujas5: added constant object for this

db/versions/0105-migration.sql
lotas: You are correct. I guess I was thinking about the other use case where we would fetch the list of all changes particular clientId did on a system. We can ignore this for now

thakkarujas5:
Continue: Added audit history for the following API's:

10) Set secret
11) Remove secret
12) create hook
13) update hook
14) remove hook

services/secrets/src/api.js
```diff
@@ -50,10 +57,25 @@ builder.declare({
...
+    await this.db.fns.insert_secrets_audit_history(
+      name,
+      'secret',
+      await req.clientId(),
+      item === undefined ? AUDIT_ENTRY_TYPE.SECRET.CREATED : AUDIT_ENTRY_TYPE.SECRET.UPDATED
+    );
+
+    res.reply({});
+  } catch (e) {
+    throw e;
```
lotas, line 77: this is probably redundant if we don't send custom response
Maybe we don't need try/catch here at all, as api has an error-catch middleware

db/versions/0105-migration.sql
```diff
@@ -0,0 +1,15 @@
...
+
+  GRANT select, insert, update, delete ON audit_history to $db_user_prefix$_auth;
```
lotas, line 13: since hooks is using this, we'll need to grant something to the `_hooks` user. 
Plus I think we can use just a subset of permissions.

`auth` should only `select, insert, delete` (udpates should not be enabled at all as we want to ensure that records are not being changed, but we need to keep `delete`  for the expire job)
`hooks` - only `insert` 
`secrets` - only `insert`

db/versions/0105.yml
```diff
@@ -0,0 +1,99 @@
...
+          created
+        ) VALUES (
+          entity_id_in,
+          entity_type_in,
+          client_id_in,
+          action_type_in,
+          now()
+        );
+      end;
+  get_audit_history:
```
lotas, line 77: I suggest we also use pagination here to avoid unexpected "big queries" 
And for this let's also add ordering, I guess it makes sense to show most recent first, so order by created please.

services/auth/schemas/v1/get-entity-history-response.yml
```diff
@@ -0,0 +1,8 @@
+$schema: "/schemas/common/metaschema.json#"
+title:              Get Entity History Response
+description: |
+Response from getEntityHistory endpoint containing audit history entries
+type: array
+items:
+  type: string
+  description: A formatted string containing client_id, action_type and timestamp
```
lotas, line 8: is this a final version? 
Let's define an object containing all those three properties and nothing else

services/auth/src/api.js
```diff
@@ -301,6 +327,36 @@ builder.declare({
...
+
+  const auditHistory = await this.db.fns.get_audit_history(
+    entityId,
+    entityType,
+  );
+
+  const result = [];
+
+  auditHistory.forEach((entry) => {
+    result.push(`${entry.client_id} ${entry.action_type} ${entityId} at ${entry.created}`);
```
lotas, line 355: let's return normal array of objects and don't format it
Would be more useful for clients

thakkarujas5: API now returns continuation token along with array of objects.

services/auth/schemas/v1/get-entity-history-response.yml
thakkarujas5: updated the schema to return array of objects and continuation token

db/versions/0105.yml
thakkarujas5: changed the stored function to support pagination and ordering

services/auth/src/api.js
```diff
@@ -301,6 +336,35 @@ builder.declare({
   return res.reply(result);
 });
 
+builder.declare({
+  method: 'get',
+  route: '/audit/:entityType/:entityId',
+  query: {
+    prefix: /^[A-Za-z0-9!@/:.+|_-]+$/, // should match clientId above
```
lotas, line 343: "above"? :)
```suggestion
```

services/auth/src/main.js
```diff
@@ -227,6 +227,15 @@ const load = Loader({
       });
     },
   },
+
+  'purge-audit-history': {
+    requires: ['db', 'monitor'],
+    setup: ({ db, monitor }, ownName) => {
+      return monitor.oneShot(ownName, async () => {
+        await db.fns.purge_audit_history(fromNow('-30d'));
```
lotas, line 235: can we please add it to the `config.yml`?  `auditHistoryRetentionDays` or something

services/auth/src/api.js
thakkarujas5: removed

services/auth/src/main.js
thakkarujas5: added to config.yml

services/auth/src/api.js
```diff
@@ -301,6 +336,34 @@ builder.declare({
...
+  name: 'getEntityHistory',
+  category: 'Audit',
+  output: 'get-entity-history-response.yml',
+  scopes: 'auth:audit-history:<entityType>',
+  stability: 'stable',
+  title: 'Get Entity History',
+  description: [
+    'Get entity history based on entity type and entity name',
+  ].join('\n'),
+}, async function(req, res) {
```
lotas, line 354: I think we need to be strict here about what types of entities this endpoint supports.
Someone might want to trigger this method for anything, like worker, queue, etc and if this doesn't return an error, it might look like there's just no entries. 

We can either add `params: { entityType: /(supported|list|entities)/ } ` 
or just check in the body first thing, and send back `404` no found or `400` invalid request argument. maybe 404 sounds better

changelog/issue-5438.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: major
+reference: issue 5438
+---
+Introduce audit history for clients, roles, secrets, hooks
```
lotas, line 5:
```suggestion
Added audit history tracking for clients, roles, secrets, and hooks. 
History can be queried using `auth.getEntityHistory(type, entityId)` and is retained for 30 days.
```
maybe something like this

db/versions/0105.yml
```diff
@@ -0,0 +1,114 @@
...
+          action_type_in,
+          now()
+        );
+      end
+  insert_secrets_audit_history:
+    description: |
+      Insert an audit history entry for a given entity.
+    mode: write
+    serviceName: secrets
+    args: entity_id_in text, entity_type_in text, client_id_in text, action_type_in text
```
lotas, line 36: I think we should drop `entity_type_in` parameter from specialized functions. This way secret service wouldn't be adding values for unrelated entities. 
Same with hooks one. 
It looks a bit strange

db/versions/0105.yml
```diff
@@ -0,0 +1,114 @@
...
+      begin
+        INSERT INTO audit_history (
+          entity_id,
+          entity_type,
+          client_id,
+          action_type,
+          created
+        ) VALUES (
+          entity_id_in,
+          entity_type_in,
```
lotas, line 71: same as function above, let's hardcode entity type in here to avoid confusions, and also we could name `entity_id_in` to `hook_id_in` to make it clear

services/auth/config.yml
```diff
@@ -115,6 +117,9 @@ defaults:
   # new relic config (processed in services/prelude.js)
   newRelic: !env:json:secret:optional NEW_RELIC
 
+production:
+  app:
+    auditHistoryRetentionDays:  !env AUDIT_RETENTION_DAYS
```
lotas, line 122:
```suggestion
    auditHistoryRetentionDays:  !env:optional AUDIT_RETENTION_DAYS
```

I think it should be optional since we already have a default value for it

services/auth/config.yml
```diff
@@ -33,6 +33,8 @@ defaults:
     # Similarly, delay before deleting expired clients
     clientExpirationDelay:      '-1 hour'
 
+    auditHistoryRetentionDays:  '-30d'
```
lotas, line 36: Maybe we need to change names here or the values that we pass. 
Currently name has "days" in it but so does value. One is probably enough.

It will later be used with the helper function `fromNow` which understands more than just days, so someone could pass `-1h` as a value, and name would be slightly misleading. 
Or someone might think that since it is days, I'll just pass `30` it will throw an error. 

Probably just `auditHistoryRetention` or `auditHistoryExpiration` would be a less confusing name

services/auth/src/main.js
```diff
@@ -227,6 +229,16 @@ const load = Loader({
       });
     },
   },
+
+  'purge-audit-history': {
+    requires: ['cfg', 'db', 'monitor'],
+    setup: ({ cfg, db, monitor }, ownName) => {
+      return monitor.oneShot(ownName, async () => {
+        console.log("AUDIT_ENV: ", cfg.app.auditHistoryRetentionDays);
```
lotas, line 237:
```suggestion
```
If it's just for the debug, don't forget to remove. If you want to log something just use `monitor`

services/auth/src/monitor.js
```diff
@@ -0,0 +1,17 @@
+import { MonitorManager } from 'taskcluster-lib-monitor';
```
lotas, line 1: This is probably fine to keep a copy in every service, but we can also define it once in a single `builtins` place - https://github.com/taskcluster/taskcluster/blob/28c7af13a54a4309bab0f8b92a81a9d4d9c9b97d/libraries/monitor/src/builtins.js

services/hooks/src/main.js
```diff
@@ -31,7 +31,7 @@ const load = loader({
   monitor: {
     requires: ['process', 'profile', 'cfg'],
     setup: ({ process, profile, cfg }) => MonitorManager.setup({
-      serviceName: 'github',
+      serviceName: 'hooks',
```
lotas, line 34: 🎉  thanks!

services/secrets/src/api.js
```diff
@@ -50,9 +57,29 @@ builder.declare({
...
   const { name } = req.params;
   const { secret, expires } = req.body;
+
+  const [item] = await this.db.fns.get_secret(name);
+
   await this.db.fns.upsert_secret(name, this.db.encrypt({
     value: Buffer.from(JSON.stringify(secret), 'utf8'),
   }), new Date(expires));
+
+  console.log("Monitor: ", this.monitor.log);
```
lotas, line 67:
```suggestion
```

db/versions/0106-downgrade.sql
```diff
@@ -0,0 +1,8 @@
+begin
+
+    REVOKE select, insert, update, delete ON audit_history FROM $db_user_prefix$_auth;
+    REVOKE select, insert, update, delete ON audit_history FROM $db_user_prefix$_hooks;
+    REVOKE select, insert, update, delete ON audit_history FROM $db_user_prefix$_secrets;
```
lotas, line 5:
```suggestion
    REVOKE select, insert, update, delete ON audit_history FROM $db_user_prefix$_secrets;
    REVOKE select, insert, update, delete ON audit_history FROM $db_user_prefix$_worker_manager;
```

thakkarujas5: Added revoke for worker manager

services/secrets/src/api.js
thakkarujas5: removed log

services/auth/src/main.js
thakkarujas5: removed log

services/auth/config.yml
thakkarujas5: kept it as auditHistoryRetention

services/auth/config.yml
thakkarujas5: made the env optional

lotas (approved):
Awesome! Thank you very much for the work :+1:

PR #7421, opened 2024-12-16
WIP: adding types

lotas:
experimenting with auto-generated types and tsconfig support with js files


db/src/utils.js
```diff
@@ -1,53 +0,0 @@
-import _ from 'lodash';
```
lotas, line 1: not used anywhere

petemoore (approved):
I'm happy to rubber stamp this, but I can't really claim to know much about it. Maybe @matt-boris / @Eijebong could scan over it too (or someone with a bit more typescript experience than me)?

@types/globals.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module 'lodash' {
```
gregtatum, line 1: This definitely works for small dependency usage. Optionally you can add packages like: https://www.npmjs.com/package/@types/lodash to do all the heavy lifting, but it will add to your dependency list. Typescript's module resolution will find anything placed in `node_modules/@types`, which is where the dependency will be stored.

db/src/main.js
```diff
@@ -1,6 +1,6 @@
 import util from 'util';
 import chalk from 'chalk';
-import { upgrade, downgrade } from 'taskcluster-db';
+import { upgrade, downgrade } from './upgrade.js';
```
gregtatum, line 3: I'm not sure how this resolution works in the runtime code, but you can teach typescript about these types of things with the `paths` property in the tsconfig. I use it in my personal project for absolute paths, e.g. https://github.com/gregtatum/indie-web/blob/main/src/logic/app-logic.ts#L1

db/src/versions.js
```diff
@@ -25,6 +29,11 @@ const run = async command => {
   });
 };
 
+/**
+ * @param {number|string} fromVersion
+ * @param {number|string} toVersion
+ * @param {{ runGit?: boolean }} opts
```
gregtatum, line 35: `Partial<{ runGit: boolean }>` is a bit more idiomatic for options.

libraries/api/src/api.js
```diff
@@ -31,35 +30,35 @@ export default class API {
     assert(options.monitor, 'monitor option is required');
     assert(!options.referencePrefix, 'referencePrefix is now deprecated!');
 
-    options = _.defaults({}, options, {
+    options = Object.assign({}, {
```
gregtatum, line 33: I think Object spread works here and may be a big simpler.

e.g. something like:

```js
options = {
  ...options,
  inputLimit: '10mb',
  allowedCORSOrigin: '*',
}
```

Also in TS, the type inference will work better if you assign it to a new name.

```js
const resolvedOptions = {
  ...options,
  inputLimit: '10mb',
  allowedCORSOrigin: '*',
}
```

libraries/api/src/middleware/errors.js
```diff
@@ -1,5 +1,8 @@
 import { reportError } from '../error-reply.js';
 
+/**
+ * @typedef {Record<string, number>} ErrorCodes
```
gregtatum, line 4: This just defines ErrorCodes, but doesn't assign it to the variable.

```suggestion
 * @type {Record<string, number>}
```

libraries/api/src/middleware/queries.js
```diff
@@ -1,17 +1,22 @@
...
-
 /**
  * Validate query-string against query.
  *
  * Query-string options not specified in options will not be allowed. But it's
  * optional if a request carries any query-string parameters at all.
+ *
+ * @template {Record<string, any>} TContext
+ * @param {{
+ *   entry: import('../../@types/index.d.ts').APIEntryOptions<TContext>,
```
gregtatum, line 9: If the imports become too cumbersome you could create a global namespace you import once through a `@typedef`, and then do something like `T.APIEntryOptions<TContext>`.

libraries/postgres/@types/fns.d.ts
```diff
@@ -0,0 +1,3393 @@
+// Generated type definitions for DB functions
```
gregtatum, line 1: If this is a third party library there may be a `@types` module available. If this is generated locally to the repo it would be nice to include instructions in the header how to regenerate it.

libraries/postgres/src/Method.js
```diff
@@ -2,10 +2,26 @@ import { strict as assert } from 'assert';
 import path from 'path';
 import { loadSql } from './util.js';
 
+/**
```
gregtatum, line 5: 👍 Just a quick note that this all looks like idiomatic JSDoc TypeScript usage.

tsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+    "checkJs": true,
+    // Only type check, don't emit files.
+    "noEmit": true,
+    // Allow esnext syntax. Otherwise the default is ES5 only.
+    "target": "esnext",
+    "lib": ["esnext", "dom"],
+    "esModuleInterop": true,
+    "moduleResolution": "NodeNext"
+  },
+  // Add a @ts-check comment to a JS file to start type checking it.
```
gregtatum, line 19: Is this statement accurate? I think it was copied from my tsconfig.json. I don't see any `@ts-check` statements in the diff.

gregtatum (commented):
Looking like a good first pass at adding types! I left a few comments here and there.

@types/globals.d.ts
lotas: Yeah, I didn't want to pull more dependencies, at least not for the lodash, as it's usage is minimal in the codebase and we are in the process of getting rid of it. Modern ES allows to achieve most of that natively.

db/src/main.js
lotas: yeah, I think this was done by mistake in the past. `package.json` uses [`link:` feature](https://github.com/taskcluster/taskcluster/blob/641db13f513c7a773bf342befdb670c5e8b803ca/package.json#L111)
here we don't even need any special named paths. 
Taskcluster services use this since all services and libraries were moved into a single monorepo many years ago to link common libraries that are also located inside same repo

libraries/api/src/api.js
lotas: Thanks! Is new name works better because TS doesn't need to merge types of two different object, and just infer from the new one?

libraries/api/src/middleware/errors.js
lotas: oh nice catch, thank you!
I think I added this to import from another file but forget to define it as type here as well. Fixed

libraries/api/src/middleware/queries.js
lotas: Oh nice! I didn't know you could "import" whole file like this with typedef 👍  cool

tsconfig.json
lotas: Ha! Well spotted, thanks. Yes, I think I copied this file from you 🙈

petemoore (approved):
Thanks @lotas and @gregtatum! No objections from my side. :-)

PR #7424, opened 2024-12-19
fix(generate): generic worker readme generation fixed to linux amd64

matt-boris:
Fixes an issue with `yarn generate` causing a diff from what's on `main`. This was due to generic worker config splitting out into linux-specific config, apart from posix.


lotas (approved):
Thanks!

Issue #7128, opened 2024-07-09
generic worker deletes file from preserved cache before task starts


bhearsum:
I've debugging a strange issue in https://github.com/mozilla/firefox-translations-training/issues/538 that occurred when we switched some workers from docker-worker to generic-worker/d2g. What we experienced was tasks intermittently failing because `run-task` (part of taskgraph) decided that their caches were corrupt and unusable. More specifically, it found a cache directory that was not empty, and didn't contain the expected `.cacherequires` file. This commonly happened when two of the recently migrated jobs ran on the same worker in a row.

After a bunch of debugging and [instrumentation](https://github.com/taskcluster/taskcluster/compare/main...bhearsum:taskcluster:cachelog?expand=1), I managed to confirm a few things:
- The cache is correctly preserved after the first task ends
- By the time the cache is mounted in the second task, `.cacherequires` has disappeared
- generic-worker is almost certainly responsible for deleting it, based on this auditd entry

A concrete example.

[Task 1](https://firefox-ci-tc.services.mozilla.com/tasks/bpOccPtrRy2ds5SPKw48Uw/runs/16), instrumented to show the state of the caches after preserving them at the end of the task. We find this at the end of the log (`cache target` is `/caches/XNBc6oDyS1mvopJKCnCc0w`):
```
[taskcluster 2024-07-08T23:47:03.163Z] === Task Finished ===
[taskcluster 2024-07-08T23:47:03.163Z] Task Duration: 28.334137312s
[taskcluster 2024-07-08T23:47:03.163Z] [mounts] before preserve
[taskcluster 2024-07-08T23:47:03.163Z] [mounts] entries in cache src:
[taskcluster 2024-07-08T23:47:03.163Z] [mounts] - .cachelog
[taskcluster 2024-07-08T23:47:03.163Z] [mounts] - .cacherequires
[taskcluster 2024-07-08T23:47:03.163Z] [mounts] d vcs/
[taskcluster 2024-07-08T23:47:03.163Z] [mounts] couldn't read entries in target
[taskcluster 2024-07-08T23:47:03.163Z] [mounts] entries in cache target:
[taskcluster 2024-07-08T23:47:03.163Z] [mounts] Preserving cache: Moving "/home/task_172048185446460/cache0" to "caches/XNBc6oDyS1mvopJKCnCc0w"
[taskcluster 2024-07-08T23:47:03.163Z] [mounts] couldn't read entries in src
[...]

PR #7149, opened 2024-07-12
fix(generic-worker)!: only chown writable directory cache files owned by previous task user to new task user

petemoore:
UPDATE: This is the third, and hopefully last part, of the fix for #7128.

The complete fix includes:
* #7170
* #7144
* #7149 (this PR)

PR #7428, opened 2024-12-24
Fix worker caches with d2g

### LINKED ISSUE #7404: Permission error on caches
Since 75.0.0, I'm getting frequent permission errors when using caches.

```
[cache 2024-12-01T18:31:58.737Z] cache /builds/worker/checkouts exists; requirements: gid=1000 uid=1000 version=1
Traceback (most recent call last):
  File "/usr/local/bin/run-task", line 1359, in <module>
    sys.exit(main(sys.argv[1:]))
             ^^^^^^^^^^^^^^^^^^
  File "/usr/local/bin/run-task", line 1182, in main
    purge = configure_cache_posix(
            ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/bin/run-task", line 503, in configure_cache_posix
    write_audit_entry(audit_path, b"used")
  File "/usr/local/bin/run-task", line 329, in write_audit_entry
    with open(path, "ab") as fh:
         ^^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: '/builds/worker/checkouts/.cachelog'
```

An example task: https://taskcluster.bananium.fr/tasks/HfpTR10xQWqJTuDVJq-PWg/runs/0

Eijebong:
Force all task users to share the same subuid map

This fixes the cache issues we've been having with podman with #7149
applied.

The core idea here is that on podman, caches permissions for UIDs
(outside of 1000) are mapped to the user subuid range on the host. So UID 0 becomes
100000, 1 becomes 100001 and so on. When remounting the cache in the
next container, the reverse operation happens to get the right UID in
the container, 100000 on the host becomes 0 in the container.

However, by default, users do not share subuid ranges, `adduser` will
always use a new subuid range (100000 for the first user, +64k for the
second and so on). Which means that the mapping wouldn't work on the
second job and the cache would just get mounted as `nobody:nogroup`.

This commit changes the user creation so that every task user shares the
same subuid range.

Fixes #7404

---

First commit is #7149 re-applied without the changelog file.

### DIFF
--- a/changelog/issue-7404.md
+++ b/changelog/issue-7404.md
@@ -1,0 +1,6 @@
+audience: users
+level: minor
+reference: issue 7404
+---
+Re-apply the patch to fix docker cache issues and fix the issues when using
+podman as the container engine.
--- a/workers/generic-worker/mounts.go
+++ b/workers/generic-worker/mounts.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"os/user"
 	"path/filepath"
 	"sort"
 	"time"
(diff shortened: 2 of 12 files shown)


Eijebong:
CC @bhearsum

bhearsum:
I gave this is a try out with Translations (using the docker container engine) and it seems to work fine!

Eijebong:
I switched all my workers to this this weekend and have not seen a single cache related error. (90% podman tasks)

bhearsum:
This might also be the last piece of the puzzle for https://github.com/taskcluster/taskcluster/pull/7428?

Eijebong:
Did you mean to link another issue?

bhearsum:
> Did you mean to link another issue?

Indeed! https://github.com/taskcluster/taskcluster/issues/7128 is what was intended; I'll update the earlier comment as well.

Eijebong:
I don't think I've ever seen #7128 happen. But now that I read about it I don't understand why...

bhearsum:
> I don't think I've ever seen #7128 happen. But now that I read about it I don't understand why...
> 
> Edit: Oh it's because you did end up changing it to `--remove-home`, duh

Yeah, I guess that issue looks tangential on the surface, too. After we dug in and realized it was cache related it ended up being a place to track other related things too...

Eijebong:
Yeah I think after this we can close #7128 I don't know of any other issue related to caches and if they arise we can track them on their own.

matt-boris (approved):
LGTM, thanks Bastien for tackling and testing this! 🎉

Will you please rebase with `main` and we can get this merged?

PR #7439, opened 2025-01-07
feat: Cloud Armor and BackendConfig

bkochendorfer:
Allow attaching a Cloud Armor Policy to a BackendConfig and assigning that BackendConfig as the Default for the ingress


bqbn:
Related to https://github.com/mozilla-services/cloudops-infra/pull/6172

bkochendorfer:
I'm not sure I am following the test failures here. Any tips would be appreciated :smiley:

lotas:
@bkochendorfer I'll check. There is a bit of an automation happening for k8s templates in taskcluster repo - we autogenerate them based on services definitions. 
Let me check how to correctly apply it.

lotas:
I had to modify our template generators a bit to work around some nuances of this process. 

We use services definitions to generate all necessary configs for each. And we use `jsone` to pre-process it, which can't handle go templates which helm charts are using, so there is a bit of magic around this part

lotas (approved):

lotas (approved):

PR #7446, opened 2025-01-09
fix(generic-worker): give task user access to readonly mounts

matt-boris:
Fixes an issue that was brought up with the initial commit in https://github.com/taskcluster/taskcluster/pull/7428. I'm not sure why this code was pulled out.

>Generic Worker: fixes permissions issues with ReadOnlyDirectory mounts.

### DIFF
--- a/changelog/RretABTnSM6y7D-IgcBNbA.md
+++ b/changelog/RretABTnSM6y7D-IgcBNbA.md
@@ -1,0 +1,4 @@
+audience: worker-deployers
+level: patch
+---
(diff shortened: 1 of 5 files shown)


petemoore (approved):
I think we can go with this now to unblock the current issue, but we should follow up with an issue that something unknown is going wrong. Archive files should be extracted as the task user, and thus all files they contain should also be owned by the task user. The fact we have to change ownership suggests that files are not being extracted as the task user, which is worrying.

Issue #6916, opened 2024-03-19
Migrate `taskcluster-dev` GCP project to use Artifact Registry


matt-boris:
The project currently uses GCP's Container Registry during the [cloudbuild](https://github.com/taskcluster/taskcluster/blob/main/cloudbuild.yaml) for [Taskcluster Dev](https://dev.alpha.taskcluster-dev.net/), which was deprecated on May 15, 2023, and will be fully replaced by Artifact Registry on March 18, 2025. GCP laid out a nice set of documentation to follow for the migration [here](https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr).

PR #7449, opened 2025-01-10
chore: remove delete old image step in gcp cloudbuild

matt-boris:
I've migrated our Container Registry usage over to Artifact Registry and found this small cleanup we can add now that I've added a cleanup policy to our Artifact Registry repository:
Screenshot 2025-01-10 at 3 03 23 PM

Screenshot 2025-01-10 at 3 06 32 PM

Fixes https://github.com/taskcluster/taskcluster/issues/6916.

### DIFF
--- a/changelog/VujHkx4yTsCmExAvFY_ZKQ.md
+++ b/changelog/VujHkx4yTsCmExAvFY_ZKQ.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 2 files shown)


lotas (approved):
oh nice thank you

matt-boris:
We can tweak the number of versions to keep as we see fit. I saw no harm in setting to 5 at first. Probably safe to even set to 1.

PR #7452, opened 2025-01-13
feat(w-m): Increase queueInactivityTimeout min to 1200

### LINKED ISSUE #7443: Worker-manager killing workers executing tasks because of queueInactivityTimeout
We introduced zombie workers killing in https://github.com/taskcluster/taskcluster/pull/6333 
but didn't consider the fact that really low `queueInactivityTimeout` values could kill the worker while it is working on a task.

Worker manager should additionally check if worker has zero active claims at that point. 

Another option would to always increase this value by a "default reclaim timeout", so if tasks are considered claim expired after 20 min, we could always include 20m extra

lotas:
Default claim timeout constant is set to 20 min.
Former value of minimum for queueInactivityTimeout was incorrectly set to 1min which conflicted with workers being killed while working on tasks.

Fixes #7443

### DIFF
--- a/changelog/issue-7443.md
+++ b/changelog/issue-7443.md
@@ -1,0 +1,8 @@
+audience: worker-deployers
+level: minor
+reference: issue 7443
+---
+
+Worker-pool's lifecycle `queueInactivityTimeout` minimum allowed value is increased
+to `1200` (20min) to avoid having workers being incorrectly considered idling
+while they were working on a task.
--- a/changelog/issue-7443.md
+++ b/changelog/issue-7443.md
@@ -1,5 +1,5 @@
 audience: worker-deployers
-level: minor
+level: major
 reference: issue 7443
 ---
 
--- a/generated/references.json
+++ b/generated/references.json
@@ -575,8 +575,8 @@
       "properties": {
         "queueInactivityTimeout": {
           "default": 7200,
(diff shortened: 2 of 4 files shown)


changelog/issue-7443.md
```diff
@@ -0,0 +1,5 @@
+audience: worker-deployers
+level: minor
+reference: issue 7443
+---
+<!-- replace this text with your changelog entry.  See dev-docs/best-practices/changelog.md for help writing changelog entries. -->
```
petemoore, line 5: Can you add a description here?

lotas: haha, no.. you were too fast to review, thanks
i was struggling with my shell setup, created several changelog entries, and when noticed, removed the ones that contained valid input .. 
should be fine now

changelog/issue-7443.md
```diff
@@ -0,0 +1,8 @@
+audience: worker-deployers
+level: minor
```
matt-boris, line 2: Would this technically be a major change here since if someone has their config set to something less than 1200 before this change, after the change, they'll get a schema validation error, right?

lotas: good question! I'm not sure, this is simply a validation from my perspective, so shouldn't break anything. 
Change my mind :)

matt-boris (approved):

PR #7455, opened 2025-01-16
chore: add d2g device generic worker tests

matt-boris:
https://github.com/taskcluster/community-tc-config/pull/901 will also be needed.

Uses tests from https://github.com/taskcluster/taskcluster/blob/main/workers/docker-worker/test/integration/device_link_test.js


petemoore (approved):
This is awesome, thanks so much Boris!!

PR #7457, opened 2025-01-21
Log cloud api call metrics for scanner and provisioner

lotas:
Logging cloud api call metrics for provisioner and scanner. 
This includes min/max/median values and breakdown by status.

Attempt to understand how (in)effective cloud provider calls are


matt-boris (approved):
Nice! Thanks Yarik 👍🏻

PR #7459, opened 2025-01-22
Drop users in task containers when using d2g with the interactive feature

Eijebong:
https://github.com/user-attachments/assets/6aef095a-8588-44c8-8810-2a9e82095dc7

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -337,14 +337,8 @@ func command(dwPayload *dockerworker.DockerWorkerPayload, dwImage Image, gwArtif
 	commands = append(
 		commands,
 		runString,
+		"exit_code=$?",
 	)
-
-	if containerName != "" {
-		commands = append(
-			commands,
-			"exit_code=$?",
-		)
-	}
 
 	commands = append(
 		commands,
@@ -358,13 +352,11 @@ func command(dwPayload *dockerworker.DockerWorkerPayload, dwImage Image, gwArtif
 		)
 	}
 
-	if containerName != "" {
-		commands = append(
-			commands,
-			"docker rm -v "+containerName,
-			`exit "${exit_code}"`,
-		)
-	}
+	commands = append(
+		commands,
+		"docker rm -v "+containerName,
+		`exit "${exit_code}"`,
+	)
 
 	return [][]string{
 		{
@@ -379,12 +371,8 @@ func runCommand(containerName string, dwPayload *dockerworker.DockerWorkerPayloa
 	command := strings.Builder{}
 	// Docker Worker used to attach a pseudo tty, see:
 	// https://github.com/taskcluster/taskcluster/blob/6b99f0ef71d9d8628c50adc17424167647a1c533/workers/docker-worker/src/task.js#L384
-	switch containerName {
-	case "":
-		command.WriteString("docker run -t --rm")
-	default:
-		command.WriteString(fmt.Sprintf("timeout -s KILL %v docker run -t --name %v", dwPayload.MaxRunTime, containerName))
-	}
+	command.WriteString(fmt.Sprintf("timeout -s KILL %v docker run -t --name %v", dwPayload.MaxRunTime, containerName))
+
 	// Do not limit resource usage by the containerName. See
 	// https://docs.docker.com/reference/cli/docker/container/run/
 	command.WriteString(" --memory-swap -1 --pids-limit -1")
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -33,6 +33,9 @@ type (
 		String() (string, error)
 		LoadCommands() []string
 	}
+	ConversionInfo struct {
+		ContainerName string
+	}
 )
 
 func ConvertTaskDefinition(dwTaskDef json.RawMessage, config map[string]interface{}, scopeExpander scopes.ScopeExpander) (json.RawMessage, error) {
@@ -58,7 +61,7 @@ func ConvertTaskDefinition(dwTaskDef json.RawMessage, config map[string]interfac
 		return nil, fmt.Errorf("cannot unmarshal Docker Worker payload: %v", err)
 	}
 
-	gwPayload, err := ConvertPayload(dwPayload, config)
+	gwPayload, _, err := ConvertPayload(dwPayload, config)
 	if err != nil {
 		return nil, fmt.Errorf("cannot convert Docker Worker payload: %v", err)
 	}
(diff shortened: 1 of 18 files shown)


Eijebong:
Setting as draft, I assumed docker would be a strict superset of podman utility wise and nope... It doesn't support `--condition` on `docker wait`

Eijebong:
Alright, this now works for both docker and podman, should be ready for reviews

workers/generic-worker/insecure.go
```diff
@@ -26,17 +27,35 @@ func secure(configFile string) {
...
+		cmd = []string{d2gConversionInfo.ContainerEngine, "exec", "-it", d2gConversionInfo.ContainerName, "/bin/bash"}
+	} else {
+		cmd = []string{"bash"}
+	}
+
+	return task.newCommandForInteractive(cmd, ctx)
+}
+
+func (task *TaskRun) generateInteractiveIsReadyCommand(d2gConversionInfo *d2g.ConversionInfo, ctx context.Context) (*exec.Cmd, error) {
+	cmd := []string{"/bin/bash", "-cx", "/bin/[ \"`/usr/bin/" + d2gConversionInfo.ContainerEngine + " container inspect -f '{{.State.Running}}' " + d2gConversionInfo.ContainerName + "`\" = \"true\" ]"}
```
petemoore, line 43: Are we sure podman/docker are always in `/usr/bin` ? If we want to enforce this, we should probably update docs. Isn't it sufficient that they are in `PATH`? System tools like `rm` we tend to specify absolute paths for, but I wonder if different linux distros will put them in different places, or if some workers might want to use a locally built version of podman/docker etc.

petemoore (approved):
Looks great, many thanks Bastien!

workers/generic-worker/insecure.go
Eijebong: Yeah so the issue here is that there are 2 sets of environment variables available to us. The task one which the user would be able to modify. (and `PATH="/path/to/where/my/mounts/are:..."` means they could run that interactively and bypass being dropped in the container.
The other set of env variable available at this point are the workers which might contain sensitive data?

We could just create our own set with just `PATH=/bin:/usr/bin:/sbin/usr/sbin` if you prefer but that doesn't help with non standard installation paths.

Eijebong: I guess we could take the worker environment and filter everything but the `PATH` out 🤔

petemoore: I think `/bin/bash` is launched in a subshell of generic-worker, and will inherit the PATH of the parent process, which should contain the docker/podman parent directory. Is it not sufficient to just launch `d2gConversionInfo.ContainerEngine` as a command, does the full path need to be specified? My assumption is that the worker process will be configured with a PATH containing all the tools it needs.

Eijebong: I'll try getting the path from the worker process and ignoring everything else in the env. This sounds like the safest thing

Eijebong:
Both windows tests are failing with `cgo: C compiler "gcc" not found: exec: "gcc": executable file not found in %PATH%` which I highly doubt is caused by this PR?

lotas:
> Both windows tests are failing with `cgo: C compiler "gcc" not found: exec: "gcc": executable file not found in %PATH%` which I highly doubt is caused by this PR?

yeah, this is a known issue, Boris was trying to find a fix for it. New images were build few days ago, and looks like some regions were not correctly setup.

matt-boris:
Try rebasing with `main` to trigger new tasks on that pool, hoping the updated image fixes it. Also, you'll need to pick up our updated macOS tasks (new mac pool).

tools/d2g/d2g.go
matt-boris, line 1: Now that `containerName` will always be defined, will you please remove the unnecessary `switch` statement at the top of the `runCommand()` function? 🙏🏻 

https://github.com/taskcluster/taskcluster/pull/7459/files#diff-4ecf833adef40715069e50a184adecf5e50562d65c6b45ce4ad326df97003ebfL379-L384

matt-boris: Here's another instance 🙏🏻

https://github.com/taskcluster/taskcluster/pull/7459/files#diff-4ecf833adef40715069e50a184adecf5e50562d65c6b45ce4ad326df97003ebfL339-L344

matt-boris: Final one that I see at least...

https://github.com/taskcluster/taskcluster/pull/7459/files#diff-4ecf833adef40715069e50a184adecf5e50562d65c6b45ce4ad326df97003ebfL358-L364

tools/d2g/d2gtest/testdata/testcases/devices_tests.yml
matt-boris, line 1: Thank you for going through all of these test cases... Not so fun.

workers/generic-worker/insecure.go
```diff
@@ -26,17 +27,41 @@ func secure(configFile string) {
...
+		cmd = []string{"bash"}
+	}
+
+	return task.newCommandForInteractive(cmd, env, ctx)
+}
+
+func (task *TaskRun) generateInteractiveIsReadyCommand(d2gConversionInfo *d2g.ConversionInfo, ctx context.Context) (*exec.Cmd, error) {
+	pathEnv := os.Getenv("PATH")
+	env := []string{"PATH=" + pathEnv}
+	cmd := []string{"/bin/bash", "-cx", "/bin/[ \"`/usr/bin/docker container inspect -f '{{.State.Running}}' " + d2gConversionInfo.ContainerName + "`\" = \"true\" ]"}
```
matt-boris, line 50: This is slick!

My only concern here is the path to the `docker` binary. I wonder if some workers may have a different path for some reason. Wdyt about using `which docker` instead of the hardcoded path here?

workers/generic-worker/multiuser_posix.go
```diff
@@ -50,17 +51,41 @@ func (task *TaskRun) generateCommand(index int) error {
...
+		cmd = []string{"bash"}
+	}
+
+	return task.newCommandForInteractive(cmd, env, ctx)
+}
+
+func (task *TaskRun) generateInteractiveIsReadyCommand(d2gConversionInfo *d2g.ConversionInfo, ctx context.Context) (*exec.Cmd, error) {
+	pathEnv := os.Getenv("PATH")
+	env := []string{"PATH=" + pathEnv}
+	cmd := []string{"/bin/bash", "-cx", "/bin/[ \"`/usr/bin/docker container inspect -f '{{.State.Running}}' " + d2gConversionInfo.ContainerName + "`\" = \"true\" ]"}
```
matt-boris, line 74: Same comment I made in `workers/generic-worker/insecure.go`.

matt-boris (changes requested):
This is awesome, thanks so much for taking the time to tackle this @Eijebong! 🎉

I've got a few cleanup requests and some small suggestions/questions. Thank you!

matt-boris (approved):
Great, thanks so much! 🎉

PR #7460, opened 2025-01-23
fix(generic-worker): unset unexpected cached username

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/7218.

>Generic Worker: Unset cached interactive username when we unexpectedly receive a non-task username.
>
>This will fix errors like: `interactive username gdm does not match task user task_173764785573833`.

### DIFF
--- a/workers/generic-worker/gdm3/gdm3.go
+++ b/workers/generic-worker/gdm3/gdm3.go
@@ -7,26 +7,9 @@ import (
 )
 
 var (
-	automaticLogin        = regexp.MustCompile(`^\s*AutomaticLogin\s*=`)
-	automaticLoginEnable  = regexp.MustCompile(`^\s*AutomaticLoginEnable\s*=`)
-	automaticLoginReplace = regexp.MustCompile(`^\s*AutomaticLogin\s*=\s*(\S*)\s*$`)
+	automaticLogin       = regexp.MustCompile(`^\s*AutomaticLogin\s*=`)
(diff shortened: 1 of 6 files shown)


workers/generic-worker/gdm3/gdm3.go
```diff
@@ -57,6 +52,9 @@ func SetAutoLogin(username string, source []byte) (output []byte) {
 					"# Set by generic-worker",
 					"AutomaticLoginEnable = true",
 					"AutomaticLogin = "+username,
+					"TimedLoginEnable = true",
+					"TimedLogin = "+username,
+					"TimedLoginDelay = 10",
```
lotas, line 57: does this number has any background, maybe worth living some context why `10`
do you know if it is something performance related? If we allow to run it on a slower VM, would it take longer than this for gdm to do its job and prevent race condition?

lotas (dismissed):
🤞

workers/generic-worker/runtime/runtime_other.go
```diff
@@ -54,6 +54,7 @@ func WaitForLoginCompletion(timeout time.Duration, username string) (err error)
 		}
 		if interactiveUsername != username {
 			log.Printf("WARNING: user %v appears to be logged in but was expecting %v.", interactiveUsername, username)
+			cachedInteractiveUsername = ""
```
matt-boris, line 57: This line is the fix, rest is just cleanup of unused code.

lotas (approved):
LGTM!

PR #7461, opened 2025-01-23
Fix d2g container deletion to not leak anonymous volumes

Eijebong:
Before this commit, when a docker container had a volume declared that wasn't explicitly mounted by the task, it would leak it as it would create an anonymous docker volume for every single one. This changes the `rm` command to remove those anonymous volumes.

Note that the unnamed task container path used `docker run --rm` which takes care of those anonymous volumes on its own.


petemoore (approved):
Many thanks Bastien!!

matt-boris (commented):
Awesome stuff, thanks Bastien! Glad it works on podman too 😉

Issue #7465, opened 2025-01-23
Worker Manager: HTTP 500 when removing worker and recreating


petemoore:
When working on #7464 I wanted to remove a worker and recreate it, to work around the issue with the worker config being cached at worker creation time.

I was able to remove the worker, but when recreating it, I got a HTTP 500 error.

```
pmoore@Petes-iMac:~ $ taskcluster api workerManager removeWorker proj-taskcluster/dev-windows pmoore delltower1
{}
```

```
pmoore@Petes-iMac:~ $ cat << EOF | taskcluster api workerManager createWorker proj-taskcluster/dev-windows pmoore delltower1
> {
>   "capacity": 1,
>   "expires": "2030-01-01T00:00:00Z",
>   "providerInfo": {
>     "staticSecret": "********************************************"
>   }
> }
> EOF
Error: API Error 500: InternalServerError
Internal Server Error, incidentId fb2a0325-0774-430a-8ca4-96b0fe529c81.

---

* method:     createWorker
* errorCode:  InternalServerError
* statusCode: 500
* time:       2025-01-23T13:00:15.795Z
```

lotas:
```
error: duplicate key value violates unique constraint "workers_pkey"
    at /app/node_modules/pg/lib/client.js:535:17
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Object.query (file:///app/libraries/postgres/src/Database.js:650:18)
    at async file:///app/libraries/postgres/src/Database.js:145:21
    at async Database._withClient (file:///app/libraries/postgres/src/Database.js:658:14)
    at async collection.<computed> [as create_worker] (file:///app/libraries/postgres/src/Database.js:133:21)
    at async Worker.create (file:///app/services/worker-manager/src/data.js:393:21)
    at async StaticProvider.createWorker (file:///app/services/worker-manager/src/providers/static.js:32:7)
    at async Proxy.<anonymous> (file:///app/services/worker-manager/src/api.js:596:14)
```

seems like removal doesn't immediately remove the record, but only marks as deleted.

From the docs:

    Remove an existing worker.  The precise behavior of this method depends
    on the provider implementing the given worker.  Some providers
    do not support removing workers at all, and will return a 400 error.
    Others may begin removing the worker, but it may remain available via
    the API (perhaps even in state RUNNING) afterward.

I assume you were using `static` provider, which only marks it as `STOPPED`. 

[...]

PR #7469, opened 2025-01-24
feat(w-m): createWorker correctly handles unique key errors

lotas:
Before it would return `500` if worker with the same `workerId` already existed

Related to #7465 issue


petemoore (approved):
This looks great! Can the test also check that the status code is 409?

lotas:
> This looks great! Can the test also check that the status code is 409?

Thanks, added it too!

petemoore (approved):
Perfect, many thanks!

PR #7471, opened 2025-01-24
fix(d2g): pull latest docker image tag

### LINKED ISSUE #7463: Tasks using d2g and the docker container engine never pull image tags
Since d2g doesn't clear out images between runs, tags linger and when a new task reuses an image with the same tag, the image doesn't get pulled. On a long running worker that means that someone using for example the `debian:12` image will always be stuck on the image that got pulled for the first task.

It would be nice if d2g added a `docker pull {image}` as its first command in the generated payload.

matt-boris:
Fixes #7463.

>D2G: Pull docker image as initial command to ensure latest image version is used during task execution.

### DIFF
--- a/changelog/issue-7463.md
+++ b/changelog/issue-7463.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 7463
+---
+D2G: Pull docker image as initial command to ensure latest image version is used during task execution.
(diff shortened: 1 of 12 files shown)


petemoore (approved):
lgtm, thanks boris and bastien!

Eijebong:
Deployed and tested, works perfectly, thank you

PR #7476, opened 2025-01-28
[Bug 1935142] Include URL query parameters in trigger hook API endpoints

### LINKED ISSUE #7475: Include URL parameters in triggerHookWithToken API endpoint
I'm working on Mozilla's [code review bot](https://github.com/mozilla/code-review) along with @marco-c & Aryx.

We want to use the `triggerHookWithToken` API endpoint to trigger [this hook](https://firefox-ci-tc.services.mozilla.com/hooks/project-relman/code-review-testing) by using a Phabricator build plan.

Unfortunately, Phabricator build plans do not send any JSON payload (as [found out by dkl here](https://bugzilla.mozilla.org/show_bug.cgi?id=1935142#c3)). But Phabricator can send URL query parameters instead; we have been using that feature for the past years [to react to Phabricator events](https://github.com/mozilla/libmozevent/blob/master/libmozevent/phabricator.py#L26-L29).

Would you consider a patch that extends the `triggerHookWithToken` payload with the request query parameters ?
My idea is to extends the [`payload ` there](https://github.com/taskcluster/taskcluster/blob/9a65b8323cf9685747e852d4d4f798ecd0e254e2/services/hooks/src/api.js#L557) so that it becomes:

```javascript
const payload = {...req.body, ...{req.queryParameters}}
```

This seems the least invasive option to benefit from the JSON-e validation at the next stage of the endpoint.

La0:
Bugzilla Bug: [1935142](https://bugzilla.mozilla.org/show_bug.cgi?id=1935142)

Github Issue: Fixes #7475

### DIFF
--- a/services/hooks/src/api.js
+++ b/services/hooks/src/api.js
@@ -446,7 +446,7 @@ builder.declare({
 
   await req.authorize({ hookGroupId, hookId });
 
-  const payload = { ...req.body, ...{queryParameters: req.query}};
+  const payload = { ...req.body, ...{ queryParameters: req.query } };
   const clientId = await req.clientId();
   const hook = hookUtils.fromDbRows(await this.db.fns.get_hook(hookGroupId, hookId));
 
@@ -554,7 +554,7 @@ builder.declare({
     'task template.',
   ].join('\n'),
 }, async function(req, res) {
-  const payload = { ...req.body, ...{queryParameters: req.query}};
+  const payload = { ...req.body, ...{ queryParameters: req.query } };
   const { hookGroupId, hookId } = req.params;
 
   const hook = hookUtils.fromDbRows(await this.db.fns.get_hook(hookGroupId, hookId));
--- a/services/hooks/src/api.js
+++ b/services/hooks/src/api.js
@@ -446,7 +446,7 @@ builder.declare({
 
   await req.authorize({ hookGroupId, hookId });
 
-  const payload = req.body;
+  const payload = { ...req.body, ...{queryParameters: req.query}};
   const clientId = await req.clientId();
   const hook = hookUtils.fromDbRows(await this.db.fns.get_hook(hookGroupId, hookId));
 
@@ -554,7 +554,7 @@ builder.declare({
     'task template.',
   ].join('\n'),
 }, async function(req, res) {
-  const payload = req.body;
+  const payload = { ...req.body, ...{queryParameters: req.query}};
   const { hookGroupId, hookId } = req.params;
 
   const hook = hookUtils.fromDbRows(await this.db.fns.get_hook(hookGroupId, hookId));
--- a/services/hooks/src/api.js
+++ b/services/hooks/src/api.js
@@ -446,7 +446,7 @@ builder.declare({
 
   await req.authorize({ hookGroupId, hookId });
 
-  const payload = { ...req.body, ...{ queryParameters: req.query } };
+  const payload = { ...req.body, queryParameters: req.query };
   const clientId = await req.clientId();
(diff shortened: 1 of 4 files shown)


services/hooks/src/api.js
```diff
@@ -446,7 +446,7 @@ builder.declare({
 
   await req.authorize({ hookGroupId, hookId });
 
-  const payload = req.body;
+  const payload = { ...req.body, ...{ queryParameters: req.query } };
```
lotas, line 449:
```suggestion
  const payload = { ...req.body, queryParameters: req.query };
```

I think this can be simplified a bit, or there are something less obvious happening here?

lotas (approved):
Looks great, thank you very much for the patch!
(Ignore failing g-w tests on windows, this is likely because of the recent image build)

services/hooks/src/api.js
```diff
@@ -554,7 +554,7 @@ builder.declare({
     'task template.',
   ].join('\n'),
 }, async function(req, res) {
-  const payload = req.body;
+  const payload = { ...req.body, ...{ queryParameters: req.query } };
```
lotas, line 557:
```suggestion
  const payload = { ...req.body, queryParameters: req.query };
```

services/hooks/src/api.js
La0: JS is not my strong suit :smile:

ui/docs/reference/core/hooks/firing-hooks.mdx
```diff
@@ -69,6 +69,8 @@ object is also supplied in the JSON-e context as `context`:
 }
 ```
 
+Tha payload includes a `queryParameters` mapping with optional URL query parameters.
```
vrigal, line 72:
```suggestion
The payload includes a `queryParameters` mapping with optional URL query parameters.
```
Same typo a few lines below.

lotas:
> @lotas Is there a simple way to add a unit test triggering the hook (similar to [this one](https://github.com/taskcluster/taskcluster/blob/main/services/hooks/test/api_test.js#L559)) but using a url instead of passing objects ?

I don't think there is, we test all api endpoints this way.

lotas:
So passing query would be not possible, unless it would be defined on the api level. 
@La0 are query params known in advance, or those could take any keys and values? 
If so, then let's keep those tests as is. 
If keys are known, they would have to be added to the api definition of the endpoint (`query: { list of keys and validation }`) and then it would be possible to pass them to the client like:
`await helper.hooks.triggerHookWithToken('foo', 'bar', res.token, { location: 'New Zealand' }, { a: 1 });`

La0:
Keys are not known in advance (they could be anything really), so I'll leave it as-is.

Thanks a lot for the quick replies, I just removed the draft status on this PR.

lotas (approved):

lotas:
Thanks for the patch and quick fixes! 
This will reach fxci by the time of the next scheduled release, which is usually every second Thursday, so around 7th of February. 
If this is something critical, we can discuss scheduling sooner release

La0:
Awesome, thanks for the merge.

>  This will reach fxci by the time of the next scheduled release, which is usually every second Thursday, so around 7th of February.

We are OK with this timeframe, I need to adapt the Taskcluster hook definition and client code on my side too.

--- 

Is there a Taskcluster testing instance where this will be deployed before fxci so I can make some tests in real conditions ?

lotas:
> Is there a Taskcluster testing instance where this will be deployed before fxci so I can make some tests in real conditions ?

yes, it is automatically gets deployed to our dev instance: https://dev.alpha.taskcluster-dev.net/ 
reach out to me in slack or matrix if you need account created there

and after that it will first will be deployed here: https://community-tc.services.mozilla.com/ (once we make a release)

PR #7480, opened 2025-01-29
feat(api): Allow endpoints to define query wildcard

lotas:
Dynamic query validation allows endpoints receive any query parameters that are not known in advance, but still can be validated.

Added `QUERY_WILDCARD` symbol to APIBuilder for dynamic query string validation. This enables:

- Validation of arbitrary query parameters using a single pattern
- Mixed validation of both specific and wildcard query parameters
- Support for endpoints that accept dynamic or user-defined query parameters

Example usage:
```javascript
builder.declare({
  query: {
    [QUERY_WILDCARD]: /^[\w-]+$/,     // Pattern for any unspecified query parameter
    specificParam: /^specific-value$/ // Pattern for a specific parameter
  }
});
```

When function is passed, it allows to validate query argument name as well:
```javascript
builder.declare({
  query: {
    [QUERY_WILDCARD]: (value, key) => key.startsWith('x-'), // Allow only query parameters starting with 'x-'
  }
});

### DIFF
--- a/changelog/GBPeHsVYT5KkmHwRdJ10zg.md
+++ b/changelog/GBPeHsVYT5KkmHwRdJ10zg.md
@@ -1,0 +1,28 @@
+audience: developers
+level: minor
+---
+
+Added `QUERY_WILDCARD` symbol to APIBuilder for dynamic query string validation. This enables:
+
+- Validation of arbitrary query parameters using a single pattern
+- Mixed validation of both specific and wildcard query parameters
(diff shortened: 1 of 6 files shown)


lotas:
oops.. i forgot about clients.. they don't know anything about it (yet)

lotas:
I'll close this one since original issue with phabricator was resolved, and this PR lacks client support for wildcard query params

PR #7482, opened 2025-01-31
chore(d2g)!: drop `podman` support

matt-boris:
>D2G: Drop support for `podman`. Docker worker `payload.capabilities.containerEngine` is removed. Generic Worker config `d2gConfig.containerEngine` is removed.

cc @aerickson @jwmoss @jschwartzentruber @pyoor @glandium fyi

I've also created https://github.com/mozilla-releng/fxci-config/pull/274 to remove the soon-to-be incompatible Generic Worker config.


glandium:
Does that imply podman is going to be removed from workers?

matt-boris:
> Does that imply podman is going to be removed from workers?

We can definitely leave it installed on the image. We understand some people may want to continue using podman, it's just become too much overhead to manage and maintain both engines.

lotas (approved):
Hard to say no to -472 lines :)

petemoore (approved):
Looks perfect!

PR #7502, opened 2025-02-05
feat(generic-worker): set a TASK_WORKDIR env

ahal:
It's often the case that we need to find an absolute path to various things in a task's payload (e.g in the env or command). This is currently difficult with generic-worker because the task user's home directory is not known when creating the task definition.

We've done some hacky workarounds in the run-task script like this: https://github.com/taskcluster/taskgraph/blob/d4e75591a98c12e0fbb0060190860f23b7c39d0b/src/taskgraph/run-task/run-task#L1260

But it occurred to me that if generic-worker could just set the env, none of this would be necessary.


ahal:
Note I didn't bother to figure out how to compile / run tests on this, thus why it's in draft. Hoping to rely on CI for this one.

matt-boris (approved):
LGTM, thanks ahal

petemoore:
What is wrong with `taskworkdir = os.getcwd()`? The working directory is already set in the process, I'm not sure there is a lot of advantage duplicating this in another environment variable.

ahal:
Good question, I should have provided some more context in the PR description :)

That's what we're doing currently, see:
https://github.com/taskcluster/taskgraph/blob/d4e75591a98c12e0fbb0060190860f23b7c39d0b/src/taskgraph/run-task/run-task#L1038

This works great as long as the task's command is using a shell. E.g, we can do:
```
payload:
  command: ["bash", "-c", "mkdir $TASK_WORKDIR"]
```

(And most of our tasks do go through a shell).

But for tasks that don't use the shell, it's not possible to reference `$TASK_WORKDIR` in their commands or environment. Now you might be wondering, why would this patch help? On Windows, `generic-worker` calls `set` in a batch script and then runs the task's command. This apparently allows us to reference any environment variables that `generic-worker` defines like this:
```
payload:
  command: ["some/cmd", "--output", "%TASK_WORKDIR%/artifacts"]
```

So the point of this was to help get rid of a bunch of hacks we have to work around this (see link in the description).

ahal:
Sure, that's what we've done for awhile, but what if you need an absolute path?
```
payload:
  env:
    PIP_CACHE_DIR: .cache/pip
  command:
    - pip install foo
    - cd some-dir
    - pip install bar
```

In this example pip will use two different cache dirs. I'd prefer to always use absolute directories wherever possible to guard against the task changing to a new directory.

ahal:
Another consideration is tasks that don't use `run-task`. These tasks don't get the convenience of a `TASK_WORKDIR` environment variable and need to keep track of the cwd to infer where things should go. Guaranteeing that `$TASK_WORKDIR` is always set is just a nicer user experience for task authors all around.

petemoore:
Thanks for the extra context. I hadn't thought about the case of multiple commands. Indeed when there are multiple commands, and the directory can be changed, it is a burden for the task to have to keep track of the working directory of the first command, so I think I'm fine with this. Probably my mistake was supporting multiple commands in the task payload (Jonas warned me about this back in the day!). But this was mostly to try to improve the experience on Windows, where it isn't trivial to exit from a batch script if something fails...
Maybe we can change it to TASK_WORK_DIR just for cleanliness, but I can do this in a separate PR.

ahal:
I mean, a single command could still run a batch / shell script that changes directories halfway through. So I think this would be a consideration regardless.

PR #7503, opened 2025-02-05
Remove "cloud.google.com/backend-config" from ingress

bqbn:
This annotation should be applied to a "Service" object, not an Ingress. The proper use of it is documented in
https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-configuration#configuring_ingress_features

### DIFF
--- a/infrastructure/k8s/templates/ingress.yaml
+++ b/infrastructure/k8s/templates/ingress.yaml
@@ -17,7 +17,6 @@ metadata:
     kubernetes.io/tls-acme: '{{ if .Values.certManagerClusterIssuerName }}true{{ end }}'
     kubernetes.io/ingress.class: '{{ if eq .Values.ingressType "nginx" }}nginx{{ end }}'
     ingress.kubernetes.io/force-ssl-redirect: '{{ if .Values.certManagerClusterIssuerName }}true{{ end }}'
-    '{{ if .Values.cloudArmorPolicy }}cloud.google.com/backend-config{{ else }}taskcluster/empty{{ end }}': '{{ if .Values.cloudArmorPolicy }}{"default": "taskcluster-backendconfig"}{{ end }}'
 spec:
   tls:
     - secretName: '{{ .Values.ingressTlsSecretName | default "" }}'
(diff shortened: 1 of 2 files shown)


lotas:
thanks @bqbn 
So if it is being removed from the ingress, is this PR missing the service object, should we add it there instead? 

And if it is applied on some other level, maybe we no longer need the rest of the changes that went in https://github.com/taskcluster/taskcluster/pull/7439/files and could be reverted? Wdyt?

bqbn:
> So if it is being removed from the ingress, is this PR missing the service object, should we add it there instead?

As far as this PR is concerned, a `Service` object is not needed. For some historical reason I'm not aware of, Taskcluster is using the `Service` object that is created by the `iprepd-nginx` template here: https://github.com/mozilla-services/cloudops-infra/tree/master/libs/iprepd-nginx/k8s/charts/iprepd-nginx/templates

> And if it is applied on some other level, maybe we no longer need the rest of the changes that went in https://github.com/taskcluster/taskcluster/pull/7439/files and could be reverted? Wdyt?

This is discussed in https://github.com/mozilla-services/cloudops-infra/pull/6233

lotas:
@bqbn since, [ingress.yml](https://github.com/mozilla-services/cloudops-infra/blob/60ba7e924d390dc246d11dc69280ae2eb839d467/projects/taskcluster/k8s/chart_overrides/taskcluster/ingress.yaml) is also being overriden, maybe this PR doesn't need to land? 

I see the issue is that we leave cloudArmor feature added #7439 partially working if we drop those lines. So I'd rather not land this one, or revert the 7439 completely

bqbn:
> @bqbn since, [ingress.yml](https://github.com/mozilla-services/cloudops-infra/blob/60ba7e924d390dc246d11dc69280ae2eb839d467/projects/taskcluster/k8s/chart_overrides/taskcluster/ingress.yaml) is also being overriden, maybe this PR doesn't need to land?
>
> I see the issue is that we leave cloudArmor feature added https://github.com/taskcluster/taskcluster/pull/7439 partially working if we drop those lines. So I'd rather not land this one, or revert the 7439 completely

Yeah, I agree. Reverting 7439 seems to make more sense.

And I'll close this PR as well.

PR #7504, opened 2025-02-06
fix(generic-worker): prune docker images during garbage collection

### LINKED ISSUE #7462: Tasks ran with d2g and the docker engine leave all of their images behind
When running a task with d2g and the docker container engine, tasks will pull the image they require to run but won't clean it up, ever. This means that on, a long running worker, you might get something that looks a bit like this:

```
REPOSITORY                                          TAG       IMAGE ID       CREATED          SIZE
ghcr.io/eijebong/archipelago-yaml-checker           next-79   fb0356dc6b67   59 minutes ago   741MB
ghcr.io/eijebong/archipelago-yaml-checker           next-77   0b3f903a0d92   14 hours ago     849MB
ghcr.io/eijebong/archipelago-yaml-checker           next-78   5f5263ae0b34   14 hours ago     849MB
ghcr.io/eijebong/taskcluster-images/decision-task   main      98e3fcbf71e7   9 days ago       320MB
```

It would be nice if it could keep only the last X used images or something like that.

matt-boris:
Fixes #7462.

>Generic Worker (D2G): prune docker images during garbage collection, if needed.

### DIFF
--- a/changelog/issue-7462.md
+++ b/changelog/issue-7462.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 7462
+---
+Generic Worker (D2G): prune docker images during garbage collection, if needed.
--- a/workers/generic-worker/garbagecollector.go
+++ b/workers/generic-worker/garbagecollector.go
@@ -1,6 +1,10 @@
 package main
 
-import "fmt"
+import (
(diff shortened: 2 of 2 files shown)


petemoore (approved):
Thanks Boris

PR #7505, opened 2025-02-07
feat(generic-worker)!: add `runTaskAsCurrentUser` task feature, remove `runTasksAsCurrentUser` generic worker config

matt-boris:
>Generic Worker: feature `runTaskAsCurrentUser` (note: `Task` not `Tasks`) has been added to replace the previous global task config setting `runTasksAsCurrentUser` (which is no longer supported). Worker pools can elect to enable or disable the feature with boolean config setting `enableRunTaskAsCurrentUser`. Tasks with the feature enabled (`task.payload.features.runTaskAsCurrentUser = true`) require scope `generic-worker:run-task-as-current-user:<provisionerID>/<workerType>`.
>
>This change was introduced in order that access to this privileged feature are guarded not only by worker config settings, but also by task scopes, and furthermore the feature must be explicitly requested, in order that tasks do not unintentionally inherit the feature by virtue of overgenerous scopes or unintentionally running on a pool with the feature enabled.

https://github.com/taskcluster/community-tc-config/pull/915 also need to be merged. I've already applied the config so this PR would pass CI.


workers/generic-worker/artifacts_multiuser_test.go
```diff
@@ -54,6 +54,7 @@ func TestChainOfTrustUpload(t *testing.T) {
 	// Chain of trust is not allowed when running as current user
 	// since signing key cannot be secured
 	if config.RunTasksAsCurrentUser {
+		td.Scopes = append(td.Scopes, "generic-worker:run-tasks-as-current-user:"+td.ProvisionerID+"/"+td.WorkerType)
```
petemoore: I would prefer that we do this [here](https://github.com/taskcluster/taskcluster/blob/3249015448f795d30ebbc3c3304c3b6d86c39284/workers/generic-worker/helper_multiuser_test.go#L43) so that it does not need to be duplicated for every test.

workers/generic-worker/artifacts_multiuser_test.go
```diff
@@ -283,6 +284,7 @@ func TestProtectedArtifactsReplaced(t *testing.T) {
...
```
petemoore: See above.

workers/generic-worker/artifacts_multiuser_test.go
```diff
@@ -383,6 +385,7 @@ func TestPrivilegedFileUpload(t *testing.T) {
 	td := testTask(t)
 
 	if config.RunTasksAsCurrentUser {
+		td.Scopes = append(td.Scopes, "generic-worker:run-tasks-as-current-user:"+td.ProvisionerID+"/"+td.WorkerType)
```
petemoore: Ditto.

workers/generic-worker/os_groups_multiuser_test.go
```diff
@@ -95,6 +95,7 @@ func TestOSGroupsRespected(t *testing.T) {
 	}
 
 	if config.RunTasksAsCurrentUser {
+		td.Scopes = append(td.Scopes, "generic-worker:run-tasks-as-current-user:"+td.ProvisionerID+"/"+td.WorkerType)
```
petemoore: See first comment.

petemoore (commented):
This is a great start Boris, many thanks!

workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -297,6 +297,18 @@ oneOf:
             `exception/malformed-payload`.
 
             Since: generic-worker 54.5.0
+        runTaskAsCurrentUser:
+          type: boolean
+          title: Run task as current user
+          description: |-
+            If true, users will still be created for tasks, but
+            tasks will be executed as the current OS user.
```
petemoore, line 305:
```suggestion
            If `true`, task commands will be executed as the
            user currently running Generic Worker (typically
            `root` or `LocalSystem`), rather than as the
            dedicated task user created for the task. The task
            user account will still be created, and is
            available for the task to use.
```

workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -297,6 +297,18 @@ oneOf:
...
+
+            Requires scope generic-worker:run-tasks-as-current-user:<provisionerID>/<workerType>.
```
petemoore, line 307:
```suggestion
            Requires scope `generic-worker:run-tasks-as-current-user:<provisionerID>/<workerType>`.
```

workers/generic-worker/README.md
```diff
@@ -264,7 +266,10 @@ and reports back results to the queue.
                                             be executed as the task user, rather than the
                                             current user (that runs the generic-worker process).
           runTasksAsCurrentUser             If true, users will still be created for tasks, but
-                                            tasks will be executed as the current OS user. [default: false]
+                                            tasks will be executed as the current OS user.
+                                            Requires scope generic-worker:run-tasks-as-current-user:<provisionerID>/<workerType>.
```
petemoore, line 270:
```suggestion
                                            Requires scope generic-worker:run-task-as-current-user:<provisionerID>/<workerType>.
```

workers/generic-worker/schemas/multiuser_windows.yml
```diff
@@ -254,6 +254,18 @@ properties:
...
+          title: Run task as current user
+          description: |-
+            If true, users will still be created for tasks, but
+            tasks will be executed as the current OS user.
+
+            Requires scope generic-worker:run-tasks-as-current-user:<provisionerID>/<workerType>.
+            Tasks submitted without this scope will be resolved
+            as `exception/malformed-payload`.
+
+            Since: generic-worker 81.0.0
```
petemoore, line 268: See comment above.

workers/generic-worker/usage_multiuser.go
```diff
@@ -18,7 +18,10 @@ func headlessTasksUsage() string {
 func runTasksAsCurrentUserUsage() string {
 	return `
           runTasksAsCurrentUser             If true, users will still be created for tasks, but
-                                            tasks will be executed as the current OS user. [default: false]`
+                                            tasks will be executed as the current OS user.
+                                            Requires scope generic-worker:run-tasks-as-current-user:<provisionerID>/<workerType>.
+                                            Tasks submitted without this scope will be resolved
+                                            as exception/malformed-payload. [default: false]`
```
petemoore, line 24: I think we can lose `runTasksAsCurrentUserUsage` entirely since it is no longer a worker feature, but a task feature instead.

matt-boris: Yep this is what I'm working on 👍🏻

lotas (approved):
👍

workers/generic-worker/usage_windows.go
```diff
@@ -78,7 +78,9 @@ func enableTaskFeatures() string {
           enableRDP                         Enables the RDP feature to be used in the task
                                             payload. [default: true]
           enableRunAsAdministrator          Enables the RunAsAdministrator feature to be used in
-                                            the task payload. [default: true]`
+                                            the task payload. [default: true]
+          enableRunTaskAsCurrentUser		Enables the Run Task As Current User feature to be
```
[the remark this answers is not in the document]
matt-boris: Yeah, not sure what the issue is. My guess is github UI formatting (I personally don't see it on my github). But here's a pic of the added line in my editor and you can see it's the same as the above lines.

Screenshot 2025-02-11 at 9 21 40 AM

lotas (dismissed):

workers/generic-worker/helper_multiuser_test.go
```diff
@@ -40,14 +39,13 @@ func engineTestSetup(t *testing.T, testConfig *gwconfig.Config) {
...
 	testConfig.HeadlessTasks = true
-	testConfig.RunTasksAsCurrentUser = os.Getenv("GW_TESTS_RUN_AS_CURRENT_USER") != ""
 	// Needed for tests that don't call RunWorker()
 	// but test methods/functions directly
 	taskUserCredentials, err := StoredUserCredentials(filepath.Join(cwd, "next-task-user.json"))
 	if err != nil {
 		t.Fatalf("Could not fetch task user credentials: %v", err)
 	}
-	pd, err := process.NewPlatformData(testConfig.RunTasksAsCurrentUser, testConfig.HeadlessTasks, taskUserCredentials)
+	pd, err := process.NewPlatformData(testConfig.HeadlessTasks, taskUserCredentials)
```
petemoore, line 48: I don't remember which tests use `pd` (only a very few) but we should find out by not setting it here, running the CI, and seeing which tests fail. Those tests may or may not rely on `runTaskAsCurrentUser` being set in the task payload, but at this point we do not have a task payload, so the test may need to be adjusted to compensate. Most tests work by submitting a task, then claiming it and executing it, and validating results. However, setting of `taskContext` below is for the unit tests that don't have that workflow, which is only a few.

workers/generic-worker/insecure.go
```diff
@@ -217,3 +217,6 @@ func addEngineDebugInfo(m map[string]string, c *gwconfig.Config) {
 
 func addEngineMetadata(m map[string]interface{}, c *gwconfig.Config) {
 }
+
+func (task *TaskRun) ensurePlatformData() {
```
petemoore, line 221: A doc comment for this method would be useful, to clarify what its purpose is.

workers/generic-worker/main.go
```diff
@@ -1017,6 +1017,9 @@ func (task *TaskRun) Run() (err *ExecutionErrors) {
 	}
 	log.Printf("Running task %v/tasks/%v/runs/%v", config.RootURL, task.TaskID, task.RunID)
 
+	// reset platform data if needed (running tasks as current OS user)
+	task.ensurePlatformData()
```
petemoore, line 1021: I think this should be part of the task feature?

workers/generic-worker/multiuser.go
```diff
@@ -34,18 +34,16 @@ var (
...
+	exitOnError(CANT_SECURE_CONFIG, secureError, "Not able to secure config file %q", configFile)
 }
 
 func rebootBetweenTasks() bool {
 	return !config.HeadlessTasks
 }
 
 func PostRebootSetup(taskUserCredentials *gwruntime.OSUser) {
-	pd, err := process.NewPlatformData(config.RunTasksAsCurrentUser, config.HeadlessTasks, taskUserCredentials)
+	pd, err := process.NewPlatformData(config.HeadlessTasks, taskUserCredentials)
```
petemoore, line 46: I've forgotten how this `pd` is used, but this looks like it might introduce a change in behaviour. I wonder if pd should be part of the task struct, rather than a global now.

workers/generic-worker/multiuser_linux.go
```diff
@@ -19,6 +19,7 @@ func platformFeatures() []Feature {
 		&InteractiveFeature{},
 		&LoopbackAudioFeature{},
 		&LoopbackVideoFeature{},
+		&RunTaskAsCurrentUserFeature{},
```
petemoore, line 22: I suspect having this is up higher in the list might be safer, if it changes `pd` and the other features depend on `taskContext.pd`.

workers/generic-worker/multiuser_posix.go
```diff
@@ -151,12 +151,12 @@ func (task *TaskRun) EnvVars() []string {
 	taskEnv["TASK_WORKDIR"] = taskContext.TaskDir
 	taskEnv["TASK_GROUP_ID"] = task.TaskGroupID
 	taskEnv["TASKCLUSTER_ROOT_URL"] = config.RootURL
-	if config.RunTasksAsCurrentUser {
+	if task.Payload.Features.RunTaskAsCurrentUser {
```
petemoore, line 154: i feel like this should be part of the feature now

workers/generic-worker/multiuser_posix.go
```diff
@@ -151,12 +151,12 @@ func (task *TaskRun) EnvVars() []string {
...
 		taskEnv["TASK_USER_CREDENTIALS"] = ctuPath
 	}
 	if runtime.GOOS == "linux" && !config.HeadlessTasks {
 		taskEnv["DISPLAY"] = ":0"
-		if !config.RunTasksAsCurrentUser {
+		if !task.Payload.Features.RunTaskAsCurrentUser {
```
petemoore, line 159: again, ideally this would be part of the feature

workers/generic-worker/multiuser_test.go
```diff
@@ -12,12 +12,12 @@ import (
...
 )
 
-// TestWhoAmI tests that the correct user is running the task, based on value of config setting RunTasksAsCurrentUser
+// TestWhoAmI tests that the correct user is running the task, based on value of payload feature toggle RunTaskAsCurrentUser
 func TestWhoAmI(t *testing.T) {
 	setup(t)
 
 	payload := GenericWorkerPayload{
-		Command:    goRun("whoami.go", strconv.FormatBool(config.RunTasksAsCurrentUser)),
+		Command:    goRun("whoami.go", strconv.FormatBool(false)),
```
petemoore, line 20:
```suggestion
		Command:    goRun("whoami.go", "false"),
```

workers/generic-worker/multiuser_test.go
```diff
@@ -27,12 +27,29 @@ func TestWhoAmI(t *testing.T) {
...
 
-func TestPrivilegedGenericWorkerBinaryFailsWorker(t *testing.T) {
+func TestWhoAmIAsCurrentUser(t *testing.T) {
 	setup(t)
+	config.EnableRunTaskAsCurrentUser = true
 
-	if config.RunTasksAsCurrentUser {
-		t.Skip("Skipping since we're testing if the generic-worker binary is executable by the task user.")
+	payload := GenericWorkerPayload{
+		Command:    goRun("whoami.go", strconv.FormatBool(true)),
```
petemoore, line 35:
```suggestion
		Command:    goRun("whoami.go", "true"),
```

workers/generic-worker/multiuser_windows.go
```diff
@@ -110,7 +111,7 @@ func (task *TaskRun) prepareCommand(index int) *CommandExecutionError {
 		contents += setEnvVarCommand("TASK_WORKDIR", taskContext.TaskDir)
 		contents += setEnvVarCommand("TASK_GROUP_ID", task.TaskGroupID)
 		contents += setEnvVarCommand("TASKCLUSTER_ROOT_URL", config.RootURL)
-		if config.RunTasksAsCurrentUser {
+		if task.Payload.Features.RunTaskAsCurrentUser {
```
petemoore, line 114: ideally this would be part of the feature

workers/generic-worker/os_groups_multiuser.go
```diff
@@ -19,7 +19,7 @@ func (osGroups *OSGroups) Start() *CommandExecutionError {
 	if len(groupNames) == 0 {
 		return nil
 	}
-	if config.RunTasksAsCurrentUser {
+	if osGroups.Task.Payload.Features.RunTaskAsCurrentUser {
```
petemoore, line 22: I suspect we should still add the task user to the groups, even if running as current user, as the docs say that the task user is still created, but the process is executed as current user. But this can be for a different PR.

workers/generic-worker/run_task_as_current_user.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (r *RunTaskAsCurrentUserTask) RequiredScopes() scopes.Required {
+	return scopes.Required{{
+		"generic-worker:run-task-as-current-user:" + config.ProvisionerID + "/" + config.WorkerType,
+	}}
+}
+
+func (r *RunTaskAsCurrentUserTask) Start() *CommandExecutionError {
+	return nil
```
petemoore, line 53: this is where we should update taskContext.pd. This was previously done in main, but now this is a feature, this is exactly where it would go.

petemoore (commented):
It is looking really good! Overall it is pretty perfect, I think though that the logic of the feature needs to be moved into the Start() function of the feature itself (i.e. setting taskContext.pd). That is pretty much the only job of the feature, to change the taskContext so that processes are executed as the current user using the default PlatformData. Thanks Boris for all your hard work, especially the painstaking task of going through all the tests... It introduced a bit of copy/paste, but on reflection I think this is better than putting everything into shared functions, because it makes the tests much harder to read. So I think it was the right choice to mostly copy/paste them.

workers/generic-worker/README.md
```diff
@@ -200,6 +200,8 @@ and reports back results to the queue.
                                             task payload. [default: true]
           enableLoopbackVideo               Enables the Loopback Video feature to be used in the
                                             task payload. [default: true]
+          enableRunTaskAsCurrentUser        Enables the Run Task As Current User feature to be
+                                            used in the task payload. [default: true]
```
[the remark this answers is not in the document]
matt-boris: We discussed this in Toronto, leaving all generic worker features to be enabled by default

workers/generic-worker/README.md
```diff
@@ -260,11 +262,10 @@ and reports back results to the queue.
...
                                             Administrator. Furthermore, even if
-                                            runTasksAsCurrentUser is true, the script will still
-                                            be executed as the task user, rather than the
-                                            current user (that runs the generic-worker process).
-          runTasksAsCurrentUser             If true, users will still be created for tasks, but
-                                            tasks will be executed as the current OS user. [default: false]
+                                            payload.features.runTaskAsCurrentUser is true,
+                                            the script will still be executed as the task user,
+                                            rather than the current user (that runs the
+                                            generic-worker process).
```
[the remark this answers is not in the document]
matt-boris: Thanks, nice catch. Was sorta blindly replacing references here! My bad.

workers/generic-worker/usage_windows.go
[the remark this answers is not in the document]
matt-boris: My bad! Nice catch guys. Not sure why VS Code displays the spaces as tabs 😅

workers/generic-worker/run_task_as_current_user_windows.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/taskcluster/taskcluster/v80/workers/generic-worker/process"
+)
+
+func (r *RunTaskAsCurrentUserTask) resetPlatformData() {
+	r.task.pd = &process.PlatformData{
+		LoginInfo: &process.LoginInfo{},
+	}
+	for _, c := range r.task.Commands {
+		var token syscall.Token
+		c.SysProcAttr.Token = token
```
petemoore, line 15: can this be done in one line, with e.g. `c.SysProcAttr.Token = syscall.Token(0)`?

workers/generic-worker/run_task_as_current_user.go
```diff
@@ -50,8 +50,19 @@ func (r *RunTaskAsCurrentUserTask) RequiredScopes() scopes.Required {
 }
 
 func (r *RunTaskAsCurrentUserTask) Start() *CommandExecutionError {
+	r.resetPlatformData()
+	r.setTaskUserCredentialsEnvVar()
+	r.removeXDGRuntimeDirEnvVar()
```
petemoore, line 55: maybe this should be `r.platformSpecificActions()`, and linux uses it to remove XDG_RUNTIME_DIR and windows just has nothing to do? to avoid having a method about XDG runtime in windows code...

workers/generic-worker/usage_multiuser_other.go
```diff
@@ -6,6 +6,6 @@ func enableTaskFeatures() string {
 	return `
           enableInteractive                 Enables the Interactive feature to be used in the
                                             task payload. [default: true]
-		  enableRunTaskAsCurrentUser		Enables the Run Task As Current User feature to be
-		  									used in the task payload. [default: true]`
+		  enableRunTaskAsCurrentUser        Enables the Run Task As Current User feature to be
```
petemoore, line 9: This still has tabs before `enableRunTaskAsCurrentUser`.

petemoore (commented):
Almost there!!!! 🎉

workers/generic-worker/usage_multiuser_other.go
matt-boris: Ugh 🤦🏻‍♂️

petemoore (approved):
Love it!

PR #7509, opened 2025-02-11
Fix the `completed` flag for `taskcluster group list`

Eijebong:
The way filters work for this command is that for each task in the returned list from taskcluster, it checks if the status of the task is contained in the flags. A completed task in taskcluster has its status set to `completed` [1], not `complete` so the previous flag always returned an empty list.

[1]: https://github.com/taskcluster/taskcluster/blob/107b4228071ed9b949e0281f72ae177336e9c776/services/queue/schemas/v1/task-status.yml#L37


lotas (approved):
Nice catch! Thanks

PR #7512, opened 2025-02-12
build(deps): upgrade to go1.24.2

matt-boris:
>Upgrades to go1.24.2.

Also, due to this new version, we can now use the `omitzero` struct tag, so jsonschema2go was updated to use this on non-default fields.


matt-boris:
Not sure why the `TestWhoAmIAsCurrentUser` test is failing, but I went ahead and built a staging image off of these changes and ran `whoami` both as task user and current user and got proper output.

https://community-tc.services.mozilla.com/tasks/T1mycd4EQcurFGDjkkwAAg/runs/0 as current user

https://community-tc.services.mozilla.com/tasks/P2BRKXaFSK-PJUomZp6HHQ/runs/0 as task user

I'll look into rewriting the test, as there were some bigger changes to `user.Current()` on windows in the go1.24 release https://tip.golang.org/doc/go1.24#osuserpkgosuser.

matt-boris:
Test failing due to https://github.com/golang/go/issues/73471.

petemoore (approved):
Mammoth job! Perfectly separated into individual commits too, in case we need to revert any of them. Love it.

PR #7515, opened 2025-02-13
Add a way to automatically update d2g test expectations

### LINKED ISSUE #7479: Add a way to update d2g test expecations
Right now when adding anything to d2g that applies to all commands, test expectations have to be updated manually which is a pain. It would be nice if there was a way to update those automatically with what d2g is currently doing

Eijebong:
Running tests with D2G_UPDATE_TEST_EXPECTATIONS=true will update all test
expectations.

Unfortunately, that means that all test expectations got reformatted
because `yaml.JSONToYAML` now has a word in how those expectations are
formatted

I also made CI run tests with this so formatting wouldn't drift off because of manual updates (reading this diff will make it painfully obvious why we don't want it to happen...)

The one issue I see with this is that it forces default values to be in those test cases when they weren't there before but I don't know how we could avoid this.

Fixes #7479


tools/d2g/d2gtest/d2g_test.go
```diff
@@ -110,18 +110,23 @@ func testSuite(schema string, path string) func(t *testing.T) {
 		// not being applied to slices
 		defaults.SetDefaults(&d)
 		unmarshalYAML(t, &d, path)
-		for _, tc := range d.TestSuite.PayloadTests {
+		for i := range d.TestSuite.PayloadTests {
+			tc := &d.TestSuite.PayloadTests[i]
```
petemoore, line 114: This does no harm, but what was the reason to change?

tools/d2g/d2gtest/d2g_test.go
```diff
@@ -241,6 +252,24 @@ func unmarshalYAML(t *testing.T, dest interface{}, path string) {
...
 }
 
+func marshalYAML(t *testing.T, src interface{}, path string) {
+	t.Helper()
+	j, err := json.Marshal(src)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	result, err := yaml.JSONToYAML(j)
```
petemoore, line 262: This is really nice. Such a shame JSONToYAML doesn't offer options for customising line wrapping etc. But I think it is still worth it, even if it makes the test yamls harder to read.

petemoore (approved):
Thanks Bastien!

tools/d2g/d2gtest/d2g_test.go
Eijebong: Because iterating over `d.TestSuite.PayloadTests` directly copies every test case instead of getting a reference to it. And I need that reference to modify the test case when it changes later on

PR #7516, opened 2025-02-13
Bug 1946490: Mount /tmp as tmpfs in containers with d2g 

Eijebong:
I'm not sure why docker doesn't do this by default but it sounds like a no-brainer to me. Some tasks might have some disk heavy things that could be done on a tmpfs and there's no way to mount one after the container has been created.

Part of [Bug 1946490](https://bugzilla.mozilla.org/show_bug.cgi?id=1946490).

Note that this contains #7515 as I really didn't feel like updating all those test expectations manually, this PR is only about the last commit.


Eijebong:
CC @hneiva

petemoore (commented):
I don't have strong feelings, but recognise this is potentially a change in behaviour that could negatively impact some tasks in FXCI. I would propose that if we do this, we use a non-standard location that is a little more explicit, such as `/tmp-as-exec` in order that users actively choose to opt in, rather than it being an unexpected side effect.

petemoore:
CC @jcristau

jcristau:
I think we should punt on this until after we've actually deployed d2g in fxci, since this is not a parity gap with docker-worker.

petemoore:
I'll put back to DRAFT while we are punting, and then we can mark it ready for review again when we want to come back to it (just so it doesn't stay in our queue).

Eijebong:
I'll close this. If we want to revisit we probably want to make it opt-in for tasks (although not in that `/tmp-as-exec` way... A feature or an explicit mount would work)

The main issue that no one's raised here is that this caps /tmp to 50% of the worker's RAM instead of the size of the root disk.

PR #7522, opened 2025-02-13
fix(generic-worker): properly configure env vars when running as current user

### LINKED ISSUE #7521: Generic Worker: `TASK_USER_CREDENTIALS` env var not set when using `runTaskAsCurrentUser`
This is happening ever since https://github.com/taskcluster/taskcluster/pull/7505 landed and was released in v81.0.0.

matt-boris:
Fixes #7521.

>Generic Worker: fixes an issue introduced in v81.0.0 where `TASK_USER_CREDENTIALS` env var wasn't written to the task's environment if `task.payload.features.runTaskAsCurrentUser` was enabled.

### DIFF
--- a/workers/generic-worker/run_task_as_current_user_posix.go
+++ b/workers/generic-worker/run_task_as_current_user_posix.go
@@ -4,6 +4,7 @@ package main
 
 import (
 	"runtime"
+	"strings"
 	"syscall"
 
 	"github.com/taskcluster/taskcluster/v81/workers/generic-worker/process"
@@ -18,10 +19,22 @@ func (r *RunTaskAsCurrentUserTask) resetPlatformData() {
 
 func (r *RunTaskAsCurrentUserTask) platformSpecificActions() {
 	if r.task.Payload.Env == nil {
-		return
+		r.task.Payload.Env = make(map[string]string)
 	}
+
+	r.task.Payload.Env["TASK_USER_CREDENTIALS"] = ctuPath
+	r.task.setVariable("TASK_USER_CREDENTIALS", ctuPath)
 
 	if runtime.GOOS == "linux" && !config.HeadlessTasks {
 		delete(r.task.Payload.Env, "XDG_RUNTIME_DIR")
+	}
+	var newEnv []string
+	for _, c := range r.task.Commands {
+		for _, e := range c.Env {
+			if !strings.HasPrefix(e, "XDG_RUNTIME_DIR=") {
+				newEnv = append(newEnv, e)
+			}
+		}
+		c.Env = newEnv
 	}
 }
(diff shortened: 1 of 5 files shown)


workers/generic-worker/run_task_as_current_user_posix.go
```diff
@@ -18,10 +19,22 @@ func (r *RunTaskAsCurrentUserTask) resetPlatformData() {
...
+	r.task.Payload.Env["TASK_USER_CREDENTIALS"] = ctuPath
 	if runtime.GOOS == "linux" && !config.HeadlessTasks {
 		delete(r.task.Payload.Env, "XDG_RUNTIME_DIR")
 	}
+
+	var newEnv []string
+	for _, c := range r.task.Commands {
+		c.SetEnv("TASK_USER_CREDENTIALS", ctuPath)
+		for _, e := range c.Env {
+			if !strings.Contains(e, "XDG_RUNTIME_DIR") {
```
petemoore, line 34:
```suggestion
			if !strings.HasPrefix(e, "XDG_RUNTIME_DIR=") {
```

workers/generic-worker/run_task_as_current_user_posix.go
```diff
@@ -18,10 +19,22 @@ func (r *RunTaskAsCurrentUserTask) resetPlatformData() {
 
 func (r *RunTaskAsCurrentUserTask) platformSpecificActions() {
 	if r.task.Payload.Env == nil {
-		return
+		r.task.Payload.Env = make(map[string]string)
```
petemoore, line 22: I think r.task.Payload.Env is meant to be what was in task.payload.env, so I think we shouldn't mutate that. I think we should just update the commands. I remember saying it is ok because the original is in task.Definition - but thinking more about this, that is because task.Definition has Payload as just a json.RawMessage, and we want to have a Generic Worker Payload struct, so we extract it from the RawMessage and put it in the Generic Worker Payload object. But that should really just be a copy of what is in the task, rather than mutating it. If we need to add env vars to the commands, we should do that just in the command objects.
We should check if we are mutating anything else inside task.Payload anywhere, because I think that would be a mistake (otherwise we can't distinguish between what was in the original task definition or not, unless we unpack the entire task definition again, which feels wrong).

petemoore (commented):
Let's talk about this tomorrow. Thanks Boris!

workers/generic-worker/run_task_as_current_user_posix.go
matt-boris: I know what you're saying here and I did think about this. We'll probably just need to do a little reworking here. The reason why I mutated `r.task.Payload.Env` is because the interactive feature (which starts _after_ this new task feature) calls out to `TaskRun.EnvVars()` which grabs from _that_ map. I noticed, also, that chain of trust (starts _before_ this new task feature) calls this method when trying to `cat` out the private signing key (which I determined doesn't _really_ need all of the task's defined env vars.

petemoore (approved):
Let's go with this, it is a great improvement!

PR #7524, opened 2025-02-14
fix(release): run as current user

matt-boris:
Was removed from the worker config here https://github.com/taskcluster/community-tc-config/commit/b84b8e2ac98b3fab1ab72d31a06d76a75b8ff301#diff-23fc71490cc090eec860515efc9788fad76e67b9559b9420316af4a48ef23713L39 and we now need to use the task feature.


petemoore (approved):
Thanks!

PR #7526, opened 2025-02-14
chore(generic-worker): fix linting issue

matt-boris:


petemoore (approved):

workers/generic-worker/multiuser_linux_test.go
```diff
@@ -30,10 +32,6 @@ func TestTaskUserCredentialsEnvVarIsWrittenAsCurrentUser(t *testing.T) {
 	)
 
 	_ = submitAndAssert(t, td, payload, "completed", "completed")
-	logtext := LogText(t)
-	if strings.Contains(logtext, "Error: TASK_USER_CREDENTIALS not set") {
-		t.Fatalf("Was expecting log file to not contain \"Error: TASK_USER_CREDENTIALS not set\", but it does: \n%v", logtext)
-	}
```
petemoore, line 36: don't you need to still keep this?

test/go-lint.sh
```diff
@@ -8,7 +8,13 @@ if [ -n "${unformatted_files}" ]; then
...
+rc=0
 for engine in multiuser insecure; do
   echo "Running golangci-lint for ${engine} engine.."
   golangci-lint run --build-tags "${engine}"
+  current_rc=$?
+  if [ $current_rc -ne 0 ]; then
+    rc=$current_rc
+  fi
 done
+exit $rc
```
petemoore, line 20: can you not just change line 1 to `#!/bin/bash -e` ? this way, it will fail straight away, so should save some compute time.

workers/generic-worker/multiuser_linux_test.go
```diff
@@ -60,8 +61,4 @@ func TestXDGRuntimeDirEnvVarIsRemovedAsCurrentUser(t *testing.T) {
 	)
 
 	_ = submitAndAssert(t, td, payload, "failed", "failed")
-	logtext := LogText(t)
-	if !strings.Contains(logtext, "Success: XDG_RUNTIME_DIR not set") {
-		t.Fatalf("Was expecting log file to contain \"Success: XDG_RUNTIME_DIR not set\", but it doesn't: \n%v", logtext)
-	}
```
petemoore, line 66: (here too)

test/go-lint.sh
matt-boris: Wouldn't this fail on the first failure then? I was hoping to run both engines. Either way is fine by me.

petemoore (approved):
Nice one!

petemoore (approved):
Sweet!

workers/generic-worker/multiuser_linux_test.go
```diff
@@ -3,47 +3,21 @@
...
-			"echo", "${XDG_RUNTIME_DIR:?Success: XDG_RUNTIME_DIR not set}",
-		}},
+		Command: goRun(
+			"check-env.go",
+			"XDG_RUNTIME_DIR",
+			"/run/user/1000",
+		),
 		Env: map[string]string{
 			"XDG_RUNTIME_DIR": "/run/user/1000",
 		},
```
petemoore, line 23: i don't think you want to specify it in the environment here, because the feature likely overwrites it with a different value (i.e. not /run/user/1000) and then it will not match, and the task will fail, so the test will pass. I think in this case, your previous implementation might have been better, if it was working. Just because we don't know what the value will be...

workers/generic-worker/multiuser_test.go
```diff
@@ -47,6 +48,31 @@ func TestWhoAmIAsCurrentUser(t *testing.T) {
 	_ = submitAndAssert(t, td, payload, "completed", "completed")
 }
 
+func TestTaskUserCredentialsEnvVarIsWrittenAsCurrentUser(t *testing.T) {
+	setup(t)
+	config.EnableRunTaskAsCurrentUser = true
```
petemoore, line 53: config is a global, do we need a defer block to restore the previous value? i don't remember. it is probably set in setup(t) but not all tests call setup(t) I think, so it feels dangerous to modify it but not restore it in a teardown.

workers/generic-worker/multiuser_linux_test.go
matt-boris: The feature _should_ entirely remove this variable though.

petemoore (approved):
Fingers crossed the CI passes, looks good!

PR #7533, opened 2025-02-18
chore(tests): add tc-proxy d2g test with refreshing creds

matt-boris:
Would've caught https://github.com/taskcluster/taskcluster/issues/7568 before it was released, but adding for additional coverage!

### DIFF
--- a/changelog/O_T64JpoSey4NKOs6FVojg.md
+++ b/changelog/O_T64JpoSey4NKOs6FVojg.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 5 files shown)


petemoore (commented):
Is there a reason not to enable d2g by default in tests, in the `setup()` method? Perhaps we could just disable it for any tests that explicitly need it disabled?

matt-boris:
Yeah that's what I started this PR off by doing! Still just a WIP, need to get around to writing more tests, obviously...

petemoore (approved):
This is great!

I've made a test push and created #7574 as a draft PR with this new test added, but the fix reverted which it aims to detect, to make sure the test really catches the failure that we experienced that led to the creation of this test in the first place. If the CI in that draft PR fails due to this new testing failing, all good! :-)

PR #7534, opened 2025-02-18
fix(generic-worker): clean up cache ACLs and fix permission issues on windows

### LINKED ISSUE #7532: Permission issues on windows caches
During my investigation on what went wrong in [bug 1948412](https://bugzilla.mozilla.org/show_bug.cgi?id=1948412), I went down the rabbit hole of windows caches permissions and there are a few things there that are wrong.

1. The cache folder retains all previous ACLs from all the previous users. (see first screenshot, you can count how many attempts I needed to figure all of this out)
2. The owner of the cache folder remains the one that first created it. (1005 in this screenshot)

Image

3. Those rights are not given recursively and instead the cache relies on files within having rights inheritance on
4. Pip doesn't create its cache with inheritance on.

Image

All those screenshots were made while a task was running as the following user, with that cache mounted

Image

Task description:

```yaml
provisionerId: windows
workerType: ci11
taskQueueId: windows/ci11
schedulerId: taskcluster-ui
projectId: none
taskGroupId: EnyorDdzRbKIzPtO5J4rWQ
dependencies: []
requires: all-completed
routes: []
priority: lowest
retries: 5
created: '2025-02-18T11:44:00.542Z'
deadline: '2025-02-18T14:44:00.542Z'
expires: '2026-02-18T14:44:00.542Z'
scopes:
  - generic-worker:cache:test
payload:
  env:
    PIP_CACHE_DIR: .task-cache/pip
  mounts:
    - cacheName: test
      directory: .task-cache/pip
  command:
    - '"C:\Program Files\Python313\python.exe" -m venv venv'
    - .\venv\Scripts\activate
    - >-
      pip install --timeout 120 --no-index --find-links
      https://pypi.pub.build.mozilla.org/pub/ vcversioner==2.16.0.0
  maxRunTime: 3600
metadata:
  name: example-task
  owner: name@example.com
  source: http://taskcluster/tasks/create
tags: {}
extra: {}
```

The machine is a freshly installed windows 11 (home edition 😩, don't do that), python 3.13.2, generic worker 81.0.0 with a slight modification regarding `Remote Desktop Users` (see #7531 for why) although this is irrelevant as the issue first appeared on mozilla machines running 77.3.1 on win11.

matt-boris:
Fixes #7532.

>Generic Worker (windows): fix cache ownership issues. Clean up ACLs so prior task users aren't referenced anymore.

### DIFF
--- a/workers/generic-worker/multiuser_windows.go
+++ b/workers/generic-worker/multiuser_windows.go
@@ -540,8 +540,29 @@ func PreRebootSetup(nextTaskUser *gwruntime.OSUser) {
 	}
 }
 
-func changeOwnershipInDir(dir string, currentOwnerUID string, newOwnerUsername string) error {
-	return host.Run("icacls", dir, "/grant:r", newOwnerUsername+":(OI)(CI)F")
+func changeOwnershipInDir(dir, newOwnerUsername string, cache *Cache) error {
+	if dir == "" || newOwnerUsername == "" || cache == nil {
+		return fmt.Errorf("directory path, new owner username, and cache must not be empty")
+	}
+
+	// Do nothing if the current owner is the same as the new owner
+	if cache.OwnerUsername == newOwnerUsername {
+		return nil
+	}
+
+	// Reset to inherited permissions only, recursively
+	err := host.Run("icacls", dir, "/reset", "/t")
+	if err != nil {
+		return fmt.Errorf("failed to reset permissions on dir %v: %v", dir, err)
+	}
+
+	// Grant full control to new owner, adding to inherited permissions
+	err = host.Run("icacls", dir, "/grant", newOwnerUsername+":(OI)(CI)F")
+	if err != nil {
+		return fmt.Errorf("failed to grant permissions to %v on dir %v: %v", newOwnerUsername, dir, err)
+	}
+
+	return nil
 }
 
 func convertNilToEmptyString(val interface{}) string {
--- a/changelog/issue-7532.md
+++ b/changelog/issue-7532.md
@@ -1,0 +1,5 @@
+audience: general
+level: patch
+reference: issue 7532
+---
+Generic Worker (windows): fix cache ownership issues. Clean up ACLs so prior task users aren't referenced anymore.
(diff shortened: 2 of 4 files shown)


matt-boris:
Tested successfully on a staging worker https://community-tc.services.mozilla.com/provisioners/proj-taskcluster/worker-types/gw-windows-2022-staging/workers/westus/vm-p7iq1gktqyqhez8sfsyueqskiobvz0stuek?sortBy=started&sortDirection=desc

workers/generic-worker/multiuser_windows.go
```diff
@@ -540,8 +540,29 @@ func PreRebootSetup(nextTaskUser *gwruntime.OSUser) {
...
+	// Do nothing if the current owner is the same as the new owner
+	if cache.OwnerUsername == newOwnerUsername {
+		return nil
+	}
+
+	// Reset to inherited permissions only, recursively
+	err := host.Run("icacls", dir, "/reset", "/t")
+	if err != nil {
+		return fmt.Errorf("failed to reset permissions on dir %v: %v", dir, err)
+	}
```
petemoore, line 557: I think this could potentially cause unexpected behaviour. Software, like pip, may wish to set explicit permissions on files, and Generic Worker by overruling these decisions, could cause unexpected behaviour. For example, suppose you run a tool like pip which populates a directory with content, and sets some specific permissions against some of those files. Those permissions would now be lost.
I think a safer approach would be mirroring what we do on posix platforms, which is just to recursively replace references from the old task user to the new one. I feel like an approach like this is likely to have fewer surprises, than discarding file permissions that the software that uses the cache has intentionally set.

I'm open to discussing it though.

@jcristau and @Eijebong, what do you think?

matt-boris: I'd argue that since generic worker gives the task user full control over the files/dirs in the task dir, that it's safe to reset permissions, recursively and grant the task user full control over the cache.

matt-boris: Going to try testing with `icacls <dir> /substitute <oldSID> <newSID>` here https://github.com/taskcluster/taskcluster/compare/main...matt-boris/gwFixWindowsCaches2#diff-9dca1899ff574724c40fa4df18e288b5da8c242ebf4a78e9ad9b4dc5b0f58eae

petemoore (approved):
OK, this is certainly better than the current situation, so maybe we can roll with this for now, given that currently it is totally broken in this scenario.

PR #7510, opened 2025-02-12
D2G: record imageHash and imageArtifactHash in chain of trust certificate

petemoore:
Fixes #7014

PR #7538, opened 2025-02-21
no-op test envvar rename HGI -> GHI

petemoore:
From https://github.com/taskcluster/taskcluster/pull/7510#discussion_r1964199790


matt-boris (approved):
You didn't need to fix this 😄 LGTM though!

PR #7536, opened 2025-02-20
chore: modernize go code

matt-boris:
Achieved by running `go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -test ./...` as instructed in https://pkg.go.dev/golang.org/x/tools/gopls@v0.18.0/internal/analysis/modernize.

Modernizations made in this PR:

- `replacing an if/else conditional assignment by a call to the built-in min or max functions added in go1.21;`
- `replacing interface{} by the 'any' type added in go1.18;`
- `replacing append([]T(nil), s...) by slices.Clone(s) or slices.Concat(s), added in go1.21;`
- `replacing a loop around an m[k]=v map update by a call to one of the Collect, Copy, Clone, or Insert functions from the maps package, added in go1.21;`
- `replacing []byte(fmt.Sprintf...) by fmt.Appendf(nil, ...), added in go1.19;`
- `replacing a 3-clause for i := 0; i < n; i++ {} loop by for i := range n {}, added in go1.22;`

PR #7539, opened 2025-02-21
Test whether setting purge caches exit codes causes a nil pointer

petemoore:
Testing whether this is the underlying cause of generic-worker panics in #7536 ...


petemoore:
This will be incorporated into #7536 directly.

PR #7542, opened 2025-02-25
feat(w-m): Fix edge case with empty launch configs

lotas:
Spotted in logs on dev for worker pool without launch configs

```
TypeError: Cannot read properties of undefined (reading 'length')
    at AzureProvider.provision (file:///app/services/worker-manager/src/providers/azure/index.js:312:58)
```


matt-boris (approved):
👍🏻

PR #7547, opened 2025-02-26
feat(generic-worker): adds `optional` payload artifact field

### LINKED ISSUE #7545: Generic Worker payload artifact option to make artifact optional

matt-boris:
Fixes #7545.

>Generic Worker: adds `optional` field to payload artifacts to ignore any artifact upload errors, for example, if the artifact isn't known to exist at the end of a task but you don't want the task to resolve as `failed/failed`. This makes the transition from Docker Worker --> Generic Worker (through d2g) more seamless, as Docker Worker does not resolve tasks as `failed/failed` if the artifact doesn't exist.

>D2G: sets each payload artifact as `optional` so tasks won't resolve as `failed/failed` if the artifact doesn't exist, like Docker Worker does.

### DIFF
--- a/workers/generic-worker/d2g_test.go
+++ b/workers/generic-worker/d2g_test.go
@@ -83,6 +83,59 @@ func TestD2GWithValidDockerWorkerPayload(t *testing.T) {
 	}
 }
 
+func TestD2GArtifactDoesNotExist(t *testing.T) {
+	setup(t)
+	testTime := tcclient.Time(time.Now().AddDate(0, 0, 1))
+	image := map[string]any{
+		"name": "ubuntu:latest",
+		"type": "docker-image",
+	}
+	imageBytes, err := json.Marshal(image)
+	if err != nil {
+		t.Fatalf("Error marshaling JSON: %v", err)
+	}
+	payload := dockerworker.DockerWorkerPayload{
+		Command: []string{
+			"/bin/bash",
+			"-c",
+			"echo hello world > existingArtifactPath",
+		},
+		Image:      json.RawMessage(imageBytes),
+		MaxRunTime: 30,
+		Artifacts: map[string]dockerworker.Artifact{
+			"existingArtifact": {
+				Path:    "existingArtifactPath",
+				Type:    "file",
+				Expires: testTime,
+			},
+			"nonExistingArtifact": {
+				Path:    "nonExistingArtifactPath",
+				Type:    "file",
+				Expires: testTime,
+			},
+		},
+	}
+	defaults.SetDefaults(&payload)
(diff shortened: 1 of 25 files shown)


lotas (approved):
LGTM

workers/generic-worker/schemas/multiuser_windows.yml
```diff
@@ -182,6 +182,15 @@ properties:
...
             Since: generic-worker 16.2.0
+        optional:
+          title: Optional artifact
+          type: boolean
+          default: false
+          description: |-
+            If `true`, the artifact is optional. If the optional artifact doesn't exist and
+            fails to be uploaded, the task will not resolve as `failed/failed`.
+
+            Since: generic-worker 83.1.0
```
matt-boris, line 193: Force-pushed to update the version in the schemas

workers/generic-worker/d2g_test.go
```diff
@@ -83,6 +83,59 @@ func TestD2GWithValidDockerWorkerPayload(t *testing.T) {
...
+	case "insecure:linux":
+		_ = submitAndAssert(t, td, payload, "exception", "malformed-payload")
+		logtext := LogText(t)
+		t.Log(logtext)
+		if !strings.Contains(logtext, "task payload contains unsupported osGroups: [docker]") {
+			t.Fatal("Was expecting log file to contain 'task payload contains unsupported osGroups: [docker]'")
+		}
+	default:
+		_ = submitAndAssert(t, td, payload, "exception", "malformed-payload")
+	}
```
petemoore, line 136: would be nice here to confirm that the existing artifact is present, and the nonexisting one is not, if possible

workers/generic-worker/schemas/insecure_posix.yml
```diff
@@ -177,6 +177,15 @@ oneOf:
               encoding to all the files contained in the directory.
 
               Since: generic-worker 16.2.0
+          optional:
+            title: Optional artifact
+            type: boolean
+            default: false
+            description: |-
+              If `true`, the artifact is optional. If the optional artifact doesn't exist and
```
petemoore, line 185: Technically, I think this should be phrased more like "if the file or directory doesn't exist, the artifact won't be created". The artifact isn't the file/dir itself, it is something that is created from the file/dir (and e.g. has its own attributes).

petemoore (commented):
This is fantastic, but I think one key test is missing, which is testing an optional directory artifact. At the moment the docs and tests only cover file artifacts, but I think it would be good to make sure the docs and tests cover both cases. Thanks Boris!

petemoore (approved):
Great work again, Boris!

PR #7549, opened 2025-02-27
feat(auth): Store hook id with group in audit

lotas:
This improves recently merged #7419 with 2 things:

1. we missed `hookId` is not sufficient to identify hook - `hookGroupId` was added
2. Replaces `worker_pool` with `worker-pool`  for the sake of consistency (we mostly use `-` everywhere external and `_` is mostly internal or db level)


petemoore (dismissed):
Looks good, thanks Yarik!

PR #7551, opened 2025-02-27
fix(generic-worker): don't `chown` loopback devices

### LINKED ISSUE #7411: generic-worker: consider setting an acl instead of chown for device nodes
generic-worker's loopback video feature calls makeFileOrDirReadWritableForUser on the device node, which changes its uid/gid.  In the d2g/docker context, this means the device node is then only accessible to root inside the container, whereas if the ownership wasn't changed then users in the video group can access it, breaking expectations of docker-worker tasks.

This can be worked around in the task (see https://bugzilla.mozilla.org/show_bug.cgi?id=1922533), but it might not be necessary if instead of using chown, we used setfacl or similar to grant/revoke access to the task user.

matt-boris:
Fixes #7411.

>Generic Worker: no longer `chown` loopback video/audio devices to the task user. Explicitly change group of the devices to `video`/`audio`, respectively, so that users in those groups may still access them.

### DIFF
--- a/changelog/issue-7411.md
+++ b/changelog/issue-7411.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 7411
+---
+Generic Worker: no longer `chown` loopback video/audio devices to the task user. Explicitly change group of the devices to `video`/`audio`, respectively, so that users in those groups may still access them.
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -637,6 +637,10 @@ func envMappings(dwPayload *dockerworker.DockerWorkerPayload, config map[string]
 		additionalEnvVars = append(additionalEnvVars, "TASKCLUSTER_PROXY_URL")
 	}
 
+	if dwPayload.Capabilities.Devices.LoopbackVideo && config["allowLoopbackVideo"].(bool) {
+		additionalEnvVars = append(additionalEnvVars, "TASKCLUSTER_VIDEO_DEVICE")
+	}
+
 	envVarNames := []string{}
 	for envVarName := range dwPayload.Env {
 		envVarNames = append(envVarNames, envVarName)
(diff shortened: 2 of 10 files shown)


matt-boris:
After this patch, you can see that [audio](https://community-tc.services.mozilla.com/tasks/BLDyPNADSA6TuIku8fGz_A/runs/1/logs/public/logs/live.log#L137) and [video](https://community-tc.services.mozilla.com/tasks/BLDyPNADSA6TuIku8fGz_A/runs/1/logs/public/logs/live.log#L39) devices are owned by root but in their respective groups (`audio` and `video`)

petemoore (commented):
This looks reasonable to me. I'd like @jcristau to have a look over it too before we merge. Are the `audio` and `video` groups standard on Linux, or do we need to also document them as requirements in the installation guide? Or should we add a startup check that they exist if the feature is used, and resolve as malformed-payload with an explanation that the worker doesn't support the feature because it is missing the os group "video" or "audio"? This would prevent the worker from panicking and being unusable. But if the groups always exist, just like `root` always exists, maybe we don't need to do this. Thanks Boris!

Another thing that would be _really_ great, to help avoid that we release this, and discover that it still isn't working under d2g for some other reason, would be to add a d2g test for audio and video. If we can demonstrate in the CI already that everything lines up end-to-end so that docker worker payloads using audio/video integrate properly with the generic-worker feature, we might be able to save ourselves a few iterations of releasing patches. Thanks Boris, awesome job! :-)

matt-boris:
> ... add a d2g test for audio and video. If we can demonstrate in the CI already that everything lines up end-to-end so that docker worker payloads using audio/video integrate properly with the generic-worker feature, we might be able to save ourselves a few iterations of releasing patches.

Already done here :) https://github.com/taskcluster/taskcluster/pull/7455

petemoore (approved):
Love it! TDD FTW!

PR #7564, opened 2025-03-04
feat(g-w): for d2g tasks, have the tc proxy listen on the docker brid…

jcristau:
…ge interface instead of loopback

### DIFF
--- a/workers/generic-worker/taskcluster_proxy.go
+++ b/workers/generic-worker/taskcluster_proxy.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"os/exec"
+	"strings"
 
 	tcclient "github.com/taskcluster/taskcluster/v83/clients/client-go"
 	"github.com/taskcluster/taskcluster/v83/internal/scopes"
@@ -39,6 +41,7 @@ type TaskclusterProxyTask struct {
 	taskclusterProxy         *tcproxy.TaskclusterProxy
 	task                     *TaskRun
 	taskStatusChangeListener *TaskStatusChangeListener
+	taskclusterProxyAddress  string
 }
 
 func (l *TaskclusterProxyTask) ReservedArtifacts() []string {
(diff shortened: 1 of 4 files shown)


workers/generic-worker/taskcluster_proxy.go
```diff
@@ -57,9 +60,19 @@ func (l *TaskclusterProxyTask) RequiredScopes() scopes.Required {
 }
 
 func (l *TaskclusterProxyTask) Start() *CommandExecutionError {
+	if l.task.D2GInfo != nil {
+		out, err := exec.Command("docker", "network", "inspect", "bridge", "--format", "{{ index (index .IPAM.Config 0 ) \"Gateway\"}}").Output()
+		if err != nil {
+			return executionError(internalError, errored, fmt.Errorf("could not determine docker bridge IP address: %s", err))
+		}
+		l.taskclusterProxyAddress = strings.TrimSpace(string(out))
```
matt-boris, line 68: This can be simplified/standardized a bit by using our `host.CombinedOutput()` helper, and you'd be able to print from standard error too in your error block.

```suggestion
		out, err := host.CombinedOutput("docker", "network", "inspect", "bridge", "--format", "{{ index (index .IPAM.Config 0 ) \"Gateway\"}}")
		if err != nil {
			return executionError(internalError, errored, fmt.Errorf("could not determine docker bridge IP address. Output: %s. Error: %v", out, err))
		}
		l.taskclusterProxyAddress = strings.TrimSpace(out)
```

matt-boris (commented):
This looks good to me, thanks so much Julien 👍🏻

I've included one suggestion around calling out to `exec.Command()`.

workers/generic-worker/taskcluster_proxy.go
matt-boris: Curious, why don't you want stderr?

jcristau: The intent here is to capture the ip address on stdout.  If docker happens to print other stuff on stderr for whatever reason, mixing that in with stdout isn't going to end well?

matt-boris (approved):
Changes LGTM, thanks!

I'd like to give @petemoore a chance to take a look here as well before merging.

PR #7565, opened 2025-03-04
feat(d2g): stop using the host network

### LINKED ISSUE #7552: D2G: create ipv6 configuration
On Docker Worker we used to set up IPv6 at e.g. https://github.com/mozilla-platform-ops/monopacker/blob/193e4d8a002f7972406c872ebc6d41011eda35b9/scripts/docker-worker-linux/90-set-up-ephemeral-disks.sh#L156, which is required for some webrtc tests. The fact that we're lacking that config, plus the fact that we use `--net=host` in D2G, means that now fails. We may need to revisit the `--net=host` decision, or require that workers have IPv6 suitably configured on the host network.

@jcristau can you provide a suitable Docker Worker test payload that should pass if the networking is set up as required, and fails if it is not? My idea is that we first add this as a test, and then validate any fix we make causes the test to pass. Thanks!

jcristau:
Github Bug/Issue: Fixes #7552 (this doesn't create any IPv6 configuration, but it means if such a config is set up in the docker bridge config, it gets used)

Also fixes https://github.com/taskcluster/taskcluster/issues/7320.


jcristau:
Depends on  #7564.

matt-boris (dismissed):
Changes LGTM, thanks so much! And thanks @Eijebong for helping test this with a decision task as well 🎉

Note, this also addresses https://github.com/taskcluster/taskcluster/issues/7320 and _may_ fix https://github.com/taskcluster/taskcluster/issues/6797. I'll update the description to auto-close the former.

petemoore (dismissed):
LGTM, thanks Julien! ❤️

matt-boris:
> Do we need to add any extra tests, or do the existing d2g tests give enough coverage? (I don't remember if we have any D2G tests that explicitly test e.g. taskcluster-proxy - @matt-boris do you remember?)

We've got this test already https://github.com/taskcluster/taskcluster/blob/6f6b9d510502f22f654262b980c6a6376afd3984/workers/generic-worker/d2g_test.go#L194-L229 for sure

matt-boris:
All set to rebase this one now @jcristau! Thanks so much.

matt-boris (approved):
Thanks 👍🏻

PR #7567, opened 2025-03-05
Routinely write file-caches.json and directory-caches.json to disk

### LINKED ISSUE #7427: Generic worker running with headlessTasks doesn't write its cache state unless something goes wrong
If you run a generic worker with `headlessTasks: true`, the `directory-cache.json` file never gets written to which means that the cache state is lost on restart (for example on a worker restart because of an update)

After staring at it for a bit, I think I figured out why.

Features persist their state through a `defer` function here: https://github.com/taskcluster/taskcluster/blob/b756a6e95bbb50a1d67b264eafd97fd4124230dc/workers/generic-worker/main.go#L411 in the `RunWorker` function.

The function loops forever, trying to get tasks and has many ways of getting returned from which would call the deferred function. When headless tasks are on though, the only way to exit that function is for something to go wrong or for the worker to receive a signal.

Unfortunately, go doesn't run deferred functions when it exits because it gets a signal. You can easily view that with this:

```go
package main

import (
	"fmt"
)

func main() {
	defer func() {
		fmt.Println("Deferred function called")
	}()

	// Uncomment then press ^C to stop the program
	// for {}
}
```

```
[eijebong@plutonium] ~  >>> go run main.go
Deferred function called
[eijebong@plutonium] ~  >>>
# Edited the file to uncomment the for line
[eijebong@plutonium] ~  >>> go run main.go
^Csignal: interrupt
[eijebong@plutonium] ~  >>>
```

petemoore:
The write to disk occurs after each task that uses the mounts feature.

Fixes #7427.

### DIFF
--- a/workers/generic-worker/fileutil/fileutil.go
+++ b/workers/generic-worker/fileutil/fileutil.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 
 	"github.com/mholt/archiver/v3"
+	"github.com/taskcluster/slugid-go/slugid"
 )
 
 func WriteToFileAsJSON(obj any, filename string) error {
@@ -28,8 +29,14 @@ func WriteToFileAsJSON(obj any, filename string) error {
 	} else {
 		log.Printf("Saving file %v", filename)
 	}
-
-	return os.WriteFile(filename, append(jsonBytes, '\n'), 0644)
+	tempFilename := filename + "-" + slugid.Nice()
+	if err := os.WriteFile(tempFilename, append(jsonBytes, '\n'), 0644); err != nil {
+		return fmt.Errorf("failed to write to temp file: %w", err)
+	}
+	if err := os.Rename(tempFilename, filename); err != nil {
+		return fmt.Errorf("failed to rename temp file: %w", err)
+	}
+	return nil
 }
 
 func CalculateSHA256(file string) (hash string, err error) {
--- a/workers/generic-worker/mounts.go
+++ b/workers/generic-worker/mounts.go
@@ -147,19 +147,6 @@ func (taskMount *TaskMount) Error(message string) {
 	taskMount.task.Error("[mounts] " + message)
 }
 
-func (feature *MountsFeature) PersistState() (err error) {
-	err = fileutil.WriteToFileAsJSON(&fileCaches, "file-caches.json")
-	if err != nil {
-		return
-	}
-	err = fileutil.WriteToFileAsJSON(&directoryCaches, "directory-caches.json")
-	if err != nil {
-		return
-	}
-	err = fileutil.SecureFiles("file-caches.json", "directory-caches.json")
-	return
-}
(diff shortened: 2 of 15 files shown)


workers/generic-worker/mounts.go
```diff
@@ -467,6 +467,10 @@ func (taskMount *TaskMount) Stop(err *ExecutionErrors) {
 			err.add(Failure(e))
 		}
 	}
+	err.add(executionError(internalError, errored, fileutil.WriteToFileAsJSON(&fileCaches, "file-caches.json")))
```
Eijebong, line 470: https://github.com/taskcluster/taskcluster/blob/8cad35306ffe8ecb8442b5fcd7af567ee1eef676/workers/generic-worker/mounts.go#L150 shouldn't this get removed too then?

matt-boris: Faster than our CI 😉

workers/generic-worker/mounts.go
```diff
@@ -467,6 +467,10 @@ func (taskMount *TaskMount) Stop(err *ExecutionErrors) {
...
+	err.add(executionError(internalError, errored, fileutil.WriteToFileAsJSON(&directoryCaches, "directory-caches.json")))
+	err.add(executionError(internalError, errored, fileutil.SecureFiles("file-caches.json", "directory-caches.json")))
+	return
```
matt-boris, line 473:
```suggestion
```

workers/generic-worker/mounts.go
petemoore: whoops, thought i had deleted that 😬 
thanks!

matt-boris (dismissed):

workers/generic-worker/fileutil/fileutil.go
```diff
@@ -28,8 +29,14 @@ func WriteToFileAsJSON(obj any, filename string) error {
 	} else {
 		log.Printf("Saving file %v", filename)
 	}
-
-	return os.WriteFile(filename, append(jsonBytes, '\n'), 0644)
+	tempFilename := filename + "-" + slugid.Nice()
+	if err := os.WriteFile(tempFilename, append(jsonBytes, '\n'), 0644); err != nil {
```
lotas, line 33: what is the reason behind writing to a temp file first and then moving it?

Eijebong: So the cache mapping file doesn't get corrupted if the worker gets interrupted mid-write.

lotas (approved):

workers/generic-worker/fileutil/fileutil.go
petemoore: Note, this could also be the cause of https://mozilla.sentry.io/issues/3332121580/?project=6462337

```
*json.SyntaxError

invalid character '\x00' looking for beginning of value
```

PR #7573, opened 2025-03-06
feat(w-m): Fixes edge-case with provisioner

lotas:
CapacityPerInstance=2 and while(toSpawn) didn't work well together.. when w-p has minCapacity: 1, while(toSpawn) underflows and goes to negative infinity (capped at wp max capacity)


matt-boris (approved):
Thanks! LGTM :shipit:

PR #7578, opened 2025-03-11
fix(g-w): better detection of the docker bridge ip address

jcristau:
In the presence of more than one address on the docker bridge network, we were assuming the gateway could be found on the first network, which is not necessarily the case.  Instead, we now get all gateways from `docker network inspect`, and select the first ipv4 one (ipv6 addresses would need more massaging to work in URLs, so we skip them).

### DIFF
--- a/workers/generic-worker/taskcluster_proxy.go
+++ b/workers/generic-worker/taskcluster_proxy.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net"
 	"net/http"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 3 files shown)


workers/generic-worker/taskcluster_proxy.go
```diff
@@ -57,11 +58,28 @@ func (l *TaskclusterProxyTask) RequiredScopes() scopes.Required {
...
+				continue
+			}
+			if ipAddress.To4() == nil {
+				continue
+			}
+			l.taskclusterProxyAddress = v
+			break
+		}
+		if l.taskclusterProxyAddress == "" {
+			return executionError(internalError, errored, fmt.Errorf("could not determine docker bridge IP address: no ipv4 gateway found among % v", gateways))
```
Eijebong, line 81: `%v` ?

Eijebong:
Ran a decision task with this: https://taskcluster.bananium.fr/tasks/CHdPow6TSsugxvSOqC_HvA/runs/0/logs/live/public/logs/live.log

I don't have workers with an ipv6 on the docker bridge so I can't test that but at least this doesn't break existing things

workers/generic-worker/tcproxy/tcproxy.go
```diff
@@ -85,5 +85,5 @@ func waitForPortToBeActive(ipAddress string, port uint16) error {
 		}
 		time.Sleep(100 * time.Millisecond)
 	}
-	return fmt.Errorf("timeout waiting for taskcluster-proxy port %v to be active", port)
+	return fmt.Errorf("timeout waiting for taskcluster-proxy %v:%v to be active", ipAddress, port)
```
matt-boris, line 88: 👍🏻 thanks

Eijebong:
With an IPv6 on the bridge: https://taskcluster.bananium.fr/tasks/XRZemGE6RzKiDapiQ3lo1g/runs/0/logs/public/logs/live.log

```
[root@tc-worker-small-2 ~]# docker inspect  bridge
[
    {
        "Name": "bridge",
        "Id": "ae30901807e9c38a51b3961b7c5d17f219272480f96fd3d6b1d4e87bec29f585",
        "Created": "2025-03-11T18:23:06.342684879+01:00",
        "Scope": "local",
        "Driver": "bridge",
        "EnableIPv6": true,
        "IPAM": {
            "Driver": "default",
            "Options": null,
            "Config": [
                {
                    "Subnet": "172.17.0.0/16",
                    "Gateway": "172.17.0.1"
                },
                {
                    "Subnet": "fd15:4ba5:5a2b:100a::/64",
                    "Gateway": "fd15:4ba5:5a2b:100a::1"
                }
            ]
        },
```

Eijebong (approved):

matt-boris (approved):
LGTM, thank you!

PR #7583, opened 2025-03-12
fix(d2g): add `audio`/`video` os-groups and scopes

### LINKED ISSUE #7581: audio group is not added to D2G tasks using loopback-audio
**Describe the bug**
The following warning is emitted when specifying loopback-audio device:

> [taskcluster:warn 2025-03-12T06:13:48.708Z] The 'audio' group is not in the list of OS groups. Consider adding so that the loopback audio devices will work as expected.

**To Reproduce**

1. Launch a D2G task with loopback-audio
2. Observe warning in the live log

**Expected behavior**
Since D2G sets up the OS Groups used by the g-w task, it should add the audio group if necessary (which the warning seems to imply).

**Taskcluster version**
83.2.4

 - OS: linux
 - Cloud: gcp

**Additional context**
Example task: https://community-tc.services.mozilla.com/tasks/VGauu0_BRNKPKwMjBvQbew/runs/0
This is also missing snd-aloop, which I will log separately.

matt-boris:
Fixes #7581.

Should have been handled in https://github.com/taskcluster/taskcluster/pull/7551.

>D2G: add `audio`/`video` os-groups and scopes needed when the Docker Worker task payload requests these loopback devices.

### DIFF
--- a/changelog/issue-7581.md
+++ b/changelog/issue-7581.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 7581
+---
+D2G: add `audio`/`video` os-groups and scopes needed when the Docker Worker task payload requests these loopback devices.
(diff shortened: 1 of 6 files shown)


petemoore (dismissed):
lgtm, thanks boris!

Is it worth having a couple of generic-worker d2g tests for this too? Maybe in a separate PR?

jcristau:
I don't think this is correct FWIW, there shouldn't be a need for the task user to be a member of these groups, the warning is just wrong in the d2g case.

matt-boris:
Thanks @jcristau nice catch. I incorrectly read the issue report for this, thinking they ran into an issue because the task user was not in the group. Maybe I can make it so that warning only appears for non-d2g tasks.

Issue #6894, opened 2024-03-08
generic worker should handle OOM better


bhearsum:
At the moment, if a task runs something that consumes all of the system RAM, generic-worker tends to get OOM-killed, resulting in the task failing with `CLAIM_EXPIRED`. This is a bit unfortunate, because this state can happen for a variety of reasons.

Ideally, we'd avoid generic-worker getting OOM-killed entirely (@petemoore - I believe you have ideas about this).

A smaller measure, if possible, might be to detect the OOM kill, and either log a clear message to the log or resolve the task with a particular state.

Issue #6464, opened 2023-08-03
Place limits on task processes so that OOM is reported rather than crashes worker


petemoore:
Sometimes tasks take more memory (physical and virtual) than is available, and cause the worker to crash, meaning that no task logs get published, and it is very difficult to understand what the cause was.

Instead, we should probably place limits on task processes so that if a task process runs out of memory, Generic Worker can still operate correctly, report the issue, upload logs, etc.

On Windows this would be handled via [Job Objects](https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects), or [cgroups](https://en.wikipedia.org/wiki/Cgroups) on Linux. Not sure about macOS - maybe ulimit will help since typically Generic Worker is invoked from a bash script.

Note though, firefox itself makes extensive use of Job Objects on Windows, and I suspect may take advantage of many cgroups features on Linux. Therefore, introducing limits on the processes Generic Worker launches may have subtle/unwanted effects on tasks that themselves use similar features. I'm kind of scared to even look into it, after the fun of [bug 1303455](https://bugzilla.mozilla.org/show_bug.cgi?id=1303455) but there is also a good chance that if we get it right, it could be a really good asset to the product, and improve reliability and eventually lead to improved performance monitoring etc.

PR #7588, opened 2025-03-13
feat(generic-worker): monitor memory usage during tasks and help prevent OOMs

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/6464.
Fixes https://github.com/taskcluster/taskcluster/issues/6894.

>Generic Worker: adds memory usage monitoring during tasks and reports average and peak memory used, in addition to the system's total available memory.
>
>If the total percentage of memory used exceeds 90% for 5 consecutive measurements at 0.5s intervals, the worker will abort the task to prevent OOM crashes and errors. If `disableOOMProtection` (default `false`) is set to `true` in the worker configuration, the worker will continue to monitor and report on memory usage, but will not abort the task if memory consumption is high.
>
>Resource monitoring can be disabled with worker config `enableResourceMonitor` (default `true`) or per task via `payload.features.resourceMonitor` (default `true`).

### DIFF
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -11,6 +11,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -215,6 +216,7 @@ func loadConfig(configFile *gwconfig.File) error {
 			CachesDir:                      "caches",
 			CheckForNewDeploymentEverySecs: 1800,
 			CleanUpTaskDirs:                true,
+			DisableOOMProtection:           false,
 			DisableReboots:                 false,
 			DownloadsDir:                   "downloads",
 			EnableChainOfTrust:             true,
@@ -820,13 +822,38 @@ func (task *TaskRun) IsIntermittentExitCode(c int64) bool {
 }
 
 func (task *TaskRun) ExecuteCommand(index int) *CommandExecutionError {
+	c := task.Commands[index]
 	task.Infof("Executing command %v: %v", index, task.formatCommand(index))
-	log.Print("Executing command " + strconv.Itoa(index) + ": " + task.Commands[index].String())
+	log.Print("Executing command " + strconv.Itoa(index) + ": " + c.String())
 	cee := task.prepareCommand(index)
 	if cee != nil {
 		panic(cee)
 	}
-	task.result = task.Commands[index].Execute()
+	go c.MonitorResources(func(previouslyWarned bool) bool {
+		if config.DisableOOMProtection {
+			if !previouslyWarned {
+				task.Warn("Sustained memory usage above 90%!")
+				task.Warn("OOM protections are disabled, continuing task...")
+			}
+			return false
+		} else {
+			task.Warn("Sustained memory usage above 90%!")
+			task.Warn("Aborting task to prevent OOM issues...")
+		}
+		err := task.StatusManager.Abort(
+			&CommandExecutionError{
+				Cause:      errors.New("task aborted due to sustained memory usage above 90%"),
+				Reason:     internalError,
+				TaskStatus: aborted,
+			},
+		)
+		if err != nil {
+			task.Warnf("Error when aborting task: %v", err)
+		}
+		return true
+	})
+	task.result = c.Execute()
+	task.result.GatherUsage(c)
 	if ae := task.StatusManager.AbortException(); ae != nil {
 		return ae
 	}
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -11,7 +11,6 @@ package main
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -63,7 +62,7 @@ var (
 	config         *gwconfig.Config
 	serviceFactory tc.ServiceFactory
 	configFile     *gwconfig.File
-	Features       []Feature
+	features       []Feature
 
 	logPath   = filepath.Join("generic-worker", "live_backing.log")
 	debugInfo map[string]string
@@ -73,14 +72,15 @@ var (
 )
 
 func initialiseFeatures() (err error) {
-	Features = []Feature{
+	features = []Feature{
 		&LiveLogFeature{},
 		&TaskclusterProxyFeature{},
 		&OSGroupsFeature{},
 		&MountsFeature{},
+		&ResourceMonitorFeature{},
 	}
-	Features = append(Features, platformFeatures()...)
-	for _, feature := range Features {
+	features = append(features, platformFeatures()...)
+	for _, feature := range features {
 		log.Printf("Initialising task feature %v...", feature.Name())
 		err := feature.Initialise()
 		if err != nil {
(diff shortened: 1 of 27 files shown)


workers/generic-worker/main.go
```diff
@@ -820,13 +822,38 @@ func (task *TaskRun) IsIntermittentExitCode(c int64) bool {
...
+			return false
+		} else {
+			task.Warn("Sustained memory usage above 90%!")
+			task.Warn("Aborting task to prevent OOM issues...")
+		}
+		err := task.StatusManager.Abort(
+			&CommandExecutionError{
+				Cause:      errors.New("task aborted due to sustained memory usage above 90%"),
+				Reason:     internalError,
+				TaskStatus: aborted,
```
bhearsum, line 847: Does `aborted` cause a rerun to happen automatically? (I presume so...just double checking!)

jcristau: I don't think we should conflate this with workerShutdown, we don't necessarily want to auto-retry on OOM IMO.

matt-boris: That's also a good point Julien.

petemoore: Agree, `failed` is probably the correct task status.

workers/generic-worker/main.go
```diff
@@ -820,13 +822,38 @@ func (task *TaskRun) IsIntermittentExitCode(c int64) bool {
...
+			}
+			return false
+		} else {
+			task.Warn("Sustained memory usage above 90%!")
+			task.Warn("Aborting task to prevent OOM issues...")
+		}
+		err := task.StatusManager.Abort(
+			&CommandExecutionError{
+				Cause:      errors.New("task aborted due to sustained memory usage above 90%"),
+				Reason:     internalError,
```
petemoore, line 846: Reason should be `""` (since failures don't contain a reason)

workers/generic-worker/main.go
petemoore: TaskStatus should be `failed` (so task resolves as failed, not as exception).

petemoore (commented):
Some initial feedback, before I get to the rest of it...

petemoore (commented):
This looks really good! A summary of my thoughts:
* We need to make it clear that the memory measurements are full-system measurements, not pertaining solely to the processes the task executes
* It would be nice if this was a task feature, since we have the structures in place to enable/disable task features, and keep the code separated nicely
* It would be nice to be able to disable the monitoring if we don't want it, rather than only being able to disable the task cancellation. I think it should be an optional feature of the process library. Just because, over time we may want to add more and more different types of monitoring, and it isn't clear that we would always want all the things running all the time, for anything that launches a process (e.g. internally we launch processes as the task user to do things, and we probably don't care about monitoring system memory usage when we do those things).

That said, this is a huge step forward, so thanks very much!

workers/generic-worker/main.go
```diff
@@ -820,13 +822,37 @@ func (task *TaskRun) IsIntermittentExitCode(c int64) bool {
...
+				TaskStatus: failed,
+			},
+		)
+		if err != nil {
+			task.Warnf("Error when aborting task: %v", err)
+		}
+		return true
+	})
+	task.result = c.Execute()
+	task.result.GatherUsage(c)
```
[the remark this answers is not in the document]
matt-boris: You don't get the benefit of seeing resource usage per payload command if this is a task feature.

workers/generic-worker/process/multiuser_windows.go
```diff
@@ -123,6 +125,8 @@ func NewCommandNoOutputStreams(commandLine []string, workingDirectory string, en
 func (c *Command) Kill() (killOutput string, err error) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
+	// abort even if process hasn't started
+	close(c.abort)
```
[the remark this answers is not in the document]
matt-boris: Just moved it up before the nil process check to keep in line with what all the other platforms do.

workers/generic-worker/process/process.go
```diff
@@ -126,7 +156,86 @@ func (r *Result) Verdict() string {
...
+	val := float64(bytes)
+	if val < 1024*1024 {
+		return fmt.Sprintf("%f B", val)
+	} else if val < 1024*1024*1024 {
+		mb := val / (1024 * 1024)
+		return fmt.Sprintf("%.2f MiB", mb)
+	} else {
+		gb := val / (1024 * 1024 * 1024)
+		return fmt.Sprintf("%.2f GiB", gb)
+	}
```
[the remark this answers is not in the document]
matt-boris: No 😄 Thanks for catching.

workers/generic-worker/process/multiuser_windows.go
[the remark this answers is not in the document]
matt-boris: This makes it so this case can complete successfully still:

```go
	case <-c.abort:
		r.SystemError = fmt.Errorf("process aborted")
		r.Aborted = true
```

workers/generic-worker/process/process.go
```diff
@@ -7,29 +7,41 @@ import (
...
+		Duration    time.Duration
+		Aborted     bool
+		KernelTime  time.Duration
+		UserTime    time.Duration
+		Usage       *ResourceUsage
+	}
+
+	ResourceUsage struct {
+		AverageMemoryUsed    uint64
+		PeakMemoryUsed       uint64
```
petemoore, line 41: Part of me feels like this should be AverageSystemMemoryUsed / PeakSystemMemoryUsed to make it clearer that it isn't process memory.

workers/generic-worker/schemas/insecure_posix.yml
```diff
@@ -263,27 +263,36 @@ oneOf:
...
+            posix platform (FreeBSD, macOS), the task will resolve as
+            `exception/malformed-payload`.
+
+            Since: generic-worker 54.5.0
+        resourceMonitor:
+          type: boolean
+          title: Resource monitor
+          description: |-
+            The resource monitor feature enables the monitoring of system memory usage
+            of the task.
```
petemoore, line 292: How about something like:

The resource monitor features reports Peak (System) Memory Used, Average (System) Memory Used and Total Available Memory in the task log for each task command executed. It also will abort any task command that causes the available system memory to be reduced to less than or equal to 10% of the total system memory for five consecutive measurements at 0.5s intervals. When this happens, the task will be resolved as failed.

Note, I haven't double checked that the terms I have written here match exactly, and I'm not 100% sure if the feature performs both metrics monitoring and task abortion, but I think so - but please check you agree with my description here before committing it, or if you can find a more elegant explanation, that is also great. I just wanted to get across that the feature really needs to tell the user in detail what it is doing, to leave no open questions for the user, such as:
* what type of monitoring is this?
* is this just monitoring, or will it actively abort my tasks?
* if they are aborted, will the task be resolved as failure or exception?
* if memory usage is reported, where do i see the results?
* what specific memory indicators are provided?
* are the counters for system memory or process memory?
etc etc
In many ways, the docs are the most important part, as this is usually where users stumble, if they don't understand what a feature is, or what it is really doing, or how to use it etc.

petemoore (commented):
This is awesome, getting it as a task feature is brilliant, and makes the code super modular. Thank you so much! Great job, this is a really nice feature! 🚀

petemoore (approved):
Great work!

PR #7590, opened 2025-03-16
Audit History UI

### LINKED ISSUE #7548: Audit history improvements
After implementing #5438 we can build some new functionality on top: 
1. Allow to query for the actions for a particular clientId (what entities were changed by this client)
2. Show history in the UI

thakkarujas5:
Fixes #7548

### DIFF
--- a/ui/src/components/SecretForm/index.jsx
+++ b/ui/src/components/SecretForm/index.jsx
@@ -20,6 +20,7 @@ import Button from '../Button';
 import SpeedDial from '../SpeedDial';
 import DatePicker from '../DatePicker';
 import SpeedDialAction from '../SpeedDialAction';
+import AuditHistorySpeedDialAction from '../AuditHistorySpeedDialAction';
 import { secret } from '../../utils/prop-types';
 
 @withStyles(theme => ({
@@ -262,6 +263,11 @@ export default class SecretForm extends Component {
               <ContentSaveIcon />
             </Button>
             <SpeedDial>
+              <AuditHistorySpeedDialAction
+                entityName="secret"
+                entityId={secretName}
+                disabled={loading}
+              />
               <SpeedDialAction
                 requiresAuth
                 tooltipOpen
--- a/ui/src/utils/client.js
+++ b/ui/src/utils/client.js
@@ -1,4 +1,4 @@
-import { Queue, WorkerManager } from '@taskcluster/client-web';
+import { Auth, Queue, WorkerManager } from "@taskcluster/client-web";
 
 export const getClient = ({ Class, user, ...options }) => {
   return new Class({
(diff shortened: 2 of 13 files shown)


ui/package.json
```diff
@@ -82,6 +82,7 @@
     "resolve-pathname": "^3.0.0",
     "snake-case": "^3.0.3",
     "subscriptions-transport-ws": "^0.11.0",
+    "taskcluster-client-web": "^83.9.1",
```
lotas: We changed package name to `@taskcluster/client-web` and it is already present in the package. json above, so you can drop all package/yarn.lock changes

ui/src/components/SecretForm/index.jsx
```diff
@@ -262,6 +265,21 @@ export default class SecretForm extends Component {
               <ContentSaveIcon />
             </Button>
             <SpeedDial>
+              <SpeedDialAction
```
lotas: nit: this looks like it could be a dedicated component where only entityName and entityId are varying across the usage, since it is being used x4 times

ui/src/utils/client.js
```diff
@@ -40,3 +40,31 @@ export const changeTaskGroupPriority = async ({
...
+  return auth.getEntityHistory(entityType, entityId, {
+    limit,
+  });
+};
+
+export const getClientAuditHistory = async (clientId, user, { limit }) => {
+  const auth = getClient({
+    Class: Auth,
+    user,
+    authorizedScopes: [`auth:audit-history:${clientId}`],
```
lotas: this is probably a typo, needs client-audit-history:
https://github.com/taskcluster/taskcluster/blob/9654803c6c0874113bc4b55c34f28068d3427b72/services/auth/src/api.js#L395 
```suggestion
    authorizedScopes: [`auth:client-audit-history:${clientId}`],
```

ui/src/views/AuditHistory/ViewClientAuditHistory/index.jsx
```diff
@@ -0,0 +1,136 @@
...
+    width: '60px',
+    minWidth: '60px',
+    maxWidth: '60px',
+    whiteSpace: 'nowrap',
+  },
+});
+
+@withStyles(useStyles)
+@withAuth
+export default class ViewClientAuditHistory extends Component {
```
lotas, line 36: It looks like it is  very similar to the `ViewAuditHistory` component where different loading function is used. 
Maybe we could reuse it and just pass the loading function to both pages?

lotas (changes requested):

ui/src/views/AuditHistory/ViewClientAuditHistory/index.jsx
thakkarujas5: so one thing I realized is that for audit history component from the speed dial and from the sidebar require different columns to be viewed.

speed dial : serial_no, action_type, client_id, created
sidebar: serial, action_type, entity_type, entity_id, created.

It was my mistake that I kept columns same in speed dial and sidebar components. For sidebar showing client id did not make sense as the user already knows what their client id is and hence display other columns.

Updating you of the change, can make it into the same component but would require passing the column names to render and all, happy to make the change in case you still feel we should make it into one component.

ui/src/utils/client.js
thakkarujas5: done.

ui/src/components/SecretForm/index.jsx
thakkarujas5: done

ui/src/views/AuditHistory/ViewClientAuditHistory/index.jsx
thakkarujas5: Side bar component:

image

Speed dial component:

image

ui/package.json
thakkarujas5: done

thakkarujas5:
I dont see many tests in the ui directory, should I add them for the new components or skip?

lotas:
> I dont see many tests in the ui directory, should I add them for the new components or skip?

yeah, we had some migrated to jest years ago, and recently switched to vitest. Plus we also have playwright now 
Up to you, vitest could ensure that components are rendering properly (without network interaction part)

lotas (approved):
Looks good to me, thanks!

lotas (approved):

PR #7591, opened 2025-03-18
Generic Worker: don't update /etc/subgid since podman support dropped

petemoore:
I think this is left over from podman days, and no longer needed.

@Eijebong can you confirm?

Thanks!

### DIFF
--- a/changelog/Qgfg20wZSni3PRa0gxb84A.md
+++ b/changelog/Qgfg20wZSni3PRa0gxb84A.md
@@ -1,0 +1,3 @@
+audience: general
+level: silent
+---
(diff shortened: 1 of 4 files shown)


Eijebong (approved):
Indeed, this is no longer required, it was necessary for caches to work with podman but doesn't have any other use

matt-boris (dismissed):
Thanks, good catch. I missed this when I ripped out podman support :)

matt-boris:
Oh, will you please apply this commit https://github.com/taskcluster/taskcluster/pull/7588/commits/f81aa517b96c6563750022acfaca43068dce6a1d to fix the broken test?

matt-boris (approved):

Issue #7368, opened 2024-11-01
Generic worker seems to unnecessarily copy artifacts to `/tmp` before uploading them


Eijebong:
I was wondering why one of my worker was getting OOM'ed when building a docker container. Turns out generic worker copies artifacts in `/tmp` through `copyToTempFileAsTaskUser` -> `gwCopyToTempFile` before uploading them which seems fairly useless.

I also got a weird output, showing the artifact in `/tmp` twice which might be a red herring.

```
471M    /tmp/artifact0.zst1997748626
471M    /tmp/artifact0.zst3096491912
```

This was all with a d2g for which the copy from the container to `cwd` is also unnecessary as one could stream the file out of the container with `podman cp container:path -` which would stream a tar archive of the files.

petemoore:
It is possible for a task command to create a subprocess detached from the parent, that continues to run after the parent process has completed. This subprocess can continue writing to an artifact while generic-worker is publishing it. Since the file is changing over time, the calcuated hashsum of the file may no longer be valid when the upload completes, since it has changed since the upload started. Also the Content-Length header may be reported inaccurately for the same reason. In order to ensure that the file is not mutated during the process of calculating hash sums, content lengths, and publishing the file, a copy is taken. There may be other ways to achieve the same, by scanning for open file handles, but that may be more complicated or require different implementations across the platforms that we support (macOS, Linux, Windows, FreeBSD). Therefore a copy was introduced as a simple but effective solution. If there is a different way to guarantee file immutability, we could adapt the approach.

PR #7598, opened 2025-03-25
feat(d2g): add volume mount artifact type

### LINKED ISSUE #7594: d2g tasks cannot preserve artifacts when a graceful shutdown happens
I recently hit my first spot termination on a d2g-enabled instance, and found that it didn't upload any artifacts. The tail end of the log looked as follows:
```
[taskcluster:error] Aborting task...
[taskcluster 2025-03-19T22:00:20.268Z] === Task Finished ===
[taskcluster 2025-03-19T22:00:20.268Z] Task Duration: 31h3m57.387703653s
[taskcluster:error] Uploading error artifact public/build from file artifact0 with message "Could not read directory '/home/task_174230977146133/artifact0'", reason "file-missing-on-worker" and expiry 2026-03-06T03:01:17.662Z
[taskcluster:error] TASK FAILURE during artifact upload: file-missing-on-worker: Could not read directory '/home/task_174230977146133/artifact0'
[taskcluster 2025-03-19T22:00:20.535Z] [mounts] Preserving cache: Moving "/home/task_174230977146133/cache0" to "caches/d5XjZt6uTlukpeL32gz93g"
[taskcluster 2025-03-19T22:00:20.711Z] [mounts] Preserving cache: Moving "/home/task_174230977146133/cache1" to "caches/RwZCbQNXTder5KbT4wOh4Q"
[taskcluster 2025-03-19T22:00:20.880Z] Uploading link artifact public/logs/live.log to artifact public/logs/live_backing.log with expiry 2026-03-06T03:01:17.662Z
[taskcluster:error] graceful termination requested, without time to finish tasks
```

Because `artifact0` is [created as part of the task payload command](https://github.com/taskcluster/taskcluster/blob/e943cb6e359bc8bac61200c93d525f0627cf2694/tools/d2g/d2g.go#L371-L374), and spot terminations [do not allow time to finish tasks](https://github.com/taskcluster/taskcluster/blob/e943cb6e359bc8bac61200c93d525f0627cf2694/workers/generic-worker/main.go#L1167-L1177), we end up aborting the task, and never running the artifact copying logic.

matt-boris:
Fixes #7594.

>Docker Worker (D2G): adds `volume` type for artifacts. This is strictly used for D2G purposes only. Use this type to have D2G volume mount your artifact path instead of `docker cp`'ing the artifact at the end of the task run. This can be useful under spot termination instances where the `docker cp` command doesn't get a chance to run, instead a volume mount will have the files on the host ready for upload as soon as the spot termination requests comes in.

### DIFF
--- a/workers/generic-worker/schemas/multiuser_posix.yml
+++ b/workers/generic-worker/schemas/multiuser_posix.yml
@@ -605,8 +605,19 @@ oneOf:
     artifacts:
       type: object
       title: Artifacts
-      description: >-
+      description: |-
         Artifact upload map example: ```{"public/build.tar.gz": {"path": "/home/worker/build.tar.gz", "expires": "2016-05-28T16:12:56.693817Z", "type": "file"}}```
+        Artifacts can be an individual `file`, a `directory` containing
+        potentially multiple files with recursively included subdirectories,
+        or a `volume` which will create a volume mount from the
+        host to the running container. Unlike `directory` artifacts, the
+        `volume` directory will already exist as the task starts. Since the
+        artifacts will be created directly on the host, they do not need to
+        be copied from the container to the host prior to being published,
+        so perform more efficiently, and simplify the d2g-generated task payload.
+        Moreover, in the case of time-critical spot terminations, tasks have
+        more chance of successfully publishing volume artifacts than directory
+        artifacts, due to the efficiency gain.
       additionalProperties:
         $ref: "#/definitions/artifact"
     supersederUrl:
@@ -925,7 +936,7 @@ definitions:
         enum:
         - file
         - directory
-        - volume # used only in d2g translation
+        - volume
       path:
         title: Location of artifact in container, as an absolute path.
         type: string
--- a/workers/generic-worker/schemas/multiuser_posix.yml
+++ b/workers/generic-worker/schemas/multiuser_posix.yml
@@ -925,6 +925,7 @@ definitions:
         enum:
         - file
         - directory
+        - volume # used only in d2g translation
       path:
         title: Location of artifact in container, as an absolute path.
         type: string
--- a/changelog/issue-7594.md
+++ b/changelog/issue-7594.md
@@ -1,0 +1,5 @@
+audience: users
+level: minor
+reference: issue 7594
+---
+Docker Worker (D2G): adds `volume` type for artifacts. This is strictly used for D2G purposes only. Use this type to have D2G volume mount your artifact path instead of `docker cp`'ing the artifact at the end of the task run. This can be useful under spot termination instances where the `docker cp` command doesn't get a chance to run, instead a volume mount will have the files on the host ready for upload as soon as the spot termination requests comes in.
--- a/generated/references.json
+++ b/generated/references.json
@@ -9617,7 +9617,7 @@
               "additionalProperties": {
                 "$ref": "#/definitions/artifact"
               },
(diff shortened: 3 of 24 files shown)


bhearsum:
I tested this as best I could on my own worker. I wasn't able to simulate a preemption event (due to my worker being configured as `static`), but I was able to verify that `volume` artifacts still uploaded things correctly. Eg: https://firefox-ci-tc.services.mozilla.com/tasks/MsjW9cPETA2nIL-mmfKogg/runs/0 and https://firefox-ci-tc.services.mozilla.com/tasks/H-REkImlTjWXV2ztOn-xdw/runs/6. I'm fairly confident this would solve the issue in a preemption event as well, seeing as these files are visible on the host machine immediately after creation.

matt-boris:
@Eijebong mind taking a look here (especially so at the second commit)? I'm trying to be sure I haven't screwed anything up for the interactive feature and auto-drop into the running container feature that you added. 🙏🏻

Eijebong:
I think with this we can close #7368?

Looks good to me. I'm assuming the fact that you didn't default all directory artifacts to be changed into volumes is because of permission differences within the container?

petemoore:
> I think with this we can close #7368?
> 
> Looks good to me. I'm assuming the fact that you didn't default all directory artifacts to be changed into volumes is because of permission differences within the container?

The reason not to switch all directory artifacts to volume artifacts is that they have slightly different behavior, which may or may not be significant, depending on the task. With a volume artifact, the directory will exist when the task starts, but with a directory artifact, it won't.

workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -911,6 +911,7 @@ definitions:
         enum:
         - file
         - directory
+        - volume # used only in d2g translation
```
petemoore, line 914: We need to provide at least a basic description of what a volume artifact is, because people are not likely to find it in the old release notes in the future. There is currently a `title` but not a `description` in this section, so this would be a great place to describe the difference between a directory artifact and a volume artifact.

The description can say that volume mount is the recommended/modern way to declare a directory artifact. The directory will be created as a volume mount from the host inside the container, so unlike a directory artifact, it will already exist when the task starts. Since the artifacts will be created directly on the host, they do not need to be copied from the container to the host prior to being published, so perform more efficiently, and simplify the generated task payload. Moreover, in the case of time-critical spot terminations, tasks have more chance of successfully publishing volume artifacts than directory artifacts, due to the efficiency gain.

workers/generic-worker/schemas/multiuser_posix.yml
```diff
@@ -591,8 +591,19 @@ oneOf:
     artifacts:
       type: object
       title: Artifacts
-      description: >-
+      description: |-
         Artifact upload map example: ```{"public/build.tar.gz": {"path": "/home/worker/build.tar.gz", "expires": "2016-05-28T16:12:56.693817Z", "type": "file"}}```
+        Artifacts can be an individual `file`, a `directory` containing
+        potentially multiple files with recursively included subdirectories,
+        or a `volume` (d2g only) which will create a volume mount from the
```
petemoore, line 598: Since this section of the schema is for d2g only, i think you can safely remove `(d2g only)`. Or you can write `(not supported in the original docker worker implementation)`. But probably just removing the brackets is the least confusing, I would say.

petemoore (dismissed):

petemoore (approved):
awesome!

PR #7317, opened 2024-10-08
fix(d2g): use `KILL` signal on `timeout` command instead of `--init`

matt-boris:
Fixes #7309.

>D2G: No longer pass `--init` to the `docker run ...` command. This was breaking docker image build tasks that Taskgraph creates. To kill the running docker container, we now pass `-s KILL` to the `timeout` command.

Issue #7309, opened 2024-10-02
D2G: using `--init` with `docker` fails build docker image tasks


matt-boris:
Relevant lines of code:
```bash
+ timeout 7200 docker run --init -t --name taskcontainer --memory-swap -1 --pids-limit -1 -e CONTEXT_PATH -e CONTEXT_TASK_ID -e DOCKER_BUILD_ARGS -e DOCKER_IMAGE_ZSTD_LEVEL -e HASH -e IMAGE_NAME -e MOZ_AUTOMATION -e PROJECT -e RUN_ID -e SCCACHE_DISABLE -e TASKCLUSTER_INSTANCE_TYPE -e TASKCLUSTER_ROOT_URL -e TASKCLUSTER_WORKER_LOCATION -e TASK_GROUP_ID -e TASK_ID -e VCS_BASE_REPOSITORY -e VCS_HEAD_REPOSITORY -e VCS_HEAD_REV -e VCS_REPOSITORY_TYPE mozillareleases/image_builder:5.1.0@sha256:7fe70dcedefffffa03237ba5d456d42e0d7461de066db3f7a7c280a104869cd5
Unable to find image 'mozillareleases/image_builder:5.1.0@sha256:7fe70dcedefffffa03237ba5d456d42e0d7461de066db3f7a7c280a104869cd5' locally
docker.io/mozillareleases/image_builder@sha256:7fe70dcedefffffa03237ba5d456d42e0d7461de066db3f7a7c280a104869cd5: Pulling from mozillareleases/image_builder
5a90e6335300: Pulling fs layer
5a90e6335300: Verifying Checksum
5a90e6335300: Download complete
5a90e6335300: Pull complete
Digest: sha256:7fe70dcedefffffa03237ba5d456d42e0d7461de066db3f7a7c280a104869cd5
Status: Downloaded newer image for mozillareleases/image_builder@sha256:7fe70dcedefffffa03237ba5d456d42e0d7461de066db3f7a7c280a104869cd5
[build-image] Downloading context.
[build-image] Building image.
INFO[0000] Resolved base name golang:1.23.2-bookworm to golang 
INFO[0000] Retrieving image manifest golang:1.23.2-bookworm 
INFO[0000] Retrieving image golang:1.23.2-bookworm from registry index.docker.io 
INFO[0000] Retrieving image manifest golang:1.23.2-bookworm 
INFO[0000] Returning cached image manifest              
[...]

PR #7600, opened 2025-03-26
D2G: use 'timeout -s TERM -k 10 --preserve-status

### LINKED ISSUE #7599: d2g: exceeding maxRuntime should not exit 137
When a docker-worker task exceeds maxRuntime, we get something like:
```
[taskcluster:error] Task timeout after 1800 seconds. Force killing container.
[taskcluster 2025-03-20 19:05:01.741Z] === Task Finished ===
[taskcluster 2025-03-20 19:05:01.741Z] Unsuccessful task run with exit code: -1 completed in 1990.763 seconds
```

With d2g:
```
bash: line 2: 296230 Killed                  timeout -s KILL 1800 docker run -t --name taskcontainer_H1Gd9mGsRj2KpKVbJQkWfw [...]
+ exit_code=137
+ docker cp taskcontainer_H1Gd9mGsRj2KpKVbJQkWfw:/builds/worker/workspace/logs/ artifact0
+ docker cp taskcontainer_H1Gd9mGsRj2KpKVbJQkWfw:/builds/worker/artifacts/ artifact1
+ docker cp taskcontainer_H1Gd9mGsRj2KpKVbJQkWfw:/builds/worker/workspace/build/blobber_upload_dir/ artifact2
+ docker rm -v taskcontainer_H1Gd9mGsRj2KpKVbJQkWfw
Error response from daemon: cannot remove container "/taskcontainer_H1Gd9mGsRj2KpKVbJQkWfw": container is running: stop the container before removing or force remove
+ exit 137
[taskcluster 2025-03-25T20:45:45.232Z]    Exit Code: 137
[taskcluster 2025-03-25T20:45:45.232Z]    User Time: 496.462ms
[taskcluster 2025-03-25T20:45:45.232Z]  Kernel Time: 1.487484s
[taskcluster 2025-03-25T20:45:45.232Z]    Wall Time: 30m17.163911038s
[taskcluster 2025-03-25T20:45:45.232Z]       Result: FAILED
[taskcluster 2025-03-25T20:45:45.232Z] === Task Finished ===
```

This changes the exit code, and if the task is set to retry on 137 (e.g. because it expects to hit OOM intermittently), we'll rerun instead of just setting the task status to failed.

petemoore:
Fixes #7599

Hopefully!


petemoore:
Ah, this might not be compatible with #7317, as it looks like it requires `--init` being passed to `docker run`...

```
# with --init works fine
$ timeout -s TERM -k 30 --preserve-status 5 docker run --init --rm --name test_oom alpine sh -c 'sleep 60'; echo $?
143

# without --init, it doesn't
$ timeout -s TERM -k 30 --preserve-status 5 docker run --rm --name test_oom alpine sh -c 'sleep 60'; echo $?
Killed: 9
137
```

matt-boris:
No, I don't remember. You could try adding `--init` then just make a change to one of the dockerfiles inside `taskcluster/docker` and you should be able to reproduce the issue from https://github.com/taskcluster/taskcluster/issues/7309.

petemoore:
Now I think about it, I think we decided we didn't want `--init` because this deviated from docker worker behaviour, and we probably don't want to introduce that.

So given that we don't want to use `--init` we are forced to use SIGKILL, otherwise the signal will not be handled. In this case, what we need is, if the docker run command exits with code 137 by itself, that we exit with code 137 too. However, if the timeout occurs, we want to kill the process, but we want our wrapper command (`timeout`, currently) to exit with a different code, even though the underlying command (`docker run`) will exit with code 137.

It doesn't seem that the `timeout` command supports this workflow. It exits with 137 if the underlying command exits with code 137 _or_ if it ends up killing the underlying command itself.

```
pmoore@Peters-MacBook-Pro-2:~ $ timeout -s KILL --preserve-status 5 docker run --rm alpine sh -c 'sleep 60'; echo $?
Killed: 9
137
pmoore@Peters-MacBook-Pro-2:~ $ timeout -s KILL 5 docker run --rm alpine sh -c 'sleep 60'; echo $?
Killed: 9
137
pmoore@Peters-MacBook-Pro-2:~ $ 
```

In other words, you can't see from the exit code, if timeout killed it, or something else killed it (e.g. linux OOM killer).

So I suspect we'll need to find an alternative to the `timeout` command, if we want to fix this issue. *sigh*

matt-boris:
Closing now that https://github.com/taskcluster/taskcluster/issues/7599 is fixed! 👍🏻

PR #7601, opened 2025-03-26
chore: bump `golangci-lint` major version

matt-boris:
Migrated config file with `golangci-lint migrate` command. Fixed remaining staticcheck issues afterwards.


.golangci.yml
```diff
@@ -1,109 +1,28 @@
...
-#   - containedctx
-#   - contextcheck
-#   - cyclop
-#   - deadcode
-#   - decorder
-#   - depguard
-#   - dogsled
-#   - dupl
-#   - dupword
-#   - durationcheck
```
petemoore, line 17: i wonder if we should review what all these linters throw up, just in case some of them are useful?

clients/client-shell/cmds/task/fetchers.go
```diff
@@ -85,9 +85,10 @@ func confirmMsg(command string, credentials *tcclient.Credentials, args []string
 
 		response = strings.ToLower(strings.TrimSpace(response))
 
-		if response == "y" || response == "yes" {
+		switch response {
+		case "y", "yes":
 			return true
-		} else if response == "n" || response == "no" {
+		case "n", "no", "":
```
petemoore, line 91: Nice fix adding the empty string! I assume you spotted that, and it wasn't automatic?

Should we also add "N" here, since the question makes it look like a capital N is accepted? If we do that, we should probably add Y to the case above, for consistency. I notice above in actors.go the capital form is accepted, so it would make it consistent with that too.

workers/generic-worker/helper_test.go
```diff
@@ -80,8 +80,8 @@ func scheduleNamedTask[P GenericWorkerPayload | dockerworker.DockerWorkerPayload
 		// horrible hack here, until we have jsonschema2go generating pointer types...
 		//
 		//////////////////////////////////////////////////////////////////////////////////
-		b = bytes.Replace(b, []byte(`"expires":"0001-01-01T00:00:00.000Z",`), []byte{}, -1)
-		b = bytes.Replace(b, []byte(`,"expires":"0001-01-01T00:00:00.000Z"`), []byte{}, -1)
+		b = bytes.ReplaceAll(b, []byte(`"expires":"0001-01-01T00:00:00.000Z",`), []byte{})
+		b = bytes.ReplaceAll(b, []byte(`,"expires":"0001-01-01T00:00:00.000Z"`), []byte{})
```
petemoore, line 84: Is this still needed? I thought there was some omitzero struct tag we could use or something... But I forget the details.

petemoore (approved):
Nice, thanks Boris!

clients/client-shell/cmds/task/fetchers.go
matt-boris: Tbh, this was GitHub copilot finishing up my switch statement and me not noticing until I put the PR up for review, like "huh, that's a a nice little addition to default to no if they don't enter anything."

Probably not _too_ big a deal, but we could followup in another PR if you think we should.

workers/generic-worker/helper_test.go
matt-boris: `omitzero` comes into play with go1.24. Which I still have a WIP PR opened up for https://github.com/taskcluster/taskcluster/pull/7512. Seems to be something to do with the new implementation of `user.Current()` on Windows that's screwing up our tests https://tip.golang.org/doc/go1.24#osuserpkgosuser. Would love to hear your thoughts on this if you ever get a chance to look into that PR and the test failures! 🙏🏻

PR #7607, opened 2025-03-31
fix(decision): don't remove `artifacts/actions.json` so CoT can still verify

matt-boris:
Prevents issues like:
```bash
export TASKCLUSTER_ROOT_URL='https://community-tc.services.mozilla.com/' && verify_cot --task-type decision a2d7wQF5QxeHl4l_wZFX5Q
scriptworker.exceptions.CoTError: 'path public/actions.json not in decision a2d7wQF5QxeHl4l_wZFX5Q chain of trust artifacts! This is likely a bug in a2d7wQF5QxeHl4l_wZFX5Q: docker-worker does not fail if a declared artifact was never generated by the task. Please download and make sure the artifact public/actions.json is the one you expect. If it is, then please reach out to the Release Engineering team. For more information: https://github.com/taskcluster/taskgraph/issues/47'
```

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -212,7 +212,7 @@ tasks:
               --head-ref='${head_ref}'
               --head-rev='${head_sha}'
               --head-tag='${head_tag}' &&
-              rm artifacts/actions.json # taskcluster is not using actions.json
+              echo "{}" > artifacts/actions.json # taskcluster is not using actions.json, setting to {} so CoT can still verify
 
           artifacts:
             "public":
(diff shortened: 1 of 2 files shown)


JohanLorenzo (approved):
LGTM! @Eijebong proposed this implementation. I don't know if it's going to work but the risk of making this change is low so let's land it and follow-up if needed.

Eijebong (approved):
https://taskcluster.bananium.fr/tasks/UE6zztOaTwuoNgnV0GQ6pQ/runs/0 FYI taskcluster looks perfectly happy with it, the original actions still work.

This doesn't solve the case when the decision task fails (at which point actions.json will still be written by taskgraph and will prevent you from calling rerun, but that was already the case with the `rm` way of doing this)

lotas (approved):
lgtm

PR #7624, opened 2025-04-05
Add logs to task failure email

### LINKED ISSUE #7197: On task failure notifications could include extra details
It might be useful to include few lines from logs or exit code of the task code to differentiate failure reasons

thakkarujas5:
Fixes #7197

### DIFF
--- a/services/notify/src/handler.js
+++ b/services/notify/src/handler.js
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import jsone from 'json-e';
 import { consume } from 'taskcluster-lib-pulse';
 import libUrls from 'taskcluster-lib-urls';
+import utils from './utils.js';
 
 /** Handler listening for tasks that carries notifications */
 class Handler {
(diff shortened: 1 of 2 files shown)


services/notify/src/handler.js
```diff
@@ -94,6 +94,9 @@ class Handler {
     // Load task definition
     let taskId = status.taskId;
     let task = await this.queue.task(taskId);
+    let artifact = await this.queue.latestArtifact(taskId, 'public/logs/live.log');
```
djmitche, line 97: In some installs, access to this log may be restricted, and it may contain sensitive information. So a few concerns here;
 - Check that `this.queue` has the correct credentials to represent a world-readable (e.g., email) view of the logs
 - If not, fail gracefully
 - Perhaps allow this to be disabled for a deployment, even if the log is "public" in terms of TC scopes (e.g., if the deployment is behind a firewall that would otherwise prevent the email recipients from reading the log)

petemoore:
Hi @thakkarujas5 are you still working on this? Thanks.

thakkarujas5:
Hi @petemoore , my work is a bit on hold, in case anyone else is working on the same issue, I'll close this PR.

PR #7626, opened 2025-04-07
Native GW support for taskcluster-proxy on docker-bridge network

### LINKED ISSUE #7625: Generic Worker: support taskcluster-proxy on docker-bridge network
This was previously only possible via D2G, but exposing as a native Generic Worker feature means tasks can use it that e.g. use Generic Worker features not available under Docker Worker.

petemoore:
Fixes #7625

### DIFF
--- a/workers/generic-worker/host/host.go
+++ b/workers/generic-worker/host/host.go
@@ -3,6 +3,7 @@
 package host
 
 import (
+	"io"
 	"log"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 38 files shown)


workers/generic-worker/taskcluster_proxy.go
```diff
@@ -57,12 +57,13 @@ func (l *TaskclusterProxyTask) RequiredScopes() scopes.Required {
...
 func (l *TaskclusterProxyTask) Start() *CommandExecutionError {
-	if l.task.D2GInfo != nil {
-		out, err := exec.Command("docker", "network", "inspect", "bridge", "--format", "{{range .IPAM.Config}}{{if .Gateway}}{{.Gateway}} {{end}}{{end}}").Output()
+	switch l.task.Payload.TaskclusterProxyInterface {
+	case "docker-bridge":
+		out, err := host.CombinedOutput("docker", "network", "inspect", "bridge", "--format", "{{range .IPAM.Config}}{{if .Gateway}}{{.Gateway}} {{end}}{{end}}")
 		if err != nil {
 			return executionError(internalError, errored, fmt.Errorf("could not determine docker bridge IP address: %s", err))
 		}
-		gateways := strings.Split(strings.TrimSpace(string(out)), " ")
```
matt-boris, line 65: I had initially suggested this in https://github.com/taskcluster/taskcluster/pull/7564#discussion_r1979583155 but @jcristau preferred the `exec.Command().Output()` method since he explicitly did not want stderr to be included here.

matt-boris (commented):
I like this change Pete, thanks. Added one small comment about switching the `exec.Command().Output()` over to `host.CombinedOutput()`.

workers/generic-worker/taskcluster_proxy.go
matt-boris: Maybe we can make a new helper that you can get stdout and stderr separately assigned in one call? I think having docker's stderr in the `executionError` might be beneficial here.

workers/generic-worker/host/host.go
```diff
@@ -12,14 +13,53 @@ import (
...
+	}
+
+	if err := cmd.Wait(); err != nil {
+		log.Printf("Error running command: %v", err)
+		log.Printf("Standard output:\n%s", stdoutBytes)
+		log.Printf("Standard error:\n%s", stderrBytes)
+		return "", err
+	}
+
+	return string(stdoutBytes), nil
```
matt-boris, line 62: This doesn't seem to match the docstring?

```go
// Output is like CombinedOutput but only returns Standard Error output.
```

suggests this should be:
```suggestion
	return string(stderrBytes), nil
```

Or is the docstring wrong? Or am I reading this incorrectly? 😄

workers/generic-worker/host/host.go
```diff
@@ -12,14 +13,53 @@ import (
...
 }
 
 // CombinedOutput is equivalent to exec.Command(name, arg...).CombinedOutput()
 // but with logging.
 func CombinedOutput(name string, arg ...string) (combinedOutput string, err error) {
-	return RunCommand(exec.Command(name, arg...))
+	return runCommand(exec.Command(name, arg...))
+}
+
+// Output is like CombinedOutput but only returns Standard Error output. If an
```
matt-boris, line 26:
```suggestion
// Output is like CombinedOutput but only returns Standard Out output. If an
```

workers/generic-worker/host/host.go
matt-boris: I think, based on how you're using this function, that the docstring is wrong. I've made a suggestion to fix.

workers/generic-worker/host/host.go
```diff
@@ -36,7 +76,7 @@ func RunBatch(allowFail bool, commands ...[]string) (err error) {
 			}
 		}
 	}
-	return err
+	return nil
```
matt-boris, line 79: Did you do this just for safety/sanity? (and in other places)

matt-boris (commented):
Changes LGTM Pete. I think there's just a small docstring issue in the new `host.Output()` function you made.

workers/generic-worker/host/host.go
matt-boris: Feel free to reword if my suggestion doesn't sound great `...Out output...` 😬

workers/generic-worker/host/host.go
petemoore: Well spotted! I've applied your fix to the docstring. Thanks!

Issue #7623, opened 2025-04-03
generic-worker parallel upload behaviour is unclear


bhearsum:
(This is going to be slightly nebulous, but for now I just want to file this to capture some state...)

Over in https://github.com/mozilla/translations/issues/1067#issuecomment-2776647845 I'm investigating a spot termination that didn't have time to upload all of its artifacts. The worker was running generic-worker 64.2.6 which appears to be the first version with [parallel upload support](https://github.com/taskcluster/taskcluster/commit/31dbbbbfee14b055139ee434a17c13bb018885c2).

Here's some log excerpts:
```
'2025-04-03T00:21:05.703482682Z"00:21:04 PUT /firefoxcitc-public-artifacts/SlOFvvBJQOCkvsCSdx0m-g/1/public/build/model.npz.decoder.yml
'2025-04-03T00:21:06.698549871Z"00:21:04 PUT /firefoxcitc-public-artifacts/SlOFvvBJQOCkvsCSdx0m-g/1/public/build/opustrainer.log
'2025-04-03T00:21:06.698549871Z"00:21:04 PUT /firefoxcitc-public-artifacts/SlOFvvBJQOCkvsCSdx0m-g/1/public/build/model.npz.yml
'2025-04-03T00:21:06.698549871Z"00:21:04 PUT /firefoxcitc-public-artifacts/SlOFvvBJQOCkvsCSdx0m-g/1/public/build/config.opustrainer.yml
'2025-04-03T00:21:07.700546829Z"00:21:04 PUT /firefoxcitc-public-artifacts/SlOFvvBJQOCkvsCSdx0m-g/1/public/build/model.npz.best-bleu-detok.npz.decoder.yml
'2025-04-03T00:21:07.700546829Z"00:21:04 PUT /firefoxcitc-public-artifacts/SlOFvvBJQOCkvsCSdx0m-g/1/public/build/model.npz.best-chrf.npz.decoder.yml
'2025-04-03T00:21:07.700546829Z"00:21:04 PUT /firefoxcitc-public-artifacts/SlOFvvBJQOCkvsCSdx0m-g/1/public/build/config.opustrainer.yml.state
'2025-04-03T00:21:07.700546829Z"00:21:04 PUT /firefoxcitc-public-artifacts/SlOFvvBJQOCkvsCSdx0m-g/1/public/build/model.npz.progress.yml
[...]

PR #7627, opened 2025-04-07
chore(generic-worker): add artifact upload worker logs

matt-boris:
Will help in giving additional insight for #7623.

cc @bhearsum

### DIFF
--- a/changelog/A_K0AWj_S--QETrODMypFA.md
+++ b/changelog/A_K0AWj_S--QETrODMypFA.md
@@ -1,0 +1,4 @@
+audience: worker-deployers
+level: patch
+---
+Generic Worker: adds additional worker logs around the artifact upload process to help debug artifact upload performance.
--- a/workers/generic-worker/artifacts/s3.go
+++ b/workers/generic-worker/artifacts/s3.go
@@ -44,13 +44,24 @@ type S3Artifact struct {
 // use this cleanup function.
 func (s3Artifact *S3Artifact) createTempFileForPUTBody() (string, func()) {
 	var fileLock *flock.Flock
+	removeFileLock := func() {
+		log.Printf("Removing exclusive file lock on %v...", s3Artifact.ContentPath)
+		_ = fileLock.Unlock()
+	}
+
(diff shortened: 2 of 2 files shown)


petemoore (approved):
I wonder if this makes sense as a permanent change, or better for a one-off diagnosis? My only worry about landing this is we will get these logs for every future task artifact published, so this could generate quite a lot of noise which might make logs harder to read and generally noisier. Perhaps we just need to deploy it once to staging and look at the results?

Or we could land, and revert in a future release?

bhearsum:
> I wonder if this makes sense as a permanent change, or better for a one-off diagnosis? My only worry about landing this is we will get these logs for every future task artifact published, so this could generate quite a lot of noise which might make logs harder to read and generally noisier. Perhaps we just need to deploy it once to staging and look at the results?
> 
> Or we could land, and revert in a future release?

This might partly depend if we want diagnostics from a full real world failure, or if a somewhat simulated repro will do. If we need the former, we need to go to prod with it.

bhearsum:
> > I wonder if this makes sense as a permanent change, or better for a one-off diagnosis? My only worry about landing this is we will get these logs for every future task artifact published, so this could generate quite a lot of noise which might make logs harder to read and generally noisier. Perhaps we just need to deploy it once to staging and look at the results?
> > Or we could land, and revert in a future release?
> 
> This might partly depend if we want diagnostics from a full real world failure, or if a somewhat simulated repro will do. If we need the former, we need to go to prod with it.

And actually, either way I'm going to need this in a production worker pool...I wasn't able to get GPU tasks running in my own worker on stage. I need full on images from https://github.com/mozilla-platform-ops/worker-images for it.

Perhaps we could make the logging conditional in some way though? eg: we could increase the verbosity of the logs from translations workers? Or enable these specific things with an env var or config entry?

petemoore:
I'd say let's go with it as it is, with the intention of adding verbosity settings in future (something @matt-boris has also proposed). Note, this isn't a huge amount of logging per artifact, so overall i would only expect a marginal increase in log size overall.

PR #7654, opened 2025-04-14
feat(ui): Fix third-party login issue, improve error handling

lotas:
Improves error handling and fixes some issues with 3rd party login

### DIFF
--- a/services/web-server/src/main.js
+++ b/services/web-server/src/main.js
@@ -149,7 +149,6 @@ const load = loader(
         const server = new ApolloServer({
           schema,
           formatError,
-          tracing: true,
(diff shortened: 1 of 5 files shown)


services/web-server/src/main.js
```diff
@@ -149,7 +149,6 @@ const load = loader(
         const server = new ApolloServer({
           schema,
           formatError,
-          tracing: true,
```
lotas, line 152: typescript says this option does not exist.. I tend to believe, since this was added > 7y ago

Eijebong: https://github.com/apollographql/apollo-server/commit/06a9532d3fda6ac3637748dc9aa5b0697703032f

Yup, typescript is right

Eijebong (approved):

PR #7658, opened 2025-04-15
fix(generic-worker): clean up temp files

matt-boris:
>Generic Worker: remove temp file created while adding additional data to chain of trust file.


lotas (approved):
LGTM

PR #7660, opened 2025-04-16
Generic Worker: move remaining task features into dedicated go types

### LINKED ISSUE #7716: Generic Worker - refactor task run steps into task features
The code that Generic Worker calls to run tasks performs several explicit startup steps, iterates through task feature `Start` methods, task feature `Stop` methods, then performs other explicit teardown steps. The explicit startup and teardown steps can also be moved into task features, so that the task run code is simplified and the feature code is more isolated.

petemoore:
Fixes #7716

Some final feature cleanup in preparation for adding launch agent support.


matt-boris (commented):
Most of the files in our main package are named with underscores separating out individual words. I like standardizing on all task features having filenames ending in `_feature.go` like you've done in this PR (minus the underscore) as well as how we've done for d2g, artifact, and metadata. Do you mind renaming your new files to include the `_` between the feature name and `feature.go`? 🙏🏻

May also be a good time to _add_ `_feature` suffix to the end of any other task features we may have (interactive, livelog, loopback audio/video, mounts, os groups, rdp, resource monitor, run task as current user, run as admin, tc proxy)? Wdyt? I think this could help navigating through our large main package, files being immediately clear that they're task features.

I'll dive further into your changes now! Just wanted to call this out in the meantime.

workers/generic-worker/main.go
```diff
@@ -81,10 +81,10 @@ func initialiseFeatures() (err error) {
 	}
 	features = append(features, platformFeatures()...)
 	for _, feature := range features {
-		log.Printf("Initialising task feature %v...", feature.Name())
```
matt-boris, line 84: I agree with this cleanup, thanks. 👍🏻 Since not all features are _task_ features.

matt-boris (approved):
This is excellent stuff here Pete! Nice!! 🎉 

Thanks so much for splitting it up into separate commits. Made reviewing _much_ easier. LG(reat)TM

PR #7661, opened 2025-04-16
fix(generic-worker): remove redundant file copy during artifact upload process

### LINKED ISSUE #7368: Generic worker seems to unnecessarily copy artifacts to `/tmp` before uploading them

matt-boris:
Fixes #7368.

>Generic Worker: increase performance of artifact uploads by removing a redundant file copy operation.

The additional logs should give more insight into #7623.

### DIFF
--- a/changelog/issue-7368.md
+++ b/changelog/issue-7368.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 7368
+---
+Generic Worker: increase performance of artifact uploads by removing a redundant file copy operation.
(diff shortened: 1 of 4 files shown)


matt-boris:
We were very close beforehand...

```bash
git diff -C30 062d86a9f4386d82a42a6a9c66b8f6611238ab0d..3deebda01af41eb69380bfa7e9a17bfa9ffd132c workers/generic-worker/artifacts/s3.go
```

petemoore (approved):
Love it

PR #7662, opened 2025-04-16
chore(ui): small navigation and filtering tweaks

matt-boris:
>UI: don't show `Requested` filter option on `Queue Workers` view, link to `W-M Workers` filtered view when clicking on `Requested Capacity` or `Stopping Capacity` buttons, don't redirect to Worker Manager Worker Pools page after saving Worker Pool config changes.


lotas (approved):
Cool, thanks!

PR #7665, opened 2025-04-22
fix(docs): update gw installation docs for macos

matt-boris:


petemoore (approved):
Love it! Thanks so much!

PR #7672, opened 2025-04-24
fix(d2g): remove unneeded os groups when using kvm

### LINKED ISSUE #7671: d2g: don't add the task user to the libvirt and kvm groups
Since the switch from podman to docker, the task user doesn't need to be in extra groups to use `--device=/dev/kvm`.

matt-boris:
Fixes #7671.

>D2G: removes unneeded `kvm` and `libvirt` OS groups for generic worker task user if KVM device is requested.

### DIFF
--- a/changelog/issue-7671.md
+++ b/changelog/issue-7671.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 7671
+---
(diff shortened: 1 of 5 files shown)


petemoore (approved):
lgtm, thanks!

PR #7673, opened 2025-04-24
fix(worker-runner): retry worker registration

matt-boris:
>Worker Runner: retries to register a worker 3 times and will shutdown if unsuccessful.

### DIFF
--- a/changelog/CmIHaI1ASc-83oryuNbZPg.md
+++ b/changelog/CmIHaI1ASc-83oryuNbZPg.md
@@ -1,0 +1,4 @@
+audience: worker-deployers
+level: patch
+---
+Worker Runner: retries to register a worker 3 times and will shutdown if unsuccessful.
--- a/tools/worker-runner/runner/run.go
+++ b/tools/worker-runner/runner/run.go
@@ -3,6 +3,7 @@ package runner
 import (
 	"fmt"
 	"log"
+	"time"
 
 	"github.com/taskcluster/taskcluster/v83/tools/worker-runner/cfg"
 	"github.com/taskcluster/taskcluster/v83/tools/worker-runner/errorreport"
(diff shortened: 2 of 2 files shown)


petemoore:
I don't think this is needed. I believe the issue is that Worker Manager is returning a 4xx status code when it can't retrieve the intermediate cert, rather than a 5xx status code, which signals to the client (worker runner) not to retry. By fixing Worker Manager to return a server error code (503 or even 504), worker runner should take care of retries automatically. Worker runner shouldn't retry requests that return 4xx status codes, so is behaving correctly.

matt-boris:
Ok sure, thanks @petemoore makes sense. But there's still a part of this that I'm wondering if we should include, which is shutting down the worker on error? https://github.com/taskcluster/taskcluster/pull/7673/files#diff-1a754b2cdf693c16ddcc2904d047f05ec1fde4441375f2672877713d6be0761dR83

That way, worker manager would know to spawn a new VM much sooner than `queueInactivityTimeout`.

petemoore:
> Ok sure, thanks @petemoore makes sense. But there's still a part of this that I'm wondering if we should include, which is shutting down the worker on error? https://github.com/taskcluster/taskcluster/pull/7673/files#diff-1a754b2cdf693c16ddcc2904d047f05ec1fde4441375f2672877713d6be0761dR83
> 
> That way, worker manager would know to spawn a new VM much sooner than `queueInactivityTimeout`.

For posterity:

We discussed on Slack that whenever worker runner exits due to a failure, it should probably shut down the worker, if it is a worker manager spawned (i.e. cloud) worker, in order that it is terminated and not incurring further costs. This is a more general case than the one mentioned here, and would therefore encompass it.

Hardware (static/standalone) workers probably should not shut down by default on failure, although it may be an option to preserve running costs.

Note, this is how generic worker behaved before worker runner was created.

PR #7676, opened 2025-04-28
Add windows support for the interactive feature

Eijebong:
The main code here is the conpty implementation because none of the implementation I found online worked properly.


workers/generic-worker/interactive/interactivejob_windows.go
```diff
@@ -0,0 +1,28 @@
+//go:build windows
+
```
matt-boris, line 2:
```suggestion
```

Safe to remove due to filename.

workers/generic-worker/multiuser_windows.go
```diff
@@ -30,6 +33,7 @@ func platformFeatures() []Feature {
 		&RunTaskAsCurrentUserFeature{},
 		&RDPFeature{},
 		&RunAsAdministratorFeature{}, // depends on (must appear later in list than) OSGroups feature
+		&InteractiveFeature{},
```
matt-boris, line 36:
```suggestion
```

This can now be moved to the global features list across all platforms https://github.com/taskcluster/taskcluster/blob/023387cb237ff46f60a3a8838504e850c356c20d/workers/generic-worker/main.go#L74-L79.

Please be sure to remove `InteractiveFeature` from the various `platformFeatures()` functions. Looks like `workers/generic-worker/insecure_other.go` can safely be removed entirely, also.

workers/generic-worker/interactive/conpty.go
```diff
@@ -0,0 +1,289 @@
+//go:build windows
+
```
matt-boris, line 2: Mind just updating the filename to `conpty_windows.go`? This'll also make it clear from the filetree within the repo this code is specific to windows just from the filename.

```suggestion
```

workers/generic-worker/gwconfig/config_windows.go
```diff
@@ -3,12 +3,14 @@ package gwconfig
 import "testing"
 
 type PublicPlatformConfig struct {
+	EnableInteractive        bool `json:"enableInteractive"`
```
matt-boris, line 6: This can now live in the global `PublicConfig` and set the default in `loadConfig()`. We'll need to clean up the other files that separate out this flag based on the platform.

Looks like the following changes would be needed here:
* Removed from `workers/generic-worker/gwconfig/config_linux.go`
* I believe `workers/generic-worker/gwconfig/config_other.go` can entirely be removed

workers/generic-worker/multiuser_windows.go
Eijebong: `vet: ./main.go:82:30: undefined: platformFeatures` can't remove the file no, have to return an empty list of features.

matt-boris (approved):
Changes LGTM, thanks for the cleanup too! Glad we now have the interactive feature across the board! 🎉 Awesome work, as always, thanks so much.

petemoore (approved):
Looks good to me, thanks Bastien! Note, we do have a win32 package where until now we have placed added all the win32 APIs that we use, however, I can see an argument for keeping it directly in the interactive package too, to keep it local to the package. Many thanks!

PR #7679, opened 2025-04-29
Add API to fetch client audit history

### LINKED ISSUE #7548: Audit history improvements

thakkarujas5:
Fixes #7548 

- As a precursor to UI changes, an API must be introduced to fetch a client's audit history.
- This API will fetch entityId, entityType, action_type and created against a client_id.


changelog/issue-7548.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: minor
+reference: issue 7548
+---
+Add api  in auth service to fetch entityId, entityType, actionType, created against a clientId
```
lotas, line 5:
```suggestion
Auth service  exposes `ListAuditHistory` endpoint for retrieving audit history by entity ID, entity type, and client ID.
```
Let's be more specific for the changelog

services/auth/src/api.js
```diff
@@ -376,6 +386,35 @@ builder.declare({
 
   return res.reply({ auditHistory: rows.map(c => auditToJson(c)), continuationToken });
 });
+
+builder.declare({
+  method: 'get',
+  route: '/audit-logs/:clientId',
```
lotas, line 392: this route is a bit inconsistent with existing one (audit vs audit-logs)
```
  route: '/audit/:entityType/:entityId',
```

Plus, maybe it should be defined as a client-specific route:
```
route: '/clients/:clientId/audit',
```

db/versions/0111.yml
```diff
@@ -0,0 +1,30 @@
+version: 111
+description: fetch audit history for a client
+methods:
```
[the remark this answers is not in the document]
thakkarujas5: added migration and downgrade for index

services/auth/schemas/v1/get-client-history-response.yml
```diff
@@ -0,0 +1,36 @@
...
+description: |
+  Response from listAuditHistory endpoint containing audit history entries for a client
+type: object
+properties:
+  auditHistory:
+    type: array
+    items:
+      type: object
+      properties:
+        entity_id:
```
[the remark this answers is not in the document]
thakkarujas5: changed the output schema with camelCase

db/versions/0111.yml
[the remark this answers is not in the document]
lotas: thanks

services/auth/src/api.js
```diff
@@ -374,8 +376,45 @@ builder.declare({
     maxLimit: 1000,
   });
 
+  const rowsWithEntityInformation = rows.map((row) => {
+    return {
+      ...row,
+      entity_id: entityId,
+      entity_type: entityType,
```
lotas, line 383:
```suggestion
      entityId,
      entityType,
```

services/auth/schemas/v1/get-client-history-response.yml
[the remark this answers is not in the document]
lotas: thank you! looks much cleaner and better :)

services/auth/src/api.js
thakkarujas5: so in the getEntityHistory and listAuditHistory are both using the same auditToJson function. In getEntityHistory I am injecting the values of entity_id and entity_type. But in listAuditHistory the values are coming from the db. In the table the casing is entity_id and entity_type. So in order to be able to use the same auditToJson function in both api's keep this casing.

lotas: Yes, but shouldn't it be camelcased here?

db/versions/0111.yml
```diff
@@ -0,0 +1,65 @@
...
+        action_type text,
+        created timestamptz
+      )
+    body: |-
+      begin
+        -- Handle case where no search criteria is provided
+        if client_id_in is null and (entity_id_in is null or entity_type_in is null) then
+          raise exception 'At least client_id or both entity_id and entity_type must be provided';
+        end if;
+        if client_id_in is not null and (entity_id_in is null or entity_type_in is null) then
```
lotas, line 37: I'm wondering if we should make it that complicated here. 
It makes sense to check if entity type is provided when entity id is in, but otherwise we can allow filter by both at the same time, why not.

we can use similar approach like in https://github.com/taskcluster/taskcluster/blob/511974dacf86b36e8096f39498da1cbeedbfe7a8/db/versions/0105.yml#L427-L439

```sql
(audit_history.client_id = client_id_in OR client_id_in IS NULL) AND
(
   (audit_history.entity_id = entity_id_in AND audit_history.entity_type = entity_type_in) OR entity_id_in IS NULL
)
```
(needs checking ^ :)

lotas (approved):
This is awesome! Thank you so much for the feature and the updates

PR #7681, opened 2025-04-29
feat(d2g): convert d2g to generic worker task feature

### LINKED ISSUE #7599: d2g: exceeding maxRuntime should not exit 137

### LINKED ISSUE #7678: d2g: ensure the container is stopped on timeout
`timeout -s KILL $maxruntime docker run ...` ends up sending terminating the `docker run` command on timeout, but leaving the container itself running.  That causes the subsequent `docker rm` to fail (which presumably also means we leak the container and its volumes?).

matt-boris:
Fixes #7678.
Fixes #7599.

>D2G: pre and post-task-processing (image pulling/loading/saving, copying artifacts out of container, creating chain of trust additional data file, removing container/volumes, handling max runtime) now happens within the D2G Task Feature in Generic Worker, as opposed to within the resulting translated task payload. This slims the translated task payload to only the `docker run ...` command.

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -22,8 +22,6 @@ import (
 	"github.com/taskcluster/shell"
 )
 
-var MaxArtifactCopyDuration int64 = 900
-
 type (
 	DockerImageName     string
 	IndexedDockerImage  dockerworker.IndexedDockerImage
@@ -35,11 +33,18 @@ type (
 		ImageLoader() ImageLoader
 	}
 	ImageLoader interface {
-		LoadCommands() []string
-		ChainOfTrustCommands() []string
+		LoadCommand() string
+		ChainOfTrustCommand() string
 	}
 	ConversionInfo struct {
 		ContainerName string
+		CopyArtifacts []CopyArtifact
+		Image         Image
+	}
+	CopyArtifact struct {
+		Name     string
+		SrcPath  string
+		DestPath string
 	}
 	FileImageLoader struct {
 		Image Image
@@ -49,7 +54,11 @@ type (
 	}
 )
 
-func ConvertTaskDefinition(dwTaskDef json.RawMessage, config map[string]any, scopeExpander scopes.ScopeExpander) (json.RawMessage, error) {
+func ConvertTaskDefinition(
+	dwTaskDef json.RawMessage,
+	config map[string]any,
+	scopeExpander scopes.ScopeExpander,
+) (json.RawMessage, error) {
 	var gwTaskDef json.RawMessage
 	var parsedTaskDef map[string]any
 	err := json.Unmarshal(dwTaskDef, &parsedTaskDef)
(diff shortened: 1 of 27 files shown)


workers/generic-worker/d2g_feature.go
```diff
@@ -0,0 +1,134 @@
...
+
+	for _, artifact := range dtf.task.D2GInfo.CopyArtifacts {
+		out, e := host.Output(
+			"docker",
+			"cp",
+			fmt.Sprintf("%s:%s", dtf.task.D2GInfo.ContainerName, artifact.SrcPath),
+			filepath.Join(taskContext.TaskDir, artifact.DestPath),
+		)
+		if e != nil {
+			dtf.task.Warnf("Artifact %q not found at %q: %v\n%v", artifact.Name, artifact.SrcPath, e, out)
```
matt-boris, line 120: Just log a warning if the artifact isn't found in the container, as docker worker does: https://github.com/taskcluster/taskcluster/blob/main/workers/docker-worker/src/features/artifacts.js#L52-L55

lotas:
Does the feature name has to mention `d2g` in it's name? Or can it be also called `DockerFeature` or something?

tools/d2g/d2g.go
```diff
@@ -443,12 +404,18 @@ func runCommand(containerName string, dwPayload *dockerworker.DockerWorkerPayloa
 	// note, dwImage.String() is already shell escaped
 	command.WriteString(" " + dwImage.String())
 	command.WriteString(" " + shell.Escape(dwPayload.Command...))
-	return command.String(), nil
+	return [][]string{
+		{
+			"/usr/bin/env",
```
lotas, line 409: is it still necessary to use env bash to run this? It looks like it is just `docker run` now, or can it contain more commands?

lotas (commented):
Tremendous refactoring! LGTM

tools/d2g/d2g.go
petemoore: i think this is just lifting what we had, essentially. there will be further cleanup in future PRs i believe that simplifies this and several other things (such as not shelling out to a script to perform logic that can be performed directly in go, for logic that is now part of the feature, and no longer part of the task commands).

petemoore (approved):
Thanks Boris!

PR #7698, opened 2025-05-04
Properly reap the proxy process when stopping the feature

### LINKED ISSUE #5786: generic-worker leaves zombie taskcluster-proxy processes behind
image

Version: 44.2.2

Eijebong:
On posix systems, when killing a child process, the parent process has to call `wait(2)` on it or the child process will be left in a "zombie" state until the parent exits, at which point PID 1 will do that.

For generic worker, it means that every time it starts a proxy, it leaks one zombie as it's never waiting for it until it gets restarted which in some cases could be never (headless multiuser, insecure engine).

By simply calling `Wait` on the proxy process after killing it, it now ends up calling `wait(2)` on it, reaping it. We simply ignore the actual return status and print an error in case something went horribly wrong.

I got bitten by this this weekend where after processing a few hundred tasks, linux was just refusing to start new processes leading to the worker being unable to complete any task.

Fixes #5786

### DIFF
--- a/changelog/issue-5786.md
+++ b/changelog/issue-5786.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: issue 5786
+---
+Properly reap proxy processes when stopping the taskcluster proxy feature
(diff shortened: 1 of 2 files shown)


matt-boris (approved):
Nice, LGTM thank you.

PR #7699, opened 2025-05-05
Fix azure provider registerWorker

### LINKED ISSUE #7685: node-forge and Azure certificates duality
**Subject: Chain verification fails due to caStore state mutation when handling duplicate Subject DNs**

**Context:**
In PR #7674, we updated our bundled Azure Intermediate CAs, removing expired ones and adding the currently used set (Microsoft Azure RSA TLS Issuing CA 03, 04, 07, 08).

**Problem Description:**
Azure provides these intermediate CAs in two forms for compatibility:
1.  **Direct-signed:** Signed directly by `Microsoft RSA Root Certificate Authority 2017`.
2.  **Cross-signed (`-xsign`):** Signed by a DigiCert CA (ultimately chaining to `DigiCert Global Root G2`).

Crucially, for a given CA number (e.g., CA 03), both the direct-signed and cross-signed certificates share the **exact same Subject DN**, but have **different Issuer DNs** and were signed by different keys.

Example (`CA 03`):
*   `microsoft_azure_rsa_tls_issuing_ca_03.pem` (Direct-signed)
    *   Subject: `O=Microsoft Corporation, CN=Microsoft Azure RSA TLS Issuing CA 03, C=US`
    *   Issuer: `O=Microsoft Corporation, CN=Microsoft RSA Root Certificate Authority 2017, C=US`
*   `microsoft_azure_rsa_tls_issuing_ca_03_xsign.pem` (Cross-signed)
    *   Subject: `O=Microsoft Corporation, CN=Microsoft Azure RSA TLS Issuing CA 03, C=US`
    *   Issuer: `O=DigiCert Inc, OU=www.digicert.com, CN=DigiCert Global Root G2, C=US` *(or an intermediate under it)*

The `node-forge` `pki.createCaStore()` uses a hash of the **Subject DN** as the primary key (`caStore.certs[cert.subject.hash]`). When both certificate versions (direct and cross-signed) are added via `caStore.addCertificate`, they generate the **same subject hash**. The store handles this by storing an **array** of certificates under the colliding hash key.

The core issue arises within `pki.verifyCertificateChain`. When attempting to verify a certificate (`cert`), it retrieves potential issuers using `caStore.getIssuer(cert)`. If multiple potential issuers exist due to the subject hash collision (i.e., `getIssuer` returns the stored array), the verification loop iterates through them:

```javascript
// Inside pki.verifyCertificateChain...
var parents = parent; // parent might be the array from getIssuer
if(!forge.util.isArray(parents)) {
  parents = [parents];
}
while(!verified && parents.length > 0) {
  // !!! MUTATION !!!
  parent = parents.shift(); // This removes an element from the 'parents' array
  try {
    verified = parent.verify(cert);
  } catch(ex) { /* ... */ }
}
```

If the `parents` variable holds a direct reference to the array stored within the `caStore`, the `parents.shift()` call **mutates the `caStore`'s internal state** by removing the certificate being tested in that iteration.

**Symptoms:**
1.  **Verification Failure:** When `pki.verifyCertificateChain` is called, if the *first* potential parent certificate removed by `shift()` is not the correct one to verify the signature, the loop continues. If the *correct* parent certificate is eventually found *later* in the array, the verification might succeed *on this first call*. However, if the correct parent was tried *first* and failed verification (perhaps due to an unrelated issue or simply being the wrong key), or if the correct parent isn't present *at all* initially, the verification fails with `"Error verifying certificate chain” - “Certificate signature is invalid.”`.
2.  **Inconsistent State:** **Crucially, subsequent calls to `pki.verifyCertificateChain` (or manual lookups like `caStore.getIssuer`) involving the same issuer subject DN will operate on a modified `caStore`**. Certificates previously removed by `shift()` will be missing. This leads to inconsistent behavior: a verification might succeed once but fail later, or vice-versa, depending on the order of operations and which certificates remain in the store's array. We confirmed this by observing changes in the number of stored certificates before and after calls to `verifyCertificateChain`.

**Expected Behavior:**
1.  `caStore.getIssuer` should reliably find and return *all* certificates in the store whose Subject matches the target Issuer DN.
2.  **`pki.verifyCertificateChain` must perform verification without modifying the state of the `caStore` it is given.** It should operate on a *copy* of the potential parents array (e.g., `var parents = (parent ? [].concat(parent) : []);`) to avoid mutating the original store via the `shift()` operation.

**Workaround:**
Currently, we can either reload all certificates after each check, or give a cloned CA Store to the `verifyCertificateChain()` function to avoid mutations

lotas:
The problem is that we load both direct-signed and cross-signed certificates which have same subject but different issuers.

node-forge has [a bug](https://github.com/digitalbazaar/forge/issues/1003) which allows store to be mutated during verification process, which is not an expected behavior.

To fix this, we clone store for those checks to prevent mutations and unexpected failures

Fixes #7685

### DIFF
--- a/changelog/issue-7685.md
+++ b/changelog/issue-7685.md
@@ -1,0 +1,8 @@
+audience: worker-deployers
+level: patch
+reference: issue 7685
+---
+
+Fixes node-forge issue with certificates being removed during chain verification process,
+which allowed Azure `registerWorker()` calls fail after some time.
+This happened when multiple certs had same subject hash but different issuers.
--- a/services/worker-manager/src/providers/azure/index.js
+++ b/services/worker-manager/src/providers/azure/index.js
@@ -14,7 +14,7 @@ import msRestAzure from '@azure/ms-rest-azure-js';
 import { ApiError, Provider } from '../provider.js';
 import { CloudAPI } from '../cloudapi.js';
 import { loadCertificates } from './azure-ca-certs/index.js';
-import { nicerId, dnToString, workerConfigWithSecrets, getCertFingerprint, getAuthorityAccessInfo } from './utils.js';
+import { nicerId, dnToString, workerConfigWithSecrets, getCertFingerprint, getAuthorityAccessInfo, cloneCaStore } from './utils.js';
 
 /** @typedef {import('../../data.js').WorkerPoolStats} WorkerPoolStats */
 
@@ -506,7 +506,15 @@ export class AzureProvider extends Provider {
 
     // verify that the embedded certificates have proper chain of trust
     try {
-      forge.pki.verifyCertificateChain(this.caStore, [crt]);
+      // Verification can mutate store certificates when Azure uses
+      // multiple certificates with the same hash but different issuer chains
+      // (direct-signed vs cross-signed), potentially causing future request failures
+      // https://github.com/digitalbazaar/forge/issues/1003
+      // https://github.com/taskcluster/taskcluster/issues/7685
+      forge.pki.verifyCertificateChain(
+        cloneCaStore(this.caStore),
+        [crt],
+      );
     } catch (err) {
       this.monitor.log.registrationErrorWarning({
         message: 'Error verifying certificate chain',
--- a/services/worker-manager/src/providers/azure/utils.js
+++ b/services/worker-manager/src/providers/azure/utils.js
@@ -126,3 +126,31 @@ export function getAuthorityAccessInfo(cert) {
   });
   return accessDescriptions;
 }
+
+/**
+ * Creates a clone of a node-forge CA certificate store.
+ * The clone will have its own internal certificate map, preventing
+ * modifications (like array shifts during verification) in the clone
+ * from affecting the original store.
+ * Note: The certificate objects themselves are shared by reference,
+ * assuming they are treated as immutable by the relevant operations.
+ *
+ * @param {object} originalCaStore - The node-forge CA store to clone.
+ * @returns {object} A new, independent CA store object.
+ */
+export function cloneCaStore(originalCaStore) {
+  if (!originalCaStore || typeof originalCaStore.certs !== 'object') {
+    throw new Error("Invalid input: Not a CA store object.");
+  }
+
+  const newCaStore = forge.pki.createCaStore();
+
+  for (const value of Object.values(originalCaStore.certs)) {
+    const certs = Array.isArray(value) ? value : [value];
+    for (const cert of certs) {
+      newCaStore.addCertificate(cert);
+    }
+  }
+
+  return newCaStore;
+}
--- a/services/worker-manager/test/provider_azure_test.js
+++ b/services/worker-manager/test/provider_azure_test.js
@@ -5,7 +5,7 @@ import assert from 'assert';
 import helper from './helper.js';
 import { FakeAzure } from './fakes/index.js';
 import { AzureProvider } from '../src/providers/azure/index.js';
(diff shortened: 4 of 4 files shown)


services/worker-manager/test/fixtures/azure_signature_good.json
```diff
@@ -0,0 +1,7 @@
...
...5Fbbv1pNdhWnuw3mV5X42w8Y5b9YpDoWY3BApqfxCqyEvMxNx/eM9+Qmdn9c4shGD9G5bxkfhsjDR63jt8+fV6psNyLB3Pe2nIo/vKmKcm/tvkvsvkMTY/JRLfE8VWQghGyCnUOM+V/hX57o4ZJy73Lb0rVxFA9SsEWXEZ8mcIOQkV5ZAOJl9u/wXZ+AJgbqYzLwBIgqnvtcebU5aw9jswoPBF1Ftgw8Sat7H6bMRClJha7WUKlsTqtDrZP2eyJWyMYUmVdTXwIQLLeXQICusgG80P4Lu0uhITRE/QEU3T8sdFNtMyMR7jmM24eEko5JhTFyxhzaCNC1iN2ubBMJkZ6SX+hVtzzk3Wah1yi8LoaIkyT/hHZYHTidWgaZCgaCG4emTOoD8r6A+TbXxC9gFyo2wHNfZsuqIZShxixuNeNHk8em7X7S1RRq0vY/RijDYcdZWK2IadmaMeE2fXxnGPHa9kLJrO0X7vuV3b1bRmLYGlqDITInRj+iaURwIDuNgi+m4UHz+oLyKnqQmwM6/cOuBG0tpZMtik9eCbhO4I3hQtY62JS3lI4H+Xfq22TEO5L0rJUvcksY+e7+vBjqx55GqbnwfnTXgwdH2A8LpxIDqufZZFlAuxiFZQx0WGMYIBnzCCAZsCAQEwdDBdMQswCQYDVQQGEwJVUzEeMBwGA1UEChMVTWljcm9zb2Z0IENvcnBvcmF0aW9uMS4wLAYDVQQDEyVNaWNyb3NvZnQgQXp1cmUgUlNBIFRMUyBJc3N1aW5nIENBIDAzAhMzAcsTK05ojcNkJTLGAAAByxMrMA0GCSqGSIb3DQEBCwUAMA0GCSqGSIb3DQEBAQUABIIBAFpe9LfcwbXphl8/BJXOY+gdQICgPsvLtR+ctWhI/PPjeSVCON3LY/TwmG5XirRMngYk0hlFJ87mPUQShpD6mhLFnvtpx1n/MxKPUqO1tDgvNKYoUDp1I0wjfz182h2k/Vu+YCB55Vjns/0pgo2XCZfvg4fo7nLJf1zB6uaCleqNLdzQJyAMK/xWaGHfVIzRfiBj8RpcfedFB3p7jtlBBCPV5EIkCpXNgRQglWWCDfW7NYIrNuZiNwN5L0oh+fW8xlikpi+QGsLjRbo3UTItE+5unTxwzw8aD7yf+eoUk3fZOXDRq4bzaM7KkHyqx9gKqG0+hXJJq32a/vAqt6QbxKw=",
```
lotas, line 3: moved good signature and vmId to a single file, so only this needs to be updated when cert expires

petemoore (approved):
Thanks Yarik, looks good to me.

petemoore (approved):
Awesome job, thanks Yarik!

matt-boris:
Great work here, Yarik! 🎉 Thank you

PR #7710, opened 2025-05-08
Several UI fixes

### LINKED ISSUE #7474: SpeeDial (hamburger menu) sometimes closes too soon
https://github.com/user-attachments/assets/df34d8e8-5494-484a-9bfb-4b00bc8c85bf

Under certain conditions menu can close sooner than expected

### LINKED ISSUE #7686: Retriggering a `very low` priority task fails with `data/priority must be equal to one of the allowed values`
Image

The network request shows that priority is set to `very_low` in the call 

Image

Which is not a valid value according to the schema

https://github.com/taskcluster/taskcluster/blob/340ef67be880cf172a3d5f61a206eef576f18630/services/queue/schemas/v1/task.yml#L131-L139

I'm assuming that if the priority was `very-high` it would get the same issue.

lotas:
* feat(ui): Fix retrigger issue with undrescores in priority
* feat(ui): Fix hamburger menu closing too quickly

Fixes #7686 
Fixes #7474

### DIFF
--- a/changelog/issue-7474.md
+++ b/changelog/issue-7474.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 7474
+---
+
+UI: fixes hamburger menu closing too quickly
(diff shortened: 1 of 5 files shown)


matt-boris (approved):
👍🏻

PR #7711, opened 2025-05-08
feat(github): Github-issue-comment event is properly simulated

### LINKED ISSUE #7575: Simulating a `github-issue-comment` on the `Debug tc.yml` page always fails
To reproduce:

- Go here https://community-tc.services.mozilla.com/tcyaml-debug?url=https%3A%2F%2Fraw.githubusercontent.com%2Ftaskcluster%2Ftaskcluster%2Frefs%2Fheads%2Fmain%2F.taskcluster.yml
- Click on the `Simulate a custom event` select
- Select `github-issue-comment.created` or `github-issue-comment.edited`

```
GraphQL error: Schema Validation Failed! Rejecting Schema: https://community-tc.services.mozilla.com/schemas/github/v1/render-taskcluster-yml-input.json# Errors: * data/fakeEvent must NOT have additional properties: "action" * data/fakeEvent/type must be equal to one of the allowed values * data/fakeEvent/type must be equal to one of the allowed values * data/fakeEvent/type must be equal to one of the allowed values * data/fakeEvent must match exactly one schema in oneOf line: 23 column: 56079
```

The error message is always the same, whether the YAML is taskcluster's, one of mine or even an empty document so I think something's wrong with the `fakeEvent`

lotas:
TcYamlDebugger and github.renderTaskclusterYml() knows how to render github-issue-comment events

It would also inject "body.taskcluster_comment" extracted from the comment's command "/taskcluster cmd" which should be accessible in yaml as `${event.taskcluster_comment}`

This value can be overridden in UI by sending `{ comment: { body: '/taskcluster something-else' }}` in the Extra content

Fixes #7575


matt-boris (approved):
LGTM, thank you!

Issue #7628, opened 2025-04-07
feature request: g-w: create metadata file at end of task


aerickson:
On Bitbar and Lamdabtest (where we run Android integration tests) we integrate with 3rd-party CI systems and drive them to complete Taskcluster work. It would be super handy for debugging if there were an easy way of logging which task just ran on a host. 

Currently, we'd need to watch job output and find the taskID / FxCI Url. 

Ideally, we could just run 

```bash
$ cat /tmp/last_tc_job.toml | grep TC_JOB_URL
TC_JOB_URL=https://firefox-ci-tc.services.mozilla.com/tasks/Klc17PU-TMmo4axlfihKJQ/runs/0
```

petemoore:
I like this! We should do this. :-)

toml is possible, json would be another possibility. The details we can iron out.

Note, in the interim, if you don't want to scrape worker logs, another data source could be [workermanager.getWorker](https://community-tc.services.mozilla.com/docs/reference/core/worker-manager/api#getWorker) API call, whose response includes the property `recentTasks` (which is the list of the 20 most recent tasks claimed by the worker). However, a metadata file would be better to avoid the network call and additional load on the services, just highlighting this as an option for an intermediate measure if you need something now that doesn't involve log scraping.

PR #7714, opened 2025-05-08
feat(generic-worker): add metadata task feature for data on recent task run

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/issues/7628.

>Generic Worker: adds a `Metadata` feature (controlled with worker config `enableMetadata` [default: `true`], not controllable in the task payload) that writes out a file `generic-worker-metadata.json` (in the current working directory of the generic worker process) containing information about the last task run.
>
>Currently, the file will look something like this:
>
>```json
>{
>	"lastTaskUrl": "https://firefox-ci-tc.services.mozilla.com/tasks/Klc17PU-TMmo4axlfihKJQ/runs/0"
>}
>```
>
>Additional data may be added to this file in future releases.

### DIFF
--- a/changelog/issue-7628.md
+++ b/changelog/issue-7628.md
@@ -1,0 +1,15 @@
+audience: worker-deployers
+level: minor
+reference: issue 7628
+---
+Generic Worker: adds a `Metadata` feature (controlled with worker config `enableMetadata` [default: `true`], not controllable in the task payload) that writes out a file `generic-worker-metadata.json` (in the current working directory of the generic worker process) containing information about the last task run.
+
+Currently, the file will look something like this:
+
+```json
+{
+	"lastTaskUrl": "https://firefox-ci-tc.services.mozilla.com/tasks/Klc17PU-TMmo4axlfihKJQ/runs/0"
+}
+```
+
+Additional data may be added to this file in future releases.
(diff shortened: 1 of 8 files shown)


changelog/issue-7628.md
```diff
@@ -0,0 +1,15 @@
...
+
+Currently, the file will look something like this:
+
+```json
+{
+	"lastTaskUrl": "https://firefox-ci-tc.services.mozilla.com/tasks/Klc17PU-TMmo4axlfihKJQ/runs/0"
+}
+```
+
+Additional data may be added to this file in future releases.
```
petemoore, line 15: Let's explicitly state that the feature cannot be enabled/disabled in the task, as that might be confusing (it confused me at first). Alternatively the changelog doesn't need to mention that it is a task feature, it could just call it a feature, because normally task features relate to what is inside `task.payload.features` and are enabled/disabled with a boolean there, which is not the case here.

petemoore (approved):
Looks great Boris! Just one small nit...

petemoore (approved):
Thanks Boris!

changelog/issue-7628.md
matt-boris: Addressed in the first paragraph of the changelog (with a force commit):

>Generic Worker: adds a `Metadata` feature (controlled with worker config `enableMetadata` [default: `true`], not controllable in the task payload) that writes out a file `generic-worker-metadata.json` (in the current working directory of the generic worker process) containing information about the last task run.

PR #7715, opened 2025-05-08
Remove unused unzip code

petemoore:
I believe this code is unused. In #7675 @odaysec proposed a security fix for Zip-Slip, and when I went to review the fix, I noticed that we are not using this code anymore, so prefer to remove it entirely.

Let's see if the CI still passes when we move it...


matt-boris (approved):
LGTM if it passes CI 👍🏻 Thanks.

PR #7721, opened 2025-05-12
Upgrade  nock to 14 to support native fetch

lotas:
Checking if we can upgrade nock to 14 as it supports native 'fetch' mocking

### DIFF
--- a/package.json
+++ b/package.json
@@ -159,7 +159,7 @@
     "mock-fs": "^4.8.0",
     "mockdate": "^3.0.2",
     "moment": "^2.30.1",
-    "nock": "13.5.6",
+    "nock": "14.0.4",
     "open-editor": "^3.0.0",
     "qlobber": "^8.0.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -159,7 +159,7 @@
     "mock-fs": "^4.8.0",
     "mockdate": "^3.0.2",
     "moment": "^2.30.1",
-    "nock": "13.5.6",
+    "nock": "14.0.4",
```
matt-boris, line 162: If this works, maybe we shouldn't pin to this version?

```suggestion
    "nock": "^14.0.4",
```

matt-boris: Sorry, figured this was probably a WIP but put in my initial review just cuz :)

lotas: no worries.. I also found another way to do what I was planning to do, so I'll just close it for now and create an issue to do upgrade later

PR #7723, opened 2025-05-14
pass /dev/nvidia* devices to docker run

bhearsum:
This is motivated by https://github.com/mozilla/translations/issues/1117, where we're seeing GPUs disappear after a systemd daemon reload happens (https://github.com/NVIDIA/nvidia-container-toolkit/issues/48 has additional background on this).

The most straightforward workaround for this seems to be to pass the nvidia devices explicitly when running a container. Other workarounds exist (such as disabling systemd cgroup management), but I'm not exactly clear on what the implications are of doing so, eg: would it break other things this image is used for.

At the moment, this is obviously hardcoded behaviour. I can imagine that it might be preferable to remove that and eg: allow the task to provide a prefix to search for.


matt-boris (approved):
Changes LGTM! Thanks for adding the tests too.

Only thing in my mind right now is wondering if passing `--gpus all` _and_ `--device=/dev/nvidia*` devices has any conflicts/downsides?

bhearsum:
> Changes LGTM! Thanks for adding the tests too.
> 
> Only thing in my mind right now is wondering if passing `--gpus all` _and_ `--device=/dev/nvidia*` devices has any conflicts/downsides?

I can tell you it works without issue in my minimal test. I can test some other scenarios if you like.

petemoore (approved):
Looks great, many thanks!

PR #7726, opened 2025-05-15
build(deps): upgrade to node 22.15.1

matt-boris:
>Upgrades to Node.js v22.15.1 (SECURITY).
>
>More info [here](https://nodejs.org/en/blog/vulnerability/may-2025-security-releases).


petemoore (approved):
Great Scott!

Issue #7630, opened 2025-04-08
generic-worker seemingly hung during preemption handling


bhearsum:
generic-worker version: 83.4.0

I found this worker which noticed a spot termination request and seemed to begin handling it, but the last message we got from it was `Received task status change: Aborted`:
```
2025-04-07 15:30:37.704
2025-04-07T14:30:37.656982+00:00 translations-1-b-linux-large-gcp-d2g-3-bnnoazo3rgkrletzdqctew start-worker[2361]: 2025/04/07 14:30:37 GCP Metadata Service says termination is imminent
2025-04-07 15:30:37.704
2025-04-07T14:30:37.657249+00:00 translations-1-b-linux-large-gcp-d2g-3-bnnoazo3rgkrletzdqctew start-worker[2361]: 2025/04/07 14:30:37 polling for termination-time
2025-04-07 15:30:37.704
2025-04-07T14:30:37.657353+00:00 translations-1-b-linux-large-gcp-d2g-3-bnnoazo3rgkrletzdqctew start-worker[2361]: 2025/04/07 14:30:37 Got graceful-termination request with finish-tasks=false
2025-04-07 15:30:37.704
2025-04-07T14:30:37.657393+00:00 translations-1-b-linux-large-gcp-d2g-3-bnnoazo3rgkrletzdqctew start-worker[2361]: 2025/04/07 14:30:37 Killing process tree with parent PID 2536... (0xc0001200f0)
2025-04-07 15:30:37.704
2025-04-07T14:30:37.657493+00:00 translations-1-b-linux-large-gcp-d2g-3-bnnoazo3rgkrletzdqctew start-worker[2361]: 2025/04/07 14:30:37 Process tree with parent PID 2536 killed.
2025-04-07 15:30:37.704
2025-04-07T14:30:37.657541+00:00 translations-1-b-linux-large-gcp-d2g-3-bnnoazo3rgkrletzdqctew start-worker[2361]: 2025/04/07 14:30:37 Notifying listener taskcluster-proxy of state change
2025-04-07 15:30:37.704
2025-04-07T14:30:37.657581+00:00 translations-1-b-linux-large-gcp-d2g-3-bnnoazo3rgkrletzdqctew start-worker[2361]: 2025/04/07 14:30:37 Received task status change: Aborted
```

After that, we had many more messages from the system itself, eg:
```
2025-04-07 15:30:39.315
[...]

PR #7727, opened 2025-05-15
feat(generic-worker): parallelize file scanning for artifacts

matt-boris:
Will improve artifact upload performance for tasks with more than one artifact included in their payload. Could help in cases like https://github.com/taskcluster/taskcluster/issues/7630.

>Generic Worker: improves artifact upload process by parallelizing the scanning of the system for file/directory artifacts from the payload.


petemoore (approved):
Awesome job, thanks!

PR #7729, opened 2025-05-20
feat(ui): Fix signed url issue without credentials

### LINKED ISSUE #4943: 'buildSignedUrl missing required credentials' shown after user signed in
**Describe the bug**
When I sign into taskcluster e.g. at https://firefox-ci-tc.services.mozilla.com/tasks/GYRCoAqMRe2su7shpTBJjQ , a red error message 'buildSignedUrl missing required credentials' will be shown and the page has to be reloaded to see information about the task. Might be related to the task/repository currently shown.

lotas:
This happened because user session is being initialized twice. First time it is initialized right after successful callback from external provider, and at this point credentials are not being issued. Second time it is initialized when AuthController detects that credentials are missing, so it fetches them and updates user session with valid credentials.

So between those two events error happened and tried to sign a url without valid credentials.

Fixes #4943

### DIFF
--- a/changelog/issue-4943.md
+++ b/changelog/issue-4943.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 4943
+---
+
+Fixes UI error "buildSignedUrl missing required credentials" shown directly after login due to missing credentials.
--- a/ui/src/utils/getArtifactUrl.js
+++ b/ui/src/utils/getArtifactUrl.js
@@ -11,7 +11,7 @@ const { Queue, Index } = require('taskcluster-client-web');
 export function getArtifactUrl({ user, taskId, runId, name }) {
   const queue = getClient({ Class: Queue, user });
 
-  if (user) {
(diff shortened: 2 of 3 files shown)


changelog/issue-4943.md
```diff
@@ -0,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 4943
+---
+
+Fixes UI error "buildSignedUrl missing required credentials" shown directly after login due to missing credentials.
```
petemoore, line 6: Do you know when this bug was introduced? Could we add a note to the change log to explicitly say which release introduced this bug, so users know if they have it and need to update or not?

petemoore (approved):
LGTM!

changelog/issue-4943.md
lotas: I think it was there for 5+ years. Only happens when user session changes. Hard do tell if that feature caused it or some other auth related one did

PR #7734, opened 2025-05-27
chore(deps): use `go tool` directive

matt-boris:
New in go1.24 https://tip.golang.org/doc/go1.24#go-command


matt-boris:
Successful staging release https://community-tc.services.mozilla.com/tasks/arUlVAn9RoGMIWmzx_xdjQ/runs/0

petemoore (approved):
Looks great! Can't believe we were still installing gox!

PR #7739, opened 2025-05-30
fix: modernize logger interface

emmanuel-ferdman:
## PR Summary
The `logger.warn()` method is deprecated since Python2.7 and replaced with `logger.warning()`. It leads to those warnings:
```python
DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
```


changelog/pr-7739.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
+level: patch
+reference: pr 7739
```
lotas:
```suggestion
```

this is only needed if there's a ticket created. So we can either drop it, or add an issue and reference it here :)

PR #7761, opened 2025-06-04
fix(d2g): ensure GPUs before adding GPU flags

matt-boris:
It's _possible_ that this may help https://github.com/mozilla/translations/issues/1117.

>D2G: Runs `nvidia-smi` to ensure that the `/dev/nvidia-uvm` and `/dev/nvidia-uvm-tools` devices are loaded and passed as `--device` flags.

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -6,6 +6,7 @@ package d2g
 import (
 	"encoding/json"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -16,6 +17,7 @@ import (
 	"github.com/taskcluster/taskcluster/v84/internal/scopes"
 	"github.com/taskcluster/taskcluster/v84/tools/d2g/dockerworker"
 	"github.com/taskcluster/taskcluster/v84/tools/d2g/genericworker"
+	"github.com/taskcluster/taskcluster/v84/workers/generic-worker/host"
 
 	"slices"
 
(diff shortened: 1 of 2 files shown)


lotas (dismissed):

petemoore (commented):
If I am reading this right, with this patch, if for some reason `nvidia-smi --query-gpu=count --format=csv,noheader` does not return a count of cpus >= 1, we won't add all the gpu flags to the `docker run` command, even if `allowGPUs` is `true`. I don't see how this solves the issue, because best case, we stop passing gpu flags through to the `docker run` command, which means the docker container will not have access to the GPUs it was expecting to have access to. Or is this patch intended to solve an issue for tasks on workers that intentionally have no GPUs, but the worker pool has been configured to allow d2g tasks to use them?

If the issue is due to a race condition of the GPUs not being available at the time the docker container starts, I think the code would need to wait for them to be available, or timeout if they never become available?

tools/d2g/d2g.go
```diff
@@ -704,3 +705,19 @@ func (ril *RegistryImageLoader) ChainOfTrustCommand() string {
 		ril.Image.String(),
 	)
 }
+
+func hasGPUs() bool {
+	if testing.Testing() {
+		return true
+	}
+
+	out, err := host.Output("nvidia-smi", "--query-gpu=count", "--format=csv,noheader")
```
petemoore: If running this command not only checks GPUs are present, but actually causes them to appear under `/dev` as a desired side-effect, we should definitely add a code comment here to explain that, or even change function name to something like `discoverNvidiaGPUs` or `enableNvidiaGPUs`. If running the command doesn't have side-effects like this, then maybe rename it to `hasNvidiaGPUs` since I think this is nvidia only...

Issue #7498, opened 2025-02-04
Excessive resource usage while watching a big task group


Eijebong:
I left [this task group](https://firefox-ci-tc.services.mozilla.com/tasks/groups/f7kRd3ZeSz-OLC8_omvUvQ) in the background while doing something else and when coming back to it, found out that the page got extremely slow...

Turns out firefox got hungry for RAM and CPU.
![Image](https://github.com/user-attachments/assets/30212b17-842c-4797-984e-e91ca9a4b2a0)

`about:memory`

![Image](https://github.com/user-attachments/assets/51e82d2a-fcf6-469e-8bd7-9e525aa7fd4e)

Strings because I found this a bit suspicious:

![Image](https://github.com/user-attachments/assets/66971355-f3d8-4127-9c4c-6b8312e5e4eb)

lotas:
those 10mb strings on the screenshot looks scary.. maybe it's some json-serialized objects used as caching keys  🤔

lotas:
I think those must be coming from the `memoize()` - https://github.com/taskcluster/taskcluster/blob/95a302b70b83afd194154b98b6362721700ffcf3/ui/src/components/TaskGroupTable/index.jsx#L54-L81   or `sorted()` https://github.com/taskcluster/taskcluster/blob/95a302b70b83afd194154b98b6362721700ffcf3/ui/src/components/TaskGroupTable/index.jsx#L29-L40 

which is coming from this library: https://github.com/caiogondim/fast-memoize.js 

since statuses change frequently and the task group is quite large, i think it would try to memoize every possible state along the way 😬  which is likely causing _some of the memory leaks_ here 
maybe there's more

lotas:
i guess we could disable caching or substitute full string with some crc32/md5 hash instead: https://github.com/taskcluster/taskcluster/blob/95a302b70b83afd194154b98b6362721700ffcf3/ui/src/components/TaskGroupTable/index.jsx#L75-L80

lotas:
I'm still debugging this, could reproduce it locally. 
the main issue is not rendering 1000 tasks, but handling updates. And it's not yet clear  to me what part exactly causes memory drain, but it's highly likely in react or the ui libraries we use. 

I can reproduce it locally with small amount of tasks and then frequent updates to the task group. Causes memory spike (probably too much re-rendering and keeping old copies in shadow dom), but after 5-10s it goes back to normal (GC?) 

[...]

PR #7767, opened 2025-06-10
Optimize memory usage in UI

lotas:
For some table data serialized keys might include hundreds if not
thousands of keys concatenated, which sometimes produced 10mb+ strings
in memory. Also default cache implementation of fast-memoize library
didn't limit the number of items stored in it, so it allowed the cache
to grow unlimited.

Fixes: #7498 (hopefully)


ui/src/components/TaskGroupProgress/index.jsx
```diff
@@ -208,6 +153,62 @@ export default class TaskGroupProgress extends Component {
     previousTaskGroupId: this.props.taskGroupId,
   };
 
+  getStatusCount = memoize(
```
lotas, line 156: just moving it inside the component, so when component is gone, the memoized function with the attached cache storage is also gone

matt-boris (approved):
LGTM, was there no configuration options on the `fast-memoize` library?

matt-boris:
Sorry, I read over the fact that you're still using `fast-memoize`, looked like you just implemented your own cache 😄 Definitely LGTM, feel free to merge.

PR #7775, opened 2025-06-10
feat(generic-worker): fine-tune auto-abortion of tasks

### LINKED ISSUE #7769: generic-worker: print resource usage summary on task abort
When a task finishes normally, g-w prints some stats on resource usage.  It would be nice if that was also available when the task is aborted (e.g. due to oom protection or going over maxRuntime).

### LINKED ISSUE #7770: generic-worker: allow tweaking the thresholds for resource monitor
Right now we abort the task if memory usage is >90% for a couple of minutes.  For workers with lots of memory 10% can still be plenty, so it'd be nice to be able to tweak that in terms of 1) how long we can be over, 2) what the threshold is both relative to total memory and in absolute terms, e.g. to not abort if there's still >500M free.

matt-boris:
Fixes #7769.

>Generic Worker: resource monitor will print out its usage summary after aborting the task.

Fixes #7770.

>Generic Worker: adds additional resource monitoring auto-abortion configuration to better fine-tune how your worker aborts running task processes.
>
>  * `absoluteHighMemoryThreshold`: The minimum amount of available memory (in bytes) required before considering task abortion. If available memory drops below this value, it may trigger an abort. Default: `524288000` (500MiB).
>  * `relativeHighMemoryThreshold`: The percentage of total system memory usage that, if exceeded, contributes to the decision to abort the task. Default: `90`.
>  * `allowedHighMemoryDurationSecs`: The maximum duration (in seconds) that high memory usage conditions can persist before the task is aborted. Default: `5`.
>
>Generic Worker will auto-abort a task if the total system memory used percentage is greater than `relativeHighMemoryThreshold` _AND_ the available memory is less than `absoluteHighMemoryThreshold` for `numberOfConsecutiveMeasurements` measurements at 0.5s intervals, unless `disableOOMProtection` is enabled.

### DIFF
--- a/workers/generic-worker/usage.go
+++ b/workers/generic-worker/usage.go
@@ -131,6 +131,25 @@ and reports back results to the queue.
         ** OPTIONAL ** properties
         =========================
 
+          absoluteHighMemoryThreshold       Number of bytes the resource monitor uses to
+                                            determine when to abort a task due to high
+                                            memory usage. The is an absolute number of bytes
+                                            needed of available memory before aborting the task.
+                                            For example, if the value is 524288000, then the worker will
+                                            abort a task if the memory available is 200MiB or less
+                                            for longer than allowedHighMemoryDurationSecs seconds.
+                                            Can be used in conjunction with relativeHighMemoryThreshold.
+                                            Does nothing if disableOOMProtection is set to true.
+                                            [default: 524288000] (500MiB)
+          allowedHighMemoryDurationSecs     The number of seconds the resource monitor will
+                                            allow the system memory usage to be above the high
+                                            memory thresholds (see absoluteHighMemoryThreshold
+                                            and relativeHighMemoryThreshold) before aborting
+                                            the task. If the memory usage is above the high
+                                            memory thresholds for longer than this time, the
+                                            worker will abort the task. Does nothing if
+                                            disableOOMProtection is set to true.
+                                            [default: 5]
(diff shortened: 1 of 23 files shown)


workers/generic-worker/usage.go
```diff
@@ -222,6 +234,11 @@ and reports back results to the queue.
                                             [default: 0]` + loopbackDeviceNumbers() + `
           maxTaskRunTime                    The maximum value allowed for maxRunTime on generic-worker payloads.
                                             [default: 86400]
+          numberOfConsecutiveMeasurements   The number of consecutive resource usage measurements
```
lotas: as a user I would probably understand "seconds" better than "measurements" and it wouldn't need an extra explanation how often those measurements run. `measurementDurationSeconds`? :)

petemoore (dismissed):
Looks good! I wonder at what point, the algorithm will also be configurable! Nice job. :-)

matt-boris:
@jcristau would do you mind checking over the new options and let me know what you think here? 🙏🏻  I'm much more confident in the OOM killer now with these changes. I'd be interested to see how these default settings work on that one test you ran a couple days ago 10 times over.

jcristau:
Seems like a good improvement, thanks @matt-boris.  (I suspect 500MB is still more than necessary, but we'll be able to tweak and experiment, so it's a reasonable first step.)

PR #7782, opened 2025-06-13
feat(ui): TaskGroup page memory and rendering optimizations

### LINKED ISSUE #7498: Excessive resource usage while watching a big task group

lotas:
This is not a complete fix, but a big step forward. Page stays responsive under frequent updates, and memory grows considerably slower compared to the existing state. 

* Limit default maxsize for memoized functions
* Uses React.memo for row rendering to avoid creating new objects when data is not be changed. This seems to improve memory consumption
* Calculate TaskGroup counts in the main component (This will avoid passing huge lists of tasks between components)
* TaskGroupTable uses batched updates for the subscriptions to avoid frequent updates

Fixes #7498 (again)

### DIFF
--- a/changelog/issue-7498.md
+++ b/changelog/issue-7498.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 7498
+---
+
+Further UI memory optimizations for the TaskGroup page to avoid using too much memory during intensive updates.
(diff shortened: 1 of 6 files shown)


matt-boris (approved):
LGTM, thanks Yarik 👍🏻

PR #7787, opened 2025-06-17
Fix lint issue

petemoore:
Hit [this linting issue](https://github.com/taskcluster/taskcluster/runs/44240899798) while working on something unrelated.


matt-boris (commented):
Cool! TIL about `strings.CutPrefix`. Added in go1.20 https://pkg.go.dev/strings#CutPrefix

PR #7788, opened 2025-06-17
feat(github): Github service reacts to empty tasks

### LINKED ISSUE #7786: The github service should comment or react on a github comment if it doesn't generate any task
Right now if you comment on a PR with `/taskcluster foobar` and that ends up not generating any task, it looks like the github service missed the webhook as there's no 👍 reaction on the comment and you have to look at the logs of the service to figure out that in fact, `intree config for mozilla-releng/fxci-config@98061c3d61bff9b2659e69ed0ecdaf6a715ae4a8 compiled with zero tasks. Skipping`.

This happened here for example https://github.com/mozilla-releng/fxci-config/pull/424#issuecomment-2971485446

It would be nice if the github service could at least react to the comment and maybe even add a comment along the line of "This comment didn't generate any task, make sure you wrote the correct thing".

lotas:
If `/taskcluster xx` command didn't produce any tasks - comment would receive a reaction

Fixes #7786

### DIFF
--- a/services/github/src/handlers/job.js
+++ b/services/github/src/handlers/job.js
@@ -239,6 +239,17 @@ export async function jobHandler(message) {
     }
     if (!graphConfig.tasks || graphConfig.tasks.length === 0) {
       debug(`intree config for ${organization}/${repository}@${sha} compiled with zero tasks. Skipping.`);
+
+      // If triggered by a comment, let everyone know we couldn't create tasks
+      if (message.payload.details['event.type'].startsWith('issue_comment')) {
+        await this.createComment({
+          instGithub, organization, repository, pullNumber, sha, debug,
+          body: {
+            summary: 'No Taskcluster jobs started for this command',
+            details: 'Task graph produced empty list of tasks',
+          },
+        });
+      }
       return;
     }
   } catch (e) {
--- a/changelog/issue-7786.md
+++ b/changelog/issue-7786.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 7786
+---
+
+Github service comments back when taskcluster command in issue comment didn't produce any tasks
--- a/services/github/test/handler_test.js
+++ b/services/github/test/handler_test.js
@@ -981,6 +981,23 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
         assert.equal(build.state, 'pending');
       });
 
+      test('valid issue_comment (user is collaborator) no tasks created', async function () {
+        instGithub.setTaskclusterYml({
+          owner: 'taskcluster',
+          repo: 'tc-dev-integration-test',
+          ref: COMMIT_SHA,
+          content: {
+            ...validYamlCommentsJson,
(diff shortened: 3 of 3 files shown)


services/github/src/handlers/job.js
```diff
@@ -239,6 +239,17 @@ export async function jobHandler(message) {
...
       debug(`intree config for ${organization}/${repository}@${sha} compiled with zero tasks. Skipping.`);
+
+      // If triggered by a comment, add a reaction to indicate no tasks were generated
+      if (message.payload.details['event.type'].startsWith('issue_comment')) {
+        await this.addCommentReaction({
+          instGithub,
+          organization,
+          repository,
+          commentId: message.payload.body.comment.id,
+          reaction: '-1',
```
lotas: i wonder if 👎  is too passive aggressive here lol 
@Eijebong wdyt? 

we are only limited to those few emojis https://docs.github.com/en/rest/reactions/reactions?apiVersion=2022-11-28

matt-boris: Could we just add a comment instead? So it's clear.

Eijebong: Yeah, that's not ideal. Maybe 👀 would be good? "I saw the comment and nothing else". The other use of that reaction is when someone doesn't have rights, but then it also puts a comment.

lotas: yes, this was an initial idea: https://github.com/taskcluster/taskcluster/issues/7786#issuecomment-2979977388 

maybe adding a comment is better to inform that nothing happened (instead of showing fingers)

matt-boris: Yeah I like always reacting the same with 👀 and then commenting based on what's going on. 👍🏻

matt-boris: I guess _just_ 👀 would be fine too. It just shows you that tc's acknowledged the command at least

Eijebong: At least if nothing happens it tells you "this isn't github's fault for once"

lotas: if we put 👀  only it might not tell much if handler dies or jobs fail to be created.. having something distinctive like thumbs down or a comment would work better imho. We can do both :)

lotas: alright, let's just leave a comment, less confusion and more clear what's going on :)

services/github/src/handlers/job.js
```diff
@@ -239,6 +239,17 @@ export async function jobHandler(message) {
     }
     if (!graphConfig.tasks || graphConfig.tasks.length === 0) {
       debug(`intree config for ${organization}/${repository}@${sha} compiled with zero tasks. Skipping.`);
+
+      // If triggered by a comment, add a reaction to indicate no tasks were generated
```
Eijebong: That comment is wrong now

lotas: lol .. too many things are happening at once 🙈  thanks

matt-boris (approved):

PR #7791, opened 2025-06-18
chore: hide confusing `corepack yarn` logs

matt-boris:

PR #7793, opened 2025-06-18
chore: fix `client-web` kind.yml

matt-boris:
@Eijebong noticed here https://github.com/taskcluster/taskcluster/pull/7791#discussion_r2154659772.


lotas (approved):
🙈

PR #7807, opened 2025-06-30
UI: Fix state for task view page

### LINKED ISSUE #6951: action tasks fire against the correct task group in some cases
Over in https://github.com/mozilla/firefox-translations-training/issues/250 we're working on ensuring that action tasks work when fired from a task group created by another action task. (In this case, we mostly care about "Cancel All", but the same issue occurs for other actions.)

When we follow these steps:
- Go to decision task group
- Initiate a Train action (this creates a new task group with a bunch of tasks)
- Navigate to one of the Tasks created from it
- Go to _that_ task's group
- Use "Cancel All" from there

...the "Cancel All" action ends up firing against the original decision task group instead of the new task group. If you refresh the page before firing the action, it runs correctly against the new task group.

This suggests to me that there's some data not being propgated or updated correctly, most likely when navigating from the Train action to one of the tasks it creates, or from there to the task group view.

I spent some time in the React Dev Tools and couldn't find an obvious cause: all of the `props` in the TaskGroup view seemed to be correct, although I am far from an expert here.

I wasn't able to dig into the graphql queries at all - I'm suspicious that we may not be firing the same queries originally as we do when refreshing the task group view though.

lotas:
* Changes state management for Task View page to avoid using "stale" actions when task changes
* Update Speed Dial (menu) to close when navigating away 

Fixes #6951

### DIFF
--- a/ui/src/views/Tasks/ViewTask/index.jsx
+++ b/ui/src/views/Tasks/ViewTask/index.jsx
@@ -131,43 +131,14 @@ export default class ViewTask extends Component {
     const {
       data: { task },
     } = props;
-    const taskActions = [];
-    const actionInputs = state.actionInputs || {};
-    const actionData = state.actionData || {};
 
     if (taskId !== state.previousTaskId && task) {
-      const { taskActions: actions } = task;
-
       updateTaskIdHistory(taskId);
 
-      actions &&
-        actions.actions.forEach(action => {
-          const schema = action.schema || {};
-
-          // if an action with this name has already been selected,
-          // don't consider this version
-          if (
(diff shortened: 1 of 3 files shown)


ui/src/views/Tasks/ViewTask/index.jsx
```diff
@@ -131,43 +131,14 @@ export default class ViewTask extends Component {
     const {
       data: { task },
     } = props;
-    const taskActions = [];
```
lotas, line 134: there's no need to keep everything in state, since it can be derived from props

PR #7826, opened 2025-07-02
fix(worker-manager): use azure next gen methods

matt-boris:
Follow-up to https://github.com/taskcluster/taskcluster/pull/7799.


matt-boris:
Got lots of information from https://github.com/Azure/azure-sdk-for-js/blob/main/documentation/MIGRATION-guide-for-next-generation-management-libraries.md.

PR #7830, opened 2025-07-08
chore: remove `lower-case` and `upper-case` deps

matt-boris:
Will be able to close https://github.com/taskcluster/taskcluster/pull/7189 after this lands.


lotas (approved):
😇

PR #7841, opened 2025-07-09
feat(generic-worker): add `status` subcommand

### LINKED ISSUE #7840: Create `status` generic worker subcommand
> After a few discussions, the requested feature has changed. The new request is for a generic-worker sub-command that will return:
> 
> - Status: if a task is currently running.
> - And if running the current task ID
> 
> The format of the response needs to be parsable, so something like easily identified key-value pairs.  

 _Originally posted by @markcor in [#7737](https://github.com/taskcluster/taskcluster/issues/7737#issuecomment-3053327342)_

matt-boris:
Fixes #7840.

>Generic Worker: adds `status` subcommand to output task ID if a task is currently being executed.
>
>Example output while a task is running:
>
>```bash
>$ generic-worker status
>{
>  "currentTaskIds": [
>    "fz7IO4uCTtevuLBoq8Qz3w"
>  ]
>}
>```
>
>Example output while worker is idle:
>
>```bash
>$ generic-worker status
>{
>  "currentTaskIds": []
>}
>```

### DIFF
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -1096,6 +1096,8 @@ func exitOnError(exitCode ExitCode, err error, logMessage string, args ...any) {
 	log.Printf("Root cause: %v", err)
 	log.Printf("%#v (%T)", err, err)
 	combinedErr := fmt.Errorf("%s, args: %v, root cause: %v, exit code: %d", logMessage, args, err, exitCode)
-	errorreport.Send(WorkerRunnerProtocol, combinedErr, debugInfo)
+	if WorkerRunnerProtocol != nil {
+		errorreport.Send(WorkerRunnerProtocol, combinedErr, debugInfo)
+	}
 	os.Exit(int(exitCode))
 }


workers/generic-worker/main.go
```diff
@@ -66,8 +66,9 @@ var (
 	logPath   = filepath.Join("generic-worker", "live_backing.log")
 	debugInfo map[string]string
 
-	version  = internal.Version
-	revision = "" // this is set during build with `-ldflags "-X main.revision=$(git rev-parse HEAD)"`
+	version          = internal.Version
+	revision         = "" // this is set during build with `-ldflags "-X main.revision=$(git rev-parse HEAD)"`
+	workerStatusPath = filepath.Join(os.TempDir(), "worker-status.json")
```
lotas, line 71: I wonder if `os.TempDir()` is always the same for all users? Is it possible that file would be written by a different user? maybe they would end up in different temp dirs?

matt-boris: Checking out https://pkg.go.dev/os#TempDir, you can see that on unix, it returns `$TMPDIR` if non-empty, else `/tmp`, windows uses `GetTempPath`.

This `os.TempDir()` call happens outside the context of a task, so it'd be checking for those values on the host environment, so that should stay consistent, no matter the user/task being run.

matt-boris: I suppose if a _task_ defines a temp dir env var that's not the typical path and `generic-worker status` is run within the context of the task payload, this may cause issues. But I think that use case doesn't make too much sense, tbh since running `generic-worker status` will happen outside of a task.

petemoore:
Given that we agreed to implement this (apologies, I was on PTO during the meeting, and hadn't seen that we had agreed) I do not want to block, we can go with it as is. I think the worker metrics that we have might offer a different path with a bit more flexibility, and we could potentially write those events to a separate event log file if desired. But if we've promised this, it is working, and unblocks Release Operations, we can go with it.

matt-boris:
Couldn't reopen, so created https://github.com/taskcluster/taskcluster/pull/7850.

PR #7850, opened 2025-07-14
feat(generic-worker): add `status` subcommand

### LINKED ISSUE #7840: Create `status` generic worker subcommand

matt-boris:
Fixes #7840.

>Generic Worker: adds `status` subcommand to output task ID if a task is currently being executed.
>
>Example output while a task is running:
>
>```bash
>$ generic-worker status
>{
>  "currentTaskIds": [
>    "fz7IO4uCTtevuLBoq8Qz3w"
>  ]
>}
>```
>
>Example output while worker is idle:
>
>```bash
>$ generic-worker status
>{
>  "currentTaskIds": []
>}
>```

### DIFF
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -66,8 +66,9 @@ var (
 	logPath   = filepath.Join("generic-worker", "live_backing.log")
 	debugInfo map[string]string
 
-	version  = internal.Version
-	revision = "" // this is set during build with `-ldflags "-X main.revision=$(git rev-parse HEAD)"`
+	version          = internal.Version
+	revision         = "" // this is set during build with `-ldflags "-X main.revision=$(git rev-parse HEAD)"`
+	workerStatusPath = filepath.Join(os.TempDir(), "worker-status.json")
 )
 
 func initialiseFeatures() (err error) {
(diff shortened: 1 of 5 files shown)


workers/generic-worker/main.go
```diff
@@ -892,6 +903,12 @@ func (task *TaskRun) Run() (err *ExecutionErrors) {
 
 	err = &ExecutionErrors{}
 
+	workerStatus := &WorkerStatus{
+		CurrentTaskIDs: []string{task.TaskID},
+	}
+	err.add(executionError(internalError, errored, fileutil.WriteToFileAsJSON(workerStatus, workerStatusPath)))
+	defer os.Remove(workerStatusPath)
```
lotas, line 910: does this Run method lives as long as the task is running?

lotas (approved):
lgtm

workers/generic-worker/main.go
matt-boris: Good question. Yes, you can see in the `RunWorker()` function how `task.Run()` is called sequentially and has the `taskStart` and `taskFinish` events before/after the call.

https://github.com/taskcluster/taskcluster/blob/main/workers/generic-worker/main.go#L471-L476

PR #7851, opened 2025-07-14
feat(d2g): cache loaded docker images

### LINKED ISSUE #7849: Artifacts images are not cached when using d2g
docker worker used to check if the image was already there and loaded based on the artifact hash. https://github.com/taskcluster/taskcluster/blob/d30f3d3620c6f4c2c47e1930d21886c9bf68f59f/workers/docker-worker/src/docker/artifact_image.js#L234-L243

d2g doesn't do this which means that for every task using an artifact image, it loads the image again. To make things worse, it doesn't load the image from artifact storage but instead copies it in the work directory of the user **then** loads it, which basically doubles the amount of disk IO.

matt-boris:
Fixes #7849.

>Generic Worker (D2G): caches loaded docker images.

### DIFF
--- a/workers/generic-worker/d2g_feature.go
+++ b/workers/generic-worker/d2g_feature.go
@@ -4,11 +4,17 @@ package main
 
 import (
 	"fmt"
+	"log"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/taskcluster/taskcluster/v87/internal/scopes"
+	"github.com/taskcluster/taskcluster/v87/workers/generic-worker/fileutil"
 	"github.com/taskcluster/taskcluster/v87/workers/generic-worker/process"
 )
+
+var imageCache ImageCache
 
 type (
 	D2GFeature struct {
@@ -17,6 +23,8 @@ type (
 	D2GTaskFeature struct {
 		task *TaskRun
 	}
+
+	ImageCache map[string]string
 )
 
 func (df *D2GFeature) Name() string {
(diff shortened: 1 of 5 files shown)


workers/generic-worker/d2g_feature.go
```diff
@@ -51,28 +62,58 @@ func (dtf *D2GTaskFeature) RequiredScopes() scopes.Required {
...
+			panic(err)
+		}
+
+		key = fmt.Sprintf("%x", hash.Sum(nil))
 	}
 
-	imageID := strings.TrimSpace(string(out))
+	imageID := imageCache[key]
+
+	if imageID == "" {
```
Eijebong, line 79: I have not tested this yet, but I feel like this is missing a check that the image actually exists in docker? What if it gets loaded and written to the cache and then gets garbage collected by https://github.com/taskcluster/taskcluster/blob/7929345de55279fad979f0885fb43c62fed77668/workers/generic-worker/garbagecollector.go#L73

matt-boris: Great call, thank you! Forgot about that in the garbage collector.

matt-boris: I wonder if it would be sufficient to just remove the cached images state file that is managed in `d2g_feature.go` on garbage collection. All d2g tasks run through `docker run`, it's not like they could inject some `docker image prune` command. See any downsides?

Eijebong: Given we prune all images on garbage collection I cannot think of a downside to pruning the cache state file with it. But I might be missing something

matt-boris: Handled in the `runGarbageCollection()` function here https://github.com/taskcluster/taskcluster/pull/7851/files?w=1#diff-99318b70558b8ee5be5ba23ce1b3bd90f2eb30b68f810e6c474e8f51489ffeafR78-R81

PR #7857, opened 2025-07-16
feat(queue): Fixes number of claimed tasks for the task queue

### LINKED ISSUE #7856: Queue reports incorrect number of claimed tasks
`queue.taskQueueCounts()` returns incorrect number of claimed tasks when tasks are being reclaimed. 

This is because how internal data is being handled in `queue_claimed_tasks` - when task is being claimed or reclaimed, new entry is being added to the table.

`queue_claimed_tasks_count` db function is not filtering for distinct `(task_id, run_id)`

Two ways of fixing this is:
1. add distinct `task_id, run_id` to the query
2. when task is being claimed/reclaimed - replace existing record

Second approach is more complicated because it might require schema changes and tweaking QueueService internals. I don't remember if there was a reason for not doing so initially

lotas:
Fixes #7856 

also improves query used in metrics collector to use postgres tuples instead of string concats


petemoore (dismissed):
Looks good to me, thanks yarik!

db/versions/0113.yml
```diff
@@ -0,0 +1,80 @@
...
+      begin
+        return (
+          select count(distinct (task_id, run_id))
+          from queue_claimed_tasks
+          where task_queue_id = task_queue_id_in
+            and taken_until > now()
+        );
+      end
+
+  queue_worker_stats:
```
lotas, line 23: this is the same query as in v112, but just uses `DISTINCT (task,run)` instead of string concat

matt-boris (approved):
👍🏻

PR #7868, opened 2025-07-22
feat(ui): Fix incorrect form handling for custom task actions

lotas:
Fixes few issues with form inputs handling that were introduced in #7807

### DIFF
--- a/ui/src/views/Tasks/ViewTask/index.jsx
+++ b/ui/src/views/Tasks/ViewTask/index.jsx
@@ -191,6 +191,7 @@ export default class ViewTask extends Component {
     dialogError: null,
     caches: null,
     selectedCaches: null,
+    formInputs: null,
   };
 
   listener = null;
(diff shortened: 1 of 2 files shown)


ui/src/views/Tasks/ViewTask/index.jsx
```diff
@@ -252,12 +253,14 @@ export default class ViewTask extends Component {
   }
 
   handleActionClick = name => () => {
-    const { action } = this.state.actionData[name];
+    const { actionData, actionInputs } = this.getTaskActionsData();
```
lotas, line 256: this was moved out of state, so `this.state.actionData` was a leftover from the last refactoring

ui/src/views/Tasks/ViewTask/index.jsx
```diff
@@ -252,12 +253,14 @@ export default class ViewTask extends Component {
...
+    const { action } = actionData[name];
 
     this.setState({
       dialogError: null,
       dialogOpen: true,
       selectedAction: action,
+      formInputs: actionInputs[name] ?? '',
```
lotas, line 263: `actionInputs[name]` is just a string with pre-filled action payload

matt-boris (dismissed):

matt-boris (approved):

lotas:
i tested it twice locally 🤞 
those actions are really hard to test and reproduce, since they are only used by some moz tasks

PR #7887, opened 2025-07-23
fix(worker-manager): terminate aws workers that are in stopped state

### LINKED ISSUE #7680: Worker Manager (AWS): Provider does not differentiate between a `stopped` and `terminated` VM
If an AWS instance stops for whatever reason, the AWS provider will not fully delete/terminate it.

GCP, on the other hand, _always_ calls the delete method in this case https://github.com/taskcluster/taskcluster/blob/0150bff352a0e2f75a28e491e9b22219fa516b2e/services/worker-manager/src/providers/google.js#L407

matt-boris:
Fixes #7680.

>Worker Manager (AWS): terminates workers that are in the stopped state.

### DIFF
--- a/services/worker-manager/src/providers/aws.js
+++ b/services/worker-manager/src/providers/aws.js
@@ -337,6 +337,9 @@ export class AwsProvider extends Provider {
 
           case 'terminated':
           case 'stopped':
+            await this._enqueue(`${region}.modify`, () => this.ec2s[region].send(new TerminateInstancesCommand({
+              InstanceIds: [worker.workerId.toString()],
+            })));
             await this.onWorkerStopped({ worker });
             state = Worker.states.STOPPED;
             break;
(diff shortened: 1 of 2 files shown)


services/worker-manager/src/providers/aws.js
```diff
@@ -388,7 +391,7 @@ export class AwsProvider extends Provider {
     try {
       const region = worker.providerData.region;
       result = await this._enqueue(`${region}.modify`, () => this.ec2s[region].send(new TerminateInstancesCommand({
-        InstanceIds: [worker.workerId],
+        InstanceIds: [worker.workerId.toString()],
```
matt-boris, line 394: For consistency across the file, similar to what's used in the `checkWorker` method https://github.com/taskcluster/taskcluster/blob/4674a75b3e4653890f6f3303245dcf30acf0a4e0/services/worker-manager/src/providers/aws.js#L325

services/worker-manager/src/providers/aws.js
```diff
@@ -337,6 +337,9 @@ export class AwsProvider extends Provider {
 
           case 'terminated':
           case 'stopped':
+            await this._enqueue(`${region}.modify`, () => this.ec2s[region].send(new TerminateInstancesCommand({
```
matt-boris, line 340: Important to note from https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/ec2/command/TerminateInstancesCommand/:

>Shuts down the specified instances. This operation is idempotent; if you terminate an instance more than once, each call succeeds.

lotas (approved):

PR #7896, opened 2025-08-01
Remove neutrino, use webpack/jest directly

### LINKED ISSUE #5265: Migrate off of Neutrino
[Neutrino](https://github.com/neutrinojs/neutrino) is a blocker to us for easily upgrading vulnerable dependencies, as well as upgrading webpack, loaders, etc. which Neutrino is not compatible with. We will lose out on security, performance enhancement, and new feature updates to these dependencies as long as we rely on this stale project.

Neutrino has last been released in December of 2020, so it's safe to say it's hardly managed at this point.

We should find a new, up-to-date solution or just do the things Neutrino was doing for us ourselves.

Eijebong:
This is mostly a direct port of the configuration that was generated by neutrino so it might not be the most optimal but at least now we have control over it.

Github Bug/Issue: Fixes #5265


petemoore:
This looks fine to me, but I think I'll wait for Yarik to review it. Thanks Bastien!

PR #7907, opened 2025-08-11
fix(worker-manager): use standard sku public IPs for azure

### LINKED ISSUE #7903: Add Support for Public IP Address SKU in Azure Worker Manager
While worker manager supports `publicIp` for each worker through worker configuration, it seems to default to a `basic` public ip sku. While this is sufficient when provisioning workers with a Public IP Address (PIP), after September 30, 2025, Microsoft will no longer support the [basic sku](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/public-ip-addresses).

This issue is to request assigning a `standard` pip when worker manager provisions a worker in Azure.

matt-boris:
Fixes #7903.

>Worker Manager (Azure): uses `Standard` SKU public IP addresses, as `Basic` will be retired on September 30, 2025. [Official announcement](https://azure.microsoft.com/en-us/updates?id=upgrade-to-standard-sku-public-ip-addresses-in-azure-by-30-september-2025-basic-sku-will-be-retired)


petemoore (approved):
Looks perfect, thank you!

PR #7924, opened 2025-08-19
fix(worker-runner): protocol pipe now works on windows

matt-boris:
>Worker Runner: adds additional logging around sending `graceful-termination` request to worker.
>
>Worker Runner (windows): fixes protocol pipe connection so that Generic Worker can communicate with Worker Runner. This allows `graceful-termination` requests to be properly sent and received, among other message types. You must include `--with-worker-runner` in your Generic Worker service configuration on the `run` subcommand.

### DIFF
--- a/tools/worker-runner/provider/azure/metadata_test.go
+++ b/tools/worker-runner/provider/azure/metadata_test.go
@@ -53,7 +53,7 @@ func testServer() *httptest.Server {
 		}
 
 		if r.URL.Path == "/metadata/instance" {
-			if apiVersion != "2019-08-15" {
+			if apiVersion != "2025-04-07" {
 				w.WriteHeader(400)
 				fmt.Fprintf(w, "Bad API version")
 				return
@@ -80,7 +80,7 @@ func testServer() *httptest.Server {
 			return
 		}
 		if r.URL.Path == "/metadata/scheduledevents" {
-			if apiVersion != "2017-11-01" { // note: different from other endpoints!
+			if apiVersion != "2020-07-01" { // note: different from other endpoints!
 				w.WriteHeader(400)
 				fmt.Fprintf(w, "Bad API version")
 				return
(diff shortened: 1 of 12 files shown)


tools/worker-runner/provider/azure/metadata_test.go
```diff
@@ -105,7 +105,7 @@ func testServer() *httptest.Server {
 		}
 
 		if r.URL.Path == "/metadata/attested/document" {
-			if apiVersion != "2019-08-15" {
+			if apiVersion != "2025-04-07" {
```
lotas, line 108: quite a jump into the future :) 

I hope this doesn't bring any new unexpected signatures to the worker manager, but you probably ran this already and it worked just fine?

matt-boris: Just used as a query param for fetching the metadata service, shouldn't affect worker manager at all 👍🏻 And yes, I've run plenty of tests where this new API version is hit many times, didn't affect wm at all.

https://github.com/taskcluster/taskcluster/blob/main/tools/worker-runner/provider/azure/metadata.go#L81

lotas: cool, thanks 🙏

lotas (approved):
lgtm

petemoore (approved):
Looks good, thanks Boris!

djmitche:
Is https://stackoverflow.com/questions/58483293/behavior-of-win32-named-pipe-in-duplex-mode a good summary of the deadlock issue here?

matt-boris:
> Is https://stackoverflow.com/questions/58483293/behavior-of-win32-named-pipe-in-duplex-mode a good summary of the deadlock issue here?

Yes, this is exactly what I read that helped things out. Since we use `winio.ListenPipe` it didn't seem to me at the time that it could be set up for asynchronous mode, which is why I went with the two pipe solution. Worked first try after I introduced the second pipe so I stopped trying any more with the async I/O 😅

PR #7932, opened 2025-08-27
UI: w-m page optimizations and Hooks new look

### LINKED ISSUE #7796: the /hooks page in the UI takes a long time to load
Loading https://firefox-ci-tc.services.mozilla.com/hooks makes a graphql query that blocks displaying the page and takes about 30s.

### LINKED ISSUE #7928: worker-manager page can fail to load data and return 504 Gateway Time-out instead
**Describe the bug**
Multiple people noticed https://firefox-ci-tc.services.mozilla.com/worker-manager failed to load data because the request to https://firefox-ci-tc.services.mozilla.com/graphql returned failure status 504 Gateway Time-out

lotas:
## Worker manager UI:

Loading error stats separately to decouple main worker pools query from the extra error stats which may slow down the page load times

Also adds few indexes to the slower queries that are used frequently.

## Hooks UI

Hooks are split in two pages now: 
1. Hook Groups
2. Hooks for specific hook group
+ breadcrumbs for easier navigation

This way we avoid loading all at once and allow better navigation / link sharing

image

image

image

Fixes #7928 
Fixes #7796


services/hooks/src/api.js
```diff
@@ -48,13 +47,9 @@ builder.declare({
     'This endpoint will return a list of all hook groups with at least one hook.',
   ].join('\n'),
 }, async function(req, res) {
-  const groups = new Set();
-  const hooks = (await this.db.fns.get_hooks(null, null, null, null)).map(hookUtils.fromDb);
```
lotas, line 52: this was loading all hooks before to find unique hook group ids on the client

services/hooks/src/api.js
```diff
@@ -63,7 +58,6 @@ builder.declare({
   route: '/hooks/:hookGroupId',
   name: 'listHooks',
   scopes: 'hooks:list-hooks:<hookGroupId>',
-  idempotent: true,
```
lotas, line 66: typescript complained that it's not used

ui/src/views/Hooks/routes.js
matt-boris, line 1: Wonder if we'll need to update some of our Fastly WAF allowed routes?

lotas: good thinking, but I didn't introduce new routes, so existing rules should be enough:

https://github.com/mozilla/webservices-infra/blob/b8f52d3ebb57f5ff5aab05f4f0c5397af3f3a4cd/taskcluster/tf/modules/waf_rules/main.tf#L46-L47

PR #7933, opened 2025-08-27
generic-worker/d2g: remove support for the dind and dockerSave features

jcristau:
dind was not actually implemented as far as I can tell.  dockerSave seems to be unused and broken, so rather than try to fix a feature we don't need, stop pretending it exists.

Github Bug/Issue: Fixes #XXXX

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -355,18 +355,6 @@ func artifacts(dwPayload *dockerworker.DockerWorkerPayload) []genericworker.Arti
 		gwArtifacts[i] = *gwArt
 	}
 
-	if dwPayload.Features.DockerSave {
-		gwArt := new(genericworker.Artifact)
-		defaults.SetDefaults(gwArt)
-
-		gwArt.Name = "public/dockerImage.tar.gz"
-		gwArt.Path = "image.tar.gz"
-		gwArt.Type = "file"
-		gwArt.Optional = true
-
-		gwArtifacts = append(gwArtifacts, *gwArt)
-	}
-
 	return gwArtifacts
 }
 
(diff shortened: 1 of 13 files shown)


workers/generic-worker/schemas/multiuser_posix.yml
matt-boris, line 1: I'd be fine completely removing the `dind` and `dockerSave` fields from the schemas, then you wouldn't even have to manually check for them from within d2g. Leave it up to the payload validation steps generic worker already has in place to resolve the task as exception/malformed-payload.

tools/d2g/d2g.go
```diff
@@ -488,6 +471,13 @@ func setFeatures(dwPayload *dockerworker.DockerWorkerPayload, gwPayload *generic
 		gwPayload.Features.BackingLog = false
 		gwPayload.Features.LiveLog = false
 	}
+	if dwPayload.Features.Dind {
+		return fmt.Errorf("the dind feature is not supported")
+	}
+	if dwPayload.Features.DockerSave {
+		return fmt.Errorf("the dockerSave feature is not supported")
+	}
+	return nil
```
matt-boris: Not necessary if we just remove the fields from the schemas.

workers/generic-worker/d2g_feature.go
matt-boris, line 1: Nice cleanup 🧹 👍🏻

workers/generic-worker/d2g_test.go
```diff
@@ -1103,3 +1103,23 @@ func D2GChainOfTrustHelper(t *testing.T, image d2g.Image, taskDependencies []str
 		t.Fatalf("Expected vs actual mismatch. Expected: %#v, Actual: %#v, Raw: %s", expected, cotCert, certString)
 	}
 }
+
+func TestD2GDind(t *testing.T) {
```
matt-boris: Thanks for the test! This _should_ continue to work properly after removing the fields from the schemas.

workers/generic-worker/schemas/multiuser_posix.yml
jcristau: I was thinking we don't need to reject tasks whose payload says `dind: false` for example.  (And actually given that the schema only allows `false`, the checks I added within d2g are probably not necessary.)

matt-boris: Understood, thanks for the explanation. I think if you remove the manual checks you added, this should be good. Would be interested in what @petemoore thinks about what to do here too.

matt-boris: Yeah the more I think of it, I think I'd prefer to just entirely remove it from the schema, especially so since we've checked the usage in FxCI and there is none. Let's not leave around any cruft.

matt-boris (approved):
Nice, LGTM 👍🏻

workers/generic-worker/d2g_test.go
jcristau: Turns out it doesn't, because `dockerworker.FeatureFlags(Dind: true)` is no longer a thing.

matt-boris (dismissed):

matt-boris (approved):
🧹

PR #7934, opened 2025-08-27
Fix: allow dots in GitHub repository names (closes #7776)

### LINKED ISSUE #7776: The github/builds API returns 500s if any repository in the return value contains a .
Traceback

```
2025-06-10T20:30:19.437Z api:schema Output schema validation error:
Schema Validation Failed!
Rejecting Schema: https://taskcluster.bananium.fr/schemas/github/v1/build-list.json#
Errors:
  * data/builds/0/repository must match pattern "^([a-zA-Z0-9-_%]*)$"
  * data/builds/1/repository must match pattern "^([a-zA-Z0-9-_%]*)$"
  * data/builds/2/repository must match pattern "^([a-zA-Z0-9-_%]*)$"
  * data/builds/12/repository must match pattern "^([a-zA-Z0-9-_%]*)$"
  * data/builds/14/repository must match pattern "^([a-zA-Z0-9-_%]*)$"
  * data/builds/847/repository must match pattern "^([a-zA-Z0-9-_%]*)$"
  * data/builds/848/repository must match pattern "^([a-zA-Z0-9-_%]*)$"
```

I checked, and the first build is:

```
{
  organization: 'Eijebong',
  repository: 'bananium.rs',
  sha: '2fce8f036a8fcf60ff8504c0e90f20490eaab3cb',
  state: 'pending',
  taskGroupId: 'Q11sCcJ5Twyfk-FMMckMOw',
  eventType: 'push',
  eventId: 'aa9ba544-46b2-11ef-97f6-dba446c1f39c',
  created: '2024-07-20T16:11:07.054Z',
  updated: '2024-07-20T16:11:07.054Z',
  pullRequestNumber: undefined
}
```

Relevant code:

https://github.com/taskcluster/taskcluster/blob/979b93219a15e4cc6c2f567e52e8411ccd375b85/services/github/schemas/v1/build-list.yml#L29

https://github.com/taskcluster/taskcluster/blob/979b93219a15e4cc6c2f567e52e8411ccd375b85/services/github/schemas/constants.yml#L7

pinakiz:
Fix: allow `.` in github-identifier-pattern

This updates the regex in the schema so that repository identifiers
can include the `.` character. Previously, such identifiers were rejected.

Changelog:
  - Updated `github-identifier-pattern` regex to include `.` .

Github Issue: Fixes #7776


petemoore (changes requested):
Many thanks @pinakiz!

Unfortunately the fix will need to be a bit more sophisticated. See line 5 and 6 above your change, it says

> Specifically, the length limitation and the fact that identifiers can't contain dots `.` is critical.

The issue is that github repositories appear in AMQP routing keys, for example in [Pull Request](https://community-tc.services.mozilla.com/docs/reference/integrations/github/exchanges#pullRequest) exchanges.

You see the github repository is the third (index 2) entry in the routing key, so having a `.` in the name would break the message binding.

In order to fix that we would need to change the routing key to be an encoded form of the github repository name, for example, which could also break existing clients and software that relies on the raw (unencoded) repository name being used in the routing key.

Your fix would have been fine, if it wasn't for this, but this rather complicates matters. Probably the real solution is for all routing key entries to escape the `.` somehow, but that would be a major change.

However, we appreciate your offer of a contribution here, and sorry that it did not work out this time!

Many thanks again.

pinakiz:
@petemoore Thanks a lot for the detailed explanation 🙏. Makes perfect sense about the routing key issue , I definitely learned something new here! Looking forward to contributing again in the future.

lotas:
hey @pinakiz I checked code as well, and found two places that say that dots are actually  might be fine:

https://github.com/taskcluster/taskcluster/blob/28d3aa0d5bc4ebe98c1728223984bc868bc99da3/services/github/src/exchanges.js#L16-L32


And there's a sanitization that is being used in API:

https://github.com/taskcluster/taskcluster/blob/28d3aa0d5bc4ebe98c1728223984bc868bc99da3/services/github/src/api.js#L21-L25

So it might be that this restriction on the schema level can be lifted, but the challenge here would be to actually test it, and see if existing tests are covering all cases where this might happen. 

Maybe we can reopen this and make sure sanitization is used everywhere

petemoore:
Ah, I just saw in the description of `repository` in my link earlier:

> The GitHub repository which had an event. All periods have been replaced by % - such that foo.bar becomes foo%bar - and all other special characters aside from - and _ have been stripped.

So indeed, tests here are key, to ensure the dots are replaced as desired, and anything consuming messages from the message bus are listening for the converted repository name, rather than the raw name.

An orthogonal concern from this comment above (not from the change in this PR) is about two different repositories sharing the same identifier (e.g. repositories `foo` and `foo^` both being indexed under `foo`). If the repository name is used as a key, and isn't guaranteed to be unique, this could potentially be exploited, or just cause corrupted data (where records overwrite each other).

services/github/schemas/constants.yml
```diff
@@ -4,7 +4,7 @@
 # limitation, the identifiers won't be useful as routing keys in RabbitMQ
 # topic exchanges. Specifically, the length limitation and the fact that
 # identifiers can't contain dots `.` is critical.
```
lotas: we should update this comment to tell that github service is doing sanitization and all dots are being replaced

services/github/schemas/constants.yml
```diff
@@ -4,7 +4,7 @@
...
-github-identifier-pattern:     "^([a-zA-Z0-9-_%]*)$"
+github-identifier-pattern:     "^([a-zA-Z0-9-_.%]*)$"
```
lotas, line 9: can you please run `yarn generate` to update all autogenerated schemas as well? Or just let me know, I can run it

lotas: Also, I didn't spot any tests that would cover such checks. Would you be comfortable checking if you could add some new tests for repositories with dots? 
Might be tricky to setup data for the tests though, I suggest starting with [api_test.js](https://github.com/taskcluster/taskcluster/blob/2b64dc5dd3e63bed3c8a5d2b5b4922b78e91c14b/services/github/test/api_test.js)

Thanks

pinakiz: Should I allow '%' ? Since ' . ' is later transformed into '%', allowing '%' could create ambiguity.

lotas: image
looks like github is already strict with repository names, I wonder how '%' got there in the first place

pinakiz: The git_Identifier_pattern also allows whitespace-only strings. Is this something we should be concerned about?"

lotas: I think we only care about the dot at this point, thank you!

Sorry, I forgot about this issue :)
Can you @pinakiz  please rebase and squash your comments, there were quite some changes since.

Thank you!

PR #7935, opened 2025-08-28
CI: stop running Generic Worker Insecure Linux CI tests as root

petemoore:
Only the Generic Worker multiuser engine should be tested as root, not the insecure engine.


lotas (dismissed):
lgtm

PR #7940, opened 2025-09-01
feat(w-m): update launch config always

### LINKED ISSUE #7918: Worker Manager: some fields in `workerManager` config are not dynamic and should not be ignored
`publicIp`, for example is a field within `workerManager` that should go into the calculation of a launch config ID. Because it's not, changes to this field (`publicIp`) do not propagate through to the worker pool config if it's the only change, since the launch config ID remains the same.

lotas:
Initial implementation was not updating configuration if the "main" part wasn't changed. But that left "workerManager" object that was supposed to be dynamic be excluded from updates, and the only way to update it was to make sure new launchConfigId is being created. Since we wanted to make "workerManager" part dynamic  (can be updated without changing launchConfigId), we update now configuration even when the main part isn't change, but workerManager part could change.

Fixes #7918


db/versions/0116.yml
```diff
@@ -0,0 +1,97 @@
...
+            ) THEN
+              RAISE EXCEPTION 'Launch config with ID `%` already exists with different configuration',
+                wp_launch_config_id
+                USING ERRCODE = 'unique_violation';
+            END IF;
+            -- make sure it is not archived and workerManager specific fields are updated
+            UPDATE worker_pool_launch_configs
+            SET is_archived = false,
+                last_modified = now(),
+                configuration = config
```
lotas, line 59: this is the only part changed from the original version https://github.com/taskcluster/taskcluster/blob/9b333a9427dc637c1bd67fbe47db004a203aa14e/db/versions/0105.yml#L85-L90

matt-boris (approved):

PR #7945, opened 2025-09-09
fix(d2g): pass env vars using `--env-file`

### LINKED ISSUE #7938: d2g: don't pass environment variables from the d-w payload directly to g-w
d2g currently sets `gwPayload.Env` to `dwPayload.Env`.  However this environment should only be passed to processes running in the docker container (e.g. with `docker run --env` or `--env-file`.  This is potentially problematic since the task user is privileged when running a d2g payload.

matt-boris:
Fixes #7938.

>D2G: pass Docker Worker environment variables to container using `--env-file` in `docker run` command.

cc @jcristau

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -190,6 +190,12 @@ func ConvertPayload(
 
 	setArtifacts(dwPayload, gwPayload)
 
+	if gwPayload.Env == nil {
+		gwPayload.Env = map[string]string{}
+	}
+	envVarsWithNewlines := ""
+	conversionInfo.EnvVars, envVarsWithNewlines = envMappings(dwPayload, gwPayload.Env, config)
+
 	gwWritableDirectoryCaches := writableDirectoryCaches(dwPayload.Cache)
 	dwImage, err := imageObject(&dwPayload.Image)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


changelog/issue-7938.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 7938
+---
+D2G: no longer pass Docker Worker env vars through to Generic Worker env vars, only supply them as flags to the `docker run` command.
```
jcristau: This doesn't look like it's actually passing the env vars, though?

matt-boris: We were always passing in the `docker run` command https://github.com/taskcluster/taskcluster/blob/31ad80ee8a3605e089821cdfdf12c67db2ead1f2/tools/d2g/d2g.go#L670-L673

matt-boris: Nvm, I see what you're saying. I forgot we removed the value of the env var to the field being passed in. Moving to draft.

matt-boris:
Test task here https://community-tc.services.mozilla.com/tasks/IB2fz09RQFG74R5fyBubxA works well!

tools/d2g/d2g.go
```diff
@@ -400,7 +408,8 @@ func runCommand(
 			}
 		}
 	}
-	command.WriteString(envMappings(dwPayload, config))
+	// Use env file that's created by D2G task feature
+	command.WriteString(fmt.Sprintf("%s --env-file env.list", envVarsWithNewlines))
```
jcristau, line 412: It seemed odd to me that we're doing different things for different variables here depending on the value.  On the other hand given moby/moby#12997 I guess env-file is not usable if we don't want to outright reject embedded newlines.  Disappointing.

tools/d2g/d2g.go
```diff
@@ -189,6 +190,12 @@ func ConvertPayload(
 
 	setArtifacts(dwPayload, gwPayload)
 
+	if gwPayload.Env == nil {
+		gwPayload.Env = map[string]string{}
+	}
+	envVarsWithNewlines := ""
+	conversionInfo.EnvVars, envVarsWithNewlines = envMappings(dwPayload, gwPayload.Env, config)
```
jcristau, line 197: That looks like some variables from the env are still getting passed through to the gwPayload?  IMO we should have a different mechanism here that keeps gwPayload.Env clear, to make sure the payload can't impact any process's environment outside the container.

matt-boris: I'll patch this in a future PR to pass along `-e <name>=<value>` to make it so we're not writing to gw's env.

PR #7946, opened 2025-09-09
fix(d2g): remove calls out to bash for image loading and cot additional data file

matt-boris:
>D2G: remove calls out to bash shells to load docker image and create chain of trust additional data file.

This work combined with https://github.com/taskcluster/taskcluster/pull/7933 will have all calls out to bash shells removed from d2g.

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -30,12 +30,7 @@ type (
 	DockerImageArtifact dockerworker.DockerImageArtifact
 	Image               interface {
 		FileMounts() ([]genericworker.FileMount, error)
-		String() string
-		ImageLoader() ImageLoader
-	}
-	ImageLoader interface {
-		LoadCommand() string
-		ChainOfTrustCommand() string
+		String(shellEscape bool) string
 	}
 	ConversionInfo struct {
 		ContainerName string
@@ -419,8 +414,7 @@ func runCommand(
 		}
 	}
 	command.WriteString(envMappings(dwPayload, config))
-	// note, dwImage.String() is already shell escaped
-	command.WriteString(" " + dwImage.String())
+	command.WriteString(" " + dwImage.String(true))
 	command.WriteString(" " + shell.Escape(dwPayload.Command...))
 	return [][]string{
 		{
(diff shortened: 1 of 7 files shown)


tools/d2g/d2g.go
```diff
@@ -30,12 +30,7 @@ type (
 	DockerImageArtifact dockerworker.DockerImageArtifact
 	Image               interface {
 		FileMounts() ([]genericworker.FileMount, error)
-		String() string
-		ImageLoader() ImageLoader
-	}
-	ImageLoader interface {
-		LoadCommand() string
-		ChainOfTrustCommand() string
+		String(shellEscape bool) string
```
petemoore, line 33: Is there a reason to do the shell escape inside the function, rather than shell escape the return value? I think the implementation is consistent across implementers, they apply `shell.Escape` to returned value if `shellEscape` is set to `true`, so I'm not sure why this logic should be copied across implementers?

matt-boris: It's just not how we were doing it beforehand. I've cleaned it up, anyway

https://github.com/taskcluster/taskcluster/pull/7946/files#diff-4ecf833adef40715069e50a184adecf5e50562d65c6b45ce4ad326df97003ebfL422-L423

matt-boris: Actually, it doesn't look like this'll work because indexed and artifact images are returning ``"${D2G_IMAGE_ID}"`` and we don't want that to be shell escaped because then it turns from:

```
        - docker run -t --name taskcontainer --memory-swap -1 --pids-limit -1 --add-host=localhost.localdomain:127.0.0.1
          -e ABC -e GHI -e RUN_ID -e TASKCLUSTER_INSTANCE_TYPE -e TASKCLUSTER_ROOT_URL
          -e TASKCLUSTER_WORKER_LOCATION -e TASK_GROUP_ID -e TASK_ID "${D2G_IMAGE_ID}"
          foo bar
```

to

```
        - docker run -t --name taskcontainer --memory-swap -1 --pids-limit -1 --add-host=localhost.localdomain:127.0.0.1
          -e ABC -e GHI -e RUN_ID -e TASKCLUSTER_INSTANCE_TYPE -e TASKCLUSTER_ROOT_URL
          -e TASKCLUSTER_WORKER_LOCATION -e TASK_GROUP_ID -e TASK_ID '"${D2G_IMAGE_ID}"'
          foo bar
```

Which I don't believe would work.

matt-boris: Yeah, d2g tests fail because of this https://community-tc.services.mozilla.com/tasks/Ta2nps8PRpukuV54V6LaKw/runs/0/logs/public/logs/live.log#L11531

PR #7950, opened 2025-09-11
Tracking more metrics 

### LINKED ISSUE #7942: Track more metrics with prometheus
Initial implementation covered most important parts, but there's still some other we can cover to track:

- http calls and response statuses
- auth failures
- websocket connections
- queue failed tasks resolutions (claim expired, deadline, etc) https://github.com/taskcluster/taskcluster/blob/a41bfd5d98081cf387e0925e4111882199bd711a/services/queue/src/api.js#L1676

lotas:
1. Collecting api calls metrics
    Renamed `path`->`name` because urls can get quite big, and come with greater variety. This explodes tracked buckets for each. Keeping it down to the api method name avoids having too many entries, plus it is probably safer not to store req.path as is to avoid unexpected injections of malicious stuff
2. enabling metrics exposing for all service that serve api endpoints 
3. auth server logs success/failure counts for auth attempts

Fixes #7942


libraries/monitor/src/builtins.js
```diff
@@ -166,9 +166,9 @@ MonitorManager.registerMetric('httpRequestsTotal', {
   description: 'Total number of HTTP requests',
   labels: {
     method: 'HTTP method',
-    path: 'HTTP path',
+    name: 'API method name',
```
lotas, line 169: those metrics wasn't exposed before, so change of label name is not really breaking anything

libraries/monitor/src/builtins.js
```diff
@@ -179,9 +179,9 @@ MonitorManager.registerMetric('httpRequestDurationSeconds', {
   description: 'HTTP request duration in seconds',
   labels: {
     method: 'HTTP method',
-    path: 'HTTP path',
+    name: 'API method name',
     status: 'Response status code',
-    service: 'Serice name',
+    service: 'Service name',
```
matt-boris, line 184: 🦅👁️

services/hooks/src/main.js
```diff
@@ -100,19 +100,24 @@ const load = loader({
   },
 
   api: {
-    requires: ['cfg', 'db', 'schemaset', 'taskcreator', 'monitor', 'publisher', 'pulseClient'],
-    setup: ({ cfg, db, schemaset, taskcreator, monitor, publisher, pulseClient }) => builder.build({
```
matt-boris, line 104: I noticed `pulseClient` was removed here, is that intentional?

matt-boris (commented):
Changes look good to me. Left one question.

services/hooks/src/main.js
lotas: yes, typescript complained it was never used - below not passed to builder, and I checked api.js - it wasn't used there as well

matt-boris (approved):

PR #7953, opened 2025-09-11
fix(d2g): remove shell call for `docker run` command

matt-boris:
>D2G: remove call out to bash shell that ran the `docker run` command. Directly call `docker` as the resulting Generic Worker payload command.

cc @jcristau

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -369,6 +369,8 @@ func runCommand(
 		containerName = fmt.Sprintf("%s_%s", containerName, slugid.Nice())
 	}
 
+	// Docker Worker used to attach a pseudo tty, see:
+	// https://github.com/taskcluster/taskcluster/blob/6b99f0ef71d9d8628c50adc17424167647a1c533/workers/docker-worker/src/task.js#L384
 	args := []string{"docker", "run", "-t", "--name", containerName}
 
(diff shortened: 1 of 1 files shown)


matt-boris:
Screenshot 2025-09-18 at 9 53 27 AM


In order to review the code changes easier, I'd recommend disabling viewing the `*.yml` test case files.

tools/d2g/d2g.go
```diff
@@ -370,53 +369,50 @@ func runCommand(
 		containerName = fmt.Sprintf("%s_%s", containerName, slugid.Nice())
 	}
 
-	command := strings.Builder{}
-
-	// Docker Worker used to attach a pseudo tty, see:
-	// https://github.com/taskcluster/taskcluster/blob/6b99f0ef71d9d8628c50adc17424167647a1c533/workers/docker-worker/src/task.js#L384
```
petemoore: i think it would be good to keep this comment, because it explains the `-t`.

petemoore (dismissed):
This looks great, and this clears a lot of the path in preparation for removing d2g tool entirely, getting a lot of logic out of the d2g tool and into the d2g feature. So thank you for all your effort here!

lotas (approved):
Thanks!

Issue #4986, opened 2021-09-02
GitHub pull requests with with `null` for `head.repo` fail to process


jwhitlock:
**Describe the bug**

When processing a GitHub pull request event, sometimes the `head.repo` object is `null`:

https://github.com/taskcluster/taskcluster/blob/27783611cc09478b99668abc688a1a5b34298a11/services/github/src/api.js#L26

This causes event processing to fail with `Cannot read property 'name' of null`

There is not enough debugging info to determine what event caused the error. The breadcrumbs in sentry may provide a clue.

**To Reproduce**

I do not know how to reproduce this. 

**Taskcluster version**

According to Sentry, this was first seen in community-tc in version 44.0.0, and is present in 44.2.2 (latest), seen 5 times in 2 months:

https://sentry.prod.mozaws.net/share/issue/f378a736276442a7afba00962daa5

A similar version is marked resolved due to inactivity in firefox-ci, from 37.3.0 to 43.2.0, and seen 23 times over a year:

https://sentry.prod.mozaws.net/share/issue/a514712054ee477b9a06daeb35166470/

**Additional context**

Issue #3455 mentions other missing data items from GitHub APIs.

PR #7958, opened 2025-09-16
Fix TypeError when GitHub database rows are missing

### LINKED ISSUE #3455: Handle missing rows in GitHub service
https://sentry.prod.mozaws.net/operations/taskcluster-firefox-ci/issues/9479421/
```
TypeError: Cannot read property 'sha' of undefined
  File "/app/services/github/src/handlers.js", line 911, col 5, in Handlers.taskDefinedHandler
    sha,
  File "internal/process/task_queues.js", line 97, col 5, in processTicksAndRejections
```

https://sentry.prod.mozaws.net/operations/taskcluster-firefox-ci/issues/9479428/
```
TypeError: Cannot destructure property 'organization' of 'build' as it is undefined.
  File "/app/services/github/src/handlers.js", line 486, col 8, in Handlers.statusHandler
    let {organization, repository, sha, event_id, event_type, installation_id} = build;
  File "internal/process/task_queues.js", line 97, col 5, in processTicksAndRejections
```

nagarwal-godaddy:
## Description
Fixes #3455 by adding defensive checks for missing database rows in GitHub service handlers.

## Problem
The GitHub service throws TypeErrors when:
- Tasks are created by periodic cron hooks (no associated GitHub build)
- Database queries return empty results
- GitHub API returns null for certain fields

## Solution
- Check if build data exists before destructuring
- Return early with debug messages for missing builds
- Add validation for required build properties
- Handle null values in GitHub webhook payloads

## Testing
- Added tests for all missing data scenarios
- Verified existing tests still pass
- Tested with simulated cron-triggered tasks

## Related Issues
- Fixes #3455
- Related to #4986 (null head.repo handling)


services/github/src/api.js
```diff
@@ -29,24 +29,30 @@ function sanitizeGitHubField(field) {
 //
 // See https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pullrequestevent
 function getPullRequestDetails(eventData) {
+  const pr = eventData.pull_request;
+
+  // Handle case where head.repo is null (referenced in issue #4986)
+  const headRepoName = pr.head.repo ? pr.head.repo.name : 'unknown';
```
lotas: This will not work, unfortunately. If `repo.name` is missing we can't just assume it is `"unknown"` since this even would be passed later to the status handler. 

We need to find out why it is empty in the first place, and stop processing event if it is missing, not just suppress the warnings this way

services/github/src/handlers/status.js
```diff
@@ -70,6 +70,14 @@ export async function statusHandler(message) {
   }
 
   const { organization, repository, sha, event_id, event_type, installation_id } = build;
+
+  // Validate required build properties
+  if (!sha || !organization || !repository) {
```
lotas: This doesn't make any sense -  db Schema for `github_builds` doesn't allow null values on any of those fields

services/github/src/api.js
nagarwal-godaddy: The `head.repo` can be null when the source repository of a PR is deleted or the repo becomes private/inaccessible. Should we look at capturing this early in the webhook processing?

lotas: Yes, I think we should validate it much sooner, and don't create this even in the first place if we know we would not be able to process it correctly. (Not able to checkout non-existing repo in handler)

services/github/src/api.js
```diff
@@ -29,24 +29,28 @@ function sanitizeGitHubField(field) {
 //
 // See https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pullrequestevent
 function getPullRequestDetails(eventData) {
+  const pr = eventData.pull_request;
+
+  // Since validation ensures head.repo exists, we can safely access it
+  // No need for fallback values anymore
```
lotas: This comment implies that something existed here before :) 
Can we leave this function as it was before, since it's not changing anything in terms of functionality?

nagarwal-godaddy: Done.

lotas:
We really appreciate the effort you put into this! 
Your PR was the catalyst for a very productive discussion on our side, and it helped us realize that a more architectural, schema-based solution is the right path forward - #7965. 
We also took another look at our production logs and noticed this particular `TypeError` hasn't occurred in several months, so it thankfully doesn't appear to be a critical issue. This gives us the confidence that it's better to take the time to implement the more robust fix rather than merging a patch right now.

PR #7961, opened 2025-09-17
Zip slip worker runner

petemoore:
Originally raised in #7917.

Many thanks @ptrgits!


yarn.lock
lotas, line 1: I don't see any package.json related changes, pushing yarn.lock is probably a mistake. Can you check if your local tooling is up to date, and maybe revert this file?

petemoore: whoopsie

matt-boris (approved):

PR #7976, opened 2025-09-25
feat(github): add webhook payload validation with JSON schemas (#7965)

### LINKED ISSUE #7965: Add payload schema validation for GitHub webhooks
This is related to #3455 issue

The endpoint responsible for ingesting webhooks from GitHub currently performs minimal to no validation on the incoming request payloads. This makes our service vulnerable to errors if GitHub sends an event with an unexpected structure, a missing field, or if a malformed test event is sent. Our code can throw unhandled exceptions, causing events to be dropped silently or with unhelpful error messages.

This issue was identified during the review of #7958, where it was decided that a systematic, schema-based approach is preferable to adding ad-hoc manual checks.

**Proposed Solution**

We should implement a validation layer at the entry point of the GitHub webhook handler. 

- Create JSON schemas for the GitHub event payloads we officially support (e.g., push, pull_request, check_suite, etc.). We should only define the fields that our application actually depends on.
- Use existing input schema validation for this
- Since this is not intended for clients, this schema doesn't need to be published and accessible in generated clients (tbd)

nitishagar:
# Github Bug/Issue: Fixes #7965
Implements JSON schema validation for GitHub webhook payloads to prevent TypeErrors from malformed or unexpected event structures. Validation is controlled via feature flags `WEBHOOK_VALIDATION_ENABLED` and `WEBHOOK_VALIDATION_BLOCK` for gradual rollout. 

# Testing
- Added tests for all missing data scenarios
- Verified existing tests still pass


services/github/config.yml
```diff
@@ -25,6 +25,9 @@ defaults:
 
   webhook:
     secret: !env:list:secret WEBHOOK_SECRET
+    validation:
```
lotas: I don't think configuration change is needed for this. Our goal is to discard events that don't contain fields we depend on. So this should be on by default and applied always

libraries/references/src/validate.js
```diff
@@ -29,6 +29,15 @@ const UNREFERENCED_SCHEMAS = [
   { service: 'generic-worker', schema: 'multiuser_posix.json#' },
   { service: 'generic-worker', schema: 'docker_posix.json#' },
   { service: 'docker-worker', schema: 'v1/payload.json#' },
+
+  // schemas for GitHub webhook payload validation (dynamically loaded at runtime)
```
lotas: Can you please explain why this needs  to be listed here as unreferenced schema? does generation fail? 
I thought that the best approach would be to define single `github-webhook` schema, that would be handled by standard library. So it would be referenced from github/src/api.js file

And that github-webhook would be OneOf the payloads we expect

services/github/src/api.js
```diff
@@ -277,6 +286,60 @@ builder.declare({
     }
   }
 
+  // Add webhook payload validation
+  const schemaMap = {
```
lotas: I think we would prefer to have single generic approach to api handler input schema validation, using standard library feature (input: 'schema-name') (as some other endpoints are using)
This way we won't need any of the custom code below.

You can likely keep those separate files as is, and have a top-level github-webhook-event.yml that would reference them with OneOf rules

lotas (changes requested):
Thanks for adding this! I left few requests how we can improve this. Goal is still to ensure valid payload reaches code, and everything necessary for validation is already implemented in libraries, no need to add new config keys, or custom code to validate schemas.

libraries/references/src/validate.js
nitishagar: I had tried dynamic loading earlier, but using a single schema is definitely much cleaner. Thanks for the tip!

tools/jsonschema2go/jsonschema.go
```diff
@@ -453,6 +453,11 @@ func (jsonSubSchema *JsonSubSchema) typeDefinition(disableNested bool, enableDef
 				extraPackages["tcclient \"github.com/taskcluster/taskcluster/v90/clients/client-go\""] = true
 			}
 		}
+	case "null":
```
lotas: We probably don't want to change go schemas for this, as clients will not be calling webhook endpoint, and this is a separate problem. We'll need to check what's the best way to separate those from schema generation for clients

services/github/src/api.js
```diff
@@ -218,6 +218,7 @@ export default builder;
 builder.declare({
   method: 'post',
   route: '/github',
+  input: 'github-webhook-event.yml',
```
lotas, line 221: nice, this is much cleaner 👍

services/github/schemas/v1/github-webhook-event.yml
```diff
@@ -0,0 +1,317 @@
...
+  # installation ID for GitHub App
+  installation:
+    type: object
+    properties:
+      id:
+        type: integer
+        minimum: 1
+    additionalProperties: true
+
+  # Pull request event specific validation
```
lotas: I liked previous approach where we have separate json schemas for each event, and then the main one would be simply `OneOf` choice. 
When we put everything into a single schema, it sure will work, but would be too broad and permissive. Of course, we can't control what github might send in a webhook, but I think it would be cleaner if we could separate properties by event type (just like the checks in code are doing currently). 
Does it make sense?

clients/client-go/tcgithub/tcgithub.go
```diff
@@ -135,9 +135,9 @@ func (github *Github) Version() error {
 // release, check run or pull request.
 //
 // See #githubWebHookConsumer
-func (github *Github) GithubWebHookConsumer() error {
+func (github *Github) GithubWebHookConsumer(payload *GitHubWebhookEvent) error {
```
lotas: @petemoore we talked about it before that some endpoints are not intended for clients at all. Should we separate / exclude such from client-side code generation and don't even show there?

services/github/schemas/v1/github-webhook-event.yml
nitishagar: Yeah, that makes sense. I had initially started the oneOf implementation that way, but it felt more like a wrapper around my original work, and I was expecting to find more commonality in the schemas—which didn’t really turn out to be the case.

Looking at it now, it does seem much clearer to keep separate schemas and apply oneOf over them. I’ll need a bit more time to circle back to this. I’ll try to take a pass at it over the weekend, otherwise it might be a little while.

Thanks again for taking the time to review.

tools/jsonschema2go/jsonschema.go
nitishagar: I noticed clients were making changes for github-webhook. Since this is server-side, do we actually need client having github-webhook at all?

In the last update, I reverted those changes, and also made the github-webhook private: https://github.com/taskcluster/taskcluster/pull/7976/files#diff-c4a0e4765607c1291cfe9120e8af6f5d69eaf08acef820ae0e90a5ccc5c824bdR227

@lotas could you confirm if this assumption is correct?

services/github/schemas/v1/github-webhook-event.yml
```diff
@@ -0,0 +1,15 @@
+$schema: "/schemas/common/metaschema.json#"
+title: "GitHub Webhook Event"
+description: |
+  Validates GitHub webhook event payloads to prevent errors from missing or
+  malformed fields. Uses OneOf pattern to strictly validate each event type
+  with its own schema, preventing TypeErrors identified in PR #7958.
+
+oneOf:
```
lotas, line 14: Thanks! I think this is a much cleaner approach

services/github/src/api.js
```diff
@@ -218,11 +218,13 @@ export default builder;
...
   name: 'githubWebHookConsumer',
   scopes: null,
   title: 'Consume GitHub WebHook',
   category: 'Github Service',
   stability: 'stable',
+  noPublish: true, // Webhook endpoint is server-side only, not called by clients
```
lotas, line 227: brilliant! I forgot this option even exists :)

tools/jsonschema2go/jsonschema.go
lotas: great! Thanks @nitishagar for spotting `noPublish` - this is exactly what we need for this endpoint, since it is not inteded for taskcluster clients (only integrations)

changelog/issue-7965.md
```diff
@@ -0,0 +1,5 @@
+audience: general
+level: minor
```
lotas: @matt-boris @petemoore what do you guys think about the level? This PR will drop `githubWebhookConsumer()` from client libs, so maybe it's a breaking one. But it was not supposed to be used by any clients, so likely `minor` is right? wdyt?

services/github/schemas/v1/webhook-pull-request-payload.yml
```diff
@@ -0,0 +1,137 @@
...
+title: "GitHub Pull Request Webhook Payload"
+description: |
+  Validates GitHub pull request event webhook payload.
+  Strictly validates fields that caused issues in PR #7958.
+
+type: object
+properties:
+  action:
+    type: string
+    enum: ["opened", "closed", "reopened", "synchronize", "edited"]
```
lotas: This is likely unnecessary. 
Github lists a lot more of actions, and if we would enforce this limited set, we might break some existing tasks that rely on different actions
 https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request

services/github/schemas/v1/webhook-check-run-payload.yml
```diff
@@ -0,0 +1,111 @@
...
+        type: string
+        pattern: "^[a-f0-9]{40}$"
+      status:
+        type: string
+        enum: ["queued", "in_progress", "completed"]
+      conclusion:
+        oneOf:
+          - type: "null"
+          - type: string
+            enum: ["success", "failure", "neutral", "cancelled", "timed_out", "action_required", "skipped"]
```
lotas: from the docs: https://docs.github.com/en/webhooks/webhook-events-and-payloads#check_run 
they list at least 11 options there

services/github/schemas/v1/webhook-check-run-payload.yml
```diff
@@ -0,0 +1,111 @@
...
+    properties:
+      id:
+        type: integer
+        minimum: 1
+      head_sha:
+        type: string
+        pattern: "^[a-f0-9]{40}$"
+      status:
+        type: string
+        enum: ["queued", "in_progress", "completed"]
```
lotas: github docs list 4 options here: https://docs.github.com/en/webhooks/webhook-events-and-payloads#check_run

services/github/schemas/v1/webhook-push-payload.yml
```diff
@@ -0,0 +1,86 @@
...
+  repository:
+    type: object
+    properties:
+      full_name:
+        type: string
+        pattern: "^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$"
+      clone_url:
+        type: string
+        format: uri
+        pattern: "^https://github\\.com/"
```
lotas: can we please drop this pattern everywhere? 

I see it's not used here, and I believe some enterprise github installations might have own domains? 
https://github.com/octokit/webhooks/blob/main/payload-schemas/api.github.com/common/repository.schema.json#L318

changelog/issue-7965.md
```diff
@@ -0,0 +1,5 @@
...
+reference: issue 7965
+---
+GitHub service now validates webhook payloads against a schema to prevent TypeErrors from missing or malformed fields. This ensures that webhook handlers receive properly structured data, the webhook now returns a 400 status and does not create an unprocessable event.
```
lotas, line 5: since we dropped webhook endpoint from publishing in schemas, and essentially dropped from clients, we should probably leave a note here that clients are no longer able to use this

services/github/test/webhook_test.js
```diff
@@ -87,7 +89,156 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
   // Bad data: should all return 400 responses
   statusTest('Push without secret', 'webhook.push.no_secret.json', 400);
   statusTest('Unknown Event', 'webhook.unknown_event.json', 400);
-  statusTest('Push with bad secret', 'webhook.push.bad_secret.json', 403);
-  statusTest('Release with bad secret', 'webhook.release.bad_secret.json', 403);
+  statusTest('Push with bad secret', 'webhook.push.bad_secret.json', 400); // Schema validation happens before signature check
```
lotas: I think we might be loosing our tests - we wanted to check signature validation here, and the fact that status code changed, means that schema is too strict (or our sample webhook data might be wrong). 
Can you please make sure that both payloads get us `403`?

changelog/issue-7965.md
nitishagar: Added line 7 --> The `githubWebHookConsumer` endpoint has been removed from client libraries as it was not intended for client use (only for GitHub webhook integrations).

services/github/schemas/v1/webhook-push-payload.yml
nitishagar: Added a test for enterprise domain as well: https://github.com/taskcluster/taskcluster/pull/7976/files#diff-71ca69908f181ea1f058d958cb1655c20a720474430e1723cba3d17358a36734R69

services/github/test/webhook_test.js
nitishagar: Makes sense, this is moved back to `403` - https://github.com/taskcluster/taskcluster/pull/7976/files#diff-71ca69908f181ea1f058d958cb1655c20a720474430e1723cba3d17358a36734R93

nitishagar:
Thanks @lotas for the feedback. I'm still working thru some of the changes. Will ping here once complete.

services/github/test/pulse_test.js
```diff
@@ -49,7 +49,9 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
     test(params.testName, async function() {
       // Trigger a pull-request message
       let res = await helper.jsonHttpRequest('./test/data/webhooks/' + params.jsonFile);
-      res.connection.destroy();
+      if (res.connection) {
```
lotas: do you know why this was needed? Where tests failing here somehow?

lotas: I tried locally and it works without this check, so maybe don't need to have it? 
And if there was a reason to wrap it this way, we can also use `res.connection?.destroy()`

services/github/schemas/v1/webhook-check-run-payload.yml
```diff
@@ -0,0 +1,131 @@
...
+      GitHub may add new action types; validation accepts any non-empty string.
+  check_run:
+    type: object
+    properties:
+      id:
+        type: integer
+        minimum: 1
+      head_sha:
+        type: string
+        pattern: "^[a-f0-9]{40}$"
```
lotas: One more thing please, before we will merge this - let's make it a bit more maintainable by adding constants as with some other validations, that can be reused (and changed in single place once needed):

for example 
```
pattern: {$const: github-sha-pattern}
```

and defining it in the https://github.com/taskcluster/taskcluster/blob/6e519409066831d5e75015318032eda69385c069/services/github/schemas/constants.yml

We can do the same for `full_name` below (or maybe even full `repository` 

Thank you again for working on this

services/github/test/webhook_test.js
```diff
@@ -106,4 +113,157 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
+        assert.ok(payload.body.ref, `${file} should have ref field`);
+        assert.ok(payload.body.commits, `${file} should have commits field`);
+      } else {
+        assert.ok(!payload.body.ref || payload.body.ref === undefined,
+          `${file} should not have ref field`);
+      }
+    }
+  });
+
+  // Note: Enum flexibility (accepting new action/status/conclusion values) is tested via
```
lotas: I think we can drop those comments

services/github/test/pulse_test.js
nitishagar: This was part of a prior change I did. I moved this to `res.connection?.destroy()` for safe side.

services/github/schemas/v1/webhook-check-run-payload.yml
nitishagar: Makes sense, this would be much clearner. thanks.

nitishagar:
@lotas Updated based on feedback.

lotas (dismissed):
Thanks @nitishagar for your time and patience :)
It is  looking good from my side.  I'll ask my team mates to have another quick look before we merge this.

services/github/schemas/v1/webhook-push-payload.yml
```diff
@@ -0,0 +1,90 @@
...
+  ref:
+    type: string
+    pattern: "^refs/(heads|tags)/.*$"
+  before:
+    type: string
+    pattern: {$const: github-sha-pattern}
+  after:
+    type: string
+    pattern: {$const: github-sha-pattern}
+  repository:
```
lotas, line 17: oh, I think I found another one:

https://github.com/taskcluster/taskcluster/blob/d5f1f87bf784b0de37f529ceedf0777724e1a76d/services/github/src/api.js#L439-L440

is using `repository.name` for all related events (all that contains `repository`), but here it is not being marked as required

And since it is being used in most of the new schemas, maybe we should also put whole repository into constants for reuse purposes?

also using [this schema](https://github.com/octokit/webhooks/blob/main/payload-schemas/api.github.com/common/repository.schema.json) for reference shows `name` should be marked as required

nitishagar: Updated the above.

lotas: Awesome! Thanks

changelog/issue-7965.md
matt-boris: I think it's fine to switch over to `major` just because we are removing this capability from the clients, even if they weren't supposed to be used at all, better to be safe here and mark the breaking change, even though it shouldn't matter.


```suggestion
level: major
```

nitishagar:
@lotas any update on this?

changelog/issue-7965.md
lotas: yes, let's make it major @nitishagar  as Matt suggests

nitishagar:
Thanks for the review @matt-boris. Updated this to `major` bump.
cc. @lotas

PR #7977, opened 2025-09-25
Azure ARM template deployments

lotas:
This feature adds ARM Templates support for azure launch configs.
Current implementation is using resource by resource provisioning (and deprovisioning) which becomes very slow with the rising number of workers to check. 

ARM Templates allow "single" async operation (start + check status calls) to trigger deployment and let Azure resource manager figure out full provisioning. 
And if template is crafted correctly (`deleteOption: Delete` for resources), it would also allow to deprovision instance quickly, removing all dependent resources automatically.

Main changes:

- new schema to support arm deployment
- new sdk is used to deploy templates (either inline or linked to a template spec)
- all resources are extracted from successful or failed deployment for later deprovisioning
- arm templates are required to have few specific parameters to be defined, otherwise deployment will fail: (`location`, `customData`, `vmName`, `computerName`, `adminPassword`, `adminUsername`)
- after successful deployment, deployment itself is being deleted (to avoid having too many deployments in history)
- it is possible to override RG for deployment `armDeploymentResourceGroup` (providerConfig.resourceGrup is used by default)

When using arm templates, full worker creation happens at provision time (similarly to aws/gcp), so worker scanner process is only updating internal data (worker.providerData)  and handles deprovisioning part. 

Implements: #7979

### DIFF
--- a/changelog/issue-7979.md
+++ b/changelog/issue-7979.md
@@ -1,5 +1,5 @@
 audience: worker-deployers
-level: major
+level: minor
 reference: issue 7979
 ---
 
--- a/services/worker-manager/src/providers/azure/index.js
+++ b/services/worker-manager/src/providers/azure/index.js
@@ -530,7 +530,12 @@ export class AzureProvider extends Provider {
           workerId: worker.workerId,
           workerGroup: workerGroup,
           armDeployment,
-          params: deploymentProperties.parameters, // safe to log adminpassword of the failed deployment?
+          params: {
+            ...deploymentProperties.parameters,
+            adminUsername: '***',
+            adminPassword: '***',
+            customData: '***',
+          },
         },
         launchConfigId: worker.launchConfigId,
       });
@@ -551,18 +556,18 @@ export class AzureProvider extends Provider {
 
     // update worker after successful or failed deployment with resources for later deprovisioning
     const extractDeployedResourcesAndUpdateWorker = async (modifier) => {
-      const depoloyedResources = await this.#extractResourcesFromDeployment(worker, monitor);
+      const deployedResources = await this.#extractResourcesFromDeployment(worker, monitor);
 
       await worker.update(this.db, worker => {
         modifier(worker);
         ['vm', 'nic', 'ip', 'disks'].forEach(resourceType => {
           if (resourceType === 'disks') {
             worker.providerData.disks ??= [];
-            worker.providerData.disks.push(...depoloyedResources.disks);
+            worker.providerData.disks.push(...deployedResources.disks);
           } else {
             worker.providerData[resourceType] = {
               ...(worker.providerData[resourceType] || {}),
-              ...depoloyedResources[resourceType],
+              ...deployedResources[resourceType],
             };
           }
         });
@@ -609,7 +614,8 @@ export class AzureProvider extends Provider {
           worker.providerData.provisioningComplete = true;
         });
 
-        // Clean up deployment to avoid hitting the 800 deployments per resource group limit
+        // Clean up deployment to avoid hitting the 800 deployments limit:
+        // https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#azure-management-group-limits
         // This should be safe because vm is already running
         // operation will start and will keep "deployment.operation" location in db until full deprovision of the worker
         await this.deprovisionResource({


package.json
```diff
@@ -24,6 +24,8 @@
     "@aws-sdk/s3-request-presigner": "^3.864.0",
     "@azure/arm-compute": "^22.4.0",
     "@azure/arm-network": "^34.0.0",
+    "@azure/arm-resources": "^7.0.0",
+    "@azure/arm-resourcesdeployments": "1.0.0-beta.1",
```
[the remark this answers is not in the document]
lotas: `@azure/arm-resourcesdeployments@npm:^1.0.0: No candidates found`
.. we'll have to wait

services/worker-manager/src/providers/azure/index.js
```diff
@@ -171,151 +232,495 @@ export class AzureProvider extends Provider {
...
+
+        await extractDeployedResourcesAndUpdateWorker(worker => {
+          worker.providerData.deployment.id = deployment.id;
+          worker.providerData.deployment.operation = undefined;
+          worker.providerData.deployment.outputs = outputs;
+          worker.providerData.vm.name = vmName;
+          worker.providerData.provisioningComplete = true;
+        });
+
+        // Clean up deployment to avoid hitting the 800 deployments per resource group limit
```
matt-boris: Maybe worth adding in the docs that point to this? https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits

services/worker-manager/src/providers/azure/index.js
```diff
@@ -171,151 +232,495 @@ export class AzureProvider extends Provider {
...
+      await this.reportError({
+        workerPool,
+        kind: 'creation-error',
+        title: 'Failed to create ARM deployment',
+        description: err.message,
+        extra: {
+          workerId: worker.workerId,
+          workerGroup: workerGroup,
+          armDeployment,
+          params: deploymentProperties.parameters, // safe to log adminpassword of the failed deployment?
```
matt-boris: 👀  Is it safe??

matt-boris: Could do something like:

```js
const safeParams = {
  ...deploymentProperties.parameters,
  adminPassword: { value: '[REDACTED]' },
  adminUsername: { value: '[REDACTED]' },
  customData: { value: '[REDACTED]' },
};

...

extra: {
  workerId: worker.workerId,
  workerGroup: workerGroup,
  armDeployment,
  params: safeParams,
},
```

services/worker-manager/src/providers/azure/index.js
```diff
@@ -171,151 +232,495 @@ export class AzureProvider extends Provider {
...
+   * @param {{ worker: Worker, monitor: import('@taskcluster/lib-monitor').Monitor }} opts
+   */
+  async #checkARMDeployment({ worker, monitor }) {
+    if (worker.providerData.provisioningComplete) {
+      return true;
+    }
+
+    // update worker after successful or failed deployment with resources for later deprovisioning
+    const extractDeployedResourcesAndUpdateWorker = async (modifier) => {
+      const depoloyedResources = await this.#extractResourcesFromDeployment(worker, monitor);
```
matt-boris:
```suggestion
      const deployedResources = await this.#extractResourcesFromDeployment(worker, monitor);
```

services/worker-manager/src/providers/azure/index.js
```diff
@@ -171,151 +232,495 @@ export class AzureProvider extends Provider {
...
+
+      await worker.update(this.db, worker => {
+        modifier(worker);
+        ['vm', 'nic', 'ip', 'disks'].forEach(resourceType => {
+          if (resourceType === 'disks') {
+            worker.providerData.disks ??= [];
+            worker.providerData.disks.push(...depoloyedResources.disks);
```
matt-boris:
```suggestion
            worker.providerData.disks.push(...deployedResources.disks);
```

services/worker-manager/src/providers/azure/index.js
```diff
@@ -171,151 +232,495 @@ export class AzureProvider extends Provider {
...
+          } else {
+            worker.providerData[resourceType] = {
+              ...(worker.providerData[resourceType] || {}),
+              ...depoloyedResources[resourceType],
```
matt-boris:
```suggestion
              ...deployedResources[resourceType],
```

changelog/issue-7979.md
```diff
@@ -0,0 +1,7 @@
+audience: worker-deployers
+level: major
```
matt-boris:
```suggestion
level: minor
```

I _think_ this isn't breaking, should be safe at `minor` level, but please correct me if I'm wrong!

matt-boris (commented):
Awesome improvements here Yarik! Looking forward to getting this shipped out! 💪🏻 

Left a few comments/suggestions :)

services/worker-manager/src/providers/azure/index.js
lotas: right.. good question. I think it doesn't hurt, since we generate random username/password for each machine and here we log it only when deployment itself fails, which means machine is unlikely to be there.. but true, i'll just redact them as you suggest 👍

services/worker-manager/src/providers/azure/index.js
lotas: 👁️ 🦅

lotas: image
this is the legit word :D

changelog/issue-7979.md
lotas: agree.. although schema changes but does it in a backwards compatible way, I think we can leave it minor

services/worker-manager/src/providers/azure/index.js
matt-boris: I think maybe a commonly misspelled way to write out `deployed` 😄

services/worker-manager/src/providers/azure/index.js
matt-boris: Maybe in rare case we have some bug where an admin's creds are reused for some reason? I think it's fine not logging these anyway tbh..

matt-boris (approved):
🎉 🚢

PR #7978, opened 2025-09-25
fix(d2g): remove on exit status retries 125, 128 codes

matt-boris:
>D2G: removes exit codes 125, 128 from `payload.onExitStatus.retry` array, as docker pulls now happen outside of the payload being run (inside the D2G task feature startup).


lotas (approved):
lgtm

PR #8016, opened 2025-10-15
chore(ci): use `uv` and `trixie`

matt-boris:
>Switches CI tasks to use `uv` tool and to the latest stable Debian release (`trixie`).


lotas (approved):
oh wow, it is faster indeed. Excellent idea

PR #8019, opened 2025-10-15
feat(w-m): Azure provider skips checking already deleted resources

lotas:
When VM was being deprovisioned it would check resource by resource and delete if necessary: vm -> nic -> ip -> disks. Each of them would first query api to see if resource is 404 to know for sure it was gone. But checks were insufficient in place to prevent those resources from being queried over and over. So vm would be checked always, nic x-1 times, ip x-2 times, etc.
By adding clear indication "deleted: true" we can skip consequent checks and avoid calling azure api

### DIFF
--- a/services/worker-manager/src/providers/azure/index.js
+++ b/services/worker-manager/src/providers/azure/index.js
@@ -1125,7 +1125,7 @@ export class AzureProvider extends Provider {
   /*
    * deprovisionResource attempts to delete a resource and verify deletion
    * if the resource has been verified deleted
-   *   * sets providerData[resourceType].id = false, signalling it has been deleted
+   *   * sets providerData[resourceType].deleted = true, signalling it has been deleted
    *   * returns true
    *
    */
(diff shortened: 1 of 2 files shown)


services/worker-manager/src/providers/azure/index.js
```diff
@@ -1125,7 +1125,7 @@ export class AzureProvider extends Provider {
   /*
    * deprovisionResource attempts to delete a resource and verify deletion
    * if the resource has been verified deleted
-   *   * sets providerData[resourceType].id = false, signalling it has been deleted
+   *   * sets providerData[resourceType].deleted = true, signalling it has been deleted
```
petemoore, line 1128: i like this a lot! explicit `deleted` field is much nicer than setting `id` to `false`. ++

petemoore (approved):
Great improvement, thanks Yarik!

services/worker-manager/src/providers/azure/index.js
```diff
@@ -1149,6 +1149,15 @@ export class AzureProvider extends Provider {
       resourceName: typeData.name,
     });
 
+    if (typeData?.deleted === true) {
+      // since setting id to false was a signal that the resource was already deleted
```
matt-boris: This comment can likely be cleaned up now since you've renamed the field

matt-boris (dismissed):
Changes LGTM, thank you. Nice find. I noticed one comment seems a bit outdated due to using the new, `deleted` field.

PR #8018, opened 2025-10-15
chore(client-py)!: drop py3.9, add py3.14 support/tests, client ci improvements

matt-boris:
>Client (python): Removes python 3.9 support. Adds support/testing for python 3.14.

`ghcr.io/astral-sh/uv:python3.9-trixie` is `1.64GB` compared to the new, in-tree python docker image that'll be built is `783MB`

This also removes a lot of duplicated task definitions in the `client/kind.yml` file by using Taskgraph's [Matrix transform](https://taskcluster-taskgraph.readthedocs.io/en/stable/reference/transforms/matrix.html).

PR #8021, opened 2025-10-16
fix(client-py): declare what python versions are supported

### LINKED ISSUE #8020: [client-py] Declare which Python versions are supported in the Python client
Taskgraph recently added a dependency on the Python client, and in so doing, we accidentally dropped support for Python 3.8 due to client-py only supporting Python 3.9 and above.

The reason Taskgraph's dependency resolution didn't complain, is the Python client doesn't actually declare which Python versions it supports so the resolver thinks that it supports all versions. It does set some classifiers, but those are not anything official and resolvers don't use them.

To remedy this, we should switch `client-py` to use the new pyproject.toml standard and set the `requires-python` key. There may be a way to do this using `setup.py` still, but IMO best to just migrate to modern practices while we're at it.

matt-boris:
Fixes #8020.

>Client (python): declares what python versions are supported. Moves to modern standards with `pyproject.toml` and `uv` usage.

### DIFF
--- a/clients/client-py/Makefile
+++ b/clients/client-py/Makefile
@@ -5,15 +5,15 @@ UV := $(shell which uv)
 
 .PHONY: test
 test: devel
-	uv run pytest
+	$(UV) run pytest
 
 .PHONY: lint
 lint: devel
-	uv run ruff check taskcluster test
+	$(UV) run ruff check taskcluster test
 
 .PHONY: format
 format: devel
-	uv run ruff format taskcluster test
+	$(UV) run ruff format taskcluster test
 
 .PHONY: update
 update: docs
@@ -28,7 +28,7 @@ clean:
 
 .PHONY: docs
 docs: devel
-	uv run --with grip grip --export README.md
+	$(UV) run --with grip grip --export README.md
 	@echo "Now, upload README.html wherever docs go!"
 
 .PHONY: devel
@@ -37,8 +37,8 @@ devel:
 		echo "uv not found, installing..."; \
 		curl -LsSf https://astral.sh/uv/install.sh | sh; \
 	fi
-	uv sync --all-extras
+	$(UV) sync --all-extras
 
 .PHONY: install
 install:
-	uv pip install -e .
+	$(UV) pip install -e .
--- a/clients/client-py/pyproject.toml
+++ b/clients/client-py/pyproject.toml
@@ -35,21 +35,7 @@ dependencies = [
     "async_timeout>=2.0.0",
 ]
 
-[project.optional-dependencies]
-test = [
-    "pytest",
-    "pytest-cov",
-    "pytest-mock",
-    "pytest-asyncio",
-    "httmock",
-    "mock",
-    "flake8",
-    "psutil",
-    "hypothesis",
-    "coverage",
-    "aiofiles",
-    "httptest",
-]
+[dependency-groups]
 dev = [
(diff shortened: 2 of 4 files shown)


taskcluster/kinds/client/kind.yml
```diff
@@ -80,7 +80,8 @@ tasks:
       command: >-
         eval $(python test/client-library-secrets.py) &&
         cd clients/client-py &&
-        TOXENV=py39 uvx tox
+        uv sync --all-extras &&
+        uv run pytest
```
ahal, line 83: `uv run` does an implicit sync already so no need to call sync separately. `run` also has an `--all-extras` flag (which I think might be the default even? But verify that before taking my word for it :p)

clients/client-py/pyproject.toml
```diff
@@ -0,0 +1,110 @@
...
+    "flake8",
+    "psutil",
+    "hypothesis",
+    "coverage",
+    "aiofiles",
+    "httptest",
+    "tox",
+    "ruff",
+    "grip",
+]
```
ahal, line 55: You want these to be `dependency-groups`, not optional dependencies:
https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies

Optional dependencies are for consumers who want to install optional features and would allow people to do `pip install taskcluster[dev]` (which isn't what you want here).

taskcluster/kinds/client/kind.yml
ahal: See my other comment, you only needed `--all-extras` because you were using optional dependencies by accident. After switching to groups, this flag won't be needed.

clients/client-py/pyproject.toml
ahal: Also up to you, but I probably wouldn't bother having separate `dev` and `test` groups.

ahal (approved):
Looks reasonable to me, thanks!

taskcluster/kinds/client/kind.yml
```diff
@@ -80,7 +80,7 @@ tasks:
       command: >-
         eval $(python test/client-library-secrets.py) &&
         cd clients/client-py &&
-        TOXENV=py39 uvx tox
+        uv run pytest
   py310:
```
ahal, line 84: Completely unrelated drive-by comment, but Taskgraph supports Github actions-like matrices you could use for this:
https://github.com/taskcluster/taskgraph/blob/main/taskcluster/kinds/test/kind.yml#L45

matt-boris: Was toying around with this yesterday! https://github.com/taskcluster/taskcluster/pull/8018/files#diff-ee26b47dfb4e74101a5c3a3f8651c9327b5d03078a351e6e655edf27bcfa3b9a

clients/client-py/Makefile
```diff
@@ -1,31 +1,44 @@
...
+VENV := .venv
 PYTHON := $(VENV)/bin/python
 APIS_JSON=$(PWD)/apis.json
-TOX := $(shell which tox)
+UV := $(shell which uv)
 
 .PHONY: test
 test: devel
-	$(TOX)
+	uv run pytest
```
lotas: why not `$(UV)` here since there is a variable on line 4 for this?

lotas (approved):
Awesome, looks modern now :)

PR #8028, opened 2025-10-20
fix(release): publish python client on release

matt-boris:
>Client (python): fixes release of client. v91.0.3 did not work in fixing.

### DIFF
--- a/clients/client-py/release.sh
+++ b/clients/client-py/release.sh
@@ -26,6 +26,7 @@ rm -rf build/
 if ! command -v uv &> /dev/null; then
     echo "uv not found, installing..."
     curl -LsSf https://astral.sh/uv/install.sh | sh
+    source $HOME/.local/bin/env
 fi
 
 # Build the package using uv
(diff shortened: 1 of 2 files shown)


clients/client-py/release.sh
```diff
@@ -26,6 +26,7 @@ rm -rf build/
 if ! command -v uv &> /dev/null; then
     echo "uv not found, installing..."
     curl -LsSf https://astral.sh/uv/install.sh | sh
+    source $HOME/.local/bin/env
```
petemoore, line 29: Nit: `"${HOME}/.local/bin/env"` (i.e. with quotation marks) would be preferred, but since `$HOME` contains no spaces, this is probably ok. If the `curl` fails, the `env` file also might not exist, and not sure if the script would exit (I think maybe `pipefail` might be needed for that) but that isn't something new introduced in this PR.

matt-boris: Thanks, I just copy/pasted exactly what uv spit out in the logs if I wanted to run right after installation.

PR #8033, opened 2025-10-23
feat(w-m): Properly pass tags to arm template deployment

lotas:
Template parameters should be passed as { value: xx }, yarik.. and not like plain objects as you would have imagined .. pay attention

Also introduces `workerManager.keepDeployment` - expecting deployment inputs/outputs in azure console might be helpful to debug template issues, so adding a flag to keep them no matter of the vm state.
Note: deployment will not be created at all if template+parameters are not even accepted by azure api (missing parameter for example)


changelog/issue-7979-1.md
matt-boris, line 1: I think it's probably worth jotting some stuff in here and making it non-silent. Maybe patch or even another minor since we're adding the keepDeployment option?

lotas: agree, thanks

PR #8057, opened 2025-11-03
fix(generic-worker): properly log out windows user syscall token

matt-boris:
While looking into https://github.com/taskcluster/taskcluster/issues/8083, we noticed the user token was improperly logged out as a quoted string instead of a hexadecimal value.

>Generic Worker (Windows): properly logs out `win32.LoadUserProfile()` errors with the user's `syscall.Token` in hex format instead of a quoted string.


lotas (approved):
Thanks!

PR #8069, opened 2025-11-07
chore(client-py): add python3.9 support back in

### LINKED ISSUE #8068: `client-py`: add python 3.9 support back in
Taskgraph needs to support python 3.9 still, so we should add it back in so they don't need to pin their taskcluster version so it never gets updated.

matt-boris:
Fixes #8068.

>Client (python): adds python 3.9 support back.

### DIFF
--- a/changelog/issue-8068.md
+++ b/changelog/issue-8068.md
@@ -1,0 +1,5 @@
+audience: general
+level: minor
+reference: issue 8068
+---
+Client (python): adds python 3.9 support back due to Taskgraph still needing to support it.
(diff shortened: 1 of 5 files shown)


changelog/issue-8068.md
```diff
@@ -0,0 +1,5 @@
+audience: general
+level: minor
+reference: issue 8068
+---
+Client (python): adds python 3.9 support back.
```
lotas: maybe it's useful to include here why we are doing here? 
"because taskgraph/gecko still uses it smth smth"

matt-boris: Updated 👍🏻 Thanks

lotas (approved):
thank you

PR #8071, opened 2025-11-08
feat(docs): add client-side multi-language API examples

### LINKED ISSUE #8049: UI: Include client library examples in the api calls
Currently API docs list all the endpoints and the input/output schemas.

Sometimes it is difficult to translate those into actual sdk calls or how to pass those arguments correctly, or even what is the exact http endpoint to be used with curl.

Suggestion to extend documentation with api call samples for:
- curl
- go client
- python (sync & async)
- node client
- web client
- rust client

Since we generate client libraries, we can also be able to provide usage examples

nitishagar:
The API reference documentation now includes code examples for every endpoint in multiple languages:
- **curl** - Raw HTTP requests
- **Go** - Go client library
- **Python** - Synchronous and asynchronous client libraries
- **Node.js** - Node.js client library
- **Web** - Browser/web client library
- **Rust** - Rust client library
- **Shell** - Taskcluster CLI (taskcluster-cli)

Tascluster Screenshot

Each example demonstrates how to:
- Set up the client with authentication (both explicit credentials and environment variables)
- Call the API endpoint with proper parameters
- Handle errors appropriately
- Process the response

Examples are generated dynamically in the browser from API metadata, keeping the documentation lean while providing comprehensive, up-to-date code samples. Examples include syntax highlighting and can be copied to clipboard directly from the documentation.

Github Bug/Issue: Fixes #8049

### DIFF
--- a/changelog/issue-8049.md
+++ b/changelog/issue-8049.md
@@ -1,0 +1,23 @@
+audience: general
+level: minor
+reference: issue 8049
+---
+
+#### API Documentation Now Includes Code Examples
+
+The API reference documentation now includes code examples for every endpoint in multiple languages:
+- **curl** - Raw HTTP requests
+- **Go** - Go client library
+- **Python** - Synchronous and asynchronous client libraries
+- **Node.js** - Node.js client library
+- **Web** - Browser/web client library
+- **Rust** - Rust client library
+- **Shell** - Taskcluster CLI (taskcluster-cli)
+
+Each example demonstrates how to:
+- Set up the client with authentication (both explicit credentials and environment variables)
+- Call the API endpoint with proper parameters
+- Handle errors appropriately
+- Process the response
+
+Examples are generated dynamically in the browser from API metadata, keeping the documentation lean while providing comprehensive, up-to-date code samples. Examples include syntax highlighting and can be copied to clipboard directly from the documentation.
--- a/ui/src/utils/api-examples/go.js
+++ b/ui/src/utils/api-examples/go.js
@@ -1,0 +1,162 @@
+/**
+ * Generate Go client library examples for API endpoints
+ */
+
+import {
+  PLACEHOLDERS,
+  getPlaceholderValue,
+  requiresAuth,
+  capitalize,
+  formatPayloadJson,
+} from './helpers';
+
+/**
+ * Convert hyphenated service name to valid Go identifier
+ * @param {string} name - Service name (e.g., 'worker-manager')
+ * @returns {string} Valid Go identifier (e.g., 'workerManager')
+ */
+function formatGoIdentifier(name) {
+  if (!name.includes('-')) {
+    return name;
+  }
+
+  const parts = name.split('-');
+
+  return (
+    parts[0] +
+    parts
+      .slice(1)
+      .map(capitalize)
+      .join('')
+  );
+}
+
+/**
+ * Generate a Go example for an API endpoint
+ *
+ * @param {string} serviceName - Service name (e.g., 'queue')
+ * @param {string} apiVersion - API version (e.g., 'v1')
+ * @param {object} entry - API entry metadata
+ * @param {string} version - Taskcluster version (default: '93')
+ * @param {object} payloadExample - Example payload object (optional)
+ * @returns {string} Go code example
+ */
+export default function generateGoExample(
+  serviceName,
+  apiVersion,
+  entry,
+  version = '93',
+  payloadExample = null
+) {
+  const methodName = capitalize(entry.name);
+  const varName = formatGoIdentifier(serviceName);
+  const packageName = `tc${serviceName.toLowerCase().replace(/-/g, '')}`;
+  const params = entry.args.map(arg => `"${getPlaceholderValue(arg)}"`);
+  // Build imports
+  const imports = ['fmt', 'log'];
+
+  if (requiresAuth(entry)) {
+    imports.push('os');
+  }
+
+  if (entry.input) {
+    imports.push('encoding/json');
+  }
+
+  const importStatements = imports.map(imp => `	"${imp}"`).join('\n');
+  // Build client initialization
+  let clientInit;
+
+  if (requiresAuth(entry)) {
+    clientInit = `// Option 1: Explicit credentials
+	creds := &tcclient.Credentials{
+		ClientID:    "${PLACEHOLDERS.clientId}",
+		AccessToken: "${PLACEHOLDERS.accessToken}",
+	}
+	${varName} := ${packageName}.New(creds, "${PLACEHOLDERS.rootUrl}")
+
+	// Option 2: Credentials from environment variables
+	// Requires: TASKCLUSTER_ROOT_URL, TASKCLUSTER_CLIENT_ID, TASKCLUSTER_ACCESS_TOKEN
+	// creds := &tcclient.Credentials{
+	//     ClientID:    os.Getenv("TASKCLUSTER_CLIENT_ID"),
+	//     AccessToken: os.Getenv("TASKCLUSTER_ACCESS_TOKEN"),
+	// }
+	// ${varName} := ${packageName}.New(creds, os.Getenv("TASKCLUSTER_ROOT_URL"))`;
+  } else {
+    clientInit = `// No authentication required for this endpoint
+	${varName} := ${packageName}.New(nil, "${PLACEHOLDERS.rootUrl}")`;
+  }
+
+  // Build API call
+  let apiCall;
+
+  if (entry.input) {
+    let payloadCode;
+
+    if (payloadExample) {
+      // Use actual payload example with JSON unmarshaling
+      const jsonPayload = formatPayloadJson(payloadExample, 2);
+
+      payloadCode = `// Create request payload - see schema at:
+	// ${PLACEHOLDERS.rootUrl}/schemas/${serviceName}/${entry.input}
+	payloadJSON := []byte(\`${jsonPayload}\`)
+
+	payload := &${packageName}.${methodName}Request{}
+	if err := json.Unmarshal(payloadJSON, payload); err != nil {
+		log.Fatalf("Failed to parse payload: %v", err)
+	}`;
+    } else {
+      payloadCode = `// Create request payload
+	// See schema at: ${PLACEHOLDERS.rootUrl}/schemas/${serviceName}/${entry.input}
+	payload := &${packageName}.${methodName}Request{
+		// Fill in required fields according to the schema
+	}`;
+    }
+
+    apiCall = `${payloadCode}
+
+	// Call the API method
+	result, err := ${varName}.${methodName}(${params.join(', ')}, payload)`;
+  } else {
+    apiCall = `// Call the API method
+	result, err := ${varName}.${methodName}(${params.join(', ')})`;
+  }
+
+  // Build response handling
+  let responseHandling;
+
+  if (entry.output && entry.output !== 'blob') {
+    responseHandling = `// Handle successful response
+	fmt.Printf("Success: %+v\n", result)`;
+  } else if (entry.output === 'blob') {
+    responseHandling = `// Handle blob response
+	fmt.Printf("Downloaded %d bytes\n", len(result))`;
+  } else {
+    responseHandling = `// Method completed successfully
+	fmt.Println("Success")`;
+  }
+
+  // Assemble the complete example
+  const example = `package main
+
+import (
+${importStatements}
+
+	tcclient "github.com/taskcluster/taskcluster/v${version}/clients/client-go"
+	"github.com/taskcluster/taskcluster/v${version}/clients/client-go/${packageName}"
+)
+
+func main() {
+	${clientInit}
+
+	${apiCall}
+	if err != nil {
+		// Handle error
+		log.Fatalf("API call failed: %v", err)
+	}
+
+	${responseHandling}
+}`;
+
+  return example;
+}
--- a/ui/src/utils/api-examples/helpers.js
+++ b/ui/src/utils/api-examples/helpers.js
@@ -1,0 +1,164 @@
+/**
+ * Shared helper functions and constants for API example generators
+ */
+
+/**
+ * Standard placeholder values for consistent examples
+ */
+export const PLACEHOLDERS = {
+  rootUrl: 'https://tc.example.com',
+  clientId: 'your-client-id',
+  accessToken: 'your-access-token',
+  taskId: 'dSlITZ4yQgmvxxAi4A8fHQ',
+  taskGroupId: 'dSlITZ4yQgmvxxAi4A8fHQ',
+  runId: '0',
+  workerGroup: 'my-worker-group',
+  workerId: 'my-worker',
+  workerPoolId: 'my-worker-pool',
+  workerType: 'my-worker-type',
+  provisionerId: 'my-provisioner',
+  artifactName: 'public/build/target.tar.gz',
+  hookGroupId: 'project-taskcluster',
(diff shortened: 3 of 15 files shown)


ui/src/utils/api-examples/go.js
```diff
@@ -0,0 +1,100 @@
...
+  example += `import (\n`;
+  example += `\t"fmt"\n`;
+  example += `\t"log"\n`;
+
+  if (requiresAuth(entry)) {
+    example += `\t"os"\n`;
+  }
+
+  example += `\n`;
+  example += `\ttcclient "github.com/taskcluster/taskcluster/v92/clients/client-go"\n`;
```
lotas: Ideally we should take `/v92/` from one of the configs or substitute at run-time in docs probably to avoid having stale version number here

ui/src/utils/api-examples/index.js
```diff
@@ -0,0 +1,48 @@
...
+      pythonAsync: generatePythonExample(serviceName, apiVersion, entry, true),
+      node: generateNodeExample(serviceName, apiVersion, entry),
+      web: generateWebExample(serviceName, apiVersion, entry),
+      rust: generateRustExample(serviceName, apiVersion, entry),
+      shell: generateShellExample(serviceName, apiVersion, entry),
+    };
+
+    return examples;
+  } catch (error) {
+    // Silently return null on error to gracefully skip this entry
```
lotas: do we know what kind of errors we are suppressing here? Maybe it's better to let it bubble up? Might cause some head scratching in future if one of the functions start failing. 

One extra question here - why do we want to generate all examples at once, if we only display 1 at any single time in the tabs? Maybe we can pass target language here, and only generate example that is being displayed to avoid rendering everything in browser? Wdyt?

infrastructure/tooling/src/generate/generators/service-refs.js
matt-boris, line 1: Let's keep the changes limited to what was necessary for this PR please! This file can be reverted

changelog/issue-8049.md
```diff
@@ -0,0 +1,25 @@
...
+
+Each example demonstrates how to:
+- Set up the client with authentication (both explicit credentials and environment variables)
+- Call the API endpoint with proper parameters
+- Handle errors appropriately
+- Process the response
+
+Examples are generated dynamically in the browser from API metadata, keeping the documentation lean while providing comprehensive, up-to-date code samples. Examples include syntax highlighting and can be copied to clipboard directly from the documentation.
+
+This addresses issue #8049.
```
matt-boris:
```suggestion
```

Redundant, since you're referencing it correctly at the top on line 3, a released changelog will properly link to the issue already for us.

nitishagar:
> agree with @petemoore - I was about to add that it would be great to generate the actual payload as well, since we already render schema in docs for those calls.

Thanks, @petemoore. I noticed the same right after pasting the screenshot. The main challenge with client-side payload generation was handling the complexity around:

* Resolving `$ref` references across multiple schema files while preventing circular dependencies
* Producing contextually meaningful placeholder values instead of generic ones (e.g., realistic `taskIds`, timestamps, or scope patterns)

@lotas — I’ve implemented an approach inspired by what @petemoore mentioned. The updates include:

1. **Payload generation** (`payload-generator.js`) that:

   * Creates example payloads from JSON schemas with all required properties
   * Uses contextual placeholders for Taskcluster-specific values (`taskId`, `workerPoolId`, `clientId`, etc.)
   * Handles timestamp fields (`deadline`, `expires`, `created`)
   * Provides meaningful values for common patterns (`scopes`, `routes`, `owners`)
   * Resolves `$ref` with circular reference detection
2. **Schema URL references** included in examples for better documentation
3. **Lazy loading** to keep performance optimal despite the added complexity

Would love your thoughts on this approach.

ui/src/utils/api-examples/helpers.js
```diff
@@ -0,0 +1,164 @@
...
+export function getPlaceholderValue(paramName) {
+  // Direct mappings
+  if (PLACEHOLDERS[paramName]) {
+    return PLACEHOLDERS[paramName];
+  }
+
+  // Pattern-based mappings
+  if (paramName.endsWith('Id')) {
+    // For IDs that aren't in our mapping, use a slugid format
+    return 'dSlITZ4yQgmvxxAi4A8fHQ';
```
lotas: why not use the above-defined placeholders.taskId here? 

Also, it is possible just to use slugid in UI: https://github.com/taskcluster/taskcluster/blob/043af7ec3780fdbbd83ec5dcb1739e8ef6a565e0/ui/src/utils/slugid.js#L42

ui/src/utils/api-examples/index.js
```diff
@@ -0,0 +1,205 @@
...
+      generateWebExample(serviceName, apiVersion, entry, payloadExample),
+    rust: () =>
+      generateRustExample(serviceName, apiVersion, entry, payloadExample),
+    shell: () =>
+      generateShellExample(serviceName, apiVersion, entry, payloadExample),
+  };
+  const generator = generators[language];
+
+  if (!generator) {
+    // Unknown language requested
```
lotas: I would probably prefer an assert here and early failure or even exception, if someone is using this function incorrectly to avoid silencing errors

ui/src/utils/api-examples/index.js
```diff
@@ -0,0 +1,205 @@
...
+    return generateSingleExample(
+      language,
+      serviceName,
+      apiVersion,
+      entry,
+      version
+    );
+  }
+
+  // Generate all examples (used for backward compatibility if needed)
```
lotas: We should probably have some rules for llm agents to avoid leaving too many obvious comments and also mentioning "backward compatibility" in files that are just being introduced to the codebase. 

From developer point of view this comment will raise many questions

ui/src/utils/api-examples/index.js
```diff
@@ -0,0 +1,205 @@
...
+  // Note: In practice, this shouldn't be used with lazy loading enabled
```
lotas: What do you mean by this comment?

ui/src/utils/api-examples/payload-generator.js
```diff
@@ -0,0 +1,563 @@
...
+  // If meaningful value (not wrapped in <>), use it
+  if (placeholder && !placeholder.startsWith('<')) {
+    return placeholder;
+  }
+
+  // Additional context-aware mappings
+  const lowerName = propertyName.toLowerCase();
+
+  // Taskcluster-specific ID patterns
+  if (lowerName.includes('taskid') || lowerName === 'task') {
```
lotas: maybe `task` in some context will be an actual task object, no the ID?

ui/src/utils/api-examples/payload-generator.js
```diff
@@ -0,0 +1,563 @@
...
+
+  if (lowerName.includes('taskgroupid') || lowerName === 'taskgroupid') {
+    return PLACEHOLDERS.taskGroupId;
+  }
+
+  if (lowerName.includes('workerid')) {
+    return PLACEHOLDERS.workerId;
+  }
+
+  if (lowerName.includes('workerpoolid') || lowerName === 'workerpool') {
```
lotas: here probably too, `workerPool` might be an actual object

ui/src/utils/api-examples/payload-generator.js
```diff
@@ -0,0 +1,563 @@
...
+
+/**
+ * Resolve a $ref to its actual schema definition
+ * @param {string} ref - Reference like "worker-pool-full.json#/path"
+ * @param {object} currentSchema - Current schema for internal refs
+ * @param {Array} allSchemas - All available schemas
+ * @param {Set} visitedRefs - Track visited refs to prevent cycles
+ * @returns {object|null} Resolved schema or null
+ */
+function resolveRef(ref, currentSchema, allSchemas, visitedRefs = new Set()) {
```
lotas: Probably would be good to add some unit-tests. This seems to be an important function that resolves references, since we want to have correct data in examples. 

I noticed locally it might not be doing this correctly at the moment:

image

`capacity` is a `$ref` but not resolved here. 

And other refs seems to have same issues:
image

ui/src/utils/api-examples/payload-generator.js
```diff
@@ -0,0 +1,563 @@
...
+  if (schema.$id) {
+    if (visited.has(schema.$id)) {
+      return `<${schema.$id}>`;
+    }
+
+    visited.add(schema.$id);
+  }
+
+  // Handle oneOf - pick the first option
+  if (schema.oneOf && Array.isArray(schema.oneOf)) {
```
lotas, line 559: @nitishagar one question, did you check if there are any existing solutions to parse complex json schemas and generate some fake data? 

Maybe we can invert this problem and generate example payloads separately to avoid doing all of this here? 
`yarn generate` currently is building all those references. I wonder if we can integrate new step there to see if we can generate some meaningful data once. 
Just an idea.

ui/src/utils/api-examples/payload-generator.js
```diff
@@ -0,0 +1,563 @@
...
+      allSchemas
+    );
+  }
+
+  // Handle allOf - merge all schemas (simplified)
+  if (schema.allOf && Array.isArray(schema.allOf)) {
+    const merged = {};
+
+    schema.allOf.forEach(subSchema => {
+      // eslint-disable-next-line no-use-before-define
```
lotas, line 587: why this eslint rule? What is exactly being used here before define?

lotas:
Thank you for the updates! I think we are moving in the right direction

> Thanks, @petemoore. I noticed the same right after pasting the screenshot. The main challenge with client-side payload generation was handling the complexity around:
> 
>     * Resolving `$ref` references across multiple schema files while preventing circular dependencies
>     * Producing contextually meaningful placeholder values instead of generic ones (e.g., realistic `taskIds`, timestamps, or scope patterns)

I mentioned in one of the comments above - did you check any existing libraries doing this? Maybe we can reuse something. 
`taskId`, `taskGroupId`, etc are well-defined keywords across the api, so should be simple to match by name and either generate `slugid()/niceid()` or just hardcode some value


> @lotas — I’ve implemented an approach inspired by what @petemoore mentioned. The updates include:
> 
>     1. **Payload generation** (`payload-generator.js`) that:
>     2. **Schema URL references** included in examples for better documentation
>     3. **Lazy loading** to keep performance optimal despite the added complexity

Thanks! This is a nice addition. I spotted few issues with it and left a comment about adding some tests maybe, or checking why it can't resolve references at the moment




Can we please also make this a bit less "stretched": 
Aligning the borders, making tabs condensed, and maybe smaller fonts to fit more on the screen
image

ui/src/utils/api-examples/index.js
nitishagar: Good catch! I've moved this now to throw error.
```
throw new Error(`Unknown language requested: ${language}`);  // ✅ Now throws error
```

ui/src/utils/api-examples/index.js
nitishagar: 🤕 That is a constant headache. Removed this and the below one: https://github.com/taskcluster/taskcluster/pull/8071#discussion_r2513804362

ui/src/utils/api-examples/payload-generator.js
nitishagar: Addressed these in following format:
```
  // ✅ Only applies to string types now
  if (type === 'string' && lowerName.includes('taskid')) {
    return PLACEHOLDERS.taskId;
  }

  if (type === 'string' && lowerName.includes('workerpoolid')) {
    return PLACEHOLDERS.workerPoolId;
  }
```

Note: All Taskcluster-specific ID patterns (`taskId`, `taskGroupId`, `workerId`, `workerPoolId`, `provisionerId`,
  `workerType`, `clientId`, `accessToken`, `hookGroupId`, `hookId`) now have `type === 'string'` guards.

ui/src/utils/api-examples/payload-generator.js
nitishagar: Addressed below.

ui/src/utils/api-examples/payload-generator.js
nitishagar: I've added this under a dedicated test suite to prevent such issues. The suite currently covers:

**$ref Resolution (6 tests):**

* ✅ Internal $ref (`#/definitions/person`)
* ✅ External $ref (`person.json`)
* ✅ External $ref with path (`definitions.json#/definitions/metadata`)
* ✅ Nested $refs (3 levels deep: task → person → address)
* ✅ Circular $refs (handled without infinite loop)
* ✅ $refs containing other $refs (top → mid → base)

@lotas Let me know if there’s any additional case you think we should include.

ui/src/utils/api-examples/payload-generator.js
nitishagar: This one is required. The Circular Dependency:
```bash
  // File order (top to bottom):

  1. generateArrayExample()     // Line ~353
     ├─ calls generateExampleValue()  ❌ Not defined yet!

  2. generateObjectExample()     // Line ~407
     ├─ calls generateExampleValue()  ❌ Not defined yet!

  3. generateExampleValue()      // Line ~488
     ├─ calls generateArrayExample()  ✅ Already defined
     └─ calls generateObjectExample() ✅ Already defined
```

**Why this happens:**
These functions call each other:

* `generateExampleValue()` routes work to the right generator.
* `generateArrayExample()` calls it for each item.
* `generateObjectExample()` calls it for each property.

That mutual recursion triggers “use before define” because all three are written as **function expressions**, which aren’t hoisted.

**Why we can’t reorder:**
Moving one above the others just breaks a different call, and switching to declarations or classes would make the code more awkward.

The `eslint-disable` seems like the cleanest approach here. The functions are safely defined at runtime, and mutual recursion is expected here.

ui/src/utils/api-examples/payload-generator.js
nitishagar: I looked into other options `json-schema-faker` and `openapi-sampler` seem to be the actively maintained ones. The tradeoffs with each are:

- json-schema-faker: Depends on @apidevtools/json-schema-ref-parser, which adds ~200 KB.
- openapi-sampler: Requires pre-dereferencing, which introduces an extra build step.

Given that, I stayed with the current implementation. It’s around ~20 KB and adds a small amount of page-load overhead, but avoids the build-time complexity.

nitishagar:
@lotas / @matt-boris When you get a chance, can you take another look. We should have most of the feedback addressed.

ui/src/utils/api-examples/payload-generator.js
lotas: sounds good! thanks

dev-docs/best-practices/docs.md
```diff
@@ -60,3 +60,53 @@ Developer documentation can be found at:
 * [`/dev-docs`](../) - higher-level developer documentation that crosses boundaries between services and libraries
 
 * [`/dev-docs/best-practices`](./) - guidelines we've agreed on for Taskcluster development, in an effort to keep the codebase consistent and permit developers to move easily throughout the codebase.
+
+## API Code Examples
```
lotas: I'm not sure if this section is really useful

nitishagar:
> @nitishagar thanks for the updates! I think it looks much better now.
> 
> Would you be able to check the layout please, I think leaving tabs as is would not be user friendly. They take 2 pages, where in fact they can all fit on one screen: (plus paddings for the code example container is trimmed): image
> 
> (feel free to use some other controls instead of tabs to make it look compact)

Thanks @lotas. Moved this to a drop-down select: 
Screenshot 2025-11-28 at 11 14 55 AM

PR #8073, opened 2025-11-12
Rename docs section 'Workers' to 'Worker Manager Configuration'

petemoore:
I renamed this to Worker Manager Configuration to be consistent with the section Queue Configuration. This makes more sense to me because this is about how e.g. SRE should configure Worker Manager, rather than how e.g. RelOps should configure workers.


ui/docs/manual/deploying/worker-manager-config.mdx
lotas, line 1: Since we don't have any kind of redirects in place, maybe it's worth leaving `workers.mdx` blank with a link to `Moved to [worker-manager-config.mdx]`? 
This page might have been included in some docs already, would be good to not leave a dead end in form of 404

matt-boris: What if we just keep the filename the same and change its contents?

petemoore: That would also be possible but has disadvantage that filename/url is still inappropriate. I think the way it is implemented it gets the new appropriate name, and if you use the inappropriate one, you at least see it is renamed to the appropriate name, so there is no question about what the config relates to (i.e. the service configuration, not the worker configuration).

lotas (dismissed):

PR #8098, opened 2025-11-17
fix(generic-worker): adds checks to `CreateUserProfile` to ensure path exists

### LINKED ISSUE #8083: Workers coming up with failure "LoadUserProfile: The device is not ready."
Happening across multiple Windows OS versions, e.g.:

* https://community-tc.services.mozilla.com/worker-manager/proj-taskcluster%2Fgw-windows-2022-gui/errors
* https://community-tc.services.mozilla.com/worker-manager/proj-taskcluster%2Fgw-windows-11-24h2-staging/errors

May be related to moving the image building to packer, since time-wise this seems to be new since moving.

Typical error looks like:

```
LoadUserProfile('̈', &{Size:56 Flags:1 Username:0xc000433050 ProfilePath:<nil> DefaultPath:<nil> ServerName:<nil> PolicyPath:<nil> Profile:0}): LoadUserProfile: The device is not ready.
```

Originates from https://github.com/taskcluster/taskcluster/blob/0fe066c9a95e48b26b582824b00d97d0b0f96181/workers/generic-worker/process/logininfo_multiuser_windows.go#L64.

Initial thoughts: maybe packer creates an image which does something special on first bootup?

Since headlessTasks is `false` at least in `proj-taskcluster/gw-windows-2022-gui` this failure is probably happening when Generic Worker tries to remap the AppData folder for the new task user it has created for the subsequent task run (after the reboot), i.e. not for the currently logged in task user, since LoadUserProfile is called to get an auth token in the case there is not an interactive desktop session for the user.

matt-boris:
Fixes #8083. Followup to #8084.

>Generic Worker (windows): adds retry logic around `CreateUserProfile` method to ensure the task user's profile path is created successfully before continuing on to `LoadUserProfile`.

### DIFF
--- a/workers/generic-worker/d2g_test.go
+++ b/workers/generic-worker/d2g_test.go
@@ -950,7 +950,7 @@ func TestD2GChainOfTrustNamedDockerImage(t *testing.T) {
 
 	expected := DockerWorkerPartialChainOfTrustCert{
 		Environment: PartialEnvironment{
-			ImageHash:         "sha256:d1c5c701d837b59386563dc702932d200adc41f8bd99a1de451e6fcf378dbfa5",
+			ImageHash:         "sha256:9db1327bff0f7565c2e658371ec726bc2475fa30fce7366f6b1579894fd621d6",
 			ImageArtifactHash: "",
 			Region:            "test-worker-group",
 		},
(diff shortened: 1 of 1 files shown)


lotas (approved):
thanks

Issue #7147, opened 2024-07-12
RFC191: worker should fetch launch config status to know if it should shutdown/restart


lotas:
This should replace existing `deploymentId` logic that was tracking changes on the worker-pool configuration side and allowed worker to make a decision if it should shutdown/restart in order to fetch new config values or just stop.

Static workers would know that something changed as well, and would be able to pull new configuration.

This can be implemented together with `should-worker-shutdown` logic, that would allow to guarantee `minCapacity` is met, where w-m would be able to ask worker to stay up longer.

lotas:
I'm removing this from the backlog for now to clarify if this is needed at all, or existing `deploymentId` logic is enough.
cc @petemoore

PR #8112, opened 2025-11-20
feat(w-m): Experimental shouldWorkerTerminate() endpoint

lotas:
Related to #7147 (we can consider fixed once the Worker Manager side is updated with real policies).

>**Breaking changes:**
>- The Generic Worker configuration properties `deploymentId` and `checkForNewDeploymentEverySecs` are no longer supported and must not be used
>- Generic Worker exit code 70 now indicates "Worker Manager advised termination" instead of "non-current deployment ID"
>
>**New features:**
>- New experimental `shouldWorkerTerminate` API endpoint in Worker Manager allows workers running with worker runner (`--with-worker-runner`) to query whether they should terminate
>- New scope `worker-manager:should-worker-terminate:<workerPoolId>/<workerGroup>/<workerId>` required to call this endpoint

### DIFF
--- a/changelog/issue-7147.md
+++ b/changelog/issue-7147.md
@@ -1,0 +1,11 @@
+audience: worker-deployers
+level: major
+reference: issue 7147
+---
+**Breaking changes:**
+- The Generic Worker configuration properties `deploymentId` and `checkForNewDeploymentEverySecs` are no longer supported and must not be used
+- Generic Worker exit code 70 now indicates "Worker Manager advised termination" instead of "non-current deployment ID"
+
+**New features:**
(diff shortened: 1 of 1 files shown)


matt-boris (approved):
I built a staging ubuntu image off of this branch and successfully ran tasks with the worker without the no-op from worker manager getting in the way of things (terminating the instance).

Note: we'll need to remove this block from being generated https://github.com/taskcluster/community-tc-config/blob/336ef5cd3f997508a4e9ea37a41723de79fbb135/generate/workers.py#L788-L802 and it doesn't appear that `checkForNewDeploymentEverySecs` is used at all, so we're fine on that side of things.

PR #8114, opened 2025-11-20
Increase the buffer size on websocktunnel

Eijebong:
This will hopefully help with catching up on livelogs when reading from a task that has been started for a while.


lotas (approved):
sure, why not. Thanks!

PR #8116, opened 2025-11-21
feat(livelog): Increase throughput for large logs

lotas:
This should theoretically improve streaming really big live logs. Websocktunnel was already upgraded to have 4Mb buffer size (#8114)

cc @Eijebong


Eijebong:
Which I think is around what we were seeing before (cc @jcristau).

PR #8119, opened 2025-11-24
Use proper URL encoding in the go client for path parameters

### LINKED ISSUE #7765: Artifacts uploaded with a space in their name get uploaded with a plus sign replacing the space
Example of task:

```yaml
retries: 5
created: '2025-06-06T15:22:33.367Z'
deadline: '2025-06-06T18:22:33.367Z'
expires: '2026-06-06T18:22:33.367Z'
taskQueueId: ap/linux-small
projectId: none
tags: {}
scopes: []
payload:
  image: debian:buster-slim
  command:
    - /bin/bash
    - '-c'
    - mkdir /foo && cd foo && echo 1 > "a b"
  artifacts:
    public/foo:
      path: /foo
      type: directory
  maxRunTime: 90
extra: {}
dependencies: []
metadata:
  name: example-task
  owner: name@example.com
  source: https://taskcluster.bananium.fr/tasks/create
schedulerId: taskcluster-ui
```

Run: https://taskcluster.bananium.fr/tasks/a98zOls3S2WkZik_3iX68A

Image

CC @djmitche

Eijebong:
The go client was using `QueryEscape` to escape path parameters, however, that is not the correct function to escape paths (`PathEscape` as the name implies is better suited for that).

`QueryEscape` follows the `application/x-www-form-urlencoded` spec, which encodes more than what paths require but also replaces spaces by `+`. The over-encoding isn't an issue since the server would decode that, but the space replacement means that for example, a secret with a space in it could never get retrieved by the go client (as it would request `a+b` instead of `a b`.)

This behavior was very visible when uploading artifacts with a space in their name as they'd end up stored with a `+` instead. It gets worse when you had an artifact named `a b` and `a+b` as it would upload the second one on top of the first. All services in the client were affected but since the taskcluster CLI uses its own URL builder for API calls (where it calls `strings.ReplaceAll(url.QueryEscape(v), "+", "%20")` to work around this issue), the only product really affected by this change is generic-worker.

I added a simple test using the secrets service because this is the easiest way I figured out how to test the behavior without a complicated setup.

This is a breaking change because anyone relying on space in artifacts being replaced by a `+` will have to remove their workarounds for it.

It's also worth noting that this brings the go client in line with the other clients (js, py, rs) which were already all doing the right thing.

Fixes #7765

### DIFF
--- a/changelog/issue-7765.md
+++ b/changelog/issue-7765.md
@@ -1,0 +1,5 @@
+audience: users
+level: major
+reference: issue 7765
+---
+Spaces in artifact names are now correctly preserved instead of being replaced with `+`
--- a/clients/client-go/codegenerator/model/api.go
+++ b/clients/client-go/codegenerator/model/api.go
@@ -347,10 +347,10 @@ func (entry *APIEntry) generateDirectMethod(apiName string) string {
 	content += queryCode
 	content += "\tcd := tcclient.Client(*" + entry.Parent.apiDef.ExampleVarName + ")\n"
 	if entry.OutputURL != "" {
-		content += "\tresponseObject, _, err := (&cd).APICall(" + apiArgsPayload + ", \"" + strings.ToUpper(entry.Method) + "\", \"" + strings.ReplaceAll(strings.ReplaceAll(entry.Route, "<", "\" + url.QueryEscape("), ">", ") + \"") + "\", new(" + entry.Parent.apiDef.schemas.SubSchema(entry.OutputURL).TypeName + "), " + queryExpr + ")\n"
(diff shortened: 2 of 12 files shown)


petemoore:
Mystery solved! We started using `url.QueryEscape` in [October 2015](https://github.com/taskcluster/taskcluster-client-go/pull/3/commits/0a41a2e8a821240d11282dde8bc3f3aef92424d1#diff-8fde4806495341655b0d6528e0428756b393ef99a7602c8bd5b5ac0482a63cdcL362-R371); at that time `url.PathEscape` did not exist. It was added to the standard library a year later in [October 2016](https://github.com/golang/go/issues/13737#issuecomment-254391851). I guess `url.QueryEscape` was the closest thing we had at the time, without rolling our own, and maybe we overlooked the `' ' -> '+'` translation and thought it was sufficient.

PR #8126, opened 2025-11-27
feat(w-m): Improve GCP error handling during provision

lotas:
After @googleapis/compute upgrade, errrors were passed differently, so they failed to be reported to worker pool

### DIFF
--- a/services/worker-manager/src/providers/google.js
+++ b/services/worker-manager/src/providers/google.js
@@ -84,6 +84,22 @@ export class GoogleProvider extends Provider {
     this.oauth2 = new gcpCompute.auth.OAuth2();
   }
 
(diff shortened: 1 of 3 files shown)


services/worker-manager/src/providers/google.js
```diff
@@ -84,6 +84,22 @@ export class GoogleProvider extends Provider {
     this.oauth2 = new gcpCompute.auth.OAuth2();
   }
 
+  #extractGoogleApiErrors(err) {
+    if (err.response?.data?.error?.errors && Array.isArray(err.response.data.error.errors)) {
```
lotas, line 88: err.response is Gaxios (google request library) response, which contains extra info here

PR #8127, opened 2025-11-28
fix(ui): update task duration in real-time when filtering by status

### LINKED ISSUE #8104: Task Group viewer task durations not updating when filtering by task status
If you look at tasks in the task group viewer, the Duration of running tasks are updated in real time. However, if you filter by Running status (blue) then you see the running tasks, but the Duration field is not updated in real time, and is frozen at the value when you clicked on the blue "Running" box.

nitishagar:
When filtering tasks by status (e.g., clicking 'Running'), the Duration component was not updating in real-time. This was because react-window reuses component instances at each index, and when filtering changed which task appeared at a given index, the Duration component's interval was not properly reset.

The fix adds a key prop to the Duration (TimeDiff) component based on taskId and start time, ensuring proper component lifecycle management when the displayed task changes due to filtering or reordering.

Also adds unit tests for the Duration component covering:
- Fixed duration display (completed tasks)
- Real-time updates (running tasks)
- Key prop handling for task switching
- Interval cleanup when task completes

Fixes #8104

### DIFF
--- a/ui/src/components/Duration/index.test.jsx
+++ b/ui/src/components/Duration/index.test.jsx
@@ -1,0 +1,114 @@
+import React from 'react';
+import { render, act } from '@testing-library/react';
+import Duration from './index';
+
+// Mock timers for testing setInterval behavior
+jest.useFakeTimers();
+
+describe('Duration component', () => {
+  afterEach(() => {
+    jest.clearAllTimers();
+  });
+
+  it('should render duration with fixed offset (completed task)', () => {
+    const from = '2025-01-01T00:00:00.000Z';
+    const offset = '2025-01-01T00:05:30.000Z';
+    const { container } = render(<Duration from={from} offset={offset} />);
+
+    expect(container.textContent).toBe('05m 30s ');
+  });
+
+  it('should start interval when offset is not provided (running task)', () => {
+    const from = new Date(Date.now() - 65000).toISOString(); // 65 seconds ago
+    const { container } = render(<Duration from={from} />);
+
+    expect(container.textContent).toMatch(/01m \d+s/);
+
+    act(() => {
+      jest.advanceTimersByTime(1000);
+    });
+
+    expect(container.textContent).toMatch(/01m \d+s/);
+  });
+
+  it('should update in real-time when no offset is provided', () => {
+    const originalDate = global.Date;
+    const startTime = new originalDate('2025-01-01T00:01:00.000Z').getTime();
+    let currentTime = startTime;
+
+    class MockDate extends originalDate {
+      constructor(...args) {
+        if (args.length === 0) {
+          super(currentTime);
+        } else {
+          super(...args);
+        }
+      }
+
+      static now() {
+        return currentTime;
+      }
+    }
+
+    global.Date = MockDate;
+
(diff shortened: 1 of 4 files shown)


nitishagar:
@lotas / @matt-boris when you get a chance, can you review this change.

ui/src/components/TaskGroupTable/index.jsx
```diff
@@ -108,7 +108,15 @@ const ItemRenderer = ({ data, index, style }) => {
         component="div"
         role="cell">
         <span>
-          {run ? <TimeDiff from={run.from} offset={run.to} /> : 'n/a'}
+          {run ? (
+            <TimeDiff
+              key={`${taskGroup.taskId}-${run.from}`}
```
lotas: should we also append `run.to` to the key in such case? Tasks can be running and changing states, so if it was running when first rendered, and then finished, this key would remain the same, and would likely not be updated once the task finishes.

ui/src/components/Duration/index.test.jsx
```diff
@@ -0,0 +1,133 @@
...
+      jest.advanceTimersByTime(5000);
+    });
+  });
+
+  it('should display hours correctly', () => {
+    const from = '2025-01-01T00:00:00.000Z';
+    const offset = '2025-01-01T02:30:45.000Z';
+    const { container } = render(<Duration from={from} offset={offset} />);
+
+    // Should show 2 hours 30 minutes 45 seconds
```
lotas: I feel like most of those comments are not really needed, because code below is perfectly readable, and having those extra comments don't add much besides distraction imo

ui/src/components/TaskGroupTable/index.jsx
lotas: Actually, if this is is using react-window, I don't think this is the right component to add those keys. It should probably be the root element.

Moreover, check how this component is being used - we are applying React.memo() which means it would always produce same output as the first time it was used, and we only pass task index, which means those states would never be updated as well. 

This should be fixed differently either by dropping react memo (which might be costly) or always passing some extra props to it (like state or run times to force re-render it)

lotas (changes requested):
Thanks for the PR! I feel like this should be fixed differently, because the root cause is not addressed

ui/src/components/Duration/index.test.jsx
nitishagar: Fair point - the test names and assertions are self-explanatory.

ui/src/components/TaskGroupTable/index.jsx
nitishagar: Spot on. The key on `TimeDiff` wouldn't help because `React.memo` prevents the parent from re-rendering - the key change would never be evaluated. Implemented using `itemKey` on the List component:
`itemKey={(index, data) => data.items[index]?.node.taskId}`

Added tests to verify the behavior.

nitishagar:
Thanks for the review @lotas. The React.memo analysis was correct - the original fix addressed a symptom, not the cause. Updated to use itemKey on the List component, removed the TimeDiff key, and cleaned up test comments.

Would like your feedback on the update.

PR #8132, opened 2025-12-01
doc: critical websocktunnel v94 service release notes

petemoore:
From @Eijebong on [matrix](https://matrix.to/#/!whDRjjSmICCgrhFHsQ:mozilla.org/$soYLWsRjJ0AZDBsGsMz1jfFl-xUmxvzTx_Lp19eSmSE?via=mozilla.org&via=matrix.org&via=braak.pro):

> Just to put it in writing somewhere, I had updated my workers but not websocktunnel and live logs were all broken after streaming one chunk. Updated websocktunnel and now everything's working fine (and fast).

@Eijebong can you check you are happy with my wording and that it accurately reflects what you experienced? Thanks!


Eijebong (approved):
I would probably not write "Websock Tunnel" but the rest looks correct to me.

PR #8163, opened 2025-12-12
fix(d2g): garbage collector prunes entire system

### LINKED ISSUE #8153: generic-worker with d2g waits for container cleanup before resolving tasks
While investigating differences in clone time for decision tasks I ran a test that wrote a million tiny files. (https://firefox-ci-tc.services.mozilla.com/tasks/I88KnJqiR1qHqYd5rCIEog). You can see `[taskcluster 2025-12-05T16:20:09.461Z] Task Duration: 1m8.629066664s` and yet the task took 5 while minutes to resolve because docker had to remove the container and that's **slow**.

The same test on docker worker shows `[taskcluster 2025-12-05 16:19:45.476Z] Successful task run with exit code: 0 completed in 65.461 seconds` and the task being resolved immediately.

Relevant code:

d-w marks the container for deletion here after the task finishes and resolves the task

https://github.com/taskcluster/taskcluster/blob/34be876dfac542c7d9e60124af5634d64f1c8e77/workers/docker-worker/src/task.js#L686-L691

The deletion actually happens here:

https://github.com/taskcluster/taskcluster/blob/34be876dfac542c7d9e60124af5634d64f1c8e77/workers/docker-worker/src/gc.js#L185-L189

g-w deletes the container immediately in d2g_feature's Stop():

https://github.com/taskcluster/taskcluster/blob/34be876dfac542c7d9e60124af5634d64f1c8e77/workers/generic-worker/d2g_feature.go#L246-L256

and the task only gets resolved after because

https://github.com/taskcluster/taskcluster/blob/34be876dfac542c7d9e60124af5634d64f1c8e77/workers/generic-worker/main.go#L924

https://github.com/taskcluster/taskcluster/blob/34be876dfac542c7d9e60124af5634d64f1c8e77/workers/generic-worker/main.go#L971-L974

and defers run in LIFO (https://go.dev/blog/defer-panic-and-recover).

This blocks the task despite it being done already which blocks any dependent until cleanup is done which is irrelevant to the task itself.

matt-boris:
Fixes #8153.

>Generic Worker (d2g): anonymous volumes from the task container are no longer removed at the end of a task run in order to resolve the task sooner. The garbage collector was updated to take care of these instead.

cc @Eijebong

### DIFF
--- a/workers/generic-worker/garbagecollector.go
+++ b/workers/generic-worker/garbagecollector.go
@@ -60,34 +60,51 @@ func runGarbageCollection(r Resources) error {
 		if r.Empty() {
 			break
 		}
+
 		err = r.EvictNext()
 		if err != nil {
 			return err
 		}
+
 		currentFreeSpace, err = freeDiskSpaceBytes(taskContext.TaskDir)
 		if err != nil {
 			return fmt.Errorf("could not calculate free disk space in dir %v due to error %#v", taskContext.TaskDir, err)
 		}
 	}
-	if currentFreeSpace < requiredFreeSpace {
-		if config.D2GEnabled() {
-			err := host.Run("docker", "image", "prune", "--all", "--force")
-			if err != nil {
-				return fmt.Errorf("could not run docker image prune to garbage collect due to error %#v", err)
-			}
-			err = os.Remove("d2g-image-cache.json")
-			if err != nil {
-				return fmt.Errorf("could not remove d2g-image-cache.json due to error %#v", err)
-			}
-			currentFreeSpace, err = freeDiskSpaceBytes(taskContext.TaskDir)
-			if err != nil {
-				return fmt.Errorf("could not calculate free disk space in dir %v due to error %#v", taskContext.TaskDir, err)
-			}
+
+	if currentFreeSpace < requiredFreeSpace && config.D2GEnabled() {
+		err := host.Run("docker", "volume", "prune", "--all", "--force")
+		if err != nil {
+			return fmt.Errorf("could not run docker volume prune to garbage collect due to error %#v", err)
+		}
+
+		currentFreeSpace, err = freeDiskSpaceBytes(taskContext.TaskDir)
+		if err != nil {
+			return fmt.Errorf("could not calculate free disk space in dir %v due to error %#v", taskContext.TaskDir, err)
 		}
 	}
+
+	if currentFreeSpace < requiredFreeSpace && config.D2GEnabled() {
+		err := host.Run("docker", "system", "prune", "--all", "--force")
+		if err != nil {
+			return fmt.Errorf("could not run docker system prune to garbage collect due to error %#v", err)
+		}
+
+		err = os.Remove("d2g-image-cache.json")
+		if err != nil {
+			return fmt.Errorf("could not remove d2g-image-cache.json due to error %#v", err)
+		}
+
+		currentFreeSpace, err = freeDiskSpaceBytes(taskContext.TaskDir)
+		if err != nil {
+			return fmt.Errorf("could not calculate free disk space in dir %v due to error %#v", taskContext.TaskDir, err)
+		}
+	}
+
 	if currentFreeSpace < requiredFreeSpace {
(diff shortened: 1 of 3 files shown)


workers/generic-worker/garbagecollector.go
```diff
@@ -69,20 +69,30 @@ func runGarbageCollection(r Resources) error {
...
+		// First try volume pruning - cheaper option that preserves image cache
+		err := host.Run("docker", "volume", "prune", "--all", "--force")
+		if err != nil {
+			return fmt.Errorf("could not run docker volume prune to garbage collect due to error %#v", err)
+		}
+		currentFreeSpace, err = freeDiskSpaceBytes(taskContext.TaskDir)
+		if err != nil {
+			return fmt.Errorf("could not calculate free disk space in dir %v due to error %#v", taskContext.TaskDir, err)
+		}
+	}
```
Eijebong: I would move that above the previous check and run it on every iteration of the garbage collector? 🤔

matt-boris: Ah yes, good call. Fixed 👍🏻

Eijebong: Actually, now that I think about it, for short lived workers it's probably better to not run it every time since there's a good chance that it would never be necessary. And since this blocks the worker from picking any other task while it's pruning...

matt-boris: Yep that's fine with me 👍🏻

workers/generic-worker/garbagecollector.go
```diff
@@ -60,34 +60,51 @@ func runGarbageCollection(r Resources) error {
...
-			err = os.Remove("d2g-image-cache.json")
-			if err != nil {
-				return fmt.Errorf("could not remove d2g-image-cache.json due to error %#v", err)
-			}
-			currentFreeSpace, err = freeDiskSpaceBytes(taskContext.TaskDir)
-			if err != nil {
-				return fmt.Errorf("could not calculate free disk space in dir %v due to error %#v", taskContext.TaskDir, err)
-			}
+
+	if currentFreeSpace < requiredFreeSpace && config.D2GEnabled() {
```
petemoore, line 75: Potentially we may wish to do this if D2G isn't enabled, because if a task runs a docker command, it could fill up the system, even if we are not using D2G (i.e. if a task runs command `docker ....` and includes the necessary entry in `osGroups`). However, we cannot be sure docker is installed if D2G isn't enabled, so we can also leave this in here for now, but something to think about later if non-d2g workers start running out of disk space.

workers/generic-worker/garbagecollector.go
```diff
@@ -60,34 +60,51 @@ func runGarbageCollection(r Resources) error {
...
+		err := host.Run("docker", "volume", "prune", "--all", "--force")
+		if err != nil {
+			return fmt.Errorf("could not run docker volume prune to garbage collect due to error %#v", err)
```
petemoore, line 78: I don't think we want to return here if docker volume cannot run, we should carry on with other clean up.

workers/generic-worker/garbagecollector.go
```diff
@@ -60,34 +60,51 @@ func runGarbageCollection(r Resources) error {
...
+		currentFreeSpace, err = freeDiskSpaceBytes(taskContext.TaskDir)
+		if err != nil {
+			return fmt.Errorf("could not calculate free disk space in dir %v due to error %#v", taskContext.TaskDir, err)
+		}
+	}
+
+	if currentFreeSpace < requiredFreeSpace && config.D2GEnabled() {
+		err := host.Run("docker", "system", "prune", "--all", "--force")
+		if err != nil {
+			return fmt.Errorf("could not run docker system prune to garbage collect due to error %#v", err)
```
petemoore, line 90: if this fails, e.g. because it half-completes, i'm not sure we want to leave d2g-image-cache.json file on disk. Maybe we should we delete d2g-image-cache.json regardless?

workers/generic-worker/garbagecollector.go
matt-boris: This panics anyway if it returns an error and doesn't make a difference either way

lotas (approved):

PR #8165, opened 2025-12-15
Let worker-runner unregister workers when they shutdown with idle timeout

Eijebong:
Currently on idle timeout, the worker decides to shut the machine down itself by using `shutdown now` and then `exit 68`. This leads to the machine shutting down while the worker is still registered, which lets worker-manager think that the worker is still alive until the cleanup loop passes by and realizes that the VM is gone which can take up to 15+mn. On pools that are not very active this can mean that we're waiting 15+mn for a new worker to spawn because there's only one task in the queue but worker manager thinks that there is a worker for it.

The root cause is that the exit code isn't read by worker-manager at all, when a worker exits with a non 0 code, it just exits itself without calling any exit handler, leaving the worker registered. Since the worker itself decides that its time has come but cannot handle its registration, worker-manager never even tries to unregister it.

Now one might think "oh that's easy, we'll just handle exit code 68 and unregister in that case" and it might work but here comes the second issue. The worker tells the machine to shut down before exiting. Which SIGTERM's worker-runner. So it most likely won't even have time to read the exit code from generic-worker, let alone make an HTTP request to unregister it.

To avoid racy behavior like this we could let the worker use `shutdown -h +1` to give a full minute to worker-manager but while looking at this, I saw that worker-manager actually exposes a `shutdown` command which allows the worker to tell it to unregister the worker which will make worker-manager terminate it.

### DIFF
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -34,6 +34,7 @@ import (
 	"github.com/taskcluster/taskcluster/v95/internal"
 	"github.com/taskcluster/taskcluster/v95/internal/mocktc/tc"
 	"github.com/taskcluster/taskcluster/v95/internal/scopes"
+	"github.com/taskcluster/taskcluster/v95/tools/workerproto"
 	"github.com/taskcluster/taskcluster/v95/workers/generic-worker/artifacts"
 	"github.com/taskcluster/taskcluster/v95/workers/generic-worker/errorreport"
 	"github.com/taskcluster/taskcluster/v95/workers/generic-worker/expose"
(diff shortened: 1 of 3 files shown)


Eijebong:
I have no good way of testing this so hopefully it just works...

workers/generic-worker/main.go
```diff
@@ -193,7 +194,15 @@ func main() {
 		case IDLE_TIMEOUT:
 			logEvent("instanceShutdown", nil, time.Now())
 			if config.ShutdownMachineOnIdle {
-				host.ImmediateShutdown("generic-worker idle timeout")
+				// If running with worker-runner, send shutdown message so it can
+				// unregister the worker before shutting down. Otherwise, shut down directly.
+				if WorkerRunnerProtocol.Capable("shutdown") {
```
matt-boris, line 199: I think we need to `AddCapability("shutdown")` to this function https://github.com/taskcluster/taskcluster/blob/main/workers/generic-worker/workerrunner.go#L81

Nice find on this `shutdown` message capability too 👍🏻 Thanks

Eijebong: Arf, good find!

PR #8166, opened 2025-12-15
fix(generic-worker): shutdown with worker runner in all shutdown cases

matt-boris:
>Generic Worker (with worker-runner): properly unregisters a worker when it exits due to internal error or non-current deployment ID. Followup to #8165.


petemoore (approved):
This looks awesome! Have we tested that it works in a real deployment, i.e. that Worker Runner implements the shutdown properly? I can see it is difficult to add tests here for a full integration test, but I guess we should try it manually, just to be sure it really works?

PR #8168, opened 2025-12-16
feat(w-m): Stopping capacity is no longer used in estimations

### LINKED ISSUE #8161: Azure provider stopping capacity
Since we started using ARM templates, workers are now being fully provisioned with one call in the provisioner. 
This no longer puts extra load on worker scanner azure that needs to deprovision workers in the same loop as new workers being created. 

Back in https://github.com/taskcluster/taskcluster/pull/6103 we added `stopping` capacity into estimator as a limitation to how many stopping+requested we should handle. 
We probably no longer need to count them as "running" since they will not be blocking the provisioning. 

However, it might still be a subject to some quota limits.

lotas:
Since ARM is deploying every resource on the Azure side we are no longer blocked by having to both provision and deprovision resources in the same loop.
Another fail-safe mechanism was added in launch configs that would lower LC weight if it would have increased rate of errors.

Fixes #8161

### DIFF
--- a/changelog/issue-8161.md
+++ b/changelog/issue-8161.md
@@ -1,0 +1,8 @@
+audience: worker-deployers
+level: minor
+reference: issue 8161
+---
+
+Estimator no longer includes `stoppingCapacity` to calculate how many workers needs to be created.
+This is no longer necessary since Azure started using ARM templates for deployment, which are no longer blocked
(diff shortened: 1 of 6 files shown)


lotas:
Checking if launch configs provide the fail-safe mechanisms that stoppingCapacity was doing before it was implemented

PR #8177, opened 2025-12-29
feat(w-m): Ensure workers are being launched for some edge cases

### LINKED ISSUE #8176: Worker manager launch config selector issues with single LC
It was reported that for some worker pools that have single launch config worker manager might not provision any workers at all.
This is likely happening because weight is being adjusted (based on number of errors and currently running capacity), which means that single LC can have weight less than `1`, which might not be selected most of the times

Happens a lot to https://community-tc.services.mozilla.com/worker-manager/proj-fuzzing%2Fwindows-pool110 and https://community-tc.services.mozilla.com/worker-manager/proj-fuzzing%2Fwindows-pool81

TODO: 
- check if this is a single-lc edge case, or something else. 
- make sure LC is always returned if global wp capacity limits are not met yet

lotas:
When worker pool has only one launch config and it contain some errors plus capacity is non zero, dynamic weight might go below zero, which would leave zero available launch configs even though capacity is not yet reached. We check if that's the only available config with remaining capacity and adjust the weight to let it be selected

Github Bug/Issue: Fixes #8176

### DIFF
--- a/changelog/issue-8176.md
+++ b/changelog/issue-8176.md
@@ -1,0 +1,6 @@
+audience: worker-deployers
+level: patch
+reference: issue 8176
+---
+
+Worker-manager ensures workers are spawned for single launch config worker pools when adjusted weight might drop below zero with remaining capacity.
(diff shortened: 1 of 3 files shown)


changelog/issue-8176.md
```diff
@@ -0,0 +1,6 @@
+audience: worker-deployers
+level: minor
```
matt-boris:
```suggestion
level: patch
```

I think it's fine just at patch level

petemoore (approved):

matt-boris (approved):

PR #8178, opened 2025-12-30
feat(web): Web-server subscriptions using classic queues

lotas:
Quorum queues are durable and provide high availability at a higher memory usage cost, and for this particular UI updates use-case we don't need to use it. Classic should be enough


petemoore (approved):
LGTM, thanks Yarik!

PR #8193, opened 2026-01-12
feat(w-m): Properly detect zombie worker using queue information

lotas:
In some cases worker entity was reloaded from db without queue information, which led to isZombie() check incorrectly considering such worker as not doing anything and getting killed.

Bug was introduced when we deployed those arm template refactorings, it shuffled db calls a bit in code, which resulted in queue fields being wiped out, and zombie decision was made without those --> always killing after `queueInactivityTimeout`

Fixes bug 2006698

### DIFF
--- a/changelog/bug-2006698.md
+++ b/changelog/bug-2006698.md
@@ -1,0 +1,8 @@
+audience: worker-deployers
+level: patch
+reference: bug 2006698
+---
+
+Fixes worker-manager killing zombie workers when queue data is missing.
+For long running tasks that took longer than queueInactivityTimeout to finish was a high risk of getting claim-expired
+because w-m incorrectly assumed worker is not doing anything (missing queue information)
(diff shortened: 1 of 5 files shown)


db/test/fns/worker_manager_test.js
```diff
@@ -875,6 +875,11 @@ suite(testing.suiteName(), function () {
       const rows = await db.fns.get_worker_3('wp/id', 'w/group', 'w/id');
       assert.equal(rows[0].provider_id, 'provider2');
       assert.equal(rows[0].state, 'requested');
+      // remove queue fields
+      delete rows[0].first_claim;
+      delete rows[0].last_date_active;
+      delete rows[0].quarantine_until;
+      delete rows[0].recent_tasks;
```
matt-boris: Is this still needed?

PR #8194, opened 2026-01-13
Don't panic if we can't remove d2g-image-cache.json and it's not there

Eijebong:
This can happen if you start the worker with low disk space and that file is already missing or in my case if the previous garbage collection put you just below the threshold for the worker to decide to system prune, and then an unrelated event puts you back above with no task running in between.


matt-boris (approved):
Thank you! This LGTM, thanks for always help find our little edge cases 😄

PR #8201, opened 2026-01-14
fix(ci): test goreleaser in CI

matt-boris:


matt-boris:
The first commit purposely doesn't pull in the latest goreleaser update so that it shows how the CI will fail for us in the future. Second commit brings in the update to the tool and should pass successfully.

PR #8212, opened 2026-01-20
fix(tc-cli): `task log` command falls back to use backing log

matt-boris:
>Client (shell): `taskcluster task log` command falls back to use `live_backing.log` if `live.log` is unavailable.

cc @ahal

### DIFF
--- a/clients/client-shell/cmds/task/fetchers.go
+++ b/clients/client-shell/cmds/task/fetchers.go
@@ -212,11 +212,26 @@ func runLog(credentials *tcclient.Credentials, args []string, out io.Writer, fla
 		return fmt.Errorf("could not fetch the logs of task %s because it's in a %s state", taskID, state)
 	}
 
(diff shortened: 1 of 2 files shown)


clients/client-shell/cmds/task/fetchers.go
```diff
@@ -212,23 +212,38 @@ func runLog(credentials *tcclient.Credentials, args []string, out io.Writer, fla
...
+			continue
+		}
+		break
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode/100 != 2 {
+		return fmt.Errorf("received unexpected response code %v", resp.StatusCode)
+	}
+
```
petemoore: I think this code block should stay in its original position, so that if you get a non-2xx response code, you can see what the response body contained. It may contain relevant information for the user.

matt-boris: Fixed in my latest push.

petemoore (approved):
Great job, thanks Boris!

PR #8213, opened 2026-01-21
feat(generic-worker): add `hideCmdWindow` feature to windows payload

### LINKED ISSUE #8012: Generic Worker newCommand on Windows displaying persistent cmd.exe window
**Describe the bug**
When Generic Worker runs a task on Windows using `command_000000.bat` there is a persistent `cmd.exe` window in the background of each task user that runs. This was identified on windows performance hardware where `cmd.exe` window was in the foreground of the screen consuming screen real estate covering UI tests in Firefox.

My attempts at discovering this points to something having to do with [CREATE_NEW_CONSOLE](https://github.com/taskcluster/taskcluster/blob/main/workers/generic-worker/process/multiuser_windows.go#L90) and how we may need to add [new constants ](https://github.com/taskcluster/taskcluster/blob/main/workers/generic-worker/win32/win32_windows.go#L70) such as `CREATE_NO_WINDOW` that would suppress this window. 

**To Reproduce**

Generic worker config:

```
{
  "accessToken": "foobar",
  "cachesDir": "C:\\cache",
  "cleanUpTaskDirs": true,
  "clientId": "project/releng/generic-worker/win11-64-24h2-hw",
  "disableOOMProtection": true,
  "disableReboots": false,
  "downloadsDir": "C:\\downloads",
  "ed25519SigningKeyLocation": "C:\\generic-worker\\ed25519-private.key",
  "idleTimeoutSecs": 0,
  "livelogExecutable": "C:\\generic-worker\\livelog.exe",
  "numberOfTasksToRun": 9999999,
  "provisionerId": "releng-hardware",
  "publicIP": "10.49.65.101",
  "requiredDiskSpaceMegabytes": 10240,
  "rootURL": "https://firefox-ci-tc.services.mozilla.com",
  "runAfterUserCreation": "C:\\generic-worker\\task-user-init.cmd",
  "sentryProject": "generic-worker",
  "shutdownMachineOnIdle": false,
  "shutdownMachineOnInternalError": false,
  "taskclusterProxyExecutable": "C:\\generic-worker\\taskcluster-proxy.exe",
  "taskclusterProxyPort": 8080,
  "tasksDir": "C:\\",
  "workerGroup": "mdc1",
  "workerId": "nuc13-100",
  "workerLocation": "{\"cloud\":\"standalone\",\"host\":\"datacenter\"}",
  "workerType": "win11-64-24h2-hw",
  "workerTypeMetadata": {
    "machine-setup": {
      "config": "https://github.com/mozilla-platform-ops/ronin_puppet/tree/cloud_windows"
    }
  },
  "wstAudience": "firefoxcitc",
  "wstServerURL": "https://firefoxci-websocktunnel.services.mozilla.com"
}
```

Execute a command that launches a background screen on windows and observe a lingering `cmd.exe` window either in the background or foreground of the task user running.

**Expected behavior**
The `cmd.exe` window should be hidden or suppressed and not impacting UI tasks.

**Taskcluster version**
88.1.3

matt-boris:
Fixes #8012.

>Generic Worker (windows): adds `task.payload.features.hideCmdWindow` [default: `false`] to hide the `cmd.exe` window that appears during task execution. This may be useful if the `cmd.exe` window gets in the way of GUI applications while running tasks. Please note: if your task needs to allocate new consoles (with `AllocConsole()`, for example), it will not be able to if you set this to `true`.

### DIFF
--- a/workers/generic-worker/multiuser_windows_test.go
+++ b/workers/generic-worker/multiuser_windows_test.go
@@ -95,6 +95,62 @@ func TestNoCreateFileMappingError(t *testing.T) {
 	_ = submitAndAssert(t, td, payload, "completed", "completed")
 }
 
+// TestHideCmdWindowEnabled verifies that when hideCmdWindow feature is enabled,
+// the spawned process does not have a console window attached.
+// This uses the GetConsoleWindow() Win32 API which returns NULL when no console
+// is attached (i.e., when CREATE_NO_WINDOW flag is used).
+func TestHideCmdWindowEnabled(t *testing.T) {
+	setup(t)
+
+	payload := GenericWorkerPayload{
+		Command:    goRun("check-console-window.go", "true"),
+		MaxRunTime: 30,
+		Features: FeatureFlags{
+			HideCmdWindow: true,
+		},
+	}
+	defaults.SetDefaults(&payload)
+	td := testTask(t)
+
+	_ = submitAndAssert(t, td, payload, "completed", "completed")
+}
+
+// TestHideCmdWindowDisabled verifies that when hideCmdWindow feature is disabled
+// (the default), the spawned process has a console window attached.
+// This uses the GetConsoleWindow() Win32 API which returns a non-NULL handle
+// when a console is attached (i.e., when CREATE_NEW_CONSOLE flag is used).
+func TestHideCmdWindowDisabled(t *testing.T) {
+	setup(t)
+
+	payload := GenericWorkerPayload{
+		Command:    goRun("check-console-window.go", "false"),
+		MaxRunTime: 30,
+		Features: FeatureFlags{
+			HideCmdWindow: false,
+		},
+	}
+	defaults.SetDefaults(&payload)
+	td := testTask(t)
+
+	_ = submitAndAssert(t, td, payload, "completed", "completed")
+}
+
+// TestHideCmdWindowDefault verifies that when hideCmdWindow feature is not
+// explicitly set, the default behavior is to have a console window attached.
+func TestHideCmdWindowDefault(t *testing.T) {
+	setup(t)
+
+	payload := GenericWorkerPayload{
+		Command:    goRun("check-console-window.go", "false"),
+		MaxRunTime: 30,
+		// Features.HideCmdWindow not set - should default to false
+	}
+	defaults.SetDefaults(&payload)
+	td := testTask(t)
+
+	_ = submitAndAssert(t, td, payload, "completed", "completed")
+}
+
 func TestDesktopResizeAndMovePointer(t *testing.T) {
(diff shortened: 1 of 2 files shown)


workers/generic-worker/schemas/multiuser_windows.yml
```diff
@@ -315,7 +315,20 @@ properties:
...
+        title: Hide cmd.exe window
+        description: |-
+          If `true`, the command window for each command will be hidden.
+          This is useful for tasks that run GUI applications to prevent
+          the command window from obscuring the application window.
+
+          If your task needs to allocate new consoles (with
+          AllocConsole(), for example) then you should not use this.
+
+          Since: generic-worker 96.1.0
```
petemoore, line 335: I think it would be good to highlight the particular process flags that are set, i.e. that when it is false, you get CREATE_NEW_CONSOLE|CREATE_NEW_PROCESS_GROUP, and when it is true, you get CREATE_NO_WINDOW|CREATE_NEW_PROCESS_GROUP and ideally with a hyperlink to e.g. https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags to explain what that means.

petemoore: I also wonder if we should add one test that makes sure that when the process is launched, that no window is created when false, but is created when true, e.g. if there is some way in win32 to iterate over the windows associated to a process etc...

matt-boris: Both of these have been addressed in followup commits!

workers/generic-worker/schemas/multiuser_windows.yml
```diff
@@ -315,7 +315,26 @@ properties:
...
+        description: |-
+          If `true`, the command window for each command will be hidden.
+          This is useful for tasks that run GUI applications to prevent
+          the command window from obscuring the application window.
+
+          The process creation flags used when this is set to `true`
+          are CREATE_NEW_PROCESS_GROUP|CREATE_NO_WINDOW. If this is
+          set to `false`, the flags used are CREATE_NEW_PROCESS_GROUP|CREATE_NEW_CONSOLE.
+          More info about these flags can be found here:
+          https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags
```
petemoore: just a small point - these docs are markdown, so you can write e.g. `[here](https://.....)`

matt-boris: Fixed here https://github.com/taskcluster/taskcluster/pull/8213/changes/8e65449dabe3bb121ac930e325bce0131f9dd198

petemoore (approved):
Awesome, thanks!

PR #8226, opened 2026-01-26
fix(k8s): specify security context settings

matt-boris:
>Adds `securityContext.runAsNonRoot: true` and `securityContext.allowPrivilegeEscalation: false` to all k8s Deployments and CronJobs. Containers are all now run as non-root, `node` user (UID/GID 1000).

### DIFF
--- a/infrastructure/references/references.sh
+++ b/infrastructure/references/references.sh
@@ -2,9 +2,9 @@
 
 set -e
 
-# relativize the schemas and references to a writable directory
+# relativize the schemas and references
 cd /app
-node infrastructure/references/relativize.js generated/references.json /tmp/references
+node infrastructure/references/relativize.js generated/references.json /references
 
 # start nginx with standalone config
 exec nginx -c /app/infrastructure/references/nginx.conf -g 'daemon off;'
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,6 +68,9 @@ COPY --from=build --chown=1000:1000 /base/app /app
 ENV HOME=/app
 WORKDIR /app
 
+# Create /references directory for references service with proper ownership
+RUN mkdir -p /references && chown 1000:1000 /references
+
 # use non-root, node user
 # https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user
 USER 1000
--- a/infrastructure/references/nginx.conf
+++ b/infrastructure/references/nginx.conf
@@ -48,7 +48,7 @@ http {
         listen 80;
         server_name localhost;
         server_name_in_redirect off;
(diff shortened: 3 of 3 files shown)


infrastructure/tooling/templates/k8s/deployment.yaml
```diff
@@ -27,11 +27,15 @@ spec:
...
       terminationGracePeriodSeconds: 210
+      securityContext:
+        runAsNonRoot: true
       containers:
       - name: ${projectName}-${lowercase(procName)}
         image: '{{ .Values.dockerImage }}'
         imagePullPolicy: Always
         args: ['${serviceName}/${procName}']
+        securityContext:
+          allowPrivilegeEscalation: false
```
petemoore, line 38: killjoy

infrastructure/k8s/templates/taskcluster-auth-cron-purgeAuditHistory.yaml
```diff
@@ -21,12 +21,16 @@ spec:
...
           imagePullSecrets: {{ if .Values.imagePullSecret }}{{ toJson (list (dict "name" .Values.imagePullSecret)) }}{{ else }}[]{{ end }}
+          securityContext:
+            runAsNonRoot: true
           containers:
             - name: taskcluster-auth-purgeaudithistory
               image: '{{ .Values.dockerImage }}'
               imagePullPolicy: Always
               args:
                 - auth/purgeAuditHistory
+              securityContext:
```
lotas, line 32: I wonder if we really need to specify this for every container, looks like spec-level securityContext might be enough:
https://kubernetes.io/docs/concepts/security/pod-security-standards/

> The container fields may be undefined/nil if the pod-level spec.securityContext.runAsNonRoot is set to true. 

if I read this correctly

lotas: ah sorry, those are different keys

infrastructure/references/references.sh
```diff
@@ -2,10 +2,9 @@
 
 set -e
 
-# relativize the schemas and references
+# relativize the schemas and references to a writable directory
 cd /app
-node infrastructure/references/relativize.js generated/references.json /references
+node infrastructure/references/relativize.js generated/references.json /tmp/references
```
lotas: why couldn't we just create `/references` writable/readable by `1000` user?

matt-boris: Addressed in https://github.com/taskcluster/taskcluster/pull/8226/changes/e24d0b5a0149c1963d37c6a87d55477eefc3e5a5!

PR #8228, opened 2026-01-26
github: add the task id to the task status header

jcristau:
It's occasionally useful to be able to copy/paste the taskId without having to fish it out of links.


petemoore:
wfm too, thanks!

PR #8248, opened 2026-02-05
feat(w-m): Provider workerGroup label for some metric

### LINKED ISSUE #8247: Add location/region/zone labels to exposed worker-manager metrics
Currently we only add providerId/workerPoolId labels to the exposed metrics. 
It might be helpful for worker operators to have breakdown by worker region. 

But we'd need to refactor codebase a bit to calculate properly per location

lotas:
For worker deployers it might be useful to see breakdown by region/location of the workers to track costs more effectively.

Fixes #8247

### DIFF
--- a/services/worker-manager/src/data.js
+++ b/services/worker-manager/src/data.js
@@ -240,6 +240,8 @@ export class WorkerPoolStats {
     this.totalErrors = 0;
     this.capacityByLaunchConfig = new Map();
     this.errorsByLaunchConfig = new Map();
+    /** @type {Map<string, { existingCapacity: number, requestedCapacity: number, stoppingCapacity: number }>} */
+    this.capacityByWorkerGroup = new Map();
   }
 
   forProvision() {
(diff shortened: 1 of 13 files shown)


services/worker-manager/src/data.js
```diff
@@ -286,26 +296,36 @@ export class WorkerPoolStats {
...
 
-    if (isStopping) {
-      this.stoppingCapacity += worker.capacity;
-    } else {
-      const requestedCapacity = isRequested ? worker.capacity : 0;
-      const existingCapacity = isQuarantined ? 0 : worker.capacity;
+    const stoppingCapacity = isStopping ? worker.capacity : 0;
+    const requestedCapacity = isRequested ? worker.capacity : 0;
+    const existingCapacity = isStopping || isQuarantined ? 0 : worker.capacity;
+    const quarantineCapacity = isQuarantined ? worker.capacity : 0;
```
matt-boris, line 302: Looks like you fixed a bug here in this refactoring/clean up!

Previously, `existingCapacity` would be set to 0 then that value was added to `quarantineCapacity`, not properly counting the `quarantineCapacity` :) Nice!

matt-boris (approved):
LGTM, thank you!

services/worker-manager/src/data.js
lotas: indeed! :)

PR #8252, opened 2026-02-05
fix(references): empty `/references` instead of removing

### LINKED ISSUE #8251: `taskcluster-references-web` permission denied error removing `/references`
Issue due to https://github.com/taskcluster/taskcluster/pull/8226 running pods as non-root, `node` user now.

matt-boris:
Followup to #8226.

Fixes #8251.

>Fixes deployment issue where `taskcluster-references-web` did not have access to remove the `/references` directory.


libraries/references/src/uri-structured.js
```diff
@@ -4,7 +4,13 @@ import path from 'path';
...
 
 export const writeUriStructured = async ({ directory, serializable }) => {
-  await rimraf(directory);
+  // Delete contents of directory rather than the directory itself
+  // This is important when running as non-root user and the directory
+  // is at the filesystem root (e.g., /references)
+  await rimraf(path.join(directory, '*'), { glob: true });
+
+  // Ensure the directory exists
+  await mkdirp(directory);
```
lotas: This is probably redundant since dirs are created individually below, plus command above doesn't remove it

matt-boris: True, just removed. Thanks

PR #8255, opened 2026-02-06
build(deps): upgrade uv and taskgraph decision image

matt-boris:
>Bumps `uv` to v0.10.0 and taskgraph decision task image to v18.1.0.


lotas (approved):
Thanks!

PR #8260, opened 2026-02-07
fix(ui): bypass GraphQL for createTask, call Queue API directly (fixes #8171)

### LINKED ISSUE #8171: Creating a task via UI fails (payload too large) while python Queue API succeeds with same payload
**Describe the bug**
Because the taskcluster UI for "Edit" (and Create) task uses graphql to submit the task, we are sometimes hitting the payload limit for graphql.  Though the payload is legitimately limited, since large payloads in graphql could be indicative of an attempted attack, the fact that the direct create task API works fine suggests we have a bug to fix here.

**To Reproduce**

1. Go to 'community tc'
2. Create a new task via the UI (see below)
4. See error in UI

<details><summary>Task Payload for reproduction</summary>
<p>

Use https://loremipsum.io/generator/?n=500&t=p to generate a 500 paragraph lorem ipsum, and then base64 encode it and paste it into the `extra.large_lorem_ipsum_b64` field.

<pre>
taskQueueId: proj-misc/tutorial
schedulerId: taskcluster-ui
created: '2025-12-17T19:56:07.123Z'
deadline: '2025-12-17T22:56:07.123Z'
payload:
  image: ubuntu:latest
  command:
    - /bin/bash
    - '-c'
    - for ((i=1;i<=60;i++)); do echo $i; sleep 1; done
  maxRunTime: 90
metadata:
  name: example-task
  owner: name@example.com
  source: https://community-tc.services.mozilla.com/tasks/create
extra:
  large_lorem_ipsum_b64: <LARGE_IPSUM_ENCODED>
</pre>

</p>
</details> 
<summary>

**Expected behavior**
(Note I didn't test this EXACT payload via RestAPI, since I arbitrarily created this lorem ipsum b64 to demonstrate the payload error, I'm unable to publicly share the internal use task that I experienced this on)

**Taskcluster version**
`v95.0.0`

nitishagar:
## Summary

- The UI's `createTask` calls (loaner, retrigger, submitTaskAction, CreateTask page) now call the Queue REST API directly via `@taskcluster/client-web` instead of routing through GraphQL
- This fixes 413 errors when creating tasks with payloads larger than 100KB (the default `bodyParser.json` limit) — up to the Queue API's 10MB limit
- Also fixes a latent bug where `authorizedScopes` passed via GraphQL `options` were silently dropped by the resolver; the Queue client now receives them correctly via `getClient({ authorizedScopes: ... })`
- Reverts the server-side body-size limit increases from the previous approach (no changes to `createApp.js` vs `main`)
- Deletes the now-unused `createTask.graphql` mutation file

## Approach

Per architect review by @lotas in #8171: skipping the GraphQL layer for task creation avoids widening the Apollo attack surface (increased body limits have historically been a source of security issues in Apollo).

The UI already uses `@taskcluster/client-web` for other endpoints (e.g. `WorkerManager.removeWorker`). Task creation follows the same `getClient()` pattern.

## Test plan

- [ ] `cd ui && yarn test` — all 52 suites / 137 tests pass (including new `submitTaskAction.test.js`)
- [ ] `cd ui && yarn lint` — no errors
- [ ] `cd services/web-server && yarn test` — `validation_test.js` passes with original 413 + `PayloadTooLargeError` assertions
- [ ] Manually verify task creation with a large payload (>100KB) succeeds in the UI

GitHub Bug/Issue: Fixes #8171

### DIFF
--- a/ui/src/views/Tasks/TaskGroup/index.jsx
+++ b/ui/src/views/Tasks/TaskGroup/index.jsx
@@ -50,6 +50,7 @@ import CopyToClipboardListItem from '../../../components/CopyToClipboardListItem
 import DateDistance from '../../../components/DateDistance';
 import sealTaskGroupQuery from './sealTaskGroup.graphql';
 import cancelTaskGroupQuery from './cancelTaskGroup.graphql';
+import { AuthContext } from '../../../utils/Auth';
 
 const initialTaskGroupActions = [
   {
@@ -155,6 +156,8 @@ const updateTaskGroupIdHistory = id => {
   },
 }))
 export default class TaskGroup extends Component {
+  static contextType = AuthContext;
+
   static calculateStatusCountStatic(taskGroup) {
     const statusCount = {
       completed: 0,
@@ -620,6 +623,7 @@ export default class TaskGroup extends Component {
         form,
         action,
         apolloClient,
+        user: this.context.user,
       });
 
       return taskId;
--- a/ui/src/views/Tasks/ViewTask/index.jsx
+++ b/ui/src/views/Tasks/ViewTask/index.jsx
@@ -12,6 +12,7 @@ import Checkbox from '@material-ui/core/Checkbox';
 import dotProp from 'dot-prop-immutable';
 import jsonSchemaDefaults from 'json-schema-defaults';
 import { dump } from 'js-yaml';
+import { Queue } from '@taskcluster/client-web';
 import HammerIcon from 'mdi-react/HammerIcon';
 import CreationIcon from 'mdi-react/CreationIcon';
 import PencilIcon from 'mdi-react/PencilIcon';
@@ -52,6 +53,8 @@ import removeKeys from '../../../utils/removeKeys';
 import parameterizeTask from '../../../utils/parameterizeTask';
 import { nice } from '../../../utils/slugid';
 import Link from '../../../utils/Link';
+import { getClient } from '../../../utils/client';
+import { AuthContext } from '../../../utils/Auth';
 import submitTaskAction from '../submitTaskAction';
 import taskQuery from './task.graphql';
 import taskSubscription from './taskSubscription.graphql';
@@ -60,8 +63,6 @@ import rerunTaskQuery from './rerunTask.graphql';
 import cancelTaskQuery from './cancelTask.graphql';
 import purgeWorkerCacheQuery from './purgeWorkerCache.graphql';
 import pageArtifactsQuery from './pageArtifacts.graphql';
-import createTaskQuery from '../createTask.graphql';
-import { AuthContext } from '../../../utils/Auth';
 
 const updateTaskIdHistory = id => {
   if (!VALID_TASK.test(id)) {
@@ -116,6 +117,8 @@ const getCachesFromTask = task =>
   }),
 })
 export default class ViewTask extends Component {
+  static contextType = AuthContext;
+
   static getDerivedStateFromProps(props, state) {
     const taskId = props.match.params.taskId || '';
     const {
@@ -295,6 +298,7 @@ export default class ViewTask extends Component {
         form: formInputs,
         action,
         apolloClient: client,
+        user: this.context.user,
       });
 
       return taskId;
@@ -473,13 +477,9 @@ export default class ViewTask extends Component {
     this.preRunningAction();
 
     try {
-      await this.props.client.mutate({
-        mutation: createTaskQuery,
-        variables: {
-          taskId,
-          task,
-        },
-      });
+      const queue = getClient({ Class: Queue, user: this.context.user });
+
+      await queue.createTask(taskId, task);
 
       return taskId;
     } catch (error) {
@@ -789,13 +789,9 @@ export default class ViewTask extends Component {
     this.preRunningAction();
 
     try {
-      await this.props.client.mutate({
-        mutation: createTaskQuery,
-        variables: {
-          taskId,
-          task,
-        },
-      });
+      const queue = getClient({ Class: Queue, user: this.context.user });
+
+      await queue.createTask(taskId, task);
 
       return taskId;
     } catch (error) {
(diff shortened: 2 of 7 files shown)


services/web-server/test/graphql/validation_test.js
```diff
@@ -21,12 +21,11 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
             }
           `,
         });
+        assert.fail('Expected query to fail');
```
lotas: I don't think this is the right test to change. It was created to test something different . Please keep `PayloadTooLargeError` test alive (probably by increasing repeat counter. 
And I think that extra test won't be needed, 400 is tested below already

services/web-server/src/servers/createApp.js
lotas: I think this might no longer be applied because we limit bodyParser at 10mb now

services/web-server/test/graphql/validation_test.js
```diff
@@ -11,20 +12,22 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
 
   suite('GraphQL Validation', function() {
     test('max tokens in request', async function() {
-      const client = helper.getHttpClient({ suppressErrors: true });
+      // Send a payload exceeding the 10mb body-parser limit.
+      // Uses got directly to avoid gql parsing overhead for large payloads.
+      const response = await got.post(
```
lotas: Do you know what was the issue with gql? and why we want to avoid this? If you keep original code and just change the value what is happening then?

services/web-server/test/graphql/validation_test.js
```diff
@@ -11,20 +12,22 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
+          throwHttpErrors: false,
+          responseType: 'json',
+        },
+      );
 
-      try {
-        await client.query({
-          query: gql`
-            query {
-              ${' a '.repeat(100000)}
```
lotas, line 20: Oh, I just realized that test was testing something completely different - https://github.com/taskcluster/taskcluster/blob/c6c55c5e44ac7dc7a650c52a8c2cdf60cb968946/services/web-server/src/main.js#L158

max tokens is limited to 100000 here, so this test shouldn't be changed

services/web-server/test/graphql/validation_test.js
```diff
@@ -11,20 +12,22 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
...
     test('max tokens in request', async function() {
-      const client = helper.getHttpClient({ suppressErrors: true });
+      // Send a payload exceeding the 10mb body-parser limit.
+      // Uses got directly to avoid gql parsing overhead for large payloads.
+      const response = await got.post(
+        `http://localhost:${helper.serverPort}/graphql`,
+        {
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({
+            query: `query { ${'a'.repeat(11 * 1024 * 1024)} }`,
```
lotas: Please note that original test contained spaces around ' a ' for a reason - that test was testing 'maxTokens'

services/web-server/test/graphql/validation_test.js
nagarwal-godaddy: The `got` switch was to avoid gql parsing a >10MB string client-side, but that's only needed if we're testing the body parser limit directly. Since we're keeping the original 100k-repeat test (~300KB — well within gql's comfort zone), I've reverted back to using gql + helper client. No issues.

ui/src/views/Tasks/submitTaskAction.js
```diff
@@ -3,15 +3,23 @@ import { satisfiesExpression } from 'taskcluster-lib-scopes';
...
 import ajv from '../../utils/ajv';
+import { getClient } from '../../utils/client';
 
-export default async ({ task, form, action, apolloClient, taskActions }) => {
+export default async ({
+  task,
+  form,
+  action,
+  apolloClient,
+  user,
```
lotas, line 20: Please check https://github.com/taskcluster/taskcluster/blob/25c072b6fde83080794c017947301af0b98237fc/ui/src/views/Tasks/TaskGroup/index.jsx#L629-L635 

TaskGroup is still calling this without a user context

nitishagar:
Thanks @lotas, all addressed:

- **Validation test** — fully reverted. Original `' a '.repeat(100000)` test is back as-is, testing maxTokens as intended. No extra test added.
- **bodyParser limit** — removed the change since the 10mb limit already covers this.
- **TaskGroup context** — good catch. `TaskGroup/index.jsx` now pulls user from `AuthContext` and passes it through to `submitTaskAction`, so the Queue client gets proper credentials.

Also added unit tests for `submitTaskAction` and rebased onto latest main.

nitishagar:
Gentle ping @lotas — all feedback was addressed on 2026-04-09 (TaskGroup context threading, bodyParser revert, validation-test revert, plus unit tests for `submitTaskAction`). Happy to answer anything if more changes are wanted.

ui/src/views/Tasks/TaskGroup/index.jsx
lotas, line 1: Why do we need to modfiy this file?

ui/src/views/Tasks/ViewTask/index.jsx
```diff
@@ -127,6 +128,8 @@ const getCachesFromTask = task =>
   }),
 })
 export default class ViewTask extends Component {
+  static contextType = AuthContext;
```
lotas, line 120: This file also has `<AuthContext.Consumer>` https://github.com/taskcluster/taskcluster/blob/c7b63836684775ee88c5c53c2706cf2a476faa8a/ui/src/views/Tasks/ViewTask/index.jsx#L963-L972 

Since AuthContext is being injected this way, we can drop  it from the `<TaskDetailsCard>` wrapper and just use `this.context.user`

lotas:
> @lotas Updated based on the feedback. Please take a look when you get a chance.

sorry for the delayed response, I think it looks much better, thank you. Just one small refactor ask to use newly injected context user. Thanks!

lotas:
@nitishagar can you please rebase your branch, there are some conflicts 🙏  Thank you

nitishagar:
@lotas Took me a lot more time to get back 😥 
Have rebased changes.

lotas (dismissed):
I checked it locally, it works fine, thanks
Also rebased the branch since there were lots of changes

PR #8262, opened 2026-02-07
fix(ui): show warning when live updates disabled due to missing web:read-pulse scope

### LINKED ISSUE #8070: Add an indication when the graphql websocket gets closed due to the lack of `web:read-pulse` scope
When a user (or anonymous) doesn't have that scope, there's currently no indication whatsoever that the information on the page is out of date (on a task group for example, the running time for tasks continues to count up despite the tasks being done already since their status doesn't get refreshed) which is very confusing for people like me that don't remember that they don't have that scope unless logged in.

The only place where you can find that this is wrong is by looking at the websocket traffic in the network inspector. It would be nice to have some kind of indication on the page telling the user that they might not be receiving updates.

### LINKED ISSUE #8261: webpack-dev-server crashes with ECONNRESET when proxying WebSocket /subscription endpoint
When running the UI locally with `yarn start` (webpack-dev-server) pointed at a remote Taskcluster deployment (e.g., community-tc), the dev server crashes shortly after compilation with an unhandled `ECONNRESET` error:

```
node:events:486
      throw er; // Unhandled 'error' event
      ^

Error: read ECONNRESET
    at TCP.onStreamRead (node:internal/stream_base_commons:216:20)
Emitted 'error' event on Socket instance at:
    at emitErrorNT (node:internal/streams/destroy:170:8)
    at emitErrorCloseNT (node:internal/streams/destroy:129:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
```

### Cause

The `/subscription` WebSocket proxy in `ui/webpack.config.js:91-95` has no error handlers:

```javascript
"/subscription": {
  ws: true,
  changeOrigin: true,
  target: proxyTarget.replace(/^http(s)?:/, "ws$1:"),
},
```

When the target rejects the WebSocket connection (e.g., community-tc returning `InsufficientScopes` for missing `web:read-pulse` scope), the TCP connection resets. The underlying `http-proxy` (v1.18.1, via `http-proxy-middleware` v0.19.1) emits an `'error'` event on the socket with no listener attached, which crashes the Node.js process.

This is a [known issue](https://github.com/webpack/webpack-dev-server/issues/1642) with webpack-dev-server 3.x on Node.js 10+.

### Fix

Add `onError` and `onProxyReqWs` handlers to the `/subscription` proxy config:

```javascript
"/subscription": {
  ws: true,
  changeOrigin: true,
  target: proxyTarget.replace(/^http(s)?:/, "ws$1:"),
  onError: function (err, req, res) {
    console.warn('[WS Proxy Error]', err.code, err.message);
  },
  onProxyReqWs: function (proxyReq, req, socket) {
    socket.on('error', function (err) {
      console.warn('[WS Socket Error]', err.code, err.message);
    });
  },
},
```

### Notes

- This only affects local development (`yarn start` / webpack-dev-server). Production uses nginx which handles connection resets gracefully.
- The docker-compose dev mode is also unaffected since it proxies through nginx internally.
- Reproducible on macOS (arm64) with Node.js v24.x by running `TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com yarn start` in `ui/`.
- Related to #8070 -- the same missing-scope scenario triggers both the silent UI failure and this dev server crash.

nitishagar:
## Summary

- Add a visible warning banner when the GraphQL WebSocket connection is closed due to missing `web:read-pulse` scope
- Track subscription connection errors in the `App` component and propagate to `Main` for display
- Add server-side test verifying WebSocket rejection when scope is missing

Fixes #8070
Fixes #8261

## Problem

When a user (or anonymous visitor) lacks the `web:read-pulse` scope, the GraphQL WebSocket connection is silently closed. The UI continues displaying stale data with no indication — task timers count up indefinitely, task states never refresh. The only way to discover this is by inspecting WebSocket traffic in browser devtools.

## Solution

**`ui/src/App/index.jsx`**: In the existing `connectionCallback` that already catches `InsufficientScopes` errors, added `this.setState({ subscriptionError: error })` to surface the error to the UI. Passes `subscriptionError` as a prop to `Main`.

**`ui/src/App/Main.jsx`**: Renders a persistent warning `Snackbar` (amber/warning variant) when `subscriptionError` is present. The banner is dismissible per session and resets on page refresh or user change (since `Main` remounts via its `key` prop).

**`services/web-server/test/graphql/tasks_graphql_test.js`**: Added test that verifies the WebSocket connection is properly rejected with `InsufficientScopes` when the `web:read-pulse` scope is missing. Confirms the server contract the UI fix depends on.

## Test plan

- [ ] Open a TaskGroup page without logging in — warning banner appears
- [ ] Banner message: "Live updates are not available because you are missing the web:read-pulse scope. Task information on this page may be out of date. Try signing in or refreshing the page."
- [ ] Clicking X dismisses the banner for the session
- [ ] Refreshing the page shows the banner again
- [ ] Logging in with `web:read-pulse` scope and refreshing — no banner
- [ ] Pages without subscriptions (e.g., `/auth/roles`) do not show the banner
- [ ] Existing web-server subscription tests still pass

### DIFF
--- a/changelog/issue-8261.md
+++ b/changelog/issue-8261.md
@@ -1,0 +1,5 @@
+audience: developers
+level: patch
+reference: issue 8261
+---
+Fix webpack-dev-server crash (ECONNRESET) when proxying WebSocket `/subscription` endpoint during local UI development.
--- a/ui/webpack.config.js
+++ b/ui/webpack.config.js
@@ -92,6 +92,14 @@ module.exports = (_, { mode }) => ({
         ws: true,
         changeOrigin: true,
         target: proxyTarget.replace(/^http(s)?:/, "ws$1:"),
+        onError: function(err, req, res) {
+          console.warn("[WS Proxy Error]", err.code, err.message);
+        },
+        onProxyReqWs: function(proxyReq, req, socket) {
+          socket.on("error", function(err) {
+            console.warn("[WS Socket Error]", err.code, err.message);
+          });
+        },
       },
       "/api/web-server": {
         target: proxyTarget,


ui/src/App/Main.jsx
```diff
@@ -53,11 +54,21 @@ import isLoggedInQuery from './isLoggedIn.graphql';
...
     error: null,
+    subscriptionError: null,
+  };
+
+  state = {
+    subscriptionWarningDismissed: false,
+  };
+
+  handleSubscriptionWarningClose = () => {
+    this.setState({ subscriptionWarningDismissed: true });
```
lotas, line 71: I think this is a good option, if permissions are missing we would show this error in a Snackbar.
User might click close/dismiss and we would save in state that it was dismissed.
So user will not be shown this again (until page is hard refreshed I think)
Probably not an issue for this kind of error 👍

PR #8265, opened 2026-02-09
Don't try to listen on port 80 in containers

Eijebong:
cc760d1333829605ea99a8fc198f8cb245d475fb changed containers from running entrypoints as root to using the node user. On linux, listening on port 80 is a root only operation by default (which can be overriden through `/proc/sys/net/ipv4/ip_unprivileged_port_start`, but most likely isn't).

This commit fixes both the dev environment and deployments outside of environments setting `ip_unprivileged_port_start=0`. This isn't a breaking change since all ports are remapped by either docker or the `Service` in k8s so nothing changes from an outside perspective


lotas:
yes, this file is not mapped into container, and it wasn't released  yet, so makes sense. 
However, it still tries to write access logs and error logs to a location which it doesn't have permission to, checking..

PR #8269, opened 2026-02-10
TaskGroup profiler & task log profiler

### LINKED ISSUE #7374: Add support for viewing tasks and taskgroups in the profiler
While working on [Firefox Translations](https://github.com/mozilla/translations), I have been building some experimental [Taskcluster Tools](https://gregtatum.github.io/taskcluster-tools/) for use with training.

https://github.com/gregtatum/taskcluster-tools

One of the tools that would probably be useful to a general Taskcluster audience is integration with the Firefox Profiler.

I have this built as the [Task Profiler](https://gregtatum.github.io/taskcluster-tools/src/taskprofiler/). It takes a Task ID or a Task Group ID, grabs the information from the Taskcluster API, and converts it into the profiler format. From there it opens a new window and uses window.postMessage to inject it into the profiler UI.

lotas:
Server-side integration of the task group profiler - this moves existing UI component from #7374 from ui into web-server, to render those profiles on the backend

Profiles could be loaded using `https://profiler.firefox.com/from-url/<encoded-profile-endpoint-url>` scheme

Two new REST endpoints added in the web-server:
- `/api/web-server/v1/task-group/:taskGroupId/profile`
- `/api/web-server/v1/task/:taskId/profile`

they use queue client to fetch task data, generate Firefox Profiler JSON server-side, and return it with CORS headers

Fixes #7374


ui/src/views/Tasks/routes.js
matt-boris, line 1: We should be sure to add these new routes to the Fastly WAF's allowed routes list 👍🏻

ui/src/views/Tasks/ViewTask/index.jsx
```diff
@@ -537,6 +544,30 @@ export default class ViewTask extends Component {
...
+    const { taskId } = this.props.match.params;
+    const profileUrl = `${window.env.TASKCLUSTER_ROOT_URL}/api/web-server/v1/tasks/${taskId}/profile`;
+    const profilerUrl = `https://profiler.firefox.com/from-url/${encodeURIComponent(
+      profileUrl
+    )}`;
+
+    window.open(profilerUrl, '_blank');
+  };
+
+  handleSnackbarOpen = ({ message, variant = 'success', open }) => {
```
matt-boris: Doesn't look like this is called anywhere

services/web-server/src/profiler/profile.js
```diff
@@ -0,0 +1,307 @@
...
+            owner: task.task.metadata.owner,
+            description: task.task.metadata.description,
+            source: task.task.metadata.source,
+            retries: `${runId + 1} / ${retries}`,
+            state: run.state,
+            reasonCreated: run.reasonCreated,
+            reasonResolved: run.reasonResolved,
+            taskURL: `https://${url.host}/tasks/${taskId}/runs/${runId}`,
+            taskGroup: `https://${url.host}/tasks/groups/${task.task.taskGroupId}`,
+            liveLog: `https://${url.host}/tasks/${taskId}/runs/${runId}/logs/live/public/logs/live.log`,
```
matt-boris: I wonder if we should include a `backingLog` field here as well? In case there's no `live.log`.

services/web-server/src/api.js
```diff
@@ -0,0 +1,161 @@
...
+    'Generate a Firefox Profiler–compatible profile from a task\'s log output.',
+    'Parses `public/logs/live.log` (or `live_backing.log`) for timing data.',
+  ].join('\n'),
+}, async function(req, res) {
+  const { taskId } = req.params;
+  const queue = this.clients({ rootUrl: this.rootUrl }).queue;
+
+  let task;
+  try {
+    task = await queue.task(taskId);
```
matt-boris: Looks like you're just getting the task definition here, not the status, below code would've always resulted in `isResolved` being true.

```suggestion
    task = await queue.status(taskId);
```

matt-boris (commented):
Few things I've initially noticed

ui/src/views/Tasks/routes.js
lotas: oh, good point! totally forgot about it

services/web-server/src/profiler/profile.js
lotas: good question, I'm not sure exactly how those markers are being used. I can add

services/web-server/src/api.js
lotas: oh! brilliant catch, thank you

lotas: we actually need both :)

PR #8281, opened 2026-02-16
Remove insecure engine from generic-worker

petemoore:
## Summary

- Removes the insecure engine entirely from generic-worker — all code, build tags, schemas, tests, CI tasks, and documentation
- The multiuser engine becomes the only engine; all `//go:build multiuser` tags are stripped and `engine` is hardcoded as `"multiuser"` in `main.go`
- Workers previously using the insecure engine should switch to multiuser with `headlessTasks: true`

### Changes

- **24 files deleted**: insecure engine Go files, schema, tests, and doc page
- **~47 files stripped** of `//go:build multiuser` build tags (combined tags like `multiuser && (darwin || linux || freebsd)` simplified to platform-only)
- **`build.sh`**: `install` function simplified from 3 args (engine, OS, arch) to 2 (OS, arch); insecure targets removed; `-tags` flags removed
- **`kind.yml`**: insecure CI task deleted; `ENGINE` env var and `-tags`/`--build-tags` flags removed from remaining tasks
- **`d2g_test.go`**: 15 `switch engine` statements simplified — insecure cases removed, multiuser cases made unconditional per-OS
- **UI/tooling**: insecure schema references removed from `constants.js`, `validate.js`, `generic-worker.js`, `docker-compose.js`, `.ui`; `engines.mdx` rewritten
- **Generated files**: updated via `yarn generate`

## Verification

- Cross-compiles for darwin, linux, windows, freebsd (all pass)
- `go vet ./...` passes
- `golangci-lint run ./...` reports 0 issues
- `build.sh -s` runs successfully
- `yarn generate` completes cleanly
- No remaining `//go:build multiuser` or `//go:build insecure` tags in any Go files

## Test plan

- [ ] CI passes on all platforms (linux, macOS, windows)
- [ ] Verify generic-worker binary reports correct engine name and version
- [ ] Verify `headlessTasks: true` config works as replacement for insecure engine


changelog/remove-insecure-engine.md
matt-boris, line 1: Will the release artifacts have updated names? eg. removed the `-multiuser-` from the binary

If so, this should be called out in the changelog.

workers/generic-worker/gwconfig/config_multiuser.go
matt-boris, line 1: How about we rename all the files now to clean up the `_multiuser` part?

matt-boris (commented):
It would be nice to be able to continue running generic worker tests locally, as you could with the insecure engine. Ideally, a `go test` command should be able to run against a dockerized, headless multiuser engine. Currently as the PR sits, local dev with generic worker is more challenging than before.

workers/generic-worker/gwconfig/config_multiuser.go
petemoore: i thought that would be good in a second pass, just to break it up, as these are two big changes, and it would be good to get one out and tested, before we do the second major part, i think. We can make them separate commits in the same PR, it just feels like it might be better to split it over two PRs as they are both disruptive, and we have a chance to catch issues from the steps independently from each other.

changelog/remove-insecure-engine.md
petemoore: good call, i will update docs to be explicit

matt-boris: Oh, also looks like this changelog file isn't named to our standard (eg. what a `yarn changelog` command would produce).

workers/generic-worker/gwconfig/config_multiuser.go
matt-boris: Yeah I definitely think some way of splitting it up would be nice. I'd be fine splitting across a different commit in this PR 🤷🏻

changelog/remove-insecure-engine.md
petemoore: that's right. but the glob search means it still gets included, and i think the named changelog is preferable, i.e. this is an advantage rather than a disadvantage.

PR #8282, opened 2026-02-16
feat(w-m): Should worker terminate implementation

### LINKED ISSUE #7147: RFC191: worker should fetch launch config status to know if it should shutdown/restart

lotas:
This uses worker scanner loop to collect information about all running workers and at the end of the pass run the logic to dermine which workers to keep, and which should go
This is stored in the worker.providerData, that api call is reading from.
Decision can be updated with each worker scanner loop, so based on when worker makes the call, results might be different

New `worker_manager_workers_to_terminate` metric captures this decision at the end of each scanner loop

Fixes #7147

### DIFF
--- a/changelog/issue-7147.md
+++ b/changelog/issue-7147.md
@@ -3,7 +3,7 @@ level: minor
 reference: issue 7147
 ---
 
-Worker-manager decides which workers should be kept during worker-scanner loop.
+Worker-manager now decides which workers should be kept during the worker-scanner loop, surfaced via the `shouldWorkerTerminate` API.
 Decision is being made based on several policies:
 - if launch config is archived - worker would be marked as "shouldTerminate=true"
-- if workers exceed the desired capacity (more workers than pending tasks) - workers would be marked as "shouldTerminate=true"
+- if workers exceed the desired capacity (more workers than pending tasks) - the oldest workers would be marked as "shouldTerminate=true" (newest workers are kept)
--- a/services/worker-manager/src/api.js
+++ b/services/worker-manager/src/api.js
@@ -1021,7 +1021,7 @@ builder.declare({
     return res.reply({ terminate: decision.terminate, reason: decision.reason });
   }
 
-  return res.reply({ terminate: false, reason: 'no decision yet' });
+  return res.reply({ terminate: false, reason: 'none' });
 });
 
 builder.declare({
--- a/services/worker-manager/test/api_test.js
+++ b/services/worker-manager/test/api_test.js
@@ -2377,7 +2377,7 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
       await createWorker();
       const result = await helper.workerManager.shouldWorkerTerminate(workerPoolId, workerGroup, workerId);
       assert.strictEqual(result.terminate, false);
-      assert.strictEqual(result.reason, 'no decision yet');
+      assert.strictEqual(result.reason, 'none');
     });
 
     test('non-existent worker returns 404', async function() {
--- a/generated/references.json
+++ b/generated/references.json
@@ -18168,6 +18168,20 @@
           ],
           "title": "Worker startup duration",
           "type": "histogram"
+        },
+        {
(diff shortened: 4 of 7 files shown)


services/worker-manager/src/estimator.js
```diff
@@ -39,6 +37,41 @@ export class Estimator {
       // only scale as high as maxCapacity
       Math.min(adjustedPendingTasks * scalingRatio + totalNonStopped, maxCapacity),
     );
+
+    return { totalIdleCapacity, adjustedPendingTasks, totalNonStopped, desiredCapacity };
+  }
+
+  async desiredCapacity({
```
lotas, line 44: I moved all calculations to a private `#calculateCapacity` method, and existing `simple()` + `desiredCapacity()` bot use it

changelog/issue-7147.md
```diff
@@ -0,0 +1,9 @@
+audience: worker-deployers
+level: minor
+reference: issue 7147
+---
+
+Worker-manager decides which workers should be kept during worker-scanner loop.
+Decision is being made based on several policies:
+- if launch config is archived - worker would be marked as "shouldTerminate=true"
+- if workers exceed the desired capacity (more workers than pending tasks) - workers would be marked as "shouldTerminate=true"
```
petemoore: it would be good to clarify what determines which workers are kept (not terminated) when number of workers > pending tasks, whether it is random/oldest/youngest etc

services/worker-manager/src/api.js
```diff
@@ -1016,10 +1016,12 @@ builder.declare({
...
-  return res.reply({
-    terminate: false,
-    reason: 'unclear',
-  });
+  const decision = worker.providerData.shouldTerminate;
+  if (decision) {
+    return res.reply({ terminate: decision.terminate, reason: decision.reason });
+  }
+
+  return res.reply({ terminate: false, reason: 'no decision yet' });
```
petemoore: i don't understand what this reason means

services/worker-manager/src/estimator.js
```diff
@@ -4,17 +4,15 @@ export class Estimator {
     this.monitor = monitor;
   }
 
-  async simple({
-    workerPoolId,
-    providerId,
+  #calculateCapacity({
```
petemoore, line 7: does the # mean this is a private function?

services/worker-manager/test/api_test.js
```diff
@@ -2338,4 +2338,53 @@ helper.secrets.mockSuite(testing.suiteName(), [], function (mock, skipping) {
       }
     });
   });
+
+  suite('shouldWorkerTerminate', function() {
```
petemoore, line 2342: thanks for adding this

services/worker-manager/src/estimator.js
lotas: yes

services/worker-manager/src/api.js
lotas: should we not send anything? or you mean naming is confusing because decision was to keep it actually alive?

changelog/issue-7147.md
matt-boris, line 1: I think we should include in the changelog that this is all pertaining to the `shouldWorkerTerminate` API.

changelog/issue-7147.md
lotas: yes, I forgot to add this, thanks

services/worker-manager/src/api.js
lotas: this happens when worker-scanner didn't yet process this worker, so technically, 'no decision' has been made yet

lotas: as agreed in slack, we'll use 'none' here

changelog/issue-7147.md
lotas: added, thanks

PR #8287, opened 2026-02-17
feat(hooks): add ability to specify 'taskId' when triggering hooks

ahal:


services/hooks/src/api.js
```diff
@@ -8,6 +8,8 @@ import Ajv from 'ajv';
 import addFormats from 'ajv-formats';
 import { hookUtils } from './utils.js';
 
+const SLUGID_PATTERN = /^[A-Za-z0-9_-]{8}[Q-T][A-Za-z0-9_-][CGKOSWaeimquy26-][A-Za-z0-9_-]{10}[AQgw]$/;
```
ahal: I copied this from `services/queue/src/api.js`.. Maybe there's a shared place it can live?

clients/client-go/tchooks/tchooks.go
```diff
@@ -285,14 +285,21 @@ func (hooks *Hooks) RemoveHook(hookGroupId, hookId string) error {
...
+// Optionally, a `taskId` query parameter can be provided which the hook task
+// will use. It must be unique and follow the slugid format.
+//
 // Required scopes:
 //
 //	hooks:trigger-hook:<hookGroupId>/<hookId>
 //
 // See #triggerHook
-func (hooks *Hooks) TriggerHook(hookGroupId, hookId string, payload *TriggerHookRequest) (*TriggerHookResponse, error) {
+func (hooks *Hooks) TriggerHook(hookGroupId, hookId, taskId string, payload *TriggerHookRequest) (*TriggerHookResponse, error) {
```
matt-boris: This technically is a breaking change, since optional arguments don't exist in go. I'm not sure how this has been handled in the past, but might be worth calling out the breakage in the go client in the changelog and marking at major? This may also apply to the rust client? Not 100% on that one though.

services/hooks/src/api.js
matt-boris: Yeah, looks like it's duplicated in the web-server as well.

If you're feeling ambitious, maybe we could consolidate this const to the api lib (where we import `APIBuilder` from)?

Nothing I'd block on here though. Up to you.

changelog/issue-8287.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: minor
```
matt-boris, line 2:
```suggestion
level: major
```

Only because of optional function params not being a thing in go (maybe rust too?), I think this should be marked as breaking.

Then will you please just call it out in the description here too?

matt-boris (changes requested):
Changes LGTM, thank you for implementing!

Noted the breaking client changes, just requesting the changelog be updated please 🙏🏻

services/hooks/src/api.js
lotas: I feel like it would be good for `slugid` library to expose this: https://github.com/taskcluster/slugid/blob/main/src/slugid.js 
currently it doesn't, so I don't mind copy-pasting 🙈

services/hooks/src/api.js
```diff
@@ -470,6 +472,9 @@ builder.declare({
   scopes: 'hooks:trigger-hook:<hookGroupId>/<hookId>',
   input: 'trigger-hook.yml',
   output: 'trigger-hook-response.yml',
+  query: {
```
lotas: I'm just curious why not make it part of the payload?

ahal: Hm, so the only change in this PR is that the hook service would attempt to extract `taskId` out of the rendered definition before generating one? I could work with that.

Though I like this approach because it's a bit more automatic and doesn't rely on hook authors to remember to set `taskId: ${payload.taskId}` in the hook task definition. But if you think that drawback is worth avoiding an API change, I'd be happy to switch.. Lmk!

lotas: I don't mind, it just "feels different" putting taskId in to query string, something that I don't think was used much elsewhere. It was either part of the route `/:taskId` or in the payload.

Btw, if it moved into payload, would it drop the requirement for the breaking change that Boris was pointing out?

ahal: Yes, I think so.

services/hooks/src/api.js
```diff
@@ -636,8 +642,13 @@ const triggerHookCommon = async function({ req, res, hook, payload, clientId, fi
...
   }
 
+  const options = {};
+  if (payload.taskId) {
+    options.taskId = payload.taskId;
+  }
+
   try {
-    resp = await this.taskcreator.fire(hook, context);
+    resp = await this.taskcreator.fire(hook, context, options);
```
ahal, line 658: I realized that if I extracted the `taskId` here, then we wouldn't need to rely on hook authors to set it in the JSON-e template..

But if this crosses some sort of boundary, we could instead extract it from the rendered task definition.

PR #8290, opened 2026-02-18
fix(generic-worker): avoid reclaim creds deadlock

### LINKED ISSUE #8289: [generic worker] credential refresh blocked by `queueMux` during artifact uploads
## Summary

The background goroutine that refreshes task credentials via `ReclaimTask` can be blocked from updating `task.Queue` when many artifacts are being uploaded concurrently. This causes credentials to expire, resulting in 401 errors and worker panics during artifact uploads.

## Root Cause

`task.queueMux` (a `sync.RWMutex`) is held as an RLock for the entire duration of HTTP calls like `CreateArtifact`, including httpbackoff retries (up to 15 minutes). When the reclaim goroutine obtains fresh credentials, it needs a write lock on `queueMux` to swap in the new `task.Queue` client. That write lock is blocked until all in-flight `CreateArtifact` RLocks are released.

Under high artifact load:
1. Multiple upload goroutines hold `queueMux.RLock()` during `CreateArtifact` calls
2. If the queue is under load, httpbackoff retries extend the time these RLocks are held
3. The reclaim goroutine successfully calls `ReclaimTask` and gets new credentials, but blocks on `queueMux.Lock()` waiting for upload goroutines to finish
4. Upload goroutines continue using old credentials until they expire
5. Expired credentials cause 401 (`ext.certificate.expiry < now`), which panics the worker

Additionally, `FinishArtifact` and `DownloadArtifactToFile` read `task.Queue` without any lock at all, which is a data race with the reclaim goroutine's write.

## Related

- #8023 — original issue about micro-DDoS from 100k+ artifact uploads
- #8032 — limited concurrent uploads to 100
- #8167 — reduced limit to 10

Those PRs reduced the likelihood of this occurring by limiting concurrency, but didn't fix the underlying locking issue.

matt-boris:
Fixes #8289.

>Generic Worker: fixes credential expiration during high-volume artifact uploads by narrowing the scope of the queue client lock so that credential refresh is no longer blocked by in-flight HTTP calls.

### DIFF
--- a/workers/generic-worker/artifacts.go
+++ b/workers/generic-worker/artifacts.go
@@ -84,13 +84,14 @@ func (task *TaskRun) uploadArtifact(artifact artifacts.TaskArtifact) *CommandExe
 	}
 	par := tcqueue.PostArtifactRequest(json.RawMessage(payload))
 	task.queueMux.RLock()
-	parsp, err := task.Queue.CreateArtifact(
+	queue := task.Queue
+	task.queueMux.RUnlock()
+	parsp, err := queue.CreateArtifact(
 		task.TaskID,
 		strconv.Itoa(int(task.RunID)),
 		artifact.Base().Name,
 		&par,
 	)
-	task.queueMux.RUnlock()
 	if err != nil {
 		switch t := err.(type) {
 		case *tcclient.APICallException:
@@ -147,7 +148,7 @@ func (task *TaskRun) uploadArtifact(artifact artifacts.TaskArtifact) *CommandExe
 		return ResourceUnavailable(e)
 	}
 
-	e = artifact.FinishArtifact(resp, task.Queue, task.TaskID, strconv.Itoa(int(task.RunID)), artifact.Base().Name)
+	e = artifact.FinishArtifact(resp, queue, task.TaskID, strconv.Itoa(int(task.RunID)), artifact.Base().Name)
 	if e != nil {
 		task.Errorf("Error finishing artifact: %v", e)
 		return ResourceUnavailable(e)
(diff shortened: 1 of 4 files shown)


workers/generic-worker/mounts_feature.go
```diff
@@ -887,7 +887,10 @@ func (ac *ArtifactContent) Download(taskMount *TaskMount) (file string, sha256 s
 	taskMount.Infof("Downloading %v to %v", ac, file)
 
 	var runID int64 = -1 // use the latest run
-	_, contentLength, err := taskMount.task.Queue.DownloadArtifactToFile(ac.TaskID, runID, ac.Artifact, file)
```
matt-boris, line 890: Was previously a data race with the reclaim goroutine that writes `taskMount.task.Queue`.

workers/generic-worker/artifacts.go
```diff
@@ -147,7 +148,7 @@ func (task *TaskRun) uploadArtifact(artifact artifacts.TaskArtifact) *CommandExe
 		return ResourceUnavailable(e)
 	}
 
-	e = artifact.FinishArtifact(resp, task.Queue, task.TaskID, strconv.Itoa(int(task.RunID)), artifact.Base().Name)
```
matt-boris, line 150: Was previously a data race with the reclaim goroutine that writes `task.Queue`.

workers/generic-worker/artifacts.go
```diff
@@ -84,13 +84,14 @@ func (task *TaskRun) uploadArtifact(artifact artifacts.TaskArtifact) *CommandExe
 	}
 	par := tcqueue.PostArtifactRequest(json.RawMessage(payload))
 	task.queueMux.RLock()
-	parsp, err := task.Queue.CreateArtifact(
+	queue := task.Queue
```
lotas, line 87: is there any benefit of introducing this variable?

matt-boris: Yes! I'm quickly grabbing the reference to the queue client while holding the lock, then immediately releasing the lock right afterwards. As opposed to holding the lock throughout the entire API call.

PR #8298, opened 2026-02-19
feat(docs): Improve best-practices to catch all edge-cases

lotas:
Let's collect our best practices so we could later create dedicated `tc-review-skill` for agents to verify code according to those


petemoore (approved):
Looks great, thanks Yarik!

PR #8304, opened 2026-02-20
feat(queue): add `WorkerRemovedResolver` to resolve removed worker's tasks as `exception/worker-shutdown`

### LINKED ISSUE #7472: tasks should be rerun sooner when the machine they're running on is known to be gone
While digging into some worker issues today I noticed an odd state: a worker was known to be shutdown, but the task was still marked as running. Eventually, the task was marked as claim expired, and rerun. But this begs the question: why can't we do that sooner if the machine that claimed it has already been killed? I assume this is because worker manager knows the latter, while the queue doesn't, but I haven't dug deeply!

Image

### LINKED ISSUE #7477: Consider resolving claimed tasks as exception when a worker state changes to stopped
If you terminate a worker while it's running a task (or in my case at the same moment it claims one...), when the worker finally stops, the task never gets resolved until it reaches its claim expiration.

Image

matt-boris:
Fixes #7477. Fixes #7472.

>The queue service now listens for `workerRemoved` events from worker-manager and immediately resolves any tasks claimed by that worker as `exception/worker-shutdown`, triggering an automatic retry.
Previously, when a worker disappeared (due to VM preemption, crash, or manual termination), its claimed tasks would wait up to 20 minutes for the claim to expire before being retried.
This new `workerRemovedResolver` background process runs alongside the existing claim-resolver and requires no configuration changes.

### DIFF
--- a/services/queue/src/workerremovedresolver.js
+++ b/services/queue/src/workerremovedresolver.js
@@ -76,6 +76,11 @@ class WorkerRemovedResolver {
       await this.db.fns.resolve_task(taskId, runId, 'exception', 'worker-shutdown', 'retry'),
     );
 
+    // we no longer need existing claimed queue message
+    // because we just resolved the task, so remove it to
+    // prevent it from being processed by the claim resolver
+    await this.db.fns.queue_claimed_task_resolved(taskId, runId);
+
     const run = task.runs[runId];
 
     // If run wasn't resolved to exception/worker-shutdown, it was already
--- a/services/queue/test/workerremovedresolver_test.js
+++ b/services/queue/test/workerremovedresolver_test.js
@@ -80,6 +80,12 @@ helper.secrets.mockSuite(testing.suiteName(), ['aws'], function(mock, skipping) 
     assert.equal(status.status.runs[1].state, 'pending');
     assert.equal(status.status.runs[1].reasonCreated, 'retry');
 
+    // verify claimed task record was cleaned up
+    const db = await helper.load('db');
+    const claimed = await db.fns.get_claimed_tasks_by_worker(
+      taskQueueId, 'my-worker-group-extended-extended', 'my-worker-extended-extended');
+    assert.equal(claimed.length, 0);
+
     // verify pulse messages were published
     helper.assertPulseMessage('task-exception', m =>
       m.payload.status.taskId === taskId && m.payload.runId === 0);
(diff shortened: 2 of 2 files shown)


services/queue/src/main.js
```diff
@@ -283,6 +291,27 @@ let load = loader({
...
+      'dependencyTracker', 'pulseClient', 'workerManagerEvents',
+    ],
+    setup: async ({
+      db, queueService, publisher, dependencyTracker,
+      monitor, pulseClient, workerManagerEvents,
+    }) => {
+      let resolver = new WorkerRemovedResolver({
+        db, queueService, publisher, dependencyTracker,
+        pulseClient, workerManagerEvents,
+        monitor: monitor.childMonitor('worker-removed-resolver'),
```
lotas, line 307: I think this is not going to expose metrics correctly
`childMonitor` creates new instance, so metrics are going to be registered on a wrong instance

https://github.com/taskcluster/taskcluster/blob/b8afbd883b5ff37197c377fdc599995ea6b89a73/libraries/monitor/src/monitor.js#L125

Check this example: https://github.com/taskcluster/taskcluster/blob/b8afbd883b5ff37197c377fdc599995ea6b89a73/services/worker-manager/src/main.js#L229-L245

services/queue/src/workerremovedresolver.js
```diff
@@ -0,0 +1,134 @@
...
+    });
+
+    // If a retry run was created, publish pending message
+    const newRun = task.runs[runId + 1];
+    if (newRun &&
+        task.runs.length - 1 === runId + 1 &&
+        newRun.state === 'pending' &&
+        newRun.reasonCreated === 'retry') {
+      await Promise.all([
+        this.queueService.putPendingMessage(task, runId + 1),
```
lotas, line 123: I think it's probably reasonable to remove claimed record right away since we just resolved the task, so claim resolver wouldn't need to process it. And also because the record would still show up in the api response and UI if someone will open claimed tasks

```
this.db.fns.queue_claimed_task_resolved(taskId, runId),
```

Something like this method is doing:
https://github.com/taskcluster/taskcluster/blob/b8afbd883b5ff37197c377fdc599995ea6b89a73/services/queue/src/queueservice.js#L127-L158

services/queue/src/main.js
lotas: I was wrong, childMonitor passes shared manager, which is used to register metrics, so expose will still work correctly 👍

services/queue/src/workerremovedresolver.js
matt-boris: Addressed in https://github.com/taskcluster/taskcluster/pull/8304/changes/7af4657ad2fc707a7951f7ae53b806f87fff402c

services/queue/src/workerremovedresolver.js
```diff
@@ -0,0 +1,139 @@
...
+
+    // resolve as exception/worker-shutdown with retry
+    task.updateStatusWith(
+      await this.db.fns.resolve_task(taskId, runId, 'exception', 'worker-shutdown', 'retry'),
+    );
+
+    // we no longer need existing claimed queue message
+    // because we just resolved the task, so remove it to
+    // prevent it from being processed by the claim resolver
+    await this.db.fns.queue_claimed_task_resolved(taskId, runId);
```
lotas, line 82: wouldn't it be better to put this after the check you do in 88-92 lines? 
resolve_task can fail at this point if someone else already resolved it, and we only know by doing those checks below. 
Probably doesn't change much  though if we leave it here

matt-boris: There's a potential of leaving around some orphaned claimed task

PR #8313, opened 2026-02-25
feat(docs): Launch config weight distribution playground

lotas:
This adds a playground in documentation to test how initial weights of launch config change distributions by showing how many workers out of 1000 would roughly be created for a given weight

`/docs/reference/core/worker-manager/launch-configurations`

image

image

Should give a better picture how those weights would affect distribution


matt-boris (approved):
Awesome stuff! A great addition to our docs

PR #8320, opened 2026-02-26
fix(generic-worker): clean temp livelog streaming files

### LINKED ISSUE #7652: Livelog doesn't clean up temp streaming files properly
A picture is better than words here:
Image

Image

I realized this because I had some CI job erroneously generate 250MB of logs, on workers that have 1GB of RAM. And those workers just stopped uploading logs because `/tmp` was full (which is another bug tbh).

matt-boris:
Fixes #7652.

>Generic Worker & Livelog: fix livelog temporary streaming files not being cleaned up. The livelog process creates a temp directory per stream, but since the generic worker kills it with SIGKILL, the process never has a chance to clean up. Fixed by having the generic worker create a dedicated temp directory for each livelog process (via the new `LIVELOG_TEMP_DIR` env var) and removing it after the process is killed.

### DIFF
--- a/changelog/issue-7652.md
+++ b/changelog/issue-7652.md
@@ -1,0 +1,5 @@
+audience: worker-deployers
+level: minor
+reference: issue 7652
+---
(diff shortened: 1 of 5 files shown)


changelog/issue-7652.md
```diff
@@ -0,0 +1,5 @@
+audience: worker-deployers
+level: minor
```
matt-boris, line 2: Made this `minor` and not `patch` due to the new `LIVELOG_TEMP_DIR` env var that can be used when running livelog.

Issue #8101, opened 2025-11-18
Store artifact size


lotas:
Currently artifacts are uploaded/finished without the file size (content length)

It might be useful to store artifact size (s3/object) for perf monitoring

Eijebong:
This has been added. Some follow ups in #8531 and #8537

PR #8322, opened 2026-02-27
feat(g-w): reporting artifact size at creation time

lotas:
Generic-worker will send contentLength of the original artifact before compression to the queue.createArtifact

This will be a breaking change for worker deployers, since it depends on the schema validation in services, ath needs to be deployed before this.

Second part of #8101

### DIFF
--- a/changelog/issue-8101.md
+++ b/changelog/issue-8101.md
@@ -1,0 +1,6 @@
+audience: worker-deployers
+level: major
+reference: issue 8101
+---
+Generic worker now reports artifact file sizes to the queue via the optional `contentLength` parameter when creating artifacts. This applies to all file-based artifacts (S3 and object) as well as log artifacts. The reported size is the original file size on disk, before any encoding such as gzip compression.
+This feature relies on the API changes introduced in [v96.7.0](https://github.com/taskcluster/taskcluster/blob/main/CHANGELOG.md#v9670) and will fail schema validation if services are not upgraded.
--- a/workers/generic-worker/artifacts.go
+++ b/workers/generic-worker/artifacts.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net"
 	"net/url"
+	"os"
 	"runtime"
 	"strconv"
 	"strings"
@@ -40,12 +41,17 @@ func createDataArtifact(
 	contentType string,
 	contentEncoding string,
 ) artifacts.TaskArtifact {
+	var contentLength int64
+	if info, err := os.Stat(contentPath); err == nil {
+		contentLength = info.Size()
+	}
 	if config.CreateObjectArtifacts {
 		// note that contentEncoding is currently ignored for object artifacts
 		return &artifacts.ObjectArtifact{
-			BaseArtifact: base,
-			Path:         path,
-			ContentType:  contentType,
+			BaseArtifact:  base,
+			Path:          path,
+			ContentType:   contentType,
+			ContentLength: contentLength,
 		}
 	}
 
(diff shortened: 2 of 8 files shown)


changelog/issue-8101.md
```diff
@@ -0,0 +1,6 @@
+audience: worker-deployers
+level: major
+reference: issue 8101
+---
+Generic worker now reports artifact file sizes to the queue via the optional `contentLength` parameter when creating artifacts. This applies to all file-based artifacts (S3 and object) as well as log artifacts. The reported size is the original file size on disk, before any encoding such as gzip compression.
+This feature relies on the API changes introduced in [v96.7.0](https://github.com/taskcluster/taskcluster/blob/main/CHANGELOG.md#v9670) and will fail schema validation if services are not upgraded.
```
matt-boris, line 6: Thanks for calling this out! 👍🏻

workers/generic-worker/artifacts.go
```diff
@@ -39,13 +40,15 @@ func createDataArtifact(
 	contentPath string,
 	contentType string,
 	contentEncoding string,
+	contentLength int64,
```
matt-boris: Instead of passing `contentLength` everywhere this function is called and duplicating code for getting the content length, we could just calculate it in this function. I _think_ we'd technically want to do the `os.Stat()` on `contentPath` here, as that's the artifact content that's uploaded iirc (since we first copy to a temp file, then upload that).

Doing this would help clean up this PR and make it simple/straightforward with minimal changes, thank you!

workers/generic-worker/artifacts/object.go
```diff
@@ -18,13 +18,17 @@ type ObjectArtifact struct {
 	Path string
 	// ContentType is used in the Content-Type header.
 	ContentType string
+	// ContentLength is the original file size in bytes.
+	// Sent to the queue for monitoring purposes.
```
matt-boris:
```suggestion
	// ContentLength is the original file size in bytes, before any
	// encoding (e.g. gzip). Sent to the queue for monitoring purposes.
```

Mind just making the doc comment the same as in `workers/generic-worker/artifacts/s3.go` please?

workers/generic-worker/artifacts.go
```diff
@@ -55,10 +58,15 @@ func createDataArtifact(
...
 		ContentEncoding: contentEncoding,
+		ContentLength:   contentLength,
 	}
 }
 
 func (task *TaskRun) uploadLog(name, path string) *CommandExecutionError {
+	var contentLength int64
+	if info, err := os.Stat(path); err == nil {
+		contentLength = info.Size()
+	}
```
petemoore: Agree with Boris here: `uploadLog` is a wrapper around `task.uploadArtifact` which is a general purpose artifact upload function, so the content length calculation could happen inside that function, rather than `uploadLog`.

workers/generic-worker/artifacts.go
lotas: good idea 👍

workers/generic-worker/artifacts.go
lotas: Moving calculation to createDataArtifact as Boris suggested

workers/generic-worker/artifacts_test.go
```diff
@@ -877,27 +897,27 @@ func TestPublicDirectoryArtifact(t *testing.T) {
 
 	taskID := submitAndAssert(t, td, payload, "completed", "completed")
 
-	queue := serviceFactory.Queue(nil, config.RootURL)
```
matt-boris, line 880: Awesome cleanup here! Thanks so much!

PR #8324, opened 2026-02-27
feat(queue): Fix pagination issues with big task groups

### LINKED ISSUE #8323: Task doesn't appear on the task group page
As reported by @marco-c in matrix:

> https://firefox-ci-tc.services.mozilla.com/tasks/FzWJ02RQRSC0dXB3-SN5jQ is part of https://firefox-ci-tc.services.mozilla.com/tasks/groups/F61bIEKnQqWTvbJYX3aKXg, but if I search for its name in the search box, it isn't found

Also @Eijebong noticed:
> It's also not scheduled and the task group reports 0 unscheduled tasks

lotas:
What happened is that some very large Task Groups would return duplicate overlapping tasks in listTaskGroup response with pagination, while hiding some other tasks from the response.

Fixes #8323


db/versions/0122.yml
```diff
@@ -0,0 +1,65 @@
+version: 122
+description: Fix listTaskGroup pagination stability by adding ORDER BY task_id
+methods:
+  get_tasks_by_task_group_projid:
```
matt-boris, line 4: Is this fine to do? Overwrite the current db method? I thought typically we just add some incremented counter suffix?

lotas: yes, we change the name when signature changes - either arguments change or resulting table. Here everything stays the same, just the `ORDER` is added, so we didn't change the signature, just the internals :) No clients would suffer from this change

PR #8327, opened 2026-02-27
fix(generic-worker): stay alive at idle timeout if wm says to

### LINKED ISSUE #8326: Generic Worker does not check with Worker Manager before shutting down due to idle timeout
This is a crucial missing piece in getting `minCapacity` truly working for worker pools.

cc @ahal thank you for bringing this up!

matt-boris:
Fixes #8326.

>Generic Worker: when running with worker-runner, the worker now checks with Worker Manager before shutting down due to idle timeout. If Worker Manager says the worker is still needed (e.g., to satisfy `minCapacity`), the idle timer resets instead of shutting down. Workers not running with worker-runner are unaffected.

### DIFF
--- a/internal/mocktc/mocktc.go
+++ b/internal/mocktc/mocktc.go
@@ -88,24 +88,15 @@ func Marshal(req *http.Request, payload any) {
 	}
 }
 
-// Mocks holds references to all mock service instances, allowing tests to
-// configure their behavior.
-type Mocks struct {
-	WorkerManager *WorkerManager
-}
-
-func ServiceProviders(t *testing.T, baseURL string) ([]httputil.ServiceProvider, *Mocks) {
+func ServiceProviders(t *testing.T, baseURL string) []httputil.ServiceProvider {
 	t.Helper()
-	wm := NewWorkerManager(t)
 	return []httputil.ServiceProvider{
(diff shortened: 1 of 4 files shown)


internal/mocktc/mocktc.go
```diff
@@ -88,15 +88,24 @@ func Marshal(req *http.Request, payload any) {
...
+			NewIndexProvider(NewIndex(t)),
+			NewQueueProvider(NewQueue(t, baseURL)),
+			NewSecretsProvider(NewSecrets()),
+			NewWorkerManagerProvider(wm),
+			mocks3.New(t),
+			NewObjectProvider(NewObject(t, baseURL)),
+		}, &Mocks{
+			WorkerManager: wm,
+		}
 }
```
petemoore: I haven't understood why we need a new Mocks object. The implementations in the mocktc packages are already mocks, so we already had a mock Worker Manager?

matt-boris: It was to keep a reference to the worker manager mock instance in order to configure the `ShouldWorkerTerminate()` method by returning `Terminate: true` after N calls. I can just switch it to use a package level var instead to remove all this.

PR #8329, opened 2026-02-27
fix(worker-manager): `shouldWorkerTerminate` uses new `targetCapacity` estimator for decisioning

### LINKED ISSUE #8328: Worker scanner's termination formula inflates target capacity with existing worker counts
If `minCapacity` is set to 1, for example, an instance will come up and say up, properly (with https://github.com/taskcluster/taskcluster/pull/8327 applied), but if `minCapacity` is ever updated, to say, 0, that worker should get a terminate respond from worker manager, but does not.

matt-boris:
Fixes #8328.

>Worker Manager: the worker scanner now uses a dedicated target capacity formula for termination decisions based on pending tasks, claimed tasks, and `minCapacity`/`maxCapacity`. Previously, the provisioning formula was reused, which inflated the target by existing worker counts, so idle workers were never terminated even when `minCapacity` was lowered to 0.

### DIFF
--- a/services/worker-manager/src/worker-scanner.js
+++ b/services/worker-manager/src/worker-scanner.js
@@ -2,7 +2,7 @@ import _ from 'lodash';
 import Iterate from '@taskcluster/lib-iterate';
 import taskcluster from '@taskcluster/client';
 import { paginatedIterator } from '@taskcluster/lib-postgres';
-import { Worker, WorkerPool, WorkerPoolStats } from './data.js';
+import { Worker, WorkerPool } from './data.js';
 
 /**
  * Make sure that we visit each worker relatively frequently to update its state
@@ -56,8 +56,7 @@ export class WorkerScanner {
 
     this.monitor.info(`WorkerScanner providers filter: ${this.providersFilter.cond} ${this.providersFilter.value}`);
 
-    // Phase 1: Check workers and accumulate per-pool stats
-    const poolStats = new Map();
+    // Phase 1: Check workers and collect termination candidates
     const poolCandidates = new Map();
 
     const fetch =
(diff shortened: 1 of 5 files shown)


services/worker-manager/src/worker-scanner.js
```diff
@@ -131,12 +131,18 @@ export class WorkerScanner {
...
+        // workers. The provisioner's formula adds totalNonStopped and
+        // later subtracts it to get toSpawn, but here we compare
+        // directly against running workers.
         const desiredCapacity = await this.estimator.desiredCapacity({
           workerPoolId: poolId,
           minCapacity: pool.config.minCapacity ?? 0,
           maxCapacity: pool.config.maxCapacity ?? 0,
           scalingRatio: pool.config.scalingRatio ?? 1.0,
-          workerInfo: poolStats.get(poolId).forProvision(),
+          workerInfo: { existingCapacity: 0, stoppingCapacity: 0, requestedCapacity: 0 },
```
lotas: Hmm. I 'm not sure this is the right way though. 
https://github.com/taskcluster/taskcluster/blob/a173240cf3b1648fc58bc8655442319a41be2a20/services/worker-manager/src/estimator.js#L17-L39

If we pass zeroes here, it means that calculation will not account for claimed tasks, which means that pending tasks will always be "maxed out", which means that we will always be granting way more workers this "immunity" to leave longer

services/worker-manager/src/worker-scanner.js
```diff
@@ -131,12 +131,18 @@ export class WorkerScanner {
...
 
+        // For termination decisions we pass zeroed worker info so the
+        // estimator returns the raw target capacity based on demand
+        // (pending tasks) and minCapacity, not inflated by existing
+        // workers. The provisioner's formula adds totalNonStopped and
+        // later subtracts it to get toSpawn, but here we compare
+        // directly against running workers.
         const desiredCapacity = await this.estimator.desiredCapacity({
           workerPoolId: poolId,
           minCapacity: pool.config.minCapacity ?? 0,
```
lotas, line 130: here we always use the latest `minCapacity`, not sure what exactly the changelog line means: 
> Worker Manager: the `shouldWorkerTerminate` API now uses the latest `minCapacity` setting

lotas:
I think I understand the issue better now, looks like using `desiredCapacity` for the should worker terminate logic was not correct. We should think about something better, because for every case where pending < existing we will have this issue - more workers would be "requested to stay"

PR #8348, opened 2026-03-04
Document start-worker (worker runner) installation steps when using generic-worker

petemoore:
## Summary

This is a carry-over from #5635, which was accidentally closed due to an unintentional force-push.

- Updates the generic-worker installation docs to include `start-worker` (worker runner) download/installation steps across all platforms (Windows, macOS, Linux)
- Modernizes the "build from source" instructions to reflect the current monorepo structure
- Removes outdated references and cleans up formatting
- Adds `installing.mdx` to the release task file list

## Test plan

- [ ] Verify the docs render correctly in the UI
- [ ] Confirm download/installation steps are accurate for each platform

### DIFF
--- a/generated/docs-search.json
+++ b/generated/docs-search.json
@@ -3809,20 +3809,6 @@
   },
   {
     "element": "h2",
-    "id": "acquire-taskcluster-credentials-for-running-tests",
-    "path": "/reference/workers/generic-worker/installing",
-    "subtitle": "Acquire taskcluster credentials for running tests",
-    "title": "Installing Generic-Worker"
-  },
-  {
-    "element": "h2",
-    "id": "create-a-test-job",
-    "path": "/reference/workers/generic-worker/installing",
-    "subtitle": "Create a test job",
-    "title": "Installing Generic-Worker"
-  },
-  {
-    "element": "h2",
     "id": "installation-on-worker-machines",
     "path": "/reference/workers/generic-worker/installing",
     "subtitle": "Installation on worker machines",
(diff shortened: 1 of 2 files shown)


infrastructure/tooling/src/release/tasks.js
```diff
@@ -216,6 +216,7 @@ export default ({ tasks, cmdOptions, credentials }) => {
         'internal/**',
         'taskcluster/kinds/generic-worker/**',
         'tools/**',
+        'ui/docs/reference/workers/generic-worker/installing.mdx',
```
matt-boris: I like the idea of adding this here, but I don't believe it's doing anything for us, because `installing.mdx` uses lines like:

```bash
curl -L "https://github.com/taskcluster/taskcluster/releases/download/v<VERSION>/start-worker-linux-amd64" > start-worker
```

Which won't get replaced during release with the latest major version. I think this PR should also update `installing.mdx` some more to take advantage of this replacement, if needed (I see you've removed the `go get` commands that could have taken advantage of this replacement).

infrastructure/tooling/src/release/tasks.js
```diff
@@ -196,6 +196,12 @@ export default ({ tasks, cmdOptions, credentials }) => {
         contents.replace(/download\/v[0-9.]*\/taskcluster-/g, `download/v${requirements['release-version']}/taskcluster-`));
       changed.push(shellreadme);
 
+      const installingMdx = 'ui/docs/reference/workers/generic-worker/installing.mdx';
+      utils.status({ message: `Update ${installingMdx}` });
+      await modifyRepoFile(installingMdx, contents =>
+        contents.replace(/download\/v[0-9.]+\//g, `download/v${requirements['release-version']}/`));
```
matt-boris, line 202: Awesome stuff 🎉

PR #8366, opened 2026-03-13
Log SIGINT signal on graceful shutdown in generic-worker

### LINKED ISSUE #7901: generic-worker: SIGINT caught for graceful shutdown but not logged
On my system, the SIGINT signal is caught, but as it took 16 sec to shutdown on my system, and there was no log info about it, I thought the system didn't catch it at all.

I think a simple fix would be to log some "shutting down gracefully" when CTRL+C is caught so at least we know something is happening.

nitishagar:
## Summary

Fixes #7901.

When the generic-worker receives `SIGINT` (Ctrl+C), it now logs a message indicating
graceful shutdown has been initiated. Previously, no message was logged, which left
users unsure whether the signal had been caught — especially confusing when shutdown
takes several seconds.

This mirrors the existing behavior for `SIGTERM`, which already logs:
> "Received SIGTERM, initiating graceful termination"

## Changes
- `workers/generic-worker/main.go` — add log message in SIGINT handler (1 line)
- `changelog/issue-7901.md` — add changelog entry

## Test Plan
- Build generic-worker and verify `SIGINT` (Ctrl+C) now prints the log message
- Existing tests pass: `cd workers/generic-worker && go test ./...`

### DIFF
--- a/changelog/issue-7901.md
+++ b/changelog/issue-7901.md
@@ -1,0 +1,7 @@
+audience: worker-deployers
+level: silent
+reference: issue 7901
+---
+The generic-worker now logs a message when it receives SIGINT (Ctrl+C), making it
+clear that the signal was received and the worker is exiting. Previously, no message was logged,
+leaving users unsure whether the signal had been received.
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -580,6 +580,7 @@ func RunWorker() (exitCode ExitCode) {
 		select {
 		case <-wait5Seconds.C:
 		case <-sigInterrupt:
+			log.Println("Received SIGINT, worker is exiting")
 			return WORKER_STOPPED
 		}
 	}


workers/generic-worker/main.go
```diff
@@ -580,6 +580,7 @@ func RunWorker() (exitCode ExitCode) {
 		select {
 		case <-wait5Seconds.C:
 		case <-sigInterrupt:
+			log.Println("Received SIGINT, initiating graceful shutdown")
```
matt-boris: The "graceful termination" terminology isn't really appropriate here because no tasks would be running at this point and there's no use of the `graceful` package, unlike for the `SIGTERM` case. I think it'd be sufficient to just log that `SIGINT` was received and the worker is exiting (with code 71/`WORKER_STOPPED`).

changelog/issue-7901.md
```diff
@@ -0,0 +1,7 @@
+audience: worker-deployers
+level: silent
+reference: issue 7901
+---
+The generic-worker now logs a message when it receives SIGINT (Ctrl+C), making it
+clear that graceful shutdown has been initiated. Previously, no message was logged,
```
matt-boris:
```suggestion
clear that the signal was received and the worker is exiting. Previously, no message was logged,
```

PR #8368, opened 2026-03-13
Apply hooks search filter before pagination

### LINKED ISSUE #7972: hooks ui search only searches what is currently visible
eg: on https://firefox-ci-tc.services.mozilla.com/hooks, if I search for `translations` I get `project-translations`, but none of the nested things. This is especially confusing/annoying when looking for cron hooks, which are in the `project-releng` group but apply to other projects. (Something that is not obvious, and easy to forget.)

nitishagar:
## Summary

Fixes #7972.

Refactors the hooks search to use a dedicated `searchHooks` API endpoint instead of
overloading `listHookGroups`. This was done per reviewer feedback in
https://github.com/taskcluster/taskcluster/pull/8368#issuecomment-4066913975.

**Root cause:** The previous approach added a `search` query param to `listHookGroups`,
which didn't match its natural contract (it returns groups but was filtering by hook IDs).
The initial page load was also slow due to N+1 requests when fetching hook IDs per group.

**Fix:**
1. New `db/versions/0123.yml`: adds `search_hooks(search_term_in, page_size, page_offset)` —
   a single `ILIKE` query across `hook_group_id` and `hook_id`, returns full hook rows
2. New `GET /hooks/search?q=<term>` API endpoint (`searchHooks`) — returns matching
   hook definitions across all groups (same shape as `listHooks`)
3. `listHookGroups` reverted to its clean contract (no search param, uses `get_hook_groups()`)
4. GraphQL: `hookGroups` no longer takes `search`; new `searchHooks(query: String!): [Hook]` added
5. UI: when a `?search=` URL param is present, fetches from `searchHooks` and renders a flat
   `HooksListTable`; otherwise renders the normal `HookGroupsTable`

Initial page load is fast (group IDs only). Searches trigger a single efficient DB query.

## Changes

**DB:**
- `db/versions/0123.yml` — new `search_hooks` function (replaces `get_hook_groups_2`)
- `db/test/fns/hooks_test.js` — tests for `search_hooks`
- `db/test/versions/0123_test.js` — version smoke test

**API (`services/hooks`):**
- `services/hooks/schemas/v1/search-hooks-response.yml` — new schema (same shape as `listHooks`)
- `services/hooks/src/api.js` — reverted `listHookGroups`, added `searchHooks` at `/hooks/search`
- `services/hooks/test/api_test.js` — tests for `searchHooks`

**GraphQL/Web-server:**
- `services/web-server/src/graphql/Hooks.graphql` — removed `search` from `hookGroups`, added `searchHooks`
- `services/web-server/src/loaders/hooks.js` — new `hookSearch` DataLoader
- `services/web-server/src/resolvers/Hooks.js` — new `searchHooks` resolver

**UI:**
- `ui/src/views/Hooks/ListHookGroups/hookGroups.graphql` — simplified (no `$search`)
- `ui/src/views/Hooks/ListHookGroups/searchHooks.graphql` — new query
- `ui/src/views/Hooks/ListHookGroups/index.jsx` — two-mode rendering

**Generated (from `yarn generate`):**
- `generated/references.json`, `generated/db-schema.json`
- `libraries/postgres/@types/fns.d.ts`, `db/fns.md`, `db/schema.md`, `db/versions/README.md`
- All client libraries updated with `searchHooks` method

## Test Plan
- [ ] `cd db && yarn test` — DB function tests pass
- [ ] `cd services/hooks && yarn test` — API tests pass including new `searchHooks` suite
- [ ] Navigate to `/hooks` — groups table loads normally (no regression)
- [ ] Type a search term and submit — flat list of matching hooks appears
- [ ] Search matches hook IDs inside groups whose name doesn't match
- [ ] Search is case-insensitive (e.g. `PROJECT` matches `project-releng`)
- [ ] Clear search — groups table reappears


ui/src/views/Hooks/ListHookGroups/hookGroups.graphql
```diff
@@ -1,5 +1,8 @@
 query HookGroups {
   hookGroups {
     hookGroupId
+    hooks {
```
lotas: This feels like the right way to fix the given search issue, but the reason why we have separate HookGroups query is because it is fast. 
We had to create a new HookGroups view specifically to make first page load fast - original query takes <300ms to respond with all groups. 

Now, when we include hooks inside that is suddenly becomes much slower - takes ~6s to load

lotas (changes requested):
We cannot accept current changes as it would be a regression. 
Plus having client-side only search doesn't help here. 

We should probably think about implementing a server-side filtering, or only include nested hooks when search is present. 

Another idea on top of my head would be to use different view for the search purposes

lotas:
hey @nitishagar thanks for the refactor, I think this a step in the right direction, since we remove the expensive nested hook loading.

That said, I think we are still overloading listHookGroups with search behavior that does not match its natural ocntract. We filter for both hook group and hook id, but return only hook group. That shape existed mostly because current UI was optimized that way.

I think a cleaner approach would be to introduce a dedicated hook search endpoint and add new UI element that would display filtered hooks. 

`searchHooks` - `GET /hooks/search?q=xx` that will return same as `listHooks` is currently reporting but without single hook group constraint.

nitishagar:
Hi @lotas, thanks for the detailed review. I've reworked the approach based on your feedback.

**What changed:**

Instead of including `hooks { hookId }` in the `hookGroups` query (which caused the ~6s regression), the fix now uses a dedicated server-side search endpoint:

- Added `GET /hooks/search?q=<term>` REST endpoint (new `searchHooks` API method) backed by a `search_hooks(term)` PostgreSQL function that does a case-insensitive substring match (`ILIKE`) across both `hook_group_id` and `hook_id`
- Added `searchHooks(query: String!): [Hook]` to the GraphQL schema with resolver and DataLoader in web-server
- The UI now uses two separate queries: the existing fast `hookGroups` query (returns only group IDs, ~300ms) for the initial page load, and the new `searchHooks` query (full hook data) only when the user has typed a search term
- The switch is URL-driven: `?search=<term>` in the URL activates the search query; no search param uses the fast groups query
- Clearing the search now removes the `search` param from the URL entirely

**Performance:** Initial page load is unchanged (still uses the fast groups-only query). Search triggers one efficient DB query instead of N+1 requests.

**Tests:** Added DB function tests (`db/test/fns/hooks_test.js`) and API tests (`services/hooks/test/api_test.js`) covering scope enforcement, case-insensitive matching, cross-group results, and empty query handling.

nitishagar:
Thanks @lotas for the detailed feedback! The PR has been updated to address all your concerns:

**Server-side search implemented**: Instead of client-side filtering, the PR now adds a dedicated `searchHooks` GraphQL resolver backed by the existing `hooks.searchHooks({ q: query })` REST API endpoint. This keeps two distinct code paths:
- **No search**: Uses the fast `hookGroups` query (only fetches `hookGroupId`, <300ms as before) — no regression
- **Search active**: Uses a new `searchHooks(query: String!)` GraphQL query that calls the hooks service server-side and returns matching hooks rendered in a flat list via `HooksListTable`

The previous client-side filtering approach has been removed entirely. The `hookGroups.graphql` still only queries `hookGroupId`, so first page load performance is unchanged.

The CI failure (`ui-browser-test`) is due to worker `claim-expired` (infra timeout on the CI runners), not a test code issue.

services/hooks/src/api.js
```diff
@@ -54,6 +54,32 @@ builder.declare({
...
+  stability: 'stable',
+  description: [
+    'Search for hooks by a query string that matches hook group ID or hook ID',
+    '(case-insensitive substring match). Returns at most 1000 results.',
+    '',
+    'This endpoint requires the `hooks:list-hooks:` scope.',
+  ].join('\n'),
+}, async function(req, res) {
+  const { q } = req.query;
+  const hookRows = await this.db.fns.search_hooks(q || '', 1000, 0);
```
petemoore, line 78: the new endpoint seems to silently cap at 1000 results, which feels like a footgun for future problems, even if this is documented. can we not use pagination here? we do that in other APIs... thanks!

nitishagar:
Hey @petemoore 

Before I push the fix, wanted to sanity-check the shape with you, @lotas, and @matt-boris since there are a few design choices worth a sec of your time.

**Approach:** use the existing `paginateResults` helper in `@taskcluster/lib-api` to replace the hardcoded `1000, 0`. Handful of calls to make:

1. **Scope — full stack vs. REST-only.** The endpoint is only consumed by `ui/src/views/Hooks/ListHookGroups`, and the GraphQL layer currently returns a flat `[Hook]`. Leaning toward threading pagination all the way through (REST → web-server GraphQL + DataLoader → UI load-more) so the UI isn't silently capped either. If you'd rather see that as a follow-up PR, happy to split.

2. **Default page size = 100, `maxLimit` = 1000.** Nudges callers toward pagination without breaking anyone relying on the existing behavior. Alternative would be keeping 1000 as the default and making pagination purely opt-in — open to either.

3. **No new DB version.** `search_hooks` (`db/versions/0123.yml`) already takes `page_size_in` / `page_offset_in`; the handler was just passing literals. Fix lives in the API layer and above.

4. **Token format.** Standard offset-mode `paginateResults` (Hashids-wrapped integer), matching `listLastFires`, `listTaskGroup`, `listWorkerPools`.

5. **Schema/description.** `search-hooks-response.yml` gets optional `continuationToken` (following the `list-lastFires-response.yml` pattern); description updated; `yarn generate` output committed.

Also worth asking while we're here — should `listHooks` / `listHookGroups` get the same treatment, or is that out of scope? Appreciate any thoughts.

services/hooks/src/api.js
```diff
@@ -54,6 +54,42 @@ builder.declare({
   return res.reply({ groups });
 });
 
+/** Search hooks across all groups **/
+builder.declare({
+  method: 'get',
+  route: '/hooks/search',
+  name: 'searchHooks',
+  scopes: 'hooks:list-hooks:',
```
lotas: Existing `listHooks()` call expects `hooks:list-hooks:<hookGroupId>` scope, which means that we could allow users without necessary scope to see those in search output

Maybe having a `hooks:search-hooks` could be a right approach, or doing those checks in code to see if user role allows us accessing those hook groups

lotas:
sorry for the delayed response @nitishagar I agree that pagination is necessary to avoid partial results. 
I think the bigger question here is how to make sure we are not exposing more than user scopes allow. 
Some deployments may include `hooks:list-hooks:*` in the `anonymous` role, which will make this search endpoint work correctly, but some users might not have all scopes and would be limited to the `hooks:list-hooks:<hookGroupId>` hooks only. We need to make sure we observe those. 
I think in most simple case we can do check in code  - see what scopes are accessible, and if all we just run the db query as is, and if there are some scopes with specific group names - we would have to limit search to those only

nitishagar:
@lotas Your concern about per-group scope exposure makes sense, this update restores the per-group scope filtering that had been lost when pagination was added.

**What changed:**

`searchHooks` now enforces scopes in the handler instead of via a bare `hooks:list-hooks:` gate, mirroring `listHooks`'s per-group model:
- `hooks:list-hooks:*` (or broader) → search all groups (no DB restriction).
- `hooks:list-hooks:<hookGroupId>` for specific groups → search restricted to those groups only.
- `hooks:list-hooks:project-*` (prefix-wildcard) → restricted to the real group names the prefix matches.
- No `hooks:list-hooks:` scope at all → `InsufficientScopes`.

The restriction is pushed down into the DB function as an `allowed_groups_in text[]` parameter (`null` = no restriction), so it is AND-ed with the search term a per-group caller using `q=%` or `q=_` (LIKE wildcards) still only ever sees their own groups. Pagination (`paginateResults`, `maxLimit: 1000`, `continuationToken`) is preserved and composes correctly with the filter.

nitishagar:
Fix for the failing `ui-lint-test-build` check: the PR had deleted `ui/src/views/Hooks/ListHookGroups/__snapshots__/index.test.jsx.snap`, so jest in CI mode couldn't write the missing snapshot and the suite failed. Restored the snapshot file (unchanged from `main` — the rendered fragment matches) and amended it into the commit. Full `lint && test --coverage && build` sequence verified locally: 149 tests / 12 snapshots pass.

nitishagar:
Follow-up: the branch was conflicting with `main` — main has since taken its own `db/versions/0127.yml` (queue priority fix) and `0128`. Rebased onto current `main` as a single commit and renumbered this PR's migration to **0129** (`db/versions/0129.yml`, test, README, `db/fns.md`, `generated/db-schema.json` regenerated via `yarn generate`). Locally: db suite 600 passing, hooks suite 135 passing.

Heads-up that `service-worker-manager` fails on this and other PRs right now with 24 `provider_azure_test.js` "Signature validation error" failures — that's the known fixture expiry tracked in #8933, unrelated to this change.

lotas:
Sorry with the delayed response, we are in the process of a bigger migration to remove graphql layer, which has a higher priority. Once it is done, we can get back to this PR to keep non-graphql bits. 
Thank you

PR #8389, opened 2026-03-25
fix(github): stream backing log in extractLog to prevent OOM on large artifacts

### LINKED ISSUE #8388: GitHub check run never updated for tasks with very large logs
## Summary

When a Taskcluster task produces a very large log (e.g. ~98MB), the GitHub service's status handler fails to update the corresponding GitHub check run. The check run remains stuck as `in_progress` on GitHub indefinitely, even though the task has completed successfully in Taskcluster.

## Impact

This is currently blocking [wpt.fyi dashboard](https://wpt.fyi) Interop scores from updating. The `wpt-firefox-nightly-testharness-6` check run is stuck `in_progress` on **every single commit** to [web-platform-tests/wpt](https://github.com/web-platform-tests/wpt) — confirmed across 20+ consecutive commits.

Example: https://github.com/web-platform-tests/wpt/runs/68257594353
- GitHub status: `in_progress` (never updated)
- Taskcluster status: `completed` with exit code 0
- Task ID: `Bs29W3K_TNyao-sRjREyLw`
- Task log size: **~98MB / 1,037,375 lines**

For comparison, `testharness-5` (which updates correctly) has an ~8MB log.

The large log is caused by excessive WebRTC debug messages in the test output, specific to this test chunk.

## Root Cause Analysis

Two interacting bugs in the GitHub service:

### 1. No size limit on artifact download (`services/github/src/handlers/requestArtifact.js`)

The `requestArtifact` function downloads the entire artifact into memory as a string via `superagent`:

```javascript
const res = await utils.throttleRequest({ url, method: 'GET' });
// ...
return res.text.toString(); // entire 98MB buffered in memory
```

This is then passed to `extractLog()` which only uses the first 20 and last 200 lines — the other ~1 million lines are downloaded for nothing. For a 98MB log, this likely causes an OOM or timeout in the handler.

### 2. Fire-and-forget handler means no retry (`services/github/src/handlers/index.js:142-158`)

The `callHandler` wrapper does not return the handler's promise:

```javascript
const callHandler = (name, handler) => {
  const timedHandler = this.monitor.timedHandler(`${name}listener`, handler.bind(this));
  return (message) => {
    this._handlerStarted(name);
    timedHandler.call(this, message).catch(async err => {
      await this.monitor.reportError(err);
      return err;
    }).then((err = null) => {
      this._handlerFinished(name, !!err);
      // ...
    });
    // Returns undefined — NOT the promise
  };
};
```

Because the Pulse consumer (`libraries/pulse/src/consumer.js:206`) awaits `handleMessage(msg)` and gets `undefined`, it immediately ACKs the message. If the handler subsequently fails (e.g. OOM during the 98MB download), the message is already acknowledged and will never be retried. The check run update is silently lost.

## Suggested Fix

For the artifact download, one of:
- Use HTTP `Range` requests to fetch only the first/last N kilobytes of the log
- Stream the artifact response and truncate during download rather than buffering the entire body
- Add a `Content-Length` check and skip the download if it exceeds a threshold (the log is only used for the check run output text, which is capped at 60KB anyway)

For the handler, consider returning the promise from `callHandler` so the Pulse consumer can nack/retry on failure, or add an explicit retry mechanism for check run updates.

## Reproduction

Any task that produces a log larger than ~50-100MB and is integrated via the GitHub service will exhibit this behavior. The `wpt-firefox-nightly-testharness-6` task reproduces it on every run.

## Environment

community-tc (but the bug exists in the code for all environments)

petemoore:
## Summary

Change `extractLog` to accept an async iterable (readable stream) instead of a string. It streams the response line-by-line, keeping only the first 20 and last 200 lines in memory via a head buffer and a circular tail buffer. The formatted output is identical to before.

- The backing log is now fetched via `fetch()` and streamed directly to `extractLog`, instead of being downloaded entirely into memory via `requestArtifact`
- Other artifact fetches (custom text, annotations JSON) continue using `requestArtifact` unchanged
- The helper functions `tailLog`, `extractTailLinesFromLog`, and `extractHeadLinesFromLog` are removed as they are no longer needed
- No new files — all changes are in existing files

Fixes #8388

## Test plan

- [x] All 265 github service tests pass
- [x] `extractLog` tests updated to pass in-memory streams via `Readable.from()`
- [x] Tests cover: small logs, large logs, long single lines, empty streams, ANSI stripping, accurate hidden line count

### DIFF
--- a/services/github/src/utils.js
+++ b/services/github/src/utils.js
@@ -160,95 +160,102 @@ export const ansi2txt = (src) => {
 };
 
 /**
- * Github checks API call is limited to 64kb
- * @param {string} log
- * @param {number} maxLines
- * @param {number} maxPayloadLength
- * @returns string
- */
-export const tailLog = (log, maxLines = 250, maxPayloadLength = 30000) => {
-  return ansi2txt(log).substring(log.length - maxPayloadLength)
-    .split('\n')
-    .slice(-maxLines)
-    .join('\n');
-};
-
-/**
+ * Github checks API call is limited to 64kb.
  *
- * @param {string} logString
- * @param {number} maxPayloadLength
- * @param {number} tailLines
- * @returns string or null
- */
-export const extractTailLinesFromLog = (logString, maxPayloadLength, tailLines) => {
-
-  if (tailLines === 0) {
-    return null;
-  }
-
-  const tailLog = logString.split("\n").slice(-tailLines).join("\n");
-
-  if (logString.length <= maxPayloadLength) {
-    return tailLog;
-  }
-
-  let tailLogMaxPayload = logString.slice(-maxPayloadLength);
-  const newLinePosition = tailLogMaxPayload.indexOf('\n');
-  tailLogMaxPayload = tailLogMaxPayload.slice(newLinePosition + 1);
-
-  return tailLog.length <= tailLogMaxPayload.length ? tailLog : tailLogMaxPayload;
-};
-
-/**
+ * Streams a log from an async iterable (e.g. a fetch response body),
+ * extracting the first headLines and last tailLines with bounded memory.
+ * Uses a head buffer and a tail buffer that evicts old entries, so memory
+ * usage is O(headLines + tailLines) regardless of log size.
  *
- * @param {string} logString
- * @param {number} maxPayloadLength
- * @returns string
+ * @param {AsyncIterable} stream - async iterable yielding byte chunks
+ * @param {number} headLines - number of lines to keep from the start
+ * @param {number} tailLines - number of lines to keep from the end
+ * @param {number} maxPayloadLength - maximum output length in bytes
+ * @returns {Promise<string>} formatted log extract
  */
-export const extractHeadLinesFromLog = (logString, maxPayloadLength) => {
-
-  if (logString.length <= maxPayloadLength) {
-    return logString;
-  }
-
-  const headLog = logString.slice(0, maxPayloadLength);
-  const lastNewLinePosition = headLog.lastIndexOf('\n');
-  return headLog.substring(0, lastNewLinePosition);
-};
-
-/**
- * Github checks API call is limited to 64kb
- * @param {string} log
- * @param {number} headLines
- * @param {number} tailLines
- * @param {number} maxPayloadLength
- * @returns string
- */
-export const extractLog = (log, headLines = 20, tailLines = 200, maxPayloadLength = 30000) => {
-  const logString = ansi2txt(log);
-  const lines = logString.split('\n');
+export const extractLog = async (stream, headLines = 20, tailLines = 200, maxPayloadLength = 30000) => {
   const LOG_BUFFER = 42;
+  const decoder = new TextDecoder('utf-8', { stream: true });
+  const head = [];
+  const tail = [];
+  let totalLines = 0;
+  let headFull = false;
+  let leftover = '';
 
-  if (lines.length <= headLines + tailLines && logString.length <= maxPayloadLength) {
-    return logString;
+  for await (const chunk of stream) {
+    const text = leftover + decoder.decode(chunk, { stream: true });
+    const lines = text.split('\n');
+    leftover = lines.pop();
+
+    for (const line of lines) {
+      const clean = ansi2txt(line);
+      totalLines++;
+      if (!headFull) {
+        head.push(clean);
+        if (head.length >= headLines) {
+          headFull = true;
+        }
+      } else {
+        tail.push(clean);
+        if (tail.length > tailLines) {
+          tail.shift();
+        }
+      }
+    }
   }
 
-  const headLogArray = lines.slice(0, headLines);
-  const headLog = headLogArray.join('\n');
+  // Flush any remaining bytes from the decoder
+  const finalText = leftover + decoder.decode();
+  if (finalText) {
+    const clean = ansi2txt(finalText);
+    totalLines++;
+    if (!headFull) {
+      head.push(clean);
+    } else {
+      tail.push(clean);
+      if (tail.length > tailLines) {
+        tail.shift();
+      }
+    }
+  }
 
+  const headLog = head.join('\n');
+  const tailLog = tail.join('\n');
+  const fullLog = tailLog ? headLog + '\n' + tailLog : headLog;
+
+  // Small log: return full content if it fits
+  if (totalLines <= headLines + tailLines && fullLog.length <= maxPayloadLength) {
+    return fullLog;
+  }
+
+  // If head alone exceeds the payload budget, truncate at a line boundary
   if (maxPayloadLength <= headLog.length) {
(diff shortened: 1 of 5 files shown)


services/github/src/handlers/requestArtifact.js
```diff
@@ -9,12 +85,12 @@ export async function requestArtifact(artifactName, { taskId, runId, debug, inst
       authorizedScopes: scopes,
     });
     const url = limitedQueueClient.buildSignedUrl(limitedQueueClient.getArtifact, taskId, runId, artifactName);
-    const res = await utils.throttleRequest({ url, method: 'GET' });
+    const { status, text } = await requestArtifact.fetchArtifactTail(url);
```
lotas: This is slightly changing the behavior. we used to have 20 head lines also to display what kind of worker and environment this is: 
https://github.com/taskcluster/taskcluster/blob/57b58d9a1077d63fbcbe54037f3cc0c85737977b/services/github/src/utils.js#L227

So while this overall refactoring makes sense for the tail of the log, I think we need to decide what to do with the header now, since it will not be present

lotas: oh.. also it was used to load all artifacts, including annotations and custom text: 
https://github.com/taskcluster/taskcluster/blob/57b58d9a1077d63fbcbe54037f3cc0c85737977b/services/github/src/handlers/status.js#L135-L139

so we can't use this approach here directly because it would truncate `customCheckRunAnnotations.json` and `customCheckRunText.md` most likely

services/github/src/handlers/requestArtifact.js
```diff
@@ -1,4 +1,80 @@
...
+async function fetchArtifactTail(url) {
+  for (let attempt = 1; attempt <= MAX_RETRIES; attempt++) {
+    let response;
+    try {
+      response = await fetch(url, {
+        method: 'GET',
+        headers: { 'Range': `bytes=-${TAIL_BYTES}` },
+        redirect: 'follow',
+      });
+    } catch (e) {
```
lotas: probably should be checked what kind of error is that. can be 404 so no need to retry

also google says invalid range is possible:

```
Invalid Ranges: If the range is completely out of bounds (e.g., if negative parsing errors occur, as mentioned on [GitHub](https://github.com/golang/go/issues/40940)), the server should return a 416 Range Not Satisfiable.
```

services/github/src/handlers/requestArtifact.js
```diff
@@ -1,4 +1,80 @@
...
+    }
+
+    if (response.status >= 500) {
+      // Consume body to free resources before retry
+      await response.body?.cancel();
+      if (attempt < MAX_RETRIES) {
+        await new Promise(r => setTimeout(r, 2 ** attempt * RETRY_BASE_DELAY));
+        continue;
+      }
+      return { status: response.status, text: '' };
```
lotas: do we hide response body on purpose? `text: ''` I believe this was used before to reporterror on the response body

services/github/src/handlers/fetchLogTail.js
```diff
@@ -0,0 +1,125 @@
...
+ * ran. It is fetched after the task completes, not during execution.
+ *
+ * Returns a string suitable for passing to extractLog(), which extracts head
+ * and tail lines from it. For small logs that fit within HEAD_BYTES, a single
+ * request suffices.
+ *
+ * @param {string} url - Signed URL to the backing log artifact
+ * @returns {string} - Log text (head + tail, or full content if small)
+ */
+async function fetchBackingLogTail(url) {
```
lotas: is this still fetching only Tail as name implies? Comment says "head and tail"

services/github/src/handlers/fetchLogTail.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  // 200 means the server returned the full body (artifact is small or Range
+  // not supported). Stream it but cap at HEAD_BYTES + TAIL_BYTES.
+  if (headResponse.status === 200) {
+    return await readTail(headResponse.body, HEAD_BYTES + TAIL_BYTES);
+  }
+
+  // 206 means we got a partial response — read the head portion
+  const headText = await readTail(headResponse.body, HEAD_BYTES);
```
lotas: does it make sense to reuse this function for the head of the log? wouldn't it take the last x bytes/lines only? 

Difficult to tell if it would, because tests are missing for those newly added functions

services/github/src/handlers/fetchBackingLog.js
```diff
@@ -0,0 +1,125 @@
+// Sizes for the head and tail Range requests. extractLog uses 20 head lines
+// and 200 tail lines with a max payload of 60KB, so these are generous.
+const HEAD_BYTES = 50 * 1024; // 50KB for first 20 lines
```
lotas: This is probably too optimistic since we know we are always limited by the 64Kb payload size of the github request

Maybe we should be more conservative as we won't need 50+512Kb for sure. 
Head lines are probably like 5Kb at most?  
And considering we would need some of the total 64kb limit for the whole payload, it would probably safe to load way less than 512Kb for tail as well

https://github.com/taskcluster/taskcluster/blob/57b58d9a1077d63fbcbe54037f3cc0c85737977b/services/github/src/handlers/utils.js#L53-L60

services/github/src/handlers/fetchBackingLog.js
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  if (tailResponse.status === 200) {
+    // Server doesn't support Range — we got the full body again.
+    // Stream it, keeping only the last HEAD_BYTES + TAIL_BYTES.
+    return await streamWithLimit(tailResponse.body, HEAD_BYTES + TAIL_BYTES);
+  }
+
+  const tailText = await streamWithLimit(tailResponse.body, TAIL_BYTES);
+  // For logs between ~512KB and ~562KB, the head and tail ranges may overlap,
```
lotas: are you sure this comment is still valid? 
smaller logs might indeed introduce overlapping lines though

services/github/src/utils.js
```diff
@@ -160,95 +160,102 @@ export const ansi2txt = (src) => {
...
+
+  for await (const chunk of stream) {
+    const text = leftover + decoder.decode(chunk, { stream: true });
+    const lines = text.split('\n');
+    leftover = lines.pop();
+
+    for (const line of lines) {
+      const clean = ansi2txt(line);
+      totalLines++;
+      if (!headFull) {
```
lotas, line 193: nice! head full of messages

services/github/src/utils.js
```diff
@@ -160,95 +160,102 @@ export const ansi2txt = (src) => {
...
+        head.push(clean);
+        if (head.length >= headLines) {
+          headFull = true;
+        }
+      } else {
+        tail.push(clean);
+        if (tail.length > tailLines) {
+          tail.shift();
```
lotas, line 201: smart!

PR #8398, opened 2026-03-26
feat(d2g): minor performance improvements

matt-boris:
>D2G: performance improvements to `docker run` for d2g tasks. Adds `--pull=never` to skip redundant registry checks (image is already loaded), `--log-driver=none` to eliminate duplicate log writes, and parallelizes artifact extraction from stopped containers.

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -398,6 +398,18 @@ func runCommand(
 	// https://docs.docker.com/reference/cli/docker/container/run/
 	args = append(args, "--memory-swap", "-1", "--pids-limit", "-1")
 
+	// The d2g task feature always ensures that the image is
+	// available locally (via docker pull or docker load) before running
+	// the container. Skip the local image lookup since we know the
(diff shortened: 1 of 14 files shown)


tools/d2g/d2g.go
```diff
@@ -398,6 +398,18 @@ func runCommand(
 	// https://docs.docker.com/reference/cli/docker/container/run/
 	args = append(args, "--memory-swap", "-1", "--pids-limit", "-1")
 
+	// The d2g task feature always ensures that the image is
+	// available locally (via docker pull or docker load) before running
+	// the container. Skip the registry check to avoid redundant
+	// network round-trips.
```
jcristau: nit: I don't think there's a registry check here, the default is `missing` so docker would just be looking up the image in its local cache and finding it there and never hitting the network?

matt-boris: Fixed the comment, thanks for pointing out 👍🏻

PR #8413, opened 2026-03-31
feat(release): publish npm packages from gha

### LINKED ISSUE #8013: Look into updating how we publish our nodejs clients to NPM
Due to changes described in https://github.blog/changelog/2025-09-29-strengthening-npm-security-important-changes-to-authentication-and-token-management/ I think we need to create a new path forward for publishing our two clients `@taskcluster/client` and `@taskcluster/client-web` to NPM during our release (using `yarn release`). Unlimited token expiration is no longer a thing, so we'll either need to update our release token every 90 days or look into using trusted publishing https://docs.npmjs.com/trusted-publishers

In the short term, I've created a new, granular access token to replace our legacy token that'll be removed/disabled automatically in November by NPM. This token expires on 2026/01/11.

matt-boris:
Fixes #8013.

Will be able to test the GitHub Action with a dry run npm publish once this lands.

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -1,0 +1,66 @@
+name: Publish NPM Packages
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Release tag to publish (e.g. v99.0.0)'
+        required: true
+      dry-run:
+        description: 'Dry run (do not actually publish)'
+        type: boolean
+        default: true
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    # Only run on the main repo, not forks or staging
+    if: github.repository == 'taskcluster/taskcluster'
+    permissions:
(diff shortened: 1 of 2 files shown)


infrastructure/tooling/src/build/tasks/release.js
```diff
@@ -126,53 +129,77 @@ export default ({ tasks, cmdOptions, credentials, baseDir, logsDir }) => {
...
+      const dir = path.join(REPO_ROOT, 'clients/client');
 
-      await npmPublish({
-        dir: path.join(REPO_ROOT, 'clients/client'),
-        apiToken: credentials.npmToken,
-        logfile: path.join(logsDir, `publish-clients-client.log`),
-        utils });
+      await execCommand({
+        dir,
+        command: ['npm', 'pack'],
```
lotas, line 147: do we need `yarn` or `npm` here?

matt-boris: I think `npm` is good since we were using it for publishing, it was using its internal `pack` anyway :)

PR #8415, opened 2026-03-31
fix: use actual npm pack output for tarball filename

matt-boris:
npm pack uses the version from package.json, which may not match the release-version (e.g., during staging releases where package.json is not bumped). Capture the tarball filename from npm pack's stdout instead of constructing it from release-version.

Part of #8013

### DIFF
--- a/infrastructure/tooling/src/build/tasks/release.js
+++ b/infrastructure/tooling/src/build/tasks/release.js
@@ -132,25 +132,24 @@ export default ({ tasks, cmdOptions, credentials, baseDir, logsDir }) => {
     title: `Pack clients/client for npm`,
     requires: [
       'clean-artifacts-dir',
-      'release-version',
     ],
     provides: [
       'npm-client-artifact',
     ],
     run: async (requirements, utils) => {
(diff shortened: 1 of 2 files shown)


infrastructure/tooling/src/build/tasks/release.js
```diff
@@ -132,25 +132,24 @@ export default ({ tasks, cmdOptions, credentials, baseDir, logsDir }) => {
...
         dir,
         command: ['npm', 'pack'],
         utils,
+        keepAllOutput: true,
         logfile: path.join(logsDir, 'pack-clients-client.log'),
       });
 
-      // npm pack produces taskcluster-client-<version>.tgz in the working directory
-      const tarball = `taskcluster-client-${version}.tgz`;
+      // npm pack prints the tarball filename on the last line of stdout
```
lotas, line 151: sounds fragile :) but I'm not sure if we could/should add  more checks around the output

matt-boris: I know.. I don't _love it_ but it should work ok for us.. :/

PR #8417, opened 2026-04-01
fix: bound OAuth2 token scopes to registered client scopes

matt-boris:
The OAuth2 grant callbacks only intersected the decision response scopes (ares.scope) with the user's scopes, but not with the registered client's allowed scopes. A tampered consent form POST could request scopes beyond what the client was registered for, bounded only by the user's own scopes.

Now ares.scope is intersected with registeredClient.scope before the user scope intersection, and a warning is logged when the decision response scopes differ from the authorization request scopes.


services/web-server/src/servers/oauth2.js
```diff
@@ -56,6 +56,15 @@ export default (cfg, db, strategies, auth, monitor) => {
 
     const userScopes = (await auth.expandScopes({ scopes: currentUser.scopes() })).scopes;
 
+    if (!_.isEqual([...ares.scope].sort(), [...areq.scope].sort())) {
+      monitor.log.oauth2ScopeMismatch({
```
lotas: shouldn't we also throw an error here on scope mismatch?

PR #8465, opened 2026-04-08
Fix blank page on pending/claimed tasks when errors occur

### LINKED ISSUE #7802: The "pending tasks" page doesn't display errors when they happen
To reproduce:

Go here while unauthenticated: https://firefox-ci-tc.services.mozilla.com/provisioners/releng-hardware/worker-types/gecko-t-osx-1400-r8/pending-tasks

No error is printed out on the page despite the graphql query returning:

```
{"errors":[{"message":"This request requires Taskcluster credentials that satisfy the following scope expression:\n\n```\nqueue:pending-list:releng-hardware/gecko-t-osx-1400-r8\n```\n\n","locations":[{"line":2,"column":3}],"path":["listPendingTasks"]},{"message":"Worker pool does not exist\n\n","locations":[{"line":37,"column":3}],"path":["WorkerPool"]}],"data":{"listPendingTasks":null,"WorkerPool":null}}
```

nitishagar:
## Summary

- The pending tasks (and claimed tasks) pages showed a blank page instead of an error when GraphQL returned errors alongside a missing-worker-pool error.
- Root cause: `wpMissing` used `error.message.includes(...)` on the combined error string, which matched even when an unrelated error (e.g. insufficient scopes) was also present — silencing the `ErrorPanel`.
- Fix: check `error.graphQLErrors.every(...)` so the error panel is only suppressed when **all** errors are the expected "Worker pool does not exist" case.

## Test plan

- [ ] Visit a pending tasks page for a provisioner/worker-type where the worker pool does not exist in worker-manager **and** you lack the required scopes — previously blank, now shows the scope error.
- [ ] Visit a pending tasks page where the worker pool simply does not exist in worker-manager but the query succeeds — error panel still suppressed (no regression).
- [ ] Same two scenarios for the claimed tasks page.

Closes #7802

### DIFF
--- a/changelog/7802.md
+++ b/changelog/7802.md
@@ -1,0 +1,7 @@
+level: patch
+audience: users
+reference: issue 7802
+---
+The pending tasks and claimed tasks pages now correctly display errors (such as
+insufficient scopes) instead of showing a blank page when the worker pool is
+also absent from worker-manager.
--- a/ui/src/views/Provisioners/ClaimedTasks/index.jsx
+++ b/ui/src/views/Provisioners/ClaimedTasks/index.jsx
@@ -102,8 +102,13 @@ export default class WMViewClaimedTasks extends Component {
     } = this.props;
     const { provisionerId, workerType } = this.props.match.params;
     // Claimed tasks could exist for the pools that are not managed by w-m
-    // so one of the request would fail with errors
-    const wpMissing = error?.message?.includes('Worker pool does not exist');
+    // so one of the requests would fail with errors; only suppress the error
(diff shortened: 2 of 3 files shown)


petemoore:
This looks great! I tested. Before:

Screenshot 2026-04-17 at 16 24 36

After:

Screenshot 2026-04-17 at 16 25 31

I also checked claimed tasks:

Screenshot 2026-04-17 at 16 26 49

Looks perfect, thanks very much!

@Eijebong does this also look ok to you?

Thanks

petemoore (approved):

PR #8466, opened 2026-04-08
Fix invalid curl and tc-cli examples in API reference

### LINKED ISSUE #8198: UI API references provide invalid instructions for curl/tc-cli
See for example:

Image
Image

The curl example suggests using a bearer token, but auth in taskcluster is implemented using hawk. Curl examples should probably not include setting Authorization header and should instead advise the user that curl can be used by tasks when taskcluster-proxy feature is enabled.

The taskcluster cli version for this particular API endpoint should not be:

```
taskcluster api github latest --owner "<owner>" --repo "<repo>" --branch "<branch>"
```

but instead should use positional rather than named arguments, i.e.:

```
taskcluster api github latest "<owner>" "<repo>" "<branch>"
```

I haven't yet validated all the other client implementation details, I wonder if we can automate the testing of the generated code somehow. Or at least for a subset of the endpoints.

nitishagar:
## Summary

- The curl examples in the API reference were showing `Authorization: Bearer` headers, which is not valid for Taskcluster (Taskcluster uses Hawk authentication). The examples now explain that `taskcluster-proxy` should be used instead.
- The `taskcluster` CLI examples were showing path arguments as named flags (e.g. `--taskId "value"`) when the CLI actually requires positional arguments. The examples now use the correct positional syntax.
- The CLI examples also used `--input -` for stdin body input, but no such flag exists; the CLI reads from stdin directly. This is now corrected.

## Test plan

- [ ] Visit an API reference page for an authenticated endpoint, check the curl tab — should no longer show `Authorization: Bearer`, instead shows taskcluster-proxy guidance
- [ ] Visit an API reference page for an endpoint with path parameters (e.g. queue's `task`), check the Shell tab — should show positional args (e.g. `taskcluster api queue task "dSlITZ4yQgmvxxAi4A8fHQ"`) not named flags
- [ ] For an endpoint with a request body, the shell example should use `<<'EOF'` redirection without `--input -`

Fixes: #8198

### DIFF
--- a/changelog/8198.md
+++ b/changelog/8198.md
@@ -1,0 +1,11 @@
+level: patch
+audience: developers
+reference: issue 8198
+---
+The API reference pages now show correct example commands for curl and the
+taskcluster CLI. The curl examples previously showed an invalid
+`Authorization: Bearer` header (Taskcluster uses Hawk authentication); they
+now explain that taskcluster-proxy should be used instead. The taskcluster CLI
+examples previously showed path arguments as named flags (`--taskId value`)
+when the CLI actually requires positional arguments; they are now shown
+correctly.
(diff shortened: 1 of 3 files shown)


lotas (approved):
Thank you

PR #8471, opened 2026-04-09
fix(worker-manager): scanner scanPrepare/scanCleanup isolation

### LINKED ISSUE #8470: Worker Scanner TypeError: `Cannot read properties of undefined (reading 'westus3')`
>TypeError
Cannot read properties of undefined (reading 'westus3')
TypeError: Cannot read properties of undefined (reading 'westus3') at AzureProvider.checkWorker (file:///app/services/worker-manager/src/providers/azure/index.js:1535:55) at process.processTicksAndRejections (node:internal/process/task_queues:104:5) at async WorkerScanner.scan (file:///app/services/worker-manager/src/worker-scanner.js:75:11) at async Iterate.handler (file:///app/services/worker-manager/src/worker-scanner.js:35:9)
Service
worker-manager
Version
99.0.3

matt-boris:
Fixes #8470.

>Worker Manager: fix a race condition where concurrent worker scanners could reset each other's tracking state via scanPrepare, causing a TypeError in the Azure provider's checkWorker. Each scanner now only calls scanPrepare/scanCleanup on the providers it is responsible for.

### DIFF
--- a/services/worker-manager/src/worker-scanner.js
+++ b/services/worker-manager/src/worker-scanner.js
@@ -23,6 +23,10 @@ export class WorkerScanner {
     this.providers = providers;
     this.monitor = monitor;
     this.providersFilter = providersFilter;
+    this._filteredProviderIds = new Set(
+      (providersFilter.value || '').split(',').filter(Boolean),
+    );
+    this._filterIsInclusive = providersFilter.cond === '=';
     this.estimator = estimator;
     this.iterate = new Iterate({
       maxFailures: 10,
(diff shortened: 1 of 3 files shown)


services/worker-manager/src/worker-scanner.js
```diff
@@ -51,8 +55,13 @@ export class WorkerScanner {
...
   }
 
+  #ownsProvider(provider) {
+    const match = this._filteredProviderIds.has(provider.providerId);
+    return this._filterIsInclusive ? match : !match;
+  }
+
   async scan() {
-    await this.providers.forAll(p => p.scanPrepare());
+    await this.providers.forAll(p => this.#ownsProvider(p) && p.scanPrepare());
```
lotas, line 64: I think the issue here is that loop takes longer than expected, so the new run starts when the old one is still running.  I think we should do something similar to what was done in the provisioner -fail when active loop running is detected. I have a PR for this ready soon

PR #8509, opened 2026-04-17
build(deps): upgrade to node 24.15.0 and yarn 4.14.1

matt-boris:
>Upgrades to Node.js v24.15.0 and yarn 4.14.1

>Yarn will not execute the postinstall scripts from third-party packages when installing the project. This change helps to reduce supply-chain risks by preventing potentially malicious scripts from running automatically.
>
>Note that you also have the ability to disable scripts on a per-package basis using `dependenciesMeta`, or to re-enable a specific script by combining `enableScripts` and `dependenciesMeta`.


.yarnrc.yml
```diff
@@ -1,3 +1,8 @@
+approvedGitRepositories:
+  - "**"
+
+enableScripts: true
+
```
matt-boris, line 5: Running `yarn` once upgraded to v4.14.1 auto-added these new (likely required?) fields.

.yarnrc.yml
```diff
@@ -1,3 +1,8 @@
+approvedGitRepositories:
+  - "**"
+
+enableScripts: true
```
petemoore: Yarn 4.14 introduced `approvedGitRepositories` and is pushing toward `enableScripts: false` as hardened defaults, specifically to close two supply-chain vectors that the current PR opts back out of.

Suggesting we set `enableScripts: false` and omit `approvedGitRepositories` entirely. No git-protocol deps exist in the repo (grepped every `package.json` and `yarn.lock`), so the Yarn default of rejecting git deps is the right behavior — no wildcard needed.

Proven out in #8515: with exactly this change applied and no other modifications, CI is green. No package in the tree actually needs install scripts.

```suggestion
enableScripts: false
```

changelog/LKbRhIo-RNqx93Z72voF4A.md
```diff
@@ -0,0 +1,4 @@
+audience: general
+level: patch
+---
+Upgrades to Node.js v24.15.0 and yarn 4.14.1
```
petemoore, line 4: If you accept the `.yarnrc.yml` suggestion above, the behavioral change for contributors is worth calling out in the changelog: anyone adding a dep that declares install scripts will now need to add a `dependenciesMeta` entry in their `package.json`.

```suggestion
Upgrades to Node.js v24.15.0 and yarn 4.14.1. Yarn install lifecycle scripts are now disabled by default to reduce supply-chain risk; packages needing `preinstall`/`install`/`postinstall` scripts must be allowlisted via `dependenciesMeta.<pkg>.built` in the consuming `package.json`.
```

petemoore (changes requested):
Thanks for driving this bump, Boris! I've got a concern about the two new `.yarnrc.yml` settings that I'd like to resolve before this lands — blocking with request-changes so auto-merge doesn't take it in the meantime.

`approvedGitRepositories: "**"` and `enableScripts: true` both opt out of Yarn 4.14's new hardened defaults, and together they permanently wildcard-open two supply-chain vectors:

- **`approvedGitRepositories: "**"`** — permits any future git-protocol dependency without review. The repo currently has zero git-protocol deps (verified by grepping every `package.json` and `yarn.lock`), so omitting the key entirely (Yarn's default — reject all) is the safer starting point.
- **`enableScripts: true`** — permits any package to run `preinstall`/`install`/`postinstall` scripts. Historically a major attack surface (shai-hulud, event-stream, xz backdoor, countless crypto-miner drops, etc.). The modern pattern is `enableScripts: false` + per-dep allowlist via `dependenciesMeta.<pkg>.built` in the consuming `package.json`.

I prototyped the lockdown in #8515 and verified CI stays green with **zero** entries in the allowlist. Every package Yarn flags during install (`ssh2`, `cpu-features`, `@newrelic/native-metrics`, `@newrelic/fn-inspect`, `protobufjs`, `@apollo/protobufjs`, `es5-ext`, `bufferutil`, `utf-8-validate`, `fsevents`, `core-js-pure`) is either an optional native speedup with a pure-JS fallback or a known no-op/ad postinstall — none are required for correctness.

Left two inline suggestions below that you can accept with one click each to roll the lockdown into this PR. I'll close #8515 as superseded once those are in.

.yarnrc.yml
matt-boris: This PR doesn't put us in any worse of a security situation than what we were already in. This could've easily been addressed in a future PR, but I'm happy to do so in this one.

matt-boris: We've also no clue if this breaks our release process.

changelog/LKbRhIo-RNqx93Z72voF4A.md
matt-boris: It's only postinstall scripts that this affects https://yarnpkg.com/configuration/yarnrc#enableScripts

matt-boris:
Running a staging release here https://community-tc.services.mozilla.com/tasks/PDz3n7zSRZeMBZwsp-VZ3A

Edit: success ✅

petemoore (approved):
Looks good, thanks!

changelog/adNEk4DQQW2DonOBdgZH2w.md
```diff
@@ -0,0 +1,6 @@
+audience: developers
+level: patch
+---
+Yarn will not execute the postinstall scripts from third-party packages when installing the project. This change helps to reduce supply-chain risks by preventing potentially malicious scripts from running automatically.
```
lotas, line 4: awesome!

PR #8512, opened 2026-04-18
Fix nginx log dir permissions for non-root container

### LINKED ISSUE #8270: References nginx logs with non-root issue
But I noticed references is still affected probably by some other default values: (likely existed before this PR also)

```
references-1  | nginx: [alert] could not open error log file: open() "/var/lib/nginx/logs/error.log" failed (13: Permission denied)
references-1  | 2026/02/10 11:32:14 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /app/infrastructure/references/nginx.conf:2
```

_Originally posted by @lotas in https://github.com/taskcluster/taskcluster/issues/8265#issuecomment-3877069547_
            

Base image needs to create /var/log/nginx/logs and grant proper permissions for reference's nginx server

nitishagar:
Github Bug/Issue: Fixes #8270

## Summary
- Fix permission error where nginx cannot open its default error log at `/var/lib/nginx/logs/error.log` when the container runs as non-root (UID 1000)
- Add `RUN mkdir -p /var/lib/nginx/logs && chown -R 1000:1000 /var/lib/nginx` to the Dockerfile before `USER 1000`
- nginx opens the compiled-in default log path *before* reading the config that redirects to stderr — this ensures the directory is writable at startup

## Test plan
- [ ] `docker build -t tc-test .` completes without error
- [ ] `docker compose up references` starts with no `Permission denied` alert for `/var/lib/nginx/logs/error.log`
- [ ] `/__heartbeat__` returns `{}` from the references container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,6 +71,9 @@ WORKDIR /app
 # Create /references directory for references service with proper ownership
 RUN mkdir -p /references && chown 1000:1000 /references
 
+# Allow nginx to write its default log path when running as non-root
+RUN mkdir -p /var/lib/nginx/logs && chown -R 1000:1000 /var/lib/nginx
+
 # use non-root, node user
 # https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user
 USER 1000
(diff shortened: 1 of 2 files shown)


matt-boris (approved):
LGTM, thanks for addressing 👍🏻

PR #8514, opened 2026-04-18
Add gzip compression middleware to API layer

### LINKED ISSUE #8347: Add compression to the API layer
Currently API is not using compression, and sends json as is. 
Some of those are very large, hundreds of Kb and sometimes Mb

It would probably be good to use compression with some threshold to skip smaller payloads, but force it for larger:

```
app.use(compression({
  level: 6, 
  threshold: 1024, 
}));
```

nitishagar:
Github Bug/Issue: Fixes #8347

## Summary
- Add the `compression` npm package to `@taskcluster/lib-api` and register it as the first middleware on every API router
- Uses `level: 6` (default gzip) and `threshold: 1024` (skip compression for responses < 1 KB)
- All microservices automatically benefit without per-service changes since they all use `lib-api`
- No double-compression risk with `web-server` — `compression()` skips responses that already have `Content-Encoding` set

## Test plan
- [ ] `cd libraries/api && yarn test` passes (includes new compression tests)
- [ ] `cd services/queue && yarn test` passes (spot-check a service using lib-api)
- [ ] `yarn lint` passes
- [ ] Large API responses (e.g. listWorkerPools) return `Content-Encoding: gzip` when client sends `Accept-Encoding: gzip`

### DIFF
--- a/changelog/issue-8347.md
+++ b/changelog/issue-8347.md
@@ -1,0 +1,5 @@
+audience: deployers
+level: minor
+reference: issue 8347
+---
+HTTP responses from Taskcluster API services are now gzip-compressed when the client sends `Accept-Encoding: gzip`. Compression is applied at the `@taskcluster/lib-api` router level with a 1 KB threshold, so small payloads are sent uncompressed.
--- a/libraries/api/src/api.js
+++ b/libraries/api/src/api.js
@@ -1,3 +1,4 @@
+import compression from 'compression';
 import express from 'express';
 import assert from 'assert';
 import libUrls from 'taskcluster-lib-urls';
@@ -83,6 +84,8 @@ export default class API {
 
     // Create router
     const router = express.Router({ caseSensitive: true });
+
+    router.use(compression());
 
     // Allow CORS requests to the API
     if (allowedCORSOrigin) {
--- a/libraries/api/test/compression_test.js
+++ b/libraries/api/test/compression_test.js
@@ -1,0 +1,120 @@
+import assert from 'assert';
+import http from 'http';
+import { gzipSync, gunzipSync } from 'zlib';
+import request from 'superagent';
+import { APIBuilder } from '../src/index.js';
+import helper from './helper.js';
+import libUrls from 'taskcluster-lib-urls';
+import testing from '@taskcluster/lib-testing';
+
+suite(testing.suiteName(), function() {
+  const u = path => libUrls.api(helper.rootUrl, 'test', 'v1', path);
+
+  const builder = new APIBuilder({
+    title: 'Test Api',
+    description: 'Test api for compression',
+    serviceName: 'test',
+    apiVersion: 'v1',
+    context: [],
+  });
+
+  // Endpoint returning a payload large enough to exceed the 1024-byte threshold
+  builder.declare({
+    method: 'get',
+    route: '/large-payload',
+    name: 'largePayload',
+    scopes: null,
+    title: 'Large Payload',
+    category: 'API Library',
+    stability: APIBuilder.stability.stable,
+    description: 'Returns a large JSON payload',
+  }, function(req, res) {
+    res.reply({ data: 'x'.repeat(2000) });
+  });
+
+  // Endpoint returning a payload below the 1024-byte threshold
+  builder.declare({
+    method: 'get',
+    route: '/small-payload',
+    name: 'smallPayload',
+    scopes: null,
+    title: 'Small Payload',
+    category: 'API Library',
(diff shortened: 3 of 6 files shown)


matt-boris:
Hi @nitishagar, will you please rebase your changes with `main` and we'll review?

nitishagar:
Rebased on `main` — ready for review. Thanks @matt-boris!

matt-boris:
You're going to need to run `yarn` (or `yarn install`) at the repo root and commit up the `yarn.lock` file change.

And please update the branch again with `main`

nitishagar:
Updated — merged latest `main` and committed the regenerated `yarn.lock`. Thanks!

petemoore (changes requested):
The approach is sound — drop compression in once at lib-api and every service benefits — but a few things worth addressing before merging.

### 1. Redundant compression in `web-server`
`services/web-server/src/servers/createApp.js:81` already does `app.use(compression())`. With this PR, every web-server API request now passes through compression middleware twice — once at the app level, once at the lib-api router level.

It won't double-compress the bytes (the inner middleware sets `Content-Encoding`, the outer bails out), but it is redundant work — wrapping `res.write`/`res.end` and re-checking headers on every request. The PR description acknowledges no double-compression risk, but the redundancy itself isn't addressed. Either drop the `app.use(compression())` from `createApp.js` here, or open a follow-up — but please at least call it out.

### 2. Test coverage gaps
The two tests check headers only. Worth adding:
- **Client without `Accept-Encoding: gzip`** → response should not be compressed. Currently we only test the threshold branch, not the content-negotiation branch.
- **Compressed body actually decodes** → superagent transparently decompresses, so a corrupted body wouldn't be caught. Easy: assert `res.body.data.length === 2000`.
- **Pre-set `Content-Encoding` is preserved** → directly verifies the "no double compression" claim. Relevant to the manually-gzipped profile endpoint at `services/web-server/src/api.js:173-178`.

### 3. Configuration values are just the defaults
```js
compression({ level: 6, threshold: 1024 })
```
Both `level: 6` (zlib default) and `threshold: 1024` are the package defaults. If we don't intend to deviate, `compression()` is clearer and tracks any future upstream default changes. If we *do* intend to pin them, a one-line comment explaining why would help.

### Minor / non-blocking
- `router.use(compression(...))` is placed before `corsHeaders` and `cacheHeaders`. Probably fine (CORS preflight is small, cache-headers don't touch the body) but worth a quick sanity check on header-order interactions.
- No streaming/SSE endpoints currently use lib-api routers, so the middleware's buffering won't break anything today — but worth keeping in mind for future endpoints.

### Looks good
- `compression` correctly placed in `dependencies`.
- Default filter skips already-compressed content types, so artifact-style payloads aren't wastefully re-compressed.
- The manually-gzipped profile endpoint will be safely skipped by the new middleware.
- Changelog entry is clear and correctly tagged.

nagarwal-godaddy:
Thanks @petemoore — all addressed:

1. **Redundant compression in web-server**: dropped \`app.use(compression())\` from \`services/web-server/src/servers/createApp.js\`. Since \`/graphql\` doesn't route through the lib-api router, I added \`compression()\` specifically at the \`/graphql\` mount point in \`main.js\`. Everything else either goes through lib-api (now compressed) or is small (login redirects).
2. **Test coverage**: added three new tests in \`libraries/api/test/compression_test.js\`:
   - Client without \`Accept-Encoding: gzip\` → no compression, body round-trips
   - Large payload → asserts decoded \`body.data.length === 2000\` (catches corrupted bodies)
   - Pre-set \`Content-Encoding\` endpoint → fetches raw bytes via \`http.get\`, gunzips once, verifies plaintext (catches any double-compression)
3. **Configuration defaults**: replaced \`compression({ level: 6, threshold: 1024 })\` with bare \`compression()\` since both were defaults.

Also fixed two CI failures uncovered by the rebase:
- \`meta-tests\`: removed the \`dependencies\` block from \`libraries/api/package.json\` (monorepo forbids workspace-local deps; \`compression\` is already in root \`package.json\`)
- \`service-github\`: 16 badge tests asserted on exact \`content-length\`, which becomes \`undefined\` under chunked encoding once compression kicks in — switched to \`res.body.length\` (decoded size)

Squashed into a single commit. Local runs: lib-api 142✓, web-server 131✓, github 54✓.

lotas (approved):
lgtm

petemoore (approved):
lgtm, thanks

nitishagar:
@petemoore Any way to re-run this multi-user ubuintu test?

PR #8515, opened 2026-04-18
build(deps): upgrade to node 24.15.0 and yarn 4.14.1, harden install scripts

petemoore:
## Summary

Supersedes #8509. Bumps to Node 24.15.0 and Yarn 4.14.1 (same as that PR), but
instead of opting out of Yarn 4.14's new hardened defaults, this PR opts *in*:

- **`enableScripts: false`** — skip lifecycle scripts for installed packages
  by default. Supply-chain attacks have increasingly ridden in through
  `postinstall`/`preinstall` hooks; disabling them globally and allowlisting
  per-dependency via `dependenciesMeta.<pkg>.built` in the consuming
  `package.json` is the modern hardened posture.

- **`approvedGitRepositories` is omitted** entirely — the repo has zero
  git-protocol dependencies today (verified by grepping every `package.json`
  and `yarn.lock`). Without this key, Yarn refuses any future git dep unless
  it's added to a narrow allowlist, which is the right default. #8509 set it
  to ``"**"`` — a permanent wildcard allowing any git repo.

## Why not #8509?

#8509 adds:

```yaml
approvedGitRepositories:
  - "**"
enableScripts: true
```

Both are opt-outs of Yarn 4.14's new defaults. The changelog doesn't mention
this, the PR description doesn't mention this, and the change is buried in a
300-line mostly-yarn-binary diff that auto-merge will land without scrutiny.
We should lock things down by default and allowlist explicitly.

## What did Yarn flag during install?

Running \`yarn install\` at root + \`ui\` + \`clients/client\` with scripts disabled
surfaced these packages requesting scripts. Each was assessed:

| Package | Script purpose | Required? |
|---|---|---|
| \`@apollo/protobufjs\` | Regenerates minimal.js | No (shipped) |
| \`@newrelic/fn-inspect\` | Native addon build | Optional (monitoring) |
| \`@newrelic/native-metrics\` | Native addon build | Optional (monitoring) |
| \`cpu-features\` | Native CPU feature detection (ssh2) | Optional (speedup) |
| \`es5-ext\` | Ad/marketing message | No |
| \`protobufjs\` | Regenerates minimal.js | No (shipped) |
| \`ssh2\` | Native SSH bindings | Optional (pure-JS fallback) |
| \`bufferutil\` | Native \`ws\` speedup | Optional (pure-JS fallback) |
| \`core-js-pure\` | Ad/marketing message | No |
| \`fsevents\` | macOS file watching | Optional (fallback) |
| \`utf-8-validate\` | Native \`ws\` speedup | Optional (pure-JS fallback) |

**None are required for correctness.** Started with zero allowlist entries.
If CI surfaces a case where a script is actually needed, we'll add the
specific package to \`dependenciesMeta\` with justification.

## Test plan

- [x] \`yarn install --immutable\` at root succeeds
- [x] \`yarn install --immutable\` in \`clients/client\` succeeds
- [x] \`yarn install --immutable\` in \`ui\` succeeds
- [x] \`yarn lint\` passes
- [x] Representative library tests (\`libraries/api\`, \`libraries/pulse\`) pass
- [x] CI green across all services, workers, clients, Docker build
- [x] UI browser-test CI task passes
- [x] Dockerfile build succeeds

Closed in favour of the inline suggestions on #8509 — see the [review there](https://github.com/taskcluster/taskcluster/pull/8509#pullrequestreview-4134329951).


petemoore:
Closing — this PR served its purpose as a proof-of-concept that the Yarn install-script lockdown works with zero allowlist entries (CI went green with 71/71 checks passing).

The lockdown is now being proposed back to #8509 as two inline review suggestions (`.yarnrc.yml` and `changelog/`), so the fix lands on Boris's bump PR directly rather than competing with it.

Closed in favour of the review on #8509.

PR #8516, opened 2026-04-20
feat(client-rust): Add `content_length` parameter to AWF

### LINKED ISSUE #8869: client-rust: Add `content_length` parameter to AWF
There is currently no way for the `AsyncWriterFactor` to know what size the expected artifact will be. This is relevant for example if you want to build a interactive progress bar that displays the download status of the artifact.

See https://github.com/taskcluster/taskcluster/pull/8516.

michaelvanstraten:
Github Bug/Issue: Fixes #8869


petemoore:
@michaelvanstraten Many thanks for the draft PR. Are you still working on it? Normally we would not provide initial feedback on a PR until it is out of draft. Thanks.

michaelvanstraten:
Hi @petemoore, sorry for the slow response.

The reason why I kept this as a draft is because I was unsure if we would end up settling on the `taskcluster` crate and since this is a breaking change I did not want to get things moving on your side. 

I am more certain that we are going to continue to use the crate now.

Do you think this breaking change would be acceptable. It is kind of specific to our use case and I would understand if you would like to kept the current API.

clients/client-rust/download/src/factory.rs
```diff
@@ -21,7 +24,13 @@ pub struct CursorWriterFactory<T>(Cursor<T>);
 
 #[async_trait]
 impl AsyncWriterFactory for CursorWriterFactory<Vec<u8>> {
-    async fn get_writer<'a>(&'a mut self) -> Result<Box<dyn AsyncWrite + Unpin + 'a>> {
+    async fn get_writer<'a>(
+        &'a mut self,
+        content_length: Option<u64>,
+    ) -> Result<Box<dyn AsyncWrite + Unpin + 'a>> {
+        if let Some(content_length) = content_length {
+            self.0.get_mut().reserve(content_length as usize);
```
Eijebong: I think this line is wrong in 2 different ways.

First of, if the content length is bogus (i.e a very large number), you just oom. Given this is used pretty much exclusively to fetch stuff from fxci it doesn't *really* matter but I just wanted to mention it.

But most importantly you're clearing *after* reserve, which AFAICT, on a retry would double the allocation.

IMO you should just drop the reserve. Rust vecs allocate by doubling, which means over a 100MB artifact you'd get ~14 reallocs (`math.log2(100*2**20) - math.log2(8*2**10) == 13.643856189774723`, assuming worst case scenario where the buffer from tokio starts at 8kB). The cost of that is going to be absolutely dwarfed by network and disk I/O.

changelog/issue-8869.md
```diff
@@ -0,0 +1,5 @@
+audience: general
+level: minor
```
Eijebong: This should be major, this is a breaking change

changelog/issue-8869.md
```diff
@@ -0,0 +1,5 @@
...
+reference: issue 8869
+---
+The client-rust `AsyncWriterFactory` now needs to be `Sync` and the `get_writer` method now takes an optional `content_length` parameter that indicates what size the writer should expect.
```
Eijebong: Did you mean `Send` here?

clients/client-rust/download/src/geturl.rs
```diff
@@ -40,6 +42,11 @@ pub(crate) async fn get_url(
         Ok(res) => res,
     };
 
+    let mut writer = match writer_factory.get_writer(res.content_length()).await {
```
Eijebong: This is an undocumented change of behavior. Before, the writer factory was called on every attempt, truncating the file even if the request ended up failing. Now it only calls `get_writer` when the HTTP request actually goes through and gets a 200. Not saying this is wrong but it might be worth a mention in the changelog/trait doc for `AsyncWriterFactory::get_writer` (and maybe explain why the change in the commit message for future you when you're reading this 6y from now to remember why you made that decision).

clients/client-rust/download/src/hashing.rs
```diff
@@ -52,6 +42,26 @@ impl<'f, AWF: AsyncWriterFactory> HasherAsyncWriterFactory<'f, AWF> {
...
 }
 
+#[async_trait]
+impl<'f, AWF> AsyncWriterFactory for HasherAsyncWriterFactory<'f, AWF>
+where
+    AWF: AsyncWriterFactory + Send,
+{
+    async fn get_writer<'a>(
+        &'a mut self,
+        content_length: Option<u64>,
```
Eijebong: Might be worth documenting here that this being `None` is normal and is going to happen for objects that have a non `identity` `Content-encoding`.

Reqwest sends `Accept-Encoding: gzip,br` in our case (because we enable both features, but they might also be enabled because something else in the tree would enable them, rust unifies features across the full compile. Which means it can technically send more encodings). When it gets something back with a non identity encoding, it yeets the content-length into oblivion if there's one (see https://docs.rs/reqwest/0.13.4/reqwest/struct.ClientBuilder.html#method.gzip for example)

You can easily verify this:

```
> curl -H "Accept-encoding: gzip" -L -v -o /dev/null https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/Ex3-sQI8Ta2zlU4Mv8LQiw/runs/0/artifacts/public%2Flogs%2Flive_backing.log

< HTTP/2 200
< content-type: text/plain; charset=utf-8
< x-guploader-uploadid: AKhXX7K15AKR6MooPvONYEO76gKAo60FWxa2MLffCEvNJ11s4Ytp8q6GfiR7ELK9F4bAR_eH
< date: Tue, 14 Jul 2026 19:45:06 GMT
< cache-control: public,max-age=604800
< last-modified: Thu, 09 Jul 2026 23:29:55 GMT
< etag: "5e0eb74e56ae8ef964da01209628d20e"
< vary: Accept-Encoding
< x-goog-generation: 1783639795164064
< x-goog-metageneration: 1
< x-goog-stored-content-encoding: gzip
< x-goog-stored-content-length: 72169
< content-encoding: gzip
< x-goog-hash: crc32c=L1dQpw==
< x-goog-hash: md5=Xg63Tlaujvlk2gEglijSDg==
< x-goog-storage-class: STANDARD
< accept-ranges: bytes
< content-length: 72169
< access-control-allow-origin: *
< access-control-expose-headers: *
< server: UploadServer
< x-cache-status: miss
< alt-svc: clear
```

Vs

```
> curl -H "Accept-encoding: gzip" -L -v -o /dev/null https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/Ex3-sQI8Ta2zlU4Mv8LQiw/runs/0/artifacts/public%2Fbuild%2Fsccache.tar.zst

< HTTP/2 200
< content-type: application/zstd
< x-guploader-uploadid: AKhXX7L4ylv27Qd0KcIFM21J_0_VXymcsw_Mef7IBA0-Mh9zfgMcC67bM-5NzbJFy9uDC_sy
< date: Tue, 14 Jul 2026 19:43:28 GMT
< cache-control: public,max-age=604800
< last-modified: Thu, 09 Jul 2026 23:29:54 GMT
< etag: "7a63e5fce3ba4c62567f5c40d65d8669"
< x-goog-generation: 1783639793968071
< x-goog-metageneration: 1
< x-goog-stored-content-encoding: identity
< x-goog-stored-content-length: 13957410
< x-goog-hash: crc32c=bFBp7Q==
< x-goog-hash: md5=emPl/OO6TGJWf1xA1l2GaQ==
< x-goog-storage-class: STANDARD
< accept-ranges: bytes
< content-length: 13957410
< access-control-allow-origin: *
< access-control-expose-headers: *
< server: UploadServer
< x-cache-status: miss
< alt-svc: clear
<

```

Repro with:

Cargo.toml
```
[package]
name = "foo"
version = "0.1.0"
edition = "2024"

[dependencies]
reqwest = { version = "0.13.4", features = ["gzip"] }
tokio = { version = "1.52.3", features = ["full"] }
```

```rs
const GZIPPED: &str =
"https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/Ex3-sQI8Ta2zlU4Mv8LQiw/runs/0/artifacts/public%2Flogs%2Flive_backing.log";
const IDENTITY: &str =
"https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/Ex3-sQI8Ta2zlU4Mv8LQiw/runs/0/artifacts/public%2Fbuild%2Fsccache.tar.zst";

async fn get(url: &str) {
    let res = reqwest::get(url).await.unwrap();
    let cl = res.content_length();
    let actual = res.bytes().await.unwrap().len();
    println!("{url} content-length = {cl:?} actual len = {actual}");
}

#[tokio::main]
async fn main() {
    get(GZIPPED).await;
    get(IDENTITY).await;
}
```

Result:

```
https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/Ex3-sQI8Ta2zlU4Mv8LQiw/runs/0/artifacts/public%2Flogs%2Flive_backing.log content-length = None actual len = 686569
https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/Ex3-sQI8Ta2zlU4Mv8LQiw/runs/0/artifacts/public%2Fbuild%2Fsccache.tar.zst content-length = Some(13957410) actual len = 13957410
```

Eijebong (changes requested):
Would be nice to add some tests too

changelog/issue-8869.md
michaelvanstraten: Hm, I thought that in rust you can do a breaking change in a minor version but okay. I'll change it to major.

changelog/issue-8869.md
michaelvanstraten: Yes :=)

clients/client-rust/download/src/factory.rs
michaelvanstraten: Sure thing, I just added it here to have some potential use of the parameter.

clients/client-rust/download/src/hashing.rs
michaelvanstraten: Would you be okay with also reading some of the less standard headers such as `x-goog-stored-content-length`?

changelog/issue-8869.md
Eijebong: You can if the version is `0.X.Y`, the the minor version changing is indicative of a breaking change. This is a very rust specific thing and we're way over `v0` in this repo :p

clients/client-rust/download/src/hashing.rs
Eijebong: I don't think that'll help. It's the compressed size too, google doesn't send the uncompressed one. For that one example there's a factor 9.5x between those and the writer will only ever get decompressed data. For your use case it means you'd be at 100% progress when you're in fact 10% into the download. And for the case where it's identity encoded then it serves no purpose since content-length doesn't get dropped.

changelog/issue-8869.md
michaelvanstraten: Okay, thank you for the clarification.

clients/client-rust/download/src/geturl.rs
michaelvanstraten: I added a comment in the commit message.

clients/client-rust/download/src/hashing.rs
michaelvanstraten: Okay

michaelvanstraten:
@Eijebong do you mind if I fix some of the doc comment warnings in this PR or do you want a second PR for this?

changelog/issue-8869.md
```diff
@@ -0,0 +1,6 @@
+audience: general
```
Eijebong: Nit:

```suggestion
audience: users
```

I think general might be too broad? Wish we had a `client-users` audience right there 🤷

clients/client-rust/download/src/factory.rs
```diff
@@ -11,7 +11,14 @@ use tokio::io::{AsyncSeekExt, AsyncWrite, AsyncWriteExt};
 pub trait AsyncWriterFactory {
     /// Get a fresh [AsyncWrite] object, positioned at the point where downloaded data should
     /// be written.
-    async fn get_writer<'a>(&'a mut self) -> Result<Box<dyn AsyncWrite + Unpin + 'a>>;
+    ///
+    /// The `content_length` parameter will have the content length of the artifact, if it is known.
+    /// See [content_length](reqwest::Response::content_length) on the `reqwest` response object
+    /// for more information on the matter.
```
Eijebong:
```suggestion
    /// The `content_length` parameter holds the response's `Content-Length` if it's known (see
    /// [content_length](reqwest::Response::content_length)). It's only a hint intended for sizing
    /// the writer, and must not be relied upon. It might be `None` for chunked or decompressed
    /// responses, and for a compressed body it reflects the compressed size, so the number of bytes
    /// actually written may differ.
```

I think it's worth hammering here that it might be missing or not match the actual written size.

clients/client-rust/download/src/geturl.rs
```diff
@@ -56,11 +63,64 @@ pub(crate) async fn get_url(
...
+
+        match get_url(&server.data_url(), &mut factory).await {
+            RetriableResult::Ok(_) => {}
+            _ => panic!("expected a successful fetch"),
+        }
+
+        assert_eq!(factory.content_lengths, vec![Some(12)]);
+        assert_eq!(&factory.buf, b"hello, world");
+    }
+
```
Eijebong:
```suggestion

#[tokio::test]
async fn get_writer_skipped_on_retriable_error_then_called_on_retry() {
    let server = FakeDataServer::new(false, &[500, 200]);
    let mut factory = SpyWriterFactory::default();

    match get_url(&server.data_url(), &mut factory).await {
        RetriableResult::Retriable(_) => {}
        _ => panic!("expected a retriable failure"),
    }
    assert!(factory.content_lengths.is_empty());
    assert!(factory.buf.is_empty());

    match get_url(&server.data_url(), &mut factory).await {
        RetriableResult::Ok(_) => {}
        _ => panic!("expected a successful fetch"),
    }
    assert_eq!(factory.content_lengths, vec![Some(12)]);
    assert_eq!(&factory.buf, b"hello, world");
}

```

I'd add a test that encodes the new behavior wrt the factory only being called when a response actually arrives. Test name is a mouthful, feel free to name it something better 🤣

Eijebong (commented):
Just a few things but this in a better shape now.

changelog/issue-8869.md
michaelvanstraten: I think client users is good. Or maybe client-consumers.

Eijebong (approved):
Thank you

michaelvanstraten:
@Eijebong could you merge this, I cant.

Eijebong:
Right. I set it to automerge but it wasn't rebased and required an extra check. I rebased and I'll try to make sure it goes through this time, sorry

Eijebong (approved):

Eijebong:
qdfjksdhqfmkjqshdfmksjqdhfmkjh I pushed main to this branch instead of the right patch when rebasing T_T

Issue #8059, opened 2025-11-04
ARM Deployment related errors


lotas:
Needs investigation in community-tc, several errors happen after and during deployments:

- [  ] `The Resource 'Microsoft.Compute/virtualMachines/vm-i9wihqfnrw4vcckykwbqqbaxs4sdr96wxue' under resource group 'rg-tc-eng-worker-manager-VMs' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix (Code: ResourceNotFound)`
- [ ] `Unable to edit or replace deployment ‘deploy-qvyxrtyfrlg8lt1exm4tjwm5t6kqqss6s4f’: previous deployment from ‘11/4/2025 1:44:51 PM’ is still active (expiration time is ‘11/11/2025 1:43:56 PM’, correlation id is ‘edf36736-edd7-430c-98ac-8b8247c68f4a’). Please see https://aka.ms/arm-deploy for usage details.`
- [ ] `Nic(s) in request is reserved for another Virtual Machine for 180 seconds. Please provide another nic(s) or retry after 180 seconds. Reserved VM: /subscriptions/8a205152-b25a-417f-a676-80465535a6c9/resourceGroups/rg-tc-eng-worker-manager-VMs/providers/Microsoft.Compute/virtualMachines/vm-b0hpjhckrduvk0kbltvaqgdfkdcktdryijr”`

https://community-tc.services.mozilla.com/worker-manager/proj-taskcluster%2Fgw-windows-2022-gui/errors

<img width="1776" height="482" alt="Image" src="https://github.com/user-attachments/assets/24173429-e1d8-4c92-9c2d-60f4905d7477" />

Some of those are likely due to the related https://github.com/taskcluster/community-tc-config/issues/965

PR #8518, opened 2026-04-21
feat(w-m): Fix azure stuck resource deletion

### LINKED ISSUE #8517: Workers in STOPPING accumulate indefinitely when cloud resources are removed out-of-band
**Describe the bug**

Workers in the Azure provider get stuck in `STOPPING` when their backing Azure resource is deleted outside of worker-manager's control, usually from Spot preemption, ARM cascade-delete, or manual cleanup. Those workers still count toward `maxCapacity` in the estimator, so the pool fills up with dead entries that block new provisioning.

On `gecko-t/win11-64-25h2` (FXCI) at `2026-04-20T21:55:29Z`, 1,185 of 1,185 workers in `state=stopping` had no matching Azure VM in the pool's resource group. The oldest had `lastModified=2026-04-20T16:00:46.879Z`, stuck for 5h 55m. `maxCapacity` is 2000, so those 1,185 phantoms take up 59% of the cap. `estimator.js` computes `totalNonStopped = existingCapacity + stoppingCapacity` and clamps new provisioning against `maxCapacity`, so only 1 new VM was being requested while 1,680 tasks waited.

This is a bug in the shared `deprovisionResource` helper in `services/worker-manager/src/providers/azure/index.js`, not a pool-specific or VM-specific symptom. When a resource still has an `id`, the helper goes straight to `client.beginDelete(...)`. The `get()` branch already treats `404` as "resource is already gone; mark deleted and continue", but the `beginDelete()` branch only handles `409 "previous deployment still active"`. When Azure has already removed the resource, `beginDelete()` returns `404`, the error re-throws, the outer `deprovisionResources` handler catches it, logs a `deletion-error`, and the worker stays in `STOPPING`.

Because this logic lives in the shared helper, the missing `404` handling affects every resource type that uses it: `vm`, `nic`, `ip`, `disks[]`, and ARM `deployment` cleanup.

ARM templates with `deleteOption: 'Delete'` amplify the bug but are not the root cause. A single Spot preemption can cascade one VM delete into VM + NIC + IP + disk removal on the Azure side, so every later `deprovisionResource` call for that worker hits the missing `beginDelete(...)->404` path.

**To Reproduce**

Works on any FXCI Azure pool with active Spot churn. `gecko-t/win11-64-25h2` is a reliable example.

1. Check the pool's own view of its state:
   ```bash
   taskcluster api workerManager workerPool gecko-t/win11-64-25h2 \
     | jq '{currentCapacity, runningCount, stoppingCount, requestedCount}'
   ```
2. Paginate `listWorkersForWorkerPool` to collect every `stopping` workerId into `tc_stopping.txt` (full script below).
3. Query Azure for the VMs present in the pool's resource group:
   ```bash
   az vm list --resource-group rg-tc-gecko-t-win11-64-25h2 --query "[].name" -o tsv | sort > az_vms.txt
   ```
4. Diff:
   ```bash
   echo "Ghosts: $(comm -23 <(sort tc_stopping.txt) az_vms.txt | wc -l)"
   echo "Real stoppers: $(comm -12 <(sort tc_stopping.txt) az_vms.txt | wc -l)"
   ```
5. A large fraction of TC-tracked `stopping` workers will have no matching Azure VM. Their `lastModified` timestamps are hours old, and the count does not decrease.

Full reproducer:
```bash
set -euo pipefail
POOL="gecko-t/win11-64-25h2"
RG="rg-tc-gecko-t-win11-64-25h2"
TMP=$(mktemp -d)

TOKEN=""
: > "$TMP/tc_stopping.txt"
while :; do
  if [ -z "$TOKEN" ]; then
    RESP=$(taskcluster api workerManager listWorkersForWorkerPool "$POOL" --limit=500)
  else
    RESP=$(taskcluster api workerManager listWorkersForWorkerPool "$POOL" --limit=500 --continuationToken "$TOKEN")
  fi
  echo "$RESP" | python3 -c "
import json, sys
d = json.load(sys.stdin)
for w in d.get('workers', []):
    if w.get('state') == 'stopping':
        print(w['workerId'])
" >> "$TMP/tc_stopping.txt"
  TOKEN=$(echo "$RESP" | python3 -c "import json,sys; print(json.load(sys.stdin).get('continuationToken',''))")
  [ -z "$TOKEN" ] && break
done

sort -u "$TMP/tc_stopping.txt" > "$TMP/tc.sorted"
az vm list -g "$RG" --query "[].name" -o tsv | sort -u > "$TMP/az.sorted"

echo "TC stopping workers: $(wc -l < "$TMP/tc.sorted")"
echo "Azure VMs in RG:     $(wc -l < "$TMP/az.sorted")"
echo "Ghosts:              $(comm -23 "$TMP/tc.sorted" "$TMP/az.sorted" | wc -l)"
echo "Real stoppers:       $(comm -12 "$TMP/tc.sorted" "$TMP/az.sorted" | wc -l)"
```

**Expected behavior**

When `client.beginDelete(...)` returns `404`, the resource is already gone. `deprovisionResource` should treat that exactly the same way it already treats `client.get(...)` returning `404`: cleanup for that resource is complete.

On `404` from `beginDelete(...)`, the provider should:

- clear `operation`
- set `id = false`
- set `deleted = true`
- return `true`

This change should be made in the shared `deprovisionResource` helper so it applies uniformly to `vm`, `nic`, `ip`, `disks[]`, and ARM `deployment` cleanup.

With that change, STOPPING workers whose Azure resources were removed out-of-band can continue through `deprovisionResources` and transition to `STOPPED` on later scans.

**Acceptance criteria**

- `beginDelete(...)` `404` no longer leaves workers stuck in `STOPPING`
- affected workers drain to `STOPPED` without manual DB cleanup
- unit tests cover the delete-branch `404` path on the shared helper
- the fix is implemented in `deprovisionResource`, not in a VM-only special case

**Non-goals / separate follow-up**

- Changing estimator handling of `stoppingCapacity` in [#8161](https://github.com/taskcluster/taskcluster/issues/8161) is not required to fix this bug. It may be a separate mitigation for capacity pressure, but it does not repair the reap path.
- Scanner-side expiry hardening is worth considering as defense in depth, but it is not the root fix for this issue.

**Taskcluster version**

99.1.0 (commit `c76d61efe4bdc1a05bcec848739cd41ebf061f01`)

**Screenshots**

N/A. The worker-manager view for [gecko-t/win11-64-25h2](https://firefox-ci-tc.services.mozilla.com/worker-manager/gecko-t%2Fwin11-64-25h2) shows a large and growing `stoppingCount` relative to `runningCount`, but the definitive signal is the TC-vs-Azure diff above.

- OS: N/A (server-side bug in the worker-manager service, `azure2` provider)
- Browser if applicable: N/A
- Version if applicable: N/A

Observed against FXCI: `firefox-ci-tc.services.mozilla.com`, Azure subscription `108d46d5-fe9b-4850-9a7d-8c914aa6c1f0`.

**Additional context**

Related issues:

- [#8161 — Azure provider stopping capacity](https://github.com/taskcluster/taskcluster/issues/8161): separate mitigation idea for estimator behavior; not required to fix this bug.
- [#8441 — Azure ARM Deployment - remove on register](https://github.com/taskcluster/taskcluster/issues/8441): different symptom (the 800-deployments-per-RG limit).
- [#8387 — Azure provider hits 800 ARM deployment limit](https://github.com/taskcluster/taskcluster/issues/8387): different symptom in the same deployment-lifecycle area.
- [#5291 — Removing Azure workers takes 10-30 minutes](https://github.com/taskcluster/taskcluster/issues/5291): nearest historical match for reap lag, but scoped to bounded delay. The bug here is unbounded.

Code references (pinned to commit `c76d61e`):

- State machine: `removeWorker` at [`azure/index.js:1993-2010`](https://github.com/taskcluster/taskcluster/blob/c76d61efe4bdc1a05bcec848739cd41ebf061f01/services/worker-manager/src/providers/azure/index.js#L1993-L2010) transitions `REQUESTED|RUNNING -> STOPPING`.
- STOPPING handler in `checkWorker`: [`azure/index.js:1650-1653`](https://github.com/taskcluster/taskcluster/blob/c76d61efe4bdc1a05bcec848739cd41ebf061f01/services/worker-manager/src/providers/azure/index.js#L1650-L1653) calls `deprovisionResources`.
- Shared deprovision helper: [`azure/index.js:1889-1989`](https://github.com/taskcluster/taskcluster/blob/c76d61efe4bdc1a05bcec848739cd41ebf061f01/services/worker-manager/src/providers/azure/index.js#L1889-L1989).
- Existing query-branch `404` handling: [`azure/index.js:1932-1950`](https://github.com/taskcluster/taskcluster/blob/c76d61efe4bdc1a05bcec848739cd41ebf061f01/services/worker-manager/src/providers/azure/index.js#L1932-L1950).
- Missing delete-branch `404` handling: [`azure/index.js:1960-1972`](https://github.com/taskcluster/taskcluster/blob/c76d61efe4bdc1a05bcec848739cd41ebf061f01/services/worker-manager/src/providers/azure/index.js#L1960-L1972).
- ARM outputs populating `vm.id`, `nic.id`, `ip.id`, and `disks[].id`, which routes the deprovision path through the vulnerable delete branch: [`azure/index.js:692-698`](https://github.com/taskcluster/taskcluster/blob/c76d61efe4bdc1a05bcec848739cd41ebf061f01/services/worker-manager/src/providers/azure/index.js#L692-L698).
- Estimator's use of `stoppingCapacity` against `maxCapacity`: [`estimator.js:28, 38, 123`](https://github.com/taskcluster/taskcluster/blob/c76d61efe4bdc1a05bcec848739cd41ebf061f01/services/worker-manager/src/estimator.js#L28).

Live data from `gecko-t/win11-64-25h2`, snapshot at `2026-04-20T21:55:29Z`:

| metric | value |
|---|---|
| `maxCapacity` | 2000 |
| `currentCapacity` | 1999 |
| `runningCount` | 813 |
| `requestedCount` | 1 |
| `stoppingCount` | 1185 |
| Azure VMs in `rg-tc-gecko-t-win11-64-25h2` | 791 |
| Ghost count (TC stopping, no Azure VM) | 1185 (100%) |
| Oldest stopping `lastModified` | `2026-04-20T16:00:46.879Z` (355 min stuck) |
| Pending tasks | 1680 |
| Effective scale-up ceiling (`max - stopping`) | 815 (40.8%) |
| Requested (what worker-manager can actually spawn) | 1 |

The pool runs across 11 Azure regions on `Standard_F8s_v2` Spot VMs. All 11 regions are active (`isArchived=false`) with weights between 0.3 and 1.0. Errors spread roughly evenly: `canadacentral` 22 at the high end, `westus3` 3 at the low end over the last 7 days. No single region is hot enough to explain the behavior.

The worker-manager error list for this pool spans `2026-04-13T01:52:36.741Z` through `2026-04-20T19:00:47.735Z` (94 entries across the ~7-day retention window):

- 53 x `Operation execution has been preempted by a more recent operation` (Spot churn. Normal by itself, but every preemption is a chance to create a ghost.)
- 27 x OS provisioning timeouts (VMs that never booted, got marked STOPPING, then stuck)
- 9 x template deployment validation failures (one-off)
- 5 x one-shot failures (quota, internal allocation, Windows undeployable, and so on)

Progression across snapshots on 2026-04-20, same pool:

| snapshot (UTC) | stoppingCount | ghost_count | ghost_pct | real_stopping |
|---|---|---|---|---|
| ~18:57 | 1682 | 1067 | 63.4% | 615 |
| ~20:00 | 1675 | 1067 | 63.7% | 608 |
| ~21:00 | 1185 | 1185 | 100.0% | 0 |
| 21:55:29 | 1185 | 1185 | 100.0% | 0 |

Real stoppers drained to zero over roughly three hours. The ghost cohort stayed flat. If the reap path only makes progress when the Azure resource still exists, this is the progression you get, and the missing `beginDelete(...)->404` branch in `deprovisionResource` explains it.

**Why ghosts persist until fixed**

There is a daily cron `expire-workers` (see `services/worker-manager/procs.yml`, runs at `01:30 UTC`) that calls `Worker.expire -> db.fns.expire_workers`:

```sql
delete from workers where workers.expires < expires_in;
```

That cron does not clean up these ghosts because `worker-scanner` keeps extending `expires` for non-stopped workers:

```js
if (worker.expires < taskcluster.fromNow('1 week')) {
    worker.expires = taskcluster.fromNow('8 days');
  });
}
```

So each scan cycle, a ghost:

1. Gets visited by the scanner.
2. Goes through `provider.checkWorker -> deprovisionResources -> beginDelete(...)->404 -> throw -> outer catch`, and stays `STOPPING`.
3. Has `worker.expires` bumped back to 8 days out.
4. Survives the daily `expire-workers` cron.

This explains why ghosts persist, and a scanner-level guard would be reasonable defense in depth. But that should be treated as a separate hardening follow-up. The proper fix for this issue remains handling `404` from `beginDelete(...)` in the shared `deprovisionResource` helper.

lotas:
deprovisionResource didn't handle deletion of resources that were removed already. After the fixes introduced in #8059 404 check was not handled properly in the delete call. As a result some workers stayed in the stopping state forever because that error wasn't handled

Fixes: #8517

### DIFF
--- a/changelog/issue-8517.md
+++ b/changelog/issue-8517.md
@@ -1,0 +1,5 @@
+audience: deployers
+level: patch
+reference: issue 8517
+---
+Fixed Azure provider workers getting stuck in `STOPPING` indefinitely when their backing Azure resources (VM, NIC, IP, disks, or ARM deployment) were deleted out-of-band (e.g. Spot preemption, ARM cascade delete). The `deprovisionResource` helper now treats a `404` from `beginDelete` the same way as a `404` from `get`: mark the resource as deleted and let the reap path continue.
--- a/services/worker-manager/src/providers/azure/index.js
+++ b/services/worker-manager/src/providers/azure/index.js
@@ -1906,6 +1906,15 @@ export class AzureProvider extends Provider {
       resourceName: typeData.name,
     });
 
+    const markDeleted = () => worker.update(this.db, worker => {
+      const resource = index !== undefined
+        ? worker.providerData[resourceType][index]
+        : worker.providerData[resourceType];
+      resource.operation = undefined;
+      resource.id = false;
+      resource.deleted = true;
+    });
+
     if (typeData?.deleted === true) {
       // if resource was already deleted we don't have to query api by name again to make sure it is 404
       // and avoid being queried multiple times during deprovision cycles
@@ -1932,18 +1941,7 @@ export class AzureProvider extends Provider {
       } catch (err) {
         if (err.statusCode === 404) {
           debug(`resource ${typeData.name} not found; removing its id and marking as deleted`);
-          await worker.update(this.db, worker => {
-            if (index !== undefined) {
-              worker.providerData[resourceType][index].operation = undefined;
-              worker.providerData[resourceType][index].id = false;
-              worker.providerData[resourceType][index].deleted = true;
-            } else {
-              worker.providerData[resourceType].operation = undefined;
-              worker.providerData[resourceType].id = false;
-              worker.providerData[resourceType].deleted = true;
-            }
-          });
-
+          await markDeleted();
           return true;
         }
         throw err;
@@ -1963,6 +1961,13 @@ export class AzureProvider extends Provider {
           typeData.name,
         ));
       } catch (err) {
+        if (err.statusCode === 404) {
+          // resource was deleted out-of-band (e.g. Spot preemption or ARM cascade delete);
+          // treat the same way as the get()->404 branch above so the worker can progress
+          debug(`resource ${typeData.name} already absent on delete; marking as deleted`);
+          await markDeleted();
+          return true;
+        }
         if (err.statusCode === 409 &&
             /previous deployment.*still active/i.test(err.message ?? '')) {
           debug('deployment still active; will retry deletion later');
--- a/services/worker-manager/test/fakes/azure.js
+++ b/services/worker-manager/test/fakes/azure.js
@@ -193,6 +193,15 @@ class ResourceManager {
   }
 
   /**
+   * Remove a fake resource directly, simulating out-of-band deletion
+   * (e.g. Spot preemption or ARM cascade delete)
+   */
+  removeFakeResource(resourceGroupName, name) {
+    const key = `${resourceGroupName}/${name}`;
+    this._resources.delete(key);
+  }
+
+  /**
    * Modify a fake resource
    */
   modifyFakeResource(resourceGroupName, name, modifier) {
--- a/services/worker-manager/test/provider_azure_test.js
+++ b/services/worker-manager/test/provider_azure_test.js
@@ -1788,6 +1788,55 @@ helper.secrets.mockSuite(testing.suiteName(), [], function(mock, skipping) {
       // check that there's a request to delete the disk (by name)
       assert.deepEqual(await fake.computeClient.disks.getFakeRequestParameters('rgrp', diskName), {});
     });
+
+    test('beginDelete 404 is treated as already-deleted (ghost worker reaps)', async function() {
+      // Seed the worker with ids for every resource so deprovision goes down
+      // the beginDelete branch, then remove the backing fakes to simulate
+      // Azure having already deleted them out-of-band (Spot preemption,
+      // ARM cascade delete, manual cleanup, etc.).
+      await makeResource('ip', true);
+      await makeResource('nic', true);
+      await makeResource('disks', true, 0);
+      await makeResource('disks', true, 1);
+      await makeResource('vm', true);
+      await worker.update(helper.db, worker => {
+        worker.state = 'running';
+      });
+
+      await provider.removeWorker({ worker, reason: 'test' });
+
+      // remove backing resources so beginDelete throws 404 for each
+      fake.computeClient.virtualMachines.removeFakeResource('rgrp', vmName);
+      fake.networkClient.networkInterfaces.removeFakeResource('rgrp', nicName);
+      fake.networkClient.publicIPAddresses.removeFakeResource('rgrp', ipName);
+      fake.computeClient.disks.removeFakeResource('rgrp', 'disks0');
+      fake.computeClient.disks.removeFakeResource('rgrp', 'disks1');
+
+      // reset errors so we can assert no deletion-error is recorded
+      provider.errors = provider.errors || {};
+      provider.errors[workerPoolId] = [];
+
+      // drive the reap loop. Each call advances one resource (VM -> NIC -> IP -> disks).
+      for (let i = 0; i < 6; i++) {
+        await provider.deprovisionResources({ worker, monitor });
+      }
+
+      const [reloaded] = await helper.getWorkers();
+      assert.equal(reloaded.state, 'stopped', 'ghost worker should progress to stopped');
+      assert.equal(reloaded.providerData.vm.deleted, true);
+      assert.equal(reloaded.providerData.nic.deleted, true);
+      assert.equal(reloaded.providerData.ip.deleted, true);
+      assert.equal(reloaded.providerData.disks[0].deleted, true);
+      assert.equal(reloaded.providerData.disks[1].deleted, true);
+      assert.equal(reloaded.providerData.vm.id, false);
+      assert.equal(reloaded.providerData.nic.id, false);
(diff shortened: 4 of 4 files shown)


matt-boris (approved):
Thanks!

PR #8519, opened 2026-04-21
feat(w-m): Restrict intermediate certificates downloads

lotas:
Checking if intermediate cert URL belongs to allowed hosts

Bugzilla Bug:
[2028956](https://bugzilla.mozilla.org/show_bug.cgi?id=2028956)

### DIFF
--- a/services/worker-manager/src/providers/azure/index.js
+++ b/services/worker-manager/src/providers/azure/index.js
@@ -4,6 +4,7 @@ import taskcluster from '@taskcluster/client';
 import forge from 'node-forge';
 import crypto from 'crypto';
 import got from 'got';
+import net from 'net';
 import { rootCertificates } from 'tls';
 import { WorkerPool, Worker } from '../../data.js';
 import azureApi from './azure-api.js';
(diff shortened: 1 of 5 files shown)


services/worker-manager/src/providers/azure/index.js
```diff
@@ -49,6 +50,46 @@ const failProvisioningStates = new Set(['Failed', 'Deleting', 'Canceled', 'Deall
 
 const DEPLOYMENT_METHOD_ARM = 'arm-template';
 const maxInstanceView404Streak = 2;
+// Per Azure Certificate Authority details, these are the HTTP AIA hosts
+// clients may need to reach for Azure certificate chain building.
```
matt-boris, line 54: Might be worth linking to their docs here too https://learn.microsoft.com/en-us/azure/security/fundamentals/azure-certificate-authority-details?tabs=root-and-subordinate-cas-list

matt-boris (dismissed):
Changes LGTM, thank you! Left one small suggestion to add microsoft's docs link to an inline comment

services/worker-manager/src/providers/azure/index.js
lotas: thanks! Will add

matt-boris (approved):

PR #8520, opened 2026-04-21
fix(worker-runner): Tighten permissions on runner config file

matt-boris:
worker-runner now ensures its config file (runner.yml / worker-runner.json) is readable only by its owner before reading, chmod'ing to 0600 (or tightening to an owner-only ACL on Windows) and warning if it was previously group- or world-readable. Closes an exposure where a task on a static-provider worker could read the staticSecret and impersonate the worker via registerWorker.

References bug 2032277.

>worker-runner now tightens the permissions of its configuration file (typically `runner.yml` / `worker-runner.json`) to be readable only by its owner before reading it, and logs a warning if the file was previously group- or world-readable. This closes an exposure where a task running on a worker using the `static` provider could read the `staticSecret` out of a loosely-permissioned runner config and impersonate the worker via `registerWorker`. Worker deployers using the static provider should update their provisioning so the runner config is created with mode `0600` (or the equivalent owner-only ACL on Windows) from the start.

### DIFF
--- a/tools/worker-runner/perms/perms_posix.go
+++ b/tools/worker-runner/perms/perms_posix.go
@@ -41,7 +41,12 @@ func ReadPrivateFile(filename string) ([]byte, error) {
 // verifyPrivateToOwner verifies that the given file can only be read by the
 // file's owner, returning an error if this is not the case, or cannot be
 // determined.  Returns an error satisfying os.IsNotExist when the file does
-// not exist.
+// not exist.  A file is considered private if no group or other permission
+// bits are set AND the owner-read bit is set (e.g. 0400, 0500, 0600, 0700);
+// operators who deliberately tighten permissions beyond 0600 should not be
+// forced to loosen them, but an owner-unreadable config (e.g. 0200) is
+// treated as a provisioning error because worker-runner must be able to
+// read the file.
 func verifyPrivateToOwner(filename string) error {
 	stat, err := os.Stat(filename)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


tools/worker-runner/perms/perms_posix.go
```diff
@@ -54,8 +54,40 @@ func verifyPrivateToOwner(filename string) error {
...
+	if err != nil {
+		return false, err
+	}
+
+	uid := int(stat.Sys().(*syscall.Stat_t).Uid)
+	if uid != os.Getuid() {
+		return false, fmt.Errorf("%s has incorrect owner id %d (expected %d); refusing to tighten permissions on a file owned by another user", filename, uid, os.Getuid())
+	}
+
+	if stat.Mode().Perm() == os.FileMode(0600) {
```
Eijebong: This won't detect the file as "secure" if perms are `0400` which seems wrong.

tools/worker-runner/perms/perms_posix.go
```diff
@@ -54,8 +54,40 @@ func verifyPrivateToOwner(filename string) error {
...
+		return false, nil
+	}
+
+	if err := os.Chmod(filename, 0600); err != nil {
```
Eijebong: Because of the above, this will actually widen the permission on the file if they were 0400 before.

Eijebong: (I also think that 0400 would be more obvious on the intent to set on that file although I'm not sure how much more value there is there since that service usually runs as root)

Eijebong: Technically there's a TOCTOU here which doesn't really matter since we're tightening rights, but given the amount of reports lately for non issues I think it's worth fixing before we make everyone lose more of their time.

If the file gets replaced between by a symlink then you'll set permissions on the target path. It doesn't matter because the result is "congrats you gave u+rw
 to a file owned by said user" which could always do that on their own.

I think it's worth changing to something alone the lines of:

```go
f, err := os.OpenFile(filename, os.O_RDONLY|syscall.O_NOFOLLOW, 0)
if err != nil {
        return false, err
}
defer f.Close()

stat, err := f.Stat()
if err != nil {
        return false, err
}

uid := int(stat.Sys().(*syscall.Stat_t).Uid)
if uid != os.Getuid() {
		return false, fmt.Errorf("%s has incorrect owner id %d (expected %d); refusing to tighten permissions on a file owned by another user", filename, uid, os.Getuid())
}

wasLoose := false
if stat.Mode().Perm() != os.FileMode(0600) {
        if err := f.Chmod(0600); err != nil {
                return false, fmt.Errorf("could not chmod %s to 0600: %w", filename, err)
        }
        wasLoose = true
}
```

Eijebong: To check that the file is `0X00` mode:

```go
if stat.Mode().Perm()&0o077 == 0 {
    return false, nil
}
```

Eijebong (commented):
You probably want to set those to 0600 (or 0400) to avoid polluting test output:

https://github.com/taskcluster/taskcluster/blob/e97ba08943266681d716fdfd3d9acb0550eaba00/tools/worker-runner/runner/run_test.go#L66
https://github.com/taskcluster/taskcluster/blob/e97ba08943266681d716fdfd3d9acb0550eaba00/tools/worker-runner/runner/run_test.go#L101
https://github.com/taskcluster/taskcluster/blob/e97ba08943266681d716fdfd3d9acb0550eaba00/tools/worker-runner/runner/run_test.go#L135
https://github.com/taskcluster/taskcluster/blob/e97ba08943266681d716fdfd3d9acb0550eaba00/tools/worker-runner/runner/run_test.go#L160

---

lotas (approved):

PR #8522, opened 2026-04-22
Use `on-resolved` in the publish route for notifications instead of on-any

### LINKED ISSUE #3684: replace `on-any` with `on-resolved` in monorepo taskgraph
Once a release containing #3662 has been deployed to community cluster, we should update our usage of `on-any` in `taskcluster/ci/release/kind.yml` to `on-resolved`. Since the former is deprecated now.

Eijebong:
On-any has been deprecated by #3662, we should show the example and stop using deprecated stuff. On-resolved does the exact same thing but isn't deprecated

Fixes #3684

### DIFF
--- a/changelog/issue-3684.md
+++ b/changelog/issue-3684.md
@@ -1,0 +1,3 @@
+level: silent
+reference: issue 3684
+---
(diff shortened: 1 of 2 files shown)


lotas (approved):
Eagle eye

PR #8527, opened 2026-04-22
fix(queue): prevent orphaned pending tasks when Pulse publish fails

### LINKED ISSUE #8525: Tasks can become "pending" orphans invisible to workers when Pulse publish fails during run transitions
**Describe the bug**

A task can end up with its latest run in state `pending` in the `tasks` table while having **no corresponding row** in `queue_pending_tasks`. In that state:

- The task appears `pending` in `queue.task(taskId).status` and on the task view UI.
- No worker ever claims it, because `queue_pending_tasks_get` doesn't see it.
- `queue.pendingTasks(...)` / `queue.taskQueueCounts(...)` report 0 for the task queue, because they read only from `queue_pending_tasks`. The "Pending Tasks" column in the provisioner/worker-type UI is therefore also blind to it.

The task sits orphaned until `deadline`, at which point `resolve_task_at_deadline` resolves it as `deadline-exceeded`. The only user-side workaround is `queue.cancelTask`.

**Observed instance**

firefox-ci task `Se7L14RuS7-F3clTVAS6IA` on `gecko-3/b-linux`:

- run 0: `exception / worker-shutdown` at `2026-04-22T14:01:15.701Z`
- run 1: `pending` (retry), scheduled at the same instant — never claimed.
- `GET /api/queue/v1/pending/gecko-3/b-linux` → `pendingTasks: 0`
- `GET /api/queue/v1/task-queues/gecko-3/b-linux/counts` → `{pendingTasks: 0, claimedTasks: 7}`

**Root cause**

Every code path that transitions a run to `pending` does two non-atomic steps:

1. **DB write:** updates `tasks.runs` to contain the new pending run (DB fns `resolve_task`, `check_task_claim`, `schedule_task`, etc.).
2. **Post-commit:** `Promise.all([queueService.putPendingMessage(task, runId), publisher.taskPending(...)])` — the first inserts into `queue_pending_tasks`, the second publishes a Pulse message.

If the Pulse publish rejects (e.g. RabbitMQ `sendDeadline` timeout under load), `Promise.all` rejects. The `queue_pending_tasks` insert may not have committed, but the tasks-table write already has. When the caller retries the API call, the DB function sees the run is no longer `running` and returns a no-op → `putPendingMessage` is never re-attempted.

For the Pulse-consumer-driven `workerremovedresolver` path, NACK + redelivery makes it worse: `get_claimed_tasks_by_worker` returns empty on redelivery (the task is no longer claimed) and nothing ever requeues it.

There is no reconciler that cross-checks `tasks.runs` against `queue_pending_tasks`.

**Affected code paths (all share the same non-atomic shape)**

- `reportException` retry — `services/queue/src/api.js` ~L1993-L2010
- `workerremovedresolver` retry — `services/queue/src/workerremovedresolver.js` ~L121-L135
- `claimresolver` retry on claim-expired — `services/queue/src/claimresolver.js` ~L164-L175
- `createTask` initial pending — `services/queue/src/api.js` ~L988-L996
- `scheduleTask` / `rerunTask` — `services/queue/src/api.js` ~L1160-L1170
- `dependencyTracker.scheduleTask` — `services/queue/src/dependencytracker.js` ~L277-L287

**Proposed fix**

Move the `queue_pending_tasks` insert into the same SQL bodies that create the pending run, so the two writes commit in one transaction. Concretely, extend `resolve_task`, `check_task_claim`, `schedule_task`, and any other DB fn that introduces a pending run to also perform the equivalent of `queue_pending_tasks_add` when a pending run is created. Remove the now-redundant `putPendingMessage` calls at the six JS callsites. Pulse `taskPending` publishing stays as best-effort (it is a notification; the DB is authoritative).

Nice-to-have on top: a reconciler that periodically finds tasks whose latest run is `pending` with no matching `queue_pending_tasks` row and repairs them — both as defense-in-depth and to self-heal existing orphans.

**Workaround for a currently-stuck task**

`queue.cancelTask <taskId>` — it updates `tasks.runs` directly and doesn't depend on `queue_pending_tasks`. Alternatively, wait for `deadline` → `deadline-exceeded`.

**Additional context**

- Short-term mitigation for the RabbitMQ-load trigger landed recently: bump `pulse sendDeadline` from 12s → 30s.
- Unknown: how many tasks are currently orphaned on firefox-ci. Can be counted with a query intersecting `tasks` (latest run pending, deadline > now) with `queue_pending_tasks`.

matt-boris:
## Summary

Fix two related races that could leave a Taskcluster task in an inconsistent state when a Pulse publish failed during task creation or a run state transition.

The pattern in both cases — a DB write followed by a separate Pulse publish, with a JS-layer step in between — is a legacy of Taskcluster's pre-Postgres era (Azure Queue + Azure Table), when cross-system atomicity was physically impossible. After the Postgres migration, atomicity became feasible but the call shape didn't change. Under current Pulse/RabbitMQ load the failure modes are no longer rare enough to ignore.

Fixes #8525.

## Observed impact

~65 orphaned tasks across 16 queues on firefox-ci from a single ~17-minute RabbitMQ load window on 2026-04-22, including 17 decision tasks that blocked downstream graphs. The pulse `sendDeadline` bump in #8524 reduces the trigger frequency; this PR closes the underlying race.

## Change shape

**db/versions/0124.yml**
- New helper `queue_pending_tasks_add_for_task(task_queue_id, priority, deadline, task_id, run_id)` — caller passes the values it already has from its `FOR UPDATE` SELECT, no redundant lookup.
- `schedule_task`, `rerun_task`, `resolve_task`, `check_task_claim` each `PERFORM` the helper inline after their `UPDATE tasks`, so the `queue_pending_tasks` row commits in the same transaction as the `tasks.runs` update.
- New `create_task_atomic` (replaces JS-layer `create_task_projid` + `putDeadlineMessage`): atomically inserts the task row and its `queue_task_deadlines` tracking row. `create_task_projid` is deprecated. Uses `ON CONFLICT ... DO UPDATE SET ...` so any pre-existing orphan deadline row is overwritten with current metadata (avoids a deadline-resolver mismatch on `(task_id, deadline)`).

**JS layer**
- HTTP handlers in `services/queue/src/api.js` (`createTask`, `scheduleTask`, `rerunTask`, `reportException`) wrap their post-commit Pulse publishes in try/catch and report a `'warning'` via `monitor.reportError`. The DB is authoritative; a Pulse failure no longer fails the API call.
- Background resolvers (`claimresolver`, `workerremovedresolver`, `dependencytracker.scheduleTask`) intentionally do **not** wrap their publishes — these preserve their documented at-least-once semantics where a throw triggers NACK + redelivery and re-attempts the publish.
- Six `queueService.putPendingMessage(...)` callsites are dropped (now atomic in DB). `QueueService.prototype.putPendingMessage` itself is retained for test-setup use.
- `services/queue/src/main.js`: `dependencyTracker` and `worker-removed-resolver` no longer take `queueService` (no longer used).
- `Task.create()` takes a new `deadlineDelaySeconds` arg and calls `create_task_atomic`.

>Fix two related races that could leave a Taskcluster task in an inconsistent state when a Pulse publish failed during task creation or a run state transition:
>
>- A run could transition to `pending` in `tasks.runs` without a corresponding `queue_pending_tasks` row, making the task invisible to workers and to the "pending tasks" UI/API counts. Transitions of a run to `pending` (`schedule_task`, `rerun_task`, `resolve_task`, `check_task_claim`) now commit the `queue_pending_tasks` row in the same database transaction as the `tasks.runs` update.
>- A task could be inserted into `tasks` without a corresponding `queue_task_deadlines` row, leaving it untracked by the deadline resolver. `createTask` now inserts both rows atomically inside `create_task_atomic`.
>
>In the queue's HTTP API handlers (`createTask`, `scheduleTask`, `rerunTask`, `reportException`), Pulse `taskPending` / `taskException` / `taskDefined` publishes that follow these now-atomic DB commits are best-effort: the database is the source of truth, so a Pulse publish failure no longer fails the operation. Consumers that need exact-once notification should treat Pulse as advisory and read `queue.task(taskId).status` for the authoritative state. Background resolvers (`claim-resolver`, `worker-removed-resolver`, `dependency-resolver`) preserve their pre-existing at-least-once publish semantics by continuing to fail the handler on Pulse error so redelivery re-attempts the publish.


db/versions/0124.yml
matt-boris, line 1: The function diffs for `schedule_task`, `rerun_task`, `resolve_task`, and `check_task_claim` are easiest to view in the [`db/fns.md`](https://github.com/taskcluster/taskcluster/pull/8527/changes#diff-d1a8f094059f92ccb4d3564a8bd078d4716289c19fd36f6027d804294d0d9b44).

db/versions/0124.yml
```diff
@@ -0,0 +1,368 @@
...
+    body: |-
+      declare
+        t_task_queue_id text;
+        t_priority task_priority;
+        t_deadline timestamptz;
+        priority_int integer;
+      begin
+        -- Caller holds FOR UPDATE on this row; the read is a convenience,
+        -- not a locking operation.
+        select tasks.task_queue_id, tasks.priority, tasks.deadline
```
lotas: I wonder if this could be passed as an argument to avoid extra select call? 
`schedule_task` could fetch same values in its `FOR UPDATE` query maybe.

services/queue/src/api.js
```diff
@@ -986,13 +986,8 @@ builder.declare({
...
   if (runZeroState === 'pending') {
-    await Promise.all([
-      // Put message into the task pending queue
-      this.queueService.putPendingMessage(task, runId),
-
-      // Publish message to pulse
-      this.publisher.taskPending({ status, task: taskPulseContents, runId }, task.routes),
-    ]);
+    // queue_pending_tasks insert is now atomic inside schedule_task (db v124).
+    await this.publisher.taskPending({ status, task: taskPulseContents, runId }, task.routes);
```
lotas: I wonder if we should turn it into a best-effort notification only with `try/catch` 
Since we already have everything in db, tasks would be claimed eventually, so the caller wouldn't need to retry in case of such pulse failure? Wdyt?

services/queue/src/api.js
```diff
@@ -1999,14 +1992,12 @@ builder.declare({
...
-    await Promise.all([
-      this.queueService.putPendingMessage(task, runId + 1),
-      this.publisher.taskPending({
-        status,
-        task: taskPulseContents,
-        runId: runId + 1,
-      }, task.routes),
-    ]);
+    // queue_pending_tasks insert is now atomic inside resolve_task (db v124).
+    await this.publisher.taskPending({
```
lotas: probably same for retries - we can wrap it and not fail the response

db/versions/0124.yml
```diff
@@ -0,0 +1,368 @@
...
+          where tasks.task_id = task_id_in;
+
+        if t_task_queue_id is null then
+          return;
+        end if;
+
+        -- Skip if the task's deadline has already passed; the deadline
+        -- resolver will resolve it as `deadline-exceeded`. Matches the
+        -- pre-fix JS behavior of `QueueService.putPendingMessage`.
+        if t_deadline < now() then
```
lotas: Slight change of the previous behavior, but probably not a big deal:

https://github.com/taskcluster/taskcluster/blob/7a6665de01b4ad102704c8151eed55e4e35bcf14/services/queue/src/queueservice.js#L281-L289

maybe it was only for tests or some really weird edge cases

services/queue/src/dependencytracker.js
```diff
@@ -275,14 +275,12 @@ class DependencyTracker {
     // Put message into pending queue, and publish message to pulse,
     // if the initial run is pending
     if (task.runs && task.runs[0].state === 'pending') {
-      await Promise.all([
-        this.queueService.putPendingMessage(task, 0),
```
lotas, line 279: we can also drop that dependency in main.js - it is no longer used here

services/queue/src/workerremovedresolver.js
```diff
@@ -124,14 +124,12 @@ class WorkerRemovedResolver {
         task.runs.length - 1 === runId + 1 &&
         newRun.state === 'pending' &&
         newRun.reasonCreated === 'retry') {
-      await Promise.all([
-        this.queueService.putPendingMessage(task, runId + 1),
```
lotas, line 128: same here, `queueService` no longer needed to be injected

lotas:
great stuff! 
I recently stumbled upon this PR https://github.com/taskcluster/taskcluster/pull/8535 which was also fixing aftermath of the same problem. 
Maybe it's worth including deadline messages, since they are always created at task creation time and should be atomic as well?

services/queue/src/dependencytracker.js
matt-boris: Good eye, thanks, addressed!

db/versions/0124.yml
```diff
@@ -0,0 +1,481 @@
+version: 124
+description: Atomic queue_pending_tasks and queue_task_deadlines inserts in createTask and pending-run transitions
+methods:
+  create_task_atomic:
```
lotas, line 6: afaiu this is replacing `create_task_projid`? Should that be deprecated, or it is still used for tests

matt-boris: Good eye. It's still used for tests, so that's why I didn't deprecate it

matt-boris: I'll deprecate it and use `db.deprecatedFns` to access it still in the tests.

lotas (approved):
Many thanks!

PR #8530, opened 2026-04-23
fix(w-m): always pre-flight GET in Azure deprovisionResource

### LINKED ISSUE #8526: Azure provider: deprovisionResource wastes a scan cycle per resource because the pre-flight get() is skipped when typeData.id is set
### Summary

On `gecko-t/win11-64-25h2` (FXCI, v99.1.1), STOPPING workers have **median age 3.7h, P90 4.1h, P99 4.9h**, and **97% of them (1080/1112) have no matching Azure VM** — they are ghosts waiting on the scanner to walk them through vm → nic → ip → disks → deployment sequentially. `estimator.js` counts all of them against `maxCapacity` (see #8161), so the pool reports 1997/2000 "used" capacity while pending tasks pile up.

This is a logic bug in `deprovisionResource` at [`services/worker-manager/src/providers/azure/index.js:1978-2005`](https://github.com/taskcluster/taskcluster/blob/main/services/worker-manager/src/providers/azure/index.js#L1978-L2005):

```js
let shouldDelete = false;
if (!typeData.id) {               // skips pre-flight when id is set (true on first visit)
  try {
    let { provisioningState } = await ... client.get(...);
    if (!(['Deleting','Deallocating','Deallocated'].includes(provisioningState))) {
      shouldDelete = true;
    }
  } catch (err) {
    if (err.statusCode === 404) { await markDeleted(); return true; }  // fast path
    throw err;
  }
}

if (typeData.id || shouldDelete) {    // fires beginDelete() when typeData.id alone is truthy
  ... beginDelete ...
}
```

When ARM cascade-delete (`deleteOption: 'Delete'` on the VM) has already removed the NIC/IP/disk before the scanner visits, `typeData.id` is still set in the worker's providerData — so we skip the `get()` pre-flight and fire a useless `beginDelete()`. The next cycle's `get()` finally returns 404 and we move on. Two scan cycles per resource × five resources × ~44 min cadence = 3-6h reap tail.

### Why the #8517 fix doesn't rescue this path

[#8517](https://github.com/taskcluster/taskcluster/issues/8517) (closed in v99.1.1) added `if (err.statusCode === 404)` on the `beginDelete()` catch block, intended to short-circuit when Azure says "already gone". Per Microsoft's own published contracts for the relevant resource types, DELETE on VMs/disks/NICs/public IPs is an **asynchronous long-running operation** with only three defined success codes:

| Code | Meaning |
|---|---|
| `200 OK` | Resource existed and was deleted |
| `202 Accepted` | Deletion accepted, poll `Location` |
| `204 No Content` | **Resource does not exist** (the idempotent case) |

This response set is documented verbatim on [Virtual Machines - Delete](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/delete) and [Disks - Delete](https://learn.microsoft.com/en-us/rest/api/compute/disks/delete). The [Microsoft.Network spec for networkInterfaces and publicIPAddresses DELETE](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/network/resource-manager/Microsoft.Network/Network/stable/2025-05-01/virtualNetwork.json) uses the same three response codes (`200`, `202`, `204`, `default`) with `x-ms-long-running-operation: true`. `404 Not Found` is **not** a defined response for these operations — the idempotent "already gone" reply is 204 from the LRO's final poll. ARM's own overview confirms 404 is the expected reply to the *GET* convergence probe after a successful delete, not to a repeat DELETE itself ([Delete Resource Groups and Resources](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/delete-resource-group#after-deletion)).

Production logs line up with the spec. Across `gecko-t/win11-64-25h2` for the last 24h:

```
jsonPayload.Fields.message ~ "already absent on delete"   →   0 events
jsonPayload.Type = "azure-instance-view-repeated-404"     →   0 events
worker-removed reason = "vm not found in state running"   →   428 events   (all via GET 404)
```

The #8517 fast-path is deployed but cannot fire in practice — it's waiting for a DELETE-404 that the spec says Azure won't emit. The reliable 404 signal is **GET**, not DELETE.

### Proposed fix

Always pre-flight with `get()`:

```js
let shouldDelete = false;
try {
  const { provisioningState } = await this._enqueue('query', () => client.get(
    worker.providerData.resourceGroupName,
    typeData.name,
  ));
  debug(`found provisioningState ${provisioningState}`);
  if (!(['Deleting','Deallocating','Deallocated'].includes(provisioningState))) {
    shouldDelete = true;
  }
} catch (err) {
  if (err.statusCode === 404) {
    debug(`resource ${typeData.name} not found; marking as deleted`);
    await markDeleted();
    return true;
  }
  throw err;
}

if (shouldDelete) {     // no longer gated on typeData.id
  // ... existing beginDelete block unchanged ...
}
return false;
```

Effect:
- Resource already gone via cascade → GET 404 → markDeleted → loop continues same cycle.
- Resource in `Deleting` state → `shouldDelete=false`, return false → wait, don't re-fire beginDelete (current code DOES re-fire because `typeData.id` alone satisfies the gate).
- Resource healthy → GET 200 → beginDelete as today.

Expected reap time drops from 2 cycles/resource to ~1 cycle/resource, i.e. **~50% off the STOPPING tail**.

Cost: +1 ARM read per resource per first visit. Amortized across ~11 configured regions, well inside ARM read quotas.

### References

- [Virtual Machines - Delete (REST API)](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/delete)
- [Disks - Delete (REST API)](https://learn.microsoft.com/en-us/rest/api/compute/disks/delete)
- [Microsoft.Network spec — networkInterfaces/publicIPAddresses DELETE](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/network/resource-manager/Microsoft.Network/Network/stable/2025-05-01/virtualNetwork.json)
- [Delete Resource Groups and Resources (ARM convergence semantics)](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/delete-resource-group#after-deletion)
- [Microsoft Azure API Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md) — "return 204-No Content for DELETE even if the resource does not exist; do not return 404"

### Related issues

- [#8517](https://github.com/taskcluster/taskcluster/issues/8517) — closed; fast-path added but dead code in production. This issue is the follow-up to fix the missed scenario.
- [#8161](https://github.com/taskcluster/taskcluster/issues/8161) — estimator counts STOPPING toward `maxCapacity`; amplifies this bug's visibility as "pool full, no scale-up".
- [#8441](https://github.com/taskcluster/taskcluster/issues/8441) — ARM 800-deployments-per-RG limit. Ghost deployments consume quota. Observed ~20 provisioning failures with `exceeds the quota of '800'` on this pool in the last 24h, directly caused by the reap lag.
- [#5291](https://github.com/taskcluster/taskcluster/issues/5291) — historical "Azure workers take 10-30 min to remove" report; same family of symptoms, now manifests as the longer tail seen here.

matt-boris:
Fixes #8526.

>Fixed the Azure provider's `deprovisionResource` wasting a worker-scanner cycle per resource when the backing VM/NIC/IP/disk had already been removed out-of-band (e.g. ARM cascade-delete via `deleteOption: 'Delete'`, Spot preemption). Previously the pre-flight `GET` was skipped whenever the worker still had a stored `id`, so the scanner fired a no-op `beginDelete` first and only discovered the resource was gone on the following cycle. The helper now always performs the pre-flight `GET`, so a missing resource is marked deleted immediately and the reap chain continues in a single cycle, shortening the `STOPPING` tail for affected Azure pools.

### DIFF
--- a/changelog/issue-8526.md
+++ b/changelog/issue-8526.md
@@ -1,0 +1,5 @@
+audience: deployers
+level: patch
+reference: issue 8526
+---
+Fixed the Azure provider's `deprovisionResource` wasting a worker-scanner cycle per resource when the backing VM/NIC/IP/disk had already been removed out-of-band (e.g. ARM cascade-delete via `deleteOption: 'Delete'`, Spot preemption). Previously the pre-flight `GET` was skipped whenever the worker still had a stored `id`, so the scanner fired a no-op `beginDelete` first and only discovered the resource was gone on the following cycle. The helper now always performs the pre-flight `GET`, so a missing resource is marked deleted immediately and the reap chain continues in a single cycle, shortening the `STOPPING` tail for affected Azure pools.
--- a/services/worker-manager/src/providers/azure/index.js
+++ b/services/worker-manager/src/providers/azure/index.js
@@ -1976,33 +1976,38 @@ export class AzureProvider extends Provider {
     debug(`deprovisionResource for ${resourceType} with index ${index}`);
 
     let shouldDelete = false;
-    // lookup resource by name
-    if (!typeData.id) {
-      try {
-        let { provisioningState } = await this._enqueue('query', () => client.get(
-          worker.providerData.resourceGroupName,
-          typeData.name,
-        ));
-        // resource could be successful, failed, etc.
-        // we have not yet tried to delete the resource
-        debug(`found provisioningState ${provisioningState}`);
-        if (!(['Deleting', 'Deallocating', 'Deallocated'].includes(provisioningState))) {
-          shouldDelete = true;
-        }
-      } catch (err) {
-        if (err.statusCode === 404) {
-          debug(`resource ${typeData.name} not found; removing its id and marking as deleted`);
-          await markDeleted();
-          return true;
-        }
-        throw err;
+    // Always look up the resource by name before attempting a delete. The
+    // pre-flight GET catches resources that have already been removed out of
+    // band (e.g. ARM cascade-delete via `deleteOption: 'Delete'`, Spot
+    // preemption, manual cleanup) and lets us mark them deleted without
+    // spending an extra scanner cycle on a no-op beginDelete. It also lets us
+    // skip re-firing beginDelete on a resource already in the Deleting state.
+    try {
+      const { provisioningState } = await this._enqueue('query', () => client.get(
+        worker.providerData.resourceGroupName,
+        typeData.name,
+      ));
+      // resource could be successful, failed, etc.
+      debug(`found provisioningState ${provisioningState}`);
+      if (!(['Deleting', 'Deallocating', 'Deallocated'].includes(provisioningState))) {
+        shouldDelete = true;
       }
+    } catch (err) {
+      if (err.statusCode === 404) {
+        debug(`resource ${typeData.name} not found; removing its id and marking as deleted`);
+        await markDeleted();
+        return true;
+      }
+      throw err;
     }
 
-    // NB: possible resource leak if we don't require `return true`
-    // we don't check operation status: no differentiating between
-    // operation => create and operation => delete
-    if (typeData.id || shouldDelete) {
+    // Callers treat `return false` as "still deleting, keep waiting"; a
+    // missed `markDeleted`/`return true` above would keep the worker stuck
+    // in STOPPING across additional scanner cycles (Azure-side resources
+    // are already gone by this point, so this leaks scanner work, not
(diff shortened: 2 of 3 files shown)


lotas (approved):
Great addition, thanks!

PR #8535, opened 2026-04-24
Enforce unique `task_id` in `queue_task_deadlines` and catch no-ops early in the deadline resolver

### LINKED ISSUE #8529: Cannot read properties of undefined (reading 'reasonResolved')
https://mozilla.sentry.io/issues/6954496208/?project=6459390&referrer=metric-issue-contributing-issues-issue-stream

```bash
TypeError: Cannot read properties of undefined (reading 'reasonResolved')
    at DeadlineResolver.handleMessage (/app/services/queue/src/deadlineresolver.js:139:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at ? (/app/services/queue/src/deadlineresolver.js:105:9)
    at Promise.all (index 18)
    at DeadlineResolver.poll (/app/services/queue/src/deadlineresolver.js:101:5)
    at Promise.all (index 0)
    at Iterate.handler (/app/services/queue/src/deadlineresolver.js:76:9)
    at Iterate.single_iteration (/app/libraries/iterate/src/index.js:104:7)
    at Iterate.iterate (/app/libraries/iterate/src/index.js:135:9)
```

Eijebong:
Enforce unique `task_id` in `queue_task_deadlines`

`createTask` is supposed to be idempotent but was inserting the same values in `queue_task_deadlines` when called multiple times with the same input.

This caused some issue due to a pulse outage which lead to a lot of `createTask` retries and later on to tasks to actually hit their deadline.

What ended up happening is that the deadline resolver (which is running with replicas > 1 in prod) picked up the same task multiple times in parallel, the first instance cancelled the task properly and continued on happily, the other instances tried to cancel the task, saw
that it was already gone, and hit a secondary bug where the task cancellation isn't checked, and hit an issue where the last run was undefined.

From the deadline resolver's code:

```js
// This updates the local task object with the runs created by the
// cancellation iif a cancellation happened
task.updateStatusWith(await this.db.fns.cancel_task(taskId, 'deadline-exceeded'));

// If a cancellation failed (because the task was already cancelled for
// example, this returns undefined)
let run = _.last(task.runs);

// Kaboom
if (run.reasonResolved === 'deadline-exceeded' && // [...]
```

While that bug on its own is worth fixing, we shouldn't end up with duplicate rows in `queue_task_deadlines` in the first place, so this commit adds a UNIQUE constraint on `task_id`. A task can only ever have one deadline date.

Regarding the migration, I used the same pattern as `0091` where I delete old duplicates *before* changing the uniqueness on the table (see 932c49b7b5). The rest is pretty self explanatory, change the primary key from `(task_id, created, deadline)` to `task_id`. The previous PK was added without any explanation by e1368f3e9a with a bunch of others, presumably to match production data which might have had duplicates already at the time.

Worth noting that while I think the idempotency should lie on the DB side by using `ON CONFLICT(task_id) DO NOTHING` on the insertion function, I chose to stay consistent with the rest of the codebase and catch the uniqueness issue on the javascript side instead.

---

Catch cancellation no-ops early in the deadline resolver

`cancel_task` returns an empty result set when the task is already in a terminal state (resolved by another path between `Task.get` and the cancel call). The previous code blindly called `_.last(task.runs)` on the unchanged task, got `undefined`, and crashed reading
`run.reasonResolved`. Instead, we now detect the no-op via `updateStatusWith`'s return value, log a warning, and remove the deadline message as there's nothing left to cancel in that case.

Fixes #8529


services/queue/src/deadlineresolver.js
```diff
@@ -131,7 +131,14 @@ class DeadlineResolver {
       return remove();
     }
 
-    task.updateStatusWith(await this.db.fns.cancel_task(taskId, 'deadline-exceeded'));
+    const updated = task.updateStatusWith(
+      await this.db.fns.cancel_task(taskId, 'deadline-exceeded'),
+    );
+
+    if (!updated) {
```
matt-boris, line 138: https://github.com/taskcluster/taskcluster/blob/main/db/versions/0028.yml#L492-L498

If the task has already resolved, there's nothing to cancel, so `cancel_task` returns an empty result, so `updateStatusWith()` will return `false` and we'll continue into this block and report the error, which I don't think is right. Maybe just a debug statement instead?

matt-boris (approved):
Good stuff, ty 👍🏻

PR #8536, opened 2026-04-24
Avoid a 500 when a triggered hook produces no task

### LINKED ISSUE #8534: 500, Cannot read properties of undefined (reading 'taskId') when using the new github service hook to create a decision task
```
 TypeError:Cannot read properties of undefined (reading 'taskId')
File "/app/services/hooks/src/api.js" , line 668 , col 32 , in Proxy.triggerHookCommon
     const taskId = resp.status.taskId;
File "node:internal/process/task_queues" , line 104 , col 5 , in process.processTicksAndRejections 
```

https://github.com/Eijebong/Archipelago-index/commit/b044d8f004465e263cb95c7596721001c9157b73#commitcomment-183297669

Eijebong:
`TaskCreator.fire` was returning an empty object, which is truthy in JS, on the declined path.  defeating the `if (!resp)` guard in `triggerHookCommon` and causing a 500 by trying to read `resp.status.taskId` on an undefined `resp.status` whenever a hook's template evaluated to nothing. Return undefined instead.

Fixes #8534

### DIFF
--- a/changelog/issue-8534.md
+++ b/changelog/issue-8534.md
@@ -1,0 +1,7 @@
+audience: users
+level: patch
+reference: issue 8534
+---
+Fix a 500 raised from hooks.triggerHook when a hook's task template evaluates
+to nothing. The endpoint now correctly replies with an empty object in that
+case.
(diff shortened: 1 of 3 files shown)


matt-boris (approved):
Nice, thanks :)

PR #8548, opened 2026-04-28
chore(tools): replace exec with execFile to avoid shell interpretation

matt-boris:


matt-boris:
Just out of abundance of caution, I ran a staging release here too which was successful.

https://community-tc.services.mozilla.com/tasks/FQURi0AqQieHjFTVMEynTg/runs/1

lotas (approved):

PR #8550, opened 2026-04-29
Update sha2 to 0.11.0 in the rust client

Eijebong:
I'm really unhappy with this. `sha2` 0.11 pulls in `digest` 0.11, which replaced its dependency on `generic-array` with `hybrid-array`. The old dependency shipped a `LowerHex` impl on its array type which the new one doesn't. This means we cannot use `format!("{:x}", ...)` anymore and have to resort to either writing a helper ourselves to render it or use what the authors are recommending which is their own [`base16ct`]. I really feel bad about adding a leftpad style dependency here but the alternative is duplicating a handrolled helper which might be slower in both the upload and download crate which isn't great either.

In the end, the crate is small and maintained by the same team so I think it's fine. But not happy.

Closes #8426

### DIFF
--- a/changelog/CEf8xPOSQL-BrK7H7n1Kgw.md
+++ b/changelog/CEf8xPOSQL-BrK7H7n1Kgw.md
@@ -1,0 +1,4 @@
+audience: users
+level: patch
+---
+Update sha2 dependency in the rust client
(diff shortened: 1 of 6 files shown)


matt-boris (approved):
🤷🏻 looks fine to me, thanks for addressing.

PR #8553, opened 2026-04-30
Update got to 15.0.3

Eijebong:
From the `got` changelog:

> response.rawBody and promise.buffer() now return a Uint8Array. Buffer
> is a subclass of Uint8Array, so most code will continue to work, but
> strict type checks will need updating.

Apparently our use case isn't "most cases" and stopped working because we're calling `JSON.parse` on the buffer which works for `Buffer` (the previous type of `rawBody`), but doesn't work so well on `Uint8Array` (the new type). As you can see below, the uint array `toString` method is spitting out bytes as a comma separated string.

```
> JSON.parse(Buffer.from('{}'))
{}
> JSON.parse(new Uint8Array([123, 125]))
Uncaught:
SyntaxError: Unexpected non-whitespace character after JSON at position 3 (line 1 column 4)
> (new Uint8Array([123, 125])).toString()
'123,125'
```

The solution is fairly simple, we're using `got` with `responseType: 'text'` since f0fbe98da8, and `body` has always contained the decoded string, the original code probably assumed that further transformations were made to it and used rawBody but there's literally no point for this.

Closes #8419


clients/client/src/client.js
```diff
@@ -117,8 +117,8 @@ export const makeRequest = async function(client, method, url, payload, query) {
     hooks: {
       afterResponse: [res => {
         // parse the body, if one was given (Got's `responseType: json` fails to check content-type)
-        if (res.rawBody.length > 0 && (res.headers['content-type'] || '').startsWith('application/json')) {
-          res.body = JSON.parse(res.rawBody);
+        if (res.body && (res.headers['content-type'] || '').startsWith('application/json')) {
```
petemoore, line 120: Nit: would be worth capturing the WHY in a code comment, so a future maintainer doesn't undo the fix thinking `rawBody` is the more natural choice.

```suggestion
        // use res.body (not rawBody) — got v15+ returns Uint8Array for rawBody, which breaks JSON.parse
        if (res.body && (res.headers['content-type'] || '').startsWith('application/json')) {
```

changelog/HhTcAvUOSHeYHIU1M0pUYw.md
```diff
@@ -0,0 +1,3 @@
+audience: users
+level: silent
```
petemoore: Nit: silent changelogs don't need an `audience` field.

```suggestion
level: silent
```

petemoore (dismissed):
Looks reasonable to me, thanks Bastien!

clients/client/src/client.js
Eijebong: IMO comments like this are more likely to bitrot than ever be useful. Clearly tests failed when it was using `rawBody` and my commit message should be more than enough for anyone to understand why. Commit messages don't get out of date. Comments do.

changelog/HhTcAvUOSHeYHIU1M0pUYw.md
Eijebong: Oh I didn't know that. I always just `yarn changelog` and pick the last one in my history then change the fields. So they're always all there.

petemoore: Ha, no worries, it is relatively new: #8358

matt-boris (approved):
Nice, thanks for tackling!

PR #8554, opened 2026-04-30
Wait on tasks running `go generate` before running go-mod-tidy in `yarn generate`

Eijebong:
I think this is causing an intermittent failure in CI where the error looks like this:

```
go: finding module for package github.com/taskcluster/taskcluster/v99/infrastructure/k8s/templates
go: github.com/taskcluster/taskcluster/v99/infrastructure/k8s/templates: module github.com/taskcluster/taskcluster@latest found (v24.3.1+incompatible), but does not contain package github.com/taskcluster/taskcluster/v99/infrastructure/k8s/templates
```

This import path is completely nonsensical and it's clear that something went wrong.

The `go-mod-tidy` task in `yarn generate` wasn't declaring any dependency and this was scheduled pretty much immediately. At the same time, there were 4 different tasks running `go generate` which was rewriting go files. It's fairly easy to understand why that might be a problem if one of those generation tasks rewrites a file while the tidy one is reading it.

### DIFF
--- a/changelog/GBGkg7L8SViaF75B-SRUgA.md
+++ b/changelog/GBGkg7L8SViaF75B-SRUgA.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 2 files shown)


petemoore (approved):
Awesome, thanks!

Issue #7722, opened 2025-05-12
Upgrade nock to v14+


lotas:
Nock 14 supports mocking fetch, but some test are failing

_Originally posted by @lotas in https://github.com/taskcluster/taskcluster/pull/7721#discussion_r2085290428_

PR #8567, opened 2026-04-30
Update nock to 14.0.13 in clients/client

Eijebong:
While the test itself doesn't use nock directly, the fact that other tests surrounding it do means that the request from got was going through nock's monkeypatched http modules anyway.And nock 14 started letting `write ECANCELED` errors bubble up instead of swallowing them. The test was trying to talk https to a plain TCP server which was wrong in the first place. It was only passing because nock was eating the error and got was timing out as expected.

Closes #8557

### DIFF
--- a/changelog/O-ejoMW-T62u6CYS5rBHcw.md
+++ b/changelog/O-ejoMW-T62u6CYS5rBHcw.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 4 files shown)


matt-boris:
Also closes https://github.com/taskcluster/taskcluster/issues/7722 :)

Eijebong:
Not really, there's another nock in the root packages.json (looking at it right now). I did however forget the changelog file

matt-boris:
Oh :rip: just noticed that 😭

Eijebong:
Yeah and looking at it... It's not as easy as this one was

matt-boris (approved):

PR #8568, opened 2026-04-30
Update nock to 14.0.13

### LINKED ISSUE #7722: Upgrade nock to v14+

Eijebong:
This updates nock and fixes worker-manager tests that were failing with version 14.

The test was hitting www.microsoft.com with a 1ms timeout which always triggered before the connection got established, resulting in the socket destruction. Nock 13 swallowed the resulting "connect ECANCELED". Nock 14 doesn't and it leaks as an uncaught exception instead. This is basically the same failure mode as #8567.

Instead of relying on hitting microsoft's servers I replaced it with a local http server that never responds, the connection is instantaneous, nock doesn't get angry at it failing and everything works smoothly. The one downside is that we're now mocking `downloadBinaryResponse` which I'm not a fan of, but we're testing the catch path anyway with this test so it's fine.

Fixes #7722
Closes #8565


package.json
```diff
@@ -165,7 +165,7 @@
     "mock-fs": "^4.8.0",
     "mockdate": "^3.0.2",
     "moment": "^2.30.1",
-    "nock": "13.5.6",
+    "nock": "14.0.13",
```
matt-boris:
```suggestion
    "nock": "^14.0.13",
```

Not sure why it was previously pinned, but this is probably fine to not pin?

Eijebong: Yeah probably, thanks for noticing

matt-boris (approved):
LGTM, thanks! Lol @ the branch name, as always with you 😄

PR #8576, opened 2026-05-04
chore(generic-worker): prefer x/sys/windows + syscall constants over locally redefined Win32 values

matt-boris:
>Replaced locally-redefined Win32 constants in `workers/generic-worker` with their equivalents from `golang.org/x/sys/windows` and `syscall`. No behavior change.


Eijebong (approved):
Looks correct.

PR #8578, opened 2026-05-04
Security: Missing HTTP timeout in GetURL function

tomaioo:
## Summary

Security: Missing HTTP timeout in GetURL function

## Problem

**Severity**: `Medium` | **File**: `clients/client-go/internal/geturl.go:L38`

The GetURL function in internal/geturl.go uses http.Get() without setting a timeout. This can lead to indefinite blocking if the remote server is unresponsive. The httpbackoff library may handle some cases, but the initial http.Get call has no timeout.

## Solution

Use http.Client with a configured Timeout instead of http.Get(). Create a client: client := &http.Client{Timeout: 30 * time.Second} and use client.Get(url) instead of http.Get(url)

## Changes

- `clients/client-go/internal/geturl.go` (modified)


clients/client-go/internal/geturl.go
```diff
@@ -35,7 +36,8 @@ func GetURL(httpBackoffClient *httpbackoff.Client, url string, writeSeeker io.Wr
 			// straight away
 			return
 		}
-		resp, tempError = http.Get(url)
+		client := &http.Client{Timeout: 30 * time.Second}
```
matt-boris: The [`http.Client.Timeout`](https://pkg.go.dev/net/http#Client) is a timeout for the entire request, including reading the body, so I don't think this is a correct fix. How it currently is implemented, a timeout could be hit while streaming a large artifact, which would be a bad experience/outcome.

I think if we'd like to fix the issue of `indefinite blocking if the remote server is unresponsive`, then we should look at setting the [`http.Client.Transport`](https://pkg.go.dev/net/http#Transport) settings like `ResponseHeaderTimeout` which accounts for the time to wait for the server's response headers, but _not_ the time to read the response body.

A minimal fix could look something like:

```go
var downloadHTTPClient = func() *http.Client {
    t := http.DefaultTransport.(*http.Transport).Clone()
    t.ResponseHeaderTimeout = 30 * time.Second
    return &http.Client{Transport: t}
}()
```

Preferably hoisted up to the package level (so it's not instantiated on each retry) and used via `downloadHTTPClient.Get(url)`.

Something to note though is this same issue appears here https://github.com/taskcluster/taskcluster/blob/443c9d06a371cd6d4221e56750e15f60e9e700c6/clients/client-go/tcobject/download.go#L102 so it would be nice to share the single client from `clients/client-go/internal/` to get a fix in over there as well.

matt-boris (changes requested):
Hi, thanks so much for your contribution! I've requested changes as the PR currently stands because it could potentially timeout during the streaming of a large artifact. I gave some suggestions on how to fix, alongside of an additional request.

Will you please also run `yarn changelog` to create a changelog file for this fix, that will address the failing `meta-changelog-pr` CI check.

PR #8579, opened 2026-05-04
feat(ui): display artifact sizes when reported by workers

### LINKED ISSUE #8537: Display artifact sizes in UI
since this information is available in API, we can also show it in the UI 

Note: not all workers submit this, so some records would not have any sizes

matt-boris:
Closes #8537

Adds a `contentLength` field to the GraphQL Artifact type and surfaces it in the task-runs and indexed-task artifact lists as a muted size label next to each artifact link. Workers that don't report a size render a blank cell.

>The web UI now displays artifact sizes in the task-runs and indexed-task artifact lists, when reported by the worker. Workers that do not report a size show a blank size column.

For https://community-tc.services.mozilla.com/tasks/RBPq5vCdRWSY3jz8M3_SFg, the UI changes produces:

Screenshot 2026-05-04 at 5 31 04 PM

### DIFF
--- a/ui/src/utils/formatBytes.js
+++ b/ui/src/utils/formatBytes.js
@@ -1,0 +1,23 @@
+const UNITS = ['B', 'KB', 'MB', 'GB', 'TB', 'PB'];
+
+export default function formatBytes(bytes) {
+  if (bytes === null || bytes === undefined || Number.isNaN(bytes)) {
+    return '';
+  }
+
+  if (bytes < 1) {
+    return `${bytes} B`;
+  }
+
+  const exponent = Math.min(
+    Math.floor(Math.log10(bytes) / 3),
+    UNITS.length - 1
+  );
+  const value = bytes / 1000 ** exponent;
+  const formatted =
+    exponent === 0 || value >= 100
+      ? value.toFixed(0)
+      : value.toFixed(1).replace(/\.0$/, '');
+
+  return `${formatted} ${UNITS[exponent]}`;
+}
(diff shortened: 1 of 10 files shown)


ui/src/utils/formatBytes.js
```diff
@@ -0,0 +1,23 @@
+const UNITS = ['B', 'KB', 'MB', 'GB', 'TB', 'PB'];
```
Eijebong, line 1: This is "wrong" for your math. Since you're doing power of 2 calculations it should be `KiB`, `MiB` and so on.

Or switch to 10 based math and use:

```js
new Intl.NumberFormat('en', {
  notation: 'compact',
  style: 'unit',
  unit: 'byte',
  unitDisplay: 'narrow',
}).format(bytes);
```

instead. Unfortunately I don't think that can do power of 2 calcs :/

matt-boris: Ugh hate all the different units based on the math being used. Thanks

matt-boris: I'll do base 1000 cuz that's what GitHub shows for release artifacts and I think S3 uses this as well. 🤷🏻

petemoore (approved):
This is a huge UX improvement!

Nit: it would be nice to have a test that only 3 significant figures are used, e.g. with a file size that is 185748752 that shows size as 186MB (rather than 185.748752 MB). I think all the tests at the moment match the exact size that they are displayed as. The manual test with your screenshot above nicely shows that this is the current behaviour though. Another thing that might be nice is a hover-over to show the exact file size (because sometimes you care about the size more precisely). But this is certainly good enough to go, thanks so much!

PR #8587, opened 2026-05-06
fix(w-m): use keyset pagination in scanner and provisioner

### LINKED ISSUE #8586: worker-manager scanner & provisioner use offset-based pagination, can mark sole workers `over capacity` and double-count capacity stats
## Summary

The worker-manager has two `paginatedIterator` call sites that use **offset-based** pagination over `get_non_stopped_workers_with_launch_config_scanner`. Both can yield the same worker row twice when the underlying result set is mutated mid-scan (a row inserted with a `worker_id` that sorts before the current cursor shifts subsequent offsets by one, causing the last row of a previous page to be returned again at the start of the next page).

- `services/worker-manager/src/worker-scanner.js` — duplicate rows in `candidates` cause `#evaluatePolicies` to assign one `terminate:false,'needed'` and one `terminate:true,'over capacity'` to the same `workerId`. Both decisions get persisted; last-write-wins on the same row leaves the worker flagged for termination. The worker calls `shouldWorkerTerminate`, gets `true`, exits. Active harm.
- `services/worker-manager/src/provisioner.js` — duplicate rows are double-counted by `WorkerPoolStats.updateFromWorker`, inflating `existingCapacity` / `runningCapacity` / `requestedCapacity` / per-launch-config and per-worker-group breakdowns. Inflated stats cause `estimator.simple()` to under-provision. Silent, only visible on pools larger than the page size.

The library README already warns about this hazard (`libraries/postgres/README.md`, lines 367–370) — *"An 'offset' into a set of results that is rapidly changing size is not accurate, and can easily return the same row twice or skip a row. … Use index-based pagination in such cases."* The worker-manager code simply hasn't been migrated.

## Evidence

- Production worker `mozilla-t/pre-commit/us-central1-a/8950556086655973951`: only running worker in a `minCapacity=1` pool, peak concurrent running over 24h = 1, `decided_at=2026-05-06T17:04:12Z` with `terminate:true, reason:'over capacity'`. The algorithm cannot produce that output for a sole candidate when `targetCapacity >= 1`, so `candidates` must have contained two or more entries pointing at the same `workerId`.
- Fleet-wide: across non-stopped workers in worker-manager, `provider_data->'shouldTerminate'->>'reason'` distribution is `over capacity: 435157`, `needed: 103547`, `launch config archived: 5791`. The 4:1 over-capacity:needed ratio is dominated by single-worker pools where the duplication-driven mis-marking fires more often than legitimate over-capacity.
- The bug presents most loudly on small pools (sole-worker pools see every duplication event as a kill signal). Large pools mask it because they can absorb a misclassified candidate among many `needed` ones.

This was previously partially captured in #8473 (framing it as scanner-staleness), but the scanner is fast — fxci-level1-gcp p95 since-last-check is approximately 8 minutes, max 18 minutes. Staleness is not the cause.

matt-boris:
Offset-based pagination over the workers table could yield duplicate rows when concurrent inserts shifted the result set, causing the scanner to mis-terminate the sole running worker in minCapacity pools and the provisioner to silently under-provision large pools. Switch both to keyset pagination via new DB function 0125.

Fixes #8586

>Fixes a bug in worker-manager where the worker-scanner could incorrectly mark a running worker as `over capacity` and tell it to terminate, even when it was the sole running worker in a pool with `minCapacity >= 1`. Caused by offset-based pagination yielding duplicate rows when other workers were inserted mid-scan, leading to conflicting termination decisions for the same worker. The provisioner had the same root cause, silently over-counting `existingCapacity` on pools with more than 1000 non-stopped workers and under-provisioning them. Both call sites now use keyset pagination via a new function `get_non_stopped_workers_with_launch_config_scanner_after` in DB version 0125. The most visible symptom was excessive worker churn on small pools.


db/versions/0125.yml
```diff
@@ -0,0 +1,71 @@
...
+            case
+              when providers_filter_cond_in = '='
+                then workers.provider_id = ANY(string_to_array(providers_filter_value_in, ','))
+              when providers_filter_cond_in = '<>'
+                then workers.provider_id <> ALL(string_to_array(providers_filter_value_in, ','))
+            end
+            ) and
+          (after_worker_pool_id_in is null or
+            (workers.worker_pool_id, workers.worker_group, workers.worker_id) >
+            (after_worker_pool_id_in, after_worker_group_in, after_worker_id_in))
```
petemoore, line 79: **Generated TS types are inconsistent for the three `after_*_in` args.**

The type generator at `infrastructure/tooling/src/utils/db.js:279-282` only marks an arg nullable if `<arg> is null` appears in the function body (or the arg is `page_size_in` / `page_offset_in`). With this body, only `after_worker_pool_id_in` is null-checked explicitly, so the generated `fns.d.ts` ends up with:

```ts
   after_worker_pool_id_in: string | null,
   after_worker_group_in: string,    // not nullable
   after_worker_id_in: string        // not nullable
```

But `paginatedIterator` calls `fetch(size, after)` with `{ after_worker_pool_id_in: null, after_worker_group_in: null, after_worker_id_in: null }` for the first page (see `libraries/postgres/src/util.js:98`), which violates those types. Runtime is correct because the SQL short-circuits on the first `is null`, but the types are wrong and any positional/typed caller would be tripped up by them.

Simplest fix is to make the `is null` predicate cover all three args so the generator picks them up — runtime semantics unchanged because of short-circuit evaluation:

```
  (after_worker_pool_id_in is null or after_worker_group_in is null or after_worker_id_in is null or
    (workers.worker_pool_id, workers.worker_group, workers.worker_id) >
    (after_worker_pool_id_in, after_worker_group_in, after_worker_id_in))
```

After regenerating, all three should come out as `?: string | null`.

db/test/fns/worker_manager_test.js
```diff
@@ -765,6 +765,111 @@ suite(testing.suiteName(), function () {
...
+
+      const last1 = page1[page1.length - 1];
+      const page2 = await db.fns.get_non_stopped_workers_with_launch_config_scanner_after(
+        null, null, null, null, null, 10,
+        last1.worker_pool_id, last1.worker_group, last1.worker_id);
+
+      // Must NOT contain id3 (already returned) and must NOT contain id1
+      // (sorts before the cursor). Must contain id5 and id7.
+      const ids = page2.map(r => r.worker_id);
+      assert.deepEqual(ids, ['id5', 'id7'],
```
petemoore, line 869: Neither new test exercises `providers_filter_cond_in` / `providers_filter_value_in`, even though that's the worker-scanner's actual call path. Worth a small case that mixes the providers filter with keyset pagination — easy to do here, and would catch any regression in the WHERE clause ordering.

db/versions/0125.yml
```diff
@@ -0,0 +1,71 @@
...
+    description: |-
+      Get non-stopped workers filtered by the optional arguments, ordered by
+      `(worker_pool_id, worker_group, worker_id)`. Uses keyset pagination via
+      the `after_*_in` parameters: pass nulls to fetch the first page, then
+      pass the last row's `(worker_pool_id, worker_group, worker_id)` to fetch
+      the next page. Unlike the offset-based variant
+      `get_non_stopped_workers_with_launch_config_scanner`, this is robust to
+      concurrent inserts and deletes during a long-running scan — the cursor
+      moves strictly forward through a unique key, so each row is returned at
+      most once.
```
petemoore, line 16: Worth spelling out the keyset trade-off explicitly here: rows whose key sorts **before** the current cursor (e.g. a worker inserted with a `worker_id` lex-less than the last-yielded row) are silently skipped for the rest of the current scan. Your second test demonstrates exactly this behavior, but a future reader skimming "robust to concurrent inserts" might assume "every row that exists at any point during the scan is returned", which isn't the case. For worker-scanner / provisioner this is fine — they re-scan periodically — but documenting it prevents the next caller from being surprised.

db/versions/0125.yml
```diff
@@ -0,0 +1,71 @@
+version: 125
+description: keyset pagination variant of get_non_stopped_workers_with_launch_config_scanner to avoid duplicate or skipped rows when the result set changes during a multi-page scan
+methods:
+  get_non_stopped_workers_with_launch_config_scanner_after:
```
petemoore, line 6: Optional / follow-up: now that both production callers have moved to the `_after` variant, consider marking `get_non_stopped_workers_with_launch_config_scanner` `deprecated: true` in this version (per the pattern in 0105/0114). The two remaining callers are `db/test/fns/worker_manager_test.js:702` and `services/worker-manager/test/data_test.js:40`, both tests — they'd need to migrate to `db.deprecatedFns.*` or to the new function. Not a blocker, but signalling intent now is cheap.

petemoore (commented):
Solid fix — diagnosis is correct (offset-pagination duplicates rows under concurrent inserts) and the keyset cursor `(worker_pool_id, worker_group, worker_id)` aligns with the `workers` primary key, so it's index-supported. Reuses the existing `paginatedIterator` `indexColumns` path, which is nice.

One issue worth fixing before merge — the generated TypeScript types for the new function are inconsistent (inline). Plus a couple of small suggestions on tests and docs.

Nice touch on the test that demonstrates the "insert behind cursor" semantics — that subtlety deserves to be exercised.

petemoore (approved):
Thanks Boris, looks great!

Issue #8589, opened 2026-05-07
add ESLint rule to prevent offset-pagination scan loops (followup to #8587)


petemoore:
The bug class fixed in #8586 / #8587 was an offset-paginated full-set scan loop in service code. While the keyset-pagination fix addressed the two known instances (worker-scanner and provisioner), the same pattern can reappear any time someone writes:

```js
let offset = 0;
while (true) {
  const rows = await this.db.fns.someListFn(..., size, offset);
  if (!rows.length) break;
  // process rows
  offset += size;
}
```

Under concurrent inserts/deletes, this loop yields duplicate or skipped rows. The audit that found #8588 confirmed the pattern is rare today (one remaining instance), but the same shape will keep being attractive to write — it's the obvious naïve translation of "paginate through everything."

## Proposal

Add a single \`no-restricted-syntax\` entry to the existing \`.eslintrc.yml\` matching the AST shape of this bug exactly, with a message explaining the issue and pointing at \`paginatedIterator\` from \`@taskcluster/lib-postgres\` as the correct replacement.

The rule should match:
- A \`let\`/\`var\` declaration initialised to \`0\` (the counter)
- A loop containing an \`await\` call to a \`db.fns.*\` member function with the counter as a trailing argument
- An \`<counter> += <number>\` increment in the same loop

This keeps the false-positive rate near zero — that exact AST shape *is* the bug — while still catching the next contributor who reaches for the obvious wrong pattern.

## Why ESLint vs. a separate CI script

Lives in the existing config, no plugin scaffolding, runs in the same \`yarn lint\` pass developers already have wired into their editors, and surfaces the violation inline at the offending line rather than at end-of-CI.

## Dependencies

Depends on #8588 being merged first — otherwise the rule would fail on the existing \`scoperesolver.js\` instance.

## References

- Bug class: #8586
- First fix: #8587
- Second fix (audit follow-up): #8588

Eijebong:
While Yarik's experimentation didn't go as amazingly in #8372 as biome's marketing promised, getting out of the eslint ecosystem should still be a goal because half of it isn't maintained.

IMO we shouldn't dig ourselves deeper in that ecosystem for this. That rule might've caught something years ago when the class of bugs was introduced and then copy/pasted around but:

- We're not adding one of those every week
- The bug is now fixed, no more risk of copy/pasting something wrong
- Honestly, this class of bug is very well known and should get caught at review time by anyone  with a background in relational DB engineering

[...]

Issue #8588, opened 2026-05-07
auth scoperesolver uses offset pagination during cache reload — same root cause as #8586


petemoore:
While auditing call sites for the offset-pagination bug class fixed by #8587, I found that `services/auth/src/scoperesolver.js:193-200` uses the same dangerous pattern:

```js
let offset = 0;
while (true) {
  const rows = await this.db.fns.get_clients(null, 1000, offset);
  if (rows.length === 0) {
    break;
  } else {
    offset += 1000;
  }
  // ... process clients
}
```

This loop walks the full `clients` table to rebuild the in-memory client cache used for scope resolution. Concurrent client inserts/deletes during the reload shift rows between pages, so a client can be silently skipped (or returned twice) — exactly the same offset-pagination race fixed in #8586/#8587 for the worker-scanner.

## Severity

Lower than the worker-scanner case — this is an eventually-consistent cache, not a termination decision — but real:
- A skipped client briefly fails authentication for that client until the next reload.
- A duplicated client wastes work and slightly distorts the resolver's cache.
- The race repeats every reload cycle whenever clients are being created/expired during the same window.

## Fix

Same shape as #8587: add a keyset-paginated DB function `get_clients_after` (in DB version 0126), then migrate the scoperesolver scan loop to use it via `paginatedIterator` from `@taskcluster/lib-postgres`. The existing `get_clients` will be marked deprecated, and the auth API handler (`auth/src/api.js`) and `static-clients.js` migrated to the new function as well.

## Audit context

The full audit found six internal full-set scan loops in `services/`. Five are already safe (keyset, self-terminating, or DB-side single-call). This is the only remaining one with the offset-pagination race.

A separate follow-up will add an ESLint rule banning the offset-loop pattern to prevent regressions.

## References

- Original bug: #8586
- Original fix (worker-scanner / provisioner): #8587

PR #8590, opened 2026-05-07
fix(auth): use keyset pagination in scoperesolver client cache reload

petemoore:
Same offset-pagination bug class as #8586/#8587, this time in the auth service. The scoperesolver's client-cache reload walks the `clients` table with `let offset = 0; while (true) { ... offset += 1000 }` — concurrent client inserts/expirations during a reload shift rows between pages, so a client can be silently skipped or returned twice.

Symptom: transient authentication failures for newly-created clients until the next reload cycle (default cache-reload interval). Lower severity than the worker-scanner case but the same root cause.

Adds keyset-paginated DB function `get_clients_after` in version 0126 (mirroring 0125's `get_non_stopped_workers_with_launch_config_scanner_after` from #8587). All three live callers of `get_clients` are migrated:

- `services/auth/src/scoperesolver.js` — internal full-set scan, the actual bug. Now uses `paginatedIterator` from `@taskcluster/lib-postgres`.
- `services/auth/src/api.js` (`listClients` endpoint) — switches `paginateResults` to `indexColumns: ['client_id']`. User-visible continuation-token format changes (offset hashids → base64 keyset), but tokens are documented as opaque and `paginateResults` already gracefully falls back to "start from beginning" on undecodable tokens, so cross-deploy clients aren't broken.
- `services/auth/src/static-clients.js` — trivial.

`get_clients` is marked `deprecated: true`; with no live callers it remains accessible only via `db.deprecatedFns` for backwards compatibility.

Tests: existing offset-pagination tests are replaced with equivalent keyset tests in `db/test/fns/auth_test.js`, plus a new test that exercises the keyset cursor's robustness to a mid-scan insert.

## References

- Bug class: #8586
- Original fix (worker-manager): #8587
- This issue: #8588
- Follow-up lint rule (depends on this merging): #8589

## Test plan

- [x] `yarn lint` — passes on changed files
- [x] `yarn generate` — schema and types regenerated cleanly
- [x] `db/test/fns/auth_test.js` — 25/25 passing (incl. 4 new keyset tests)
- [x] `services/auth` — 601/601 passing locally
- [x] CI green
- [x] Pete review


services/auth/src/api.js
matt-boris, line 1: https://github.com/taskcluster/taskcluster/blob/b892f15cc35626094029e8d29c22618a5e90e0d7/services/auth/src/api.js#L217-L220

The API description is now a bit stale. A `continuationToken` is now only returned when there's for sure more results to return.

matt-boris (commented):
Changes LGTM, holding off on a formal approval until a small API description is fixed. Everything else looks solid, thanks.

lotas (changes requested):
I think we don't really need this change, same as the similar pr about clients. We don't have so many clients, so reloading issues are highly likely do not exist, because on every change new pulse message is fired, so it would reload all clients as long as events are coming in. 

I believe we should think about the ways to reduce the load on the system during such updates, because 50+ pods fetching all clients and roles all at once is adding a lot of stress (which is currently swallowed by the compute we have)

upd: this is probably ok for other use cases outside of internal scope resolver. If someone is using pagination to retrieve those it could hit those issues. 

https://github.com/taskcluster/taskcluster/pull/8590/changes#r3203562870 is this still worth addressing before merging though?
p.s. yarn generate ;)

PR #8591, opened 2026-05-07
chore(lint): ban offset-paginated full-set scan loops over db.fns.*

petemoore:
Adds a `no-restricted-syntax` ESLint rule matching the exact AST shape of the bug class fixed in #8586 / #8587 / #8588: a loop containing both an `await db.fns.X(...)` call and an `<id> += <literal>` counter increment. The rule message points new contributors at `paginatedIterator` from `@taskcluster/lib-postgres` as the correct keyset alternative.

## Selector

```
:matches(WhileStatement, DoWhileStatement, ForStatement, ForInStatement, ForOfStatement)
  :has(AwaitExpression CallExpression MemberExpression[property.name="fns"])
  :has(AssignmentExpression[operator="+="][right.type="Literal"])
```

A loop is flagged only when **both** conditions hold simultaneously inside the loop body:

1. An `await db.fns.X(...)` call (member access via `.fns`).
2. An `<id> += <literal>` increment (the counter advancement).

The combined match has near-zero false-positive rate: this combination is *exactly* the offset-pagination bug shape and is virtually never the right thing to write deliberately.

## Verified across the codebase

After running `yarn lint` against all `services/`, `libraries/`, and `db/` JS, only one site matched: `db/test/versions/0122_test.js`, where the test deliberately exercises the offset-paginated `get_tasks_by_task_group_projid` function in a no-concurrent-writes environment to verify stable `ORDER BY` ordering. Suppressed inline with a one-line `eslint-disable-next-line` comment explaining the intent.

The rule was tested manually against fixtures covering:
- the bug pattern (fires correctly)
- healthy keyset pagination using `paginatedIterator` (no false positive)
- counter-only loop with `+=` literal but no `db.fns.*` call (no false positive)
- `db.fns.*` call with non-literal `+=` (e.g. `cumulativeWeight += config.weight` from `launch-config-selector.js` — no false positive)

## Stacked on #8590

This branch is based on the commit from #8590 (the scoperesolver fix). Until #8590 is merged, this PR's diff includes both commits; once #8590 lands the diff will clean up to just this rule + suppression. **Do not merge this until #8590 is merged.**

## References

- Bug class: #8586
- First fix (worker-manager): #8587
- Second fix (scoperesolver): #8588 / #8590
- This issue: #8589

## Test plan

- [x] `yarn lint` — full repo passes
- [x] Manual fixture tests confirm rule fires on bug pattern, not on healthy variants
- [x] CI green
- [x] Pete review


lotas:
I don't think we need to change anything here, existing query uses page size `1000` , and currently community-tc has `29` clients, and fxci has `424`

Plus I believe that whole architecture around those full reloads should be likely changed, since `auth` service is the most heavy one, we could have >50 pods running at the same time, and on each client change it would fire those expensive full reads. Which makes it even worse when several clients are updated at once through config change. 

So I'd probably not touch this for now

petemoore:
OK, let's drop this then, thanks.

PR #8594, opened 2026-05-08
docs(worker-manager): drop unused Azure permissions from example role

### LINKED ISSUE #8595: Revise Azure deployment docs to clarify minimum role and resource-group strategy
## Background

A recent Matrix conversation with Callek revealed that the Azure provider deployment docs were carrying a number of `Actions` in the example custom role that the worker-manager has never actually exercised — including `Microsoft.Support/*` and `Microsoft.Insights/alertRules/*`, which his security team flagged as unjustified. The original example role (PR #2197, Jan 2020) was a near-verbatim copy of Microsoft's built-in *Virtual Machine Contributor* role plus a couple of additions, and was never trimmed as the worker-manager evolved.

A second gap surfaced during the audit: there is no documentation of the **resource-group strategy** trade-offs. With ARM-template provisioning (PR #7977, Oct 2025) a launch configuration can use `armDeploymentResourceGroup` to deploy each pool — or each launch configuration — into its own resource group, with worker-manager auto-creating the resource group on demand. This is a designed feature with real operational benefits (per-pool auditing, cleanup, quota isolation, RBAC), but using it requires `Microsoft.Resources/subscriptions/resourceGroups/write` on the role, and Azure RBAC alone has no way to scope that grant to a name prefix. The pattern needs to be documented alongside the recommended Azure Policy pairing that bounds it.

## Scope

1. Audit every `Action` in the example custom Azure role in `ui/docs/manual/deploying/worker-manager-config.mdx` against the worker-manager codebase and typical worker ARM templates; drop entries that are not actually used.
2. Reframe the example as the *base set* the worker-manager itself exercises, with launch-configuration ARM templates allowed to require additional `Actions` on top, and link to Microsoft's resource-provider-operations reference so deployers can look up additions.
3. Add a "Resource Group Strategy" section to `ui/docs/reference/core/worker-manager/azure.mdx` covering the single-shared-RG vs per-launch-config-RG patterns, when to use each, and the recommended Azure Policy pattern for bounding `resourceGroups/write`.
4. Update the stale "within a single Resource Group" line at the top of `azure.mdx` that predates ARM-template support.

## Status

Implementation in PR #8594.

petemoore:
Closes #8595

### DIFF
--- a/ui/docs/manual/deploying/worker-manager-config.mdx
+++ b/ui/docs/manual/deploying/worker-manager-config.mdx
@@ -194,31 +194,33 @@ An Azure-based provider is configured in `providers` with the following structur
 }
 ```
 
-Create a new resource group, in which all resources will be created (note that the location of the resource group is not related to the location of the workers).
+Create a new resource group, which will be used as the default destination for the worker resources this provider creates (note that the location of the resource group is not related to the location of the workers).
+Launch configurations may also opt into per-pool resource groups via `armDeploymentResourceGroup`; see [Resource Group Strategy](/docs/reference/core/worker-manager/azure#resource-group-strategy) for the trade-offs and the recommended Azure Policy pairing.
 
 Create a new "App Registration" as well, but do not create secrets for it yet.
 It must first be granted appropriate permissions.
 See [here](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal) for information on creating Azure AD Applications, assigning roles, and getting credentials.
 
 Azure-based providers need permissions to manage virtual machines, network interfaces, disks, and public IP addresses.
-The following is an example custom Azure role definition that encompasses the scope of permissions an Azure-based provider needs.
-Add this under your subscription -> "IAM" -> "Create a Custom Role", and start from JSON using content like this:
+The example custom role below is the **base set of `Actions`** that the worker-manager itself exercises — i.e. the permissions required regardless of which launch configurations you define.
+A worker pool's launch configuration may reference additional resources via its ARM template (e.g. availability sets, scale sets, custom NSG rules, boot-diagnostics storage accounts, or any other Azure resource type), and the role will need the corresponding `Actions` added on top before that pool can provision.
+The full list of available actions per resource provider is at [Azure resource provider operations](https://learn.microsoft.com/en-us/azure/role-based-access-control/resource-provider-operations).
+
+Add the role under your subscription -> "IAM" -> "Create a Custom Role", and start from JSON using content like this:
 
 ```json
 {
   "Name": "Worker-Manager Azure Provider",
   "IsCustom": true,
-  "Description": "Based on Virtual Machine Contributor + image access, delete access",
+  "Description": "Minimum permissions for the Taskcluster worker-manager Azure provider.",
   "Actions": [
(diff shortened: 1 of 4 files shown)


petemoore:
cc @Callek

ui/docs/manual/deploying/worker-manager-config.mdx
```diff
@@ -194,31 +194,30 @@ An Azure-based provider is configured in `providers` with the following structur
...
 ```json
 {
   "Name": "Worker-Manager Azure Provider",
   "IsCustom": true,
-  "Description": "Based on Virtual Machine Contributor + image access, delete access",
+  "Description": "Minimum permissions for the Taskcluster worker-manager Azure provider.",
   "Actions": [
     "Microsoft.Authorization/*/read",
-    "Microsoft.Compute/availabilitySets/*",
-    "Microsoft.Compute/locations/*",
```
lotas, line 215: I wonder if it was necessary in the past to poll long running operations:
https://github.com/lotas/taskcluster/blob/7a1a02e4b4eb4dd734d15244e209575e169f9bc8/services/worker-manager/src/providers/azure/index.js#L1280-L1286

might be irrelevant since we switched to newer sdk with a different operation url format

lotas:
```
"operation": "https://management.azure.com/subscriptions/123-cdef/operationresults/eyJ......n0
```
I wonder if this url requires some documented permission as well

ui/docs/manual/deploying/worker-manager-config.mdx
```diff
@@ -229,17 +228,20 @@ Add this under your subscription -> "IAM" -> "Create a Custom Role", and start f
     "Microsoft.Network/virtualNetworks/read",
     "Microsoft.Network/virtualNetworks/subnets/join/action",
     "Microsoft.Resources/deployments/*",
-    "Microsoft.Resources/subscriptions/resourceGroups/read",
-    "Microsoft.Storage/storageAccounts/listKeys/action",
-    "Microsoft.Storage/storageAccounts/read",
-    "Microsoft.Support/*"
```
lotas, line 235: We should probably add `templateSpecs/*/read` something role, since we added those 😬  I forgot to check that back then

lotas (changes requested):

ui/docs/manual/deploying/worker-manager-config.mdx
petemoore: Whoopsie. Good call - it looks like we still need:

- `Microsoft.Compute/locations/operations/read`
- `Microsoft.Network/locations/operations/read`
- `Microsoft.Network/locations/operationResults/read`

Adding back...

ui/docs/manual/deploying/worker-manager-config.mdx
petemoore: note, i added it to the doc comments below rather the template scopes, as this is opt-in

petemoore:
@lotas Are you happy with this now?

ui/docs/manual/deploying/worker-manager-config.mdx
lotas: looks good to me, thanks

lotas (approved):
Thanks! LGTM

PR #8555, opened 2026-04-30
Remove docker-worker from the monorepo

petemoore:
## Summary

- Deletes `workers/docker-worker/`, `ui/docs/reference/workers/docker-worker/`, and `tools/worker-runner/worker/dockerworker/`.
- Moves the docker-worker payload schema to `tools/d2g/schemas/docker-worker/v1/payload.yml` so d2g (the only remaining consumer) owns it. The published service-schema URL `/schemas/docker-worker/v1/payload.json` is unchanged.
- Cleans up build/CI tooling that referenced the deleted tree (node-version generator, minify task, dependabot config, root `package.json` lint scripts).
- Removes the `'docker-worker'` entry from the UI's `TASK_PAYLOAD_SCHEMAS`. The two callers in `ui/src/views/Tasks/CreateTask/index.jsx` that used `'docker-worker'` as a fallback now fall back to `'generic-multi-posix'`.
- Reframes user-facing documentation to make clear that legacy docker-worker payloads run on generic-worker via d2g. The payload examples themselves stay valid — generic-worker's posix schemas accept the docker-worker format as a `oneOf` branch.
- Mozilla CI's `taskcluster/config.yml` and `taskcluster/src/job.py` keep their `implementation: docker-worker` annotations (they describe the payload format their transforms emit, not the runtime). New comments make this explicit.
- Reword the queue API's error-artifact description from docker-worker to generic-worker.
- Major-level changelog entry covering deployer impact (worker-runner removal, `SITE_SPECIFIC.tutorial_worker_schema` migration, schema-path move with stable public URL).
- d2g, the `docker-worker:*` scope namespace, and tasks using the docker-worker payload format are unchanged.

Closes #8437.

## Test plan

- [x] `yarn lint` — clean
- [x] `yarn lint:go` — clean (both `multiuser` and `insecure` engines)
- [x] `yarn lint:py` — clean
- [x] `yarn generate` — produces only the expected auto-generated diffs (regenerated clients reflect the queue API reword; regenerated docs index reflects the deleted reference docs)
- [x] `tools/d2g` `go generate ./...` produces byte-identical generated types
- [x] `tools/d2g`, `tools/worker-runner`, `workers/generic-worker` (D2G/Mounts subset) Go tests pass
- [x] CI green

PR #8596, opened 2026-05-08
Remove the interactive viewer for the now removed docker-worker

Eijebong:
Docker-worker was removed by #8555, the v1 of the interactive viewer was specially made for it and is thus useless now. Just in case someone would somehow stumble their way onto this, I added a nice message telling them that they should be using generic-worker instead.

I initially wrote the changelog as a `major` change with some reservation since docker-worker was already removed and thus this wasn't exactly a supported feature getting removed. Upon agreement in the review, I downgraded it to `patch`.

### DIFF
--- a/changelog/bV-iyQTcQdadZeiclzug2A.md
+++ b/changelog/bV-iyQTcQdadZeiclzug2A.md
@@ -1,0 +1,4 @@
+audience: users
+level: patch
+---
+Removed the interactive shell viewer for docker-worker in the UI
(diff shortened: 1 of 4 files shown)


changelog/bV-iyQTcQdadZeiclzug2A.md
```diff
@@ -0,0 +1,4 @@
+audience: users
+level: major
```
petemoore: agree this could be `patch` since docker worker no longer supported

petemoore (dismissed):
Thanks Bastien for the cleanup!

changelog/bV-iyQTcQdadZeiclzug2A.md
Eijebong: Downgraded to patch

petemoore (approved):

PR #8598, opened 2026-05-08
Replace the broken scope diff viewer with a maintained dependency

Eijebong:
The `DiffTextArea` component used by `RoleForm` and `ClientForm` to preview scope edits was rendering incorrectly and was always showing "Files changed (0)". Now because it's using `react-gh-like-diff` which is very out of date, has been unmaintained for 6 years, I didn't bother trying to fix it in place and instead opted into replacing it with `react-diff-viewer-continued` which itself is a maintained version of `react-diff-viewer`. I tried to keep what I think it must've looked like based on the previous styling.

The second objective was also to get rid of `react-gh-like-diff` because it only supported react 16 which would be blocking the update to react 17+. That's one less blocker there.

Now, `react-diff-viewer-continued` pulls in `@emotion/{css,react}` as transitive dependencies which makes this patch look a lot scarier than it is. Yes those are new dependencies, but they're requirements to update MUI past v4 (which has been EOL since 2021...). So this is just paying that cost slightly earlier rather than introducing a dependency we'd otherwise avoid.

I also had to tell babel about `react-diff-viewer-continued` because it ships modern syntax (`??`, `?.`). The existing
`nullish-coalescing-operator` and `optional-chaining` plugins lower it to syntax that webpack 4 can parse. This entry can be removed once webpack is upgraded to 5+.

---

#### Before / After

image

image


matt-boris (approved):
Nice, thanks! LGTM

PR #8599, opened 2026-05-08
fix warnings on ui server startup

petemoore:
Reduces some of the noise on ui server startup.

### DIFF
--- a/ui/src/views/Quickstart/index.jsx
+++ b/ui/src/views/Quickstart/index.jsx
@@ -485,7 +485,7 @@ export default class QuickStart extends Component {
                 onChange={this.handleInputChange}
                 fullWidth
                 multiline
-                rows={3}
+                minRows={3}
                 value={taskDescription}
               />
             </ListItem>
@@ -646,7 +646,7 @@ export default class QuickStart extends Component {
           <Button
             spanProps={{ className: classes.resetButtonSpan }}
             tooltipProps={{ title: 'Reset Form & File' }}
-            variant="round"
+            variant="circular"
             onClick={this.handleReset}
             color="secondary">
             <RestartIcon />
--- a/changelog/hUuuatxukFZvel7dBty3.md
+++ b/changelog/hUuuatxukFZvel7dBty3.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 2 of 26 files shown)


ui/src/views/Quickstart/index.jsx
```diff
@@ -646,7 +646,7 @@ export default class QuickStart extends Component {
           <Button
             spanProps={{ className: classes.resetButtonSpan }}
             tooltipProps={{ title: 'Reset Form & File' }}
-            variant="round"
+            variant="circular"
```
Eijebong, line 649: Before/After on `/quickstart`

Image

I'm actually surprised you're not getting warnings about this because:

https://github.com/taskcluster/taskcluster/blob/36251c824740f42e6aa97be3a7e916f1658ac3f3/ui/src/components/Button/index.jsx#L39

This also changes the component because of https://github.com/taskcluster/taskcluster/blob/36251c824740f42e6aa97be3a7e916f1658ac3f3/ui/src/components/Button/index.jsx#L87

Eijebong:
```
> rg variant=\"circul | wc -l
1
> rg variant=\"round | wc -l
29
```

This is also inconsistent with the rest of the codebase which uses round everywhere.

Eijebong (dismissed):

ui/src/views/Quickstart/index.jsx
petemoore: OK, the warning was coming from:
https://github.com/mui/material-ui/blob/a9715056ab92595f3e62b045837edbeddc359579/packages/material-ui/src/Fab/Fab.js#L221-L229

we have this wrapper:
https://github.com/taskcluster/taskcluster/blob/e15efa7b495e35edd94352e507835a68f5cecbf1/ui/src/components/Button/index.jsx#L87-L88

so when i updated to use the newer term, we ended up with the default `MuiButton`. 🙄

I've fixed the wrapper now to respect the new term, and visually checked the UI that it now renders properly, and can confirm the warning has disappeared. Thanks for the review!

lotas (dismissed):
LGTM also tested locally, didn't notice any major issues

lotas (approved):

PR #8600, opened 2026-05-08
Drop our dependency on `json-loader`

Eijebong:
This has been redundant with webpack since webpack v2 (2016). Their readme says:

> Since webpack >= v2.0.0, importing of JSON files will work by
> default. You might still want to use this if you use a custom file
> extension.

The repository for it is archived since 2018, before it was even introduced in this project (b5a07368291a598f0c2c4289225f3a7974d44b91).

The only non trivial use is for the contributors page which needed custom file extension support for the .all-contributorsrc file. But that was also fixed *before* `json-loader` was introduced in this repo (The ability to specify a rule type for a file was added in webpack 4, which this repo was on before b5a07368291a598f0c2c4289225f3a7974d44b91 already) so it had no reason to be here in the first place.

### DIFF
--- a/changelog/IFi3e7EsTU2flROVUy76sg.md
+++ b/changelog/IFi3e7EsTU2flROVUy76sg.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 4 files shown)


matt-boris (approved):
Be gone! Thanks

PR #8601, opened 2026-05-08
Drop a few UI tests related dependencies

Eijebong:
Drop our dependency on react-addons-test-utils/mocha/chai

React-addons-test-utils was useless from the beginning. It was added alongside enzyme in 6c92bd4497a324cceac59a7a5a16414de234164c (karma -> jest migration) but is only useful with react 15 (from the package description: `This package is deprecated as of version 15.5.0, TestUtils have been moved to react-dom/test-utils`).
I'm guessing it was added by mistake and then got overlooked. Enzyme was already using `react-dom/test-utils`.

Mocha/chai have been useless since 2022. Both came in as part of the monorepo migration and were used at the time. Then in 6c92bd4497a324cceac59a7a5a16414de234164c, the test runner switched from `karma` to `jest` which also dropped the use of `chai` and `mocha`. However the patch didn't drop those dependencies.

No mention left of any of these in the codebase, tests are still passing, let's drop everything.


matt-boris (approved):
Nice

PR #8602, opened 2026-05-08
chore: add node/go upgrade scripts

matt-boris:
>Adds scripts to automate Node.js and Go version upgrades.

### DIFF
--- a/changelog/node-go-upgrade-script.md
+++ b/changelog/node-go-upgrade-script.md
@@ -1,0 +1,4 @@
+audience: developers
+level: patch
+---
+Adds scripts to automate Node.js and Go version upgrades.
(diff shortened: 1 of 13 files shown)


Eijebong:
Any reason to handroll a bash script instead of reusing the tooling that already exists for similar tasks in `infrastructure/tooling`?

matt-boris:
Good point, I'll update it to use the existing tooling. Going to keep a small shell wrapper to call out to `gvm install/use` and `nvm install/use`

lotas:
This is awesome! Thanks! 
It works, although I'm probably blacklisted by github :D 
image

lotas (approved):

Issue #8606, opened 2026-05-08
worker-manager: verify Azure cycle-time improvement after #8574 fix is deployed


petemoore:
## Goal

After the inline `beginDelete` fix from #8574 is deployed, re-collect the lifecycle-gap and cycle-time metrics for the same Mozilla Azure pools and confirm whether p50 actually moved as predicted.

## Predicted impact (from pre-deploy analysis)

Comments on #8574 captured baseline measurements on `gecko-t/win11-64-25h2`, fx-ci, 6 h on 2026-05-08 07:30–13:30 UTC:

- **API-call lifecycle gap (worker-removed → first scanner observation of STOPPING):** p50 14.9 min, p90 29.4 min
- **Cycle (worker-removed → worker-stopped):** p50 40.8 min, p90 81.6 min
- **Decomposition:** lifecycle gap ≈ ½ scanner pass; cleanup ≈ 1 scanner pass; cycle ≈ 1.5 passes
- **Predicted post-fix:** lifecycle gap → ~0; cycle p50 → ~26 min (1 cleanup pass alone) = **~37% reduction** for the API-call case

## What to do

1. After the fix is deployed (worker-manager release containing #8574 fix), wait at least 24 h for the system to stabilize.
2. Pull the same metrics over a window of comparable length and time-of-day to the baseline (off-peak Pacific, 6 h, single pool `gecko-t/win11-64-25h2`).
3. Compute and post both the lifecycle gap and cycle-time distributions to #8574 as a follow-up comment.
4. Compare against the prediction:
   - If cycle p50 dropped to ~26 min → fix worked as expected; close.
   - If cycle p50 dropped less than predicted → investigate (e.g., did pass duration change too? did the predicate skip too many cases?).
   - If no change → roll back; the model was wrong.
5. Optionally extend to a multi-pool / multi-day window for a more robust number.

## Reproduction

The methodology is documented in #8574 (jmoss's original) and in the validation comment. tc-logview filters and jq snippets are in the validation thread.

## Related

- #8574 — original issue + fix PR
- #8605 — replay harness (would let us answer this question in CI in seconds, but doesn't replace post-deploy validation against real production)

lotas:
looks like it was deployed over a month ago, do you want to follow up on this @petemoore ?

Issue #8605, opened 2026-05-08
worker-manager: scanner replay harness driven by captured production traces


petemoore:
## Goal

Build a deterministic, in-process scanner replay harness that drives `WorkerScanner` over a captured trace of real production worker lifecycle events, so we can measure scanner-architecture changes (e.g. inline `beginDelete`, STOPPING fast lane, per-worker parallelism) against realistic load **without spawning Azure VMs**.

## Inputs

Captured from tc-logview for an arbitrary window/pool/provider:

- `worker-requested`, `worker-running`, `worker-removed`, `worker-stopped` — per-worker timeline
- `cloud-api-metrics` — per-pass Azure API call latency distribution
- `scan-seen` — per-pass per-pool worker visibility, real pass durations
- `cloud-api-paused` / `cloud-api-resumed` — queue throttling events (relevant for incident replays)

## Components

1. **Capture CLI** — wrap tc-logview into a single command that pulls all required signals for a given window/pool and writes a self-contained trace bundle (jsonl files + manifest).
2. **Replay harness** — runs the actual `WorkerScanner` + a fake Azure provider + a fake DB. Fake clock advances to drive scanner passes. Fake Azure provider implements per-resource state machines (`Running` → on `beginDelete` → `Deleting` → after sampled latency → `404`) calibrated from the trace's observed cycle times.
3. **Reference trace(s)** — small captured high-load and low-load traces committed under `services/worker-manager/test/traces/` for regression testing.
4. **Test asserting cycle behaviour** — e.g. "with the inline-beginDelete fix, p50 cycle on the high-load trace is < X minutes" — runs in CI as a permanent regression artifact.

## Fidelity boundaries

- **High fidelity:** worker arrival/removal timeline, scanner pass durations, queue throttling.
- **Medium fidelity:** Azure API call latencies (sampled from aggregated p50/p95/p99 per period — no per-call timing in logs).
- **Modeled (low fidelity):** per-resource Azure state transitions during cleanup (NIC/IP/disks Running→Deleting→404). Calibrated from observed cycle times rather than measured directly.

## Why

Today the only way to validate a scanner change against real load is a production deploy followed by 24-48h of observation (see #8606). A replay harness:

- gives a CI-visible "did this change improve cycle time" answer in seconds rather than days;
- lets us re-run captured incident traces against architectural alternatives (STOPPING fast lane, parallelism) before committing to one;
- protects against regressions when worker-scanner internals are refactored.

## Out of scope

[...]

PR #8607, opened 2026-05-08
fix(w-m): submit Azure VM beginDelete inline in removeWorker (#8574)

### LINKED ISSUE #8574: worker-manager: Azure VM deletion waits for next scanner pass after removeWorker
## Summary

In the Azure provider, `workerManager.removeWorker` marks a worker `STOPPING` but does not start deleting the VM. The first Azure-side `virtualMachines.beginDelete()` call happens later, when `worker-scanner` reaches that worker.

That works functionally, but it leaves a billed-running tail after the worker process has exited. The scanner walks every non-stopped worker (REQUESTED, RUNNING, and STOPPING), so a freshly STOPPING worker can wait a long time before its VM delete is even submitted.

Mozilla production data for Apr 14-30, 2026 shows the shape of the problem:

- 44 Windows Azure worker pools
- 208,457 worker removals
- 91.0% removed via `workerManager.removeWorker API call`
- p50 delay from `worker-removed` to first VM deprovision log: about 30 min
- p90 delay: about 1.5 h
- 143,374 lifecycle-gap VM-hours over 17 days

This is not primarily a `deprovisionResources` bug. When the scanner reaches a STOPPING VM, `deprovisionResources()` passes `this.computeClient.virtualMachines` into `deprovisionResource()`, which then calls `client.beginDelete(...)`. For that resource type, this is the VM `virtualMachines.beginDelete()` call.

Current v99.2.0 also already has the Azure GET/404 convergence fix from #8530, and the earlier delete-branch 404 handling from #8518. Those fixes help the cleanup path converge once the scanner gets there. The gap here is before that: the first VM delete request waits for the scanner instead of starting from the API path that requested removal.

## To reproduce

This should show up in any busy Taskcluster deployment using the Azure provider.

1. Pick an Azure worker pool with steady termination volume, roughly >50 removals/hour.
2. Pull `worker-manager.worker-removed` events where `reason="workerManager.removeWorker API call"`.
3. Pull `worker-manager-workerscanner-azure` logs with `jsonPayload.Fields.message="deprovisionResource for vm with index undefined"`.
4. Join on `workerId`.
5. Compute `delay = vm_deprovision_event_time - worker_removed_event_time`.

For Mozilla's production data, the distribution looked like this:

| Delay | Count | Share |
|---|---:|---:|
| <15 min | 58,090 | 28.9% |
| 15 min-1 h | 106,392 | 52.9% |
| 1-2 h | 32,563 | 16.2% |
| 2-4 h | 3,661 | 1.8% |
| 4-8 h | 219 | 0.1% |
| 8 h+ | 6 | <0.01% |

The `vm-deprovision` log used here fires at the start of `deprovisionResource()` for the VM resource type, before the Azure delete call. So this measures scanner first-visit latency, not Azure delete latency.

## Expected behavior

When a worker calls `workerManager.removeWorker`, worker-manager should submit the Azure VM delete promptly for workers that have a provisioned VM.

The scanner should still reconcile state, retry failures, and finish cleanup. It just should not be the normal path for the first VM delete request.

## Taskcluster version

Observed on v99.2.0 (`bc2efe2827`). The same deferred-delete path appears in 99.0.3, 99.1.0, and 99.1.1.

## Platform

- Service: worker-manager
- Provider: Azure (`azure2`)
- Mozilla deployment: worker-manager on GKE in `moz-fx-webservices-high-prod`
- Client OS/browser: N/A

GCP and AWS may have a similar pattern, but the data above is only for Azure.

## Code references

API handler:

```js
// services/worker-manager/src/api.js:982
await provider.removeWorker({ worker, reason: 'workerManager.removeWorker API call' });
```

Azure `removeWorker`:

```js
// services/worker-manager/src/providers/azure/index.js:2056-2072
async removeWorker({ worker, reason }) {
  const shouldEmit = [Worker.states.REQUESTED, Worker.states.RUNNING].includes(worker.state);
  if (shouldEmit) {
    await this.onWorkerRemoved({ worker, reason });
  }
  // transition from either REQUESTED or RUNNING to STOPPING, and let the
  // worker scanner take it from there.
  await worker.update(this.db, w => {
    const now = new Date();
    if ([Worker.states.REQUESTED, Worker.states.RUNNING].includes(w.state)) {
      w.lastModified = now;
      w.state = Worker.states.STOPPING;
    }
    w.providerData.reasonRemoved ??= reason;
  });
}
```

Scanner handoff:

```js
// services/worker-manager/src/providers/azure/index.js:1701-1703
if (worker.state === states.STOPPING) {
  await this.deprovisionResources({ worker, monitor });
  return;
}
```

The scanner loop is in `services/worker-manager/src/worker-scanner.js:29-127`. It sleeps 20 seconds between passes, but each pass scans all non-stopped workers for that provider.

Current VM delete path in v99.2.0:

- `deprovisionResources()` passes the VM client into the shared helper: [`azure/index.js:2090-2095`](https://github.com/taskcluster/taskcluster/blob/v99.2.0/services/worker-manager/src/providers/azure/index.js#L2090-L2095)
- The shared helper does the pre-flight GET/404 check: [`azure/index.js:1985-1999`](https://github.com/taskcluster/taskcluster/blob/v99.2.0/services/worker-manager/src/providers/azure/index.js#L1985-L1999)
- The shared helper submits the delete request with `client.beginDelete(...)`: [`azure/index.js:2010-2018`](https://github.com/taskcluster/taskcluster/blob/v99.2.0/services/worker-manager/src/providers/azure/index.js#L2010-L2018)

The GET/404 behavior was added by the merged PR #8530 (`fix(w-m): always pre-flight GET in Azure deprovisionResource`). The earlier delete-branch 404 fallback was added by #8518 (`feat(w-m): Fix azure stuck resource deletion`).

## Proposed fix

Start the VM delete in Azure `removeWorker` after the worker is moved to `STOPPING`, but only when the VM is known to exist. The scanner remains the fallback and verifier.

Sketch:

```js
async removeWorker({ worker, reason }) {
  const previousState = worker.state;
  const shouldEmit = [Worker.states.REQUESTED, Worker.states.RUNNING].includes(worker.state);
  if (shouldEmit) {
    await this.onWorkerRemoved({ worker, reason });
  }

  await worker.update(this.db, w => {
    const now = new Date();
    if ([Worker.states.REQUESTED, Worker.states.RUNNING].includes(w.state)) {
      w.lastModified = now;
      w.state = Worker.states.STOPPING;
    }
    w.providerData.reasonRemoved ??= reason;
  });

  // Avoid racing ARM deployments that have not finished resource extraction.
  const vmReadyForDelete =
    previousState === Worker.states.RUNNING ||
    worker.providerData.provisioningComplete === true;

  if (vmReadyForDelete && worker.providerData.vm?.name && !worker.providerData.vm.deleted) {
    this._enqueue('query', () => this.computeClient.virtualMachines.beginDelete(
      worker.providerData.resourceGroupName,
      worker.providerData.vm.name,
    )).catch(err => {
      const monitor = this.workerMonitor({
        worker,
        extra: {
          resourceGroupName: worker.providerData.resourceGroupName,
          vmName: worker.providerData.vm.name,
        },
      });
      monitor.debug({
        message: 'failed to start VM deletion from removeWorker; scanner will retry',
        error: err.message,
        code: err.code,
        statusCode: err.statusCode,
      });
    });
  }
}
```

Notes:

- Current Azure worker-manager queues deletes through `_enqueue('query', ...)`; there is no `delete` CloudAPI queue today.
- The provisioning guard matters. A REQUESTED ARM worker can move to STOPPING before deployment/resource extraction has finished. In that case, the scanner should keep handling cleanup once deployment state is known.
- The inline call should not wait for the long-running Azure delete to finish. It only needs to submit the delete request. If that fails, the existing scanner path retries later.

## Expected impact

This should move the billed-running tail from "next scanner visit" to "right after `removeWorker` marks the worker STOPPING."

For Mozilla's Windows Azure pools, the target is:

- `delay_p50_h`: about 30 min -> sub-minute / low-minutes
- `lifecycle_gap_share`: 45% -> roughly 15-25%, pending rollout measurement

## Related follow-ups

Related existing issues:

- #5291 reported the same family of Azure removal lag symptoms. This issue narrows the cause to the deferred first VM delete after `removeWorker`.
- #8161 covers STOPPING workers counting against capacity. That can amplify the impact, but it is separate from starting VM deletion promptly.
- #8517 fixed workers stuck after out-of-band resource deletion. It was fixed by merged PR #8518.
- #8526 fixed the Azure provider's pre-flight GET behavior so cascaded child-resource cleanup converges faster. It was fixed by merged PR #8530.

1. Split STOPPING workers into a fast scanner lane. A `state = STOPPING` scanner with a shorter wait time would bound the fallback path when the eager delete fails or is skipped.

2. Make ARM cascade cleanup more explicit in `deprovisionResources`. Mozilla's ARM template sets `deleteOption: 'Delete'` on NIC, OS disk, and public IP. Current v99.2.0 already handles cascaded 404s well, so this is scanner-throughput cleanup rather than the main fix.

3. Audit the Windows fallback shutdown path. `workers/generic-worker/host/host_windows.go` uses `shutdown.exe /s`, which puts an Azure VM into Stopped (Allocated), still billed for compute. That path only runs if `RemoveWorker` fails, but it is worth checking logs for `Falling back to system shutdown`.

## Data sources

- Splunk `index=azure_audit` for Azure VM CREATE/DELETE events
- GCP Cloud Logging via `tc-logview` for worker-manager events
- Yardstick / Prometheus (`gcp-v2-prod`) for capacity and queue metrics
- Taskcluster source at v99.2.0 (`bc2efe2827`)

A longer Mozilla-side investigation writeup exists and can be shared if useful.

petemoore:
## Summary

Closes #8574 — Azure provider's `removeWorker` now submits the VM `beginDelete` inline instead of waiting for the next worker-scanner pass to observe `state=STOPPING` before initiating the cloud-side delete. Brings Azure to parity with GCP (`google.js:197`) and AWS (`aws.js:399`).

Built on top of jmoss's draft from the issue (kept as-is in commit 1, attribution preserved). Three commits on top: a tightening of his guard predicate, tests, and a changelog entry. See per-commit messages for rationale.

## Pre-deploy validation

Independent measurement on `gecko-t/win11-64-25h2` (busiest Windows Azure pool, fx-ci) over 6 h on 2026-05-08, methodology and full numbers in the [validation comment](https://github.com/taskcluster/taskcluster/issues/8574#issuecomment-4407662481):

- Lifecycle gap (worker-removed → first scanner observation of STOPPING): **p50 14.9 min**
- Cycle (worker-removed → worker-stopped): **p50 40.8 min**
- Decomposes cleanly as `½ scanner pass (gap)` + `1 scanner pass (cleanup)`. Scanner pass duration: p50 25.6 min.
- Predicted with this fix: cycle p50 → ~26 min (one cleanup pass alone). **~37% reduction** for the API-call case (which is ~80% of removals on this pool).
- Same-window GCP comparison: cycle p50 **2.6 min**. Azure is **15.7×** slower at the median today.

## Hypotheses considered and ruled out for the same window

- **Queue saturation**: 0 `cloud-api-paused` events. p-queue never throttled.
- **Azure API latency / rate-limiting**: 54,904 calls, 0 failures, 0 retries, p50 1.78 s.

## Out of scope

The remaining ~63% of cycle time (the cleanup scanner pass) scales linearly with worker count. Two follow-ups:

- #8605 — scanner replay harness driven by captured production traces (lets us validate scanner-architecture changes in CI without spawning real Azure VMs)
- #8606 — verify the cycle-time reduction in production after this fix is deployed
- #8609 — handle Azure beginDelete 409 from concurrent inline + scanner deletes (rare race window opened by this fix; track post-deploy and address if non-trivial)


services/worker-manager/src/providers/azure/index.js
```diff
@@ -2068,7 +2084,35 @@ export class AzureProvider extends Provider {
       }
       // additionally store removal reason
       w.providerData.reasonRemoved ??= reason;
+      // Match deprovisionResource's convention: id is unset once a delete
+      // request is in flight. Keeps the `worker.providerData.vm.id` gate in
+      // deprovisionResources correct on the next scanner pass and avoids
+      // confusing other readers.
+      if (vmReadyForDelete) {
+        w.providerData.vm.id = false;
```
lotas, line 2092: Isn't it a bit risky, if the actual `beginDelete` fails, we would mark id as removed and force checkWorker to do an extra get request to find this id out

changelog/issue-8574.md
```diff
@@ -0,0 +1,5 @@
+audience: deployers
+level: minor
+reference: issue 8574
+---
+The Azure provider in worker-manager now submits VM `beginDelete` inline from `removeWorker`, removing one scanner cycle of latency before cloud-side deletion. Brings Azure to parity with GCP and AWS; the worker-scanner remains the fallback and verifier.
```
lotas, line 5: Something worth keeping in mind those api calls might get delayed and timed out now because azure frequently does so, which means the whole response might take several minutes. 
If worker-runner/generic-worker rely on this response to be quick to do some other cleanup before termination, we might need to rethink this approach.

lotas (approved):

services/worker-manager/src/providers/azure/index.js
petemoore: I think we're ok. `vm.id` is the full resource id which gets set to false once the delete is triggered. `vm.name` remains set to the resource name, `vm.deleted` is false, state is `STOPPING`. on the next scanner pass deprovisionResource uses `vm.name`, not `vm.id`. So having lost `vm.id` doesn't force an extra lookup. the vm would then presumably be in succeeded state (if delete failed), so `shouldDelete` is then true, and beginDelete gets tried again...

changelog/issue-8574.md
petemoore: the `beginDelete` call isn't awaited, so i think we are also ok here. also the workers don't currently rely on the response.

PR #8608, opened 2026-05-08
Playwright tests for UI

### LINKED ISSUE #6866: E2E UI tests
We introduced Cypress some time ago but it was never running in CI 

We should either fully adopt it and write more tests that would run in CI or get rid of it and use something different like playwrite or puppeteer?

petemoore:
Fixes #6866.

Note, the new task will also save screenshots as artifacts, see e.g. https://community-tc.services.mozilla.com/tasks/TWED7fgOQki34yBXIMWXcA/runs/0#artifacts

This also renames check/task `ui-browser-test` to `ui-lint-test-build` (since it was not actually testing the browser) and adds the task `ui-smoke` so when this lands, we should update the required github checks list for merging a PR to include the `ui-smoke`, `ui-lint-test-build`, and remove `ui-browser-test`.


ui/smoke-ci.sh
```diff
@@ -0,0 +1,29 @@
...
+DEV_LOG=/tmp/dev-server.log
+yarn start > "$DEV_LOG" 2>&1 &
+DEV_PID=$!
+
+# Wait for the initial webpack build to finish, not just for the port
+# to open — webpack-dev-server starts serving before all chunks are
+# emitted, which races the smoke harness on a cold CI worker.
+timeout 300 bash -c "until grep -q 'Compiled successfully' '$DEV_LOG'; do sleep 2; done"
+
+yarn smoke --base "http://localhost:${PORT}"
```
lotas: why do we use `TASKCLUSTER_ROOT_URL` above? Shouldn't it be also in the base?

ui/smoke.mjs
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env node
```
lotas: I wonder why you decided to go this way of running tests? 
[Playwright supports writing proper test()](https://playwright.dev/docs/writing-tests) functions that can test the behavior also - something that cypress did as well: click on items, submit form, expect text to be present. 

I'm a bit worried that this is also hard to support and extend in it's current form

Maybe we can re-use same `test/e2e` folder to write some simple test() cases for those pages and let playwright do the most of heavy lifting this script is doing? `npx playwright tests/e2e` (or just adding a proper [playwright.config](https://playwright.dev/docs/test-configuration) - I think it would be much better to extend in the future. Wdyt?

ui/package.json
```diff
@@ -16,10 +16,11 @@
     "test:watch": "cross-env TZ=Europe/Berlin jest --watch",
     "lint": "eslint --cache --format codeframe --ext js,jsx src",
     "lint:fix": "eslint --cache --format codeframe --fix --ext js,jsx src",
+    "smoke": "npx playwright test --config playwright.config.mjs",
```
lotas: I'm semi-sure if it's a dependency you can just leave `playwright test` - yarn should figure out there might be a binary (it should likely have same name)

ui/smoke-ci.sh
```diff
@@ -0,0 +1,29 @@
...
+DEV_LOG=/tmp/dev-server.log
+yarn start > "$DEV_LOG" 2>&1 &
+DEV_PID=$!
+
+# Wait for the initial webpack build to finish, not just for the port
+# to open — webpack-dev-server starts serving before all chunks are
+# emitted, which races the smoke harness on a cold CI worker.
+timeout 300 bash -c "until grep -q 'Compiled successfully' '$DEV_LOG'; do sleep 2; done"
+
+BASE_URL="http://localhost:${PORT}" yarn smoke
```
lotas, line 29: shouldn't it use TASKCLUSTER_ROOT_URL?

ui/test/smoke/routes.spec.mjs
```diff
@@ -0,0 +1,43 @@
...
+  { path: '/docs/manual', name: 'docs-manual' },
+  { path: '/docs/tutorial/hello-world', name: 'docs-tutorial' },
+  { path: '/docs/reference', name: 'docs-reference' },
+  { path: '/tasks', name: 'tasks' },
+  { path: '/provisioners', name: 'provisioners' },
+  { path: '/quickstart', name: 'quickstart' },
+];
+
+for (const route of ROUTES) {
+  test(`${route.name} (${route.path}) renders without errors`, async ({ page, issues }) => {
```
lotas, line 17: 👍 nice, i like this approach way more

ui/package.json
petemoore: Indeed, good spot! :-)

ui/smoke-ci.sh
petemoore: so the BASE_URL tells playwright which URL to hit to load the UI, i.e. the local dev server. i think if we set this to TASKCLUSTER_ROOT_URL it would test the production UI, not the local checked out one

ui/test/smoke/routes.spec.mjs
petemoore: well thanks for flagging to use the test harness, and the page object pattern was johan's idea :-)

lotas (approved):
LGTM! Nice addition with screenshots and browseable reports!

lotas:
image
this will need to be removed from required once we merge

PR #8615, opened 2026-05-08
fix(deps): remove `sentry-api` to resolve critical vulns

matt-boris:
Fixes https://github.com/taskcluster/taskcluster/security/dependabot/458
Fixes https://github.com/taskcluster/taskcluster/security/dependabot/461


lotas (approved):
LGTM, thanks! 
Looks like all usages of sentry api were included, but I don't know how to live-test it

petemoore (approved):
LGTM too

PR #8627, opened 2026-05-14
fix(github): ensure Github service can fetch custom check run artifac…

### LINKED ISSUE #8625: Github service failing to fetch task artifacts when the tasks have scopes that it does not
See e.g. https://github.com/mozilla-releng/fxci-config/commit/db49043fa6be214e5276e7ead387b9a5954ff962#commitcomment-185117604 - here the task has `assume:hook-id:lint/pre-commit-v1` but the github service doesn't, so the request by the github service to fetch the task artifact fails because the request is signed by the github service and includes the task's scopes as "authorized scopes".

However, the task's scopes are not relevant here, whether the github service should be able to read the artifacts should depend on what has been granted to the github service. Usually the anonymous scopes should be enough, so signing would not be necessary at all, but in the case of a locked down deployment and a private github repo, an operator may choose to grant the github service read access to e.g. task logs even if the anonymous scope doesn't allow it, so making the request as the github service is probably the correct solution here.

ahal:
…ts from hook tasks
Previously the requests to fetch this artifact were made with a
limitedQueueClient that had `assume:hook-id:<name>` in the
authorizedRoles. The problem is that `static/taskcluster/github` client
does not have permission to assume hook-id roles!

Nor do we want to grant it permission. Because unlike
`queue.createTask`, the caller of `hooks.triggerHook` does not need a
superset of the scopes of the task being created (controlled privilege
escalation). So granting `assume:hook-id:*` would be giving the Github
service lots of privileges it doesn't need.

Instead, let's explicitly grant get-artifact scopes to the Github
service for the three well-defined artifact constants we know it needs
to fetch. This way we don't need to rely on the anonymous role or task
roles having them by happenstance.

Github Bug/Issue: Fixes #8625


services/github/src/handlers/requestArtifact.js
```diff
@@ -1,14 +1,24 @@
...
+  // out any assume:hook-id:* scopes before restricting the request. If only
+  // hook scopes remain after filtering, fall back to an unauthenticated
+  // request.
+  const filteredScopes = scopes?.filter(s => !s.startsWith('assume:hook-id:'));
+  if (filteredScopes?.length > 0) {
+    const limitedQueueClient = queueClient.use({ authorizedScopes: filteredScopes });
+    return limitedQueueClient.buildSignedUrl(limitedQueueClient.getArtifact, taskId, runId, artifactName);
+  }
+  return queueClient.buildUrl(queueClient.getArtifact, taskId, runId, artifactName);
+}
```
matt-boris: What about first trying unauthenticated and if that fails, try as the github service with a get artifact scope (`queue:get-artifact:<artifactName>`)?

Then we could grant `queue:get-artifact:<artifactName>` or some broader wildcard to `static/taskcluster/github`.

I feel like doing it this way would better solve the root issue, as the task's scopes should not be relevant to whether the github service can read the artifact. That way, deployers could just specify exactly the scopes they want their github service to have and it only uses them if the anonymous get of the artifact first fails.

Also, what if there are other scopes in the filteredScopes list that the github service cannot satisfy?

If we go that route, we should probably update the docs https://docs.taskcluster.net/docs/reference/integrations/github/checks that mention you should add the artifact read scopes to the task's scopes.

services/github/src/constants.js
```diff
@@ -85,6 +85,8 @@ export const PUBLISHERS = {
 export const CHECKLOGS_TEXT = 'View logs in Taskcluster';
 export const CHECKRUN_TEXT = 'View task in Taskcluster';
 export const CHECK_TASK_GROUP_TEXT = 'View task group in Taskcluster';
+// Keep the `get-artifact` grants in services/auth/src/static-scopes.json
+// synchronized with these constants.
```
matt-boris: 👍🏻 thanks!

ahal:
I'm fairly confident this is the correct fix now. The commit messages explains things but I'll paste it here for convenience:

---

Previously the requests to fetch this artifact were made with a
`limitedQueueClient` that had `assume:hook-id:<name>` in the
authorizedRoles. The problem is that `static/taskcluster/github` client
does not have permission assume `hook-id` roles!

Nor do we want to grant it permission. Because unlike
`queue.createTask`, the caller of `hooks.triggerHook` does not need a
superset of the scopes of the task being created (controlled privilege
escalation). So granting `assume:hook-id:*` would be giving the Github
service lots of privileges it doesn't need.

Instead, let's explicitly grant `get-artifact` scopes to the Github
service for the three well-defined artifact constants we know it needs
to fetch. This way we don't need to rely on the anonymous role or task
roles having them by happenstance.

matt-boris (commented):
This LGTM, thank you! Let's get https://docs.taskcluster.net/docs/reference/integrations/github/checks updated now and I think this will be all good to go then 👍🏻

matt-boris (approved):
Nice, this looks great to me

lotas (approved):

PR #8633, opened 2026-05-18
Remove use of `.map().flat()` in favor of `.flatMap`/`map`

Eijebong:
This fixes a biome warning (`complexity/useFlatMap`).

There were only two callsites flagged by biome. The first one actually requires the flat map; `group.hooks` is an array so the change is easy enough, switch to `flatMap`, enjoy the one saved allocation.

The second callsite actually did not need the flat map so I just dropped the `flat` part of it. Both the API and graphQL schemas type `hookGroupId` as a string. This has been working because, despite strings being iterable, they're not flattened:

```
>  ["ab", "cd"].flat()
[ 'ab', 'cd' ]
> [["ab"], ["cd"]].flat()
[ 'ab', 'cd' ]
```

(for reference, python's closest equivalent,
`itertools.chain.from_iterable`, would do `["ab", "cd"]` -> `['a', 'b', 'c', 'd']`)

After digging a bit, that code was introduced in 1dca82decf but was already useless back then and was probably copy/pasted from neighbouring code.

### DIFF
--- a/changelog/d9uYuwDuSbWIwF9dtbBLNQ.md
+++ b/changelog/d9uYuwDuSbWIwF9dtbBLNQ.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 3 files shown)


lotas (approved):
Cool! Thanks for the fixes

PR #8634, opened 2026-05-18
Replace usage of `new Date().getTime()` with `Date.now()`

Eijebong:
This fixes a biome lint (`complexity/useDateNow`).

This avoids allocating a Date object just to extract ms from it. The behavior is otherwise unchanged.

### DIFF
--- a/changelog/b4q6llG6QGWq-hcSByinwQ.md
+++ b/changelog/b4q6llG6QGWq-hcSByinwQ.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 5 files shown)


lotas (approved):
Thanks! Are you trying to fix all biome warnings one by one? :) how many are there?

Eijebong:
I'm fixing what can be fixed in tiny patches independently to avoid slamming a +5000/-5000 PR. Helps with reviews, helps with me testing these patches.

PR #8651, opened 2026-05-19
Fix biome's noUnusedVariables lint

Eijebong:
This is better reviewed commit by commit. It fixes 2 bugs that were surfaced by the lint and ignores all other unused variables as those were ignored on purpose.


matt-boris (approved):
Nice, ty!

PR #8652, opened 2026-05-19
Remove useless outer `<React.Fragment/>`

Eijebong:
This was found by biome's noUselessFragments lint. It change no behavior, those fragments were in fact not useful
(`<Fragment><elem>...</elem></Fragment>` can just be `<elem>...</elem>`)

This commit is better viewed with whitespaces hidden since a lot of it is just reindenting the content of the fragments down by one.

### DIFF
--- a/changelog/aeHd2o95ROypj26wmmv95g.md
+++ b/changelog/aeHd2o95ROypj26wmmv95g.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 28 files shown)


matt-boris (dismissed):
Much easier to review with whitespace changes hidden. This LGTM, thanks.

matt-boris (approved):

PR #8656, opened 2026-05-19
Derive go example version from the build instead of a hardcoded '93' and remove unused function arguments in doc generation

Eijebong:
This removes 12 unused arguments from the doc example generators which were added in ced6eaa3830f033d2e082b92c585a7dff1c52bd8. Both `version` and `apiVersion` were only ever used by a single generator but were threaded to every single one of them for some reason. On top of that, version was hardcoded to `93` which means go examples were showing imports for `v93` which is already 7 major versions behind.

This was found while fixing biomes `noUnusedFunctionParameters` lint.

### DIFF
--- a/changelog/IYvZA22yTE64owyDAnRaOQ.md
+++ b/changelog/IYvZA22yTE64owyDAnRaOQ.md
@@ -1,0 +1,4 @@
+audience: users
+level: patch
+---
+Use the current version for taskcluster in API code examples instead of a hardcoded '93'.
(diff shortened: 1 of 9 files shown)


matt-boris (dismissed):
Thanks for the cleanup 🙏🏻

matt-boris (approved):

petemoore (approved):
Tell us honestly Bastien, did you write this, or did you get claude to write it for you?

PR #8666, opened 2026-05-20
feat(github): display artifact size in check results

### LINKED ISSUE #8662: Display artifact size in github check results
Since we have it now, we can also include it on the github pages

https://github.com/taskcluster/taskcluster/blob/dc7b40dcec628323bcaac551deebf503988f2bee/services/github/src/handlers/status.js#L226-L237

mahavirkunwar:
Github Bug/Issue: Fixes #8662


services/github/src/handlers/status.js
```diff
@@ -221,18 +221,28 @@ export async function statusHandler(message) {
...
       }
 
+      // A quick helper function to format bytes nicely (e.g., 1024 -> "1 KB")
+      const formatBytes = (bytes) => {
+        if (!bytes || bytes === 0) return '0 Bytes';
+        const k = 1024;
+        const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB'];
+        const i = Math.floor(Math.log(bytes) / Math.log(k));
+        return parseFloat((bytes / Math.pow(k, i)).toFixed(2)) + ' ' + sizes[i];
+      };
```
petemoore: We have something similar in ui/src/utils/formatBytes.js - I wonder if that can be reused? I think there we use "B" instead of "Bytes", not that it really matters. Also reusing isn't critical, it is only a few lines that are unlikely to change.

In the other version we have:
```
const UNITS = ['B', 'KB', 'MB', 'GB', 'TB', 'PB'];
```

so maybe we should just align on B vs Bytes. Or maybe we should just copy that implementation so we are certain they are consistent. Not sure if we already have any means to share code across ui and services.

@lotas / @Eijebong what do you think?

Eijebong: https://github.com/taskcluster/taskcluster/pull/8579#discussion_r3184866292

Could also use what I suggested there (which honestly should probably have been used for the service also since it ended up doing base 1k instead of base 2, but it's irrelevant here)

mahavirkunwar: I will check this and update PR. I will use the existing functions.

mahavirkunwar:
Hi everyone, thanks for the reviews and patience! I've updated the PR to use the native Intl.NumberFormat approach that @Eijebong suggested over in #8579, which handles the formatting cleanly. I also resolved the nodejs linting errors and generated the changelog snippet.

Let me know if anything else is needed!

services/github/src/handlers/status.js
```diff
@@ -221,18 +221,30 @@ export async function statusHandler(message) {
         output.addText(`### Artifacts`);
       }
 
+      // A helper function using native Intl to format bytes (base 1000)
+      const formatBytes = (bytes) => {
+        if (!bytes || bytes === 0) {return '0 B';}
```
lotas: @mahavirkunwar can you please update this logic to match what we do on the UI:

https://github.com/taskcluster/taskcluster/blob/8a350dbe154eacd52d67814da1a1d299be15aabe/ui/src/utils/formatBytes.js#L4-L10

mahavirkunwar: @lotas I will update the PR.

lotas: @mahavirkunwar thanks, did you by a chance forgot to push it? :) I think the version is still the same

lotas:
Thanks @mahavirkunwar 
I tried to run tests locally after adding sizes to the mock data https://github.com/taskcluster/taskcluster/blob/f7038b2bab5f8ed2fa7ee793402521ec72c3275f/services/github/test/handler_test.js#L184 

```js
        for (let i = 0; i < options.limit; i++) {
          artifacts.push({
            name: `artifact-${i}`,
            size: 2 ** i,
          });
        }
```

tests are obviously failing, but what I was checking is the resulting output:

```diff
-\- [artifact-0 (1 byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-0)
-\- [artifact-1 (2 byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-1)
-\- [artifact-2 (4 byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-2)
-\- [artifact-3 (8 byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-3)
-\- [artifact-4 (16 byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-4)
-\- [artifact-5 (32 byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-5)
-\- [artifact-6 (64 byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-6)
-\- [artifact-7 (128 byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-7)
-\- [artifact-8 (256 byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-8)
-\- [artifact-9 (512 byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-9)
-\- [artifact-10 (1K byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-10)
-\- [artifact-11 (2K byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-11)
-\- [artifact-12 (4.1K byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-12)
-\- [artifact-13 (8.2K byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-13)
-\- [artifact-14 (16K byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-14)
-\- [artifact-15 (33K byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-15)
-\- [artifact-16 (66K byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-16)
-\- [artifact-17 (131K byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-17)
-\- [artifact-18 (262K byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-18)
-\- [artifact-19 (524K byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-19)
-\- [artifact-20 (1M byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-20)
-\- [artifact-21 (2.1M byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-21)
-\- [artifact-22 (4.2M byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-22)
-\- [artifact-23 (8.4M byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-23)
-\- [artifact-24 (17M byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-24)
-\- [artifact-25 (34M byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-25)
-\- [artifact-26 (67M byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-26)
-\- [artifact-27 (134M byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-27)
-\- [artifact-28 (268M byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-28)
-\- [artifact-29 (537M byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-29)
-\- [artifact-30 (1.1B byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-30)
-\- [artifact-31 (2.1B byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-31)
-\- [artifact-32 (4.3B byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-32)
-\- [artifact-33 (8.6B byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-33)
-\- [artifact-34 (17B byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-34)
-\- [artifact-35 (34B byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-35)
-\- [artifact-36 (69B byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-36)
-\- [artifact-37 (137B byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-37)
-\- [artifact-38 (275B byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-38)
-\- [artifact-39 (550B byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-39)
-\- [artifact-40 (1.1T byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-40)`
-\- [artifact-41 (2.2T byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-41)
-\- [artifact-42 (4.4T byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-42)
-\- [artifact-43 (8.8T byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-43)
-\- [artifact-44 (18T byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-44)
-\- [artifact-45 (35T byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-45)
-\- [artifact-46 (70T byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-46)
-\- [artifact-47 (141T byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-47)
-\- [artifact-48 (281T byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-48)
-\- [artifact-49 (563T byte)](https://tc-tests.example.com/tasks/banana/runs/0/artifact-49)
+\- [artifact-0](https://tc-tests.example.com/tasks/banana/runs/0/artifact-0)
+\- [artifact-1](https://tc-tests.example.com/tasks/banana/runs/0/artifact-1)
+\- [artifact-2](https://tc-tests.example.com/tasks/banana/runs/0/artifact-2)
+\- [artifact-3](https://tc-tests.example.com/tasks/banana/runs/0/artifact-3)
+\- [artifact-4](https://tc-tests.example.com/tasks/banana/runs/0/artifact-4)
+\- [artifact-5](https://tc-tests.example.com/tasks/banana/runs/0/artifact-5)
+\- [artifact-6](https://tc-tests.example.com/tasks/banana/runs/0/artifact-6)
+\- [artifact-7](https://tc-tests.example.com/tasks/banana/runs/0/artifact-7)
+\- [artifact-8](https://tc-tests.example.com/tasks/banana/runs/0/artifact-8)
+\- [artifact-9](https://tc-tests.example.com/tasks/banana/runs/0/artifact-9)
+\- [artifact-10](https://tc-tests.example.com/tasks/banana/runs/0/artifact-10)
+\- [artifact-11](https://tc-tests.example.com/tasks/banana/runs/0/artifact-11)
```

That's an unusual way to write sizes  `4 byte`, `16K byte`, `2.1M byte` ? Shouldn't it be using `MB`, `KB` etc

Can you also add a proper tests to cover this once the formatting is fixed? 
Thanks!

lotas (changes requested):
Thanks!

lotas:
I had to rebase because there were some conflicts in handler_test 
Also extracted helper function into a separate util

lotas (approved):

PR #8668, opened 2026-05-21
Use a `Set` instead of a `Map` with `undefined` values in `TaskGroup`

Eijebong:
`this.tasks` was declared as a `Map` but only ever used for membership checks (insertion was `set(key)`, no value which defaulted to `undefined`). This is a sign that it's a `Set` wearing a `Map` costume.

Convert it to an actual `Set` and fix what prompted me to look at this code while I'm touching it, `forEach` was returning whatever `this.tasks.set(...)` was returning which triggered biome's `useIterableCallbackReturn`.

This map was never needed and should've been a set in 3c330c362ad7348a6eee38ea868f412cd2b28116 already.

### DIFF
--- a/changelog/Wr2mIsbrRqeqOOnskEjmdQ.md
+++ b/changelog/Wr2mIsbrRqeqOOnskEjmdQ.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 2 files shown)


lotas (approved):
sweet. thanks

PR #8671, opened 2026-05-21
Fix a few issues on the changelog UI page

Eijebong:
See individual commits. Basically fixes a bunch of warnings thrown by react/MUI because of bad usage of APIs, missing IDs. Fixed 2 biome a11y lints and fixed the filters to properly reflect what the filters are in the URL and to update when clicking on a version number in the changelog.


lotas (approved):
Awesome, thanks

PR #8676, opened 2026-05-22
Make TC proxy tests use a dynamically allocated port and don't panic on failure to bind

Eijebong:
Fixes a bug where a collision on bind would panic and segfault because of a code change that didn't adapt the test for it and greatly reduces the chance of port collision by dynamically allocating the port during the test rather than hardcoding it to :34570.

See individual commit messages for more details.


lotas (approved):
nice!

PR #8678, opened 2026-05-22
Remove dead Image module in documentation

Eijebong:
This was never actually used. It got introduced as a copy of `imageFactory` in f2aa9af2b11865592e6806a3deaff83ca5736076 where it was already unused. Then 6b8a86f4eae2db1c696f8af13aabd104b19c2278 removed the image factory entirely because it wasn't working, didn't use `Image` and left the file behind.


Eijebong:
Oh, it's imported directly from mdx files, which I didn't grep. Whoops

PR #8679, opened 2026-05-22
Remove our `yarn minify` handrolled lockfile minifier

Eijebong:
This has been broken for a few years already (since 163794b682daaf8ff522b6511cc48ac6f3282b4b). It was using `yarn-minify` which only shrinks yarn v1 lockfiles.

Yarn's `dedupe` command is now built in and covers the same job, so drop the `minify` tool entirely instead of trying to fix it. Given it's been broken for 3 years, I don't think anyone's going to miss it.

Also ran `yarn dedupe` on all our lockfiles.


lotas (approved):
I'm not sure I ever used it

Issue #8372, opened 2026-03-17
Replace eslint with biome?


lotas:
Biome is written in rust and comes with performance improvements:
https://biomejs.dev/guides/migrate-eslint-prettier/

Currently we are using eslint on multiple levels, and eslint from v9+ is no longer supporting existing yaml config format

Also client-web was using broken version already:
https://github.com/taskcluster/taskcluster/pull/8334

PR #8680, opened 2026-05-22
Switch ui/ linting from eslint to biome

Eijebong:
This should be reviewed commit by commit. I tried keeping lint fixes split up and justified.

Note that a lot of issues discovered during this migration have already been merged, the fixes in here are tiny enough that I didn't feel like they warranted their own PR.

The first commit does the scaffolding, switches eslint to biome with a configuration that matches somewhat. The two commits after that are autogenerated with biome. Each commit after that fixes/silences specific lints.

Part of #8372, doesn't close it yet as we're still using eslint in services/clients.


lotas (approved):
Tremendous job! Thanks!

PR #8504, opened 2026-04-16
feat(w-m): Track Azure rate limiting errors

lotas:
Exposes two new metrics and logs with detailed information about ocurred rate limiting errors

Fixes: #8497

PR #8684, opened 2026-05-27
feat(w-m): add Azure ARM deployment error metrics

jwmossmoz:
Exposes a new worker-manager Azure Prometheus counter for ARM deployment creation failures and failed deployment operations.

The new `worker_manager_azure_arm_deployment_errors_total` metric increments once for each failed ARM deployment creation or failed deployment operation. It includes labels for `providerId`, `workerPoolId`, `workerGroup`, Azure error code/state details, target resource type, VM size, and priority so deployers can chart which pools, regions, VM sizes, and Azure error codes are producing ARM deployment failures in Yardstick/Grafana.

Follows the Azure observability shape from #8504.

## Testing

- [x] `git diff --check`
- [x] `NODE24_BIN_DIR=$(dirname "$(npx -y -p node@24.16.0 which node)") && PATH="$NODE24_BIN_DIR:$PATH" TEST_DB_URL=postgresql://postgres@localhost:5432/taskcluster-test yarn generate`
- [x] `NODE24_BIN_DIR=$(dirname "$(npx -y -p node@24.16.0 which node)") && PATH="$NODE24_BIN_DIR:$PATH" yarn workspace @taskcluster/worker-manager exec eslint src/providers/azure/index.js src/monitor.js test/provider_azure_test.js`
- [x] `NODE24_BIN_DIR=$(dirname "$(npx -y -p node@24.16.0 which node)") && PATH="$NODE24_BIN_DIR:$PATH" TEST_DB_URL=postgresql://postgres@localhost:5432/taskcluster-test yarn workspace @taskcluster/worker-manager exec mocha test/provider_azure_test.js --grep "records metric when ARM deployment creation fails|reports worker-pool error when ARM deployment fails"` - 2 passing
- [x] `NODE24_BIN_DIR=$(dirname "$(npx -y -p node@24.16.0 which node)") && PATH="$NODE24_BIN_DIR:$PATH" TEST_DB_URL=postgresql://postgres@localhost:5432/taskcluster-test yarn workspace @taskcluster/worker-manager exec mocha test/provider_azure_test.js` - 148 passing

### DIFF
--- a/services/worker-manager/src/monitor.js
+++ b/services/worker-manager/src/monitor.js
@@ -545,6 +545,27 @@ MonitorManager.registerMetric('azureThrottleCount', {
   registers: ['provision', 'scan'],
 });
 
+MonitorManager.registerMetric('azureArmDeploymentError', {
+  name: 'worker_manager_azure_arm_deployment_errors_total',
+  type: 'counter',
+  title: 'Azure ARM deployment errors',
+  description: 'Count of Azure ARM deployment creation failures and failed deployment operations by worker pool, location, and Azure error details',
+  labels: {
+    providerId: 'ID of the provider',
+    workerPoolId: 'The worker pool ID',
+    workerGroup: 'Worker group (region/zone/location)',
+    errorKind: 'Worker-manager error kind',
+    errorCode: 'Azure error code',
+    statusCode: 'Operation status code from the resource provider',
+    provisioningState: 'Azure provisioning state',
+    provisioningOperation: 'Azure provisioning operation',
+    targetResourceType: 'Azure target resource type',
+    vmSize: 'Azure VM size',
+    priority: 'Azure VM priority',
+  },
+  registers: ['provision', 'scan'],
+});
+
 MonitorManager.registerMetric('azureRateLimitRemaining', {
   name: 'worker_manager_azure_ratelimit_remaining',
   type: 'gauge',
--- a/services/worker-manager/src/providers/azure/index.js
+++ b/services/worker-manager/src/providers/azure/index.js
@@ -49,6 +49,8 @@ const InstanceStates = {
 const failProvisioningStates = new Set(['Failed', 'Deleting', 'Canceled', 'Deallocating']);
 
 const DEPLOYMENT_METHOD_ARM = 'arm-template';
+const UNKNOWN_METRIC_LABEL = 'unknown';
+const MAX_METRIC_LABEL_LENGTH = 200;
 const maxInstanceView404Streak = 2;
 // Per Azure Certificate Authority details, these are the HTTP AIA hosts
 // clients may need to reach for Azure certificate chain building.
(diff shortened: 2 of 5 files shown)


services/worker-manager/src/providers/azure/index.js
```diff
@@ -92,6 +94,81 @@ export function isAllowedAiaLocation(location) {
...
+    }
+
+    if (Array.isArray(current.details) && current.details.length > 0) {
+      queue.push(...current.details);
+    } else if (current.code || current.message || current.target) {
+      errors.push(current);
+    }
+  }
+
+  return errors.length > 0 ? errors : [{}];
```
lotas: this is probably incorrect to return empty object here `[{}]`
Or is it intended to force `azureArmDeploymentError` metric without the error?

jwmossmoz: > this is probably incorrect to return empty object here `[{}]` Or is it intended to force `azureArmDeploymentError` metric without the error?

Ah, the intent was to fallback to a top-level error or unknown. I updated the code!

services/worker-manager/src/monitor.js
```diff
@@ -545,6 +545,29 @@ MonitorManager.registerMetric('azureThrottleCount', {
   registers: ['provision', 'scan'],
 });
 
+MonitorManager.registerMetric('azureArmDeploymentError', {
+  name: 'worker_manager_azure_arm_deployment_errors_total',
+  type: 'counter',
+  title: 'Azure ARM deployment errors',
+  description: 'Count of Azure ARM deployment failures by worker pool, location, and Azure error details',
+  labels: {
```
lotas, line 553: I think we need to possibly reduce some of those.

[Prometheus docs](https://prometheus.io/docs/practices/naming/) warn that:
```
CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.
```

And some of those are possibly unbounded and belong to the logs instead.

services/worker-manager/src/providers/azure/index.js
```diff
@@ -92,6 +94,81 @@ export function isAllowedAiaLocation(location) {
...
+    error?.response?.parsedBody?.error,
+    error?.parsedBody?.error,
+    error?.body?.error,
+    error?.error,
+    error,
+  ];
+  const seen = new Set();
+  const errors = [];
+
+  while (queue.length > 0) {
```
lotas, line 136: this method probably deserves some unit-test to see what happens. Besides, if there are several errors nested under single failure this would produce several errors, that would increment counter more than once for a single failed arm deployment. If that's okay, we can ignore it

jwmossmoz: Good catch, fixing this!

lotas:
@jwmossmoz I think I missed that this PR was updated since I last reviewed it. Can you please run `yarn generate` to make tests happy?

lotas (approved):

PR #8685, opened 2026-05-27
Bug 2042324 - Fix authorization code lifetime in the oauth2 provider

Eijebong:
This fixes 2 issues with the lifetime of the authorization code given out by the taskcluster oauth2 provider.

The authorization codes expiry was checked against the expiry time of the resulting credentials instead of the code's expiry which means that in most cases the code was valid for a day instead of 10 minutes. Easy fix, check against the right field.

Authorization codes were not deleted after being used to issue access tokens (relying on the expiration cron to remove them). The fix there is a bit more involved because there was no method to delete the code. I went with a `consume` method which enforces the single use by removing the row from the database when we get it.

See each commit for more details.


lotas (approved):
Thanks!

PR #8690, opened 2026-05-28
Fix `on-defined` routes not firing notifications

### LINKED ISSUE #8689: Notification service failed to send `on-defined` pulse events
I'm working with Kyle from Pernosco to switch them over to using notify service routes rather than the full queue exchange.

[This task](https://firefox-ci-tc.services.mozilla.com/tasks/b1MfEBDaSC-fGZYfJNQhSg) defines:
```
notify.pulse.pernosco-v1.on-defined
```

But I was monitoring the notification exchange and no event came in. Events for `on-running` came through just fine. Per the [docs](https://docs.taskcluster.net/docs/reference/core/notify/usage#filters), on-defined is supposed to work.

Eijebong:
Routes like `notify.pulse.foo.on-defined` are supposed to send a message on the `foo` routing key of the notify service pulse exchange when the task gets defined but it was never wired up correctly.

aa326d8218b0477d061d838fe608acc7a8dad973 added the possibility to add `on-defined` but never tested it and missed that a task that got created would have `unscheduled` as its state and not `defined`.

I fixed the problem by matching `defined` as decider to the `unscheduled` state of task. If a task switched to the `unscheduled` state then it just got created.

Fixes #8689

### DIFF
--- a/services/notify/test/handler_test.js
+++ b/services/notify/test/handler_test.js
@@ -67,6 +67,12 @@ helper.secrets.mockSuite(testing.suiteName(), ['aws'], function(mock, skipping) 
     ],
   };
 
+  const definedStatus = {
+    ..._.cloneDeep(baseStatus),
+    state: 'unscheduled',
+    runs: [],
+  };
+
   let monitor;
   suiteSetup('create handler', async function() {
     if (skipping()) {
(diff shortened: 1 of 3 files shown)


services/notify/test/handler_test.js
```diff
@@ -67,6 +67,12 @@ helper.secrets.mockSuite(testing.suiteName(), ['aws'], function(mock, skipping)
     ],
   };
 
+  const definedStatus = {
+    ..._.cloneDeep(baseStatus),
```
lotas, line 71: my new favorite textmoji dropped `..._.`

lotas (approved):
Thanks

PR #8691, opened 2026-05-29
fix(client): use semver range for engines.node in published clients

### LINKED ISSUE #8660: taskcluster-client (npm): engines.node uses exact version instead of semver range since 55.3.3
## Summary

Since `taskcluster-client@55.3.3`, the `engines.node` field in the published npm package uses an exact Node.js version instead of a semver range. This causes `yarn install --frozen-lockfile` to fail with an engine incompatibility error even when the actual code is perfectly compatible with the installed Node.js version.

## Example error

```
error taskcluster-client@55.3.4: The engine "node" is incompatible with this module. Expected version "18.18.0". Got "18.20.8"
error Found incompatible module.
```

## Affected versions

| Version | `engines.node` |
|---------|---------------|
| 55.3.2  | `>=v8.17.0` ✅ |
| 55.3.3  | `18.18.0` ❌ |
| 55.3.4  | `18.18.0` ❌ |
| 56.0.0  | `18.18.0` ❌ |
| 57.0.0  | `18.18.1` ❌ |
| 60.0.0  | `18.19.0` ❌ |
| 80.0.0  | *(empty)* |
| 87.1.3  | `22.17.1` ❌ |

The exact version in the field appears to match the Node.js version used at publish time, suggesting the build/publish tooling is capturing the current Node.js version verbatim rather than writing a `>=` range.

## Impact

Any project using `yarn --frozen-lockfile` (or `npm install --engine-strict`) against a `node:18-alpine` Docker image (which ships Node.js 18.20.8) will fail when `taskcluster-client` resolves to `>=55.3.3`. This is affecting [taskcluster/taskgraph#954](https://github.com/taskcluster/taskgraph/pull/954).

## Expected behaviour

The `engines.node` field should use a semver range like `>=18.18.0` or `^18` so that compatible Node.js versions aren't rejected.

nitishagar:
## What

The `taskcluster-client` and `taskcluster-client-test` npm packages declared `engines.node` as an exact version (e.g. `24.15.0`) rather than a compatible range. This meant that any consumer running a compatible but different patch version of Node.js would see `engine-strict` install failures.

The generator in `node-version.js` now prefixes the version with `>=` when writing `engines.node` into the two published client packages, so any Node.js version at or above the declared minimum is accepted.

The `ui` and `docker-worker` packages retain the exact version because the meta-consistency check asserts `ui.engines.node === root engines.node`.

## Why

Fixes #8660. Consumers such as taskgraph encountered broken `yarn install --frozen-lockfile` when their Node.js patch version differed from the one baked into the client package.

## How verified

- `clients/client/package.json` `engines.node` changed from `24.15.0` → `>=24.15.0`
- `clients/client-test/package.json` `engines.node` changed from `24.15.0` → `>=24.15.0`
- `ui/package.json` and `workers/docker-worker/package.json` remain unchanged at `24.15.0`
- `yarn lint` passes


Eijebong (changes requested):
I wouldn't widen the range that much. We can't even run tests with node 26 right now. If there's a way to say "24 or lower" it's probably better? Although there's an argument for having a lower bound but that'd require testing that in CI too which is a bit overkill for the 2 external users of this.

```
> nvm use 26
*Now using node v26.2.0 (npm v11.13.0)
> yarn test
[@taskcluster/auth]: Process started
[@taskcluster/auth]: file:///home/eijebong/code/taskcluster/node_modules/mocha/node_modules/yargs/yargs:3
[@taskcluster/auth]: const {applyExtends, cjsPlatformShim, Parser, Yargs, processArgv} = require('./build/index.cjs')
[@taskcluster/auth]:                                                                     ^
[@taskcluster/auth]:
[@taskcluster/auth]: ReferenceError: require is not defined in ES module scope, you can use import instead
[@taskcluster/auth]:     at file:///home/eijebong/code/taskcluster/node_modules/mocha/node_modules/yargs/yargs:3:69
[@taskcluster/auth]:     at ModuleJobSync.runSync (node:internal/modules/esm/module_job:543:37)
[@taskcluster/auth]:     at ModuleLoader.importSyncForRequire (node:internal/modules/esm/loader:365:47)
[@taskcluster/auth]:     at loadESMFromCJS (node:internal/modules/cjs/loader:1692:24)
[@taskcluster/auth]:     at Module._compile (node:internal/modules/cjs/loader:1855:5)
[@taskcluster/auth]:     at Object..js (node:internal/modules/cjs/loader:2013:10)
[@taskcluster/auth]:     at Module.load (node:internal/modules/cjs/loader:1596:32)
[@taskcluster/auth]:     at Module._load (node:internal/modules/cjs/loader:1398:12)
[@taskcluster/auth]:     at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
[@taskcluster/auth]:     at Module.require (node:internal/modules/cjs/loader:1619:12)
[@taskcluster/auth]:
[@taskcluster/auth]: Node.js v26.2.0
[@taskcluster/auth]: Process exited (exit code 1), completed in 0s 139ms
```

PR #8692, opened 2026-05-29
fix(worker-manager): report consistent runningDuration across terminal events

### LINKED ISSUE #8346: `runningDuration` is being reported twice for worker that is being removed and stopping
We use same helper to log worker lifecycle events: https://github.com/taskcluster/taskcluster/blob/0bd9f844f89d79b27814e3a9c23b6af5869311d4/services/worker-manager/src/providers/provider.js#L281-L283 

They both produce different values https://github.com/taskcluster/taskcluster/blob/0bd9f844f89d79b27814e3a9c23b6af5869311d4/services/worker-manager/src/providers/provider.js#L342-L346

example showing RUNNING_DURATION and two different values
```
$  tc-logview query --where "workerId=8799521697670995739" --since 2h                                                                                                                                                                                          
Auto-detected environment: community-tc
Matched types: registration-error-warning, task-claimed, task-reclaimed, worker-removed, worker-requested, worker-running, worker-stopped
Filter: resource.labels.cluster_name="taskcluster-communitytc-v1" AND (jsonPayload.Type="registration-error-warning" OR jsonPayload.Type="task-claimed" OR jsonPayload.Type="task-reclaimed" OR jsonPayload.Type="worker-removed" OR jsonPayload.Type="worker-requested" OR jsonPayload.Type="worker-running" OR jsonPayload.Type="worker-stopped") AND jsonPayload.Fields.workerId=8799521697670995739 AND timestamp>="2026-03-04T10:22:00Z" AND timestamp<="2026-03-04T12:22:00Z"
Time: 2026-03-04T10:22:00Z to 2026-03-04T12:22:00Z
Querying GCP Cloud Logging...

TIMESTAMP             SERVICE         TYPE              PROVIDER_ID                  REASON                               REGISTRATION_DURATION  RUN_ID  RUNNING_DURATION  TASK_ID                 TASK_QUEUE_ID               TERMINATE_AFTER  WORKER_AGE  WORKER_GROUP   WORKER_ID            WORKER_POOL_ID
2026-03-04T12:13:40Z  worker-manager  worker-requested  community-tc-workers-google                                                                                                                                            1772972019805                us-central1-b  8799521697670995739  proj-mozci/compute-smaller
2026-03-04T12:14:29Z  worker-manager  worker-running    community-tc-workers-google                                       49.045                                                                                                                            us-central1-b  8799521697670995739  proj-mozci/compute-smaller
2026-03-04T12:15:59Z  queue           task-claimed                                                                                               0                         PTSz2ZNhRA2R6gf-hB08UA  proj-mozci/compute-smaller                               us-central1-b  8799521697670995739
2026-03-04T12:18:39Z  worker-manager  worker-removed    community-tc-workers-google  workerManager.removeWorker API call                                 250.069                                                                                299.116     us-central1-b  8799521697670995739  proj-mozci/compute-smaller
2026-03-04T12:19:25Z  worker-manager  worker-stopped    community-tc-workers-google                                                                      295.3                                                                                  344.347     us-central1-b  8799521697670995739  proj-mozci/compute-smaller
5 entries
```

We need to make sure that we record metric once to avoid pushing confusing values to metrics, but also be consistent with reporting

nitishagar:
## What

When a worker transitions through both `worker-removed` and `worker-stopped`, the `runningDuration` structured-log field was reported with two different values because each handler captured its own `Date.now()` timestamp, seconds or minutes apart.

This fix introduces a `_ensureStoppedAt` helper on `Provider` that persists `stoppedAt` on the first terminal event if it is not already set. Both terminal handlers now derive `runningDuration` from this persisted value, so the field is identical across both log lines. The static provider's direct log path receives the same fix.

The Prometheus `workerLifetime` metric is unaffected — it was already guarded against double-recording via `_recordWorkerStopped`'s idempotent flag.

## Why

Fixes #8346. The inconsistent values made it impossible to correlate log lines for the same worker and confused operators diagnosing worker lifetime anomalies.

## How verified

- `services/worker-manager` Mocha suite passes including:
  - existing "does not double-record lifetime" test (still green)
  - existing "includes workerAge and runningDuration on workerStopped/workerRemoved" tests
  - new test asserting equal `runningDuration` across both terminal events for the same worker
- `yarn lint` passes
- `git diff` is limited to `provider.js`, `static.js`, `provider_test.js`, and the changelog


lotas (changes requested):
Thanks for digging into this! I think current changes treats a non-bug as a bug. 

`worker-removed` and `worker-stopped` aren't one event logged twice - they're distinct moments (we requested removal vs. the scanner observed the instance gone), and the `workerLifetime` metric is already deduped by the `stoppedAt` guard. 

The only thing that differed was the log `runningDuration`, because each handler computes `now − registeredAt` at a different time - and that gap (often large on Azure, where deprovisioning is slow) is real termination latency. 

Freezing both lines to the first event's `stoppedAt` makes the authoritative terminal event report a stale value just to look consistent. I'd rather flip it: emit anchored, non-overlapping segments from persisted timestamps:
- `registrationDuration` (already have it in worker-registered)
- `runningDuration` (registered->removed)
- and a new `deprovisionDuration` (removed->stopped) - each at the event where it becomes final. 

That removes the inconsistency, needs only a set-once `removedAt` in the existing `providerData` jsonb.

We could add `deprovisionDuration` here: https://github.com/taskcluster/taskcluster/blob/fe314d92315c5ca248e56d34494eac8f42eb7122/services/worker-manager/src/monitor.js#L48-L62

nitishagar:
Reworked per your review feedback. The single-frozen-timestamp approach is gone; this now implements the anchored, non-overlapping duration model.

**Duration model**
- `removedAt` is persisted write-once in `providerData.workerManager` when removal is requested.
- `runningDuration` = `removedAt − registeredAt`, reported on **both** `workerRemoved` and `workerStopped` (anchored at the removal moment, consistent by construction).
- New `deprovisionDuration` = `stoppedAt − removedAt`, reported **only** on `workerStopped` (the authoritative terminal event), where it has a real value. It's omitted from `workerRemoved` because `stoppedAt` doesn't exist yet at remove time.

**Concurrency (the race in the previous `_ensureStoppedAt`)**
The old helper read `stoppedAt` *outside* the update and returned its own local `now`, so two concurrent terminal events could each emit a different "first" timestamp. The new `_ensureRemovedAt` / `_ensureStoppedAt` each run their read-compare-write **inside** the `worker.update` optimistic-concurrency callback (etag-guarded UPDATE + retry on stale) and **re-read the persisted value after the update resolves**. Under concurrent remove + stop, the loser's write is discarded and reloaded, so both callers converge on the winner's timestamp.

Added a dedicated test for this (`concurrent onWorkerRemoved + onWorkerStopped cannot emit different first timestamps`) that loads two independent `Worker` instances of the same row and fires both events via `Promise.all` against the real test Postgres, so the etag/P0004 retry is genuinely exercised.

nitishagar:
Fix for the failing `meta-generate` check: the updated monitor log-type fields in `worker-manager/src/monitor.js` weren't reflected in `generated/references.json` (`runningDuration` description changes, new `deprovisionDuration` field, `workerStopped` version bump 2 → 3). Re-ran `yarn generate` and amended the result into the merge commit. The earlier `client-go` failure looks like a live-integration flake — no Go/API surface is touched by this PR and it passes on the re-run.

nitishagar:
Note: `service-worker-manager` is currently red across PRs with 24 `provider_azure_test.js` "Signature validation error" failures — that's the known fixture expiry tracked in #8933 and predates/unrelated to this change.

nitishagar:
The `service-worker-manager` failures turned out not to be the #8933 flake after all: this branch was still carrying the pre-refresh Azure signature fixtures from its July base, which now fail validation against the current date (verified locally — 24 `provider_azure_test.js` failures on the old tree, 0 on current `main`). Rebased onto current `main` as a single commit (which brings in the refreshed `azure_signature_good.json` + CA certs) and re-amended the `yarn generate` output for `generated/references.json`. Full worker-manager suite now passes locally: 501 passing, 0 failing.

PR #8693, opened 2026-05-29
feat(tooling): support Helm v4 for dev k8s rendering

### LINKED ISSUE #8259: Support helm v4
Helm v4 is out https://helm.sh/docs/overview/

But we only claim to support v2 and v3 https://github.com/taskcluster/taskcluster/blob/7e335dd50c7d98dfe083b4c5a2ee64406544fc0b/infrastructure/tooling/src/dev/helm.js#L59

Let's look into what we need to do (if anything) to support v4 and update this check.

nitishagar:
## What

The Helm version detection in `infrastructure/tooling/src/dev/helm.js` only accepted v2 and v3, throwing an error for anything else. This change parses the major version from the detected Helm version string and accepts v3 **or higher**, so Helm v4 (and future major versions) work without changes.

Additionally, `infrastructure/k8s/Chart.yaml` previously had no `apiVersion` field, which caused Helm to treat it as a legacy v1 chart. Helm v4 is stricter about this, so `apiVersion: v2` is declared explicitly.

The downstream `helm-version === 2` check for the `-n` flag is unaffected — v3/v4/v5+ return their actual major number, so the v2-only codepath is correctly skipped.

## Why

Fixes #8259. Helm v4 was released and users running it against this repo received an error: "Must use supported helm version (2 or 3)."

## How verified

- `yarn lint` passes
- `git diff` is limited to `helm.js`, `Chart.yaml`, and the changelog
- No `=== 3` equality checks for `helm-version` exist in the codebase; only the `=== 2` check remains, which is correct


infrastructure/tooling/src/dev/helm.js
```diff
@@ -60,12 +60,14 @@ const actions = [
       const match = /(SemVer|Version):"(v[0-9]+.[0-9]+.[0-9]+-?[^"]*)"/g.exec(res);
       if (!match) {
         throw new Error(`Could not determine helm version from: ${res}`);
-      } else if (match[2].includes('v3')) {
-        return { 'helm-version': 3 };
-      } else if (match[2].includes('v2')) {
+      }
+      const major = parseInt(match[2].replace(/^v/, '').split('.')[0], 10);
+      if (major >= 3) {
```
lotas: I think this check should be explicit (supported versions >=2 <=4), and the the one below is not needed
```js
        return { 'helm-version': 2 };
```

changelog/issue-8259.md
```diff
@@ -0,0 +1,6 @@
+audience: deployers
+level: patch
+reference: issue 8259
+---
+Development tooling now supports Helm v4 (and any future v3+ major) for rendering k8s
```
lotas: I would probably leave out "any future v3+" because those might introduce breaking changes, it's better to leave support for what's actually working and tested at the moment

infrastructure/k8s/Chart.yaml
```diff
@@ -1,3 +1,4 @@
+apiVersion: v2
```
lotas, line 1: `apiVersion: v2` requires Helm v3+ so keeping support of v2 in code will be misleading.
We should either make this conditional (for v3+) or drop overall support for v2

lotas (changes requested):

changelog/issue-8259.md
nitishagar: The changelog now says "supports Helm v3 and v4" only, and dev-docs/dev-deployment.md is updated to match ("Helm 3 or 4 installed").

infrastructure/k8s/Chart.yaml
nitishagar: Went with the second option; dropped v2 support entirely: the gate rejects < 3, the dead === 2 → -n branch is deleted, and the changelog + dev-docs both say v3/v4. Since there's no v1 chart in the repo, making it conditional would just preserve dead code, so removing it was the cleaner consistent resolution.

infrastructure/tooling/src/dev/helm.js
nitishagar: The gate is now explicit and bounded to >= 3 && <= 4 with a message naming "3 or 4", and the helm-version === 2 path (including the command.push('-n') branch) is removed entirely. Dropped v2 rather than just narrowing it, per your point 3 below.

PR #8710, opened 2026-06-03
Bug 2042324 - Bind TC credential issuance through oauth to the access token's lifetime

Eijebong:
Taskcluster was happily issuing credentials for an expired access token if the requested credentials wouldn't have expired yet and the daily cronjob hadn't cleaned it. This is the same class of bug as what 5e609add5151e231d42024102a6707b10c917c72 fixed (I even looked at this condition there and missed that it was incomplete), we were checking against the resulting credentials lifetime instead of the access token that was passed to us. In this case though we check against both, we don't want to give credentials that have expired already (makes no sense).

Note that the diff looks weird for the test because git tries to be smart, I didn't touch the test above the new one.

### DIFF
--- a/services/web-server/src/servers/oauth2.js
+++ b/services/web-server/src/servers/oauth2.js
@@ -326,8 +326,10 @@ export default (cfg, db, strategies, auth, monitor) => {
     }
 
     // Although we eventually delete expired rows, that only happens once per day
-    // so we need to check that the accessToken is not expired.
-    if (new Date(entry.client_details.expires) < new Date()) {
+    // so we need to check that the access token is not past its own lifetime, as
+    // well as that the requested credentials have not already expired.
+    const now = new Date();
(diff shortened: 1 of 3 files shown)


services/web-server/src/servers/oauth2.js
```diff
@@ -326,8 +326,10 @@ export default (cfg, db, strategies, auth, monitor) => {
     }
 
     // Although we eventually delete expired rows, that only happens once per day
-    // so we need to check that the accessToken is not expired.
-    if (new Date(entry.client_details.expires) < new Date()) {
+    // so we need to check that the access token is not past its own lifetime, as
+    // well as that the requested credentials have not already expired.
+    const now = new Date();
+    if (new Date(entry.expires) < now || new Date(entry.client_details.expires) < now) {
```
lotas, line 332: this looks so obvious .. in the hindsight .. 👍

lotas (approved):
Thank you

PR #8712, opened 2026-06-04
Upgrade node and go to the latest available

lotas:
node `24.16.0`
go `go1.26.4`


Eijebong:
https://community-tc.services.mozilla.com/tasks/XXv9DORhTP6DIZfCQuv8zA/runs/0/logs/public/logs/live.log#L3280

Wonder what happened here. I'll try to think about investigating this next week.

Eijebong (approved):

Eijebong (changes requested):
That node bump is actually what breaks CI (hangs on playwright install).

Eijebong:
https://github.com/taskcluster/taskcluster/pull/8715 should fix the issue

Eijebong (approved):

PR #8715, opened 2026-06-08
Bump playwright to 1.60.0

Eijebong:
This version fixes a bug with more recent versions of nodejs where the browser download could end up hanging at 100%. See https://github.com/microsoft/playwright/pull/40747 and linked issues there for some details.


lotas (approved):
awesome, thanks for a quick fix!

PR #8719, opened 2026-06-09
Fix a typo in the github service intree scope normalization tests

Eijebong:
The test was comparing two strings which means that code block was dead. That `'key'` is quite clearly a typo given it's defined on the line above. I reordered one of the fixtures to get the scopes out of order on purpose so the code would need to be exercised for tests to pass.

### DIFF
--- a/services/github/test/intree_test.js
+++ b/services/github/test/intree_test.js
@@ -94,7 +94,7 @@ suite(testing.suiteName(), function() {
         assert.equal(config.tasks.length, count);
       }
       for (let key of Object.keys(expected)) {
-        if ('key' === 'scopes') {
+        if (key === 'scopes') {
           expected[key].sort();
         }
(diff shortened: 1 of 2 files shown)


services/github/test/intree_test.js
```diff
@@ -94,7 +94,7 @@ suite(testing.suiteName(), function() {
         assert.equal(config.tasks.length, count);
       }
       for (let key of Object.keys(expected)) {
-        if ('key' === 'scopes') {
```
lotas, line 97: wow, what a typo? 👍

lotas (approved):
thanks

services/github/test/intree_test.js
Eijebong: Yeah, biome was not happy about that :p I'm surprised eslint wasn't screaming tbh

lotas: good job, biome!

PR #8723, opened 2026-06-10
Use the `node:` protocol for built-in module imports

Eijebong:
This applies biome's `useNodejsImportProtocol` lint across all services/libs, prefixing builtin imports with `node:`.

Unlike the web UI, this code is server only and must never have its builtins polyfilled. The `node:` prefix enforces that. If any of it is ever pulled into a browser bundle, the import fails loudly instead of silently resolving to a webpack shim.


lotas (approved):
awesome!

PR #8725, opened 2026-06-10
Use `Array.isArray` instead of `instanceof Array` to fix biome's useIsArray

Eijebong:
In practice this is going to make no difference whatsoever. But biome calls it out because in theory you could get an array from a different domain (an iframe), which would get checked against the Array prototype of the current domain and differ. In this case, that's never going to happen, but this follows what MDN
[recommends](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#instanceof_vs._array.isarray), works, and silences a biome warning.

### DIFF
--- a/changelog/fpsSGZHLSzOeusHyRqelCg.md
+++ b/changelog/fpsSGZHLSzOeusHyRqelCg.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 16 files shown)


lotas (approved):
nice, TIL about iframe realms and array is not being array

PR #8730, opened 2026-06-10
Always provide the base when using parseInt

Eijebong:
This avoids javascript parsing strings as a wrong number instead of raising an exception and fixes biome's useParseIntRadix lint.

In practice all of those are internal except for the worker manager ones, and every single one of those should always be base 10 so behavior is unchanged


lotas (approved):
10

PR #8734, opened 2026-06-10
fix: add response body limit and close body in Request

MatrixNeoKozak:
The `Request` method in `clients/client-go/http.go` lacked a size limit for reading response bodies, creating a potential Out of Memory (OOM) vulnerability when processing large API responses. Additionally, the method failed to close the response body, leading to resource leaks (e.g., unclosed file descriptors/connections). 

This change:
1. Introduces a 10MB limit on the response body read using `io.LimitReader` to prevent OOM attacks.
2. Ensures the response body is explicitly closed using `defer` when a response is received, mitigating resource leaks.
3. Removes dead, commented-out code related to response dumping to improve maintainability.

These changes enhance the robustness and reliability of the Taskcluster Go client.


lotas:
Why is lock file removed?

lotas:
what problem does this PR solve? Is there an open issue for this?
If you believe this is a valid problem and is affecting someone/something, please file an issue/bug first.
Thank you

MatrixNeoKozak:
Thanks for the feedback. The lock file removal was unintentional — it was a side effect of how the branch was prepared and should not have been included in the diff. Apologies for that noise.

On the core change: the concern is that reading an unbounded API response body has no size limit, which could cause OOM in a scenario where a server returns an unexpectedly large or malicious payload. The missing `defer resp.Body.Close()` is a resource leak — if the body is never closed, the underlying TCP connection stays open and won't be returned to the pool.

I'll file a proper issue first as you suggested, and can reopen or recreate the PR against that issue with the lock file properly restored.

PR #8735, opened 2026-06-11
Fix typos where `,` was used instead of `;` as the end of line

Eijebong:
This fixes biome's `noCommaOperator` lint. Those typos didn't hide any bug and behavior is unchanged.


lotas (approved):
wow, why didn't it fail in the past

```js
> 1,2,3
3
```
TIL https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_operator

PR #8739, opened 2026-06-11
Remove useless escapes in regexes

Eijebong:
Because we're using literal regexes (and not `new Regex("..")`), there's a few places where escape sequences are not necessary. Removing them makes regexes more "readable".

Fixes biome's noUselessEscapeInRegex lint


lotas (approved):
\ok

PR #8740, opened 2026-06-11
Use optional chain when possible

Eijebong:
Replace `a && a.foo` with `a?.foo` which is more idiomatic javascript.

Fixes biome's useOptionalChain lint


Eijebong:
That test failure is real... Wonder why it's not accepting that syntax though

lotas (approved):
awe?.some, thanks

PR #8743, opened 2026-06-11
Fix documentation for the secrets API mentioning tasks/provisioners instead of secrets

Eijebong:
This was copy/pasted from the tasks API docs and the author forgot to change some references


Eijebong:
CC @jcristau

lotas (approved):
nice catch, thanks

PR #8745, opened 2026-06-12
Use arrow functions instead of `function() {}`

Eijebong:
These are less verbose but mostly, they avoid the huge javascript footgun of having this not being inherited from the parent scope and instead being the function itself which is pretty much never what you want...

There was only one case where the fix broke something and it was in a mock for google oauth2 in worker-manager because we're using that function as a constructor which arrow functions can't be used as (IMO, also a win...). I applied `useArrowFunction` in this commit and reverted that one change manually which will explain why a second run would get that different. Since biome's not in yet I decided to keep the `biome-ignore` rule for a later commit where it'll actually make sense.

Fixes biome's useArrowFunction lint

### DIFF
--- a/changelog/DjlRPVHFRZyF9ax9SfPLdg.md
+++ b/changelog/DjlRPVHFRZyF9ax9SfPLdg.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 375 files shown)


lotas (approved):
it is a bit unrealistic to check each refactoring, so I hope biome/eslint/ tests caught everything else 🤞

PR #8747, opened 2026-06-12
Remove useless constructors in worker-manager tests

Eijebong:
Those constructors are exactly what JS does, biome doesn't like them, doesn't change any behavior

Fixes biome's noUselessConstructor lint


lotas (approved):
super!

PR #8748, opened 2026-06-12
Remove useless `this` aliases in the index service

Eijebong:
Because #8745 transformed those functions into arrow functions, we don't need the weird `that = this` alias anymore which biome catches and complain about.

Fix biome's noUselessThisAlias lint


lotas (dismissed):
lol. that.queue is the same as this.queue. they are equal .. is there another language with similar fun? :D

PR #8749, opened 2026-06-12
Remove useless escape sequences in strings

Eijebong:
Fixes biome's noUselessEscapeInString lint

### DIFF
--- a/infrastructure/tooling/src/meta/checks/node-pg.js
+++ b/infrastructure/tooling/src/meta/checks/node-pg.js
@@ -9,7 +9,7 @@ export const tasks = [{
   run: async () => {
     // This checks one of the tc-lib-postgres security invariants, that
     // services are not using postgres directly
-    for (const pattern of ['require\(.pg\)', '_withClient']) {
+    for (const pattern of ['require(.pg)', '_withClient']) {
       try {
         const res = await execFileAsync('git', ['grep', pattern, '--', 'services/']);
         // if the grep succeeded, then something matched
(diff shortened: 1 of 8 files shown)


infrastructure/tooling/src/meta/checks/node-pg.js
```diff
@@ -9,7 +9,7 @@ export const tasks = [{
   run: async () => {
     // This checks one of the tc-lib-postgres security invariants, that
     // services are not using postgres directly
-    for (const pattern of ['require\(.pg\)', '_withClient']) {
+    for (const pattern of ['require(.pg)', '_withClient']) {
```
lotas, line 12: hm. this is passed to grep. i wonder what it would match `require(['"]pg)` - but there's no closing one :) 

plus, I forgot to update this one when I did switch to esmodules 

I checked following locally: 
```suggestion
    for (const pattern of ['require(.pg.)', 'import.* from .pg.;', '_withClient']) {
```

lotas (changes requested):

infrastructure/tooling/src/meta/checks/node-pg.js
lotas: or we can turn it into separate issue if you'd like to avoid rebasing 300 commits, since it was broken already

lotas (approved):
I'll approve because bug existed before, but we'll need to create new issue if this is not applied

infrastructure/tooling/src/meta/checks/node-pg.js
lotas: Filed #8761

PR #8750, opened 2026-06-12
Use template strings instead of concatenating when possible

Eijebong:
Fixes biome's useTemplate lint


lotas (approved):
too big to review manually every line, I hope tests caught everything

PR #8755, opened 2026-06-12
feat(ui): show metadata for recently viewed tasks and task groups

### LINKED ISSUE #8711: Improve Tasks/Task Group UI pages
Currently pages like:
https://firefox-ci-tc.services.mozilla.com/tasks
https://firefox-ci-tc.services.mozilla.com/tasks/groups

show last 5 task ids and nothing more. Those might be tasks opened 5 seconds ago or 5 years ago, and contain no additional clues to what it might be 

Image

Those are kept in browser storage. 
It would be much more helpful to show extra information in here and at least include some timestamps: 
- metadata.name
- metadata.source (truncated probably)
- taskQueueId
- created/deadline 

And for the task groups probably task counts and resolved/unresolved state

The balance should be between storing too much in the browser cache vs making this a live tracking page with background refreshes. We probably don't want to overstep into Treeherder territory, but from the user perspective it would be way more easier to reason what those opened tasks were.

We could also increase the number of items we display/store to 20-50 probably

nitishagar:
## Summary

The `/tasks` and `/tasks/groups` pages previously listed the last 5 recently-viewed items as bare IDs with no context. This PR stores richer metadata at view time and displays up to 20 entries ordered by recency.

### Before
- 5 bare task/group IDs with no context about what they were

### After
- Up to 20 entries, ordered by when you last visited
- **Tasks**: shows task name (falls back to ID), status label, task queue, age, and truncated source URL
- **Task groups**: shows decision-task name, task queue, age, source URL, task count breakdown (completed/failed/running/pending), and a resolved/running indicator

### Design decisions

- **Store-at-view-time snapshot**: metadata is captured when the page is first loaded, not refreshed in the background. Keeps browser storage small and avoids replicating live task tracking (Treeherder territory).
- **20-item limit**: up from 5; a `viewedAt` secondary index on both Dexie tables makes `orderBy('viewedAt').reverse().limit(20)` efficient.
- **Legacy record graceful degradation**: existing v1 records lack `viewedAt` and are excluded by the index query — they age out naturally as the user browses and new rich records fill in. Mentioned in tests.
- **Snapshot counts for task groups**: status counts may be partial if not all pages were fetched when the group was viewed — acceptable for a history snapshot.

### Possible follow-up (out of scope)
A comment on the issue suggested a "my recent tasks" filtered by owner. That would require storing `task.metadata.owner` and filtering; noted here for future work.

### Changes
- `ui/src/utils/db.js`: `db.version(2)` adds `viewedAt` secondary index
- `ui/src/views/Tasks/ViewTask/index.jsx`: `updateTaskIdHistory` stores name, source, taskQueueId, created, deadline, state, viewedAt
- `ui/src/views/Tasks/TaskGroup/index.jsx`: `updateTaskGroupIdHistory` stores decision-task metadata + statusCount + viewedAt
- `ui/src/views/Tasks/NoTask/index.jsx`: queries by viewedAt, renders rich list
- `ui/src/views/Tasks/NoTaskGroup/index.jsx`: queries by viewedAt, renders rich list with status counts
- Tests added for both display components (jest.mock for IndexedDB, covers rich + legacy + empty states)

Fixes #8711

### DIFF
--- a/ui/src/utils/db.js
+++ b/ui/src/utils/db.js
@@ -9,6 +9,13 @@ db.version(1).stores({
   taskDefinitions: 'metadata.name, created',
 });
 
+db.version(2).stores({
+  taskIdsHistory: 'taskId, viewedAt',
+  taskGroupIdsHistory: 'taskGroupId, viewedAt',
+  userPreferences: '',
+  taskDefinitions: 'metadata.name, created',
+});
+
 /**
  * A wrapper around indexDB. If you're not dealing with collections,
  * use localForage or localStorage instead.
--- a/ui/src/utils/db.js
+++ b/ui/src/utils/db.js
@@ -9,12 +9,39 @@ db.version(1).stores({
   taskDefinitions: 'metadata.name, created',
 });
 
-db.version(2).stores({
-  taskIdsHistory: 'taskId, viewedAt',
-  taskGroupIdsHistory: 'taskGroupId, viewedAt',
-  userPreferences: '',
-  taskDefinitions: 'metadata.name, created',
-});
+db.version(2)
+  .stores({
+    taskIdsHistory: 'taskId, viewedAt',
+    taskGroupIdsHistory: 'taskGroupId, viewedAt',
+    userPreferences: '',
+    taskDefinitions: 'metadata.name, created',
+  })
+  .upgrade(trans =>
+    // Back-fill viewedAt on records written before v2, which lacked the index.
+    // Without this, legacy records are silently absent from the new
+    // `.orderBy('viewedAt')` index and never appear in the recent lists again.
+    // Sentinel 0 sorts last under the descending recency order, so legacy
+    // entries render but age out as the user browses — they are never dropped
+    // (review comment discussion_r3579292033).
+    trans
+      .table('taskIdsHistory')
+      .toCollection()
+      .modify(rec => {
+        if (rec.viewedAt === undefined) {
+          rec.viewedAt = 0;
+        }
+      })
+      .then(() =>
+        trans
+          .table('taskGroupIdsHistory')
+          .toCollection()
+          .modify(rec => {
+            if (rec.viewedAt === undefined) {
+              rec.viewedAt = 0;
+            }
+          })
+      )
+  );
 
 /**
  * A wrapper around indexDB. If you're not dealing with collections,
--- a/ui/src/utils/db.js
+++ b/ui/src/utils/db.js
@@ -17,12 +17,7 @@ db.version(2)
     taskDefinitions: 'metadata.name, created',
   })
   .upgrade(trans =>
-    // Back-fill viewedAt on records written before v2, which lacked the index.
-    // Without this, legacy records are silently absent from the new
-    // `.orderBy('viewedAt')` index and never appear in the recent lists again.
-    // Sentinel 0 sorts last under the descending recency order, so legacy
-    // entries render but age out as the user browses — they are never dropped
-    // (review comment discussion_r3579292033).
+    // Back-fill viewedAt, or pre-v2 records stay absent from the new index.
     trans
       .table('taskIdsHistory')
       .toCollection()
--- a/ui/src/views/Tasks/NoTaskGroup/index.jsx
+++ b/ui/src/views/Tasks/NoTaskGroup/index.jsx
@@ -1,13 +1,17 @@
 import React, { Component } from 'react';
+import { sum } from 'ramda';
 import { withStyles } from '@material-ui/core/styles';
 import Typography from '@material-ui/core/Typography';
 import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
+import ListItemText from '@material-ui/core/ListItemText';
 import ListSubheader from '@material-ui/core/ListSubheader';
 import LinkIcon from 'mdi-react/LinkIcon';
 import Dashboard from '../../../components/Dashboard';
 import HelpView from '../../../components/HelpView';
 import Search from '../../../components/Search';
+import DateDistance from '../../../components/DateDistance';
+import StatusLabel from '../../../components/StatusLabel';
 import Link from '../../../utils/Link';
 import db from '../../../utils/db';
 
@@ -20,6 +24,19 @@ import db from '../../../utils/db';
     display: 'flex',
     justifyContent: 'space-between',
   },
+  secondaryLine: {
+    display: 'flex',
+    alignItems: 'center',
+    gap: theme.spacing(1),
+    flexWrap: 'wrap',
+  },
+  source: {
+    textOverflow: 'ellipsis',
+    overflowX: 'hidden',
+    whiteSpace: 'nowrap',
+    maxWidth: 260,
+    display: 'inline-block',
+  },
 }))
 export default class NoTaskGroup extends Component {
   state = {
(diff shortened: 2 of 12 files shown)


lotas:
Thank you very much @nitishagar  I was testing it locally and it looks great so far! 
image

We store more data indeed now and display it in a nice way.

This looks great on a narrow/mobile view, but I wonder if we could make it responsive also by using extra space to fill the line? Use more columns to display all that info instead of cramping it into a narrow single column? 


TaskGroup page does not show anything here but maybe that's a known issue, and is a separate bug
image
upd: taskgroup is a separate issue I commented below, it depends on `taskActions` which is optional

ui/src/views/Tasks/ViewTask/index.jsx
```diff
@@ -63,12 +63,21 @@ import pageArtifactsQuery from './pageArtifacts.graphql';
...
 
-  db.taskIdsHistory.put({ taskId: id });
+  db.taskIdsHistory.put({
+    taskId: id,
+    name: task?.metadata?.name,
+    source: task?.metadata?.source,
+    taskQueueId: task?.taskQueueId,
+    created: task?.created,
+    deadline: task?.deadline,
+    state: task?.status?.state,
```
lotas, line 82: I think it is fine to store the actual state, but worth keeping in mind it might change at any point. So when we look at this page we probably want to make it obvious that this was the state at the time we looked at this task: 
Additionally  we could re-fetch statuses for each given task and display actual one, and this could be a separate issue from this likely. 
I just want to avoid the confusion when person would look at this and wonder why my task is still Pending/Running/etc 

Image

ui/src/views/Tasks/TaskGroup/index.jsx
```diff
@@ -220,7 +228,11 @@ export default class TaskGroup extends Component {
       taskGroupId !== state.previousTaskGroupId &&
       taskActions
     ) {
```
lotas: Ah, I know what the issue here is - `taskActions` are not present everywhere

So this condition should be split into two - we could drop `&& taskActions` so we always update `taskGroupIdHistory` and only do `taskActions.actions` below if `taskActions` are defined, this way we'd always store updated task group Id

Image

Btw, we don't show those newly stored columns here

lotas (changes requested):

nitishagar:
@lotas Addressed comments on the PR:
- split the taskActions guard so groups are always recorded in history
- added "viewed N ago" + "recorded at view time" labels so the stored state/status reads as a point-in-time snapshot rather than live 

Also made the recent lists responsive (two columns on wide screens, one on mobile).
**Mobile:** 
pr8755_tasks_narrow
**Web:** 
pr8755_tasks_wide

lotas:
> @lotas Addressed comments on the PR:
>     * split the taskActions guard so groups are always recorded in history
>     * added "viewed N ago" + "recorded at view time" labels so the stored state/status reads as a point-in-time snapshot rather than live
> Also made the recent lists responsive (two columns on wide screens, one on mobile). 


thank you for the updates and adding "viewed N days ago" feature, I like it. 

But I think we need to tweak the UI a bit, currently it doesn't look good:
- mobile view is displaying 4 column layout where each column is broken into multiple lines - it is really difficult to understand. Most of the responsive mobile would display those row-wise, not column-wise. 
- wide view is also not great:
  * there is no reason to break columns into multiple lines (like "1 day<nl>ago", etc)
  * having two tables side by side also "a unique solution", there should be 1 task per line
  * `taskId` is not even seen anymore? can we please always have it 

Thanks!

nitishagar:
@lotas Thanks for the detailed feedback. Reworked both lists from the card grid to the standard `DataTable` component used elsewhere in the UI (Clients, Hooks, Roles, …):

- **One entry per line**: single full-width table, no more side-by-side columns
- **`taskId` / `taskGroupId` always visible**: first column, monospace with a copy-to-clipboard button
- **No mid-phrase line breaks**: date/age cells are `nowrap`; the table scrolls as a unit instead of wrapping cell contents
- **Mobile is row-wise**: the app's global responsive table styles (`MuiTableCell` stacks block-wise below the `md` breakpoint) kick in automatically, so each field renders on its own line, same as every other table in the UI
- Kept the "viewed N ago" column and the "recorded at view time; may be stale" tooltip on the status chips

Verified against community-tc with the dev server: console and Network tab clean on both pages.

**Wide — /tasks:**
tasks_wide

**Wide — /tasks/groups:**
groups_wide

**Mobile — /tasks:**
tasks_mobile

**Mobile — /tasks/groups:**
groups_mobile

ui/src/utils/db.js
```diff
@@ -9,6 +9,13 @@ db.version(1).stores({
   taskDefinitions: 'metadata.name, created',
 });
 
+db.version(2).stores({
+  taskIdsHistory: 'taskId, viewedAt',
+  taskGroupIdsHistory: 'taskGroupId, viewedAt',
+  userPreferences: '',
+  taskDefinitions: 'metadata.name, created',
+});
```
lotas: I'm not sure how to fully test it, but likely what will happen here is newly added `viewedAt` is not backfilled, so all existing entries would be silently dropped I believe according to docs 

https://dexie.org/docs/Version/Version.upgrade()

we can add some default values with `.upgrade()` to make sure we don't loose anything

ui/src/views/Tasks/NoTaskGroup/index.jsx
```diff
@@ -39,6 +45,106 @@ export default class NoTaskGroup extends Component {
...
+
+    return (
+      <TableRow key={taskGroupId}>
+        <CopyToClipboardTableCell
+          tooltipTitle={taskGroupId}
+          textToCopy={taskGroupId}
+          text={<code>{taskGroupId}</code>}
+        />
+        <TableCell>
+          {name ? (
```
lotas: We only add name if decision task was present according to: https://github.com/taskcluster/taskcluster/pull/8755/changes#diff-a7ecbd68649fbfa12bd81eb0187a3fe77919ab3ef2df07d90136600038881df7R104

We should probably fallback to taskGropuId if it wasn't set to not show blank unclickable cell

lotas:
this is awesome, thanks @nitishagar I checked locally and it renders great indeed.
I only left two very small comments about the default value migration and empty name check for task groups without decision task (decision task is something that has taskId matching taskGroupId, but it's not a must, so might not be present)

ui/src/utils/db.js
nagarwal-godaddy: Added a `db.version(2).upgrade(trans => ...)` that back-fills `viewedAt: 0` on legacy records (those without a `viewedAt`) in both `taskIdsHistory` and `taskGroupIdsHistory`, so existing entries re-enter the new `viewedAt` index instead of being silently dropped. The upgrader runs once inside the version change transaction and is a no-op on already-v2 databases.

ui/src/views/Tasks/NoTaskGroup/index.jsx
nagarwal-godaddy: Addressed in this push: the Name cell now falls back to a clickable link to the task-group id when no name is stored, instead of rendering a blank unclickable cell. Applied symmetrically to both NoTaskGroup and NoTask (mirroring the existing `name || id` convention in TaskDetailsCard).

lotas:
I rebased locally and made few changes, because it would be faster this way. I think the PR is in a really good shape now, thanks for working on it

lotas (dismissed):

PR #8756, opened 2026-06-12
fix(github): create build records for all unique taskGroupIds

### LINKED ISSUE #8751: Github service only tracks taskGroupId of first defined task
After parsing the `tasks` list, we only insert `tasks[0]` into the Github builds table:
https://github.com/taskcluster/taskcluster/blob/c6c4718dc1a0c8df58034a3355c5e5febafc686e/services/github/src/handlers/job.js#L400

This means if a `.taskcluster.yml` defines more than one task, none of the subsequent ones will be displayed in Github. As a workaround, they could all use the same taskGroupId, but in many cases that might not be desirable.

Instead we should loop over all unique taskGroupIds and insert a record into the builds table for each.

nitishagar:
## Problem

When a `.taskcluster.yml` defines tasks with different `taskGroupId` values, the GitHub service only recorded `tasks[0]`'s `taskGroupId` in the `github_builds` table and only published `taskGroupCreationRequested` for that first group. All other task groups were invisible to GitHub checks and statuses — the build record was never created, so no status was ever posted.

## What changed

**`services/github/src/handlers/job.js`**

The `graphConfig.tasks` processing block now:
1. Builds an ordered `Map<taskGroupId, routes>` over all tasks, keeping the routes of the first task in each group.
2. Creates one `github_builds` row per unique `taskGroupId` (parallel, like the hooks path).
3. Calls `createTasks` unchanged (still a single call with all tasks).
4. Calls `cancelPreviousTaskGroups` once with the first new build. Sibling builds from the same event share the same `event_id` and are already excluded by the existing `event_id` filter inside `cancelPreviousTaskGroups`, so no interface change was needed there.
5. Publishes `taskGroupCreationRequested` once per unique `taskGroupId` (mirroring the hooks code path).

No database schema or API changes are needed — `github_builds` already supports multiple rows per SHA with different `task_group_id` values, and downstream handlers (`status.js`, `deprecatedStatus.js`, `taskGroupCreation.js`) look up builds by `taskGroupId` and need no changes.

## Test coverage

- New fixture `valid-yaml-v1-multi-group.json`: 3 tasks — task1 and task2 in group A, task3 in group B.
- `multi-group yml creates one build record per unique taskGroupId` — verifies 2 `github_builds` rows (one per group), both `pending`.
- `multi-group yml publishes taskGroupCreationRequested once per unique group` — verifies exactly 2 publishes, one per group ID.
- `multi-group yml calls cancelPreviousTaskGroups once for pull_request` — verifies the function is called exactly once regardless of group count.
- `does not cancel sibling task groups from the same event` (in `cancelPreviousTaskGroups` unit suite) — verifies that builds from the same event (same `event_id`) are not cancelled when `cancelPreviousTaskGroups` is called for one of them.

## Maintainer note

I followed the hooks code path (`job.js:335-392`) as the structural model, which already publishes `taskGroupCreationRequested` once per hook. I'd welcome feedback on whether publishing once per unique `taskGroupId` is the right behaviour, or whether there are downstream consumers that expect a single publish per job event.

Fixes #8751

### DIFF
--- a/services/github/src/handlers/job.js
+++ b/services/github/src/handlers/job.js
@@ -436,32 +436,37 @@ export async function jobHandler(message) {
 
   // Create tasks (if present)
   if (graphConfig.tasks && graphConfig.tasks.length > 0) {
-    let routes;
-    let taskGroupId;
-
+    const taskGroupMap = new Map();
     try {
-      routes = graphConfig.tasks[0].task.routes;
-      taskGroupId = graphConfig.tasks[0].task.taskGroupId;
+      for (const { task } of graphConfig.tasks) {
+        if (!taskGroupMap.has(task.taskGroupId)) {
+          taskGroupMap.set(task.taskGroupId, task.routes);
+        }
+      }
     } catch (e) {
       return await this.createExceptionComment({ debug, instGithub, organization, repository, sha, error: e });
     }
 
-    const build = await createGithubBuildRecord({
-      context,
-      organization,
-      repository,
-      sha,
-      taskGroupId,
-      groupState,
-      installationId: message.payload.installationId,
-      eventType: message.payload.details['event.type'],
-      eventId: message.payload.eventId,
-      pullNumber,
-      debug,
-    });
+    const builds = await Promise.all(
+      [...taskGroupMap.keys()].map(taskGroupId =>
+        createGithubBuildRecord({
+          context,
+          organization,
+          repository,
+          sha,
+          taskGroupId,
+          groupState,
+          installationId: message.payload.installationId,
+          eventType: message.payload.details['event.type'],
+          eventId: message.payload.eventId,
+          pullNumber,
+          debug,
+        }),
+      ),
+    );
 
     try {
-      debug(`Creating tasks for ${organization}/${repository}@${sha} (taskGroupId: ${taskGroupId})`);
+      debug(`Creating tasks for ${organization}/${repository}@${sha} (${taskGroupMap.size} task group(s))`);
       await this.createTasks({ scopes: graphConfig.scopes, tasks: graphConfig.tasks });
     } catch (e) {
       debug(`Creating tasks for ${organization}/${repository}@${sha} failed! Leaving comment on Github.`);
@@ -470,28 +475,32 @@ export async function jobHandler(message) {
 
     // Only cancel previous tasks after we have successfully created new ones.
     // Cancel existing builds for non-default branches.
+    // All sibling builds from this event share the same event_id, so they are already
+    // excluded by the event_id filter inside cancelPreviousTaskGroups.
     if (graphConfig.autoCancelPreviousChecks !== false) {
       if (pullNumber || message.payload.body.ref !== defaultBranch) {
-        await this.cancelPreviousTaskGroups({ instGithub, debug, newBuild: build });
+        await this.cancelPreviousTaskGroups({ instGithub, debug, newBuild: builds[0] });
       }
     }
 
-    try {
-      debug(`Publishing status exchange for ${organization}/${repository}@${sha} (${groupState})`);
-      await context.publisher.taskGroupCreationRequested(
-        {
-          taskGroupId,
-          organization: organization.replace(/\./g, '%'),
-          repository: repository.replace(/\./g, '%'),
-        },
-        routes
-      );
-    } catch (e) {
-      debug(`Failed to publish to taskGroupCreationRequested exchange.
-      Parameters: ${taskGroupId}, ${organization}, ${repository}, ${routes}`);
-      debug(`Stack: ${e.stack}`);
-      return debug(`Failed to publish to taskGroupCreationRequested exchange
-      for ${organization}/${repository}@${sha} with the error: ${stringify(e, null, 2)}`);
+    for (const [taskGroupId, routes] of taskGroupMap.entries()) {
+      try {
+        debug(`Publishing status exchange for ${organization}/${repository}@${sha} (${groupState}, taskGroupId: ${taskGroupId})`);
+        await context.publisher.taskGroupCreationRequested(
+          {
+            taskGroupId,
+            organization: organization.replace(/\./g, '%'),
+            repository: repository.replace(/\./g, '%'),
+          },
+          routes
+        );
+      } catch (e) {
+        debug(`Failed to publish to taskGroupCreationRequested exchange.
+        Parameters: ${taskGroupId}, ${organization}, ${repository}, ${routes}`);
+        debug(`Stack: ${e.stack}`);
+        return debug(`Failed to publish to taskGroupCreationRequested exchange
+        for ${organization}/${repository}@${sha} with the error: ${stringify(e, null, 2)}`);
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


lissyx:
How can we give it a try run on e.g. https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-pr&selectedTaskRun=QWZirOO5QAeNzYD0Y_EHNw.0&revision=87250356048646dcfd44aa5c424eba431ba1dc5f ?

ahal:
> How can we give it a try run on e.g. https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-pr&selectedTaskRun=QWZirOO5QAeNzYD0Y_EHNw.0&revision=87250356048646dcfd44aa5c424eba431ba1dc5f ?

There's not a great way to do that.

For the record, I think this PR has been ignored because it might break assumptions in other parts of the Taskcluster services. So basically someone needs to spend time digging and understanding what the ramifications of a change like this might be (if anything). I think the feature request is sound though.

If this is becoming a high priority, that's a conversation we can have. But the Taskcluster team is also short staffed right now and there is a surge of security reports they are dealing with.

(and just for context, I am not on the Taskcluster team)

nitishagar:
Thanks @ahal

I audited the ramifications against current `main` (`0c1171d5`) and the PR head. My understanding:

_The data model is already task-group keyed, so creating one row per unique `taskGroupId` is sound and needs no DB/API migration._

The one material caveat is the **default Statuses API path**, and it's a behavior this PR *introduces*: today only `tasks[0]`'s group posts a status, so nothing collides. With one build per group, every group posts the **same** status context for the same SHA, and GitHub keeps only the latest per context so a later group's success can hide another group's failure. The opt-in `reporting: checks-v1` path has no such collision.

And, Treeherder should be unaffected; this PR doesn't change task definitions or their queue routes, and Treeherder ingests Queue Pulse events rather than `github_builds` (to the best of my understanding).

The open decision is how to handle the Statuses collision. Two options I see:
1. Ship as-is and accept that multi-group repos on the default Statuses API get a single shared group-level status (last-write-wins).
2. Make the group-level status distinguishable per group (e.g. include `taskGroupId` in the context), or only emit the extra Statuses posts when not using `checks-v1`.

I'm leaning toward (1) for this PR since it's confined and unblocks multi-group visibility, with (2) as a follow-up if needed but would like a pass on that call from @lotas / @petemoore / @matt-boris.

lotas:
I dived a bit into history and posted findings here: https://github.com/taskcluster/taskcluster/issues/8751#issuecomment-5129759484 
so the only thing that might break from multiple tgs is the deprecated statuses API, which are listening to the task group resolved event, and would fail/succeed on the first group to resolve

services/github/src/handlers/job.js
```diff
@@ -436,32 +436,37 @@ export async function jobHandler(message) {
...
-    let routes;
-    let taskGroupId;
-
+    const taskGroupMap = new Map();
     try {
-      routes = graphConfig.tasks[0].task.routes;
-      taskGroupId = graphConfig.tasks[0].task.taskGroupId;
+      for (const { task } of graphConfig.tasks) {
+        if (!taskGroupMap.has(task.taskGroupId)) {
+          taskGroupMap.set(task.taskGroupId, task.routes);
```
lotas: it is possible that some tasks would define different routes, and here we only take first task from the group. Not sure why the assumption before was "routes from the first task". 

I think we should have a Set here to collect all unique routes also

services/github/src/handlers/job.js
```diff
@@ -470,28 +475,32 @@ export async function jobHandler(message) {
...
+            organization: organization.replace(/\./g, '%'),
+            repository: repository.replace(/\./g, '%'),
+          },
+          routes
+        );
+      } catch (e) {
+        debug(`Failed to publish to taskGroupCreationRequested exchange.
+        Parameters: ${taskGroupId}, ${organization}, ${repository}, ${routes}`);
+        debug(`Stack: ${e.stack}`);
+        return debug(`Failed to publish to taskGroupCreationRequested exchange
```
lotas: we should probably not return early here but let all events be published (to not block other groups)

services/github/test/data/yml/valid-yaml-v1-multi-group.json
```diff
@@ -0,0 +1,101 @@
+{
+  "version": 1,
+  "tasks": [
+    {
+      "taskGroupId": "AAAAAAAAAAAAAAAAAAAAAA",
```
lotas, line 5: let's use valid slugid for the tests also - AAA is not a valid one (`taskcluster.slugid()`)

services/github/test/handler_test.js
```diff
@@ -722,6 +764,84 @@ helper.secrets.mockSuite(testing.suiteName(), [], (mock, skipping) => {
       assert.equal(build.state, 'pending');
     });
 
+    test('multi-group yml creates one build record per unique taskGroupId', async () => {
+      const GROUP_A = 'AAAAAAAAAAAAAAAAAAAAAA';
```
lotas: please use `taskcluster.slugid()` to generate valid slugid

lotas (changes requested):
We should also update relevant documentation in https://github.com/taskcluster/taskcluster/blob/a4a715059175615f12bb4436e07da4bcd6157850/ui/docs/reference/integrations/github/taskcluster-yml-v1.mdx#L148-L156 

```
.. multiple task groups are supported under `reporting: checks-v1`, and  legacy Statuses API cannot distinguish them (single commit-status context per sha)
```

services/github/src/handlers/job.js
```diff
@@ -436,32 +436,37 @@ export async function jobHandler(message) {
...
-      sha,
-      taskGroupId,
-      groupState,
-      installationId: message.payload.installationId,
-      eventType: message.payload.details['event.type'],
-      eventId: message.payload.eventId,
-      pullNumber,
-      debug,
-    });
+    const builds = await Promise.all(
```
lotas: I'm thinking maybe it would be better to create builds sequentially because createGithubBuildRecord might throw UNIQUE_VIOLATION, and uses bunch of asserts after which would be easier to validate if created one by one (to avoid race conditions)

https://github.com/taskcluster/taskcluster/blob/a4a715059175615f12bb4436e07da4bcd6157850/services/github/src/handlers/job.js#L56-L66

```js
const builds = [];
for (const taskGroupId of groups.keys()) {
  builds.push(await createGithubBuildRecord({
   ...
```

nitishagar:
Thanks @lotas! All addressed.

Routes are now collected into a Set per group, so every task's routes make it into the
publish (there's a test with a route-less first task and two siblings declaring different
routes, so first-task-wins would visibly fail). The publish loop no longer returns early,
failures are logged and the remaining groups still publish. Switched to your sequential
for-loop for creating builds; a throw still aborts before createTasks, so redelivery goes
through the existing exact-match recovery.

Tests now generate `GROUP_A`/`GROUP_B` with `taskcluster.slugid()` and inject them into the
fixture (which keeps readable placeholders), and there's a small guard test that the ids
actually match the slugid pattern.

lotas (approved):
@nitishagar I think we are in a good shape, thanks for following up with fixes, appreciate it!

Thank you

PR #8759, opened 2026-06-14
Ensure the hooks service cannot fail to serialize errors from the queue

### LINKED ISSUE #8758: When task creation fails from a hook due to the queue service, the error might fail to serialize
Taskcluster was throwing weird looking errors that went away on their own when sheriffs were trying to run hooks: `Retrigger failed with Decision task: VAdR99J3T6Oa0LoBN4El9A: Error: While firing hook: TypeError: Converting circular structure to JSON --> starting at object with constructor 'Agent' | property 'sockets' -> object with constructor 'Object' | property 'taskcluster-queue:80:' -> object with constructor 'Array' | ... | property '_httpMessage' -> object with constructor 'ClientRequest'`.

After having a look, it's actually this, https://github.com/taskcluster/taskcluster/blob/120c26f95febba16e37b66af4e5995a3bcaa610d/services/hooks/src/taskcreator.js#L145

The hook service failed to contact the queue service and when trying to serialize the issue, we get some error about said error not being JSON serializable which we bubble up as is.

Quick repro:

```
import got from 'got';
import http from 'node:http';
const agent = new http.Agent({ keepAlive: true });
try {
  await got('http://192.0.2.1:80/queue', {
    method: 'POST', agent: { http: agent },
    timeout: { request: 1200 }, json: { x: 1 }, retry: { limit: 0 },
  });
} catch (err) {
  console.log(err)
  JSON.stringify(err);
}
```

-> 

```
file:///work/taskcluster/repro.mjs:11
  JSON.stringify(err);
       ^

TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Agent'
    |     property 'sockets' -> object with constructor 'Object'
    |     property '192.0.2.1:80:' -> object with constructor 'Array'
    |     ...
    |     property '_httpMessage' -> object with constructor 'ClientRequest'
    --- property 'agent' closes the circle
    at JSON.stringify (<anonymous>)
    at file:///work/taskcluster/repro.mjs:11:8
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

Eijebong:
When `queue.createTask` failed with some network error, the hooks service tried to `JSON.stringify` it to store it in the `lastFire` table. But errors from `got` contain a circular reference which makes them non JSON serializable. This serialization failure ends up being reported to the client.

Instead of serializing the entire object, pick out the useful fields that would be necessary to debugging why it's failing in the first place and serialize those instead. I also made sure if the error is ever `null` that it can't fail because javascript is awesome and `typeof null` is `object`, as one would definitely expect.

Fixes #8758

### DIFF
--- a/services/hooks/src/taskcreator.js
+++ b/services/hooks/src/taskcreator.js
@@ -141,10 +141,17 @@ export class TaskCreator {
         // of the LastFire table
         let lfError;
 
-        if (typeof err === 'object') {
-          lfError = JSON.stringify(err, null, 2);
+        if (err && typeof err === 'object') {
+          lfError = JSON.stringify({
+            name: err.name,
+            message: err.message,
+            code: err.code,
+            statusCode: err.statusCode ?? err.response?.statusCode,
+            url: err.options?.url?.href,
+            body: err.body,
+          }, null, 2);
         } else {
-          lfError = err.toString();
+          lfError = String(err);
         }
         if (lfError.length > 256 * 1024 / 2) {
           lfError = lfError.substring(0, 256 * 1024 / 2);
--- a/services/hooks/test/taskcreator_test.js
+++ b/services/hooks/test/taskcreator_test.js
@@ -226,6 +226,47 @@ suite(testing.suiteName(), () => {
       throw new Error('should have seen an error from .fire');
     });
 
+    test('firing a hook where the queue throws a network error stores a serialized error', async () => {
+      const hook = await createTestHook([], { firedBy: '${firedBy}' });
+      const taskId = taskcluster.slugid();
+
+      // Build an error with the same shape a failed `got` request carries
+      const circErr = new Error('connect ECONNREFUSED taskcluster-queue:80');
+      circErr.code = 'ECONNREFUSED';
+      const agent = { sockets: {} };
+      const clientRequest = { agent };
+      agent.sockets['taskcluster-queue:80:'] = [{ _httpMessage: clientRequest }];
+      circErr.options = {
+        agent: { http: agent, https: undefined, http2: undefined },
+      };
+
+      creator.fakeCreate = false;
+      const realCreateTask = taskcluster.Queue.prototype.createTask;
+      taskcluster.Queue.prototype.createTask = async () => { throw circErr; };
+
+      let caught;
+      try {
+        await creator.fire(hook, { firedBy: 'me' }, { taskId });
+      } catch (err) {
+        caught = err;
+      } finally {
+        taskcluster.Queue.prototype.createTask = realCreateTask;
+      }
+
+      assert.ok(caught, 'expected fire() to throw');
+      assert.ok(caught === circErr, 'should rethrow the original queue error');
+
+      const [lf] = await helper.db.fns.get_last_fire(
+        hook.hookGroupId,
+        hook.hookId,
+        taskId,
+      );
+      assume(lf.result).to.equal('error');
+      assume(lf.error).to.match(/ECONNREFUSED/);
+      assume(lf.fired_by).to.equal('me');
+      assertFireLogged({ firedBy: 'me', taskId, result: 'failure' });
+    });
+
     test('firing a real task that sets its own task times works', async () => {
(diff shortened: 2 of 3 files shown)


services/hooks/src/taskcreator.js
```diff
@@ -141,10 +141,18 @@ export class TaskCreator {
...
-        if (typeof err === 'object') {
-          lfError = JSON.stringify(err, null, 2);
+        if (err && typeof err === 'object') {
+          lfError = JSON.stringify({
+            name: err.name,
+            message: err.message,
+            code: err.code,
+            statusCode: err.statusCode ?? err.response?.statusCode,
+            url: err.options?.url?.href,
+            body: err.body,
```
hneiva, line 151: could this ever contain a token/secret of some sort?

lotas: very unlikely, unless the original request sent something. Queue service "normally" doesn't send back secrets

lotas: but, it might contain some unserializable entities on it's own, so maybe worth catching it also

services/hooks/src/taskcreator.js
```diff
@@ -141,10 +141,18 @@ export class TaskCreator {
...
+            stack: err.stack,
```
lotas: I would probably not persist stacktrace in db/ui, and instead only logged it

services/hooks/test/taskcreator_test.js
```diff
@@ -226,6 +226,45 @@ suite(testing.suiteName(), () => {
...
+      const hook = await createTestHook([], { firedBy: '${firedBy}' });
+      const taskId = taskcluster.slugid();
+
+      // Build an error with the same shape a failed `got` request carries
+      const circErr = new Error('connect ECONNREFUSED taskcluster-queue:80');
+      circErr.code = 'ECONNREFUSED';
+      const agent = { sockets: {} };
+      const clientRequest = { agent };
+      agent.sockets['taskcluster-queue:80:'] = [{ _httpMessage: clientRequest }];
+      circErr.agent = agent;
```
lotas: this is an incorrect place to put agent, it should be in `options`:

```js
import taskcluster from '@taskcluster/client';

const queue = new taskcluster.Queue({
  rootUrl: 'http://127.0.0.1:1',
  retries: 0,
  credentials: { clientId: 'x', accessToken: 'y' },
});

try {
  await queue.createTask(taskcluster.slugid(), {});
} catch (err) {
  console.log('ownKeys', Object.keys(err));
  console.log('hasTopAgent', Object.hasOwn(err, 'agent'));
  console.log('optionsAgentKeys', Object.keys(err.options?.agent || {}));

  try {
    JSON.stringify(err);
    console.log('stringify ok');
  } catch (e) {
    console.log('stringify error', e.message);
  }
}
```

shows:

```
ownKeys [ 'name', 'code', 'input', 'timings', 'options' ]
hasTopAgent false
optionsAgentKeys [ 'http', 'https', 'http2' ]
stringify error Converting circular structure to JSON
    --> starting at object with constructor 'Agent'
    |     property 'sockets' -> object with constructor 'Object'
    |     property '127.0.0.1:1:' -> object with constructor 'Array'
    |     ...
    |     property '_httpMessage' -> object with constructor 'ClientRequest'
    --- property 'agent' closes the circle
```

lotas (changes requested):

services/hooks/src/taskcreator.js
Eijebong: I don't think it can though. It's just a string.

services/hooks/test/taskcreator_test.js
Eijebong: Right. Missed a level and didn't realize because it was circular already

services/hooks/src/taskcreator.js
Eijebong: Don't think we want to log it though. Because any user error would also end up here? I'll just remove it.

lotas (approved):
Thanks

PR #8766, opened 2026-06-15
Don't return and assign from functions in a single expression

Eijebong:
This is usually hiding actual typos (missing `=` for example) and is always a bit unclear just to save what's most of the time a single line of code.

The hintpoller one is substantial because the previous bespoke implementation was something that you'd normally only ever see in C code golf. There should be no behavior change but it's slightly more readable now.

Fix biome's noAssignInExpressions lint


lotas (approved):
good ones!

PR #8767, opened 2026-06-15
Fix biome's useBlockStatements lint

Eijebong:
I have no idea why this got eslint ignored to save a pair of curly braces. This isn't the kind of braces you want to ignore to not get bankrupt...


services/worker-manager/src/providers/aws.js
```diff
@@ -116,7 +116,10 @@ export class AwsProvider extends Provider {
     let toSpawnCounter = toSpawn;
     for await (const lc of shuffledConfigs) {
       const config = lc.configuration;
-      if (toSpawnCounter <= 0) break; // eslint-disable-line
```
lotas, line 119: lol, using ~30 chars to save on two brackets

lotas (approved):

PR #8768, opened 2026-06-15
Don't return values from callbacks that are expecting no return value

Eijebong:
These callbacks are passed to iterable methods (forEach, map, ...) which don't expect said callbacks to return any value. But because they're written as single expression arrow functions, they were returning implicitly which biome flags as suspicious. Easy but noisy solution, just wrap bodies in blocks.

Fix biome's useIterableCallbackReturn lint. This diff is better viewed when hiding whitespaces as git tries to be clever but is making it very hard to understand what actually happens...

### DIFF
--- a/changelog/Zk5dNq2HS6qC6O5idnxTsw.md
+++ b/changelog/Zk5dNq2HS6qC6O5idnxTsw.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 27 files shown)


lotas:
yeah, that's an arguable one, I think for me it's easier to understand when it's not creating a function/closure, less to type and to read, but if biome thinks it's better not to, we can keep it

lotas (approved):

PR #8769, opened 2026-06-15
Remove unused function arguments or mark them as unused with a _ prefix

Eijebong:
There's 4 big classes of fixes in there.

Anything related to `cfg` as a dependency in pretty much every service is vestigial from previous use/copy-paste between services. `cfg` was very often declared and then unused so I just removed it when unnecessary.
We were passing a mock object around a lot in tests and then not using it. I removed those.

The rest is a case by case. When I could trace all calls and the last argument of a function was unused, I removed it unless it was necessary to keep it for documentation and/or interface declaration purposes. The rest I prefixed with a `_`.

This was mostly done by going service by service, having biome autofix everything by prefixing a `_`, running tests and then reviewing them one by one on what could get removed vs what had to stay.

### DIFF
--- a/changelog/Ze9Iki6OTK-tQWLTZ0YkLA.md
+++ b/changelog/Ze9Iki6OTK-tQWLTZ0YkLA.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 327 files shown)


lotas (approved):
epic! thank you

PR #8781, opened 2026-06-16
Actually throw an error if a fake client is used in a prod environment

Eijebong:
The previous code was creating the error but never throwing it so clients in fake mode and production NODE_ENV would just happily continue working


petemoore (approved):
Nice catch!

PR #8788, opened 2026-06-19
Fix auth test in client-web not actually checking its expectation

Eijebong:
Because the `expect` line returns a Promise, in the previous form it was never resolved which means the test was never checking anything. Instead, the testing framework expects us to return the promise so it can await it itself before calling the test complete. This makes the test actually check something which revealed that the expectation was wrong.

### DIFF
--- a/changelog/C0lf15N9S0epOChdCYnUJA.md
+++ b/changelog/C0lf15N9S0epOChdCYnUJA.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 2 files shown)


petemoore (approved):
Wow, good find! Did you find it using a tool to search for code not returning promises, or did you discover it because the thing it was meant to protect was broken?

Eijebong:
I found it because while switching tests to vitest it started screaming at me

Issue #8113, opened 2025-11-20
Worker-manager should check for orphaned/dangling cloud resources


lotas:
At the moment, the worker manager's operational logic only checks for the resources it knows should exist. While the deprovisioning routine is designed to remove all associated infrastructure (instances, disks, IP addresses, NICs), intermittent cloud failures or race conditions can sometimes leave behind dangling or leaked resources. This leads to unnecessary cloud costs and resource exhaustion.

Worker-manager scanner loop should be extended to perform a resource reconciliation sweep:

* Query the cloud provider for all resources identifiable as being created by this worker manager (e.g., by matching specific resource tags or naming conventions).
* Compare this list against the resources actively tracked or known by the worker manager's current state.
* Identify all unaccounted resources (i.e., those found in the cloud but not tracked by the worker manager).
* Log these resources and/or trigger a cleanup routine to safely remove them.

PR #8790, opened 2026-06-23
feat(w-m): fast-track azure vm deletion when resources are cascade-deleted

### LINKED ISSUE #8161: Azure provider stopping capacity

lotas:
Since Azure supports fast resource removal with the VM (`deleteOption: 'Delete'`) and our ARM templates are enforcing this, we could save some time by not trying to deprovision each resource manually as before.

We also cannot rely/control what template is used, so we have to check the actual VM model and NIC response from Azure when worker registers. Only after checking that and making sure all resources are "cascadable" we could short-circuit deprovision loop.

Deprovision now becomes: 1. delete VM. 2. if cascade exists, mark all done.
This would skip NIC/IP/Disks steps

Additionally we log new metric and log type to see how it performs.

This should reduce total number of STOPPING workers that are usuallly blocking worker-scanner-azure.

Fixes #8161 - instead of excluding stopping capacity from estimator, we just do this

### DIFF
--- a/services/worker-manager/src/providers/azure/index.js
+++ b/services/worker-manager/src/providers/azure/index.js
@@ -58,6 +58,24 @@ const InstanceStates = {
 const failProvisioningStates = new Set(['Failed', 'Deleting', 'Canceled', 'Deallocating']);
 
 const DEPLOYMENT_METHOD_ARM = 'arm-template';
+const CASCADE_PROBE_TIMEOUT_MS = 30000;
+
+/**
+ * All identity comparisons normalize ARM-id casing
+ * to avoid VM/NIC model and deployment operations mismatch
+ * @param {String|Object} id
+ */
+const normalizeArmId = id => (typeof id === 'string' ? id.toLowerCase() : id);
+
+/** @param {Record<string, unknown>} resource } */
+const markResourceGone = resource => {
+  if (resource) {
+    resource.operation = undefined;
+    resource.id = false;
+    resource.deleted = true;
+  }
+};
+
 const UNKNOWN_METRIC_LABEL = 'unknown';
 const MAX_METRIC_LABEL_LENGTH = 200;
 const maxInstanceView404Streak = 2;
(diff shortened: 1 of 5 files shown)


services/worker-manager/src/providers/azure/index.js
```diff
@@ -2320,42 +2510,58 @@ export class AzureProvider extends Provider {
       if (!vmDeleted || worker.providerData.vm.id) {
         return;
       }
-      const nicDeleted = await this.deprovisionResource({
```
lotas, line 2323: this part stays as is, it is just moved under `if .. else { }` block

services/worker-manager/src/providers/azure/index.js
```diff
@@ -2320,42 +2510,58 @@ export class AzureProvider extends Provider {
...
-      for (let i = 0; i < worker.providerData.disks.length; i++) {
-        const success = await this.deprovisionResource({
+      // Cascade fast path: `deleteOption: 'Delete'` resources will be deleted with the VM.
+      // Skip the per-resource GET/delete walk entirely only when the tracked resources are
+      // exactly the VM-owned set, so a standalone disk/IP still gets deleted
+      const useFastPath = worker.providerData.cascade?.all === true && this.#cascadeTrackedMatches(worker);
+      const teardownMode = useFastPath ? 'fast' : 'slow';
+
+      if (useFastPath) {
+        // Mark the cascaded resource records deleted as we trust cascade delete option
```
petemoore, line 2521: thanks yarik! what happens if the deletes fail? is this a fire-and-forget or is there something that would later drop those resources if the deletes fail?

petemoore (commented):
Thanks for adding all the tests! This looks pretty solid, the only thing I'm not sure about is how clean up works if deletes fail...

services/worker-manager/src/providers/azure/index.js
lotas: so we do this check specifically to make sure that resources got "deleteOption: Delete" 
And if all do, then it's on the Azure's resource manager to clean them up

And indeed, once we confirmed that vm is deleted, it might be the case that resources are still "present" on azure side, but I don't think we should worry about this (not at least here)

We had a long standing - #8113 to cover that gap

petemoore (approved):
Thanks Yarik. This looks pretty robust, thanks for the extra tests, great job. :-)

PR #8791, opened 2026-06-23
Transfer ownership of mounts to the task on windows

### LINKED ISSUE #6561: [generic-worker] Change ownership of caches to current task
**Describe the bug**
As of Git 2.35.2, they introduced a security feature to mitigate a CVE on multi-user setups:
https://github.blog/2022-04-12-git-security-vulnerability-announced/#cve-2022-24765

Basically Git will error out if the checkout dir isn't owned by the current user. This impacts Firefox-CI because we commonly cache the checkout dir of the Github repo. So it will be owned by the first task user to check it out, and then the second task user will fail as it does not own the checkout.

**To Reproduce**

Here's an example task that hit this issue:
https://firefox-ci-tc.services.mozilla.com/tasks/cVeBuKGhQkGDqzsoz_utkA/runs/0/logs/public/logs/live.log

Relevant log:
```
[taskcluster 2023-09-18T18:20:35.232Z] Executing command 0: C:/mozilla-build/python3/python3.exe run-task --mozillavpn-checkout=./build/src/ --task-cwd build/src/windows/installer -- cmd.exe /c build_prod.cmd
D:\task_169506064099579>C:/mozilla-build/python3/python3.exe run-task --mozillavpn-checkout=./build/src/ --task-cwd build/src/windows/installer -- cmd.exe /c build_prod.cmd 
[setup 2023-09-18T18:20:35.398Z] run-task started in D:\task_169506064099579
[setup 2023-09-18T18:20:35.398Z] Invoked by command: --mozillavpn-checkout=./build/src/ --task-cwd build/src/windows/installer -- cmd.exe /c build_prod.cmd
[setup 2023-09-18T18:20:35.398Z] Python version: 3.9.10
[vcs 2023-09-18T18:20:35.401Z] WARNING: vcs checkout path (D:\task_169506064099579\build\src) not in cache or volume; performance will likely suffer
[vcs 2023-09-18T18:20:35.401Z] executing ['git', 'fetch', '--tags', '--force', 'https://github.com/mozilla-mobile/mozilla-vpn-client', 'refs/heads/main']
[vcs 2023-09-18T18:20:35.414Z] fatal: detected dubious ownership in repository at 'D:/task_169506064099579/build/src'
[vcs 2023-09-18T18:20:35.414Z] 'D:/task_169506064099579/build/src' is owned by:
[vcs 2023-09-18T18:20:35.414Z] 	'S-1-5-21-867665065-978626105-2054798079-1000'
[vcs 2023-09-18T18:20:35.414Z] but the current user is:
[vcs 2023-09-18T18:20:35.414Z] 	'S-1-5-21-867665065-978626105-2054798079-1001'
[vcs 2023-09-18T18:20:35.414Z] To add an exception for this directory, call:
[vcs 2023-09-18T18:20:35.414Z] 
[vcs 2023-09-18T18:20:35.414Z] 	git config --global --add safe.directory D:/task_169506064099579/build/src
```

**Expected behavior**

We worked around this by disabling the `safe.directory` feature as the error recommends:
https://github.com/taskcluster/taskgraph/pull/346

However the safety check seems like it could actually prove useful. I think `generic-worker` should handle updating cache ownership to the current user which would avoid this issue. Though workers that run more than one task at a time might prove to be a challenge :/

Eijebong:
The multiuser engine on windows was granting full access to mounts (yes, that `/grant:r` is to grant `(OI)(CI)F`, full access, by *replacing* the ACL for the given user, not read access... I don't want to talk about how much I got confused by this) but not ownership. This led to those mounts working fine despite being owned by the wrong user, but git didn't like that at all and it was complaining about dubious ownership.

So now we also use `icacls /setowner` to transfer ownership properly which mirrors what we do on POSIX platforms (`chown -R`). I left the full access grant because ownership doesn't imply that you'll have any right on the file (yay windows).

Note that the new `icacls` uses `/T` for ~~Trecursion~~ recursion, which the `/grant` one doesn't. The reason is that while ACLs are applied recursively by `(OI)(CI)`, ownership isn't and doesn't have any similar mechanism, so we have to pay the price of walking the tree. This is the same thing as our `chown -R` so hopefully it won't lead to big caches being really slow because of it...

Writing a test for this was a huge PITA. Two reasons for that, the test suite always reuses the same next-user, and the first cache creation is always done by said next-user.

I worked around this by running a task that does nothing but creates the cache entry, filling said cache from the testing code and giving everything to admin (via its SID, because windows is amazing, so the admin group name is localized and `Administrators` doesn't work for all locales). Then rerunning the task and checking that the cache ownership flip flops between Admin/next-user.

Fixes #6561

### DIFF
--- a/workers/generic-worker/mounts_multiuser_windows_test.go
+++ b/workers/generic-worker/mounts_multiuser_windows_test.go
@@ -1,0 +1,85 @@
+//go:build multiuser
+
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/mcuadros/go-defaults"
+	"github.com/taskcluster/taskcluster/v100/workers/generic-worker/host"
+)
+
+func ownerOf(t *testing.T, path string) string {
+	t.Helper()
+	out, err := host.Output("powershell", "-NoProfile", "-NonInteractive", "-Command",
+		"(Get-Acl -LiteralPath '"+path+"').Owner")
+	if err != nil {
+		t.Fatalf("could not read owner of %q: %v", path, err)
+	}
+	owner := strings.TrimSpace(out)
+	return owner[strings.LastIndex(owner, `\`)+1:]
+}
+
+func TestWritableDirectoryCacheReclaimsOwnership(t *testing.T) {
+	setup(t)
+
+	mounts := []MountEntry{
+		&WritableDirectoryCache{
+			CacheName: "banana-cache",
+			Directory: "ownership-cache",
+		},
+	}
+
+	payload := GenericWorkerPayload{
+		Mounts:     toMountArray(t, &mounts),
+		Command:    helloGoodbye(),
+		MaxRunTime: 180,
+	}
+	defaults.SetDefaults(&payload)
+
+	runTask := func() {
+		td := testTask(t)
+		td.Scopes = append(td.Scopes, "generic-worker:cache:banana-cache")
+		_ = submitAndAssert(t, td, payload, "completed", "completed")
+	}
+
+	cacheLocation := func() string {
+		entries := directoryCaches["banana-cache"]
+		if len(entries) == 0 {
+			t.Fatal("expected a persisted banana-cache entry after the task ran")
+		}
+		return entries[0].Location
+	}
+
+	runTask()
+	location := cacheLocation()
+
+	nested := filepath.Join(location, "sub", "file.txt")
+	if err := os.MkdirAll(filepath.Dir(nested), 0700); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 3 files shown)


workers/generic-worker/mounts_multiuser_windows_test.go
```diff
@@ -0,0 +1,85 @@
+//go:build multiuser && windows
```
petemoore: nit: the `&& windows` is already implied by the filename, which go natively honours.

petemoore (dismissed):
Nice job Bastien, also fighting the test framework, and adding a test. Nice to close the loop on this old issue!

workers/generic-worker/mounts_multiuser_windows_test.go
Eijebong: Oops, thanks

PR #8792, opened 2026-06-23
Switch client-web tests from karma to vitest/playwright

Eijebong:
Karma is deprecated so we needed to change the test runner anyway. Webpack is very outdated and would require a rewrite also. It's also very overkill for what we're doing here (we just need a way to polyfill a few things to run tests, we're not distributing anything). Vitest does everything at once by leveraging the vite plugin infrastructure which pulls in a lot less stuff than webpack.

We already had a playwright infrastructure for the UI which I reused, I just kept tests running on firefox as it's what was there before. (bit ironic that the UI tests are running on chromium).


petemoore (approved):
Thanks for the cleanup Bastien, this looks good to me!

PR #8794, opened 2026-06-23
Remove the decision task extra python requirements

Eijebong:
While looking at dependabot related stuff, I saw that it was complaining about not being able to build the dependency graph because of python. From what I gathered from public dependabot issues, it looks related to us having both a uv workspace and a "normal" one. Which made me look at the decision task dependencies. It's running python 3.11+, which means we can use the standard library tomllib instead of pulling an extra dependency and we can just remove the whole requirements.* + the dependabot config.


Eijebong:
Failure is unrelated (#8795)

petemoore (approved):

PR #8805, opened 2026-06-24
Remove our direct dependency on chai/chai-as-promised in client-web

Eijebong:
When I switched to vitest in #8792, I missed that vitest had a dependency on chai (another version from ours too...) and was exposing it in their public API through their `expect` wrappers. So I removed our direct dependency on chai so we can't get weird dependency resolution in the future.

Vitest also allows us to use their own `.resolves` and `.rejects` from `@vitest/expect` which means we can remove our dependency on `chai-as-promised` as well and remove that hacky `assertRejected`.

It's not as big a win as I'd like dependency count wise but I'll take it.

### DIFF
--- a/changelog/TZIkgUiFTNKdLSfUIIVOLg.md
+++ b/changelog/TZIkgUiFTNKdLSfUIIVOLg.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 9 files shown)


petemoore (approved):
Looks good, thanks!

PR #8806, opened 2026-06-24
Update query-string to 9 in client-web

Eijebong:
Query-string v8 switched to pure ESM, removing named imports from their entry point, only re-exporting their functions as `default` which forces us to import the module and then pick functions from it instead of being able to only get said functions on import.

Closes #8798


petemoore (approved):
Thanks Bastien!

PR #8807, opened 2026-06-24
Cleanup very outdated dependencies from client-web

Eijebong:
Used `yarn up -R node-gyp semver` which clears out a very large chunk of outdated transitive dependencies. This should clear out quite a few security issue false positives (this is all dev dependencies anyway).


petemoore (approved):
Awesome job!

PR #8808, opened 2026-06-24
Don't emit success github statuses when a task group still has pending tasks

### LINKED ISSUE #8009: Github check passed before task group completed
**Describe the bug**
I opened https://github.com/MozillaSecurity/orion/pull/699, which created a GH check backed by task group https://community-tc.services.mozilla.com/tasks/groups/FdeDG71mSz2woSfGvyDjIg to run CI tasks. 

That PR was set to auto-merge on success, and it did merge at [2025-10-09T19:33Z](https://github.com/MozillaSecurity/orion/pull/699#event-20192646436).  The [last task run](https://community-tc.services.mozilla.com/tasks/PebtPJ86RhufWDVRsUQSBw/runs/1) did not complete until 2025-10-09T19:40:26.186Z (~7 minutes later). 

**To Reproduce**
N/A

**Expected behavior**
Github checks should not pass until the task group is successful.

**Taskcluster version**
v90.0.0

**Additional context**
There were two possibly relevant Github service events today:
- https://www.githubstatus.com/incidents/k7bhmjkblcwp
- https://www.githubstatus.com/incidents/kk58nfytx0c7

It is probable this was caused by one of those, although we are outside their declared event window. Regardless, there is an opportunity to check if our GH interface could be more robust to mitigate during incidents like this.

Also, the final task in the above group did have a [failed run 0](https://community-tc.services.mozilla.com/tasks/PebtPJ86RhufWDVRsUQSBw/runs/0), and it failed because of a truncated artifact download. It's possible this is an unrelated network glitch.

Eijebong:
A PR using github statuses instead of checks could have its status from taskcluster reported as success before the task group would actually be completed.

The queue service emits a `task-group-resolved` event as soon as all tasks in the group have had a resolved run (and then every time a run resolves). The github service listens to that, and when it happens, grabs all tasks, checks for failures and computes the status from it. Except that if a task had been retried and was either pending or running, it would still mark the status as success. Instead, we now also check if there's any task in a non final state and downgrade the status from success to pending in that case.

While statuses are deprecated and we could just ignore the issue, the fix is simple enough and the implications are that a broken PR might get auto merged because a task got retried on any project using statuses instead of checks.

Fixes #8009

### DIFF
--- a/changelog/issue-8009.md
+++ b/changelog/issue-8009.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 8009
+---
+Fixed a bug where the github status API was marked resolved as `success` before any re-ran tasks were resolved
--- a/services/github/src/handlers/deprecatedStatus.js
+++ b/services/github/src/handlers/deprecatedStatus.js
@@ -37,18 +37,28 @@ export async function deprecatedStatusHandler(message) {
       usesChecks = true;
     }
 
+    let hasUnresolvedTask = false;
     const params = {};
     do {
       const group = await this.queueClient.listTaskGroup(message.payload.taskGroupId, params);
       params.continuationToken = group.continuationToken;
 
       for (let i = 0; i < group.tasks.length; i++) {
-        if (['failed', 'exception'].includes(group.tasks[i].status.state)) {
+        const taskState = group.tasks[i].status.state;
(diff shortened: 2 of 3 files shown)


petemoore (approved):
Thanks for doing this. Seems like low risk to accept since probably few people are using it. Nice job!

PR #8811, opened 2026-06-26
Update js-yaml to 4.2.0

Eijebong:
Version 4.2.0 of js-yaml fixed its block mapping parser to support tagged mapping keys (`!env FOO: bar` would now work for example when before it would've thrown on `!env FOO`). To achieve this, they're parsing nodes twice and throwing one of them away, which ends up calling our type `construct` twice. Since we were collecting potential env variable names as a side effect of the node construction to be able to return a listing of it for documentation purpose, every single env variable was inserted twice.

This whole thing does not spark any joy but I've tried a few things and nothing came up that I liked and this is the least invasive change that I could get to keep things working.

Namely, I tried making the constructor polymorphic and return a different value type with a unique `Symbol` as a key and then walking the final tree returned by js-yaml and collecting those items, but it ended up being quite complex *and* slightly wrong because it was ignoring keys (*technically* you could have `!env FOO` as a key name and it would've missed those). Because it was fragile, made the diff a lot bigger and was in my opinion as vile as the current code, I ended up scrapping it.

I also tried using the listener hooks from js-yaml but no luck there either, they're getting fired twice, would need their own dedupe and blergh....

So I went the "simple" way, by deduping values *after* they're inserted in the `vars`. Note that the dedupe has to be on the full descriptor (`type`, `name`, `flags`) so the same variable being used multiple times under different types survives twice. It also changes the behavior slightly by not emitting duplicates at all anymore, which isn't a problem, all dowstream consumers were already deduping things themselves, but I added a test to encode this behavior so we don't get surprised in the future.

I am not happy about this but I'm out of ideas.

Closes #8773

### DIFF
--- a/libraries/config/src/index.js
+++ b/libraries/config/src/index.js
@@ -9,6 +9,19 @@ import buildSchema from './schema.js';
 
 const __dirname = new URL('.', import.meta.url).pathname;
 
+const dedupeEnvVars = vars => {
+  const seen = new Set();
+  return vars.filter(({ type, var: name, optional, secret }) => {
+    // None of the fields can contain a | so this is safe to use as a key
+    const key = `${type}|${name}|${optional}|${secret}`;
+    if (seen.has(key)) {
+      return false;
+    }
+    seen.add(key);
+    return true;
+  });
+};
+
 const REPO_ROOT = path.join(__dirname, '../../../');
 
 const config = ({
@@ -54,7 +67,7 @@ const config = ({
 
     // If the user requested environment variables list, return it and stop
     if (envVars) {
-      return envVars;
+      return dedupeEnvVars(envVars);
     }
 
     // Add defaults to list of configurations if present
--- a/changelog/RwoJMxL5RkSAP1gEu3ZMQw.md
+++ b/changelog/RwoJMxL5RkSAP1gEu3ZMQw.md
@@ -1,0 +1,2 @@
+level: silent
+---
--- a/libraries/config/test/config_test.js
+++ b/libraries/config/test/config_test.js
@@ -176,6 +176,23 @@ suite(testing.suiteName(), () => {
     ]);
   });
 
+  test('load !env listing dedupes repeated descriptors', () => {
+    const vars = config({
+      serviceName: 'test',
+      files: [{ path: path.join(__dirname, 'test-env-dupes.yml'), required: true }],
+      env: {},
+      getEnvVars: true,
+    });
+
+    assume(vars).deep.equals([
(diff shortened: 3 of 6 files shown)


libraries/config/src/index.js
```diff
@@ -9,6 +9,18 @@ import buildSchema from './schema.js';
 
 const __dirname = new URL('.', import.meta.url).pathname;
 
+const dedupeEnvVars = vars => {
+  const seen = new Set();
+  return vars.filter(({ type, var: name, optional, secret }) => {
+    const key = `${type}|${name}|${optional}|${secret}`;
```
petemoore, line 16: Do we need to escape `${type}` or `${name}` or `${optional}` or `${secret}`? What I'm wondering - it looks like `Set.has(expr)` take a regexp or something similar, so this concatenated string looks like it might be at risk if the constituent parts aren't properly escaped (we had so many of these issues raised recently in security incidents). Or maybe those strings have no risk of holding special characters - but even then, explicitly escaping them even if it is a no-op helps future-proof and demonstrate safety. Or if they can't carry dangerous symbols we could just add a comment to say so.

petemoore (commented):
Thanks Bastien! Sounds like you had to go through a lot of pain, and on a very hot day! I'm sorry to hear that.

libraries/config/src/index.js
Eijebong: No? Unless Set is a magic thing that isn't a set somehow (would barely be surprised at this point). https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has 

The only collision that would be possible is values had | in them, and that's not possible here. We could stringify the array instead if you feel like that's better but short of being able to hash tuples (which unfortunately don't exist), we can't do much better AFAIK.

petemoore: I guess optional and secret are limited to true or false, env var name can't contain a pipe, and the type can't either. JSON.stringify would work, but maybe it is overkill. It just jumped out at me that it would break if any of the vars included a pipe. maybe just an explicit comment like `none of these vars can include | symbol, so safe as a separator` or something like that?

petemoore (dismissed):
Thanks Bastien! It looks very good.

My inline comment is a suggestion, if you disagree, you can leave it as it is. Being explicit that something _isn't_ a problem could save someone the headspace of having to think through it and reach the same conclusion.

libraries/config/src/index.js
Eijebong: Yeah I'll just add a comment then 👍

Issue #4959, opened 2021-08-11
Python client tests not running in CI?


petemoore:
In #4953 I discovered that `clients/client-py/Makefile` doesn't seem to be called by CI, since it refers to a file that doesn't exist (`devDep.py`).

When running locally, some `test_upload_download_real.py` tests fail, probably because they require valid taskcluster credentials with particular assigned scopes to run. Perhaps these tests were intended run against the original Object Service that was later replaced by a completely different Object Service, I'm not sure.

In any case, we should investigate if some/all of these tests should run in CI and integrate them, and remove tests that are no longer applicable, or at least try to find out why they are not part of the current CI. Also we should find out what the `devDep.py` script was for.

PR #8812, opened 2026-06-26
Remove old scripts/makefile from client-py and prune useless dependencies

Eijebong:
These are not necessary when they're as much typing as the equivalent `uv` command. They're also inconsistent (lint.sh puts linting/formatting as one command, but the makefile splits them). I would also prefer never to see a `curl | bash` ever again in my life. IMO these provide no extra value and are just footguns as proven by https://github.com/taskcluster/taskcluster/issues/4959 and the fact that CI is essentially running `uv run uv run ruff` right now...

I kept `release.sh` because that has actual logic. For the rest, just call `uv run ruff check` and/or `uv run ruff format` instead of `./lint.sh` or `make lint`/`make format`.

Also cleaned up dependencies because `make docs` was the only consumer for `grip` (and was not useful), and `flake8` had escaped pruning from #8021


petemoore (approved):
Nice cleanup, thanks!

PR #8813, opened 2026-06-26
Fix the `pg` require meta check

### LINKED ISSUE #8761: node-pg incorrect check
> hm. this is passed to grep. i wonder what it would match `require(['"]pg)` - but there's no closing one :) 
> 
> plus, I forgot to update this one when I did switch to esmodules 
> 
> I checked following locally: 
> ```suggestion
>     for (const pattern of ['require(.pg.)', 'import.* from .pg.;', '_withClient']) {
> ``` 

 _Originally posted by @lotas in [#8749](https://github.com/taskcluster/taskcluster/pull/8749/changes#r3412574626)_

Two fixes:
- require(pg) without closing quotes
- missed import check when switched to esmodules

Eijebong:
There were two issues there. `require("pgfoo")` would've matched and it wasn't checking anything anymore since the switch to esmodules.

Fixes #8761

### DIFF
--- a/changelog/issue-8761.md
+++ b/changelog/issue-8761.md
@@ -1,0 +1,3 @@
+level: silent
+reference: issue 8761
+---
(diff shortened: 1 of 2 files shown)


petemoore (approved):
Awesome, thanks Bastien!

PR #8816, opened 2026-06-26
Worker-manager scanner concurrency

### LINKED ISSUE #8815: Add bounded concurrency to (Azure) worker scanner
The (Azure) worker scanner currently checks non-stopped workers sequentially, so loop duration scales almost directly with worker count. The timing report shows loop duration correlates strongly with scanned workers (r = +0.952), and slow loops are driven by thousands of sequential Azure calls rather than slow per-call latency.

As a first-order improvement, make the (Azure) scan loop process workers with bounded, configurable concurrency, for example 10-20 checkWorker calls in flight. This should reduce provisioning-surge scan loops from tens of minutes to a few minutes while keeping Azure API pressure controlled by the existing rate limiter.

We should roll this out conservatively and monitor `cloud-api-paused`, Azure failures/retries, and scanner loop duration, since scanner, provisioner, and web processes share the same Azure subscription budget.

| workers scanned | loops | median loop time | median time / worker |
|---|---|---|---|
| 0–300 | 58 | **64 s** | 540 ms |
| 300–700 | 26 | **382 s** | 925 ms |
| 700–1200 | 11 | **1,719 s** | 2,067 ms |
| 1,200+ | 12 | **2,986 s** | 1,884 ms |

lotas:
First commit just extracts the logic of a single worker check, second puts it into concurrency mode

Metrics and logs show that there is a strong correlation between slow
scan loops and the number of workers. So the only way to lower those
times is to make requests in parallel.

Rate limiting will like not be the limiting factor here, as current
usage shows to be staying below those.

2 is chose as a safe default which should gives plenty of headroom for
all cloud providers with likely very modest improvements, but it would
allow to experiment on the live system with bigger than that numbers, to
adjust to the optimal worker count/load to make scanner loops much
faster

Fixes #8815


services/worker-manager/src/worker-scanner.js
```diff
@@ -74,67 +84,90 @@ export class WorkerScanner {
...
-        // Only RUNNING, non-quarantined workers are candidates for termination decisions
-        const isQuarantined = worker.quarantineUntil && worker.quarantineUntil > new Date();
-        if (worker.state === Worker.states.RUNNING && !isQuarantined) {
-          if (!poolCandidates.has(poolId)) {
-            poolCandidates.set(poolId, []);
-          }
-          poolCandidates.get(poolId).push(worker);
-        }
+  /** @param {Worker} worker */
+  async #checkOneWorker(worker) {
```
lotas, line 121: this was extracted in a first commit from the paginatedIterator loop, insides are identical except for `poolCandidates` being moved to `this.poolCandidates`

services/worker-manager/src/worker-scanner.js
```diff
@@ -74,67 +84,90 @@ export class WorkerScanner {
...
+
+    try {
+      for await (const row of paginatedIterator({
+        fetch,
+        indexColumns: ['worker_pool_id', 'worker_group', 'worker_id'],
+        size: 500,
+      })) {
+        launch(Worker.fromDb(row));
+        if (inFlight.size >= this.concurrency) {
+          await Promise.race(inFlight);
```
lotas, line 108: that's the whole concurrency here - we check if inflight size is at max and wait for the first task to finish before adding a new one

services/worker-manager/test/worker_scanner_test.js
```diff
@@ -299,6 +299,204 @@ helper.secrets.mockSuite(testing.suiteName(), [], (mock, skipping) => {
...
+      },
+    };
+    const fakeProviders = {
+      forAll: async fn => {
+        await fn(fakeProvider);
+      },
+      get: id => (id === 'testing1' ? fakeProvider : undefined),
+    };
+
+    const scanner = new (await import('../src/worker-scanner.js')).WorkerScanner({
```
Eijebong: Any reason to not import this normally?

services/worker-manager/src/worker-scanner.js
```diff
@@ -80,61 +79,65 @@ export class WorkerScanner {
...
+        return;
+      }
+      try {
+        await withTimeout(
+          provider.checkWorker({ worker }),
+          60_000, // 1 minute
+          `checkWorker timed out for ${worker.workerPoolId}/${worker.workerId}`
         );
+      } catch (err) {
+        this.monitor.reportError(err); // Just report it and move on so this doesn't block other providers
```
Eijebong: nit: that comment is wrong now

services/worker-manager/src/worker-scanner.js
Eijebong: Yeah, commit 1 should be reviewed without whitespace changes, makes it much easier to see what's going on

services/worker-manager/test/worker_scanner_test.js
lotas: indeed. thanks, will import once

Eijebong (dismissed):

services/worker-manager/src/worker-scanner.js
lotas: whoopsie. thanks

Eijebong (approved):

PR #8817, opened 2026-06-29
Fix the octokit secondary rate limiter by telling it to do something

### LINKED ISSUE #7773: The github worker throttling does not seem to work well enough
https://github.com/Eijebong/Archipelago-index/commit/928b064038099c99187ffd953054fc17f66cefdb

This was pushed when my instance was on 84.1.1

Looks like it detects those but doesn't throttle on it?

```
{"EnvVersion":"2.0","Fields":{"message":"POST /repos/Eijebong/Archipelago-index/check-runs - 403 in 139ms"},"Hostname":"taskcluster-github-worker-c747b7b7c-6hscx","Logger":"taskcluster.github.octokit","Pid":1,"Severity":6,"Timestamp":1749566856289000000,"Type":"monitor.generic","message":"POST /repos/Eijebong/Archipelago-index/check-runs - 403 in 139ms","serviceContext":{"service":"github","version":"84.1.1"},"severity":"INFO"}
{"EnvVersion":"2.0","Fields":{"message":"SecondaryRateLimit detected for request POST /repos/{owner}/{repo}/check-runs"},"Hostname":"taskcluster-github-worker-c747b7b7c-6hscx","Logger":"taskcluster.github.octokit","Pid":1,"Severity":4,"Timestamp":1749566856291000000,"Type":"monitor.generic","message":"SecondaryRateLimit detected for request POST /repos/{owner}/{repo}/check-runs","serviceContext":{"service":"github","version":"84.1.1"},"severity":"WARNING"}
{"EnvVersion":"2.0","Fields":{"checkRunStatus":"queued","debugId":"id-1679","eventId":null,"event_id":"b9bc25fa-4609-11f0-8871-018d86653b41","installationId":null,"installation_id":52832514,"message":"creating comment on Eijebong/Archipelago-index@928b064038099c99187ffd953054fc17f66cefdb","owner":"Eijebong","repo":"Archipelago-index","sha":"928b064038099c99187ffd953054fc17f66cefdb","taskGroupId":"C0kjiXDqSaW79bHy-elsCg","taskId":"WUvs1kQKQFmbViMX2Ks8kg","v":1},"Hostname":"taskcluster-github-worker-c747b7b7c-6hscx","Logger":"taskcluster.github.handlers","Pid":1,"Severity":7,"Timestamp":1749566856292000000,"Type":"handler-debug","message":"creating comment on Eijebong/Archipelago-index@928b064038099c99187ffd953054fc17f66cefdb","serviceContext":{"service":"github","version":"84.1.1"},"severity":"DEBUG"}
{"EnvVersion":"2.0","Fields":{"message":"request"},"Hostname":"taskcluster-github-worker-c747b7b7c-6hscx","Logger":"taskcluster.github.octokit","Pid":1,"Severity":7,"Timestamp":1749566856299000000,"Type":"monitor.generic","message":"request","serviceContext":{"service":"github","version":"84.1.1"},"severity":"DEBUG"}
```

Eijebong:
When I added back the rate limiter in #7732, I mostly copied the example from their readme because it matched the original implementation which was also wrong (and probably got backed out because of the exact same bug).

It was still a lot better because it forced *some* delay between requests and was still detecting primary rate limits and backing off on that. But not returning true from `onSecondaryRateLimit` means this was doing **nothing** except logging that we hit the limit and then going through and throwing, which lead to the service putting a comment on the commit/PR, which helped exhaust the app write quota which means the next write would hit the limit again and yeah... Not great.

I chose to not add an upper bound on the number of retries because secondary rate limits should be very short lived and resolve on their own and we'd rather not lose events going towards github as those might leave checks in a stuck state which might prevent merging PRs for example. If it becomes an issue, we should review the architecture and stop ACK-ing pulse messages *before* calling github's API.

(If only there was some way to encode that this function needed to make a choice and return a value... We could call it a type system for example...)

Fixes #7773

### DIFF
--- a/changelog/issue-7773.md
+++ b/changelog/issue-7773.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: issue 7773
+---
+The github service will now respect secondary rate limits and retry those
+requests instead of just commenting that they failed.
--- a/services/github/src/github-auth.js
+++ b/services/github/src/github-auth.js
@@ -74,6 +74,7 @@ export default async ({ cfg, monitor }) => {
       },
       onSecondaryRateLimit: (_retryAfter, options, octokit) => {
         octokit.log.warn(`SecondaryRateLimit detected for request ${options.method} ${options.url}`);
+        return true;
       },
     },
     retry: {
--- a/services/github/test/github-auth_test.js
+++ b/services/github/test/github-auth_test.js
@@ -56,7 +56,36 @@ suite(testing.suiteName(), () => {
       await gh.getAppGithub();
       await gh.getInstallationGithub(100);
     });
+
+    test('secondary rate limits get retried instead of throwing', async () => {
+      nock('https://api.github.com:443')
+        .get('/app')
+        .reply(403, { message: 'secondary rate limit' }, { 'retry-after': '0.001' })
+        .get('/app')
+        .reply(200, { id: 12345 });
+
+      const monitor = await helper.load('monitor');
+      const gh = await githubAuth({
+        monitor,
+        cfg: {
+          github: {
+            credentials: {
+              appId: 12345,
+              privatePEM: FAKE_KEY,
+            },
+          },
+        },
+      });
+      const ghApp = await gh.getAppGithub();
+      const res = await ghApp.apps.getAuthenticated();
+
+      assert.equal(res.status, 200);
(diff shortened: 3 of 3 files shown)


lotas (approved):
nice catch! thanks

PR #8818, opened 2026-06-30
Make client-web a conformant ESM module

Eijebong:
In today's things I really didn't want to have to learn, while playing around with moving UI tests to vitest (as a continuation/rework of #8580), I discovered that the client-web package was using ESM syntax (`import`/`export`) but was very much not a valid ESM module. It was just working so far because webpack decided that it should work somehow.

According to the node documentation:

- ES module relative specifiers require a file extension: "A file   extension must be provided when using the `import` keyword to resolve relative or absolute specifiers"[1] and "the file extension is always necessary for these"[2]
- A package using ES module syntax should declare its module system via the "type" field rather than relying on Node's ambiguous CommonJS-first syntax-detection fallback. [3] [4]
- Entry points should be defined via "exports" [5]

This PR fixes all of these and updates the import in the UI (the only consumer here). Note that the UI will still be using the old package since it uses the released version but the change is necessary for newer versions as this is a breaking change for client-web.

Changes are very boring, this is mostly "add js extension to file, update imports to use the new path". The one interesting thing is the `crypto-js` that I had to update because it's a CJS module and ES modules can't import named exports from it.

[1]: https://nodejs.org/api/esm.html#mandatory-file-extensions
[2]: https://nodejs.org/api/esm.html#terminology
[3]: https://nodejs.org/api/packages.html#type
[4]: https://nodejs.org/api/packages.html#determining-module-system
[5]: https://nodejs.org/api/packages.html#package-entry-points


lotas (approved):
awesome, thanks! I should have done it when I migrated the rest of the base to ESM

PR #8819, opened 2026-06-30
Fix a few things zizmor is unhappy about

Eijebong:
Some sha not matching the version in the comment on the same line (the horror :scream:). And fix one false positive in the release workflow that could potentially have been an issue had we decided to install a package in said workflow because they're enabling package caches by default in `setup-node`. In practice even that would probably have been impossible but eh, costs nothing to disable the cache and silence a warning.


Eijebong:
They should try and make these comments less useful... 

Turns out you can run zizmor locally and there it's actually useful and even has an autofix...

lotas (approved):
nice, thanks

PR #8827, opened 2026-07-01
Relax the node version requirement for `taskcluster-client` to `^major`

### LINKED ISSUE #7331: Client node version management
Problem:

[Dependabot PR]((https://github.com/taskcluster/fast-azure-storage/pull/201/checks?check_run_id=31574594546)) in a different project failed because tc client needs a newer Node version, but the repo was still on Node 20.16.

Tried updating .tc.yml to use the latest Node, but that caused another issue when the repo was moved to a newer version of Node: [Check run failure](https://github.com/taskcluster/fast-azure-storage/pull/202/checks?check_run_id=31605603125).

Suggestions:

Maybe we don’t need strict minor versioning for Node. We could enforce `>=LTS.0.0` and only increase the minimum minor version if there’s a breaking change.

I’ve noticed that I always have to manually pull new versions locally after a Node upgrade. We could add a git hook to make sure the latest version is used automatically.

### LINKED ISSUE #8660: taskcluster-client (npm): engines.node uses exact version instead of semver range since 55.3.3

Eijebong:
Users of our published libraries are currently stuck on the exact same version as the one we use for development which makes it painful to update as they have to keep their node version in sync with us. This relaxes the requirement to only require the same major version as the one we're using, trusting node to not introduce breaking changes during a major cycle.

I didn't relax the requirement further because we already have proof that node breaking changes will break this client (current tests don't pass with node 26).

Note that right now I only relaxed the requirement for taskcluster-client. We also publish client-web, but that one doesn't currently have an engine bound so I didn't add one. Other packages aren't published so they're unaffected.

Fixes #7331
Fixes #8660
Closes #8691

### DIFF
--- a/changelog/issue-8660.md
+++ b/changelog/issue-8660.md
@@ -1,0 +1,5 @@
+audience: users
+level: minor
+reference: issue 8660
+---
+Relax the required node version for `taskcluster-client` from `24.17.0` to `^24`
--- a/clients/client/package.json
+++ b/clients/client/package.json
@@ -23,7 +23,7 @@
     "nock": "^14.0.13"
   },
   "engines": {
-    "node": "24.17.0"
+    "node": "^24"
   },
   "type": "module",
   "files": [
--- a/infrastructure/tooling/src/generate/generators/node-version.js
+++ b/infrastructure/tooling/src/generate/generators/node-version.js
@@ -47,13 +47,20 @@ tasks.push({
       contents.replace(/^( *NODE_VERSION *= *")[0-9.]+(")$/m, `$1${nodeVersion}$2`)
     );
 
-    for (const file of ['ui/package.json', 'clients/client/package.json', 'clients/client-test/package.json']) {
+    for (const file of ['ui/package.json', 'clients/client-test/package.json']) {
       utils.status({ message: file });
       await modifyRepoJSON(file, contents => {
         contents.engines.node = nodeVersion;
         return contents;
       });
     }
+
+    // Relax the engine requirement for published packages
+    utils.status({ message: 'clients/client/package.json' });
+    await modifyRepoJSON('clients/client/package.json', contents => {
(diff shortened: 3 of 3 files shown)


petemoore (approved):
LGTM, thanks!

PR #8828, opened 2026-07-01
Migrate the biome config to 2.5.1

Eijebong:
When running biome 2.5.1 with our config it complains quite loudly about `recommended: true` being deprecated. Just ran `biome migrate --write` which fixed it by using the new way of signaling that.


petemoore (approved):
LGTM, thanks!

PR #8836, opened 2026-07-01
Drop `application/graphql` support as a valid content type for web-server

Eijebong:
We were supporting that content type through `body-parser-graphql` which has not seen any update in 8 years and is now making updating `body-parser` itself in web-server impossible as it expects body-parser 1, not 2. The API changed slightly and it ends up double consuming the body which doesn't work very well as one can expect.

Since `application/graphql` isn't even mentioned in the graphql over HTTP spec[1] and nothing in this repo uses it, we can just drop support for it.

The 1mb limit that was set on the body parser was dead code as `bodyParser.json()` was running first without respecting that limit (it has its own limit set to 100kB).

[1]: https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md

### DIFF
--- a/changelog/BqX5LHLxQtaT4JhK_jUhuw.md
+++ b/changelog/BqX5LHLxQtaT4JhK_jUhuw.md
@@ -1,0 +1,6 @@
+audience: users
+level: major
+---
+Drop support for `application/graphql` as the request content type on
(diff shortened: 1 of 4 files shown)


lotas (approved):
very good

petemoore (approved):
WFM, thanks!

PR #8837, opened 2026-07-02
Update taskcluster/client-web to 101.0.0

Eijebong:
The migration job was already done with the breaking change in 7c8bc735ef4c158d98e456b8c5334099f9f53b47


Eijebong:
I hate it here... qs 0.9 uses `??` and `&&=` which webpack 4 doesn't understand. So this'll go into the "yeet webpack into the sun" PR

PR #8838, opened 2026-07-02
Convert some commonJS to ESM in the UI

Eijebong:
`exports.foo` -> `export const foo` and `require` -> `import`.

Found those while looking at migrating to vite and they can land on their own as webpack 4 is perfectly happy with them.

There are a few more CJS things that need to be converted but I cannot (namely a `require.context` and a few `require()`) but I can't do that as fixing them breaks or can't be done while keeping feature parity under webpack 4.


lotas (approved):
ty

PR #8839, opened 2026-07-03
feat(w-m): Remove stale check and change level to notice for transient gcp errors

lotas:
403 is already handled by the previous check.
To avoid logging all 5xx from GCP as warning, it is now a notice for a first try. Request would be retried in 1s

### DIFF
--- a/services/worker-manager/test/provider_google_test.js
+++ b/services/worker-manager/test/provider_google_test.js
@@ -904,4 +904,41 @@ helper.secrets.mockSuite(testing.suiteName(), [], (mock, skipping) => {
       helper.assertPulseMessage('worker-running', m => m.payload.launchConfigId === worker.launchConfigId);
     });
   });
+
+  suite('errorHandler', () => {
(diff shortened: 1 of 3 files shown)


services/worker-manager/test/provider_google_test.js
```diff
@@ -904,4 +904,42 @@ helper.secrets.mockSuite(testing.suiteName(), [], (mock, skipping) => {
...
+      err.code = 403;
+      const result = provider.cloudApi.errorHandler({ err, tries: 0 });
+      assert.equal(result.reason, 'rateLimit');
+      assert.equal(result.backoff, 50); // 1 * 50
+      assert.equal(result.level, 'notice');
+    });
+
+    test('non-4xx/5xx errors are rethrown', () => {
+      const err = new Error('nope');
+      err.code = 404;
```
Eijebong: `non-4xx` eh?

lotas: eagle eye :D oops

Eijebong (approved):

PR #8841, opened 2026-07-06
feat(w-m): Handle checkWorker timeouts as a warning not as an error

### LINKED ISSUE #8664: Investigate checkWorker Timeouts - https://mozilla.sentry.io/issues/7417554128/?project=6459385
Errors like `checkWorker timed out for proj-fuzzing/linux-pool105/626068830237694292`

Seem to be equally split between azure non azure. 
Maybe it's helping overall, maybe not. Needs deep dive and  investigation of particular workers 

https://mozilla.sentry.io/issues/7417554128/?project=6459385

Similar happening ~20/day for azure
https://mozilla.sentry.io/issues/7417587033/

lotas:
Most of the failures happen in bursts and logs show that it's happening on the cloud provider side where instance.get or instanceView calls hang or take longer than the 60s timeout.

We emit those as a warning instead of error, since worker will be retried on the next run.

Additionally, we now pass the abort signal to also abort requests that might have stayed pending after we abandon them in Promise.race()

Fixes #8664

### DIFF
--- a/services/worker-manager/src/worker-scanner.js
+++ b/services/worker-manager/src/worker-scanner.js
@@ -2,7 +2,7 @@ import Iterate from '@taskcluster/lib-iterate';
 import taskcluster from '@taskcluster/client';
 import { paginatedIterator } from '@taskcluster/lib-postgres';
 import { Worker, WorkerPool } from './data.js';
-import { withTimeout } from './util.js';
+import { withTimeout, TimeoutError } from './util.js';
 
 /**
  * Make sure that we visit each worker relatively frequently to update its state
@@ -19,6 +19,7 @@ export class WorkerScanner {
     providersFilter = {},
     estimator,
     concurrency = 1,
+    checkWorkerTimeoutMs = 60_000,
   }) {
     this.WorkerPool = WorkerPool;
     this.providers = providers;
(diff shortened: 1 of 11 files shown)


services/worker-manager/src/worker-scanner.js
```diff
@@ -19,13 +19,15 @@ export class WorkerScanner {
     providersFilter = {},
     estimator,
     concurrency = 1,
+    checkWorkerTimeoutMs = 60_000,
```
Eijebong, line 22: Guessing this is here for tests only as it's not wired up to the config?

Eijebong (approved):
Really fun stuff. I find it a bit weird that we're not passing the abortSignal down to every single call but I guess a `get` is a much more frequent operation than any of the others which is why it's the one standing out.

services/worker-manager/src/worker-scanner.js
lotas: yes, this is for tests only. Wasn't sure if it's worth placing into a service config or not

Eijebong: I have no opinion on that. Just wanted to call it out in case it was something you missed

lotas: alright, thanks! yes. I'll merge it as is then

lotas:
> Really fun stuff. I find it a bit weird that we're not passing the abortSignal down to every single call but I guess a `get` is a much more frequent operation than any of the others which is why it's the one standing out.

yes, I think passing it to mutating calls would not be a great idea.. but those are cheap and those are the ones that hang most likely, so if we abort the check for a given worker, request should also be stopped

PR #8845, opened 2026-07-08
Bug 2053178 - Make sure v1 github policy is respected when a fork uses v0

Eijebong:
The scopes list for a rendered TC YML comes from the full rendering for a given version of a taskcluster.yml, which in the case of a PR, is whatever is at the PR HEAD. So a fork with a v0 YAML at their HEAD will unconditionally get v0 scopes which includes `:pull-request` which is the trusted version, ignoring the v1 policy upstream despite the PR correctly being considered untrusted and tasks_for being set to `PULL_REQUEST_UNTRUSTED`.

The fix is fairly easy although it feels a bit dirty. We have `tasks_for` in the payload (which, while not exposed to the user in v0, is still present). That field already contains the trust level decision made against the base repo's policy, it just isn't respected by v0 because it didn't have that concept of trusted PRs. So now if we have a tasks_for that is untrusted, we honor it even with v0, and give the untrusted scope instead of reverting to the `:pull-request` one.

### DIFF
--- a/changelog/cG4RCE2BT9ibjBCNQmTM_g.md
+++ b/changelog/cG4RCE2BT9ibjBCNQmTM_g.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: bug 2053178
+---
+Fix a way for PRs from forks to bypass the `public_restricted` policy by
+declaring `version: 0` in their own taskcluster.yml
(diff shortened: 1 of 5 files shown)


lotas (dismissed):

petemoore (dismissed):
Thanks!

lotas (approved):

petemoore (approved):

PR #8846, opened 2026-07-08
fix(changelog): don't assume PR is running against taskcluster/taskcluster

petemoore:
While working on integrating https://github.com/mozilla/blender with taskcluster, I set up a staging taskcluster repo at https://github.com/petemoore/taskcluster to test the dependabot approval process before we applied it to the production repo. In the process, I discovered that we check which files changed in a PR to decide whether a changelog is necessary. Since we perform a shallow clone (depth 20) we may not have enough information to locally determine what the full set of changed files is in the PR. Therefore we use octokit to query the PR directly to find the list of changed files (rather than computing locally). The flow is basically, find out what PR we are on, and then query github for the list of changed files in _that_ PR. However, the code was assuming that the PR was raised against github.com/taskcluster/taskcluster. This does not work when you have a staged repo. It is also a little fragile if orgs or repos get renamed, or a third party forks taskcluster as the basis for their own project etc.

So this PR looks up the real owner and repo names, and logs them too. We should see when the CI runs for this PR, that it reports that it looks up the diff for the exact PR that gets created when I submit this pull request! To be extra safe, I also created a local PR to my fork, which correctly identified the PR (`Checking changelog requirement for petemoore/taskcluster#392`) in https://github.com/petemoore/taskcluster/pull/392/checks?check_run_id=85878636336.

### DIFF
--- a/infrastructure/tooling/src/changelog/index.js
+++ b/infrastructure/tooling/src/changelog/index.js
@@ -236,14 +236,19 @@ export class ChangeLog {
   }
 }
 
-const check_pr = async pr => {
+const check_pr = async prUrl => {
   const octokit = new Octokit();
-  const options = octokit.pulls.listFiles.endpoint.merge({
-    owner: 'taskcluster',
-    repo: 'taskcluster',
-    pull_number: pr,
-  });
-  const files = await octokit.paginate(options, response => response.data.map(({ filename }) => filename));
+  if (!prUrl) {
+    throw new Error(
+      'Cannot check the changelog: --pull-request-url is unset. This check should only run in github pull requests.'
+    );
+  }
+  console.log(`Checking changelog requirement for ${prUrl}`);
+  const files = await octokit.paginate(`GET ${prUrl}/files`, response => response.data.map(({ filename }) => filename));
+  if (!files.length) {
+    throw new Error(`no changed files found for ${prUrl}; the pull request lookup returned nothing`);
+  }
+  console.log(`changed files (${files.length}):\n${files.map(f => `  ${f}`).join('\n')}`);
 
   // files that do not require a changelog entry if they are the only thing changed.  This
   // is similar to the list in .gitattributes., along with yarn and package.json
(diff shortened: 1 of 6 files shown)


petemoore:
Seems to work!

https://github.com/taskcluster/taskcluster/pull/8846/checks?check_run_id=85881627975 shows:

```
Checking changelog requirement for taskcluster/taskcluster#8846
```

infrastructure/tooling/src/changelog/index.js
```diff
@@ -238,9 +238,19 @@ export class ChangeLog {
 
 const check_pr = async pr => {
   const octokit = new Octokit();
+  // matches both https and ssh git urls
+  const repoMatch = (process.env.GITHUB_BASE_REPO_URL || '').match(/github\.com[/:]([^/]+)\/([^/]+?)(?:\.git)?\/?$/);
```
lotas: I'm a bit confused here - you are using process env? 
this is not part of the config, and the only other mention is using task payload:
https://github.com/lotas/taskcluster/blob/59058d24c64c4f8509884b077ba938c3664ad009/services/github/src/tc-yaml.js#L76 

was it supposed to be coming from pr payload?

lotas: ah, I see it's injected in taskgraph transformer

lotas: all good, I thought it's the server side code, but it's a tooling that runs in the tasks. all good

lotas (dismissed):

infrastructure/tooling/src/changelog/index.js
petemoore: good catch, i can pass this through the decision task, like we do with the PR number

petemoore:
Output still looks good:

```
Checking changelog requirement for https://api.github.com/repos/taskcluster/taskcluster/pulls/8846
changed files (6):
  .taskcluster.yml
  changelog/Sqfg5YYQTi-f0APJmNSUSQ.md
  infrastructure/tooling/src/changelog/index.js
  infrastructure/tooling/src/main.js
  taskcluster/kinds/meta/kind.yml
  taskcluster/src/transforms/__init__.py
PR https://api.github.com/repos/taskcluster/taskcluster/pulls/8846 OK
```

lotas (approved):
lgtm

PR #8847, opened 2026-07-08
Move material-ui-json-schema-viewer to this repository

Eijebong:
We haven't been maintaining it as a dependency for years, if there are external users of it, I sure hope that they forked it and updated dependencies... In any case, this dep is currently blocking the react update in this repository and I don't see any value in contorting ourselves to keep it externally, so move it in here. After this merges, we can archive the original repository.

The code in `ui/src/components/JsonSchemaViewer/` is exactly what was in the original repo, minus the theme.js file which is not used and one biome annotation.

Original repository is at
https://github.com/taskcluster/material-ui-json-schema-viewer/

### DIFF
--- a/changelog/axQ8AobzQPuAl5tPW-P0fQ.md
+++ b/changelog/axQ8AobzQPuAl5tPW-P0fQ.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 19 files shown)


lotas (approved):
yay, -1 dependency

petemoore (approved):
Sweet! This is a great project organisation simplification. Thanks Bastien!

PR #8850, opened 2026-07-08
Remove the apollo-cache-persist dependency

Eijebong:
This was supposed to add persistence to the cache so it survives hard refreshes (not navigation, that's a different cache layer which we still have). Except that it never actually called `restore()` making it a no-op. Well not exactly a no-op because the write side of it has been working fine so we've been caching query results into users indexdb for years without ever using them.

Here's the funnier part here and why I'm removing this dependency instead of calling restore. Every single query I checked that might maybe have benefited from this has a `fetchPolicy` set to `network-only` which means it wouldn't even look at the cache in the first place...

Add to that that we want to get rid of graphql, I ain't even going to think about it, it's been broken for years, into the trash it goes :put_litter_in_its_place:

### DIFF
--- a/changelog/VCOiLvopRs2lZwMw5Tp0iQ.md
+++ b/changelog/VCOiLvopRs2lZwMw5Tp0iQ.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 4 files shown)


lotas (approved):
wonderful!

PR #8851, opened 2026-07-09
Update to apollo/client v3

Eijebong:
Most of this is very mechanical renaming of imports from their own package to the `@apollo/client` one.

The mocking also needed some changes, apollo-client-mock doesn't support v3 and switching to the alternative (`MockProvider`) requires rewriting test snapshots and would make reviewing this a lot more painful. Since there's currently talks about removing graphql entirely, I chose to take the easy route here and basically replicate what the package was doing by using `graphql-tools/mock` and `graphql-tools/schema` directly.

The one breaking change from v3 we hit is that the cache freezes values, and we were modifying some of those by mutating them in place with `.sort()`. We now make a copy before in those 2 places.

The final piece in here is the change for possible fragments, also used for the caching. The old way was taking the whole schema, the new one only needs possible types so I simplified and renamed the script/file. I'm also fairly confident that this will help because that file had not been regenerated in quite some time and the diff between what I extracted from the current fragments and when I actually regenerated it was quite substantial.

One thing of note, we still have some graphql ws related dependencies, I did not touch those as my only goal right now is to unblock the react updates, which apollo-client was a blocker for since it required react 16 (I'm fairly certain I could've pushed past to 17 with no issues and just ignore the peer deps warnings but I don't intend on stopping at react 17 so we might as well do things "right" from the start).


lotas (approved):
awesome, thanks

PR #8856, opened 2026-07-09
Remove google analytics support

Eijebong:
AFAICT, this has never been used in prod. It got introduced in 2018 by e035521b680396907ece04680587ad1d18a3f5c7 but I can't find any evidence of it ever being used in production. Neither deployments (community / fxci) is currently trying to make any outbound request (except for auth0/gravatar), their window.env contains an empty `GA_TRACKING_ID`. Since this is pretty much dead code, the dep hasn't been touched for 4 years and *will* block us from updating to react 19 (latest supported is 18), just put the whole tracking analytics thing in the trash where it belongs.


lotas (approved):

petemoore (approved):
Thanks!

PR #8860, opened 2026-07-09
feat(ui): #8855 change priority of pending task

### LINKED ISSUE #8855: Add "Change Task Priority" feature in the UI
Since we added #8191 support for this in the API, we could also add this on a task level in the UI

For the Task page: `queue.changeTaskPriority`
For the Task Group page: `queue.changeTaskGroupPriority`

There should be a shared dialog component for both that would display a list (dropdown) of available priorities and "Change priority" button

Note: We should use queue api directly, not the web server graphql

petemoore:
I tested locally, and it seems to work well.

Fixes #8855.

### DIFF
--- a/changelog/change-task-priority-ui.md
+++ b/changelog/change-task-priority-ui.md
@@ -1,0 +1,4 @@
+audience: general
+level: minor
+---
+The task view now has a "Change Priority" action to change a task's priority directly from the UI.
--- a/changelog/change-task-priority-ui.md
+++ b/changelog/change-task-priority-ui.md
@@ -1,4 +1,5 @@
 audience: general
 level: minor
+reference: issue 8855
 ---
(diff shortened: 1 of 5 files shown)


changelog/change-task-priority-ui.md
```diff
@@ -0,0 +1,4 @@
+audience: general
+level: minor
```
lotas, line 2:
```suggestion
level: minor
reference: issue 8855
```

can we please link it to the issue also

lotas (dismissed):
LGTM. thanks

changelog/change-task-priority-ui.md
petemoore: whoops! good catch

lotas (approved):

PR #8862, opened 2026-07-10
Bug 2053989 - Document github comments tasks as elevating privileges

Eijebong:
This is the intended behavior for github comments. It shifts the trust process from the origin of the code to the reviewer, allowing them to run tasks for non collaborators after reviewing code from a PR.


lotas (dismissed):
thanks for the quick update

lotas (approved):

PR #8865, opened 2026-07-10
fix(queue): remove support for deprecated task priority 'normal'

### LINKED ISSUE #8858: Changing a task's priority to `normal` triggers a crash / incident
While working on #8855 i discovered that if I set a task's to priority `normal` a queue service incident is raised.

petemoore:
Fixes #8858

The `normal` priority was deprecated [9 years ago](https://github.com/taskcluster/taskcluster-queue/pull/166/changes), time to remove it for real!


db/versions/0127.yml
```diff
@@ -0,0 +1,55 @@
+version: 127
+description: Remove the `else 0` fallback for priority `normal` in `queue_pending_tasks_add_for_task`
+methods:
+  queue_pending_tasks_add_for_task:
```
lotas, line 4: I wonder if we really need to redefine this since there are few gates even before it reaches this point

lotas (approved):
Thanks

db/versions/0127.yml
lotas: oh crap, didn't realize it has automerge 😬 .. too late I guess

petemoore: i think it is cleaner not to have it, so if anything were to slip through here it would fail loudly rather than silently land with a lower priority

PR #8872, opened 2026-07-15
Add a way to disable the database version check

Eijebong:
I personally run a local postgres installation on arch which is 3 major versions ahead and I find it easier to just remove the check rather than bother running pg 15 instead every time. I finally got tired enough of seeing that message to add a way out of the check...

Note that while tests are passing with pg 18, `yarn generate` will generate a tiny diff in the documentation because of some constraints difference in database schemas. But CI will catch the diff anyway and go red if someone ever commits it.


lotas (approved):
LGTM, thanks

PR #8873, opened 2026-07-15
Relax the go version check from `yarn generate`

Eijebong:
In 32fbdf38addff1e9bd95af10c85e0f62db2a993b, I relaxed the go version check to not check for suffixes on versions to allow devs to use their package manager's version. I'm running into `yarn generate` being annoying again because my local golang updated to 1.26.5... I didn't go as far (yet) as what I'd really want to, because I don't know how common compiler breakage is with go, having never seen any issue, but I'm definitely tired of removing that check all the time. I think this is a safe middle ground.

### DIFF
--- a/changelog/Nvah8zb7TomLkyTi1wTUvg.md
+++ b/changelog/Nvah8zb7TomLkyTi1wTUvg.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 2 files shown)


lotas (approved):
sgtm!

PR #8875, opened 2026-07-16
Feat/d2g capability scopes

lotas:
This brings `disableSeccomp` capability handling on par with docker worker (checking scopes) and also improves scope checks or large set of scopes

Bugzilla Bug: [2055467](https://bugzilla.mozilla.org/show_bug.cgi?id=2055467)

### DIFF
--- a/tools/d2g/d2g.go
+++ b/tools/d2g/d2g.go
@@ -275,6 +275,12 @@ func validateDockerWorkerScopes(
 		)
 	}
 
+	if dwPayload.Capabilities.DisableSeccomp {
+		requiredScopes = append(requiredScopes,
+			[]string{"docker-worker:capability:disableSeccomp"},
+		)
+	}
+
 	if dwPayload.Capabilities.Devices.HostSharedMemory {
 		requiredScopes = append(requiredScopes,
 			[]string{"docker-worker:capability:device:hostSharedMemory"},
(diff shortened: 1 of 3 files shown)


tools/d2g/d2g.go
```diff
@@ -266,55 +266,77 @@ func validateDockerWorkerScopes(
...
-	}
-	if !scopesSatisfied {
-		return nil, fmt.Errorf("d2g task requires scopes:\n\n%v\n\nbut task only has scopes:\n\n%v\n\nYou probably should add some scopes to your task definition", requiredScopes, expandedScopes)
+	for _, required := range requiredScopes {
+		scopesSatisfied, err := expandedScopes.Satisfies(required, dummyExpander)
+		if err != nil {
+			return nil, fmt.Errorf("error expanding scopes: %v", err)
+		}
+		if !scopesSatisfied {
+			return nil, fmt.Errorf("d2g task requires scopes:\n\n%v\n\nbut task only has scopes:\n\n%v\n\nYou probably should add some scopes to your task definition", required, expandedScopes)
```
Eijebong: This is a regression in the error returned to the user? It used to return all scopes and now only returns the first one that doesn't match?

petemoore: Agree, although now it is rather complicated as we have a three dimensional slice 😬. Too bad the Scopes package uses DNF and not CNF - that is because originally the services declared scope requirements in DNF form, and then later it was changed to the current form. I guess we can just join the outer array with 'AND'.

lotas: true, let me update the message to include all scopes

Eijebong (approved):
Changes look correct.

PR #8876, opened 2026-07-16
Remove coverage dependencies in services/UI

Eijebong:
Those have been write only for at least 5 years. Nothing is consuming them, let's remove them. They can easily be re-added with up to date dependencies if we want them back at some point. See individual commits for reasoning on both


lotas (approved):
nice, thanks!
yeah, I don't remember ever seeing those coverages (besides in the task log directly)

PR #8877, opened 2026-07-16
Replace our dependency on md-directory with a handrolled glob -> map

Eijebong:
md-directory has not been updated for 6 years, depends on stuff that is older than that, polluting our lockfile for what is essentially used as a glob mapped to `gray-matter::matter`... Adding 500 lines of lockfile to avoid 5 lines of trivial code isn't worth it.


lotas (approved):
that's awesome

Issue #8088, opened 2025-11-14
Github service errors in logs


lotas:
Noticed in logs (just some random, without high frequency)

1. max response size
```
Error: Maximum response size reached

at .IncomingMessage.<anonymous> ( /app/node_modules/superagent/lib/node/index.js:1096 )
at .Stream.emit ( node:events:531 )
at .Stream.emit ( node:domain:489 )
at .Unzip.<anonymous> ( /app/node_modules/superagent/lib/node/unzip.js:47 )
at .Unzip.emit ( node:events:519 )
at .Unzip.emit ( node:domain:489 )
at .addChunk ( node:internal/streams/readable:561 )
at .readableAddChunkPushByteMode ( node:internal/streams/readable:512 )
at .Readable.push ( node:internal/streams/readable:392 )
at .Zlib.processCallback ( node:zlib:536 ) 
```

2. deprecation notice:

```
2025-11-14 15:03:29.679
Deprecation: [@octokit/request-error] `error.code` is deprecated, use `error.status`.
2025-11-14 15:03:29.679
    at RequestError.get (/app/node_modules/@octokit/core/node_modules/@octokit/request-error/dist-node/index.js:57:21)
2025-11-14 15:03:29.680
    at Monitor.reportError (file:///app/libraries/monitor/src/monitor.js:312:17)
2025-11-14 15:03:29.680
    at file:///app/services/github/src/handlers/index.js:148:30
```

3. Invalid request

```
code: 422
message: "Invalid request.

Only 65535 characters are allowed; 45687097 were supplied."
name: "HttpError"
owner: "web-platform-tests"
repo: "wpt"
sha: "7e4e59f2748fc1be13271d86d30728366084b0ba"
stack: "HttpError: Invalid request.

Only 65535 characters are allowed; 45687097 were supplied.
    at /app/node_modules/@octokit/core/node_modules/@octokit/request/dist-node/index.js:86:21
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async requestWithGraphqlErrorHandling (file:///app/node_modules/@octokit/plugin-retry/dist-bundle/index.js:36:20)
    at async Job.doExecute (/app/node_modules/bottleneck/light.js:405:18)"
status: 422
```

PR #8881, opened 2026-07-17
Use `status` instead of `code` to read github API return statuses

Eijebong:
The package that octokit uses (`octokit/request-error`) was giving both `code` and `status` since their original release with `code` being deprecated for pretty much as long as docker worker... When it got updated in 3b026c629eaffe2c28f539da798ce090bbb5ed59, code disappeared because their v6 removed it [1] but the guards were never updated. Tests passed since we're mocking those results and had both on the mocked result.

I found this by looking at sentry and seeing the github service throwing 500s for badges when it was trying to get results for a repository that didn't have taskcluster enabled (i.e with TC checks missing).

While looking for an open issue with this, I found #8088 which is 1/3rd about this, showing `Deprecation: [@octokit/request-error] `error.code` is deprecated, use `error.status`.

[1]: https://github.com/octokit/request-error.js/releases/tag/v6.0.0

### DIFF
--- a/changelog/dVRYQKw7SXCu2kgml6GHYw.md
+++ b/changelog/dVRYQKw7SXCu2kgml6GHYw.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+---
+Fetching a github status badge for a repository/branch that doesn't exist will
+now return a proper newrepo badge instead of a 500.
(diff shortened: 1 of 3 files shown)


Eijebong:
The sentry issue in question: `TASKCLUSTER-SERVICES-COMMUNITYTC-EB`.

lotas (approved):
Thanks for jumping into fixing this

PR #8884, opened 2026-07-20
Remove our dependency on dockerode and unused functions in docker tooling

Eijebong:
Pulling in a dependency for what we're doing here is way over the top. The only thing it provides over a subprocess is a progress bar iif the image is missing, the working tree is clean and the repo is at the last released tag.

On top of that, the file was already a mishmash of dockerode and docker commands anyway, meaning that if anyone was actually using this command, they had to use docker as a command already. The alternative here is updating the dependency but given the usage I'd rather take the switch to subprocesses.

The one technical regression here is that docker being down would return `false` from `dockerImageExists` instead of throwing since we can't detect a missing image from a command failure. The error would then come up from the `pull` instead. But the behavior and message would be correct so I don't think it's worth even trying, if the daemon isn't up, the image doesn't exist.

I also removed the `dockerRun` function in tooling. It's been unused since we stopped publishing docker worker in
3009d952f64544ab72f636e674ddbc54b67d1d5f but wasn't removed back then.

### DIFF
--- a/changelog/SAbCJGo8Q366rDCGA6AwnA.md
+++ b/changelog/SAbCJGo8Q366rDCGA6AwnA.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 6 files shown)


lotas (approved):
nice

PR #8891, opened 2026-07-21
chore(security): Update SECURITY guidelines

lotas:
to reflect current Mozilla direction and process and avoid contributors submitting sensitive bugs directly here


petemoore (approved):
Looks good, thanks Yarik!

PR #8892, opened 2026-07-21
feat(generic-worker): Added custom mime-types mapping for .jsonl

benoitgoupilleau:
## What

When looking at the artifacts of a test in treeherder, the artifacts in json lines format are currently returned with the mime type octet stream which triggers a download of the file instead of displaying it in a new tab. Buy returning the correct mime type, it will be displayed in a new tab using the JSON viewer

## Why
We are adding a new artifact for each test which will be in the json lines format and we are adding support to display it in the JSON viewer in firefox
Bugzilla Bug: [2055774](https://bugzilla.mozilla.org/show_bug.cgi?id=2055774)


workers/generic-worker/artifacts.go
```diff
@@ -25,7 +25,8 @@ var (
 
 		// keys *must* be lower-case
 
-		".log": "text/plain",
+		".log":   "text/plain",
+		".jsonl": "application/jsonlines",
```
Eijebong: https://jsonlines.org/ This says it should be `application/jsonl`. Any reason to pick `application/jsonlines` over it?

changelog/bug-2055774.md
```diff
@@ -0,0 +1,5 @@
+audience: worker-deployers
+level: silent
+reference: bug 2055774
+---
+The artifacts worker now returns the correct mime type for json lines files instead of octet stream
```
Eijebong: I'm not sure what you mean here by `artifacts worker`. Did you mean generic worker? Something along the line of "Generic worker will now upload .jsonl artifacts with the correct mime type (application/jsonl) [...]" might read better?

workers/generic-worker/artifacts.go
Eijebong: I also found `application/x-ndjson` in https://bugzilla.mozilla.org/show_bug.cgi?id=1603986  although I'd rather take the `jsonl` one if that's what people are actually going with.

Eijebong (changes requested):
Mind adding a test? It should be fairly easy in https://github.com/taskcluster/taskcluster/blob/d7cd97849a127cb755689296672dec093c96b7c9/workers/generic-worker/artifacts_test.go to use `validateArtifacts` for that.

workers/generic-worker/artifacts.go
benoitgoupilleau: thanks for the feedback. as it says it's not yet standardised. I have mostly seen jsonlines but probably it's better to follow the maybe future standard

workers/generic-worker/artifacts.go
```diff
@@ -25,7 +25,8 @@ var (
 
 		// keys *must* be lower-case
 
-		".log": "text/plain",
+		".log":   "text/plain",
+		".jsonl": "application/jsonl",
```
Eijebong, line 29: I'm still bummed about this but I also don't want to block on it forever...

Re-reading the issue in more details, it looks like it might end up as something completely different despite what the original message says or even not end up standard at all.

And your firefox patch doesn't even seem to have considered it as being valid so maybe I shouldn't have jumped on that `application/jsonl`. I misread that issue as a "we're way on the way to standardization with that name" when in fact, nothing's been done and they can't even agree on a name.

But it's also not the end of the world if we have to change this later on since it currently doesn't do anything anyway...

What's your thoughts on that since you probably have more experience with jsonl in the wild than me?

benoitgoupilleau: I think i'll keep the application/jsonl as it seems that it gonna be the standard. And yeah won't be a massive change to change that later. I'll more make sure that firefox can display them

Eijebong: Ok then I'll take it as is, thanks for dealing with me :)

Eijebong (approved):
Just leaving it here so it persists in something github won't hide by default. That mime type is not standard and might change in the future. Nothing's relying on this and if it ever gets standardized, just change it to whatever it should be.

PR #8898, opened 2026-07-21
Update express from 4 to 5

Eijebong:
Most of this is mechanical changes in the route format where I followed their guide here https://expressjs.com/en/guide/migrating-5/#path-route-matching-syntax

Namely, `/:foo?` -> `{/:foo}` and `/:foo(*)` -> `/*foo`. I also had to escape the `:` in `declareWithTrailingColon` or it would fail and try to parse a parameter there. I'm not sure that method is still necessary (it's been 3y, I hope people are not using routes with trailing `:` on worker pool names anymore...) but that's out of scope so I kept it working as is.

The change in route means I had to update our code that extracts parameters from them to generate clients, that's trivially testable by checking that `yarn generate` outputs the same before/after.

On change that actually impacts us in a way I couldn't fix at the root (and I don't think we should), splat params are now passed down as arrays split on `/` instead of strings. So "/scopes/foo/bar" would now pass `["foo", "bar"]` instead of `"foo/bar"`. Rather than adapting all routes, I added a simple `flattenParams` middleware that just reproduces the previous behaviour. When I did try, I found that it was hitting the issue in the authorization middleware *anyway* since scope checking requires those parameters to be `foo/bar`, so adding a middleware before solves both issues.

Another one is that `req.query` is no longer writable. The `queryValidator` middleware depended on being able to remove invalid parameters from the query so they wouldn't end up in logs but because express reparses the args instead on every single access, that wasn't working very well for us. I just replaced the `query` property on the request with the parsed / sanitized object, which has the double effect of removing invalid parameters *and* preventing express from parsing params over and over (I seriously have no idea why one would do that but I'm not gonna ask questions).

The only resulting change for users is that `/secret/` like routes don't work anymore and returns a 404 instead of a 403 (it used to pass an empty secret to the route and now doesn't match it anymore). We could restore that by duplicating the route definition but none of the routes using a `/` splat path (secrets, worker-pool, pending, purge-cache, task-queues) would benefit from it so I decided not to.

Closes #8897


Eijebong:
ARgh, these tests were failing locally because of ipv6 diffs and I didn't see the query difference... I'll investigate

Eijebong:
And that new failure is also probably due to this... I hate it here

libraries/api/src/middleware/queries.js
```diff
@@ -16,7 +16,17 @@ export const queryValidator = ({ entry }) => {
   return (req, res, next) => {
     /** @type {string[]} */
     const errors = [];
-    Object.entries(req.query || {}).forEach(([key, value]) => {
+    const sanitized = { ...(req.query || {}) };
+
+    // Replace the express read-only query with our own property so it can be
+    // mutated later
+    Object.defineProperty(req, 'query', {
```
lotas, line 23: neat

lotas (approved):
awesome, thanks! lgtm

PR #8899, opened 2026-07-22
Update action version comments in github workflows

Eijebong:
Seriously... If this continues being a pain like this it might be worth considering trying to remove those comments and/or find a way to make zizmor stop complaining about them...


petemoore (approved):
lol, so v7 = v7.0.0

PR #8908, opened 2026-07-22
Update email-templates to 13.0.1

Eijebong:
The update made it so styles were not inlined anymore and instead pulled into a `<style>` tag at the top. I'm way too scared of email to take that change so instead I changed the config we're passing to email-templates to match the previous rendering and added a test to lock it in. If anyone disagrees I'm happy to change that and take the new rendering but I wouldn't do so without checking the results on a bunch of webmails and mobile and desktop mail clients.

Closes #8707

### DIFF
--- a/changelog/PZ4_rUMISRCxq4ayKR4g-A.md
+++ b/changelog/PZ4_rUMISRCxq4ayKR4g-A.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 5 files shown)


lotas (approved):
I happened to be passing by, nice stuff, thanks!

petemoore (approved):

PR #8909, opened 2026-07-22
Enforce triggerSchema validation for pulse-triggered hooks

### LINKED ISSUE #8867: Hooks get triggered even though payload doesn't match `trigger_schema`
**Describe the bug**

I have a hook definition that specifies a `trigger_schema` where the accepted values are a subset of the valid values. I'd expect the hook to fire only if the payload is valid for the `trigger_schema`. However, I have found that, when the hook fires based on Pulse messages (with bindings less tight than the `trigger_schema`), even non-conformant payload lead to an execution of the hook.

**To Reproduce**

1. In fxci-config, I have this in the `trigger_schema` https://github.com/mozilla-releng/fxci-config/blob/e4a832166be4b3ca8dbd267a69dfb0cdfef12196/hooks.yml#L227, and those bindings https://github.com/mozilla-releng/fxci-config/blob/e4a832166be4b3ca8dbd267a69dfb0cdfef12196/hooks.yml#L198-L200
2. Send a message on a matching queue+routing_key with a `body.action` not in the list [e.g., add or remove a reviewer on a PR in on https://github.com/mozilla-firefox/infra-testing, such as https://github.com/mozilla-firefox/infra-testing/pull/43]

**Expected behavior**
The hook is not triggered.

**Taskcluster version**
101.1.0 (https://firefox-ci-tc.services.mozilla.com)

lotas:
feat(hooks): triggerSchema is used against Pulse payloads before creating a task

this brings parity between direct hook invocation and pulse-triggered
events. `triggerSchema` enforces both paths, and if hook payload fails
validation, no tasks will be created.

If hook was created with default triggerSchema (additionalProperties:
false) it would likely fail now, hook owners would need to update it

Fixes #8867

this is a breaking change, and there are some hooks that needs closer look before this goes live
high risk candidates are `hg-push/*` hooks

### DIFF
--- a/changelog/issue-8867.md
+++ b/changelog/issue-8867.md
@@ -1,0 +1,9 @@
+audience: users
+level: major
+reference: issue 8867
+---
+Pulse-triggered hooks now validate matching pulse message payloads against the hook's `triggerSchema` before creating a task.
+If a pulse message matches the hook's `bindings` but fails `triggerSchema` validation, the message is discarded and no task is created.
+This is a breaking change: previously `triggerSchema` was only enforced on the API and webhook paths, and pulse messages fired the hook regardless of their payload.
+Validation is unconditional, including for hooks that did not set a `triggerSchema`: the default schema only accepts an empty payload, so such a hook will no longer fire on pulse messages that carry a payload.
+Before upgrading, review all hooks with pulse `bindings` and make sure each `triggerSchema` accepts the pulse payloads that should still create tasks, for example by allowing expected properties or using `additionalProperties: true`.
--- a/clients/client-go/tchooks/types.go
+++ b/clients/client-go/tchooks/types.go
@@ -73,6 +73,14 @@ type (
 		// Additional properties allowed
 		Task json.RawMessage `json:"task"`
 
+		// JSON Schema used to validate payloads supplied when firing the hook.
+		// API calls and webhooks are rejected when their payloads fail validation.
+		// Pulse messages that match the hook's bindings are discarded without
+		// creating a task when their payloads fail validation. When omitted, this
+		// defaults to a schema that only accepts an empty payload, so a hook
+		// without an explicit triggerSchema will not fire on Pulse messages that
+		// carry a payload.
+		//
 		// Default:    {
 		//               "additionalProperties": false,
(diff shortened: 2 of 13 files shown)


Eijebong (dismissed):
It's nice to have this documented now but IMO we should validate the schema and find a way to make this debuggable.

lotas:
> It's nice to have this documented now but IMO we should validate the schema and find a way to make this debuggable.

do you have anything in mind? Maybe we could reuse notify service somehow. Before we introduce "last hook misfires" entity to track every rejected invocation

Eijebong:
I don't have any good idea no. Got plenty of bad ones. Something in the UI would be nice but it'd be so noisy for real hooks that I don't think it's feasible. Unless we implement some debug mode for hooks but then that's so overkill for something that one person is going to be using once every 5 years

lotas:
I like the debug mode idea.. and maybe the simplest one would be to create task always but if validation fails use `built-in/fail` pool to leave trace? or maybe we could introduce similar built-in for this specific usecase

petemoore (commented):
I think the notification mechanism is an important question, but I think it is independent of the requirement for schema validation. I see the pulse-triggered hooks being validated differently to api triggered hooks as a problem though. I don't see a good argument for the two triggering mechanisms to have differing behaviour. Documenting the difference is better than not documenting it, so I'm not giving an r-, but I think we should still fix this. Ultimately the fact the queue.createTask has schema validation too is some protection, but being able to supply a triggerSchema and it not being used, seems wrong (iiuc).

lotas:
after long discussion in slack we agreed that triggerSchema should also be validating pulse events. I'll update the pr with new proposed changes

petemoore (changes requested):
I reviewed #8915 first and a lot of the changes i reviewed there came from here - apologies, since the review now lives in the other PR. But see the comments in https://github.com/taskcluster/taskcluster/pull/8915#pullrequestreview-4772808810. Thanks Yarik!

This is a really nice piece of work, only the issue about default behaviour when no triggerSchema is provided I think is wrong, the rest looks pretty perfect! :-)

lotas:
let me close it then in favor of #8915 since it contains both

PR #8910, opened 2026-07-22
Worker Runner (Windows): wait for Windows Setup completion on start up

Callek:
## Summary

Adds an optional worker-runner startup gate that prevents Windows workers
from registering with worker-manager or accepting work before Windows Setup
has completed the Sysprep `specialize` and OOBE sequence.

Set the following in the worker-runner configuration to enable it:

```yaml
waitForWindowsImageState: true
```

The option is provider-independent and can be used for Windows workers on
AWS, Azure, Google Cloud, or other supported providers.

## Symptoms of the issue this fixes

Windows workers claimed their first task and resulted in CLAIM_EXPIRED frequently, subsequent tasks had no such issue. Said second task would generally start well before the CLAIM ended up actually expiring, suggesting a reboot of the system.

## Root cause

A generalized Windows image can start automatic services while Windows Setup
is still processing its first-boot configuration.

During an observed EC2 first launch, Windows progressed through these image
states:

```text
IMAGE_STATE_GENERALIZE_RESEAL_TO_OOBE
-> IMAGE_STATE_UNDEPLOYABLE
-> IMAGE_STATE_SPECIALIZE_RESEAL_TO_OOBE
-> restart
-> IMAGE_STATE_UNDEPLOYABLE
-> IMAGE_STATE_COMPLETE
```

The Worker Runner service started while the image was
`IMAGE_STATE_UNDEPLOYABLE` or
`IMAGE_STATE_SPECIALIZE_RESEAL_TO_OOBE`. It registered the instance with
worker-manager, allowing Generic Worker to claim a task before Windows
performed the planned Setup restart. The restart then interrupted the task.

The restart was requested by the Windows Setup `specialize` pass and was not
caused by Windows Update or an unexpected operating-system failure.

## Changes

When `waitForWindowsImageState` is enabled, worker-runner now:

- Reads the 64-bit Windows Setup registry value at:

  ```text
  HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Setup\State\ImageState
  ```

- Polls every 10 seconds until the value is:

  ```text
  IMAGE_STATE_COMPLETE
  ```

- Retries registry-read failures rather than allowing the worker to start
  prematurely.
- Performs the check before loading cached runner state, configuring the cloud
  provider, registering with worker-manager, retrieving secrets, or starting
  the worker.
- Leaves existing behavior unchanged by default.
- Treats the setting as a no-op on non-Windows systems.

The polling logic is separated from the Windows registry implementation so it
can be unit tested without requiring a Windows host.

## Expected result

A newly launched generalized Windows instance may still perform its normal,
planned Setup restart.

However, worker-runner will not register the instance or start Generic Worker
until Windows reports `IMAGE_STATE_COMPLETE`. Tasks therefore cannot be
claimed during the first-boot Setup sequence and interrupted by its restart.

## Validation

- [x] Added a unit test covering registry-read errors and intermediate Setup
      states before `IMAGE_STATE_COMPLETE`.
- [x] Verified runner configuration parsing.
- [x] Ran `go test ./tools/worker-runner/...`.
- [x] Ran `go vet ./tools/worker-runner/...`.
- [x] Cross-compiled the runner test package for Windows.

## Not Done

- I have no tested this worker runner go binary on my Windows Workers directly, as our deployment story is strictly tied to taskcluster release versions

## Extra Reviewer Notes
- This code was primarily generated via AI-Assisted-Development using GitHub Copilot with GPT-5.6 Sol.
- I am not a skilled go developer, and have barely done go work, please validate against best practices that my AI tools didn't follow.

### DIFF
--- a/changelog/worker-runner-windows-image-state.md
+++ b/changelog/worker-runner-windows-image-state.md
@@ -1,0 +1,7 @@
+level: minor
+audience: worker-deployers
+---
+Worker-runner can now wait for Windows Sysprep and OOBE to complete before it
+registers with worker-manager or starts the worker. Set
+`waitForWindowsImageState: true` in the runner configuration to prevent tasks
+from being claimed before a first-launch OOBE reboot.
--- a/tools/worker-runner/cfg/runnerconfig.go
+++ b/tools/worker-runner/cfg/runnerconfig.go
@@ -11,12 +11,13 @@ import (
 // RunnerConfig defines the configuration for taskcluster-worker-starter.  See the usage
 // string for field descriptions
 type RunnerConfig struct {
-	Provider             ProviderConfig             `yaml:"provider"`
-	WorkerImplementation WorkerImplementationConfig `yaml:"worker"`
-	WorkerConfig         *WorkerConfig              `yaml:"workerConfig"`
-	Logging              *LoggingConfig             `yaml:"logging"`
-	GetSecrets           bool                       `yaml:"getSecrets"`
-	CacheOverRestarts    string                     `yaml:"cacheOverRestarts"`
+	Provider                 ProviderConfig             `yaml:"provider"`
+	WorkerImplementation     WorkerImplementationConfig `yaml:"worker"`
+	WorkerConfig             *WorkerConfig              `yaml:"workerConfig"`
+	Logging                  *LoggingConfig             `yaml:"logging"`
+	GetSecrets               bool                       `yaml:"getSecrets"`
+	CacheOverRestarts        string                     `yaml:"cacheOverRestarts"`
+	WaitForWindowsImageState bool                       `yaml:"waitForWindowsImageState"`
 }
 
 // LoadRunnerConfig loads a worker-runner configuration file.
--- a/tools/worker-runner/cfg/runnerconfig_test.go
+++ b/tools/worker-runner/cfg/runnerconfig_test.go
@@ -40,6 +40,7 @@ func TestLoadConfig(t *testing.T) {
 	assert.Equal(t, "ec2", runnercfg.Provider.ProviderType, "should read providerType correctly")
 	assert.Equal(t, 10.0, runnercfg.WorkerConfig.MustGet("x"), "should read workerConfig correctly")
 	assert.Equal(t, true, runnercfg.GetSecrets, "getSecrets should default to true")
+	assert.True(t, runnercfg.WaitForWindowsImageState, "waitForWindowsImageState should be read correctly")
 }
 
 // TestLoadConfig_InsecurePerms verifies that LoadRunnerConfig tightens the
--- a/tools/worker-runner/cfg/test-config.yml
+++ b/tools/worker-runner/cfg/test-config.yml
@@ -1,4 +1,5 @@
 provider:
     providerType: 'ec2'
+waitForWindowsImageState: true
 workerConfig:
     x: 10
--- a/tools/worker-runner/runner/image_state.go
+++ b/tools/worker-runner/runner/image_state.go
@@ -1,0 +1,21 @@
+package runner
+
+import "log"
+
+const imageStateComplete = "IMAGE_STATE_COMPLETE"
+
+func waitForImageStateComplete(readImageState func() (string, error), wait func()) {
+	for {
+		imageState, err := readImageState()
+		if err != nil {
+			log.Printf("Could not read Windows image state: %v; retrying", err)
+		} else if imageState == imageStateComplete {
+			log.Printf("Windows image state is %s", imageStateComplete)
+			return
+		} else {
+			log.Printf("Windows image state is %q; waiting for %s", imageState, imageStateComplete)
+		}
+
+		wait()
+	}
+}
--- a/tools/worker-runner/runner/image_state_other.go
+++ b/tools/worker-runner/runner/image_state_other.go
@@ -1,0 +1,5 @@
+//go:build !windows
+
+package runner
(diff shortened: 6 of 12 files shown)


petemoore (commented):
I think this issue isn't specific to Worker Runner, it is a problem for _any_ service that should only run once Windows Setup has completed.

It looks like `%WINDIR%\Setup\Scripts\SetupComplete.cmd` is designed exactly for this purpose. It runs as `LocalSystem` when Windows Setup is complete. I think the correct fix here is to update the taskcluster docs to say if using SysPrep (e.g. for a cloud worker) then set Worker Runner to start on demand, and add these two lines to `%WINDIR%\Setup\Scripts\SetupComplete.cmd` (note: the space between `start=` and `auto` is required):

```cmd
sc config "worker runner" start= auto
sc start "worker runner"
```

This starts Worker Runner after Windows Setup has completed, and sets it to auto-launch on boot. The reason I prefer this solution, is that:
1) It is general purpose for any service that should only run once Windows Setup has completed (which may apply to more services than just Worker Runner).
2) There could be multiple system setup steps or tools that need to run before Worker Runner is launched, we shouldn't bake knowledge of all of them into Worker Runner; the scheduling should simply take care of ensuring that setup steps are completed before services are started that depend on them.

We can meet up to discuss if you like.

Callek:
> I think this issue isn't specific to Worker Runner, it is a problem for _any_ service that should only run once Windows Setup has completed.

Not necessarily wrong, but admittedly the docs for worker-runner do state to launch in auto-start with nssm.

> It looks like `%WINDIR%\Setup\Scripts\SetupComplete.cmd` is designed exactly for this purpose. It runs as `LocalSystem` when Windows Setup is complete. I think the correct fix here is to update the taskcluster docs to say if using SysPrep (e.g. for a cloud worker) then set Worker Runner to start on demand, and add these two lines to `%WINDIR%\Setup\Scripts\SetupComplete.cmd` (note: the space between `start=` and `auto` is required):
> 
> ```batchfile
> sc config "worker runner" start= auto
> sc start "worker runner"
> ```

I looked into this a bit, and a few notes: `SetupComplete.cmd` may not be available in all cases (such as oem windows client versions, as opposed to windows server). Seems like Windows 11 Arm on Azure would support it though (at least in a quick check) -- additionally SetupComplete.cmd can be modified/updated by `sysprep` itself, which makes placing the worker-runner commands "at the end" a bit hard if you follow the general guidance of how you should `sysprep` [which is let the system capture happen *immediately* after running sysprep] - This can occur directly, or even when calling the `EC2Launch` wrapper as AWS documents

> This starts Worker Runner after Windows Setup has completed, and sets it to auto-launch on boot. The reason I prefer this solution, is that:
>     1. It is general purpose for any service that should only run once Windows Setup has completed (which may apply to more services than just Worker Runner).
>     2. There could be multiple system setup steps or tools that need to run before Worker Runner is launched, we shouldn't bake knowledge of all of them into Worker Runner; the scheduling should simply take care of ensuring that setup steps are completed before services are started that depend on them.

And absolutely, if you are already doing multiple system steps or tools that need to run before Worker Runner is launched, you shouldn't use autostart and should instead put worker-runner in the ordered list of how those tools run.

As far as I can tell, every working windows system should be in this IMAGE State when its ready, and this configuration is optional but allows the `auto start` process, as documented, work effectively on systems that need to reboot due to `specialize` like this.

This was never a problem for us in the past, but we recently updated taskcluster versions and *something* made the time between system boot and generic-worker claiming work smaller, which is what exposed this.

> We can meet up to discuss if you like.

I'm happy to chat more, if useful.

=====

A note as helped by my AI:
> `SetupComplete.cmd` is itself executed by Windows Setup. Microsoft does not document that ImageState is already `IMAGE_STATE_COMPLETE` when the script begins. It is a late Setup hook, and Setup might complete after the script returns.

=====

I do currently have a wrapper script in place to do effectively exactly what this patch does, so making a release with this is not especially time-pressing.

Eijebong:
I agree with Pete here. This does not feel like something either worker-runner or generic-worker should be concerned about. When you know you have a machine that's going to reboot, you don't start a service which is going to be badly affected by said reboot (because in this case, it'll pick up a task here and get interrupted), it's you starting the service early that's at fault, not the service itself for doing what it was intended to do.

What I do agree on though is that debugging this sucks. And discovery for the solution too.
I think a **tiny** paragraph in `ui/docs/reference/workers/generic-worker/installing.mdx` to tell people about `SERVICE_AUTO_START` and `SetupComplete.cmd` would be a much more preferable solution than adding polling of a registry key in worker-runner. Because at that point I'd argue that generic-worker on its own has the same problem, there's nothing saying that you have to use worker-runner, it's perfectly valid to run generic-worker on its own.

petemoore:
@Callek Based on the feedback above, I'll close this in favour of a Windows Installation doc update, to add some specific guidance around Windows Update / SysPrep integration when creating cloud images. It sounds like you have a working solution for your environment, so hopefully this will not cause you any problems. Thanks!

Callek:
> @Callek Based on the feedback above, I'll close this in favour of a Windows Installation doc update, to add some specific guidance around Windows Update / SysPrep integration when creating cloud images. It sounds like you have a working solution for your environment, so hopefully this will not cause you any problems. Thanks!

No problem, thank you for the update!  I admittedly was on the fence about "does this belong in worker-runner/generic-worker" but landed here.  Either way, my wrapper script is indeed keeping things working on my end.

PR #8911, opened 2026-07-22
migrate secrets from graphql to rest

### LINKED ISSUE #8901: Migrate Secrets page from GraphQL to REST
The Secrets UI currently uses GraphQL for listing, viewing, creating, updating, and deleting secrets. Migrate secrets page to use the Secrets client from @taskcluster/client-web.

Parent Issue: #8874

thakkarujas5:
Github Issue: Fixes #8901 

* Change ViewSecrets page to use REST
* Change ViewSecret page to use REST
* Remove GraphQL files


thakkarujas5:
have cherry picked the commit of decorator PR to unblock development of this PR, will take the rebase once decorator PR is merged.

thakkarujas5:
Previously the search in secrets page was powered by the loader (https://github.com/taskcluster/taskcluster/blob/main/services/web-server/src/loaders/secrets.js#L12).

Since now we directly call the list endpoint so have moved the same logic of filtering in ui/src/utils/searchFilter.js. 

Can inline this in case we dont want to keep it as a separate function.

ui/src/views/Secrets/ViewSecret/index.jsx
```diff
@@ -35,13 +23,48 @@ export default class ViewSecret extends Component {
...
 
+  componentDidMount() {
+    if (!this.props.isNewSecret) {
+      this.fetchSecret();
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    // react-router reuses this instance across /secrets/:secret, so a
+    // back/forward navigation between two secrets only changes the params --
```
lotas: I think this is a made up reason somehow, because you can't jump between secrets with some navigation on that page. only if you enter in URL manually. 
Needs local testing though, because I'm 70% certain Going back from secret to listing and then selecting a new one would reuse the same component

ui/src/views/Secrets/ViewSecret/index.jsx
```diff
@@ -35,13 +23,48 @@ export default class ViewSecret extends Component {
...
+  };
+
   handleDeleteSecret = name => {
     this.setState({ dialogError: null, loading: true });
 
-    return this.props.client.mutate({
-      mutation: deleteSecretQuery,
-      variables: { name },
-    });
+    const client = this.props.createTaskclusterClient({ Class: Secrets });
```
lotas: used at least x3 times, can be put into a property probably

```js
get secretsClient() {
  return this.props.createTaskclusterClient({ Class: Secrets });
}
..
handleDeleteSecret = name => {
  ..
  this.secretsClient.remove(name);
}
```

thakkarujas5: done

ui/src/views/Secrets/ViewSecret/index.jsx
thakkarujas5: so do we want to rule out the possibility that secrets page cant be navigates by editing url or should i just put this case in the comment?

lotas: Probably better to test first, it feels like it will be a very rare usecase, but maybe I'm wrong. Let's keep it

ui/src/components/PaginatedDataTable/index.jsx
```diff
@@ -0,0 +1,304 @@
...
+    marginTop: theme.spacing(-1),
+    marginBottom: theme.spacing(1),
+    marginLeft: theme.spacing(2),
+    marginRight: theme.spacing(2),
+    width: '96%',
+  },
+}));
+
+/**
+ * A paginated table over a plain array of items.
```
lotas, line 46: We could probably keep the previous version here that mentioned that it is a `ConnectionDataTable` counterpart for rest client and that the other one would be dropped once we fully migrate

changelog/issue-8901.md
```diff
@@ -0,0 +1,5 @@
+audience: developers
+level: patch
+reference: issue 8901
+---
+Replace Secrets page GraphQL calls with REST
```
lotas:
```suggestion
UI Secrets page switches from GraphQL to direct service calls
```

nit: maybe something like this

ui/src/components/SecretsTable/index.jsx
```diff
@@ -37,55 +31,46 @@ const iconSize = 16;
...
+      );
     },
     {
-      serializer: ([secretsConnection, sortBy, sortDirection]) => {
-        const ids = sorted(secretsConnection.edges);
-
-        return `${ids.join('-')}-${sortBy}-${sortDirection}`;
-      },
+      serializer: ([secrets, sortBy, sortDirection]) =>
+        `${[...secrets].sort(sort).join('-')}-${sortBy}-${sortDirection}`,
```
lotas: hmm. this will probably be a long string, we patched memoize at some point to not use those full strings, but turn it into hash instead. 

But also if secrets is just a simple `string[]` we won't gain much by using memoize here

also we are sorting everything twice essentially, plus wasting memory.
I say drop it completely here (memoize layer)

ui/src/utils/usePaginatedResource.js
```diff
@@ -0,0 +1,93 @@
...
+  const [state, setState] = useState({
+    items: [],
+    continuationToken: null,
+    loading: true,
+    error: null,
+  });
+  const [page, setPage] = useState(0);
+  const [reloadCount, setReloadCount] = useState(0);
+
+  const tokens = useRef([null]);
```
lotas: I think this is a bit unclear what a `tokens` here stand for. Naming is hard, as always.

maybe something like `pageContinuationTokens`?

ui/src/utils/usePaginatedResource.js
```diff
@@ -0,0 +1,93 @@
...
+
+    const continuationToken = tokens.current[page];
+
+    if (continuationToken) {
+      options.continuationToken = continuationToken;
+    }
+
+    fetch(options).then(
+      response => {
+        if (id !== requestId.current) {
```
lotas, line 50: Is this to avoid concurrent fetches? Or to avoid jumping to next page before current didn't load? Would be good to add some comment when this might happen

lotas: probably worth keeping, this might happen when we loaded few pages and then navigating back and forth, so only processing when current page is expected

ui/src/views/Secrets/ViewSecret/index.jsx
lotas: couldn't reproduce it locally because page reloads completely on navigation, maybe due to some other issues

lotas: Ah, so the `/secrets/:secret` route stays mounted across param-only changes, so switching secrets in place would keep the old one on screen. No in-app link currently does that today (detail pages are only reachable from the list)

we can either keep it for future when someone would add extra navigation and fix the comment to avoid confusion, or just drop it

changelog/issue-8901.md
thakkarujas5: changed

ui/src/components/SecretsTable/index.jsx
thakkarujas5: removed the memoize layer

ui/src/utils/usePaginatedResource.js
thakkarujas5: changed

ui/src/components/PaginatedDataTable/index.jsx
thakkarujas5: added back the comment

ui/src/views/Secrets/ViewSecret/index.jsx
thakkarujas5: dropped the componentDidUpdate

ui/src/views/Secrets/ViewSecrets/index.jsx
```diff
@@ -131,72 +75,104 @@ export default class ViewSecrets extends Component {
...
+          error,
+          page,
+          hasNextPage,
+          hasPreviousPage,
+          nextPage,
+          previousPage,
+          reload,
+        }) => {
+          const secrets = searchTerm
+            ? items.filter(name => name.includes(searchTerm))
```
lotas: this should probably be case insensitive to match what we had on the graphql server side

https://github.com/taskcluster/taskcluster/blob/6ba361152cb1c940b32da640a31a5bb1b203898b/services/web-server/src/utils/searchFilter.js#L11

ui/src/views/Secrets/ViewSecrets/index.jsx
```diff
@@ -131,72 +75,104 @@ export default class ViewSecrets extends Component {
...
-                This will delete the secret {deleteSecretName}.
-              </Typography>
-            }
-            confirmText="Delete Secret"
-          />
-        )}
-      </Dashboard>
+      <PaginatedResource
+        fetch={this.fetchSecrets}
+        payload={{ limit: VIEW_SECRETS_PAGE_SIZE }}
```
lotas: we should likely include `searchTerm` here as well to make sure we reload properly

ui/src/utils/usePaginatedResource.js
```diff
@@ -0,0 +1,97 @@
...
+          return;
+        }
+
+        setState({
+          items: select(response),
```
lotas, line 55: probably minor but I wonder if this could throw some error and we'll not be catching it properly

we could additionally add a catch block
```js
fetch(options).then(...).catch(// some erorr handling if select goes wrong)
```

ui/src/views/Secrets/ViewSecrets/index.jsx
thakkarujas5: payload is the body passed to the api as it is along with continutation token, since search is not part of api payload so have not added it

ui/src/views/Secrets/ViewSecrets/index.jsx
thakkarujas5: filtering it now by converting both sides to lowercase

ui/src/utils/usePaginatedResource.js
thakkarujas5: added catch error handling.

ui/src/components/PaginatedDataTable/index.jsx
```diff
@@ -0,0 +1,308 @@
...
+   *
+   * This currently works on single column tables (lists); rendering of
+   * multiple columns will be out of sync with header widths.
+   */
+  lazyRender: bool,
+  /** Height in pixels of a single row. */
+  rowHeight: number,
+};
+
+PaginatedDataTable.defaultProps = {
```
lotas: I'm so sorry.. react really hates devs who want to support anything written with it.. 

https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-deprecated-react-apis

they deprecated/deprecating both defaultProps and propTypes.. 
`defaultProps` belongs to the default values in the component function directly, it would also make reading and supporting much simpler, since defaults would live next to the variable declaration

but propTypes are different .. they recommend using type checkers (and you probably don't want to extend this small graphql removal into a typescript migration also xD)

so let's keep `propTypes` for now (we are on react17, so will continue working) and only merge defaultProps into function arg defaults

ui/src/views/Secrets/ViewSecrets/index.jsx
```diff
@@ -11,57 +10,50 @@ import Search from '../../../components/Search';
...
-    };
-  },
-})
 @withStyles(theme => ({
   plusIconSpan: {
     ...theme.mixins.fab,
   },
 }))
+@withTaskclusterClient
+@withPaginatedResource({
```
lotas, line 25: nice! this is a much smaller change than then <Dashboard> wrapping! Thanks

ui/src/components/PaginatedDataTable/index.jsx
thakkarujas5: done! added them in function arguments

lotas (approved):
Awesome!

PR #8913, opened 2026-07-23
feat: add withTaskclusterClient decorator for REST clients

### LINKED ISSUE #8912: UI: Authentication for REST Clients
Currently the UI REST Client uses user credentials passed from the component props. This can lead to incorrect client initialization as the auth is not loaded at that point of time.

Figure out a way to make it similar to authLink used in GraphQL

thakkarujas5:
Github Issue: Fixes #8912

### DIFF
--- a/changelog/issue-8912.md
+++ b/changelog/issue-8912.md
@@ -1,0 +1,6 @@
+audience: developers
+level: patch
+reference: issue 8912
+---
+UI adds withTaskclusterClient decorator to create rest client removing the need to using props
+for rest client creation
--- a/ui/src/App/index.jsx
+++ b/ui/src/App/index.jsx
@@ -19,6 +19,8 @@ import CssBaseline from '@material-ui/core/CssBaseline';
 import Main from './Main';
 import { ToggleThemeContext } from '../utils/ToggleTheme';
 import { AuthContext } from '../utils/Auth';
+import { TaskclusterClientContext } from '../utils/TaskclusterClient';
+import { getClient } from '../utils/client';
 import db from '../utils/db';
 import reportError from '../utils/reportError';
 import ErrorPanel from '../components/ErrorPanel';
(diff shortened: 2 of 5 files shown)


ui/src/views/Secrets/ViewSecrets/index.jsx
```diff
@@ -45,6 +48,14 @@ export default class ViewSecrets extends Component {
     dialogError: null,
   };
 
+  async componentDidMount() {
+    try {
```
thakkarujas5: kept this for demonstrating how the decorator works, will remove it at the time of PR approval

thakkarujas5:
Screenshot 2026-07-23 at 10 30 24 PM

thakkarujas5:
credentialAgent breaks buildSignedUrlSync ( https://github.com/taskcluster/taskcluster/blob/main/clients/client-web/src/Client.js#L190-L192)

As views would migrate to createTaskclusterClient, `getArtifactUrl` method (getArtifactUrl.js) uses the older way of making queueClient.

Should we initialize the client by ading the credentails agent in every component so that each component has the choice which type of client they want to make?

changelog/issue-8912.md
```diff
@@ -0,0 +1,6 @@
+audience: developers
+level: minor
```
lotas:
```suggestion
level: patch
```
I don't think it's changing much at the moment, maybe  patch is enough?

changelog/issue-8912.md
```diff
@@ -0,0 +1,6 @@
...
+reference: issue 8912
+---
+add withTaskclusterClient decorator to create rest client removing the need to using props
```
lotas:
```suggestion
UI adds withTaskclusterClient decorator to create rest client removing the need to using props
```

we can make it obvious for changelog reader what is going on here

changelog/issue-8912.md
thakkarujas5: changed :)

changelog/issue-8912.md
thakkarujas5: changed

lotas (approved):
Thanks, LGTM

PR #8914, opened 2026-07-23
Support pod annotations in the Helm chart

Callek:
## Summary

Add support for applying arbitrary annotations to pods created by the Taskcluster Helm chart. This is useful for deployment-specific integrations and policies such as Karpenter's `karpenter.sh/do-not-disrupt` annotation.

The chart now supports:

- A global `podAnnotations` map applied to every Taskcluster Deployment and CronJob pod.
- A `podAnnotations` map on each service process for workload-specific annotations.
- Additive merging of the two maps, with process-specific values taking precedence when both maps define the same key.

For example:

```yaml
podAnnotations:
  example.com/global: foo
  karpenter.sh/do-not-disrupt: "true"

queue:
  procs:
    claimResolver:
      podAnnotations:
        example.com/process: bar
        karpenter.sh/do-not-disrupt: "false"
```

The `claimResolver` pods receive both `example.com/global` and `example.com/process`, with their process-specific Karpenter value. Other Taskcluster pods receive the global annotations.

## Implementation

- Add global and per-process `podAnnotations` values to the generated Helm values schema.
- Apply the merged annotations to pod templates generated for both Deployments and CronJobs.
- Preserve the chart-managed `checksum/secret` annotation on Deployment pod templates so secret changes continue to trigger rollouts.
- Use generator replacements for the Helm annotation map because the JSON-e source templates must remain valid YAML before Helm evaluates the values.
- Regenerate the chart templates, default values, example development configuration, values-schema references, and deployment documentation.
- Add a deployer-facing changelog entry.

Annotation values are constrained to strings, matching Kubernetes' annotation value type.

## Testing

- Regenerated all Kubernetes chart templates and verified a second generation produced an identical diff.
- Regenerated the references target and verified its embedded values schema matches `infrastructure/k8s/values.schema.json`.
- Rendered representative Deployment and CronJob templates with Helm and parsed the resulting YAML.
- Verified global annotations are inherited, process annotations are additive, and matching process keys override global values.
- Verified the Deployment secret checksum cannot be replaced by a user annotation.
- Verified an unconfigured CronJob renders a valid empty annotations map.
- Ran `biome check` on the modified generator and `git diff --check`.


infrastructure/tooling/src/generate/generators/k8s.js
```diff
@@ -272,10 +272,15 @@ const renderTemplates = async (name, vars, procs, templates) => {
       readinessPath,
       labels: labels(`taskcluster-${name}`, proc),
     };
+    // A literal << in deployment.yaml would be resolved by js-yaml before Helm renders the annotation map.
+    // Replace the merge-key sentinel after dumping instead.
+    const podAnnotationsConfig = `mergeOverwrite (dict) (.Values.podAnnotations | default dict) (.Values.${context.configName}.procs.${context.configProcName}.podAnnotations | default dict)`;
```
lotas: At the moment we still support helm v2 (although there is an open #8693 to drop v2), it would break since `mergeOverwrite` is not supported there

Only `merge` is available: https://helm.sh/community/incubator/#merge

infrastructure/tooling/src/generate/generators/k8s.js
```diff
@@ -272,10 +272,15 @@ const renderTemplates = async (name, vars, procs, templates) => {
...
     const replacements = {
       REPLICA_CONFIG_STRING: `{{ int (.Values.${context.configName}.procs.${context.configProcName}.replicas) }}`,
       IMAGE_PULL_SECRETS_STRING:
         '{{ if .Values.imagePullSecret }}{{ toJson (list (dict "name" .Values.imagePullSecret)) }}{{ else }}[]{{ end }}',
+      POD_ANNOTATIONS_MERGE_KEY: '<<',
```
lotas: I'm a bit afraid of using such yaml constructs `<<:` which might be library-specific at parsing/rendering so might break some tools like ArgoCD (which is used by mozilla) 
additionally native kubectl does not support such kind of merging

I think we can get rid of this and just merge whole annotation object with `merge` to keep it helmv2 compatible and avoid using syntax that might break some platforms:

```
{{ merge (dict "checksum/secret" (include (print $.Template.BasePath "/${projectName}-secret.yaml") . | sha256sum)) (....procs.<proc>.podAnnotations | default dict) (.Values.podAnnotations | default dict) | toJson }}
```

(for deployments we want to keep checksum/secret always)

lotas (changes requested):
Thanks @Callek 
I think this requires few tweaks before we can land it to avoid some issues with yaml rendering.
It's unfortunate that we have so many layers here.. js generated json-e template rendered into yaml train 😬

infrastructure/tooling/src/generate/generators/k8s.js
Callek: Didn't realize we still supported helm v2, will adjust

infrastructure/tooling/src/generate/generators/k8s.js
Callek: Yea, the reason I did it this way, is that i wasn't sure that we'd want to move the secret checksum bit into the k8s.js rendering logic. But this is doable

infrastructure/tooling/src/generate/generators/k8s.js
lotas: yeah. the risk is quite low probably .. but just in case 🤞 
thanks

lotas (approved):
Thanks for the updates, LGTM

PR #8915, opened 2026-07-24
Hooks triggerSchema is used to validate pulse events from now on, UI provides easy debugger

### LINKED ISSUE #8867: Hooks get triggered even though payload doesn't match `trigger_schema`

lotas:
feat(hooks): triggerSchema is used against Pulse payloads before creating a task

this brings parity between direct hook invocation and pulse-triggered
events. triggerSchema enforces both paths, and if hook payload fails
validation, no tasks will be created.

If hook was created with default triggerSchema (additionalProperties:
false) it would likely fail now, hook owners would need to update it

Fixes https://github.com/taskcluster/taskcluster/issues/8867

this is a breaking change, and there are some hooks that needs closer look before this goes live
high risk candidates are hg-push/* hooks

image
image
image


changelog/issue-8867.md
```diff
@@ -0,0 +1,9 @@
+audience: users
+level: major
+reference: issue 8867
+---
+Pulse-triggered hooks now validate matching pulse message payloads against the hook's `triggerSchema` before creating a task.
+If a pulse message matches the hook's `bindings` but fails `triggerSchema` validation, the message is discarded and no task is created.
+This is a breaking change: previously `triggerSchema` was only enforced on the API and webhook paths, and pulse messages fired the hook regardless of their payload.
+Validation is unconditional, including for hooks that did not set a `triggerSchema`: the default schema only accepts an empty payload, so such a hook will no longer fire on pulse messages that carry a payload.
+Before upgrading, review all hooks with pulse `bindings` and make sure each `triggerSchema` accepts the pulse payloads that should still create tasks, for example by allowing expected properties or using `additionalProperties: true`.
```
petemoore: I'm not sure setting `additionalProperties: true` is good advice here, because if there is a schema, and the validation is failing because there are additional properties, the fix is probably to fix/investigate rather than ignore it.

services/hooks/schemas/v1/create-hook-request.yml
```diff
@@ -32,6 +32,14 @@ properties:
...
     type:                 object
   triggerSchema:
+    description: |
+      JSON Schema used to validate payloads supplied when firing the hook.
+      API calls and webhooks are rejected when their payloads fail validation.
+      Pulse messages that match the hook's bindings are discarded without
+      creating a task when their payloads fail validation. When omitted, this
+      defaults to a schema that only accepts an empty payload, so a hook
+      without an explicit triggerSchema will not fire on Pulse messages that
+      carry a payload.
```
petemoore, line 42: This default I think is the opposite of what we want. If you don't supply a schema, anything should be allowed. I think this will break everything that doesn't have a triggerSchema, and triggerSchema is optional, but effectively if you didn't supply it, you wouldn't get any hooks firing.

services/hooks/schemas/v1/hook-definition.yml
```diff
@@ -17,6 +17,11 @@ properties:
       a task definition that is submitted to the Queue service.
     type:                 object
   triggerSchema:
+    description: |
+      JSON Schema used to validate payloads supplied when firing the hook.
+      API calls and webhooks are rejected when their payloads fail validation.
+      Pulse messages that match the hook's bindings are discarded without
+      creating a task when their payloads fail validation.
```
petemoore, line 24: Do we need two different definitions of triggerSchema? Can they be shared. They differ here in detail, it would be good for them to be aligned and from the same source.

services/hooks/src/monitor.js
```diff
@@ -21,3 +21,25 @@ MonitorManager.register({
...
+    firing the hook. This log type is sampled to avoid being noisy.
+  `,
+  fields: {
+    hookGroupId: 'The group ID of the hook that matched the pulse message',
+    hookId: 'The ID of the hook that matched the pulse message',
+    exchange: 'The pulse exchange that delivered the message',
+    routingKey: 'The pulse routing key that delivered the message',
+    reason: 'The reason the pulse message was discarded',
+    validationError: 'The triggerSchema validation error that caused the discard',
+    discardedCount: 'Number of pulse messages discarded by this listener for this reason since it was created',
```
petemoore, line 42: How many listeners are there? Does these totals get aggregated somewhere?

services/hooks/test/listeners_test.js
```diff
@@ -361,6 +372,111 @@ helper.secrets.mockSuite(testing.suiteName(), [], (mock, skipping) => {
       ]);
     });
 
+    test('discards pulse message when hook has the default triggerSchema', async () => {
```
petemoore, line 375: I think this is the opposite of what we want to do.

ui/docs/reference/core/hooks/firing-hooks.mdx
```diff
@@ -101,6 +106,12 @@ task payload is rendered with the following context:
 }
 ```
 
+To debug why a Pulse-triggered hook is not firing, open the hook in the UI and
+click **Debug bindings**. The drawer listens on the hook's saved bindings and
+shows, for each incoming Pulse message, whether its payload passes the hook's
+`triggerSchema` or is rejected (with the exact validation error). This requires
+the `web:read-pulse` scope.
```
petemoore, line 113: Nice docs, thanks!

ui/docs/reference/core/hooks/README.mdx
```diff
@@ -22,7 +22,9 @@ By default, the new task has a `taskGroupId` equal to its `taskId`, as is the co
 
 ### Trigger Schema
 
-When a hook is triggered by an API call or a webhook, the user-provided payload is validated against the hook's `triggerSchema`, and the call rejected if validation fails.
+When a hook is triggered by an API call, webhook, or Pulse message, the provided payload is validated against the hook's `triggerSchema`.
+API calls and webhooks are rejected if validation fails; matching Pulse messages are discarded without creating a task.
+A hook that does not define a `triggerSchema` uses the default schema, which only accepts an empty payload, so such a hook will not fire on Pulse messages that carry a payload.
```
petemoore, line 27: Same here (we should not have schema enforcement if triggerSchema not supplied).

petemoore (changes requested):

services/hooks/schemas/v1/create-hook-request.yml
lotas: yes, but default was `additionalProperties:false` before, so every hook saved as is would inherit this:
https://github.com/taskcluster/taskcluster/blob/f5e69d1813269b0071c4167bcd09a4e961a1285d/services/hooks/schemas/v1/create-hook-request.yml#L44

Are you proposing we flip the default in schema now to `true` now?

changelog/issue-8867.md
lotas: agree, we can drop this recommendation

services/hooks/schemas/v1/hook-definition.yml
lotas: ah, you mean different descriptions? 
I think create-hook one contains extra bit of info about the default value. Is this what you think should be aligned?

services/hooks/src/monitor.js
lotas: there is 1 listener per hook, which listens to all selected exchanges.

There is this counter also  which should log (in theory) total count: https://github.com/taskcluster/taskcluster/blob/f5e69d1813269b0071c4167bcd09a4e961a1285d/services/hooks/src/listeners.js#L101

services/hooks/test/listeners_test.js
lotas: yes, but we didn't change the default so far. and default was to not allow any additional properties

ui/docs/reference/core/hooks/README.mdx
lotas: the problem is that all existing hooks got created with the default value. so we can't change it without a migration. and even if we would, it would probably break all the hooks that created with the exact same schema 🤔 .. because there is no way to differentiate from the default value at this point

ui/docs/reference/core/hooks/firing-hooks.mdx
lotas: we could include that cron triggered hooks don't need triggerSchema at all?

services/hooks/schemas/v1/create-hook-request.yml
lotas: https://github.com/taskcluster/taskcluster-hooks/pull/66  that was introduced to say that payload should be an empty object {}

changelog/issue-8867.md
lotas: or use the ui debugger

lotas: "we did a review of community/fxci hooks with pulse bindings and trigger schema, and they seem to not contain the {additionalProperties: false}"

petemoore (approved):
Looks great, thanks Yarik!

PR #8916, opened 2026-07-24
Bug 2056592 - Fix auth role modification silently ignoring write conflicts

Eijebong:
`modifyRoles` placed a `return` within its retry loop right after the `try/catch` block that handles whether or not it should retry which means it never did. That means that if it ever had conflicted (P0004 from the database), it would throw in `db.fns.modify_roles`, ignore the error in the `catch`, and then return as if it succeeded instead of looping and trying again. Which means that a role that got removed or modified would stay as it was before but reported to the client as having changed. On top of the state not changing, the caller for `modifyRoles` inserts an audit event that now might not match what actually happened (or didn't).

While not ideal, in practice this is fairly meaningless since the state isn't cached but re-read from database all the time. So any role modification that would've hit this didn't get modified/deleted but stayed as is in database, which means they would just not show as modified/deleted in the UI/API and since all modifications are made through tooling that resolves states based on differences, it would resolve eventually.

The bug was introduced in 3fd060c69c265ecbdf1ef6e6690906dd453c2f5b when the roles moved from the azure style entity tables to a regular postgres one and this commit reverts the behavior to the previous one with the exception that it returns a 500 instead of a 409 when it can't resolve a conflict. Fixing that is unfortunately not entirely trivial and is probably not worth the hassle. I'd rather have it start failing loudly if it does and try fixing the conflict source rather than forcing clients to retry things in the hope that it eventually maybe resolves when stars align.

### DIFF
--- a/changelog/bug-2056592.md
+++ b/changelog/bug-2056592.md
@@ -1,0 +1,6 @@
+audience: users
+level: patch
+reference: bug 2056592
+---
+Fixed a bug in the auth service where a concurrent modification conflict while
+creating, updating, or deleting a role was silently reported as success
--- a/services/auth/src/data.js
+++ b/services/auth/src/data.js
@@ -17,13 +17,13 @@ export const modifyRoles = async (db, modifier) => {
       });
       await modifier({ roles });
       await db.fns.modify_roles(JSON.stringify(roles), etag);
+      return; // success!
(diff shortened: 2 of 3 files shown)


lotas (approved):
Nicely spot!

PR #8917, opened 2026-07-24
Bug 2056597 - Fix worker panic when declaring a directory artifact with an absolute path

Eijebong:
A task declaring a `directory` artifact with an absolute path was triggering a panic during artifacts collection because we ended up calling `filepath.Rel(absPathDeclaredInTask, relPathFromTaskDir)` which errored out since it makes no sense to try and find the relative path between an absolute path and a relative path (relative to what?). The walk then treated that error as a bug and panicked with a message along the line of `Rel: can't make ../foo/bar relative to /tmp/foo`, taking down the worker process.

Instead we now resolve the original path to an absolute one (even when it's declared as relative in the task) and then use the absolute path given to us by the walk function to compute the relative path. Since the call always ends up being between 2 absolute paths, it cannot fail.

Because we're now using absolute paths we can also get rid of the first `subPath` calculation which funnily enough fixes a similar panic that would've been exposed by not panicking in the other place if the task directory and the artifact path were on different volumes as `Rel` fails to calculate cross volume relative paths as is somehow not described in filepath documentation. (see
https://cs.opensource.google/go/go/+/refs/tags/go1.26.5:src/path/filepath/path.go;l=205 if you want an explanation as to why it fails and enjoy reading stdlib internals instead of what should actually be documentation)

### DIFF
--- a/changelog/bug-2056597.md
+++ b/changelog/bug-2056597.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: bug 2056597
+---
+Fixed a worker panic when declaring a directory artifact as an absolute path
--- a/workers/generic-worker/artifact_feature.go
+++ b/workers/generic-worker/artifact_feature.go
@@ -186,13 +186,9 @@ func (atf *ArtifactTaskFeature) FindArtifacts() {
 				payloadArtifacts = append(payloadArtifacts, errArtifact)
 				break
 			}
+			absBasePath := fileutil.AbsFrom(taskDir, basePath)
 			walkFn := func(path string, d os.DirEntry, incomingErr error) error {
-				subPath, err := filepath.Rel(taskDir, path)
-				if err != nil {
-					// this indicates a bug in the code
-					panic(err)
-				}
-				relativePath, err := filepath.Rel(basePath, subPath)
+				relativePath, err := filepath.Rel(absBasePath, path)
 				if err != nil {
 					// this indicates a bug in the code
 					panic(err)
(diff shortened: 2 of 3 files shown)


petemoore (approved):
This is a great find! I tested locally without the patch and indeed hit the panic. Leaving here for posterity. Great fix!

```
panic: Rel: can't make ../../../../../../../../../../var/folders/yn/n8p0gq9d37v254psqgqt41480000gn/T/TestDirectoryArtifactUploadFromAbsolutePath2964584615 relative to /var/folders/yn/n8p0gq9d37v254psqgqt41480000gn/T/TestDirectoryArtifactUploadFromAbsolutePath2964584615

goroutine 140 [running]:
github.com/taskcluster/taskcluster/v102/workers/generic-worker.(*ArtifactTaskFeature).FindArtifacts.func1.1({0x4dad19a34380?, 0x1005ab520?}, {0x101354b88, 0x4dad19ba0430}, {0x0, 0x0})
	/Users/pmoore/git/taskcluster/workers/generic-worker/artifact_feature.go:198 +0x4a4
path/filepath.walkDir({0x4dad19a34380, 0x66}, {0x101354b88, 0x4dad19ba0430}, 0x4dad19ac7e80)
	/opt/homebrew/Cellar/go/1.26.5/libexec/src/path/filepath/path.go:311 +0x44
path/filepath.WalkDir({0x4dad19a34380, 0x66}, 0x4dad19ac7e80)
	/opt/homebrew/Cellar/go/1.26.5/libexec/src/path/filepath/path.go:401 +0x74
github.com/taskcluster/taskcluster/v102/workers/generic-worker.(*ArtifactTaskFeature).FindArtifacts.func1({{0x0, 0x0}, {0x0, 0x0}, {0x1406f400, 0xee1f55bbb, 0x0}, {0x4dad19b5c060, 0xe}, 0x0, ...})
	/Users/pmoore/git/taskcluster/workers/generic-worker/artifact_feature.go:237 +0x3bc
created by github.com/taskcluster/taskcluster/v102/workers/generic-worker.(*ArtifactTaskFeature).FindArtifacts in goroutine 47
	/Users/pmoore/git/taskcluster/workers/generic-worker/artifact_feature.go:244 +0xf0
exit status 2
FAIL	github.com/taskcluster/taskcluster/v102/workers/generic-worker	3.106s
```

PR #8918, opened 2026-07-24
Fix expired client purge recording a 'created' audit entry

Eijebong:
When purging an expired client, the code properly logged the client as expired through monitor audit events but was inserting the event as created in database.

The bug was introduced in 5a347f63e42543c5e8c0821ed5e17fcec834c1d7, probably due to a bad copy/paste, when the audit trail for purged clients was first added. So every client deleted by expiration since then has been recorded as having been created at the moment it was removed.

Unfortunately the test that was supposed to help catch this didn't because it called `assume` without any assumption added to it (`assume()` is a no-op unless called with a terminator like `.is.something`/`.includes`). The original commit probably meant `assume().includes()` but instead put the `.includes` inside of the `assume()` call, which didn't fail because `Array.includes` is also a thing and ended up resolving to `assume(false)` which did nothing.

### DIFF
--- a/changelog/HLbEt32VQKGZtDELPlw_Dg.md
+++ b/changelog/HLbEt32VQKGZtDELPlw_Dg.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+---
+Fixed a bug in the auth service where purging an expired client recorded the
+deletion in the audit history as `created` instead of `expired`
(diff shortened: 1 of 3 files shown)


lotas (approved):
whoopsie, good one! thanks

PR #8919, opened 2026-07-24
Bug 2057491 - Fix a typo in the scope resolver assume prefix regex

Eijebong:
This is pretty self explanatory, the original commit meant to use `?:` (non capturing group), used `:?` (optional `:`). This means that it allowed a bunch of strings starting with `:` as being the equivalent to `assume`, which, while not doing anything on their own are just a general footgun that we don't need.


petemoore (approved):
Amazing find! Indeed, the regexp was wrong.

PR #8920, opened 2026-07-24
Fix scope resolve tests misparsing roles.json and silently not doing anything

Eijebong:
Found this while working on the fix for bug 2057491. Since 3fd060c69c265ecbdf1ef6e6690906dd453c2f5b, the role ID column was renamed to `role_id` on the scopes side but was still `roleId` in the roles.json dump (and in the current API). Tests in scoperesolver_test rely on parsing that dump and creating scopes out of them to actually test something and ended up instead with 600 `assume:undefined` because javascript is a perfectly cromulent language that will happily destructure non existing fields on an object to `undefined` and then format it like it's a normal Tuesday.

The fix is trivial, I just map the old `roleId` to `role_id` to fix the mismatch between what the API returns and what the role builder expects.

### DIFF
--- a/changelog/D7LqsNyUT8i730yA-GtNug.md
+++ b/changelog/D7LqsNyUT8i730yA-GtNug.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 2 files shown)


lotas (approved):
oh boi.. thanks

PR #8921, opened 2026-07-24
Deduplicate our client from row building in scope resolver

Eijebong:
We're building clients from database rows in two places in the exact same way, extract that to a function. The reason I saw this is because one of those was calling `toString` on the access token while the other was leaving it as a `Buffer`. It ends up not mattering because the places it gets used doesn't differentiate between both (it's used to build HMACs only and both Buffer and utf-8 strings end up hashing the same).


petemoore (approved):
Nice refactor.

PR #8922, opened 2026-07-24
Make sure we throw the right error when client cert signatures have an invalid length

Eijebong:
`crypto.timingSafeEqual` throws when the inputs are not the same length which with a message that bubbles back up all the way to the user with `Input buffers must have the same byte length` instead of `ext.certificate.signature is not valid`. The length is known in advance so this doesn't change anything security wise but it changes the categorization we make for metrics in that case from `other` to `invalid_signature` and gives a better error message to the user.

### DIFF
--- a/services/auth/test/signaturevalidator_test.js
+++ b/services/auth/test/signaturevalidator_test.js
@@ -685,6 +685,23 @@ suite(testing.suiteName(), () => {
   );
 
   testWithTemp(
+    'invalid: signature too short for temp creds',
+    {
+      id: 'root',
+    },
+    (id, key, certificate) => {
+      certificate.signature = 'baguette';
+      return {
+        authorization: {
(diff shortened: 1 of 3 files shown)


services/auth/test/signaturevalidator_test.js
```diff
@@ -684,6 +684,23 @@ suite(testing.suiteName(), () => {
     failed('ext.certificate.signature is not valid')
   );
 
+  testWithTemp(
+    'invalid: signature too short for temp creds',
+    {
+      id: 'root',
+    },
+    (id, key, certificate) => {
+      certificate.signature = 'baguette';
```
lotas, line 693: 🥖 🥖 🥖 🥖

lotas (approved):
🥖

PR #8923, opened 2026-07-24
Fix regex capture typos

Eijebong:
Those are both the same typo as #8919 in different places. Turns out the database one did have an impact on what documentation gets generated.


petemoore (approved):
Amazing! Maybe we should just delete docs, as clearly nobody reads them 😆

db/schema.md
```diff
@@ -110,6 +103,7 @@ CREATE TABLE azure_queue_messages (
 );
 ALTER TABLE azure_queue_messages
     ADD CONSTRAINT azure_queue_messages_pkey PRIMARY KEY (message_id);
+CREATE INDEX azure_queue_messages_inserted ON azure_queue_messages USING btree (queue_name, inserted);
```
lotas, line 106: ha, we should probably drop this table

PR #8924, opened 2026-07-24
Fix the renderTaskclusterYml endpoint to return 400 instead of 500 on templating errors

Eijebong:
`InvalidInput` isn't what's expected to return a 400, `InputError` is. So the former fell back through and ended up as an unknown error -> 500.

Fixes sentry issue `TASKCLUSTER-SERVICES-COMMUNITYTC-EB`


lotas (approved):
Thank you

PR #8926, opened 2026-07-27
Close websocktunnel connections when a client misbehaves

Eijebong:
Since websockettunnel clients are all on the trusted side of the tunnel, they **MUST** respect the protocol, anything else is a bug in wst-client and should be fixed. Before this, any bug would just get swallowed silently by the tunnel. We now close the connection instead, making it visible. It shouldn't change anything since I'm sure the client is bug free (right?)


lotas (approved):
not very familiar with websocket protocol, but lgtm  😬

PR #8931, opened 2026-07-28
feat(github, web-server): Stop following reference redirect artifacts in services

lotas:
While fetching live log and actions.json services were following redirects even if artifact had 'reference' type.
That could lead to a possible SSRF where github service or web-server would blindly follow artifact redirect url.

This would only download managed artifacts (the ones that resolve to stored s3 artifact on queue side)

Bugzilla Bug: [2056618](https://bugzilla.mozilla.org/show_bug.cgi?id=2056618)

### DIFF
--- a/clients/client/src/download.js
+++ b/clients/client/src/download.js
@@ -26,6 +26,8 @@ const s3 = async ({ url, streamFactory, retryCfg }) => {
     let contentType = 'application/binary';
     try {
       const src = got.stream(url, { retry: { limit: 0 } });
+      // aborted request emits again after pipeline() already rejected, crashes if unhandled
+      src.on('error', () => {});
       src.on('response', res => {
         contentType = res.headers['content-type'] || contentType;
       });
@@ -58,6 +60,8 @@ const getUrl = async ({ object, name, resp, streamFactory, retryCfg }) => {
     try {
       responseUsed = true;
       const src = got.stream(resp.url, { retry: { limit: 0 } });
+      // aborted request emits again after pipeline() already rejected, crashes if unhandled
+      src.on('error', () => {});
       src.on('response', res => {
         contentType = res.headers['content-type'] || contentType;
       });
--- a/clients/client/src/download.js
+++ b/clients/client/src/download.js
@@ -8,7 +8,7 @@ import { clients } from './client.js';
 const makeRetryCfg = ({ retries, delayFactor, randomizationFactor, maxDelay }) => ({
   retries: retries === undefined ? 5 : retries,
   delayFactor: delayFactor === undefined ? 100 : delayFactor,
-  randomizationFactor: randomizationFactor === undefined ? randomizationFactor : 0.25,
+  randomizationFactor: randomizationFactor === undefined ? 0.25 : randomizationFactor,
   maxDelay: maxDelay === undefined ? 30 * 1000 : maxDelay,
 });
 
--- a/clients/client/src/download.js
+++ b/clients/client/src/download.js
@@ -12,6 +12,15 @@ const makeRetryCfg = ({ retries, delayFactor, randomizationFactor, maxDelay }) =
   maxDelay: maxDelay === undefined ? 30 * 1000 : maxDelay,
 });
 
+// got reports an HTTP status on `err.response.statusCode`, while Taskcluster client errors carry
+// it on `err.statusCode`; normalize onto the latter so callers need only check one place.
+const withStatusCode = err => {
+  if (err.statusCode === undefined && err.response?.statusCode !== undefined) {
+    err.statusCode = err.response.statusCode;
+  }
+  return err;
+};
+
 const s3 = async ({ url, streamFactory, retryCfg }) => {
   return await retry(retryCfg, async retriableError => {
     let contentType = 'application/binary';
@@ -122,46 +131,63 @@ export const download = async ({
   }
 };
 
-export const downloadArtifact = async ({
-  taskId,
-  runId,
-  name,
-  queue,
-  streamFactory,
-  retries,
-  delayFactor,
-  randomizationFactor,
-  maxDelay,
-}) => {
+// when managedOnly is set, a `reference` artifact is rejected before its URL is read.  The queue
+// resolves `link` artifacts server-side, so the storageType seen here is always the final one.
+const downloadArtifactImpl = async (
+  { taskId, runId, name, queue, streamFactory, retries, delayFactor, randomizationFactor, maxDelay },
+  managedOnly
+) => {
   const retryCfg = makeRetryCfg({ retries, delayFactor, randomizationFactor, maxDelay });
 
-  const artifact = await (runId === undefined
-    ? queue.latestArtifact(taskId, name)
-    : queue.artifact(taskId, runId, name));
+  try {
+    const artifact = await (runId === undefined
+      ? queue.latestArtifact(taskId, name)
+      : queue.artifact(taskId, runId, name));
 
-  switch (artifact.storageType) {
-    case 'reference':
-    case 's3': {
-      return await s3({ url: artifact.url, streamFactory, retryCfg });
+    switch (artifact.storageType) {
+      case 'reference':
+      case 's3': {
+        if (managedOnly && artifact.storageType === 'reference') {
+          const err = new Error(
+            'Refusing to download a `reference` artifact: its URL is supplied by the task.  Use ' +
+              '`downloadArtifact` if fetching a task-supplied URL is intended.'
+          );
+          err.code = 'ArtifactStorageTypeRejected';
+          err.storageType = 'reference';
+          throw err;
+        }
+
+        return await s3({ url: artifact.url, streamFactory, retryCfg });
+      }
+
+      case 'object': {
+        const object = new clients.Object({
+          rootUrl: queue._options._trueRootUrl,
+          credentials: artifact.credentials,
+        });
+        return await download({ name: artifact.name, object, streamFactory, ...retryCfg });
+      }
+
+      case 'error': {
+        const err = new Error(artifact.message);
+        err.code = 'ArtifactError';
+        err.reason = artifact.reason;
+        throw err;
+      }
+
+      default:
+        throw new Error(`Unsupported artifact storageType '${artifact.storageType}'`);
     }
-
-    case 'object': {
-      const object = new clients.Object({
-        rootUrl: queue._options._trueRootUrl,
-        credentials: artifact.credentials,
-      });
-      return await download({ name: artifact.name, object, streamFactory, ...retryCfg });
-    }
-
-    case 'error': {
-      const err = new Error(artifact.message);
-      err.reason = artifact.reason;
-      throw err;
-    }
-
-    default:
-      throw new Error(`Unsupported artifact storageType '${artifact.storageType}'`);
+  } catch (err) {
+    throw withStatusCode(err);
   }
 };
 
-export default { download, downloadArtifact };
+export const downloadArtifact = async options => await downloadArtifactImpl(options, false);
+
+// as downloadArtifact, but refuses `reference` artifacts, whose content is fetched from a URL of
+// the publishing task's choosing.  Callers reading artifacts of untrusted tasks should use this,
+// so that a task cannot direct them at an arbitrary URL reachable from their network position.
+export const downloadManagedArtifact = async options => await downloadArtifactImpl(options, true);
+
+export default { download, downloadArtifact, downloadManagedArtifact };
--- a/clients/client/src/download.js
+++ b/clients/client/src/download.js
@@ -145,9 +145,8 @@ const downloadArtifactImpl = async (
       : queue.artifact(taskId, runId, name));
 
     switch (artifact.storageType) {
-      case 'reference':
-      case 's3': {
-        if (managedOnly && artifact.storageType === 'reference') {
+      case 'reference': {
+        if (managedOnly) {
           const err = new Error(
             'Refusing to download a `reference` artifact: its URL is supplied by the task.  Use ' +
               '`downloadArtifact` if fetching a task-supplied URL is intended.'
@@ -156,7 +155,9 @@ const downloadArtifactImpl = async (
           err.storageType = 'reference';
           throw err;
         }
-
+        return await s3({ url: artifact.url, streamFactory, retryCfg });
+      }
+      case 's3': {
         return await s3({ url: artifact.url, streamFactory, retryCfg });
       }
 
--- a/services/github/src/handlers/requestArtifact.js
+++ b/services/github/src/handlers/requestArtifact.js
@@ -1,8 +1,31 @@
 import utils from '../utils.js';
 
-export function buildArtifactUrl(queueClient, { taskId, runId, artifactName }) {
-  const limitedQueueClient = queueClient.use({ authorizedScopes: [`queue:get-artifact:${artifactName}`] });
-  return limitedQueueClient.buildSignedUrl(limitedQueueClient.getArtifact, taskId, runId, artifactName);
+/**
+ * Build the message shown to the repository as a comment when an artifact cannot be fetched.
+ */
+function artifactErrorMessage(artifactName, err) {
+  const requiredScope = `queue:get-artifact:${artifactName}`;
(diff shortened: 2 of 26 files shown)


services/github/test/utils_test.js
```diff
@@ -531,4 +534,110 @@ suite(testing.suiteName(), () => {
...
+          consume: async stream => {
+            for await (const chunk of stream) {
+              return chunk.length;
+            }
+          },
+        })
+      );
+
+      assert.equal(partial, body.length);
+    });
```
Eijebong: This only passes because the write happens to only ever write one chunk. If the body was big enough to get split or was writing 2 chunks explicitly, flushing in between then that would throw instead of giving you a partial result.

clients/client/src/download.js
Eijebong: So, you didn't introduce this but backoff is actually broken with retryCfg if you pass an undefined randomizationFactor (or omit it) because this ternary is reversed. Passing undefined sets it to undefined, passing any other value sets it to 0.25.

That undefined then gets transformed into a NaN in `delay *= Math.random() * 2 * rf + 1 - rf`.

Since the github service is about to depend on this working instead of its own `throttleRequest`, you'll want to fix it.

services/web-server/src/api.js
```diff
@@ -123,44 +123,27 @@ builder.declare(
...
-      if (bytesRead > MAX_LOG_SIZE) {
-        return res.reportError('InputTooLarge', 'Log exceeds 200MB size limit', {});
+    try {
+      for await (const line of readTaskLogLines(log)) {
+        profileBuilder.addLine(line);
       }
-      profileBuilder.addLine(line);
+    } catch (err) {
+      if (err.code === 'LogTooLarge') {
+        return res.reportError('InputTooLarge', err.message, {});
```
Eijebong, line 152: Urgh I hate javascript. Load bearing strings like this are the *worst*... Nothing actionable here, I think it's correct but URGH

services/web-server/src/api.js
```diff
@@ -123,44 +123,27 @@ builder.declare(
...
         {}
       );
     }
 
-    // Check Content-Length if available
-    const contentLength = parseInt(response.headers.get('content-length') || '0', 10);
-    if (contentLength > MAX_LOG_SIZE) {
-      return res.reportError('InputTooLarge', 'Log exceeds 200MB size limit', {});
-    }
-
```
Eijebong, line 153: Bit sad to lose the early exit here but now that I'm thnking about it, I'm assuming that this actually never worked since content-encoding would've been `gzip` or whatever we were using and clients would strip it or be completely wrong anyway (since text compresses well, the actual logs would've needed to be gigabytes for this to trip).

services/web-server/src/api.js
Eijebong, line 109: Might be worth adding here that the task needs to be done? Or that the log needs to not be a reference if we care about ever supporting other worker implementations.

services/web-server/src/api.js
```diff
@@ -123,44 +123,27 @@ builder.declare(
...
-    }
+    const log = await openTaskLog({ queue, taskId });
 
-    if (!response) {
+    if (!log) {
       return res.reportError(
         'ResourceNotFound',
         'Could not fetch task log. The task may not have logs or they may have expired.',
         {}
       );
```
Eijebong, line 141: WDYT of something along the line of this instead?

```suggestion
if (!log) {
  const taskIsRunning = ['running', 'pending', 'unscheduled'].includes(status.state);
  return res.reportError('ResourceNotFound',
    taskIsRunning
      ? 'Profiling is only supported on resolved tasks. Wait for the task to finish and retry'
      : 'Could not fetch task log. The task may not have logs or they may have expired.', {});
}
```


I don't think I missed unresolved states, but I'd like to once again voice my hatred of javascript and load bearing strings

services/github/src/handlers/requestArtifact.js
```diff
@@ -1,58 +1,69 @@
...
+  }
 }
 
 /**
  * Helper to request artifacts from statusHandler.
  */
 export async function requestArtifact(artifactName, { taskId, runId, debug, instGithub, build }) {
   try {
-    const url = buildArtifactUrl(this.queueClient, { taskId, runId, artifactName });
-    const res = await utils.throttleRequest({ url, method: 'GET' });
```
Eijebong, line 14: This was the only user of `throttleRequest` so we can probably remove it?

services/github/src/utils.js
```diff
@@ -1,11 +1,89 @@
...
+    taskId,
+    runId,
+    name: artifactName,
+    queue: limitedQueueClient,
+    streamFactory,
+    retries,
+  });
+};
+
+export const downloadArtifactAsText = async ({ queueClient, taskId, runId, artifactName }) => {
```
Eijebong, line 28: Isn't this `downloadArtifactToBuffer(...).toString()`?

clients/client/src/download.js
```diff
@@ -122,46 +131,63 @@ export const download = async ({
...
-  retries,
-  delayFactor,
-  randomizationFactor,
-  maxDelay,
-}) => {
+// when managedOnly is set, a `reference` artifact is rejected before its URL is read.  The queue
+// resolves `link` artifacts server-side, so the storageType seen here is always the final one.
+const downloadArtifactImpl = async (
+  { taskId, runId, name, queue, streamFactory, retries, delayFactor, randomizationFactor, maxDelay },
+  managedOnly
```
Eijebong, line 142: Any reason to make this positional? Forcing it to be named avoid having a random boolean attached to it that you have to follow to the definition of the function to figure out what it means.

clients/client/src/download.js
```diff
@@ -122,46 +131,63 @@ export const download = async ({
...
+      case 's3': {
+        if (managedOnly && artifact.storageType === 'reference') {
+          const err = new Error(
+            'Refusing to download a `reference` artifact: its URL is supplied by the task.  Use ' +
+              '`downloadArtifact` if fetching a task-supplied URL is intended.'
+          );
+          err.code = 'ArtifactStorageTypeRejected';
+          err.storageType = 'reference';
+          throw err;
+        }
```
Eijebong: This reads very awkwardly with the outer switch. You can enter this arm for both types and then check again. I'd extract the whole check out of it tbh

Eijebong (changes requested):

clients/client/src/download.js
lotas: what a sharp eye! I'll flip it here

clients/client/src/download.js
lotas: the only reason was to  have two distinct exported methods instead. I can also keep the old version and just add a named option I guess

clients/client/src/download.js
lotas: yeah, I missed that. will move managed check inside the 'reference' case instead to make it explicit

services/github/src/handlers/requestArtifact.js
lotas: well spotted! Indeed, it was only used for artifact

services/github/src/utils.js
lotas: it is, but `downloadArtifactToBuffer` lives in the `web-server` part of the codebase.. Probably would be good to place it into shared library.
For consistency I'll name them both the same

lotas: actually they are a bit different, so I'd leave it as is for now

services/github/test/utils_test.js
lotas: Ah, indeed.. sending bigger buffer forces chunking and it throws, good catch.
Looks like if got request didn't have an on'error' handler, everything would rethrow an error after client stops consuming. adding a fix

services/web-server/src/api.js
lotas: are you trying to say that you want to rewrite it in Typescript?

Eijebong: :crab:

lotas: maybe .. I've heard bun was rewritten from zig to rust _for only $150k_ worth of tokens

services/web-server/src/api.js
lotas: oh indeed

services/web-server/src/api.js
lotas: Agree. I'll adjust wording. I also just realized we never specify runId.. so if task has several we only check latest.. probably fine for now

lotas: I also realized we could do this check earlier and not even try to read the logs in the first place, since we won't be able to read them

services/github/src/handlers/requestArtifact.js
```diff
@@ -1,58 +1,69 @@
...
+    return prefix.concat('Make sure the artifact exists on the worker or other location.');
+  }
+
+  switch (err.statusCode) {
+    case 403:
+      return prefix.concat(
+        `The GitHub service does not have permission to read this artifact. Make sure the artifact is public or grant the GitHub service the \`${requiredScope}\` scope.`
+      );
+    default:
+      return err.message ? prefix.concat(err.message) : prefix;
```
Eijebong: This is the kind of stuff that's somehow making me think that go errors are sane...

Shouldn't this branch not shove the exception in a comment that's about to be public? I can see a transient DNS error leaking internal hostnames for example (not that it super matters but 🤷). The previous behavior was   to just ignore any exception that didn't have a `error?.message` entirely.

Eijebong: After much further reading, I *think* this is worse than I originally thought.

There are a "few" paths along the way of `utils.downloadArtifactAsText` that can throw.

1) `const artifact = await (runId === undefined ? queue.latestArtifact(taskId, name) : queue.artifact(taskId, runId, name));` at `clients/client/src/download.js:147`
2) `clients/client/src/download.js:165` through `s3(...);`
  2a) `const dest = await streamFactory();` at `clients/client/src/download.js:34`
  2b) `await pipeline(src, dest);` at `clients/client/src/download.js:35`, carrying errors from `got.stream(url)` at `clients/client/src/download.js:28` and from `dest`
  2c) `throw err;` at `clients/client/src/download.js:40`, the fatal exit for an `HTTPError` under 500
  2d) `throw retriableError;` at `clients/client/src/retry.js:25`, the exit for everything else once retries are exhausted

3) `clients/client/src/download.js:173` through `download(...)`
  3a) `const object = new clients.Object({...});` at `clients/client/src/download.js:169`
  3b) `const resp = await object.startDownload(name, { acceptDownloadMethods });` at `clients/client/src/download.js:129`
  3c) `throw new Error('Could not negotiate a download method');` at `clients/client/src/download.js:134`
  3d) `resp = await object.startDownload(name, {...});` at `clients/client/src/download.js:56`, on URL renewal
  3e) `const dest = await streamFactory();` at `clients/client/src/download.js:68`
  3f) `await pipeline(src, hashStream, dest);` at `clients/client/src/download.js:70`, carrying errors from `got.stream(resp.url)` at `clients/client/src/download.js:62`
  3g) `throw err;` at `clients/client/src/download.js:76`, the fatal exit for an `HTTPError` under 500
  3h) `throw retriableError;` at `clients/client/src/retry.js:25`, the exit for everything else
  3i) `verifyHashes(observedHashes, resp.hashes);` at `clients/client/src/download.js:85`, throwing at `:101` on mismatch or `:110` when no acceptable algorithm is present
  
4) `clients/client/src/download.js:160` which throws a handcrafted error
5) `clients/client/src/download.js:180` which throws an error containing the artifact errors from the queue service
6) `clients/client/src/download.js:184` which throws a handcrafted error with no code (although it should pretty much be unreachable).

---

A few of those are ok:

- 1 is perfectly fine. They're the same errors you'd get if you called the APIs.
- 4 is fine, it's a handcrafted error, we special case it in the description to give a more relevant error message to the user, bit duplicated but fine.
- 5 is fine today because you're special casing `ArtifactError`, otherwise it'd include the artifact creation message details straight from database (see `services/queue/src/artifacts.js:1002`, you can follow details straight up to the database fetch, untransformed, which I think would be fine but AFAICT it's task controlled so not great either)
- 6 is fine, it should be unreachable anyway

The remaining 2 aren't:

- 2a is fine, the only call uses a  factory that's ours (`services/github/src/utils.js:36`) and can't throw. But it's a callback and nothing prevents it from throwing anything
- 2b is the main problem on this branch. We're using got 15 which builds `HTTPError.message` as `Request failed with status code ${statusCode} (${statusMessage}): ${method} ${stripUrlAuth(url)}` (`clients/client/node_modules/got/dist/source/core/errors.js:74`), and `stripUrlAuth` only clears username/password, so the query string survives. If in a private bucket or if the public bucket uses signed URLs, we're leaking those. Transport errors go here too, which might leak internal hostnames/IPs. Neither carries `.body`/`.code` which are what we match on, so both land in the default branch and `services/github/src/handlers/requestArtifact.js:27` publishes the message straight up to the commit.
- 2c and 2d are both fine. Worth noting however 2d contains the >=500 case, which `services/github/src/handlers/requestArtifact.js:59` excludes from the monitor report while still posting the comment. So we're publicly reporting the error but not internally. I *think* it's fine since the user is much more interested in "stuff didn't work because AWS dies" than we are but yeah...

- 3a is fine, only a misconfiguration throws
- 3b and 3d are fine, same as 1, except they come from the object service and not  the queue.
- 3c is fine, handcrafted, no code.
- 3e is fine, same as 2a.
- 3f is the main problem here. It's the same thing as 2b, except the URL is the object service's signed download URL this time.
- 3g and 3h are the fine, very similar to 2c and 2d.
- 3i is fine, but it has no `code`/`statusCode` and more importantly, no `body`, so it looks exactly like a transport error which is painful

---

Now because of 2b and 3f, we don't want to comment errors verbatim. Because 3i, we cannot rely on `body` to be present (it's only present in TC services errors because of `clients/client/src/client.js:424`, which would be a fine way to only ever report those in messages and let the rest ride a "something went wrong path").

It's a clusterfuck (pardon my french), and my head hurt by now. What I'd suggest is something along the line of this:

```js
switch (err.code) {
  case 'ArtifactStorageTypeRejected': return prefix.concat('Something something reference artifacts are not fetched');
  case 'ArtifactError': return prefix.concat('Make sure the artifact exists on the worker or other location.');
}

// Only errors from the API contain `err.body`
if (err.body?.message) {
  return err.statusCode === 403
    ? prefix.concat(`Something something \`${requiredScope}\` scope.`)
    : prefix.concat(err.body.message);
}

return prefix.concat('Internal error. The artifact could not be downloaded');
```

We lose 3i to the bottom generic arm here but the alternative is going into download.js and slap codes everywhere...

I would also always report it internally despite what I said earlier with it being fine. Because the generic arm completely loses what the actual issue is and we'd have no way of correlating that to anything.

Sorry for the wall of text

Eijebong (changes requested):
Can't find a way to add a comment to a file that you didn't change in the PR...

On `ui/src/views/Tasks/ViewTask/index.jsx:1081`, wdyt about adding somehting like `FabProps={{ disabled: ['unscheduled', 'pending', 'running'].includes(task.status.state) }}` so the profiler button is disabled until the task is resolved? Could even change the tootltipTitle based on the same state to be "Profile Task Log (wait until it's done)" or whatever wording.

services/github/src/handlers/requestArtifact.js
Eijebong: image

lotas: oh wow, excellent sleuthing, thanks again for leading the investigation and review 😬

lotas:
> Can't find a way to add a comment to a file that you didn't change in the PR...
> 
> On `ui/src/views/Tasks/ViewTask/index.jsx:1081`, wdyt about adding somehting like `FabProps={{ disabled: ['unscheduled', 'pending', 'running'].includes(task.status.state) }}` so the profiler button is disabled until the task is resolved? Could even change the tootltipTitle based on the same state to be "Profile Task Log (wait until it's done)" or whatever wording.

great idea! Adding this to match backend behavior

services/github/src/handlers/requestArtifact.js
lotas: I will include your proposed corrections, and also reporting error internally now always too

Eijebong (approved):

PR #8941, opened 2026-07-30
Fix claimWork's long poll timers, clear them and stop failing to pool them

Eijebong:
`claimWork` arms a 20s timer for each request, bound to its `accept` which is the base of how long polling works in the queue. That timer was never cleared, even when long polling did return a task early. Clear it in a `finally` block so it's always released. This avoids having to duplicate the release between return paths and has the benefit of releasing the timer if the handler ever throws.

I simplified the whole timer handling by removing the `sleep20Seconds` hack which was supposed to pool timers by creating new ones every 2s. Except that it never did because it never set `_lastTime` so `time - _lastTime > 2000` was always true. That behavior was added in 290861a46d63a49d06d4975c78855f0e4bb5c515 which comes from an unreviewed PR, 9 years ago where there's no justification for doing this except "reduce number of timers". No numbers, no profiling. Given this never worked I'm guessing that timers were never an issue and that this was at best guess work. It also did the exact opposite of what it says because it removed a `clearTimeout` that was doing what I'm adding in this commit.

As such, I didn't even try to fix pooling and instead explicitly give each connection its own timer. This keeps the current behavior but does so in a way that doesn't lie about intention.

Funnily enough, this came up in 8b6799c67d4c97a62cdd67ea65b3f363f9edac38 but I didn't see it back then when I made `_lastTime` const which was a clear indicator that something was wrong.

This commit diff is better viewed by ignoring whitespaces

### DIFF
--- a/changelog/fBw2096USZq5PF1eW2y1-A.md
+++ b/changelog/fBw2096USZq5PF1eW2y1-A.md
@@ -1,0 +1,2 @@
+level: silent
+---
--- a/services/queue/src/api.js
+++ b/services/queue/src/api.js
@@ -1439,17 +1439,7 @@ builder.declare(
   }
 );
 
-// Hack to get promises that resolve after 20s without creating a setTimeout
-// for each, instead we create a new promise every 2s and reuse that.
-const _lastTime = 0;
-let _sleeping = null;
-const sleep20Seconds = () => {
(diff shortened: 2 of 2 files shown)


petemoore (approved):
Thanks Bastien, nice cleanup.

Issue #8945, opened 2026-07-30
Garbage collection blocks task mounts while removing caches


Eijebong:
`grbageCollection` takes `cacheMutex` and holds it via defer across `trimPoolExcess()` and the entire directory cache eviction loop which prevents mounts from happening at the same time. This is only blocking on capacity > 1.

Note that there's an apparent obvious fix, which is to just release the lock between evictions, but it's unsafe. `SortedResources` filters InUse, so releasing the lock between evictions lets a task `AcquireCache` an entry from the snapshot and start mounting it before the collector deletes it.

`trimPoolExcess` has the exact same problem.

Issue #8944, opened 2026-07-30
A failed cache eviction leaks disk space permanently


Eijebong:
Fixing #8942 means making cache eviction failures non fatal. Unfortunately, the easy and obvious fix here (evicting on disk before removing from the map) cannot be applied that easily, a few call sites rely on the eviction being done unconditionally and the ones that evict a broken entry out of the pool would leave it sitting there, marked in use.

Issue #8943, opened 2026-07-30
Garbage collection can delete a file cache a running task is using with capacity > 1


Eijebong:
`FileCacheMap.SortedResources` returns every entry, (unlike `CacheMap.SortedResources` which filters `!cache.InUse`). Nothing ever tracks file cache usage. 

So with capacity > 1 the main loop collector can delete a downloaded archive while another task is mounting it which would lead to the task dying if it was about to extract it for example.

PR #8946, opened 2026-07-30
Fix a bunch of garbage collection related regressions from the capacity > 1 PR

### LINKED ISSUE #8942: Workers keep checking disk space and trying to GC while a task is running since v100.0.0
This is a regression from #8278 where it went from doing:

- check disk space
- wait for claim
- run
- **wait here for completion**
- cleanup
- goto "check disk space"

To

- Check disk space (crash if we can't get below the required minimum, which we can't because the task is currently using the disk space)
- try claim
- if claim: queue run (non blocking)
- goto check disk space

---

Multiple issues here:

We shouldn't be trying to cleanup disk space to claim a task if we don't have capacity to claim one. This completely breaks the worker if a task ends up using enough disk during its execution to go below the minimum.

We definitely shouldn't be panicking if we can't get enough disk space from GC **while a task is still running**

We shouldn't be logging disk space in the function that measures it. It's spamming logs every 5s.

Eijebong:
I was already pretty unhappy about the state of a few things post #8278 and this didn't change my mind at all. I had my eyes very much set on `git revert` while unwinding this whole garbage collection stuff...

I would recommend reviewing this commit by commit as each of them fixes something different. I added changelog entries for changes that are relevant to users/worker deployers.

Note that while fixing these, I found 3 more regressions from that one PR that I didn't have the courage to fix. (#8943, #8944 and #8945).

There's one commit in here that fixes a bug that predates #8278, I included it in the same stack because I was touching relevant code anyway and it's a very small change.

Fixes #8942

### DIFF
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -611,23 +611,35 @@ mainLoop:
 			}
 		}
 
+		canClaimTask := claimCount > 0
+
+		// Ensure there is enough disk space *before* claiming a task.
+		if canClaimTask {
+			// Pass tasksRunning so docker prune is skipped when tasks may
+			// have loaded images that are not yet running in a container.
+			tasksRunning := !taskManager.IsIdle()
+			err := garbageCollection(tasksRunning)
+			if err != nil {
+				if !tasksRunning {
+					// If we're not running any task and have no way to claim
+					// one, something is wrong, panic
+					panic(err)
+				}
+
+				log.Printf("Not claiming any task: %v", err)
+				canClaimTask = false
+			}
+		}
+
 		// Make sure at least 5 seconds pass between tcqueue.ClaimWork API
 		// calls. Only back off if we could actually claim a task during that
 		// cycle.
 		var claimBackoff <-chan time.Time
-		if claimCount > 0 || taskManager.IsIdle() {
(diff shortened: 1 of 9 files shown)


workers/generic-worker/main.go
```diff
@@ -611,23 +611,35 @@ mainLoop:
...
+		canClaimTask := claimCount > 0
+
+		// Ensure there is enough disk space *before* claiming a task.
+		if canClaimTask {
+			// Pass tasksRunning so docker prune is skipped when tasks may
+			// have loaded images that are not yet running in a container.
+			tasksRunning := !taskManager.IsIdle()
+			err := garbageCollection(tasksRunning)
+			if err != nil {
+				if !tasksRunning {
```
lotas, line 623: I wonder if it is possible for this state to change during `garbageCollection` loop, since it's taking "some time" to finish? 
Does it need an extra idle check here, or tasks cannot be claimed in between? (not familiar how it works)

Eijebong: It can change yes, but only from running -> idle, not the other way around. So it can never change in a way where we'd panic when we shouldn't. Worst case a task resolves in between and we skip the panic on this round, wait 5s and try again.

lotas (approved):
LGTM

PR #8948, opened 2026-07-31
feat(w-m): worker manager refuses to refresh credentials when worker is not running or expired

### LINKED ISSUE #8947: reregisterWorker call ignores worker state and expiry
worker manager `reregisterWorker` ignores state and expiry of a worker

`registerWorker` enforces expiry and provider state==REQUESTED but `reregisterWorker` checks only row-existence + secret match.

lotas:
In some edge cases it was possible for a worker to be stopping/stopped/expired and still request refreshed credentials.

Fixes #8947

### DIFF
--- a/services/worker-manager/src/api.js
+++ b/services/worker-manager/src/api.js
@@ -1294,6 +1294,14 @@ builder.declare(
       return res.reportError('InputError', 'Could not generate credentials for this secret', {});
     }
 
+    if (worker.state !== Worker.states.RUNNING) {
+      return res.reportError('InputError', `Worker ${workerGroup}/${workerId} is not running`, {});
+    }
+
+    if (worker.expires < new Date()) {
+      return res.reportError('InputError', `Worker ${workerGroup}/${workerId} has expired`, {});
+    }
+
     // worker has not been registered yet
     if (!worker.secret) {
       return res.reportError('InputError', 'Could not generate credentials for this secret', {});
--- a/services/worker-manager/src/api.js
+++ b/services/worker-manager/src/api.js
@@ -1294,14 +1294,6 @@ builder.declare(
       return res.reportError('InputError', 'Could not generate credentials for this secret', {});
     }
 
-    if (worker.state !== Worker.states.RUNNING) {
-      return res.reportError('InputError', `Worker ${workerGroup}/${workerId} is not running`, {});
-    }
-
-    if (worker.expires < new Date()) {
-      return res.reportError('InputError', `Worker ${workerGroup}/${workerId} has expired`, {});
-    }
-
     // worker has not been registered yet
     if (!worker.secret) {
       return res.reportError('InputError', 'Could not generate credentials for this secret', {});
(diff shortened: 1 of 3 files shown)


services/worker-manager/src/api.js
```diff
@@ -1294,6 +1294,14 @@ builder.declare(
...
     }
 
+    if (worker.state !== Worker.states.RUNNING) {
+      return res.reportError('InputError', `Worker ${workerGroup}/${workerId} is not running`, {});
+    }
+
+    if (worker.expires < new Date()) {
+      return res.reportError('InputError', `Worker ${workerGroup}/${workerId} has expired`, {});
+    }
+
```
Eijebong: I'd move this *below* the secret check below? Right now this would allow probing worker existence/state (not that it really matters but given why we're doing this, might as well reduce future noise)

Eijebong: (Errors are opaque on purpose, and I think it's fine to have them non opaque after the secret's been checked. Not before, which validates that point)

services/worker-manager/src/api.js
```diff
@@ -1294,6 +1294,14 @@ builder.declare(
       return res.reportError('InputError', 'Could not generate credentials for this secret', {});
     }
 
+    if (worker.state !== Worker.states.RUNNING) {
+      return res.reportError('InputError', `Worker ${workerGroup}/${workerId} is not running`, {});
+    }
+
+    if (worker.expires < new Date()) {
```
Eijebong: This actually exposes some bug, which probably doesn't matter much, but it might be worth having a better look next week.

`expires` is hardcoded to 1 week here (and in every other provider, just taking google as the example):

https://github.com/taskcluster/taskcluster/blob/85c75bc86bf9001d154b363fbdff814445d0fdaa/services/worker-manager/src/providers/google.js#L369

It's never written back to except for the static provider:

https://github.com/taskcluster/taskcluster/blob/85c75bc86bf9001d154b363fbdff814445d0fdaa/services/worker-manager/src/providers/static.js#L49

The `expireWorkers` cronjob will clear those rows which would fail on the `!worker` above before when a worker would re-register for the first time after a week (+ 1 day for the cronjob to run) modulo their `reregistrationTimeout`.

Now they'll fail to re-register immediately on the first re-registration after a week, even if the cronjob has not run yet.

Both behaviors are wrong and completely ignore the `providerData.terminateAfter` that is supposed to be the worker's lifetime.

AFAICT, unless worker-runner decides to stop, currently if we hit that `expires` and the cronjob runs, the worker continues running until its next reregistration without worker-manager even knowing about it, which means it can't loop over it and shut it down.

services/worker-manager/src/api.js
lotas: Alright, yes, I'm not sure if it's worth to keep the same error message for all checks.. It's not like the worker api could expose same information for free already

Eijebong: Yeah I think moving those checks below the secret check means that only a worker with a valid secret can check if it's still alive. Then you can keep those error messages.

services/worker-manager/src/api.js
lotas: You probably missed this: (worker-scanner is doing this job)
https://github.com/taskcluster/taskcluster/blob/42efb0e2a04c7049ace88647a52befb56c76bcf3/services/worker-manager/src/worker-scanner.js#L151-L159

scanner is checking all non stopped workers and updates its expires by +8 days if it's less than 7

Once it is fully stopped, it will be gone after that expires

Eijebong: I did miss that yes. Nevermind me then, thanks

Eijebong (approved):

PR #8971, opened 2026-08-03
add new websocket server to receive pulse messages

### LINKED ISSUE #9126: Add new websocket server to receive pulse messages

thakkarujas5:
Fixes #9126


ui/src/views/Tasks/ViewTask/index.jsx
```diff
@@ -224,22 +223,28 @@ export default class ViewTask extends Component {
       this.unsubscribe();
     }
 
-    const unsubscribe = subscribeToMore({
-      document: taskSubscription,
-      variables: {
-        taskId,
+    const unsubscribe = subscribeToNamedEvents(
```
lotas, line 226: I see method description says `queue` is default, but I'd rather not have it as default and instead made it required

plus add here `service: 'queue'` better be explicit and avoid unexpected behavior when defaults change

ui/src/utils/pulseListener.js
```diff
@@ -1,46 +1,133 @@
...
+  );
+
+/**
+ * Subscribe to Pulse events by name rather than by raw binding, via the
+ * /subscription/named endpoint: the server expands the named events into the
+ * matching exchanges, filtered by the routing key. Pass a non-empty
+ * `subscriptions` array of event names (e.g. `['taskDefined',
+ * 'taskCompleted']`) and a `routingKey` object of fields to match (e.g.
+ * `{ taskGroupId }`); omitted routing-key fields are wildcarded. `service`
+ * selects whose events the names refer to and defaults to 'queue'. Returns a
```
lotas, line 119: the problem is that this default is set on the web-server side.
Let's always require `service` and don't assign a default out of convenience 

fail early

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -1,89 +1,50 @@
...
+
+export default ({ cfg, server, pulseEngine, clients, authFactory, monitor }) => {
+  // One endpoint per subscription kind, differing only in how subscribe frames
+  // resolve to bindings: /subscription/raw takes pre-resolved exchange/pattern
+  // bindings (the Pulse debugger), /subscription/named takes event names the
+  // server resolves via the service events clients. Both speak the same frame
+  // protocol; separate endpoints let auth requirements differ per kind.
+  const endpoints = new Map([
+    ['/subscription/raw', resolveRawBindings],
+    ['/subscription/named', frame => resolveNamedBindings(frame, clients)],
```
lotas, line 17: I wonder if we could avoid having separate endpoints.

all clients expose `$ServiceEvents` in addition to `$Service` so we could always do something like:
```js
import { QueueEvents } from '@taskcluster/client-web';

subscribeToNamedEvents({
   subscriptions: [
         QueueEvents.taskDefined('#.taskGroupId.#'),
         QueueEvents.taskRunning('#.taskGroupId.#'),
 . ..

```

https://github.com/taskcluster/taskcluster/blob/84a0477a44a63d17cf1dc54b66250526e9d407eb/clients/client-web/src/clients/QueueEvents.js#L19-L23

services/web-server/src/servers/SubscriptionConnection.js
```diff
@@ -0,0 +1,265 @@
...
+      default:
+        this.sendError({ code: 'ProtocolError', message: `Unknown frame type: ${frame.type}` });
+    }
+  }
+
+  async handleConnectionInit(_frame) {
+    clearTimeout(this.connectionInitTimeout);
+
+    try {
+      // const credentials = frame.authorization ? decryptToken(frame.authorization) : null;
```
lotas, line 181: this is temporary right?

services/web-server/src/servers/resolveBindings.js
```diff
@@ -0,0 +1,79 @@
...
+//
+// A malformed frame throws an Error with `code: 'ProtocolError'`, which the
+// connection reports to the client without tearing down the socket.
+
+const protocolError = message => Object.assign(new Error(message), { code: 'ProtocolError' });
+
+const isPlainObject = value => typeof value === 'object' && value !== null && !Array.isArray(value);
+
+export const resolveRawBindings = ({ bindings }) => {
+  if (!Array.isArray(bindings) || bindings.length === 0) {
```
lotas, line 20: let's additionally pick up some arbitrary max number of subscriptions to allow, to avoid hacker one reports ;)

let's cap it at 100 bindings? I'm not sure we'd need more. Even

services/web-server/src/servers/resolveBindings.js
```diff
@@ -0,0 +1,79 @@
...
+]);
+
+export const resolveNamedBindings = ({ service = 'queue', subscriptions, routingKey }, clients) => {
+  const eventsService = EVENTS_SERVICES.get(service);
+
+  if (!eventsService) {
+    throw protocolError(`unknown events service: ${service}`);
+  }
+
+  if (!Array.isArray(subscriptions) || subscriptions.length === 0) {
```
lotas, line 51: same here, some top boundary please

services/web-server/src/servers/createSubscriptionServer.js
lotas: ah, I see those are used already below. nvm. this was just thinking outloud

services/web-server/src/servers/createSubscriptionServer.js
```diff
@@ -1,89 +1,50 @@
...
-                  new ErrorReply({
-                    code: 'InsufficientScopes',
-                    message: message,
-                    details: {
-                      required: ['web:read-pulse'],
-                    },
-                  })
-                );
-              }
+    wss.handleUpgrade(req, socket, head, ws => {
```
lotas, line 41: we need to make sure we only allow expected origins to subscribe
Something like this:
https://websocket.org/guides/security/#origin-validation-implementation 

`verifyClient()`

ui/src/utils/pulseListener.js
```diff
@@ -1,46 +1,133 @@
...
+    }
+  };
+
+  ws.onerror = () => {
+    if (!torn) {
+      onError(new Error('WebSocket connection error'));
+    }
+  };
+
+  ws.onclose = event => {
```
lotas, line 81: we probably want to reconnect to match existing functionality

ui/src/utils/pulseListener.js
```diff
@@ -1,46 +1,133 @@
...
+  // origin when relative (dev: proxied by Vite) and is used as-is when absolute
+  // (deployed: 'https://host/subscription').
+  const base = window.env?.GRAPHQL_SUBSCRIPTION_ENDPOINT || '/subscription';
+  const url = new URL(
+    `${base.replace(/\/$/, '')}/${endpointPath}`,
+    window.location.href
+  );
+
+  // http(s) -> ws(s); a same-origin relative path inherits the page protocol.
+  url.protocol = url.protocol === 'https:' ? 'wss:' : 'ws:';
```
lotas, line 21: minor. but someone might give url like:
```
GRAPHQL_SUBSCRIPTION_ENDPOINT=wss://events.example.com/subscription
```
and it would be downgraded to `ws`

safer to do a map check:

```js
  const websocketProtocols = {
    'https:': 'wss:',
    'http:': 'ws:',
    'wss:': 'wss:',
    'ws:': 'ws:',
  };

  url.protocol = websocketProtocols[url.protocol];
```

ui/src/utils/pulseListener.js
```diff
@@ -1,46 +1,133 @@
...
+        break;
+      case 'error':
+        onError(new Error(frame.message ?? 'Pulse subscription error'));
+        break;
+      default:
+        break;
+    }
+  };
+
+  ws.onerror = () => {
```
lotas, line 75: a failed connect fires `onerror` *and* `onclose(1006, wasClean:false)`, so `onError` is invoked twice for one failure.

services/web-server/src/servers/SubscriptionConnection.js
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    this.isAlive = false;
+    this.ws.ping();
+  }
+
+  send(frame) {
+    return new Promise((resolve, reject) => {
+      if (this.ws.readyState !== this.ws.OPEN) {
+        reject(new Error('Socket is not open'));
```
lotas, line 122: this is causing rabbitmq to nack and requeue the message. This can repeatedly redeliver same  message. 

it might be better to handle this in UI: unsubscribe and move on:
lines 241-244:

```js
 message =>
    this.send({
      type: FRAME_TYPES.DATA,
      subscriptionId,
      message,
    }).catch(err => {
      this.monitor.reportError(err);

      // The socket cannot receive further messages.
      this.pulseEngine.unsubscribe(subscriptionId);
      this.subscriptions.delete(subscriptionId);

      if (this.ws.readyState === this.ws.OPEN) {
        this.ws.close(CLOSE_CODES.INTERNAL_ERROR, 'WebSocket delivery failed');
      }

      // Intentionally don't rethrow: PulseEngine will ack rather than requeue.
    })
```

lotas (changes requested):
important:

1. authentication on connection and `web:read-pulse` check
2. resource limits - 100 per call
3. validate origin - verifyclient (was missing before but will be good)
4. `artifacts_graphql_test.js:55` test might be broken - calls removed helpers
5. closing after 6 hours without reconnection - pulseListener should reconnect if needed
6. ws/wss downgrade in pulseListener 
7. double onError/onClose error (needs exactly once termination)
8. ws.send() failure when socket is closing  to avoid nack() (subscriptionConnection:122)

chores:
lockfiles, changelogs, generate..

ui/src/utils/pulseListener.js
```diff
@@ -1,46 +1,194 @@
...
+    reconnectAttempts += 1;
+    reconnectTimer = setTimeout(connect, delay);
+  };
+
+  function connect() {
+    subscriptionId = null;
+    // A failed connect fires `onerror` *and* `onclose(1006)`; guard so a single
+    // failure reports through onError only once.
+    let failureReported = false;
+    ws = new WebSocket(getEventsWsUrl(endpointPath));
```
lotas, line 103: this is currently missing authentication part

The old client sent:
```
Authorization: `Bearer ${btoa(JSON.stringify(user.credentials))}`
```

lotas:
I think most of the comments are resolved, and I was able to successfully test it locally. 
Auth is passed correctly, disconnects on sign out, and manages both raw and named subscriptions correctly.

I only pushed a small fix to avoid crashing server with incorrect URL provided.

Great job! Thank you

lotas (approved):
All tested locally and reviewed, no critical issues found for this particular migration

PR #8979, opened 2026-08-04
fix: changes roles page to rest

### LINKED ISSUE #8978: Migrate Roles page from GraphQL to REST

thakkarujas5:
Github Issue: Fixes #8978


thakkarujas5:
In ViewRoles there were two components managing entire ViewRoles page, changed them to a single component, can revert it back to previous way if required.

ui/src/components/RolesTable/index.jsx
```diff
@@ -37,58 +28,29 @@ const iconSize = 16;
...
-    onPageChange: func.isRequired,
-    /** A search term to refine the list of roles */
+    roleIds: arrayOf(string).isRequired,
     searchTerm: string,
+    loading: bool,
+    page: number.isRequired,
+    hasNextPage: bool,
+    hasPreviousPage: bool,
+    onNextPage: func.isRequired,
+    onPreviousPage: func.isRequired,
```
lotas: since we will be tossing this navigation-related shapes around, maybe it's worth to add them to the `utils/prop-types.js`
so we'd only  `...pagination` or something

lotas:
> In ViewRoles there were two components managing entire ViewRoles page, changed them to a single component, can revert it back to previous way if required.

by checking history, I think it was added to add pagination

I tested everything locally, roles listing, creation, deletion works fine, same as before, so refactoring done well!

there's a different bug with pagination and searching, and that's because we do filter client side.. so searching for a string, and going to next page introduces "a weird behavior", which existed before. (i'll check if there's a bug for it)

lotas (approved):

PR #8989, opened 2026-08-05
chore(biome): migrate version in schema

lotas:
For some reason it complains when you run lint that biome config is of different version.
yarn.lock was upgraded to 2.5.5 at some point

Running `yarn biome migrate --write` fixes this. 
Probably needs to be a part of biome upgrade or yarn generate at some point

UPD: if we set schema to `./node_modules/@biomejs/biome/configuration_schema.json` then we won't have to deal with future upgrades at all, it will always point to the same as the installed version. I think it's fine-ish

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.5.1/schema.json",
+  "$schema": "./node_modules/@biomejs/biome/configuration_schema.json",
   "root": true,
   "files": {
     "includes": [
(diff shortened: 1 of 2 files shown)


biome.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.5.1/schema.json",
+  "$schema": "./node_modules/@biomejs/biome/configuration_schema.json",
```
lotas, line 2: either use local version always, or we have to run `biome migrate` for every version bump in the future

Eijebong: Not entirely sure how I feel about this but it seems to work fine :shrug:

Eijebong (approved):

biome.json
lotas: me neither.. but bumping this file each time biome upgrades is also **too much work** 😬

PR #8991, opened 2026-08-06
Third-party login updates

lotas:
1. make `client_id` check mandatory as per RFC
2. introduce optional PKCE S256 support (configured on a client level)
3. strict validation schema for `REGISTERED_CLIENTS` (web-server might fail to start if it fails)
4. fix multiple `redirectUri` parsing for CORS
5. update documentation to use modern @taskcluster/client and ESM modules + extend examples

(better reviewed commit-by-commit)

Bugzilla Bug: [2060854](https://bugzilla.mozilla.org/show_bug.cgi?id=2060854)


petemoore:
CC @djmitche

services/web-server/schemas/v1/registered-clients.yml
```diff
@@ -0,0 +1,66 @@
...
+        Scopes this client is allowed to request.
+      type: array
+      items:
+        type: string
+    redirectUri:
+      description: |
+        URIs this client may be redirected back to after authorization.
+      type: array
+      items:
+        type: string
```
petemoore, line 32: Can we add something like `pattern: '^https?://'` here too for safety? or maybe `format: uri` - or both? Might need to call [addFormats](https://ajv.js.org/guide/formats.html) too if we use `format: uri`. Is it worth adding a test with an invalid url just to check if it causes a crash?

services/web-server/schemas/v1/registered-clients.yml
```diff
@@ -0,0 +1,66 @@
...
+    requirePkce:
+      description: |
+        If true, authorization requests from this client must use PKCE.  Only
+        meaningful for the authorization-code flow.
+      type: boolean
+    maxExpires:
+      description: |
+        Longest lifetime of credentials issued to this client, as a duration
+        such as `1 year`.
+      type: string
```
petemoore, line 48: lexically I guess it is difficult to validate that this is an actual valid duration, but maybe we can add a test for an invalid duration?

services/web-server/src/validateConfig.js
```diff
@@ -0,0 +1,31 @@
...
+
+  const details = validateRegisteredClientsSchema.errors
+    .map(({ instancePath, message, params }) => {
+      const property = params.additionalProperty ? ` (${params.additionalProperty})` : '';
+      return `${instancePath || '/'} ${message}${property}`;
+    })
+    .join('; ');
+
+  throw new TypeError(`Invalid registeredClients configuration: ${details}`);
+};
```
petemoore, line 74: might be nice to also check that clientIds are unique here

djmitche (approved):
I definitely don't remember this code well enough to go line-by-line, but these are good fixes in general. Fixing the missed check for client-id is a good call!

services/web-server/schemas/v1/registered-clients.yml
lotas: yeah, good all. adding it

services/web-server/src/validateConfig.js
lotas: nice catch! will add

lotas: included! thanks

services/web-server/schemas/v1/registered-clients.yml
lotas: yeah, good call. also empty value produces current timestamp, so we'd be issuing tokens that are already expired 🤦 
good catch. added check in validate config

services/web-server/schemas/v1/registered-clients.yml
petemoore: Are you sure you added it?

lotas: must be cosmic rays, sorry.. even commit message said I did.. 
now it's here

petemoore (approved):
Awesome job, thanks Yarik!

PR #8997, opened 2026-08-07
feat(ui): encode log artifact name in the URL

lotas:
Fixes UI not showing some task runs with "strange" artifact names.

Example:  https://community-tc.services.mozilla.com/tasks/UgexocfcRe6RfLwffwqkxg

it won't be even opened in the UI 
image

while with the fix it renders 
image 

Related to: https://bugzilla.mozilla.org/show_bug.cgi?id=2072159


Eijebong (changes requested):
I have no idea what this is trying to actually fix. And from testing what I'm assuming it is fixing, it looks like it's not.

---

500s web-server:
```
    - name: public/logs/100%.log
      path: hello.log
      type: file
```

because the queue ain't happy

```
2026-08-10T13:26:45.098Z api:schema Output schema validation error:
Schema Validation Failed!
Rejecting Schema: http://taskcluster/schemas/queue/v1/get-artifact-response.json#
Errors:
  * data/url must match format "uri"
```

---

Redirects to `/foo.log` which I thought you were trying to solve
```
    - name: public/logs/../../../../../foo.log
      path: hello.log
      type: file
```

---

This ends up in a UI traceback:

```
logs:
  live: public/logs/%25foo.log
artifacts:
  - name: public/logs/%25foo.log
    path: x.log
    type: file
```

```
Uncaught URIError: malformed URI sequence
```

It's really contrived, only accessible through a link (not the TC UI) but that's accessible through github check links AFAICT.

lotas:
good examples, so actually currently if you submit those artifacts:

https://community-tc.services.mozilla.com/tasks/UgexocfcRe6RfLwffwqkxg

it won't be even opened in the UI 
image

while with the fix it renders 
image

lotas:
ah, I actually misunderstood your comment initially, I thought you were trying to say such artifacts cannot be created, but they can, and they break the UI if they get created, so fix is probably better to keep 

github link point is valid, I'll send a follow-up PR to encode artifact name segments correctly 

current version:
`public/logs/100%.log` old one crashed while rendering react router link, this fix should help
`public/logs/../../../foo.log` should be rejected as a log-viewer route 
`public/logs/%25foo.log` fixed that regression

ui/src/components/TaskRunsCard/index.test.jsx
```diff
@@ -98,3 +98,63 @@ it('should render TaskRunsCard', () => {
 
   expect(asFragment()).toMatchSnapshot();
 });
+
+it('should link an unsafe log artifact name to the artifact url', () => {
+  const originalEnv = window.env;
```
Eijebong: This whole env dance can be removed, it's already set in vitest.setup.js (I checked, removed the 3 lines and the test still passes)

ui/src/views/Tasks/TaskLog/index.jsx
```diff
@@ -55,17 +56,17 @@ export default class TaskLog extends Component {
     // for compatibility, if `name` is an encoded URL-shaped thing, try to
     // extract the artifact name.
     if (rawName.startsWith('https%3A')) {
```
Eijebong: Now this checks the wrong thing, the name arrives only half decoded. (and `%3A`
 is left alone which makes this wrong)


The following crashes the log viewer with `URL parameter name expected a string but was provided type "undefined"`
```
payload:
  command:
    - - /bin/bash
      - '-c'
      - echo "hello" > a.log
  artifacts:
    - name: https:foo.log
      path: a.log
      type: file
      contentType: text/plain
  maxRunTime: 120
```

You probably want to do something along the line of:

```js
const decoded = decodeArtifactName(rawName);
const match = decoded.startsWith('https:') ? /.*\/api\/queue\/v1\/task\/[^/]{22}\/runs\/\d+\/artifacts\/([^?]+)/.exec(decoded) : null;
const name = match ? decodeArtifactName(match[1]) : decoded;
```

Eijebong (changes requested):

ui/src/views/Tasks/TaskLog/index.jsx
lotas: i think we can completely drop this compat check which was added 6! (not factorial 6, just 6 exclamation mark) years ago
https://github.com/taskcluster/taskcluster/commit/3a7725b2784ca04ceb744bbc0c9e8d9884d45b7d

and only keep
```js
const name = decodeArtifactName(rawName)
```

matt-boris: Lol 6 factorial - back in 1306? 😆

lotas: I wonder how did you arrive to 1306? 6!=720 😬

matt-boris: Current year minus 6! years ago 😉

lotas: aahh.. i'm sleeping already

ui/src/utils/artifactNames.js
```diff
@@ -0,0 +1,29 @@
+const isSafeLogViewerArtifactName = name =>
+  name.split('/').every(
+    segment =>
+      segment.length > 0 &&
+      // history decodes %25 before rendering hrefs, which can revive %2e dot segments
+      !/[%?#\\]/.test(segment) &&
```
Eijebong: This is breaking my brain but I'm 99% sure that `#` and `?` don't need to be blocked here.

From the queue response to here:

- `const encodedName = name.split('/').map(encodeURIComponent).join('/');` 
- `location.pathname = decodeURI(location.pathname);` (this is behind our back, `history` does this, it half decodes the path, which is *really* confusing. The updated version apparently dropped this which is good but nowhere near close to being updatable to (because it's brought by react-router))
  - `<a href={createHref(location)}>`
  - `pushState(..., href)` (sets the URL bar to href)
- [The browser normalizes the URL here]
- `decodeURIComponent(name)`
- `encodeURIComponent(param)`

On reload you enter directly into `location.pathname`. So we need `href` in the URL bar to roundtrip through `decodeURI` (which is why `%` breaks).

Basically, each segment needs to pass both:

- `decodeURIComponent(decodeURI(encodeURIComponent(segment))) === segment`
- `decodeURI(normalize(decodeURI(encodeURIComponent(segment))) === decodeURI(encodeURIComponent(segment)`

```js
const normalize = s => new URL(`/a/${s}/b`, 'https://example.com').pathname.split('/')[2];
const roundtrips = n => {
  try {
    const s = decodeURI(encodeURIComponent(n));
    return decodeURIComponent(s) === n && decodeURI(normalize(s)) === s;
  } catch {
    return false
  }
}

const segments = [];
for (let i = 0x20; i <= 0x7e; i++) {
  const c = String.fromCharCode(i);
  if (c !== '/') segments.push(`a${c}b.log`);
}
// And a few edge cases
segments.push('.', '..', '', '%2e', 'a%3Fb.log', 'normal.log');

for (const s of segments) {
  if (!roundtrips(s)) {
    console.log(`${s} didn't roundtrip`);
  }
}
```

```
$ node a.js
a%b.log didn't roundtrip
a\b.log didn't roundtrip
. didn't roundtrip
.. didn't roundtrip
%2e didn't roundtrip
a%3Fb.log didn't roundtrip
```

So basically `\`, `%`, `.` and `..` are the only things that need to go.

lotas: fair point. pushing a fix.

Note I rebased all, but only last commit is what changed

Eijebong (approved):
LGTM now, thank you. Wonder if we should do the same in `TaskArtifactRedirect` (`TaskArtifactRedire` class name because JS is awesome and it's the default export so it could have been named `Roflmao` and still work).  I don't know how to test whether or not it's a problem there but I'm assuming it is. Fine as a followup though

lotas:
`TaskArtifactRedirec` is a good enough class name 🤣 
right, we also pass it there .. I'll add a separate PR, thanks!

lotas:
https://github.com/taskcluster/taskcluster/pull/9176 here it goes

PR #9009, opened 2026-08-11
Enable ipv6 on the per-task docker network when it's enabled on the daemon

### LINKED ISSUE #8990: Generic Worker per-task Docker network disables IPv6 for d2g tasks
**Describe the bug**

Generic Worker 100.0.0 and later creates a dedicated `gw-task-*` Docker network for d2g tasks. Generic Worker creates this network without the Docker `--ipv6` option.

This change moves the task container from Docker's IPv6-enabled default bridge to an IPv4-only user-defined network. The problem occurs even when IPv6 is enabled in the Docker daemon configuration.

On Firefox CI, this causes the Tier 1 xpcshell test `netwerk/test/unit/test_network_connectivity_service.js` to fail:

```
Check DNSv6 support (expect OK) - 2 == 1
```

**To Reproduce**

1. Configure a Linux Docker worker with IPv6 enabled in the Docker daemon.
2. Run a d2g task with Generic Worker 103.0.1.
3. Confirm that Generic Worker creates a `gw-task-*` network and starts the task container with `--network gw-task-*`.
4. Run an IPv6 connectivity check in the task container.
5. Observe that the container has no IPv6 connectivity.

A production example is available in these tasks:

1. [Passing task on Generic Worker 99.2.1](https://firefox-ci-tc.services.mozilla.com/tasks/cSEm_OkkQPWJNH_zEa6vjQ)
2. [Failing backfill of the same Gecko revision on Generic Worker 103.0.1](https://firefox-ci-tc.services.mozilla.com/tasks/dp2IcsShTaWGQwMr646fZQ)

The tasks use the same Gecko revision, build artifact, task image, test command, and task group. The passing task uses Docker's default bridge. The failing task uses a `gw-task-*` network.

**Expected behavior**

The per-task Docker network must preserve the IPv6 connectivity that is available on the default bridge. A d2g task on a capacity-1 worker must not lose IPv6 connectivity after a Generic Worker upgrade.

Generic Worker should enable IPv6 when it creates the per-task network. It must configure an IPv6 subnet or address pool if Docker requires one.

**Taskcluster version**

- Generic Worker: 103.0.1
- First known bad version: 100.0.0
- Last known good version: 99.2.1

**Screenshots**

Not applicable. The task logs above show the Docker commands and the test failure.

- OS: Ubuntu 24.04, Firefox CI Google Compute worker image dated 2026-08-04
- Browser if applicable: Not applicable
- Version if applicable: Docker Engine 29.5.3

**Additional context**

The behavior started with [taskcluster#8278](https://github.com/taskcluster/taskcluster/pull/8278). Commit [4a557276](https://github.com/taskcluster/taskcluster/commit/4a5572769f549fd43a72a89b54e0e90b95e2829e) added the per-task network. The code runs `docker network create gw-task-...` without `--ipv6`.

Docker requires `--ipv6` to enable IPv6 on a user-defined network. The daemon `fixed-cidr-v6` setting configures the default bridge and does not enable IPv6 on each user-defined network. Docker 29.4.2 and 29.5.3 both default this option to false, so this is not a Docker 29.5.3 regression.

The affected Firefox CI pool has capacity 1. The description of taskcluster#8278 states that behavior for capacity-1 workers does not change, but the Docker network topology changes for these tasks.

The Generic Worker upgrade reached production in [worker-images#839](https://github.com/mozilla-platform-ops/worker-images/pull/839) and [fxci-config#1109](https://github.com/mozilla-releng/fxci-config/pull/1109).

The Firefox OS integration xpcshell task used the same image and Generic Worker version, but it ran with `--tag=os_integration`. The failing test does not have that tag, so the integration check did not run it.

A regression test should run a d2g task on the per-task network and verify IPv6 connectivity.

Eijebong:
4a5572769f549fd43a72a89b54e0e90b95e2829e moved d2g task containers off of docker's default bridge network so that tasks could have isolated networks. Those per-task networks are created without `--ipv6`, which means that unlike the default bridge that defaults to having an ipv6, task networks did not. Containers therefore lost all ipv6 capabilities which is a problem for fxci.

Pass `--ipv6` to the task network creation only if the default bridge has ipv6 enabled. I wasn't too sure about how to handle configuration for this (I don't want to have it swing the other way and have ipv6 enabled when it shouldn't be), but, tbqh, the per-task docker network feels like an implementation detail that was never communicated in the first place so as a deployer I'd expect configuration done on the docker daemon to stick. Since we have access to that through `docker network inspect` it's fairly easy to check whether or not ipv6 is enabled on the default network, and forward that config option to task networks.

This raises the minimum supported version of docker to 27 since docker didn't want to implicitly declare ipv6 subnets before that. One could technically run on older docker versions by disabling ipv6 on the default bridge but I don't think it's worth documenting this.

I went back and forth, trying to avoid that minimum version but without gutting capacity > 1 or gutting ipv6 support based on version (which I really would rather not do), but docker is very uncooperative on the subject. Versions older than 27, as noted above require passing a `--subnet` with a *unique* subnet per task. That's solvable by hashing the taskID/runID, picking up a few bytes and shoving that into an IPv6 prefix (collision chances were like 10^-14 per task for capacity 16). On the paper that sounds good enough, except that while those tasks ended up with an ipv6 they had no egress over it as docker didn't actually add a `MASQUERADE` rule for ipv6 before version 27... So the containers would boot up with an ipv6, a route saying "bridge IP is next" and packets went there to die silently. Which is a lot worse than no ipv6 since it means producers for said packets would not even be notified about them being dropped and would have to time out. And the fix is to do iptables ourselves and that's a huge fat "no thanks".

Fixes #8990

### DIFF
--- a/changelog/issue-8990.md
+++ b/changelog/issue-8990.md
@@ -1,0 +1,8 @@
+audience: worker-deployers
+level: major
+reference: issue 8990
+---
+Generic worker has ipv6 enabled again for d2g tasks if it's enabled on the
+default bridge. Due to how docker handles ipv6 on networks that are necessary
+for capacity > 1, this raises the minimum docker version supported by generic
+worker to 27.
--- a/workers/generic-worker/taskcluster_proxy_feature.go
+++ b/workers/generic-worker/taskcluster_proxy_feature.go
@@ -57,6 +57,14 @@ func sweepStaleDockerResources() {
 	}
 }
 
+func dockerDefaultBridgeHasIPv6() bool {
+	out, err := host.Output("docker", "network", "inspect", "bridge", "--format", "{{.EnableIPv6}}")
+	if err != nil {
+		return false
+	}
+	return strings.TrimSpace(out) == "true"
+}
+
 func (feature *TaskclusterProxyFeature) IsEnabled() bool {
 	return config.EnableTaskclusterProxy
 }
@@ -102,7 +110,11 @@ func (l *TaskclusterProxyTask) Start() *CommandExecutionError {
 		// Create a per-task Docker network for isolation. Each container
 		// will only be able to reach its own tc-proxy instance.
 		networkName := fmt.Sprintf("gw-task-%s-%d", l.task.TaskID[:12], l.task.RunID)
-		_, err := host.Output("docker", "network", "create", networkName)
+		createArgs := []string{"network", "create"}
+		if dockerDefaultBridgeHasIPv6() {
+			createArgs = append(createArgs, "--ipv6")
+		}
+		_, err := host.Output("docker", append(createArgs, networkName)...)
 		if err != nil {
 			return executionError(internalError, errored, fmt.Errorf("could not create Docker network %s: %s", networkName, err))
 		}


petemoore (dismissed):
Looks good to me, thanks Bastien! Should we mention in the changelog that support for ipv6 still depends on it being enabled in the default bridge?

PR #9010, opened 2026-08-11
fix: change task index page from graphql to rest

### LINKED ISSUE #9006: Migrate Task Index page from GraphQL to REST

thakkarujas5:
Github Issue: Fixes #9006


lotas:
something funky is going on when loading namespaces - I see double requests for every page i go into:
image

upd: nevermind, those are separate requests, url was truncated.. one is namespaces one is tasks

lotas (changes requested):
there are few issues I flagged in matrix

lotas (dismissed):
tested again locally, pages do work as expected, thanks!

lotas (approved):
tested again locally, works fine, thanks

PR #9013, opened 2026-08-14
fix: rename Client Audit History section

thakkarujas5:
#7590 added a section Client Audit History which sounded confusing as to which client it is referring to, fixed this.


thakkarujas5:
image

Eijebong (approved):

PR #9015, opened 2026-08-14
Add a route to the auth service to get github tokens

### LINKED ISSUE #8378: Implement ability to hand out Github tokens via API
The auth service currently supports the ability to hand out credentials for various external services like AWS, Azure, GCP and Sentry:
https://docs.taskcluster.net/docs/reference/platform/auth/api

It would be _really_ helpful if it could also hand out Github tokens.

### Problem

As we're migrating Firefox to Github, there are more and more tasks that need to interface with the Github API in some way. There are two ways we currently do this:

1. Create a personal access token and store it in a TC secret
2. Create a Github App, store the private key in a TC secret, then use this to generate an access token at task runtime

The first option sucks because you're either tying auth to someone's personal Github user and it will fail when they leave the company, or you need to create "fake" users and deal with storing their credentials.

The latter is the recommended approach, but the login process is somewhat complex:
https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation

The latter also means you need to create a new GH app anytime you want to do something. Especially when the Taskcluster instance's Github app is already installed!

### Proposal

Implement an API endpoint that handles the complexity of generating a GH App installation token for the user using the Taskcluster Github app that's already installed on the repo.

I could see an argument for this API living in either:

1. The Auth service because that's where similar endpoints already exist
2. The Github service because this would be tightly coupled to the same app that the GH service already uses

Eijebong:
Automating anything that is related to github often requires adding authentication tokens, if only to not worry about rate limiting. Currently the only way of doing this is creating a PAT, setting it as a secret and then using that in the task. It causes many issues, the token
expires, the owner of the token expires, the task needs to get a secret (even if it's a read only token) which is painful to audit.

To help with this, this commits adds a route to the auth service that is very similar to what already exists for example for GCP where the auth service owns long lived broadly scoped credentials and distributes much narrower tokens (both in scope and time) to requests, protected behind a scope that is explicit about what's getting requested.

On the implementation itself, it's fairly "simple". We have configuration for multiple named applications (the name is a taskcluster identifier that is used in scopes/calls). Anyone with credentials with scopes like: `auth:github-repo-token:<appName>/<owner>/<repo>:<perm>:<level>` can request the permission `<perm>:<level>` (matching the exact name github
gives it, see https://docs.github.com/en/rest/apps/apps?apiVersion=2026-03-10#create-an-installation-access-token-for-an-app) on `<owner>/<repo>` through the app `<appName>`. The token lifetime is currently 1h, decided by github, we have no control over that, it's the caller's responsibility to get a new token once it expires if needed as github provides no way to either extend the lifetime of a token or refresh it.

I'd recommend reviewing commit by commit and ignoring whitespaces in the commit adding caching to the installation ID as it makes the github.js changes much more readable.

Fixes #8378


ui/docs/manual/deploying/cloud-credentials.mdx
```diff
@@ -111,6 +111,38 @@ The current implementation only supports one project, with any number of allowed
 [Future work](https://bugzilla.mozilla.org/show_bug.cgi?id=1599247) will allow multiple projects.
 </Warning>
 
+## Github credentials
```
ahal, line 114: Would be nice to have some docs targeted at users of the feature as well. Likely on this page:
https://github.com/taskcluster/taskcluster/blob/main/ui/docs/reference/platform/auth/external-services-credentials.mdx

I realize we don't document any of the other external services basically at all.. but maybe this can be the one where we start :p. I was thinking just some quick examples of how you might leverage this as a task author.

(If you want to leave it for a follow-up that documents everything at once, I'm cool with that too)

Eijebong: Agreed that we should probably document this from a user perspective. It'll already have the reference implementation from the route itself but a task example would be nice. Do you have ideas? I'm thinking maybe a simple bash example "here's how to clone a private repo"? And maybe a python example on how to open a PR from a task? :thinking: Might be too complex for documentation but it's also the kind of stuff I know *I* would like to see there soo...

ahal: Those both sound good to me, though I had something less useful in mind, just enough to obtain the token. Another option could be to demonstrate an authenticated curl call against the `/octocat` endpoint.. 

But I'll leave it up to you

Eijebong: Added some. Does it match what you wanted?

ahal: Thanks, this is great!

services/auth/src/github.js
```diff
@@ -57,3 +70,128 @@ export const makeGithub = ({ cfg, monitor }) => {
...
+        for (const repo of repositories) {
+          repoPerms.push(`${repo}:${permission}:${level}`);
+        }
+      }
+
+      await req.authorize({ appName, owner, repoPerms });
+
+      const app = this.github.get(appName);
+
+      if (app === undefined) {
```
lotas, line 148: nit: I know taskcluster is mostly like this where generic "missing scope" error is the first thing you see, but I wonder if we should do this check before we authorize scopes?

services/auth/src/github.js
```diff
@@ -0,0 +1,215 @@
...
+  if (err.status !== 403) {
+    return false;
+  }
+
+  const headers = err.response?.headers ?? {};
+  return headers['retry-after'] !== undefined || headers['x-ratelimit-remaining'] === '0';
+};
+
+// lowercased owner: installation id
+export const makeInstallationCache = () => new LRU({ maxSize: 1000, maxAge: INSTALLATION_CACHE_MAX_AGE_MS });
```
petemoore, line 74: maxAge appears to be a quick-lru 6.0.0 feature, but i think we are on quick-lru 5.1.1

services/auth/src/github.js
Eijebong: Well in general security says that you should not be able to enumerate things if you don't have the rights to use said things. In the fxci case it doesn't really matter since our scopes config is public so you can just go and grab the app names there. But for a locked down deployment, that's not the case. I'm not sure what we'd gain by doing that check before either.

ui/docs/manual/deploying/cloud-credentials.mdx
```diff
@@ -111,6 +111,38 @@ The current implementation only supports one project, with any number of allowed
...
+
+The available apps are defined by the keys of the outer object, in this case just `app-name`.
+App names appear both in the method's `appName` argument and in the scopes required to call it, so they must match `^[a-zA-Z0-9_-]+$`.
+The `appId` property is the numeric app ID visible in the application settings, `privateKey` is a private key generated on that same page.
+It's a multiline string whose newlines may be written either as literal `\n` escape sequences or as actual newlines if your deployment mechanics handle that.
+The Auth service will refuse to start if a configured app is missing either property or if a configured key doesn't look like a proper one.
+
+These apps are independent of the app used by the Github service for CI, described in [Github Integration](/docs/manual/deploying/github).
+
+Requesting `<permission>: <level>` on `<owner>/<repo>` requires the scope `auth:github-repo-token:<appName>/<owner>/<repo>:<permission>:<level>`, with the owner and repository name in lowercase.
```
petemoore, line 138: At one point I think we were considering `auth:github-repo-token:<appName>/<perm>:<level>:<owner>:<repo>` so that `*` expansion could cover multiple repos. Did we decide a solution for that use case (e.g. parameterized roles)?

services/auth/src/github.js
```diff
@@ -0,0 +1,215 @@
...
+    return true;
+  }
+
+  if (err.status !== 403) {
+    return false;
+  }
+
+  const headers = err.response?.headers ?? {};
+  return headers['retry-after'] !== undefined || headers['x-ratelimit-remaining'] === '0';
+};
```
petemoore, line 71: GitHub's docs say `Retry-After` is included "if present" but tell clients to back off ≥60s when it's absent, and `x-ratelimit-remaining` reflects only the primary quota, not secondary limits. A secondary-limit 403 lacking both signals falls through to a generic re-throw -> InternalServerError (500), which per the existing test also fires a `monitor.error` alerting noise for a benign, expected GitHub-side throttle. Since
  @octokit/plugin-throttling's own `onSecondaryRateLimit` callback already implements GitHub's recommended detection, it may be better to key off that than header-sniffing.

services/auth/src/github.js
```diff
@@ -134,29 +142,38 @@ export const githubBuilder = builder => {
...
+              'TooManyRequests',
+              `The application \`${appName}\` is currently hitting rate limits`
+            );
+          }
+          throw err;
+        }
+
+        const actualAppOwner = installationInfo.data.account?.login;
+
+        if (actualAppOwner?.toLowerCase() !== owner) {
```
lotas, line 175: so before adding cache that would work on each call, but with lru cache layer, this will only be done once for its lifetime. 
Unlikely renames are happening so often, but maybe worth checking always?

services/auth/schemas/v1/github-repo-token-request.yml
```diff
@@ -0,0 +1,52 @@
...
+      single_file: { type: string, enum: ["read", "write"] }
+      statuses: { type: string, enum: ["read", "write"] }
+      vulnerability_alerts: { type: string, enum: ["read", "write"] }
+      workflows: { type: string, enum: ["write"] }
+  repositories:
+    title: "Repositories"
+    description: |
+      Names of the repositories (within `:owner`) the token gets access to.
+    type: array
+    minItems: 1
```
lotas, line 50: should we also add some `maxLimit: 500` here to match what github has and avoid the explosion of scopes check in `AllOf` 

https://docs.github.com/en/rest/apps/apps?apiVersion=2026-03-10#create-an-installation-access-token-for-an-app:~:text=Up%20to%20500%20repositories%20can%20be%20listed%20in%20this%20manner

services/auth/src/github.js
Eijebong: No we're not. #9014

ui/docs/manual/deploying/cloud-credentials.mdx
Eijebong: We talked about this at the community meeting yeah. This reads better and in prod we don't want to use star scopes because it's way too easy to give access to an org assuming "all repos are fine" until someone creates a private repo in there. With this it's much harder to think something like that

services/auth/src/github.js
lotas: indeed. nvm

services/auth/src/github.js
Eijebong: The cache has a max lifetime of one hour for this. I picked one hour just so it's not infinite, I don't think we'll ever see an org rename.

services/auth/schemas/v1/github-repo-token-request.yml
Eijebong: I missed that they had an upper limit there, nice catch

services/auth/schemas/v1/github-repo-token-request.yml
```diff
@@ -0,0 +1,52 @@
...
+description: |
+  The list of permissions and repositories to scope a repository level github token to.
+type: object
+properties:
+  permissions:
+    title: "Repository permissions"
+    description: |
+      Repository level github app permissions to request. Mapping a permission
+      name to the desired level of access. Names and levels are a direct mapping
+      from [github API permissions](https://docs.github.com/en/rest/apps/apps?apiVersion=2026-03-10#create-an-installation-access-token-for-an-app).
```
lotas, line 12: on that page the list of permissions is a bit bigger, was this intentional to support less? 

```js
[...document.querySelectorAll('#permissions-object code.text-bold.f5')].map(elm => elm.innerText)

[
  "actions",
  "administration",
  "artifact_metadata",
  "attestations",
  "checks",
  "code_quality",
  "codespaces",
  "contents",
  "dependabot_secrets",
  "deployments",
  "discussions",
  "environments",
  "issues",
  "merge_queues",
  "metadata",
  "packages",
  "pages",
  "pull_requests",
  "repository_custom_properties",
  "repository_hooks",
  "repository_projects",
  "secret_scanning_alerts",
  "secrets",
  "security_events",
  "single_file",
  "statuses",
  "vulnerability_alerts",
  "workflows",
  "custom_properties_for_organizations",
  "members",
  "organization_administration",
  "organization_custom_roles",
  "organization_custom_org_roles",
  "organization_custom_properties",
  "organization_copilot_seat_management",
  "organization_copilot_agent_settings",
  "organization_announcement_banners",
  "organization_events",
  "organization_hooks",
  "organization_personal_access_tokens",
  "organization_personal_access_token_requests",
  "organization_plan",
  "organization_projects",
  "organization_packages",
  "organization_secrets",
  "organization_self_hosted_runners",
  "organization_user_blocking",
  "email_addresses",
  "followers",
  "git_ssh_keys",
  "gpg_keys",
  "interaction_limits",
  "profile",
  "starring",
  "enterprise_custom_properties_for_organizations",
  "actions",
  "administration",
  "artifact_metadata",
  "attestations",
  "checks",
  "code_quality",
  "codespaces",
  "contents",
  "dependabot_secrets",
  "deployments",
  "discussions",
  "environments",
  "issues",
  "merge_queues",
  "metadata",
  "packages",
  "pages",
  "pull_requests",
  "repository_custom_properties",
  "repository_hooks",
  "repository_projects",
  "secret_scanning_alerts",
  "secrets",
  "security_events",
  "single_file",
  "statuses",
  "vulnerability_alerts",
  "workflows",
  "custom_properties_for_organizations",
  "members",
  "organization_administration",
  "organization_custom_roles",
  "organization_custom_org_roles",
  "organization_custom_properties",
  "organization_copilot_seat_management",
  "organization_copilot_agent_settings",
  "organization_announcement_banners",
  "organization_events",
  "organization_hooks",
  "organization_personal_access_tokens",
  "organization_personal_access_token_requests",
  "organization_plan",
  "organization_projects",
  "organization_packages",
  "organization_secrets",
  "organization_self_hosted_runners",
  "organization_user_blocking",
  "email_addresses",
  "followers",
  "git_ssh_keys",
  "gpg_keys",
  "interaction_limits",
  "profile",
  "starring",
  "enterprise_custom_properties_for_organizations"
]
```

lotas: maybe github publishes this as schema also we could re-use

lotas: ah, I guess not all applies to repositories, sorry

Eijebong: Most of these are enterprise/user/org ones but I was missing a few indeed

services/auth/src/github.js
Eijebong: Stupid github

lotas (approved):
thank you!

PR #9017, opened 2026-08-14
Don't let late launch agent results mask an aborted command

Eijebong:
This fixes both a leaked goroutine I found while investigating this and a race condition. See each commit for details


matt-boris (approved):
Nice, thanks!

PR #9019, opened 2026-08-14
fix(generic-worker): don't gc file caches in use

### LINKED ISSUE #8943: Garbage collection can delete a file cache a running task is using with capacity > 1

matt-boris:
Fixes #8943.

>Generic Worker no longer garbage collects a file cache that a running task is still using in `capacity` > 1 cases. Relatedly, when a cached download no longer matches a task's required SHA256, the stale entry is now dropped from the cache table immediately (with its deletion deferred until any tasks still using it finish) instead of being served to the task again.

### DIFF
--- a/workers/generic-worker/mounts_feature.go
+++ b/workers/generic-worker/mounts_feature.go
@@ -169,11 +169,11 @@ func trimPoolExcess() {
 }
 
 func (cm FileCacheMap) SortedResources() Resources {
-	r := make(Resources, len(cm))
-	i := 0
+	r := Resources{}
 	for _, cache := range cm {
-		r[i] = cache
-		i++
+		if !cache.InUse {
+			r = append(r, cache)
+		}
 	}
 	sort.Sort(r)
 	return r
@@ -208,6 +208,7 @@ func (cm *FileCacheMap) LoadFromFile(stateFile string, cacheDir string) {
 			delete(*cm, i)
 		} else {
 			(*cm)[i].Owner = *cm
+			(*cm)[i].InUse = false // all entries are available on startup
 		}
 	}
 }
@@ -228,6 +229,7 @@ type Cache struct {
 	// SHA256 of content, if a file (not used for directories)
 	SHA256 string `json:"sha256"`
 	// InUse indicates whether a task currently owns this pool entry.
+	// For file caches, true while any task is using the downloaded file.
 	// Persisted with json:"in_use" (rather than json:"-") because
 	// loadFromJSONFile uses DisallowUnknownFields — older state files
 	// written by prior workers contain "in_use" and would otherwise
@@ -240,6 +242,8 @@ type Cache struct {
 	// in use. ReleaseCache checks this flag and evicts the entry instead
 	// of returning it to the pool.
 	NeedsPurge bool `json:"-"`
+	// Number of tasks currently using this file cache (not used for directories)
+	UseCount int `json:"-"`
 }
 
 // Rating determines how valuable the cache is compared to others.
@@ -252,6 +256,9 @@ func (cache *Cache) Rating() float64 {
 }
 
 func (cache *Cache) Evict(taskMount *TaskMount) error {
+	if cache.UseCount > 0 {
+		return nil
+	}
 	if taskMount != nil {
 		taskMount.Infof("Removing cache %v from cache table", cache.Key)
 	}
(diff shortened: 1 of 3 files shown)


workers/generic-worker/mounts_feature.go
```diff
@@ -1010,6 +1054,7 @@ func ensureCached(fsContent FSContent, taskMount *TaskMount) (file string, sha25
 		}
 		taskMount.Infof("Found existing download of %v (%v) with SHA256 %v but task definition explicitly requires %v so deleting it", cacheKey, file, sha256, requiredSHA256)
 		cacheMutex.Lock()
+		taskMount.releaseFileCache(cachedEntry)
 		if entry, ok := fileCaches[cacheKey]; ok {
 			err = entry.Evict(taskMount)
```
Eijebong: I think this actually exposes a bug now because this eviction might not actually delete files if they're in use by another task.

So if a tasks holds the cache, something tampers with the cached files (cosmic rays, I dunno), then this calls evict which short circuits since the cache is in use.

Then, since we need to redownload the mount, we pass by:

```
val, dlErr, _ := fileCacheDownloads.Do(cacheKey, func() (any, error) {
    // Re-check cache: another goroutine may have completed the download
    // while we were waiting for the singleflight lock.
    cacheMutex.Lock()
    if entry, ok := fileCaches[cacheKey]; ok {
        cacheMutex.Unlock()
        return downloadResult{file: entry.Location, sha256: entry.SHA256}, nil
    }
```

which immediately short circuits and returns the entry as is, despite it not matching the checksum

workers/generic-worker/mounts_feature.go
```diff
@@ -1010,6 +1054,7 @@ func ensureCached(fsContent FSContent, taskMount *TaskMount) (file string, sha25
 		}
 		taskMount.Infof("Found existing download of %v (%v) with SHA256 %v but task definition explicitly requires %v so deleting it", cacheKey, file, sha256, requiredSHA256)
```
Eijebong: `so deleting it` here doesn't work anymore since the cache might be in use still

workers/generic-worker/mounts_feature.go
Eijebong: You'll want to insert it with `UseCount: 1` and insert in fileCacheRefs here I think? Otherwise a GC could run between the unlock below and the download and destroy the entry before it even has a chance to be refcounted below and that'd probably end very badly.

Eijebong: Also the `Cache` struct still says:

```go
// Number of tasks currently using this file cache (not used for directories)
UseCount int `json:"-"`
```

You'll want to change that

workers/generic-worker/mounts_feature.go
```diff
@@ -1069,6 +1121,7 @@ func ensureCached(fsContent FSContent, taskMount *TaskMount) (file string, sha25
 		err = fmt.Errorf("Download %v of %v has SHA256 %v but task definition explicitly requires %v; not retrying download as there were no connection failures and HTTP response status code was 200", file, fsContent, sha256, requiredSHA256)
 		cacheMutex.Lock()
 		if entry, ok := fileCaches[cacheKey]; ok {
+			taskMount.releaseFileCache(entry)
```
Eijebong: Because you release the lock in between, the entry might be different here if it gets evicted and recreated by another task which would lead to you removing one from the counter of another cache entry than they one you're expecting, potentially setting it to 0 (or worse, -1).

Re-reading this block, I think the best way would be to not even unlock the cache line 1114 and instead defer the unlock and let the function finish. That removes a bunch of cache unlock branching.

Eijebong (changes requested):

workers/generic-worker/mounts_feature.go
matt-boris: > You'll want to change that

@Eijebong which part are you referring to here?

Eijebong: I think I misread the comment back then because I thought it was obvious and can't find it anymore :joy:

matt-boris: No worries, ty

Eijebong (approved):

PR #9026, opened 2026-08-17
Fix a typo in `finishArtifact`'s error path

Eijebong:
I don't think taskcluster sports, let alone resports. Joke aside, this fixes an obvious typo that would make `finishArtifact` return a 500 instead of a 400 when calling it on a non object artifact.


matt-boris (approved):
Lol ty

PR #9031, opened 2026-08-18
Make JS clients throw when authorizedScopes is passed as the wrong type

Eijebong:
Since we don't have type to do this, we have to rely on manual checks and runtime errors. Previously, passing `authorizedScopes: 'foo'` didn't do *anything*. And silently. Which means a user could be expecting to make a call with a narrow set of scopes and end up doing with a much wider one (see #9030 for an example). This commit changes it from being silent to throwing an error.

I checked the taskcluster code base and apart from #9030, I didn't find any other misuse of `authorizedScopes`.

### DIFF
--- a/changelog/K0i_4srpQd2FKydDFwu02A.md
+++ b/changelog/K0i_4srpQd2FKydDFwu02A.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+---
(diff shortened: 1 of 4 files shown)


Eijebong:
Nice. Failure is #9030 which is expected.

matt-boris (approved):

PR #9032, opened 2026-08-18
Bug 2064373 - Disable slack message link previews by default

Eijebong:
I chose to disable it by default and add a setting to re-enable them. You most likely want short notifications rather than have half of a 4k screen filled with link previews that have no value whatsoever.

See https://docs.slack.dev/messaging/unfurling-links-in-messages/


matt-boris (approved):
👍🏻

PR #9034, opened 2026-08-20
Bug 2064901 - Require `github:trigger-hook` to fire hooks from .taskcluster.yml

Eijebong:
This changes the feature that was introduced by #8431 to require `github:trigger-hook:<hook>` instead of `hooks:trigger-hook:<name>`.

This fixes a security oversight of the original implementation. The feature was implemented so that we could run the same task template across trust domain / repositories / levels, all of which are derived from the github event that the github service injects into the created task. Due to how hooks work, the `hook-id:<name>` role needs to hold the sum of all scopes that every github event could trigger it for.

This would be fine except it needs to hold `trigger-hook:<itself>` since the github service could not trigger it with its own credentials unless the event role for which it was trying to trigger said hook also held that scope (which the github service held transitively since it holds `assume:repo:github.com/*`).

A decision task could then call `triggerHook` itself with a fake github event, and since the hook has scopes for every single other repo using said hook, it'd be pretty bad.

The new way of doing this is to add a marker scope `github:trigger-hook:<name>` which allows a github event to trigger a hook of the same name *through* the github service. Since said service verifies that events are legitimately from github and since it forces `event` in the context to be whatever github provided, the hook can then assume the right scopes for the event, which solves the issue described above. The github event role cannot trigger the hook manually anymore, everyone's happy.

This means however that I had to expand the roles of the static github service client to be able to trigger all hooks (and to expand scopes, that one's fine). This is a bit dicey and I'm not thrilled about it but given the very broad `assume:repo:github.com/*` right above, it's not too much of a stretch.

### DIFF
--- a/changelog/bug-2064901.md
+++ b/changelog/bug-2064901.md
@@ -1,0 +1,6 @@
+audience: users
+level: major
+reference: bug 2064901
+---
+Github hooks triggered from .taskcluster.yml now require the
+`github:trigger-hook:<name>` scope instead of `hooks:trigger-hook:<name>`.
--- a/services/auth/src/static-scopes.json
+++ b/services/auth/src/static-scopes.json
@@ -11,6 +11,8 @@
     "scopes": [
       "assume:repo:github.com/*",
       "assume:scheduler-id:taskcluster-github/*",
+      "auth:expand-scopes",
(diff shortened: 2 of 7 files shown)


lotas (approved):
nice, thanks

PR #9035, opened 2026-08-20
Bug 2062167 - Ensure each interactive session is authenticated against its own secret

Eijebong:
This is yet another fallout from the capacity > 1 PR where it was missed  that the interactive feature was using an environment *on the worker* as a way to pass the interactive session secret around instead of reading it from the interactive struct that was right there. That's been the case since the secret check got introduced in https://github.com/taskcluster/taskcluster/commit/ad2c6d5ae7ad46eeacc56ca7bc38e7e38bdfa63c.

With capacity > 1, this becomes a problem because an environment variable is just a hidden global. A worker process that would start two interactive sessions would have that global change on each session,  making the first session not able to authenticate anymore with its own  secret after the second one started. Instead, the second shell could authenticate to the first one with the new secret.

The change in code is fairly simple, as I said before, the struct was right there, just read the secret from it. I also switched the check to use a timing safe compare since I was in the area. It's a secret, it costs nothing to do this right.

Where that got a bit spicy is with tests (which is probably the reason why the environment variable was a thing in the first place). Because integration tests don't have access to the internals of the worker, they can't extract the secret from the interactive struct so I had to find another way. Thankfully extracting the secret from the shell artifact was fairly straightforward since it's a reference artifact with the secret in the URL and it required very little finagling.

### DIFF
--- a/changelog/bug-2062167.md
+++ b/changelog/bug-2062167.md
@@ -1,0 +1,5 @@
+audience: users
+level: patch
+reference: bug 2062167
+---
+Fixed a bug where multiple interactive sessions on a worker with capacity greater than 1 would override eachother's secrets.
--- a/internal/mocktc/queue.go
+++ b/internal/mocktc/queue.go
@@ -530,6 +530,17 @@ func (queue *Queue) Artifact(taskId, runId, name string) (*tcqueue.GetArtifactCo
 			return nil, err
 		}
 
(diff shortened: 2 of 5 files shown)


matt-boris (approved):
Thanks, glad this was pretty straightforward to fix 🙏🏻

PR #9036, opened 2026-08-20
chore(docs): Improve readme wording for security reporting

lotas:
we see an increased number of advisories being drafted where reports should be going through the official SECURITY.md guidelines

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,7 +6,7 @@ This code and its associated production web page are included in Mozilla's web a
 
 ## Reporting a Potential Vulnerability
 
-Please submit all security-related bugs through [HackerOne](https://hackerone.com/mozilla). Never submit security-related bugs through a GitHub Issue or by email.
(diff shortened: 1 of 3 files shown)


SECURITY.md
```diff
@@ -6,17 +6,18 @@ This code and its associated production web page are included in Mozilla's web a
...
 
 To help us triage quickly, please provide:
 
 - A clear description of the issue
 - Steps to reproduce, or a proof of concept
-- Affected versions or environments
+- Affected versions, services, deployments, or environments
 - Potential impact, including what an attacker could achieve
 - Any suggested mitigations or fixes
+- Any existing GitHub Security Advisory ID, if one was already created
```
Eijebong: Bit weird to tell people not to create a github advisory but then to link one? 🤔

lotas: yeah. some people don't read it before creating one idk. I can drop it :)

Eijebong (approved):

SECURITY.md
Eijebong: Then ~~their LLM~~they can copy the content over to h1 :shrug:.

PR #9037, opened 2026-08-20
fix: change hooks page to use direct rest endpoints

### LINKED ISSUE #9023: Migrate Hooks page from GraphQL to REST

thakkarujas5:
Github Issue: Fixes #9023


ui/src/components/HookForm/index.jsx
```diff
@@ -701,41 +700,27 @@ export default class HookForm extends Component {
             </List>
           </ListItem>
           {!isNewHook && (
-            <Fragment>
-              <ListItem>
-                <ListItemText
-                  primary="Next Scheduled Fire"
```
lotas, line 707: I think it's worth mentioning in commit message or changelog why it was removed.
we discussed this offline, and concluded that it's not worth calling deprecated method that was returning this value. 
graphql was using for historical reasons probably, but method itself was deprecated, and no other api methods return this information because it's not part of definition or last fire, so probably fine to drop it. 
it just calculated next date based on the cron. I don't think we break anything by not doing this

ui/src/views/Hooks/ListHooks/index.jsx
```diff
@@ -44,6 +36,50 @@ import Link from '../../../utils/Link';
...
+  }
+
+  fetchHooks = async () => {
+    const { hookGroupId } = this.props.match.params;
+
+    this.setState({ loading: true, error: null });
+
+    try {
+      const result = await this.hooksClient.listHooks(hookGroupId);
+      const hooks = await Promise.all(
```
lotas, line 60:
```suggestion
      const hooks = await Promise.allSettled(
```

probably single failed request shouldn't block the rest
and some groups might contain a lot of hooks, so let's load all

lotas (changes requested):

PR #9039, opened 2026-08-20
Bug 2062161 - Remove deprecated declareProvisioner, declareWorkerType and fix declareWorker 500 on empty body

Eijebong:
See each commit for details

PR #9040, opened 2026-08-20
Remove the concept of queue actions

Eijebong:
Those have not been usable for 6 years now since the only way to register them has been stubbed out since then. The UI confirms that with those nice `TODO: Action not working`. Since I didn't know about these before today and they're irrelevant anyway, into the trash it goes 🚮.

This contains #9039. We'll see if github allows me to stack them or not


Eijebong:
Who could've seen that one coming... No it doesn't allow me to stack them despite the docs saying it should.

lotas (dismissed):
r+ for all

lotas (approved):

PR #9042, opened 2026-08-21
feat(github): fix unsafe handling in as_slugid helper

lotas:
as_slugid was using `{}` object to store and lookup resolved slugids for tasks, which allowed referencing builtins like valueOf, toString, etc

### DIFF
--- a/services/github/src/tc-yaml.js
+++ b/services/github/src/tc-yaml.js
@@ -88,9 +88,11 @@ class VersionZero extends TcYaml {
   }
 
   substituteParameters(config, cfg, payload) {
+    // Ensure template parameters resolve only explicitly defined properties.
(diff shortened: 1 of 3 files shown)


services/github/src/tc-yaml.js
Eijebong, line 285: Isn't this the same problem?

lotas: I think `taskId` is defined by us, let me trace it

lotas: actually we only set a default one, but taskId might be coming from task definition yep

lotas: done, thanks!

Eijebong (approved):
Thank you

PR #9043, opened 2026-08-21
Bug 2065461 - Stop rendering emails through juice's resource inliner

Eijebong:
The notify service was rendering emails through juice which would happily fetch resources to inline them before sending the email.

The problem lies in the fact that we embed the original content as `!{content}` in the text part of email, which makes sense, we don't want to escape that, the email marks it as plain text. However, `email-templates` calls `juice.juiceResources` on every single part of the email, including that one, which ends up inlining things in the text content, and mangling the markdown formatted part of the message as well (`<https://example.com/a/b>` is perfectly valid text in markdown and would get mangled as `<https: example.com a b>` by juice. `cmd 2>&1 <input.txt` would lose the `<input.txt` part entirely, which is even worse).

Juice is only used because we need to inline CSS in the email. In 640f39c68d8fca128666b5783f3a4740efbc98bc, I mentioned being scared of removing that inlining because the email ecosystem is a clusterfuck. Turns out I was right. I tried removing juice and then beat gmail into rendering a local email (seriously, why is this so hard...) and lo and behold, it ate the `<style>` block despite mangling class names in the message which suggests that they process it... So removing juice entirely is out of the question.

So instead I switched how the rendering works to call `juice` ourselves, on the html part of the email only since that's the part that requires the CSS inlining. And I disabled `juice` through `email-templates` since there's no way to control what that does.

I have **thoughts** about the naming of that method and the option in `email-templates` that I won't be putting in there but note that I'm talking about `juice` and not `juice.juiceResources`, and that is a very important difference. `juice` only parses the content through `cheerio` and reserializes it, inlining CSS. It cannot embed any resource. `juice.juiceResources` on the other hand, while having a name that is more specific and very inconspicuous, will do that and happily fetch files and network resources in any kind of text.

The rest of the changes are that we have to include the style early through pug now instead of letting juice do it later on and that the CSS was actually invalid which ate the first rule. (`//` isn't a valid CSS comment)

### DIFF
--- a/changelog/bug-2065461.md
+++ b/changelog/bug-2065461.md
@@ -1,0 +1,7 @@
+audience: users
+level: patch
+reference: bug 2065461
+---
+Prevent the notify service from fetching files and remote URLs referenced in
+the content of the emails it sends. This also stops the plain text part of
+those emails from being mangled by the HTML processing.
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     "json-e": "^4.8.4",
     "json-parameterization": "^2.0.1",
     "jsonwebtoken": "^9.0.3",
+    "juice": "^11.1.1",
     "jwks-rsa": "^2.1.5",
     "lodash": "4.18.1",
     "loglevel": "^1.9.2",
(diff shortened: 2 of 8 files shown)


matt-boris (approved):
Looks good to me 👍🏻

PR #9044, opened 2026-08-21
Bug 2062170 - Pass proxy credentials from the worker through environment variables

Eijebong:
Since `capacity > 1`, multiple tasks could be running at the same time. On default configurations, at least on linux, a process can read the cmdline from other processes on the system so 2 tasks running concurrently could read eachothers credentials which would allow a bad actor to act as the other task.

The proxy already allows passing credentials through environment variables and that is inaccessible from other users on any given process.


matt-boris (approved):
Thanks

PR #9048, opened 2026-08-24
Bug 1917274 - Publish a pulse message when a push changes .taskcluster.yml

aminehmida:
Bugzilla Bug: [1917274](https://bugzilla.mozilla.org/show_bug.cgi?id=1917274)

Adds a new exchange, `exchange/taskcluster-github/v1/taskcluster-yml-update`, published when a push touches a repository's `.taskcluster.yml`. The ordinary `push` message is still published as well, since that exchange has other consumers.

This is the taskcluster half of [bug 1917274](https://bugzilla.mozilla.org/show_bug.cgi?id=1917274). The fxci-config half, a hook bound to this exchange that redeploys, comes later and needs this deployed to the Firefox-CI instance first. Design of record is [comment 5](https://bugzilla.mozilla.org/show_bug.cgi?id=1917274#c5), +1'd in [comment 6](https://bugzilla.mozilla.org/show_bug.cgi?id=1917274#c6).

## Detection

A push lists its changed files per commit, so the common case costs no API call. GitHub [caps that list at 2048 entries](https://docs.github.com/en/webhooks/webhook-events-and-payloads#push) and sends no count field, so a full array is the only signal that commits were dropped. When the list is full and no listed commit names the file, the service asks GitHub directly rather than assume the file was untouched. It makes two [Get repository content](https://docs.github.com/en/rest/repos/contents#get-repository-content) calls, one at each end of the push:

```
GET /repos/{owner}/{repo}/contents/.taskcluster.yml?ref={before}
GET /repos/{owner}/{repo}/contents/.taskcluster.yml?ref={after}
```

`before` is the commit the branch was on before the push, `after` is the new tip, so the pair brackets every commit in the push including the ones GitHub did not send. If the two responses differ, something changed the file and we publish. A 404 is treated as "the file did not exist at that commit" rather than as an error, which is what makes a file added or deleted in an unsent commit come out right.

The two calls are skipped when `before` or `after` is the all-zeroes SHA, which is how GitHub represents a ref that was just created or just deleted. There is nothing to read at a commit that does not exist.

## Security

A push in one repository will eventually cause a task to run in another, so the payload deliberately carries only the organization and repository names. No file contents, no commit messages, no branch, no author. Both fields are already validated and sanitized by tc-gh, and the schema sets `additionalProperties: false` so the shape is enforced rather than merely intended.

## Three things I would like pushed back on

1. **The exchange name.** `taskcluster-yml-update`. [Comment 5](https://bugzilla.mozilla.org/show_bug.cgi?id=1917274#c5) suggested `meta/taskcluster-yml/update`, but exchange slugs here are flat, so that would render as `.../v1/meta/taskcluster-yml/update`. Renaming after release is a breaking change.
2. **`organization` in the payload.** [Comment 5](https://bugzilla.mozilla.org/show_bug.cgi?id=1917274#c5) says "include the repository name in the payload" and does not mention the organization. I added it because the consuming hook keys projects by owner/repo, and because all five existing github exchanges carry it. It is not user input, so it breaks neither rule stated there, but it is a one-line revert if you disagree.
3. **The truncation fallback at all.** At a 2048-commit cap this is a close call, and plain "trust the list" is defensible. I kept the fallback because the failure it prevents is silent, but it is roughly 15 lines that will almost never run.

Not addressed, and pre-existing: GitHub [does not deliver push payloads over 25 MB](https://docs.github.com/en/webhooks/webhook-events-and-payloads) at all, so a sufficiently large push is invisible to the existing `push` exchange too.

## Tests

`services/github/test/pulse_test.js` covers the file appearing in `added`, `modified` and `removed`, a push that misses it, a branch deletion, the `push` message still firing alongside, and three truncation cases. The 2048-commit bodies are built in the test rather than stored as fixtures, which needed a small helper to re-sign a modified body.

299 passing in `services/github`.


changelog/bug-1917274.md
```diff
@@ -0,0 +1,16 @@
...
+message when a push changes a repository's `.taskcluster.yml`. The ordinary `push`
+message is still published as well.
+
+The payload carries only the organization and repository names. Consumers act on a
+repository they already know about, so a push in one repository can trigger work in
+another without any data the pushing user controls crossing the boundary.
+
+GitHub truncates the push payload's commit list at 2048 entries. When the list is
+full and no listed commit names the file, the service reads `.taskcluster.yml` at
+both ends of the push to decide.
```
Eijebong: This is an implementation detail and IMO doesn't belong in the changelog

services/github/src/api.js
```diff
@@ -23,6 +23,61 @@ function sanitizeGitHubField(field) {
...
+// count field, so a full array is the only signal that commits were dropped.
+// https://docs.github.com/en/webhooks/webhook-events-and-payloads#push
+const COMMITS_LIMIT = 2048;
+
+// `before` on a created ref, and `after` on a deleted one.
+const NULL_SHA = '0'.repeat(40);
+
+// Read `.taskcluster.yml` at a revision. A missing file is not an error, it
+// means the file did not exist there.
+async function getTaskclusterYml(instGithub, { owner, repo, ref }) {
```
Eijebong: https://github.com/taskcluster/taskcluster/blob/9d59eb59f6fd0912075c194100d8910f2da40f52/services/github/src/handlers/index.js#L554-L576

This is basically the same except `getYml` has some more error handling that seems to be here because of github knowingly misbehaving.

You probably want to extract the `getContent` into a `getRawYml` and then switch `getYml` to `getRawYml` -> `yaml.load` and the use `getRawYml` here?

services/github/src/api.js
```diff
@@ -436,6 +491,14 @@ builder.declare(
     msg.eventId = eventId;
 
     await this.publisher[publisherKey](msg);
+
+    if (eventType === EVENT_TYPES.PUSH && (await pushChangedTaskclusterYml(instGithub, body))) {
```
Eijebong: I got a bit confused by this because I didn't immediately see the `eventType` guard and thought it was firing on all events. I can see why it's not in the switch but IMO this a bit of a smell. I wonder if this whole thing should be in a handler in `handlers/job.js` instead.

Eijebong: Actually something else that would move me towards making this a job is that right now this could block the webhook handler until this is resolved. It might be slow if we have to wait for github to respond and worse, it might fail (I know github never 500s...). Failure would mean delivery would get marked as failed on github despite having triggered everything else already and someone might get tempted to just redeliver the event which would dupe all of the above.

Moving it to a job solves both issues. It unfortunately doesn't fix the "github might 500" since we apparently ACK messages even if handlers fail (https://github.com/taskcluster/taskcluster/blob/9d59eb59f6fd0912075c194100d8910f2da40f52/services/github/src/handlers/index.js#L137) which is probably wrong but very much out of scope here.

services/github/src/api.js
```diff
@@ -23,6 +23,61 @@ function sanitizeGitHubField(field) {
...
+  const commits = body.commits || [];
+  const touched = commits.some(({ added = [], removed = [], modified = [] }) =>
+    [...added, ...removed, ...modified].includes(TASKCLUSTER_YML)
+  );
+  if (touched) {
+    return true;
+  }
+
+  // A created or deleted ref has no counterpart to compare against.
+  if (commits.length < COMMITS_LIMIT || body.before === NULL_SHA || body.after === NULL_SHA) {
```
Eijebong: AFAICT, this is reimplementing `body.created` / `body.deleted`. We can use those and not need `NULL_SHA` which can the go away.

https://docs.github.com/en/webhooks/webhook-events-and-payloads#push

```
created boolean Required

Whether this push created the ref.

---

deleted boolean Required

Whether this push deleted the ref.
```

services/github/src/api.js
```diff
@@ -23,6 +23,61 @@ function sanitizeGitHubField(field) {
...
+  }
+}
+
+// Decide whether a push changed `.taskcluster.yml`.
+//
+// The commit list answers this on its own unless GitHub truncated it, in which
+// case the file may have changed in a commit we were never told about. Reading
+// the file at both ends of the push is then the only way to find out.
+async function pushChangedTaskclusterYml(instGithub, body) {
+  const commits = body.commits || [];
```
Eijebong: Curious why you would `|| []` here. The github docs says this is a mandatory parameter so I don't see how it could be missing

services/github/src/api.js
```diff
@@ -23,6 +23,61 @@ function sanitizeGitHubField(field) {
...
+  const touched = commits.some(({ added = [], removed = [], modified = [] }) =>
+    [...added, ...removed, ...modified].includes(TASKCLUSTER_YML)
```
Eijebong: One thing that got missed here, the doc says:

`A maximum of 3000 changed files will be reported per commit.`. We already have a fallback for the number of commits in a branch but we need to fallback if any commit contains more than 3k files changed too.

Something along the line of:

```
const listComplete = commits.length < COMMITS_TRUST_LIMIT &&
    commits.every(c => c.added.length + c.removed.lengthh + c.modified.length < FILES_TRUST_LIMIT)
```

Eijebong: (probably have to handle those arrays not being there, the doc doesn't mark them as required)

services/github/src/api.js
Eijebong: Doesn't this also swallow a `.taskcluster.yml` added to a branch with 2.5k+ commits on a new branch

services/github/src/api.js
Eijebong, line 323: If the HEAD of a branch is `[ci skip]` and the push contains a taskcluster yml change, it gets completely skipped right now. This kinda conflicts with my comment saying that we should move the whole thing into a job but I think it might be worth refactoring this as a precursor for this PR to still publish the event with a new field set and then ignore it in the job handler instead of completely ignoring it.

Eijebong: @lotas or @matt-boris might have thoughts on that since they both touched that feature.

The doc clearly says:

> If you push your changes and head commit will include `[ci skip]` or `[skip ci]` (case insensitive) in its message, no tasks would be generated for such commit.

Not that a message wouldn't get published

services/github/schemas/v1/taskcluster-yml-update-message.yml
```diff
@@ -0,0 +1,30 @@
+$schema: "/schemas/common/metaschema.json#"
+title:        "Taskcluster Yml Update Message"
+description: |
+  Message reporting that a push changed a repository's `.taskcluster.yml`.
+
+  The payload is deliberately limited to the identity of the repository.
+  Consumers act on a repository they already know about, so nothing an author of
+  the pushed commits controls needs to travel with the message.
+type:         object
+properties:
```
Eijebong, line 10: You probably want at least the ref name in there too so you can skip hook generation early if the change happened on a branch we don't care about.

Eijebong: Might also want the event ID for debugging purposes? 🤔

services/github/src/api.js
```diff
@@ -436,6 +491,14 @@ builder.declare(
     msg.eventId = eventId;
 
     await this.publisher[publisherKey](msg);
+
+    if (eventType === EVENT_TYPES.PUSH && (await pushChangedTaskclusterYml(instGithub, body))) {
+      await this.publisher[PUBLISHERS.TASKCLUSTER_YML_UPDATE]({
+        organization: msg.organization,
```
Eijebong: This is a sanitized version of the org which might differ from the actual org name. Same thing below with repository.

If repo name contains a `.`, you'll publish those names with a `%` instead. It's necessary for the routing key but for the payload that feels wrong to have the caller know about this.

Eijebong (changes requested):
I didn't read tests since there's quite a bit that needs changing in code

bhearsum:
I'm not the right person to review the code, but to some of the higher level questions:

>     1. **The exchange name.** `taskcluster-yml-update`. [Comment 5](https://bugzilla.mozilla.org/show_bug.cgi?id=1917274#c5) suggested `meta/taskcluster-yml/update`, but exchange slugs here are flat, so that would render as `.../v1/meta/taskcluster-yml/update`. Renaming after release is a breaking change.

I think my spec'ing out in the bug was a bit thin.

We already have a `meta/taskcluster-yml/update` exchange that includes the org and repo in the routing key. Presumably we could add new data to the existing messages here that give us what we need to filter these messages by those that contain taskcluster.yml updates, and what's needed to rebuild the hooks?

I think using a new exchange is OK as well though; I don't feel terribly strongly about this.

>     2. **`organization` in the payload.** [Comment 5](https://bugzilla.mozilla.org/show_bug.cgi?id=1917274#c5) says "include the repository name in the payload" and does not mention the organization. I added it because the consuming hook keys projects by owner/repo, and because all five existing github exchanges carry it. It is not user input, so it breaks neither rule stated there, but it is a one-line revert if you disagree.

Good catch; yes, we need the org as well to fully identify the repo.

It might also be OK to drop the org and repo from the payload, and look at the routing key instead?

> 
>     3. **The truncation fallback at all.** At a 2048-commit cap this is a close call, and plain "trust the list" is defensible. I kept the fallback because the failure it prevents is silent, but it is roughly 15 lines that will almost never run.
> 
> 
> Not addressed, and pre-existing: GitHub [does not deliver push payloads over 25 MB](https://docs.github.com/en/webhooks/webhook-events-and-payloads) at all, so a sufficiently large push is invisible to the existing `push` exchange too.

Unfortunate...but livable. In this rare scenario we end up with broken hooks like we do now, and will rebuild them by hand.

aminehmida:
@bhearsum @Eijebong I think I have addressed all your comments and now ready for review again

services/github/src/handlers/taskclusterYml.js
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Decide whether a push changed `.taskcluster.yml`.
+ *
+ * The commit list answers this on its own unless GitHub truncated it, either by
+ * dropping commits or by dropping file names within one, in which case the file
+ * may have changed somewhere we were never told about. Reading the file at both
+ * ends of the push is then the only way to find out.
+ */
+async function pushChangedTaskclusterYml(instGithub, body) {
```
Eijebong: So I checked and github is really not helpful here... Turns out on force pushes that revert a commit, it doesn't tell you the file changed since it only computes the diff between the closest common ancestor and the push.

So `git reset --hard HEAD~1 && git push -f` gives you "nothing changed".

I don't think we should do anything about this because the only way to detect this is to detect the force push and then go through the diff path which means more requests to github on every single force push. Which is not something we'd want given the rate limits they have. But it's once again showing that this is at most a "best effort" kind of thing. Probably want to document that.

services/github/src/exchanges.js
```diff
@@ -109,6 +109,31 @@ exchanges.declare({
   CCBuilder: () => [],
 });
 
+/** taskcluster.yml update exchange */
+exchanges.declare({
```
Eijebong, line 113: So one issue of doing this is that now if someone pushes a commit with `[ci skip]` on the HEAD, that message is never sent.

Eijebong: Right I had called that out but forgot. Let's ignore that for now. I opened https://github.com/taskcluster/taskcluster/issues/9121 which would fix that concern

services/github/src/api.js
Eijebong: https://github.com/taskcluster/taskcluster/issues/9121

services/github/src/exchanges.js
```diff
@@ -109,6 +109,31 @@ exchanges.declare({
...
+  exchange: 'taskcluster-yml-update',
+  name: PUBLISHERS.TASKCLUSTER_YML_UPDATE,
+  title: 'Taskcluster Yml Update Event',
+  description: [
+    "When a GitHub push event changes a repository's `.taskcluster.yml` it will",
+    'be broadcast on this exchange with the designated `organization` and',
+    '`repository` in the routing-key.',
+    '',
+    'The payload carries only the identity of the repository, so that a push in',
```
Eijebong: This isn't true now that you have the ref in there?

services/github/schemas/v1/taskcluster-yml-update-message.yml
```diff
@@ -0,0 +1,44 @@
...
+  version:    {$const: message-version}
+  organization:
+    description: |
+      The GitHub `organization` which had an event, as GitHub spells it.  The
+      routing key carries the same name with periods replaced by `%`.
+    type:         string
+    maxLength:    {$const: github-identifier-max-length}
+    pattern:      {$const: github-repo-name-pattern}
```
Eijebong: This should be `github-webhook-identifier-pattern` I think. That's what's used everywhere else for repo owners.

Eijebong: I don't think it can actually change (the pattern is there because of `[bot]` users/owners but I don't think those can have a fork. But let's keep it consistent.

Also you'll need to add a minLength: `{$const: github-identifier-min-length}` because `github-webhook-identifier-pattern` accepts length 0 by itself.

services/github/src/handlers/taskclusterYml.js
```diff
@@ -0,0 +1,77 @@
...
+ * slow or failing GitHub API call cannot fail the webhook delivery, which would
+ * invite a redelivery of an event whose tasks have already been created.
+ **/
+export async function taskclusterYmlHandler(message) {
+  const { body, eventId, installationId } = message.payload;
+  const owner = body.repository.owner.name;
+  const repo = body.repository.name;
+  const debug = makeDebug(this.monitor, { eventId, installationId, owner, repo });
+
+  const instGithub = await this.context.github.getInstallationGithub(installationId);
```
Eijebong: We should create this in `pushChangedTaskclusterYml` right before it gets used.

Right now on every push this tries to get a token and then drops it without using it most of the time. We should avoid those extra requests to github.

services/github/src/handlers/taskclusterYml.js
```diff
@@ -0,0 +1,77 @@
...
+ * The commit list answers this on its own unless GitHub truncated it, either by
+ * dropping commits or by dropping file names within one, in which case the file
+ * may have changed somewhere we were never told about. Reading the file at both
+ * ends of the push is then the only way to find out.
+ */
+async function pushChangedTaskclusterYml(instGithub, body) {
+  const commits = body.commits;
+  const touched = commits.some(({ added = [], removed = [], modified = [] }) =>
+    [...added, ...removed, ...modified].includes(TASKCLUSTER_YML_PATH)
+  );
```
Eijebong:
```js
const touched = commits.some(
  ({ added = [], removed = [], modified = [] }) =>
    modified.includes(TASKCLUSTER_YML_PATH) ||
    added.includes(TASKCLUSTER_YML_PATH) ||
    removed.includes(TASKCLUSTER_YML_PATH) 
);
```

would avoid a potentially enormous allocation

Eijebong (changes requested):
This is in a pretty good shape now. Got a few things left and you'll need to rebase but it's fairly minor

services/github/src/handlers/taskclusterYml.js
amine-mawaqit: Documented it in the JSDoc, plus a line in the exchange description so consumers see it too. Let me know if there is another/better place that

services/github/src/exchanges.js
amine-mawaqit: Thanks for opening that PR 👍

services/github/src/exchanges.js
amine-mawaqit: oops! Forgot to update that. Fixed now. Thanks

services/github/schemas/v1/taskcluster-yml-update-message.yml
amine-mawaqit: Done

amine-mawaqit:
@Eijebong Thanks for the second round of reviews. I made all the requested changes. Feel free to merge now if all looks good to you

services/github/src/handlers/taskclusterYml.js
aminehmida: Documented it in the JSDoc, plus a line in the exchange description so consumers see it too. Let me know if there is another/better place that

services/github/src/exchanges.js
aminehmida: Thanks for opening that PR 👍

services/github/src/exchanges.js
aminehmida: oops! Forgot to update that. Fixed now. Thanks

services/github/schemas/v1/taskcluster-yml-update-message.yml
aminehmida: Done

services/github/src/handlers/taskclusterYml.js
aminehmida: Good catch. Sorted now.

services/github/src/handlers/taskclusterYml.js
aminehmida: Fair point. Sorted

aminehmida:
@Eijebong Thanks for the second round of reviews. I made all the requested changes. Feel free to merge now if all looks good to you

services/github/schemas/v1/taskcluster-yml-update-message.yml
lotas: I wonder if it's worth including actor/user for this event as well.. to exclude some bots or automated updates

Eijebong (approved):
Alright. This is probably in a good enough shape now. I'm assuming we're going to hit some issues with how github's documentation is lying, but short of putting this in prod there's no good way of testing that won't make you go bald.

Issue #7369, opened 2024-11-01
Crash when the downloads directory is missing


Eijebong:
Steps to reproduce:

1. Get a task to put something in the downloads folder
2. Remove the downloads folder
3. Run the task again

Version: 72.1.0
Example task: https://taskcluster.bananium.fr/tasks/XG4ARWQHTeapX58w_BPliw

Stacktrace:

```
[taskcluster:error] [mounts] Could not fetch from task TeeslIAUTN-VlQJBv7uKQQ artifact public/image.tar.zst into file downloads/Jkl7ENGbTnued3N5t7SDiA due to open downloads/Jkl7ENGbTnued3N5t7SDiA: no such file or directory
[taskcluster 2024-11-01T20:07:44.960Z] Uploading link artifact public/logs/live.log to artifact public/logs/live_backing.log with expiry 2025-11-01T20:07:40.726Z
[taskcluster:error] [mounts] open downloads/Jkl7ENGbTnued3N5t7SDiA: no such file or directory
[taskcluster:error] goroutine 1 [running]:
[taskcluster:error] runtime/debug.Stack()
[taskcluster:error] 	/home/task_172891650450432/go/go/src/runtime/debug/stack.go:26 +0x5e
[taskcluster:error] main.(*TaskRun).Run.func2()
[taskcluster:error] 	/home/task_172891650450432/taskcluster/workers/generic-worker/main.go:995 +0x9d
[taskcluster:error] panic({0xa71140?, 0x1078e40?})
[taskcluster:error] 	/home/task_172891650450432/go/go/src/runtime/panic.go:785 +0x132
[taskcluster:error] github.com/taskcluster/taskcluster/v72/workers/generic-worker/process.(*Result).ExitCode(...)
[taskcluster:error] 	/home/task_172891650450432/taskcluster/workers/generic-worker/process/process.go:41
[taskcluster:error] main.(*TaskMount).shouldPurgeCaches(0x0?)
[taskcluster:error] 	/home/task_172891650450432/taskcluster/workers/generic-worker/mounts.go:449 +0x3e
[taskcluster:error] main.(*TaskMount).Stop(0xc0003d0480, 0xc00036e768)
[taskcluster:error] 	/home/task_172891650450432/taskcluster/workers/generic-worker/mounts.go:424 +0x32
[taskcluster:error] main.(*TaskRun).Run.func3({{0xc6c650?, 0xc0003d0480?}, {0xc6d480?, 0x10b31e0?}})
[...]

PR #9051, opened 2026-08-25
fix(generic-worker): don't leak disk space when a cache deletion fails

### LINKED ISSUE #8944: A failed cache eviction leaks disk space permanently

matt-boris:
Fixes #8944.

>Generic Worker no longer leaks disk space when cache files remain on disk without a cache table entry. Garbage collection and worker startup now delete anything in the caches directory that the worker does not know about, so a failed deletion (or a leftover from a crash) is retried instead of occupying space forever.

### DIFF
--- a/workers/generic-worker/mounts_feature.go
+++ b/workers/generic-worker/mounts_feature.go
@@ -414,8 +414,8 @@ func (cm *CacheMap) LoadFromFile(stateFile string, cacheDir string) {
 func (feature *MountsFeature) Initialise() error {
 	cacheMutex.Lock()
 	defer cacheMutex.Unlock()
-	fileCaches.LoadFromFile("file-caches.json", config.CachesDir)
-	directoryCaches.LoadFromFile("directory-caches.json", config.DownloadsDir)
+	fileCaches.LoadFromFile("file-caches.json", config.DownloadsDir)
+	directoryCaches.LoadFromFile("directory-caches.json", config.CachesDir)
 	sweepUnknownContent(config.CachesDir, directoryCaches)
 	return nil
 }
--- a/workers/generic-worker/mounts_feature.go
+++ b/workers/generic-worker/mounts_feature.go
@@ -257,6 +257,30 @@ func (cache *Cache) Rating() float64 {
 	return float64(cache.LastUsed.Unix())
 }
 
+// sweepUnknownContent deletes anything in dir that is not a cache Location.
+// Failed deletions are retried on the next call (see #8944).
+func sweepUnknownContent(dir string, caches CacheMap) {
+	keep := map[string]bool{}
+	for _, entries := range caches {
+		for _, cache := range entries {
+			keep[filepath.Base(cache.Location)] = true
+		}
+	}
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return
+	}
+	for _, entry := range entries {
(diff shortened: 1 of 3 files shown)


workers/generic-worker/mounts_feature.go
```diff
@@ -46,6 +46,11 @@ var (
 
 	// fileCacheDownloads deduplicates concurrent downloads for the same cache key.
 	fileCacheDownloads singleflight.Group
+
+	// undeletedCacheLocations holds cache content that failed to be deleted
+	// from disk, so that garbage collection can retry it. See
+	// removeCacheContent. Protected by cacheMutex.
+	undeletedCacheLocations []string
```
Eijebong: This isn't enough as-is. If the worker panics/reboots, you lose that and orphan anything that was still in it.

Eijebong (changes requested):
I don't think this would work well enough. The list isn't persisted so it would be lost anyway on restart (and if a cache is ever orphaned because it doesn't have a `Cache` entry it's also leaked forever, although I'm not sure that can happen). Also I had not realized but downloads have the same kind of problem. Any download that gets aborted leaks forever. And I would really like to avoid piling yet another way of clearing file on top of this.

What would be nice would be a sweep on both caches and downloads and removing what we don't know about. It would make the current PR entirely redundant but a lot more robust. It's a problem though because downloads happen without holding the cache mutex so we can't distinguish between a download still in flight and one that's orphaned. It would also allow fixing #7369 by making the reconciliation bidirectional.

I wonder if we could do the sweep on caches only and do downloads as a followup, it probably requires downloading them in a "staging" folder or as a .partial, deleting those on start and ignoring them during runtime.

workers/generic-worker/mounts_feature.go
matt-boris: Yeah I can just take the entirely different approach of changing the mounts feature to wipe anything in caches/downloads dirs that we don't know about

workers/generic-worker/mounts_feature.go
```diff
@@ -390,8 +414,9 @@ func (cm *CacheMap) LoadFromFile(stateFile string, cacheDir string) {
 func (feature *MountsFeature) Initialise() error {
 	cacheMutex.Lock()
 	defer cacheMutex.Unlock()
-	fileCaches.LoadFromFile("file-caches.json", config.CachesDir)
-	directoryCaches.LoadFromFile("directory-caches.json", config.DownloadsDir)
+	fileCaches.LoadFromFile("file-caches.json", config.DownloadsDir)
+	directoryCaches.LoadFromFile("directory-caches.json", config.CachesDir)
```
matt-boris, line 418: Tiny bugfix here - swapped the worker cache dirs

lotas: I'm not sure what kind of magic is this, can't trace why they are created in different directories

Eijebong: ftr I'd get that in its own commit in case it ends up breaking things we can just revert that

lotas (approved):

PR #9052, opened 2026-08-26
fix(generic-worker): don't track `cacheMutex` ownership with a bool in `Mount` `recover()`

matt-boris:
>Generic Worker: writable directory cache `Mount` no longer tracks `cacheMutex` ownership with a side boolean in its panic recover path. Acquire/create runs under a lock that is always released on return, and recover always takes the lock after that.

### DIFF
--- a/workers/generic-worker/mounts_feature.go
+++ b/workers/generic-worker/mounts_feature.go
@@ -129,7 +129,7 @@ func ReleaseCache(entry *Cache) {
 
 // newPoolEntry creates a new in-use cache pool entry for the given cache name.
 // Caller must hold cacheMutex.
-func newPoolEntry(cacheName string) (*Cache, error) {
+func newPoolEntry(cacheName string) *Cache {
 	basename := slugid.Nice()
 	file := filepath.Join(config.CachesDir, basename)
 	return &Cache{
@@ -139,7 +139,23 @@ func newPoolEntry(cacheName string) (*Cache, error) {
 		Owner:    directoryCaches,
 		Key:      cacheName,
 		InUse:    true,
-	}, nil
+	}
+}
+
+// acquireOrCreateCache returns an in-use pool entry for name, creating one if
+// none is available.
+//
+// The entry is never nil. The bool is true if this call created the entry,
+// false if an existing pool entry was reused.
+func acquireOrCreateCache(name string) (*Cache, bool) {
+	cacheMutex.Lock()
+	defer cacheMutex.Unlock()
+	if entry := AcquireCache(name); entry != nil {
+		return entry, false
+	}
+	entry := newPoolEntry(name)
+	directoryCaches[name] = append(directoryCaches[name], entry)
+	return entry, true
 }
 
 // trimPoolExcess evicts available entries beyond config.Capacity per cache name.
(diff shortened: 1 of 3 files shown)


workers/generic-worker/mounts_feature.go
```diff
@@ -744,113 +764,77 @@ func (f *FileMount) FSContent() (FSContent, error) {
...
+		return err
 	}
 
-	lock()
-	entry := AcquireCache(w.CacheName)
+	// released is set before taking cacheMutex in releaseEntry so a panic
+	// during Evict cannot make the recover handler lock again.
+	released := false
+	releaseEntry := func(evict bool) error {
+		if released || entry == nil {
```
Eijebong: `entry` here can never be nil

Eijebong: "On success the entry is never nil." from the comment on `acquireOrCreateCache`

workers/generic-worker/mounts_feature.go
```diff
@@ -744,113 +764,77 @@ func (f *FileMount) FSContent() (FSContent, error) {
...
-		mutexHeld = false
+	entry, created, err := acquireOrCreateCache(w.CacheName)
+	if err != nil {
+		return err
 	}
 
-	lock()
-	entry := AcquireCache(w.CacheName)
+	// released is set before taking cacheMutex in releaseEntry so a panic
+	// during Evict cannot make the recover handler lock again.
```
Eijebong: I don't think this is true. The `defer cacheMutex.Unlock()` would run on panic *before* any `recover()` since the closure frame would unwind first.

The only necessary thing here is that `released` is set before anything could panic since that's the only reentrant path into `releaseEntry`

workers/generic-worker/mounts_feature.go
```diff
@@ -142,6 +142,26 @@ func newPoolEntry(cacheName string) (*Cache, error) {
...
+// On success the entry is never nil. The bool is true if this call created the
+// entry, false if an existing pool entry was reused. On error, entry is nil
+// and the bool is false.
+func acquireOrCreateCache(name string) (*Cache, bool, error) {
+	cacheMutex.Lock()
+	defer cacheMutex.Unlock()
+	if entry := AcquireCache(name); entry != nil {
+		return entry, false, nil
+	}
+	entry, err := newPoolEntry(name)
```
Eijebong: This can't fail. `newPoolEntry` just above is `return &Cache, nil`. Might be worth cleaning that up here. Not sure why it's even returning an error...

workers/generic-worker/mounts_feature.go
```diff
@@ -744,113 +764,77 @@ func (f *FileMount) FSContent() (FSContent, error) {
...
+	if err != nil {
+		return err
 	}
 
-	lock()
-	entry := AcquireCache(w.CacheName)
+	// released is set before taking cacheMutex in releaseEntry so a panic
+	// during Evict cannot make the recover handler lock again.
+	released := false
+	releaseEntry := func(evict bool) error {
```
Eijebong: I really don't like boolean args. Can we make this 2 closures instead?

Something like:

```go
evictEntry := func() error {
        if released {
                return nil
        }
        released = true
        delete(taskMount.poolEntries, w)
        cacheMutex.Lock()
        defer cacheMutex.Unlock()
        return entry.Evict(taskMount)
}

returnEntryToPool := func() {
        if released {
                return
        }
        released = true
        delete(taskMount.poolEntries, w)
        cacheMutex.Lock()
        defer cacheMutex.Unlock()
        ReleaseCache(entry)
}
```

It also avoids that weird `releaseErr` being conditionally set.

workers/generic-worker/mounts_feature.go
```diff
@@ -744,113 +764,77 @@ func (f *FileMount) FSContent() (FSContent, error) {
...
+		released = true
+		cacheMutex.Lock()
+		defer cacheMutex.Unlock()
+		var releaseErr error
+		if evict {
+			releaseErr = entry.Evict(taskMount)
+		} else {
+			ReleaseCache(entry)
+		}
+		delete(taskMount.poolEntries, w)
```
Eijebong, line 919: AFAICT, this doesn't need to be in the lock. My suggestion above already gets it out of the locked section

Eijebong (approved):
Thank you for fixing that

PR #9055, opened 2026-08-27
feat(ci): Switch taskcluster.yml to public_restricted

lotas:
This isolates tasks and tests for non-collaborators to only run the ones that don't involve using secrets and use pull-request-untrusted role

This should land together with
https://github.com/taskcluster/community-tc-config/pull/998

Bug: [2066797](https://bugzilla.mozilla.org/show_bug.cgi?id=2066797)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -3,43 +3,44 @@ version: 1
 reporting: checks-v1
 policy:
   pullRequests: public_restricted
+  allowComments: collaborators
 autoCancelPreviousChecks: true
 tasks:
   - $let:
       trustDomain: taskcluster
       # We set some of these directly in tc because we don't have some mozilla-specific concepts
       ownerEmail: taskcluster-internal@mozilla.com
-      # Keep writable caches from untrusted code out of the trusted level-1
-      # namespace. The Community-TC role enforces the corresponding scopes.
+      # Match Firefox CI trust levels: external pull requests use level 1,
+      # while trusted code and collaborator-approved runs use level 3.
       level:
         $if: 'tasks_for == "github-pull-request-untrusted"'
-        then: 0
-        else: 1
+        then: 1
+        else: 3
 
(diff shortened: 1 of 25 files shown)


taskcluster/src/target_tasks.py
```diff
@@ -11,6 +19,36 @@ def target_tasks_taskcluster_branches(full_task_graph, parameters, graph_config)
...
+        has_secret_scope = any(is_secret_scope(scope) for scope in task.task.get("scopes", []))
+        if parameters["tasks_for"] == UNTRUSTED_PULL_REQUEST and has_secret_scope:
+            logger.info("Skipping secret-dependent task %s in an untrusted pull request", task.label)
+            return False
+        return True
 
     return [label for label, t in full_task_graph.tasks.items() if filter(t)]
+
+
+@verifications.add("target_task_graph")
```
petemoore: nice

.taskcluster.yml
```diff
@@ -2,14 +2,19 @@
...
   - $let:
       trustDomain: taskcluster
       # We set some of these directly in tc because we don't have some mozilla-specific concepts
       ownerEmail: taskcluster-internal@mozilla.com
-      level: 1
+      # Keep writable caches from untrusted code out of the trusted level-1
+      # namespace. The Community-TC role enforces the corresponding scopes.
+      level:
+        $if: 'tasks_for == "github-pull-request-untrusted"'
+        then: 0
```
petemoore: is this right? i though level 1 was untrusted and level 3 was trusted?

Eijebong: FTR I'd rather we match fxci here. With 1 being untrusted and 3 being trusted. Having L1 be trusted for taskcluster and untrusted for everyone else would be very confusing. Especially true as we merge projects into fxci

lotas: agreed, applying same 1-3 level scheme

PR #9057, opened 2026-08-27
Don't ignore upload failures for optional artifacts

### LINKED ISSUE #9056: Optional artifacts should fail a task if they're present but fail to upload
Right now, we ignore all failure on upload for optional artifacts:

```go
e := task.uploadArtifact(taskArtifact)
 if e != nil {
     // we don't care about optional artifacts failing to upload
     if taskArtifact.Base().Optional {
         return nil
     }
     uploadErrChan <- e
 }
```

However, an optional artifact means that it might be missing from what the task produces, not that we should accept s3 closing the connection on us mid upload. We should differentiate the two error types and fail if the artifact is present but failed to upload for another reason.

Eijebong:
An artifact being marked as optional means that the task is allowed to not produce it. When it does produce said artifact though, it should fail the task if it fails to upload it for any given reason.

The previous behavior ignored any upload error if the artifact was marked as optional. That covers every single artifact uploaded by a task using d2g since it marks them all as optional which means that it would resolve as complete despite the artifact being there and S3 saying no.

This changes that behavior to continue ignoring artifact upload failures when the artifact is marked as optional except if it's present on disk. Artifacts that are missing are resolved early on to `ErrorArtifact` so we can just detect that and ignore the upload error in that case.

I moved some code around to avoid matching on `ErrorArtifact` twice, but the behavior is unchanged except for the "optional & content is there" case where it now fails the task.

Fixes #9056

### DIFF
--- a/workers/generic-worker/artifacts_test.go
+++ b/workers/generic-worker/artifacts_test.go
@@ -762,6 +762,33 @@ func TestMissingOptionalFileArtifactDoesNotFailTest(t *testing.T) {
 	expectedArtifacts.Validate(t, taskID, 0)
 }
 
+func TestOptionalArtifactUploadFailureFailsTask(t *testing.T) {
+	if os.Getenv("GW_TESTS_USE_EXTERNAL_TASKCLUSTER") != "" {
+		t.Skip("This test requires mock services")
+	}
+
+	setup(t)
+
+	expires := tcclient.Time(time.Now().Add(time.Minute * 30))
+	payload := GenericWorkerPayload{
+		Command:    copyTestdataFile("SampleArtifacts/_/X.txt"),
+		MaxRunTime: 30,
+		Artifacts: []Artifact{
+			{
+				Path:     "SampleArtifacts/_/X.txt",
+				Expires:  expires,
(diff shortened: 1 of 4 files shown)


matt-boris (dismissed):

workers/generic-worker/artifacts_test.go
```diff
@@ -762,6 +762,29 @@ func TestMissingOptionalFileArtifactDoesNotFailTest(t *testing.T) {
 	expectedArtifacts.Validate(t, taskID, 0)
 }
 
+func TestOptionalArtifactUploadFailureFailsTask(t *testing.T) {
+	setup(t)
```
matt-boris, line 770: You'll need to add something like this so this only runs against the mocks https://github.com/taskcluster/taskcluster/blob/ac99a93b4924049536d2bdbf160909da6751974a/workers/generic-worker/main_test.go#L53-L55

```suggestion
	if os.Getenv("GW_TESTS_USE_EXTERNAL_TASKCLUSTER") != "" {
		t.Skip("This test requires mock services")
	}

	setup(t)
```

PR #9059, opened 2026-08-30
fix: claimed tasks and pending tasks pages to REST

### LINKED ISSUE #8992: Migrate Claimed Tasks and Pending Tasks pages from GraphQL to REST

thakkarujas5:
Fixes #8992


lotas (approved):
Thanks! 
I tested this locally, it works. 
There is some code duplication regarding reading worker pool info for navigation, but we can refactor this later. Thanks

PR #9061, opened 2026-08-30
fix: view provisioners and view worker types pages to REST

### LINKED ISSUE #9060: Migrate ViewProvisioners and ViewWorkerTypes pages from GraphQL to REST

thakkarujas5:
Fixes #9060


thakkarujas5:
The previous ViewWorkerTypes page used internal rest endpoint queue.pendingTasks to get the pending task count for this page.

On conversion to REST I found out that this endpoint is deprecated and instead used the endpoint queue.taskQueueCounts. But the problem is that this endpoint required the scope queue:claimed-count:<taskQueueId> which is not provided to anonymous role in community instance, so it would fail on remote.

lotas:
> On conversion to REST I found out that this endpoint is deprecated and instead used the endpoint queue.taskQueueCounts. But the problem is that this endpoint required the scope queue:claimed-count: which is not provided to anonymous role in community instance, so it would fail on remote.

yes, if old endpoints didn't check permissions, and new ones do, we should follow that. UI can display 'n/a' if scopes are missing

lotas (approved):
tested several times locally, added small batching with immediate set state calls, to update UI as responses arrive. lgtm

PR #9068, opened 2026-08-31
fix(tests): use ephemeral ports via `testing.getFreePort()` to avoid intermittent `EADDRINUSE` failures

matt-boris:
This should be the proper fix for #3665.

>Service and library tests now allocate ephemeral ports via `testing.getFreePort()`, avoiding intermittent `EADDRINUSE` failures when multiple suites start HTTP servers.


libraries/api/test/context_test.js
```diff
@@ -55,8 +55,10 @@ suite(testing.suiteName(), () => {
       serviceName: 'test',
       folder: path.join(__dirname, 'schemas'),
     });
+    const port = await testing.getFreePort();
```
petemoore, line 58: Do we need synchronisation around here to avoid races? I don't understand the threading model well enough in node.js to be able to assess here if there is a race.

But my instinct says this must be a race, because it could even be a different process on the system that then takes the port. Shouldn't ports always be allocated atomically, i.e. you essentially request and bind to the port in a single call?

libraries/api/test/helper.js
```diff
@@ -29,14 +33,17 @@ teardown(() => {
 });
 
 /**
- * Set up a testing server on port 23525 serving the given API.
+ * Set up a testing server on an ephemeral port serving the given API.
  */
 export const setupServer = async ({ builder, context }) => {
+  const port = await testing.getFreePort();
+  helper.rootUrl = `http://127.0.0.1:${port}`;
```
petemoore, line 40: same here

petemoore (approved):
So it looks like an improvement to me, but still inherently racey. I don't think we should ever fetch an available port number and then later bind to it, I think it should always be an atomic request to bind to a free port, and then we get back the port number that was bound to. But perhaps that is a bigger rework of the code, I'm not sure.

In any case, presumably this will be _less_ intermittent than it was, but I suspect it does not fully solve the issue. Happy for others to pitch in, e.g. @Eijebong loves solving TOCTOU races! 🤓

libraries/api/test/context_test.js
matt-boris: No in-process race, node is single-threaded, and each call is a real kernel bind. You're right that another process could grab the port in the gap. The atomic fix (listen(0) + server.address()) doesn't fit well here, because the helpers need the rootUrl before the server starts. The window is tiny, and it's far better than the hardcoded ports, so I've documented it as a known race. If it ever flakes, retrying on EADDRINUSE would be an easy follow-up.

petemoore (approved):
Thanks Boris. This is certainly improvement, thanks for working on it! Indeed refactoring code to avoid the race entirely would be a big undertaking, so let's go with it like this.

PR #9069, opened 2026-08-31
fix(generic-worker): avoid nil store in `ExecutionErrors.add`

matt-boris:
The `e == nil` branch dereferenced a nil pointer. `append` already handles a nil slice.

### DIFF
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -1114,11 +1114,7 @@ func (e *ExecutionErrors) add(err *CommandExecutionError) {
 	if err == nil {
 		return
 	}
-	if e == nil {
-		*e = ExecutionErrors{err}
-	} else {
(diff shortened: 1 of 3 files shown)


workers/generic-worker/main.go
```diff
@@ -1114,11 +1114,7 @@ func (e *ExecutionErrors) add(err *CommandExecutionError) {
 	if err == nil {
 		return
 	}
-	if e == nil {
```
Eijebong, line 1117: Hilarious. I'm guessing this was supposed to be `if *e == nil`

Eijebong (approved):

workers/generic-worker/main.go
matt-boris: ... yep I'd imagine so lol

PR #9070, opened 2026-08-31
Return errors from `host.Run*` instead of swallowing them.

Eijebong:
This fixes three regressions from 712433340f which started returning nil there instead of errors for some reason. That means that any `host.Run*` might have been failing silently, leaving, for example, worker credentials unprotected when they were supposed to be chmod'ed at the worker start.

I added some simple tests to make sure this doesn't happen again


matt-boris (dismissed):
Ugh, ty. Thanks for the tests as well.

matt-boris (dismissed):

matt-boris (approved):

PR #9075, opened 2026-09-01
fix(generic-worker): drain remaining task completions on shutdown

matt-boris:
Shutdown waited on a wait group that only advanced in `RemoveTask`,
which never ran if the main loop had already returned. Shutdown
paths now receive completions until no tasks remain, so the worker
cannot hang and still records those tasks.

>Generic Worker no longer deadlocks on shutdown or when interrupted with Ctrl+C / `SIGINT` while a task is running. Shutdown waits on task completions until no tasks remain, instead of blocking on a wait group that only advanced when those completions were processed.

### DIFF
--- a/changelog/Tm_0Nl58SyKoOu6f2QhuZA.md
+++ b/changelog/Tm_0Nl58SyKoOu6f2QhuZA.md
@@ -1,0 +1,4 @@
+audience: worker-deployers
+level: patch
+---
+Generic Worker no longer deadlocks on shutdown or when interrupted with Ctrl+C / `SIGINT` while a task is running. Shutdown waits on task completions until no tasks remain, instead of blocking on a wait group that only advanced when those completions were processed.
--- a/workers/generic-worker/main.go
+++ b/workers/generic-worker/main.go
@@ -494,12 +494,9 @@ func RunWorker() (exitCode ExitCode) {
 		log.Printf("WARNING: failed to remove old task directories/users: %v", err)
 	}
 
-	// processCompletion records bookkeeping for a finished task and
-	// reports back what the main loop should do next. Extracted from
-	// inline drain logic so both the top-of-loop drain and the
-	// wait-window completion path can share a single implementation
-	// (avoiding the prior re-put-on-channel pattern, which relied on
-	// a brittle buffer-size / sender-count invariant).
+	// processCompletion applies a completion and reports what the main loop
+	// should do next. Bookkeeping is in recordCompletion so drainUntilIdle
+	// uses the same path.
 	type completionAction int
 	const (
 		completionContinue completionAction = iota
@@ -508,15 +505,23 @@ func RunWorker() (exitCode ExitCode) {
 		completionTasksComplete
 		completionRebootRequired
 	)
-	processCompletion := func(result taskCompletionResult) completionAction {
+	recordCompletion := func(result taskCompletionResult) {
 		taskManager.RemoveTask(result.taskID)
 		tasksResolved++
 		lastActive = time.Now()
+	}
+	drainUntilIdle := func() {
+		for !taskManager.IsIdle() {
+			recordCompletion(<-taskCompleteChan)
+		}
+	}
+	processCompletion := func(result taskCompletionResult) completionAction {
+		recordCompletion(result)
 
 		if result.workerShutdown {
 			log.Printf("Task %s requested worker shutdown, aborting other tasks...", result.taskID)
 			graceful.Terminate(false) // Abort other tasks immediately
-			taskManager.WaitForAll()
+			drainUntilIdle()
 			return completionWorkerShutdown
 		}
 
@@ -529,7 +534,7 @@ func RunWorker() (exitCode ExitCode) {
 		log.Printf("Resolved %v tasks in total so far%v.", tasksResolved, remainingTaskCountText)
 		if remainingTasks == 0 {
 			log.Printf("Completed all task(s) (number of tasks to run = %v)", config.NumberOfTasksToRun)
-			taskManager.WaitForAll()
+			drainUntilIdle()
 			if checkWhetherToTerminate() {
 				return completionWorkerManagerShutdown
 			}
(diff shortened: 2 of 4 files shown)


workers/generic-worker/task_manager.go
```diff
@@ -51,14 +51,19 @@ func (tm *TaskManager) AddTask(task *TaskRun) {
...
+// RemoveTask unregisters a task.
 // Updates the worker status file with remaining running task IDs.
 func (tm *TaskManager) RemoveTask(taskID string) {
 	tm.Lock()
 	defer tm.Unlock()
 	if _, exists := tm.runningTasks[taskID]; exists {
 		delete(tm.runningTasks, taskID)
-		tm.wg.Done()
 		tm.lastActive = time.Now()
 		tm.updateWorkerStatusLocked()
```
Eijebong, line 60: Don't we want to call this also?

AFAICT the only thing that this changes is the status file which only `g-w status` reads so it's not the end of the world, but temp files might survive reboots and that command might lie then.

Although that poses a problem because the `delete` above won't be called.

matt-boris: It's a good callout and something I was thinking about. I erred on the side of just not deadlocking vs having a bit of potential stale data. I can re-look into it if you think I should.

Eijebong: Well it's now code that's wrong so yeah, probably worth fixing?

matt-boris: Fixed in https://github.com/taskcluster/taskcluster/pull/9075/changes/7b3828cdbfa72269e2de3f6880384c5d2ae736fa.

workers/generic-worker/task_manager_test.go
```diff
@@ -117,11 +125,55 @@ func TestTaskManagerLastActive(t *testing.T) {
...
+
+	select {
+	case <-done:
+	case <-time.After(1 * time.Second):
+		t.Fatal("WaitForAll timed out")
+	}
+
+	require.True(t, tm.IsIdle())
+	_, err = os.Stat(workerStatusPath)
+	require.True(t, os.IsNotExist(err), "status file should be removed when no tasks remain")
```
Eijebong: Nit: `require.NoFileExists(t, workerStatusPath)`

https://pkg.go.dev/github.com/stretchr/testify/require#NoFileExists

workers/generic-worker/task_manager.go
```diff
@@ -87,9 +92,24 @@ func (tm *TaskManager) IsIdle() bool {
...
 }
 
-// WaitForAll blocks until all running tasks have completed.
+// WaitForAll blocks until all running tasks have completed, then unregisters
+// any tasks still in the map. Shutdown skips the main-loop completion drain,
+// so without this the worker status file would keep listing finished tasks
+// (and generic-worker status would report them as still running).
 func (tm *TaskManager) WaitForAll() {
 	tm.wg.Wait()
+	tm.clearRunningTasks()
```
Eijebong: I think this should be in a defer in `RunWorker`? Avoids the side effect in here and also covers the panic path which doesn't call this.

Wouldn't cover SIGKILL or someone pulling the plug but these you can't do anything about to begin with.

workers/generic-worker/main.go
```diff
@@ -743,6 +743,7 @@ mainLoop:
 							t.Error(fmt.Sprintf("Internal worker error (panic): %v", r))
 						}
 						portManager.ReleasePorts(t.TaskID)
+						taskManager.FinishTask()
 						workerShutdown := errors != nil && errors.WorkerShutdown()
 						taskCompleteChan <- taskCompletionResult{
```
Eijebong: I think this is still racy (`FinishTask` is called before sending to `taskCompleteChan`, meaning `wg.Wait()` can unblock before the message is even written to the channel).

Every single path that calls `WaitForAll` and returns out can now drop the last completion since it'll be written to the channel after the mainLoop returns out and none of those paths drain the channel. That's 2 cases, either the completion isn't even written or it's sitting in there.

I think a better fix would be to write to the channel before calling FinishTask and then drain the channel in the wait path and counting them in `tasksResolved` / `tm.RemoveTask`

matt-boris: I'm going to rework and entirely remove the waitgroup

matt-boris: Reworked in https://github.com/taskcluster/taskcluster/pull/9075/changes/2754c686c9eb3313ef732d969e90167d56f522a7. Before merge, I'll squash into the previous two commits. Left for now for ease of re-reviewing. Thank you for the reviews! 🙏🏻

workers/generic-worker/task_manager.go
```diff
@@ -87,9 +84,12 @@ func (tm *TaskManager) IsIdle() bool {
...
 }
 
-// WaitForAll blocks until all running tasks have completed.
-func (tm *TaskManager) WaitForAll() {
-	tm.wg.Wait()
+// WaitForAll applies each completion until no registered tasks remain.
+// apply must call RemoveTask, otherwise this never returns.
+func (tm *TaskManager) WaitForAll(completions <-chan taskCompletionResult, apply func(taskCompletionResult)) {
+	for !tm.IsIdle() {
+		apply(<-completions)
```
Eijebong: Hrmf. Not a fan of this. Could we move this into the `RunWorker` loop directly? Something like

```
drainUntilIdle := func() {
        for !taskManager.IsIdle() {
                recordCompletion(<-taskCompleteChan)
        }
}
```

Callback soup becomes very undebuggable real quick.


I think it means some tests have to go which is IMO not the end of the world, although there might be a way to still get this tested through some `RunWorker` level shutdown test? 🤔

Eijebong (commented):
I think this is a much nicer solution. Just one thing about the placement of the drain loop

workers/generic-worker/task_manager.go
matt-boris: Good point, addressed with https://github.com/taskcluster/taskcluster/pull/9075/changes/09a5dffe0b9bc8c2f21932ff45f9dd19f4b6daf2. It's nice having a fresh set of eyes on it to notice these things needing refactored.

Eijebong: That's what reviews are for :joy:

changelog/Tm_0Nl58SyKoOu6f2QhuZA.md
```diff
@@ -0,0 +1,4 @@
+audience: worker-deployers
+level: patch
+---
+Generic Worker no longer deadlocks on shutdown when `capacity` is greater than 1, or when interrupted with Ctrl+C / `SIGINT` while a task is running. Shutdown waits on task completions until no tasks remain, instead of blocking on a wait group that only advanced when those completions were processed.
```
Eijebong: Nit: This also happens at capacity == 1

Eijebong (dismissed):
LGTM now with a squash

Eijebong (approved):
Thank you

PR #9092, opened 2026-09-02
change worker manager worker pools and workers pages to use rest api

### LINKED ISSUE #9074: Migrate WMViewWorkerPools and WMViewWorkers pages from GraphQL to REST

thakkarujas5:
Fixes #9074


lotas (approved):
Thanks! Tested everything locally, thanks for adding ?tab to the urls

I also updated fetching of pending tasks to batch them to avoid firing hundreds of requests at once

PR #9097, opened 2026-09-02
fix(github): race condition in task group canceling

### LINKED ISSUE #9058: Github events concurrent processing issue
From matrix:
> Working with GitHub Stacks, I keep getting “EXCEPTION” errors on TC tasks when rebasing the stack (this is the [error comment](https://github.com/mozilla/bugbug/commit/8b120150153768245aa9e59d8f5193370cf0ea09#commitcomment-197970138)). Do you have a similar observation? 

Looks like currently it is not possible because of the errors like: https://github.com/mozilla/bugbug/commit/8b120150153768245aa9e59d8f5193370cf0ea09#commitcomment-197970138

```
Task group Csah2OJ3TcacgqLVLiu2Ig is sealed and does not accept new tasks.

    method: createTask
    errorCode: RequestConflict
    statusCode: 409
    time: 2026-08-27T17:28:46.886Z
```

Support didn't exist so far and the assumption was that single PR/commit results in single check suite or task group that got sealed. 
This is somehow not true for stacked prs.

Needs further investigation and implementation

lotas:
events from github might arrive at any pace, prs can be rebased, force pushed, etc very fast, and when that happens we might process them one by one in a way they will interfere.
Without checking build creation date we might accidentaly seal or cancel newer build that happened around the same time while the tasks are still being created. Which results in errors such as "task group is sealed"

It was also reported to a commit instead of PR because pull_request_number wasn't passed properly to the handler

Fixes #9058


services/github/src/handlers/index.js
```diff
@@ -348,7 +364,10 @@ class Handlers {
           build =>
             build.task_group_id !== newTaskGroupId &&
             build.event_id !== eventId &&
-            includedEventTypes.includes(build.event_type)
+            includedEventTypes.includes(build.event_type) &&
+            // slower older handler must never supersede a newer delivery
+            (build.created < newBuildCreated ||
+              (build.created.getTime() === newBuildCreated.getTime() && build.event_id < eventId))
```
lotas: `eventId` is UUID here coming from github.. and I hope they are holding temporal piece of info so we could determine who came first

services/github/src/handlers/job.js
```diff
@@ -69,7 +70,7 @@ async function createGithubBuildRecord({
     assert.equal(build.sha, sha);
     assert.equal(build.event_type, eventType);
     assert.equal(build.event_id, eventId);
-    return build;
+    return { ...build, pull_number: build.pull_request_number };
```
lotas, line 72: we need more types here.. js on its own cannot guess that `pull_request_number` is the `pull_number`

services/github/src/handlers/index.js
```diff
@@ -234,6 +234,16 @@ class Handlers {
...
       } catch (err) {
+        // as tasks are being created sequentially and it is likely something seals task group in between
+        // when this happens we don't want to fail the whole process, but just acknowledge and stop
+        if (
+          err.code === 'RequestConflict' &&
+          err.statusCode === 409 &&
+          err.message?.includes('is sealed and does not accept new tasks')
+        ) {
+          this.monitor.debug(`Task group ${t.task.taskGroupId} was sealed while tasks were being created; superseded`);
+          return;
```
Eijebong: This assumes that all tasks would be in the same group which might not be true. So it'd skip all tasks from a second group if the first one was sealed.
This goes against the work done in #8756

services/github/src/handlers/job.js
Eijebong: Unrelated but this is missing a pullNumber also. Worth a second tiny commit if you feel like fixing it

services/github/src/handlers/index.js
Eijebong: Yeah they're UUIDv1. However you can't string compare them reliably.

```
>>> a = uuid.UUID('ffb3b4b0-a6c5-11f1-bfff-00deadbeef01')
>>> b = uuid.UUID('004c4b30-a6c6-11f1-bfff-00deadbeef01')
>>> a.time
140076432753931440
>>> b.time
140076432763931440
>>> (b.time - a.time)
10000000
>>> a.time < b.time
True
>>> str(a) < str(b)
False
```

Those are 100ns ticks, given by github when the delivery is created. So decoding them and using them is probably better than our own `created`/`newBuildCreated` since those are records of when the current pod got the message from pulse, not when the event happened.

services/github/src/handlers/index.js
```diff
@@ -234,6 +234,16 @@ class Handlers {
       try {
         await limitedQueueClient.createTask(t.taskId, t.task);
       } catch (err) {
+        // as tasks are being created sequentially and it is likely something seals task group in between
+        // when this happens we don't want to fail the whole process, but just acknowledge and stop
+        if (
+          err.code === 'RequestConflict' &&
+          err.statusCode === 409 &&
```
Eijebong: Those two mean the same?

Eijebong (changes requested):

services/github/src/handlers/index.js
lotas: true.. so we probably need to keep track of which taskGroupIds were already sealed and only skip those.. 
but on the other hand if one of the taskGroups from the same github event was already sealed, it could mean the whole build is being cancelled

services/github/src/handlers/index.js
lotas: technically yes.. I can drop one

services/github/src/handlers/index.js
lotas: good catch, I can add those to tests

services/github/src/handlers/index.js
```diff
@@ -230,10 +234,26 @@ class Handlers {
...
       try {
         await limitedQueueClient.createTask(t.taskId, t.task);
       } catch (err) {
+        // as tasks are being created sequentially and it is likely something seals task group in between
+        // when this happens we don't want to fail the whole process, but just acknowledge and stop processing
+        // affected task group
+        if (err.statusCode === 409 && err.message?.includes('is sealed and does not accept new tasks')) {
+          sealedTaskGroupIds.add(taskGroupId);
+          this.monitor.debug(`Task group ${taskGroupId} was sealed while tasks were being created; superseded`);
+          continue;
```
Eijebong, line 255: I think that skipping them entirely here makes it so that `job.js:513` then publishes `taskGroupCreationRequested` for the sealed task group despite it being meaningless by then. That ends up creating statuses for the SHA which stay pending forever. Worse, if the same SHA arrives twice here (force push I'm guessing could do that) then they can race eachother and we could end up with a permanent `pending` status.

Eijebong: Not sure how much we should care here given it requires the same SHA to be a problem and github statuses have been deprecated for 20 years

lotas: well that's true.. I think that's an easy fix, createTasks could return which groups were actually created ..
although, if it would be hit in the middle of creation, should we treat this half-finished group as created or not. Probably if there's at least one task created even should be fired

Eijebong: I think it ended up being sealed it should be ignored entirely? The main problem I have with this approach is if the group got sealed for another reason than cancelling but given the timing that should not really happen.

lotas: agreed. added fix to emit events for the task groups that didn't encounter 409s

services/github/src/handlers/index.js
```diff
@@ -230,10 +234,26 @@ class Handlers {
...
+        continue;
+      }
+
       try {
         await limitedQueueClient.createTask(t.taskId, t.task);
       } catch (err) {
+        // as tasks are being created sequentially and it is likely something seals task group in between
+        // when this happens we don't want to fail the whole process, but just acknowledge and stop processing
+        // affected task group
+        if (err.statusCode === 409 && err.message?.includes('is sealed and does not accept new tasks')) {
```
Eijebong, line 272: Hrmf. I'm not sure how I like depending on some english in an error message to detect this. Once again javascript showing how painful it is to do anything. Can we at least make it so that some test would fail if this was to diverge from the queue service?

services/github/src/handlers/index.js
```diff
@@ -348,7 +375,9 @@ class Handlers {
           build =>
             build.task_group_id !== newTaskGroupId &&
             build.event_id !== eventId &&
```
Eijebong: Nit: This is redundant with `isOlderDelivery(build.event_id, eventId)`

services/github/src/handlers/index.js
```diff
@@ -13,6 +13,10 @@ import { jobHandler } from './job.js';
 import { rerunHandler } from './rerun.js';
 import { POLICIES } from './policies.js';
 import { GITHUB_BUILD_STATES } from '../constants.js';
+import { v1ToV6 } from 'uuid';
+
+// github uses UUID1 and to compare them as ordered strings we convert it to V6 which is sortable
+const isOlderDelivery = (candidateId, currentId) => v1ToV6(candidateId).localeCompare(v1ToV6(currentId)) < 0;
```
Eijebong: Nit: you can just use `<` here, it's hex

Eijebong (changes requested):
I think this is still racy in 2 places.

1) If you get 3 deliveries back to back, imagine D1 < D2 < D3, received as D3 -> D1 -> D2:

- D3 runs, finds nothing pending, continues merrily on its way
- D1 runs, finds nothing older, continues (oof)
- D2 runs, finds D1, cancels it, continues (double oof)

It's better than before where D1 would've killed D3, but it means the fix isn't complete. I think deliveries should self cancel when they find a newer one already there?

2) I think that you could try sealing a task group between the moment the DB row is written to the github service at https://github.com/taskcluster/taskcluster/blob/38eb9fc135f09dffcc880d1e178ce20dc7018849/services/github/src/handlers/job.js#L32 and the moment the queue service creates the group whenever it feels like it.

services/github/src/handlers/index.js
lotas: yes.. we could technically use different status code, but adding dedicated test in queue service is "easier". Will add 👍

lotas: pushing fix, thanks

services/github/src/handlers/index.js
lotas: refactored the whole block to make it more obvious what is about to get sealed

services/github/src/handlers/index.js
lotas: done, thanks

lotas:
> I think this is still racy in 2 places.
> 
>     1. If you get 3 deliveries back to back, imagine D1 < D2 < D3, received as D3 -> D1 -> D2:
> 
> 
>     * D3 runs, finds nothing pending, continues merrily on its way
> 
>     * D1 runs, finds nothing older, continues (oof)
> 
>     * D2 runs, finds D1, cancels it, continues (double oof)
> 
> 
> It's better than before where D1 would've killed D3, but it means the fix isn't complete. I think deliveries should self cancel when they find a newer one already there?
> 
>     2. I think that you could try sealing a task group between the moment the DB row is written to the github service at https://github.com/taskcluster/taskcluster/blob/38eb9fc135f09dffcc880d1e178ce20dc7018849/services/github/src/handlers/job.js#L32
>         and the moment the queue service creates the group whenever it feels like it.

yes, indeed, it was possible. 

What I'm doing now:
1. checking if newer builds exist (compared to current event)
2a. if newer exist it means we are redundant -> cancel itself
2b. if older builds exist (based on UUID ordering) we cancel those and letting current roll

wdyt

services/github/src/handlers/job.js
Eijebong: I think we also need to exclude these task groups getting cancelled by this from the loop below or we publish `taskGroupCreationRequested` for a group we just cancelled.

And the comment above is lying now. The function is literally doing the opposite when `newerDeliveryExists`

```
      const buildsToCancel = newerDeliveryExists
        ? relevantBuilds.filter(build => build.event_id === eventId)
```

I'd rename that method to `cancelSupersededTaskGroups` since "previous" here is lying as we could cancel ourselves.

Eijebong: I'm fairly certain that this needs to be pulled out of the `Promise.all`. When `newerDeliveryExists`, this cancels every task group of the current delivery, including sibling hooks that haven't reached `triggerHook` yet. Their `createTask` then hits a sealed task group and fails, posting a comment in the PR about it.

Eijebong (changes requested):

services/github/src/handlers/job.js
lotas: indeed. that's what happens after few rounds of reviews.. probably would be better to start from scratch :D 
Will amend

lotas: agree. will move

Eijebong (approved):
Thank you for dealing with this! Looks good now. Bonus point if you squash the history

lotas:
let me quickly merge this before you change your mind :D 
Thanks for the reviews!!

PR #9098, opened 2026-09-02
Generic Worker does not panic on bad artifact name

### LINKED ISSUE #9007: Generic worker panics when it receives a 4xx from `createArtifact` instead of resolving as `malformed-payload`
```yaml
taskQueueId: docker-compose/generic-worker
schedulerId: taskcluster-ui
created: '2026-08-10T12:50:03.453Z'
deadline: '2026-08-10T15:50:03.453Z'
payload:
  command:
    - - /bin/bash
      - '-c'
      - echo "hi" > hello.log;
  artifacts:
    - name: |-
        public/logs/a
        b.log
      path: hello.log
      type: file
  maxRunTime: 300
metadata:
  name: example-task
  owner: name@example.com
  source: http://taskcluster/tasks/create
```

Results in a worker panic:

```
panic: WORKER EXCEPTION due to response code 400 from Queue when uploading artifact &artifacts.S3Artifact{BaseArtifact:(*artifacts.BaseArtifact)(0x3ea9226214a0), Path:"/home/task_NUOauf9IThOD9_0/hello.log", ContentPath:"/tmp/hello.log1102881728", ContentEncoding:"gzip", ContentType:"text/plain", ContentLength:3} with CreateArtifact payload {"contentLength":3,"contentType":"text/plain","expires":"2027-08-10T15:48:42.877Z","storageType":"s3"} - HTTP response body: {
	  "code": "InvalidRequestArguments",
	  "message": "Invalid URL patterns:\nURL parameter 'name' given as 'public/logs/a\nb.log' must match regular expression: '/^[\\x20-\\x7e]+$/'\n\n---\n\n* method:     createArtifact\n* errorCode:  InvalidRequestArguments\n* statusCode: 400\n* time:       2026-08-10T12:48:54.078Z",
	  "requestInfo": {
	    "method": "createArtifact",
	    "params": {
	      "taskId": "NUOauf9IThOD9NUsoqm7Ng",
	      "runId": "0",
	      "name": "public/logs/a\nb.log"
	    },
	    "payload": {
	      "contentLength": 3,
	      "contentType": "text/plain",
	      "expires": "2027-08-10T15:48:42.877Z",
	      "storageType": "s3"
	    },
	    "time": "2026-08-10T12:48:54.078Z"
	  }
	}

goroutine 80 [running]:
main.(*TaskRun).classifyCreateArtifactError(0xb37a60?, {0xd0cf20, 0x3ea9226f09b0}, {0x3ea92274a070, 0x66, 0x3ea9228912f0?}, {0xcffcc0?, 0x3ea92266e648})
	/app/workers/generic-worker/artifacts.go:155 +0x7c5
main.(*TaskRun).uploadArtifact(0x3ea922501008, {0xd0cf20, 0x3ea9226f09b0})
	/app/workers/generic-worker/artifacts.go:104 +0x29d
main.(*ArtifactTaskFeature).Stop.func1()
	/app/workers/generic-worker/artifact_feature.go:113 +0x8d
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/go/pkg/mod/golang.org/x/sync@v0.22.0/errgroup/errgroup.go:93 +0x50
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 85
	/go/pkg/mod/golang.org/x/sync@v0.22.0/errgroup/errgroup.go:78 +0x95
```

And that panic isn't recovered.

It should probably resolve the task as `malformed-payload` instead. #6577 is the other half of the same problem where the queue refuses an artifact name that would probably work fine.

petemoore:
Fixes #9007


workers/generic-worker/artifacts.go
```diff
@@ -157,7 +151,15 @@ func (task *TaskRun) classifyCreateArtifactError(artifact artifacts.TaskArtifact
 			if status == deadlineExceeded || status == cancelled {
 				return nil
 			}
-			// assume a problem with the request == worker bug
+			// 4xx status code implies Queue rejected this artifact. The
+			// Queue's requirements might be stricter than Generic Worker's
+			// payload schema. For example, allowed characters in artifact name
+			// get tightened on Queue side but Generic Worker is not updated
+			// everywhere so Generic Worker payload schema might be looser.
+			if rootCause.HttpResponseCode/100 == 4 {
```
Eijebong: Thinking again about this. We probably don't want to catch 401/403 in there since that'd mean something has gone horribly wrong with scopes?

tools/d2g/schemas/docker-worker/v1/payload.yml
```diff
@@ -221,6 +221,8 @@ properties:
       Moreover, in the case of time-critical spot terminations, tasks have
       more chance of successfully publishing volume artifacts than directory
       artifacts, due to the efficiency gain.
+    propertyNames:
+      pattern: '^[\x20-\x7e]+$'
```
Eijebong, line 226: This is not valid for a task input. It needs to be `^[\x20-\x7e]*$`. The worker expects `""` to be a valid artifact name, at which point it generates one from the path of the artifact. See https://github.com/taskcluster/taskcluster/blob/2314588d85d27b3d3a91f4678e8d611a053c4d77/workers/generic-worker/artifact_feature.go#L171-L173

internal/mocktc/queue.go
```diff
@@ -105,10 +106,25 @@ func (queue *Queue) ensureArtifactMap(taskId, runId string) {
...
+		return nil, &tcclient.APICallException{
+			CallSummary: &tcclient.CallSummary{
+				HTTPResponseBody: fmt.Sprintf("Invalid URL patterns:\nURL parameter 'name' given as '%v' must match regular expression: '/^[\\x20-\\x7e]+$/'", name),
+			},
+			RootCause: httpbackoff.BadHttpResponseCode{
+				HttpResponseCode: 400,
+			},
+		}
+	}
+
```
Eijebong, line 132: This never gets hit. All the new tests are either getting the task through the worker, which rejects it due to the schema change, or handbuild API responses.

I don't think it's wrong per-se since it would catch a regression in the schema but it's also showing we're not testing something we probably should

Eijebong (changes requested):
Probably also want to do the same with `logs.live` and `logs.backing`

tools/d2g/schemas/docker-worker/v1/payload.yml
petemoore: Good spot, i had forgotten about that!

workers/generic-worker/artifacts.go
petemoore: Fall back to panic? Although I'm not sure panic helps us much, because if there is a bug, all workers will have the same bug.... What do you think?

internal/mocktc/queue.go
petemoore: My original approach was just to defer to the Queue to decide if it should be rejected (I don't particularly like duplicating the regexp across Queue and worker since in any case they can easily get out of sync, especially as workers can have a different update cadence to services), and then later I decided to patch the schema too just to avoid the latency of waiting for a task to complete before discovering there was a bad artifact name. I agree it is not bad per-se, other than that we have mocks in go that essentially implement the services in Go, rather than just using the services themselves - but that is an old problem. I mean, we are in the same repo, I feel like if we are going to mock the services, just stand up the actual services in the task. I can't think of a good way to catch this though. Although, let's say the worker regexp stopped working, and the test that has a line return in the artifact name would hit this, and get a 400 back. I'm not sure we need to test something explicitly here. What test would you add?

tools/d2g/schemas/docker-worker/v1/payload.yml
petemoore: Also, we should probably have a test for that! I'll add one. Tests should have found that.

internal/mocktc/queue.go
Eijebong: I don't have any idea how to test it. But if we're at "test the mock" then I'd just remove it.

workers/generic-worker/artifacts.go
Eijebong: I think a worker panicking because something went horribly wrong is fine? That can't happen on user error, it'd be some mismatch with auth/scopes somehow that'd return 401/403 and if those happen then we'd like to know rather than show a confusing malformed payload to a user when they can't do anything about it?

tools/d2g/schemas/docker-worker/v1/payload.yml
petemoore: After thinking about this some more, I think the following artifact should be considered valid, with name being treated as matching path:

```yml
payload:
  artifacts:
    - path: public/my/path
      type: file
```

whereas this should resolve as `exception/malformed-payload`, since name is explicitly specified as an invalid value (rather than just being omitted):

```yml
payload:
  artifacts:
    - name: ''
      path: public/my/path
      type: file
```

Do you agree? I think this would be cleanest/most correct. We could make the error message explicit that if the intention is for name to match the path, either do not include name in the task payload at all, or set it equal to path. However, go probably won't make it easy for us to differentiate the two cases (empty string vs omitted entirely) so maybe it isn't worth the effort.

Eijebong: That'd be a breaking change and while I agree that someone passing `name: ''` probably did so by mistake there might be some tasks out there depending on that these days. Since it doesn't seem trivial either I'd say not worth the effort?

internal/mocktc/queue.go
petemoore: So it turns out we can test it by not specifying an artifact name, and only a path, and then have the path have a non-printable ascii character. So I test this now explicitly with an artifact without a `name` and with `path` as `unzulässiges-Zeichen.txt`. Clever, hey?! :-D

tools/d2g/schemas/docker-worker/v1/payload.yml
petemoore: agree

workers/generic-worker/artifacts.go
```diff
@@ -157,7 +157,18 @@ func (task *TaskRun) classifyCreateArtifactError(artifact artifacts.TaskArtifact
...
-			// assume a problem with the request == worker bug
+			// 4xx status code implies Queue rejected this artifact. The
+			// Queue's requirements might be stricter than Generic Worker's
+			// payload schema. For example, allowed characters in artifact name
+			// get tightened on Queue side but Generic Worker is not updated
+			// everywhere so Generic Worker payload schema might be looser.
+			// 401/403 are excluded here: they signal a worker/credentials
+			// problem rather than something the task submitter can fix, so
+			// they fall through to the panic below instead.
+			if rootCause.HttpResponseCode/100 == 4 && rootCause.HttpResponseCode != 401 && rootCause.HttpResponseCode != 403 {
```
Eijebong, line 163: I don't think 429 should land in there. (408 maybe neither?)

I don't think the queue can emit a 429 but we probably want to special case them to fail the task as intermittent instead, which is better than panicking or blaming the payload because the network/clownflare/fastly decided to not.

workers/generic-worker/artifacts_test.go
```diff
@@ -789,6 +790,247 @@ func TestOptionalArtifactUploadFailureFailsTask(t *testing.T) {
 	_ = submitAndAssert(t, td, payload, "exception", "resource-unavailable")
 }
 
+// TestInvalidArtifactNameFailsAsMalformedPayload verifies that an artifact
+// name which doesn't match the pattern the Queue enforces is now caught by
+// generic-worker's own payload schema validation, failing the task
+// immediately as malformed-payload rather than running to completion.
+// See https://github.com/taskcluster/taskcluster/issues/9007
+func TestInvalidArtifactNameFailsAsMalformedPayload(t *testing.T) {
```
Eijebong, line 801: AFAICT, this doesn't test what the comment says (you can remove the schema changes and it still passes). Also "now caught" is a LLM-ism and that comment makes no sense outside of that diff

workers/generic-worker/artifacts_test.go
```diff
@@ -789,6 +790,247 @@ func TestOptionalArtifactUploadFailureFailsTask(t *testing.T) {
...
+		t.Fatalf("Error reading source file: %v", err)
+	}
+	actualData := getArtifactContent(t, taskID, "SampleArtifacts/_/X.txt")
+	if string(expectedData) != string(actualData) {
+		t.Fatalf("Artifact content mismatch: expected %d bytes, got %d bytes", len(expectedData), len(actualData))
+	}
+}
+
+// TestInvalidLiveLogNameFailsAsMalformedPayload verifies that logs.live is
+// now subject to the same printable-ASCII pattern as artifact names, so a
```
Eijebong, line 988: Same LLM-ism here

workers/generic-worker/artifacts_test.go
```diff
@@ -789,6 +790,247 @@ func TestOptionalArtifactUploadFailureFailsTask(t *testing.T) {
...
+	defaults.SetDefaults(&payload)
+
+	td := testTask(t)
+	_ = submitAndAssert(t, td, payload, "exception", "malformed-payload")
+}
+
+// TestClassifyCreateArtifactError4xxIsMalformedPayload unit tests
+// classifyCreateArtifactError directly, to verify that a 4xx response from
+// the Queue's CreateArtifact endpoint is classified as malformed-payload
+// rather than causing an unrecovered panic.
```
Eijebong, line 824: And one more LLM-ism here. The previous code is not visible outside of this diff so this makes no sense to a future reader

workers/generic-worker/artifacts_test.go
```diff
@@ -789,6 +790,247 @@ func TestOptionalArtifactUploadFailureFailsTask(t *testing.T) {
 	_ = submitAndAssert(t, td, payload, "exception", "resource-unavailable")
 }
 
+// TestInvalidArtifactNameFailsAsMalformedPayload verifies that an artifact
+// name which doesn't match the pattern the Queue enforces is now caught by
+// generic-worker's own payload schema validation, failing the task
+// immediately as malformed-payload rather than running to completion.
+// See https://github.com/taskcluster/taskcluster/issues/9007
```
Eijebong, line 800: I really don't think we need that much repetition of the issue number. The tests are very explicit at what they're doing and if anyone wonders why they can blame and read the blame.

I'd remove those comments that are just repeating what the code is doing (the test name is enough) and add one on `TestEmptyArtifactNameDerivesFromPath` because it's not evident that the map roundtrip exists because `Name` is `omitempty`.

Eijebong (changes requested):

workers/generic-worker/artifacts.go
Eijebong: Discussed a bit out of band. The queue can't emit those and mapping them to something that wouldn't make things worse by hammering the queue in the case of a 429 isn't trivial. Let's keep it this way, it's not like we're locking ourselves into this, it's fairly easy to change the behavior.

Eijebong (approved):
Thanks!

PR #9099, opened 2026-09-02
Drop dotenv-cli as a dev dependency

Eijebong:
It stopped being used by the UI build in 2019
(c7011aa95b2c955514c3b0881830c1c8e6e6838c) and the only other user disappeared in 2021 (3fc8fbeedf63ed1333db8aee226aeb2eeed34a33).

The script reads a single variable which isn't a secret (`GRAPHQL_ENDPOINT`) and the documentation mentions to use an inline env variable anyway which defeats the point.

Closes #9084


lotas (approved):
yay! minus few deps

PR #9118, opened 2026-09-02
fix: change profile, quickstart, third party login, tc debug yaml pages from GraphQL to REST

### LINKED ISSUE #9117: Migrate Profile, Quickstart, TcYamlDebug, ThirdPartyLogin pages from GraphQL to REST

thakkarujas5:
Fixes #9117


services/web-server/src/loaders/github.js
```diff
@@ -1,32 +0,0 @@
-import DataLoader from 'dataloader';
-
-export default ({ github }, _isAuthed, _rootUrl, _monitor, _strategies, _req, _cfg, _requestId) => {
```
lotas, line 3: I think this is the last place that used `clients.github` so it's no longer needed in https://github.com/taskcluster/taskcluster/blob/7dfdf156fca5f4c3e1264aecc00c3c985475653d/services/web-server/src/clients.js#L6 

but I guess we can drop that file as well after we remove the last graphql loader

lotas (approved):
Tested locally, refactoring looks to be correct! thanks

PR #9124, opened 2026-09-04
Local dev env fixes

lotas:
* bump `tc-admin` to latest `5.3.1`
* apply CORS headers always to avoid confusing network related errors with actual service responses (masking bad auth for example)


Eijebong:
Weird. b4ecec24a9a256f8a41f42c234fdce1794f6c582 is yours and should've fixed issues you'd get locally? What exactly were you seeing?

lotas:
> Weird. [b4ecec2](https://github.com/taskcluster/taskcluster/commit/b4ecec24a9a256f8a41f42c234fdce1794f6c582) is yours and should've fixed issues you'd get locally? What exactly were you seeing?

```
✘ Contain... Error response from daemon: image with reference taskcluster/tc-admin:5.2.0 was found but its platform (linux/arm64) does not match the specified platform (linux/amd64) 
```

ah sorry. might be my overrides or env vars. but regardless bumping it to 5.3.1

https://hub.docker.com/r/taskcluster/tc-admin/tags  we do indeed ship for both architectures since 5.2.0

Eijebong:
Mind fixing the commit message for the tc-admin bump then? I'm assuming you had a local 5.2.0 that wasn't multiarch. Bumping it is very fine but let's not have lying commit messages and make future us groan at it next time we have some issue with the tc-admin image

Eijebong (approved):
Thx

Issue #9065, opened 2026-08-31
ViewWorkerTypes page need a endpoint for returning pending tasks count


thakkarujas5:
Currently the ViewWorkerTypes page does a N+1 query to fetching the pending tasks.

We need a new endpoint which fetches all the pending tasks count from all worker type in bulk at one shot.

queue.taskQueueCounts in the queue endpoint which fetches data for one worker type at a time, we need a bulk alternate endpoint for this.

lotas:
This is also needed for the dashboard as well, since we are fetching for each pool individually, and that would send a lot of requests

Main issue is figuring out scopes. If anonymous includes pending/claimed that's easy, if not, we need to check which pools are accessible to the client and only return those

PR #9128, opened 2026-09-07
change dashboard page to use rest api's

### LINKED ISSUE #9127: Migrate Dashboard page from GraphQL to REST

thakkarujas5:
Fixes #9127


ui/src/components/StatusDashboard/StatsFetcher.jsx
```diff
@@ -1,44 +1,180 @@
...
-import rolesQuery from './roles.graphql';
-import secretsQuery from './secrets.graphql';
-import wmPoolsQuery from './wmPools.graphql';
-import wmPoolsErrorsQuery from './wmPoolsErrors.graphql';
+import useResource from '../../utils/useResource';
+import fetchAllPages from '../../utils/fetchAllPages';
+import { TaskclusterClientContext } from '../../utils/TaskclusterClient';
+import { DASHBOARD_STATS_PAGE_SIZE } from '../../utils/constants';
+
+const REFRESH_INTERVAL = 30 * 1000;
```
lotas, line 15: that's going to hurt. I think we should add dedicated w-m or queue service endpoint to return summaries. As it stands, it will send ~400 requests every 30s on community and probably > 600 on fxci

ui/src/components/StatusDashboard/StatsFetcher.jsx
```diff
@@ -1,44 +1,180 @@
...
+        fetchAllPages(
+          options => workerManager.listWorkerPoolsStats(options),
+          response => response.workerPoolsStats,
+          pageOptions
+        ),
+      ]);
+      const statsByPool = new Map(
+        stats.map(poolStats => [poolStats.workerPoolId, poolStats])
+      );
+      const pendingTasks = await fetchPendingTasks(
```
lotas, line 79: maybe the easiest way is to actually send `pendingCount` again with the `listWorkerPoolStats` 🤔

lotas: but then again, it's not concern of the w-m.. so new queue endpoint with summaries is preferred, but figuring out which pools are accessible for the current actor is going to be fun

lotas (changes requested):
Blocked by  https://github.com/taskcluster/taskcluster/issues/9065
upd: https://github.com/taskcluster/taskcluster/pull/9129 
upd: no longer blocked

ui/src/components/StatusDashboard/StatsFetcher.jsx
lotas: ok, since worker manager page is doing the same, we'll keep it as is, and I will send a follow up PR to do a single counts call instead

lotas (approved):
Thanks!
I will accept and merge it as is, and for the both worker-manager, dashboard pages will send a follow up PR to use single counts endpoint to avoid doing 1 per worker pool to optimize.

PR #9129, opened 2026-09-07
feat(queue): Introduce `taskQueueCountsBatch` endpoint for aggregations

### LINKED ISSUE #9065: ViewWorkerTypes page need a endpoint for returning pending tasks count

lotas:
UI is requesting pending and claimed counts in several places and so far it was only possible by doing individual requests. This follows same semantics as recently introduced batched tasks and tasks statuses. Caller needs to pass the exact list of task queues to be returned, and should have both `queue:pending-count:<taskQueueId>` and `queue:claimed-count:<taskQueueId>` scopes for every requested task queue.

Fixes #9065


matt-boris (approved):
LGTM, thanks.

PR #9130, opened 2026-09-07
Bug 2069456 - Use safefs when reading/writing during command preparation on windows workers

Eijebong:
This is fairly trivial since safefs already had the required primitives


petemoore (approved):
Looks good, thanks Bastien!

PR #9132, opened 2026-09-08
Bug 2069332 - Don't take artifact source path from the task user's stdout

### LINKED ISSUE #6677: [Generic Worker] Windows syscalls log to stdout
> Let's put here that it appears that the syscall logs are sent to stdout, even though the code appears to send to standard error, and that it is an open issue why this is the case (we could leave a TODO, or even create an issue to keep track of this - it could bite us at some point).

_Originally posted by @petemoore in https://github.com/taskcluster/taskcluster/pull/6673#discussion_r1388173119_

Eijebong:
When copying an artifact as a user on windows through `g-w copy-to-temp-file`, the worker asked for that copy without giving a destination and instead read that back from the *last line* of the user controlled process stdout. That stdout is a pipe and a stray process could write to it since it's entirely task owned, appending a single line and get the worker to then copy an entirely different file.

Posix platforms are just as redirectable for two reasons. There `\n` is a valid character in a filename, and the stray process could just read stdout as well and steal the actual filename and then write its own.

Instead of doing that which is very fragile to begin with, decide on the file path in the privileged process and then go cat the file as the user through `g-w cat-file`. This lead to me having to refactor quite a bit of the artifacts upload because cleanup became quite messy since it was unclear who owned that temporary file.

I had tried adding a `--dest` flag to the `copy-to-temp-file` command at first since it seemed like an easier solution, unfortunately windows had something to say about this because temp directories are scoped per user, so I couldn't easily get the worker to create a temp directory that the task user could write to. So I tried putting it in a reserved directory in the task folder but then we had to ignore that directory on upload in case the task would try to upload the entire task home as artifact... Anyway, it was painful, had many edge cases that I really didn't like. So I came up with this, which doesn't have any of those issues. Better, the task never even sees the actually uploaded file or file structure, so unlike the previous implementation that relied entirely on safefs to not follow planted junctions/links, now we also have default windows ACLs providing some security after the copy.

If you're attentive you'll notice that this removes a very scary workaround, `copy-to-temp-file` used to print some unrelated things on stdout which would corrupt the resulting path. It was never found out why this was happening so I had to dig up some history. Before ed5e9c1f48ec2a4469ed3e3fb10dc7208c12f7b1, which fixed this exact pollution on `--version`, logs from the worker were output to stdout instead of stderr. After that commit, logs go to stderr which fixes the corruption and thus doesn't impact `cat-file`. I still added a test for binary files artifacts and check that there's no stray data in there. I was both concerned about that workaround before digging and about windows messing linebreaks. Both concerns are now properly tested.

There's a few bonus points in this refactor, first of all, when gzipping an artifact, we avoid a full copy. Before this, the file was copied in a temp file, then gzipped in a second one. Since we're streaming the content, we can just pass it directly through gzip. That's one less copy.

Secondly, the code used to copy all files to tmp and then upload them (which is the reason why cleanup was so complex). It'll now copy and then upload in one go, which means the maximum disk space required for tmp isn't the entirety of all artifacts anymore but just space for 10 at a time. This reduces peak usage and should also start uploads earlier which should speed things up quite a bit since copy/upload will now happen in parallel.

Third, artifacts temporary copies are now stored in the worker's temporary directory instead of the user's. Which means the user can't fiddle with them anymore. I don't think this saves us from anything since the content is owned by the user already, but I feel better about internal copies not being in a user owned directory.

Now one change in behavior that I chose to take rather than bend the code in weird ways (it's bent enough as it is...). When uploading an optional artifact, it'll now fail the task if the artifact fails to upload because of a permission issue. That goes against the original implementation of optional artifacts in
5062e7d82b28be3078356394f5787b859c33bf88.
However, neither the commit message nor the documentation state that the intent of resolving tasks failing to upload artifacts due to permission issues was on purpose. I'm assuming it was made to match docker-worker 1-1, but I don't think that this case can happen in normal tasks without someone trying to mess with reserved artifacts or trying to upload files that can't get uploaded anyway, in which case we'd just hide the bug (for a d2g task that generates a 0600 root owned file declared as an artifact, the task will now fail rather than pass and ignore the file). IMHO this is an improvement on behavior, dropping artifacts that do exist silently hides task bugs. It also follows the direction we've taken with 7044b6a9726854f865a1a70683e5bb75be635ca2.

Closes #6677 on the way.

### DIFF
--- a/changelog/bug-2069332-1.md
+++ b/changelog/bug-2069332-1.md
@@ -1,0 +1,7 @@
+audience: users
+level: major
+reference: bug 2069332
+---
+When uploading an optional artifact, if it's not readable (or encounters any
+unreadable file for directory artifacts), the task will now fail instead of
+silently omitting that file.
--- a/changelog/bug-2069332.md
+++ b/changelog/bug-2069332.md
@@ -1,0 +1,10 @@
+audience: worker-deployers
+level: major
+reference: bug 2069456
+---
+When uploading artifacts, Generic Worker multiuser engine will now create
+temporary files as the worker user (`root`/`LocalSystem`) and stream content
+into them as the task user.
+
+The internal `generic-worker copy-to-temp-file` command has been replaced with
+`generic-worker cat-file`
--- a/ui/docs/reference/workers/generic-worker/usage.mdx
+++ b/ui/docs/reference/workers/generic-worker/usage.mdx
@@ -20,7 +20,7 @@ and reports back results to the queue.
                                             [--worker-runner-protocol-pipe PIPE]
     generic-worker show-payload-schema
     generic-worker new-ed25519-keypair      --file ED25519-PRIVATE-KEY-FILE
-    generic-worker copy-to-temp-file        --copy-file COPY-FILE
+    generic-worker cat-file                 --cat-file CAT-FILE
     generic-worker create-file              --create-file CREATE-FILE
     generic-worker create-dir               --create-dir CREATE-DIR
     generic-worker unarchive                --archive-src ARCHIVE-SRC --archive-dst ARCHIVE-DST --archive-fmt ARCHIVE-FMT
@@ -43,9 +43,8 @@ and reports back results to the queue.
                                             compliant private/public key pair. The public
                                             key will be written to stdout and the private
                                             key will be written to the specified file.
-    copy-to-temp-file                       This will copy the specified file to a temporary
-                                            location and will return the temporary file path
-                                            to stdout. Intended for internal use.
+    cat-file                                This will write the contents of the specified
+                                            file to stdout. Intended for internal use.
     create-file                             This will create a file at the specified path.
                                             Intended for internal use.
     create-dir                              This will create a directory (including missing
@@ -74,7 +73,7 @@ and reports back results to the queue.
                                             to. The parent directory must already exist.
                                             If the file exists it will be overwritten,
                                             otherwise it will be created.
-    --copy-file COPY-FILE                   The path to the file to copy.
+    --cat-file CAT-FILE                     The path to the file to write to stdout.
     --create-file CREATE-FILE               The path to the file to create.
     --create-dir CREATE-DIR                 The path to the directory to create.
     --archive-src ARCHIVE-SRC               The path to the archive file to unarchive.
@@ -397,7 +396,7 @@ and reports back results to the queue.
            spot termination notice, and therefore has shut down.
     73     The config provided to the worker is invalid.
     75     Not able to create an ed25519 key pair.
-    76     Not able to copy --copy-file to a temporary file.
+    76     Not able to write --cat-file to stdout.
     77     Not able to apply required file access permissions to the generic-worker config
            file so that task users can't read from or write to it.
     78     Not able to connect to --worker-runner-protocol-pipe.
--- a/workers/generic-worker/artifact_feature.go
+++ b/workers/generic-worker/artifact_feature.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"mime"
 	"os"
@@ -111,12 +112,30 @@ func (atf *ArtifactTaskFeature) Stop(err *ExecutionErrors) {
 				task.Warnf("Not uploading artifact %v found in task.payload.artifacts section, since this will be uploaded later by %v", taskArtifact.Base().Name, feature)
 				return nil
 			}
+
+			// Read the content *before* telling the queue the artifact
+			// exists, so that a file the task user cannot read is published
+			// as an error artifact rather than as a failed upload.
+			defer taskArtifact.DiscardContent()
+			if e := taskArtifact.PrepareContent(); e != nil {
+				if _, ok := errors.AsType[artifacts.ContentError](e); !ok {
+					uploadErrChan <- executionError(internalError, errored, fmt.Errorf("could not copy artifact %v to temp directory: %w", taskArtifact.Base().Name, e))
+					return nil
+				}
+				taskArtifact = &artifacts.ErrorArtifact{
+					BaseArtifact: taskArtifact.Base(),
+					Message:      fmt.Sprintf("Couldn't read file '%s': %v", taskArtifact.SourcePath(), e),
+					Reason:       "file-not-readable-on-worker",
+					Path:         taskArtifact.SourcePath(),
+				}
+			}
+
 			e := task.uploadArtifact(taskArtifact)
 			errArtifact, isErrArtifact := taskArtifact.(*artifacts.ErrorArtifact)
 			// An artifact we never got readable content for is published as an
 			// error artifact. Optional means we're fine with the file not
 			// existing, not that we're fine with failing to upload content
-			if isErrArtifact && errArtifact.Optional {
+			if isErrArtifact && errArtifact.Optional && errArtifact.Reason != "file-not-readable-on-worker" {
 				return nil
 			}
 
@@ -255,14 +274,13 @@ func (atf *ArtifactTaskFeature) FindArtifacts() {
 	atf.artifacts = payloadArtifacts
 }
 
-// File should be resolved as an S3Artifact if file exists as file and is
-// readable, otherwise i) if it does not exist as a "file-missing-on-worker" ErrorArtifact,
-// or ii) if it cannot be read by the task user, as a "file-not-readable-on-worker" ErrorArtifact,
-// otherwise if it exists as a directory, as an "invalid-resource-on-worker" ErrorArtifact.
-// A directory should resolve as `nil` if directory exists as directory and is readable,
-// otherwise i) if it does not exist or ii) cannot be read, as a "file-missing-on-worker"
-// ErrorArtifact, otherwise if it exists as a file, as
-// "invalid-resource-on-worker" ErrorArtifact
+// File should be resolved as an S3Artifact if file exists as a file, otherwise
+// i) if it does not exist as a "file-missing-on-worker" ErrorArtifact,
+// otherwise if it exists as a directory, as an "invalid-resource-on-worker"
+// ErrorArtifact. A directory should resolve as `nil` if directory exists as
+// directory and is readable, otherwise i) if it does not exist or ii) cannot
+// be read, as a "file-missing-on-worker" ErrorArtifact, otherwise if it exists
+// as a file, as "invalid-resource-on-worker" ErrorArtifact
 // TODO: need to also handle "too-large-file-on-worker"
 func resolve(base *artifacts.BaseArtifact, artifactType, path, contentType, contentEncoding string, pd *process.PlatformData, taskDir string) artifacts.TaskArtifact {
 	fullPath := fileutil.AbsFrom(taskDir, path)
@@ -311,16 +329,6 @@ func resolve(base *artifacts.BaseArtifact, artifactType, path, contentType, cont
 			BaseArtifact: base,
 			Message:      fmt.Sprintf("File artifact '%s' exists but is not a regular file (%v)", fullPath, fileinfo.Mode().Type()),
 			Reason:       "invalid-resource-on-worker",
-			Path:         path,
-		}
-	}
-
-	tempPath, err := copyToTempFileAsTaskUser(fullPath, pd, taskDir)
-	if err != nil {
-		return &artifacts.ErrorArtifact{
-			BaseArtifact: base,
-			Message:      fmt.Sprintf("Could not copy file '%s' to temporary location as task user: %v", fullPath, err),
-			Reason:       "file-not-readable-on-worker",
 			Path:         path,
 		}
 	}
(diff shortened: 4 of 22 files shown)


workers/generic-worker/artifacts_multiuser.go
```diff
@@ -3,23 +3,69 @@
...
+		if errors.As(err, &exitErr) && exitErr.ExitCode() == int(CANT_CAT_FILE) {
+			return out.written, artifacts.ContentError{Err: readErr}
+		}
+		return out.written, readErr
 	}
+	return out.written, nil
+}
 
-	return strings.TrimSpace(string(output)), nil
+type recordingWriter struct {
```
Eijebong, line 58: This I'm not super happy about. But I couldn't find any other way to distinguish a "can't read artifact" and "the artifact copy failed for any other given reason"

lotas:
> Uh oh! Looks like an error!
> 
> Taskcluster-GitHub attempted to cancel previously created task groups with following scopes:
> 
> ```
> assume:repo:github.com/taskcluster/taskcluster:*, queue:seal-task-group:taskcluster-github/*, queue:cancel-task-group:taskcluster-github/*
> ```
> 
> Client ID static/taskcluster/github does not have sufficient scopes and is missing the following scopes:
> 
> ```
> queue:seal-task-group:taskcluster-level-3/Dyftiu7sQaWotuHCNGHqaw
> ```
> 
> This request requires the client to satisfy the following scope expression:
> 
> ```
> queue:seal-task-group:taskcluster-level-3/Dyftiu7sQaWotuHCNGHqaw
> ```
> 
>     * method:     sealTaskGroup
> 
>     * errorCode:  InsufficientScopes
> 
>     * statusCode: 403
> 
>     * time:       2026-09-08T11:42:24.538Z

whoops https://github.com/taskcluster/community-tc-config/pull/1007

petemoore:
> There's a few bonus points in this refactor, first of all, when gzipping an artifact, we avoid a full copy. Before this, the file was copied in a temp file, then gzipped in a second one. Since we're streaming the content, we can just pass it directly through gzip. That's one less copy.

If I remember correctly, originally we did not copy and then gzip, and did it in one shot too. But we discovered an issue, which is if a zombie process continues to write to a file, the file can continue to change after the task has resolved. There was a race in this case, e.g. I think in chain of trust where we would take a hash of a file, but by the time we had then gzip compressed it, the content had changed. Or maybe it was a different race, but what I remember is we got bitten by files that changed after the last task command had completed. We decided at the time the best solution was to copy the file somewhere else, to ensure that file didn't change when performing further operations on it like zipping it up, or taking a hash of it for chain of trust, etc. The git history will probably have some context here if you take a look. @matt-boris may also have some opinions/memories on this topic.

Eijebong:
Well it doesn't change that. It still makes a copy. It just gzips it on the fly instead of doing copy -> gzipped copy. If a task was still writing to the file it'll end up with the same result, just with one less copy. Also I'm fairly certain I fixed that bug anyway without knowing it even existed in 5562f302eee4c17fa8be91bbf7fa242f4b3aa801

changelog/bug-2069332.md
```diff
@@ -0,0 +1,5 @@
+audience: worker-deployers
+level: patch
+reference: bug 2069332
+---
+When uploading artifacts, generic-worker will now create temporary files as the worker and stream content into it as the user.
```
petemoore:
```suggestion
When uploading artifacts, Generic Worker multiuser engine will now create temporary files as the worker user (`root`/`LocalSystem`) and stream content into them as the task user.
```

workers/generic-worker/artifacts/s3.go
```diff
@@ -23,86 +23,77 @@ type S3Artifact struct {
...
+	s3Artifact.bodyPath, s3Artifact.bodySHA256, s3Artifact.ContentLength, err = s3Artifact.createTempFileForPUTBody()
+	return err
+}
+
+func (s3Artifact *S3Artifact) DiscardContent() {
+	if s3Artifact.bodyPath != "" {
+		os.Remove(s3Artifact.bodyPath)
+		s3Artifact.bodyPath = ""
+		s3Artifact.bodySHA256 = ""
+	}
```
petemoore, line 51: if this is meant to be safe for concurrent access a mutex may be helpful here. although i can't imagine why the s3artifact would get reused..

workers/generic-worker/artifacts/s3.go
```diff
@@ -23,86 +23,77 @@ type S3Artifact struct {
...
-	tmpFile, err := os.CreateTemp("", baseName)
+	tmpFile, err := os.CreateTemp("", "artifact-")
 	if err != nil {
-		panic(err)
+		return "", "", 0, err
 	}
-	defer tmpFile.Close()
+	defer func() {
+		if err != nil {
+			tmpFile.Close()
```
petemoore, line 66: can tmpFile.Close() also return an error? if so, doesn't that get swallowed?

workers/generic-worker/artifacts/s3.go
```diff
@@ -23,86 +23,77 @@ type S3Artifact struct {
...
+			os.Remove(tmpFile.Name())
```
petemoore: does this return an error? should this be `err = os.Remove(tmpFile.Name())` ?

workers/generic-worker/artifacts/s3.go
```diff
@@ -23,86 +23,77 @@ type S3Artifact struct {
...
-	source, err := openContent(s3Artifact.ContentPath, s3Artifact.Path)
-	if err != nil {
-		panic(err)
-	}
-	defer source.Close()
 	hasher := sha256.New()
-	if _, err := io.Copy(io.MultiWriter(target, hasher), source); err != nil {
-		panic(err)
+	if contentLength, err = s3Artifact.Content.WriteContent(io.MultiWriter(target, hasher)); err != nil {
+		return "", "", 0, err
```
petemoore: note ~~if you didn't shadow `err`,~~ `return` would be sufficient here, although the more verbose form does no harm

workers/generic-worker/artifacts/s3.go
```diff
@@ -23,86 +23,77 @@ type S3Artifact struct {
...
+}
+
+// Streams the artifact content into a temporary file owned by the worker,
+// applying ContentEncoding on the way. The path of the temporary file is
+// returned, alongside the sha256 and the length of the uncompressed content
+// that got written.
+func (s3Artifact *S3Artifact) createTempFileForPUTBody() (path string, sha256sum string, contentLength int64, err error) {
 	baseName := filepath.Base(s3Artifact.Path)
-	tmpFile, err := os.CreateTemp("", baseName)
+	tmpFile, err := os.CreateTemp("", "artifact-")
```
petemoore, line 60: ~~i think this is going to shadow `err` but no big deal~~

UPDATE: Seems I was wrong about this

changelog/bug-2069332-1.md
```diff
@@ -0,0 +1,7 @@
+audience: users
+level: major
+reference: bug 2069332
+---
+When uploading an optional artifact, if it's not readable (or encounters any
+unreadable file for directory artifacts), the task will now fail instead of
+silently omitting that file.
```
petemoore, line 7: It would be good to add a test for this

workers/generic-worker/artifacts/s3.go
petemoore: actually, since error is already non-nil, maybe better to just log new error and return the original error

workers/generic-worker/artifacts/s3.go
Eijebong: Nah, each artifact is confined to its own goroutine. There's no sharing between them so this is safe as is.

workers/generic-worker/artifacts/s3.go
Eijebong: It was already swallowed before and was actually double closing. This is on purpose, we don't want to hide the actual error, this is a best effort thing.

workers/generic-worker/artifacts/s3.go
Eijebong: Same thing here yeah, we don't want to hide the original error. This one I guess we could log since it'd leave files behind

changelog/bug-2069332-1.md
Eijebong: That's pretty much what the `TestPrivilegedOptionalFileUploadDoesNotFailTest` -> `TestPrivilegedOptionalFileUploadFailsTask` flip is

Eijebong: I guess I could add a test for the directory walk case though

ui/docs/reference/workers/generic-worker/usage.mdx
```diff
@@ -20,7 +20,7 @@ and reports back results to the queue.
                                             [--worker-runner-protocol-pipe PIPE]
     generic-worker show-payload-schema
     generic-worker new-ed25519-keypair      --file ED25519-PRIVATE-KEY-FILE
-    generic-worker copy-to-temp-file        --copy-file COPY-FILE
+    generic-worker cat-file                 --cat-file CAT-FILE
```
matt-boris, line 23: I'm torn whether to call this "breaking" change out in one of the changelogs...

matt-boris: Oh small nit.. probably worth updating CLAUDE.md at this point

https://github.com/taskcluster/taskcluster/blob/5b76c14a50dea918fd8cf41ece314a890acaf6ea/workers/generic-worker/CLAUDE.md?plain=1#L166

workers/generic-worker/artifact_feature.go
```diff
@@ -111,12 +112,32 @@ func (atf *ArtifactTaskFeature) Stop(err *ExecutionErrors) {
...
+				}
+			}
+
 			e := task.uploadArtifact(taskArtifact)
 			errArtifact, isErrArtifact := taskArtifact.(*artifacts.ErrorArtifact)
 			// An artifact we never got readable content for is published as an
 			// error artifact. Optional means we're fine with the file not
 			// existing, not that we're fine with failing to upload content
-			if isErrArtifact && errArtifact.Optional {
+			if isErrArtifact && errArtifact.Optional && !contentUnreadable {
```
matt-boris: Nit: I don't love tracking if the content is readable in this separate boolean (especially with the naming, `contentUnreadable` makes checks like `!contentUnreadable` awkward/hard to read/follow). Why not just use the `Reason` field on your check?


```suggestion
			if isErrArtifact && errArtifact.Optional && errArtifact.Reason == "file-missing-on-worker" {
```

workers/generic-worker/artifacts.go
```diff
@@ -174,25 +169,3 @@ func (task *TaskRun) classifyCreateArtifactError(artifact artifacts.TaskArtifact
...
-		// Example output:
-		// 2023/11/07 19:56:06Z Making system call GetProfilesDirectoryW with args: [C0000C15F0 C00027E980]
-		// 2023/11/07 19:56:06Z   Result: 1 0 The operation completed successfully.
-		// C:\Windows\SystemTemp\TestPrivilegedFileUpload664016823956663638
-		outputLines := strings.Split(tempFilePath, "\n")
-		tempFilePath = strings.TrimSpace(outputLines[len(outputLines)-1])
-	}
-
-	return
-}
```
matt-boris, line 198: Good riddance... 😭

Thank you!

ui/docs/reference/workers/generic-worker/usage.mdx
Eijebong: Oh yeah I'll fix the claude md, wouldn't want it to hallucinate commands now would we...

On the breaking change part yeah, I dunno. I can mark it breaking even though that command is internal. Numbers are free. I just didn't want to alarm anyone wanting to update since it's just a worse `cp`

workers/generic-worker/artifact_feature.go
Eijebong: Mhhh I had this exact code at some point and I got rid of it, I don't remember why, I'll have a look because the reason I did might not be real anymore. That code changed at least 5 times, I struggled really hard to get it into a place I liked

matt-boris: Lol I just also noticed I incorrectly commented `!contentReadable` (what my brain _wants to_ read 😄). Just edited the initial comment...

ui/docs/reference/workers/generic-worker/usage.mdx
Eijebong: Oh wait I already marked the d2g breakage as major... I'll mark the other one too and call that out just in case

workers/generic-worker/artifact_feature.go
Eijebong: We wanted `file-not-readable-on-worker`, not `file-missing-on-worker` but yeah. Switched

matt-boris (dismissed):
Nice, thanks. LGTM

petemoore (dismissed):
Looks great, thanks Bastien!

PR #9133, opened 2026-09-08
fix(d2g): re-pull named images and run by digest

### LINKED ISSUE #8004: Non artifact docker images don't get pulled anymore before the task runs
This is a regression from 85a2287d04cdb66637e5867c66998a52014d82f0, before that, the worker would always try to pull docker images, getting updates if the tag was pointing to something different than what the workers has in store.

Cache looks like this for those images:

```
  "ghcr.io/eijebong/taskcluster-images/decision-task:main": {
    "id": "ghcr.io/eijebong/taskcluster-images/decision-task:main",
    "name": "ghcr.io/eijebong/taskcluster-images/decision-task:main"
  },
```

And the code bypasses the docker pull step if the name of the image is present at all in the cache. It should instead bypass that check if the image is not an artifact one.

matt-boris:
Registry cache hits skipped `docker pull`, so a
poisoned local tag could persist across tasks.
Always re-pull `name:tag` images and pin
`docker run` to the content digest.

Fixes bug 2069337. Also re-fixes #8004.

>D2G: named/registry docker images are always re-pulled and `docker run` pins them by digest, so a poisoned local tag cannot persist across tasks.


Eijebong:
FTR I'm not sure this is worth the complexity. The prerequisite for polluting the local image registry is to be able to run docker on the bare worker (no d2g). If you can do this you're root on the worker. It's much easier to just replace the docker binary or generic-worker itself to get something out of it than overwrite an image.

This is where d2g loses entirely compared to having a proper docker worker

Edit: NVM, turns out this also re-fixes #8004 because it got broken by capacity > 1

workers/generic-worker/d2g_feature.go
```diff
@@ -170,35 +170,30 @@ func (dtf *D2GTaskFeature) Start() *CommandExecutionError {
...
-			// same image (see https://github.com/taskcluster/taskcluster/issues/7967)
-			idLine, _, _ := strings.Cut(strings.TrimSpace(string(out)), "\n")
-			imageID = strings.TrimPrefix(idLine, "sha256:")
+		cmd, err = process.NewCommandNoOutputStreams([]string{
+			"docker",
+			"images",
+			"--no-trunc",
+			"--quiet",
+			imageName,
+		}, taskDir, []string{}, dtf.task.pd)
```
Eijebong, line 180: This ain't good enough.

```
[eijemoz@mozium] ~/foo  >>> docker pull archlinux
Using default tag: latest
latest: Pulling from library/archlinux
41e227706d01: Already exists
195e507860df: Already exists
Digest: sha256:b944cc65c5f28665dfd5fdbf5ed2997c88f5bb4a0aefac7ee8a7ef01893e5ed9
Status: Downloaded newer image for archlinux:latest
docker.io/library/archlinux:latest
[eijemoz@mozium] ~/foo  >>> docker images --no-trunc --quiet archlinux
sha256:84e88cacd0e42ea6f23b45c9760c2942b05421cdd76ef33d3011c284bf29cf23

[eijemoz@mozium] ~/foo  >>> printf 'FROM archlinux:latest\nRUN echo "surprise baguette" > /oops\n' > Dockerfile
[eijemoz@mozium] ~/foo  >>> docker build -t archlinux:gift .
[+] Building 0.0s (6/6) FINISHED                                                                                               docker:default
 => [internal] load build definition from Dockerfile                                                                                     0.0s
 => => transferring dockerfile: 96B                                                                                                      0.0s
 => [internal] load metadata for docker.io/library/archlinux:latest                                                                      0.0s
 => [internal] load .dockerignore                                                                                                        0.0s
 => => transferring context: 2B                                                                                                          0.0s
 => [1/2] FROM docker.io/library/archlinux:latest                                                                                        0.0s
 => CACHED [2/2] RUN echo "surprise baguette" > /oops                                                                                    0.0s
 => exporting to image                                                                                                                   0.0s
 => => exporting layers                                                                                                                  0.0s
 => => writing image sha256:25a7ef48761d48b5a36df1f6a43fa0cd6d5af835ec06beefdc9822d0f4fcf5fb                                             0.0s
 => => naming to docker.io/library/archlinux:gift   

[eijemoz@mozium] ~/foo  >>> docker images --no-trunc --quiet archlinux
sha256:25a7ef48761d48b5a36df1f6a43fa0cd6d5af835ec06beefdc9822d0f4fcf5fb # My gift
sha256:84e88cacd0e42ea6f23b45c9760c2942b05421cdd76ef33d3011c284bf29cf23 # The actual image
```

Ordering won't work, it's going to list the latest "created" first always and that's also easily falsifiable.

Eijebong: It's not enough twice over.

```
[eijemoz@mozium] ~/foo  >>> docker images --no-trunc --quiet archlinux
[eijemoz@mozium] ~/foo  >>> docker image pull archlinux@sha256:b944cc65c5f28665dfd5fdbf5ed2997c88f5bb4a0aefac7ee8a7ef01893e5ed9
docker.io/library/archlinux@sha256:b944cc65c5f28665dfd5fdbf5ed2997c88f5bb4a0aefac7ee8a7ef01893e5ed9: Pulling from library/archlinux
41e227706d01: Already exists
195e507860df: Already exists
Digest: sha256:b944cc65c5f28665dfd5fdbf5ed2997c88f5bb4a0aefac7ee8a7ef01893e5ed9
Status: Downloaded newer image for archlinux@sha256:b944cc65c5f28665dfd5fdbf5ed2997c88f5bb4a0aefac7ee8a7ef01893e5ed9
docker.io/library/archlinux@sha256:b944cc65c5f28665dfd5fdbf5ed2997c88f5bb4a0aefac7ee8a7ef01893e5ed9
[eijemoz@mozium] ~/foo  >>> docker images --no-trunc --quiet archlinux@sha256:b944cc65c5f28665dfd5fdbf5ed2997c88f5bb4a0aefac7ee8a7ef01893e5ed9
[eijemoz@mozium] ~/foo  >>>
```

`archlinux@sha256:b944cc65c5f28665dfd5fdbf5ed2997c88f5bb4a0aefac7ee8a7ef01893e5ed9` is a perfectly cromulent image to be using

Eijebong: I think both can be fixed with `docker image inspect --format={{.Id}} <imageName>`

```
[eijemoz@mozium] ~/foo  >>> docker image inspect --format={{.Id}} archlinux@sha256:b944cc65c5f28665dfd5fdbf5ed2997c88f5bb4a0aefac7ee8a7ef01893e5ed9
sha256:84e88cacd0e42ea6f23b45c9760c2942b05421cdd76ef33d3011c284bf29cf23
```

```
[eijemoz@mozium] ~/foo  >>> docker image pull archlinux
Using default tag: latest
latest: Pulling from library/archlinux
Digest: sha256:b944cc65c5f28665dfd5fdbf5ed2997c88f5bb4a0aefac7ee8a7ef01893e5ed9
Status: Downloaded newer image for archlinux:latest
docker.io/library/archlinux:latest
[eijemoz@mozium] ~/foo  >>> docker build -t archlinux:gift .
[+] Building 0.0s (6/6) FINISHED                                                                                                                                                                                                                                             docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                                   0.0s
 => => transferring dockerfile: 96B                                                                                                                                                                                                                                                    0.0s
 => [internal] load metadata for docker.io/library/archlinux:latest                                                                                                                                                                                                                    0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                                                      0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                                        0.0s
 => [1/2] FROM docker.io/library/archlinux:latest                                                                                                                                                                                                                                      0.0s
 => CACHED [2/2] RUN echo "surprise baguette" > /oops                                                                                                                                                                                                                                  0.0s
 => exporting to image                                                                                                                                                                                                                                                                 0.0s
 => => exporting layers                                                                                                                                                                                                                                                                0.0s
 => => writing image sha256:25a7ef48761d48b5a36df1f6a43fa0cd6d5af835ec06beefdc9822d0f4fcf5fb                                                                                                                                                                                           0.0s
 => => naming to docker.io/library/archlinux:gift                                                                                                                                                                                                                                      0.0s
[eijemoz@mozium] ~/foo  >>> docker image inspect --format={{.Id}} archlinux
sha256:84e88cacd0e42ea6f23b45c9760c2942b05421cdd76ef33d3011c284bf29cf23 # That's :latest, correct
```

workers/generic-worker/d2g_feature.go
Eijebong: I don't know why github isn't allowing me to make a code suggestion here...

This whole thing could be:

```go
image, loadedImage, loadErr := dtf.loadImageLocked(key, loadImage, isImageArtifact)
if loadErr != nil {
    return loadErr
}
```

and you can remove the `image := dtf.imageCache[key]` above. It's a nice cleanup and avoids a random unnamed boolean.

Eijebong: Probably want to gate that on `allowCacheHit` to avoid polluting the cache with stuff that can never be hit

Eijebong (changes requested):

workers/generic-worker/d2g_feature.go
```diff
@@ -170,63 +168,38 @@ func (dtf *D2GTaskFeature) Start() *CommandExecutionError {
...
+			"docker",
+			"image",
+			"inspect",
+			"--format={{.Id}}",
+			imageName,
+		}, taskDir, []string{}, dtf.task.pd)
+		if err != nil {
+			return nil, executionError(internalError, errored, fmt.Errorf("[d2g] could not create process to determine docker image ID: %v", err))
+		}
+		out, err = cmd.Output()
```
Eijebong, line 192: Sorry, I should've caught this in the first review. There's a TOCTOU here.


This does:

```
docker load < task-image.tar   # or docker pull -q imageName
# short period here where or a concurrent d2g task could load an image that ends up with the same tag but isn't correct. (Or something doing `docker tag evil imageName` which is the original problem that got reported even though it's very unlikely to happen. The artifact path is a real problem)
docker image inspect imageName
```

workers/generic-worker/d2g_feature.go
```diff
@@ -170,63 +168,38 @@ func (dtf *D2GTaskFeature) Start() *CommandExecutionError {
 			if !imageNameFound {
 				return nil, executionError(internalError, errored, fmt.Errorf("[d2g] could not determine docker image name from docker load output:\n%v", string(out)))
 			}
+		} else {
+			imageName = key
```
Eijebong: `imageName := lines[0]` is dead code. You reassign `imageName` in all branches that exit without an error. The comment above also needs updating

Eijebong: Something like this?

```go
var imageName string
if isImageArtifact {
      imageNameFound := false
      // Find the first line with "Loaded image: " prefix
      // (see https://github.com/taskcluster/taskcluster/issues/7969)
      for line := range strings.SplitSeq(strings.TrimSpace(string(out)), "\n") {
              if name, found := strings.CutPrefix(line, "Loaded image: "); found {
                      imageName = name
                      imageNameFound = true
                      break
              }
      }
      if !imageNameFound {
              return nil, executionError(internalError, errored, fmt.Errorf("[d2g] could not determine docker image name from docker load output:\n%v", string(out)))
      }
} else {
      imageName = key
}
```

workers/generic-worker/d2g_feature.go
Eijebong:
```
> docker pull ubuntu
Using default tag: latest
latest: Pulling from library/ubuntu
08f5f5b2a2b0: Pulling fs layer
6a7c4f6d8c38: Pulling fs layer
6a7c4f6d8c38: Download complete
08f5f5b2a2b0: Download complete
08f5f5b2a2b0: Pull complete
6a7c4f6d8c38: Pull complete
Digest: sha256:513c074113a871b51a8d16ab445c88779d6452d937a164fb5cc479f32668a41d
Status: Downloaded newer image for ubuntu:latest
docker.io/library/ubuntu:latest

> docker pull archlinux
Using default tag: latest
latest: Pulling from library/archlinux
Digest: sha256:b944cc65c5f28665dfd5fdbf5ed2997c88f5bb4a0aefac7ee8a7ef01893e5ed9
Status: Image is up to date for archlinux:latest
docker.io/library/archlinux:latest
```

You probably want to parse this, get the Digest and inspect with that? Not sure if there's a catch there though. My testing shows it's there all the time even when pulling is a no-op (the arch pull) so it's probably fine.

However that won't work for artifacts and there I don't have any great idea at the moment.

```
> docker load < image.tar.zst
Loaded image: lint:latest
```

We could grab it from manifest.json but I'm not very happy about how expensive that is because we need to decompress the whole thing. I guess we could do this and the load at the same time, dunno if zst is parallel or not...

```
> du -sh image.tar.zst
312M	image.tar.zst
> time tar xOf image.tar.zst --occurrence=1 manifest.json | jq '.[0].Config'
"ca03cb6f10721facf7cee09835f18fd6fa8f240da9030289a0afc37cae07ba95.json"
tar xOf image.tar.zst --occurrence=1 manifest.json  0.64s user 0.28s system 163% cpu 0.558 total
jq '.[0].Config'  0.00s user 0.00s system 0% cpu 0.558 total
> docker load < image.tar.zst
Loaded image: lint:latest
> docker image inspect --format="{{.Id}}" lint:latest
sha256:ca03cb6f10721facf7cee09835f18fd6fa8f240da9030289a0afc37cae07ba95
```

Eijebong (changes requested):

workers/generic-worker/d2g_feature.go
matt-boris: Going to push this up into my recent commit, then I'll address the TOCTOU in another commit in the PR.

workers/generic-worker/d2g_feature.go
matt-boris: I'll address the registry images first in its own commit, then will dig around for a nice way to do it with artifact images. Thanks for helping look into what's possible here.

workers/generic-worker/d2g_feature.go
```diff
@@ -164,7 +164,20 @@ func (dtf *D2GTaskFeature) Start() *CommandExecutionError {
 				return nil, executionError(internalError, errored, fmt.Errorf("[d2g] could not determine docker image name from docker load output:\n%v", string(out)))
 			}
 		} else {
-			imageName = key
+			var dgst string
```
Eijebong: Any reason to save 2 whole characters here and make me think about what dgst means?

workers/generic-worker/d2g_feature.go
matt-boris: I've some other thoughts for artifact images as well:

* Use a mutex to only allow for one artifact image load at a time (easiest)
* Use a teereader so every time docker load reads stdin, also writes those bytes to a pipe. Then a goroutine reads the other end of the pipe and parses manifest.json. (this avoids two decompresses)
* Remove the archive metadata (`RepoTags` in `manifest.json`, `repositories`, `index.json`) so that a docker load returns something like: `Loaded image ID: sha256:e5f44b4f...`

Eijebong: A mutex doesn't fix the original problem that was reported though.

I don't think you can remote the archive metadata without opening it? And we shouldn't depend on that preexisting IMO

The tee might work though? I have concerns about memory usage but worth looking at

Eijebong (commented):
Love that because I had something queued in review I have to post a full review to respond to your comment. 12/10 UX right there

workers/generic-worker/d2g_feature.go
matt-boris: I've gotta figure out a more durable path forward because the example you gave only works for storage backends of overlay2. On my mac, using overlayfs, I get this:

```bash
➜  du -sh image.tar.zst
353M    image.tar.zst
➜  time tar xOf image.tar.zst manifest.json | jq '.[0].Config'
"13c5108c2dfa99f4f52ee1bb0dacfb9e1ac25190bcd02fffeb28612694702487.json"
tar xOf image.tar.zst manifest.json  0.95s user 0.39s system 140% cpu 0.957 total
jq '.[0].Config'  0.00s user 0.00s system 0% cpu 0.956 total
➜  docker load < image.tar.zst
Loaded image: ci:latest
➜  docker image inspect --format="{{.Id}}" ci:latest
sha256:b627a1db04d6636c9191bc4b90ebe51f1a1d668fe855ead8bb03aa8164fb446b
```

I don't want to create the solution just based off of what we configure our docker engine to be https://github.com/mozilla-platform-ops/worker-images/blob/e39d337db780ae52cb909bfa68f489f8714b9eb2/scripts/linux/common/docker-config.sh#L17

I think I'm leaning towards rewriting the name metadata in the archive before load so that `docker load` provides us with the image id we can use to run.

PR #9135, opened 2026-09-09
feat(ci): trigger trusted CI for Dependabot PRs

matt-boris:
Dependabot PRs only get untrusted (level 1) checks under public_restricted, so auto-merge can stall on a graph that never runs. Comment `/taskcluster run` after approval so the trusted graph starts.

### DIFF
--- a/.github/workflows/dependabot-automerge.yml
+++ b/.github/workflows/dependabot-automerge.yml
@@ -24,6 +24,11 @@ jobs:
         env:
           PR_URL: ${{ github.event.pull_request.html_url }}
           GITHUB_TOKEN: ${{ secrets.DEPENDABOT_AUTOMERGE_TOKEN }}
+      - name: Trigger trusted CI
+        run: gh pr comment "$PR_URL" --body "/taskcluster run"
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-automerge.yml
```diff
@@ -24,6 +24,12 @@ jobs:
         env:
           PR_URL: ${{ github.event.pull_request.html_url }}
           GITHUB_TOKEN: ${{ secrets.DEPENDABOT_AUTOMERGE_TOKEN }}
+      - name: Trigger trusted CI
+        if: ${{ contains(fromJson('["version-update:semver-patch", "version-update:semver-minor"]'), steps.metadata.outputs.update-type) }}
```
Eijebong: Wonder if we'd want to run this when it's a major also? I understand not approving/merging but running CI should be fine?

matt-boris: Oh good point

matt-boris: Removed the conditional

Eijebong (approved):

PR #9139, opened 2026-09-09
chore(auth): Drop deprecated and no longer used Azure endpoints in auth service

lotas:
This dates back to the pre-postgres times when everything was stored in azure tables (storage) and auth service was also giving out credentials on demand. Methods were deprecated long ago.

Another reason is that tests relied on
`project/taskcluster/testing/azure` secret which holds credentials that are no longer valid (tenant was removed), so effectively dead code in tests also since it wasn't tested.

Breaking change for the deployers since helm values would likely fail validation if they still contain `auth.azure_accounts`, so it needs to be dropped before the upgrade.

Bugzilla Bug: [2066797](https://bugzilla.mozilla.org/show_bug.cgi?id=2066797)


services/auth/test/helper.js
```diff
@@ -65,7 +61,7 @@ helper.secrets = new Secrets({
 
 helper.loadJson = async filename => JSON.parse(await fs.readFile(path.join(__dirname, filename), 'utf8'));
 
-helper.withCfg = (mock, skipping) => {
+helper.withCfg = (_mock, skipping) => {
```
Eijebong: IIRC everywhere else this happened I just removed the mock parameter instead

lotas: dropping, thanks

Eijebong (approved):
🚮

petemoore (commented):
Looks great, thanks Yarik!

PR #9140, opened 2026-09-09
fix(tests): expect ACL on loopback video devices in both gw engines

matt-boris:
Worker images now install v4l2loopback via DKMS
(see https://github.com/mozilla-platform-ops/worker-images/pull/910), so `/dev/videoN` consistently has POSIX ACLs on insecure as well as multiuser. The engine split from #7551 is no longer needed.


Eijebong (approved):
Ah I'm guessing that's the failure in #9141 
Makes sense, thank you

PR #9145, opened 2026-09-10
Migrate CI from community-tc to firefox-ci

petemoore:
Migration from community-tc to firefox-ci.


petemoore:
Note, after this lands, we should create a staging release, before we create a real release. I'm creating an issue for that!

lotas (approved):
LGTM

PR #9151, opened 2026-09-11
Disable fsync, full_page_writes and synchronous_commit on postgresql on CI

Eijebong:
We've had a few tasks that failed due to timeouts and the failures were very suspiciously slow around database operations. The clone was also slower than usual on those tasks which makes me think that the disks on those workers were not in a happy place. We can't fix slow disks but we can try to write a bit less and stop waiting for said writes.

The best thing would obviously be to move the database to tmpfs instead, but d2g doesn't allow for that so we take the next best thing.

We don't need [fsync] (or any of its [full_page_writes], [synchronous_commit] friends) on CI, if the database dies it dies forever anyway so there's literally no drawback to disabling it, and it should at least move those writes out of hot paths. See [non-durability] docs for postgresql where they recommend exactly that.

[non-durability]: https://www.postgresql.org/docs/15/non-durability.html
[fsync]: https://www.postgresql.org/docs/15/runtime-config-wal.html#GUC-FSYNC
[full_page_writes]: https://www.postgresql.org/docs/15/runtime-config-wal.html#GUC-FULL-PAGE-WRITES
[synchronous_commit]: https://www.postgresql.org/docs/15/runtime-config-wal.html#GUC-SYNCHRONOUS-COMMIT

### DIFF
--- a/changelog/LwumgGt5Shuu02EAYbhq2w.md
+++ b/changelog/LwumgGt5Shuu02EAYbhq2w.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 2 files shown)


Eijebong:
Well I can't get a slow worker but this shouldn't hurt and would put the database mostly out of the way in case we get timeouts again.

matt-boris (approved):

PR #9152, opened 2026-09-11
Fix flaky email notify test

Eijebong:
That test uses a shared SES queue for all jobs running at the same time. The previous code was eagerly deleting everything it found in that queue as soon as it found one message which means 2 tests running concurrently would make one of them fail since its message would've gotten stolen. It also means that a test could pass despite having broken email sending since it could get emails from another concurrent test.

Since I don't want to split queues, both because I have no idea if we have creds that can do that and mostly because I don't want to leak them if a test ever fails to cleanup, I instead went this way. Every message is now sent to `success+<randomID>-<seq>@` and the check loop ignores *anything* that doesn't match. So two tests can't just take eachother messages. This removes both the failure case and fixes the problem where one test could pass for the wrong reason.

Messages that do end up orphaned get cleared up by AWS after `MessageRetentionPeriod` which should be set to something sensible on the test queue.

### DIFF
--- a/changelog/ZLmktQtcSqyJwhPlDkP_Pw.md
+++ b/changelog/ZLmktQtcSqyJwhPlDkP_Pw.md
@@ -1,0 +1,2 @@
+level: silent
+---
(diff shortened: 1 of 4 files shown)


matt-boris (approved):
Thanks, lgtm

PR #9153, opened 2026-09-11
chore: Recreate the azure-attested-document hook on firefox-ci

### LINKED ISSUE #9149: Recreate the azure-attested-document hook on firefox-ci
community-tc-config defines a weekly hook (`azure-attested-document`) that runs on `proj-taskcluster/gw-windows-2022-gui`, fetches the Azure IMDS attested document and instance metadata, and publishes them at index route `project.taskcluster.azure.attested-document.latest`, so `services/worker-manager/test/fixtures/azure_signature_good.json` can be refreshed when its leaf certificate expires (~6 months; see #8533).

It was not migrated. Recreate it in fxci-config `hooks.yml` (hook group `project-taskcluster`) with the same schedule, pool and index route, then update the `curl` in #8533 / the hook description to the firefox-ci index URL.

petemoore:
Fixes #9149

Should be landed _after_ https://github.com/mozilla-releng/fxci-config/pull/1207 and _before_ https://github.com/taskcluster/community-tc-config/pull/1012.


taskcluster/src/target_tasks.py
```diff
@@ -37,6 +37,17 @@ def filter(task):
...
 
 
+@register_target_task("azure-attested-document")
+def target_tasks_azure_attested_document(full_task_graph, parameters, graph_config):
+    """The weekly cron job from .cron.yml: just the attested-document task."""
+    return [
+        label
+        for label, t in full_task_graph.tasks.items()
+        if standard_filter(t, parameters)
+        and t.attributes.get("cron-job") == "azure-attested-document"
```
ahal, line 47: You could just check `label == azure-attested-document`, but this is fine too, especially if you expect to need to add more tasks to this cron in the future.

petemoore: I kind of like the protection that the current mechanism works even if more cron tasks are added, as you noted, so I would prefer to leave it as is. Aside: I think if we were to check the label directly it would need to be `azure-attested-document-fetch` rather than `azure-attested-document`.

Eijebong (approved):

PR #9154, opened 2026-09-14
Fix dependencies and dependents list heights

### LINKED ISSUE #9142: Dependencies / Dependents list should have identical maximum height and indication if they are scrollable
**Describe the bug**
For https://firefox-ci-tc.services.mozilla.com/tasks/R7qPQZGsRwmvCqAq9zNMnw is two-column view (15'' screen), the left column contains lists for 'Dependencies' and 'Dependents'. The 'Dependencies' list holds more elements than it can show but there is no indicator about the additional elements unless one moves the mouse over it. Combined with the taller 'Dependents' list, one can assume all elements shown are all elements there are.

Suggestion: Use a maximum height for 'Dependencies' which only displays half of the last visible element.

s1h2r3e4-alt:
## Summary

Fixes #9142

- Added an optional maxHeight prop to ConnectionDataTable.
- Added a scrollable maximum-height style.
- Applied a matching 400px height to the Dependents list.
- Ensured Dependencies and Dependents lists have consistent heights and scrolling behavior.

## Testing

- Focused Biome check passed for both modified files.
- UI tests: 188 passed; one unrelated snapshot failure remains in ListHookGroups.


s1h2r3e4-alt:
Hi maintainers, this is my first contribution to Taskcluster.
The PR is ready for review. Some workflows are currently awaiting maintainer approval.
Thank you!

lotas:
Hi @s1h2r3e4-alt Thanks for adding the fix

This PR is missing a changelog entry briefly describing the fix and referencing the issue, please check https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/changelog.md 

UI tests are currently failing because snapshots needs to be updated before (`yarn test:update-snapshots`)

s1h2r3e4-alt:
Updated the PR with the requested changelog entry and refreshed UI snapshots. Thank you for the feedback!

lotas:
image

I tested this locally, it looks great!

To address some other comments mentioned in the original ticket, can you please check if we could also put the count next to the Dependents? Just like the Dependencies are showing? 
And if it will be possible to add sorting by name to make it even better?

Thanks!

s1h2r3e4-alt:
Thanks for testing it!

Yes, I’ll look into adding the total Dependents count next to the heading and enabling sorting by task name as well.

s1h2r3e4-alt:
I’ve added sorting by task name for the Dependents list.

For the Dependents count, I found that the current GraphQL TasksConnection and queue API don’t expose a total count, and the connection is paginated. Would you prefer extending the API to provide the total count, or should the count be limited to the currently loaded page?

lotas:
> I’ve added sorting by task name for the Dependents list.

can you please push this change (I dont' see it)

> For the Dependents count, I found that the current GraphQL TasksConnection and queue API don’t expose a total count, and the connection is paginated. 

thanks for looking into it. I thought we had already the counts in place, but if not, let's not do that part here to avoid creeping scope. Plus we are actively removing graphql layer.

So please push your sorting fix and updated snapshots, and this fix would be complete. Thank you

s1h2r3e4-alt:
The sorting fix has been pushed in commit 8ac9ba0. Thanks!

lotas:
Thanks! sorting part works fine :) 
image
the only issue is that filter now disappears if I go to second page, but this was probably an existing regression, not something this change introduced
image

Feel free to rebase once again and run `yarn test:update-snapshots` to update snapshots (to make failing test happier)

s1h2r3e4-alt:
Hi @lotas, I’ve rebased the branch onto the latest main, updated the UI snapshots, and pushed the changes. The sorting fix is included as well. Thank you!

matt-boris:
@allcontributors please add @s1h2r3e4-alt for code

PR #9159, opened 2026-09-15
chore: improve local dev experience

lotas:
UI dev container can be stale and not pick up latest packages installed

I've got `vite is not installed` because I don't rebuild dev containers often enough, with this it will be doing automatically and help avoid staring at "500 nginx error page"


petemoore (approved):
Awesome, thanks Yarik!

PR #9160, opened 2026-09-15
Bug 2072198 - Fix mounting caches on windows when they contain hardlinks with nLinks > 1

Eijebong:
This fixes a regression from 2865a7efd49520cae97f5dbb0e65d1ad5f9c9137 which added a refusal when mounting caches containing hardlinks if said hardlinks were not owned by the previous task user. Unfortunately I tested the case where I'd hardlink something from outside the cache into it, but not a hardlink from the cache into the same cache.

What happens then is that we end up with nLinks == 2, the chown runs on mount, chowns the file to the new user on the first one it encounters since it correctly belongs to the previous task user and then it encounters the second one. And it goes "I've never seen this user in my life, absolutely not" when in fact, the file is already owned by the target we're trying to chown to.

The fix is trivial and matches what I did on posix in 441c19ef1917c03c8cd9422294bcd1918fdbb2ed. If the link is already owned by the target, just allow it through.


lotas (approved):
looks right, thanks

PR #9161, opened 2026-09-15
Bug 2071940 - Use LSA to store the password for the next user on windows

Eijebong:
For some reason, winlogon allows reading the password for the next user from cleartext in the registry and for some other reason, that key is world readable. And that's the start of a story of me losing my mind on windows related problems... Again.

At least that drawback is documented at
https://learn.microsoft.com/en-us/troubleshoot/windows-server/user-profiles-and-logon/turn-on-automatic-logon

> The autologon feature is provided as a convenience. However, this
> feature may be a security risk. If you set a computer for autologon,
> anyone who can physically obtain access to the computer can gain access
> to all the computer's contents, including any networks it is connected
> to. Additionally, when autologon is turned on, the password is stored in
> the registry in plain text. The specific registry key that stores this
> value can be remotely read by the Authenticated Users group

There's also a very shy mention of the solution on that page, which we can follow to
https://learn.microsoft.com/en-us/windows/win32/secauthn/protecting-the-automatic-logon-password

And here goes madness.

Golang of course doesn't expose any of this. I wonder what the point is of having a windows module when from experience it's missing 80% of whatever one would use.

The documentation *lies*.

> Note that if Winlogon cannot find a password stored by the
> LsaStorePrivateData function, it will use the DefaultPassword value of
> the Winlogon key (if it exists) for the automatic logon password.

This is literally backwards. Tested both with the registry containing a valid and invalid passwords. In both case LSA gets completely ignored.

So yeah, back to writing unsafe code to wrap those APIs. At least the patch ended up being fairly simple if you ignore the sanity I lost in the process. This is basically the example given on their docs translated 1-1 into go.

The one thing that could surprise anyone is the `DeleteValue` which is here so workers don't get stuck with an invalid password in the registry. Passwords *obviously* are also not deleted by winlogon when a user gets removed. So an invalid password would linger in there if a worker was used on a version before this one, and would just stay there forever (or if the image build steps use this also). And since winlogon doesn't even read the LSA value if there's a registry one, the worker would stay stuck on login forever.

On the bright side, users can read that value but cannot write to it so there's no problem with a task setting an invalid value there and blocking a worker


workers/generic-worker/win32/lsa_windows.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"fmt"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+// Nothing LSA related is exposed in x/sys/windows, because why would it be...
```
matt-boris, line 11: 😭

petemoore (commented):
This looks very good, but to just call out that I think this deviates a little from our other win32 call sites that wrap the calls with some logging. Although the logging probably would not add a lot of value, it is something I added at the beginning to help troubleshoot if things go wrong.

Also I think we put the rest of the win32 api calls in the win32 package (`workers/generic-worker/win32`).

matt-boris (approved):
Looks reasonable, thank you for the fix

Eijebong:
> Also I think we put the rest of the win32 api calls in the win32 package (workers/generic-worker/win32).

Well it's in the win32 package. Assuming you meant `win32/win32_windows.go`. There's precedent there to not be doing this (win32acl_windows) and I like this better than adding to a file that's already way too long with win32 internals.

I don't think logging anything here would help because we don't want to log the password and the caller already logs enough

PR #9162, opened 2026-09-15
generic-worker: support preloaded writable directory caches

### LINKED ISSUE #8809: generic-worker: support preloaded writable directory caches from worker config
**Describe the feature request**

Generic-worker should support adopting existing directories as writable directory caches when the worker starts.

Right now, a writable directory cache is only created after a task asks for it in `payload.mounts`. That means the first task on a fresh worker always starts with an empty cache, even if the machine image already contains the data we want cached.

For Firefox CI, this makes workers spend a long time cloning Gecko before the task can do useful work. We could bake a checkout into the image, but the only way to make generic-worker use it today is to pre-create `directory-caches.json` ourselves. That file is generic-worker internal state, so image builds should not have to write it directly.

**Concrete use case**

Firefox CI Windows `run-task` jobs mount a writable Gecko checkout cache at `build/`. On cold workers, profile-server tasks clone Gecko before profileserver starts.

Example task:

- task: `V9HeyYLSRoG65_Mv3j5mYw`
- pool: `gecko-1/win11-a64-25h2-builder-alpha`
- mount: `cacheName=gecko-level-1-checkouts-sparse`, `directory=build`
- command: `run-task --gecko-checkout=build/src --gecko-sparse-profile=build/sparse-profiles/profile-generate`
- cold log: `[mounts] No existing writable directory cache 'gecko-level-1-checkouts-sparse'`
- clone bundle: `6963601216` bytes, about 6.96 GB

I sampled 50 recent similar Treeherder jobs for job `574801572`.

- 38 cold-cache jobs had VCS perf data
- cold clone median: 19.9 minutes
- cold VCS overall median: 21.9 minutes
- warm directory-cache VCS overall median: 2.2 minutes

The image build can pay the checkout cost once. The first task on each new worker should be able to reuse that checkout instead of cloning from scratch.

**Expected behavior**

Add a generic-worker config field for preloaded directory caches. For example:

```json
{
  "preloadedDirectoryCaches": [
    {
      "cacheName": "gecko-level-1-checkouts-sparse",
      "location": "C:\\worker-runner\\cache-seeds\\gecko-hg-sparse"
    }
  ]
}
```

At startup, after loading `directory-caches.json`, generic-worker would:

1. check whether each configured seed directory exists;
2. add it to the `directoryCaches` pool for the configured `cacheName`;
3. leave normal `WritableDirectoryCache.Mount()` logic to move it into the task directory.

If a seed path is missing, generic-worker should warn and continue. This is a cache optimization, not a hard requirement. Image tests can enforce that expected seed paths exist.

**Why this belongs in generic-worker**

Worker-manager already passes `workerConfig` through to worker-runner and generic-worker. It should not need to understand Gecko cache names or image-local paths.

Generic-worker owns writable directory caches: `directory-caches.json`, `CacheMap`, `AcquireCache`, and `WritableDirectoryCache.Mount`. The small supported hook is in `MountsFeature.Initialise()`, after:

```go
directoryCaches.LoadFromFile("directory-caches.json", config.DownloadsDir)
```

**Additional context**

Firefox taskgraph computes checkout cache names:

- `taskcluster/gecko_taskgraph/transforms/job/common.py`: `get_cache_name()` returns `checkouts-sparse` for hg sparse-profile jobs.
- `taskcluster/gecko_taskgraph/transforms/task.py`: generic-worker cache names are prefixed as `{trust_domain}-level-{level}-{name}`.

For `gecko-1` level-1 hg sparse checkout tasks, that becomes `gecko-level-1-checkouts-sparse`.

I did not find an existing issue for image-baked writable directory cache seeds. Related issues:

- #4990: writable caches modified by interactive tasks
- #7163: cache misses should not be fatal
- #7530: Windows cache directory creation failure
- #7650: Windows cache ownership reset issue
- #7427: cache state persistence with headlessTasks

jwmossmoz:
Fixes #8809.

Add `preloadedDirectoryCaches` to register image directories with the existing writable cache system. Existing caches take precedence; missing directories are skipped. Registration does not copy files or add a state file.

Tested registration, cache reuse, and purge on macOS. Windows, Linux, and FreeBSD multiuser builds pass.


Eijebong (changes requested):
Really wish you'd have run this through us **before** having claude output 900 lines of code...

I think this is much more complicated than it needs to be. I don't see why we could just have the `preloadedCaches` config, then something like this:

```golang
directoryCaches.LoadFromFile("directory-caches.json", config.CachesDir)
for _, p := range config.PreloadedCaches {
    if len(directoryCaches[p.Name]) > 0 {
        continue                      // existing cache wins
    }
    if fi, err := os.Stat(p.Dir); err != nil || !fi.IsDir() {
        continue
    }
    directoryCaches[p.Name] = []*Cache{{
        Location: p.Dir, Key: p.Name, Owner: directoryCaches,
        Created: time.Now(), InUse: false, LastUsed: time.Now(),
    }}
}
```

And you're done. First task moves the cache away so you don't have to worry about it being gone. It also avoids a copy (+ whatever you're doing because CopyFS is also the wrong tool for this) of hundreds of thousands of files on boot which is kind of the whole point of the feature.

If you want an error for a misconfigured cache, I'd emit it when tasksResolved == 0 && the cache doesn't exist && the seed dir is missing. Anything other than that is the cache having been evicted another way. And it avoids storing yet another state file that we have to care about.

If we need cross volume cache mounts to work (on POSIX we don't support cross volume renames right now), then we should implement it separately as a prerequisite for this

ui/docs/reference/workers/generic-worker/usage.mdx
```diff
@@ -288,6 +288,11 @@ and reports back results to the queue.
                                             [default: 524288000] (500MiB)
           numberOfTasksToRun                If zero, run tasks indefinitely. Otherwise, after
                                             this many tasks, exit. [default: 0]
+          preloadedDirectoryCaches          List of objects with cacheName and location. Each
+                                            location is an absolute path to a seed directory.
+                                            At startup, copy it into cachesDir if no cache
+                                            exists for that name.
+                                            [default: []]
```
petemoore, line 296: Is there a reason not to prepare the caches directory directly? I'm not sure what value having to copy it into place adds, as it would happen on first run and never again, so probably just loading it in place would also work and be cheaper?

tools/worker-runner/runner/run.go
```diff
@@ -39,9 +39,11 @@ func Run(configFile string) (state run.State, err error) {
 		}
 	}
 
-	state.Lock()
-	state.WorkerConfig = state.WorkerConfig.Merge(runnercfg.WorkerConfig)
-	state.Unlock()
+	if !runCached {
```
petemoore: what was the reason to add this check?

tools/worker-runner/runner/run_test.go
```diff
@@ -130,6 +130,9 @@ getSecrets: false
 cacheOverRestarts: %s
 workerConfig:
   fromFirstRun: true
+  preloadedDirectoryCaches:
```
petemoore: this looks like a generic-worker feature, but the implementation is `dummy`, and it seems to be adding logic about generic-worker to worker-runner (whereas worker-runner is meant to be agnostic to worker implementation as much as possible)

petemoore (commented):
Thanks for the PR. I think Generic Worker already effectively supports preloaded caches, since it stores a record of caches in a json file to disk, which persists across runs. If this was populated in the image build directly, I think you would already have the feature? I think the image build process just needs to populate it (directory-caches.json), and it would JustWork™ already, without code changes?

jwmossmoz:
Thanks for the feedback @Eijebong @petemoore! I'll continue iterating on this.

PR #9163, opened 2026-09-15
fix(generic-worker): stop interpolating osGroups into PowerShell on Windows

matt-boris:
Fixes bugs 2071937 and 2027573.

>Generic Worker on Windows no longer interpolates `payload.osGroups` into a PowerShell `-Command` string when adding or removing the task user from OS groups. Membership is updated with `NetLocalGroupAddMembers` / `NetLocalGroupDelMembers`, so group names cannot break out of a PowerShell single-quoted literal (including via Unicode quotation marks U+2018–U+201B).


lotas (dismissed):
THanks!

workers/generic-worker/win32/netapi_windows.go
```diff
@@ -0,0 +1,59 @@
...
+	r0, _, _ := proc.Call(
+		0, // local computer
+		uintptr(unsafe.Pointer(groupPtr)),
+		uintptr(localGroupMembersInfoLevel3),
+		uintptr(unsafe.Pointer(&info)),
+		1,
+	)
+	runtime.KeepAlive(groupPtr)
+	runtime.KeepAlive(memberPtr)
+	runtime.KeepAlive(info)
```
Eijebong: Those are useless. proc.Call has `//go:uintptrescapes` (and no, despite what your LLM is going to tell you the blank line between that and the function decl is not a problem or I'd have fixed it already)

workers/generic-worker/win32/netapi_windows.go
```diff
@@ -0,0 +1,59 @@
...
+	procNetLocalGroupDelMembers = netapi32.NewProc("NetLocalGroupDelMembers")
+)
+
+const localGroupMembersInfoLevel3 = 3
+
+type localGroupMembersInfo3 struct {
+	domainAndName *uint16
+}
+
+func netLocalGroupMembers(proc *LazyProcWrapper, fn, group, member string) error {
```
Eijebong: AFAICT, `fn` here is useless? `proc.LazyProc.Name` is the exact same thing?

Eijebong (changes requested):
https://github.com/taskcluster/taskcluster/blob/b0195cb382de5715f284ee9844a6241823f5a568/workers/generic-worker/runtime/runtime_windows.go#L39

You missed that one

matt-boris:
> https://github.com/taskcluster/taskcluster/blob/b0195cb382de5715f284ee9844a6241823f5a568/workers/generic-worker/runtime/runtime_windows.go#L39
> 
> You missed that one

@Eijebong I wanted to keep this PR more scoped to the issue, I'm happy to follow up in other spots with the new helpers, but everywhere else currently does not take in user input.

I've also force-pushed the changes from your review into the initial commit since they're all small enough. Thanks for the review.

Eijebong (approved):
Fine with me in a followup for the missing one 👍

PR #9168, opened 2026-09-17
Don't require database options for yarn commands that don't need them

Eijebong:
Right now when using `yarn db:new`, one must provide both `ADMIN_DB_URL` and `USERNAME_PREFIX` for no reason whatsoever (same for `db:renumber`) which hints that the method might actually be touching a live database when it does, in fact, not.

The database credentials were gotten in `yarn db:*` before checking whether we need them or not so I just moved it inside of the cases where they're actually needed.


petemoore (approved):
Thanks for the cleanup. Indeed, only upgrade and downgrade need the credentials, good catch.

PR #9169, opened 2026-09-17
auth: add support for multiple gcp projects (bug 1599247)

jcristau:
The config schema already allowed this, so just lift the restriction and keep a mapping of project to credentials and allowed service account names.

Bugzilla Bug: [1599247](https://bugzilla.mozilla.org/show_bug.cgi?id=1599247)


services/auth/src/main.js
```diff
@@ -192,40 +192,43 @@ const load = Loader(
...
-
-        if (projectIds.length === 0) {
-          return { googleapis, auth: {}, credentials: {}, allowedServiceAccounts: [] };
+        const allowedProjects = cfg.gcpCredentials.allowedProjects || {};
+
+        // Build a map of projectId -> per-project auth configuration.  Any
+        // number of projects may be configured (including none, in which case
+        // this service still starts up correctly but the gcpCredentials
+        // endpoint has nothing to grant).
+        const projects = {};
```
Eijebong: Having PTSD right now. If you want a fun fact about javascript:

```
> const projects = {}
> !!projects['__proto__']
true
> !!projects['anything']
false
```

That means that your line in gcp.js:28 will pass for `__proto__` as the name. Arguably very low since it requires holding a `auth:gcp:access-token:__proto__/...` scope.

`const projects = Object.create(null)` fixes it here, (`const project = Object.hasOwn(this.gcp.projects, projectId) ? this.gcp.projects[projectId] : undefined;` in gcp.js but I prefer the first one since it removes the footgun entirely).

Eijebong: Also the worst that can happen here is a 500 since it's just a read but it's how you end up with surprise injection and/or RCE because

```
> projects['__proto__']['foo'] = 'js is awesome'
'js is awesome'
> {}.foo
'js is awesome'
```

services/auth/test/helper.js
Eijebong: This is just `projectId` above

services/auth/test/helper.js
```diff
@@ -343,26 +343,44 @@ helper.withGcp = (mock, skipping) => {
...
+      };
     } else {
       // For testing, we expect a GCP service account defined with the "Service Account Token Creator"
       // role.  In CI, this is the "auth-granter" service account in the "taskcluster-tests" project.
       // It issues credentials for itself, so the allowedServiceAccounts must be (in order)
       // [<service account email>, invalid@mozilla.com].
-      const { credentials, allowedServiceAccounts } = await load('gcp');
+      const { projects } = await load('gcp');
+      const projectId = Object.keys(projects)[0];
+      const { credentials, allowedServiceAccounts } = projects[projectId];
```
Eijebong: `credentials` here can go if you replace `credentials.project_id` below. And that's the last consumer of this so you can then remove it in `main.js:220`

services/auth/src/main.js
```diff
@@ -192,40 +192,43 @@ const load = Loader(
...
-        assert(projectIds.length <= 1, 'at most one GCP project is supported');
-
-        if (projectIds.length === 0) {
-          return { googleapis, auth: {}, credentials: {}, allowedServiceAccounts: [] };
+        const allowedProjects = cfg.gcpCredentials.allowedProjects || {};
+
+        // Build a map of projectId -> per-project auth configuration.  Any
+        // number of projects may be configured (including none, in which case
+        // this service still starts up correctly but the gcpCredentials
+        // endpoint has nothing to grant).
```
Eijebong: This feels a bit much for something that goes `projects = {}\n for(entry of allowedProjects) { projects[id] = ... }`

services/auth/src/gcp.js
```diff
@@ -25,27 +25,24 @@ export const gcpBuilder = builder =>
...
         version: 'v1',
-        auth: this.gcp.auth,
+        auth: project.auth,
       });
 
       try {
         const response = await iamcredentials.projects.serviceAccounts.generateAccessToken({
           // NOTE: the `-` here represents the projectId, and uses the projectId
-          // from this.gcp.auth, which is why we verified those match above.
+          // from project.auth, whose project_id we verified matches above.
```
Eijebong: It's not above anymore, it's on startup?

Eijebong (changes requested):
Just a few minor things but looking pretty good

Eijebong:
/taskcluster run

I hate it here

Eijebong (changes requested):
Sorry, should've caught this the first time:

```
ui/docs/manual/deploying/cloud-credentials.mdx
102:[Future work](https://bugzilla.mozilla.org/show_bug.cgi?id=1599247) will allow multiple projects.
```

Rest looks good now, thank you

Eijebong (approved):
Thank you

PR #9170, opened 2026-09-17
Don't leave ACL entries behind when securing a file on windows

Eijebong:
`SecureFiles` ran `icacls <path> /grant:r Administrators:(GA) /inheritance:r`, which didn't do as much as one would hope... `/inheritance:r` removes the inherited entries, and `/grant:r` replaces the granted rights of the trustee it names. An entry sitting on the file itself for anyone else is neither of those, so it survives untouched. This means that a file carrying `Everyone:(F)` (full access to everyone) comes out of `SecureFiles` as:

```
BUILTIN\Administrators:(F)
Everyone:(F)
```

which is not what the name promises nor what happens on posix (chmod 0600). It happened to be enough in practice, because a freshly created file gets its entries by inheritance and those do get dropped, but nothing about the command enforces that.

Ownership was also never changed through `SecureFiles`, and an owner always keeps both `READ_CONTROL` and `WRITE_DAC`, so whoever owned the file could grant themselves access back at any point.

So we now build the descriptor ourselves instead of asking icacls to edit one: a DACL holding a single entry, full control to the Administrators group, protected so that nothing is inherited into it, and ownership handed to that same group. And it's applied through a handle from `safefs` because we needed the token rights extension that's granted by it anyway so it now refuses to resolve a path through a junction or a link, so this can't be redirected onto another file the way the icacls call could.

To get this to work without the pain of having to do recursive grants, I decided refuse directories. Securing one never protected what it contained anyway, everyone gets to skip the traverse check on the way down, so a file is reached by its own DACL alone regardless of the directory above it. The old icacls call was making windows propagate the inheritance cut to descendants, stripping whatever they inherited through that directory so it wasn't entirely useless but it was also way weaker than one would expect. In the end this is nothing anyway, the worker only ever secures files except in tests where it was securing the task's GOPATH to make the worker binary unreadable. I had to rewrite that part by adding a `makeDirTaskUserProof` which basically reuses whatever the old icacls was doing.

Since icacls apparently can't do anything correctly, I also had to rewrite the `/reset` call to actually reset the security descriptor to whatever it was before instead of what it seems to be doing, "just enable inheritance and call it day"


petemoore:
> In the end this is nothing anyway, the worker only ever secures files except in tests where it was securing the task's GOPATH to make the worker binary unreadable. I had to rewrite that part by adding a makeDirTaskUserProof which basically reuses whatever the old icacls was doing.

I suspect only the worker binary itself needs securing, so could this code not be updated to just secure the worker binary rather than the directory it is in, and then securing directories could be dropped entirely?

Eijebong:
> I suspect only the worker binary itself needs securing, so could this code not be updated to just secure the worker binary rather than the directory it is in, and then securing directories could be dropped entirely?

Doesn't work. It only worked because nobody could read the file. And when I did secure the file only, I had to reset permissions which ended up being a nightmare, I still have no idea why resetting it without inheritance was still somehow shoving an `(I)` flag on it. I couldn't reproduce locally and IMO not fucking with the binary directly is better anyway

workers/generic-worker/fileutil/fileutil_windows.go
```diff
@@ -1,35 +1,71 @@
...
+	}
+	if err := sd.SetControl(windows.SE_DACL_PROTECTED, windows.SE_DACL_PROTECTED); err != nil {
+		return fmt.Errorf("could not protect the DACL in the security descriptor: %w", err)
+	}
+
 	for _, path := range filepaths {
-		err = host.Run("icacls", path, "/grant:r", "Administrators:(GA)", "/inheritance:r")
-		if err != nil {
-			return
+		if err := secureFile(path, sd); err != nil {
```
petemoore, line 49: It may be useful to log permissions changes for auditing purposes (originally we did not log permissions changes in task logs but this was requested as an audit trail).

workers/generic-worker/fileutil/fileutil_windows_test.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/taskcluster/taskcluster/v110/workers/generic-worker/host"
+	"golang.org/x/sys/windows"
+)
+
+func TestSecureFilesDropsExplicitEntries(t *testing.T) {
```
petemoore, line 12: Is this a test that would fail with the old behaviour, but now passes with the explicit DACL replacement?

workers/generic-worker/safefs/safefs_windows.go
```diff
@@ -21,6 +21,9 @@ const (
 	traverseAccess uint32 = windows.FILE_READ_ATTRIBUTES | windows.FILE_LIST_DIRECTORY | windows.SYNCHRONIZE
 )
 
+// This is winnt.h's FILE_ALL_ACCESS, which is missing from x/sys/windows...
```
petemoore, line 24: lol

workers/generic-worker/README.md
```diff
@@ -460,10 +460,10 @@ There are a few environment variables that you can set to influence the tests:
 Set to a non-empty string if you wish to skip all tests that require python to
 be installed.
 
-### `GW_SKIP_MOZILLA_BUILD_TESTS`
+### `GW_SKIP_MSYS_TESTS`
```
petemoore, line 463: Changing this would affect anywhere that env var is set - are you sure all places that were using GW_SKIP_MOZILLA_BUILD_TEST are known, and are they all changed to use GW_SKIP_MSYS_TESTS now? This is a better name, so a nice improvement. In any case, if we haven't, it should show up, as it is opt-in skip rather than opt-out, which protects us somewhat.

petemoore (approved):
Looks nice!

workers/generic-worker/README.md
Eijebong:
```
[eijemoz@mozium] ~/code/taskcluster5 (win-secure-files) >>> rg -i MOZILLA_BUILD_TESTS
[eijemoz@mozium] ~/code/taskcluster5 (win-secure-files) >>>
```

:shrug: I can't think of anywhere outside the repo that would use it. It can be useful to run the test suite locally (although good luck on that anyway...)

workers/generic-worker/fileutil/fileutil_windows_test.go
Eijebong: Yeah exactly. The old code didn't remove explicit entries

workers/generic-worker/fileutil/fileutil_windows.go
Eijebong: I mean, this is on worker startup, nothing to do with tasks themselves?

Eijebong: I'll merge this as is to unblock CI for other PRs, if you feel like this warrants a log line I'm happy to see a follow up PR for that. As it is I personally don't see the point of adding more noise to the logs for something that's gonna be a no-op in 100% of cases.

PR #9171, opened 2026-09-17
feat(w-m): Fix unique violation detection in static worker create

lotas:
This fixes an old artifact of pre-postgres that checked for `EntityAlreadyExists` code instead of proper postgres error code

Provider will now throw a valid worker exists error in case of collision and api will report it as `409` conflict.

Updated endpoint documentation to explain this

Additionally removes irrelevant uniqueness check for error reporting, since there are no unique fields and handler assignes unique errorId each time, making this check completely useless

Bugzilla Bug: [2072160](https://bugzilla.mozilla.org/show_bug.cgi?id=2072160)


matt-boris (dismissed):
👍🏻

services/worker-manager/src/providers/provider.js
```diff
@@ -485,18 +485,7 @@ export class Provider {
         launchConfigId,
       });
 
-      try {
```
lotas, line 488: this was useless since errorId was always unique here

matt-boris (dismissed):

services/worker-manager/src/api.js
```diff
@@ -846,6 +846,10 @@ declareWithTrailingColon(
       'Create a new worker.  This is only useful for worker pools where the provider',
       'does not create workers automatically, such as those with a `static` provider',
       'type.  Providers that do not support creating workers will return a 400 error.',
+      'Worker IDs cannot be reused after removal.',
+      'If the ID belongs to a stopped worker or conflicts with the existing worker,',
+      'this method returns a 409 worker.',
```
Eijebong: Presumably a 409 error?

lotas: good catch! thanks

lotas:
updated api doc in new commit and rebased (first commit unchanged)

Eijebong (approved):

PR #9174, opened 2026-09-18
Don't return expired indices from `index.findTasksAtIndex`

Eijebong:
The issue is that when an index expires, it's not immediately removed from the database, instead a cron comes and cleans up every once in a while. So for a moment after expiration, the database contains invalid rows that are expired (and most likely also point to an expired task) which is wrong to return.

I dug up a bit because I was confused why the db function test was saying it should return expired rows. I think this all stems from the azure -> postgres migration (2076ecf4a491b91e573510d8498f52237062da49) where every single row getter (`get_indexed_task`, `get_index_namespace`) got that test but then has the index service filter expired tasks on the javascript side whereas getters that return lists filter it on the SQL side (`get_indexed_tasks`, `get_index_namespaces`). When `get_tasks_from_indexes_and_namespaces` got implemented in #6915, every test inserted got copied from the one just above (the ones for the singular return equivalent), including the fact that it was supposed to return expired rows. And then nothing on the JS side is filtering them.

I couldn't find any rationale that would indicate that this was intentional, I cannot find of a single reason why we would want that behavior. I even found a bug in a test that depended on the bug in the query (the expired row had a higher rank than a non expired one so the non expired insert never did anything and the test still passed because it was returning the expired row), with a comment clearly indicating that it was supposed to not return expired rows so let's just call this a bug and fix it, matching every other function that can return an indexed task already.


lotas (approved):
LGTM! Thanks

matt-boris (commented):
Thanks for fixing! 👍🏻

PR #9175, opened 2026-09-18
Indexing a task now ignores expired entries with a higher rank on insertion

Eijebong:
This fixes a very narrow issue where inserting an index entry with a lower rank than an existing one that is expired but not cleaned up yet would wrongly end up being a no-op.


lotas (approved):
lgtm, thanks!

there is likely an existing race condition when two concurrent inserts happen and last write would always win, but it was always here

PR #9176, opened 2026-09-18
Improve artifact name decoding in UI and Github service

lotas:
- redirect artifact names in UI decode artifact name properly
- GitHub service now encodes each artifact name as a single route parameter to preserve URL-special characters
- Plus a typo fix in class name

Related to #8997


petemoore (commented):
This looks good! It looks like a UI fix only, iiuc. Do you have an example of an artifact name, how it rendered before, and how it renders now? It might be worth adding that to the changelog or PR comments or git commit message for posterity. I see we call a decode, but it wasn't immediately obvious how this presented in the UI. Just one example might be nice, to make it clearer what types of symbols would be interpreted incorrectly due to which missing decoding step etc. But looks good! Thanks Yarik!

petemoore (dismissed):

Eijebong (changes requested):
Hold on, I think this is incomplete

lotas:
https://github.com/taskcluster/taskcluster/blob/56ac4c8e4aa801cc395c7b67261b82068b84667e/ui/src/views/Tasks/routes.js#L61

here is how route is being defined
it is used externally, maybe treeherder is using those, and also github service when building artifact links:
https://github.com/taskcluster/taskcluster/blob/56ac4c8e4aa801cc395c7b67261b82068b84667e/services/github/src/handlers/utils.js#L251-L257

which reminds me it was flagged already in some comment because it is not escaping it properly yet .. I can add the fix here as they kind of compliment each other

ui/src/views/Tasks/TaskArtifactRedirect/index.test.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { MemoryRouter, Route } from 'react-router-dom';
+import TaskArtifactRedirect from './index';
+
+vi.mock('../../../utils/Auth', () => ({ withAuth: Component => Component }));
+
+it('preserves reserved characters through the router', () => {
+  const taskId = 'eR1kMya2SruyMaRMZguROg';
+  const name = 'public/file?#.txt';
```
Eijebong: `const name = 'public/file?#%26.txt';` this doesn't work, it won't work on any valid `%XY`. And I don't think you can fix that by encoding harder for the same reason as #8997 (`%2526` revives `%26`)

I honestly can't find a solution that doesn't involve dropping the name out of the path and breaking everything... This is really annoying and i hate this garbage ecosystem.

Eijebong (changes requested):
Sorry, had to request changes before it got merged incomplete and I wasn't sure yet :p

ui/src/views/Tasks/TaskArtifactRedirect/index.test.jsx
lotas: yes, good catch! 
so as route param gets decoded by react-router before, we could use `window.location.pathname` to read "the original" and decode it ourselves.

Also updated github url to encode full name before to preserve any ?# bits

services/github/src/handlers/utils.js
Eijebong: Doesn't this need something like this too?

```suggestion
  libUrls.ui(rootUrl, `/tasks/${taskId}/runs/${runId}/logs/live/${encodeURIComponent(liveLogName)}`);
```

Eijebong (commented):
I think we missed one place entirely because it's on the other side of the file, otherwise lGTM

services/github/src/handlers/utils.js
Eijebong: I would love to know how you're seeing a blank line at line 13 github...

lotas:
grrrr @Eijebong i realized I also missed TaskLog page.. which needs same kind of attention as the artifact redirect one

plus github handler utils contained two different ui url building routines.. unified them a bit for consistency and updated tests since encoding artifact names is producing different strings

changelog/issue-8997.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 8997
+---
+UI Artifact redirects and GitHub check-run artifact links now preserve names containing URL-special characters.
```
lotas:
```suggestion
UI artifact redirects, log views, and GitHub check-run artifact links now preserve names containing URL-special characters.
```

services/github/src/handlers/status.js
```diff
@@ -17,12 +17,12 @@ import { requestArtifact } from './requestArtifact.js';
...
   taskUI,
   makeDebug,
-  taskLogUI,
+  taskLiveLogUI,
   GithubCheck,
   getTimeDifference,
   taskGroupUI,
-  buildUrl,
-  buildLogUrl,
+  taskAritfactUI,
```
petemoore: i beg your pardon

services/github/src/handlers/utils.js
```diff
@@ -9,8 +8,12 @@ export const taskUI = (rootUrl, taskGroupId, taskId) =>
   );
 export const taskGroupUI = (rootUrl, taskGroupId) =>
   libUrls.ui(rootUrl, `${rootUrl === 'https://taskcluster.net' ? '' : '/tasks'}/groups/${taskGroupId}`);
-export const taskLogUI = (rootUrl, runId, taskId, liveLogName = 'public/logs/live.log') =>
-  libUrls.ui(rootUrl, path.join(`/tasks/${taskId}/runs/${runId}/logs/live/`, liveLogName));
+export const taskLiveLogUI = (rootUrl, runId, taskId, liveLogName = 'public/logs/live.log') =>
```
petemoore: `rootUrl, taskId, runId` would be more consistent order with other helpers

ui/src/utils/artifactNames.js
```diff
@@ -9,6 +9,19 @@ const isSafeLogViewerArtifactName = name =>
       segment !== '..'
   );
 
+// react-router's history decodes the pathname before populating route params
+// which turns '%2526' into '%26' and looses information
```
petemoore: looses -> loses

changelog/issue-8997.md
```diff
@@ -0,0 +1,5 @@
+audience: users
+level: patch
+reference: issue 8997
```
petemoore: https://github.com/taskcluster/taskcluster/pull/8997 is a PR not an issue. Bug is 2072159.

ui/src/views/Tasks/TaskLog/index.test.jsx
```diff
@@ -57,3 +57,12 @@ it('preserves malformed percent encoding in a direct log link', () => {
     'public/logs/%foo.log'
   );
 });
+
+it('preserves an artifact name encoded as one route parameter', () => {
```
petemoore, line 63: this reuses the MemoryRouter helper, so does not test the fix (`window.location.pathname` stays `/`, the prefix lookup in `getRawArtifactName` fails, and the code takes the old fallback path).

petemoore (changes requested):
Maybe PR title should be "Preserve URL-special characters in artifact names".

Client-side navigation still loses `%25`. Any `<Link>` or `history.push` goes through `decodeURI` before `pushState`, so the address bar ends up with `%26` where `%2526` was intended, and `window.location.pathname` is no longer raw.

Other than that, looks pretty good! :-)

services/github/src/handlers/status.js
lotas: .. you fix one typo, you bring one in .. typos must flow..

changelog/issue-8997.md
lotas: true, thanks

services/github/src/handlers/utils.js
lotas: ouch. thanks, nice catch

ui/src/views/Tasks/TaskLog/index.test.jsx
lotas: indeed. switched to use browser router instead, thanks

petemoore (approved):
Nice job!

Eijebong (approved):
A history cleanup would be much appreciated :)

PR #9177, opened 2026-09-18
Fix circular import breaking the aio python client

### LINKED ISSUE #9172: python client tries to import missing `Object`
https://github.com/taskcluster/taskcluster/blob/05fad0540c000a79504004a6bba11db04cd75673/clients/client-py/taskcluster/aio/download.py#L31 tries to import `Object` from the same module, which doesn't exist. It looks like might it's part of the generated python code instead? https://github.com/taskcluster/taskcluster/blob/05fad0540c000a79504004a6bba11db04cd75673/clients/client-py/taskcluster/generated/aio/object.py#L16

Eijebong:
For all python is good at, it's not very good at figuring out how to deal with circular imports... In this case what happens is that import `taskcluster.helper` before importing anything in `taskcluster.aio` left `taskcluster.aio` in a broken state.
In its `__init__.py` it does `from _client_importer import *`, which checks if the module is already in `sys.modules` and when it is, copies its `__dict__` in `taskcluster.aio`. And when you import `_client_importer` from *anywhere* else, it inserts an empty module in `sys.modules`, then while evaluating the module, imports `taskcluster.aio` which snapshots an empty dict, and ends up not reimporting the clients. And you end up with:

```
>>> import taskcluster.helper, taskcluster.aio
>>> hasattr(taskcluster.aio, "Queue")
False
>>> import taskcluster.download
Traceback (most recent call last):
  File "<python-input-2>", line 1, in <module>
    import taskcluster.download
  File "/home/eijemoz/.local/lib/python3.14/site-packages/taskcluster/download.py", line 16, in <module>
    from .aio import download as aio_download
  File "/home/eijemoz/.local/lib/python3.14/site-packages/taskcluster/aio/download.py", line 28, in <module>
    from . import Object
ImportError: cannot import name 'Object' from 'taskcluster.aio' (/home/eijemoz/.local/lib/python3.14/site-packages/taskcluster/aio/__init__.py)
```

The fix at least is somewhat trivial, just import `taskcluster` and `taskcluster.aio` and let them do the import of the `_client_importer` private module, which breaks the loop. I say somewhat trivial because it's technically incomplete, you can still break imports by going out of your way and importing that private module yourself, but fixing that is much more invasive as it requires moving a bunch of code (`asyncclient`, `asyncutils`, `retry`) out of `taskcluster.aio` into a helper module and I'd rather not.

The sync version never had this issue because python always imports the root module first anyway, so importing `taskcluster.anything` always imports `taskcluster` completely before touching `anything`

Fixes #9172


matt-boris (approved):
LGTM 👍🏻

PR #9178, opened 2026-09-18
fix(generic-worker): stop Windows caches inheriting Users access

matt-boris:
On pools with cachesDir at a drive root, writable-directory-cache trees
inherited BUILTIN\Users:(OI)(CI)(RX). grantFullControl only merged a Full
Control ACE and never set SE_DACL_PROTECTED, so a later task without
generic-worker:cache:* could read leftover cache contents.

Replace the DACL and protect it when granting a task user the tree. Give
cachesDir and downloadsDir a worker-only protected DACL at startup so new
entries no longer inherit Users from the drive root.

Fixes bug 2071833.

>Generic Worker on Windows multiuser now replaces (instead of merging) DACLs when granting a task user control of task and writable-directory-cache trees, and protects those DACLs from inheritance so leftover `BUILTIN\Users` / Everyone ACEs cannot follow the cache. `cachesDir` and `downloadsDir` are given a protected worker-only DACL at startup so they no longer inherit those ACEs from a drive root.


workers/generic-worker/mounts_feature.go
```diff
@@ -1027,6 +1026,13 @@ func (w *WritableDirectoryCache) Unmount(taskMount *TaskMount) error {
 	return nil
 }
 
+func returnCacheToPool(oldpath, newpath string) error {
+	if err := RenameCrossDevice(oldpath, newpath); err != nil {
+		return err
+	}
+	return secureCachePoolEntry(newpath)
```
Eijebong: Do we have any idea how slow that is? I'm a bit concerned because `Unmount` is happening *before* a task gets resolved. Which means if this takes minutes you sit there for that long, on the task deadline's budget, with the task done already and waiting for this to happen when the followup tasks could already have been going.

workers/generic-worker/mounts_feature.go
```diff
@@ -1027,6 +1026,13 @@ func (w *WritableDirectoryCache) Unmount(taskMount *TaskMount) error {
...
+}
```
Eijebong: This leaves a window where the cache is renamed and unprotected.

workers/generic-worker/win32fs_multiuser_windows.go
```diff
@@ -59,24 +62,41 @@ func (g *granter) mayTake(links uint32, owner *windows.SID) bool {
...
+// and takes ownership as username. Descendants are updated if recurse is set.
 func grantFullControl(path, username string, recurse bool) error {
+	sid, _, _, err := windows.LookupSID("", username)
+	if err != nil {
+		return fmt.Errorf("could not look up SID for user %q: %w", username, err)
+	}
+	return grantTo(path, sid, recurse, true)
+}
+
+func grantTo(path string, sid *windows.SID, recurse, takeOwnership bool) error {
```
Eijebong: Not a huge fan of this. We're not granting anything ot anyone with that change. We're removing the DACL for non `SYSTEM`.

Eijebong (changes requested):
The implementation looks mostly correct minus the time where the cache stays unprotected and a few things I don't like about the code but I don't think the `secureCachePoolEntry` dance is necessary at all. Left some comments if it turns out I'm wrong because I looked at it anyway but hear me out. 

The problem happens because of two things. 1) The root of those workers is a disk that has `BUILTIN\Users:(OI)(CI)(RX)` for some reason, which `D:\caches` inherits as is. 2) When creating a task directory we grant it to the task user but we never replace the existing ACLs, we only merge it. And since on the worker we're testing on, it's on the same disk, it also inherits `BUILTIN\Users:(I)(OI)(CI)(RX)`. So any cache (or download) created inherits `BUILTIN/Users: (I)(RX)`.

The solution seems fairly easy. Keep the ACL replacement of this patch, it's correct. Protect `cachesDir`/`downloadsDir` on startup (we already call `grantFullControl` on the task directory when it gets created). That should do the trick and you don't have to pay for a recursive grant on unmount as well as mounting.

workers/generic-worker/win32fs_multiuser_windows.go
matt-boris: I know.. I wasn't a fan here either. Can rethink/rework it a bit so I'm not abusing the side effect of using this helper.

matt-boris:
https://github.com/taskcluster/taskcluster/pull/9189 landing will make this simpler.

